PR #5187, opened 2022-11-28
fix: prevent balances refreshed when not changing account address

### LINKED ISSUE #5175: Provide an option not to automatically rescan balances with every modification
## Abstract

In the current version, whenever I change an address name, or make a new tag, my whole portfolio is rescanned and while it's being rescanned, all options to change names or tags are greyed out until the scan is completed. This makes the app utterly unusable, as to change names or takes for 100+ addresses it can take many hours of waiting time. 

## Motivation

It'd be really great if the automatic rescanning is slightly delayed/postponed. For example, this is how some other crypto tax apps work: they have a grace period of a few minutes until your portfolio is recalculated, giving the user more time to add/update his data before scheduling a complete rescan and tax recalculation.

lukicenturi:
Closes #5175

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`5187` Fix issue where the blockchain balances were refreshed, when only changing the label or the tags of an account.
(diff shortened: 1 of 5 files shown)


kelsos (approved):
lgtm, thank you @lukicenturi

PR #5203, opened 2022-11-30
Fix interactio with sqlite transactions

nebolax:
This PR solves 2 bugs that cause the same problem.

1. In some database upgrades (especially important is globaldb v2_v3 upgrade which might take a long time) `cursor.executescript` was used which (as also stated here https://github.com/rotki/rotki/blob/develop/rotkehlchen/db/drivers/gevent.py#L112) always issues a commit. This causes inconsistent database state if user closes the app during an upgrade.

2. But that's only half of the problem. The other thing is that by default sqlite3 (not sqlite, sqlite3 python library) begins a transaction when it sees `INSERT/REPLACE/UPDATE/DELETE` statement but with statements like `CREATE TABLE` or `DROP TABLE`, etc. it works differently and no transaction begins automatically. Also there was no explicit `BEGIN TRANSACTION` in `write_ctx` context manager and this means that everywhere where we had `with conn.write_ctx() as cursor: cursor.execute('CREATE TABLE ...`), the table was created immediately, ignoring the context. And it was also leading to db inconsistency.


LefterisJP:
Okay so it seems that CREATE TABLE and DROP TABLE are immediately commited for the `DEFERRED` state which is our default. Unlike all other statements. 

So thats ... annoying and I would consider it a bug but it's how it works. Let me think how to properly address this.

LefterisJP (commented):
One thing you are missing here is to change the isolation level.

As I wrote above your assertion was not correct on what python sqlite module does. By default `solation_level` is set to `''` which means `DEFERRED`.

That means that only the DML statements start a transaction automagically.

We can control that.

We can set the isolation level to `None`, effectively stopping this behaviour. This happens here:
https://github.com/rotki/rotki/blob/e8e671a36e1aa7791b3e8996c20b6e019f3168ca/rotkehlchen/db/drivers/gevent.py#L245-L248

Add `isolation_level=None` to the `connect()` method.

This way you won't have competing transactions opening. As you can see many of your tests failed with `cannot start a transaction within a transaction`. I suspect that this may address them.


I tested that this "should" work for at least the CREATE/DROP table scenario by playing with a super simple script

```python
from gevent import monkey  # isort:skip
monkey.patch_all()  # isort:skip
from rotkehlchen.db.drivers.gevent import DBConnection, DBConnectionType, DBCursor
from rotkehlchen.logging import TRACE, add_logging_level, configure_logging
from rotkehlchen.tests.utils.args import default_args

add_logging_level('TRACE', TRACE)
configure_logging(default_args())
connection = DBConnection('test.db', DBConnectionType.USER, 0)




with connection.write_ctx() as cursor:
    cursor.execute('PRAGMA KEY=123')
    # cursor.execute('DROP TABLE foo')
    # cursor.execute('DROP TABLE boo')
    # cursor.execute('CREATE TABLE foo(a INTEGER NOT NULL PRIMARY KEY)')
    connection._conn.execute('CREATE TABLE foo(a INTEGER NOT NULL PRIMARY KEY)')
    cursor.execute('INSERT INTO foo(a) VALUES(?)', (1,))
    cursor.execute('CREATE TABLE boo(a INTEGER NOT NULL PRIMARY KEY, b STRING NOT NULL)')
    # cursor.execute('INSERT INTO foo(a) VALUES(?)', (1,))
    # cursor.execute('INSERT INTO boo(a, b) VALUES(?, ?)', (1, 'fff'))
    raise ValueError('foo')
```

Playing with the above and the isolation level, recreating the DB each time and then checking if something was written on it or not helps figure this out.

rotkehlchen/globaldb/upgrades/v2_v3.py
```diff
@@ -546,7 +544,8 @@ def migrate_to_v3(connection: 'DBConnection') -> None:
         # database and are not EVM (matic tokens, Otimism tokens, etc) + their variants in
         # other chains. And populates them properly via sql statements
         with open(dir_path / 'data' / 'globaldb_v2_v3_assets.sql', 'r') as f:
-            sql_sentences = f.read()
-            cursor.executescript(sql_sentences)
-        connection.commit()
+            raw_sql_sentences = f.read()
+            per_table_sentences = raw_sql_sentences.split('\n\n')
+            for sql_sentences in per_table_sentences:
```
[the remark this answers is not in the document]
yabirgb: The executescript here was used since that sql file had 3 insert for the three tables that contain information about the evm assets. What would be nice is to revert the changes if something goes wrong for some reason since otherwise the state becomes inconsistent. (say that you set information in 2 of the 3 tables for some reason). In that case we should rollback i guess

rotkehlchen/globaldb/upgrades/v2_v3.py
```diff
@@ -519,7 +517,7 @@ def migrate_to_v3(connection: 'DBConnection') -> None:
         );
         """)
 
-        cursor.executescript('PRAGMA foreign_keys=off;')
+        cursor.execute('PRAGMA foreign_keys=off;')
```
yabirgb, line 520: I'm not sure about this one here. Doesn't it need a commit? I believe not and then it was my mistake

rotkehlchen/data/globaldb_v2_v3_assets.sql
```diff
@@ -719,6 +719,7 @@ INSERT INTO assets(identifier, type, name) VALUES
 ("eip155:56/erc20:0x1e155e26085Be757780B45a5420D9f16a938f76b", "C", "KaraStar KARA"),
 ("eip155:56/erc20:0x86a53fcd199212FEa44FA7e16EB1f28812be911D", "C", "Inflation Hedging Coin"),
 ("eip155:56/erc20:0x4ffA6BB42d1a1A6d8e79935CcF1457d55deCff3f", "C", "REV3AL");
+
```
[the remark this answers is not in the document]
nebolax: For simplicity of the code in `v2_v3` globaldb upgrade. After adding these lines it is possible to split statements by `\n\n` and execute them via `.execute` instead of `.executescript`

rotkehlchen/db/upgrades/v28_v29.py
```diff
@@ -39,7 +39,7 @@ def _create_new_tables(conn: 'DBConnection') -> None:
     FOREIGN KEY(tx_hash, log_index) REFERENCES ethtx_receipt_logs(tx_hash, log_index) ON DELETE CASCADE ON UPDATE CASCADE,
     PRIMARY KEY(tx_hash, log_index, topic_index)
     );""")  # noqa: E501
-    conn.executescript("""CREATE TABLE IF NOT EXISTS nfts (
+    conn.execute("""CREATE TABLE IF NOT EXISTS nfts (
```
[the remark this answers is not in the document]
nebolax: Yeah, probably makes sense

LefterisJP (commented):
### Your approach

So as discussed in the discord you should undo and redo the entire PR. Almost all of these 1700 lines are not needed and are a mistake.

We should absolutely not open a single cursor and use it as much as possible. This approach goes against the idea of what a cursor is. A cursor is not a connection. I think the quickest way here is to kill all your new changes and start from scratch. You really should have asked about this as it made you waste time.

### The problem

The reason why you did this is not a silly one. I understand why you tried this approach. It's just not the right approach.

The problem is that once a python process with sqlite and `isolation level=None` opens a transaction it can not do that again until the transaction has closed.

This is something serious as the way we do stuff won't scale. So your idea was to somehow use some kind of god-cursor object :sweat_smile: 

That is the wrong approach though because:

- It's error prone
- Makes the code really hard to read and have to pass that god object everywhere
- Will still fail in edge cases
- Will still fail in a multi-greenlet environment as we can't predict if another greenlet wll open another write cursor. At some point someone will.

### The right approach.

I think ... it's surprisingly easy and elegant. It's close to a question you asked about savepoints and also to your usage of savepoints when you wrote in the PR that it was "too deep in the code" so you used a savepoint.

What I suggest is the following:

```diff
diff --git a/rotkehlchen/db/drivers/gevent.py b/rotkehlchen/db/drivers/gevent.py
index a4314cb75..64e194e28 100644
--- a/rotkehlchen/db/drivers/gevent.py
+++ b/rotkehlchen/db/drivers/gevent.py
@@ -237,15 +237,16 @@ class DBConnection:
         CONNECTION_MAP[connection_type] = self
         self._conn: UnderlyingConnection
         self.in_callback = gevent.lock.Semaphore()
+        self.in_transaction = False
         self.connection_type = connection_type
         self.sql_vm_instructions_cb = sql_vm_instructions_cb
         # We need an ordered set. Python doesn't have such thing as a standalone object, but has
         # `dict` which preserves the order of its keys. So we use dict with None values.
         self.savepoints: Dict[str, None] = {}
         if connection_type == DBConnectionType.GLOBAL:
-            self._conn = sqlite3.connect(path, check_same_thread=False)
+            self._conn = sqlite3.connect(path, isolation_level=None, check_same_thread=False)
         else:
-            self._conn = sqlcipher.connect(path, check_same_thread=False)  # pylint: disable=no-member  # noqa: E501
+            self._conn = sqlcipher.connect(path, isolation_level=None, check_same_thread=False)  # pylint: disable=no-member  # noqa: E501
         self._set_progress_handler()
         self.minimized_schema = None
         if connection_type == DBConnectionType.USER:
@@ -318,15 +319,22 @@ class DBConnection:
     @contextmanager
     def write_ctx(self) -> Generator['DBCursor', None, None]:
         cursor = self.cursor()
-        try:
-            yield cursor
-        except Exception:
-            self._conn.rollback()
-            raise
+        if self.in_transaction is True:
+            with self.savepoint_ctx() as cursor:
+                yield cursor
+                return
         else:
-            self._conn.commit()
-        finally:
-            cursor.close()
+            cursor.execute('BEGIN TRANSACTION')
+            self.in_transaction = True
+            try:
+                yield cursor
+            except Exception:
+                self._conn.rollback()
+                raise
+            else:
+                self._conn.commit()
+            finally:
+                cursor.close()
 
     @contextmanager
     def savepoint_ctx(
```

What that does is:
1. Remember when the connection is in a transaction by a boolean. (if there is a better way I am not sure but a quick check did not show one for python + sqlite).
2. If we are already in a transaction and another write ctx request comes in, don't do it. Open a savepoint context instead.

It's simple and elegant. And will have the added effect that basically we use simple transactions when needing it, and the moment the code nests (even if we did not plan to nest) the rollback functionality is nestable too and will only roll the changes back to where they are required.

Kind of "free savepoints only when needed".

I tested it with this code:

```python
from gevent import monkey  # isort:skip
monkey.patch_all()  # isort:skip
from rotkehlchen.db.drivers.gevent import DBConnection, DBConnectionType, DBCursor
from rotkehlchen.logging import TRACE, add_logging_level, configure_logging
from rotkehlchen.tests.utils.args import default_args

add_logging_level('TRACE', TRACE)
configure_logging(default_args())
connection = DBConnection('test.db', DBConnectionType.USER, 0)




with connection.write_ctx() as cursor:
    cursor.execute('PRAGMA KEY=123')
    cursor.execute('CREATE TABLE foo(a INTEGER NOT NULL PRIMARY KEY)')
    cursor.execute('INSERT INTO foo(a) VALUES(?)', (1,))

    try:
        with connection.write_ctx() as cursor2:
            cursor2.execute('INSERT INTO foo(a) VALUES(?)', (2,))
            cursor.execute('CREATE TABLE boo(a INTEGER NOT NULL PRIMARY KEY, b STRING NOT NULL)')
            cursor2.execute('INSERT INTO boo(a, b) VALUES(?, ?)', (1, 'fff'))
            raise ValueError('foo')
    except ValueError:
        print('Oh well, inner write_ctx failed, but lets let outer run to end and commit')

with connection.write_ctx() as cursor3:
    cursor3.execute('INSERT INTO foo(a) VALUES(?)', (3,))
```

The result here is that the cursor1 stuff go in nicely.
Then a nested write_ctx pops up, it rolls back due to an error and both the table creation and the additions don't go in.
Then finally a new root write_ctx starts and writes.

The result is:

```sql
SQLite version 3.39.2 2022-07-21 15:24:47 (SQLCipher 4.5.2 community)
Enter ".help" for usage hints.
sqlite> PRAGMA KEY="123";
ok
sqlite> .schema
CREATE TABLE foo(a INTEGER NOT NULL PRIMARY KEY);
sqlite> SELECT * from foo;
1
3
sqlite> 
```

LefterisJP:
Pushed what we discussed. I will check the tests and re-review the PR.

Forcing single transaction and switching from long-living write_cursor objects we should probably also do, but will open another issue for this. It's gonna take some time and don't want to combine in this PR

rotkehlchen/db/dbhandler.py
```diff
@@ -344,11 +390,22 @@ def get_setting(self, cursor: 'DBCursor', name: Literal['premium_should_sync'])
     def get_setting(self, cursor: 'DBCursor', name: Literal['main_currency']) -> AssetWithOracles:
         ...
 
+    @overload
+    def get_setting(self, cursor: 'DBCursor', name: Literal['ongoing_upgrade_from_version']) -> Optional[int]:  # noqa: E501
```
LefterisJP, line 394: you don't need a new overload here. version already does int

rotkehlchen/db/dbhandler.py
```diff
@@ -231,26 +236,67 @@ def __init__(
...
+                'Your encrypted database is in a half-upgraded state and there was no backup '
+                'found. Please open an issue on our github or contact us in our discord server.',
+            )
+
+        backup_to_use = sorted(found_backups)[-1]  # Use latest backup
+        shutil.copyfile(
+            os.path.join(self.user_data_dir, backup_to_use),
+            os.path.join(self.user_data_dir, 'rotkehlchen.db'),
+        )
+        self.msg_aggregator.add_warning(
```
LefterisJP, line 294: This functionality could also be in another PR. We should talk about this and probably have the user agree to this from the UI. This is a bit more complicated than this.

rotkehlchen/db/upgrade_manager.py
```diff
@@ -242,4 +250,5 @@ def _perform_single_upgrade(
 
         # Upgrade success all is good
         with self.db.user_write() as cursor:
+            self.db.delete_setting(write_cursor=cursor, name='ongoing_upgrade_from_version')
```
LefterisJP, line 253: no need for a function to just do this. directly call the DELETE here.

rotkehlchen/db/upgrades/v28_v29.py
```diff
@@ -99,13 +94,13 @@ def upgrade_v28_to_v29(db: 'DBHandler', progress_handler: 'DBUpgradeProgressHand
...
-    cursor.executemany(
-        'INSERT INTO xpub_mappings(address, xpub, derivation_path, account_index, derived_index) '
+    write_cursor.executemany(
+        'INSERT INTO xpub_mappings(address, xpub, derivation_path, account_index, derived_index) '  # noqa: E501
         'VALUES(?, ?, ?, ?, ?);',
         xpub_mappings_data,
     )
     progress_handler.new_step()
-    cursor.execute("""
+    write_cursor.execute("""
```
LefterisJP, line 103: you are making way way too many changes by renaming cursor here. Makes it very difficult to review and make sure you did not fuck something up. Please don't do that. In your PRs touch as little things as possible. It's really hard to review.

rotkehlchen/globaldb/handler.py
```diff
@@ -57,14 +59,72 @@
...
+    with connection.read_ctx() as cursor:
+        ongoing_upgrade_from_version = _get_setting_value(
+            cursor=cursor,
+            name='ongoing_upgrade_from_version',
+            default_value=-1,
+        )
+    if ongoing_upgrade_from_version == -1:
+        return connection  # We are all good
+
+    # Otherwise replace the db with a backup and relogin
```
LefterisJP, line 86: same comment as in the other function for the user db

rotkehlchen/globaldb/handler.py
```diff
@@ -57,14 +59,72 @@
 """
 
 
-def initialize_globaldb(dbpath: Path, sql_vm_instructions_cb: int) -> DBConnection:
-    """Initialie globaldb. May raise DBSchemaError if GlobalDB's schema is malformed."""
+def _initialize_and_check_abandoned_upgrades(
```
LefterisJP, line 62: The abandoned upgrades name is very wrong. You mean unfinished? In progress?

rotkehlchen/globaldb/handler.py
```diff
@@ -106,13 +174,15 @@ class GlobalDBHandler():
...
     _data_directory: Optional[Path] = None
+    _msg_aggregator: Optional['MessagesAggregator'] = None
     _packaged_db_conn: Optional[DBConnection] = None
     conn: DBConnection
 
     def __new__(
             cls,
             data_dir: Optional[Path] = None,
             sql_vm_instructions_cb: Optional[int] = None,
+            msg_aggregator: Optional[MessagesAggregator] = None,
```
LefterisJP, line 185: makes no sense. Leave that out. Message aggregator should not be an argument to the GlobalDB. Remove it.

rotkehlchen/globaldb/handler.py
```diff
@@ -139,10 +210,11 @@ def packaged_db_conn() -> DBConnection:
             # mypy does not recognize the initialization as that of a singleton
             return GlobalDBHandler()._packaged_db_conn  # type: ignore
 
-        packaged_db_path = Path(__file__).resolve().parent.parent / 'data' / 'global.db'
+        global_dir = Path(__file__).resolve().parent.parent / 'data'
         packaged_db_conn = initialize_globaldb(
-            dbpath=packaged_db_path,
+            global_dir=global_dir,
             sql_vm_instructions_cb=DEFAULT_SQL_VM_INSTRUCTIONS_CB,
+            msg_aggregator=GlobalDBHandler()._msg_aggregator,  # type: ignore  # Definitely is not None here  # noqa: E501
```
LefterisJP, line 217: remove

rotkehlchen/globaldb/handler.py
```diff
@@ -124,12 +194,13 @@ def __new__(
         """
         if GlobalDBHandler.__instance is not None:
             return GlobalDBHandler.__instance
-
         assert data_dir is not None, 'First instantiation of GlobalDBHandler should have a data_dir'  # noqa: E501
         assert sql_vm_instructions_cb is not None, 'First instantiation of GlobalDBHandler should have a sql_vm_instructions_cb'  # noqa: E501
+        assert msg_aggregator is not None, 'First instantiation of GlobalDBHandler should have a msg_aggregator'  # noqa: E501
```
LefterisJP, line 199: that makes no sense :>

rotkehlchen/rotkehlchen.py
```diff
@@ -132,6 +132,7 @@ def __init__(self, args: argparse.Namespace) -> None:
         GlobalDBHandler(
             data_dir=self.data_dir,
             sql_vm_instructions_cb=self.args.sqlite_instructions,
+            msg_aggregator=self.msg_aggregator,
```
LefterisJP, line 135: no. Remove

rotkehlchen/globaldb/utils.py
```diff
@@ -17,3 +17,14 @@ def _get_setting_value(cursor: 'DBCursor', name: str, default_value: int) -> int
         return default_value
 
     return int(result[0][0])
+
+
+def _add_setting_value(write_cursor: 'DBCursor', name: str, value: int) -> None:
```
LefterisJP, line 22: what's the point of making a function like this? Where you use it only once?

rotkehlchen/globaldb/utils.py
```diff
@@ -17,3 +17,14 @@ def _get_setting_value(cursor: 'DBCursor', name: str, default_value: int) -> int
...
+    write_cursor.execute(
+        'INSERT OR REPLACE INTO settings(name, value) VALUES(?, ?)',
+        (name, value),
+    )
+
+
+def _delete_setting_value(write_cursor: 'DBCursor', name: str) -> None:
```
LefterisJP, line 29: same here ...

rotkehlchen/globaldb/upgrades/manager.py
```diff
@@ -8,7 +8,13 @@
 from rotkehlchen.utils.misc import ts_now
 from rotkehlchen.utils.upgrades import UpgradeRecord
 
-from ..utils import GLOBAL_DB_VERSION, MIN_SUPPORTED_GLOBAL_DB_VERSION, _get_setting_value
+from ..utils import (
+    GLOBAL_DB_VERSION,
+    MIN_SUPPORTED_GLOBAL_DB_VERSION,
+    _add_setting_value,
+    _delete_setting_value,
```
LefterisJP, line 15: importing `_` functions? Wtf? Do you even know what `_` means? It's supposed to be private to the module. This is silly.

rotkehlchen/globaldb/updates.py
```diff
@@ -402,8 +402,12 @@ def perform_update(
         assert data_directory is not None, 'data directory should be initialized at this point'
         global_db_path = data_directory / 'global_data' / 'global.db'
         with TemporaryDirectory() as tmpdirname:
-            tempdbpath = Path(tmpdirname) / 'temp.db'
-            connection = initialize_globaldb(tempdbpath, GlobalDBHandler().conn.sql_vm_instructions_cb)  # noqa: E501
+            tmpdir = Path(tmpdirname)
+            connection = initialize_globaldb(
+                global_dir=tmpdir,
+                sql_vm_instructions_cb=GlobalDBHandler().conn.sql_vm_instructions_cb,
+                msg_aggregator=GlobalDBHandler()._msg_aggregator,  # type: ignore  # msg_aggregator is definitely not None here  # noqa: E501
```
LefterisJP, line 409: remove this ....

LefterisJP (commented):
Shit man .... I told you to rewrite and keep simple. This PR is still so bad ... mixing so many different things and touching and doing thing it should not be doing.

rotkehlchen/db/dbhandler.py
LefterisJP: Ah okay no no. You do. This is Optional int.

rotkehlchen/tests/unit/test_globaldb.py
```diff
@@ -505,7 +505,7 @@ def test_globaldb_pragma_foreign_keys(globaldb):
     fails we know that something has changed and we will need to adjust our strategy.
     """
     cursor = globaldb.conn.cursor()
-    cursor.execute('PRAGMA foreign_keys = OFF;')
```
LefterisJP, line 508: also this test fails in multiple places. You did not bother to actually run the tests of the code you were touching?

rotkehlchen/globaldb/updates.py
```diff
@@ -402,8 +402,11 @@ def perform_update(
         assert data_directory is not None, 'data directory should be initialized at this point'
         global_db_path = data_directory / 'global_data' / 'global.db'
         with TemporaryDirectory() as tmpdirname:
-            tempdbpath = Path(tmpdirname) / 'temp.db'
-            connection = initialize_globaldb(tempdbpath, GlobalDBHandler().conn.sql_vm_instructions_cb)  # noqa: E501
+            tmpdir = Path(tmpdirname)
+            connection, _ = initialize_globaldb(
+                global_dir=tmpdir,
```
LefterisJP, line 407: why did you change the logic to pass the root dir and not the path?

**Here you introduced a bug that was quite hard to find**.

Thank god for our tests.

You were supposed to initialize a tmpdir/temp.db database with this statement but you do tmpdir/global.db

LefterisJP:
Hmm something in this PR really makes tests super slow.

LefterisJP:
ah I think I see. It's a single test that simply never ends. Weird. Thought pytest would have a limit for a single test.

rotkehlchen/db/drivers/gevent.py
```diff
@@ -304,17 +328,33 @@ def read_ctx(self) -> Generator['DBCursor', None, None]:
...
-            raise
-        else:
-            self._conn.commit()
-        finally:
-            cursor.close()
+    def write_ctx(self, commit_ts: bool = False) -> Generator['DBCursor', None, None]:
+        with self.critical_section():
+            cursor = self.cursor()
+            if self._conn.in_transaction is True:
+                with self.savepoint_ctx() as cursor:
```
LefterisJP, line 335: For this: We have a follow-up: https://github.com/rotki/rotki/issues/5222

rotkehlchen/globaldb/updates.py
LefterisJP: fixed

rotkehlchen/tests/unit/test_globaldb.py
LefterisJP: fixed

PR #5204, opened 2022-11-30
Small shit

LefterisJP:
- Remove possibility of circular import from GlobalDBHandler
- Add location in binance pair error message


yabirgb (commented):
There is one test that needs to be checked `test_binance_query_pairs`. The circurlar import did you just prevent it or something triggered it?

LefterisJP:
> There is one test that needs to be checked `test_binance_query_pairs`. The circurlar import did you just prevent it or something triggered it?

The trigger was trying to test binance query pairs in a script and just importing globaldb handler alone triggerred it.

LefterisJP:
All green now. Merging.

PR #5205, opened 2022-12-01
Allow to filter by multiple addresses in decoding endpoint

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/api/v1/resources.py
+++ b/rotkehlchen/api/v1/resources.py
@@ -95,6 +95,7 @@ from rotkehlchen.api.v1.schemas import (
     ManualPriceRegisteredSchema,
     ManualPriceSchema,
     ModifyEvmTokenSchema,
+    MultipleAddressAsyncSchema,
     NameDeleteSchema,
     NamedEthereumModuleDataSchema,
     NamedOracleCacheCreateSchema,
@@ -503,12 +504,19 @@ class EthereumTransactionsResource(BaseMethodView):
 
 
 class EthereumTransactionsDecodingResource(BaseMethodView):
-    post_schema = AsyncQueryArgumentSchema()
+    post_schema = MultipleAddressAsyncSchema()
 
     @require_loggedin_user()
     @use_kwargs(post_schema, location='json_and_query')
-    def put(self, async_query: bool) -> Response:
-        return self.rest_api.decode_pending_ethereum_transactions(async_query=async_query)
+    def post(
+            self,
+            async_query: bool,
+            evm_addresses: Optional[list[ChecksumEvmAddress]],
+    ) -> Response:
+        return self.rest_api.decode_pending_ethereum_transactions(
+            async_query=async_query,
+            evm_addresses=evm_addresses,
+        )
 
 
 class EthereumAirdropsResource(BaseMethodView):
--- a/rotkehlchen/api/v1/schemas.py
+++ b/rotkehlchen/api/v1/schemas.py
@@ -2394,6 +2394,10 @@ class AssetsImportingFromFormSchema(Schema):
     file = FileField(allowed_extensions=['.zip', '.json'], required=True)
 
 
+class MultipleAddressAsyncSchema(AsyncQueryArgumentSchema):
+    evm_addresses = fields.List(EthereumAddressField(), load_default=None)
+
+
 class ReverseEnsSchema(AsyncIgnoreCacheQueryArgumentSchema):
     ethereum_addresses = fields.List(EthereumAddressField(), required=True)
 
--- a/rotkehlchen/chain/evm/decoding/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/decoder.py
@@ -272,14 +272,21 @@ class EVMTransactionDecoder():
 
         return sorted(events, key=lambda x: x.sequence_index, reverse=False)
 
-    def get_and_decode_undecoded_transactions(self, limit: Optional[int] = None) -> None:
+    def get_and_decode_undecoded_transactions(
+            self,
+            limit: Optional[int] = None,
+            addresses: Optional[list[ChecksumEvmAddress]] = None,
+    ) -> None:
         """Checks the DB for up to `limit` undecoded transactions and decodes them.
+        If a list of addresses is provided then only the transactions involving those
+        addresses are decoded.
 
         This is protected by concurrent access from a lock"""
         with self.undecoded_tx_query_lock:
             hashes = self.dbevmtx.get_transaction_hashes_not_decoded(
                 chain_id=self.evm_inquirer.chain_id,
                 limit=limit,
+                addresses=addresses,
             )
             self.decode_transaction_hashes(ignore_cache=False, tx_hashes=hashes)
 
--- a/rotkehlchen/db/evmtx.py
+++ b/rotkehlchen/db/evmtx.py
@@ -5,7 +5,7 @@ from rotkehlchen.chain.ethereum.constants import ETHEREUM_BEGIN, GENESIS_HASH
 from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
 from rotkehlchen.chain.evm.structures import EvmTxReceipt, EvmTxReceiptLog
 from rotkehlchen.db.constants import HISTORY_MAPPING_STATE_DECODED
-from rotkehlchen.db.filtering import EvmTransactionsFilterQuery
+from rotkehlchen.db.filtering import EvmTransactionsFilterQuery, TransactionsNotDecodedFilterQuery
 from rotkehlchen.db.history_events import DBHistoryEvents
 from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.logging import RotkehlchenLogsAdapter
@@ -36,7 +36,8 @@ from rotkehlchen.constants.limits import FREE_ETH_TX_LIMIT
 
 TRANSACTIONS_MISSING_DECODING_QUERY = (
     'evmtx_receipts AS A LEFT OUTER JOIN evm_tx_mappings AS B ON A.tx_hash=B.tx_hash '
-    'AND A.chain_id=B.chain_ID WHERE B.tx_hash is NULL'
+    'AND A.chain_id=B.chain_ID LEFT JOIN evm_transactions AS C on '
+    'A.tx_hash=C.tx_hash '
 )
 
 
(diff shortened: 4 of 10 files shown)


rotkehlchen/api/v1/schemas.py
```diff
@@ -2775,3 +2775,7 @@ def make_nft_filter_query(  # pylint: disable=no-self-use
             'ignore_cache': data['ignore_cache'],
             'filter_query': filter_query,
         }
+
+
+class MultipleAddressAsyncSchema(AsyncQueryArgumentSchema):
+    addresses = DelimitedOrNormalList(EthereumAddressField(), load_default=None)
```
nebolax, line 2781: Also I guess this schema should become a parent for this one https://github.com/rotki/rotki/blob/eb51706e9e59f0cf5627e499b7c1e55032eb6059/rotkehlchen/api/v1/schemas.py#L2408

rotkehlchen/db/evmtx.py
```diff
@@ -271,14 +272,18 @@ def get_transaction_hashes_not_decoded(
...
         """Get transaction hashes for the transactions that have not been decoded.
         Optionally by chain id.
         If the limit argument is provided then it is used in the SQL query with
         the default order.
         """
         querystr = 'SELECT A.tx_hash from ' + TRANSACTIONS_MISSING_DECODING_QUERY
-        bindings = []
+        bindings: List[Union[ChecksumEvmAddress, int]] = []
+        if addresses is not None:
+            bindings += addresses + addresses
```
nebolax, line 285: Why here to use `addresses + addresses` while in the other place (below) `addresses * 2` :thinking:

rotkehlchen/db/evmtx.py
```diff
@@ -271,14 +272,18 @@ def get_transaction_hashes_not_decoded(
...
+            querystr += f' AND (C.from_address IN ({",".join("?"*len(addresses))}) OR C.to_address IN ({",".join("?"*len(addresses))}))'  # noqa: E501
         if chain_id is not None:
```
nebolax, line 287: Since here filtering happens by many parameters (not just `addresses`) then maybe will make sense to use a proper filter class here?

rotkehlchen/db/evmtx.py
```diff
@@ -35,7 +35,8 @@
 
 TRANSACTIONS_MISSING_DECODING_QUERY = (
     'evmtx_receipts AS A LEFT OUTER JOIN evm_tx_mappings AS B ON A.tx_hash=B.tx_hash '
-    'AND A.chain_id=B.chain_ID WHERE B.tx_hash is NULL'
+    'AND A.chain_id=B.chain_ID LEFT JOIN evm_transactions AS C on '
+    'A.tx_hash=C.tx_hash WHERE B.tx_hash is NULL'
```
nebolax, line 39: Hmmmm I am a bit confused here. What sense does `WHERE B.tx_hash is NULL` make if it `evm_tx_mappings.tx_hash` can never be null? https://github.com/rotki/rotki/blob/1c29fb705f38c097d9da4f88a4ef98725cb80e55/rotkehlchen/db/schema.py#L530

Maybe the check should be then that `evmtx_receipts.tx_hash` is not in `evm_tx_mappings.tx_hash` ?

rotkehlchen/db/evmtx.py
```diff
@@ -271,14 +272,18 @@ def get_transaction_hashes_not_decoded(
...
             chain_id: Optional[ChainID],
             limit: Optional[int],
+            addresses: Optional[List[ChecksumEvmAddress]],
     ) -> List[EVMTxHash]:
         """Get transaction hashes for the transactions that have not been decoded.
         Optionally by chain id.
         If the limit argument is provided then it is used in the SQL query with
         the default order.
         """
         querystr = 'SELECT A.tx_hash from ' + TRANSACTIONS_MISSING_DECODING_QUERY
```
nebolax, line 282: Question. Should this function return transactions that do have receipts queried but are not decoded yet or both (wither when there is no receipt or when the receipt is queried but the transaction is not decoded)?

rotkehlchen/tasks/manager.py
```diff
@@ -425,7 +425,7 @@ def _maybe_decode_evm_transactions(self) -> Optional[gevent.Greenlet]:
         lock acquired.
         """
         dbevmtx = DBEvmTx(self.database)
-        amount_of_tx_to_decode = dbevmtx.count_hashes_not_decoded()
+        amount_of_tx_to_decode = dbevmtx.count_hashes_not_decoded(addresses=None)
```
nebolax, line 428: Hm, I guess this should be `number_of_tx_to_decode`, right?

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -272,14 +272,19 @@ def decode_transaction(
 
         return sorted(events, key=lambda x: x.sequence_index, reverse=False)
 
-    def get_and_decode_undecoded_transactions(self, limit: Optional[int] = None) -> None:
+    def get_and_decode_undecoded_transactions(
+            self,
+            limit: Optional[int] = None,
+            addresses: Optional[List[ChecksumEvmAddress]] = None,
+    ) -> None:
         """Checks the DB for up to `limit` undecoded transactions and decodes them.
```
nebolax, line 280: I guess the docstring needs an update

rotkehlchen/db/evmtx.py
```diff
@@ -271,14 +272,18 @@ def get_transaction_hashes_not_decoded(
             self,
             chain_id: Optional[ChainID],
             limit: Optional[int],
+            addresses: Optional[List[ChecksumEvmAddress]],
     ) -> List[EVMTxHash]:
         """Get transaction hashes for the transactions that have not been decoded.
```
nebolax, line 277: I guess the docstring needs an update

rotkehlchen/db/evmtx.py
```diff
@@ -290,10 +295,16 @@ def get_transaction_hashes_not_decoded(
             cursor.execute(querystr, bindings)
             return [make_evm_tx_hash(x[0]) for x in cursor]
 
-    def count_hashes_not_decoded(self) -> int:
+    def count_hashes_not_decoded(self, addresses: Optional[List[ChecksumEvmAddress]]) -> int:
         """Count the number of transactions queried that have not been decoded"""
```
nebolax, line 299: I guess the docstring needs an update

nebolax (commented):
Also api docs need adjustments

rotkehlchen/db/evmtx.py
yabirgb: Is because of the type discrepancy between `addresses * 2` and `addresses + addresses` but I agree that looks confusing. WIll try to use the same in both places

rotkehlchen/api/v1/resources.py
```diff
@@ -503,12 +504,15 @@ def delete(self) -> Response:
...
 
 class EthereumTransactionsDecodingResource(BaseMethodView):
-    post_schema = AsyncQueryArgumentSchema()
+    post_schema = MultipleAddressAsyncSchema()
 
     @require_loggedin_user()
     @use_kwargs(post_schema, location='json_and_query')
-    def put(self, async_query: bool) -> Response:
-        return self.rest_api.decode_pending_ethereum_transactions(async_query=async_query)
+    def put(self, async_query: bool, addresses: Optional[List[ChecksumEvmAddress]]) -> Response:
```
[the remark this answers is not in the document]
yabirgb: It should be post, thanks

rotkehlchen/db/evmtx.py
yabirgb: For that we have `get_transaction_hashes_no_receipt`. The usage of this functions is to decode the transactions that have not been decoded yet but we have the receipts already

rotkehlchen/db/evmtx.py
yabirgb: the `tx_hash` gets null when you do the join. in the table it can't be null but it is when you do the join that you the filtering

rotkehlchen/api/v1/schemas.py
yabirgb: I have abstracted it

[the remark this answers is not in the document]
yabirgb: No need indeed, was the copy paste. Is only for post, yes

rotkehlchen/db/evmtx.py
```diff
@@ -36,7 +36,8 @@
 
 TRANSACTIONS_MISSING_DECODING_QUERY = (
     'evmtx_receipts AS A LEFT OUTER JOIN evm_tx_mappings AS B ON A.tx_hash=B.tx_hash '
-    'AND A.chain_id=B.chain_ID WHERE B.tx_hash is NULL'
+    'AND A.chain_id=B.chain_ID LEFT JOIN evm_transactions AS C on '
```
LefterisJP, line 39: Why was B.tx_hash is NULL removed? In general why the change here? 

Reading later: Is it due to the filter you have created?

rotkehlchen/db/filtering.py
```diff
@@ -169,6 +169,30 @@ def prepare(self) -> tuple[list[str], list[Any]]:
...
+    and chain.
+    """
+    addresses: Optional[list[ChecksumEvmAddress]]
+    chain_id: Optional[ChainID]
+
+    def prepare(self) -> tuple[list[str], list[Any]]:
+        query_filters = ['B.tx_hash is NULL']
+        bindings: list[Union[int, ChecksumEvmAddress]] = []
+        if self.addresses is not None:
+            bindings += self.addresses + self.addresses
```
LefterisJP, line 186: No need to append to an emty list. Easier to understand:

```suggestion
            bindings = [*self.addresses, *self.addresses]
```

rotkehlchen/db/filtering.py
```diff
@@ -169,6 +169,30 @@ def prepare(self) -> tuple[list[str], list[Any]]:
...
+            query_filters.append(f'(C.from_address IN ({",".join("?"*len(self.addresses))}) OR C.to_address IN ({",".join("?"*len(self.addresses))}))')  # noqa: E501
```
LefterisJP, line 187: This is too long. you can calculate the question marks once and input them in the query filter.

rotkehlchen/tests/api/test_ethereum_transactions.py
```diff
@@ -1484,3 +1489,57 @@ def test_decoding_missing_transactions(rotkehlchen_api_server: 'APIServer') -> N
     result = assert_proper_response_with_result(response)
     outcome = wait_for_async_task(rotkehlchen_api_server, result['task_id'])
     assert outcome['result']['decoded_tx_number'] == 0
+
+
+def test_decoding_missing_transactions_by_address(rotkehlchen_api_server: 'APIServer') -> None:
+    """Test that decoding all pending transactions works fine when a filter by addres is set"""
```
LefterisJP, line 1495:
```suggestion
    """Test that decoding all pending transactions works fine when a filter by address is set"""
```

LefterisJP (commented):
some smaller commenhts

rotkehlchen/db/evmtx.py
nebolax: Yeah, because it is checked in the filter
image

LefterisJP: tx_hash is not unique

nebolax:
The only failed test is for Avalanche and thus is not related + it passed locally

LefterisJP (approved):
LGTM

PR #5210, opened 2022-12-02
Fix error editing tokens related to underlying assets

yabirgb:
The query for the underlying tokens was wrong and was returning the address of the parent token instead of the addresses of the underlying assets

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -13,6 +13,7 @@ Changelog
 * :bug:`2934` Users will now be able to deploy the docker image on a subpath instead of the root domain.
 * :bug:`-` Users won't get errors querying prices due the usage of defi oracles for querying prices of NFTs.
 * :bug:`-` Fix an edge-case problem with synchronizing database with the server.
+* :bug:`-` Users will be able to edit EVM tokens again and the information about underlying tokens will be correctly displayed.
 * :bug:`-` Fix an issue with nginx websocket proxying on docker.
 
 * :release:`1.26.1 <2022-11-04>`
--- a/rotkehlchen/globaldb/handler.py
+++ b/rotkehlchen/globaldb/handler.py
@@ -250,24 +250,31 @@ class GlobalDBHandler():
         underlying_tokens: Dict[str, List[Dict[str, str]]] = defaultdict(list)
         prepared_filter_query, bindings = filter_query.prepare()
         parent_query = """
-        SELECT A.identifier AS identifier, A.type, B.address, B.decimals, A.name, C.symbol, C.started, C.forked, C.swapped_for, C.coingecko, C.cryptocompare, B.protocol, B.chain, B.token_kind, D.notes, D.type AS custom_asset_type
-        FROM assets as A
+        SELECT A.identifier AS identifier, A.type, B.address, B.decimals, A.name, C.symbol,
+        C.started, C.forked, C.swapped_for, C.coingecko, C.cryptocompare, B.protocol, B.chain,
+        B.token_kind, D.notes, D.type AS custom_asset_type FROM
+        assets as A
         LEFT JOIN common_asset_details AS C ON C.identifier = A.identifier
         LEFT JOIN evm_tokens as B ON B.identifier = A.identifier
         LEFT JOIN custom_assets as D ON D.identifier = A.identifier
-        """  # noqa: E501
+        """
         query = f'SELECT * FROM ({parent_query}) {prepared_filter_query}'
-        underlying_tokens_query = """
-        SELECT assets.identifier AS identifier, address, token_kind, parent_token_entry, weight, name, symbol, type, chain
-        FROM underlying_tokens_list
-        INNER JOIN assets ON assets.identifier = parent_token_entry
-        INNER JOIN evm_tokens ON assets.identifier=evm_tokens.identifier
-        INNER JOIN common_asset_details ON common_asset_details.identifier = assets.identifier
-        """  # noqa: E501
+        # Get the identifier of the EVM tokens in the filter and query the underlying tokens where
+        # parent_token_entry is in that set of identifiers. We need to join with evm_tokens since
+        # the address column is present there. @yabirgb tested using JOIN instead of a subquery
+        # and the query increased in complexity since we need to join more tables and the test
+        # showed that the query using joins is slower than this one with the subquery.
+        # test with subquery took on average 1.17 seconds and with joins 1.53 seconds.
+        underlying_tokens_query = (
+            f'SELECT parent_token_entry, address, token_kind, weight FROM underlying_tokens_list '
+            f'LEFT JOIN evm_tokens ON underlying_tokens_list.identifier=evm_tokens.identifier '
+            f'WHERE parent_token_entry IN (SELECT identifier FROM ({query}))'
+        )
+
         with GlobalDBHandler().conn.read_ctx() as cursor:
             # get all underlying tokens
-            for entry in cursor.execute(f'SELECT * FROM ({underlying_tokens_query}) {prepared_filter_query}', bindings):  # noqa: E501
-                underlying_tokens[entry[0]].append(UnderlyingToken.deserialize_from_db((entry[1], entry[2], entry[4])).serialize())  # noqa: E501
+            for entry in cursor.execute(underlying_tokens_query, bindings):
+                underlying_tokens[entry[0]].append(UnderlyingToken.deserialize_from_db((entry[1], entry[2], entry[3])).serialize())  # noqa: E501
 
             cursor.execute(query, bindings)
             for entry in cursor:
--- a/rotkehlchen/tests/api/test_user_evm_tokens.py
+++ b/rotkehlchen/tests/api/test_user_evm_tokens.py
@@ -7,10 +7,12 @@ import requests
 
 from rotkehlchen.accounting.structures.balance import BalanceType
 from rotkehlchen.assets.asset import Asset, EvmToken, UnderlyingToken
+from rotkehlchen.assets.types import AssetType
 from rotkehlchen.balances.manual import ManuallyTrackedBalance
+from rotkehlchen.chain.ethereum.types import string_to_evm_address
 from rotkehlchen.constants import ONE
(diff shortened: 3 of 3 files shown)


rotkehlchen/globaldb/handler.py
```diff
@@ -258,16 +259,17 @@ def retrieve_assets(filter_query: 'AssetsFilterQuery') -> Tuple[List[Dict[str, A
         """  # noqa: E501
         query = f'SELECT * FROM ({parent_query}) {prepared_filter_query}'
         underlying_tokens_query = """
-        SELECT assets.identifier AS identifier, address, token_kind, parent_token_entry, weight, name, symbol, type, chain
+        SELECT parent_token_entry AS identifier, underlying_tokens_list.identifier, token_kind, weight, name, symbol, type, chain
         FROM underlying_tokens_list
-        INNER JOIN assets ON assets.identifier = parent_token_entry
+        LEFT JOIN assets ON assets.identifier = parent_token_entry
```
nebolax, line 264: Why left join and not inner?

rotkehlchen/globaldb/handler.py
```diff
@@ -258,16 +259,17 @@ def retrieve_assets(filter_query: 'AssetsFilterQuery') -> Tuple[List[Dict[str, A
         """  # noqa: E501
         query = f'SELECT * FROM ({parent_query}) {prepared_filter_query}'
         underlying_tokens_query = """
-        SELECT assets.identifier AS identifier, address, token_kind, parent_token_entry, weight, name, symbol, type, chain
+        SELECT parent_token_entry AS identifier, underlying_tokens_list.identifier, token_kind, weight, name, symbol, type, chain
```
nebolax, line 262: Why to do this renaming of `parent_token_entry` to `identifier` and then using `identifier` column as `underlying_tokens_list.identifier` ? From my perspective it just confuses me

rotkehlchen/globaldb/handler.py
```diff
@@ -258,16 +259,17 @@ def retrieve_assets(filter_query: 'AssetsFilterQuery') -> Tuple[List[Dict[str, A
...
-        INNER JOIN assets ON assets.identifier = parent_token_entry
+        LEFT JOIN assets ON assets.identifier = parent_token_entry
         INNER JOIN evm_tokens ON assets.identifier=evm_tokens.identifier
         INNER JOIN common_asset_details ON common_asset_details.identifier = assets.identifier
         """  # noqa: E501
         with GlobalDBHandler().conn.read_ctx() as cursor:
             # get all underlying tokens
             for entry in cursor.execute(f'SELECT * FROM ({underlying_tokens_query}) {prepared_filter_query}', bindings):  # noqa: E501
-                underlying_tokens[entry[0]].append(UnderlyingToken.deserialize_from_db((entry[1], entry[2], entry[4])).serialize())  # noqa: E501
+                address = caips_identifier_to_address(entry[1])
```
nebolax, line 271: Isn't it better to query address from the database?

rotkehlchen/globaldb/handler.py
nebolax, line 262: Are `name, symbol, type, chain` used anywhere?

rotkehlchen/globaldb/handler.py
yabirgb: Rewrote this section improving it after reviewing a bug with kelsos

rotkehlchen/globaldb/handler.py
yabirgb: they were needed due the filter

rotkehlchen/globaldb/handler.py
```diff
@@ -250,24 +250,26 @@ def retrieve_assets(filter_query: 'AssetsFilterQuery') -> Tuple[List[Dict[str, A
...
+            f'SELECT parent_token_entry, address, token_kind, weight FROM underlying_tokens_list '
+            f'LEFT JOIN evm_tokens ON underlying_tokens_list.identifier=evm_tokens.identifier '
+            f'WHERE parent_token_entry IN (SELECT identifier FROM ({query}))'
+        )
+
         with GlobalDBHandler().conn.read_ctx() as cursor:
             # get all underlying tokens
-            for entry in cursor.execute(f'SELECT * FROM ({underlying_tokens_query}) {prepared_filter_query}', bindings):  # noqa: E501
-                underlying_tokens[entry[0]].append(UnderlyingToken.deserialize_from_db((entry[1], entry[2], entry[4])).serialize())  # noqa: E501
+            print(underlying_tokens_query)
```
kelsos, line 270: Don't forget to remove the print here

kelsos:
I tested it and the bug with the second page having `null` underlying tokens is fixed now 👍

rotkehlchen/globaldb/handler.py
yabirgb: thanks

docs/changelog.rst
```diff
@@ -13,6 +13,7 @@ Changelog
 * :bug:`2934` Users will now be able to deploy the docker image on a subpath instead of the root domain.
 * :bug:`-` Users won't get errors querying prices due the usage of defi oracles for querying prices of NFTs.
 * :bug:`-` Fix an edge-case problem with synchronizing database with the server.
+* :bug:`-` Users will be able to edit again EVM tokens and the information about underlying tokens will be correctly displayed.
```
LefterisJP, line 16:
```suggestion
* :bug:`-` Users will be able to edit EVM tokens again and the information about underlying tokens will be correctly displayed.
```

rotkehlchen/globaldb/handler.py
```diff
@@ -250,24 +250,25 @@ def retrieve_assets(filter_query: 'AssetsFilterQuery') -> Tuple[List[Dict[str, A
...
+            f'SELECT parent_token_entry, address, token_kind, weight FROM underlying_tokens_list '
+            f'LEFT JOIN evm_tokens ON underlying_tokens_list.identifier=evm_tokens.identifier '
+            f'WHERE parent_token_entry IN (SELECT identifier FROM ({query}))'
+        )
+
         with GlobalDBHandler().conn.read_ctx() as cursor:
             # get all underlying tokens
-            for entry in cursor.execute(f'SELECT * FROM ({underlying_tokens_query}) {prepared_filter_query}', bindings):  # noqa: E501
-                underlying_tokens[entry[0]].append(UnderlyingToken.deserialize_from_db((entry[1], entry[2], entry[4])).serialize())  # noqa: E501
+            for entry in cursor.execute(f'SELECT * FROM ({underlying_tokens_query})', bindings):  # noqa: E501
```
LefterisJP, line 270: This part of the code is really complicated and to be hnest I don't get any of these queries very well. Can't they be written in a better way?

For example why this complexity here?

```python
f'SELECT * FROM ({underlying_tokens_query})', bindings
```
Can't you just use the actual query? Why do you do SELECT * from the other query again?

rotkehlchen/tests/api/test_user_evm_tokens.py
```diff
@@ -609,3 +610,97 @@ def test_add_non_ethereum_token(rotkehlchen_api_server):
     assert token.symbol == 'XYZ'
     assert token.chain == ChainID.BINANCE
     assert token.protocol == 'my-own-protocol'
+
+
+@pytest.mark.parametrize('use_clean_caching_directory', [True])
+def test_adding_custom_evm_token(rotkehlchen_api_server):
+    """Test that the information for an evm token is correcrtly processed in the API response"""
```
LefterisJP, line 617:
```suggestion
    """Test that the adding an evm token with underlying tokens is correctly processed by the API"""
```

rotkehlchen/tests/api/test_user_evm_tokens.py
```diff
@@ -609,3 +610,97 @@ def test_add_non_ethereum_token(rotkehlchen_api_server):
     assert token.symbol == 'XYZ'
     assert token.chain == ChainID.BINANCE
     assert token.protocol == 'my-own-protocol'
+
+
+@pytest.mark.parametrize('use_clean_caching_directory', [True])
+def test_adding_custom_evm_token(rotkehlchen_api_server):
```
LefterisJP, line 616:
```suggestion
def test_adding_evm_token_with_underlying_token(rotkehlchen_api_server):
```

rotkehlchen/globaldb/handler.py
yabirgb: Indeed was my fault, is not needed the `SELECT from` and is there from the previous query. Regarding the queries I'll explain what each one do

rotkehlchen/globaldb/handler.py
```diff
@@ -250,24 +250,28 @@ def retrieve_assets(filter_query: 'AssetsFilterQuery') -> Tuple[List[Dict[str, A
...
-        INNER JOIN evm_tokens ON assets.identifier=evm_tokens.identifier
-        INNER JOIN common_asset_details ON common_asset_details.identifier = assets.identifier
-        """  # noqa: E501
+        # Get the identifier of the EVM tokens in the filter and query the underlying tokens where
+        # parent_token_entry is in that set of identifiers. We need to join with evm_tokens since
+        # the address column is present there.
+        underlying_tokens_query = (
+            f'SELECT parent_token_entry, address, token_kind, weight FROM underlying_tokens_list '
+            f'LEFT JOIN evm_tokens ON underlying_tokens_list.identifier=evm_tokens.identifier '
+            f'WHERE parent_token_entry IN (SELECT identifier FROM ({query}))'
```
LefterisJP, line 268: Subqueries are generally a non-performant way to retrieve something when compared to a JOIN. And here your subquery is huge. It's ALL of our assets. I think this should be turned to a join. Would also be fun to compare this query versus a join version instead of the subquery.

Depends on the DB optimizer and the particular query, but wonder how much of a diff it would be.

kelsos: This approach is my fault since I was thinking only from the perspective of the application as the consumer. 

When the application is the consumer, this subquery is never all the assets because of pagination. The pagination can be at most 100 items and not more. 

FYI In the case of the 10-100 items per page, the API response time is around 40-90ms

PR #5211, opened 2022-12-02
Adds explicit returns types, removes adex, fixes unintended type erasure

kelsos:
Closes #(issue_number)

## Checklist


frontend/app/src/types/reports.ts
```diff
@@ -97,15 +96,27 @@ export interface ReportError {
...
-export const Reports = Entries(z.array(Report));
+export const Reports = z.object({
+  entries: z.array(Report),
+  entriesFound: z.number(),
+  entriesLimit: z.number()
+});
 
 export type Reports = z.infer<typeof Reports>;
 
-export const ProfitLossReportOverview = Entries(z.array(ProfitLossOverview));
```
kelsos, line 104: @lukicenturi I realized that these types were actually breaking type inference since we didn't pass generics in the method.

I spend some time to figure out I we can do something the preserves the types properly but it didn't work, and I decided against it. 

We can try again in the future to see if it is possible.

PR #5212, opened 2022-12-02
Optimism refactor step 2

LefterisJP:


LefterisJP:
All failed tests are only due to covalent.

"Error message: backend queue is full and cannot accept request and text {"data":null,"error":true,"error_message":"backend queue is full and cannot accept request","error_code":507}'}"

This is still happening and so they faill locally. But this is unrelated to the PR.

PR #5213, opened 2022-12-03
Enforce typing in contract constants

LefterisJP:


LefterisJP:
All beautifully green. Merging.

PR #5214, opened 2022-12-03
Adds more returns types and introduces eslint cache

kelsos:
Closes #(issue_number)

## Checklist


kelsos:
@LefterisJP if this change works, code analysis and eslint will run when `develop` has changes.

If it works, this will create a baseline for code analysis and a baseline cache for frontend PRs. I am trying the `--cache` option locally, which saves a lot of waiting time. 

I had no idea it existed, I was looking for ways to optimize performance today and I came across an article :)

If it doesn't I just have to revert the `rotki_ci.yml` modifications

LefterisJP:
I see thanks. Anyway we can apply this to backend PRs? Or just talking about the eslint thing?

kelsos:
I need to check how it will work on develop, I might have to do a follow up if things don't work but the idea is that we will run only code-analysis for the baseline to compare (both backend/frontend) and also create a lint cache that PRs can use.

LefterisJP (commented):
everything seems green here

PR #5215, opened 2022-12-04
Upgrade packages and enforce all UPXX ruff rules

LefterisJP:
Upgrade packages and enforce all UPXX ruff rules

Check individual commits for details.


LefterisJP:
Only failed test is: `test_remove_avax_blockchain_account` due to covalent.

PR #5218, opened 2022-12-04
Remove a misleading error log in trades querying

LefterisJP:


LefterisJP:
Only frontend 

```
 │ ✖  accounts-balances/blockchain-balanc      08:50        5        -        5        -        - │
  │    es.spec.ts                   

```

failed. We would need to see why this is happening, but not related to the PR

PR #5220, opened 2022-12-05
Completes migration to zod schemas

kelsos:
Closes #(issue_number)

## Checklist


kelsos:
@lukicenturi There is another thing that would probably make sense as part of this PR but it would also require some testing.

This would be to globally enable the `snake_case` transformer and the custom `paramSerializer`. I need to check it to make sure that this doesn't create any problems.

frontend/app/src/types/defi/compound.ts
```diff
@@ -1,34 +1,41 @@
...
 
-type CompoundRewards = Readonly<Record<string, CompoundReward>>;
+const CompoundRewards = z.record(CompoundReward);
 
-interface CompoundLending extends HasBalance {
-  readonly apy: string | null;
-}
+const CompoundLending = z.object({
+  balance: Balance,
+  apy: z.string().nullish()
```
lukicenturi, line 18: `nullish()` is `| null | undefined`. 
Are you sure?

frontend/app/src/types/defi/compound.ts
```diff
@@ -38,39 +45,45 @@ export const COMPOUND_EVENT_TYPES = [
...
-export type CompoundProfitAndLoss = Readonly<
-  Record<string, CompoundAssetProfitAndLoss>
->;
+const CompoundEvent = z.object({
+  eventType: CompoundEventType,
+  address: z.string(),
+  blockNumber: z.number(),
+  timestamp: z.number(),
+  asset: z.string(),
+  toAsset: z.string().nullable(),
```
lukicenturi, line 58: `nullable()` is `| null`.
But the previous scheme is `| undefined` which should be`optional()` (?)

frontend/app/src/services/axios-tranformers.ts
```diff
@@ -98,28 +72,21 @@ export const axiosNoRootCamelCaseTransformer: AxiosResponseTransformer = (
...
-      try {
-        result = JSON.parse(data, reviver);
-        // eslint-disable-next-line no-empty
-      } catch (e: any) {}
-    }
-    return result;
-  };
+const jsonTransformer: AxiosResponseTransformer = (data, headers) => {
+  let result = data;
+  const isJson = headers?.['content-type'] === 'application/json';
```
lukicenturi, line 77: ```const isJson = headers?.['content-type']?.includes('application/json');```

lukicenturi: some api have this 
image

lukicenturi (changes requested):

frontend/app/src/types/defi/compound.ts
kelsos: It was wrong though, I tried optional but the response was returning null instead.

frontend/app/src/types/defi/compound.ts
kelsos: I can play around a bit to see what the response is. We also need to go through the backend responses and see where it can be nullable or optional at some point.

In general, the backend returns either one or the other in most cases.

frontend/common/src/defi/aave/index.ts
```diff
@@ -43,58 +51,69 @@ export enum AaveLendingEventType {
...
+
+export type AaveLendingEvent = z.infer<typeof AaveLendingEvent>;
+
+export const AaveLiquidationEvent = z.object({
+  eventType: z.literal(AaveBorrowingEventType.LIQUIDATION),
+  collateralBalance: Balance,
+  collateralAsset: z.string(),
+  principalAsset: z.string(),
+  principalBalance: Balance
+}).merge(AaveEvent);
```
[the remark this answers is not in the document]
kelsos: Let me see

frontend/app/src/types/defi/compound.ts
kelsos: It is nullable.

kelsos:
> @lukicenturi There is another thing that would probably make sense as part of this PR but it would also require some testing.
> 
> This would be to globally enable the `snake_case` transformer and the custom `paramSerializer`. I need to check it to make sure that this doesn't create any problems.

I will split this to a new PR.

frontend/common/src/defi/aave/index.ts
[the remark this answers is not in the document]
kelsos: Yes, all of the `Borrowing` events should be Based on the `AaveBaseEvent` schema. It is fixed now

kelsos:
I think I got the aave schema correct now, it also works fine.

lukicenturi (approved):
LGTM!

PR #5224, opened 2022-12-06
Updates job checker action

kelsos:
Closes #(issue_number)

## Checklist

## Notes

Initially, due to a misconception, the action was setting an environment variable. 
Since environment variables are not actually available outside of a job's steps and 
you have to re-export this was changed to an output for consistency:
https://github.com/rotki/action-job-checker/pull/3

The version bumping happened because otherwise the regular ci would break


LefterisJP:
kelsos from discord explaining what this does

> I changed the action to use set output instead of setting the env var 
> since env variables are not available outside of the job it doesn't make sense to use an env var
> but initially I didn't know that

Issue #5221, opened 2022-12-06
Ask user if they should resume from backup after botched DB upgrade


LefterisJP:
## Abstract

At https://github.com/rotki/rotki/pull/5203/ the ability to revert to an automatic backup taken right before a botched upgrade is added. As per https://github.com/rotki/rotki/pull/5203/files#r1039927071 and a further discussion we had in discord we should probably add an interaction with the user there.

Essentially asking them if they want to revert to that upgrade or quit.

Investigate if same can be done for the global DB upgrade.

## Task

- [ ] Do it
- [ ] Also a subtask is to actually add unit tests for this functionality, as the above PR has no tests and so we can't be sure it actually does  what it should.

LefterisJP:
This should have been closed on the last release. @lukicenturi when moving something to "Done" please close it.

PR #5225, opened 2022-12-06
Add tests for ongoing upgrades

nebolax:
Part of #5221 

Add tests for ongoing upgrades


rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1339,3 +1342,46 @@ def test_old_versions_raise_error(user_data_dir):  # pylint: disable=unused-argu
...
+    with pytest.raises(DBUpgradeError):
+        _init_db_with_target_version(
+            target_version=34,
+            user_data_dir=user_data_dir,
+            msg_aggregator=msg_aggregator,
+        )
+
+    # Add a backup
+    shutil.copy(
+        os.path.join(
```
LefterisJP, line 1372: as I wrote in the previous PR, please don't use the os module. Use the more moden `pathlib` module

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1339,3 +1342,46 @@ def test_old_versions_raise_error(user_data_dir):  # pylint: disable=unused-argu
             msg_aggregator=msg_aggregator,
         )
     assert 'Your account was last opened by a very old version of rotki' in str(upgrade_exception)
+
+
+def test_ongoing_upgrades(user_data_dir):
```
LefterisJP, line 1347: Call them unfinished upgrades. As ongoing is a misleading title. It's not ongoing. It started, but was stopped in the middle.

```suggestion
def test_unfinished_upgrades(user_data_dir):
```

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1339,3 +1342,46 @@ def test_old_versions_raise_error(user_data_dir):  # pylint: disable=unused-argu
...
+        )
+
+    # Add a backup
+    shutil.copy(
+        os.path.join(
+            os.path.dirname(os.path.dirname(os.path.realpath(__file__))),
+            'data',
+            'v33_rotkehlchen.db',
+        ),
+        os.path.join(user_data_dir / f'{ts_now()}_rotkehlchen_db_v33.backup'),
```
LefterisJP, line 1377: funny thing is you are actually using the pathlib module here :D. The `/` operator does this. And then you turn it back to a string to push to os.path.join
```suggestion
        user_data_dir / f'{ts_now()}_rotkehlchen_db_v33.backup',
```

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1339,3 +1342,46 @@ def test_old_versions_raise_error(user_data_dir):  # pylint: disable=unused-argu
...
+    )
+
+    db = _init_db_with_target_version(  # Now the backup should be used
+        target_version=34,
+        user_data_dir=user_data_dir,
+        msg_aggregator=msg_aggregator,
+    )
+    # Check that there is no setting left
+    with db.conn.read_ctx() as cursor:
+        assert db.get_setting(cursor, 'ongoing_upgrade_from_version') is None
```
LefterisJP, line 1387: Can you also add a check that the backup was used? Probably that the size ofthe DB is as expected? Cause I think normal empty test DB should be smaller than v33_rotkehlchen.db

rotkehlchen/tests/unit/test_globaldb_upgrades.py
```diff
@@ -141,3 +147,33 @@ def test_upgrade_v2_v3(globaldb):
             ('BIFI', 'BIFI'),
         )
         assert cursor.fetchone()[0] == 0
+
+
+@pytest.mark.parametrize('globaldb_upgrades', [[]])
+@pytest.mark.parametrize('globaldb_version', [2])
+@pytest.mark.parametrize('target_globaldb_version', [2])
+@pytest.mark.parametrize('reload_user_assets', [False])
+def test_ongoing_upgrades(globaldb: GlobalDBHandler):
```
LefterisJP, line 156: same comment for ongoing/unfinished here

rotkehlchen/tests/unit/test_globaldb_upgrades.py
```diff
@@ -141,3 +147,33 @@ def test_upgrade_v2_v3(globaldb):
...
+    globaldb.add_setting_value(  # Pretend that an upgrade was started
+        name='ongoing_upgrade_from_version',
+        value=2,
+    )
+    # There are no backups, so it is supposed to raise an error
+    with pytest.raises(DBUpgradeError):
+        create_globaldb(globaldb._data_directory, 0)
+
+    # Add a backup
+    shutil.copy(
```
LefterisJP, line 167: same comments for pathlib

rotkehlchen/tests/unit/test_globaldb_upgrades.py
```diff
@@ -141,3 +147,33 @@ def test_upgrade_v2_v3(globaldb):
...
+            'data',
+            'v2_global.db',
+        ),
+        globaldb._data_directory / 'global_data' / f'{ts_now()}_global_db_v2.backup',  # type: ignore  # _data_directory is definitely not null  # noqa: E501
+    )
+
+    globaldb = create_globaldb(globaldb._data_directory, 0)  # Now the backup should be used
+    assert globaldb.used_backup is True
+    # Check that there is no setting left
+    assert globaldb.get_setting_value('ongoing_upgrade_from_version', -1) == -1
```
LefterisJP, line 179: same comment for checking the size ... or some other way ... to check that the DB is the backup and not something else

LefterisJP (commented):
Some comments left. nice and easy to review PR.

rotkehlchen/tests/db/test_db_upgrades.py
nebolax: Ah okay, will change. I just followed the approach that I saw in other places of the tests

rotkehlchen/tests/db/test_db_upgrades.py
nebolax: The sizes of the databases are identical since same files are used for the initial db and the backup db, but I added a flag to the backup

rotkehlchen/tests/db/test_db_upgrades.py
LefterisJP: yean I get it. We still use os.path.join in some places. We can phase it out slowly.

rotkehlchen/tests/db/test_db_upgrades.py
LefterisJP: Cool. That should work.

LefterisJP (approved):
Great thanks. LGTM.

PR #5228, opened 2022-12-07
Cryptocom cashback reversal

nebolax:
Fix importation of cryptocom cashback reversal

Until this PR cashback reversal (which happens when a user receives cashback for using their cryptocom card but then cryptocom decides that the user doesn't deserve this cashback and deducts it) was treated as a Trade with type Buy.

Now I made it a Ledger action with type Expense

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -15,6 +15,7 @@ Changelog
 * :bug:`-` Fix an edge-case problem with synchronizing database with the server.
 * :bug:`-` Users will be able to edit EVM tokens again and the information about underlying tokens will be correctly displayed.
 * :bug:`-` Fix an issue with nginx websocket proxying on docker.
+* :bug:`-` Now `Card Cashback Reversal` entries from cryptocom csv will be imported correctly.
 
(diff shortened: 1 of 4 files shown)


docs/changelog.rst
```diff
@@ -15,6 +15,7 @@ Changelog
 * :bug:`-` Fix an edge-case problem with synchronizing database with the server.
 * :bug:`-` Users will be able to edit EVM tokens again and the information about underlying tokens will be correctly displayed.
 * :bug:`-` Fix an issue with nginx websocket proxying on docker.
+* :bug:`-` Now `Card Cashback Reversal` entries from cryptocom csv will be imported correctly.
```
yabirgb, line 18: Can be changed in other PR since all the tests are green

```suggestion
* :bug:`-` Now `Card Cashback Reversal` entries from cryptocom csv will be correctly imported.
```

yabirgb (approved):
Approving, it makes sense. We didn't have it in the tests and probably was a guess based on the data we had. Thanks. @LefterisJP can you check?

LefterisJP (approved):
LGTM

PR #5232, opened 2022-12-08
Remove `geventwebsocket` debug logs

prettyirrelevant:
Fixes https://discord.com/channels/657906918408585217/1050071468152733736/1050076437652181062

Example:
```sh
[06/12/2022 10:23:49 WAT] INFO geventwebsocket.handler 127.0.0.1 - - [2022-12-06 10:23:49] "GET /api/1/users HTTP/1.1" 200 339 0.002334
[06/12/2022 10:23:49 WAT] DEBUG geventwebsocket.handler Initializing WebSocket
[06/12/2022 10:23:49 WAT] DEBUG geventwebsocket.handler Validating WebSocket request
```

### DIFF
--- a/rotkehlchen/api/server.py
+++ b/rotkehlchen/api/server.py
@@ -463,6 +463,8 @@ class APIServer():
             environ={'rotki_notifier': self.rotki_notifier},
             error_log=wsgi_logger,
         )
+        # this is to prevent littering logs with geventwebsocket upgrade messages
+        logging.getLogger('geventwebsocket.handler').setLevel(logging.ERROR)
 
         if 'pytest' not in sys.modules:  # do not check
             if __debug__:


LefterisJP:
Asked Isaac in discord if we can confirm a websocket is being re-used and not re-initialized each time to which he responded:

1. https://github.com/wwwtyro/gevent-websocket/blob/655216d9defa6d98933a39a1e48294613f971cb6/geventwebsocket/handler.py#L34
^ the method called for an actual websocket connection. It stores a dict of key-value for each client it connects to.

2. https://github.com/wwwtyro/gevent-websocket/blob/655216d9defa6d98933a39a1e48294613f971cb6/geventwebsocket/handler.py#L90
^ where it tries to upgrade every request to websocket request. If it fails, it uses the default wsgi handler. Otherwise, it calls (1).

Just putting it here to document that we do indeed always use an existing websocket and these logs are just noise and should be removed.

Issue #4872, opened 2022-09-23
Frontend changes meta issue


kelsos:
## Abstract

I have a few things in mind that we should do in the frontend.
This is a meta issue to keep track of and proceed with. **We can create and link issues when needed.**

## Motivation

Improve the day-to-day situation (linting/editor performance) etc.

## Specification

- [x] Split `rotkehlchen-api` to smaller utility classes  https://github.com/rotki/rotki/blob/172c1ad9d055200eeeeb6dd93e4f063739f9b9e8/frontend/app/src/services/rotkehlchen-api.ts#L73 (#5259) (#5578)

> An example of the approach can be found here:
https://github.com/rotki/rotki/blob/172c1ad9d055200eeeeb6dd93e4f063739f9b9e8/frontend/app/src/services/session/queried-addresses-api.ts#L13

- [x] Try to keep typescript files around ~20 imports ~400loc (check if there is an ESLint plugin for that) 
- [x] Slowly convert components to use `setup` (ESLint now warns but it will turn to an error when this is done.)
- [x] move any exported type, utility classes, and consts out of the `/composables` and `stores` directories. (When done we can enable auto-import for these)(#5571)
- [x]  Finish removing the `keys` approach for auto conversion of strings to BigNumber and replace all cases with zod schemas (balance keys have been marked as deprecated (#5220)
- [x] Introduce layouts with [vite-plugin-vue-layouts](https://github.com/JohnCampionJr/vite-plugin-vue-layouts), split loading/login/account creation to different views
- [x] Evaluate if it makes sense to at some point migrate to [vite-plugin-pages](https://github.com/hannoeru/vite-plugin-pages) #5075
- [x] Replace usage of vuetify validation using `rules` with vuelidate (#5230)
- [x] Check if it makes sense to move to `pnpm` (#5237).
- [x] Update to unified confirmation dialog (#5249)

> I will add more stuff in the future but this is just to keep track of things
@lukicenturi if you also have any ideas please add them here.

kelsos:
There are some problems with `vite-plugin-pages` for our use-case, but the views have been renamed go pages and they follow a similar structure.

After updating to vue 3 and vue-router 4.x we can instead move to [unplugin-vue-router](https://github.com/posva/unplugin-vue-router) which uses a similar folder structure as the pages plugin and has some generation for routes that would actually help.

kelsos:
I started this with the multi evm addition but we can now slowly start:

[...]

PR #5237, opened 2022-12-09
Updates frontend eslint config

kelsos:
Part of #4872

## Checklist

## TODO

- [x] Resolve situation with refreshing package-lock.json or move to pnpm
- [x] Update to Cypress 12 > Ignoring since cypress 12 has a breaking change with test isolation
- [ ] Introduce type-checking for tests


kelsos:
@rotki/rotki-devs So until we release 1.27 we will have to have different behavior between `bugfixes` and `develop`. This PR replaces `npm` with [`pnpm`](https://pnpm.io) as the used package manager.

### Bugfixes

For `bugfixes` it is business as usual when switching from develop:

```bash
npm run clean:modules
npm ci
npm run dev
```

### Develop

For `develop` after this PR we will now use `pnpm` as the package manager which means that there are a few extra steps you need to follow.

```bash
# If you never have installed pnpm before
npm install -g pnpm@7 # It should install pnpm>=7.18.1
## Then
pnpm run clean:modules
pnpm install
pnpm run dev
```

After installing `pnpm` in most cases you just need to replace `npm` with `pnpm` when starting the commands.

If the commands are in a subproject `-w workspace` should be replaced with `--filter project_name`. *(It is in the `package.json`)*

It is also in the documentation but some of the commands are:

```bash
pnpm run lint
pnpm run lint:fix
pnpm run build
pnpm run --filter rotki test:unit
pnpm run --filter rotki test:integration-ci 
```

PR #5244, opened 2022-12-12
Make sure to query receipts before decoding

yabirgb:
The transactions were not being correctly decoded because the receipt logs were not stored locally. Since this endpoint is used to decode everything pending I opted for downloading the missing receipts. In the worst case this can be a slow task but is needed to obtain accurate history.

Also we have a periodic task that does this in the background so it shouldn't be in the average case a slow task

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -3355,6 +3355,11 @@ class RestAPI():
             evm_addresses: Optional[list[ChecksumEvmAddress]],
     ) -> dict[str, Any]:
         dbevmtx = DBEvmTx(self.rotkehlchen.data.db)
+        ethereum_manager = self.rotkehlchen.chains_aggregator.ethereum
+        # make sure that all the receipts are already queried
+        ethereum_manager.transactions.get_receipts_for_transactions_missing_them(
+            addresses=evm_addresses,
+        )
         amount_of_tx_to_decode = dbevmtx.count_hashes_not_decoded(
             addresses=evm_addresses,
             chain_id=ChainID.ETHEREUM,
@@ -3373,6 +3378,17 @@ class RestAPI():
             async_query: bool,
             evm_addresses: Optional[list[ChecksumEvmAddress]],
     ) -> Response:
+        """
+        This method should be called after querying ethereum transactions and does the following:
+        - Query missing receipts
+        - Decode ethereum transactions
+
+        It can be a slow process and this is why it is important to set the list of addresses
+        queried per module that need to be decoded.
+
+        This logic is executed by the frontend in pages where the set of transactions needs to be
+        up to date, for example, the liquity module.
+        """
         if async_query is True:
             return self._query_async(
                 command=self._decode_pending_ethereum_transactions,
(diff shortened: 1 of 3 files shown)


rotkehlchen/api/rest.py
```diff
@@ -3340,6 +3340,11 @@ def _decode_pending_ethereum_transactions(
             evm_addresses: Optional[list[ChecksumEvmAddress]],
     ) -> dict[str, Any]:
         dbevmtx = DBEvmTx(self.rotkehlchen.data.db)
+        ethereum_manager = self.rotkehlchen.chains_aggregator.ethereum
+        # make sure that all the receipts are already queriedsssss
```
LefterisJP, line 3344: Queriedsssss?

rotkehlchen/api/rest.py
```diff
@@ -3358,6 +3363,17 @@ def decode_pending_ethereum_transactions(
...
     ) -> Response:
+        """
+        This method should be called after querying ethereum transactions and does the following:
+        - Query missing receipts
+        - Decode ethereum transactions
+
+        It can be a slow process and this is why it is important to set the list of addresses
+        queried per module that need to be decoded.
+
+        This logic is executed by the frontend in pages where the set of transactions needs to be
```
LefterisJP, line 3374: Question: if this method is to be only queried by the front-end, does it not also query the ethereum transactions? Or does the fronted have to make sure transactions are queried? Why not keep it simple and do it all in one place? Or perhaps I am misunderstanding

rotkehlchen/chain/evm/transactions.py
```diff
@@ -466,7 +466,11 @@ def get_or_query_transaction_receipt(
...
         return tx_receipt  # type: ignore  # tx_receipt was just added in the DB so should be there  # noqa: E501
 
-    def get_receipts_for_transactions_missing_them(self, limit: Optional[int] = None) -> None:
+    def get_receipts_for_transactions_missing_them(
+            self,
+            limit: Optional[int] = None,
+            addresses: Optional[list[ChecksumEvmAddress]] = None,
+    ) -> None:
         """
         Searches the database for up to `limit` transactions that have no corresponding receipt
```
LefterisJP, line 475: Add a mention about address filtering here in the docstring

rotkehlchen/api/rest.py
yabirgb: We have this as two queries since the fronted reuses the same logic they have for the ethereum transactions view. So the ethereum transaction query logic is not only used here. What is only used in this page is the `decode all pending transactions`

rotkehlchen/chain/evm/transactions.py
```diff
@@ -466,19 +466,33 @@ def get_or_query_transaction_receipt(
...
         and for each one of them queries the receipt and saves it in the DB.
 
         It's protected by a lock to not enter the same code twice
         (i.e. from periodic tasks and from pnl report history events gathering)
+
+        If the addresses argument is provided then is used to filter the transactions missing
```
LefterisJP, line 481:
```suggestion
        If the addresses argument is provided then it is used to filter the transactions missing
```

LefterisJP (commented):
Leaving a small typo comment but also the stuff we discussed in discord need to be figured out before proceeding

LefterisJP (approved):
LGTM as a temporary solutoin as discussed. I am rerunning the frontend tests as they failed as usual

PR #5246, opened 2022-12-12
Fix notes for token transfers between owned accounts

nebolax:
Fix notes for token transfers between owned accounts

Previously direction of the transfer was reversed.

### DIFF
--- a/rotkehlchen/chain/ethereum/decoding/base.py
+++ b/rotkehlchen/chain/ethereum/decoding/base.py
@@ -77,7 +77,7 @@ class BaseDecoderTools():
             event_type = HistoryEventType.TRANSFER
             location_label = from_address
             counterparty = to_address
-            verb = 'Send'
+            verb = 'Transfer'
         elif tracked_from:
             if to_exchange is not None:
                 event_type = HistoryEventType.DEPOSIT
@@ -131,7 +131,7 @@ class BaseDecoderTools():
         amount_raw_or_token_id = hex_or_bytes_to_int(tx_log.data)
         if token.token_kind == EvmTokenKind.ERC20:
             amount = token_normalized_value(token_amount=amount_raw_or_token_id, token=token)
-            if event_type == HistoryEventType.SPEND:
+            if event_type in {HistoryEventType.SPEND, HistoryEventType.TRANSFER}:
                 notes = f'{verb} {amount} {token.symbol} from {location_label} to {counterparty}'
             else:
                 notes = f'{verb} {amount} {token.symbol} from {counterparty} to {location_label}'
@@ -151,7 +151,7 @@ class BaseDecoderTools():
             amount = ONE
             name = 'ERC721 token' if token.name == '' else token.name
             extra_data = {'token_id': token_id, 'token_name': name}
-            if event_type == HistoryEventType.SPEND:
+            if event_type in {HistoryEventType.SPEND, HistoryEventType.TRANSFER}:
                 notes = f'{verb} {name} with id {token_id} from {location_label} to {counterparty}'  # noqa: E501
             else:
                 notes = f'{verb} {name} with id {token_id} from {counterparty} to {location_label}'  # noqa: E501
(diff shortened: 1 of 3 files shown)


rotkehlchen/chain/ethereum/decoding/base.py
```diff
@@ -131,7 +131,7 @@ def decode_erc20_721_transfer(
         amount_raw_or_token_id = hex_or_bytes_to_int(tx_log.data)
         if token.token_kind == EvmTokenKind.ERC20:
             amount = token_normalized_value(token_amount=amount_raw_or_token_id, token=token)
-            if event_type == HistoryEventType.SPEND:
+            if verb == 'Send':
```
yabirgb, line 134: I believe it is better to avoid direct string compassion like this if we can. One of the reasons is that we can change the text in the future and this will be harder to find or if we introduce translations as codes.

Looking at the code it appears that `Send` appears when it is a `TRANSFER` or a `SPEND` so why not changing it to

```suggestion
            if event_type == (HistoryEventType.SPEND, HistoryEventType.TRANSFER):
```

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -131,3 +134,97 @@ def test_no_logs_and_zero_eth(
...
+        events = evm_transaction_decoder.decode_transaction(
+            write_cursor=cursor,
+            transaction=transaction,
+            tx_receipt=receipt,
+        )
+    assert events == [
+        HistoryBaseEntry(
+            event_identifier=HexBytes(
+                '0xbb58b36ddc027a1070131e68b915e5f0dca37767b020ed164eda681725b5ca4e',
+            ),
```
yabirgb, line 200: can't you use `evmhash` here?

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -131,3 +134,97 @@ def test_no_logs_and_zero_eth(
...
+            balance=Balance(),
+            location_label='0x4bBa290826C253BD854121346c370a9886d1bC26',
+            notes='Burned 0 ETH for gas',
+            counterparty='gas',
+            identifier=None,
+            extra_data=None,
+        ), HistoryBaseEntry(
+            event_identifier=HexBytes(
+                '0xbb58b36ddc027a1070131e68b915e5f0dca37767b020ed164eda681725b5ca4e',
+            ),
```
yabirgb, line 216: same, use the variable you have

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -131,3 +134,97 @@ def test_no_logs_and_zero_eth(
...
+            transaction=transaction,
+            tx_receipt=receipt,
+        )
+    assert events == [
+        HistoryBaseEntry(
+            event_identifier=HexBytes(
+                '0xbb58b36ddc027a1070131e68b915e5f0dca37767b020ed164eda681725b5ca4e',
+            ),
+            sequence_index=0,
+            timestamp=0,
```
yabirgb, line 202: We will have to do it anyway in the future so is good if we don't introduce type errors. Can you change this to 

```suggestion
            timestamp=Timestamp(0),
```

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -131,3 +134,97 @@ def test_no_logs_and_zero_eth(
...
+            notes='Burned 0 ETH for gas',
+            counterparty='gas',
+            identifier=None,
+            extra_data=None,
+        ), HistoryBaseEntry(
+            event_identifier=HexBytes(
+                '0xbb58b36ddc027a1070131e68b915e5f0dca37767b020ed164eda681725b5ca4e',
+            ),
+            sequence_index=74,
+            timestamp=0,
```
yabirgb, line 218: same comment about the timestamp

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -131,3 +134,97 @@ def test_no_logs_and_zero_eth(
...
+            location=Location.BLOCKCHAIN,
+            event_type=HistoryEventType.TRANSFER,
+            event_subtype=HistoryEventSubType.NONE,
+            asset=A_USDT,
+            balance=Balance(amount=FVal('38.002229')),
+            location_label=from_address,
+            notes=f'Send 38.002229 USDT from {from_address} to {to_address}',
```
LefterisJP, line 225: Actually can you also change the verb to `Transfer` in decoding/base.py? It may make sense to use a different verb too for the TRANSFER type

LefterisJP (approved):
LGTM

PR #5248, opened 2022-12-12
Narrow down supported chains, move account data to own file

LefterisJP:
Also abstract and generalize the code in few places regarding supported chains and not repeat elifs. (more places to do this left)


LefterisJP:
Frontend tests faill all the time. Not sure why. Took it to discord. Merging.

PR #5254, opened 2022-12-13
Large premium user DBs should no longer hit plaintext database error

LefterisJP:
It happened due to context switching during the attaching of the database, some stuff happening elsewhere and in the meantime the upload task starting again


LefterisJP:
I will force merge. Frontends fail in the funny way that will always fail no matter how many times I restart. Already tried 3 times.

PR #5256, opened 2022-12-13
Use different name for data directory if non-packaged

### LINKED ISSUE #4978: Think of a way to keep development accounts separately
## Abstract

Between multiple development machines and between `production`/`develop` it becomes harder to keep track of which database has been used by which branch. This can lead to partially updated databases being used.

As a proposal, we could somehow separate where the `develop` accounts and the production accounts are stored so that they are not sharing the same place.

We can always copy accounts from production to develop manually (this can go to the guide).

We should also consider how this affects nightlies and how nightlies are treated. For example we might want to treat nightlies as development to avoid having users mess accidentally with their production accounts.  

## Motivation

Helps better track which accounts are used in `develop`/`production`.

## Specification

- TDB

LefterisJP:
Fix #4978

### DIFF
--- a/rotkehlchen/config.py
+++ b/rotkehlchen/config.py
@@ -2,6 +2,7 @@ import logging
 import os
 import platform
 import shutil
+import sys
 from pathlib import Path
 
 from rotkehlchen.logging import RotkehlchenLogsAdapter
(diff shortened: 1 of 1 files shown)


LefterisJP:
Only failing test is covalent. Hmm weird. So that means no other tests failed. Okay nice. I tested that running it locally in develop uses the develop_data directory. Did not try making a package to see if it still uses the proper default dir but I guess it does. Merging.

PR #5257, opened 2022-12-13
Fix duplicated last_queried_ts

nebolax:
Fix duplicated last_queried_ts.

Before this PR each time tokens were detected, a new `last_queried_ts` entry was added into the db without removing the previous ones.

### DIFF
--- a/rotkehlchen/db/dbhandler.py
+++ b/rotkehlchen/db/dbhandler.py
@@ -1308,6 +1308,10 @@ class DBHandler:
         returned_list = []
         for (key, value) in cursor:
             if key == ACCOUNTS_DETAILS_LAST_QUERIED_TS:
+                # At the moment last_queried_timestamp is not used. It used to be a cache for the
+                # query but since we made token detection not run it is no longer used, but is
+                # written. This will probably change in the future again. Related issue:
+                # https://github.com/rotki/rotki/issues/5252
                 last_queried_ts = deserialize_timestamp(value)
             else:  # should be ACCOUNTS_DETAILS_TOKENS
                 try:
@@ -1361,10 +1365,10 @@ class DBHandler:
         )
         # Delete previous entries for tokens
         write_cursor.execute(
-            'DELETE FROM accounts_details WHERE account=? AND blockchain=? AND KEY=?',
-            (address, blockchain.serialize(), ACCOUNTS_DETAILS_TOKENS),
+            'DELETE FROM accounts_details WHERE account=? AND blockchain=? AND KEY IN(?, ?)',
+            (address, blockchain.serialize(), ACCOUNTS_DETAILS_TOKENS, ACCOUNTS_DETAILS_LAST_QUERIED_TS),  # noqa: E501
         )
-        # Timestamp will get replaced
+        # Insert new values
         write_cursor.executemany(
             'INSERT OR REPLACE INTO accounts_details '
             '(account, blockchain, key, value) VALUES (?, ?, ?, ?)',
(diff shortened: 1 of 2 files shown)


rotkehlchen/db/dbhandler.py
```diff
@@ -1247,6 +1247,10 @@ def get_tokens_for_address(
         returned_list = []
         for (key, value) in cursor:
             if key == EVM_ACCOUNTS_DETAILS_LAST_QUERIED_TS:
+                # At the moment last_queried_timestamp is not used. It used to be a cache for the
+                # query but since we made token detection not run it is no longer is used, but is
```
LefterisJP, line 1251:
```suggestion
                # query but since we made token detection not run it is no longer used, but is
```

rotkehlchen/tests/unit/test_tokens.py
```diff
@@ -73,3 +74,55 @@ def test_generate_chunks():
...
     ]
     assert generated_chunks == expected_chunks
+
+
+def test_last_queried_ts(tokens):
+    """
+    Checks that after detecting evm tokens last_queried_timestamp is updated and there
+    are no duplicates.
+    """
+    # We don't need to query chain here, so mock tokens list
```
LefterisJP, line 84:
```suggestion
    # We don't need to query the chain here, so mock tokens list
```

rotkehlchen/tests/unit/test_tokens.py
```diff
@@ -73,3 +74,55 @@ def test_generate_chunks():
...
+        tokens.db.conn.execute(
+            'UPDATE evm_accounts_details SET value=? WHERE '
+            'account=? AND chain_id=1 AND key=?',
+            (123456, address, 'last_queried_timestamp'),
+        )
+        # Make sure that it was modified
+        after_updating = tokens.db.conn.execute(
+            'SELECT key, value FROM evm_accounts_details',
+        ).fetchall()
+        assert int(after_updating[0][1]) == 123456
```
LefterisJP, line 116: Don't do this. This is really bad test code. Just move the time with pytest freezegun, a module we already use.
```suggestion
freezer.move_to('Some time in the future')
```

LefterisJP (commented):
Shouldn't this be in bugfixes?

What problem does it solve other than not writing some extra entries (which is indeed a problem ofc -- but your discord message made it sound like a bug)? If it stays here it should also have an entry in the DB upgrade to delete all old token detection timestamps AND a test for it.

If it moves to bugfixes, then the db upgrade code should still be done in develop.

rotkehlchen/db/dbhandler.py
```diff
@@ -1361,10 +1365,10 @@ def save_tokens_for_address(
         )
         # Delete previous entries for tokens
         write_cursor.execute(
-            'DELETE FROM accounts_details WHERE account=? AND blockchain=? AND KEY=?',
-            (address, blockchain.serialize(), ACCOUNTS_DETAILS_TOKENS),
+            'DELETE FROM accounts_details WHERE account=? AND KEY IN(?, ?)',
```
LefterisJP, line 1368: why did you delete the blockchain key here?

LefterisJP (commented):
Question and asking you to force push since no tests ran

rotkehlchen/db/dbhandler.py
nebolax: Ahhh, my bad, probably this happened when I was switching from develop to bugfixes (since in develop `chain_id` is used)

LefterisJP (commented):
lgtm

LefterisJP:
I have to remember to add to the DB upgrade for this.

PR #5261, opened 2022-12-14
Add optimism blockchain balance query

LefterisJP:
Also better abstract querying of blockchain balances


LefterisJP (commented):
all green. Merging.

PR #5262, opened 2022-12-14
Add grouping of multichain assets to the api balances and responses

yabirgb:
This PR addresses the problem of presenting to the frontend the information about assets in multiple chains.

For the balances it follows one schema like this one

```python
{
        'assets': {
            'USD': {'amount': '2', 'usd_value': '2'},
            'ETH': {'amount': '3', 'usd_value': '900'},
            'eip155:1/erc20:0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48': {
                'amount': '5',
                'usd_value': '5',
                'breakdown': {
                    'eip155:1/erc20:0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48': {
                        'amount': '2',
                        'usd_value': '2'
                    },
                    'eip155:10/erc20:0x7F5c764cBc14f9669B88837ca1490cCa17c31607': {
                        'amount': '3',
                        'usd_value': '3',
                    }
                },
            }
        },
        'liabilities': {
            ethaddress_to_identifier('0x6B175474E89094C44Da98b954EedeAC495271d0F'): {'amount': '5', 'usd_value': '5.1'},
            'ETH': {'amount': '0.5', 'usd_value': '150'},
        },
    }
```

and for the `/assets/mappings` we include now if the asset is part of a collection and then provide the collection details. Check the docs for a detailed example

**This PR includes a new update in the globaldb adding the collections**. The file with the groups was created using this script https://gist.github.com/yabirgb/258f29c33c59c88fc82cc8fddbb09bdd


rotkehlchen/accounting/structures/balance.py
```diff
@@ -18,6 +20,67 @@ class BalanceType(DBEnumMixIn):
...
+            # for each asset
+            collection = assets_to_collection[asset.identifier]
+            total_collections_balance[collection] += balance
+            # we need the chain id to later sort and know what assets in the set will be
+            # used to represent the others in the collection.
+            crypto_asset = asset.resolve_to_crypto_asset()
+            if isinstance(crypto_asset, EvmToken) is True:
+                sorting_key = crypto_asset.chain_id  # type: ignore  # noqa: E501 mypy fails to see that this could be EvmToken
+            else:
+                sorting_key = 0
```
yabirgb, line 67:
```suggestion
            else:
                # we need to do this to avoid issues in the case of collections for non evm tokens
                sorting_key = 0
```

rotkehlchen/tests/api/test_blockchain.py
```diff
@@ -219,7 +219,7 @@ def test_query_blockchain_balances(
             )
         else:
             outcome = assert_proper_response_with_result(response)
-
+    assert False
```
yabirgb, line 222: wtf dude

yabirgb (commented):
You left an empty file. Seriously take care when pushing

docs/api.rst
```diff
@@ -2782,69 +2782,129 @@ Querying onchain balances
...
+                  "ETH":{
+                    "amount":"10",
+                    "usd_value":"1650.53"
                   },
-                  "liabilities": {
-                      "eip155:1/erc20:0x6B175474E89094C44Da98b954EedeAC495271d0F": {"amount": "20", "usd_value": "20.35"}
+                  "eip155:1/erc20:0x6B175474E89094C44Da98b954EedeAC495271d0F":{
+                    "amount":"15",
+                    "usd_value":"15.21",
+                    "breakdown":{
```
nebolax, line 2839: Question. Why a breakdown is needed if user's owned asset belongs only to one chain?

rotkehlchen/globaldb/handler.py
```diff
@@ -430,8 +446,14 @@ def get_assets_mappings(identifiers: list[str]) -> dict[str, dict]:
                     result[entry[0]].update({'evm_chain': ChainID.deserialize_from_db(entry[3]).to_name()})  # noqa: E501
                 if entry[5] is not None:
                     result[entry[0]].update({'custom_asset_type': entry[5]})
-
-        return result
+                if entry[6] is not None:
+                    result[entry[0]].update({'collection_id': str(entry[6])})
```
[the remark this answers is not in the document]
yabirgb: Because numbers can't be keys to json

rotkehlchen/api/rest.py
```diff
@@ -1396,13 +1396,17 @@ def query_list_of_all_assets(
 
     def get_assets_mappings(self, identifiers: list[str]) -> Response:
         try:
-            asset_mappings = GlobalDBHandler().get_assets_mappings(identifiers)
+            asset_mappings, asset_collections = GlobalDBHandler().get_assets_mappings(identifiers)
```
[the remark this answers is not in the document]
yabirgb: To avoid duplication in the information and so the frontend can group them later. So for each asset that forms a group we give their group id and then the groups names/symbols

yabirgb:
@nebolax thanks! We have changed a bit the approach and after your comment we have moved the grouping from the backend to the frontend to avoid the duplications. We only are adding the collections to the database here and the assets groups in the reponse for the assets

nebolax (commented):
Just a few comments

rotkehlchen/globaldb/handler.py
```diff
@@ -430,8 +447,14 @@ def get_assets_mappings(identifiers: list[str]) -> dict[str, dict]:
...
+                    if entry[6] not in asset_collections:
+                        asset_collections[str(entry[6])] = {
```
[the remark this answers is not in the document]
nebolax: Discussed in discord. All good here. My fault -- misread the code

rotkehlchen/globaldb/handler.py
```diff
@@ -1908,3 +1931,17 @@ def get_asset_type(identifier: str, use_packaged_db: bool = False) -> AssetType:
             raise UnknownAsset(identifier)
 
         return AssetType.deserialize_from_db(type_in_db[0])
+
+    @staticmethod
+    def get_assets_collections(asset_ids: list[str]) -> dict[str, int]:
```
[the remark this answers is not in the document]
yabirgb: True is not used. The usage was removed after updating the logic. Thansk

docs/api.rst
```diff
@@ -2782,69 +2782,117 @@ Querying onchain balances
...
-                  }},
+        "result":{
+          "per_account":{
+            "BTC":{
+              "standalone":{
+                "3Kb9QPcTUJKspzjQFBppfXRcWew6hyDAPb":{
+                  "amount":"0.5",
+                  "usd_value":"3770.075"
+                },
+                "33hjmoU9XjEz8aLxf44FNGB8TdrLkAVBBo":{
```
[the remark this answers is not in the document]
yabirgb: I updated them because fixed the indentation and other stuff when changing the structure before but will undo that change

docs/api.rst
```diff
@@ -3114,35 +3162,44 @@ Get asset identifiers mappings
...
-              "eip155:1/erc20:0xB6eD7644C69416d67B522e20bC294A9a9B405B31": {
-                  "name": "0xBitcoin",
-                  "symbol": "0xBTC",
-                  "asset_type": "evm token"
+        "result":{
+            "assets":{
+              "eip155:1/erc20:0xB6eD7644C69416d67B522e20bC294A9a9B405B31":{
+                  "name":"0xBitcoin",
+                  "symbol":"0xBTC",
+                  "asset_type":"evm token"
```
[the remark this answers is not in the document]
yabirgb: yes ups, seems that when formatting I delete it. Thanks

rotkehlchen/globaldb/handler.py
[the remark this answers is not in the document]
yabirgb: It was a misunderstanding about how the collection worked

docs/api.rst
```diff
@@ -3114,35 +3114,45 @@ Get asset identifiers mappings
...
 
       {
-          "result": {
-              "eip155:1/erc20:0xB6eD7644C69416d67B522e20bC294A9a9B405B31": {
-                  "name": "0xBitcoin",
-                  "symbol": "0xBTC",
-                  "asset_type": "evm token"
+        "result":{
+            "assets":{
+              "eip155:1/erc20:0xB6eD7644C69416d67B522e20bC294A9a9B405B31":{
```
LefterisJP, line 3119: why remove the space between `:` and `{` ? This is how json is styled in this file . Same in the rest of the diff

docs/api.rst
```diff
@@ -3114,35 +3114,45 @@ Get asset identifiers mappings
...
-              "eip155:1/erc20:0xcC4eF9EEAF656aC1a2Ab886743E98e97E090ed38": {
-                  "name": "DigitalDevelopersFund",
-                  "symbol": "DDF",
-                  "chain_id": 1,
-                  "is_custom_asset": false,
-		  "asset_type": "evm token"
+              "eip155:1/erc20:0xcC4eF9EEAF656aC1a2Ab886743E98e97E090ed38":{
+                  "name":"DigitalDevelopersFund",
+                  "symbol":"DDF",
+                  "chain_id":1,
```
LefterisJP, line 3133: why did you remove the styling here? Please undo and make it like the rest of the file with a space between `:` and the other stuff. In all of the json

rotkehlchen/data/assets_collections.sql
```diff
@@ -0,0 +1,726 @@
+INSERT INTO asset_collections(id, name, symbol) VALUES
```
LefterisJP, line 1: shouldn't we do both here and below `OR IGNORE` just to be safe?

rotkehlchen/tests/unit/test_globaldb_upgrades.py
```diff
@@ -29,19 +29,48 @@
 ASSETS_IN_V2_GLOBALDB = 3095
 
 
-def _count_v2_v3_assets_inserted() -> int:
-    """Counts and returns how many assets are to be inserted by globaldb_v2_v3_assets.sql"""
-    assets_inserted_by_update = 0
+def _count_sql_file_sentences(file_name: str, skip_tables: int = 0):
```
LefterisJP, line 32: shouln't it be skip_statements? As there is no guarantee two different statements don't touch the same table?

rotkehlchen/tests/unit/test_globaldb_upgrades.py
```diff
@@ -29,19 +29,48 @@
...
             if ';' in line:
                 break
 
-    return assets_inserted_by_update - 1
+    return insertions_made - 1
+
+
+def _count_v2_v3_assets_inserted() -> int:
+    """Counts and returns how many assets are to be inserted by globaldb_v2_v3_assets.sql"""
+    return _count_sql_file_sentences('globaldb_v2_v3_assets.sql')
```
LefterisJP, line 63: Why make all those different functions with just 1 other function call in the body? Keep it simple and just use the `count_sql_file_sententces` where you need. These functions are not needed

rotkehlchen/data/assets_collections.sql
yabirgb: okey

rotkehlchen/tests/unit/test_globaldb_upgrades.py
yabirgb: Yes. I had one table per statmenent but makes more sense to rename it

rotkehlchen/tests/unit/test_globaldb_upgrades.py
yabirgb: It was to make it clearer but will change it and just call `_count_sql_file_sentences`

rotkehlchen/tests/unit/test_globaldb_upgrades.py
```diff
@@ -29,19 +29,33 @@
 ASSETS_IN_V2_GLOBALDB = 3095
 
 
-def _count_v2_v3_assets_inserted() -> int:
-    """Counts and returns how many assets are to be inserted by globaldb_v2_v3_assets.sql"""
-    assets_inserted_by_update = 0
+def _count_sql_file_sentences(file_name: str, skip_statements: int = 0):
+    """
+    Count the sql lines in scripts used during upgrades. If the skip_tables argument is provided
```
LefterisJP, line 34: Yabir my man I told you to rename it. That means also:

1. in the docstring
2. Inside the function. `skipped_tables` should be `skipped_statements`

PR #5266, opened 2022-12-15
Fix accounting without api keys

### LINKED ISSUE #5263: Don't ignore exchanges in PnL if API keys are removed
## Problem Definition

If a user removes exchange API keys from the app then trades, ledger actions, etc. from that exchange will no longer be used in PnL reports even though they are still stored in the database. That is a problem and should be fixed. The same problem does not appear for the trades view, as [query_trades](https://github.com/rotki/rotki/blob/c974019abddb4bf01d8742658c45d8fd8c6798bc/rotkehlchen/history/events.py#L135) does not check active exchanges but simply gets paginated trades in the DB.

The way to implement this would probably be by doing something similar like the trades query in the part of gathering history events for the PnL report.

## Task

- [ ] Make it so that trades from an exchange that is not tracked (no key in the app, so not part of the exchange manager) are taken into account in the PnL report.
- [ ] Write a test for it.

nebolax:
Closes #5263

### DIFF
--- a/rotkehlchen/exchanges/exchange.py
+++ b/rotkehlchen/exchanges/exchange.py
@@ -42,6 +42,7 @@ ExchangeHistorySuccessCallback = Callable[
 ]
 
 ExchangeHistoryFailCallback = Callable[[str], None]
+ExchangeHistoryNewStepCallback = Callable[[str], None]
 
 
 class ExchangeInterface(CacheableMixIn, LockableQueryMixIn):
@@ -477,45 +478,48 @@ class ExchangeInterface(CacheableMixIn, LockableQueryMixIn):
             self,
             start_ts: Timestamp,
             end_ts: Timestamp,
-            success_callback: ExchangeHistorySuccessCallback,
             fail_callback: ExchangeHistoryFailCallback,
+            new_step_data: Optional[Tuple[ExchangeHistoryNewStepCallback, str]] = None,
     ) -> None:
         """Queries the historical event endpoints for this exchange and performs actions.
-
-        In case of success passes the result to success_callback.
+        The results are saved in the DB.
         In case of failure passes the error to failure_callback
+
+        `new_step_data` argument contains callback and exchange name to be used for steps.
         """
+        if new_step_data is not None:
+            new_step_callback, exchange_name = new_step_data
+            new_step_callback(f'Querying {exchange_name} trades history')
         try:
-            trades_history = self.query_trade_history(
+            self.query_trade_history(
                 start_ts=start_ts,
                 end_ts=end_ts,
                 only_cache=False,
             )
-            margin_history = self.query_margin_history(
+            if new_step_data is not None:
+                new_step_callback(f'Querying {exchange_name} margin history')
+            self.query_margin_history(
                 start_ts=start_ts,
                 end_ts=end_ts,
             )
-            asset_movements = self.query_deposits_withdrawals(
+            if new_step_data is not None:
+                new_step_callback(f'Querying {exchange_name} asset movements history')
+            self.query_deposits_withdrawals(
                 start_ts=start_ts,
                 end_ts=end_ts,
                 only_cache=False,
             )
-            # Query (and save in the DB) any ledger actions. They will be included in history l8er
+            if new_step_data is not None:
+                new_step_callback(f'Querying {exchange_name} ledger actions history')
             self.query_income_loss_expense(
                 start_ts=start_ts,
                 end_ts=end_ts,
                 only_cache=False,
             )
-            exchange_specific_data = self.query_exchange_specific_history(  # pylint: disable=assignment-from-none  # noqa: E501
+            # No new step for exchange_specific_history since it is not used in any exchange atm.
+            self.query_exchange_specific_history(
                 start_ts=start_ts,
                 end_ts=end_ts,
             )
-            success_callback(
-                trades_history,
-                margin_history,
-                asset_movements,
-                exchange_specific_data,
-            )
-
         except RemoteError as e:
             fail_callback(str(e))
--- a/rotkehlchen/history/events.py
+++ b/rotkehlchen/history/events.py
@@ -1,6 +1,6 @@
 import logging
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, List, Tuple
+from typing import TYPE_CHECKING, List, Tuple
 
 from rotkehlchen.accounting.structures.base import HistoryBaseEntry
 from rotkehlchen.constants.misc import ZERO
@@ -14,11 +14,11 @@ from rotkehlchen.db.filtering import (
 from rotkehlchen.db.history_events import DBHistoryEvents
 from rotkehlchen.db.ledger_actions import DBLedgerActions
 from rotkehlchen.errors.misc import RemoteError
-from rotkehlchen.exchanges.data_structures import AssetMovement, MarginPosition, Trade
+from rotkehlchen.exchanges.data_structures import AssetMovement, Trade
 from rotkehlchen.exchanges.manager import SUPPORTED_EXCHANGES, ExchangeManager
 from rotkehlchen.fval import FVal
 from rotkehlchen.logging import RotkehlchenLogsAdapter
-from rotkehlchen.types import EXTERNAL_LOCATION, Location, Timestamp
+from rotkehlchen.types import Location, Timestamp
 from rotkehlchen.user_messages import MessagesAggregator
 from rotkehlchen.utils.misc import timestamp_to_date
 
@@ -37,12 +37,12 @@ log = RotkehlchenLogsAdapter(logger)
 # eth transactions
 # eth receipts
 # eth tx decoding
+# reading from the db trades, asset movements, margin positions
 # ledger actions
-# external location trades -> len(EXTERNAL_LOCATION)
 # eth2
 # base history entries
 # Please, update this number each time a history query step is either added or removed
-NUM_HISTORY_QUERY_STEPS_EXCL_EXCHANGES = 6 + len(EXTERNAL_LOCATION)
+NUM_HISTORY_QUERY_STEPS_EXCL_EXCHANGES = 7
 
 
 class EventsHistorian:
@@ -290,7 +290,10 @@ class EventsHistorian:
         """
         self._reset_variables()
         step = 0
-        total_steps = len(self.exchange_manager.connected_exchanges) + NUM_HISTORY_QUERY_STEPS_EXCL_EXCHANGES  # noqa: E501
+        total_steps = (
+            len(self.exchange_manager.connected_exchanges) * 4 +  # 4 substeps in each exchange
+            NUM_HISTORY_QUERY_STEPS_EXCL_EXCHANGES
+        )
         log.info(
             'Get/create trade history',
             start_ts=start_ts,
@@ -300,27 +303,16 @@ class EventsHistorian:
         history: List['AccountingEventMixin'] = []
         empty_or_error = ''
 
-        def populate_history_cb(
-                trades_history: List[Trade],
-                margin_history: List[MarginPosition],
-                result_asset_movements: List[AssetMovement],
-                exchange_specific_data: Any,
-        ) -> None:
-            """This callback will run for succesfull exchange history query
-
-            We don't include ledger actions here since we simply gather all of them at the end
-            """
-            history.extend(trades_history)
-            history.extend(margin_history)
-            history.extend(result_asset_movements)
-
-            if exchange_specific_data:
-                pass  # this used to be only for polo loans -- removed now. TODO: Think if needed
-
         def fail_history_cb(error_msg: str) -> None:
             """This callback will run for failure in exchange history query"""
             nonlocal empty_or_error
             empty_or_error += '\n' + error_msg
+
+        def new_step_cb(state_name: str) -> None:
+            """This callback will run for each new step in exchange history query"""
+            nonlocal step
+            step = self._increase_progress(step, total_steps)
+            self.processing_state_name = state_name
 
         for exchange in self.exchange_manager.iterate_exchanges():
             self.processing_state_name = f'Querying {exchange.name} exchange history'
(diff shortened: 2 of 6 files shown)


rotkehlchen/tests/unit/accounting/test_exchanges.py
```diff
@@ -38,3 +42,36 @@ def test_account_for_coinbase_ledger_actions(rotkehlchen_api_server_with_exchang
...
+@pytest.mark.parametrize('ethereum_accounts', [[]])
+def test_exchanges_removed_api_keys(rotkehlchen_api_server):
+    """
+    Test that if actions made on an exchange are stored in the DB but the API keys were removed
+    then the actions are still taken into account in accounting.
+    """
+    rotki = rotkehlchen_api_server.rest_api.rotkehlchen
+    with rotki.data.db.user_write() as write_cursor:
+        rotki.data.db.add_trades(write_cursor, trades=[Trade(
+            timestamp=Timestamp(1611426201),
```
yabirgb, line 57: Can you add also an asset movement? And add to the database a manual trade that doesn't belong to any exchange. location external

rotkehlchen/history/events.py
```diff
@@ -365,17 +347,34 @@ def fail_history_cb(error_msg: str) -> None:
...
+                filter_query=AssetMovementsFilterQuery.make(),
+                has_premium=True,  # we need all trades for accounting -- limit happens later
+            )
+        history.extend(asset_movements)
+        step = self._increase_progress(step, total_steps)
+
+        # Include all margin positions
+        self.processing_state_name = 'Querying trades history'
+        with self.db.conn.read_ctx() as cursor:
+            margin_positions = self.db.get_margin_positions(cursor)
```
yabirgb, line 375: Can't you just put everything under the same with statement. I believe is faster than opening and closing the cursor each time

yabirgb (commented):
A few coments and there is also a test that needs love

rotkehlchen/history/events.py
```diff
@@ -365,16 +349,31 @@ def fail_history_cb(error_msg: str) -> None:
...
+            asset_movements = self.db.get_asset_movements(
+                cursor,
+                filter_query=AssetMovementsFilterQuery.make(),
+                has_premium=True,  # we need all trades for accounting -- limit happens later
+            )
+            history.extend(asset_movements)
+            step = self._increase_progress(step, total_steps)
+
+            # Include all margin positions
+            self.processing_state_name = 'Querying trades history'
```
yabirgb, line 374: This text is duplicated

rotkehlchen/history/events.py
```diff
@@ -365,16 +349,31 @@ def fail_history_cb(error_msg: str) -> None:
...
-                external_trades = self.db.get_trades(
-                    cursor,
-                    filter_query=TradesFilterQuery.make(location=location),
-                    has_premium=True,  # we need all trades for accounting -- limit happens later
-                )
-            history.extend(external_trades)
+        # Include all trades
+        self.processing_state_name = 'Querying trades history'
+        with self.db.conn.read_ctx() as cursor:
+            trades = self.db.get_trades(
```
yabirgb, line 355: I believe there is no reason to have them here. Can you move this db queries just after querying the exchanges? I found it confusing since they are related things but they were distant in the code. Also this queries now query both excchange trades and external locations trades, can you write that in one comment?

rotkehlchen/tests/api/test_history.py
```diff
@@ -128,7 +128,7 @@ def test_query_history(rotkehlchen_api_server_with_exchanges, start_ts, end_ts):
         ),
     )
     result = assert_proper_response_with_result(response=response, status_code=HTTPStatus.OK)
-    assert len(result['missing_acquisitions']) == 10
+    assert len(result['missing_acquisitions']) == 9
```
yabirgb, line 131: why is there one less?

nebolax: The problem is that in the mocks that we have here there are many events that occur on the same timestamp (`1458994442`). Before this PR, events were added in batches per-exchange (e.g. queried events for exchange №1 -> added these events, then queried events for exchange №2 -> added these events).

With this pr's changes, now all events are added all at once and since we order only by timestamp, the order is a bit different.

yabirgb: Mmmmmm but then I guess this is a bug we had before somehow? Is not exactly a bug but uncovers an issue on that logic :thinking:

rotkehlchen/exchanges/exchange.py
```diff
@@ -473,11 +473,10 @@ def query_income_loss_expense(
...
 
-    def query_history_with_callbacks(
+    def query_history_with_fail_callback(
             self,
             start_ts: Timestamp,
             end_ts: Timestamp,
-            success_callback: ExchangeHistorySuccessCallback,
             fail_callback: ExchangeHistoryFailCallback,
     ) -> None:
         """Queries the historical event endpoints for this exchange and performs actions.
```
LefterisJP, line 482: Edit the docstring to specify that any results of querying are saved in the DB.
Also remove the success callback mention since you removed it.

rotkehlchen/history/events.py
```diff
@@ -37,12 +37,14 @@
...
+# trades
+# asset movements
+# margin positions
 # ledger actions
-# external location trades -> len(EXTERNAL_LOCATION)
 # eth2
 # base history entries
 # Please, update this number each time a history query step is either added or removed
-NUM_HISTORY_QUERY_STEPS_EXCL_EXCHANGES = 6 + len(EXTERNAL_LOCATION)
+NUM_HISTORY_QUERY_STEPS_EXCL_EXCHANGES = 9
```
LefterisJP, line 47: this is quite dirty. We probably should think of a way to make this better, not have hard coded steps like this and comments. **Not an action item** for this PR since this has been there for a while. Just mentioning it.

rotkehlchen/history/events.py
```diff
@@ -300,39 +302,50 @@ def get_history(
...
                 # We need to have history of exchanges since before the range
                 start_ts=Timestamp(0),
                 end_ts=end_ts,
-                success_callback=populate_history_cb,
                 fail_callback=fail_history_cb,
             )
             step = self._increase_progress(step, total_steps)
 
+        # Query all trades, asset movements and margin positions from the DB.
+        # Query all of them for all locations (both exchanges and other locations).
```
LefterisJP, line 321: simplify

```suggestion
        # Query all trades, asset movements and margin positions from the DB for all possible locations
```

rotkehlchen/history/events.py
```diff
@@ -300,39 +302,50 @@ def get_history(
...
+        with self.db.conn.read_ctx() as cursor:
+            # Include all trades
```
LefterisJP, line 323: few comments here which count for all steps under the cursor:

1. Since you have a string name of what you do there is no need for a comment in each step.
2. The processing state name in each step is not correct. The time you query from the DB is not the step time. Reading from the DB is near instant. The steps are to appear in the PnL report view. So this step should appear only for when querying the actual trade history from remotes.
3. Same for the increase step

rotkehlchen/tests/api/test_history.py
LefterisJP: order being a bit different is one thing, but missing an event is another. How is an event missing? Something else is off here and should be looked at.

Or you mean that the order of events in processing changed so somehow there is one less missing acquisition. If that is so then I would need you to show me exactly which event lead to this otherwise I would be afraid there is something more sinister underneath.

I guess you already spent time doing that so can you point me to that event whose order change did it? Perhaps easiest way to do is run a certain test with debugger on for both this PR and bugixes/develop.

rotkehlchen/tests/unit/accounting/test_exchanges.py
```diff
@@ -38,3 +42,75 @@ def test_account_for_coinbase_ledger_actions(rotkehlchen_api_server_with_exchang
     assert events_map[AccountingEventType.FEE] == 2
     assert events_map[AccountingEventType.ASSET_MOVEMENT] == 3
     assert sum(events_map.values()) == 11  # processed events
+
+
+@pytest.mark.parametrize('default_mock_price_value', [FVal(1.5)])
+@pytest.mark.parametrize('ethereum_accounts', [[]])
+def test_exchanges_removed_api_keys(rotkehlchen_api_server):
+    """
+    Test that if actions made on an exchange are stored in the DB but the API keys were removed
```
LefterisJP, line 51: This is not doing what it says. You should add the exchange, add the trades and then remove it. I know the result is gonna be the same but this is what you say you are testing, but this is not what the test is doing which is a problem.

rotkehlchen/tests/unit/accounting/test_exchanges.py
```diff
@@ -38,3 +42,75 @@ def test_account_for_coinbase_ledger_actions(rotkehlchen_api_server_with_exchang
...
+            location=Location.COINBASE,
+            category=AssetMovementCategory.DEPOSIT,
+            asset=A_BTC,
+            amount=ONE,
+            fee_asset=A_BTC,
+            fee=FVal(0.00001),
+            address=None,
+            transaction_id=None,
+            link='no link',
+        )])
```
LefterisJP, line 84: also add a ledger action and a margin position to cover the entire array of events that an exchange can add as of this version of rotki

LefterisJP (commented):
few comments

rotkehlchen/tests/api/test_history.py
nebolax: [initial-events-old.txt](https://github.com/rotki/rotki/files/10277609/initial-events-old.txt)
[initial-events-new.txt](https://github.com/rotki/rotki/files/10277610/initial-events-new.txt)
[processed-events-old.txt](https://github.com/rotki/rotki/files/10277672/processed-events-old.txt)
[processed-events-new.txt](https://github.com/rotki/rotki/files/10277613/processed-events-new.txt)

I attached 4 json files (they have .txt extension here since github doesn't support json files).

`initial-` files - events before accounting
`processed-` files - processed events

`-old` files - before this pr changes
`-new` files - after this pr changes

In `initial-events-old` you can see these 2 events in one order:
```json
{
        "identifier": "f0bdeb714921a9d8dd14f6ce389c710ad4564fc8c0e94b2c12ba57f87719ad2b",
        "timestamp": 1458994442,
        "location": "poloniex",
        "category": "withdrawal",
        "address": "131rdg5Rzn6BFufnnQaHhVa5ZtRU1J2EZR",
        "transaction_id": "2d27ae26fa9c70d6709e27ac94d4ce2fde19b3986926e9f3bfcf3e2d68354ec5",
        "asset": "BTC",
        "amount": "5.0",
        "fee_asset": "BTC",
        "fee": "0.5",
        "link": "1"
    },
    {
        "timestamp": 1458994442,
        "location": "kraken",
        "base_asset": "BTC",
        "quote_asset": "EUR",
        "trade_type": "buy",
        "amount": "1",
        "rate": "100",
        "fee": "0.1",
        "fee_currency": "EUR",
        "link": "11458994442",
        "notes": null
    },
```

while in `initial-events-new` the order is reversed:
```json
{
        "timestamp": 1458994442,
        "location": "kraken",
        "base_asset": "BTC",
        "quote_asset": "EUR",
        "trade_type": "buy",
        "amount": "1",
        "rate": "100",
        "fee": "0.1",
        "fee_currency": "EUR",
        "link": "11458994442",
        "notes": null
    },
    {
        "identifier": "f0bdeb714921a9d8dd14f6ce389c710ad4564fc8c0e94b2c12ba57f87719ad2b",
        "timestamp": 1458994442,
        "location": "poloniex",
        "category": "withdrawal",
        "address": "131rdg5Rzn6BFufnnQaHhVa5ZtRU1J2EZR",
        "transaction_id": "2d27ae26fa9c70d6709e27ac94d4ce2fde19b3986926e9f3bfcf3e2d68354ec5",
        "asset": "BTC",
        "amount": "5.0",
        "fee_asset": "BTC",
        "fee": "0.5",
        "link": "1"
    },
```

As the result of this in `processed-events-old` poloniex withdrawal event has no matched acquisitions
```json
{
            "type": "asset movement",
            "notes": "poloniex withdrawal",
            "location": "poloniex",
            "timestamp": 1458994442,
            "asset": "BTC",
            "free_amount": "0.0",
            "taxable_amount": "0.5",
            "price": "373.295",
            "pnl_taxable": "0",
            "pnl_free": "0",
            "cost_basis": {
                "is_complete": false,
                "matched_acquisitions": []
            }
        },
```

while in `processed-events-new` it does:
```json
{
            "type": "asset movement",
            "notes": "poloniex withdrawal",
            "location": "poloniex",
            "timestamp": 1458994442,
            "asset": "BTC",
            "free_amount": "0.0",
            "taxable_amount": "0.5",
            "price": "373.295",
            "pnl_taxable": "0",
            "pnl_free": "0",
            "cost_basis": {
                "is_complete": true,
                "matched_acquisitions": [
                    {
                        "amount": "0.5",
                        "event": {
                            "timestamp": 1458994442,
                            "full_amount": "1",
                            "rate": "100.1",
                            "index": 10
                        },
                        "taxable": true
                    }
                ]
            }
```

rotkehlchen/exchanges/exchange.py
```diff
@@ -473,49 +473,39 @@ def query_income_loss_expense(
...
                 only_cache=False,
             )
-            # Query (and save in the DB) any ledger actions. They will be included in history l8er
             self.query_income_loss_expense(
                 start_ts=start_ts,
                 end_ts=end_ts,
                 only_cache=False,
             )
-            exchange_specific_data = self.query_exchange_specific_history(  # pylint: disable=assignment-from-none  # noqa: E501
+            self.query_exchange_specific_history(  # pylint: disable=assignment-from-none
```
LefterisJP, line 506: you removed the assignment so I think that the pylint disable comment is no longer needed here

rotkehlchen/history/events.py
LefterisJP: Why did you remove the steps? I told you where they should go in my last comment.

> the steps are to appear in the PnL report view. So this step should appear only for when querying the actual trade history from remotes.

So you should probably add as many steps as are required in the callback.

rotkehlchen/tests/api/test_history.py
LefterisJP: Okay that makes sense thank you.

LefterisJP (commented):
Few more comments. 2 minor 1 needing some work

rotkehlchen/history/events.py
nebolax: I am not sure that I get you. Which callback do you mean?

> So this step should appear only for when querying the actual trade history from remotes.

There already are steps when querying exchanges. What other location do you mean?

rotkehlchen/history/events.py
```diff
@@ -300,39 +303,55 @@ def get_history(
...
         for exchange in self.exchange_manager.iterate_exchanges():
             self.processing_state_name = f'Querying {exchange.name} exchange history'
-            exchange.query_history_with_callbacks(
+            exchange.query_history_with_fail_callback(
                 # We need to have history of exchanges since before the range
                 start_ts=Timestamp(0),
                 end_ts=end_ts,
-                success_callback=populate_history_cb,
                 fail_callback=fail_history_cb,
+                new_step_callback=new_step_cb,
```
LefterisJP, line 324: Hm ... you always pass it. Why did you make it optional and added all these other if checks in the function? Wtf? Just forget what I told you in that part then. Just pass both as non optional and remove those unneeded checks.

rotkehlchen/exchanges/exchange.py
```diff
@@ -473,49 +474,52 @@ def query_income_loss_expense(
 
         return ledger_actions
 
-    def query_history_with_callbacks(
+    def query_history_with_fail_callback(
```
LefterisJP, line 477: revert to the pevious name. After the new changes the new name makes no sense

rotkehlchen/exchanges/exchange.py
```diff
@@ -477,45 +478,49 @@ def query_history_with_callbacks(
...
         """Queries the historical event endpoints for this exchange and performs actions.
-
-        In case of success passes the result to success_callback.
+        The results are saved in the DB.
         In case of failure passes the error to failure_callback
+
+        `new_step_data` argument contains callback and exchange name to be used for steps.
         """
+        if new_step_data is not None:
+            new_step_callback, exchange_name = new_step_data
```
LefterisJP, line 491:
```suggestion
        if new_step_data is not None:
            new_step_callback, exchange_name = new_step_data
            new_step_callback(f'Querying {exchange_name} trades history')
```

And delete it from below. No need to repeat the if check like this

rotkehlchen/history/events.py
LefterisJP: ah fair enough. Did not see this since it was hidden in my review view as unchanged since last view.

LefterisJP (approved):
LGTM. Am approving though frontend failed for some reason. Backend failure is a single covalent rate limit

PR #5267, opened 2022-12-16
Edit packaged contracts and add them to global DB. Also include in upgrade

LefterisJP:


rotkehlchen/tests/fixtures/accounting.py
```diff
@@ -360,10 +364,11 @@ def fixture_inquirer(
         mocked_current_prices_with_oracles,
         current_price_oracles_order,
         ignore_mocked_prices_for,
+        ethereum_manager,
```
LefterisJP, line 367: This change makes many tests that use inquirer fail, since ethereum manager uses the, the ethereum inquirer which uses the database fixture. And as such the database is already initialized with the test_user and it fails when trying to initialize a new user DB from rotkehlchen instance fixture.

yabirgb (approved):
LGTM, thanks sir

PR #5268, opened 2022-12-16
Adding optimism account, querying balances, detecting tokens works

LefterisJP:


LefterisJP (commented):
All green. Quite a few changes here, but also moving stuff around causes the line diff.

With this merged, the frontend should be able to add optimism accounts and query optimism balances.

PR #5269, opened 2022-12-18
Start of pytest tests speedup, start being more careful with networking

LefterisJP:
This PR introduces a few things and starts changing some tests to speed our test suite up by aiming to reduce network queries. Read the individual commits to see more details but what this does in a TL;DR is:

- Add pytest-socket.  By adding the `--disable-socket` cli argument in a test run there will be failure when there is an attempt to connect to the network. This helps figure out if there is any network calls in a test run and if yes where they are so they can be mocked.
- Add `--no-network-mocking` cli argument for pytest. For tests that are aware of their networking calls (tests we need to touch, such as the ones in this PR), this skips mocking and uses the actual network. This ends up being a fixture variable `network_mocking` which can be read in the tests.
- Since not all tests are properly made to mock and be aware of it, and since most networks calls are web3 related we add the parametrizable fixture `should_mock_web3`. Where this is parametrized as true all web3 stuff (web3 rpc, etherscan, other related apis are mocked. The reason this is different from the `network_mocking` is that we need to go step by step for each test and make it mockable and mock aware. So we just add this to all tests that are properly mocked. Once this has covered the entire suite we can fall back to `network_mocking` fixture and `should_mock_web3` can go away.
- Mocking web3 is annoying in each test with different patches etc. Started a "framework" of sorts where both etherscan and web3 and related apis are mocked from the very start, at the fixture initialization. And each test can parametrize the mocking on its own. This will probably evolve as it is used in more tests and replaces older mocks.

### DIFF
--- a/rotkehlchen/tests/unit/test_defi_oracles.py
+++ b/rotkehlchen/tests/unit/test_defi_oracles.py
@@ -23,7 +23,7 @@ if TYPE_CHECKING:
 
 @pytest.mark.parametrize('use_clean_caching_directory', [True])
 @pytest.mark.parametrize('should_mock_current_price_queries', [False])
-def test_uniswap_oracles_asset_to_asset(inquirer_defi):
+def test_uniswap_oracles_asset_to_asset(inquirer_defi, socket_enabled):  # pylint: disable=unused-argument  # noqa: E501
     """
     Test that the uniswap oracles return a price close to the one reported by
     coingecko.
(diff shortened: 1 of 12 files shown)


rotkehlchen/tests/unit/test_defi_oracles.py
```diff
@@ -23,7 +23,7 @@
 
 @pytest.mark.parametrize('use_clean_caching_directory', [True])
 @pytest.mark.parametrize('should_mock_current_price_queries', [False])
-def test_uniswap_oracles_asset_to_asset(inquirer_defi):
+def test_uniswap_oracles_asset_to_asset(inquirer_defi, socket_enabled):  # pylint: disable=unused-argument  # noqa: E501
```
LefterisJP, line 26: Hey @yabirgb I was not sure about these 3 tests. I marked them as `socket_enabled` which means that they should not fail even when running with socket-disabled, just because I was not sure if they can be mocked.

They probably should, maybe except the first one? I need help here to determine what to do.

LefterisJP:
Merging. All tests that failed pass locally.

PR #5270, opened 2022-12-19
Mock network for eth2 deposit related unit tests

LefterisJP:


LefterisJP:
only covalent api rate limit stopped tests again so merging.

PR #5074, opened 2022-11-01
Customizable priorities of address names sources

tomvodi:
Closes #4487 

## Checklist

- [x] The PR modified the frontend, and updated the [user guide](https://github.com/rotki/rotki/blob/develop/docs/usage_guide.rst) to reflect the changes.

PR #5275, opened 2022-12-19
WIP: Adds frontend support for optimism

kelsos:
Closes #(issue_number)

## Checklist


yabirgb:
Noticed that optimism is grayed out while ethereum is not

image

yabirgb:
About changing the name of the account it seems to be linked and only if I change it in ethereum it changes in optimism. I believe they should be linked but editing op also should edit it on ethereum. Also if you edit it in op it shows the name that the ethereum account has but in the edit field it shows a different thing

kelsos:
> Noticed that optimism is grayed out while ethereum is not
> 
> image

This is an incorrect observation. I am not sure what you expected, but applying grayscale filters to different colors will not have the same results.

The optimism logo has a red color, and the ethereum one has a blue-gray one, the grayscale filter will produce different results for them.

To verify, you can go hover the mouse over the ethereum logo and you will notice there is a really slight change when you hover, but the colors are really similar so you really have to pay attention.

kelsos:
> About changing the name of the account it seems to be linked and only if I change it in ethereum it changes in optimism. I believe they should be linked but editing op also should edit it on ethereum. Also if you edit it in op it shows the name that the ethereum account has but in the edit field it shows a different thing

A while ago we changed how the name display for addresses works.

The way the address display works for Ethereum addresses works, [see here](https://github.com/rotki/rotki/pull/5074#issuecomment-1314982049), is that the addresses use the label provided by `/api/1/names`.

This is what causes the "linked" display, and also the difference between what you see in the table and what you see when editing. This has been the default behavior since #5074 was merged.

yabirgb (approved):
All looks good for what I tested. The comments were clarified and we need to sit to discuss the issue of the names. LGTM

PR #5277, opened 2022-12-19
Multi evm account addition

LefterisJP:
For Kelsos :heart:


rotkehlchen/api/rest.py
```diff
@@ -1967,6 +1967,47 @@ def edit_xpub(
...
+    def _add_evm_accounts(
+            self,
+            account_data: list[BlockchainAccountData],
+    ) -> dict[str, Any]:
+        try:
+            added_accounts = self.rotkehlchen.add_evm_accounts(account_data=account_data)
+        except EthSyncError as e:
+            return {'result': None, 'message': str(e), 'status_code': HTTPStatus.CONFLICT}
+        except TagConstraintError as e:
+            return {'result': None, 'message': str(e), 'status_code': HTTPStatus.CONFLICT}
```
yabirgb, line 1979: This two can be combined in one single except clause

rotkehlchen/api/v1/schemas.py
```diff
@@ -1447,7 +1430,7 @@ def _transform_btc_or_bch_address(
     return address
 
 
-def _transform_eth_address(
+def _transform_evm_address(
         ethereum_inquirer: EthereumInquirer,
         given_address: str,
 ) -> ChecksumEvmAddress:
```
yabirgb, line 1436: Here is only ethereum that has `ens_lookup`. If you pass an ens name it should raise an error for chains that don't have it right?

rotkehlchen/chain/aggregator.py
```diff
@@ -1292,3 +1293,38 @@ def get_chain_manager(
...
+        If it's already added in a chain we just ignore that chain.
+
+        Returns a list of tuples of the address and the chain it was added in
+        """
+        added_accounts = []
+        for account in accounts:
+            chains: list[SUPPORTED_EVM_CHAINS]
+            if self.ethereum.node_inquirer.get_code(account) != '0x':
+                log.debug(f'Not adding {account} to all evm chains as it is a mainnet contract')
+                chains = [SupportedBlockchain.ETHEREUM]
```
yabirgb, line 1314: The case of a contract in other chains add extra complexity and slows the addition. Is that why you didn't consider it?

rotkehlchen/rotkehlchen.py
```diff
@@ -525,6 +526,48 @@ def get_blockchain_account_data(
...
+                data_type='blockchain accounts',
+            )
+            added_accounts = self.chains_aggregator.add_accounts_to_all_evm(
+                accounts=[ChecksumEvmAddress(entry.address) for entry in account_data],  # type: ignore[arg-type]  # noqa: E501 # we know it's evm address
+            )
+            for chain, address in added_accounts:
+                account_data_entry = account_data_map[address]
+                if chain != SupportedBlockchain.ETHEREUM:
+                    # for now add tags only to mainnet address
+                    # TODO: Need to make this generic so that tag mapping is blockchain+address
```
yabirgb, line 561: Create an issue for this?

yabirgb (commented):
One thing that is a bug I believe and small comments

rotkehlchen/api/rest.py
LefterisJP: good point. Thanks.

rotkehlchen/chain/aggregator.py
LefterisJP: No no. If it's a contract then you don't own the same address in any other chain. EOA you own the private key and it's repeatable in all chains. Contract account no. So for a contract account we should not do multi-evm addition.

rotkehlchen/rotkehlchen.py
LefterisJP: no need to. I will take it.

LefterisJP:
All tests pass locally so will merge. Will make a tiny PR with the suggested change.

PR #5280, opened 2022-12-20
Fix argument to xpub new addresses query in task

yabirgb:
The function was taking `blockchain` as argument that is None always in that branch of the `if` statement


LefterisJP (approved):
Lgtm

PR #5284, opened 2022-12-21
Do not timeout prematurely for substrate node connection

LefterisJP:


LefterisJP:
single test failure is a covalent timeout

PR #5285, opened 2022-12-21
Combine EthSyncerror and TagConstraintError

LefterisJP:
As per Yabir's feedback in https://github.com/rotki/rotki/pull/5277


yabirgb (approved):
Giving an approval

PR #5289, opened 2022-12-21
add missing changelog entry

prettyirrelevant:


docs/changelog.rst
```diff
@@ -11,6 +11,7 @@ Changelog
 * :bug:`5038` Premium users with big databases should no longer see the error: "Upload data to server died with exception: database plaintext is locked"
 
 * :release:`1.26.2 <2022-12-09>`
+* :bug:`4332` Price oracles are now banned temporarily after repeated failures.
```
LefterisJP, line 14: this should go to bugfixes. Also use same terminology as in the code. I already told you about it when you made the feature.
```suggestion
* :bug:`4332` Price oracles are now penalized temporarily after repeated failures.
```

LefterisJP: In fact also the location of this is wrong. I don't see it in 1.26.2. This is in develop. So 1.27.0

LefterisJP (commented):
Anyway I also solved it myself as it's easier than waiting

PR #5293, opened 2022-12-22
Make addressses names multichain

nebolax:
Make addresses names multichain


rotkehlchen/db/addressbook.py
```diff
@@ -42,20 +47,36 @@ def write_ctx(self, book_type: AddressbookType) -> Generator['DBCursor', None, N
...
-                f'SELECT address, name FROM address_book WHERE address IN ({questionmarks})',
-                addresses,
-            )
-
-        entries = []
-        for address, name in cursor:
-            entries.append(AddressbookEntry(address=ChecksumEvmAddress(address), name=name))
+            for address, blockchain in addresses:
+                query = 'SELECT address, name, blockchain FROM address_book WHERE address = ?'
+                bindings = [address]
```
yabirgb, line 67: this is ugly, you are doing one query per item in the tuple. You could build the queries and do a `executemany`. It would be even less queries if you group by blockchain and then do as it was before but grouped by blockechain. You would have as many as `len(distinct_blockchains)` queries

rotkehlchen/db/addressbook.py
```diff
@@ -83,46 +104,54 @@ def add_addressbook_entries(
     def update_addressbook_entries(
             self,
             book_type: AddressbookType,
-            entries: list[AddressbookEntry],
+            entries: list[tuple[ChecksumEvmAddress, str, Optional[SupportedBlockchain]]],
     ) -> None:
         with self.write_ctx(book_type) as write_cursor:
```
yabirgb, line 109: can you add a docstring comment explaining the behavior when the blockchain param is None and not?

rotkehlchen/db/addressbook.py
```diff
@@ -83,46 +104,54 @@ def add_addressbook_entries(
...
+                if blockchain is not None:
+                    query += ' AND blockchain = ?'
+                    bindings.append(blockchain.value)
+                write_cursor.execute(query, bindings)
                 if write_cursor.rowcount == 0:
                     raise InputError(
-                        f'Addressbook entry with address "{address}" '
-                        f'doesn\'t exist in the address book. So it cannot be deleted.',
+                        f'Addressbook entry with address "{address}" doesn\'t '
+                        f'exist in the address book. So it cannot be deleted.',
```
yabirgb, line 139: You should add the blockchain param to the error message if it is provided

yabirgb: Also left told me to not use the `\'` thing and just use double quotes

rotkehlchen/chain/ethereum/names.py
```diff
@@ -171,20 +182,25 @@ def global_addressbook_address_to_name(
...
 
 
 def hardcoded_address_to_name(
         _: DBHandler,
         address: ChecksumEvmAddress,
+        blockchain: SupportedBlockchain,
 ) -> Optional[str]:
     """Returns the name of a known address or None if there is no such address"""
+    if blockchain != SupportedBlockchain.ETHEREUM:
+        return None
```
yabirgb, line 196: We shoudl move those to the global db at some point. Maybe this release is fine and we would be able to get rid of this piece of code. But not this PR

rotkehlchen/chain/ethereum/names.py
```diff
@@ -134,21 +141,23 @@ def get_prioritized_names(
...
 ) -> Optional[str]:
     """Returns the label of an ethereum blockchain account with the given address or
     None if there is no such account or the account has no label set.
     """
     with db.conn.read_ctx() as cursor:
         return db.get_blockchain_account_label(
             cursor=cursor,
-            blockchain=SupportedBlockchain.ETHEREUM,
             address=address,
+            blockchain=blockchain,
```
yabirgb, line 153: why changing the order of the arguments? It doesn't matter that much but it reads better if you use the same order

rotkehlchen/api/rest.py
```diff
@@ -4624,12 +4633,17 @@ def delete_addressbook_entries(
...
+            addresses: list[tuple[ChecksumEvmAddress, Optional[SupportedBlockchain]]],
+    ) -> Response:
         mappings = search_for_addresses_names(
             database=self.rotkehlchen.data.db,
             addresses=addresses,
         )
-        return api_response(_wrap_in_ok_result(mappings))
+        return api_response(_wrap_in_ok_result(process_result_list(
+            [[*x] for x in mappings],
+        )))
```
yabirgb, line 4646: This is changing the response format right? It should update also the docs. Also have you talked with the frontend to discuss what format is better for them or if this one makes sense? 

Lastly don't use single letter variables, here is confusing and is not clear what `x` is just by reading this part of the code. Use something like `named_address_tuple` or something more descriptive. In other places by the context is easy but here it comes from a function call that is already complex

rotkehlchen/tests/utils/factories.py
```diff
@@ -152,14 +153,17 @@ def make_addressbook_entries() -> list[AddressbookEntry]:
...
         ),
         AddressbookEntry(
             address=to_checksum_address('0x368B9ad9B6AAaeFCE33b8c21781cfF375e09be67'),
             name='Neighbour Thomas',
+            blockchain=SupportedBlockchain.ETHEREUM,
         ),
         AddressbookEntry(
             address=to_checksum_address('0x3D61AEBB1238062a21BE5CC79df308f030BF0c1B'),
             name='Secret agent Rose',
+            blockchain=SupportedBlockchain.ETHEREUM,
```
yabirgb, line 166: can you also add another example with optimism? and test that requesting a name for optimism works as expected. Not saying it here but in the right place

rotkehlchen/db/addressbook.py
```diff
@@ -83,46 +104,54 @@ def add_addressbook_entries(
...
     ) -> Optional[str]:
         """Returns the name for a specific address.
         Returns None if either there is no name set or the address doesn't exist in database.
         """
         with self.read_ctx(book_type) as read_cursor:
             query = read_cursor.execute(
-                'SELECT name FROM address_book WHERE address=?',
-                (address,),
+                'SELECT name FROM address_book WHERE address=? AND blockchain=?',
+                (address, blockchain.value),
```
yabirgb, line 154: docstring needs to get updated

rotkehlchen/db/addressbook.py
```diff
@@ -83,46 +104,54 @@ def add_addressbook_entries(
...
             book_type: AddressbookType,
-            addresses: list[ChecksumEvmAddress],
+            addresses: list[tuple[ChecksumEvmAddress, Optional[SupportedBlockchain]]],
     ) -> None:
         with self.write_ctx(book_type) as write_cursor:
-            for address in addresses:
-                write_cursor.execute('DELETE FROM address_book WHERE address = ?', (address,))
+            for address, blockchain in addresses:
+                query = 'DELETE FROM address_book WHERE address = ?'
+                bindings: list[Any] = [address]
```
yabirgb, line 131:
```suggestion
                bindings: list[str] = [address]
```

rotkehlchen/db/addressbook.py
```diff
@@ -42,20 +47,36 @@ def write_ctx(self, book_type: AddressbookType) -> Generator['DBCursor', None, N
     def get_addressbook_entries(  # pylint: disable=no-self-use
             self,
             cursor: 'DBCursor',
-            addresses: Optional[list[ChecksumEvmAddress]] = None,
+            addresses: Optional[list[
+                tuple[ChecksumEvmAddress, Optional[SupportedBlockchain]]
+            ]] = None,
     ) -> list[AddressbookEntry]:
+        entries = []
```
yabirgb, line 54: can you also add a docstring explaining what this does regarding the arguments?

rotkehlchen/api/rest.py
```diff
@@ -4597,7 +4606,7 @@ def add_addressbook_entries(
     def update_addressbook_entries(
             self,
             book_type: AddressbookType,
-            entries: list[AddressbookEntry],
+            entries: list[tuple[ChecksumEvmAddress, str, Optional[SupportedBlockchain]]],
```
yabirgb, line 4609: Since you are repeating this type in several places maybe you can declare this as a type and reuse it

yabirgb (commented):
Some comments and things to improve

rotkehlchen/api/rest.py
nebolax: Yeah, thank you for reminding to ask frontend guys :pray: 

Didn't change the docs yet because wanted to get an agreement on the api format first

rotkehlchen/db/addressbook.py
```diff
@@ -83,46 +117,60 @@ def add_addressbook_entries(
...
+                query = 'UPDATE address_book SET name = ? WHERE address = ?'
+                bindings = [name, address]
+                if blockchain is not None:
+                    query += 'AND blockchain = ?'
+                    bindings.append(blockchain.value)
+                write_cursor.execute(query, bindings)
                 if write_cursor.rowcount == 0:
                     raise InputError(
-                        f'Addressbook entry with address "{entry.address}" and name "{entry.name}"'
+                        f'Addressbook entry with address "{address}" and name "{name}"'
```
LefterisJP, line 136: Shouldn't the error message specify a chain too?

rotkehlchen/db/addressbook.py
```diff
@@ -83,46 +117,60 @@ def add_addressbook_entries(
...
-                write_cursor.execute('DELETE FROM address_book WHERE address = ?', (address,))
+            for address, blockchain in addresses:
+                query = 'DELETE FROM address_book WHERE address = ?'
+                bindings: list[str] = [address]
+                if blockchain is not None:
+                    query += ' AND blockchain = ?'
+                    bindings.append(blockchain.value)
+                write_cursor.execute(query, bindings)
                 if write_cursor.rowcount == 0:
+                    blockchain_msg = f' and blockchain {blockchain.value}' if blockchain is not None else ''  # noqa: E501
```
LefterisJP, line 154: Why create a new variable here? Just put it in the already existing f string

LefterisJP (commented):
Lots of comments. This PR which should have been really simple is full of very badly written code and is all in all a hack.

If merged as it it would be a terrible addition to our technical debt. I left a lot of comments as to why that is pointing to specific parts of the PR's code.

But after reading all of it I also think I understand why. There seems to have been no thought at all about the feature. The developer seems to have just started writing code hacking stuff around without understanding what they are building. So let's see what should the developer have done.

#### Think what do we have until now

Right now we have the ethereum address book: https://rotki.readthedocs.io/en/latest/usage_guide.html#ethereum-address-book

a way for users to enter names for ethereum addresses and for the app to use them in the various labels such as the transactions overview.

#### What do we want to change?

We want to generalize this and make it chain-agnostic. Also, though this should now still be used only for EVM chains keep it as agnostic as possible so it used.

That means that you can add an address and also specify a chain for it. If you don't specify a chain then all "EVM" chains in this case are affected.

#### How can we do that?

This PR does it via ... well hacking around everything. It's all ugly and passing stuff here and stuff there. Not even sure if it does what it should properly as the code is hard to read, and hard to understand and test.

Why is that? Because of the way the mappings of address to names are represented in the DB.

Both DBs represent it like this before this PR:

```sql
CREATE TABLE IF NOT EXISTS address_book (
    address TEXT NOT NULL,
    blockchain TEXT NOT NULL DEFAULT "ETH",
    name TEXT NOT NULL,
    PRIMARY KEY(address, blockchain)
);
```

The PR does not touch this at all. And there lies the problem.

The schema does not represent what we want to have after this PR. What we have is either a chain mapping (which the schema can represent) or a chain agnostic mapping which it can't. It would need blockchain to be nullable, in which case we can consider it as chain agnostic.

I was expecting an addition to a DB upgrade for both DBs as I had told Alexey when I had my other PR open. Weird this did not happen.

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1092,6 +1092,8 @@ def test_upgrade_db_35_to_36(user_data_dir):  # pylint: disable=unused-argument
         ('0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12', 'hardware'),
         ('1PUrJgftNnHvvqVyEsm9DiCDQuZHCn47fQ', 'hot'),
     ]
+    cursor.execute('SELECT * FROM address_book')
+    assert cursor.fetchall() == [('0xc37b40ABdB939635068d3c5f13E7faF686F03B65', 'ETH', 'yabir secret account')]  # noqa: E501
```
LefterisJP, line 1096: Can you maybe both here and in global add 1 more? Just to see that we don't have off by one errors and that everything is transferred.

Whenever I add data to the test DB for upgrade I always add more than 1 exactly for this reason.

rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -281,40 +384,123 @@ def names_request(addresses: list[str]):
...
+            book_type=book_type,
+        ),
+        json={
+            'addresses': [
+                {'address': test_address},
+            ],
+        },
+    )
+    result = assert_proper_response_with_result(response)
+    assert result == [custom_name.serialize()]
```
LefterisJP, line 506: Shouldn't you also check the DB in this test? Or you can just move that to a unit test of the corresponding DB function.

What's more ... do we have a test for addign same name but for different chain? For example:

1. Add a name for 0xyabir in ethereum.
2. Add same name for 0xyabir in optimism.

What happens then? Do we have a test for this? I guess what "should" happen is API error.

LefterisJP (commented):
Thank you. Much better. Left some comments.

rotkehlchen/db/addressbook.py
```diff
@@ -64,65 +94,87 @@ def add_addressbook_entries(
...
         with self.write_ctx(book_type) as write_cursor:
-            for address in addresses:
-                write_cursor.execute('DELETE FROM address_book WHERE address = ?', (address,))
+            for address, blockchain in addresses:
+                query = 'DELETE FROM address_book WHERE address = ?'
+                bindings: list[str] = [address]
+                if blockchain is not None:
+                    query += ' AND blockchain = ?'
+                    bindings.append(blockchain.value)
+                write_cursor.execute(query, bindings)
```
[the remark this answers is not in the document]
yabirgb: The reason is to properly format the error when returned to the frontend. Otherwise you don't know which one failed

rotkehlchen/tests/api/test_addressbook.py
yabirgb: Why an error? Having same or different name for address `0xyabir` on different chains is allowed. Otherwise the primary key would be name and address and not blockchain and address as it is now

rotkehlchen/db/addressbook.py
```diff
@@ -64,65 +94,87 @@ def add_addressbook_entries(
...
                     ) from e
 
     def update_addressbook_entries(
             self,
             book_type: AddressbookType,
             entries: list[AddressbookEntry],
     ) -> None:
+        """
+        Updates names of addressbook entries.
+        If blockchain is None then update all entries with the specified address.
```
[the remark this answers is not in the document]
yabirgb: Yes, it doesn't make sense. Removing it. It makes sense in `add_addressbook_entries` where if the blockchain is NULL then the name is valid for all the evm chains and it doesn't make sense to have their rows in the database

rotkehlchen/db/addressbook.py
```diff
@@ -64,65 +94,87 @@ def add_addressbook_entries(
...
-                        (entry.address, entry.name),
+                        'INSERT INTO address_book (address, name, blockchain) VALUES (?, ?, ?)',
+                        entry.serialize_for_db(),
                     )
                 # Handling both private db (pysqlcipher) and global db (raw sqlite3)
                 except (dbapi2.IntegrityError, sqlite3.IntegrityError) as e:  # pylint: disable=no-member  # noqa: E501
                     raise InputError(
-                        f'Addressbook entry with address "{entry.address}" and name "{entry.name}"'
-                        f' already exists in the address book. Identifier must be unique.',
+                        f'Addressbook entry with address "{entry.address}", name "{entry.name}" '
```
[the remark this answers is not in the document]
yabirgb: In the case of the delete we don't have the address book entry instance so is the one I haven't changed. For the other two I've used the str method

rotkehlchen/db/addressbook.py
```diff
@@ -64,67 +93,83 @@ def add_addressbook_entries(
             book_type: AddressbookType,
             entries: list[AddressbookEntry],
     ) -> None:
+        """
+        Add every entry of entries to the address book table.
+        If blockchain is None then make sure that the same address doesn't appear in combination
+        with other blockchain value.
```
LefterisJP, line 99:
```suggestion
        with other blockchain values.
```

rotkehlchen/db/addressbook.py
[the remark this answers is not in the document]
LefterisJP: this is not a good enough reason. You don't need to know which specific name failed. Unless it's an explicit requirement for some reason I don't get understand, please optimize for the least number of DB writes.

rotkehlchen/tests/api/test_addressbook.py
LefterisJP: Yes okay true. But can you add a test for this? To see same name for different addresses across different chains works?

But then also what happens if:

1. I add "yabiroptimism" for "0x1" in optimism
2. I add "yabirethereum" for "0x1" in ethereum.
3. Add "yabir" for "0x1" for chain=None?

yabirgb: We had them in the inser. In the case where we insert for chain = None I have added it here `test_insert_into_addressbook`. As we agreed it deletes the other two entries and leaves the one with `chain=None`

rotkehlchen/db/addressbook.py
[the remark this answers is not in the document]
yabirgb: Found a case where this is needed. If we delete by address only it can happen that the address has more than one entry (different chains) and this might result in a error detecting addresses that couldn't be deleted because they dont' exist in the db

rotkehlchen/db/addressbook.py
```diff
@@ -64,67 +93,106 @@ def add_addressbook_entries(
...
+        with self.read_ctx(book_type) as read_cursor:
+            read_cursor.execute(
+                f'SELECT DISTINCT address FROM address_book WHERE address IN ({",".join("?"*len(chain_addresses))})',  # noqa: E501
+                list(addresses),
+            )
+            db_addresses = {row[0] for row in read_cursor}
+            if len(addresses) != len(db_addresses):
+                raise InputError(f'Addresses {addresses} are not present in the database')
+
+        delete_without_blockchain = 'DELETE FROM address_book WHERE address = ?'
```
LefterisJP, line 162: Does the API allow both addresses with name + chain and addresses with name + no chain specified in the same call? Cause this seems to complicate the logic here a bit no?

If the API allows this then should it be allowed? How does the frontend use this feature?

LefterisJP (commented):
typos and a question

LefterisJP (approved):
Will merge to unblock frontend and we should figure out the open issues in a subsequent PR

PR #5299, opened 2022-12-23
Remove pacakged db write attempt

nebolax:
Remove pacakged db write attempt

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,8 +5,8 @@ Changelog
 * :bug:`-` Connecting to substrate nodes will no longer timeout prematurely for systems with slow connections.
 * :bug:`-` Transfers between tracked accounts will now have a correct label in the UI.
 * :bug:`-` Users will be able to finish balance queries if they have assets with missing information.
-* :bug:`-` Visiting the trades page won't query now trades from exchanges that are ignored.
-* :bug:`5038` Premium users with big databases should no longer see the error: "Upload data to server died with exception: database plaintext is locked"
+* :bug:`5265` Visiting the trades page will no longer query trades from ignored exchanges.
+* :bug:`5038` Premium users with big databases should no longer see the error: "Upload data to server died with exception: database plaintext is locked".
 * :bug:`-` Tokens added by the Balancer module will now have the name field correctly set.
 * :bug:`-` If a user removes the API keys for an exchange, actions on that exchange will no longer be excluded from PnL reports.
 
--- a/rotkehlchen/globaldb/handler.py
+++ b/rotkehlchen/globaldb/handler.py
@@ -144,8 +144,9 @@ class GlobalDBHandler():
             return GlobalDBHandler()._packaged_db_conn  # type: ignore
 
         packaged_db_path = Path(__file__).resolve().parent.parent / 'data' / 'global.db'
-        packaged_db_conn = initialize_globaldb(
-            dbpath=packaged_db_path,
+        packaged_db_conn = DBConnection(
+            path=packaged_db_path,
+            connection_type=DBConnectionType.GLOBAL,
             sql_vm_instructions_cb=DEFAULT_SQL_VM_INSTRUCTIONS_CB,
         )
         GlobalDBHandler()._packaged_db_conn = packaged_db_conn
--- a/rotkehlchen/tests/unit/test_assets.py
+++ b/rotkehlchen/tests/unit/test_assets.py
@@ -1,4 +1,9 @@
+import os
+import shutil
 import warnings as test_warnings
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from unittest.mock import PropertyMock, patch
 
 import pytest
 from eth_utils import is_checksum_address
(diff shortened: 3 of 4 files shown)


docs/changelog.rst
```diff
@@ -7,6 +7,7 @@ Changelog
 * :bug:`-` Users will be able to finish balance queries if they have assets with missing information.
 * :bug:`5038` Premium users with big databases should no longer see the error: "Upload data to server died with exception: database plaintext is locked"
 * :bug:`-` If a user removes the API keys for an exchange, actions on that exchange will no longer be excluded from PnL reports.
+* :bug:`-` Rotki will not attempt to write to packaged files anymore.
```
LefterisJP, line 10:
```suggestion
* :bug:`-` Rotki will not attempt to write to the packaged DB anymore.
```

rotkehlchen/globaldb/handler.py
```diff
@@ -144,8 +144,9 @@ def packaged_db_conn() -> DBConnection:
             return GlobalDBHandler()._packaged_db_conn  # type: ignore
 
         packaged_db_path = Path(__file__).resolve().parent.parent / 'data' / 'global.db'
-        packaged_db_conn = initialize_globaldb(
-            dbpath=packaged_db_path,
+        packaged_db_conn = DBConnection(
```
LefterisJP, line 147: this needs a test

rotkehlchen/tests/unit/test_assets.py
```diff
@@ -633,3 +638,43 @@ def test_symbol_or_name(database):
...
         Asset('i-dont-exist').symbol_or_name()
+
+
+def test_load_from_packaged_db(globaldb: GlobalDBHandler):
+    """
+    Test that when resolving assets from the packaged db,
+    there is no attempts to write into the pacakged db.
+    """
+    packaged_db_path = Path(__file__).resolve().parent.parent.parent / 'data' / 'global.db'
+    backup = shutil.copy(packaged_db_path, packaged_db_path.with_suffix('.backup'))
```
LefterisJP, line 649: Why do this in the packaged db path? Create a temp directoy for this. Then you won't care about deletion.

rotkehlchen/tests/unit/test_assets.py
```diff
@@ -633,3 +639,43 @@ def test_symbol_or_name(database):
...
+                # Try to load the asset
+                CryptoAsset('BTC')
+                # Check that the packaged db was used
+                assert packaged_db_mock.call_count == 1
+
+            # Check that the packaged db was not modified
+            hash_after = file_md5(packaged_db_path)
+            assert hash_before == hash_after
+        finally:
+            shutil.copy(backup, packaged_db_path)
```
LefterisJP, line 681: why do you do this? Why wrap the entire thing in try/finally? And why at the finally copy back?

nebolax: Because to provoke the execution of the script that creates new tables, I delete settings table form the packaged db

nebolax: And in order to restore the packaged db after the db, I added this `finally:` block

LefterisJP: Okay then this entire test is very bad. Even in tests you must not write on the packaged DB. Mock it please. Also perhaps you can just see if what @yabirgb mentioned is true. Perhaps just an attempt to write is what breaks and testing the md5hash makes no sense.

rotkehlchen/tests/unit/test_assets.py
```diff
@@ -633,3 +638,25 @@ def test_symbol_or_name(database):
...
+        backup = shutil.copy(packaged_db_path, dest_file)
+
+        # mock Path parent attribute to return the destination file always
+        def parent():
+            return Path(tmpdirname)
+
+        # set the permissions for the copy of the globaldb to read only
+        backup.chmod(0o444)
+        # mock the parent method from pathlib in the initialization of the globaldb
+        with patch('pathlib.Path.parent', new_callable=PropertyMock) as mock_path:
```
LefterisJP, line 659: Can you also somehow make sure that the `dest_file` packaged_db is accessed in this test? Otherwise I gotta trust that the code you wrote is correct.

yabirgb: Added the comment about the read permissions and added a read to ensure we can read from the packaged globaldb (is well connected)

LefterisJP: Still that's not what I meant.

What you are doing now is you are testing that you can read from the packaged global DB which does not tell you anything. I mean ofc you can. If you could not this entire thing would not work.

What I told you before and in discord is to: **Determine that you are reading from the `dest_file` global DB**. Essentially this will make sure your mock works. Otherwise you are just reading the normal packaged DB and the test has no effect.

The way I can imagine this being done is:
1. Add a special setting or something, or a custom price or anything directly to dest_file using sqlite.
2. At the end when you read from the packaged DB, make sure that thing you added is there.

This will achieve what you need. Which is to determine that the mock works fine.

Don't forget to explain with comments what you are doing.

LefterisJP (approved):
LGTM so long as tests pass. Frontend failed with the bittrex api key problem again.

PR #5301, opened 2022-12-24
Add script that detects useless pylint suppresion and removes it. Use it.

LefterisJP:
As per the docstring this has unfortunately some false positives but
the fault lies with the pylint detection mechanism.


yabirgb (approved):
Failing tests are covalent API queries that failed. LGTM

PR #5303, opened 2022-12-25
Mock network for avalanche/convalent tests

LefterisJP:


LefterisJP:
Backend tests are green :heavy_check_mark: 

Merging. We need to figure out why frontend tests always fail in develop though.

PR #5306, opened 2022-12-26
Add nfts manual prices to the API latest manual prices response

yabirgb:
Return prices of NFTs stored in the database

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -4106,6 +4106,22 @@ class RestAPI():
                 'to_asset': price_entry[1],
                 'price': price_entry[2],
             })
+
+        if (nft_module := self.rotkehlchen.chains_aggregator.get_module('nfts')) is not None:
+            # query also nfts manual prices
+            nft_price_data = nft_module.get_nfts_with_price(
+                from_asset=from_asset,
+                to_asset=to_asset,
+                only_with_manual_prices=True,
+            )
+
+            for nft_data in nft_price_data:
+                prices_information.append({
+                    'from_asset': nft_data['asset'],
+                    'to_asset': nft_data['price_asset'],
+                    'price': nft_data['price_in_asset'],
+                })
+
         return api_response(_wrap_in_ok_result(prices_information), status_code=HTTPStatus.OK)
 
     def get_nfts_with_price(self, lps_handling: NftLpHandling) -> Response:
--- a/rotkehlchen/chain/ethereum/modules/nft/nfts.py
+++ b/rotkehlchen/chain/ethereum/modules/nft/nfts.py
@@ -1,6 +1,6 @@
 import logging
 from collections import defaultdict
-from typing import TYPE_CHECKING, Any, Optional, Union
+from typing import TYPE_CHECKING, Any, Optional
 
 from pysqlcipher3 import dbapi2 as sqlcipher
 
@@ -8,6 +8,7 @@ from rotkehlchen.assets.asset import Asset
 from rotkehlchen.chain.ethereum.modules.uniswap.v3.types import AddressToUniswapV3LPBalances
 from rotkehlchen.constants.assets import A_USD
 from rotkehlchen.constants.misc import ZERO
+from rotkehlchen.db.filtering import NFTFilterQuery
 from rotkehlchen.errors.asset import UnknownAsset
 from rotkehlchen.errors.misc import InputError, RemoteError
 from rotkehlchen.externalapis.opensea import NFT, Opensea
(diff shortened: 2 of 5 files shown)


rotkehlchen/api/rest.py
```diff
@@ -4106,6 +4106,26 @@ def get_manual_latest_prices(
                 'to_asset': price_entry[1],
                 'price': price_entry[2],
             })
+
+        if self.rotkehlchen.chains_aggregator.get_module('nfts') is not None:
+            # query also nfts manual prices
+            nft_price_data = self._eth_module_query(
```
LefterisJP, line 4112: Does this also query the nft prices? Or just read from the DB? From a quick look it just reads from the DB, which should be okay.

Also why go through `_eth_module_query()` and not query `get_nfts_with_price` directly? What's the reasoning of the extra work? You are actually reading just the `result` of the dict and not using the constructured api response at all.

rotkehlchen/chain/ethereum/modules/nft/nfts.py
```diff
@@ -280,36 +284,48 @@ def get_nfts_with_price(
             query_str += ' AND is_lp=?'
             bindings.append(lps_handling == NftLpHandling.ONLY_LPS)
 
+        if only_with_manual_prices is True:
```
LefterisJP, line 287: the filter is getting complicated here. Can't you build a normal filter like we do in the rest of the code?

rotkehlchen/chain/ethereum/modules/nft/nfts.py
```diff
@@ -280,36 +284,48 @@ def get_nfts_with_price(
...
-                    to_asset = Asset(to_asset_id).check_existence()
+                    target_asset = Asset(to_asset_id).check_existence()
                 except UnknownAsset:
                     log.error(
                         f'Unknown asset {to_asset_id} in custom nft price DB table. Ignoring.',
                     )
                     continue
 
-                if to_asset != A_USD:
+                if target_asset != A_USD and only_with_manual_prices is False:
```
LefterisJP, line 312: So `if target_asset != A_USD and only_with_manual_prices is True:`

then `usd_price` is not really a usd price? Isn't that broken?

rotkehlchen/tests/api/test_current_assets_price.py
```diff
@@ -527,3 +531,78 @@ def test_prices_cache_invalidation_for_manual_prices(rotkehlchen_api_server):
     # # the max diff is to account for price changes between the requests interval
     assert FVal(avax_result['assets']['AVAX'][0]).is_close(FVal(eth_result['assets']['ETH'][0]), max_diff='10')  # noqa: E501
     assert result['target_asset'] == 'USD'
+
+
+@pytest.mark.parametrize('should_mock_current_price_queries', [True])
+@pytest.mark.parametrize('ethereum_accounts', [['0xc37b40ABdB939635068d3c5f13E7faF686F03B65']])
+@pytest.mark.parametrize('ethereum_modules', [['nfts']])
+def test_get_manual_prices_with_nfts(
+    rotkehlchen_api_server: 'APIServer',
```
LefterisJP, line 540: shouldnt' this need 1 more indentation level?

rotkehlchen/api/rest.py
yabirgb: It queries from the database. Okey

rotkehlchen/tests/api/test_current_assets_price.py
yabirgb: I have now the correct indentation in the editor but might have edited it myself

rotkehlchen/chain/ethereum/modules/nft/nfts.py
yabirgb: I have changed it to make the field nullable and then the logic makes sense again. Also added the info to the docstring

rotkehlchen/db/filtering.py
```diff
@@ -852,13 +852,20 @@ def make(
 
 @dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=False)
 class DBNullFilter(DBFilter):
-    columns: list[str]
+    null_columns: Optional[list[str]] = None
```
LefterisJP, line 855: Why this change? Isn't it more expressive to just have a different filter for the not null? Do you at any point combine those? So have something with both null_columns and not_null_columns?

yabirgb: No, I don't combine them in the current filters. Will then split then it in two

LefterisJP (approved):
Perfect. Nice thing with the verb. LGTM.

rotkehlchen/db/filtering.py
LefterisJP: makes sense. Thanks!

PR #5307, opened 2022-12-27
Fix tag editing. Should now delete old tag mappings

LefterisJP:


LefterisJP:
Merging as this fixes something important. The frontend test will hopefully be fixed in a subsequent PR by @lukicenturi

PR #5308, opened 2022-12-27
fix: empty tags to null

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/accounts/AccountForm.vue
+++ b/frontend/app/src/components/accounts/AccountForm.vue
@@ -36,6 +36,7 @@ import { startPromise } from '@/utils';
 import { assert } from '@/utils/assertions';
 import { getMetamaskAddresses } from '@/utils/metamask';
 import { xpubToPayload } from '@/utils/xpub';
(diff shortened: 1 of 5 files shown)


LefterisJP (commented):
Tests are failings still. Is there something off?

lukicenturi:
> Tests are failings still. Is there something off? 

My bad, should be fine now

LefterisJP (approved):
LGTM since tests pass.

@kelsos I am taking the liberty to approve and merge to unblock our CI green-ness. If you find anything wrong please fix in subsequent PRs.

kelsos:
I was doing a check, it it would make sense for all the payload classes that deal with arrays that are send to the backend to start introducing a `NonEmptyArray<T> | null` maybe as a type.

https://stackoverflow.com/questions/56006111/is-it-possible-to-define-a-non-empty-array-type-in-typescript

kelsos:
@lukicenturi also we have a method to remove nullable fields from the payload in these cases maybe it would make sense to use them.

PR #5310, opened 2022-12-27
Update docs with instructions for running tests

yabirgb:
- Update docs
- Change order of execution for linting
- Remove ftx from assets testing in makefile

### DIFF
--- a/docs/contribute.rst
+++ b/docs/contribute.rst
@@ -398,6 +398,12 @@ We require this wrapper as a drop-in replacement of pytest due to quirks of geve
 For running the tests with a more specific usage and invocation, please refer to the `pytest <https://docs.pytest.org/en/stable/usage.html>`__ documentation.
 
 
+We run the test suite in the GitHub CI but only a subset of them since not all are needed and sometimes they suffer from rate limiting. We have some special settings to choose what tests are executed:
+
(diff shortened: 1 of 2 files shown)


docs/contribute.rst
```diff
@@ -398,6 +398,12 @@ We require this wrapper as a drop-in replacement of pytest due to quirks of geve
 For running the tests with a more specific usage and invocation, please refer to the `pytest <https://docs.pytest.org/en/stable/usage.html>`__ documentation.
 
 
+We run the test suite in the GitHub CI but only a subset of them since not all are needed and sometimes they suffer of rate limiting. We have some special settings to choose what tests are executed:
```
LefterisJP, line 401:
```suggestion
We run the test suite in the GitHub CI but only a subset of them since not all are needed and sometimes they suffer from rate limiting. We have some special settings to choose what tests are executed:
```

docs/contribute.rst
```diff
@@ -398,6 +398,12 @@ We require this wrapper as a drop-in replacement of pytest due to quirks of geve
...
+
+- ``[skip py tests]`` will not run the python backend tests regardless of the change status of the monitored directory.
```
LefterisJP, line 403:
```suggestion
- ``[skip py tests]`` will not run the python backend tests regardless of whether the backend code has been touched.
```

LefterisJP (approved):

PR #5311, opened 2022-12-28
Add name field to the balancer tokens

yabirgb:
The logic in the balancer module was creating pool tokens (`BPT`) but was not setting the name of those. The side effect of this was that when they were being used for balance queries "basic information" was missing and this was raising unexpected errors (handled already in other PR).


LefterisJP (approved):
LGTM

PR #5312, opened 2022-12-28
feat: adjust liquity staking UI

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/display/AmountCurrency.vue
+++ b/frontend/app/src/components/display/AmountCurrency.vue
@@ -1,6 +1,8 @@
 <script setup lang="ts">
-import { type PropType } from 'vue';
+import { type ComputedRef, type PropType } from 'vue';
 import { type Currency } from '@/types/currencies';
+import { useAssetInfoRetrieval } from '@/store/assets/retrieval';
+import { useAssetCacheStore } from '@/store/assets/asset-cache';
 
 const props = defineProps({
   currency: { required: true, type: Object as PropType<Currency> },
@@ -21,7 +23,7 @@ const props = defineProps({
   }
 });
 
-const { assetPadding, showCurrency, currency } = toRefs(props);
+const { asset, assetPadding, showCurrency, currency } = toRefs(props);
 const assetStyle = computed(() => {
   if (!get(assetPadding)) {
     return {};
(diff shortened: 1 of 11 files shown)


yabirgb:
functionality looks okey. Only request is that for consistency maybe you can put ETH rewards in the same position for both columns
image

So on the left LQTY and then ETH and in the right LUSD and then ETH

yabirgb (approved):
Approving on my side, functionality looks good. Needs a review of the code by kelsos still

frontend/app/src/components/display/AmountCurrency.vue
```diff
@@ -45,28 +47,43 @@ const displayAsset = computed(() => {
 
   return '';
 });
+
+const { assetSymbol } = useAssetInfoRetrieval();
+
+const symbol: ComputedRef<string> = computed(() => {
+  if (get(asset)) {
+    return get(assetSymbol(get(asset)));
```
kelsos, line 55: If `assetSymbol` takes a `MaybeRef` we should pass the asset directly

frontend/app/src/components/helper/NftDetails.vue
```diff
@@ -44,7 +44,7 @@ const loading = isLoading(Section.NON_FUNGIBLE_BALANCES);
 
 const { isPending } = useAssetCacheStore();
 const isNftDetailLoading: ComputedRef<boolean> = computed(
-  () => get(loading) || get(isPending(get(identifier)))
+  () => get(loading) || get(isPending(identifier))
```
kelsos, line 47: In my open PR I will be introducing https://vueuse.org/functions.html#category=%40Math so that we can use the `logicOr`/ `logicAnd` etc between refs

frontend/app/src/components/staking/liquity/LiquityStake.vue
```diff
@@ -1,12 +1,13 @@
...
 import StatCard from '@/components/display/StatCard.vue';
 
 defineProps({
-  stakings: {
-    required: true,
-    type: Array as PropType<AssetBalance[]>
+  stake: {
+    required: false,
+    type: Object as PropType<LiquityStakingDetail | null>,
+    default: null
```
kelsos, line 10: This is not something for this PR but for going forward can we slowly start doing:
```typescript
defineProps<{ stake: LiquityStakingDetail | null}>()
```

and if there are defaults we can use [`withDefaults`](https://vuejs.org/api/sfc-script-setup.html#default-props-values-when-using-type-declaration)

So it would look like this:
```typescript
withDefaults(
  defineProps<{ stake: LiquityStakingDetail | null}>(),
  { stake: null }
)

kelsos (approved):
We can merge, the only comment I have is minor and we can also fix in another PR

PR #5313, opened 2022-12-28
Adjust liquity balance query

yabirgb:
Fix an error when querying the staking balances for the aggregated view and also include the liquity stability pool balances

### DIFF
--- a/rotkehlchen/chain/aggregator.py
+++ b/rotkehlchen/chain/aggregator.py
@@ -56,6 +56,7 @@ from rotkehlchen.constants.assets import (
     A_ETH2,
     A_KSM,
     A_LQTY,
+    A_LUSD,
 )
 from rotkehlchen.constants.misc import ONE, ZERO
 from rotkehlchen.constants.resolver import ethaddress_to_identifier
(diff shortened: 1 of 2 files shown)


rotkehlchen/chain/aggregator.py
```diff
@@ -997,12 +998,20 @@ def _add_eth_protocol_balances(self, eth_balances: DefaultDict[ChecksumEvmAddres
...
                 collateral = deposits.collateral.balance
                 eth_balances[address].assets[A_ETH] += collateral
+
             # Get staked amounts
             liquity_staked = liquity_module.liquity_staking_balances(
                 addresses=self.queried_addresses_for_module('liquity'),
             )
             for address, staked_info in liquity_staked.items():
-                eth_balances[address].assets[A_LQTY] += staked_info['staked'].amount
+                eth_balances[address].assets[A_LQTY] += staked_info['staked'].balance
```
LefterisJP, line 1007: you need test/s for this. Tests should have caught it before.

LefterisJP (approved):
As discussed in discord, merging to unblock @kelsos in develop. But this test needs to be written as it would have noticed the problem much earlier.

PR #5314, opened 2022-12-29
Don't query trades for locations ignored

### LINKED ISSUE #5265: Querying trades of a single exchange that is ignored for sync should not pull data
## Problem Definition

rotki has a setting for each CEX that is meant to ignore syncing data with that exchange. The problem is that in the trades querying view, the frontend asks trades for each exchange separately.

In our code the function that decides whether data needs to be pulled from the exchange respects the setting when it concerns all exchanges (it iterates only connected exchanges that are not ignored), but not when querying a single one directly.

https://github.com/rotki/rotki/blob/62540a0e094b4bf9f5be87876b84dae99dc31300/rotkehlchen/history/events.py#L125-L127

This needs to change.

### System Description
Operating system: Any 
Rotki version: 1.26.2

## Task

- [ ] Make it so that the ignore_sync setting is respected when querying a single exchange.
- [ ] Write a test for it
- [ ] See if there is any other places we should respec this setting and we are not doing so.

yabirgb:
Closes #5265 

I've checked other places where trades are queried and they properly ignore the excluded locations. Added a test for the task manager to verify that the logic doesn't change there

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,7 @@ Changelog
 * :bug:`-` Connecting to substrate nodes will no longer timeout prematurely for systems with slow connections.
 * :bug:`-` Transfers between tracked accounts will now have a correct label in the UI.
 * :bug:`-` Users will be able to finish balance queries if they have assets with missing information.
+* :bug:`-` Visiting the trades page won't query now trades from exchanges that are ignored.
 * :bug:`5038` Premium users with big databases should no longer see the error: "Upload data to server died with exception: database plaintext is locked"
 * :bug:`-` Tokens added by the Balancer module will now have the name field correctly set.
 * :bug:`-` If a user removes the API keys for an exchange, actions on that exchange will no longer be excluded from PnL reports.
--- a/rotkehlchen/tests/api/test_exchanges.py
+++ b/rotkehlchen/tests/api/test_exchanges.py
@@ -1,6 +1,7 @@
 import os
 import random
 from http import HTTPStatus
+from typing import TYPE_CHECKING
 from unittest.mock import patch
 from urllib.parse import urlencode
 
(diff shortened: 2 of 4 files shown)


docs/changelog.rst
```diff
@@ -5,6 +5,7 @@ Changelog
 * :bug:`-` Connecting to substrate nodes will no longer timeout prematurely for systems with slow connections.
 * :bug:`-` Transfers between tracked accounts will now have a correct label in the UI.
 * :bug:`-` Users will be able to finish balance queries if they have assets with missing information.
+* :bug:`-` Visiting the trades page won't query now trades from exchanges that are ignored.
```
LefterisJP, line 8:
```suggestion
* :bug:`5265` Visiting the trades page will no longer query trades from ignored exchanges.
```

rotkehlchen/tests/api/test_exchanges.py
```diff
@@ -643,6 +649,34 @@ def assert_okay(response):
             api_url_for(server, "tradesresource") + '?' + urlencode(data))
         assert_okay(response)
 
+    # check that for poloniex we have information
+    with setup.binance_patch, setup.polo_patch:
+        response = requests.get(
+            url=api_url_for(server, "tradesresource"),
```
LefterisJP, line 655:
```suggestion
            url=api_url_for(server, 'tradesresource'),
```

rotkehlchen/tests/api/test_exchanges.py
```diff
@@ -643,6 +649,34 @@ def assert_okay(response):
...
+    assert len(result['entries']) != 0
+
+    # exclude poloniex as location and delete the trades from there
+    db = server.rest_api.rotkehlchen.data.db
+    poloniex_exchange = server.rest_api.rotkehlchen.exchange_manager.connected_exchanges[Location.POLONIEX][0]  # noqa: E501
+    with db.user_write() as cursor:
+        db.set_settings(cursor, ModifiableDBSettings(
+            non_syncing_exchanges=[poloniex_exchange.location_id()],
+        ))
+        db.purge_exchange_data(cursor, Location.POLONIEX)
```
LefterisJP, line 669: Why delete the data? Shouldn't the test check that they are not queries if it's ignored but data is still there?

rotkehlchen/tests/api/test_exchanges.py
```diff
@@ -643,6 +649,34 @@ def assert_okay(response):
...
+
+    # query the API and check that information is not queried again
+    with setup.binance_patch, setup.polo_patch:
+        response = requests.get(
+            url=api_url_for(server, "tradesresource"),
```
LefterisJP, line 674:
```suggestion
            url=api_url_for(server, 'tradesresource'),
```

LefterisJP (commented):
looks okay, got some questions/nitpicks

LefterisJP (approved):
Will merge.

As discussed in discord:

- Comments will be addressed in another PR
- The failing frontend test may be a bittrex api problem - we will have to see. But happened in another bugfixes PR too

PR #5315, opened 2022-12-29
fix: update balance refresh mechanism after detect tokens

### LINKED ISSUE #5158: Detect all tokens is a no-op
## Problem Definition

The detect all tokens button does not do the work it's supposed to do in v1.26.1

It's a no-op. The reason is that it's called by the frontend with `only_cache=True` so only the last known detected tokens are returned. The intention was to skip cache and actually detect tokens so it should be `only_cache=False`.

Additionally the backend should not count it as a token detection.

## Task

- [ ] Do not use `only_cache=True` on that button
- [ ] If `detect_tokens` is called with `only_cache=True` the last detection timestamp **should not be** updated

lukicenturi:
Closes #5158

## Checklist

### DIFF
--- a/frontend/app/src/store/blockchain/tokens.ts
+++ b/frontend/app/src/store/blockchain/tokens.ts
@@ -1,3 +1,4 @@
+import { Blockchain } from '@rotki/common/lib/blockchain';
 import { set } from '@vueuse/core';
 import isEqual from 'lodash/isEqual';
 import { ComputedRef, Ref } from 'vue';
@@ -7,6 +8,7 @@ import {
 } from '@/services/balances/types';
 import { api } from '@/services/rotkehlchen-api';
 import { useIgnoredAssetsStore } from '@/store/assets/ignored';
+import { useBlockchainBalancesStore } from '@/store/blockchain/balances';
 import { useEthBalancesStore } from '@/store/blockchain/balances/eth';
 import { useTasks } from '@/store/tasks';
 import { TaskMeta } from '@/types/task';
@@ -100,6 +102,20 @@ export const useBlockchainTokensStore = defineStore('blockchain/tokens', () => {
       return;
     }
     await fetchDetectedTokens();
+  });
+
+  const { isTaskRunning } = useTasks();
+  const isDetecting = isTaskRunning(TaskType.FETCH_DETECTED_TOKENS);
+
+  const { fetchBlockchainBalances } = useBlockchainBalancesStore();
+
+  watch(isDetecting, async (isDetecting, wasDetecting) => {
+    if (wasDetecting && !isDetecting) {
+      await fetchBlockchainBalances({
+        blockchain: Blockchain.ETH,
+        ignoreCache: true
+      });
+    }
   });
 
   return {
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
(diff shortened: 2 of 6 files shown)


frontend/app/src/store/blockchain/tokens.ts
```diff
@@ -102,6 +104,20 @@ export const useBlockchainTokensStore = defineStore('blockchain/tokens', () => {
     await fetchDetectedTokens();
   });
 
+  const { isTaskRunning } = useTasks();
+  const isDetecting = isTaskRunning(TaskType.FETCH_DETECTED_TOKENS);
+
+  const { fetchBlockchainBalances } = useBlockchainBalancesStore();
+
+  watch(isDetecting, async (curr, prev) => {
```
kelsos, line 112: do you think we can adjust the names here to isDetecting & wasDetecting for clarity?

frontend/app/src/store/blockchain/tokens.ts
```diff
@@ -102,6 +104,20 @@ export const useBlockchainTokensStore = defineStore('blockchain/tokens', () => {
...
+    if (prev && !curr) {
+      await fetchBlockchainBalances({
```
kelsos, line 114: So this will trigger the refresh as soon as the task is done correct?

frontend/.eslintcache
```diff
@@ -0,0 +1 @@
...arningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"3917","messages":"3918","suppressedMessages":"3919","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"3920","messages":"3921","suppressedMessages":"3922","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"3923","messages":"3924","suppressedMessages":"3925","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"3926","messages":"3927","suppressedMessages":"3928","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"3929","messages":"3930","suppressedMessages":"3931","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"3932","messages":"3933","suppressedMessages":"3934","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"3935","messages":"3936","suppressedMessages":"3937","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarning
```
kelsos, line 1: we should not include this in the PR. can you add it to `.gitignore` instead

frontend/app/src/store/blockchain/tokens.ts
lukicenturi: yeah correct

kelsos:
@lukicenturi don't forget a changelog entry here

kelsos (approved):
At least from a quick look/check it seems to work as expected.

Failed e2e test is due to bittrex.

kelsos (approved):
lgtm

PR #5316, opened 2022-12-29
Modify API endpoints to query all EVM transactions

LefterisJP:
- Modify all api endpoint to query all EVM transactions.
- Modifications also in API to request decoding of EVM transactions per chain
- Modification also in API to purge EVM transactions per chain.

For understanding of changes check the api.rst docs

Also added a test ... which should be mocked soon


LefterisJP:
Merging since backend tests passed. But this will also need frontend work to adjust to the changes.

PR #5317, opened 2022-12-30
Adjust addressbook in frontend

lukicenturi:
Closes #(issue_number)

## Checklist


lukicenturi:
The only step remain is to disable the addition of the addressbook for non-compatible chain

kelsos:
@lukicenturi other than the frontend changes the backend PR is still not merged right?

frontend/app/src/types/frontend-settings.ts
```diff
@@ -176,7 +176,7 @@ export const FrontendSettings = z.object({
       Defaults.DEFAULT_VERSION_UPDATE_CHECK_FREQUENCY
     )
   ),
-  enableEthNames: z.boolean().default(true)
+  enableAliasNames: z.boolean().default(true)
```
kelsos, line 179: The only problem with renaming the property is that the users will have to set it from scratch since we don't do any migration.

frontend/app/src/types/notes.ts
```diff
@@ -50,7 +50,7 @@ export enum NoteLocation {
   PROFIT_LOSS_REPORTS = 'PROFIT_LOSS_REPORTS',
   ASSETS = 'ASSETS',
   PRICE_MANAGER = 'PRICE_MANAGER',
-  ETH_ADDRESS_BOOK_MANAGER = 'ETH_ADDRESS_BOOK_MANAGER',
+  ADDRESS_BOOK_MANAGER = 'ADDRESS_BOOK_MANAGER',
```
kelsos, line 53: What happens if you had notes in the old location?

frontend/app/package.json
```diff
@@ -85,7 +85,7 @@
     "c8": "7.12.0",
     "cross-env": "7.0.3",
     "csv": "6.2.5",
-    "cypress": "12.2.0",
```
kelsos, line 88: This is part of the dependency update I did so you can remove it from here

frontend/app/src/store/blockchain/accounts/addresses-names.ts
```diff
@@ -0,0 +1,252 @@
...
+  type AddressBookSimplePayload,
+  type EthNames
+} from '@/types/eth-names';
+import { type TaskMeta } from '@/types/task';
+import { TaskType } from '@/types/task-type';
+import { uniqueObjects, uniqueStrings } from '@/utils/data';
+import { logger } from '@/utils/logging';
+import { isValidEthAddress } from '@/utils/text';
+
+export const useAddressesNamesStore = defineStore('addressesNames', () => {
```
kelsos, line 21: This I have to have a more detailed look :)

kelsos (commented):
So these are the comments I have so far, I need to also checkout and check it.

frontend/app/src/components/address-book-manager/AddressBookManagement.vue
```diff
@@ -47,12 +53,13 @@
...
   set(showForm, false);
   set(form, emptyForm());
+  set(enableForAllChain, false);
 };
 
 const valid = ref<boolean>(false);
 const showForm = ref(false);
 const editMode = ref<boolean>(false);
-const form = ref<EthNamesPayload>(emptyForm());
+const form = ref<AddressBookPayload>(emptyForm());
```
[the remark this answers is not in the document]
lukicenturi: formPayload should be better (?)

frontend/app/src/types/notes.ts
lukicenturi: hmm yeah they won't see it. Should we keep the old name ?

frontend/app/src/types/frontend-settings.ts
lukicenturi: should we keep the old state name?

lukicenturi: I was thinking that it's okay since the setting is easy to toggle, and also the majority will keep the value true.

frontend/app/src/types/notes.ts
kelsos: Or find a way to migrate from the old key to the new one.

lukicenturi: I think it's easier to migrate from backend.
Otherwise we can change only the enum name, but keep the value.

frontend/app/src/components/accounts/blockchain/ChainSelect.vue
```diff
@@ -46,51 +45,72 @@ const chains: SupportedChain[] = [
...
-  blockchain: {
-    required: true,
-    type: String as PropType<Blockchain>
-  },
-  disabled: {
-    required: true,
-    type: Boolean
+const props = withDefaults(
+  defineProps<{
+    modelValue?: Blockchain | null;
```
kelsos, line 50: it doesn't have to be `modelValue`,

`modelValue` is the placeholder you can replace with anything 

so for blockchain the pattern would be `blockchain` for the value and `update:blockchain` for the emit as it was before. 

and for vue3 this can then be used as a v-model -> `v-model:blockchain`

frontend/app/src/components/display/AccountDisplay.vue
```diff
@@ -28,22 +35,35 @@ const address = computed<string>(() => {
...
         <v-col cols="auto" class="pr-2">
           <v-avatar left size="28px" class="mr-0">
-            <asset-icon size="24px" :identifier="account.chain" />
+            <asset-icon
+              v-if="account.chain"
+              size="24px"
+              :identifier="account.chain"
+            />
+            <v-tooltip v-else top>
+              <template #activator="{ on }">
```
kelsos, line 62: can we rename one of the two? to avoid shadowing?

frontend/app/src/locales/en.json
```diff
@@ -835,6 +835,77 @@
...
+      }
+    },
+    "dialog": {
+      "add_title": "Add address book",
+      "edit_title": "Edit address book"
+    },
+    "form": {
+      "labels": {
+        "address": "Address",
+        "for_all_chain": "Save this name for all chain"
```
kelsos, line 878: chain/chains

frontend/app/src/locales/en.json
```diff
@@ -835,6 +835,77 @@
...
+      "fetch": {
+        "error": {
+          "message": "Failed to retrieve the address book: {message}",
+          "title": "Querying address book"
+        }
+      }
+    },
+    "dialog": {
+      "add_title": "Add address book",
+      "edit_title": "Edit address book"
```
kelsos, line 873: can you add entry at the end of those two?

frontend/app/src/components/accounts/blockchain/ChainSelect.vue
lukicenturi: but since this component sole purpose is to input blockchain (no other thing to bind), can't we use the default name `modelValue` instead `blockchain`, so later we just use `v-model` instead of `v-model:blockchain`

frontend/app/src/components/helper/HashLink.vue
```diff
@@ -105,9 +114,12 @@ const { href, onLinkClick } = useLinks(url);
...
   <div class="d-flex flex-row shrink align-center">
-    <v-avatar v-if="showIcon" size="24" class="mr-2">
-      <v-img :src="makeBlockie(displayText)" />
-    </v-avatar>
+    <!--    TODO: move makeBlockie to local, and make the size adjustable-->
+    <!--    <span>-->
+    <!--      <v-avatar v-if="showIcon && !tx" size="24" class="ml-n2 mr-2">-->
+    <!--        <v-img :src="makeBlockie(displayText)" />-->
+    <!--      </v-avatar>-->
+    <!--    </span>-->
```
kelsos, line 122: A bit more info about this? 

Btw since we talked about the performance of base64. I think in the past in another project I was working on we did a small LRU cache so the last 30-40 blockies where actually stored in memory instead of being re-calculated every time. This certainly improved performance.

Maybe we can do the same here.

lukicenturi: Yeah exactly what I was thinking, store them in memory. But I will do it in another PR, since I think it little bit complicated, and has nothing to do with this PR.

kelsos (approved):
lgtm @lukicenturi thanks.

PR #5318, opened 2022-12-30
Add llamanode as node and migration

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/tests/data_migrations/test_migrations.py
+++ b/rotkehlchen/tests/data_migrations/test_migrations.py
@@ -6,6 +6,7 @@ import pytest
 
 from rotkehlchen.constants import ONE
 from rotkehlchen.constants.assets import A_BTC, A_ETH
+from rotkehlchen.constants.misc import ZERO
 from rotkehlchen.data_migrations.manager import (
     MIGRATION_LIST,
     DataMigrationManager,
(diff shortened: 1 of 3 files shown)


LefterisJP (commented):
Looking at it though this is not what I meant. We can't possibly do the same migration as last time.

This is silly and makes it hard to do any additions. The comparison to defaults was supposed to be an one time thing and not something we do each migration.

This way we would have to keep every version's defaults as a json and this create serious difficulties in testing and updating stuff as you see from your PR. We can't do that each time we need to change the defaults.

What's the problem in doing a super simple?
1. Get all ethereum nodes.
2. Check if any of those endpoints matches the rpc of the llama one
    a) If yes do nothing
    b) if not add it to the nodes and reweigh

rotkehlchen/tests/data_migrations/test_migrations.py
```diff
@@ -400,3 +400,32 @@ def test_migration_6_own_node(rotkehlchen_api_server):
...
+
+    nodes = database.get_web3_nodes(blockchain=SupportedBlockchain.ETHEREUM)
+    # check that weight is correct for nodes
+    assert sum((node.weight for node in nodes)) == ONE
+    llama_node_in_db = False
+    for node in nodes:
+        if node.node_info.endpoint == 'https://eth.llamarpc.com':
+            llama_node_in_db = True
+            break
+    assert llama_node_in_db is True
```
LefterisJP, line 431: no need to change this, just letting you know that this is the one and only use case where `for ... else` loop construct makes sense.

```suggestion
    for node in nodes:
        if node.node_info.endpoint == 'https://eth.llamarpc.com':
            break
    else:  # not found
        raise AssertionError('llamarpc should have been found')
```

LefterisJP (approved):
LGTM so long as tests are fine.

PR #5325, opened 2022-12-30
Fix error during globaldb upgrades

yabirgb:
Fixes and error where the db version was not properly set. Also adds a fix for the assets collections upgrade where it handles the case of assets missings during the collections creation


rotkehlchen/globaldb/upgrades/v3_v4.py
```diff
@@ -256,17 +258,69 @@ def _add_optimism_contracts(cursor: 'DBCursor', eth_scan_abi_id: int, multicall_
...
+
+    We need to commit the changes here because otherwise the packaged globaldb is busy and we
+    can't detach it. At this point all the information in the upgrade should be fine.
+
+    TODO: This approach assumes that the schema in the packaged globaldb and the schema in v3 are
+    the same. When this doesn't hold anymore we should keep a copy of the old schema or ask users
+    to use an intermediate version to upgrade.
+    """
+    packaged_db_path = root_dir / 'data' / 'global.db'
+    identifiers_quotes = ",".join("?" * len(assets_ids))
```
LefterisJP, line 279:
```suggestion
    identifiers_quotes = ','.join('?' * len(assets_ids))
```

rotkehlchen/globaldb/upgrades/v3_v4.py
```diff
@@ -256,17 +258,69 @@ def _add_optimism_contracts(cursor: 'DBCursor', eth_scan_abi_id: int, multicall_
     log.debug('Exit _add_optimism_contracts')
 
 
+def _copy_assets_from_packaged_db(
+        cursor: 'DBCursor',
+        assets_ids: list[str],
+        root_dir: Path,
+) -> None:
+    """
+    Copy missing assets from the packaged globaldb to the globaldb
```
LefterisJP, line 267: Should this be only in this upgrade? Shouldn't we do this more often also? Kind of a check, if a user has done a booboo and deleted an asset that should be constant and in the DB?

rotkehlchen/globaldb/upgrades/v3_v4.py
```diff
@@ -256,17 +258,69 @@ def _add_optimism_contracts(cursor: 'DBCursor', eth_scan_abi_id: int, multicall_
...
     log.debug('Enter _add_assets_collections')
     root_dir = Path(__file__).resolve().parent.parent.parent
+    # regex used to identify assets that will be added to the collections
+    asset_regex = re.compile(r'eip155[a-zA-F0-9:\/]+')
     with open(root_dir / 'data' / 'assets_collections.sql') as f:
         raw_sql_sentences = f.read()
         per_table_sentences = raw_sql_sentences.split('\n\n')
         for sql_sentences in per_table_sentences:
+            # check if we are adding the assets
+            if 'multiasset_mappings' in sql_sentences:
```
LefterisJP, line 305: Why not just split the sql multiassets mapping into its own sql file and just do this after the for loop? This way you separate concerns better and don't have to do if checks for each loop iteration

rotkehlchen/globaldb/upgrades/v3_v4.py
```diff
@@ -256,17 +258,69 @@ def _add_optimism_contracts(cursor: 'DBCursor', eth_scan_abi_id: int, multicall_
...
+                # if not copy them from the packaged db
+                all_assets = asset_regex.findall(sql_sentences)
+                assets_to_add = []
+                for asset_identifier in all_assets:
+                    cursor.execute(
+                        'SELECT COUNT(*) FROM assets where identifier=?',
+                        (asset_identifier,),
+                    )
+                    if cursor.fetchone()[0] == 0:
+                        assets_to_add.append(asset_identifier)
```
LefterisJP, line 316: silly goose :smile_cat:!
You don't need to iterate in python and make as many DB queries as there is assets. You can just do 1 query.

The below is untested pseudocode but I guess you get the idea


```suggestion
                cursor.execute('SELECT identifier from assets WHERE identifier IN (?, ?, ? .... )', all_assets)
                existing_ids = [x[0] for x in cursor]
                assets_to_add = set(all_assets) - set(existing_ids)
```

rotkehlchen/globaldb/upgrades/v3_v4.py
```diff
@@ -256,17 +258,69 @@ def _add_optimism_contracts(cursor: 'DBCursor', eth_scan_abi_id: int, multicall_
...
+
+                try:
+                    _copy_assets_from_packaged_db(cursor, assets_to_add, root_dir)
```
LefterisJP, line 319: use kwargs

rotkehlchen/tests/unit/test_globaldb_upgrades.py
```diff
@@ -287,3 +287,25 @@ def test_unfinished_upgrades(globaldb: GlobalDBHandler):
...
+@pytest.mark.parametrize('globaldb_upgrades', [[]])
+@pytest.mark.parametrize('globaldb_version', [2])
+@pytest.mark.parametrize('target_globaldb_version', [2])
+@pytest.mark.parametrize('reload_user_assets', [False])
+def test_applying_all_upgrade(globaldb):
+    """Test globalDB upgrade from v2 to latest"""
+    # Check the state before upgrading
+    assert globaldb.get_setting_value('version', None) == 2
+    with globaldb.conn.cursor() as cursor:
+        assert cursor.execute('SELECT COUNT(*) from assets WHRE identifier="eip155:/erc20:0x32c6fcC9bC912C4A30cd53D2E606461e44B77AF2"')[0] == 0  # noqa: E501
```
LefterisJP, line 301: How is this test even passing?
```suggestion
        assert cursor.execute('SELECT COUNT(*) from assets WHERE identifier="eip155:/erc20:0x32c6fcC9bC912C4A30cd53D2E606461e44B77AF2"')[0] == 0  # noqa: E501
```

rotkehlchen/tests/unit/test_globaldb_upgrades.py
```diff
@@ -287,3 +287,25 @@ def test_unfinished_upgrades(globaldb: GlobalDBHandler):
     assert globaldb.get_setting_value('ongoing_upgrade_from_version', -1) == -1
     with globaldb.conn.read_ctx() as cursor:
         assert cursor.execute('SELECT value FROM settings WHERE name="is_backup"').fetchone()[0] == 'Yes'  # noqa: E501
+
+
+@pytest.mark.parametrize('globaldb_upgrades', [[]])
+@pytest.mark.parametrize('globaldb_version', [2])
+@pytest.mark.parametrize('target_globaldb_version', [2])
+@pytest.mark.parametrize('reload_user_assets', [False])
+def test_applying_all_upgrade(globaldb):
```
LefterisJP, line 296: we also need in between tests. To test the change you made we need to see that v2->v3 stops at v3 and does not end up putting v4, which is what the previous logic was doing. Can you make sure this is tested?

LefterisJP (commented):
Some comments.

rotkehlchen/tests/unit/test_globaldb_upgrades.py
yabirgb: yes, I've added checks to ensure that the version is properly set

rotkehlchen/globaldb/upgrades/v3_v4.py
yabirgb: indeed much better

rotkehlchen/globaldb/upgrades/v3_v4.py
yabirgb: went with the other approach improving the check

rotkehlchen/globaldb/upgrades/v3_v4.py
yabirgb: in this case it is required. If a user deletes a constant asset is more of his fault I would say. We could do this it nevertheless but I guess someone wouldn't delete an asset if there is no reason for that or by mistake

rotkehlchen/globaldb/upgrades/manager.py
```diff
@@ -99,7 +99,9 @@ def maybe_upgrade_globaldb(
             )
             write_cursor.execute(
                 'INSERT OR REPLACE INTO settings(name, value) VALUES(?, ?)',
-                ('version', str(GLOBAL_DB_VERSION)),
+                ('version', str(to_version)),
```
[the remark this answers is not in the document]
yabirgb: Because we had only one possible value and is in this upgrade where this error appears.

rotkehlchen/globaldb/upgrades/v3_v4.py
LefterisJP: What other approach? I don't understand. You still have an if check here

rotkehlchen/tests/unit/test_globaldb_upgrades.py
```diff
@@ -287,3 +289,25 @@ def test_unfinished_upgrades(globaldb: GlobalDBHandler):
...
+    with globaldb.conn.cursor() as cursor:
+        assert cursor.execute('SELECT COUNT(*) from assets WHERE identifier="eip155:/erc20:0x32c6fcC9bC912C4A30cd53D2E606461e44B77AF2"')[0] == 0  # noqa: E501
+
+    maybe_upgrade_globaldb(
+        connection=globaldb.conn,
+        global_dir=globaldb._data_directory / 'global_data',
+        db_filename=GLOBAL_DB_FILENAME,
+    )
+
+    assert globaldb.get_setting_value('version', None) == 4
```
LefterisJP, line 311: Since this should be latest replace `4` with `GLOBAL_DB_VERSION`

rotkehlchen/globaldb/upgrades/v3_v4.py
LefterisJP: Sure I am just mentioning it so we keep in mind it may make sense to do this also periodically at some point.

rotkehlchen.spec
```diff
@@ -101,6 +101,8 @@ a = Entrypoint(
         ('rotkehlchen/data/globaldb_v2_v3_assets.sql', 'rotkehlchen/data'),
         ('rotkehlchen/data/nodes.json', 'rotkehlchen/data'),
         ('rotkehlchen/data/nodes_as_of_1-26-1.json', 'rotkehlchen/data'),
+        ('rotkehlchen/data/populate_asset_collections.sql', 'rotkehlchen/data'),
```
LefterisJP, line 104: you are repeating the names here

rotkehlchen/globaldb/upgrades/v3_v4.py
```diff
@@ -256,19 +258,75 @@ def _add_optimism_contracts(cursor: 'DBCursor', eth_scan_abi_id: int, multicall_
...
+    identifiers_quotes = ','.join('?' * len(assets_ids))
+    cursor.execute(f'ATTACH DATABASE "{packaged_db_path}" AS packaged_db;')
+    cursor.execute(f'INSERT INTO assets SELECT * FROM packaged_db.assets WHERE identifier IN ({identifiers_quotes});', assets_ids)  # noqa: E501
+    cursor.execute(f'INSERT INTO evm_tokens SELECT * FROM packaged_db.evm_tokens WHERE identifier IN ({identifiers_quotes});', assets_ids)  # noqa: E501
+    cursor.execute(f'INSERT INTO common_asset_details SELECT * FROM packaged_db.common_asset_details WHERE identifier IN ({identifiers_quotes});', assets_ids)  # noqa: E501
+    cursor.execute('COMMIT;')
+    cursor.execute('DETACH DATABASE packaged_db;')
+
+
+def _populate_asset_collection(cursor: 'DBCursor', root_dir: Path) -> None:
```
LefterisJP, line 288:
```suggestion
def _populate_asset_collections(cursor: 'DBCursor', root_dir: Path) -> None:
```

setup.py
```diff
@@ -36,6 +36,8 @@
             'data/globaldb_v2_v3_assets.sql',
             'data/nodes.json',
             'data/nodes_as_of_1-26-1.json',
+            'data/populate_asset_collections.sql',
```
LefterisJP, line 39: same here ... man be careful

LefterisJP (commented):
sloppy mistakes

LefterisJP (approved):
LGTM!

PR #5328, opened 2022-12-31
Add OKX exchange integration

anukul:
Added integration for asset balances, trades and deposits & withdrawals

https://user-images.githubusercontent.com/44864521/210120836-18d40233-f3fb-4609-b0d2-8cc3a44cdaa1.mov

Note:

- Used a throwaway account for exchange fixture, since all queries must be authenticated with API credentials (including list of assets)
- Only populated a small subset of assets (those also found in other exchange mappings), would prefer to discuss before moving further on this.


rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,410 @@
...
+            self,
+            start_ts: Timestamp,
+            end_ts: Timestamp,
+    ) -> list[LedgerAction]:
+        return []  # noop for okx
+
+    def trade_from_okx(self, raw_trade: dict[str, Any]) -> Optional[Trade]:
+        try:
+            # okx api returns timestamp with milliseconds
+            timestamp = Timestamp(int(raw_trade['cTime'][:-3]))
```
yabirgb, line 309: we have a function for this `ts_ms_to_sec`

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,410 @@
...
+            [base_asset_str, quote_asset_str] = raw_trade['instId'].split('-')
```
yabirgb, line 310: this can raise an error if for some reason the format is not the expected. It should be handled

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,410 @@
...
+
+    def asset_movement_from_okx(
+            self,
+            raw_movement: dict[str, Any],
+            category: AssetMovementCategory,
+    ) -> Optional[AssetMovement]:
+        try:
+            tx_hash = raw_movement['txId']
+            # okx api returns timestamp with milliseconds
+            timestamp = Timestamp(int(raw_movement['ts'][:-3]))
```
yabirgb, line 367: use `ts_ms_to_sec` here too

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,410 @@
...
+                AssetMovementCategory.DEPOSIT,
+            )
+            if movement:
+                movements.append(movement)
+        for raw_movement in withdrawals:
+            movement = self.asset_movement_from_okx(
+                raw_movement,
+                AssetMovementCategory.WITHDRAWAL,
+            )
+            if movement:
```
yabirgb, line 294: we try to be as explicit as possible with the expressions

```suggestion
            if movement is not None:
```

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,410 @@
...
+        withdrawals = self._paginated_api_query('withdrawals', 'ts')
+
+        movements = []
+        for raw_movement in deposits:
+            movement = self.asset_movement_from_okx(
+                raw_movement,
+                AssetMovementCategory.DEPOSIT,
+            )
+            if movement:
+                movements.append(movement)
```
yabirgb, line 288:
```suggestion
            if movement is not None:
                movements.append(movement)
```

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,410 @@
...
+    def _generate_signature(self, timestamp: str, method: str, path: str, body: str) -> str:
+        prehash = timestamp + method.upper() + path + body
+        signature = hmac.new(self.secret, prehash.encode('utf-8'), hashlib.sha256)
+        return base64.b64encode(signature.digest()).decode('utf-8')
+
+    def _api_query(
+        self,
+        endpoint: Literal['currencies', 'balance', 'trades', 'deposits', 'withdrawals'],
+        options: Optional[dict] = None,
+    ) -> dict:
```
yabirgb, line 101: we use double indentation when splitting in lines the arguments

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,410 @@
...
+        """
+        raw_trades = self._paginated_api_query('trades', 'ordId', {
+            'start_ts': start_ts,
+            'end_ts': end_ts,
+        })
+
+        trades: list[Trade] = []
+        for raw_trade in raw_trades:
+            trade = self.trade_from_okx(raw_trade)
+            if trade:
```
yabirgb, line 257:
```suggestion
            if trade is not None:
```

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,410 @@
...
+
+    def query_online_income_loss_expense(
+            self,
+            start_ts: Timestamp,
+            end_ts: Timestamp,
+    ) -> list[LedgerAction]:
+        return []  # noop for okx
+
+    def trade_from_okx(self, raw_trade: dict[str, Any]) -> Optional[Trade]:
+        try:
```
yabirgb, line 307: can you add a docstring to this functions that process the information?

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,410 @@
...
+            # supports pagination
+            params.update({
+                'limit': options.get('limit', ''),
+                'after': options.get('after', ''),
+            })
+        else:
+            raise AssertionError(f'Unexpected {self.name} endpoint type: {endpoint}')
+
+        if len(params) != 0:
+            path += f'?{urlencode(params)}'
```
yabirgb, line 143: it would be better to use the functionality in requests itself https://requests.readthedocs.io/en/latest/user/quickstart/#passing-parameters-in-urls

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,410 @@
...
+            })
+        elif endpoint == 'withdrawals':
+            path = '/api/v5/asset/withdrawal-history'
+            # supports pagination
+            params.update({
+                'limit': options.get('limit', ''),
+                'after': options.get('after', ''),
+            })
+        else:
+            raise AssertionError(f'Unexpected {self.name} endpoint type: {endpoint}')
```
yabirgb, line 140: Since we use types here and the endpoint param is never set via variables I believe you can remove this error. Also is an exception that you are not handling in other places of the code

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,410 @@
...
+    ) -> bool:
+        changed = super().edit_exchange_credentials(api_key, api_secret, passphrase)
+        if api_key is not None:
+            self.session.headers.update({'OK-ACCESS-KEY': self.api_key})
+        if passphrase is not None:
+            self.session.headers.update({'OK-ACCESS-PASSPHRASE': self.passphrase})
+        return changed
+
+    def _generate_signature(self, timestamp: str, method: str, path: str, body: str) -> str:
+        prehash = timestamp + method.upper() + path + body
```
yabirgb, line 93: can you link the url where it explains how to build this one? for future references

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,410 @@
...
+            return False, 'Error validating API credentials'
+        return True, ''
+
+    def query_balances(self, **kwargs: Any) -> ExchangeQueryBalances:
+        """
+        https://www.okx.com/docs-v5/en/#rest-api-account-get-balance
+        """
+        response = self._api_query('balance')
+
+        assets_balance: DefaultDict[AssetWithOracles, Balance] = defaultdict(Balance)
```
yabirgb, line 209: why do you need a defaultdict instead of using a dict?

rotkehlchen/tests/exchanges/test_okx.py
```diff
@@ -0,0 +1,802 @@
...
+
+    warnings = mock_okx.msg_aggregator.consume_warnings()
+    errors = mock_okx.msg_aggregator.consume_errors()
+    assert len(warnings) == 0
+    assert len(errors) == 0
+
+
+def test_okx_query_trades(mock_okx: Okx):
+    # pylint: disable=unused-argument
+    def mock_okx_trades(method, url):
```
yabirgb, line 192:
```suggestion
    def mock_okx_trades(method, url):    # pylint: disable=unused-argument
```

rotkehlchen/tests/exchanges/test_okx.py
```diff
@@ -0,0 +1,802 @@
...
+            notes=None,
+        ),
+    ], (1609103082, 1672175105))
+
+    assert trades == expected_trades
+
+
+def test_okx_query_deposits_withdrawals(mock_okx: Okx):
+    # pylint: disable=unused-argument
+    def mock_okx_deposits_withdrawals(method, url):
```
yabirgb, line 655:
```suggestion
    def mock_okx_deposits_withdrawals(method, url):    # pylint: disable=unused-argument
```

rotkehlchen/tests/exchanges/test_okx.py
```diff
@@ -0,0 +1,802 @@
...
+            category=AssetMovementCategory.WITHDRAWAL,
+            timestamp=Timestamp(1670953159),
+            address='0x388c818ca8b9251b393131c08a736a67ccb19297',
+            transaction_id='0x9444b018c33c5adb58ee171bc18e61c56078495e37ae88833007a46c02b4552f',
+            asset=A_USDT,
+            amount=FVal(421.169831),
+            fee_asset=A_USDT,
+            fee=Fee(FVal(0.1)),
+            link='0x9444b018c33c5adb58ee171bc18e61c56078495e37ae88833007a46c02b4552f'),
+    ]
```
yabirgb, line 800: Should those be changed?

rotkehlchen/tests/exchanges/test_okx.py
```diff
@@ -0,0 +1,802 @@
...
+    assert exchange.name == 'okx1'
+
+
+def test_assets_are_known(mock_okx: Okx):
+    unsupported_assets: set[str] = set(UNSUPPORTED_OKX_ASSETS)
+    common_items = unsupported_assets.intersection(set(WORLD_TO_OKX.values()))
+    assert not common_items, f'OKX assets {common_items} should not be unsupported'
+
+    okx_assets = set()
+
```
yabirgb, line 37: Extra space
```suggestion
```

rotkehlchen/tests/exchanges/test_okx.py
```diff
@@ -0,0 +1,802 @@
...
+    currencies = mock_okx._query_currencies()
+    for currency in currencies['data']:
+        okx_assets.add(currency['ccy'])
+
+    sorted_assets = sorted(okx_assets)
+    for okx_asset in sorted_assets:
+        try:
+            _ = asset_from_okx(okx_asset)
```
yabirgb, line 45:
```suggestion
            asset_from_okx(okx_asset)
```

rotkehlchen/tests/exchanges/test_okx.py
```diff
@@ -0,0 +1,802 @@
...
+    common_items = unsupported_assets.intersection(set(WORLD_TO_OKX.values()))
+    assert not common_items, f'OKX assets {common_items} should not be unsupported'
+
+    okx_assets = set()
+
+    currencies = mock_okx._query_currencies()
+    for currency in currencies['data']:
+        okx_assets.add(currency['ccy'])
+
+    sorted_assets = sorted(okx_assets)
```
yabirgb, line 42: why do you need to sort them?

yabirgb (commented):
Thanks for the contribution. I checked the backend and someone else will check the frontend. Can you also add an entry in the `docs/changelog` file saying that support for OKX was added? Also wrote a question in discord (priv) about the tests

yabirgb:
Also @anukul linting is failing. Check it before pushing

rotkehlchen/exchanges/okx.py
anukul: I had started by using the `params` parameter, but changed to manually concatenating to path in order to use it to generate the signature. I admit it is a little unintuitive, but thought to use this approach after seeing it in implementations for the other exchanges.

I had also tried to directly use requests' approach of generating the URL, but it does not seem to expose that [function](https://github.com/psf/requests/blob/v2.28.1/requests/models.py#L481) for direct use and modifies the `url` attribute on the `Request` class.

There may be a way to initialize the request, use the generated path for the signature, and then attach that as a header, but it seems to diverge from the convention followed across the codebase.

Maybe this can be addressed in a subsequent PR for all instances?

rotkehlchen/tests/exchanges/test_okx.py
anukul: I have anonymized (by using random values) hopefully all instances of identifiable data - addresses, transaction hashes and unique identifiers for trades, deposits & withdrawals

rotkehlchen/tests/exchanges/test_okx.py
anukul: Oh.. I probably didn't need to do this. I think I left it in while copying this test from binance.  Sorry about that.

rotkehlchen/exchanges/okx.py
anukul: Copied the pattern used elsewhere, but I believe using a dict here throws a `KeyError` on this line below:

```py
assets_balance[asset] += Balance(
```
since this is the only way the dictionary is populated and it expects the key (or a default value) to exist.

rotkehlchen/exchanges/okx.py
anukul: Thanks for the heads up, fixed :)

rotkehlchen/exchanges/okx.py
anukul: I understand that static analysis may be able to highlight this error from the type annotation, but python will still accept this as valid code, and throw a different error after a request is made to an invalid endpoint.

How do you suggest I handle the `else` case? I see the following options:

1. Handle one of the listed endpoints in the `else` branch
2. Remove the `else` branch, and initialize `path` to an empty string

Alternatively, if we do keep this, the exception is currently thrown to the caller (in tests) and shown to the end user in the app.

LefterisJP:
Hey @anukul I am not reviewing yet but just a note. We don't merge develop back to the branch. Instead we always rebase feature branches on top of develop

lukicenturi:
The frontend part is okay.

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,418 @@
...
+        for currency_data in response['data'][0]['details']:
+            try:
+                asset = asset_from_okx(okx_name=currency_data['ccy'])
+            except (UnknownAsset, UnsupportedAsset):
+                continue
```
yabirgb, line 218: Shouldn't in this case an error message also be sent?

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,418 @@
...
+    ) -> list[LedgerAction]:
+        return []  # noop for okx
+
+    def trade_from_okx(self, raw_trade: dict[str, Any]) -> Optional[Trade]:
+        """
+        Converts a raw trade from OKX into a Trade object
+        """
+        try:
+            timestamp = ts_ms_to_sec(TimestampMS(int(raw_trade['cTime'])))
+            [base_asset_str, quote_asset_str] = raw_trade['instId'].split('-')
```
yabirgb, line 316: In python the `[]` is not needed for assignment

```suggestion
            base_asset_str, quote_asset_str = raw_trade['instId'].split('-')
```

rotkehlchen/tests/fixtures/exchanges/okx.py
```diff
@@ -0,0 +1,22 @@
...
+
+@pytest.fixture
+def mock_okx(
+        database,
+        inquirer,  # pylint: disable=unused-argument
+        function_scope_messages_aggregator,
+):
+    api_key = ApiKey('f32f48d7-74ad-41ce-8028-fcc4e4589f9c')
+    secret = ApiSecret(b'3DC350723E8200C236792784644E17A0')
+    passphrase = 'Rotki123!'
```
yabirgb, line 15: For this I'm seeing and in the other exchanges we create fake api keys for this except for gemini that we use the sandbox. Since you are mocking the requests in which test are the keys needed?

https://github.com/rotki/rotki/blob/d8e16391d5c35b471eebb0635d4334a248d06aae/rotkehlchen/tests/utils/exchanges.py#L802

yabirgb (commented):
Thanks @anukul, left some minor comments

rotkehlchen/tests/fixtures/exchanges/okx.py
anukul: OKX API requires authentication for the currencies endpoint, used in `test_assets_are_known`

yabirgb: Yes, I've checked the API and there is no way around this it seems.

LefterisJP (commented):
Hey @anukul thanks a lot for such a nice and big PR. Left some comments/questions.

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,486 @@
...
+        May raise
+        - RemoteError if the OKX API or price oracle returns an unexpected response
+        """
+        data = self._api_query_list(endpoint='balance')
+        try:
+            currencies_data = data[0]['details']
+        except (IndexError, KeyError, TypeError) as e:
+            raise RemoteError(
+                f'{self.name} balance API request failed due to unexpected response {str(e)}',
+            ) from e
```
LefterisJP, line 245: 1. Why TypeError? We never handle this anywhere. When can this be raised?
2. Just like we do in the rest of the code for KeyError you need better explanation.

```suggestion
        except (IndexError, KeyError, TypeError) as e:
            msg = str(e)
            if isinstance(e, KeyError):
                msg = f'Missing key: {msg}'
            raise RemoteError(
                f'{self.name} balance API request failed due to unexpected response {msg}',
            ) from e
```

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,486 @@
...
+    ) -> list[LedgerAction]:
+        return []  # noop for okx
+
+    def trade_from_okx(self, raw_trade: dict[str, Any]) -> Optional[Trade]:
+        """
+        Converts a raw trade from OKX into a Trade object
+        """
+        try:
+            timestamp = ts_ms_to_sec(TimestampMS(int(raw_trade['cTime'])))
+            base_asset_str, quote_asset_str = raw_trade['instId'].split('-')
```
LefterisJP, line 384: This can raise `ValueError` due to `split()` if the string does not contain `'-``.
```suggestion
            try:
                base_asset_str, quote_asset_str = raw_trade['instId'].split('-')
            except ValueError as e:
                raise DeserializationError(f'Expected pair {raw_trade["instId"]} to contain a "-"') from e
```

And then remove the ValueError handling from below. This is the only thing that can raise ValueError here, right? Errors need to make sense. Otherwise if we see the following in a user's logs we don't easily know what happened: `ValueError: not enough values to unpack (expected 2, got 1)`

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,486 @@
...
+    def query_online_income_loss_expense(
+            self,
+            start_ts: Timestamp,
+            end_ts: Timestamp,
+    ) -> list[LedgerAction]:
+        return []  # noop for okx
+
+    def trade_from_okx(self, raw_trade: dict[str, Any]) -> Optional[Trade]:
+        """
+        Converts a raw trade from OKX into a Trade object
```
LefterisJP, line 380:
```suggestion
        Converts a raw trade from OKX into a Trade object. If there is an error `None` is returned and error is logged.
```

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,486 @@
...
+            tx_hash = raw_movement['txId']
+            timestamp = ts_ms_to_sec(TimestampMS(int(raw_movement['ts'])))
+            asset = asset_from_okx(raw_movement['ccy'])
+            amount = deserialize_asset_amount(raw_movement['amt'])
+            address = deserialize_asset_movement_address(raw_movement, 'to', asset)
+            fee = Fee(ZERO)
+            if category is AssetMovementCategory.WITHDRAWAL:
+                fee = deserialize_fee(raw_movement['fee'])
+
+            return AssetMovement(
```
LefterisJP, line 451: you can move this out of the try except.

LefterisJP (commented):
Left some more things that need changing

rotkehlchen/exchanges/okx.py
anukul: `TypeError` will be thrown on accessing `data[0]['details']` if `data[0]` is not a dictionary

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,498 @@
...
+        """
+        https://www.okx.com/docs-v5/en/#rest-api-account-get-balance
+
+        May raise
+        - RemoteError if the OKX API or price oracle returns an unexpected response
+        """
+        data = self._api_query_list(endpoint='balance')
+        try:
+            currencies_data = data[0]['details']
+        except (IndexError, KeyError, TypeError) as e:
```
LefterisJP, line 243: continuing the `TypeError` discussion do you see the problem here?

Such a `TypeError` would read: `"TypeError: string indices must be integers"`

Which would translate to: "OKX balance API request failed due to unexpected response TypeError: string indices must be integers".

We won't have any information as to what went wrong or which attribute it is that failed for us.

I think it's much better if you check that `data[0]` is a `dict` here and if not raise a `RemoteError` saying that the `"details"` key of the OKX balances response was not a dict as expected, and even include the raw response so it can be studied in the logs.

What we usually do with this is:
1. Put as few info as needed in the `RemoteError` as this is usually propagated high up, caught by the API and shown to the user.
2. At the same time log an error, with the raw response `data` so that when we study the logs of the user we can see both the error and the raw response and figure out what happened.

rotkehlchen/exchanges/okx.py
```diff
@@ -0,0 +1,498 @@
...
+            address = deserialize_asset_movement_address(raw_movement, 'to', asset)
+            fee = Fee(ZERO)
+            if category is AssetMovementCategory.WITHDRAWAL:
+                fee = deserialize_fee(raw_movement['fee'])
+        except UnknownAsset as e:
+            self.msg_aggregator.add_warning(
+                f'Found {self.name} deposit/withdrawal with unknown asset '
+                f'{e.identifier}. Ignoring it.',
+            )
+            return None
```
LefterisJP, line 466: ah I see that you need to repeat the `None` this way. You can fix this by doing something like `asset_movement = None` at the start of the function, at success assign it, and at the very end return it.

So you don't need to repeat `return None` in each except.

LefterisJP (commented):
Two more small comments

rotkehlchen/exchanges/okx.py
LefterisJP: You missed to do (2).

Logging the error containing the raw response so we, the developers, can debug it.

LefterisJP (approved):
Left a small comment you forgot, but can also be added in another PR.

Other than that looks good. So will merge.

The single failing test is timeout due to ethereum remote node query

PR #5332, opened 2023-01-02
Use evm_chain at transaction serialization out of the API

LefterisJP:


LefterisJP:
Finally all green. Merging.

PR #5333, opened 2023-01-02
Fix transfers between tracked accounts

nebolax:
Fix transfers of ether and internal transactions between tracked accounts

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,9 @@
 Changelog
 =========
 
+* :bug:`-` Transfers of ether between tracked accounts will now have a correct label in the UI.
+
+
 * :release:`1.26.3 <2022-12-30>`
 * :bug:`5315` Fix issue where balance is not fully refreshed after detect tokens button pressed.
 * :bug:`-` Connecting to substrate nodes will no longer timeout prematurely for systems with slow connections.
--- a/rotkehlchen/chain/ethereum/decoding/decoder.py
+++ b/rotkehlchen/chain/ethereum/decoding/decoder.py
@@ -57,6 +57,7 @@ from .constants import (
     GOVERNORALPHA_PROPOSE_ABI,
     GTC_CLAIM,
     ONEINCH_CLAIM,
+    OUTGOING_EVENT_TYPES,
 )
 from .structures import ActionItem
 from .utils import maybe_reshuffle_events
@@ -363,7 +364,7 @@ class EVMTransactionDecoder():
                 continue
 
             event_type, location_label, counterparty, verb = direction_result
-            preposition = 'to' if verb == 'Send' else 'from'
+            preposition = 'to' if event_type in OUTGOING_EVENT_TYPES else 'from'
             events.append(HistoryBaseEntry(
                 event_identifier=tx.tx_hash,
                 sequence_index=self.base.get_next_sequence_counter(),
@@ -387,7 +388,7 @@ class EVMTransactionDecoder():
             return None
         event_type, location_label, counterparty, verb = direction_result
         amount = ZERO if tx.value == 0 else from_wei(FVal(tx.value))
-        preposition = 'to' if verb == 'Send' else 'from'
+        preposition = 'to' if event_type in OUTGOING_EVENT_TYPES else 'from'
         return HistoryBaseEntry(
             event_identifier=tx.tx_hash,
             sequence_index=self.base.get_next_sequence_counter(),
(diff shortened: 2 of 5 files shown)


rotkehlchen/chain/ethereum/decoding/decoder.py
```diff
@@ -363,7 +363,7 @@ def _maybe_decode_internal_transactions(
                 continue
 
             event_type, location_label, counterparty, verb = direction_result
-            preposition = 'to' if verb == 'Send' else 'from'
+            preposition = 'to' if event_type in {HistoryEventType.SPEND, HistoryEventType.TRANSFER} else 'from'  # noqa: E501
```
yabirgb, line 366: Shouldn't `DEPOSIT` also be here? I believe it makes sense to create a constant with the types that are sending stuff out of the account and use it here and below and in the erc20/721 transfers

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -224,3 +224,81 @@ def test_simple_erc20_transfer(
...
+]])
+def test_eth_transfer(
+        database,
+        ethereum_accounts,
+        evm_transaction_decoder,
+):
+    """
+    Data taken from
+    https://etherscan.io/tx/0x8caa7df2ebebfceb98207605e64691202b9e7498c3cccdbccb41c1600cf16e65
+    """
```
yabirgb, line 241: I believe we have some test for those but can you check also the deposit and the spend?

docs/changelog.rst
```diff
@@ -2,6 +2,9 @@
 Changelog
 =========
 
+* :bug:`-` Transfers between tracked accounts will now have a correct label in the UI.
+
```
yabirgb, line 6: I would mention that this is in the case of ETH, otherwise is confusing at it seems that it was fixed before and now again

rotkehlchen/chain/ethereum/decoding/decoder.py
nebolax: Yeah, perhaps it makes sense, thanks

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -224,3 +225,216 @@ def test_simple_erc20_transfer(
...
+    from_address = ethereum_accounts[0]
+    to_address = ethereum_accounts[1]
+    transaction = EvmTransaction(
+        tx_hash=evmhash,
+        timestamp=0,
+        block_number=0,
+        from_address=from_address,
+        to_address=to_address,
+        value=500000000000000000,
+        gas=0,
```
LefterisJP, line 253: These transactions are unrealistic. Please use actual number. 0 ETH for gas is impossible

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -224,3 +225,216 @@ def test_simple_erc20_transfer(
...
+    from_address = ethereum_accounts[0]
+    to_address = string_to_evm_address('0x38C3f1Ab36BdCa29133d8AF7A19811D10B6CA3FC')
+    transaction = EvmTransaction(
+        tx_hash=evmhash,
+        timestamp=0,
+        block_number=0,
+        from_address=from_address,
+        to_address=to_address,
+        value=500000000000000000,
+        gas=0,
```
LefterisJP, line 328: same here

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -224,3 +225,216 @@ def test_simple_erc20_transfer(
...
+    from_address = ethereum_accounts[0]
+    to_address = '0xAe2D4617c862309A3d75A0fFB358c7a5009c673F'  # Kraken 10
+    transaction = EvmTransaction(
+        tx_hash=evmhash,
+        timestamp=0,
+        block_number=0,
+        from_address=from_address,
+        to_address=to_address,
+        value=100000000000000000000,
+        gas=0,
```
LefterisJP, line 403: same

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -224,3 +225,216 @@ def test_simple_erc20_transfer(
...
+    with database.user_write() as cursor:
+        dbethtx.add_ethereum_transactions(cursor, [transaction], relevant_address=None)
+        events = evm_transaction_decoder.decode_transaction(
+            write_cursor=cursor,
+            transaction=transaction,
+            tx_receipt=receipt,
+        )
+    assert events == [
+        HistoryBaseEntry(
+            event_identifier=evmhash,
```
LefterisJP, line 426: Why isthere no gas event generated here?

nebolax: Yes, should be here, fixed. For some reason I thought that for deposits to exchanges we don't count gas (but we should count it of course)

LefterisJP (approved):
LGTM. The failed tests are time outs from APIs

Issue #4341, opened 2022-05-23
Decoded transaction events accounting settings should be configurable


LefterisJP:
## Abstract

Each module right now has a set of accounting settings for how decoded event should be treated at accounting. An example below for compound:

https://github.com/rotki/rotki/blob/48065dab4297ddcd24061c3a2eda789854782728/rotkehlchen/chain/ethereum/modules/compound/accountant.py#L19

These settings are hard-coded and not configurable by the user at the moment. This has to change.

## Specification

- Make a DB table where these settings reside.
- At `ModuleAccountantInterface` each submodule accountant should check the DB and if there is an entry there, return it. Otherwise return the hard coded defaults.
- Create an endpoint to query all accounting settings for decoded events
- Create an endpoint to edit settings for a decoded event
- UI interface to configure all these.

A2be:
Would the spec you have in mind address the [issue discussed on Discord 21 March 2023](https://discord.com/channels/657906918408585217/745172610240872508/1087750015818145812): ❓ 
```Use case: want to (try to) use Rotki to communicate my blockchain account info, accnt names, etc. activity to my tax professional.  I want T.P. to receive an export of a bunch of my Rotki DB data, but not all. (so, not send the entire DB + my DB decryption key to the T.P.).```

```US tax law allows gifts of under ~$15k/yr to be excluded from ordinary "gift tax" rulz.  Say I want to encode on the Edit event page that the Event type: Spend is a gift to another person, and not a spend where I benefit.  (thus, non taxable).```

Would it allow, for example, an `Evm Transactions/Event Subtype` to be user defined? So one might spec out a subtype of gift?  (or, `gift` or `gift <$nnnnn)?  ❔ 

Would it allow a user-spec'd subset of transactions +Rotki DB info to be exported to a tax professional who is retained to deal with the Rotki user's taxes?
  this could be: ... exported in a generic format (e.g., .csv) ❔  , OR
     ... exported in a Rotki-specific format (where the tax professional might import that data file into a TP-Rotki-instance so that they might see their client's blockchain stuff, WITHOUT the client having to share a snapshot of their entire Rotki DB + their Rotki decryption password) ❔

LefterisJP:
This would work for:

> US tax law allows gifts of under ~$15k/yr to be excluded from ordinary "gift tax" rulz.  Say one wants to encode on the Edit event page that the Event type: Spend is a gift to another person, and not a spend where I benefit.  (thus, non taxable).

For this:

[...]

PR #5336, opened 2023-01-02
Add tx settings to give users more capabilities to customize events

nebolax:
Adds some settings that allow users to customize events.

These settings don't have counterparty set and will be applied when users customize events.

### DIFF
--- a/rotkehlchen/chain/evm/accounting/aggregator.py
+++ b/rotkehlchen/chain/evm/accounting/aggregator.py
@@ -3,6 +3,7 @@ import logging
 import pkgutil
 from types import ModuleType
 from typing import TYPE_CHECKING, Union
+from rotkehlchen.accounting.ledger_actions import LedgerActionType
 
 from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.chain.ethereum.constants import MODULES_PACKAGE, MODULES_PREFIX_LENGTH
@@ -11,7 +12,7 @@ from rotkehlchen.errors.misc import ModuleLoadingError
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.user_messages import MessagesAggregator
 
-from .structures import TxEventSettings
+from .structures import TxEventSettings, TxMultitakeTreatment
 
 if TYPE_CHECKING:
     from rotkehlchen.accounting.pot import AccountingPot
@@ -104,22 +105,17 @@ class EVMAccountingAggregators():
     def __init__(self, aggregators: list[EVMAccountingAggregator]) -> None:
         self.aggregators = aggregators
 
-    def get_accounting_settings(self, pot: 'AccountingPot') -> dict[str, TxEventSettings]:
+    def _get_internal_settings(self, pot: 'AccountingPot') -> dict[str, TxEventSettings]:
         """
-        Iterate through loaded accountants and get accounting settings for each event type
-
-        This also contains the default built-in settings.
+        Returns accounting settings for events that can come from various decoders and thus
+        don't have any particular counterparty.
         """
         result = {}
-        for aggregator in self.aggregators:
-            result.update(aggregator.get_accounting_settings(pot))
-
-        # Also add the default settings
         gas_key = str(HistoryEventType.SPEND) + '__' + str(HistoryEventSubType.FEE) + '__' + CPT_GAS  # noqa: E501
         result[gas_key] = TxEventSettings(
             taxable=pot.settings.include_gas_costs,
-            count_entire_amount_spend=True,
-            count_cost_basis_pnl=True,
+            count_entire_amount_spend=pot.settings.include_gas_costs,
+            count_cost_basis_pnl=pot.settings.include_crypto2crypto,
             take=1,
             method='spend',
         )
@@ -155,6 +151,76 @@ class EVMAccountingAggregators():
             take=1,
             method='acquisition',
         )
+        return result
+
+    def _get_settings_for_editable_customizations(
+            self,
+            pot: 'AccountingPot',
+    ) -> dict[str, TxEventSettings]:
+        """
+        Returns accounting settings that allow users to customize events. Settings generated here
+        don't have any particular counterparty and thus can be easily set by the users in the UI.
+        Users are supposed to apply these settings in Evm Trasactions view.
+        """
+        result = {}
+        fee_key = str(HistoryEventType.SPEND) + '__' + str(HistoryEventSubType.FEE)
+        result[fee_key] = TxEventSettings(
+            taxable=True,
+            count_entire_amount_spend=True,
+            count_cost_basis_pnl=True,
+            method='spend',
+            take=1,
+        )
+        renew_key = str(HistoryEventType.RENEW) + '__' + str(HistoryEventSubType.NONE)
+        result[renew_key] = TxEventSettings(
+            taxable=True,
+            count_entire_amount_spend=True,
+            count_cost_basis_pnl=True,
+            method='spend',
+            take=1,
+        )
+        swap_key = str(HistoryEventType.TRADE) + '__' + str(HistoryEventSubType.SPEND)
+        result[swap_key] = TxEventSettings(
+            taxable=True,
+            count_entire_amount_spend=False,
+            count_cost_basis_pnl=True,
+            method='spend',
+            take=2,
+            multitake_treatment=TxMultitakeTreatment.SWAP,
+        )
+        airdrop_key = str(HistoryEventType.RECEIVE) + '__' + str(HistoryEventSubType.AIRDROP)
+        result[airdrop_key] = TxEventSettings(
+            taxable=LedgerActionType.AIRDROP in pot.settings.taxable_ledger_actions,
+            # count_entire_amount_spend and count_cost_basis_pnl don't matter for acquisitions.
+            count_entire_amount_spend=False,
+            count_cost_basis_pnl=False,
+            method='acquisition',
+            take=1,
+        )
+        reward_key = str(HistoryEventType.RECEIVE) + '__' + str(HistoryEventSubType.REWARD)
+        result[reward_key] = TxEventSettings(
+            taxable=True,
+            # count_entire_amount_spend and count_cost_basis_pnl don't matter for acquisitions.
+            count_entire_amount_spend=False,
+            count_cost_basis_pnl=False,
+            method='acquisition',
+            take=1,
+        )
+        return result
+
+    def get_accounting_settings(self, pot: 'AccountingPot') -> dict[str, TxEventSettings]:
+        """
+        Iterate through loaded accountants and get accounting settings for each event type
+
+        This also contains the default built-in settings.
+        """
+        result = {}
+        for aggregator in self.aggregators:
+            result.update(aggregator.get_accounting_settings(pot))
+
+        # Using | operator is safe since dicts keys are unique and won't be overriden
+        result |= self._get_internal_settings(pot)
+        result |= self._get_settings_for_editable_customizations(pot)
         return result
 
     def reset(self) -> None:
--- a/rotkehlchen/tests/unit/accounting/test_default_settings.py
+++ b/rotkehlchen/tests/unit/accounting/test_default_settings.py
@@ -1,0 +1,352 @@
+from typing import Optional
+
+import pytest
+
+from rotkehlchen.accounting.cost_basis.base import (
+    AssetAcquisitionEvent,
+    CostBasisInfo,
+    MatchedAcquisition,
+)
+from rotkehlchen.accounting.ledger_actions import LedgerActionType
+from rotkehlchen.accounting.mixins.event import AccountingEventType
+from rotkehlchen.accounting.pnl import PNL
+from rotkehlchen.accounting.pot import AccountingPot
+from rotkehlchen.accounting.structures.balance import Balance
+from rotkehlchen.accounting.structures.base import HistoryBaseEntry
+from rotkehlchen.accounting.structures.processed_event import ProcessedAccountingEvent
+from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.accounting.transactions import TransactionsAccountant
+from rotkehlchen.chain.evm.decoding.constants import CPT_GAS
+from rotkehlchen.constants import ZERO
+from rotkehlchen.constants.assets import A_DAI, A_ETH
+from rotkehlchen.constants.misc import ONE
+from rotkehlchen.db.settings import ModifiableDBSettings
+from rotkehlchen.fval import FVal
+from rotkehlchen.tests.utils.factories import make_evm_address, make_random_bytes
+from rotkehlchen.types import Location, Price, Timestamp, TimestampMS, make_evm_tx_hash
+from rotkehlchen.utils.misc import ts_sec_to_ms
+
+EXAMPLE_EVM_HASH = make_evm_tx_hash(make_random_bytes(32))
+EXAMPLE_TX_HASH_HEX = EXAMPLE_EVM_HASH.hex()  # pylint: disable=no-member  # EvmTxHash does have hex() member  # noqa: E501
+EXAMPLE_ADDRESS = make_evm_address()
+
+# Some utility timestamps. They are used for mocked prices.
+TIMESTAMP_1_SECS = Timestamp(1624395186)
+TIMESTAMP_1_MS = ts_sec_to_ms(TIMESTAMP_1_SECS)
+TIMESTAMP_2_SECS = Timestamp(1628994441)
+TIMESTAMP_2_MS = ts_sec_to_ms(TIMESTAMP_2_SECS)
+
+ETH_PRICE_TS_1 = FVal('2000')
+ETH_PRICE_TS_2 = FVal('3000')
+
+MOCKED_PRICES = {
+    'ETH': {
+        'EUR': {
+            TIMESTAMP_1_SECS: ETH_PRICE_TS_1,
+            TIMESTAMP_2_SECS: ETH_PRICE_TS_2,
+        },
+    },
+    A_DAI.identifier: {
+        'EUR': {
+            TIMESTAMP_2_SECS: Price(ONE),
+        },
+    },
+}
+
+
+@pytest.fixture(name='airdrops_taxable')
+def fixture_airdrops_taxable():
+    return True
+
+
+@pytest.fixture(name='gas_taxable')
+def fixture_gas_taxable():
+    return True
+
+
+@pytest.fixture(name='include_crypto2crypto')
+def fixture_include_crypto2crypto():
+    return True
+
+
+@pytest.fixture(name='accounting_pot')
+def fixture_accounting_pot(accountant, airdrops_taxable, gas_taxable, include_crypto2crypto):
+    pot = accountant.pots[0]
+    with pot.database.user_write() as write_cursor:
+        pot.database.set_settings(
+            write_cursor=write_cursor,
+            settings=ModifiableDBSettings(
+                taxable_ledger_actions=[LedgerActionType.AIRDROP] if airdrops_taxable else [],
+                include_gas_costs=gas_taxable,
+                include_crypto2crypto=include_crypto2crypto,
+            ),
+        )
+    with pot.database.conn.read_ctx() as cursor:
+        settings = pot.database.get_settings(cursor)
+    pot.reset(
+        settings=settings,
+        start_ts=Timestamp(0),  # timestamps don't matter since they are used only for querying
+        end_ts=Timestamp(0),  # history, which does not happen in the tests in this file
+        report_id=1,
+    )
+    return pot
+
+
+def _gain_one_ether(
+        transactions: 'TransactionsAccountant',
+        event_type: 'HistoryEventType' = HistoryEventType.RECEIVE,
+        event_subtype: 'HistoryEventSubType' = HistoryEventSubType.NONE,
+) -> None:
+    """Helper function to gain 1 ETH, so that spending events have something to spend"""
+    eth_gain_event = HistoryBaseEntry(
+        event_identifier=EXAMPLE_EVM_HASH,
+        sequence_index=0,
+        timestamp=TIMESTAMP_1_MS,
+        location=Location.BLOCKCHAIN,
+        location_label=EXAMPLE_ADDRESS,
+        asset=A_ETH,
+        balance=Balance(amount=ONE),
+        notes='Received 1 ETH',
+        event_type=event_type,
+        event_subtype=event_subtype,
+    )
+    consumed_num = transactions.process(
+        event=eth_gain_event,
+        events_iterator=iter([]),
+    )
+    assert consumed_num == 1
+
+
+def test_accounting_no_settings(accounting_pot: 'AccountingPot'):
+    """Test that if there are no settings provided, the event is not taken into account"""
+    event = HistoryBaseEntry(
+        event_identifier=EXAMPLE_EVM_HASH,
+        sequence_index=0,
+        timestamp=TimestampMS(0),
+        location=Location.BLOCKCHAIN,
+        event_type=HistoryEventType.INFORMATIONAL,
+        event_subtype=HistoryEventSubType.NONE,
+        asset=A_ETH,
+        balance=Balance(),
+    )
+    consumed_num = accounting_pot.transactions.process(event=event, events_iterator=iter([]))
+
+    assert consumed_num == 1
+    assert len(accounting_pot.pnls) == 0, 'Nothing should have happened since there were no settings'  # noqa: E501
+
+
+@pytest.mark.parametrize('event_type,event_subtype,is_taxable,airdrops_taxable', [
+    (HistoryEventType.RECEIVE, HistoryEventSubType.NONE, True, False),
+    (HistoryEventType.WITHDRAWAL, HistoryEventSubType.NONE, False, False),
+    (HistoryEventType.RECEIVE, HistoryEventSubType.AIRDROP, False, False),
+    (HistoryEventType.RECEIVE, HistoryEventSubType.AIRDROP, True, True),
+    (HistoryEventType.RECEIVE, HistoryEventSubType.REWARD, True, False),
+])
+@pytest.mark.parametrize('mocked_price_queries', [MOCKED_PRICES])
+def test_accounting_receive_settings(
+        accounting_pot: 'AccountingPot',
+        event_type: 'HistoryEventType',
+        event_subtype: 'HistoryEventSubType',
+        is_taxable: bool,
+):
+    """Test that the default accounting settings for receiving are correct"""
+    _gain_one_ether(
+        transactions=accounting_pot.transactions,
+        event_type=event_type,
+        event_subtype=event_subtype,
+    )
+    expected_event = ProcessedAccountingEvent(
+        type=AccountingEventType.TRANSACTION_EVENT,
+        notes='Received 1 ETH',
+        location=Location.BLOCKCHAIN,
+        timestamp=TIMESTAMP_1_SECS,
+        asset=A_ETH,
+        free_amount=ZERO if is_taxable else ONE,
+        taxable_amount=ONE if is_taxable else ZERO,
+        price=Price(ETH_PRICE_TS_1),
+        pnl=PNL(
+            taxable=ETH_PRICE_TS_1 if is_taxable else ZERO,
+            free=ZERO,
+        ),
+        cost_basis=None,
+        index=0,
+        extra_data={'tx_hash': EXAMPLE_TX_HASH_HEX},
+    )
+    expected_event.count_entire_amount_spend = False
+    expected_event.count_cost_basis_pnl = is_taxable
+
+    assert accounting_pot.processed_events == [expected_event]
+    assert accounting_pot.pnls.taxable == ETH_PRICE_TS_1 if is_taxable else ZERO
+    assert accounting_pot.pnls.free == ZERO
+
+
+@pytest.mark.parametrize('event_type,event_subtype,is_taxable,counterparty,gas_taxable,include_crypto2crypto', [  # noqa: E501
+    (HistoryEventType.SPEND, HistoryEventSubType.NONE, True, None, False, False),
+    (HistoryEventType.SPEND, HistoryEventSubType.FEE, True, None, False, False),
+    (HistoryEventType.SPEND, HistoryEventSubType.FEE, True, CPT_GAS, True, False),
+    (HistoryEventType.SPEND, HistoryEventSubType.FEE, True, CPT_GAS, True, True),
+    (HistoryEventType.SPEND, HistoryEventSubType.FEE, False, CPT_GAS, False, False),
+    (HistoryEventType.SPEND, HistoryEventSubType.FEE, False, CPT_GAS, False, True),
+    (HistoryEventType.DEPOSIT, HistoryEventSubType.NONE, False, None, False, False),
+    (HistoryEventType.RENEW, HistoryEventSubType.NONE, True, None, False, False),
+])
+@pytest.mark.parametrize('mocked_price_queries', [MOCKED_PRICES])
+def test_accounting_spend_settings(
+        accounting_pot: 'AccountingPot',
+        event_type: 'HistoryEventType',
+        event_subtype: 'HistoryEventSubType',
+        is_taxable: bool,
+        counterparty: Optional[str],
+        include_crypto2crypto,
+):
+    _gain_one_ether(transactions=accounting_pot.transactions)
+    spend_event = HistoryBaseEntry(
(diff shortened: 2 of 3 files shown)


rotkehlchen/chain/ethereum/accounting/aggregator.py
```diff
@@ -112,6 +113,52 @@ def get_accounting_settings(self, pot: 'AccountingPot') -> dict[str, TxEventSett
...
+        airdrop_key = str(HistoryEventType.RECEIVE) + '__' + str(HistoryEventSubType.AIRDROP)
+        airdrops_taxable = LedgerActionType.AIRDROP in pot.settings.taxable_ledger_actions
+        result[airdrop_key] = TxEventSettings(
+            taxable=airdrops_taxable,
+            count_entire_amount_spend=False,
+            count_cost_basis_pnl=False,
+            method='acquisition',
+            take=1,
+            multitake_treatment=None,
+        )
```
yabirgb, line 134: This has to respect the setting for airdrops as we do in https://github.com/rotki/rotki/blob/296f830262c3022830d84130396549a445341fd1/rotkehlchen/chain/ethereum/modules/airdrops/accountant.py#L21

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,271 @@
...
+
+@pytest.fixture(name='pot')
+def fixture_pot(accountant):
+    pot: 'AccountingPot' = accountant.pots[0]
+    with pot.database.conn.read_ctx() as cursor:
+        settings = pot.database.get_settings(cursor)
+    pot.reset(
+        settings=settings,
+        start_ts=Timestamp(0),
+        end_ts=Timestamp(100),
```
yabirgb, line 35: how are this start/end chosen. Does it affect somehow the tests?

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,271 @@
...
+        report_id=1,
+    )
+    return pot
+
+
+def gain_one_ether(transactions, evmhash):
+    eth_gain_event = HistoryBaseEntry(
+        event_identifier=evmhash,
```
yabirgb, line 43: Can you add a docstring explaining what this function does and how it is used? Also if is only used here you can add the `_` before the name

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,271 @@
...
+        location=Location.BLOCKCHAIN,
+        event_type=HistoryEventType.INFORMATIONAL,
+        event_subtype=HistoryEventSubType.NONE,
+        asset=A_ETH,
+        balance=Balance(),
+    )
+    was_skipped = False
+
+    def log_mock(text):
+        assert 'Skipping...' in text
```
yabirgb, line 75: shouldn't you also check that the skipped entry is the one you think is being skipped? Doesn't it have the identifier in the row?

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,271 @@
...
+        nonlocal was_skipped
+        was_skipped = True
+
+    log_patch = patch('rotkehlchen.accounting.transactions.log.debug', side_effect=log_mock)
+    with log_patch:
```
yabirgb, line 80: can it go in one line? if you don't need to reuse it I believe is fine to just do

```
with patch(...
```

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,271 @@
...
+from rotkehlchen.constants.misc import ONE
+from rotkehlchen.fval import FVal
+from rotkehlchen.tests.utils.history import prices
+from rotkehlchen.types import Location, Timestamp, TimestampMS, deserialize_evm_tx_hash
+
+
+@pytest.fixture(name='pot')
+def fixture_pot(accountant):
+    pot: 'AccountingPot' = accountant.pots[0]
+    with pot.database.conn.read_ctx() as cursor:
```
yabirgb, line 30: this fixture should go with the other fixtures in the `tests/fixtures` folders. Probably under `accounting`. Also what do you think about renaming this one to `accounting_pot`?

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,271 @@
...
+        notes='Send 1 ETH to 0xABC',
+        event_type=HistoryEventType.SPEND,
+        event_subtype=HistoryEventSubType.NONE,
+    )
+    consumed_num = pot.transactions.process(event=gas_event, events_iterator=None)
+    assert consumed_num == 1
+
+    acquisition_event = AssetAcquisitionEvent(
+        amount=ONE,
+        timestamp=Timestamp(1624395186),
```
yabirgb, line 171: This timestamp comes from the transaction in the method you have right? Maybe you can return it and use it instead of repeating this constant in all the tests

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,271 @@
...
+        extra_data={'tx_hash': tx_hex},
+    )
+    expected_event.calculate_pnl(count_entire_amount_spend=True, count_cost_basis_pnl=True)
+    assert pot.processed_events[-1] == expected_event
+
+
+@pytest.mark.parametrize('mocked_price_queries', [prices])
+def test_accounting_spend_settings(pot):
+    """Test that the default accounting settings for spends are correct"""
+    tx_hex = '0x282f632c7a2385d026068122dc415562a1c4e827fd54673671af1d40445d6b58'
```
yabirgb, line 149: This txhash seems to be the same in all the tests. Can you set it as a constant and state that is not relevant for the test?

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,271 @@
...
+    pot.reset(
+        settings=settings,
+        start_ts=Timestamp(0),
+        end_ts=Timestamp(100),
+        report_id=1,
+    )
+    return pot
+
+
+def gain_one_ether(transactions, evmhash):
```
yabirgb, line 41: Why do you need to pass the evmhash? is it used at some point?. The event is constant so it doesn't seem to be relevant

yabirgb (commented):
Left some comments about the implementation.

About how to organize the tests for accounting I'm not sure. Since we already mock the transactions when preparing the tests for the decoders maybe is it easier to check that the accounting is properly done there? What do you think? Also @LefterisJP could also have another idea

rotkehlchen/chain/ethereum/accounting/aggregator.py
nebolax: Hmmm I don't get you, what setting?

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
nebolax: Will add a comment into the code, but no, it does not matter, since is used only for querying history which does not happen in the tests in this file

rotkehlchen/chain/ethereum/accounting/aggregator.py
yabirgb: This one that sets if the airdrops are taxable https://github.com/rotki/rotki/blob/296f830262c3022830d84130396549a445341fd1/rotkehlchen/chain/ethereum/modules/airdrops/accountant.py#L19

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
nebolax: No, it doesn't have the identifier. But in the place where this patch is used:

```python
with log_patch:
        consumed_num = accounting_pot.transactions.process(event=event, events_iterator=None)
```

only a single event is processed, so I think there is no chance that some other event will appear there

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
nebolax: Alright. Wanted to be more consistent with the usages of patches in other tests, but I guess can also go in one line

rotkehlchen/chain/ethereum/accounting/aggregator.py
nebolax: Isn't this setting respected here? :thinking: 
image

yabirgb: I would swear it was not there but yes, that one. Sorry

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+def _gain_one_ether(transactions):
+    """Helper function to gain 1 ETH, so that spending events have what to spend"""
+    eth_gain_event = HistoryBaseEntry(
+        event_identifier=EXAMPLE_EVM_HASH,
+        sequence_index=0,
+        timestamp=TIMESTAMP_1_MS,
+        location=Location.BLOCKCHAIN,
+        location_label=string_to_evm_address('0x4bBa290826C253BD854121346c370a9886d1bC26'),
```
yabirgb, line 60: this address is being used around all the tests. Can be a constant too

yabirgb (commented):
This is also would need an update in the docs explaining this rules.

I was checking the tests and isn't this missing the tests for the rules that you created?

rotkehlchen/chain/evm/accounting/aggregator.py
```diff
@@ -155,6 +164,40 @@ def get_accounting_settings(self, pot: 'AccountingPot') -> dict[str, TxEventSett
             take=1,
             method='acquisition',
         )
+        swap_key = str(HistoryEventType.TRADE) + '__' + str(HistoryEventSubType.SPEND)
```
LefterisJP, line 167: why is this here? Don't all swaps already have their own rules in each accountant module? IF this is here then shouldn't those accountants be removed?

rotkehlchen/chain/evm/accounting/aggregator.py
```diff
@@ -123,6 +124,14 @@ def get_accounting_settings(self, pot: 'AccountingPot') -> dict[str, TxEventSett
             take=1,
             method='spend',
         )
+        fee_key = str(HistoryEventType.SPEND) + '__' + str(HistoryEventSubType.FEE)
```
LefterisJP, line 127: There is already `get_tx_event_type_identifier(HistoryEventType.SPEND, HistoryEventSubType.FEE, CPT_ONEINCH_V1): TxEventSettings(  # noqa: E501` for fees. Shoudl it exist? It has `count_entire_amount_spend` as `False` though.

rotkehlchen/chain/evm/accounting/aggregator.py
```diff
@@ -155,6 +164,40 @@ def get_accounting_settings(self, pot: 'AccountingPot') -> dict[str, TxEventSett
...
+        swap_key = str(HistoryEventType.TRADE) + '__' + str(HistoryEventSubType.SPEND)
+        result[swap_key] = TxEventSettings(
+            taxable=True,
+            count_entire_amount_spend=False,
+            count_cost_basis_pnl=True,
+            method='spend',
+            take=2,
+            multitake_treatment=TxMultitakeTreatment.SWAP,
+        )
+        airdrop_key = str(HistoryEventType.RECEIVE) + '__' + str(HistoryEventSubType.AIRDROP)
```
LefterisJP, line 176: why are you adding this here? We already have all the airdrop accountants set up and there should be no more airdrops existing outside of those.

rotkehlchen/chain/evm/accounting/aggregator.py
```diff
@@ -155,6 +164,40 @@ def get_accounting_settings(self, pot: 'AccountingPot') -> dict[str, TxEventSett
...
+        airdrops_taxable = LedgerActionType.AIRDROP in pot.settings.taxable_ledger_actions
+        result[airdrop_key] = TxEventSettings(
+            taxable=airdrops_taxable,
+            count_entire_amount_spend=False,
+            count_cost_basis_pnl=False,
+            method='acquisition',
+            take=1,
+        )
+        reward_key = str(HistoryEventType.RECEIVE) + '__' + str(HistoryEventSubType.REWARD)
```
LefterisJP, line 185: We already have this in aave1-v1, compound and votium. The only difference is aave-v1 has `count_cost_basis_pnl` as True.

The question here is why the discrepancy? Am I correct in seeing that **for acquisitions**, `count_cost_basis_pnl` is not used? So it's the same. So then again the question on why have them both in the accountants and here stays.

rotkehlchen/chain/evm/accounting/aggregator.py
```diff
@@ -155,6 +164,40 @@ def get_accounting_settings(self, pot: 'AccountingPot') -> dict[str, TxEventSett
...
+        result[reward_key] = TxEventSettings(
+            taxable=True,
+            count_entire_amount_spend=False,
+            count_cost_basis_pnl=False,
+            method='acquisition',
+            take=1,
+        )
+        renew_key = str(HistoryEventType.RENEW) + '__' + str(HistoryEventSubType.NONE)
```
LefterisJP, line 193: why make this a generic default? Why not just add it to the only place this exists, ens module as an accountant?

LefterisJP (commented):
Okay sorry I don't get the point of this PR. (I did not check the tests though)

 You seem to confuse defaults with accountants and re-add the same things that should already exist and by doing so also just waste a lot of logic. If you wanted to add missing per-module definitions do that. Don't just add all in the defaults file. That's why we creates a modular system.

Defaults only existed for things that do not have a module, and thus can't have an accountant. You probably have not understood how this is supposed to work. Both you @yabirgb and @nebolax

LefterisJP:
> Okay sorry I don't get the point of this PR. (I did not check the tests though)

Hmmm as I was thinking why you would try to do this yesterday I came up with a possible answer. Did you do all this in order to make it possible for swaps/airdrops etc. to be taken into account in a generic way when the user edits a transaction like that?

If yes ... then please guys write that somewhere in the description of the PR or in some comment. Don't say "Add some default tx settings".

Also in that case I would:

1. Add them in a separate function to separate what would be potentially editable combinations by the user from the defaults for which we don't build a module accountant.
2. Explain what they are in a docstring
3. (In the future) - they should also be part of configurability just like the accountant ones: https://github.com/rotki/rotki/issues/4341

rotkehlchen/chain/evm/accounting/aggregator.py
nebolax: Hm, I am not sure what value should `count_entire_amount_spend` have because I guess in different jurisdictions fees should be treated differently. I chose `True` here because gas by default is taxable and has `count_entire_amount_spend` set to `True`.

rotkehlchen/chain/evm/accounting/aggregator.py
```diff
@@ -157,6 +153,73 @@ def get_accounting_settings(self, pot: 'AccountingPot') -> dict[str, TxEventSett
         )
         return result
 
+    def _get_settings_for_customization(self, pot: 'AccountingPot') -> dict[str, TxEventSettings]:
```
LefterisJP, line 156: nitpick: Both in the name and in the docstring don't use customization as a word. Since that we should use for customizing the tax treatment of each different type of event when https://github.com/rotki/rotki/issues/4341 is done.

Perhaps call it editable_combinations and in the docstrings explain users can edit the transactions type/subtype to get them.

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,338 @@
+from typing import Optional
```
LefterisJP, line 1: Why not place this inside `tests/unit/accounting`? This is not a decoder test but an accounting test.

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,338 @@
...
+                taxable_ledger_actions=taxable_ledger_actions,
+                include_gas_costs=gas_taxable,
+            ),
+        )
+    with pot.database.conn.read_ctx() as cursor:
+        settings = pot.database.get_settings(cursor)
+    pot.reset(
+        settings=settings,
+        start_ts=Timestamp(0),  # timestamps don't matter since they are used only for querying
+        end_ts=Timestamp(0),  # histroy, which does not happen in the tests in this file
```
LefterisJP, line 71:
```suggestion
        end_ts=Timestamp(0),  # history, which does not happen in the tests in this file
```

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,338 @@
...
+    (HistoryEventType.RECEIVE, HistoryEventSubType.REWARD, True, False),
+])
+@pytest.mark.parametrize('mocked_price_queries', [prices])
+def test_accounting_receipts_settings(
+        accounting_pot: 'AccountingPot',
+        event_type: 'HistoryEventType',
+        event_subtype: 'HistoryEventSubType',
+        is_taxable: bool,
+):
+    """Test that the default accounting settings for recepts are correct"""
```
LefterisJP, line 141:
```suggestion
    """Test that the default accounting settings for receiving are correct"""
```

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,338 @@
...
+from rotkehlchen.constants.misc import ONE
+from rotkehlchen.db.settings import ModifiableDBSettings
+from rotkehlchen.fval import FVal
+from rotkehlchen.tests.utils.history import prices
+from rotkehlchen.types import Location, Price, Timestamp, TimestampMS, deserialize_evm_tx_hash
+from rotkehlchen.utils.misc import ts_sec_to_ms
+
+
+# The transaction hash doesn't matter in the tests in this file, so we can use any.
+EXAMPLE_TX_HASH_HEX = '0x282f632c7a2385d026068122dc415562a1c4e827fd54673671af1d40445d6b58'
```
LefterisJP, line 33: In which case you should just use a random one and not even mention anything here.We have the `make_random_bytes(42)` which can easily be turned into a hexstring.

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,338 @@
...
+EXAMPLE_EVM_HASH = deserialize_evm_tx_hash(EXAMPLE_TX_HASH_HEX)
+EXAMPLE_ADDRESS = string_to_evm_address('0x4bBa290826C253BD854121346c370a9886d1bC26')
+
+# Some utility timestamps. They don't matter here and are used only for mocked prices.
+TIMESTAMP_1_SECS = Timestamp(1624395186)
```
LefterisJP, line 38: Again same thing. If no specific value is needed no need to write it in the comment. You show it by using `make_random_timestamp()`

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,338 @@
...
+    )
+    return pot
+
+
+def _gain_one_ether(
+        transactions: 'TransactionsAccountant',
+        event_type: 'HistoryEventType' = HistoryEventType.RECEIVE,
+        event_subtype: 'HistoryEventSubType' = HistoryEventSubType.NONE,
+) -> None:
+    """Helper function to gain 1 ETH, so that spending events have what to spend"""
```
LefterisJP, line 82:
```suggestion
    """Helper function to gain 1 ETH, so that spending events have something to spend"""
```

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,338 @@
...
+        event_subtype=HistoryEventSubType.NONE,
+        asset=A_ETH,
+        balance=Balance(),
+    )
+    was_skipped = False
+
+    def log_mock(text):
+        assert 'Skipping...' in text
+        nonlocal was_skipped
+        was_skipped = True
```
LefterisJP, line 119: Don't do this. Do not rely on a log entry like this in a test. Check what is not there. So no processed events, no pnl change etc.

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,338 @@
...
+        pnl=PNL(),
+        cost_basis=None,
+        index=0,
+        extra_data={'tx_hash': EXAMPLE_TX_HASH_HEX},
+    )
+    expected_event.calculate_pnl(
+        count_entire_amount_spend=False,
+        count_cost_basis_pnl=is_taxable,
+    )
+    assert accounting_pot.processed_events == [expected_event]
```
LefterisJP, line 166: shouldn't you also check that pnl is correclty calculated? That's the whole point of accounting.

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,338 @@
...
+@pytest.mark.parametrize('event_type,event_subtype,is_taxable,counterparty,gas_taxable', [
+    (HistoryEventType.SPEND, HistoryEventSubType.NONE, True, None, False),
+    (HistoryEventType.SPEND, HistoryEventSubType.FEE, True, None, False),
+    (HistoryEventType.SPEND, HistoryEventSubType.FEE, True, CPT_GAS, True),
+    (HistoryEventType.SPEND, HistoryEventSubType.FEE, False, CPT_GAS, False),
+    (HistoryEventType.DEPOSIT, HistoryEventSubType.NONE, False, None, False),
+    (HistoryEventType.RENEW, HistoryEventSubType.NONE, True, None, False),
+])
+@pytest.mark.parametrize('mocked_price_queries', [prices])
+def test_accounting_spents_settings(
```
LefterisJP, line 178:
```suggestion
def test_accounting_spend_settings(
```

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,338 @@
...
+        accounting_pot: 'AccountingPot',
+        event_type: 'HistoryEventType',
+        event_subtype: 'HistoryEventSubType',
+        is_taxable: bool,
+        counterparty: Optional[str],
+):
+    """Test that the default accounting settings for spends are correct"""
+    # Gain some ETH so that the spending event has something to spend
```
LefterisJP, line 186: name is already descriptive enough
```suggestion
```

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,338 @@
...
+        pnl=PNL(),
+        cost_basis=cost_basis,
+        index=1,
+        extra_data={'tx_hash': EXAMPLE_TX_HASH_HEX},
+    )
+    expected_event.calculate_pnl(
+        count_entire_amount_spend=is_taxable,
+        count_cost_basis_pnl=is_taxable,
+    )
+    assert accounting_pot.processed_events[-1] == expected_event
```
LefterisJP, line 245: again same thing. You don't seem to check the PnL at all. The entire point of accounting is to have correct PnL. This is what you should check

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+@pytest.mark.parametrize('mocked_price_queries', [prices])
+def test_accounting_swap_settings(accounting_pot: 'AccountingPot'):
+    """Test that the default accounting settings for deposits are correct"""
+    # Gain some ETH so that the spending event has something to spend
```
LefterisJP, line 251:
```suggestion
```

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
```diff
@@ -0,0 +1,338 @@
...
+        pnl=PNL(free=ZERO, taxable=ZERO),
+        cost_basis=None,
+        index=2,
+        extra_data={'tx_hash': EXAMPLE_TX_HASH_HEX, 'group_id': EXAMPLE_TX_HASH_HEX + '12'},
+    )
+    expected_receive_event.calculate_pnl(
+        count_entire_amount_spend=False,
+        count_cost_basis_pnl=False,
+    )
+    assert accounting_pot.processed_events[1:] == [expected_spend_event, expected_receive_event]
```
LefterisJP, line 338: same here for pnl

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
nebolax: Yes, makes sense

rotkehlchen/tests/unit/decoders/accounting_settings/test_default_settings.py
nebolax: Will adjust the comments, but timestamps are required to be preset because they are used for getting historical prices

rotkehlchen/tests/unit/accounting/test_default_settings.py
```diff
@@ -0,0 +1,332 @@
...
+    _gain_one_ether(transactions=accounting_pot.transactions)
+    spend_event = HistoryBaseEntry(
+        event_identifier=EXAMPLE_EVM_HASH,
+        sequence_index=0,
+        timestamp=TIMESTAMP_2_MS,
+        location=Location.BLOCKCHAIN,
+        location_label=EXAMPLE_ADDRESS,
+        asset=A_ETH,
+        balance=Balance(amount=FVal(0.5)),
+        notes='Send 1 ETH to 0xABC',
```
LefterisJP, line 190: You are sneind half an ETH, no?

```suggestion
        notes='Send 0.5 ETH to 0xABC',
```

rotkehlchen/tests/unit/accounting/test_default_settings.py
```diff
@@ -0,0 +1,332 @@
...
+                    amount=FVal(0.5),
+                    event=acquisition_event,
+                    taxable=True,
+                ),
+            ],
+            is_complete=True,
+        )
+    expected_event = ProcessedAccountingEvent(
+        type=AccountingEventType.TRANSACTION_EVENT,
+        notes='Send 1 ETH to 0xABC',
```
LefterisJP, line 222:
```suggestion
        notes='Send 0.5 ETH to 0xABC',
```

rotkehlchen/tests/unit/accounting/test_default_settings.py
```diff
@@ -0,0 +1,332 @@
...
+
+    acquisition_event = AssetAcquisitionEvent(
+        amount=ONE,
+        timestamp=TIMESTAMP_1_SECS,
+        rate=Price(FVal(1862.06)),
+        index=0,
+    )
+    acquisition_event.remaining_amount = FVal(0.5)
+    cost_basis = None
+    if is_taxable:
```
LefterisJP, line 206: So some comments here:

1. Add comments, with the calculations and what it is you are testing.
2. You are mocking stuff. Use easier numbers. Why the heck use `1862.06` and `3258.55`. You know you can **and should** mock price queries. We have fixtures for this. Make sure no network calls happen in all these tests you made.
3. The spending part and the selling part are different and should be checked for in the test depending on customization. This is what `count_entire_amount` as spend is for and `taxable + crypto_to_crypto`.

rotkehlchen/tests/unit/accounting/test_default_settings.py
```diff
@@ -0,0 +1,332 @@
...
+    )
+    consumed_num = accounting_pot.transactions.process(
+        event=swap_spend_event,
+        events_iterator=iter([swap_receive_event]),
+    )
+    assert consumed_num == 2
+    acquisition_event = AssetAcquisitionEvent(
+        amount=ONE,
+        timestamp=TIMESTAMP_1_SECS,
+        rate=Price(FVal(1862.06)),
```
LefterisJP, line 282: same comments for rates here. Also for numbers you use multiple times in a test, add them in a variable

rotkehlchen/tests/unit/accounting/test_default_settings.py
```diff
@@ -0,0 +1,332 @@
...
+    assert accounting_pot.processed_events[-1] == expected_event
+    assert accounting_pot.pnls.taxable == FVal(1862.06) + expected_event.pnl.taxable
+    assert accounting_pot.pnls.free == ZERO
+
+
+@pytest.mark.parametrize('mocked_price_queries', [prices])
+def test_accounting_swap_settings(accounting_pot: 'AccountingPot'):
+    """Test that the default accounting settings for swaps are correct"""
+    _gain_one_ether(transactions=accounting_pot.transactions)
+
```
LefterisJP, line 249:
```suggestion
```

rotkehlchen/tests/unit/accounting/test_default_settings.py
```diff
@@ -0,0 +1,332 @@
...
+    expected_receive_event = ProcessedAccountingEvent(
+        type=AccountingEventType.TRANSACTION_EVENT,
+        notes='Receive 1200 DAI as the result of a swap',
+        location=Location.BLOCKCHAIN,
+        timestamp=TIMESTAMP_2_SECS,
+        asset=A_DAI,
+        free_amount=FVal(1200),
+        taxable_amount=ZERO,
+        price=Price(FVal('2.715458333333333333333333333')),
```
LefterisJP, line 322: Why this weird number? 1 DAI ~= 1$. Why make this more complicated?

rotkehlchen/tests/unit/accounting/test_default_settings.py
```diff
@@ -0,0 +1,332 @@
...
+    assert consumed_num == 2
+    acquisition_event = AssetAcquisitionEvent(
+        amount=ONE,
+        timestamp=TIMESTAMP_1_SECS,
+        rate=Price(FVal(1862.06)),
+        index=0,
+    )
+    acquisition_event.remaining_amount = ZERO
+
+    expected_spend_event = ProcessedAccountingEvent(
```
LefterisJP, line 287: You need to specify what the prices at the time are and how the processed event prices are derived. And use easier numbers, but make it make sense. At the moment it's very very hard to understand what's happening here. I can't really say if the test is correct or not.

rotkehlchen/chain/evm/accounting/aggregator.py
```diff
@@ -104,22 +105,17 @@ class EVMAccountingAggregators():
...
-            result.update(aggregator.get_accounting_settings(pot))
-
-        # Also add the default settings
         gas_key = str(HistoryEventType.SPEND) + '__' + str(HistoryEventSubType.FEE) + '__' + CPT_GAS  # noqa: E501
         result[gas_key] = TxEventSettings(
             taxable=pot.settings.include_gas_costs,
-            count_entire_amount_spend=True,
-            count_cost_basis_pnl=True,
+            count_entire_amount_spend=pot.settings.include_gas_costs,
+            count_cost_basis_pnl=pot.settings.include_gas_costs,
```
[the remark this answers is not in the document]
nebolax: Yes, makes sense, thanks :+1:

rotkehlchen/tests/unit/accounting/test_default_settings.py
```diff
@@ -0,0 +1,332 @@
...
+])
+@pytest.mark.parametrize('mocked_price_queries', [prices])
+def test_accounting_spend_settings(
+        accounting_pot: 'AccountingPot',
+        event_type: 'HistoryEventType',
+        event_subtype: 'HistoryEventSubType',
+        is_taxable: bool,
+        counterparty: Optional[str],
+):
+    # Gain some ETH so that the spending event has something to spend
```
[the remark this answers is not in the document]
nebolax: Ouch, don't know

rotkehlchen/tests/unit/accounting/test_default_settings.py
nebolax: Used already existing big structure with mocked prices and because of that prices were so ugly. But now improved.

LefterisJP (approved):
Perfect. All green and tests look correct. Merging.

PR #5340, opened 2023-01-03
Changes in ignored_actions api endpoint

LefterisJP:


LefterisJP:
Merging as is. Backend tests passed (except a timeout that passes locally).

Frontend tests will need adjustments.

PR #5346, opened 2023-01-04
Changes pnpm/cypress caching on GitHub actions

### LINKED ISSUE #5341: Make caching of cypress and pnpm cache consistent
## Abstract

```

The cypress npm package is installed, but the Cypress binary is missing.

We expected the binary to be installed here: /home/runner/.cache/Cypress/12.1.0/Cypress/Cypress

Reasons it may be missing:

- You're caching 'node_modules' but are not caching this path: /home/runner/.cache/Cypress
- You ran 'npm install' at an earlier build step but did not persist: /home/runner/.cache/Cypress

Properly caching the binary will fix this error and avoid downloading and unzipping Cypress.
```

## Motivation

With `action/setup-node` we are using it's own cache support.
With Cypress we use caching some times.

There are caches where what is restored from action/setup-node considers cypress installed, but 
~/.cache/Cypress is not properly restored

kelsos:
Closes #5341 

## Checklist

### DIFF
--- a/.github/workflows/rotki_ci.yml
+++ b/.github/workflows/rotki_ci.yml
@@ -80,21 +80,29 @@ jobs:
         uses: rotki/action-env@v1
         with:
           env_file: .github/.env.ci
-      - name: Setup pnpm
-        uses: pnpm/action-setup@v2
-        with:
-          version: ${{ env.PNPM_VERSION }}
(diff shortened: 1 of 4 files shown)


LefterisJP (approved):
LGTM

PR #5347, opened 2023-01-05
Generalize basic decoders between EVM chains

LefterisJP:
Some simple changes to make the common decoders work for multiple evm chains


LefterisJP:
Oh wow all green? Wtf.

PR #5351, opened 2023-01-05
Handle errors when querying missing receipts

yabirgb:
There was an unhandled remote error and also and unhandled integrity error

### DIFF
--- a/rotkehlchen/chain/evm/transactions.py
+++ b/rotkehlchen/chain/evm/transactions.py
@@ -489,7 +489,12 @@ class EvmTransactions(metaclass=ABCMeta):  # noqa: B024
 
             with self.database.user_write() as cursor:
                 for entry in hash_results:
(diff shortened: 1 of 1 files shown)


rotkehlchen/chain/evm/transactions.py
```diff
@@ -498,4 +503,5 @@ def get_receipts_for_transactions_missing_them(
                         )
                     except sqlcipher.IntegrityError as e:  # pylint: disable=no-member
                         if 'UNIQUE constraint failed: evmtx_receipts.tx_hash' not in str(e):
+                            log.error(f'Fialed to store transaction {entry.hex()} receipt due to {str(e)}')  # noqa: E501
```
LefterisJP, line 506:
```suggestion
                            log.error(f'Failed to store transaction {entry.hex()} receipt due to {str(e)}')  # noqa: E501
```

Issue #5176, opened 2022-11-25
Think of a way to warn users of the docker image to not expose rotki on public networks


kelsos:
## Abstract

We have a warning in our documentation, warning users to not expose rotki on public networks since the stuck was never designed to operate in such a way.

One user suggestion is to use an environment variable, to ensure that users have read the documentation before using the docker image. 

If the environment variable e.g. `ROTKI_ACCEPT_DOCKER_RISK` is not set explicitly by the user, the frontend should show a warning notifying the user of the potential risk and linking to the documentation.

We have to communicate this to DAppNode too so that the DAppNode package passes this variable to the rotki docker container.

## Motivation

Avoid having users accidentally expose their rotki instances.

## Specification

- Env var is only considered on dockerized instance otherwise it is ignored
- Update documentation with instructions on what needs to be done

PR #5354, opened 2023-01-06
Add accept_docker_risk boolean in info endpoint

LefterisJP:
This is the backend part of
https://github.com/rotki/rotki/issues/5176, essentially adding a way for the frontend to check if the user of the backend has set a specific environment variable.


LefterisJP (commented):
Will merge, since the only failed test is a timeout

PR #5356, opened 2023-01-06
Fix issue updating addressbook entries and blockchain null

yabirgb:
We were comparing a nullable column with the `=` operator and this resulted in errors while making the queries. This PR changes it to use the `IS` operator that works as `=` but also handles the `NULL` value

## Checklist

### DIFF
--- a/rotkehlchen/tests/api/test_addressbook.py
+++ b/rotkehlchen/tests/api/test_addressbook.py
@@ -6,6 +6,7 @@ import requests
 from eth_utils import to_checksum_address
 
 from rotkehlchen.chain.accounts import BlockchainAccountData
+from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.db.addressbook import DBAddressbook
 from rotkehlchen.db.ens import DBEns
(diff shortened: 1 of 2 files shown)


rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -265,10 +265,35 @@ def test_update_addressbook(
         )
         assert_error_response(
             response=response,
-            contained_in_msg='address "0x79B598976bD83a47CD8B428C824C8474311267b8" and blockchain ETH doesn\'t exist in the address book',  # noqa: E501
+            contained_in_msg='address "0x79B598976bD83a47CD8B428C824C8474311267b8" and blockchain ethereum doesn\'t exist in the address book',  # noqa: E501
             status_code=HTTPStatus.CONFLICT)
         assert db_addressbook.get_addressbook_entries(cursor) == entries_in_db_before_bad_patch
 
+    # add a new entry with no blockchain assigned
```
LefterisJP, line 272: Shouldn't you also have a test for deleting without blockchain to see this fix works for all cases? Also to get with no blockchain?

LefterisJP (approved):
LGTM

PR #5361, opened 2023-01-08
Add default values of backend args to info endpoint

LefterisJP:


docs/api.rst
```diff
@@ -10465,7 +10465,12 @@ Querying general information
...
               "data_directory": "/home/username/.local/share/rotki/data"
               "log_level": "DEBUG",
-	      "accept_docker_risk: false
+	      "accept_docker_risk: false,
+	      "backend_default_arguments": {
+	          "max-logfiles-num": 3,
+	          "max-size-in-mb-all-logs": 300,
+	          "sqlite-instructions": 5000,
+	      },
           },
```
yabirgb, line 10474: maybe we can change this in other PR since is green but the trailing `,` is something kelsos complains about

yabirgb (approved):
just a small comment that can be changed in other PR

docs/api.rst
LefterisJP: It's not just kelsos. It's the docs failing on invalid json. Weird it did not fail.

LefterisJP: I will change in another docs only PR though

PR #5362, opened 2023-01-08
Trades involving delisted bitfinex pairs are now recognized

LefterisJP:


LefterisJP:
User confirmed this works. Merging.

PR #5364, opened 2023-01-09
EVM Chain genesis, and ruff to 0.0.215 with many more checks.

LefterisJP:
- Upgrade ruff to 0.0.215. This introduces many changes in the code and some new rules due to
ruff activating many checkers, most of which we also will follow.
- Differentiate the genesis chain start logic between different evm chains.


LefterisJP:
Tests passed locally and the 2 failures are timeouts

PR #5365, opened 2023-01-09
feat: update blockie

lukicenturi:
Update blockie, move it to local

### DIFF
--- a/frontend/app/src/composables/accounts/blockie.ts
+++ b/frontend/app/src/composables/accounts/blockie.ts
@@ -1,0 +1,185 @@
+import { logger } from '@/utils/logging';
+
+// Source: https://github.com/ethereum/blockies
+// The random number is a js implementation of the Xorshift PRNG
+interface BlockieOptions {
+  /** seed used to generate icon data, default: random */
+  seed: string;
+  /** to manually specify the icon color, default: random */
+  color: string;
+  /** choose a different background color, default: random */
+  bgcolor: string;
+  /** width/height of the icon in blocks, default: 8 */
+  size: number;
+  /** width/height of each block in pixels, default: 4 */
+  scale: number;
+  /**
+   * each pixel has a 13% chance of being of a third color, default: random.
+   *
+   * Set to -1 to disable it.
+   *
+   * These "spots" create structures that look like eyes, mouths and noses.
+   */
+  spotColor: string;
+}
+
+const randSeed = new Array(4); // Xorshift: [x, y, z, w] 32 bit values
+
+const seedRand = (seed: string) => {
+  for (let i = 0; i < randSeed.length; i++) {
+    randSeed[i] = 0;
+  }
+  for (let i = 0; i < seed.length; i++) {
+    randSeed[i % 4] =
+      (randSeed[i % 4] << 5) - randSeed[i % 4] + seed.charCodeAt(i);
+  }
+};
+
+function rand() {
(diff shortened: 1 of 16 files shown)


frontend/app/src/composables/accounts/blockie.ts
```diff
@@ -0,0 +1,175 @@
...
+    if (!stored[formatted]) {
+      const canvas = createIcon({
+        seed: formatted
+      });
+
+      stored[formatted] = canvas.toDataURL('image/jpeg');
+      set(storedBlockie, stored);
+    }
+
+    return stored[formatted];
```
kelsos, line 169: The first thing we need here, is a limit and eviction.
Similar to what we do for the asset cache.

Is storing the `canvas.toDataUrl()` more efficient than storing the base64 itself? 

If I remember the performance hit is during generation

frontend/app/src/composables/accounts/blockie.ts
```diff
@@ -0,0 +1,175 @@
...
+  newOpts.size = opts.size || 8;
+  newOpts.scale = opts.scale || 4;
+  newOpts.color = opts.color || createColor();
+  newOpts.bgcolor = opts.bgcolor || createColor();
+  newOpts.spotColor = opts.spotColor || createColor();
+
+  return newOpts as BlockieOptions;
+}
+
+export function renderIcon(
```
kelsos, line 105: since we are moving this library internally then this export doesn't make much sense right?

frontend/app/src/composables/accounts/blockie.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  randSeed[0] = randSeed[1];
+  randSeed[1] = randSeed[2];
+  randSeed[2] = randSeed[3];
+  randSeed[3] = randSeed[3] ^ (randSeed[3] >> 19) ^ t ^ (t >> 8);
+
+  return (randSeed[3] >>> 0) / ((1 << 31) >>> 0);
+}
+
+export const createColor = () => {
```
kelsos, line 50: Also, this export if we are encapsulating the functionality, there is no need for `createColor` to be called outside of this composable.

frontend/app/src/composables/accounts/blockie.ts
lukicenturi: `canvas.toDataUrl()` will resulting the `base64` string

kelsos: Also, another important thing, can you also add some unit tests for the cache/eviction, we can just mock the actual blockie creation.

kelsos: > `canvas.toDataUrl()` will resulting the `base64` string. This is how the image is created.

OK thanks for clarifying, never looked at the library internals :)

kelsos: It would be preferable to have a limit and an eviction strategy for the cached blockies, similar to how we cache the asset information in the frontend.

My original comment was to add a similar mechanism here and add some test for that.

We can limit it to 100 cached entries and evict the least recently used items when we hit a limit.

lukicenturi: The limit and eviction strategy have added

kelsos (approved):

kelsos:
I rebased on top of develop, we can merge as soon as the CI passes

PR #5366, opened 2023-01-09
feat: add docker warning

### LINKED ISSUE #5176: Think of a way to warn users of the docker image to not expose rotki on public networks

lukicenturi:
Closes #5176

## Checklist

image

### DIFF
--- a/frontend/app/src/components/account-management/DockerWarning.vue
+++ b/frontend/app/src/components/account-management/DockerWarning.vue
@@ -1,0 +1,29 @@
+<script setup lang="ts">
+import { useMainStore } from '@/store/main';
+
+const { dockerRiskAccepted } = storeToRefs(useMainStore());
+const { tc } = useI18n();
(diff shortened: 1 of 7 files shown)


kelsos (approved):
lgtm @lukicenturi thank you.

Issue #5222, opened 2022-12-06
Enforce smaller transactions and single transaction per connection. Undo auto-savepoints.


LefterisJP:
## Abstract

In order to make things more explicit in how transactions are handled and fix bugs that happened, especially in DB upgrades, in https://github.com/rotki/rotki/pull/5203 we switched to using `isolation_level=None` and stop using the default python sqlite auto-commit mode.

Each write_ctx right now opens a transaction explicitly and closes it when going out, either normally or via an exception.

There is many places where we have multiple nested write_ctx()s in the codebase right now. Opening another transaction while one is progress is a BOOM-able offense.

In order to facilitiate for that and not make too many changes that PR cheats, by making any subsequently nested write_ctx actually open a savepoint under the hood.

This works fine, for 99% of the cases, but has some edge cases. These edge cases would be bad design on our part and they would not be noticed easily. Say three nested write contexts, with the 3rd commiting and the 2nd rolling back. The 3rd would also be rolled back since they were nested. And nesting may also come inadvedertedly through accessing the connection via different greenlets.

In order to get rid of all these we should keep things explicit.

## Task

- [x] Modify the write_ctx code to only do a single transaction and disallow opening a 2nd one. If trying to open a 2nd one it should wait/lock until the first one is finished. Remove the explicit savepoint logic.
- [x] Nested savepoints should work as they already do BUT only be used explicitly with the savepoint_ctx. (nothing actionable here I think -- just writing it to make it clear)
- [x] Go through code thanks to tests, and opening the app. Find all write_ctx usage that is wrong and fix it. Make sure we don't open a write_ctx without a reason, keep it as short as possible, and that it does not nest. If nesting is needed, with specific part rollback-able then this is exactly where savepoints should explicitly be used.

PR #5370, opened 2023-01-09
Enforce smaller DB transactions in evm transactions querying

LefterisJP:
- Part of https://github.com/rotki/rotki/issues/5222, where we want to make sure DB write transactions do not stay open for long as this creates problems described in the issue
- Remove old forgotten gevent.sleep(0) context switches

### DIFF
--- a/rotkehlchen/chain/evm/transactions.py
+++ b/rotkehlchen/chain/evm/transactions.py
@@ -4,7 +4,6 @@ from collections import defaultdict
 from contextlib import contextmanager
 from typing import TYPE_CHECKING, Iterator, Optional
 
-import gevent
 from gevent.lock import Semaphore
 from pysqlcipher3 import dbapi2 as sqlcipher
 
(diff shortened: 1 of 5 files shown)


rotkehlchen/chain/evm/transactions.py
```diff
@@ -264,43 +262,44 @@ def _query_and_save_internal_transactions_for_range(
...
+                        write_cursor=write_cursor,
                         transactions=[internal_tx],
                         relevant_address=None,  # no need to re-associate address
                     )
-
-                    if period.range_type == 'timestamps':
-                        assert location_string, 'should always be given for timestamps'
-                        log.debug(f'Internal {self.evm_inquirer.chain_name} transactions for {address} -> update range {period.from_value} - {timestamp}')  # noqa: E501
+                if period.range_type == 'timestamps':
+                    assert location_string, 'should always be given for timestamps'
```
yabirgb, line 284: should we keep assets like this in the code?

LefterisJP: You mean asserts? Some times you can do type ignore, some times you can do asserts. The key point is that asserts dissapear in production.

yabirgb: yes asserts. True, I forgot that it gets removed. Thanks

LefterisJP:
All green, will merge

PR #5375, opened 2023-01-10
Add post-decoding rules

nebolax:
Adds post-decoding rules that run after the main decoding loop.

### DIFF
--- a/rotkehlchen/chain/evm/decoding/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/decoder.py
@@ -2,6 +2,7 @@ import importlib
 import logging
 import pkgutil
 from abc import ABCMeta, abstractmethod
+from dataclasses import dataclass
 from types import ModuleType
 from typing import TYPE_CHECKING, Any, Callable, Optional, Protocol, Union
 
@@ -73,6 +74,28 @@ class EventDecoderFunction(Protocol):
         ...
 
 
+@dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=True)
+class DecodingRules():
+    address_mappings: dict[ChecksumEvmAddress, tuple[Any, ...]]
+    event_rules: list[EventDecoderFunction]
+    token_enricher_rules: list[Callable]  # enrichers to run for token transfers
+    post_decoding_rules: list[tuple[int, Callable]]  # rules to run after the main decoding loop
+    all_counterparties: set[str]
+
+    def __add__(self, other: 'DecodingRules') -> 'DecodingRules':
+        if not isinstance(other, DecodingRules):
+            raise TypeError(
+                f'Can only add DecodingRules to DecodingRules. Got {type(other)}',
+            )
+        return DecodingRules(
+            address_mappings={**self.address_mappings, **other.address_mappings},
+            event_rules=self.event_rules + other.event_rules,
+            token_enricher_rules=self.token_enricher_rules + other.token_enricher_rules,
+            post_decoding_rules=self.post_decoding_rules + other.post_decoding_rules,
+            all_counterparties=self.all_counterparties | other.all_counterparties,
+        )
+
+
 class EVMTransactionDecoder(metaclass=ABCMeta):
 
     def __init__(
@@ -98,7 +121,6 @@ class EVMTransactionDecoder(metaclass=ABCMeta):
         """
         self.database = database
         self.misc_counterparties = [CPT_GAS] + misc_counterparties
-        self.all_counterparties: set[str] = set()
         self.evm_inquirer = evm_inquirer
         self.transactions = transactions
         self.msg_aggregator = database.msg_aggregator
@@ -111,29 +133,41 @@ class EVMTransactionDecoder(metaclass=ABCMeta):
             is_non_conformant_erc721_fn=self._is_non_conformant_erc721,
             address_is_exchange_fn=self._address_is_exchange,
         )
-        self.event_rules = [
-            self._maybe_decode_erc20_approve,
-            self._maybe_decode_erc20_721_transfer,
-        ]
-        self.event_rules.extend(event_rules)
-        self.token_enricher_rules: list[Callable] = []  # enrichers to run for token transfers
+        self.rules = DecodingRules(
+            address_mappings={},
+            event_rules=[
+                self._maybe_decode_erc20_approve,
+                self._maybe_decode_erc20_721_transfer,
+            ],
+            token_enricher_rules=[],
+            post_decoding_rules=[],
+            all_counterparties=set(self.misc_counterparties),
+        )
+        self.rules.event_rules.extend(event_rules)
         self.value_asset = value_asset
-        self.initialize_all_decoders()
+        self.decoders: dict[str, 'DecoderInterface'] = {}
+        # Recursively check all submodules to get all decoder address mappings and rules
+        rules = self._recursively_initialize_decoders(self.chain_modules_root)
+        # Sort post decoding rules by priority (which is the first element of the tuple)
+        self.rules += rules
+        self.rules.post_decoding_rules.sort(key=lambda x: x[0], reverse=True)
         self.undecoded_tx_query_lock = Semaphore()
 
     def _recursively_initialize_decoders(
             self,
             package: Union[str, ModuleType],
-    ) -> tuple[
-            dict[ChecksumEvmAddress, tuple[Any, ...]],
-            list[Callable],
-            list[Callable],
-    ]:
+    ) -> DecodingRules:
         if isinstance(package, str):
             package = importlib.import_module(package)
-        address_results = {}
-        rules_results = []
-        enricher_results = []
+
+        results = DecodingRules(
+            address_mappings={},
+            event_rules=[],
+            token_enricher_rules=[],
+            post_decoding_rules=[],
+            all_counterparties=set(),
+        )
+
         for _, name, is_pkg in pkgutil.walk_packages(package.__path__):
             full_name = package.__name__ + '.' + name
             if full_name == __name__:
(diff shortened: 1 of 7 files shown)


rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -53,6 +53,10 @@ def enricher_rules(self) -> list[Callable]:  # pylint: disable=no-self-use
         """
         return []
 
+    def post_decoding_rules(self) -> list[Callable]:
+        """Subclasses may implement this to add post processing of the decoded events"""
```
LefterisJP, line 57:
```suggestion
        """Subclasses may implement this to add post processing of the decoded events. This will run after the normal decoding step and will only process decoded history events"""
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -244,6 +250,16 @@ def decode_by_address_rules(
 
         return result
 
+    def run_all_post_decoding_rules(
+            self,
+            transaction: EvmTransaction,
+            decoded_events: list[HistoryBaseEntry],
+            action_items: list[ActionItem],
```
LefterisJP, line 257: action items are not needed here, since they are supposed to be only used during the decoding step.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -167,23 +170,26 @@ def _recursively_initialize_decoders(
...
     def initialize_all_decoders(self) -> None:
         """Recursively check all submodules to get all decoder address mappings and rules
         """
         self.decoders: dict[str, 'DecoderInterface'] = {}
-        address_result, rules_result, enrichers_result = self._recursively_initialize_decoders(self.chain_modules_root)  # noqa: E501
+        address_result, rules_result, enrichers_result, post_decoding_results = self._recursively_initialize_decoders(self.chain_modules_root)  # noqa: E501
         self.address_mappings = address_result
         self.event_rules.extend(rules_result)
         self.token_enricher_rules.extend(enrichers_result)
+        self.post_decoding_rules.extend(post_decoding_results)
```
LefterisJP, line 192: We will also need to somehow specify order of processing here, as we need to guarantee that a certain decoder's post processing happens before some other one's.

For example I can see that you may need to require uniswap to run before cowswap or 1inch. I gave an idea for this below in the `post_decoding_rules` function

rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -53,6 +53,10 @@ def enricher_rules(self) -> list[Callable]:  # pylint: disable=no-self-use
         """
         return []
 
+    def post_decoding_rules(self) -> list[Callable]:
```
LefterisJP, line 56: In order to be able to specify order I think this should also return an int, which should indicate priority. So all post decoding rules that have priority 0 will run first. Then those with priority 1, those with 2 etc.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -167,23 +170,26 @@ def _recursively_initialize_decoders(
                     address_results.update(self.decoders[class_name].addresses_to_decoders())
                     rules_results.extend(self.decoders[class_name].decoding_rules())
                     enricher_results.extend(self.decoders[class_name].enricher_rules())
+                    post_decoding_results.extend(self.decoders[class_name].post_decoding_rules())  # noqa: E501
                     self.all_counterparties.update(self.decoders[class_name].counterparties())
 
-                recursive_addrs, recursive_rules, recurisve_enricher_results = self._recursively_initialize_decoders(full_name)  # noqa: E501
+                recursive_addrs, recursive_rules, recurisve_enricher_results, recursive_post_decoding_results = self._recursively_initialize_decoders(full_name)  # noqa: E501
```
LefterisJP, line 176: I know it was from before but please fix this typo
```suggestion
                recursive_addrs, recursive_rules, recursive_enricher_results, recursive_post_decoding_results = self._recursively_initialize_decoders(full_name)  # noqa: E501  
```

Also perhaps it makes sense to change this and add all those into a dataclass or named tuple. It's kind of silly to return a raw tuple of so many things and pass it around.

rotkehlchen/chain/evm/decoding/interfaces.py
nebolax: Makes sense, will add

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -164,28 +194,25 @@ def _recursively_initialize_decoders(
...
 
-                recursive_addrs, recursive_rules, recurisve_enricher_results = self._recursively_initialize_decoders(full_name)  # noqa: E501
-                address_results.update(recursive_addrs)
-                rules_results.extend(recursive_rules)
-                enricher_results.extend(recurisve_enricher_results)
+                recursive_results = self._recursively_initialize_decoders(full_name)
+                results += recursive_results
 
-        return address_results, rules_results, enricher_results
+        results.post_decoding_rules.sort(key=lambda x: x[1], reverse=True)
```
LefterisJP, line 206: I know what this is, but we will both forget in a week. Explain why this is sorting and what it's sorting on

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -164,28 +194,25 @@ def _recursively_initialize_decoders(
...
         """Recursively check all submodules to get all decoder address mappings and rules
         """
         self.decoders: dict[str, 'DecoderInterface'] = {}
-        address_result, rules_result, enrichers_result = self._recursively_initialize_decoders(self.chain_modules_root)  # noqa: E501
-        self.address_mappings = address_result
-        self.event_rules.extend(rules_result)
-        self.token_enricher_rules.extend(enrichers_result)
-        # update with counterparties not in any module
-        self.all_counterparties.update(self.misc_counterparties)
+        rules = self._recursively_initialize_decoders(self.chain_modules_root)
```
LefterisJP, line 213: I think this function is no longer needed. Include its contents directly in the constructor, the only place it's called.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -73,6 +74,28 @@ def __call__(
         ...
 
 
+@dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=True)
+class DecodingRules():
+    address_mappings: dict[ChecksumEvmAddress, tuple[Any, ...]]
+    event_rules: list[EventDecoderFunction]
+    token_enricher_rules: list[Callable]
+    post_decoding_rules: list[tuple[Callable, int]]
```
LefterisJP, line 82: nitpick: From a logical perspective I would expect priority to come first and the callable second, in a tuple

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -244,6 +271,15 @@ def decode_by_address_rules(
 
         return result
 
+    def run_all_post_decoding_rules(
+            self,
+            transaction: EvmTransaction,
+            decoded_events: list[HistoryBaseEntry],
+            all_logs: list[EvmTxReceiptLog],
+    ) -> None:
```
LefterisJP, line 279: Add a docstring here, explaining the format of `self.rules.post_decoding_rules` and mention that at this point it should be sorted by priority in ascending order

rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -53,6 +53,17 @@ def enricher_rules(self) -> list[Callable]:  # pylint: disable=no-self-use
         """
         return []
 
+    def post_decoding_rules(self) -> list[tuple[Callable, int]]:
+        """
+        Subclasses may implement this to add post processing of the decoded events.
+        This will run after the normal decoding step and will only process decoded history events.
+
+        This function should return a list of tuples where the first element is the function to
```
LefterisJP, line 61: This will need to be edited if you put priority as first

rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -53,6 +53,17 @@ def enricher_rules(self) -> list[Callable]:  # pylint: disable=no-self-use
...
+        run and the second element is the priority of the function. The higher the priority the
```
LefterisJP, line 62:
```suggestion
        run and the second element is the priority of the function. The higher the priority number the
```

LefterisJP (commented):
Also a few decoder related tests are failing, which you will have to adjust

rotkehlchen/chain/evm/decoding/decoder.py
nebolax: Makes sense :+1:

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -111,29 +133,41 @@ def __init__(
...
+        )
+        self.rules.event_rules.extend(event_rules)
         self.value_asset = value_asset
-        self.initialize_all_decoders()
+        self.decoders: dict[str, 'DecoderInterface'] = {}
+        # Recursively check all submodules to get all decoder address mappings and rules
+        rules = self._recursively_initialize_decoders(self.chain_modules_root)
+        # Sort post decoding rules by priority (which is the first element of the tuple)
+        self.rules += rules
+        self.rules.post_decoding_rules.sort(key=lambda x: x[0], reverse=True)
```
LefterisJP, line 153: nitpick: This should have gone above the actual sorting
```suggestion
        self.rules += rules
        # Sort post decoding rules by priority (which is the first element of the tuple)
        self.rules.post_decoding_rules.sort(key=lambda x: x[0], reverse=True)
```

LefterisJP (approved):
Approving and merging. Left a small nitpick which should be handled in another PR.

PR #5376, opened 2023-01-10
Handle assets collections in globaldb updates

yabirgb:
This PR allows to handle new assets collections and assets collections mappings in the SQL files for globaldb updates

### DIFF
--- a/rotkehlchen/globaldb/updates.py
+++ b/rotkehlchen/globaldb/updates.py
@@ -4,9 +4,10 @@ import re
 import sqlite3
 import sys
 from contextlib import suppress
+from enum import Enum, auto
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from typing import Any, Literal, NamedTuple, Optional, Union
+from typing import TYPE_CHECKING, Any, Literal, NamedTuple, Optional, Union
 
 import requests
 
@@ -14,7 +15,7 @@ from rotkehlchen.assets.asset import Asset
 from rotkehlchen.assets.resolver import AssetResolver
 from rotkehlchen.assets.types import AssetData, AssetType
 from rotkehlchen.constants.timing import DEFAULT_TIMEOUT_TUPLE
-from rotkehlchen.db.drivers.gevent import DBConnection, DBCursor
+from rotkehlchen.db.drivers.gevent import DBCursor
 from rotkehlchen.errors.asset import UnknownAsset
 from rotkehlchen.errors.misc import RemoteError
 from rotkehlchen.errors.serialization import DeserializationError
@@ -25,10 +26,22 @@ from rotkehlchen.user_messages import MessagesAggregator
 
 from .handler import GlobalDBHandler, initialize_globaldb
 
+if TYPE_CHECKING:
+    from rotkehlchen.db.drivers.gevent import DBConnection
+
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
 ASSETS_VERSION_KEY = 'assets_version'
+ASSETS_UPDATES_URL = 'https://raw.githubusercontent.com/rotki/assets/{branch}/updates/{version}/updates.sql'  # noqa: E501
+ASSET_COLLECTIONS_UPDATES_URL = 'https://raw.githubusercontent.com/rotki/assets/{branch}/updates/{version}/asset_collections_updates.sql'  # noqa: E501
+ASSET_COLLECTIONS_MAPPINGS_UPDATES_URL = 'https://raw.githubusercontent.com/rotki/assets/{branch}/updates/{version}/asset_collections_mappings_updates.sql'  # noqa: E501
+
+
+class UpdateFileType(Enum):
+    ASSETS = auto()
+    ASSET_COLLECTIONS = auto()
+    ASSET_COLLECTIONS_MAPPINGS = auto()
 
 
 def executeall(cursor: DBCursor, statements: str) -> None:
@@ -44,7 +57,7 @@ def executeall(cursor: DBCursor, statements: str) -> None:
 
 
 def _replace_assets_from_db(
-        connection: DBConnection,
+        connection: 'DBConnection',
         sourcedb_path: Path,
 ) -> None:
     cursor = connection.cursor()
@@ -55,10 +68,14 @@ def _replace_assets_from_db(
     DELETE FROM evm_tokens;
     DELETE FROM underlying_tokens_list;
     DELETE FROM common_asset_details;
+    DELETE FROM asset_collections;
+    DELETE FROM multiasset_mappings;
     INSERT INTO assets SELECT * FROM other_db.assets;
     INSERT INTO evm_tokens SELECT * FROM other_db.evm_tokens;
     INSERT INTO underlying_tokens_list SELECT * FROM other_db.underlying_tokens_list;
     INSERT INTO common_asset_details SELECT * FROM other_db.common_asset_details;
+    INSERT INTO asset_collections SELECT * FROM other_db.asset_collections;
+    INSERT INTO multiasset_mappings SELECT * FROM other_db.multiasset_mappings;
     INSERT OR REPLACE INTO settings(name, value) VALUES("{ASSETS_VERSION_KEY}",
     (SELECT value FROM other_db.settings WHERE name="{ASSETS_VERSION_KEY}")
     );
@@ -67,7 +84,7 @@ def _replace_assets_from_db(
     """)
 
 
-def _force_remote(cursor: DBCursor, local_asset: Asset, full_insert: str) -> None:
+def _force_remote_asset(cursor: DBCursor, local_asset: Asset, full_insert: str) -> None:
     """Force the remote entry into the database by deleting old one and doing the full insert.
 
     May raise an sqlite3 error if something fails.
@@ -79,6 +96,28 @@ def _force_remote(cursor: DBCursor, local_asset: Asset, full_insert: str) -> Non
     # Insert new entry. Since identifiers are the same, no foreign key constrains should break
     executeall(cursor, full_insert)
     AssetResolver().clean_memory_cache(local_asset.identifier.lower())
+
+
+def _query_file_or_rollback(url: str, connection: 'DBConnection') -> str:
+    """
+    Query the given url or rollback the connection if it was not possible to download it.
+    May raise:
+    - RemoteError if it was not possible to query GitHub
+    """
+    try:
+        response = requests.get(url=url, timeout=DEFAULT_TIMEOUT_TUPLE)
+    except requests.exceptions.RequestException as e:
+        connection.rollback()
+        raise RemoteError(f'Failed to query Github for {url} during assets update: {str(e)}') from e  # noqa: E501
+
+    if response.status_code != 200:
+        connection.rollback()
+        raise RemoteError(
+            f'Github query for {url} failed with status code '
+            f'{response.status_code} and text: {response.text}',
+        )
+
+    return response.text
 
 
 class ParsedAssetData(NamedTuple):
@@ -103,6 +142,8 @@ class AssetsUpdater():
         self.assets_re = re.compile(r'.*INSERT +INTO +assets\( *identifier *, *name *, *type *\) +VALUES\(([^\)]*?),([^\)]*?),([^\)]*?)\).*?')  # noqa: E501
         self.ethereum_tokens_re = re.compile(r'.*INSERT +INTO +evm_tokens\( *identifier *, *token_kind *, *chain *, *address *, *decimals *, *protocol *\) +VALUES\(([^\)]*?),([^\)]*?),([^\)]*?),([^\)]*?),([^\)]*?),([^\)]*?)\).*')  # noqa: E501
         self.common_asset_details_re = re.compile(r'.*INSERT +INTO +common_asset_details\( *identifier *, *symbol *, *coingecko *, *cryptocompare *, *forked *, *started *, *swapped_for *\) +VALUES\((.*?),(.*?),(.*?),(.*?),(.*?),([^\)]*?),([^\)]*?)\).*')  # noqa: E501
+        self.assets_collection_re = re.compile(r'.*INSERT +INTO +asset_collections\( *id *, *name *, *symbol *\) +VALUES +\(([^\)]*?),([^\)]*?),([^\)]*?)\).*?')  # noqa: E501
+        self.multiasset_mappings_re = re.compile(r'.*INSERT +INTO +multiasset_mappings\( *collection_id *, *asset *\) +VALUES +\(([^\)]*?), *"([^\)]+?)"\).*?')  # noqa: E501
         self.string_re = re.compile(r'.*"(.*?)".*')
         self.branch = 'master'
         if not getattr(sys, 'frozen', False):
@@ -269,7 +310,7 @@ class AssetsUpdater():
             token_kind,
         )
 
-    def _parse_full_insert(self, insert_text: str) -> AssetData:
+    def _parse_full_insert_assets(self, insert_text: str) -> AssetData:
         """Parses full insert line for an asset to give information for the conflict to the user
 
         Note: In the future this needs to be different for each version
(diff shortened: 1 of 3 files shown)


rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +305,219 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
             protocol=protocol,
         )
 
+    def _process_asset_collection(
```
nebolax, line 308: Question. In the current implementation identifiers of asset collections are just sequential integers 0, 1, 2, ...

We don't allow this yet, but if a user at some point adds a collection, and then later we run an asset collections update, then ... what identifiers will we use for the collections in the update?

Maybe we can later do something like a flag in the table that would help to distinguish, but not sure what is best to do here

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +305,219 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
+            try:
+                with connection.savepoint_ctx() as cursor:
+                    cursor.execute(full_insert)
+            except sqlite3.Error as e:
+                self.msg_aggregator.add_warning(
+                    f'Failed to edit asset collection mapping with asset {groups[1]} '
+                    f'and id {groups[0]}. Error: {str(e)}',
+                )
+
+    def _handle_asset_update(
```
nebolax, line 389: Kinda nitpicky, but why is this method called `_handle...` while the other two `_process...` ?

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +305,219 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
+        2. Try to deserialize it as an asset collection
+        3. Try to deserialize it as an asset collection mapping.
+
+        If none of the above rules sucess then the entry is skipped. In the case of assets
+        possible conflicts are handled.
+        """
         lines = text.splitlines()
         for action, full_insert in zip(*[iter(lines)] * 2):
-            if full_insert == '*':
+            if full_insert.replace(' ', '') == '*':
```
nebolax, line 469: Why is this needed?

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +305,219 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
                 full_insert = action
 
+            processed_entry = False
```
nebolax, line 472: Maybe is better to initialize `remote_asset_data` as `None` here instead of introducing a new flag?

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +305,219 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
+                    full_insert=full_insert,
+                    version=version,
                 )
                 continue
 
-            local_asset: Optional[Asset] = None
-            with suppress(UnknownAsset):
-                local_asset = Asset(remote_asset_data.identifier).check_existence()
+            # if it wasn't an asset let's check if it is an asset collection
+            try:
```
nebolax, line 491: Hmmm I would say that is better to use pure conditions instead of `try` ... `except`.

All three assets, collections and mappings functions start with regexps, so maybe we can move them here and use for determining the type of an update?

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +305,219 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
+            )[0]
+            self.conflicts.append((local_data, remote_asset_data))
+
     def _apply_single_version_update(
             self,
-            connection: DBConnection,
+            connection: 'DBConnection',
             version: int,
             text: str,
             conflicts: Optional[dict[Asset, Literal['remote', 'local']]],
```
nebolax, line 456: I would rename `conflicts` (also `self.conflicts`) if it is only for assets.

Upd: also, for example, `force_remote`, and generally all other only-assets methods

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +305,219 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
             protocol=protocol,
```
nebolax, line 305: Also I guess `_parse_full_insert` is good to be renamed to show that it is only for assets

rotkehlchen/globaldb/updates.py
```diff
@@ -477,8 +627,8 @@ def _perform_update(
                 version += 1
                 continue
 
+            url = f'https://raw.githubusercontent.com/rotki/assets/{self.branch}/updates/{version}/updates.sql'  # noqa: E501
```
nebolax, line 630: Hm, maybe move it to a constant, make something like

`'https://raw.githubusercontent.com/rotki/assets/{branch}/updates/{version}/updates.sql'`

and here just format?

rotkehlchen/tests/unit/test_globaldb_asset_updates.py
```diff
@@ -298,7 +305,7 @@ def test_parse_full_insert(assets_updater, text, expected_data, error_msg):
         assert error_msg in str(excinfo.value)
 
 
-def test_some_updates_are_malformed(assets_updater):
+def test_some_updates_are_malformed(assets_updater: AssetsUpdater) -> None:
```
nebolax, line 308: Hmmm (maybe also @LefterisJP's opinion will make sense here) we don't have typings for fixtures in other test. Why not to add them?

rotkehlchen/tests/unit/test_globaldb_asset_updates.py
```diff
@@ -324,7 +331,7 @@ def test_some_updates_are_malformed(assets_updater):
         connection=connection,
         version=999,  # doesn't matter
         text=update_text,
-        conflicts={'ETH': 'remote', 'BTC': 'remote'},
+        conflicts={A_ETH: 'remote', A_BTC: 'remote'},
```
nebolax, line 334: Just curious, why change this?

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +305,219 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
 
-            local_asset: Optional[Asset] = None
-            with suppress(UnknownAsset):
-                local_asset = Asset(remote_asset_data.identifier).check_existence()
+            # if it wasn't an asset let's check if it is an asset collection
+            try:
+                self._process_asset_collection(
+                    connection=connection,
+                    action=action,
+                    full_insert=full_insert,
```
nebolax, line 495: Hmm, I am not sure if `full_insert` is needed for collections (since we probably will never update collections, only add / delete, and then it won't be needed). But not sure.

nebolax (commented):
Thanks! Left some comments.
Also one test failed that seems to be related, but the error looks weird.
```
FAILED rotkehlchen/tests/api/test_assets_updates.py::test_simple_update[True] - AssertionError: assert 'Euro' == 'Ευρώ'
  - Ευρώ
  + Euro
```

rotkehlchen/tests/unit/test_globaldb_asset_updates.py
yabirgb: Is easier if you do the introduction of types step by step than all in one big PR. I did those because are in the same file I'm changing and is related to my changes.

rotkehlchen/tests/unit/test_globaldb_asset_updates.py
yabirgb: the type of conflicts is `asset -> str` and not `str -> str`

rotkehlchen/globaldb/updates.py
yabirgb: While writing the tests I came to a line that required this due to how it was formatted. I believe is not that painful but if not I can change. We take care of this but since we are comparing to `'*'` I believe it makes sense

rotkehlchen/tests/unit/test_globaldb_asset_updates.py
nebolax: I just didn't see then, did you change anything in any other place of the code? If not, then how was this working before?

rotkehlchen/globaldb/updates.py
nebolax: Yeah, I don't see any big problem here. The only thing is that maybe `.strip()` is better to use

yabirgb: ah good. I tough strip didn't remove if them it was after. Thanks

rotkehlchen/tests/unit/test_globaldb_asset_updates.py
yabirgb: because when comparing assets to string we do a string comparasion and we were doing 'ETH' == 'ETH' basically at some point that is what happens when you do `A_ETH == 'ETH'`

rotkehlchen/globaldb/updates.py
yabirgb: We can rename them for example when an asset changes its name

rotkehlchen/globaldb/updates.py
yabirgb: It was intentionally because this one is bit more complex and not just reading the info and putting it in the database

nebolax: Alright then :+1:

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +306,230 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
+        """
+        groups = match.groups()
+        if len(groups) != 2:
+            raise DeserializationError(
+                f'At asset DB update could not parse asset collection data out of {action}',
+            )
 
+        # check that the asset exists and so does the collection
+        Asset(groups[1]).check_existence()
+        with connection.cursor() as cursor:
```
LefterisJP, line 357: nitpick: Due to the way we have written the code it's equivalent (right now), but we want to be explicit and see it's a read only cursor
```suggestion
        with connection.read_ctx() as cursor:
```

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +306,230 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
+                except sqlite3.Error as e:
+                    self.msg_aggregator.add_warning(
+                        f'Failed to resolve conflict for {remote_asset_data.identifier} in '
+                        f'the DB during the v{version} assets update. Skipping entry. '
+                        f'Error: {str(e)}',
+                    )
+                return  # fail or succeed continue to next entry
+
+            # else can't resolve. Mark it for the user to resolve.
+            # TODO: When assets refactor is finished, remove the usage of AssetData here
```
LefterisJP, line 436: question: This todo still relevant? And I think I have asked that same question before. What do we mean by "assets refactor" here?

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +306,230 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
+
+    def _apply_single_version_update(
+            self,
+            connection: 'DBConnection',
+            version: int,
+            text: str,
+            assets_conflicts: Optional[dict[Asset, Literal['remote', 'local']]],
+    ) -> None:
+        """
+        Process each action in the queried data and process it by:
```
LefterisJP, line 452:
```suggestion
        Process each action in the queried data and try the following in order:
```

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +306,230 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
+            if (
+                (match := self.assets_re.match(full_insert)) is not None and
+                len(match.groups()) != 0
+            ):
+                remote_asset_data = None
+                try:
+                    remote_asset_data = self._parse_full_insert_assets(full_insert)
+                except DeserializationError as e:
+                    self.msg_aggregator.add_warning(
+                        f'Skipping entry during assets update to v{version} due '
```
LefterisJP, line 474: Wouldn't it make sense to add a log entry with the raw entry we skip? Since by seeing this in the logs we won't be able to know which entry it's talking about

LefterisJP (commented):
Okay so reviewed. Left some comments.

But also have an important question on approach.

Just like with this other PR you had where you combined stuff in one file I see you try to do the same thing here.

This does not scale and is unnecessary complicated. Change this.
In the remote we will have 1 file for assets, 1 for collections, 1 for mappings, 1 for whatever else we need.

This way you won't need to try to match every potential rule to each line. You just take the assets file and everything should be either an asset or if its' not it's an error. Same with the other stuff.

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +341,232 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
-            connection: DBConnection,
+            connection: 'DBConnection',
             version: int,
             text: str,
-            conflicts: Optional[dict[Asset, Literal['remote', 'local']]],
+            assets_conflicts: Optional[dict[Asset, Literal['remote', 'local']]],
+            update_file_type: UpdateFileType,
     ) -> None:
+        """
+        Process each action in the queried data and try the following in order:
```
LefterisJP, line 502: Why did you copy paste my previous docstring suggestion when you now have actually split it in 3 files? The docstring is no longer relevant and needs changing.

rotkehlchen/globaldb/updates.py
LefterisJP: you still have not done this

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +306,230 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
+                    self._process_asset_collection(
+                        connection=connection,
+                        action=action,
+                        full_insert=full_insert,
+                        collection_match=match,
+                    )
+                    continue
+                except DeserializationError as e:
+                    self.msg_aggregator.add_warning(
+                        f'Skipping entry during assets collection update to v{version} due '
```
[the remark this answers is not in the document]
LefterisJP: this either.

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +306,230 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
-                # TODO: When assets refactor is finished, remove the usage of AssetData here
-                local_data = GlobalDBHandler().get_all_asset_data(  # pylint: disable=unsubscriptable-object  # noqa: E501
-                    mapping=False,
-                    serialized=False,
-                    specific_ids=[local_asset.identifier],
-                )[0]
-                self.conflicts.append((local_data, remote_asset_data))
+                except DeserializationError as e:
+                    self.msg_aggregator.add_warning(
+                        f'Skipping entry during assets collection multimapping update to '
```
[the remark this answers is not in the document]
LefterisJP: again this neither

rotkehlchen/globaldb/updates.py
```diff
@@ -477,25 +676,43 @@ def _perform_update(
...
-                connection.rollback()
-                raise RemoteError(
-                    f'Github query for {url} failed with status code '
-                    f'{response.status_code} and text: {response.text}',
-                )
+            assets_url = ASSETS_UPDATES_URL.format(branch=self.branch, version=version)
+            asset_collections_url = ASSET_COLLECTIONS_UPDATES_URL.format(branch=self.branch, version=version)  # noqa: E501
+            asset_collections_mappings_url = ASSET_COLLECTIONS_MAPPINGS_UPDATES_URL.format(branch=self.branch, version=version)  # noqa: E501
 
+            assets_file = _query_file_or_rollback(
```
LefterisJP, line 683: I am a bit confused with this rollback mechanism. Don't we use contexts here? How does this work? Rollback to what? The way the DB is accessed in this area, after the changes that happened seems dangerous

rotkehlchen/globaldb/updates.py
```diff
@@ -300,74 +341,232 @@ def _parse_full_insert(self, insert_text: str) -> AssetData:
...
+        collection_match = self.assets_collection_re.match(full_insert)
+        if collection_match is None:
+            log.error(f'Failed to match asset collection {full_insert}')
+            raise DeserializationError(
+                f'At asset DB update could not parse asset collection data out of {action}',
+            )
+
+        groups = collection_match.groups()
+        if len(groups) != 3:
+            log.error(f'Asset collection {full_insert} does not have the excpected elements')
```
LefterisJP, line 360:
```suggestion
            log.error(f'Asset collection {full_insert} does not have the expected elements')
```

rotkehlchen/globaldb/updates.py
yabirgb: Discussed in discord. This will come in a future PR

rotkehlchen/globaldb/updates.py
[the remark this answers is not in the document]
yabirgb: Different errors are logged inside the function depending on the section that failed https://github.com/rotki/rotki/pull/5376/files#diff-17df0f14bc17cda693441ce72a132c9d77830a0e2685c8fc77f5a8f1a9a6d1dbR399

rotkehlchen/globaldb/updates.py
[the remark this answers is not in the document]
yabirgb: same as above https://github.com/rotki/rotki/pull/5376/files#diff-17df0f14bc17cda693441ce72a132c9d77830a0e2685c8fc77f5a8f1a9a6d1dbR353

rotkehlchen/globaldb/updates.py
yabirgb: For the updates we create an intermediate database and then copy the new assets from it. https://github.com/rotki/rotki/blob/25a3d39e9919917bdffcae36f84f70927fa17697/rotkehlchen/globaldb/updates.py#L405

We modify the database here https://github.com/rotki/rotki/blob/25a3d39e9919917bdffcae36f84f70927fa17697/rotkehlchen/globaldb/updates.py#L466 and sqlite creates a transaction that is not commited. The rollbacks undoes that queries. Also for the rollback at the end of the file, since what we do is `INSERT` and `UPDATE`s the transaction is not commited so it can be rolled back

LefterisJP (commented):
Only thing that failed is: `rotkehlchen/tests/api/test_evm_transactions.py::test_query_transactions`

Don't think it's related and it seems that test needs some more love as it fails recently a lot.

Will probably merge but am leaving some small ToDOs that can go to another PR.
Also we have two followups that need to happen after this:

1. The long-standing TODO that we said would happen
2. Explicit transaction

PR #5377, opened 2023-01-11
Optimism airdrop, delegate change and accountants refactor

LefterisJP:
So this is one of those PRs where you start by trying to add a small thing (optimism airdrop), and you find out a lot more things should be done and you go ahead and do them. For more details read the commits but will break it down here:

### Create per-chain EvmAccountingAggregator and a grouping class
This commit moves the core logic of aggregating per-module evm
accountants into a base class for all evm chains.

Then two different classes are instantiated. One for Ethereum and one
for Optimism. The reason for this is:

1. Each accountant also gets passed the inquirer class of the chain so
it can query things on-chain if it needs to. This would probably make
more sense only if there is archive capability.
2. Subclassing instead of generating two EvmAccountingAggregator
instances with different arguments is due to the potential of having
different logic per evm chain, just like we do with the decoders themselves.

### Add decoder for optimisim delegate change
This also touches the "Governance Propose" subtype and generalizes it
to "Governance". It was the only governance related subtype, so this
made sense. The action details are given in the `notes` part of the
event. If we ever want to create per governance actions subtypes we
can revisit this, but it probably won't be necessary.

This last one also touches frontend a tiny bit


LefterisJP:
> This changes affected other decoders as can be seen in the tests

It was due to a change I did in the `__str__()` of `FVal` to suppress scientific representation. I abstracted it out in its own commit (last commit), so it's better contained along with the test adjustments.

LefterisJP:
All green. Merging.

PR #5379, opened 2023-01-11
Anukul/supported blockchains endpoint

anukul:
New response:

```json
{
    "message": "",
    "result": [
        {
            "evm_chain_name": "ethereum",
            "id": "ETH",
            "name": "Ethereum",
            "type": "evm"
        },
        {
            "id": "ETH2",
            "name": "Beacon chain validator",
            "type": "eth2"
        },
        {
            "id": "BTC",
            "name": "Bitcoin",
            "type": "bitcoin"
        },
        {
            "id": "BCH",
            "name": "Bitcoin Cash",
            "type": "bitcoin"
        },
        {
            "id": "KSM",
            "name": "Kusama",
            "type": "substrate"
        },
        {
            "evm_chain_name": "avalanche",
            "id": "AVAX",
            "name": "Avalanche",
            "type": "evm"
        },
        {
            "id": "DOT",
            "name": "Polkadot",
            "type": "substrate"
        },
        {
            "evm_chain_name": "optimism",
            "id": "OPTIMISM",
            "name": "Optimism",
            "native_asset": "ETH",
            "type": "evm"
        }
    ]
}
```


anukul:
I see everybody is using forks. Will delete this branch after merge and go back to fork.

rotkehlchen/types.py
```diff
@@ -417,6 +417,15 @@ def get_chain_type(self) -> str:
         # else
         return 'eth2'  # the outlier
 
+    def get_name(self) -> str:
+        return SUPPORTED_BLOCKCHAIN_TO_NAME[self]
+
+    def get_native_asset(self) -> str:
```
nebolax, line 423: Right now this doesn't return an asset, but just a string. If the asset is needed, I would return `A_ETH`, `A_BTC`, etc. then.

nebolax (commented):
I would squash the commits.

Also there is a failed test that is probably related `FAILED rotkehlchen/tests/api/test_misc.py::test_query_supported_chains - AssertionError: Did not find ethereum in the supported chains result`

Also I think that a small test won't hurt.

rotkehlchen/types.py
anukul: I don't mind decoupling native asset from the enum value.
That would probably also imply always returning `native_asset` key in the response.

anukul:
Fixed (and updated) failing test, should hopefully increase coverage once all tests finish running.

Also squashed commits.

rotkehlchen/types.py
LefterisJP: We don't want `native_asset` in every entry. Only when it's different from the short name. So right now only in Optimism.

Where is this function used? If it's only used for this new api, then perhaps does not even make sense to create it. Just add this logic directly in the api endpoint and if we ever need this we would create such a fucntion here.

Cause you would need to indeed return an asset and not a string here. But returning constant assets here will complicate the code and will probably hit you with import problems as this is `types.py` a very central file.

docs/api.rst
```diff
@@ -2041,19 +2041,19 @@ Getting all supported chains
...
-                {"name": "AVAX", "type": "evm", "evm_chain_name": "avalanche"},
-                {"name": "ETH2", "type": "eth2"},
-                {"name": "DOT", "type": "substrate"},
-                {"name": "KSM", "type": "substrate"},
-                {"name": "BTC", "type": "bitcoin"},
-                {"name": "BCH", "type": "bitcoin"}
+                {"id": "ETH", "name": "Ethereum", "type": "evm", "evm_chain_name": "ethereum"},
+                {"id": "OPTIMISM", "name": "Optimism", "type": "evm", "evm_chain_name": "optimism", "native_asset": "ETH"},
+                {"id": "AVAX", "name": "Avalanche", "type": "evm", "evm_chain_name": "avalanche"},
+                {"id": "ETH2", "name": "Beacon chain validator", "type": "eth2"},
```
LefterisJP, line 2047:
```suggestion
                {"id": "ETH2", "name": "Ethereum Beacon chain", "type": "eth2"},
```

rotkehlchen/types.py
```diff
@@ -417,6 +417,15 @@ def get_chain_type(self) -> str:
         # else
         return 'eth2'  # the outlier
 
+    def get_name(self) -> str:
+        return SUPPORTED_BLOCKCHAIN_TO_NAME[self]
```
LefterisJP, line 421: you don't need this. Why are you adding a "name" here?

1. Was this needed by the frontend?
2. This is already happening by the `str()` method inherited by `SerializableEnumValueMixin2`. So you don't need a new method here nor an entire new mapping dictionary just to capitalize it. Why do they need capitalization?

rotkehlchen/types.py
anukul: updated

rotkehlchen/types.py
anukul: > Was this needed by the frontend?

yes, based on discussion in discord thread

> Why do they need capitalization?

i believe they need capitalization for display purposes here



but this can be handled on frontend and we can avoid the additonal mapping

LefterisJP: yes they should handle if if they need. Just use the str() method.

anukul: updated

LefterisJP (approved):
LGTM. No idea why the backend tests took so long to run in this case. Perhaps the multiple subsequent pushes slowed it down. Try to limit the number of pushes as I think each one creates a new CI build and am not sure if it cancels the previous one.

Which we should check. If it does not we should find a way to make it do so.

kelsos:
> LGTM. No idea why the backend tests took so long to run in this case. Perhaps the multiple subsequent pushes slowed it down. Try to limit the number of pushes as I think each one creates a new CI build and am not sure if it cancels the previous one.
> 
> Which we should check. If it does not we should find a way to make it do so.

It should auto-cancel the previous one, I have it enabled.

PR #5383, opened 2023-01-11
Events details

nebolax:
Adds an endpoint to retrieve history-base-entries' details to show to the user

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2361,9 +2361,9 @@ Querying evm transactions
                 "notes": "Burned 0.00863351371344 ETH for gas",
                 "sequence_index": 0,
                 "timestamp": 1642802807,
-                "extra_data": null
               },
-              "customized": false
+              "customized": false,
+              "has_details": false
             }, {
               "entry": {
                 "tx_hash": "0x867119d6c66cab26561ccc5775c9cd215389efb2e3832e54baed2a0a34498c4b",
@@ -2397,9 +2397,9 @@ Querying evm transactions
                     "location_label": "0xF2Eb18a344b2a9dC769b1914ad035Cbb614Fd238",
                     "notes": "Receive reward of 0.002916189466526136 ETH from Liquity's staking",
                     "counterparty": "liquity",
-                    "extra_data": null
                   },
-                  "customized": false
+                  "customized": false,
+                  "has_details": true
                 },
                 {
                   "entry": {
@@ -2418,12 +2418,9 @@ Querying evm transactions
                     "location_label": "0xF2Eb18a344b2a9dC769b1914ad035Cbb614Fd238",
                     "notes": "Stake 408.16397184365102401 LQTY in the Liquity protocol",
                     "counterparty": "liquity",
-                    "extra_data": {
-                      "staked_amount": "1477.541875499512958985",
-                      "asset": "eip155:1/erc20:0x6DEA81C8171D0bA574754EF6F8b412F2Ed88c54D"
-                    }
                   },
-                  "customized": false
+                  "customized": false,
+                  "has_details": false
                 },
                 {
                   "entry": {
@@ -2442,9 +2439,9 @@ Querying evm transactions
                     "location_label": "0xF2Eb18a344b2a9dC769b1914ad035Cbb614Fd238",
                     "notes": "Receive reward of 21.405502592141755289 LUSD from Liquity's staking",
                     "counterparty": "liquity",
-                    "extra_data": null
                   },
-                  "customized": false
+                  "customized": false,
+                  "has_details": false
                 }
               ],
               "ignored_in_accounting": false
@@ -2462,9 +2459,9 @@ Querying evm transactions
                 "notes": "Send 0.096809163374771208 ETH 0x6e15887E2CEC81434C16D587709f64603b39b545 -> 0xA090e606E30bD747d4E6245a1517EbE430F0057e",
                 "sequence_index": 1,
                 "timestamp": 1642802807,
-                "extra_data": null
               },
-              "customized": true
+              "customized": true,
+              "has_details": false
             }]
           }, {
             "entry": {
@@ -2496,9 +2493,9 @@ Querying evm transactions
                 "notes": "Burned 0.00863351371344 ETH for gas",
                 "sequence_index": 0,
                 "timestamp": 1642802807,
-                "extra_data": null
               },
-              "customized": false
+              "customized": false,
+              "has_details": false
             }]
           }],
           "entries_found": 95,
@@ -2512,7 +2509,7 @@ Querying evm transactions
    :resjson object result: A list of transaction entries to return for the given filter.
    :resjson object entry: A single transaction entry
    :resjson bool ignored_in_accounting: A boolean indicating whether this transaction should be ignored in accounting or not
-   :resjson list decoded_events: A list of decoded events for the given transaction. Each even is an object comprised of the event entry and a boolean denoting if the event has been customized by the user or not.
+   :resjson list decoded_events: A list of decoded events for the given transaction. Each event is an object comprised of the event entry and a boolean denoting if the event has been customized by the user or not. Each entry also has a `has_details` flag. If `has_details` is true, then it is possible to call /history/events/details endpoint to retrieve some extra information about the event.
    :resjson int entries_found: The number of entries found for the current filter. Ignores pagination.
    :resjson int entries_limit: The limit of entries if free version. -1 for premium.
    :resjson int entries_total: The number of total entries ignoring all filters.
@@ -12121,4 +12118,51 @@ Custom Assets
 
    :statuscode 200: Custom asset types retrieved successfully.
    :statuscode 409: No user is currently logged in. Check error message.
+   :statuscode 500: Internal rotki error.
+
+
+Events Details
+================
+
+.. http:get:: /api/(version)/history/events/details
+
+   Doing a GET on this endpoint will return the details of a history event.
+
+   **Example Request**:
+
+   .. http:example:: curl wget httpie python-requests
+
+      GET /api/1/history/events/details HTTP/1.1
+      Host: localhost:5042
+      Content-Type: application/json;charset=UTF-8
+
+      {"identifier": 137}
+
+   :reqjson int identifier: The identifier of the event to be queried.
+
+   **Example Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+          "result": {
+              "sub-swaps": [
+                  {"from_amount": "100.0", "to_amount": "0.084", "from_asset": "eip155:1/erc20:0x6B3595068778DD592e39A122f4f5a5cF09C90fE2", "to_asset": "ETH"},
+                  {"from_amount": "0.084", "to_amount": "98.2", "from_asset": "ETH", "to_asset": "eip155:1/erc20:0xdAC17F958D2ee523a2206206994597C13D831ec7"}
+              ]
+          },
+          "message": ""
+      }
+
+   :resjson object result: A dictionary with the details. It may contain one of the following amount of details.
+   :resjson list sub-swaps: A list with the details of each sub-swap. Each entry contains the following keys: from_amount, to_amount, from_asset, to_asset.
+   :resjson dict liquity-staking: Information about assets that were staked in liquity in this event.
+
+   :statuscode 200: The details were returned successfully.
+   :statuscode 400: Provided JSON is in some way malformed.
+   :statuscode 404: There is no event with the provided identifier or the event has no details to be returned.
+   :statuscode 409: No user is currently logged in.
    :statuscode 500: Internal rotki error.
--- a/rotkehlchen/accounting/structures/base.py
+++ b/rotkehlchen/accounting/structures/base.py
@@ -71,6 +71,14 @@ HISTORY_EVENT_DB_TUPLE_WRITE = tuple[
     Optional[str],  # extra_data
 ]
 
+SUB_SWAPS_DETAILS = 'sub-swaps'
+LIQUITY_STAKING_DETAILS = 'liquity-staking'
+
+ALL_DETAILS_KEYS = {
+    SUB_SWAPS_DETAILS,
+    LIQUITY_STAKING_DETAILS,
+}
+
 
 def get_tx_event_type_identifier(event_type: HistoryEventType, event_subtype: HistoryEventSubType, counterparty: str) -> str:  # noqa: E501
     return str(event_type) + '__' + str(event_subtype) + '__' + counterparty
@@ -203,6 +211,11 @@ class HistoryBaseEntry(AccountingEventMixin):
             'extra_data': self.extra_data,
         }
 
+    def serialize_without_extra_data(self) -> dict[str, Any]:
+        result = self.serialize()
+        result.pop('extra_data')
+        return result
+
     @classmethod
     def deserialize(cls, data: dict[str, Any]) -> 'HistoryBaseEntry':
         """Deserializes a dict history base entry to HistoryBaseEntry object.
@@ -276,6 +289,18 @@ class HistoryBaseEntry(AccountingEventMixin):
 
     def get_assets(self) -> list[Asset]:
         return [self.asset]
+
+    def has_details(self) -> bool:
+        if self.extra_data is None:
+            return False
+        return len(self.extra_data.keys() & ALL_DETAILS_KEYS) > 0
+
+    def get_details(self) -> Optional[dict[str, Any]]:
+        if self.extra_data is None:
+            return None
+
+        details = {k: v for k, v in self.extra_data.items() if k in ALL_DETAILS_KEYS}
+        return details if len(details) > 0 else None
 
     def process(
             self,
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -3348,7 +3348,8 @@ class RestAPI():
                         'entry': entry.serialize(),
                         'decoded_events': [
                             {
-                                'entry': x.serialize(),
+                                'entry': x.serialize_without_extra_data(),
+                                'has_details': x.has_details(),
                                 'customized': x.identifier in customized_event_ids,
                             } for x in events
                         ],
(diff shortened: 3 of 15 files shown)


rotkehlchen/accounting/mixins/event.py
```diff
@@ -76,11 +76,11 @@ def process(
         ...
 
     @abstractmethod
-    def serialize(self) -> dict[str, Any]:
+    def serialize(self, to_debug: bool) -> dict[str, Any]:
```
nebolax, line 79: This `to_debug` flag is something that I am not sure about. But I added it because when exporting a json with events for debugging, `extra_data` is needed there. While in other places it is not.

docs/api.rst
```diff
@@ -2512,7 +2509,7 @@ Querying evm transactions
    :resjson object result: A list of transaction entries to return for the given filter.
    :resjson object entry: A single transaction entry
    :resjson bool ignored_in_accounting: A boolean indicating whether this transaction should be ignored in accounting or not
-   :resjson list decoded_events: A list of decoded events for the given transaction. Each even is an object comprised of the event entry and a boolean denoting if the event has been customized by the user or not.
+   :resjson list decoded_events: A list of decoded events for the given transaction. Each event is an object comprised of the event entry and a boolean denoting if the event has been customized by the user or not. Each entry also has a `has_details` flag. If `has_details` is true, then it is possible to call /history/events/details endpoint to retrieve some useful details about the event.
```
yabirgb, line 2512:
```suggestion
   :resjson list decoded_events: A list of decoded events for the given transaction. Each event is an object comprised of the event entry and a boolean denoting if the event has been customized by the user or not. Each entry also has a `has_details` flag. If `has_details` is true, then it is possible to call /history/events/details endpoint to retrieve some extra information about the event.
```

rotkehlchen/accounting/structures/base.py
```diff
@@ -277,6 +280,17 @@ def get_identifier(self) -> str:
     def get_assets(self) -> list[Asset]:
         return [self.asset]
 
+    def has_details(self) -> bool:
+        return self.extra_data is not None and 'sub-swaps' in self.extra_data
```
yabirgb, line 284: For example the liquity events have `self.extra_data` non empty. If you set the extra condition of `sub-swaps`  the transactions in the liquity page need to handle this scenario too where the flag is false but they do have information in the extra details field. Is this something that you have discussed with the fronted?

rotkehlchen/accounting/mixins/event.py
yabirgb: it is also needed in liquity transactions for the staking page

rotkehlchen/api/v1/schemas.py
```diff
@@ -2854,3 +2854,20 @@ def make_nft_filter_query(  # pylint: disable=no-self-use
             'ignore_cache': data['ignore_cache'],
             'filter_query': filter_query,
         }
+
+
+class EventDetailsQuerySchema(Schema):
+    event_identifier = EVMTransactionHashField(required=True)
+    sequence_index = fields.Integer(required=True)
+
```
yabirgb, line 2862: shouldn't the frontend requests multiple for the same transaction? If a tx has for example 3 swaps you have to query this endpoint 3 times. Is this something agreed with the frontend?

rotkehlchen/accounting/mixins/event.py
yabirgb, line 79: this is something only needed for the history events class and that made you change the rest of the code. I would create a new method and use that one, for example `serialize_basic_information` or any other name that you find better. Then you can use it where it is needed

rotkehlchen/tests/api/test_evm_transactions.py
```diff
@@ -94,3 +105,122 @@ def test_query_transactions(rotkehlchen_api_server: 'APIServer'):
...
+        event_type=HistoryEventType.TRADE,
+        event_subtype=HistoryEventSubType.RECEIVE,
+        asset=A_USDT,
+        balance=Balance(amount=FVal(98.2)),
+        extra_data={
+            'sub-swaps': [
+                {'amount_in': '100.0', 'amount_out': '0.084', 'from_asset': A_SUSHI.identifier, 'to_asset': A_ETH.identifier},  # noqa: E501
+                {'amount_in': '0.084', 'amount_out': '98.2', 'from_asset': A_ETH.identifier, 'to_asset': A_USDT.identifier},  # noqa: E501
+            ],
+            'some-internal-data': 'some data',  # this data shouldn't be returned to the frontend
```
yabirgb, line 153: As I said there is some extra information about staking that is returned in the case of liquity to the frontend for the staking view

docs/api.rst
```diff
@@ -12122,3 +12119,50 @@ Custom Assets
...
+
+   .. http:example:: curl wget httpie python-requests
+
+      POST /api/1/assets/custom HTTP/1.1
+      Host: localhost:5042
+      Content-Type: application/json;charset=UTF-8
+
+      {"event_identifier": "0x66e3d2686193e01a4fbf0f598872236402edfe2f4efad84c4f6cdc753b8c78e3", "sequence_index": "222"}
+
+   :reqjson int[optional] event_identifier: The identifier of the event to be queried.
```
yabirgb, line 12141: is not an integer, is a string. Also is not optional

docs/api.rst
```diff
@@ -12122,3 +12119,50 @@ Custom Assets
...
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+          "result": {
+              "sub-swaps": [
+                  {"amount_in": "100.0", "amount_out": "0.084", "from_asset": A_SUSHI.identifier, "to_asset": A_ETH.identifier},  # noqa: E501
+                  {"amount_in": "0.084", "amount_out": "98.2", "from_asset": A_ETH.identifier, "to_asset": A_USDT.identifier},  # noqa: E501
+              ],
```
yabirgb, line 12156: the sequence index is optional. How is the response when there are multiples?

docs/api.rst
```diff
@@ -12122,3 +12119,50 @@ Custom Assets
...
+          },
+          "message": ""
+      }
+
+   :resjson object result: A dictionary with the details. If there is no details, an empty dictionary is returned.
+   :resjson list sub-swaps: A list with the details of each sub-swap. Each entry contains the following keys: amount_in, amount_out, from_asset, to_asset.
+
+   :statuscode 200: The details were returned successfully.
+   :statuscode 400: Provided JSON is in some way malformed.
+   :statuscode 404: There is no event with the provided event_identifier and sequence_index.
```
yabirgb, line 12166:
```suggestion
   :statuscode 404: There is no event with the provided event_identifier or sequence_index.
```

yabirgb (commented):
A few comments and questions

rotkehlchen/accounting/structures/base.py
nebolax: But the liquity information in extra_data is not shown on the frontend and is needed only for accounting on backend

yabirgb: Yes, it is shown in the frontend in the staking section. Not in the evm transactions but yes in the staking page. Is something that has to be known.

docs/api.rst
nebolax: No, it's not optional, it's required. It was a mistake in the docs

docs/api.rst
nebolax: Why this change? The events are matched by event_identifier + sequence_index

yabirgb: it was due to how the docs were. If it was a mistaken then it is an `and`. If it was optional then is `or`

rotkehlchen/api/v1/schemas.py
nebolax: Yes, the idea of returning the details is to keep as minimum data on frontend as possible. So if there are 3 swaps, users will open/expand them one-by-one, not all 3 at the same time. So to keep minimum info on the frontend they request the details only for a specific event

rotkehlchen/api/rest.py
```diff
@@ -4951,3 +4952,21 @@ def get_custom_asset_types(self) -> Response:
...
+            events = dbevents.get_history_events(
+                cursor=cursor,
+                filter_query=filter_query,
+                has_premium=True,  # doesn't matter here
+            )
+        if len(events) == 0:
+            return api_response(wrap_in_fail_result('No event found'), status_code=HTTPStatus.NOT_FOUND)  # noqa: E501
+
+        details = events[0].get_details()
+
```
yabirgb, line 4968: extra space

rotkehlchen/accounting/mixins/event.py
```diff
@@ -79,8 +79,11 @@ def process(
     def serialize(self) -> dict[str, Any]:
         ...
 
+    def _serialize_for_debugging(self) -> dict[str, Any]:
+        return self.serialize()
+
     def serialize_for_debug_import(self) -> dict[str, Any]:
-        data = self.serialize()
+        data = self._serialize_for_debugging()
```
yabirgb, line 86: I feel this is a bad name because it does reference to where it is used instead of what it does but @LefterisJP can also chime in here and give a third opinion

rotkehlchen/accounting/structures/base.py
```diff
@@ -277,6 +279,17 @@ def get_identifier(self) -> str:
...
+
+    def get_details(self) -> Optional[dict[str, Any]]:
+        details = {}
+
+        if self.extra_data is not None and 'sub-swaps' in self.extra_data:
+            details['sub-swaps'] = self.extra_data['sub-swaps']
+
+        return details if len(details) > 0 else None
```
yabirgb, line 291: this still has the problem we said about the extra info in the liquity staking events right?

rotkehlchen/accounting/mixins/event.py
nebolax: Yes, makes sense, I agree. Renamed it to `serialize_all_info`

rotkehlchen/db/upgrades/v35_v36.py
```diff
@@ -542,6 +544,31 @@ def _add_okx(write_cursor: 'DBCursor') -> None:
...
+                HistoryEventSubType.DEPOSIT_ASSET.serialize(),
+                HistoryEventSubType.REMOVE_ASSET.serialize(),
+                CPT_LIQUITY,
+            ),
+        )
+        for (identifier, serialized_extra_data) in read_cursor:
+            if serialized_extra_data is None:
+                continue
+            extra_data = json.loads(serialized_extra_data)
+            upgraded_extra_data = {'liquity-staking': extra_data}
```
yabirgb, line 563: if the `liquity-staking` and `sub-swaps` is something this deep it should be a constant somewhere

rotkehlchen/db/upgrades/v35_v36.py
```diff
@@ -542,6 +544,31 @@ def _add_okx(write_cursor: 'DBCursor') -> None:
     log.debug('Exit _add_okx')
 
 
+def _upgrade_liquity_staking_events(write_cursor: 'DBCursor', conn: 'DBConnection') -> None:
+    log.debug('Enter _upgrade_liquity_staking_events')
+    with conn.read_ctx() as read_cursor:
+        read_cursor.execute(
+            'SELECT identifier, extra_data FROM history_events WHERE type=? AND subtype IN (?, ?) AND counterparty=?',  # noqa: E501
+            (
```
yabirgb, line 552: why not filter where the extra data column is not `None` . The serialization method writes `NULL`  when the filter contains no information

rotkehlchen/accounting/structures/base.py
```diff
@@ -277,6 +279,24 @@ def get_identifier(self) -> str:
     def get_assets(self) -> list[Asset]:
         return [self.asset]
 
+    def has_details(self) -> bool:
+        return (
+            self.extra_data is not None and
+            ('sub-swaps' in self.extra_data or 'liquity-staking' in self.extra_data)
+        )
```
yabirgb, line 286: I don't see this scale easily. Why not return directly what is inside the extra_data field? This way you need to make sure always that whenever you add something that needs to be returned to also add it here but the only benefit that I see is having a common key but don't know that this can exists already by looking to the protocol in the frontend?

rotkehlchen/accounting/structures/base.py
```diff
@@ -277,6 +279,24 @@ def get_identifier(self) -> str:
...
+
+    def get_details(self) -> Optional[dict[str, Any]]:
+        if self.extra_data is None:
+            return None
+
+        details = {}
+        if 'sub-swaps' in self.extra_data:
+            details['sub-swaps'] = self.extra_data['sub-swaps']
+        if 'liquity-staking' in self.extra_data:
+            details['liquity-staking'] = self.extra_data['liquity-staking']
```
yabirgb, line 296: this can be an `elif` right? a single event belongs only to one protocol

yabirgb (commented):
comments and questions

rotkehlchen/accounting/structures/base.py
nebolax: Turned into a loop

rotkehlchen/db/upgrades/v35_v36.py
nebolax: Yes, makes sense, thanks!

rotkehlchen/db/upgrades/v35_v36.py
```diff
@@ -542,6 +545,29 @@ def _add_okx(write_cursor: 'DBCursor') -> None:
...
+                CPT_LIQUITY,
+            ),
+        )
+        for (identifier, serialized_extra_data) in read_cursor:
+            extra_data = json.loads(serialized_extra_data)
+            upgraded_extra_data = {LIQUITY_STAKING_DETAILS: extra_data}
+            write_cursor.execute(
+                'UPDATE history_events SET extra_data=? WHERE identifier=?',
+                (json.dumps(upgraded_extra_data), identifier),
+            )
```
yabirgb, line 566: This can be an execute many and do it in 1 write

yabirgb (commented):
Just one small comment

rotkehlchen/db/upgrades/v35_v36.py
```diff
@@ -542,6 +545,31 @@ def _add_okx(write_cursor: 'DBCursor') -> None:
     log.debug('Exit _add_okx')
 
 
+def _upgrade_liquity_staking_events(write_cursor: 'DBCursor', conn: 'DBConnection') -> None:
```
LefterisJP, line 548: docstring explaining why this change is happening and what you are doing

docs/api.rst
```diff
@@ -12122,3 +12119,50 @@ Custom Assets
...
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+          "result": {
+              "sub-swaps": [
+                  {"from_amount": "100.0", "to_amount": "0.084", "from_asset": A_SUSHI.identifier, "to_asset": A_ETH.identifier},  # noqa: E501
+                  {"from_amount": "0.084", "to_amount": "98.2", "from_asset": A_ETH.identifier, "to_asset": A_USDT.identifier},  # noqa: E501
```
LefterisJP, line 12154: This is invalid json copy pasta Use the correct values and make sure that there is no trailing commas. Run make docs.

```suggestion
                  {"from_amount": "100.0", "to_amount": "0.084", "from_asset": A_SUSHI.identifier, "to_asset": A_ETH.identifier},  # noqa: E501
                  {"from_amount": "0.084", "to_amount": "98.2", "from_asset": A_ETH.identifier, "to_asset": A_USDT.identifier},  # noqa: E501
```

docs/api.rst
```diff
@@ -12122,3 +12119,50 @@ Custom Assets
...
+              ],
+          },
+          "message": ""
+      }
+
+   :resjson object result: A dictionary with the details.
```
LefterisJP, line 12160:
```suggestion
   :resjson object result: A dictionary with the details. It may contain one of the following amount of details.
```

rotkehlchen/accounting/structures/base.py
```diff
@@ -277,6 +290,18 @@ def get_identifier(self) -> str:
...
+        if self.extra_data is None:
+            return False
+        return len(self.extra_data.keys() & ALL_DETAILS_KEYS) > 0
+
+    def get_details(self) -> Optional[dict[str, Any]]:
+        if self.extra_data is None:
+            return None
+        details = {k: v for k, v in self.extra_data.items() if k in ALL_DETAILS_KEYS}
+
+        return details if len(details) > 0 else None
```
LefterisJP, line 303: This is not what I asked in the previous comment. Read the diff more carefully. Wrong newline removed :)
```suggestion

        details = {k: v for k, v in self.extra_data.items() if k in ALL_DETAILS_KEYS}
        return details if len(details) > 0 else None
```

rotkehlchen/db/filtering.py
```diff
@@ -974,6 +975,8 @@ def make(
             )
         if exclude_ignored_assets is True:
             filters.append(DBIgnoredAssetsFilter(and_op=True, asset_key='asset'))
+        if identifier is not None:
```
LefterisJP, line 978: Do you really need to use the filter mechanism here? This is not filterable and it's just a `SELECT` that should match a single entry. No need to complicate things. Do a SELECT query directly and don't add more `ifs` and `args` in here.

The identifier has no place in the filtering mechanism as it's not something that can be combined with anything else.

rotkehlchen/tests/api/test_evm_transactions.py
```diff
@@ -94,3 +105,115 @@ def test_query_transactions(rotkehlchen_api_server: 'APIServer'):
...
+        contained_in_msg='No event found',
+        status_code=HTTPStatus.NOT_FOUND,
+    )
+
+    # Check that if an event is in the db, but has no details, an error is returned
+    response = requests.get(
+        api_url_for(
+            rotkehlchen_api_server,
+            'eventdetailsresource',
+        ), json={'identifier': decoded_events[0]['identifier']},
```
LefterisJP, line 203: This raises a `KeyError` in the test

LefterisJP (commented):
Left more comments. Also a test is failing. 
`rotkehlchen/tests/api/test_evm_transactions.py::test_event_with_details`

rotkehlchen/tests/api/test_evm_transactions.py
nebolax: Ouch, thanks

LefterisJP (commented):
some small comments

rotkehlchen/tests/api/test_evm_transactions.py
```diff
@@ -74,7 +85,7 @@ def test_query_transactions(rotkehlchen_api_server: 'APIServer'):
             'evmtransactionsresource',
         ), json={
             'async_query': False,
-            'limit': 20,
+            'limit': 30,
```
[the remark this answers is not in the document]
nebolax: Alright

rotkehlchen/tests/api/test_evm_transactions.py
```diff
@@ -94,3 +105,115 @@ def test_query_transactions(rotkehlchen_api_server: 'APIServer'):
 
     assert optimism_found is True
     assert ethereum_found is True
+
+
+def test_event_with_details(rotkehlchen_api_server: 'APIServer'):
```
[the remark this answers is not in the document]
nebolax: Indeed, thanks. Didn't notice that file

LefterisJP (approved):
LGTM! Merging

PR #5387, opened 2023-01-11
Enforce single DB transaction, remove implicit savepoints and keep write transactions short

LefterisJP:
Closes https://github.com/rotki/rotki/issues/5222


LefterisJP:
heh

LefterisJP (commented):
Forgive me frens, for I am about to sin. Will merge this as it fixes a lot of things and introduces the required changes for how the DB is accessed but most probably breaks develop.

Backend tests are green, but frontend not, which indicates that:
1. There is still something off
2. We are probably missing some form of test in the backend as this should also be caught there.

As discussed in discord merging now and will fix/figure out later

PR #5392, opened 2023-01-13
Add uniswap auto routers decoder

nebolax:
This PR adds decoding methods for uniswap auto routers and makes sure that if an auto router is used, events have correct order and if a swap happened in multiple steps, the steps are combined.

There are 4 different types of swaps that can happen via an auto router:
1. Swap from eth to token https://etherscan.io/tx/0xaf8755f0ab8a0cfa8901fe2a9250a8727cca54825210061aab90f34b7a3ed9ba
2. Swap from token to eth https://etherscan.io/tx/0x1b6c3fe84ed4f8f273a54c3e3f6ba80f843522c6a19220a05089104fc54b09ba
3. Swap from token to token
    a) Multiple spend events, but a single receipt https://etherscan.io/tx/0x3ae92fa63a9cf672906036beb18ece09592a8a471bd7f15e4385ca5011615e51
    b) Multiple spend events and multiple receipt events https://etherscan.io/tx/0xa4e0dbf77bf7a9721e1ba4ecf44ed6ea8dcb1c16e9e784b6fefa30749f64e7c0

There are two auto routers from uniswap. The v1 router uses only uni v3 pools and v2 router uses both uni v3 and v2 pools. This PR ensures the correct order of events for both of them.

For uni v2 swaps:
1. If a swap happened through a uni v2 router (not an auto router), decode using the old logic and keep everything as is.
2. If a swap happened through some other router / aggregator (a uniswap auto router or any other router), decode only basic info (i.e.. try to find spend and receive events, but don't change the order of the events)

For uni v3 swaps always decode only basic info.

If only basic info was decoded for swap, then we expect that some aggregator-specific functionality will later ensure the order of the events and will combine them if needed. For uniswap auto routers this logic is located in uniswap v3 decoder's post-decoding rules.

These rules check which type of swap it is and process the events accordingly. For each swap `from_asset`, `to_asset`, `from_amount` and `to_amount` are determined and then used to create proper swap events.

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/uniswap/utils.py
+++ b/rotkehlchen/chain/ethereum/modules/uniswap/utils.py
@@ -7,13 +7,19 @@ from typing import TYPE_CHECKING, Any, Callable, Optional
 import requests
 from web3.types import BlockIdentifier
 
+from rotkehlchen.accounting.structures.base import HistoryBaseEntry
+from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.assets.asset import EvmToken
 from rotkehlchen.chain.ethereum.defi.zerionsdk import ZERION_ADAPTER_ADDRESS
 from rotkehlchen.chain.ethereum.interfaces.ammswap.types import LiquidityPool
 from rotkehlchen.chain.ethereum.interfaces.ammswap.utils import _decode_result
+from rotkehlchen.chain.ethereum.modules.uniswap.constants import CPT_UNISWAP_V2, CPT_UNISWAP_V3
+from rotkehlchen.chain.ethereum.utils import asset_normalized_value
 from rotkehlchen.chain.evm.contracts import EvmContract
+from rotkehlchen.chain.evm.decoding.utils import maybe_reshuffle_events
 from rotkehlchen.chain.evm.types import WeightedNode
 from rotkehlchen.constants import ZERO
+from rotkehlchen.constants.assets import A_ETH
 from rotkehlchen.constants.misc import ONE
 from rotkehlchen.constants.resolver import ethaddress_to_identifier
 from rotkehlchen.constants.timing import DEFAULT_TIMEOUT_TUPLE
@@ -235,3 +241,85 @@ def find_uniswap_v2_lp_price(
     numerator = (token0_supply * token0_price + token1_supply * token1_price)
     share_value = numerator / total_supply
     return Price(share_value)
+
+
+def decode_basic_uniswap_info(
+        amount_sent: int,
+        amount_received: int,
+        decoded_events: list[HistoryBaseEntry],
+        counterparty: str,
+        notify_user: Callable[[HistoryBaseEntry, str], None],
+) -> None:
+    """
+    Check last three events and if they are related to the swap, label them as such.
+    We check three events because potential events are: spend, (optionally) approval, receive.
+    Earlier events are not related to the current swap.
+    """
+    spend_event, approval_event, receive_event = None, None, None
+    for event in reversed(decoded_events):
+        try:
+            crypto_asset = event.asset.resolve_to_crypto_asset()
+        except (UnknownAsset, WrongAssetType):
+            notify_user(event, counterparty)
+            return
+
+        if (
+            event.event_type == HistoryEventType.INFORMATIONAL and
+            event.event_subtype == HistoryEventSubType.APPROVE and
+            approval_event is None
+        ):
+            approval_event = event
+        elif (
+            event.balance.amount == asset_normalized_value(amount=amount_sent, asset=crypto_asset) and  # noqa: E501
+            event.event_type == HistoryEventType.SPEND and
+            # don't touch ETH since there may be multiple ETH transfers
+            # and they are better handled by the aggregator decoder.
+            event.asset != A_ETH and
+            spend_event is None
+        ):
+            event.event_type = HistoryEventType.TRADE
+            event.event_subtype = HistoryEventSubType.SPEND
+            event.counterparty = counterparty
+            event.notes = f'Swap {event.balance.amount} {crypto_asset.symbol} in {counterparty}'  # noqa: E501
+            spend_event = event
+        elif (
+            event.balance.amount == asset_normalized_value(amount=amount_received, asset=crypto_asset) and  # noqa: E501
+            event.event_type == HistoryEventType.RECEIVE and
+            event.asset != A_ETH and
+            receive_event is None
+        ):
+            event.event_type = HistoryEventType.TRADE
+            event.event_subtype = HistoryEventSubType.RECEIVE
+            event.counterparty = counterparty
+            event.notes = f'Receive {event.balance.amount} {crypto_asset.symbol} as a result of a {counterparty} swap'  # noqa: E501
+            receive_event = event
+        elif (
+            event.counterparty in {CPT_UNISWAP_V2, CPT_UNISWAP_V3} and
+            event.event_type == HistoryEventType.TRADE
+        ):
+            # The structure of swaps is the following:
+            # 1.1 Optional approval event
+            # 1.2 Optional spend event
+            # 1.3 Optional receive event
+            # 1.4 SWAP_SIGNATURE event
+            # 2.1 Optional approval event
+            # 2.2 Optional spend event
+            # 2.3 Optional receive event
+            # 2.4 SWAP_SIGNATURE event
+            # etc.
+            # So if we are at SWAP_SIGNATURE № 2 then all events that are before SWAP_SIGNATURE № 1
+            # should have already been decoded, have counterparty set and have type Trade.
+            break
+        else:
+            # If what described in the comment above is not met then it is an error.
+            log.error(
+                f'Found unexpected event {event.serialize()} during decoding a uniswap swap '
+                f'in transaction {event.event_identifier.hex()}',
+            )
+            break
+
+    # Make sure that the approval event is NOT between the spend and receive events.
+    if spend_event is not None and approval_event is not None:
+        maybe_reshuffle_events(out_event=approval_event, in_event=spend_event)
+    if spend_event is not None and receive_event is not None:
+        maybe_reshuffle_events(out_event=spend_event, in_event=receive_event)
--- a/rotkehlchen/chain/ethereum/modules/uniswap/v2/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/uniswap/v2/decoder.py
@@ -3,11 +3,17 @@ from typing import TYPE_CHECKING, Callable, Optional
 from rotkehlchen.accounting.structures.base import HistoryBaseEntry
 from rotkehlchen.assets.asset import EvmToken
 from rotkehlchen.chain.ethereum.modules.uniswap.constants import CPT_UNISWAP_V2
-from rotkehlchen.chain.ethereum.modules.uniswap.v2.common import decode_uniswap_v2_like_swap
+from rotkehlchen.chain.ethereum.modules.uniswap.utils import decode_basic_uniswap_info
+from rotkehlchen.chain.ethereum.modules.uniswap.v2.common import (
+    UNISWAP_V2_ROUTER,
+    decode_uniswap_v2_like_swap,
+)
 from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface
 from rotkehlchen.chain.evm.decoding.structures import ActionItem
 from rotkehlchen.chain.evm.structures import EvmTxReceiptLog
+from rotkehlchen.constants.misc import ZERO
 from rotkehlchen.types import EvmTransaction
+from rotkehlchen.utils.misc import hex_or_bytes_to_int
 
 if TYPE_CHECKING:
     from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
@@ -34,6 +40,34 @@ class Uniswapv2Decoder(DecoderInterface):
         self.database = ethereum_inquirer.database
         self.ethereum_inquirer = ethereum_inquirer
 
+    def _decode_basic_swap_info(
+            self,
+            tx_log: EvmTxReceiptLog,
+            decoded_events: list[HistoryBaseEntry],
+    ) -> None:
+        """
+        Decodes only basic swap info. Basic swap info includes trying to find approval, spend and
+        receive events for this particular swap but doesn't include ensuring order of events if the
+        swap was made by an aggregator.
+        """
+        # amount_in is the amount that enters the pool and amount_out the one
+        # that leaves the pool
+        amount_in_token_0, amount_in_token_1 = hex_or_bytes_to_int(tx_log.data[0:32]), hex_or_bytes_to_int(tx_log.data[32:64])  # noqa: E501
+        amount_out_token_0, amount_out_token_1 = hex_or_bytes_to_int(tx_log.data[64:96]), hex_or_bytes_to_int(tx_log.data[96:128])  # noqa: E501
+        amount_in, amount_out = amount_in_token_0, amount_out_token_0
+        if amount_in == ZERO:
+            amount_in = amount_in_token_1
+        if amount_out == ZERO:
+            amount_out = amount_out_token_1
+
+        decode_basic_uniswap_info(
+            amount_sent=amount_in,
+            amount_received=amount_out,
+            decoded_events=decoded_events,
+            counterparty=CPT_UNISWAP_V2,
+            notify_user=self.notify_user,
+        )
+
     def _maybe_decode_v2_swap(  # pylint: disable=no-self-use
             self,
             token: Optional[EvmToken],  # pylint: disable=unused-argument
@@ -44,15 +78,23 @@ class Uniswapv2Decoder(DecoderInterface):
             all_logs: list[EvmTxReceiptLog],  # pylint: disable=unused-argument
     ) -> None:
         if tx_log.topics[0] == SWAP_SIGNATURE:
-            decode_uniswap_v2_like_swap(
-                tx_log=tx_log,
-                decoded_events=decoded_events,
-                transaction=transaction,
-                counterparty=CPT_UNISWAP_V2,
-                database=self.database,
-                ethereum_inquirer=self.ethereum_inquirer,
-                notify_user=self.notify_user,
-            )
+            if transaction.to_address == UNISWAP_V2_ROUTER:
+                # If uniswap v2 router is used, then we can decode an entire swap.
+                # Uniswap v2 router is a simple router that always has a single spend and a single
+                # receive event.
+                decode_uniswap_v2_like_swap(
+                    tx_log=tx_log,
+                    decoded_events=decoded_events,
+                    transaction=transaction,
+                    counterparty=CPT_UNISWAP_V2,
+                    database=self.database,
+                    ethereum_inquirer=self.ethereum_inquirer,
+                    notify_user=self.notify_user,
+                )
+            else:
+                # Else some aggregator (potentially complex) is used, so we decode only basic info
+                # and other properties should be decoded by the aggregator decoding methods later.
+                self._decode_basic_swap_info(tx_log=tx_log, decoded_events=decoded_events)
 
     # -- DecoderInterface methods
 
--- a/rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
@@ -1,23 +1,24 @@
 import logging
-from typing import TYPE_CHECKING, Callable, Optional
+from typing import TYPE_CHECKING, Callable, NamedTuple, Optional
+from rotkehlchen.accounting.structures.balance import Balance
 
 from rotkehlchen.accounting.structures.base import HistoryBaseEntry
 from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
-from rotkehlchen.assets.asset import EvmToken
-from rotkehlchen.chain.ethereum.utils import asset_normalized_value
+from rotkehlchen.assets.asset import Asset, EvmToken
+from rotkehlchen.chain.ethereum.modules.uniswap.utils import decode_basic_uniswap_info
+from rotkehlchen.chain.evm.decoding.constants import CPT_GAS
 from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface
 from rotkehlchen.chain.evm.decoding.structures import ActionItem
-from rotkehlchen.chain.evm.decoding.utils import maybe_reshuffle_events
 from rotkehlchen.chain.evm.structures import EvmTxReceiptLog
 from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.constants.assets import A_ETH
 from rotkehlchen.constants.misc import ZERO
-from rotkehlchen.errors.asset import UnknownAsset, WrongAssetType
+from rotkehlchen.fval import FVal
 from rotkehlchen.logging import RotkehlchenLogsAdapter
-from rotkehlchen.types import EvmTransaction
-from rotkehlchen.utils.misc import hex_or_bytes_to_address, hex_or_bytes_to_int
+from rotkehlchen.types import EvmTransaction, Location
+from rotkehlchen.utils.misc import hex_or_bytes_to_int
 
-from ..constants import CPT_UNISWAP_V3
+from ..constants import CPT_UNISWAP_V2, CPT_UNISWAP_V3
 
 if TYPE_CHECKING:
     from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
@@ -31,7 +32,77 @@ log = RotkehlchenLogsAdapter(logger)
 # https://docs.uniswap.org/protocol/reference/core/interfaces/pool/IUniswapV3PoolEvents#swap
 SWAP_SIGNATURE = b'\xc4 y\xf9JcP\xd7\xe6#_)\x17I$\xf9(\xcc*\xc8\x18\xebd\xfe\xd8\x00N\x11_\xbc\xcag'  # noqa: E501
 
-UNISWAP_V3_ROUTER = string_to_evm_address('0xE592427A0AEce92De3Edee1F18E0157C05861564')
+UNISWAP_AUTO_ROUTER_V1 = string_to_evm_address('0xE592427A0AEce92De3Edee1F18E0157C05861564')
+UNISWAP_AUTO_ROUTER_V2 = string_to_evm_address('0x68b3465833fb72A70ecDF485E0e4C7bD8665Fc45')
+
+UNISWAP_ROUTERS = {UNISWAP_AUTO_ROUTER_V1, UNISWAP_AUTO_ROUTER_V2}
+
+
+class SwapData(NamedTuple):
+    from_asset: Optional[Asset]
+    from_amount: FVal
+    to_asset: Optional[Asset]
+    to_amount: FVal
+
+
+def _find_from_asset_and_amount(events: list[HistoryBaseEntry]) -> Optional[tuple[Asset, FVal]]:
+    """
+    Searches for uniswap v2/v3 swaps, detects `from_asset` and sums up `from_amount`.
+    Works only with `from_asset` being the same for all swaps.
+    """
+    from_asset: Optional[Asset] = None
+    from_amount = ZERO
+    for event in events:
+        if (
+            event.event_type == HistoryEventType.TRADE and
+            event.event_subtype == HistoryEventSubType.SPEND and
+            event.counterparty in {CPT_UNISWAP_V2, CPT_UNISWAP_V3}
+        ):
+            if from_asset is None:
+                from_asset = event.asset
+            elif from_asset != event.asset:  # We currently support only single `from_asset`.
+                return None  # unexpected event
+
+            from_amount += event.balance.amount
+
+    if from_asset is None:
+        return None
+    return from_asset, from_amount
+
+
+def _find_to_asset_and_amount(events: list[HistoryBaseEntry]) -> Optional[tuple[Asset, FVal]]:
+    """
+    Searches for uniswap v2/v3 swaps, detects `to_asset` and sums up `to_amount`.
+    Works only with `to_asset` being the same for all swaps.
+    Also works with a special case where there is only one receive event at the end.
+    """
+    to_asset: Optional[Asset] = None
+    to_amount = ZERO
+    for event in events:
+        if (
+            event.event_type == HistoryEventType.TRADE and
+            event.event_subtype == HistoryEventSubType.RECEIVE and
+            event.counterparty in {CPT_UNISWAP_V2, CPT_UNISWAP_V3}
+        ):
+            # Some swaps have many receive events. The structure is:
+            # spend1, receive1, spend2, receive2, ..., spendN, receiveN
+            # In this case they will be decoded as trades and we check them here.
+            if to_asset is None:
+                to_asset = event.asset
+            elif to_asset != event.asset:  # We currently support only single `to_asset`.
+                return None  # unexpected event
+            to_amount += event.balance.amount
+        elif event.event_type == HistoryEventType.RECEIVE and to_asset is None:
+            # Some other swaps have only a single receive event. The structure is:
+            # spend1, spend2, ..., spendN, receive
+            # In this case the receive event won't be decoded as a trade and we check it here.
+            # to_asset should be None here since it should be the only receive event.
+            to_asset = event.asset
+            to_amount = event.balance.amount
+
+    if to_asset is None:
+        return None
+    return to_asset, to_amount
 
 
 class Uniswapv3Decoder(DecoderInterface):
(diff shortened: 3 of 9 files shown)


rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +68,237 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+        )
+
+    def _decode_token_to_token_swap(self, decoded_events: list[HistoryBaseEntry]) -> SwapData:
+        from_asset: Optional[Asset] = None
+        from_amount = ZERO
+        to_asset: Optional[Asset] = None
+        to_amount = ZERO
+
+        for i in range(1, len(decoded_events)):
+            event = decoded_events[i]
```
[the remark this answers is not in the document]
yabirgb: After reading below I see that you limit the execution of this logic to trades were the user interacts with the uniswap routers so I believe is safe. But change the for loop here

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +68,237 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
-                    event.event_subtype = HistoryEventSubType.NONE
-                    event.counterparty = CPT_UNISWAP_V3
-                    event.notes = f'Refund of {event.balance.amount} {crypto_asset.symbol} in uniswap-v3 due to price change'  # noqa: E501
-
-        maybe_reshuffle_events(out_event=out_event, in_event=in_event)
+        if transaction.to_address not in UNISWAP_ROUTERS:
+            return decoded_events  # work only with the known routers for now
+
+        if (decoded_events[1].asset == A_ETH and decoded_events[1].counterparty in UNISWAP_ROUTERS):  # noqa: E501
+            if decoded_events[1].event_type == HistoryEventType.SPEND:
```
yabirgb, line 241: I would defend against index errors in case something goes wrong and the data you expect here doesn't exists

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +68,237 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+            location_label=transaction.from_address,
+            asset=to_crypto_asset,
+            balance=Balance(amount=swap_data.to_amount),
+            notes=f'Receive {swap_data.to_amount} {to_crypto_asset.symbol} as the result of a swap via uniswap auto router',  # noqa: E501
+            event_type=HistoryEventType.TRADE,
+            event_subtype=HistoryEventSubType.RECEIVE,
+            counterparty=CPT_UNISWAP_V3,
+        )
+
+        gas_event = decoded_events[0]
```
yabirgb, line 300: should you check that this is indeed the gas event?

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -271,14 +271,16 @@ def run_all_post_decoding_rules(
...
-    ) -> None:
+    ) -> list[HistoryBaseEntry]:
         """
         Runs all post-decoding rules from self.rules.post_decoding_rules.
         The post-decoding rules list consists of tuples (priority, rule) and must be sorted by
         priority in ascending order. The higher the priority number the later the rule is run.
         """
         for (_, rule) in self.rules.post_decoding_rules:
-            rule(transaction=transaction, decoded_events=decoded_events, all_logs=all_logs)  # noqa: E501
+            decoded_events = rule(transaction=transaction, decoded_events=decoded_events, all_logs=all_logs)  # noqa: E501
```
yabirgb, line 281: not related to this PR but the priority is not taken into account right now?

yabirgb (commented):
Some comments

rotkehlchen/chain/ethereum/modules/uniswap/v2/decoder.py
```diff
@@ -34,6 +42,32 @@ def __init__(
...
         self.ethereum_inquirer = ethereum_inquirer
 
+    def _decode_basic_swap_info(  # pylint: disable=no-self-use
+            self,
+            tx_log: EvmTxReceiptLog,
+            decoded_events: list[HistoryBaseEntry],
+    ) -> None:
+        # amount_in is the amount that enters the pool
+        amount_in_0, amount_in_1 = hex_or_bytes_to_int(tx_log.data[0:32]), hex_or_bytes_to_int(tx_log.data[32:64])  # noqa: E501
+        amount_in = amount_in_0
```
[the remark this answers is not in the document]
nebolax: Makes sense, thanks

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +68,237 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+                event.event_subtype = HistoryEventSubType.RECEIVE
+                event.counterparty = CPT_UNISWAP_V3
+                event.notes = f'Receive {decoded_events[-1].balance.amount} {crypto_asset.symbol} as a result of a uniswap-v3 swap'  # noqa: E501
+
+    # --- Routers methods ---
+
+    def _decode_eth_to_token_swap(self, decoded_events: list[HistoryBaseEntry]) -> SwapData:
+        send_eth_event = decoded_events[1]
+        refund_eth_event = decoded_events[2] if decoded_events[2].asset == A_ETH else None
+        receive_token_event = decoded_events[2] if refund_eth_event is None else decoded_events[3]
```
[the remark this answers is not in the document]
nebolax: Yes, will do, thanks

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +68,237 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+            to_asset=receive_token_event.asset,
+            to_amount=receive_token_event.balance.amount,
+        )
+
+    def _decode_token_to_eth_swap(self, decoded_events: list[HistoryBaseEntry]) -> SwapData:
+        from_asset: Optional[Asset] = None
+        from_amount = ZERO
+
+        for i in range(2, len(decoded_events)):
+            event = decoded_events[i]
```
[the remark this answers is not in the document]
nebolax: > this the same as

Yes, thanks. During my experiments I also needed the index, but it is not needed anymore.

> Why are you dismissing the event with index 0? and why not iterate all?

Will add a comment into the code. index `0` is the gas event, index `1` is the receipt of ETH

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
nebolax: Could make sense to make sure that the transaction was initiated by the user :+1:

rotkehlchen/chain/evm/decoding/decoder.py
nebolax: It is https://github.com/rotki/rotki/blob/develop/rotkehlchen/chain/evm/decoding/decoder.py#L153

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +68,237 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+                event.event_type == HistoryEventType.TRADE and
+                event.event_subtype == HistoryEventSubType.SPEND and
+                event.counterparty in {CPT_UNISWAP_V2, CPT_UNISWAP_V3}
+            ) is False:
+                break  # the swap is over
+
+            if from_asset is None:
+                from_asset = event.asset
+            elif from_asset != event.asset:
+                break  # the swap is over
```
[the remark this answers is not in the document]
nebolax: > If I understood correctly in each iteration it detects the asset being swap and in the next event that is not informational it breaks it

Yes

> But why doesn't it finish in the if statement before this one

Yes, it should. I added this `break` just in case something from what we expect. Perhaps I'll change it so that if something unexpected happens, the decoding fails and the user is notified

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,267 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+        delta_token_0 = hex_or_bytes_to_int(tx_log.data[0:32], signed=True)
+        delta_token_1 = hex_or_bytes_to_int(tx_log.data[32:64], signed=True)
+        if delta_token_0 > 0:
+            amount_sent, amount_received = delta_token_0, -delta_token_1
+        else:
+            amount_sent, amount_received = delta_token_1, -delta_token_0
+
+        # Check last two events and if they are related to the swap, label them as such
+        last_events = decoded_events[-2:]
+        for event in last_events:
```
yabirgb, line 93:
```suggestion
        for event in decoded_events[-2:]:
```

yabirgb (commented):
A few things to discuss. Also important, as I sent in discord for example this two transactions that use aggreagators

https://etherscan.io/tx/0x297dddacc9ea6f441d42887443199eed50406e13128e5123916f0061286f7875
https://etherscan.io/tx/0xcbe558177f62ccdb77f59b6be11e60b0a3fed1d224d5ce28d2bb6dff59447d3b

Get partially decoded and this can result in errors  during accounting. This needs to be addressed 

image

rotkehlchen/chain/ethereum/modules/uniswap/v2/decoder.py
```diff
@@ -44,15 +93,21 @@ def _maybe_decode_v2_swap(  # pylint: disable=no-self-use
             all_logs: list[EvmTxReceiptLog],  # pylint: disable=unused-argument
     ) -> None:
         if tx_log.topics[0] == SWAP_SIGNATURE:
-            decode_uniswap_v2_like_swap(
```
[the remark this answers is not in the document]
nebolax: Yep, added now :+1:

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,267 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+                swap_data = self._decode_token_to_eth_swap(decoded_events)
+
+        else:
+            # Then this should be a swap from token to token
+            swap_data = self._decode_token_to_token_swap(decoded_events)
+
+        if swap_data is None:
+            self.msg_aggregator.add_warning(
+                f'Failed to decode a uniswap swap for transaction {transaction.tx_hash.hex()}. '
+                f'Please open an issue on our github or contact us in our discord server',
```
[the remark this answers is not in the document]
nebolax: Mmm yes, in fact they can be combined, so I just put them under a single if :+1:

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,245 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+        from_asset: Optional[Asset] = None
+        from_amount = ZERO
+        to_asset: Optional[Asset] = None
+        to_amount = ZERO
 
-        The swap method has as data the delta on the pool for the assets swapped so we make sure
-        to use signed integers to detect the amounts.
+        # In this loop we sum up events that spend `from_asset` and sum up
+        # events that receive the `to_asset`. `0` is the gas event, so we start with `1`.
+        for event in decoded_events[1:]:
```
yabirgb, line 177: I wouldn't skip the first event. If we change something in the logic of how events get decoded we can end up with bugs here since you are making assumptions that are not really need. Why not just

```suggestion
        for event in decoded_events:
```

the gast event can be skipped in the first check you do.

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,245 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+            if (
+                event.event_type == HistoryEventType.INFORMATIONAL and
+                event.event_subtype == HistoryEventSubType.APPROVE
```
yabirgb, line 180: Shouldn't this skip any informational event?

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,245 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+            ):
+                continue  # skip approval events
+
+            if (
+                event.event_type == HistoryEventType.TRADE and
+                event.counterparty in {CPT_UNISWAP_V2, CPT_UNISWAP_V3}
+            ) is False:
```
yabirgb, line 187: why this is false instead of negating the condition? it reads bad

```suggestion
                event.event_type != HistoryEventType.TRADE or
                event.counterparty not in in {CPT_UNISWAP_V2, CPT_UNISWAP_V3}
            ):
```

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,245 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+                # after all SWAP events and thus is not labeled with the appropriate counterparty.
+                # In this case we check it here.
+                if to_asset is None and event.event_type == HistoryEventType.RECEIVE:
+                    to_asset = event.asset
+                    to_amount = event.balance.amount
+                else:
+                    return None  # unexpected event
+
+            elif event.event_subtype == HistoryEventSubType.SPEND:
+                if from_asset is None:
```
yabirgb, line 198: why don't you need to check the counterparties as in the previous check?

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -315,7 +317,8 @@ def decode_transaction(
             if event:
                 events.append(event)
 
-        self.run_all_post_decoding_rules(
+        events.sort(key=lambda x: x.sequence_index, reverse=False)
+        events = self.run_all_post_decoding_rules(
```
yabirgb, line 321: why sort before and not after? Is it guaranteed that `run_all_post_decoding_rules` won't change order of events? or is it documented somewhere that it shouldn't?

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,245 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+    # --- Routers methods ---
+
+    def _decode_eth_to_token_swap(
+            self,
+            decoded_events: list[HistoryBaseEntry],
+    ) -> Optional[SwapData]:
+        """
+        Decode a swap of ETH to a token. Such swap consists of 3 events:
+        1. Sending ETH to the router.
+        2. If there is a rfund, receiving ETH from the router.
```
yabirgb, line 111:
```suggestion
        2. If there is a refund, receiving ETH from the router.
```

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,245 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+    def _decode_token_to_eth_swap(
+            self,
+            decoded_events: list[HistoryBaseEntry],
+    ) -> Optional[SwapData]:
+        from_asset: Optional[Asset] = None
+        from_amount = ZERO
+
+        # In this loop we find all the events that spend the token and sum them up.
+        # `0` event is the gas event, `1` event is the ETH receipt event, so we start with `2`.
+        for event in decoded_events[2:]:
```
yabirgb, line 138: why not just iterate over all the events? If we change anything we might affect the logic without noticing

yabirgb (commented):
Thanks! A few more comments. I've checked the swaps and seems to be decoded properly now

rotkehlchen/chain/ethereum/modules/uniswap/utils.py
```diff
@@ -235,3 +239,54 @@ def find_uniswap_v2_lp_price(
...
+        amount_sent: int,
+        amount_received: int,
+        decoded_events: list[HistoryBaseEntry],
+        counterparty: str,
+        notify_user: Callable[[HistoryBaseEntry, str], None],
+) -> None:
+    """
+    Check last three events and if they are related to the swap, label them as such.
+    We check three events because potential events are: spent, (optionally) approval, receipt.
+    Earlier events are not related to the current swap.
```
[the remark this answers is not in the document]
nebolax: Yes

rotkehlchen/chain/ethereum/modules/uniswap/v2/decoder.py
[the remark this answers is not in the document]
nebolax: In uniswap v3 decoder, in method `_maybe_decode_v3_swap`.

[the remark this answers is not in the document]
nebolax: In uniswap v3 decoder, in method `_maybe_decode_v3_swap`. 
image

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
nebolax: By iterating only specific events that we need I tried to make the logic more readabale and understandable.

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
nebolax: For better redability I tried to keep code here as explicit as possible. I think that if at any point we change the place of the gas event, it will require modifying many things. Or is redability not really an argument here?

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
nebolax: Currently it works only with uniswap routers, which means that we know what types of events can appear here

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,245 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+        # In this loop we sum up events that spend `from_asset` and sum up
+        # events that receive the `to_asset`. `0` is the gas event, so we start with `1`.
+        for event in decoded_events[1:]:
+            if (
+                event.event_type == HistoryEventType.INFORMATIONAL and
+                event.event_subtype == HistoryEventSubType.APPROVE
+            ):
+                continue  # skip approval events
+
+            if (
```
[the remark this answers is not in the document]
nebolax: Because if the if above is True then `continue` is called the loop goes to the next iteration

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
nebolax: Because they are already checked by the previous check. If counterparty is not in uniswap-v2 and uniswap-v3, then the body of the `if` above will be called

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,245 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+            balance=Balance(amount=swap_data.from_amount),
+            notes=f'Swap {swap_data.from_amount} {from_crypto_asset.symbol} via uniswap auto router',  # noqa: E501
+            event_type=HistoryEventType.TRADE,
+            event_subtype=HistoryEventSubType.SPEND,
+            counterparty=CPT_UNISWAP_V3,
+        )
+
+        to_event = HistoryBaseEntry(
+            event_identifier=transaction.tx_hash,
+            sequence_index=decoded_events[-1].sequence_index,
```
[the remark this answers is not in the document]
nebolax: Yes. Refunds are handled in `_decode_eth_to_token_swap` method

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,245 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+            asset=to_crypto_asset,
+            balance=Balance(amount=swap_data.to_amount),
+            notes=f'Receive {swap_data.to_amount} {to_crypto_asset.symbol} as the result of a swap via uniswap auto router',  # noqa: E501
+            event_type=HistoryEventType.TRADE,
+            event_subtype=HistoryEventSubType.RECEIVE,
+            counterparty=CPT_UNISWAP_V3,
+        )
+
+        gas_event = decoded_events[0]
+        return [gas_event, from_event, to_event]
```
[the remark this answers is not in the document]
nebolax: Sorry, didn't get the question.

> only 3 events in the original list of decoded events

Why does the length of the original list matters here?

rotkehlchen/chain/evm/decoding/decoder.py
nebolax: The reason why I made this change is because during the main decoding loop `maybe_reshuffle_events` may replace sequence indexes of some of events. But it doesn't re-order the events in the `decoded_events` list.

This change is not mandatory for uniswap routers post-decoders, but I just thought that it will be good if post-decoders receive events in the right order.

Should the events be re-sorted after running post-decoding rules or do you maybe think that this change is not needed at all?

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,245 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+
+        to_event = HistoryBaseEntry(
+            event_identifier=transaction.tx_hash,
+            sequence_index=decoded_events[-1].sequence_index,
+            timestamp=decoded_events[-1].timestamp,
+            location=Location.BLOCKCHAIN,
+            location_label=transaction.from_address,
+            asset=to_crypto_asset,
+            balance=Balance(amount=swap_data.to_amount),
+            notes=f'Receive {swap_data.to_amount} {to_crypto_asset.symbol} as the result of a swap via uniswap auto router',  # noqa: E501
```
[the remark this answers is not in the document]
nebolax: On etherscan these routers are labeled as `Uniswap V3 Router` and `Uniswap V3 Router 2`, but officially uniswap team calls them just auto routers.

https://uniswap.org/blog/auto-router
https://uniswap.org/blog/auto-router-v2

So I thought that is better to stay in line with the official docs.

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
yabirgb: I don't see how it improves readability. Please change it

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
yabirgb: Leaving this comment open. For me is not more redeable and is not easier to understand since you need a comment to explain it

rotkehlchen/chain/evm/decoding/decoder.py
yabirgb: What you did for example can introduce an error by the developer on how the events get genearted and return them in the wrong order. I would sort at the end. About sorting before... maybe you can avoid it since is not required now?

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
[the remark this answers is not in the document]
yabirgb: Here since you check the transaction.to_address I believe is fine. In the case of a transaction that bundles multiple events you might have issues but is a case that we don't handle yet so is fine

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
yabirgb: same comment as before, for me it only gets confusing and in the other case you had to add a comment to explain why

yabirgb (commented):
Left a comment in the sorting before vs after. For the places where you skip 1 or 2 of the events it just adds complexity in my opinion for someone reading the code for little reward.

As for the `returning events in the post process rules vs modifying in place`  I prefer returning. My problem was that you were overwriting the events but later saw that this happens only if you call the router contract directly so there is no place to other events to appear so is fine.

rotkehlchen/chain/evm/decoding/decoder.py
nebolax: Perhaps yes

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
LefterisJP: > if counterparty is not in uniswap-v2 and uniswap-v3, then the body of the if above will be called

That is a very very very sneaky thing to reason about by just looking at the code. You either need to rewrite the logic to make this a lot more obvious or put small and strategic comments to explain what you are doing here.

I spent way too long a time looking at this trying to understand your reasoning.

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,240 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+
+    def _routers_post_decoding(
+            self,
+            transaction: EvmTransaction,
+            decoded_events: list[HistoryBaseEntry],
+            all_logs: list[EvmTxReceiptLog],  # pylint: disable=unused-argument
+    ) -> list[HistoryBaseEntry]:
+        """
+        Ensures that if an auto router (either v1 or v2) is used, events have correct order and
+        are properly combined (i.e. each swap consists only of one send and one receive event).
```
LefterisJP, line 225:
```suggestion
        are properly combined (i.e. each swap consists only of one spend and one receive event).
```

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,240 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
-                    event.event_subtype = HistoryEventSubType.NONE
-                    event.counterparty = CPT_UNISWAP_V3
-                    event.notes = f'Refund of {event.balance.amount} {crypto_asset.symbol} in uniswap-v3 due to price change'  # noqa: E501
-
-        maybe_reshuffle_events(out_event=out_event, in_event=in_event)
+        if transaction.to_address not in UNISWAP_ROUTERS:
+            return decoded_events  # work only with the known routers for now
+
+        if len(decoded_events) < 3 or decoded_events[0].counterparty != CPT_GAS:
+            # The minimum is 3 events: gas, spend, receive.
```
LefterisJP, line 241: Will they be at this order by this point? You should specify what is the order this function expects them to be

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,240 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+            return decoded_events
+
+        # Since we check that the to_address is a known router, we can be sure that the ETH
+        # transfer event (and other events) are parts of the swap.
+        if decoded_events[1].asset == A_ETH:
```
LefterisJP, line 246: Is decoded_events[1] guaranteed to be the spend event always at this point? If yes then specify it in the comment above or create a named variable. If used a lot a variable is better

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,240 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+            else:  # Receive
+                # This is a swap from token to eth
+                swap_data = self._decode_token_to_eth_swap(decoded_events)
+        else:
+            # Then this should be a swap from token to token
+            swap_data = self._decode_token_to_token_swap(decoded_events)
+
+        if swap_data is None or swap_data.from_asset is None or swap_data.to_asset is None:
+            self.msg_aggregator.add_warning(
+                f'Failed to decode a uniswap swap for transaction {transaction.tx_hash.hex()}. '
```
LefterisJP, line 259: Isn't this kind of message a bit verbose? Do we do this for transactions that we have errors with? Also what is the difference between this and the self.notify_user() method? Just that one calls add_warning and the other add_error?

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,240 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+
+        from_event = HistoryBaseEntry(
+            event_identifier=transaction.tx_hash,
+            sequence_index=decoded_events[1].sequence_index,
+            timestamp=decoded_events[1].timestamp,
+            location=Location.BLOCKCHAIN,
+            location_label=transaction.from_address,
+            asset=from_crypto_asset,
+            balance=Balance(amount=swap_data.from_amount),
+            notes=f'Swap {swap_data.from_amount} {from_crypto_asset.symbol} via uniswap auto router',  # noqa: E501
```
LefterisJP, line 284: Isn't this v3 specific?

```suggestion
            notes=f'Swap {swap_data.from_amount} {from_crypto_asset.symbol} via uniswap-v3 auto router',  # noqa: E501
```

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,240 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+
+        to_event = HistoryBaseEntry(
+            event_identifier=transaction.tx_hash,
+            sequence_index=decoded_events[-1].sequence_index,
+            timestamp=decoded_events[-1].timestamp,
+            location=Location.BLOCKCHAIN,
+            location_label=transaction.from_address,
+            asset=to_crypto_asset,
+            balance=Balance(amount=swap_data.to_amount),
+            notes=f'Receive {swap_data.to_amount} {to_crypto_asset.symbol} as the result of a swap via uniswap auto router',  # noqa: E501
```
LefterisJP, line 298: same versioning comment

rotkehlchen/tests/unit/decoders/test_uniswapv2.py
```diff
@@ -323,3 +324,67 @@ def test_uniswap_v2_swap_eth_returned(database, ethereum_inquirer, eth_transacti
         ),
     ]
     assert events == expected_events
+
+
+@pytest.mark.parametrize('ethereum_accounts', [['0xa931b486F661540c6D709aE6DfC8BcEF347ea437']])
```
LefterisJP, line 329: use vcr.py. That's the only reason I said you can not hand-mock the below

rotkehlchen/tests/unit/decoders/test_uniswapv3.py
```diff
@@ -491,3 +487,211 @@ def test_uniswap_v3_swap_by_aggregator(database, ethereum_inquirer, eth_transact
     ]
     assert len(events) == 3
     assert events == expected_events
+
+
+@pytest.mark.parametrize('ethereum_accounts', [['0xd6f2F8a2D6BD2f06234a95e61b55f41676CbE50d']])
```
LefterisJP, line 492: use vcr.py and remove TODO

rotkehlchen/tests/unit/decoders/test_uniswapv3.py
```diff
@@ -491,3 +487,211 @@ def test_uniswap_v3_swap_by_aggregator(database, ethereum_inquirer, eth_transact
...
+            balance=Balance(amount=FVal('841047.621362')),
+            location_label=user_address,
+            notes='Receive 841047.621362 USDC as the result of a swap via uniswap auto router',  # noqa: E501
+            counterparty=CPT_UNISWAP_V3,
+        ),
+    ]
+
+
+@pytest.mark.parametrize('ethereum_accounts', [['0xd6f2F8a2D6BD2f06234a95e61b55f41676CbE50d']])
+def test_swap_tokens_to_eth(database, ethereum_inquirer, ethereum_accounts):
```
LefterisJP, line 545: same here

rotkehlchen/tests/unit/decoders/test_uniswapv3.py
```diff
@@ -491,3 +487,211 @@ def test_uniswap_v3_swap_by_aggregator(database, ethereum_inquirer, eth_transact
...
+            asset=Asset('ETH'),
+            balance=Balance(amount=FVal('150.306972002256248665')),
+            location_label=user_address,
+            notes='Receive 150.306972002256248665 ETH as the result of a swap via uniswap auto router',  # noqa: E501
+            counterparty=CPT_UNISWAP_V3,
+        ),
+    ]
+
+
+@pytest.mark.parametrize('ethereum_accounts', [['0xCDeBA740656640fCA1A7b573e925f8C3b92f76b6']])
```
LefterisJP, line 596: same

rotkehlchen/tests/unit/decoders/test_uniswapv3.py
```diff
@@ -491,3 +487,211 @@ def test_uniswap_v3_swap_by_aggregator(database, ethereum_inquirer, eth_transact
...
+            balance=Balance(amount=FVal('2994089.109716')),
+            location_label=user_address,
+            notes='Receive 2994089.109716 USDC as the result of a swap via uniswap auto router',  # noqa: E501
+            counterparty=CPT_UNISWAP_V3,
+        ),
+    ]
+
+
+@pytest.mark.parametrize('ethereum_accounts', [['0x73264d8bE9EDDfCD25E4d54BF1b69828c9631A1C']])
+def test_swap_tokens_to_tokens_multiple_receipts(database, ethereum_inquirer, ethereum_accounts):
```
LefterisJP, line 649: same

LefterisJP (commented):
Lots of comments

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
nebolax: Uniswap official documentation names them just "auto routers", but yes, they work with uni-v3 pools, so perhaps including `v3` version in the text will make sense.

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
nebolax: The thing is that `notify_user` method is only for assets. The text that it sends is:
```
f'Could not identify asset {event.asset} decoding ethereum event in {counterparty}. '
f'Make sure that it has all the required properties (name, symbol and decimals) and '
f'try to decode the event again {event.event_identifier.hex()}.',
```

But here there is a different problem

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,240 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+            if (
+                event.event_type == HistoryEventType.TRADE and
+                event.event_subtype == HistoryEventSubType.SPEND and
+                event.counterparty in {CPT_UNISWAP_V2, CPT_UNISWAP_V3}
+            ) is False:
+                return None  # unexpected event
+
+            if from_asset is None:
+                from_asset = event.asset
+            elif from_asset != event.asset:
```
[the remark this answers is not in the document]
nebolax: These post-decoding rules are called only for uniswap auto routers. And currently they are made only for a single `from_asset` and a single `to_asset`.

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
nebolax: Mm yes, the gas event check is not needed anymore

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
nebolax: Discussed in the standup. Will change into a log.error

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +69,240 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+            self,
+            decoded_events: list[HistoryBaseEntry],
+    ) -> Optional[SwapData]:
+        """
+        Decode a swap of ETH to a token. Such swap consists of 3 events:
+        1. Sending ETH to the router.
+        2. If there is a refund, receiving ETH from the router.
+        3. Receiving tokens from the router.
+        """
+        send_eth_event = decoded_events[1]
```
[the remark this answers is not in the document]
nebolax: Changed to use explicit eth event and find other events by iterating over a loop

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
nebolax: Rewrote. Hopefully should be more understandable

rotkehlchen/chain/ethereum/modules/uniswap/utils.py
```diff
@@ -235,3 +240,58 @@ def find_uniswap_v2_lp_price(
...
+        ):
+            event.event_type = HistoryEventType.TRADE
+            event.event_subtype = HistoryEventSubType.SPEND
+            event.counterparty = counterparty
+            event.notes = f'Swap {event.balance.amount} {crypto_asset.symbol} in {counterparty}'  # noqa: E501
+            spent_event = event
+        elif (
+            event.balance.amount == asset_normalized_value(amount=amount_received, asset=crypto_asset) and  # noqa: E501
+            event.event_type == HistoryEventType.RECEIVE and
+            event.asset != A_ETH
```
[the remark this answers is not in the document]
nebolax: The thing is that we don't know tokens that were swapped without querying the blockchain. In order to reduce risks of accidentally finding an unrelated transfer the loop checks only the last 3 events `for event in decoded_events[-3:]:`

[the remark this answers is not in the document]
nebolax: Adjusted the code to make it more clear

rotkehlchen/chain/ethereum/modules/uniswap/utils.py
```diff
@@ -235,3 +241,84 @@ def find_uniswap_v2_lp_price(
...
+            event.counterparty = counterparty
+            event.notes = f'Receive {event.balance.amount} {crypto_asset.symbol} as a result of a {counterparty} swap'  # noqa: E501
+            receive_event = event
+        elif (
+            event.counterparty in {CPT_UNISWAP_V2, CPT_UNISWAP_V3} and
+            event.event_type == HistoryEventType.TRADE
+        ):
+            # The structure of swaps is the following:
+            # 1.1 Optional approval event
+            # 1.2 Optinal spend event
```
LefterisJP, line 302:
```suggestion
            # 1.2 Optional spend event
```

rotkehlchen/chain/ethereum/modules/uniswap/utils.py
```diff
@@ -235,3 +241,84 @@ def find_uniswap_v2_lp_price(
...
+            # 1.3 Optional receive event
+            # 1.4 SWAP_SIGNATURE event
+            # 2.1 Optional approval event
+            # 2.2 Optinal spend event
```
LefterisJP, line 306:
```suggestion
            # 2.2 Optional spend event
```

rotkehlchen/chain/ethereum/modules/uniswap/utils.py
```diff
@@ -235,3 +241,84 @@ def find_uniswap_v2_lp_price(
...
+            # 2.3 Optional receive event
+            # 2.4 SWAP_SIGNATURE event
```
LefterisJP, line 308: Impotant to specify there can be more

```suggestion
            # 2.4 SWAP_SIGNATURE event
            # etc.
```

rotkehlchen/chain/ethereum/modules/uniswap/utils.py
```diff
@@ -235,3 +241,84 @@ def find_uniswap_v2_lp_price(
...
+            # So if we are at SWAP_SIGNATURE № 2 then all events that are before SWAP_SIGNATURE № 1
+            # should have already been decoded, have counterparty set and have type Trade.
+            break
+        else:
+            # If what desribed in the comment above is not met then it is an error.
```
LefterisJP, line 313:
```suggestion
            # If what described in the comment above is not met then it is an error.
```

rotkehlchen/chain/ethereum/modules/uniswap/utils.py
```diff
@@ -235,3 +241,84 @@ def find_uniswap_v2_lp_price(
...
+            log.error(
+                f'Found unexpected event {event.serialize()} during decoding a uniswap swap '
+                f'in transaction {event.event_identifier.hex()}',
+            )
+            break
+
+    # Make sure that the approval event is NOT between the spent and receipt events.
```
LefterisJP, line 320:
```suggestion
    # Make sure that the approval event is NOT between the spend and receive events.
```

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -57,93 +120,192 @@ def _maybe_decode_v3_swap(  # pylint: disable=no-self-use
...
+
+        try:
+            to_crypto_asset = swap_data.to_asset.resolve_to_crypto_asset()
+        except (UnknownAsset, WrongAssetType):
+            self.notify_user(event=decoded_events[-1], counterparty=CPT_UNISWAP_V3)
+            return decoded_events
+
+        from_event = HistoryBaseEntry(
+            event_identifier=transaction.tx_hash,
+            sequence_index=decoded_events[1].sequence_index,
```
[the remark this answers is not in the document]
nebolax: Yes, forgot to change this, sorry

LefterisJP (approved):
LGTM. Merging. Failed test is a timeout

PR #5393, opened 2023-01-13
Replace requests with urllib for packaging.py

kelsos:
Closes #(issue_number)

## Checklist

### DIFF
--- a/package.py
+++ b/package.py
@@ -340,11 +340,13 @@ class WindowsPackaging:
 
         build_dir = self.__storage.build_directory
         os.chdir(build_dir)
-        url = f'https://github.com/mrx23dot/miniupnp/releases/download/miniupnpd_2_2_24/{miniupnc}'
-        zip_filename, _ = urllib.request.urlretrieve(url, Path.joinpath(build_dir, miniupnc))
-        zip_path = build_dir / zip_filename
+        zip_path = build_dir / miniupnc 
         extraction_dir = build_dir / 'miniupnpc'
         extraction_dir.mkdir(exist_ok=True)
+
+        url = f'https://github.com/mrx23dot/miniupnp/releases/download/miniupnpd_2_2_24/{miniupnc}'
+        urllib.request.urlretrieve(url, zip_path)
+
         with ZipFile(zip_path, 'r') as zip_ref:
             zip_ref.extractall(extraction_dir)
 
@@ -483,11 +485,11 @@ class MacPackaging:
         temporary_directory = self.__storage.temporary_directory
         temporary_directory.mkdir(exist_ok=True)
         wheel_file = temporary_directory / pip_wheel
-        filename, _ = urllib.request.urlretrieve(
-            f'https://github.com/rotki/rotki-build/raw/main/{pip_wheel}',
-            wheel_file,
+        urllib.request.urlretrieve(
+            url=f'https://github.com/rotki/rotki-build/raw/main/{pip_wheel}',
+            filename=wheel_file,
         )
-        return Path(filename)
+        return Path(wheel_file)
 
     def __get_versions(self, packages: list[str]) -> dict[str, str]:
         """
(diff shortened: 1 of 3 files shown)


package.py
```diff
@@ -493,17 +480,14 @@ def __download_patched_pip(self) -> Path:
...
-
-        logger.error(f'pip failed to download {pip_wheel}')
-        raise Exception(f'{pip_wheel} download failed')
+        temporary_directory = self.__storage.temporary_directory
+        temporary_directory.mkdir(exist_ok=True)
+        wheel_file = temporary_directory / pip_wheel
+        filename, _ = urllib.request.urlretrieve(
+            f'https://github.com/rotki/rotki-build/raw/main/{pip_wheel}',
+            wheel_file,
+        )
```
yabirgb, line 489: in the backend since it helps catching erros we try to be explicit with the arguments when we break the line.

```suggestion
        filename, _ = urllib.request.urlretrieve(
            url=f'https://github.com/rotki/rotki-build/raw/main/{pip_wheel}',
            filename=wheel_file,
        )
```

package.py
```diff
@@ -493,17 +480,14 @@ def __download_patched_pip(self) -> Path:
...
-            with open(wheel_file, 'wb') as file:
-                file.write(response.content)
-            return wheel_file
-
-        logger.error(f'pip failed to download {pip_wheel}')
-        raise Exception(f'{pip_wheel} download failed')
+        temporary_directory = self.__storage.temporary_directory
+        temporary_directory.mkdir(exist_ok=True)
+        wheel_file = temporary_directory / pip_wheel
+        filename, _ = urllib.request.urlretrieve(
```
yabirgb, line 486: the filename is the same as the second argument to the function call so you can omit it and use `wheel_file` instead of `filename`

```suggestion
        urllib.request.urlretrieve(
```

package.py
```diff
@@ -354,7 +341,7 @@ def setup_miniupnpc(self) -> None:
         build_dir = self.__storage.build_directory
         os.chdir(build_dir)
         url = f'https://github.com/mrx23dot/miniupnp/releases/download/miniupnpd_2_2_24/{miniupnc}'
-        zip_filename = Downloader.download(url)
+        zip_filename, _ = urllib.request.urlretrieve(url, Path.joinpath(build_dir, miniupnc))
         zip_path = build_dir / zip_filename
```
yabirgb, line 345: When setting the filename argument it is the path where the file will be saved and the returned value in this case is the same as the argument

```
>>> a, b = urllib.request.urlretrieve('https://google.com', '/home/yabirgb/google.html')
>>> a
'/home/yabirgb/google.html'
```

```suggestion
        zip_path = build_dir / miniupnc 
        urllib.request.urlretrieve(url, zip_path)
```

yabirgb (commented):
Just some minor comments

yabirgb (approved):
Code looks good and builds fine. LGTM

package.py
```diff
@@ -353,11 +340,13 @@ def setup_miniupnpc(self) -> None:
 
         build_dir = self.__storage.build_directory
         os.chdir(build_dir)
-        url = f'https://github.com/mrx23dot/miniupnp/releases/download/miniupnpd_2_2_24/{miniupnc}'
-        zip_filename = Downloader.download(url)
-        zip_path = build_dir / zip_filename
+        zip_path = build_dir / miniupnc
```
LefterisJP, line 343: how did this pass linting @kelsos @yabirgb? This fails linting in develop right now (I am fixing in my PR). It has a trailing whitespace.

yabirgb: I recall running the linter but probably is my fault chaning something. Will check this weekend how to configure the hooks. The CI doesn't trigger the linter for this file. Thanks

kelsos: Yes, make lint will fix errors, but this file is not actually watched by the detection change mechanism, so unless there are also backend changes, it can miss it. I should probably change it so that you can also specify single files to the job action

PR #5400, opened 2023-01-14
Detect which chains an evm address is active

LefterisJP:
- Add functionality to detect at which evm chains an address is active.
- At evm chain addition, only add it for chains that it is active
- For 1.27 add a data migration to duplicate ethereum addresses to optimism and avalanche only if they have had activity there

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -9172,7 +9172,7 @@ Adding EVM accounts to all EVM chains
 
 .. http:put:: /api/(version)/blockchains/evm/accounts
 
-   Doing a PUT on the EVM accounts endpoint functions just like the add blockchain accounts endpoint but adds the addresses for all evm chains.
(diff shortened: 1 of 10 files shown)


LefterisJP:
The failed tests were due to failing to contact cryptoscamdb during migration. They pass locally.

```
rotkehlchen.data_migrations.migrations.migration_3:migration_3.py:41 Main Greenlet: Failed to update the list of ignored assets during db migration. Failed to retrieve information from cryptoscamdb. Expecting value: line 1 column 1 (char 0)
```

So will merge. But ... then again this is yet another part that shows we should not be contacting cryptoscamdb at migrations! These stuff should be mocked, or data migrations should just not run everywhere.

LefterisJP:
> The failed tests were due to failing to contact cryptoscamdb during migration. They pass locally.
> 
> ```
> rotkehlchen.data_migrations.migrations.migration_3:migration_3.py:41 Main Greenlet: Failed to update the list of ignored assets during db migration. Failed to retrieve information from cryptoscamdb. Expecting value: line 1 column 1 (char 0)
> ```
> 
> So will merge. But ... then again this is yet another part that shows we should not be contacting cryptoscamdb at migrations! These stuff should be mocked, or data migrations should just not run everywhere.

Note to self: They don't run everywhere. They are specifically opened for that particular test. But still will try to mock cryptoscamdb at least.

PR #5408, opened 2023-01-15
Fixes on liquity balances

yabirgb:
Prevent 0 balances from being added when liquity module is active


LefterisJP (approved):
LGTM. Failed tests do not seem to be related.

PR #5409, opened 2023-01-15
Handle profit on genesis for eth validators

### LINKED ISSUE #5390: PnL Report is wrong for Genesis ETH2 Validator
## Problem Definition

For a validator that deposited prior to the Beacon Chain Genesis date
- the post genesis deposit event (from the 06/11/2020) is treated as if the 32 ETH were sold
- the genesis day `ETH2 staking daily PnL` value (from the 01/12/2020) in the Profit and Loss Report is too large by 32 ETH (32.016 ETH instead of 0.016 ETH) 

Bildschirmfoto 2023-01-13 um 00 04 16

One day later, the `ETH2 staking daily PnL` value is correctly calculated.
Bildschirmfoto 2023-01-13 um 00 06 01

## Logs

I can provide logs if needed via a DM.

I assume that you can reproduce for any pre-geneis deposited validator. Accordingly, I suggest that you pick a validator with a sufficiently low validator index, e.g. [Validator 999](https://beaconcha.in/validator/999) who [deposited on 05/12/2020](https://etherscan.io/tx/0x187bef85f7797f4f42534fcfa080ed28ab77491b79fe9e9be8039416eebab6bc)

### System Description

Operating system: MacOS Montery
Rotki version: 1.26.3

yabirgb:
The beaconchain stats return an amount >32 for the income tab when profit is made on the fist day after deposit. This results in errors during accounting.

Closes #5390

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -13,6 +13,7 @@ Changelog
 * :bug:`-` Fixed an edge case where users would be swamped with "This socket is already used by another greenlet" errors during websocket communication between backend and frontend.
 * :bug:`-` Transfers of ether between tracked accounts will now have a correct label in the UI.
 * :bug:`-` Trades involving delisted bitfinex pairs will now be properly read by rotki.
+* :bug:`5390` Profit for ethereum validators will now be handled correctly if the deposit was made on beacon chain genesis.
 * :feature:`784` Add support for OKX exchange
 
 * :release:`1.26.3 <2022-12-30>`
--- a/rotkehlchen/chain/ethereum/modules/eth2/utils.py
+++ b/rotkehlchen/chain/ethereum/modules/eth2/utils.py
@@ -7,7 +7,7 @@ import requests
 from bs4 import BeautifulSoup, SoupStrainer
 
 from rotkehlchen.constants.assets import A_ETH
-from rotkehlchen.constants.misc import ZERO
+from rotkehlchen.constants.misc import ONE, ZERO
 from rotkehlchen.constants.timing import DAY_IN_SECONDS, DEFAULT_TIMEOUT_TUPLE
 from rotkehlchen.errors.misc import RemoteError
 from rotkehlchen.fval import FVal
@@ -21,6 +21,9 @@ from .structures import ValidatorDailyStats
 
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
+
+DAY_AFTER_ETH2_GENESIS = Timestamp(1606780800)
+INITIAL_ETH_DEPOSIT = FVal(32)
 
 
 class ValidatorBalance(NamedTuple):
@@ -137,6 +140,13 @@ def scrape_validator_daily_stats(
                 column_pos += 1
             elif column_pos == 2:
                 pnl = _parse_fval(column.string, 'income')
+                # if the validator makes profit in the genesis day beaconchain returns a
+                # profit of deposit + validation reward. We need to subtract the deposit value
+                # to obtain the actual pnl.
+                # Example: https://beaconcha.in/validator/999/stats
+                if pnl > ONE and timestamp == DAY_AFTER_ETH2_GENESIS:
+                    pnl -= INITIAL_ETH_DEPOSIT
+
                 column_pos += 1
             elif column_pos == 3:
                 start_amount = _parse_fval(column.string, 'start amount')
--- a/rotkehlchen/tests/unit/test_eth2.py
+++ b/rotkehlchen/tests/unit/test_eth2.py
@@ -2,6 +2,7 @@ import datetime
 import re
 from contextlib import ExitStack
 from pathlib import Path
+from typing import TYPE_CHECKING
 from unittest.mock import patch
 
 import pytest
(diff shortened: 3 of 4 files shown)


rotkehlchen/chain/ethereum/modules/eth2/utils.py
```diff
@@ -137,6 +137,16 @@ def scrape_validator_daily_stats(
                 column_pos += 1
             elif column_pos == 2:
                 pnl = _parse_fval(column.string, 'income')
+                # if the validator makes profit in the first day beaconchain returns a
+                # profit of deposit + validation reward. We need to substract the deposit value
+                # to obtain the actual pnl.
+                # Example: https://beaconcha.in/validator/999/stats
+                if pnl > ONE:
```
nebolax, line 144: Kinda nitpicky since the current implementation will probably always work, but I think it will be cleaner to check if the value is greater than 32 and then subtract 32 from the pnl

nebolax (commented):
Just a couple of minor comments

rotkehlchen/chain/ethereum/modules/eth2/utils.py
yabirgb: This was my first approach but then I tough that you might miss to commit a block and then you will get penalized, in that case you balance on first day would be < 32 ETH and I guess that the API will have the same issue

nebolax: Alright then :+1:

LefterisJP: you don't commit a block. Back then it was only attestations. And maybe some empty blocks. But yes can be zero or negative if you got slashed or if your validator just does not come offline.

But this in general looks like a flawed approach.

Why not deal with day of activation? So just find the first date and do this for the first date in the export?

rotkehlchen/chain/ethereum/modules/eth2/utils.py
```diff
@@ -137,6 +137,16 @@ def scrape_validator_daily_stats(
                 column_pos += 1
             elif column_pos == 2:
                 pnl = _parse_fval(column.string, 'income')
+                # if the validator makes profit in the first day beaconchain returns a
+                # profit of deposit + validation reward. We need to subtract the deposit value
+                # to obtain the actual pnl.
+                # Example: https://beaconcha.in/validator/999/stats
+                if pnl > ONE:
+                    # Since the profit in one day is < 1 ETH get the integer part
```
LefterisJP, line 145: Does this happen only for stakers that are active at genesis? If a validator starts later will the 32 ETH deposit also show like this? If yes then I can see edge cases where this breaks

docs/changelog.rst
```diff
@@ -13,6 +13,7 @@ Changelog
 * :bug:`-` Fixed an edge case where users would be swamped with "This socket is already used by another greenlet" errors during websocket communication between backend and frontend.
 * :bug:`-` Transfers of ether between tracked accounts will now have a correct label in the UI.
 * :bug:`-` Trades involving delisted bitfinex pairs will now be properly read by rotki.
+* :bug:`5390` Profit for ethereum validators will now be handled correctly when it was made in the first day after deposit.
```
LefterisJP, line 16: Related to my other comment. This seems to only be for the genesis day. Not deposit.

```suggestion
* :bug:`5390` Profit for ethereum validators will now be handled correctly if the deposit was made on beacon chain genesis.
```

rotkehlchen/tests/unit/test_eth2.py
```diff
@@ -882,3 +887,47 @@ def test_validator_daily_stats_with_db_interaction(  # pylint: disable=unused-ar
...
+
+
+@pytest.mark.parametrize('default_mock_price_value', [FVal(1.55)])
+def test_validator_daily_stats_with_event_in_first_day(
+        network_mocking: bool,
+        price_historian: 'PriceHistorian',  # pylint: disable=unused-argument
+        function_scope_messages_aggregator: 'MessagesAggregator',
+) -> None:
+    """
+    Test that if profit returned by beaconchain on first day is > 32 ETH the app
```
LefterisJP, line 899: edit comment: I think we are talking about genesis day. Not just validator's first deposit day

rotkehlchen/chain/ethereum/modules/eth2/utils.py
LefterisJP: Ping on this one and the comments around this as pinged you in discord.

rotkehlchen/chain/ethereum/modules/eth2/utils.py
```diff
@@ -137,6 +139,16 @@ def scrape_validator_daily_stats(
...
                 pnl = _parse_fval(column.string, 'income')
+                # if the validator makes profit in the first day beaconchain returns a
+                # profit of deposit + validation reward. We need to subtract the deposit value
+                # to obtain the actual pnl.
+                # Example: https://beaconcha.in/validator/999/stats
+                if pnl > ONE and timestamp == DAY_AFTER_ETH2_GENESIS:
+                    # Since the profit in one day is < 1 ETH get the integer part
+                    # of the profit and subtract it. This covers the case of profit
+                    # and the case of missing attestations on the first day.
+                    pnl -= pnl.to_int(exact=False)
```
LefterisJP, line 150: wait what? Why not just deduct 32?What are you doing here?

rotkehlchen/tests/unit/test_eth2.py
```diff
@@ -888,3 +893,47 @@ def test_validator_daily_stats_with_db_interaction(  # pylint: disable=unused-ar
             )
         last_stat = stats[:len(expected_stats)][-1]
         assert last_stat.pnl_balance.amount == expected_stats[-1].pnl_balance.amount * FVal(0.45)  # noqa: E501
+
+
+@pytest.mark.parametrize('default_mock_price_value', [FVal(1.55)])
+def test_validator_daily_stats_with_event_in_first_day(
```
LefterisJP, line 899:
```suggestion
def test_validator_daily_stats_with_genesis_event(
```

LefterisJP (commented):
Please read my reviews

rotkehlchen/chain/ethereum/modules/eth2/utils.py
yabirgb: I had it at -32 ETH and changed it because I short circuited last night with the case of a missing attestation. - 32 is the right thing to do. CHanged it

LefterisJP (approved):
LGTM. Merging

PR #5410, opened 2023-01-16
Small fix in names search

nebolax:
Adjust 2 things in names search:

1. Blockchain labels will be searched only if blockchain is specified
2. ENS names will be searched only for Ethereum

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -70,8 +70,8 @@ from rotkehlchen.chain.bitcoin.xpub import XpubManager
 from rotkehlchen.chain.ethereum.airdrops import check_airdrops
 from rotkehlchen.chain.ethereum.modules.eth2.constants import FREE_VALIDATORS_LIMIT
 from rotkehlchen.chain.ethereum.modules.nft.structures import NftLpHandling
-from rotkehlchen.chain.ethereum.names import find_ens_mappings, search_for_addresses_names
 from rotkehlchen.chain.evm.manager import EvmManager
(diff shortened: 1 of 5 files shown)


rotkehlchen/chain/ethereum/names.py
```diff
@@ -148,6 +149,10 @@ def _blockchain_address_to_name(
     """Returns the label of an ethereum blockchain account with the given address or
     None if there is no such account or the account has no label set.
     """
+    if chain_address.blockchain is None:
```
LefterisJP, line 152: Update the docstring of the function. It should evm account and you should also explain the behaviour when not specifying chain.

rotkehlchen/chain/ethereum/names.py
```diff
@@ -205,6 +210,12 @@ def _ens_address_to_name(
         chain_address: OptionalChainAddress,
 ) -> Optional[str]:
     """Returns the ens name for an address or None if the address doesn't have one"""
+    if (
+        chain_address.blockchain is None or
+        chain_address.blockchain != SupportedBlockchain.ETHEREUM
+    ):
+        return None  # ENS names are only for ethereum
```
LefterisJP, line 217: Why though? So long as the address is the same and ENS registrar exists only in ethereum shouldn't this work the same?

I am not sure about this. Need to think a bit.

LefterisJP (commented):
Small comments. And one question. I saw you renamed a test. Shouldn't you also rename and move rotkehlchen/chain/ethereum/names.py to go under evm as it's generic and not ethereum specific?

LefterisJP (approved):
LGTM

PR #5416, opened 2023-01-17
Make extra table check log.info and remove extra_tokens at upgrade

LefterisJP:


yabirgb (approved):
Backend tests look okey. LGTM

PR #5420, opened 2023-01-17
Use specific types for querying amms

nebolax:
Don't iterate over sushiswap / uniswap events types when it is not needed

### DIFF
--- a/rotkehlchen/chain/ethereum/interfaces/ammswap/types.py
+++ b/rotkehlchen/chain/ethereum/interfaces/ammswap/types.py
@@ -124,6 +124,10 @@ class EventType(Enum):
         raise RuntimeError(f'Corrupt value {self} for EventType -- Should never happen')
 
 
+UNISWAP_EVENTS_TYPES = (EventType.MINT_UNISWAP, EventType.BURN_UNISWAP)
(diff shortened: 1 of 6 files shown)


rotkehlchen/chain/ethereum/interfaces/ammswap/types.py
```diff
@@ -124,6 +124,10 @@ def __str__(self) -> str:
         raise RuntimeError(f'Corrupt value {self} for EventType -- Should never happen')
 
 
+UNISWAP_EVENTS_TYPES = [EventType.MINT_UNISWAP, EventType.BURN_UNISWAP]
+SUSHISWAP_EVENTS_TYPES = [EventType.MINT_SUSHISWAP, EventType.BURN_SUSHISWAP]
```
yabirgb, line 128: Make those a tuple since they are faster and use less space.

```suggestion
UNISWAP_EVENTS_TYPES = (EventType.MINT_UNISWAP, EventType.BURN_UNISWAP)
SUSHISWAP_EVENTS_TYPES = (EventType.MINT_SUSHISWAP, EventType.BURN_SUSHISWAP)
```

yabirgb (commented):
There are places in the code where we do 

```
[EventType.MINT_SUSHISWAP, EventType.BURN_SUSHISWAP]
```

and similar for uniswap. Can you find them and use this constants instead?

LefterisJP (approved):
LGTM. Thanks

PR #5429, opened 2023-01-18
Assets for 1.27

yabirgb:
- Replace freezgun utility for pytest. The one I'm using as replacement is in the pytest-dev organization and this is
why I trust it. The reason was that the old was not being maintained and had deprecation warnings.
- Add new assets and mappings
- Fix issue updating assets when asset didn't exist yet locally


rotkehlchen/assets/converters.py
```diff
@@ -763,6 +772,13 @@
     'SPEL',  # Spell moon (SPEL). No information about this token
 )
 
+UNSUPPORTED_OKEX_ASSETS = {
```
LefterisJP, line 775: nitpick

```suggestion
UNSUPPORTED_OKX_ASSETS = {
```

rotkehlchen/assets/exchanges_mappings/binance.py
```diff
@@ -134,11 +134,13 @@
     evm_address_to_identifier('0x445f51299Ef3307dBD75036dd896565F5B4BF7A5', ChainID.ETHEREUM, EvmTokenKind.ERC20): 'VIDT',  # noqa: E501
     evm_address_to_identifier('0xb6EE9668771a79be7967ee29a63D4184F8097143', ChainID.ETHEREUM, EvmTokenKind.ERC20): 'CXO',  # noqa: E501
     evm_address_to_identifier('0x0202Be363B8a4820f3F4DE7FaF5224fF05943AB1', ChainID.ETHEREUM, EvmTokenKind.ERC20): 'UFT',  # noqa: E501
-    evm_address_to_identifier('0xdd974D5C2e2928deA5F71b9825b8b646686BD200', ChainID.ETHEREUM, EvmTokenKind.ERC20): 'KNC',  # noqa: E501
```
LefterisJP, line 137: where did kyber go?

LefterisJP (approved):
Nice you upgrading freezegun. I wanted to do it too at some point. Does it not require changing its usage around the tests too?

For the rest it looks okay, but I have naturally not tested every asset.

rotkehlchen/assets/exchanges_mappings/binance.py
yabirgb: We have this `COMMON_ASSETS_MAPPINGS` dictionary that is combined with the rest of mappings. I just moved it there because other exchanges needed it too

rotkehlchen/globaldb/upgrades/v3_v4.py
```diff
@@ -360,6 +361,11 @@ def _upgrade_address_book_table(cursor: 'DBCursor') -> None:
     cursor.execute('ALTER TABLE address_book_new RENAME TO address_book;')
 
 
+def _update_yearn_v1_protocol(cursor: 'DBCursor') -> None:
```
LefterisJP, line 364: this should be tested in the global DB upgrade

rotkehlchen/tests/unit/test_globaldb.py
```diff
@@ -1063,10 +1063,10 @@ def test_edit_token_with_missing_information(database):
     Test that editing a token that already exists with missing information doesn't
     raise any error and the information is updated
     """
-    token_address = string_to_evm_address('0xf53AD2c6851052A81B42133467480961B2321C09')
+    token_address = string_to_evm_address('0xc37b40ABdB939635068d3c5f13E7faF686F03B65')
     peth = get_or_create_evm_token(
         userdb=database,
-        symbol='PETH',
+        symbol='CPETH',
```
LefterisJP, line 1069: Why CPETH? Is it just any token that is supposed to be missing? Thenplease use a more exotic/random name.

LefterisJP (approved):
Perfect. All comments addressed and all tests are green. Merging.

PR #5430, opened 2023-01-18
Removes dots from docker dev build version

kelsos:
Closes #(issue_number)

## Checklist

## Notes
At least from the logs, it seemed that it didn't like what we put after -dev because it didn't seem to conform with PEP 440 or something. 

At least from a quick reading, it seems that using numbers is fine but `.` not.

### DIFF
--- a/.github/workflows/rotki_dev_builds.yml
+++ b/.github/workflows/rotki_dev_builds.yml
@@ -288,7 +288,7 @@ jobs:
         run: |
           ROTKI_VERSION=$(cat .bumpversion.cfg | grep 'current_version = ' | sed -n -e 's/current_version = //p')
           POSTFIX=$(if git describe --tags --exact-match "$REVISION" &>/dev/null; then echo ''; else echo '-dev'; fi)
-          ROTKI_VERSION=${ROTKI_VERSION}${POSTFIX}-$(date +'%Y.%m.%d')
+          ROTKI_VERSION=${ROTKI_VERSION}${POSTFIX}$(date +'%Y%m%d')
           echo "version=${ROTKI_VERSION}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/rotki_dev_builds.yml
```diff
@@ -288,7 +288,7 @@ jobs:
         run: |
           ROTKI_VERSION=$(cat .bumpversion.cfg | grep 'current_version = ' | sed -n -e 's/current_version = //p')
           POSTFIX=$(if git describe --tags --exact-match "$REVISION" &>/dev/null; then echo ''; else echo '-dev'; fi)
-          ROTKI_VERSION=${ROTKI_VERSION}${POSTFIX}-$(date +'%Y.%m.%d')
+          ROTKI_VERSION=${ROTKI_VERSION}${POSTFIX}-$(date +'%Y%m%d')
```
yabirgb, line 291:
```suggestion
          ROTKI_VERSION=${ROTKI_VERSION}${POSTFIX}$(date +'%Y%m%d')
```

yabirgb (commented):
This won't be enough. An example of a valid one would be

```
>>> is_canonical('1.26.3.dev20230114')
True
```
to get that I believe the change is

```
POSTFIX=$(if git describe --tags --exact-match "$REVISION" &>/dev/null; then echo ''; else echo '.dev'; fi)
```

and the suggestion. I used https://pypi.org/project/pep440/ to verify that is valid

.github/workflows/rotki_dev_builds.yml
kelsos: Good I will update

kelsos:
Done

yabirgb (approved):
Thanks

PR #5433, opened 2023-01-19
Adjust web3 related fixtures for all backend tests

LefterisJP:
Generalize the way web3 mocks are initialized across all types of tests and make sure that mocking happens depending on the same type of arguments across multiple tests.

This is part 1. More work coming in more commits touching more tests.

I actually started by trying to mock some tests that are not mocked, and have not done that yet :sweat_smile: 

Adjusting the fixtures to make them more reusable and make sure that they always mock some stuff like connecting to web3 nodes takes time.

More will come as either new PRs or commits to this one


LefterisJP:
Okay backend is fine so merging. Will add new stuff in new PRs.

PR #5434, opened 2023-01-19
Fix savepoint not being released on Exception

yabirgb:
The rollback of sqlite doesn't release the savepoint and we need to manually release it always. Before on exceptions we were letting the savepoint open and this had performance issues and raised errors


LefterisJP (approved):
Okay LGTM. Left 2 minor nitpicks to make clarifications but they can be added by you later.

PR #5440, opened 2023-01-19
Don't duplicate assets icons for assets in the same collection

yabirgb:
Instead of making two queries to coingecko and duplicating space in disk just create a symlink. In windows python3 works with symlinks so it should work there too

After discussing it with @LefterisJP and @nebolax I changed the approach to use a common identifier always for assets in collections


nebolax (commented):
Question. Instead of creating symlinks and iterating over related assets when reading an icon, maybe we can just keep the icons only for primary assets in collections and for other assets just use icons of the primary assets?

yabirgb:
So the problem with checking the collection is that at the moment this is fast because it will go to disk and check if we have the asset. Otherwise you would need to two extra step:

1. Query the database to get the collections
2. Check if for the asset that you want there is an icon. This is easier if you set a rule for what is the primary asset somehow but with the new collections schema this property doesn't exists

Also when store icons is a bit more complicated because you have again to check the collection and decide what is the proper name to use
My problem with changing this approach is that introduces complexity in different points, at least as how I imagine this implemented]
if you see an easier way to implement this I'm happy to hear it
on the other hand I don't see what is the problem with the symbolic links

rotkehlchen/icons.py
```diff
@@ -127,26 +130,31 @@ def get_icon(
...
-        except (UnknownAsset, WrongAssetType):
-            return None
+        # check if the asset is in a collection
+        related_assets_identifiers = set(GlobalDBHandler().get_related_assets(asset.identifier))
+
+        asset_to_query_icon = asset
+        if len(related_assets_identifiers) != 0:
+            related_assets_identifiers.add(asset.identifier)
+            # get the asset with the lowest lex order
+            collection_main_asset_id = min(related_assets_identifiers)
```
nebolax, line 140: Why no to do these checks at the DB query time?

Something like method `get_identifier_with_icon(identifier: str)` that would accept an identifier and return the asset identifier to check / query the icon for

rotkehlchen/icons.py
```diff
@@ -44,8 +44,11 @@ def __init__(self, data_dir: Path, coingecko: Coingecko) -> None:
         self.custom_icons_dir.mkdir(parents=True, exist_ok=True)
         self.failed_asset_ids: set[str] = set()
 
+    def _default_name(self, identifier: str) -> str:
```
nebolax, line 47: I am not sure that I understand. Why to create this method?

rotkehlchen/icons.py
yabirgb: I made the method general enough to be useful in other places but I can return only the min I guess

rotkehlchen/icons.py
yabirgb: Created it because I needed the name in other places too but as how I rewrote it is not needed anymore. Will undo this change

rotkehlchen/tests/data/mocks/test_coingecko/coins/dai.json
```diff
@@ -0,0 +1 @@
...ken/0xe3520349f477a5f6eb06107066048508498a291b/token-transfers","https://moonriver.moonscan.io/token/0x80a16016cc4a2e6a2caca8a4a498b1699ff0f844"],"official_forum_url":["","",""],"chat_url":["","",""],"announcement_url":["",""],"twitter_screen_name":"MakerDAO","facebook_username":"","bitcointalk_thread_identifier":null,"telegram_channel_identifier":"makerdaoOfficial","subreddit_url":"https://www.reddit.com/r/MakerDAO","repos_url":{"github":[],"bitbucket":[]}},"image":{"thumb":"https://assets.coingecko.com/coins/images/9956/thumb/4943.png?1636636734","small":"https://assets.coingecko.com/coins/images/9956/small/4943.png?1636636734","large":"https://assets.coingecko.com/coins/images/9956/large/4943.png?1636636734"},"country_origin":"","genesis_date":null,"contract_address":"0x6b175474e89094c44da98b954eedeac495271d0f","sentiment_votes_up_percentage":66.67,"sentiment_votes_down_percentage":33.33,"market_cap_rank":18,"coingecko_rank":165,"coingecko_score":40.022,"developer_score":0.0,"community_score":41.386,"liquidity_score":69.787,"public_interest_score":0.018,"public_interest_stats":{"alexa_rank":30241,"bing_matches":null},"status_updates":[],"last_updated":"2023-01-20T11:35:21.891Z"}
```
LefterisJP, line 1: As I told you in other files, we use only an absolute minimum from this data. Do not blindly copy all of it. Just copy what we need for the test mock. This way we save less data in the repo's tests

rotkehlchen/tests/data/mocks/test_coingecko/coins/ethereum.json
```diff
@@ -0,0 +1 @@
...30","sentiment_votes_up_percentage":75.62,"sentiment_votes_down_percentage":24.38,"ico_data":{"ico_start_date":"2014-07-20T00:00:00.000Z","ico_end_date":"2014-09-01T00:00:00.000Z","short_desc":"A decentralized platform for applications","description":null,"links":{},"softcap_currency":"","hardcap_currency":"","total_raised_currency":"","softcap_amount":null,"hardcap_amount":null,"total_raised":null,"quote_pre_sale_currency":"","base_pre_sale_amount":null,"quote_pre_sale_amount":null,"quote_public_sale_currency":"BTC","base_public_sale_amount":1.0,"quote_public_sale_amount":0.00074794,"accepting_currencies":"","country_origin":"","pre_sale_start_date":null,"pre_sale_end_date":null,"whitelist_url":"","whitelist_start_date":null,"whitelist_end_date":null,"bounty_detail_url":"","amount_for_sale":null,"kyc_required":true,"whitelist_available":null,"pre_sale_available":null,"pre_sale_ended":false},"market_cap_rank":2,"coingecko_rank":2,"coingecko_score":78.783,"developer_score":97.494,"community_score":72.746,"liquidity_score":95.183,"public_interest_score":0.236,"public_interest_stats":{"alexa_rank":8793,"bing_matches":null},"status_updates":[],"last_updated":"2023-01-20T12:51:55.702Z"}
```
LefterisJP, line 1: same here

rotkehlchen/tests/external_apis/test_coingecko.py
```diff
@@ -63,3 +68,64 @@ def test_assets_with_icons(icon_manager):
...
+def test_asset_icons_for_collections(icon_manager: IconManager) -> None:
+    """
+    Test that for assets in the same collection only one file is saved and
+    is later used by all the assets in the same collection
+    """
+    dai_op = Asset('eip155:10/erc20:0xDA10009cBd5D07dd0CeCc66161FC93D7c9000da1')
+    times_api_was_queried = 0
+
+    # remove icon file in case it was added by another test
+    icon_path = icon_manager.iconfile_path(A_DAI)
```
LefterisJP, line 82: will this affect production data? No right? This should go under "test_data" directory? Perhaps make sure by having the data fixture directory point to a random directory?

rotkehlchen/tests/external_apis/test_coingecko.py
```diff
@@ -63,3 +68,64 @@ def test_assets_with_icons(icon_manager):
...
+            if '279' in url:
+                icon_name = '279.png'
+            with open(test_data_folder / 'icons' / icon_name, 'rb') as f:
+                return MockResponse(
+                    status_code=HTTPStatus.OK,
+                    text=str(f.read()),
+                    headers={'Content-Type': 'image/png'},
+                )
+
+        raise AssertionError(f'Unexpected url {url} in bitpanda test')
```
LefterisJP, line 111: I can see where you copied the logic from

```suggestion
        raise AssertionError(f'Unexpected url {url} in asset collection icons test)
```

rotkehlchen/tests/external_apis/test_coingecko.py
```diff
@@ -63,3 +68,64 @@ def test_assets_with_icons(icon_manager):
...
+    with patch('requests.get', wraps=mock_coingecko):
+        icon_dai_eth = icon_manager.get_icon(A_DAI)
+
+    # check that the api was not queried again and the fail returned is the same
+    assert times_api_was_queried == 2
+    assert icon_dai_op == icon_dai_eth
+
+    # try to get an asset without collection
+    icon_manager.get_icon(A_ETH)
+    assert icon_manager.iconfile_path(A_ETH).exists()
```
LefterisJP, line 131: test api was queried again by checking `times_api_was_queried`

LefterisJP (commented):
some small comments

rotkehlchen/tests/external_apis/test_coingecko.py
yabirgb: used a clean cache directory

rotkehlchen/globaldb/handler.py
```diff
@@ -1931,3 +1931,18 @@ def get_asset_type(identifier: str, use_packaged_db: bool = False) -> AssetType:
...
+        lexicographical order.
+        """
+        with GlobalDBHandler().conn.read_ctx() as cursor:
+            cursor.execute('SELECT A.asset FROM multiasset_mappings AS A JOIN multiasset_mappings AS B ON A.collection_id=B.collection_id WHERE B.asset=? ORDER BY A.asset LIMIT 1', (identifier,))  # noqa: E501
+            result = cursor.fetchone()
+
+        if result is not None:
+            result = result[0]
+
+        return result
```
LefterisJP, line 1948:
```suggestion
        return result[0] if result is not None else None
```

rotkehlchen/tests/data/mocks/test_coingecko/coins/ethereum.json
```diff
@@ -0,0 +1 @@
...ations in a decentralized manner; while all states and data are distributed and publicly accessible. Ethereum supports smart contracts in which developers can write code in order to program digital value. Examples of decentralized apps (dapps) that are built on Ethereum includes tokens, non-fungible tokens, decentralized finance apps, lending protocol, decentralized exchanges, and much more.\r\n\r\nOn Ethereum, all transactions and smart contract executions require a small fee to be paid. This fee is called Gas. In technical terms, Gas refers to the unit of measure on the amount of computational effort required to execute an operation or a smart contract. The more complex the execution operation is, the more gas is required to fulfill that operation. Gas fees are paid entirely in Ether (ETH), which is the native coin of the blockchain. The price of gas can fluctuate from time to time depending on the network demand."},"image":{"thumb":"https://assets.coingecko.com/coins/images/279/thumb/ethereum.png?1595348880","small":"https://assets.coingecko.com/coins/images/279/small/ethereum.png?1595348880","large":"https://assets.coingecko.com/coins/images/279/large/ethereum.png?1595348880"}}
```
LefterisJP, line 1: Hmmm interesting. I see we parse the description and save it in memory in the `CoingeckoAssetData` struct. And unless I am totally mistaken this is not used anywhere except from here: https://github.com/rotki/rotki/blob/d01f565e7066bca1c216c6ebbadef2d231946345/rotkehlchen/icons.py#L87

And here we only parse the image url. Correct? So can we get read of the asset description? If yes, please:

1. Do it. This will save us a lot of memory.
2. Also remove it from these mocks. Both this and the other file

LefterisJP (approved):
LGTM. Thanks. The test error is that test we said needs rethinking as it can fail when the market moves: https://twitter.com/LefterisJP/status/1616543619682390018

PR #5441, opened 2023-01-19
Make VCR ignore responses that are not 200

yabirgb:
At the moment VCR doesn't have any filter on the response status code. What I did was adding a new step to the processing rules that doesn't store any response that doesn't have status code 200

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,8 +24,10 @@ format:
 clean:
 	rm -rf build/ dist/ rotkehlchen_py_dist/ htmlcov/ rotkehlchen.egg-info/ *.dmg frontend/app/dist/ frontend/app/build/
 
+
 docker-image:
 	packaging/docker-image.sh
+
 
 test-assets:
 	python pytestgeventwrapper.py rotkehlchen/tests/exchanges/test_binance.py::test_binance_assets_are_known
@@ -48,3 +50,7 @@ test-assets:
 	python pytestgeventwrapper.py rotkehlchen/tests/exchanges/test_independentreserve.py::test_assets_are_known
 	python pytestgeventwrapper.py rotkehlchen/tests/unit/test_zerionsdk.py::test_protocol_names_are_known
 	python pytestgeventwrapper.py rotkehlchen/tests/unit/test_zerionsdk.py::test_query_all_protocol_balances_for_account
+
+
+create-cassettes:
+	RECORD_CASSETTES=true && python pytestgeventwrapper.py -m vcr rotkehlchen/tests
--- a/rotkehlchen/tests/api/test_bitcoin.py
+++ b/rotkehlchen/tests/api/test_bitcoin.py
@@ -258,6 +258,7 @@ def test_add_delete_xpub(rotkehlchen_api_server):
     assert len(result) == 0, 'all xpub mappings should have been deleted'
 
 
+@pytest.mark.vcr
 @pytest.mark.parametrize('number_of_eth_accounts', [0])
 def test_add_delete_xpub_multiple_chains(rotkehlchen_api_server):
     # Disable caching of query results
--- a/rotkehlchen/tests/api/test_makerdao_dsr.py
+++ b/rotkehlchen/tests/api/test_makerdao_dsr.py
@@ -677,6 +677,7 @@ def test_query_historical_dsr_with_a_zero_withdrawal(
     assert len(errors) == 0
 
 
+@pytest.mark.vcr
 @pytest.mark.parametrize('ethereum_accounts', [['0xf753beFE986e8Be8EBE7598C9d2b6297D9DD6662']])
 @pytest.mark.parametrize('ethereum_modules', [['makerdao_dsr']])
 @pytest.mark.parametrize('start_with_valid_premium', [True])
--- a/rotkehlchen/tests/conftest.py
+++ b/rotkehlchen/tests/conftest.py
@@ -6,8 +6,10 @@ import tempfile
 import warnings as test_warnings
 from contextlib import suppress
 from enum import auto
+from http import HTTPStatus
+from json import JSONDecodeError
 from pathlib import Path
-from typing import Any
+from typing import TYPE_CHECKING, Any, Optional
 
 import py
 import pytest
@@ -17,6 +19,10 @@ from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.logging import TRACE, add_logging_level, configure_logging
 from rotkehlchen.tests.utils.args import default_args
 from rotkehlchen.utils.mixins.serializableenum import SerializableEnumMixin
+from rotkehlchen.utils.serialization import jsonloads_dict
+
+if TYPE_CHECKING:
+    from vcr import VCR
 
 
 TESTS_ROOT_DIR = Path(__file__).parent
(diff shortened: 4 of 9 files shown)


rotkehlchen/tests/conftest.py
```diff
@@ -158,10 +161,10 @@ def get_cassette_dir(request: pytest.FixtureRequest) -> Path:
     return Path(request.node.path).relative_to(TESTS_ROOT_DIR).with_suffix('')
 
 
-@pytest.fixture
-def vcr_config(request: pytest.FixtureRequest, default_cassette_name: str) -> dict[str, Any]:
+@pytest.fixture(scope='module')
+def vcr_config(_request: pytest.FixtureRequest) -> dict[str, Any]:
     """
-    vcrpy config
+    vcrpy configd
```
kelsos, line 167: I guess this is accidental here right?

LefterisJP: It's yabir's world-famous typos

rotkehlchen/tests/conftest.py
```diff
@@ -196,3 +191,15 @@ def vcr_cassette_dir(request: pytest.FixtureRequest) -> str:
         base_dir = default_data_directory().parent / 'cassettes'
     cassette_dir = get_cassette_dir(request)
     return str(base_dir / cassette_dir)
+
+
+@pytest.fixture(scope='module', name='vcr')
+def my_vcr(vcr: 'VCR') -> 'VCR':  # pylint: disable=redefined-outer-name
+
```
LefterisJP, line 198: can you add a docstring here explaining what this does?

rotkehlchen/tests/conftest.py
```diff
@@ -170,15 +173,7 @@ def vcr_config(request: pytest.FixtureRequest, default_cassette_name: str) -> di
...
-    rewrite_path = os.environ.get('CASSETTE_REWRITE_PATH')
-    cassette_dir = get_cassette_dir(request)
-    cassette_path = cassette_dir / default_cassette_name
-
-    if rewrite_path and (cassette_path.is_relative_to(rewrite_path) or rewrite_path == '*'):
-        record_mode = 'rewrite'
-    else:
-        record_mode = 'once'
-
+    record_mode = 'once'
```
LefterisJP, line 176: Hmmm does it make sense to have this as an opion anymore then? Remove it completely. Also the docstring needs adjusting. I don't know how vcr.py works to judge what is correct and what not. I just need it to work both locally and in the CI

rotkehlchen/tests/conftest.py
kelsos: This smells like accidentally pressing a button before moving to another file :) 

This happens to me sometimes too 😛

yabirgb: It was not accidental. The CI got stuck and I wanted to force push just to test it :see_no_evil:

kelsos: maybe rebase the branch and change the commit? 😛

Makefile
```diff
@@ -48,3 +50,7 @@ test-assets:
 	python pytestgeventwrapper.py rotkehlchen/tests/exchanges/test_independentreserve.py::test_assets_are_known
 	python pytestgeventwrapper.py rotkehlchen/tests/unit/test_zerionsdk.py::test_protocol_names_are_known
 	python pytestgeventwrapper.py rotkehlchen/tests/unit/test_zerionsdk.py::test_query_all_protocol_balances_for_account
+
+
+create-cassettes:
```
LefterisJP, line 55: I think we need to do it per test. Not for all. I would never run it for all locally and running it for all increases the chance it will record backoffs. And we need to do it locally ofc.

rotkehlchen/tests/api/test_bitcoin.py
```diff
@@ -258,6 +258,7 @@ def test_add_delete_xpub(rotkehlchen_api_server):
     assert len(result) == 0, 'all xpub mappings should have been deleted'
 
 
+@pytest.mark.vcr
```
LefterisJP, line 261: This involves balance queries on-chain. This is not repeatable.

rotkehlchen/tests/api/test_makerdao_dsr.py
```diff
@@ -677,6 +677,7 @@ def mock_get_logs(
     assert len(errors) == 0
 
 
+@pytest.mark.vcr
```
LefterisJP, line 680: Nope. This queries the history for an account and this history can change. This is not something vcr can or should record

rotkehlchen/tests/conftest.py
```diff
@@ -158,41 +164,69 @@ def get_cassette_dir(request: pytest.FixtureRequest) -> Path:
...
+        body = jsonloads_dict(response['body']['string'])
+        result = int(body['status']) == 0 and 'rate limit reached' in body['result']
+    return result
+
+
+@pytest.fixture(scope='module', name='vcr')
+def vcr_fixture(vcr: 'VCR') -> 'VCR':
+    """
+    Update VCR instance to discard error responses during the record mode.
+    This target directly etherscan that is the service we are first focusing on
```
LefterisJP, line 179:
```suggestion
    This targets directly etherscan that is the service we are first focusing on
```

rotkehlchen/tests/conftest.py
```diff
@@ -158,41 +164,69 @@ def get_cassette_dir(request: pytest.FixtureRequest) -> Path:
...
+    # pytest-deadfixtures ignore
```
LefterisJP, line 180: What's this?

rotkehlchen/tests/unit/test_tokens.py
```diff
@@ -19,6 +19,7 @@ def fixture_ethereumtokens(ethereum_inquirer, database, inquirer):  # pylint: di
     return EthereumTokens(database, ethereum_inquirer)
 
 
+@pytest.mark.vcr
```
LefterisJP, line 22: lol man no this can't work. This is detecting tokens which queries the last chain state which is always changing. This is not what VCR is for. This is not repeatable.

rotkehlchen/tests/api/test_makerdao_dsr.py
LefterisJP: Okay as discussed in the call let's go with this. Use vcr anywhere we can and avoid any complicated mocks.

rotkehlchen/tests/conftest.py
LefterisJP: answered in the call: Is for pytest dead fixtures module to ignore this.

rotkehlchen/tests/unit/test_tokens.py
LefterisJP: Okay as discussed in the call let's go with this. Use vcr anywhere we can and avoid any complicated mocks.

rotkehlchen/tests/api/test_bitcoin.py
LefterisJP: Okay as discussed in the call let's go with this. Use vcr anywhere we can and avoid any complicated mocks.

LefterisJP (approved):
LGTM. Will merge. Discussed some things in the call. The 2 small action items from here can go in another PR @yabirgb along with the changes in more tests.

PR #5450, opened 2023-01-22
feat: refetch empty page of histories

lukicenturi:
Fix edge case when deleting trade, ledger action, or other sections, that use backend pagination

For example when you have 11 trades, 
and now you are on page 2 where there is only 1 trade on that page,
then you delete that 1 trade, you get empty page.

I think the better UX here is to move to page 1 instead.

### DIFF
--- a/frontend/app/src/components/helper/CollectionHandler.vue
+++ b/frontend/app/src/components/helper/CollectionHandler.vue
@@ -1,19 +1,32 @@
 <script setup lang="ts">
-import { type PropType } from 'vue';
 import { type Collection } from '@/types/collection';
 import { getCollectionData, setupEntryLimit } from '@/utils/collection';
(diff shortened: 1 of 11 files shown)


frontend/app/src/components/helper/CollectionHandler.vue
```diff
@@ -1,20 +1,33 @@
...
+import { useFrontendSettingsStore } from '@/store/settings/frontend';
 
-const props = defineProps({
-  collection: {
-    required: true,
-    type: Object as PropType<Collection<any>>
-  }
-});
+const props = defineProps<{
+  collection: Collection<any>;
```
kelsos, line 7: For vue 3 there are some guides on how to make generics work with components, in the case of slots it would be nice to have type inference. Unfortunately, the same thing does not seem to work with 2.7

Issue #5414, opened 2023-01-17
Frontend pending issues for v1.27


kelsos:
## Problem Definition

@rotki/backend Let's add any pending issues here so that @rotki/frontend can pick them and fix them. 

## TODO 

- [x] ~~Updates explorers to support Optimism~~ Done from merged PR
- [x] Add RPC node management for Optimism - @lukicenturi https://github.com/rotki/rotki/pull/5425
- [x] Fix counterparties for the optimism decoders we added - @lukicenturi  https://github.com/rotki/rotki/pull/5384
![2023-01-17_12-41](https://user-images.githubusercontent.com/1658405/212896846-ca951927-e58a-4a1a-818d-6413081857f8.png)
Also hop decoder and generally all decoders under optimism need to have their counterparties added to the frontend

- [x] Move this multichain etherscan servicekey to tabs - @lukicenturi  https://github.com/rotki/rotki/pull/5422
![image](https://user-images.githubusercontent.com/26648140/213086621-eca1589b-19d8-4b5a-9099-b6587dc004e1.png)
- [x] Frontend part for https://github.com/rotki/rotki/pull/5426 @kelsos 
- [x] Nothing happening at restart after Global DB upgrade @lukicenturi 
- [x] Add multple addresses at evm transactions filtering @lukicenturi 
- [x] Optimism etherscan key update exposes implementation detail to confirmation. It should be replaced with a proper resource/text. @lukicenturi 
![image](https://user-images.githubusercontent.com/2269732/213770885-9eb85342-ce81-4202-815b-e120d038a804.png)

@lukicenturi if you pick any of them can you add a note next to the issue?

PR #5452, opened 2023-01-22
fix: some issues for fe

lukicenturi:
Part of https://github.com/rotki/rotki/issues/5414
FE part of #5451

## Checklist

- [x] Add multple addresses at evm transactions filtering
- [x] Optimism etherscan key update exposes implementation detail to confirmation. It should be replaced with a proper resource/text.

### DIFF
--- a/frontend/app/src/components/asset-manager/MergeDialog.vue
+++ b/frontend/app/src/components/asset-manager/MergeDialog.vue
@@ -113,15 +113,15 @@ const v$ = useVuelidate(
       <div v-if="done">{{ t('merge_dialog.done') }}</div>
 
       <v-form v-else :value="!v$.$invalid">
-        <v-text-field
+        <asset-select
           v-model="source"
-          :label="t('merge_dialog.source.label')"
-          :error-messages="v$.source.$errors.map(e => e.$message)"
           outlined
-          :disabled="pending"
-          persistent-hint
+          :error-messages="v$.source.$errors.map(e => e.$message)"
+          :label="t('merge_dialog.source.label')"
           :hint="t('merge_dialog.source_hint')"
-          @focus="clearErrors()"
+          persistent-hint
+          :disabled="pending"
+          @focus="clearErrors"
         />
         <v-row align="center" justify="center" class="my-4">
           <v-col cols="auto">
@@ -134,6 +134,7 @@ const v$ = useVuelidate(
           :error-messages="v$.target.$errors.map(e => e.$message)"
           :label="tc('merge_dialog.target.label')"
           :disabled="pending"
+          @focus="clearErrors"
         />
       </v-form>
     </card>
--- a/frontend/app/src/pages/settings/api-keys/external/index.vue
+++ b/frontend/app/src/pages/settings/api-keys/external/index.vue
@@ -14,7 +14,7 @@ import {
   type ExternalServiceKeys,
   type ExternalServiceName
 } from '@/types/user';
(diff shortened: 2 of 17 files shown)


frontend/app/src/types/history/tx.ts
```diff
@@ -83,6 +83,11 @@ export interface AddressesAndEvmChainPayload {
   readonly evmChain: string;
 }
 
+export interface AddressAndEvmChainPayload {
```
kelsos, line 86: I am pretty sure I already added a type that takes address and evmChain in the websocket PR, not sure if I exported it though. Maybe it would make sense to use that one?

frontend/app/src/store/history/transactions.ts
```diff
@@ -108,15 +107,16 @@ export const useTransactions = defineStore('history/transactions', () => {
...
-            }).toString()
-          : undefined,
-        address
+        description:
+          accounts && accounts.length > 0
+            ? t('actions.transactions.task.description', {
+                address: accounts[0].address,
+                chain: accounts[0].evmChain
+              }).toString()
+            : undefined
```
kelsos, line 119: So with this, you will get the first account if you have multiple or non if empty?

The biggest problem I can think of is when you mix different chains and addresses but for example, you could join and show multiple.

frontend/app/src/pages/settings/api-keys/external/index.vue
```diff
@@ -82,7 +82,7 @@ const save = async (serviceName: ExternalServiceName, key: string) => {
     setMessage({
       title: tc('external_services.set.success.title'),
       description: tc('external_services.set.success.message', 0, {
-        serviceName
+        serviceName: serviceName.split('_').join(' ')
```
kelsos, line 85: Can you add some capitalization too?

frontend/app/src/components/asset-manager/MergeDialog.vue
```diff
@@ -113,15 +113,15 @@ const v$ = useVuelidate(
       <div v-if="done">{{ t('merge_dialog.done') }}</div>
 
       <v-form v-else :value="!v$.$invalid">
-        <v-text-field
+        <asset-select
```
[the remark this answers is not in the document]
lukicenturi: unless we should be able to input asset identifier that doesn't registered in db, I believe it's okay.

frontend/app/src/store/history/transactions.ts
lukicenturi: this is for ignoreCache=True. We always call it inside the accounts loop
```
        const refreshAddressTxs = accountsList.map(account =>
          fetchTransactionsHandler(false, { accounts: [account] }).catch(
```

So it is guaranteed that the length of the `accounts` here is 1

kelsos: so it can be either 1 or none? maybe then we can re-evaluate it as some point

frontend/app/src/types/history/tx.ts
```diff
@@ -83,6 +83,13 @@ export interface AddressesAndEvmChainPayload {
   readonly evmChain: string;
 }
 
+export const EvmChainAddress = z.object({
+  address: z.string(),
+  evmChain: z.string()
+});
+
+export type EvmChainAddress = z.infer<typeof EvmChainAddress>;
```
kelsos, line 91: I will merge it as it is but we should maybe move the type somewhere else in the future as it is not only related to transactions

PR #5456, opened 2023-01-23
Refactor updates

yabirgb:
This PR addresses two related problems:

- First is a bug that was copying assets from the packaged globaldb to the user's globaldb during updates. This caused a bug resulting in an infinite loop trying to solve conflicts
- Second problem solved is preventing the globaldb from getting modified during the assets update. To do this we first query all the needed files and then apply the updates in a critical section

### DIFF
--- a/rotkehlchen/assets/asset.py
+++ b/rotkehlchen/assets/asset.py
@@ -75,15 +75,17 @@ class Asset:
     def serialize(self) -> str:
         return self.identifier
 
-    def check_existence(self) -> 'Asset':
+    def check_existence(self, query_packaged_db: bool = True) -> 'Asset':
         """
         If this asset exists, returns the instance. If it doesn't, throws an error.
+        When the `query_packaged_db` is set to True and the checked asset is in the list
+        of constant asset we try to copy it from the packaged global db.
         May raise:
         - UnknownAsset
         """
         # We don't need asset type, but using `get_asset_type` since it has all the functionality
         # that we need here
-        AssetResolver().get_asset_type(self.identifier)
+        AssetResolver().get_asset_type(self.identifier, query_packaged_db=query_packaged_db)
         return self
 
     def is_nft(self) -> bool:
--- a/rotkehlchen/globaldb/updates.py
+++ b/rotkehlchen/globaldb/updates.py
@@ -61,28 +61,28 @@ def _replace_assets_from_db(
         connection: 'DBConnection',
         sourcedb_path: Path,
 ) -> None:
-    cursor = connection.cursor()
-    cursor.executescript(f"""
-    ATTACH DATABASE "{sourcedb_path}" AS other_db;
-    PRAGMA foreign_keys = OFF;
-    DELETE FROM assets;
-    DELETE FROM evm_tokens;
-    DELETE FROM underlying_tokens_list;
-    DELETE FROM common_asset_details;
-    DELETE FROM asset_collections;
-    DELETE FROM multiasset_mappings;
-    INSERT INTO assets SELECT * FROM other_db.assets;
-    INSERT INTO evm_tokens SELECT * FROM other_db.evm_tokens;
-    INSERT INTO underlying_tokens_list SELECT * FROM other_db.underlying_tokens_list;
-    INSERT INTO common_asset_details SELECT * FROM other_db.common_asset_details;
-    INSERT INTO asset_collections SELECT * FROM other_db.asset_collections;
-    INSERT INTO multiasset_mappings SELECT * FROM other_db.multiasset_mappings;
-    INSERT OR REPLACE INTO settings(name, value) VALUES("{ASSETS_VERSION_KEY}",
-    (SELECT value FROM other_db.settings WHERE name="{ASSETS_VERSION_KEY}")
-    );
-    PRAGMA foreign_keys = ON;
-    DETACH DATABASE "other_db";
-    """)
+    with connection.write_ctx() as cursor:
+        cursor.executescript(f"""
+        ATTACH DATABASE "{sourcedb_path}" AS other_db;
+        PRAGMA foreign_keys = OFF;
+        DELETE FROM assets;
+        DELETE FROM evm_tokens;
+        DELETE FROM underlying_tokens_list;
+        DELETE FROM common_asset_details;
+        DELETE FROM asset_collections;
+        DELETE FROM multiasset_mappings;
+        INSERT INTO assets SELECT * FROM other_db.assets;
+        INSERT INTO evm_tokens SELECT * FROM other_db.evm_tokens;
+        INSERT INTO underlying_tokens_list SELECT * FROM other_db.underlying_tokens_list;
+        INSERT INTO common_asset_details SELECT * FROM other_db.common_asset_details;
+        INSERT INTO asset_collections SELECT * FROM other_db.asset_collections;
+        INSERT INTO multiasset_mappings SELECT * FROM other_db.multiasset_mappings;
+        INSERT OR REPLACE INTO settings(name, value) VALUES("{ASSETS_VERSION_KEY}",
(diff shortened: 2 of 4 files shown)


rotkehlchen/globaldb/updates.py
```diff
@@ -653,11 +654,16 @@ def _perform_update(
             up_to_version: Optional[int],
             updates: dict[int, dict[str, str]],
     ) -> None:
+        """
+        Apply to the db the different sql updates from the `updates` argument
+        """
         version = self.local_assets_version + 1
         target_version = min(up_to_version, self.last_remote_checked_version) if up_to_version else self.last_remote_checked_version   # noqa: E501
         # type ignore since due to check_for_updates we know last_remote_checked_version exists
         while version <= target_version:
```
nebolax, line 663: Hmm, can't this be changed to a `for` loop?

rotkehlchen/globaldb/updates.py
```diff
@@ -602,49 +600,115 @@ def perform_update(
...
+        updates = self._retrieve_update_files(
+            local_schema_version=local_schema_version,
+            infojson=infojson,
+            up_to_version=up_to_version,
+        )
+
         with TemporaryDirectory() as tmpdirname:
             tmpdir = Path(tmpdirname)
+            temp_db_name = 'temp.db'
             connection, _ = initialize_globaldb(
```
nebolax, line 615: I would rename it to something like `temp_db_connection`

rotkehlchen/globaldb/updates.py
```diff
@@ -602,49 +600,115 @@ def perform_update(
...
-
-            # otherwise we are sure the DB will work without conflicts so let's
-            # now move the data to the actual global DB
-            connection.close()
-            connection = GlobalDBHandler().conn
-            with connection.critical_section():  # assure global DB is not accessed anywhere else
-                _replace_assets_from_db(connection, tmpdir / 'temp.db')
+                connection = GlobalDBHandler().conn
+                log.debug('Replace globaldb')
+                _replace_assets_from_db(connection, tmpdir / temp_db_name)
```
nebolax, line 646: I would pass `GlobalDBHandler().conn` directly, without assigning it to the `connection` variable first

rotkehlchen/globaldb/updates.py
```diff
@@ -602,49 +600,115 @@ def perform_update(
...
-                # close the temporary DB and return the conflicts
+                self._perform_update(
+                    connection=connection,
+                    assets_conflicts=conflicts,
+                    up_to_version=up_to_version,
+                    updates=updates,
+                )
+                if len(self.conflicts) != 0:
+                    # close the temporary DB and return the conflicts
+                    connection.close()
```
nebolax, line 635: the connection gets closed anyway, so it can happen before this `if`

nebolax (commented):
Thanks. Left some comments / questions

rotkehlchen/globaldb/updates.py
```diff
@@ -138,7 +136,7 @@ class AssetsUpdater():
     def __init__(self, msg_aggregator: MessagesAggregator) -> None:
         self.msg_aggregator = msg_aggregator
         self.local_assets_version = GlobalDBHandler().get_setting_value(ASSETS_VERSION_KEY, 0)
-        self.last_remote_checked_version = None
+        self.last_remote_checked_version = -1  # integer value that represents no update
```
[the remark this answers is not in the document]
yabirgb: I changed it to avoid the weird ignore type comments since we are doing arithmetical operations with a nullable value. Changed it to -1 since it won't break the logic

rotkehlchen/globaldb/handler.py
```diff
@@ -1893,6 +1893,7 @@ def resolve_asset_from_packaged_and_store(identifier: str) -> AssetWithNameAndTy
             asset_data['swapped_for'] = getattr(asset, 'swapped_for', None)
 
         if asset_count == 0:
+            log.debug('Copying asset')
```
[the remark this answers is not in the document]
yabirgb: Is the other way around introduce and then delete but I'll squash and merge both commits

rotkehlchen/globaldb/updates.py
```diff
@@ -443,7 +443,8 @@ def _handle_asset_update(
         """
         local_asset: Optional[Asset] = None
         with suppress(UnknownAsset):
-            local_asset = Asset(remote_asset_data.identifier).check_existence()
+            # we avoid querying the packaged db to prevent the copy of constant assets
+            local_asset = Asset(remote_asset_data.identifier).check_existence(query_packaged_db=False)  # noqa: E501
```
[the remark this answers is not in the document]
yabirgb: 1. This extra condition here prevents us from introducing this issue without noticing in the future. It might happen that we forget about this constraint and the bug appears again with us not noticing.  https://discord.com/channels/657906918408585217/683245846070165504/1066110368402899065


For the bug you have to keep in mind that we have 3 globaldbs:
1. The user globaldb
2. The packaged globaldb
3. An intermediate globaldb that is a copy of 1. and where we apply the updats

The sequence for the infinite loop was:

1. Update starts
2. OP_ETH appeas in the update and by checking its existence we copy it since is in the constants assets we copy it
3. When trying to apply again the update with the conflicts solved a new one appears, `OP_ETH` and is the only conflict now
4. The state of the conflicts is not kept and since you have only resolved the conflict with `OP_ETH` the others are still there and this restarts the chain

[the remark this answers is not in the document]
nebolax: Ah yes, I see, so it gets added to the real globaldb, not the copy that is used for updating. Yeah, alright, thanks for the explanation :+1:

rotkehlchen/globaldb/updates.py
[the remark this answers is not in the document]
nebolax: Alright :+1: But then a comment on line 659 should be removed

rotkehlchen/assets/asset.py
```diff
@@ -75,15 +75,17 @@ def to_dict(self) -> dict[str, Any]:
     def serialize(self) -> str:
         return self.identifier
 
-    def check_existence(self) -> 'Asset':
+    def check_existence(self, query_packaged_db: bool = True) -> 'Asset':
         """
         If this asset exists, returns the instance. If it doesn't, throws an error.
+        When the `query_packaged_db` is set to True and the checked asset is in the list
+        of constant asset we try to copy it from the packaged global db.
```
LefterisJP, line 82:
```suggestion
        of constant assets we try to copy it from the packaged global db.
```

rotkehlchen/tests/unit/test_globaldb_asset_updates.py
```diff
@@ -440,3 +445,29 @@ def test_updates_assets_collections_errors(assets_updater: AssetsUpdater):
...
+        assets_updater.perform_update(up_to_version=999, conflicts={})
+
+    with GlobalDBHandler().conn.read_ctx() as cursor:
+        cursor.execute('SELECT * FROM asset_collections WHERE id = 99999999')
+        assert cursor.fetchall() == [(99999999, 'My custom ETH', 'ETHS')]
+        cursor.execute('SELECT * FROM multiasset_mappings WHERE collection_id = 99999999')
+        assert cursor.fetchall() == [(99999999, 'ETH'), (99999999, 'eip155:1/erc20:0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84')]  # noqa: E501
+
+    # check that we skip versions with wrong schema and that all the versions
+    # required are corretly queried.
```
LefterisJP, line 469:
```suggestion
    # required are correctly queried.
```

LefterisJP (commented):
LGTM. 2 nitpicks left for another PR. Before merging can you tell me if the test fails also locally?
`rotkehlchen/tests/api/test_balances.py::test_ethereum_tokens_detection`

PR #5459, opened 2023-01-23
Allow account removal not be delayed by transaction queries

prettyirrelevant:


rotkehlchen/rotkehlchen.py
```diff
@@ -168,6 +169,14 @@ def __init__(self, args: argparse.Namespace) -> None:
         self.shutdown_event = gevent.event.Event()
         self.migration_manager = DataMigrationManager(self)
 
+    def kill_running_tx_query_tasks(self) -> None:
+        """Attempts to kill greenlets tasks querying transactions."""
+        log.debug(f'Attempting to delete running transaction queries tasks from {self.api_task_greenlets}')  # noqa: E501
+        for task in self.api_task_greenlets:
+            if task.is_txn_query is True:
+                task.greenlet.kill()
+                self.api_task_greenlets.remove(task)
```
nebolax, line 178: Doesn't it break because of trying to delete an element from the list that is being iterated?

rotkehlchen/rotkehlchen.py
```diff
@@ -168,6 +169,14 @@ def __init__(self, args: argparse.Namespace) -> None:
         self.shutdown_event = gevent.event.Event()
         self.migration_manager = DataMigrationManager(self)
 
+    def kill_running_tx_query_tasks(self) -> None:
+        """Attempts to kill greenlets tasks querying transactions."""
```
nebolax, line 173: Why does it "attempt"? Are there any scenarios when it fails to kill the greenlets?

rotkehlchen/api/rest.py
```diff
@@ -321,7 +323,10 @@ def _query_async(self, command: Callable, **kwargs: Any) -> Response:
         )
         greenlet.task_id = task_id
         greenlet.link_exception(self._handle_killed_greenlets)
-        self.rotkehlchen.api_task_greenlets.append(greenlet)
+        self.rotkehlchen.api_task_greenlets.append(ApiTaskGreenlet(
```
nebolax, line 326: General question. Isn't it missing the tracking and killing of periodic tasks?

rotkehlchen/tests/api/blockchain/test_evm.py
```diff
@@ -284,3 +289,41 @@ def test_add_multievm_accounts(rotkehlchen_api_server):
...
+            'limit': 1000,
+            'address': ADDY,
+            'evm_chain': 'ethereum',
+        },
+    )
+    assert_ok_async_response(response)
+
+    # now delete the address in sync -- and see that it does not take time.
+    seconds = 2
+    with gevent.Timeout(seconds):
```
nebolax, line 317: I would also somehow check whether the task was indeed killed.

Also I think will make sense to explicitly check if the locks were acquired and then released

rotkehlchen/rotkehlchen.py
prettyirrelevant: If no greenlet is running at that point, perhaps the docstring can be better.

rotkehlchen/api/rest.py
prettyirrelevant: yes, adding that.

rotkehlchen/tests/api/blockchain/test_evm.py
prettyirrelevant: done.

rotkehlchen/rotkehlchen.py
nebolax: What about this?

rotkehlchen/rotkehlchen.py
```diff
@@ -168,6 +169,29 @@ def __init__(self, args: argparse.Namespace) -> None:
...
+            f'Attempting to kill running transaction queries tasks '
+            f'from api task greenlets {self.api_task_greenlets} and '
+            f'task manager running greenlets {self.task_manager.running_greenlets}',
+        )
+        for task in self.api_task_greenlets:
+            if task.is_txn_query is True:
+                task.greenlet.kill()
+                self.api_task_greenlets.remove(task)
+
+        task_names = {f'Query {blockchain.name.lower()} transactions for {account}' for account in accounts}  # noqa: E501
```
nebolax, line 189: Hmmm, this comparison by name looks brittle to me. Maybe compare with a function or add a special flag for such greenlets (like `is_txn_query`) or something like that?

rotkehlchen/api/rest.py
LefterisJP, line 326: why go through all these? And why make a wrapper with a boolean? This is a terrible approach. Absolutely no. Ask before taking such an approach.

Hacky as fuck with the extra kwargs that is checked in every `query_async` invocation.

Wow ..

rotkehlchen/tests/api/blockchain/test_evm.py
```diff
@@ -284,3 +289,48 @@ def test_add_multievm_accounts(rotkehlchen_api_server):
     assert result == [
         {'address': common_account, 'label': label, 'tags': ['metamask']},
     ]
+
+
+@pytest.mark.parametrize('ethereum_accounts', [[ADDY]])
+def test_evm_account_deletion_does_not_wait_for_pending_txn_queries(rotkehlchen_api_server) -> None:  # noqa: E501
+    """
+    Test that if transactions for an address is being queried and account removal is issued,
```
LefterisJP, line 297:
```suggestion
    Test that if transactions for an address are being queried and removal is requested for that address,
```

rotkehlchen/tests/api/blockchain/test_evm.py
```diff
@@ -284,3 +289,48 @@ def test_add_multievm_accounts(rotkehlchen_api_server):
...
+    the transactions querying greenlet is killed & the account is deleted subsequently.
```
LefterisJP, line 298:
```suggestion
    the transactions querying greenlet is killed & the account is subsequently deleted.
```

rotkehlchen/tests/api/blockchain/test_evm.py
```diff
@@ -284,3 +289,48 @@ def test_add_multievm_accounts(rotkehlchen_api_server):
...
+            'evmtransactionsresource',
+        ), json={
+            'async_query': True,
+            'only_cache': False,
+            'limit': 1000,
+            'accounts': [{'address': ADDY, 'evm_chain': 'ethereum'}],
+            'evm_chain': 'ethereum',
+        },
+    )
+    assert_ok_async_response(response)
```
LefterisJP, line 313: Don't let this test make remote queries. Since we want to emulate a long running transaction query patch one of the underlying functions of the transaction query (not the one the greenlet calls), to run an infinite loop (with gevent sleeps in between so it yields)

LefterisJP (commented):
Absolutely terrible approach. Making a wrapper and changing the logic in so many places just for this one thing and making it a special case and create technical debt for us down the line?

Lol what are you thinking???

The wrapper is not required. `Gevent` already has fields you can add. Such as we already do with the `task_id`. So if you wanted a boolean you could do it like that. And no need for a wrapper and unwrapping everywhere.

What's more this is a bad approach. The way greenlets for api are called is always the same. It spawns a greenlet that runs `self._do_query_async` that runs a command, for a task if and some kwargs.

So if you want to filter and find specific `command` you can do that by just looking at the greenlet itself. Just read the docs and understand the tools we are using.

So you could do something like:

```python
greenlet.args[0].__func__.__qualname__ == `RestAPI._get_evm_transactions`
```

to filter the right greenlet.

rotkehlchen/rotkehlchen.py
```diff
@@ -170,6 +170,24 @@ def __init__(self, args: argparse.Namespace) -> None:
         self.shutdown_event = gevent.event.Event()
         self.migration_manager = DataMigrationManager(self)
 
+    def kill_running_tx_query_tasks(self) -> None:
```
LefterisJP, line 173: wtf man no? You should kill the greenlet that queries transaction for that address you are deleting. Not for all addresses.

rotkehlchen/rotkehlchen.py
```diff
@@ -170,6 +170,24 @@ def __init__(self, args: argparse.Namespace) -> None:
...
+        """Checks for running greenlets related to transactions query and kills them."""
+        assert self.task_manager is not None
```
LefterisJP, line 175:
```suggestion
        assert self.task_manager is not None, 'task manager should have been initialized at this point'
```

rotkehlchen/rotkehlchen.py
```diff
@@ -170,6 +170,24 @@ def __init__(self, args: argparse.Namespace) -> None:
...
+        log.debug(
+            f'Attempting to kill running transaction queries tasks '
+            f'from api task greenlets {self.api_task_greenlets} and '
+            f'task manager running greenlets {self.task_manager.running_greenlets}',
+        )
+        for greenlet in self.api_task_greenlets:
+            if greenlet.args[0].__func__.__name__ == '_get_evm_transactions':
```
LefterisJP, line 182: no no. Yesterday I told you. Check the full name including the class. This way you will trigger for any class's `_get_evm_transactions`

rotkehlchen/rotkehlchen.py
```diff
@@ -170,6 +170,24 @@ def __init__(self, args: argparse.Namespace) -> None:
...
+                greenlet.kill()
+        self.api_task_greenlets = [greenlet for greenlet in self.api_task_greenlets if not greenlet.dead]  # noqa: E501
+
+        for fn, greenlets in list(self.task_manager.running_greenlets.items()):
+            if fn.__name__ == '_maybe_query_evm_transactions':
```
LefterisJP, line 187: no no no. This will kill them all. Specifically check for `evm_manager.transactions.single_address_query_transactions` and the argument being that address. And the different evm manager depending on the chain id

prettyirrelevant: `chain id` cannot be used here since `transactions.single_address_query_transactions` calls the parent class directly.

So the `__qualname__` evaluates to `EvmTransactions.single_address_query_transactions` not the chain specific child class

LefterisJP (commented):
I am taking over the PR as it's not doing what I asked after multiple reviews.

LefterisJP (approved):
all tests green, so I am merging

LefterisJP:
@prettyirrelevant and @nebolax please see the PR and understand how this should have been done in the first place. Spending time doing it wrong review after review and in the end having a complete failure of communication and understanding between us is frustrating, a waste of time and a waste of money.

PR #5461, opened 2023-01-23
Sort by rate instead of amount in HIFO

nebolax:
Sort by rate instead of amount in HIFO


prettyirrelevant (approved):

LefterisJP (approved):
LGTM too, though just ceremonial. Merging

PR #5469, opened 2023-01-24
eth staking rewards are now used in accounting

### LINKED ISSUE #5455: ETH staking rewards show as 'missing acquisition' if you try to sell them
## Problem Definition

When creating a profit/loss report for a Rotki account that just has a beacon node validator generating rewards, and a test transaction to sell just those rewards, then the report will have an error that the ETH rewards don't have a corresponding acquisition.

Expected outcome: the report should not have any errors.

I know this is theoretical at this stage because withdrawals are not enabled but I was just testing it out.

prettyirrelevant:
Closes #5455

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,6 +9,7 @@ Changelog
 * :feature:`5001` The PnL report can now be generated with the Highest-In First Out (HIFO) accounting method.
 * :feature:`1793` The PnL report can now be generated with the average cost basis accounting method.
 * :feature:`5148` Users will now see tokens detected for accounts having a DSProxy
+* :bug:`5455` ETH2 staking rewards are now taken into consideration during accounting.
 * :bug:`4332` Price oracles are now temporarily penalized after repeated failures.
 * :bug:`-` Fixed an edge case where users would be swamped with "This socket is already used by another greenlet" errors during websocket communication between backend and frontend.
 * :bug:`-` Transfers of ether between tracked accounts will now have a correct label in the UI.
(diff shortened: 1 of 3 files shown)


docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`5455` ETH2 staking rewards are now taken into consideration during accounting.
```
nebolax, line 5: Hmm it is kinda mixed here already, but if I am not mistaken, we tend to keep the list of features on top and the list of bugs in the bottom

rotkehlchen/tests/unit/test_history.py
```diff
@@ -162,3 +163,51 @@ def test_pnl_processing_with_eth2_staking_setting(accountant, db_settings):
...
+    no_message_errors(accountant.msg_aggregator)
+    if db_settings['eth_staking_taxable_after_withdrawal_enabled'] is True:
+        expected_pnls = PnlTotals({
+            AccountingEventType.STAKING: PNL(taxable=ZERO, free=ZERO),
+        })
+    else:
+        expected_pnls = PnlTotals({
+            AccountingEventType.STAKING: PNL(taxable=FVal(473.1710), free=ZERO),
+        })
+    check_pnls_and_csv(accountant, expected_pnls, None)
```
nebolax, line 213: Missing new line :(

nebolax (commented):
Just a few super minor things

LefterisJP (commented):
Hmm I don't think this is the right approach. The test is not bad but it's before we have withdrawals processing.

Think you were right in your initial assesment that nothing is wrong.

I think that the correct approach here is either:
1. Remove ETH2 as an asset completely after a particular upgrade
2. When withdrawals start existing and we add processing for them, then and only then would the accounting make sense. Since at withdrawal ETH2 (so locked ETH) becomes ETH and accounting would match.

docs/changelog.rst
LefterisJP: just fyi the order here does not really matter. It's just for us to look nice. It's rendered differently in the actual changelog: https://rotki.readthedocs.io/en/latest/changelog.html

LefterisJP:
Closing this as discussed.

PR #5476, opened 2023-01-25
improve speed in the slowest tests

yabirgb:
This PR:

- mocks the slowest tests we had
- add a new make rule `create-cassette` to record a single test
- avoids the creation of extra ethereum addresses that were getting queried in the tests

## Checklist

### DIFF
--- a/rotkehlchen/tests/api/test_ethereum_transactions.py
+++ b/rotkehlchen/tests/api/test_ethereum_transactions.py
@@ -35,6 +35,7 @@ from rotkehlchen.tests.utils.checks import assert_serialized_lists_equal
 from rotkehlchen.tests.utils.constants import TXHASH_HEX_TO_BYTES
 from rotkehlchen.tests.utils.ethereum import (
     TEST_ADDR1,
+    TEST_ADDR2,
     TEST_ADDR3,
     extended_transactions_setup_test,
     setup_ethereum_transactions_test,
@@ -1505,6 +1506,7 @@ def test_no_value_eth_transfer(rotkehlchen_api_server: 'APIServer'):
     assert result['entries'][0]['decoded_events'][0]['entry']['balance']['amount'] == '0'
 
 
+@pytest.mark.parametrize('ethereum_accounts', [[TEST_ADDR1, TEST_ADDR2]])
 def test_decoding_missing_transactions(rotkehlchen_api_server: 'APIServer') -> None:
     """Test that decoding all pending transactions works fine"""
     rotki = rotkehlchen_api_server.rest_api.rotkehlchen
@@ -1554,6 +1556,7 @@ def test_decoding_missing_transactions(rotkehlchen_api_server: 'APIServer') -> N
     assert outcome['result']['decoded_tx_number'] == {}
 
 
+@pytest.mark.parametrize('ethereum_accounts', [[TEST_ADDR1, TEST_ADDR2]])
 def test_decoding_missing_transactions_by_address(rotkehlchen_api_server: 'APIServer') -> None:
     """Test that decoding all pending transactions works fine when a filter by address is set"""
     rotki = rotkehlchen_api_server.rest_api.rotkehlchen
--- a/rotkehlchen/tests/integration/test_transactions.py
+++ b/rotkehlchen/tests/integration/test_transactions.py
@@ -1,23 +1,31 @@
+from typing import TYPE_CHECKING
+
 import pytest
 from flaky import flaky
 
 from rotkehlchen.db.evmtx import DBEvmTx
 from rotkehlchen.db.filtering import EvmTransactionsFilterQuery
 from rotkehlchen.tests.utils.ethereum import (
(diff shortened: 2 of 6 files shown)


rotkehlchen/tests/api/test_ethereum_transactions.py
```diff
@@ -1505,6 +1506,7 @@ def test_no_value_eth_transfer(rotkehlchen_api_server: 'APIServer'):
     assert result['entries'][0]['decoded_events'][0]['entry']['balance']['amount'] == '0'
 
 
+@pytest.mark.parametrize('ethereum_accounts', [[TEST_ADDR1, TEST_ADDR2]])
```
LefterisJP, line 1509: why not freeze time here and in 1559 too?

rotkehlchen/tests/integration/test_transactions.py
```diff
@@ -1,23 +1,31 @@
...
+    from rotkehlchen.chain.ethereum.transactions import EthereumTransactions
+    from rotkehlchen.db.dbhandler import DBHandler
+
 
 @flaky(max_runs=3, min_passes=1)  # failed in a flaky way sometimes in the CI due to etherscan
-@pytest.mark.parametrize(*ETHERSCAN_AND_INFURA_PARAMS)
+@pytest.mark.parametrize('ethereum_accounts', [[TEST_ADDR1, TEST_ADDR2]])
 @pytest.mark.parametrize('transaction_already_queried', [True, False])
+@pytest.mark.freeze_time('2023-01-24 22:45:45 GMT')
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
```
LefterisJP, line 23: what is this?

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -186,6 +190,7 @@ def test_maybe_schedule_exchange_query_ignore_exchanges(
 
 
 @pytest.mark.parametrize('one_receipt_in_db', [True, False])
+@pytest.mark.parametrize('ethereum_accounts', [[TEST_ADDR1, TEST_ADDR2]])
```
LefterisJP, line 193: same question here on freezing

rotkehlchen/tests/utils/ethereum.py
```diff
@@ -161,18 +162,9 @@ def setup_ethereum_transactions_test(
         one_receipt_in_db: bool = False,
         second_receipt_in_db: bool = False,
 ) -> tuple[list[EvmTransaction], list[EvmTxReceipt]]:
+    """This setup assummes that TEST_ADDR1 and TEST_ADDR2 are already present in the db"""
```
LefterisJP, line 165:
```suggestion
    """This setup assumes that TEST_ADDR1 and TEST_ADDR2 are already present in the db"""
```

LefterisJP (commented):
All right so reviewed. Some questions and also a big question mark. This took 45 mins, which is more than other runs.

rotkehlchen/tests/api/test_ethereum_transactions.py
yabirgb: it was not needed as the queries didn't depend on the time

rotkehlchen/tests/integration/test_transactions.py
yabirgb: is an option by vcr to remove a param from the query, in this case the etherscan apiKey query param

rotkehlchen/tests/utils/ethereum.py
```diff
@@ -161,18 +162,9 @@ def setup_ethereum_transactions_test(
...
     dbevmtx = DBEvmTx(database)
     tx_hash1 = deserialize_evm_tx_hash('0x692f9a6083e905bdeca4f0293f3473d7a287260547f8cbccc38c5cb01591fcda')  # noqa: E501
-    addr2 = string_to_evm_address('0x442068F934BE670aDAb81242C87144a851d56d16')
```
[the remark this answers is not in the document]
yabirgb: No, this change was because in the test `test_get_transaction_receipt` we were creating one extra account in the mocked ethereum object in addition to those checked in the test. I changed it to create explicitly only the addressed tested. This doesn't affect any other VCR test, only those using `setup_ethereum_transactions_test`

rotkehlchen/tests/integration/test_transactions.py
LefterisJP: To not record it or replay it? I guess its' nice to have but not really important as the key is already hard-coded and free :P

rotkehlchen/tests/api/test_ethereum_transactions.py
LefterisJP: Thing is a lot of things underneat may depend on time and all the queries that may be formed and recorded may be time dependent.

Result may not, but the queries sent may be. I guess if not needed it's fine but we also gotta think about maintenance and how tests will change in the future.

LefterisJP (approved):
Please work, please please please

PR #5479, opened 2023-01-25
Fix last tag deletion

nebolax:
Now if the user will want to delete all the remaining tags of a blockchain account they will be able to do so.

### DIFF
--- a/rotkehlchen/db/dbhandler.py
+++ b/rotkehlchen/db/dbhandler.py
@@ -1117,7 +1117,8 @@ class DBHandler:
         # Delete the current tag mappings for all affected accounts
         write_cursor.executemany(
             'DELETE FROM tag_mappings WHERE '
-            'object_reference = ?;', [(f'{x.chain}{x.address}',) for x in account_data],
+            # Using chain.value since it has to match the key generated in `_prepare_tag_mappings`.
+            'object_reference = ?;', [(f'{x.chain.value}{x.address}',) for x in account_data],
         )
 
         # Update the blockchain account labels in the DB
--- a/rotkehlchen/tests/api/test_tags.py
+++ b/rotkehlchen/tests/api/test_tags.py
@@ -3,7 +3,12 @@ from http import HTTPStatus
 import pytest
 import requests
 
-from rotkehlchen.tests.utils.api import api_url_for, assert_error_response, assert_proper_response
+from rotkehlchen.tests.utils.api import (
+    api_url_for,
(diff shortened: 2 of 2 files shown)


rotkehlchen/tests/api/test_tags.py
```diff
@@ -678,3 +683,72 @@ def test_delete_utilized_tag(rotkehlchen_api_server):
...
+        eth_balances=None,
+        token_balances=None,
+        btc_balances=btc_balances,
+    )
+    with setup.bitcoin_patch:
+        response = requests.put(
+            api_url_for(
+                rotkehlchen_api_server,
+                'blockchainsaccountsresource',
+                blockchain='BTC',
```
yabirgb, line 720: since you are not testing the btc account creation is this needed? I believe you can manually add the btc address to the `blockchain_accounts` table and assign the already created tag directly there avoiding to use the api and setting the balances.

I believe it should be possible

yabirgb (commented):
A few comments

rotkehlchen/db/dbhandler.py
```diff
@@ -1117,7 +1117,7 @@ def edit_blockchain_accounts(
         # Delete the current tag mappings for all affected accounts
         write_cursor.executemany(
             'DELETE FROM tag_mappings WHERE '
-            'object_reference = ?;', [(f'{x.chain}{x.address}',) for x in account_data],
+            'object_reference = ?;', [(f'{x.chain.value}{x.address}',) for x in account_data],
```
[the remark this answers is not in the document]
nebolax: Will add

rotkehlchen/tests/api/test_tags.py
```diff
@@ -678,3 +683,69 @@ def test_delete_utilized_tag(rotkehlchen_api_server):
...
+            ), json={
+                'accounts': [{
+                    'address': UNIT_BTC_ADDRESS1,
+                    'label': 'my btc miner',
+                    'tags': ['public'],
+                }],
+            },
+        )
+    assert_proper_response(response)
+    # Now delete the tag from the account
```
LefterisJP, line 726: nitpick: You do not delete the tag. To trigger the error you are editing the account in order to give it new tags and make sure old tags are gone.

Please adjust this in a future PR. Also perhaps add two tags to that btc address, and in the edit, give it one. This would be more realistic a test.

LefterisJP (approved):
LGTM with a small TODO for later.

PR #5480, opened 2023-01-25
Send websocket messages immediately in all cases

LefterisJP:
### Send websocket messages immediately 
This is important as we rely on websockets to show progress and updates, and scheduling a greenlet to send messes with that as we don't know when that greenlet will be scheduled to run.

Made it configurable so we still have the option to send asynchronously but not sure if that should be an option at all

### Add new websocket message to inform frontend for new token detection

While we are processing EVM transactions new tokens may be detected and added to the Database. Some of them can be spam tokens. Using this message we can let the frontend know which tokens are detected. Then they can in turn allow the user to see an aggregated list of all detected tokens and using that list, easily mark spam assets if any.

### DIFF
--- a/docs/websockets.rst
+++ b/docs/websockets.rst
@@ -111,3 +111,16 @@ At the user DB migrations when a new evm chain is introduced rotki will do a mig
 	    "address": "0xFeebabE6b0418eC13b30aAdF129F5DcDd4f70CeA"
 	}]
     }
+
+
+EVM Token Detection
+=======================
+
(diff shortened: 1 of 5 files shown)


LefterisJP:
I will remove the configurability. It makes no sense.

LefterisJP:
Split in two PRs to help with deugging and figure out wtf slowed down the tests.

https://github.com/rotki/rotki/pull/5485
https://github.com/rotki/rotki/pull/5486

PR #5482, opened 2023-01-25
Remove cryptoscamdb and update spam assets

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/assets/spam_assets.py
+++ b/rotkehlchen/assets/spam_assets.py
@@ -1,20 +1,12 @@
 import logging
-from json.decoder import JSONDecodeError
 from typing import TYPE_CHECKING, Any
 
-import requests
-from eth_utils import to_checksum_address
 
 from rotkehlchen.assets.asset import EvmToken
 from rotkehlchen.assets.utils import get_or_create_evm_token
 from rotkehlchen.chain.evm.types import string_to_evm_address
-from rotkehlchen.constants.resolver import ethaddress_to_identifier
-from rotkehlchen.constants.timing import DEFAULT_TIMEOUT_TUPLE
-from rotkehlchen.errors.asset import UnknownAsset
-from rotkehlchen.errors.misc import NotERC20Conformant, RemoteError
-from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.logging import RotkehlchenLogsAdapter
-from rotkehlchen.types import SPAM_PROTOCOL, ChainID, ChecksumEvmAddress
+from rotkehlchen.types import SPAM_PROTOCOL, ChainID
 
 if TYPE_CHECKING:
     from rotkehlchen.db.dbhandler import DBHandler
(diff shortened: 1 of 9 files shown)


rotkehlchen/assets/spam_assets.py
```diff
@@ -757,95 +750,181 @@
...
-                token = EvmToken(ethaddress_to_identifier(checksumed_address))
-            except UnknownAsset:
-                continue
-            if token is not None:
-                tokens_to_ignore.add(token)
-
     # Try to add custom list
     for token_address, info in KNOWN_ETH_SPAM_TOKENS.items():
         try:
             own_token = get_or_create_evm_token(
```
nebolax, line 903: There is no inquirer passed, so there will be no remote calls, so there is no need to handle RemoteError and NotERC20Conformant. Or am I missing something?

rotkehlchen/assets/spam_assets.py
```diff
@@ -757,95 +750,181 @@
...
                 protocol=SPAM_PROTOCOL,
                 decimals=info.get('decimals', 18),
                 name=info.get('name', MISSING_NAME_SPAM_TOKEN),
                 symbol=info.get('symbol', MISSING_SYMBOL_SPAM_TOKEN),
             )
         except (RemoteError, NotERC20Conformant) as e:
-            log.debug(f'Skipping {checksumed_address} due to {str(e)}')
+            log.debug(f'Skipping {token_address} due to {str(e)}')
             continue
         if own_token is not None:
```
nebolax, line 915: `get_or_create_evm_token` never returns None.

rotkehlchen/assets/spam_assets.py
```diff
@@ -757,95 +750,181 @@
...
             tokens_to_ignore.add(own_token)
```
nebolax, line 916: So I would transform this entire loop to something like
```python
for token_address, info in KNOWN_EVM_SPAM_TOKENS.items():
        tokens_to_ignore.add(get_or_create_evm_token(
            userdb=db,
            evm_address=token_address,
            chain_id=info.get('chain', ChainID.ETHEREUM),
            protocol=SPAM_PROTOCOL,
            decimals=info.get('decimals', 18),
            name=info.get('name', MISSING_NAME_SPAM_TOKEN),
            symbol=info.get('symbol', MISSING_SYMBOL_SPAM_TOKEN),
        ))
```

rotkehlchen/assets/spam_assets.py
```diff
@@ -757,95 +750,181 @@
...
-                continue
-            try:
-                token = EvmToken(ethaddress_to_identifier(checksumed_address))
-            except UnknownAsset:
-                continue
-            if token is not None:
-                tokens_to_ignore.add(token)
-
     # Try to add custom list
     for token_address, info in KNOWN_ETH_SPAM_TOKENS.items():
```
nebolax, line 901: Question. What is the reason for having the list of spam tokens as a dict and not just a simple list?

Especially with the addition of other evm chains addresses (which are the keys in this dict) are not guaranteed to be unique anymore, so having it as a pure list sounds better to me.

nebolax (commented):
There is a number of places that also have to be updated

image

rotkehlchen/assets/spam_assets.py
yabirgb: Yes, I believe there shouldn't be any query. The reason for it to be there is because this was not the previous behavior and I didn't update it. Done now

rotkehlchen/assets/spam_assets.py
yabirgb: changed it to a list. The reason was fast query before with the cryptoscamdb but is not needed anymore

nebolax (approved):
LGTM!

LefterisJP (approved):
All right so this looks fine for what it does, left a small comment.

Test passes locally.

**BUT** this approach won't scale. We discussed that these should be fetchable from the remote assets/stuff repo and we should update continuously from there. Keeping a list in the code like this is silly.

@yabirgb Can you write a TODO for this or an issue?

PR #5485, opened 2023-01-25
Send websocket messages immediately in all cases

LefterisJP:
This is important as we rely on websockets to show progress and updates, and scheduling a greenlet for sending a message, messes with that as we don't know when that greenlet will be scheduled to run.


LefterisJP:
All green. Merging.

PR #5486, opened 2023-01-25
Add new websocket message to inform frontend for new token detection

LefterisJP:
While we are processing EVM transactions new tokens may be detected and added to the Database. Some of them can be spam tokens. Using this message we can let the frontend know which tokens are detected. Then they can in turn allow the user to see an aggregated list of all detected tokens and using that list, easily mark spam assets if any.

### DIFF
--- a/docs/websockets.rst
+++ b/docs/websockets.rst
@@ -111,3 +111,16 @@ At the user DB migrations when a new evm chain is introduced rotki will do a mig
 	    "address": "0xFeebabE6b0418eC13b30aAdF129F5DcDd4f70CeA"
 	}]
     }
+
+
+EVM Token Detection
+=======================
+
+While we are processing EVM transactions new tokens may be detected and added to the Database. Some of them can be spam tokens. Using this message we can let the frontend know which tokens are detected. Then they can in turn allow the user to see an aggregated list of all detected tokens and using that list, easily mark spam assets if any.
+
+::
+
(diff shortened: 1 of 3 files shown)


docs/websockets.rst
```diff
@@ -111,3 +111,16 @@ At the user DB migrations when a new evm chain is introduced rotki will do a mig
 	    "address": "0xFeebabE6b0418eC13b30aAdF129F5DcDd4f70CeA"
 	}]
     }
+
+
+EVM Token Detection
+=======================
+
+While we are processing EVM transactions new tokens may be detected and added to the Database. Some of them can be spam tokens. Using this message we can let the frontend know which tokens are detected. Then they can in turn allow the user to see an aggregated list of all detected tokens and using that list, easily mark spam assets if any.
```
yabirgb, line 119:
```suggestion
While we are processing EVM transactions new tokens may be detected and added to the database. Some of them can be spam tokens. Using this message we can let the frontend know which tokens are detected. Then they can in turn allow the user to see an aggregated list of all detected tokens and using that list, easily mark spam assets if any.
```

yabirgb (approved):
Just a small thing in the docs that can be changed in other PR

docs/websockets.rst
LefterisJP: why do you need to capitalize this?

PR #5493, opened 2023-01-26
feat: add new detected token view

lukicenturi:
Closes #(issue_number)

## Checklist

- [x] Receive the messages from the websocket (that a new asset was detected)
- [x]  Aggregate the list of the detected assets (see if we can also persist them in local storage for a period of time/number of entries)
- [x] Group notifications for this, and add actions there
- [x] New page under asset manager where you can go by clicking the action button or navigate normally
- [x] The new page has a list of the aggregated assets where you can select which of these assets you want to ignore.
- [x] Actions in that page

- [x] Fix download zip error (found by isaac)
- [x] Show currency on currency button tooltip

### DIFF
--- a/frontend/app/src/components/asset-manager/NewlyDetectedAssetTable.vue
+++ b/frontend/app/src/components/asset-manager/NewlyDetectedAssetTable.vue
@@ -1,0 +1,167 @@
+<script setup lang="ts">
+import { type DataTableHeader } from 'vuetify';
+import { type ComputedRef, type Ref } from 'vue';
+import { useNewlyDetectedTokens } from '@/composables/assets/newly-detected-tokens';
+import { type NewDetectedToken } from '@/types/websocket-messages';
+import { useMessageStore } from '@/store/message';
+import { useIgnoredAssetsStore } from '@/store/assets/ignored';
+import { uniqueStrings } from '@/utils/data';
+import { getAddressFromEvmIdentifier } from '@/utils/assets';
+
+const { tc } = useI18n();
+
+const { tokens, removeNewDetectedTokens } = useNewlyDetectedTokens();
+
+const mappedTokens: ComputedRef<NewDetectedToken[]> = computed(() => {
+  return get(tokens).map(tokenIdentifier => ({
+    tokenIdentifier,
+    address: getAddressFromEvmIdentifier(tokenIdentifier)
+  }));
+});
+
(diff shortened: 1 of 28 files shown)


kelsos (commented):
Can we make the title/text adjust to plural singular?

image

I would avoid elevation. Is it possible to have the select/deselect actions side by side? What do you think?
image

I am not sure about the clarity of the icon buttons here.
image

Also, do you think we can infer the address for the tokens from the identifier and also display it in the table?

kelsos:
I am thinking aloud a bit here, but this notification would be more useful if it actually got displayed to the user, however, the display should not spam the user so we would want a delayed display. and we don't have such a mechanism in place.

What I mean by delayed is that for example, we could configure a notification to only display if there are no updates in the group for `x` seconds

lukicenturi:
> I am thinking aloud a bit here, but this notification would be more useful if it actually got displayed to the user, however, the display should not spam the user so we would want a delayed display. and we don't have such a mechanism in place.
> 
> What I mean by delayed is that for example, we could configure a notification to only display if there are no updates in the group for `x` seconds

I make the notification displayed like usual, but for the delayed notification, can we do it in next release, I think it will require rearranging the notification store.

lukicenturi:
> Can we make the title/text adjust to plural singular?
> 
> image
> 
> I would avoid elevation. Is it possible to have the select/deselect actions side by side? What do you think? image
> 
> I am not sure about the clarity of the icon buttons here. image
> 
> Also, do you think we can infer the address for the tokens from the identifier and also display it in the table?

Updated, except for the action icons.
Is the tooltip on the action icons helpful enough?

kelsos:
@lukicenturi can you have a look to the changes I did, I kept them in separate commits for ease.

frontend/app/src/components/asset-manager/NewlyDetectedAssetTable.vue
```diff
@@ -160,7 +160,7 @@ const ignoreTokens = async () => {
       </template>
 
       <template #item.address="{ item }">
-        <hash-link :text="item.address" />
+        <hash-link :text="item.address" no-link />
```
kelsos, line 163: I made it no-link for starters because in order for the link to work fine we need to also get the chain, which is not impossible but it would be a bit more complicated.

kelsos (approved):
lgtm thank you :)

lukicenturi:
LGTM too, thanks 👍

PR #5495, opened 2023-01-26
Make sure curve pools are queried at decoding time

LefterisJP:
This is a start of generalizing the "Decoder that needs some external data to exist at decoding time" approach.

- Made a new mixin for Decoders that require this and made sure that at start of transaction decoding whatever needs to be queried per decoder is queried.
- Split out the curve protocol cache query from EthereumManager to EthereumNodeInquirer since this is the only thing that is actually used to query the chain.
- Made sure that when querying from the decoder the EthereumNodeinquirer version is used, but if used from outside the `EthereumManager` is used which also encapsulates an update to the decoder.

### DIFF
--- a/rotkehlchen/chain/ethereum/node_inquirer.py
+++ b/rotkehlchen/chain/ethereum/node_inquirer.py
@@ -13,22 +13,35 @@ from web3 import Web3
 from rotkehlchen.chain.constants import DEFAULT_EVM_RPC_TIMEOUT
 from rotkehlchen.chain.ethereum.constants import ETHEREUM_ETHERSCAN_NODE
 from rotkehlchen.chain.ethereum.graph import Graph
+from rotkehlchen.chain.ethereum.modules.curve.pools_cache import (
+    clear_curve_pools_cache,
+    update_curve_metapools_cache,
+    update_curve_registry_pools_cache,
+)
 from rotkehlchen.chain.evm.contracts import EvmContracts
 from rotkehlchen.chain.evm.node_inquirer import WEB3_LOGQUERY_BLOCK_RANGE, EvmNodeInquirer
 from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.errors.misc import InputError, RemoteError, UnableToDecryptRemoteData
 from rotkehlchen.errors.serialization import DeserializationError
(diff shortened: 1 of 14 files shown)


rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -64,12 +64,6 @@ def post_decoding_rules(self) -> list[tuple[int, Callable]]:
         """
         return []
 
-    def reload(self) -> Mapping[ChecksumEvmAddress, tuple[Any, ...]]:  # pylint: disable=no-self-use  # noqa: E501
```
LefterisJP, line 67: renamed this from `reload()` as `reload()` is a reserved keyword in python.

rotkehlchen/chain/ethereum/node_inquirer.py
```diff
@@ -223,6 +237,36 @@ def _ens_lookup(
...
+        1. Deletes all previous cache values
+        2. Queries information about curve pools' addresses, lp tokens and used coins
+        3. Saves queried information in the cache in globaldb
+        """
+        if should_update_protocol_cache(GeneralCacheType.CURVE_LP_TOKENS) is False:
+            return False
+
+        curve_address_provider = self.contracts.contract('CURVE_ADDRESS_PROVIDER')
+        # Using savepoint to not end up having partially populated cache
+        with GlobalDBHandler().conn.savepoint_ctx() as savepoint_cursor:
```
nebolax, line 255: I know it is not related to this PR's changes, but:

Currently this savepoint will stay open for a long time (because there are remote queries).

Both `update_curve_registry_pools_cache` and `update_curve_metapools_cache` write to the db only at the very end.

So maybe change this to:
1. Query everything
2. Quickly clear the cache and write new pools

LefterisJP: Thanks for the suggestion. Let me check.

LefterisJP:
All is green. Will merge.

rotkehlchen/chain/ethereum/node_inquirer.py
LefterisJP: here: https://github.com/rotki/rotki/pull/5498

PR #5498, opened 2023-01-26
Separate curve pool cache query from DB writing

LefterisJP:
Move the querying outside of the write context. And gather the pool mappings together. Thanks to this, savepoints don't even need to be used as the DB writing is bundled together.


nebolax (approved):
LGTM!

PR #5504, opened 2023-01-27
feat: improve newly detected tokens

lukicenturi:
Closes #(issue_number)

## Checklist

- [x] Add etherscan link to address.
- [x] Add action button for each entries.
- [x] "Remove from list" -> "Accept selected tokens"
- [x] Show tx, which the token detected from

### DIFF
--- a/frontend/app/src/components/asset-manager/NewlyDetectedAssetTable.vue
+++ b/frontend/app/src/components/asset-manager/NewlyDetectedAssetTable.vue
@@ -1,22 +1,33 @@
 <script setup lang="ts">
 import { type DataTableHeader } from 'vuetify';
 import { type ComputedRef, type Ref } from 'vue';
+import { Blockchain } from '@rotki/common/lib/blockchain';
 import { useNewlyDetectedTokens } from '@/composables/assets/newly-detected-tokens';
 import { type NewDetectedToken } from '@/types/websocket-messages';
 import { useMessageStore } from '@/store/message';
 import { useIgnoredAssetsStore } from '@/store/assets/ignored';
 import { uniqueStrings } from '@/utils/data';
 import { getAddressFromEvmIdentifier } from '@/utils/assets';
+import { useSupportedChains } from '@/composables/info/chains';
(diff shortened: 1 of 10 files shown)


yabirgb:
Was checking the PR. Does it make sense to have the image generated by the tx hash? i believe we don't do it in other places 
image

lukicenturi:
> Was checking the PR. Does it make sense to have the image generated by the tx hash? i believe we don't do it in other places image

my bad, updated

yabirgb:
Another thing about UX that I noticed is that if you want to click on an accept/ignore and a new token appears the table gets updated and the entry moves. You might end up ignoring the wrong asset. Could we maybe avoid this? Not sure how difficult it is

kelsos:
> > Another thing about UX that I noticed is that if you want to click on an accept/ignore and a new token appears the table gets updated and the entry moves. You might end up ignoring the wrong asset. Could we maybe avoid this? Not sure how difficult it is
> 
> Hmm yeah I see what you mean, let's see what I can do about it

Maybe we can disable sorting completely and keep the default table default array order?

lukicenturi:
> > > Another thing about UX that I noticed is that if you want to click on an accept/ignore and a new token appears the table gets updated and the entry moves. You might end up ignoring the wrong asset. Could we maybe avoid this? Not sure how difficult it is
> > 
> > 
> > Hmm yeah I see what you mean, let's see what I can do about it
> 
> Maybe we can disable sorting completely and keep the default table default array order?

That happens because I push the new data to the head of the array.
We can move it to the end, but then if it pass 500 cache_size, the same thing still happens

yabirgb:
> That happens because I push the new data to the head of the array. We can move it to the end, but then if it pass 500 cache_size, the same thing still happens

if 500 cache size is 500 rows that is enough. How it happened to me is probably an edge case but ~20 new tokens can appear with a background task and that can move enough the table to misclick

kelsos (approved):
The code looks ok @yabirgb can you check the functionality. If function is ok merge.

PR #5505, opened 2023-01-27
Add information about where new tokens are first seen

LefterisJP:
Add new argument in `get_or_create_evm_token` which allows specifying where a token newly added to the DB was first seen.

The value can either be missing or it can be a transaction hash or a string description of an action.

This is used in the websocket messages sent to notify the frontend.

### DIFF
--- a/rotkehlchen/assets/utils.py
+++ b/rotkehlchen/assets/utils.py
@@ -1,5 +1,5 @@
 import logging
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, NamedTuple, Optional
 
 from rotkehlchen.api.websockets.typedefs import WSMessageType
 from rotkehlchen.assets.asset import Asset, AssetWithOracles, EvmToken, UnderlyingToken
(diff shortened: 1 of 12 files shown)


rotkehlchen/assets/utils.py
```diff
@@ -202,9 +211,16 @@ def get_or_create_evm_token(
...
                 GlobalDBHandler().edit_evm_token(ethereum_token)
             else:
-                userdb.msg_aggregator.add_message(  # inform frontend new token detected
+                # inform frontend new token detected
+                data = {'token_identifier': identifier}
+                if seen is not None:
+                    if seen.tx_hash is not None:
+                        data['seen_tx_hash'] = '0x' + seen.tx_hash.hex()
+                    else:  # description should have been given
+                        data['seen_description'] = seen.description  # type: ignore
```
yabirgb, line 220: Did we say that we will explain why the type ignore here? I guess that is because data is a `dict[str, str]` instead of `dict[str, Optional[str]]`

rotkehlchen/chain/ethereum/modules/balancer/utils.py
```diff
@@ -243,6 +246,7 @@ def deserialize_pool_share(
         protocol='balancer',
         decimals=18,  # All BPT tokens have 18 decimals
         underlying_tokens=pool_tokens,
+        seen=TokenSeenAt(description='Balancer pools query'),
```
yabirgb, line 249: For consistency with the other descriptions should it be

```suggestion
        seen=TokenSeenAt(description='Querying balancer pools'),
```

?

yabirgb (commented):
Two small comments

rotkehlchen/assets/utils.py
LefterisJP: It's the comment right above.

rotkehlchen/chain/ethereum/modules/balancer/utils.py
LefterisJP: Will see in a subsequent PR if it makes sense!

PR #5506, opened 2023-01-27
Adds session persistence for specific message categories

kelsos:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/address-book-manager/AddressBookTable.vue
+++ b/frontend/app/src/components/address-book-manager/AddressBookTable.vue
@@ -1,5 +1,9 @@
 <script setup lang="ts">
-import { type NotificationPayload, Severity } from '@rotki/common/lib/messages';
+import {
+  NotificationCategory,
+  type NotificationPayload,
+  Severity
+} from '@rotki/common/lib/messages';
 import { type Ref } from 'vue';
 import { type DataTableHeader } from 'vuetify';
 import { type Blockchain } from '@rotki/common/lib/blockchain';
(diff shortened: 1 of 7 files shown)


LefterisJP (commented):
Unfortunately this did not work.

When pressing "confirm" DB assets upgrade I don't see the evm address upgrade message anymore

LefterisJP (commented):
So for me:

1. I still do not get a popup for this as you said it should appear.
2. The notification now **does** persist but at least at my first try was duplicated (see screenshot)

At the 2nd try it was not. 

My assumption for that is that migration started happening but did not commit to DB before restarting?
3. I am not sure if it's related but also got this:
```
assertions.ts:10 Uncaught AssertionError: No tooltip element found
    at assert (assertions.ts:10:11)
    at Tooltip.external (NetWorthChart.vue:304:5)
    at Tooltip.update (chart.mjs:8789:24)
    at Tooltip.handleEvent (chart.mjs:9127:14)
    at Object.afterEvent (chart.mjs:9188:25)
    at callback (helpers.segment.mjs:47:15)
    at PluginService._notify (chart.mjs:4988:11)
    at PluginService.notify (chart.mjs:4975:25)
    at Chart.notifyPlugins (chart.mjs:6124:26)
    at Chart._eventHandler (chart.mjs:6151:10)
```

I think (3) is not related since I posted it earlier today too in develop.

LefterisJP (commented):
One more thing I noticed which is really weird about the notification of evm address migration appearing only as a notification in the drawer and not as a popup notification.

When the "new tokens found" alert comes and I get a popup notification, then I actually also get the old evm address migration appearing again as a popup notification that you described. So the "new tokens found" thing seems to trigger and show the evm address migration popup somehow

kelsos:
OK so I will merge and then we can figure out if this also happens to others, because I have not been able to replicate the notification not showing locally.

Maybe we can try and debug it together.

@yabirgb can you also try it and see if you get the addresses migrated message?

PR #5507, opened 2023-01-27
Allow to run tests in matrix for backend

yabirgb:
Create a matrix job for the backend to speed up CI execution. Due to limits in the req/min in the api keys I've added a random function to pick them


LefterisJP (approved):
The result is nice but shouldn't we split it in api+integration in one and then unit+db tests in another? Is there more categories?

I would like to think of the groups of tests a bit more so we also group the tests properly when developing But I don't see anything wrong with merging right now and improving later.

PR #5508, opened 2023-01-28
Add frax decoder

anukul:
Only decodes ADD_COLLATERAL event


anukul:
Can't seem to figure out why tests are failing 🤔 

Will keep this code in the source branch if someone wants to pick it up

PR #5510, opened 2023-01-29
Do not perform any migrations for a fresh user DB

LefterisJP:
In the past we had a new DB run all data migrations. This was not
needed. The only reason it was being done was to do two things:

1. Spam assets addition (Migration: 3)
2. Read and write rpc nodes to the DB (Migrations: 4, 6, 7)

This commit starts a fresh DB at latest migration, so no migrations
run for a fresh DB. Instead those actions are now set apart as actions
to perform to populate a fresh DB.

Tests and their fixtures/mocks are also adjusted properly.

What's more the `perform_migrations_at_unlock` test fixture is removed, since now a fresh user DB does not go through migrations this is not needed

### DIFF
--- a/docs/websockets.rst
+++ b/docs/websockets.rst
@@ -80,20 +80,20 @@ The messages sent by rotki when a user is logging in and a db upgrade is happeni
 
     {
         "type": "db_upgrade_status",
-	"data": {
-	    "start_version": 26,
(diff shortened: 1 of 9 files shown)


LefterisJP:
All green. Ze merge is now.

PR #5511, opened 2023-01-29
Remove post/api/1/assets/ignore api endpoint

LefterisJP:
Since we removed pulling of ignored assets from cryptoscamdb this endpoint is no longer needed.


LefterisJP:
@rotki/frontend this endpoint is no longer needed, right? Do you use it in the frontend? If so I think you should also remove it since we don't pull anything from cryptoscamdb any longer.

It may become needed again when/if we move spam assets to our own updatable repo like we have the assets at the moment.

PR #5512, opened 2023-01-29
Let matrix jobs finish if one of them fails

yabirgb:
Disable fast fail in github CI matrix job for backend tests

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast


kelsos (approved):
lgtm

PR #5514, opened 2023-01-30
Bump linting packages - Lots of ruff work and rules enabling

LefterisJP:
Check commits for details


LefterisJP:
`rotkehlchen/tests/api/blockchain/test_optimism.py::test_add_optimism_blockchain_account` should pass again now as it passes locally. It's due to the test not mocking and me having moved tokens in the address. Tokens are back to what the test expects so all good :rofl:

PR #5515, opened 2023-01-30
Replace get or create evm token for a simpler version in spam tokens

yabirgb:
This PR:

- changes the complicated logic of `get_or_create_evm_token` for somethig lighter 
- Mocks the calls for the optimism token detection test
- Adds for the last time as a hardcoded step the sync of spam assets

### DIFF
--- a/rotkehlchen/assets/spam_assets.py
+++ b/rotkehlchen/assets/spam_assets.py
@@ -1059,23 +1059,19 @@ def query_token_spam_list() -> set[Asset]:
             protocol=SPAM_PROTOCOL,
             underlying_tokens=None,
         )
-        token_exists = True
+
         try:
             evm_token.check_existence()
-        except UnknownAsset:
-            token_exists = False
-
-        if token_exists is True:
-            # make sure that the token has the spam protocol
-            db_evm_token = EvmToken(evm_token.identifier)
-            if db_evm_token.protocol != SPAM_PROTOCOL:
-                GlobalDBHandler().edit_evm_token(entry=evm_token)
-        else:
+        except UnknownAsset:  # token does not exist
             GlobalDBHandler().add_asset(
                 asset_id=evm_token.identifier,
                 asset_type=AssetType.EVM_TOKEN,
                 data=evm_token,
             )
+        else:  # token exists, make sure it has spam protocol set
+            db_evm_token = EvmToken(evm_token.identifier)
+            if db_evm_token.protocol != SPAM_PROTOCOL:
+                GlobalDBHandler().edit_evm_token(entry=evm_token)
         # save the asset instead of the EvmToken as we don't need all the extra information later
         tokens_to_ignore.add(Asset(evm_token.identifier))
 
--- a/rotkehlchen/assets/spam_assets.py
+++ b/rotkehlchen/assets/spam_assets.py
@@ -2,11 +2,13 @@ import logging
 from typing import TYPE_CHECKING, Any
 
 
-from rotkehlchen.assets.asset import EvmToken
-from rotkehlchen.assets.utils import get_or_create_evm_token
+from rotkehlchen.assets.asset import Asset, EvmToken
+from rotkehlchen.assets.types import AssetType
 from rotkehlchen.chain.evm.types import string_to_evm_address
+from rotkehlchen.errors.asset import UnknownAsset
+from rotkehlchen.globaldb.handler import GlobalDBHandler
 from rotkehlchen.logging import RotkehlchenLogsAdapter
-from rotkehlchen.types import SPAM_PROTOCOL, ChainID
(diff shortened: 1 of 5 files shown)


rotkehlchen/assets/spam_assets.py
```diff
@@ -1039,24 +1041,43 @@
...
+            if db_evm_token.protocol != SPAM_PROTOCOL:
+                GlobalDBHandler().edit_evm_token(entry=evm_token)
+        else:
+            GlobalDBHandler().add_asset(
+                asset_id=evm_token.identifier,
+                asset_type=AssetType.EVM_TOKEN,
+                data=evm_token,
+            )
+        # save the asset instead of the EvmToken as we don't need all the extra information later
+        tokens_to_ignore.add(Asset(evm_token.identifier))
```
LefterisJP, line 1080: shorter and no need to use extra variable

```suggestion
        try:
            evm_token.check_existence()
        except UnknownAsset:  # token does not exist
            GlobalDBHandler().add_asset(
                asset_id=evm_token.identifier,
                asset_type=AssetType.EVM_TOKEN,
                data=evm_token,
            )
        else:  # token exists, make sure it has spam protocol set
            db_evm_token = EvmToken(evm_token.identifier)
            if db_evm_token.protocol != SPAM_PROTOCOL:
                GlobalDBHandler().edit_evm_token(entry=evm_token)
        # save the asset instead of the EvmToken as we don't need all the extra information later
        tokens_to_ignore.add(Asset(evm_token.identifier))
```

rotkehlchen/data_migrations/migrations/migration_8.py
```diff
@@ -66,4 +67,6 @@ def data_migration_8(rotki: 'Rotkehlchen', progress_handler: 'MigrationProgressH
                 data=[{'evm_chain': str(x[0]), 'address': x[1]} for x in to_add_accounts],
             )
 
+    # Also update the spam assets
+    update_spam_assets(db=rotki.data.db)
```
LefterisJP, line 71: Add a step in the migration for this. Also add +1 in line 31

rotkehlchen/tests/api/blockchain/test_optimism.py
```diff
@@ -19,13 +19,11 @@
 TEST_ADDY = '0x9531C059098e3d194fF87FebB587aB07B30B1306'
 
 
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
```
LefterisJP, line 22: Did you mock this as I asked with the old ones? So basically did you run record after: `2023-01-30 11:14:00`?

rotkehlchen/tests/data_migrations/test_migrations.py
```diff
@@ -517,3 +524,9 @@ def assert_progress_message(msg, step_num, description) -> None:
             assert_progress_message(msg, i, None)
         else:
             assert_progress_message(msg, i, None)
+
+    # check that the spam assets have been updated
```
LefterisJP, line 528: you would need to add a tiny check for 1 more websocket message here once you do what I asked in the data migration

rotkehlchen/tests/api/blockchain/test_optimism.py
yabirgb: yes, it would fail If I didn't record the previous state

rotkehlchen/assets/spam_assets.py
yabirgb: I wanted to avoid the error to control the flow but okey, will change

rotkehlchen/tests/api/blockchain/test_optimism.py
LefterisJP: ah yes ofc. Since the test has no changes inside it at all.

rotkehlchen/assets/spam_assets.py
LefterisJP: there is no change in the flow. Same thing happens.

rotkehlchen/data_migrations/migrations/migration_8.py
```diff
@@ -65,5 +67,11 @@ def data_migration_8(rotki: 'Rotkehlchen', progress_handler: 'MigrationProgressH
                 message_type=WSMessageType.EVM_ADDRESS_MIGRATION,
                 data=[{'evm_chain': str(x[0]), 'address': x[1]} for x in to_add_accounts],
             )
+    else:
+        # The only step is update the spam assets
```
LefterisJP, line 71: nitpick to not waste space

```suggestion
    else:  # The only step is updating the spam assets
```

LefterisJP (approved):
LGTM. Left a tiny nitpick for next PR

PR #5518, opened 2023-01-30
Include optimism in the list of hidden imports

yabirgb:
Add optimism to the list of hidden imports


LefterisJP (approved):
LGTM. Thanks @yabirgb

PR #5519, opened 2023-01-30
fix: several things in frontend

lukicenturi:
Closes #(issue_number)

## Checklist

- [x] Adjust db upgrade & data migration messages
- [x] Address migration will wait until user are really logged in

### DIFF
--- a/frontend/app/src/components/status/notifications/NotificationPopup.vue
+++ b/frontend/app/src/components/status/notifications/NotificationPopup.vue
@@ -20,22 +20,24 @@ const dismissAll = async () => {
   set(visibleNotification, { ...get(visibleNotification), display: false });
 };
 
-watch(
-  queue,
-  () => {
-    const data = [...get(queue)];
-    if (!get(visibleNotification).display && data.length > 0) {
-      const next = data.shift();
-      if (!next) {
-        return;
-      }
-      nextTick(() => {
-        set(visibleNotification, next);
-      });
+const checkQueue = () => {
+  const data = [...get(queue)];
+  if (!get(visibleNotification).display && data.length > 0) {
+    const next = data.shift();
+    if (!next) {
+      return;
     }
-  },
-  { deep: true }
-);
+    nextTick(() => {
+      set(visibleNotification, next);
+    });
+  }
+};
+
+watch(queue, checkQueue, { deep: true });
+
+onMounted(() => {
+  checkQueue();
+});
 
 const { t } = useI18n();
 </script>
--- a/frontend/app/src/locales/en.json
+++ b/frontend/app/src/locales/en.json
@@ -1003,10 +1003,10 @@
     "newly_detected": {
       "accept_selected": "Accept selected tokens",
       "ignore_selected": "Ignore selected tokens",
+      "seen_during": "Seen during",
       "select_deselect_all_tokens": "Select/deselect all tokens",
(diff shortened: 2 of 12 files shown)


kelsos:
So what I see when I check out and run locally is the following

image

At least the second error is because of `notify` which for some reason is undefined

frontend/app/src/store/blockchain/accounts/migrate.ts
```diff
@@ -0,0 +1,99 @@
...
+    watch(logged, logged => {
+      checkLoggedStatus(logged);
+    });
+
+    return {
+      migratedAddresses,
+      upgradeMigratedAddresses
+    };
+  }
+);
```
kelsos, line 99: missing the hot-reload part at the end of the store.

frontend/app/src/store/blockchain/accounts/migrate.ts
```diff
@@ -0,0 +1,99 @@
...
+  type Notification,
+  NotificationCategory,
+  Severity
+} from '@rotki/common/lib/messages';
+import { type MigratedAddresses } from '@/types/websocket-messages';
+import { useSessionAuthStore } from '@/store/session/auth';
+import { startPromise } from '@/utils';
+import { useNotificationsStore } from '@/store/notifications';
+
+export const useAccountMigration = defineStore(
```
kelsos, line 12: can we rename it to `useAccountMigrationStore`?

I have a template I created in WebStorm to generate the template for a store.
It still requires some manual interaction in nested ones but I can share it if you want.

LefterisJP:
> So what I see when I check out and run locally is the following
> 
> image
> 
> At least the second error is because of `notify` which for some reason is undefined

I also see this.

LefterisJP (commented):
Another thing I see is 



As I said in discord it should say "Data migration 8 is running". Not 8 - 8.

frontend/app/src/locales/en.json
```diff
@@ -1003,10 +1003,10 @@
     "newly_detected": {
       "accept_selected": "Accept selected tokens",
       "ignore_selected": "Ignore selected tokens",
+      "seen_during_or_tx_origin": "Seen during / Tx origin",
```
LefterisJP, line 1006:
```suggestion
      "seen_during": "Seen during",
```

LefterisJP (commented):
This should be a simple "Seen during" as per the review. "Seen during" easily applies also to tx_hashes

frontend/app/src/store/blockchain/accounts/migrate.ts
lukicenturi: sure I can try the template

lukicenturi:
> Another thing I see is
> 
> 
> 
> As I said in discord it should say "Data migration 8 is running". Not 8 - 8.

Updated

frontend/app/src/components/status/notifications/NotificationPopup.vue
```diff
@@ -20,22 +20,24 @@ const dismissAll = async () => {
...
+      set(visibleNotification, next);
+    });
+  }
+};
+
+watch(queue, checkQueue, { deep: true });
+
+onMounted(() => {
+  checkQueue();
+});
```
lukicenturi, line 40: This should fix the issue where the popup of notification for the evm address migration, appear late.

PR #5520, opened 2023-01-30
Moves asset update flow directly after login

kelsos:
Closes #(issue_number)

## Checklist

## TODO

- [x] Check if the settings component still works
- [x] Check if it works for a non-premium account
- [x] Check if skip still still works as expected


LefterisJP (approved):
LGTM. Functionality looks okay to me. Left a comment also in discord: ` to find a way to clean this up (localstorage) between upgrades?`

But other than that all good.

kelsos:
> LGTM. Functionality looks okay to me. Left a comment also in discord: ` to find a way to clean this up (localstorage) between upgrades?`
> 
> But other than that all good.

I need to think about it a bit. Probably it would make sense to also include at least the major.minor version combination in the key, and on start check and cleanup the rest

PR #5522, opened 2023-01-31
Don't query chains that address is not added to

nebolax:
Closes #(issue_number)

## Checklist

- [ ] The PR modified the frontend, and updated the [user guide](https://github.com/rotki/rotki/blob/develop/docs/usage_guide.rst) to reflect the changes.

PR #5529, opened 2023-01-31
Kill evm tx greenlets on account deletion

nebolax:
This PR makes sure that all the greenlets are killed and locks are acquired when deleting an EVM account.

The main bug that caused me spent so much time is the following scenario:

1. An EVM address that has transactions in both ethereum and optimism is added.
2. Frontend queries transactions for the ethereum address. It does this by calling `/evm/transactions` API endpoint with `accounts` being a list with a single item `{'address': 'the-address', 'evm_chain': 'ethereum'}` and `evm_chain` (another standalone argument) being `null`.
3. Same happens for the optimism address.

Since `evm_chain` is `None` in both API calls, this causes the backend to [iterate](https://github.com/rotki/rotki/blob/b725daf8cc2357b690cb3e67ef1fbdc8c85fb98b/rotkehlchen/api/rest.py#L3290) through all supported evm chains and ask them for transactions. Before #5522 this caused that for both of the API calls, the transactions would be asked for both ethereum and optimism. So there were duplicated greenlets for both ethereum and optimism transactions.

These duplicated greenlets would not run simultaneously due to the locks existence, but they would queue one after another and later, when the user would try to delete a transaction, one of the greenlets would be killed and the other one would immediately re-acquire the lock and block the account deletion.

Since #5522 is merged, this is no longer an issue and the greenlets are killed fine. However, I have question. @LefterisJP why not to change the evm transactions endpoint so that it accepts only a single chain argument? I mean, currently there are 2 arguments. `accounts` which is a list of (`address`, `evm_chain`) and another, standalone argument `evm_chain` which caused these discrepancies. Can we maybe change it so that `accounts` becomes just a list of addresses and the chain is determined only by a single `evm_chain` argument?

### DIFF
--- a/rotkehlchen/rotkehlchen.py
+++ b/rotkehlchen/rotkehlchen.py
@@ -11,6 +11,7 @@ from types import MethodType
 from typing import TYPE_CHECKING, Any, DefaultDict, Literal, Optional, Union, cast, overload
 
 import gevent
+from gevent.lock import Semaphore
 
 from rotkehlchen.accounting.accountant import Accountant
 from rotkehlchen.accounting.structures.balance import Balance, BalanceType
@@ -31,6 +32,7 @@ from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
 from rotkehlchen.chain.ethereum.oracles.saddle import SaddleOracle
 from rotkehlchen.chain.ethereum.oracles.uniswap import UniswapV2Oracle, UniswapV3Oracle
 from rotkehlchen.chain.evm.accounting.aggregator import EVMAccountingAggregators
+from rotkehlchen.chain.evm.manager import EvmManager
 from rotkehlchen.chain.evm.nodes import populate_rpc_nodes_in_database
 from rotkehlchen.chain.optimism.accountant import OptimismAccountingAggregator
 from rotkehlchen.chain.optimism.manager import OptimismManager
@@ -49,7 +51,13 @@ from rotkehlchen.db.filtering import NFTFilterQuery
 from rotkehlchen.db.settings import DBSettings, ModifiableDBSettings
 from rotkehlchen.errors.api import PremiumAuthenticationError
 from rotkehlchen.errors.asset import UnknownAsset
-from rotkehlchen.errors.misc import EthSyncError, InputError, RemoteError, SystemPermissionError
+from rotkehlchen.errors.misc import (
+    EthSyncError,
+    GreenletKilledError,
+    InputError,
+    RemoteError,
+    SystemPermissionError,
+)
 from rotkehlchen.exchanges.manager import ExchangeManager
 from rotkehlchen.externalapis.beaconchain import BeaconChain
 from rotkehlchen.externalapis.coingecko import Coingecko
@@ -104,6 +112,20 @@ MAIN_LOOP_SECS_DELAY = 10
 
 ICONS_BATCH_SIZE = 3
 ICONS_QUERY_SLEEP = 60
+
+
+def kill_greenlet_and_acquire_locks(
+        stack: contextlib.ExitStack,
+        acquired_by_this_method: dict[Semaphore, bool],
+        greenlet: gevent.Greenlet,
+        locks: list[Semaphore],
+) -> None:
+    """Kills the provided greenlet and tries to acquire the locks."""
+    greenlet.kill(exception=GreenletKilledError('Killed to delete evm addresses'))
+    for lock in locks:
+        if lock.locked() is False:
+            stack.enter_context(lock)
+            acquired_by_this_method[lock] = True
 
 
 class Rotkehlchen():
@@ -173,8 +195,10 @@ class Rotkehlchen():
         self.shutdown_event = gevent.event.Event()
         self.migration_manager = DataMigrationManager(self)
 
-    def maybe_kill_running_tx_query_tasks(
+    def acquire_all_evm_tx_locks(
             self,
+            stack: contextlib.ExitStack,
+            evm_manager: 'EvmManager',
             blockchain: SupportedBlockchain,
             addresses: list[ChecksumEvmAddress],
     ) -> None:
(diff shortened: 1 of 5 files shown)


LefterisJP:
> I have question. @LefterisJP why not to change the evm transactions endpoint so that it accepts only a single chain argument? I mean, currently there are 2 arguments. `accounts` which is a list of (`address`, `evm_chain`) and another, standalone argument `evm_chain` which caused these discrepancies. Can we maybe change it so that `accounts` becomes just a list of addresses and the chain is determined only by a single `evm_chain` argument?


No, because those are different filters. The (address, chain) filter is just the address filter, with the proper chain prefix now that we are multichain. The chain_id filter is for all transactions of that chain.

> Can we maybe change it so that `accounts` becomes just a list of addresses and the chain is determined only by a single `evm_chain` argument?

This would not let you do many of the filters a user needs. Such as give me lefteris.eth in ethereum mainnet and lefteris.eth in optimism and in avalanche.

rotkehlchen/rotkehlchen.py
```diff
@@ -173,35 +195,108 @@ def __init__(self, args: argparse.Namespace) -> None:
...
+                address = greenlet.kwargs.get('address')
+                if address is None or address in addresses:
+                    kill_greenlet_and_acquire_locks(
+                        stack=stack,
+                        acquired_by_this_method=acquired_by_this_method,
+                        greenlet=greenlet,
+                        locks=[evm_manager.transactions.address_tx_locks[address]],
+                    )
+
+            for greenlet in self.task_manager.running_greenlets.get(self.task_manager._maybe_decode_evm_transactions, []):  # noqa: E501
```
LefterisJP, line 271: This is not needed. 

1. Why is it that important to kill a transaction decodign task?
2. This won't kill the transaction decoding task.Since this will kill just the scheduling event. This is where the scheduling event will actually spawn the decoding task: https://github.com/rotki/rotki/blob/97ec324c0216e55cb8aa1a69d0610ff1030a4d62/rotkehlchen/tasks/manager.py#L476

rotkehlchen/rotkehlchen.py
```diff
@@ -173,35 +195,108 @@ def __init__(self, args: argparse.Namespace) -> None:
...
+                        acquired_by_this_method=acquired_by_this_method,
+                        greenlet=greenlet,
+                        locks=[evm_manager.transactions.missing_receipts_lock],
+                    )
 
-            tx_query_task_greenlets = self.task_manager.running_greenlets.get(self.task_manager._maybe_query_evm_transactions, [])  # noqa: E501
-            for greenlet in tx_query_task_greenlets:
-                if greenlet.kwargs['address'] in addresses:
-                    greenlet.kill()
+            for greenlet in self.task_manager.running_greenlets.get(self.task_manager._maybe_schedule_evm_txreceipts, []):  # noqa: E501
```
LefterisJP, line 281: Same comment as above. You are trying to kill the scheduling event here and not the actual greenlet that will be scheduled

rotkehlchen/rotkehlchen.py
```diff
@@ -173,35 +195,108 @@ def __init__(self, args: argparse.Namespace) -> None:
...
+                    kill_greenlet_and_acquire_locks(
+                        stack=stack,
+                        acquired_by_this_method=acquired_by_this_method,
+                        greenlet=greenlet,
+                        locks=[evm_manager.transactions_decoder.undecoded_tx_query_lock],
+                    )
+
+        if all([acquired_by_this_method[lock] for lock in all_locks]) is True:
+            log.debug('Acquired all locks for killing running tx query tasks')
+        else:
```
LefterisJP, line 293: I don't understand the logic here at all. Why wait until the end to wait on acquiring the lock? What's the point of doing it this way? Just to log something if you manage to acquire all locks immediately? I don't see the point. Or the point of the existence of `acquired_by_this_method`?

LefterisJP (commented):
Left quite a few comments.

I will be honest I don't like the approach you took in `maybe_kill_running_tx_query_tasks` . It's become too complicated and bug-prone and still does not cover all the cases.

You mentioned that 
>Since https://github.com/rotki/rotki/pull/5522 is merged, this is no longer an issue

In that case if this is not solving any actual issue I would be much more wary and careful in any refactoring here. The current approach is definitely not the way.

 To proceed. Split this in two.

Take the killing greenlet and and handling of the task with `None` and empty message out. Also add to it the nitpick I asked in your last PR.

And for the locks and their acquisition separate it out in a new PR which we can think of more (and apply the comments of this PR), if you can simplify and improve the code. And if you can we can decide if it should go in this release or if we should push for later.

If you can't we can close and I will think about it at some point, since there can definitely be issues with those locks but they are non-fatal/edge cases and adding more complexity will turn them into serious errors.

rotkehlchen/rotkehlchen.py
```diff
@@ -106,6 +114,20 @@
...
+def kill_greenlet_and_acquire_locks(
+        stack: contextlib.ExitStack,
+        acquired_by_this_method: dict[Semaphore, bool],
+        greenlet: gevent.Greenlet,
+        locks: list[Semaphore],
+) -> None:
+    """Kills the provided greenlet and tries to acquire the locks."""
+    greenlet.kill(exception=GreenletKilledError('Killed to delete evm addresses'))
+    for lock in locks:
+        if lock.locked() is False:
```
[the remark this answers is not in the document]
nebolax: When a greenlet is killed, it is not guaranteed that the lock is released. For example, imagine this situation:
1. A periodic task for querying transactions runs (and it acquires a lock)
2. A request is made from frontend that also requests transactions (and it lines up in the queue for that lock)

Then we start killing greenlets. First we kill the api task. But since the lock is acquired by the periodic task, it does not free the lock, so we can't be sure here that the lock is acquirable.

rotkehlchen/rotkehlchen.py
nebolax: Mmm yeah, it is complicated. As I said here https://github.com/rotki/rotki/pull/5529#discussion_r1093063359, we may not be able to acquire a lock immediately after killing a greenlet because the lock may still be used by some other greenlet.

So to make sure that we eventually do acquire all the locks, we need this `acquired_by_this_method` dictionary.

------------

But yeah, it is complicated and not sure whether is really needed. Since the most important thing is to just kill the transactions querying greenlets, which is already in develop. And the bug with duplicated transactions-requesting greenlets is already fixed.

rotkehlchen/rotkehlchen.py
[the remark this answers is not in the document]
LefterisJP: Yes but this still does not make sense to me.

> Then we start killing greenlets

The moment you kill the greenlet, you should get the locks it has acquired. No need to complicate things. Since `greenlet.kill()` kills it asynchronously what you would need todo I guess here is kill and wait for it.

And then yet again I don't understand why not just do a blocking acquire at the very end and be done with it? Since this is what you end up doing anyway. We can take the conversation to discord also.

rotkehlchen/rotkehlchen.py
LefterisJP: This is not needed 100% and the approach is too complicated. Taking it to discord to explain it better. Read my other comment

PR #5530, opened 2023-02-01
feat: clear saved newly detected tokens on upgrade

lukicenturi:
Closes #(issue_number)

## Checklist

- [x]  clear saved newly detected tokens on upgrade
- [x] message dialog can now be closed with "Esc" or "Enter", for better UX
- [x] For merge dialog, I change the source asset to use text-field instead, at least from the docs https://rotki.readthedocs.io/en/latest/usage_guide.html#merging-two-assets, it support unknown identifier.

### DIFF
--- a/frontend/app/src/components/account-management/upgrade/UpgradeProgressDisplay.vue
+++ b/frontend/app/src/components/account-management/upgrade/UpgradeProgressDisplay.vue
@@ -3,6 +3,7 @@ import { type ComputedRef } from 'vue';
 import { type CurrentDbUpgradeProgress } from '@/types/login';
 import { useSessionAuthStore } from '@/store/session/auth';
 import DbActivityProgress from '@/components/account-management/upgrade/DbActivityProgress.vue';
+import { useNewlyDetectedTokens } from '@/composables/assets/newly-detected-tokens';
 
 const { dbUpgradeStatus, dataMigrationStatus } = storeToRefs(
   useSessionAuthStore()
@@ -51,6 +52,14 @@ const dataMigrationStatusData: ComputedRef<CurrentDbUpgradeProgress | null> =
       description: description || ''
     };
   });
+
+const { clearInternalTokens } = useNewlyDetectedTokens();
+
+onMounted(() => {
+  if (get(dbUpgradeStatus)) {
+    clearInternalTokens();
+  }
+});
 </script>
 
 <template>
--- a/frontend/app/src/components/asset-manager/MergeDialog.vue
+++ b/frontend/app/src/components/asset-manager/MergeDialog.vue
@@ -113,14 +113,16 @@ const v$ = useVuelidate(
       <div v-if="done">{{ t('merge_dialog.done') }}</div>
 
       <v-form v-else :value="!v$.$invalid">
-        <asset-select
+        <!-- We use `v-text-field` here instead `asset-select` -->
+        <!-- because the source can be filled with unknown identifier -->
+        <v-text-field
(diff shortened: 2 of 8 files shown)


frontend/app/src/components/asset-manager/MergeDialog.vue
```diff
@@ -113,14 +113,14 @@ const v$ = useVuelidate(
       <div v-if="done">{{ t('merge_dialog.done') }}</div>
 
       <v-form v-else :value="!v$.$invalid">
-        <asset-select
+        <v-text-field
```
kelsos, line 116: When we did the change I was pretty sure that we had a reason for this 😄 
Thanks for confirming. 

Can you add a comment above it so that we don't accidentally remove it in the future?

frontend/app/src/components/status/update/AssetUpdateMessage.vue
```diff
@@ -16,6 +16,7 @@ const emit = defineEmits<{
 const { versions } = toRefs(props);
 
 const partial: Ref<boolean> = ref(false);
+const maxToVersion: Ref<number> = ref(0);
```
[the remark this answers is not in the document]
lukicenturi: I found bug before, if you can upgrade from v 14 to 16.
And you activate the partial, set lowest version possible (14), and then disabled the partial
If I press update, it should upgrade to the default version (to the max / 16), but instead the value used is 14 (the partial value, which shouldn't be used because it's been disabled)

So what I do is, if we disable the partial, I set it back to the default version

frontend/app/src/components/account-management/upgrade/UpgradeProgressDisplay.vue
```diff
@@ -51,6 +52,12 @@ const dataMigrationStatusData: ComputedRef<CurrentDbUpgradeProgress | null> =
       description: description || ''
     };
   });
+
+const { clearInternalTokens } = useNewlyDetectedTokens();
+
+onBeforeMount(() => {
+  clearInternalTokens();
```
kelsos, line 59: since this, if before mount we can't check for dbUpgradeStatus, really.

Lefteris mentioned that this should not happen if it's only data migration

lukicenturi: so this is only happens if `dbUpgradeProgressData` is not null?

frontend/app/src/components/status/update/AssetUpdateMessage.vue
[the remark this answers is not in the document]
kelsos: Can't we just update the upToVersion to the remote version?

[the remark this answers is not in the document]
lukicenturi: oh yeah, just realize I can do that too, will update it

frontend/app/src/components/account-management/upgrade/UpgradeProgressDisplay.vue
kelsos: yes exactly

lukicenturi: updated

frontend/app/src/components/status/update/AssetUpdateMessage.vue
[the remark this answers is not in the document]
lukicenturi: updated

kelsos (approved):
lgtm

PR #5533, opened 2023-02-01
Handle killed greenlets

nebolax:


LefterisJP (approved):
LGTM. The failed test is due to kusama problems and is not related

PR #5535, opened 2023-02-01
Test that the price used for assets in trades is nominated using the main currency

yabirgb:
The bug in 1.26.3 was making that when the price used for assets in the PnL report  was not available in oracles and was calculated based in trade  then it was not using the main currency as reference.

This behavior changed along the changes made to the accounting process for 1.27 and this PR adds a test to verify this. The provided test fails (with the wrong price) in 1.26.3 and works fine in 1.27

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -20,6 +20,7 @@ Changelog
 * :feature:`-` Swaps made via uniswap v3 auto routers (both v1 and v2) will now be decoded correctly.
 * :feature:`3231` Optimism is now supported. Optimism balances will be shown and optimism transactions will be decoded.
 * :feature:`1756` Uniswap and sushiswap pool join/exit events are now properly decoded and taken into account during PnL report.
+* :bug:`-` Fixed an issue where price for pairs of fiat currencies was not queried properly.
 * :bug:`-` Customized ownership proportions of validators owned by eth1 addresses will now be properly respected.
 
 * :release:`1.26.3 <2022-12-30>`
--- a/rotkehlchen/tests/unit/accounting/test_misc.py
+++ b/rotkehlchen/tests/unit/accounting/test_misc.py
@@ -1,16 +1,21 @@
+from typing import TYPE_CHECKING
 import pytest
 
 from rotkehlchen.accounting.ledger_actions import LedgerAction, LedgerActionType
-from rotkehlchen.accounting.mixins.event import AccountingEventType
+from rotkehlchen.accounting.mixins.event import AccountingEventMixin, AccountingEventType
 from rotkehlchen.accounting.pnl import PNL, PnlTotals
 from rotkehlchen.constants import ONE, ZERO
-from rotkehlchen.constants.assets import A_ETH, A_EUR, A_KFEE, A_USDT
+from rotkehlchen.constants.assets import A_ETH, A_ETH2, A_EUR, A_KFEE, A_USD, A_USDT
 from rotkehlchen.exchanges.data_structures import Trade
 from rotkehlchen.fval import FVal
 from rotkehlchen.tests.utils.accounting import accounting_history_process, check_pnls_and_csv
+from rotkehlchen.tests.utils.constants import A_GBP
 from rotkehlchen.tests.utils.history import prices
 from rotkehlchen.tests.utils.messages import no_message_errors
-from rotkehlchen.types import Location, Timestamp, TradeType
+from rotkehlchen.types import AssetAmount, Fee, Location, Price, Timestamp, TradeType
+
+if TYPE_CHECKING:
+    from rotkehlchen.accounting.accountant import Accountant
 
 
 @pytest.mark.parametrize('mocked_price_queries', [prices])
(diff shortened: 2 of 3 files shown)


docs/changelog.rst
```diff
@@ -20,6 +20,7 @@ Changelog
 * :feature:`-` Swaps made via uniswap v3 auto routers (both v1 and v2) will now be decoded correctly.
 * :feature:`3231` Optimism is now supported. Optimism balances will be shown and optimism transactions will be decoded.
 * :feature:`1756` Uniswap and sushiswap pool join/exit events are now properly decoded and taken into account during PnL report.
+* :bug:`-` PnL report will now respect the profit currency value for trades when the price for assets is calculated instead of queried.
```
LefterisJP, line 23: so it was not only for manual trades? If it's only manual trades state it as it's much easier to understand from the user perspective

rotkehlchen/tests/unit/accounting/test_misc.py
```diff
@@ -180,3 +185,42 @@ def test_fees_in_received_asset(accountant, google_service):
         AccountingEventType.LEDGER_ACTION: PNL(taxable=FVal('178.615'), free=ZERO),
     })
     check_pnls_and_csv(accountant, expected_pnls, google_service)
+
+
+@pytest.mark.parametrize('should_mock_price_queries', [False])
+def test_native_currency_is_respected(accountant: 'Accountant') -> None:
```
LefterisJP, line 191: What is native currency? Do you mean profit_currency/main_currency? Do not introduce new terms. Use the term we use in the rest of the code

rotkehlchen/tests/unit/accounting/test_misc.py
```diff
@@ -180,3 +185,42 @@ def test_fees_in_received_asset(accountant, google_service):
...
+    """Verify that the price after processing trades respects the native currency. We change the
+    native currency to GBP and assert the price for a trade using a different currency as quote.
+
+    The experienced issue appears when price is not available in oracles and needs to be
+    calculated in our side.
```
LefterisJP, line 196:
```suggestion
    The issue in questionappears when price is not available in oracles and needs to be
    calculated on our side.
```

rotkehlchen/tests/unit/accounting/test_misc.py
```diff
@@ -180,3 +185,42 @@ def test_fees_in_received_asset(accountant, google_service):
...
+        ),
+    ]
+    accounting_history_process(
+        accountant=accountant,
+        start_ts=Timestamp(1436979735),
+        end_ts=Timestamp(1625001466),
+        history_list=history,
+    )
+    warnings = accountant.msg_aggregator.consume_warnings()
+    assert len(warnings) == 0
```
LefterisJP, line 222: what about checking for errors from aggregator here?

rotkehlchen/tests/unit/accounting/test_misc.py
```diff
@@ -180,3 +185,42 @@ def test_fees_in_received_asset(accountant, google_service):
         AccountingEventType.LEDGER_ACTION: PNL(taxable=FVal('178.615'), free=ZERO),
     })
     check_pnls_and_csv(accountant, expected_pnls, google_service)
+
+
+@pytest.mark.parametrize('should_mock_price_queries', [False])
```
LefterisJP, line 190: Why not mock price queries? You should mock the historical queries to make sure the test is repeatable.

rotkehlchen/tests/unit/accounting/test_misc.py
```diff
@@ -180,3 +185,42 @@ def test_fees_in_received_asset(accountant, google_service):
...
+        accountant=accountant,
+        start_ts=Timestamp(1436979735),
+        end_ts=Timestamp(1625001466),
+        history_list=history,
+    )
+    warnings = accountant.msg_aggregator.consume_warnings()
+    assert len(warnings) == 0
+    # Check that the price is correctly computed in GBP. Since the rate is 2403.20$ the
+    # GBP price should be 2403.20 * 0.741076 == 1780.9538432 being 0.741076 the ratio
+    # of USD to GBP on the trade moment
```
LefterisJP, line 225:
```suggestion
    # of USD to GBP at the moment of the trade
```

rotkehlchen/tests/unit/accounting/test_misc.py
yabirgb: I've mocked the result from xrates to verify that the logic is indeed correct by patching the price historian

rotkehlchen/tests/unit/accounting/test_misc.py
```diff
@@ -180,3 +185,44 @@ def test_fees_in_received_asset(accountant, google_service):
...
+            link=None,
+        ),
+    ]
+    accounting_history_process(
+        accountant=accountant,
+        start_ts=Timestamp(1436979735),
+        end_ts=Timestamp(1625001466),
+        history_list=history,
+    )
+    errors = warnings = accountant.msg_aggregator.consume_errors()
```
LefterisJP, line 222: dude wtf?

```suggestion
    errors = accountant.msg_aggregator.consume_errors()
```

rotkehlchen/tests/unit/accounting/test_misc.py
```diff
@@ -180,3 +185,44 @@ def test_fees_in_received_asset(accountant, google_service):
...
+    warnings = accountant.msg_aggregator.consume_warnings()
+    assert len(warnings) == len(errors) == 0
+    # Check that the price is correctly computed in GBP. Since the rate is 2403.20$ the
+    # GBP price should be 2403.20 * 0.741076 == 1780.9538432 being 0.741076 the ratio
+    # of USD to GBP at the moment of the trade
```
LefterisJP, line 227: Don't overdo it with the comments. Mocking stuff also makes these stuff a lot more explicit

```suggestion
    # GBP price should be 2403.20 * 0.741076 == 1780.9538432
```

you can even be a lot more explicit by doing this calculation exactly. By doing `history[0].rate * mocked_price_queries['USD'][GBP'] and get rid of the comment completely

LefterisJP (approved):
LGTM. Thank you.

PR #5537, opened 2023-02-01
Small tags refactor

nebolax:
Related conversation:

https://discord.com/channels/657906918408585217/683245846070165504/1067858892874321990
image

## Checklist

### DIFF
--- a/rotkehlchen/db/utils.py
+++ b/rotkehlchen/db/utils.py
@@ -260,6 +260,20 @@ def deserialize_tags_from_db(val: Optional[str]) -> Optional[list[str]]:
     return tags
 
 
+def _get_tag_reference(
+        entry: Union['ManuallyTrackedBalance', BlockchainAccountData, 'XpubData'],
+        object_reference_keys: list[
+            Literal['id', 'chain', 'address', 'xpub.xpub', 'derivation_path'],
+        ],
+) -> str:
+    reference = ''
+    for key in object_reference_keys:
+        value = rgetattr(entry, key)
+        if value is not None:  # value.value is for SupportedBlockchain
+            reference += str(value) if key != 'chain' else value.value
+    return reference
+
+
 def _prepare_tag_mappings(
         entry: Union['ManuallyTrackedBalance', BlockchainAccountData, 'XpubData'],
         object_reference_keys: list[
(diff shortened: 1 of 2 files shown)


nebolax:
Reminder to myself: don't forget to update the method that updated the tags of derived xpub addresses

rotkehlchen/db/utils.py
```diff
@@ -260,19 +260,29 @@ def deserialize_tags_from_db(val: Optional[str]) -> Optional[list[str]]:
...
+
+def _prepare_tag_mappings(
+        entry: Union['ManuallyTrackedBalance', BlockchainAccountData, 'XpubData'],
+        object_reference_keys: list[
+            Literal['id', 'chain', 'address', 'xpub.xpub', 'derivation_path'],
+        ],
+) -> list[tuple[str, str]]:
+    """Common function to prepare tag mappings. Caller has to make sure entry.tags is not None"""
+    mapping_tuples = []
+    reference = _get_tag_reference(entry, object_reference_keys)
```
prettyirrelevant, line 285: Is there a need for this new function?
It doesn't do anything different from how the previous logic works. Am I missing something?

Edit: I see you use it below.

rotkehlchen/db/utils.py
```diff
@@ -309,11 +319,10 @@ def replace_tag_mappings(
     insert_tuples = []
     delete_tuples = []
     for entry in data:
+        # Always delete current tags even if there are no new tags.
+        delete_tuples.append((_get_tag_reference(entry, object_reference_keys),))
```
prettyirrelevant, line 323: Could you please add a test to show what added benefit/improvement it brings?
Really difficult to know what side effects it introduces

prettyirrelevant (commented):
Left a comment and a question

rotkehlchen/db/utils.py
nebolax: Yes, makes sense, for some reason thought there already is one. Will add.

nebolax: I don't think there is a need in a test since it is just a refactor. It does not fix anything, but prevents us from hitting errors like the one fixed in #5538.

prettyirrelevant: oh okay then.

LefterisJP (approved):
I don't really remember our conversation from back then :sweat_smile: 

But anyway I vaguely remember something was extraneous and could be removed. Looks vaguely fine and hopefully since all current tests pass it's fine.

PR #5538, opened 2023-02-01
Update tags of derived addresses

nebolax:
Currently tags of xpubs' derived addresses are not updated when an xpub's tags are changed. But they should be updated. This PR fixes it.


LefterisJP (approved):
LGTM. Thanks

PR #5539, opened 2023-02-01
Fix refund of uni v3 router

nebolax:
Previously I have not found a transaction for uniswap v3 auto router that would have an eth refund event.

Now I stumbled upon one. Added a test and adjusted the decoding logic.


LefterisJP (approved):
LGTM. I am not sure exactly on the logic as the entire logic there is complicated, but I don't see anything super weird.

And the test transaction you included seems to pass. How was it decoded before? Was the amount of ETH swapped bigger than reality since it was missing the refund?

PR #5540, opened 2023-02-01
Avoid decoding transactions for the wrong chain

yabirgb:
@kelsos found and saw where the code was failing to decode transactions because the eth decoder was trying to decode an op transaction. We couldn't reproduce it in the wild outside his computer but a test can reproduce it easily

### DIFF
--- a/rotkehlchen/tests/unit/test_evm_tx_decoding.py
+++ b/rotkehlchen/tests/unit/test_evm_tx_decoding.py
@@ -1,3 +1,4 @@
+from typing import TYPE_CHECKING
 from unittest.mock import patch
 
 import pytest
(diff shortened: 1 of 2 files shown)


rotkehlchen/tests/unit/test_evm_tx_decoding.py
```diff
@@ -14,9 +15,82 @@
...
 from rotkehlchen.utils.hexbytes import hexstring_to_bytes
 
+if TYPE_CHECKING:
+    from rotkehlchen.chain.ethereum.transactions import EthereumTransactions
+    from rotkehlchen.chain.optimism.decoding.decoder import OptimismTransactionDecoder
+    from rotkehlchen.chain.optimism.transactions import OptimismTransactions
+    from rotkehlchen.db.dbhandler import DBHandler
+
+
+def _add_transactions_to_db(
```
LefterisJP, line 36: Why move this outside in its own function when it's only used in one place? What do you achieve?

LefterisJP (commented):
LGTM apart from some comment rewording and nitpicks. Let's see how tests do

rotkehlchen/tests/unit/test_evm_tx_decoding.py
yabirgb: It was because I wanted to reuse it but then I noticed that I could consolidate in one test. I left it as an aux function because for me it was easier to read the code without the hustle of the setup

PR #5549, opened 2023-02-02
Make the create user endpoint async

yabirgb:
This PR makes the endpoint to create user async in order to allow ws messages during the sync process in frontend


LefterisJP (commented):
you need to change api docs also. But don't add it in this PR if this is the only error. Let's see how tests do

LefterisJP (approved):
LGTM

PR #5550, opened 2023-02-02
Use token kind in cache key

nebolax:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/evm/node_inquirer.py
+++ b/rotkehlchen/chain/evm/node_inquirer.py
@@ -2,6 +2,7 @@ import json
 import logging
 import random
 from abc import ABCMeta, abstractmethod
+from itertools import zip_longest
 from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Sequence, Union
 from urllib.parse import urlparse
 
@@ -199,8 +200,9 @@ class EvmNodeInquirer(metaclass=ABCMeta):
 
         log.debug(f'Initializing {self.chain_name} inquirer. Nodes to connect {connect_at_start}')
 
-        # A cache for the erc20 contract info to not requery same one
-        self.contract_info_cache: dict[ChecksumEvmAddress, dict[str, Any]] = {}
+        # A cache for erc20 and erc721 contract info to not requery the info
+        self.contract_info_erc20_cache: dict[ChecksumEvmAddress, dict[str, Any]] = {}
+        self.contract_info_erc721_cache: dict[ChecksumEvmAddress, dict[str, Any]] = {}
         self.connect_to_multiple_nodes(connect_at_start)
 
     def connected_to_any_web3(self) -> bool:
@@ -967,7 +969,7 @@ class EvmNodeInquirer(metaclass=ABCMeta):
         if it is provided in the contract. This method may raise:
         - BadFunctionCallOutput: If there is an error calling a bad address
         """
-        cache = self.contract_info_cache.get(address)
+        cache = self.contract_info_erc20_cache.get(address)
         if cache is not None:
             return cache
 
@@ -1012,12 +1014,12 @@ class EvmNodeInquirer(metaclass=ABCMeta):
             )
             log.debug(f'{address} was succesfuly decoded as ERC20 token')
 
-        for prop, value in zip(properties, decoded):
+        for prop, value in zip_longest(properties, decoded):
             if isinstance(value, bytes):
                 value = value.rstrip(b'\x00').decode()
             info[prop] = value
 
-        self.contract_info_cache[address] = info
+        self.contract_info_erc20_cache[address] = info
         return info
 
     def get_erc721_contract_info(self, address: ChecksumEvmAddress) -> dict[str, Any]:
@@ -1032,7 +1034,7 @@ class EvmNodeInquirer(metaclass=ABCMeta):
         - BadFunctionCallOutput: If there is an error calling a bad address
         - NotERC721Conformant: If the address can't be decoded as an ERC721 contract
         """
-        cache = self.contract_info_cache.get(address)
+        cache = self.contract_info_erc721_cache.get(address)
         if cache is not None:
             return cache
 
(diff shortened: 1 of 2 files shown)


rotkehlchen/tests/unit/test_tokens.py
```diff
@@ -140,3 +147,43 @@ def test_last_queried_ts(tokens, freezer):
...
+            new=MagicMock(side_effect=AssertionError('Chain calls should not be made')),
+        )
+
+    with patch_multicall_2(ERC20_INFO_RESPONSE):
+        erc20_token_data = query_token_info(EvmTokenKind.ERC20)
+
+    with patch_multicall_2(ERC721_INFO_RESPONSE):
+        erc721_token_data = query_token_info(EvmTokenKind.ERC721)
+
+    with disable_chain_calls():
```
yabirgb, line 184: you can avoid the function in this case and just use the patch directly

rotkehlchen/tests/unit/test_tokens.py
```diff
@@ -140,3 +147,43 @@ def test_last_queried_ts(tokens, freezer):
...
         assert after_second_query[0][0] == 'last_queried_timestamp'
         assert int(after_second_query[0][1]) >= continuation
+
+
+def test_cache_is_per_token_type(ethereum_inquirer):
+    """This test makes sure that tokens info cache is per token type"""
+    def query_token_info(token_kind):
+        return _query_or_get_given_token_info(
+            evm_inquirer=ethereum_inquirer,
+            evm_address=ADDR,
```
yabirgb, line 157: instead of creating a global variable that persists you can make it here

```suggestion
            evm_address=make_evm_address(),
```

rotkehlchen/tests/unit/test_tokens.py
```diff
@@ -140,3 +147,43 @@ def test_last_queried_ts(tokens, freezer):
         assert len(after_second_query) == 1
         assert after_second_query[0][0] == 'last_queried_timestamp'
         assert int(after_second_query[0][1]) >= continuation
+
+
+def test_cache_is_per_token_type(ethereum_inquirer):
+    """This test makes sure that tokens info cache is per token type"""
+    def query_token_info(token_kind):
+        return _query_or_get_given_token_info(
```
yabirgb, line 155: can you add a small comment explaining how this is used later with the patch to return the desired response?

yabirgb (commented):
A few small requests

rotkehlchen/tests/unit/test_tokens.py
nebolax: No, can't be done like that since the address will be regenerated then for every token info request. Bu it has to be the same

yabirgb: Than maybe create it inside the test? I don't see why it should be a global variable

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -200,7 +201,7 @@ def __init__(
         log.debug(f'Initializing {self.chain_name} inquirer. Nodes to connect {connect_at_start}')
 
         # A cache for the erc20 contract info to not requery same one
-        self.contract_info_cache: dict[ChecksumEvmAddress, dict[str, Any]] = {}
+        self.contract_info_cache: dict[tuple[ChecksumEvmAddress, EvmTokenKind], dict[str, Any]] = {}  # noqa: E501
```
LefterisJP, line 204: That's not what I meant by 2 caches @nebolax 

This approach uses two keys which makes the hashing more complicated than necessary.

It's impossible to have the same address for different stuff in EVM. So the key being only the address is correct. You just need a different map per token kind. So just two different dicts. Or a double dict, but probably 2 different dicts is cleaner?

You would also need to adjust the comment above

rotkehlchen/tests/unit/test_tokens.py
```diff
@@ -140,3 +145,45 @@ def test_last_queried_ts(tokens, freezer):
         assert len(after_second_query) == 1
         assert after_second_query[0][0] == 'last_queried_timestamp'
         assert int(after_second_query[0][1]) >= continuation
+
+
+def test_cache_is_per_token_type(ethereum_inquirer):
+    """This test makes sure that tokens info cache is per token type"""
```
LefterisJP, line 151:
```suggestion
    """This test makes sure that different info cache is used per token type."""
```

rotkehlchen/tests/unit/test_tokens.py
```diff
@@ -140,3 +145,45 @@ def test_last_queried_ts(tokens, freezer):
...
+    address = make_evm_address()
+
+    def query_token_info(token_kind):
+        # Util function to request token info. Doesn't pass name, symbol or decimals because they
```
LefterisJP, line 155: This should have been a docstring, not a comment

rotkehlchen/tests/unit/test_tokens.py
```diff
@@ -140,3 +145,45 @@ def test_last_queried_ts(tokens, freezer):
...
+        return _query_or_get_given_token_info(
+            evm_inquirer=ethereum_inquirer,
+            evm_address=address,
+            name=None,
+            symbol=None,
+            decimals=None,
+            token_kind=token_kind,
+        )
+
+    # This patch method together with ERC20_INFO_RESPONSE and ERC721_INFO_RESPONSE mock tokens info
```
LefterisJP, line 166:
```suggestion
    # This patch method together with ERC20_INFO_RESPONSE and ERC721_INFO_RESPONSE mocks tokens info
```

rotkehlchen/chain/evm/node_inquirer.py
nebolax: Yes, makes sense. 2 dicts sounds better to me

rotkehlchen/tests/unit/test_tokens.py
```diff
@@ -140,3 +145,50 @@ def test_last_queried_ts(tokens, freezer):
...
+            return_value=return_value,
+        )
+
+    with patch_multicall_2(ERC20_INFO_RESPONSE):
+        erc20_token_data = query_token_info(EvmTokenKind.ERC20)
+
+    with patch_multicall_2(ERC721_INFO_RESPONSE):
+        erc721_token_data = query_token_info(EvmTokenKind.ERC721)
+
+    with patch.object(  # disable chain calls
```
LefterisJP, line 185: comment elaboration as it took me some time to understand what you do here

```suggestion
    with patch.object(  # fail if cache is not used and chain is called
```

LefterisJP (approved):
LGTM. Assuming tests pass. Got a  nitpick that can go later too

LefterisJP:
Merged. Don't forget the nitpick @nebolax for your other PRs. I definitely will forget

PR #5551, opened 2023-02-02
Fix two minor issues backend

yabirgb:
This PR:
- solves an error with message being a tuple instead of a string
- prevents the curve module from raising `NotERC20Conformant` since the curve pool tokens follow the standard. The special case was `0x6b8734ad31D42F5c05A86594314837C416ADA984`

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/curve/pools_cache.py
+++ b/rotkehlchen/chain/ethereum/modules/curve/pools_cache.py
@@ -1,3 +1,4 @@
+import logging
 from typing import TYPE_CHECKING, Optional
 
 from eth_utils import to_checksum_address
@@ -9,7 +10,9 @@ from rotkehlchen.chain.evm.contracts import EvmContract
 from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.constants import ONE
 from rotkehlchen.db.drivers.gevent import DBCursor
+from rotkehlchen.errors.misc import NotERC20Conformant
 from rotkehlchen.globaldb.handler import GlobalDBHandler
+from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.types import (
     CURVE_POOL_PROTOCOL,
     ChainID,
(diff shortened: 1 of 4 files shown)


rotkehlchen/chain/ethereum/modules/curve/pools_cache.py
```diff
@@ -113,6 +115,8 @@ def ensure_curve_tokens_existence(
                         weight=ONE,
                     )],
                     seen=TokenSeenAt(description='Querying curve pools'),
+                    name=ethaddress_to_identifier(token_address),
+                    symbol=DEFAULT_MISSING_SYMBOL,
```
nebolax, line 119: Have you checked that the symbol doesn't end up being `DEFAULT_MISSING_SYMBOL` in all the tokens? At least by looking at this line of the code https://github.com/rotki/rotki/blob/8ae08751302aeebb844de45b28b2bbd713d8079b/rotkehlchen/assets/utils.py#L59 it sounds to me that it will be preferred over the information on chain

rotkehlchen/chain/ethereum/modules/curve/pools_cache.py
```diff
@@ -113,6 +115,8 @@ def ensure_curve_tokens_existence(
                         weight=ONE,
                     )],
                     seen=TokenSeenAt(description='Querying curve pools'),
+                    name=ethaddress_to_identifier(token_address),
```
nebolax, line 118: Name being the token identifier :thinking: Does it make any sense?

rotkehlchen/chain/ethereum/modules/curve/pools_cache.py
yabirgb: Good catch. Indeed that is what would happend. I've changed to catch the error instead and skip

rotkehlchen/chain/ethereum/modules/curve/pools_cache.py
yabirgb: Replaced this but is what we do when the token is missing the name field in other places in the code

rotkehlchen/chain/ethereum/modules/curve/pools_cache.py
```diff
@@ -80,40 +75,72 @@ def ensure_curve_tokens_existence(
...
+                    )
+                except NotERC20Conformant as e:
+                    log.info(
+                        f'Skipping pool {pool_info} because {token_address} is not a '
+                        f'valid ERC20 token. {str(e)}',
+                    )
+                    continue
+
+        # finally ensure lp token exists in the globaldb
+        get_or_create_evm_token(
```
LefterisJP, line 136: will lp token always be ERC20Conformant? You checkes everything else for it.

LefterisJP (approved):
LGTM

PR #5554, opened 2023-02-03
Fixes trades refresh to only check connected exchanges

kelsos:
Closes #(issue_number)

## Checklist

- [ ] The PR modified the frontend, and updated the [user guide](https://github.com/rotki/rotki/blob/develop/docs/usage_guide.rst) to reflect the changes.

With this, the `only_cache=False` requests will happen only for connected exchanges

PR #5555, opened 2023-02-03
Not connected, not syncing exchange query won't query all

LefterisJP:
Due to a faulty if condition if an exchange that used to be connected to and is now removed is attempted to be queried, for all trades then all exchanges would be queried instead if that exchange was also in the non-syncing exchanges list.

This commit adds a test for and fixes it.

This was uncovered due to frontend querying FTX with cache_only being False while API key are removed (which is separately fixed in #5554 )


LefterisJP:
The only failed test is coingecko timeout: `rotkehlchen/tests/external_apis/test_coingecko.py::test_coingecko_historical_price` and it passed locally so merging.

PR #5556, opened 2023-02-03
Reset events

nebolax:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/db/schema.py
+++ b/rotkehlchen/db/schema.py
@@ -530,6 +530,7 @@ CREATE TABLE IF NOT EXISTS used_query_ranges (
 );
 """
 
+# Currently this table is used only to store a flag that shows whether a transaction is decoded.
 DB_CREATE_EVM_TX_MAPPINGS = """
 CREATE TABLE IF NOT EXISTS evm_tx_mappings (
     tx_hash BLOB NOT NULL,
@@ -634,6 +635,7 @@ CREATE TABLE IF NOT EXISTS history_events (
 );
 """
 
+# This table is used to store for each history event it's chain id and whether it is customized.
 DB_CREATE_HISTORY_EVENTS_MAPPINGS = """
 CREATE TABLE IF NOT EXISTS history_events_mappings (
     parent_identifier INTEGER NOT NULL,
--- a/rotkehlchen/db/upgrades/v35_v36.py
+++ b/rotkehlchen/db/upgrades/v35_v36.py
@@ -2,22 +2,23 @@ import json
 import logging
 from typing import TYPE_CHECKING
 
-from rotkehlchen.accounting.structures.base import LIQUITY_STAKING_DETAILS
-from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.chain.ethereum.modules.eth2.utils import (
     DAY_AFTER_ETH2_GENESIS,
     INITIAL_ETH_DEPOSIT,
 )
-from rotkehlchen.chain.ethereum.modules.liquity.constants import CPT_LIQUITY
 from rotkehlchen.constants.misc import ONE
+from rotkehlchen.db.constants import (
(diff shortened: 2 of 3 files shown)


nebolax:
A couple of tests failed, but they are are about assets / balances and are not related.

rotkehlchen/db/upgrades/v35_v36.py
```diff
@@ -616,6 +587,28 @@ def _fix_eth2_pnl_genesis(write_cursor: 'DBCursor') -> None:
         )
 
 
+def _reset_decoded_events(write_cursor: 'DBCursor') -> None:
+    write_cursor.execute('SELECT tx_hash from evm_transactions')
```
yabirgb, line 591: when writting such a functions always add a docstring explaining what it does

rotkehlchen/db/upgrades/v35_v36.py
```diff
@@ -616,6 +587,28 @@ def _fix_eth2_pnl_genesis(write_cursor: 'DBCursor') -> None:
...
+    tx_hashes = [x[0] for x in write_cursor]
+    write_cursor.execute(
+        'SELECT parent_identifier FROM history_events_mappings WHERE name=? AND value=?',
+        (HISTORY_MAPPING_KEY_STATE, HISTORY_MAPPING_STATE_CUSTOMIZED),
+    )
+    customized_event_ids = [x[0] for x in write_cursor]
```
yabirgb, line 597: in line 591 you are fetching all the tx hashes right? and in line 593 you are getting the customized ones and then you say 'Delete from history events where tx_hash in all and not in customized'. Can't you just do
```
ids_to_delete = list(set(all tx_hash) - set(customized))
```

and say

delete if in `ids_to_delete`. Also I'm not 100% sure but I believe doing a subquery is faster than checking if `event_identifier IN (?,....,?)`

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1202,21 +1202,8 @@ def test_upgrade_db_35_to_36(user_data_dir):  # pylint: disable=unused-argument
...
-            elif i == 1:
-                assert entry[i] == 1
-            else:
-                if entry[i] == 0:
-                    assert tx_mappings_result[idx][i] == 'decoded'
-                elif entry[i] == 1:
-                    assert tx_mappings_result[idx][i] == 'customized'
-                else:
-                    raise AssertionError(f'Unexpected value {entry[i]} in evm_tx_mappings')
-    assert cursor.execute('SELECT COUNT(*) from history_events').fetchone()[0] == 562
```
yabirgb, line 1219: all the history events are evm_transactions? If there was a customized event there should be at least one that is not right?

rotkehlchen/db/upgrades/v35_v36.py
yabirgb: Can you also state that this needs to happen after `_upgrade_events_mappings` since is altering the schema

yabirgb (commented):
Left some comments

rotkehlchen/db/upgrades/v35_v36.py
LefterisJP: This is copy paste from the develop code that is tested. I would not change it. This should have been in the docstring.

rotkehlchen/db/schema.py
```diff
@@ -634,6 +635,7 @@
 );
 """
 
+# This table is used to store for each history event it's chain id and whether it is customized.
```
LefterisJP, line 638:
```suggestion
# This table is used to store for each history event its chain id and whether it is customized.
```

LefterisJP (approved):
Should be fine. Can merge if tests are okay. But please remember the typo for next PRs. I have done such nitpicks in other PRs. I hope you remember to fix them all because I don't remember if you follow up or not.

LefterisJP:
3 tests failed with timeouts from coingecko

PR #5583, opened 2023-02-08
Do not send user message but just log warning for remote sync check

LefterisJP:
It can happen way too often that a remote is ~10 blocks away from the "source of truth" ... or that the source of truth is itself wrong.

So just log a warning in the logs and don't bother the user


LefterisJP:
nobody saw this or cared to comment so just merging

yabirgb:
Didn't see it but looks okey. Thanks sir :pray:

PR #5585, opened 2023-02-08
Multitake refactor and curve fix

nebolax:
- [x] Refactors multitake logic
- [x] Fixes curve deposits and withdrawals accounting

I renamed `TxMultitakeTreatment` to `TxSpecialTreatment` because it is no longer specified for all multitake events. For example for curve I made just a callback.


rotkehlchen/accounting/transactions.py
```diff
@@ -46,41 +66,22 @@ def process(
...
                 event=event,
-                other_events=other_events,
+                other_events=history_base_entries_iterator(events_iterator, event),
             )
 
-        if event_settings.multitake_treatment == TxMultitakeTreatment.SWAP:
+        if event_settings.special_treatment == TxSpecialTreatment.SWAP:
+            in_event = next(history_base_entries_iterator(events_iterator, event), None)
+            if in_event is None:
+                return 1
```
LefterisJP, line 80: Isn't this a serious error? I would expect a log here

rotkehlchen/accounting/transactions.py
```diff
@@ -15,6 +15,26 @@
...
+        events_iterator: Iterator[AccountingEventMixin],
+        associated_event: HistoryBaseEntry,
+) -> Iterator[HistoryBaseEntry]:
+    """
+    Takes an iterator of accounting events and transforms it into a history base entries iterator.
+    Takes associated event as an argument to be able to log it in case of errors.
+    """
+    for event in events_iterator:
+        if not isinstance(event, HistoryBaseEntry):
+            log.debug(
```
LefterisJP, line 28:
```suggestion
            log.error(
```

rotkehlchen/tests/unit/accounting/test_decoded_events.py
```diff
@@ -0,0 +1,200 @@
+# Some utility timestamps. They are used for mocked prices.
```
LefterisJP, line 1: Seems like the comment is off. And if it's for the tiemstamps after all imports also not needed

```suggestion
```

rotkehlchen/tests/unit/accounting/test_decoded_events.py
```diff
@@ -0,0 +1,200 @@
...
+import pytest
+
```
LefterisJP, line 3: Also this filename does not make much sense. Is it just for curve? Seems like it. For EVM transactions we also have `tests/unit/accounting/test_transactions.py`. What I would suggest is make a new directory under `unit/accounting` and call it `unit/accounting/evm` and put both files there. Rename this one to curve, since it concerns only curve events.

So would be something like:

```
tests/unit/accounting/evm/
 |
 | --- test_transactions.py
 | --- test_curve.py
```

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -158,6 +161,19 @@ def test_curve_deposit(database, ethereum_transaction_decoder):
             sequence_index=371,
             timestamp=1650276061000,
             location=Location.ETHEREUM,
+            event_type=HistoryEventType.RECEIVE,
```
LefterisJP, line 164: Question: In this test only the order of detected events changed, due to what you did, right? No other details changed, apart from the "extra_data" right?

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -749,3 +756,94 @@ def test_curve_remove_imbalanced(database, ethereum_transaction_decoder):
             counterparty=CPT_CURVE,
         )]
     assert expected_events == events
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('ethereum_accounts', [['0x6Bb553FFC5716782051f51b564Bb149D9946f0d2']])
+def test_deposit_multiple_tokens(ethereum_transaction_decoder, ethereum_accounts):
```
LefterisJP, line 763: Need a docstring here. Also this seems to fail in the tests.

LefterisJP (commented):
Looked it over and left some comments/questions.

Some general notes:

### Naming

The name `special_treatment` is too generic and sounds funny.
It concerns how an event and potentially its surrounding events, will be consumed for accounting. So perhaps `accounting_treatment` and `TxAccountingTreatment`  ?

### Iterator vs List

Why did you opt to pass the iterator further down instead of consuming and passing the list as it was before?

### User Experience in 1.27.1

Users of 1.27.1 would need to redecode all of their Curve related, events. Right? No other event decoding **needs to happen**.

So can't we maybe make it so that in the DB migration you just re-decode all events with the curve protocol? And leave the rest as they are. So users don't have to do it manually.

nebolax:
As for why I decided to pass an iterator - is because passing a list would still require that `take` argument, which we wanted to get rid of.

Also sometimes (like in curve) the number of `other_events` to consume is not the same across different transactions and because of that passing an iterator to the callback and letting the callback to decide how many events to consume sounded better to me.

nebolax:
About redecoding - yes, probably it should happen in a migration. And only for curve, yes.

LefterisJP:
> As for why I decided to pass an iterator - is because passing a list would still require that `take` argument, which we wanted to get rid of.

You would not need to do it. That's the length of the list. But yes figuring out the exact amount to pull into the list beforehand is the tricky part.

I don't have any objection on the iterator part. Just wanted to see the logic behind your choice.

rotkehlchen/tests/unit/decoders/test_curve.py
yabirgb: I've checked the tests and seems that the change is re order + adding the info in the extra_data field

LefterisJP (approved):
LGTM.

PR #5590, opened 2023-02-09
Change assets updater to allow spam tokens

yabirgb:
Closes #(issue_number)

## Checklist


prettyirrelevant (commented):
@yabirgb left you some comments

rotkehlchen/assets/spam_assets.py
```diff
@@ -1041,17 +1041,25 @@
...
+def _save_or_update_spam_assets(
+        db: 'DBHandler',
+        list_of_assets: list[dict[str, Any]],
+) -> set[Asset]:
     """
     Generate a set of assets that can be ignored using the list of spam
-    assets KNOWN_EVM_SPAM_TOKENS.
+    assets list_of_assets.
+
+    This function doesn't make any external query.
```
prettyirrelevant, line 1052: nitpick: since the function name changed, I don't think this disclaimer is needed.

rotkehlchen/globaldb/updates.py
```diff
@@ -643,14 +653,21 @@ def perform_update(
                 log.info('Finishing assets update. Replacing users globaldb with the updated information')  # noqa: E501
                 _replace_assets_from_db(GlobalDBHandler().conn, tmpdir / temp_db_name)
 
-            return None
+        with GlobalDBHandler().conn.critical_section():
+            # We apply in a different step since the spam assets don't require conflict resolution
+            # and make use of the logic in the globaldb class so the final db has to be used.
+            # Use a critical section to avoid another greenlet adding the same asset but is fine
+            # since we don't make any external query.
```
LefterisJP, line 660: This comment has some wording mistakes but also does not make sense. If you don't need a critical section, don't do it. If you need do it. Being future proof is also a need.

```suggestion
            # Applied in a different step since the spam assets don't require conflict resolution
            # and utilize logic of the globaldb class so the final db has to be used.
            # Use a critical section to avoid another greenlet adding spam assets at the same time
```

rotkehlchen/globaldb/updates.py
```diff
@@ -692,12 +709,31 @@ def _perform_update(
         # In this case we have conflicts. Everything should also be rolled back
         connection.rollback()
 
+    def _update_spam_assets(
+            self,
+            up_to_version: Optional[int],
+            updates: dict[int, dict[UpdateFileType, str]],
+    ) -> None:
+        """
+        Update spam assets in the globaldb. It also sync them with the list of assets in the
```
LefterisJP, line 718:
```suggestion
        Update spam assets in the globaldb. It also syncs them with the list of assets in the
```

rotkehlchen/globaldb/updates.py
```diff
@@ -692,12 +709,31 @@ def _perform_update(
...
+        user's db
+        """
+        target_version = min(up_to_version, self.last_remote_checked_version) if up_to_version else self.last_remote_checked_version   # noqa: E501
+        for version in range(self.local_assets_version + 1, target_version + 1):
+            if version not in updates:
+                continue
+
+            if len(updates[version][UpdateFileType.SPAM_ASSETS]) != 0:
+                data = json.loads(updates[version][UpdateFileType.SPAM_ASSETS])
+                log.info(f'Applying updates for spam assets from {version}. {len(data)} tokens to add')  # noqa: E501
```
LefterisJP, line 728:
```suggestion
                log.info(f'Applying update for spam assets from {version}. {len(data)} tokens to add')  # noqa: E501
```

rotkehlchen/rotkehlchen.py
```diff
@@ -444,6 +442,7 @@ def unlock_user(
             batch_size=ICONS_BATCH_SIZE,
             sleep_time_secs=ICONS_QUERY_SLEEP,
         )
+        self.assets_updater = AssetsUpdater(self.msg_aggregator, self.data.db)
```
LefterisJP, line 445: kwargs

LefterisJP (commented):
I have not checked the tests.

I checked the logic and left soem comments but I also see multiple problems:

1. Why do you need to combine versions off asset updates with thespam asset updates? Why do we need to have an "update version" for all stuff that gets bumped each time a new thing gets added. Keep in mind we will have new contracts also, new nodes and probably more stuff. Would it not make sense to keep a version per type? Or would that be more difficult for some reason?

2. I think the current logic is broken.
    - When the update happens, you add the new spam assets to the global DB and then to the currently logged in user DB.
    - Say I logout after that and log in with my other account, which does not have the spam assets mirrored in its user DB. Nothing happens. I lost them for that account.

rotkehlchen/assets/spam_assets.py
```diff
@@ -1041,17 +1041,25 @@
 ]
 
 
-def query_token_spam_list(db: 'DBHandler') -> set[Asset]:
+def _save_or_update_spam_assets(
```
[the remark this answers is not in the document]
yabirgb: I've moved it already to https://github.com/rotki/data/tree/develop/updates/spam_assets

LefterisJP (commented):
From your discord ping:

> The thing missing is having a background task but wanted to ask first where do you think is best to save the last time the updates were queried. In the user db we don't have a cache table but we do have a multisetting. Either I store it in multisetting or create a cache table as we have in the globaldb or store it in the globaldb with a per user key 


> missing is having a background task

yes I noticed. Left a comment about this.

> where do you think is best to save the last time the updates were queried

Let's see.

> store it in the globaldb with a per user key 

AAAAAAAA WTF NO 

giphy (9)


> store it in multisetting or create a cache table

Perhaps at some point it would make sense to have a cache table, but this is still not cache, right? Each one of these stuff you pull already has tables it needs to go in. What you want to do is store last time it was queried? Or last version that is stored locally? Or both?

The `multisettings` won't work for this for the reason I explained in the part where you use it. `settings` woudl work but yeah it's true it would pollute it a bit. We can perhaps do `settings` for now and then think about cleaning it up?

rotkehlchen/tests/api/test_assets_updates.py
```diff
@@ -90,6 +90,7 @@ def test_simple_update(rotkehlchen_api_server: 'APIServer', globaldb: GlobalDBHa
     INSERT INTO multiasset_mappings(collection_id, asset) VALUES (99999999, "eip155:1/erc20:0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84");
     *"""  # noqa: E501
     empty_update = {'mappings': '', 'collections': '', 'assets': ''}
+
```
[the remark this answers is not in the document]
yabirgb: I deleted the spam assets info that was here in the previous approach and added a empty line without noticing. Will remove

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -945,7 +945,6 @@ def try_insert_mapping(cur):
         new_ignored_assets_ids = cursor.execute('SELECT value FROM multisettings WHERE name="ignored_asset";').fetchall()  # noqa: E501
         for new_ignored_assets_id in expected_new_ignored_assets_ids:
             assert new_ignored_assets_id in new_ignored_assets_ids
-        assert len(new_ignored_assets_ids) >= 135
```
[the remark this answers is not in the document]
yabirgb: Since spam assets don't get added now on the upgrade and require the data repository they are not here yet. Also the line above this one checks that for the already existing ignored assets their ID gets correctly updated

rotkehlchen/rotkehlchen.py
```diff
@@ -444,6 +443,19 @@ def unlock_user(
             batch_size=ICONS_BATCH_SIZE,
             sleep_time_secs=ICONS_QUERY_SLEEP,
         )
+
+        self.assets_updater = AssetsUpdater(self.msg_aggregator)
+        self.data_updater = RotkiDataUpdater(
+            msg_aggregator=self.msg_aggregator,
+            user_db=self.data.db,
+        )
+        self.greenlet_manager.spawn_and_track(
```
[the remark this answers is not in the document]
yabirgb: Added it

rotkehlchen/tests/fixtures/rotkehlchen.py
```diff
@@ -185,6 +185,11 @@ def augmented_unlock(user, password, create_new, initial_settings):
     # does not run during tests
     size_patch = patch('rotkehlchen.rotkehlchen.ICONS_BATCH_SIZE', new=0)
     sleep_patch = patch('rotkehlchen.rotkehlchen.ICONS_QUERY_SLEEP', new=999999)
+    # don't perform checks for updates
+    rotki_updates_patch = patch(
+        'rotkehlchen.db.updates.RotkiDataUpdater.check_for_updates',
```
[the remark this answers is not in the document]
yabirgb: By default it replaces it with magickmock. With this example

```python
from unittest.mock import patch


class F1Manager:

    def start(self, version: int) -> str:
        print(f'running v{version}')

manager = F1Manager()

manager.start(1)

with patch('__main__.F1Manager.start', autospec=True):
    manager.start(9)

manager.start(3)
```

the output is

```
(rotki-main) ➜  rotki git:(spam-assets-repo) ✗ python pytest_test.py
running v1
running v3
```

tools/assets_database/main.py
```diff
@@ -95,12 +98,23 @@ def main() -> None:
...
     print('Applying updates...')
-    GlobalDBHandler(data_dir=target_directory, sql_vm_instructions_cb=0)
-    assets_updater = AssetsUpdater(msg_aggregator=MessagesAggregator)
-    conflicts = assets_updater.perform_update(
-        up_to_version=args.target_version,
-        conflicts=None,
-    )
+    msg_aggregator = MessagesAggregator()
+    # We need a user db since the spam assets also get synced in the user db but this
+    # will happen on startup for the user db since assets get updated from the globaldb
```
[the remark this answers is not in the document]
yabirgb: Yes, it adds them to the globaldb and also to the user's db ignore list

rotkehlchen/tests/fixtures/rotkehlchen.py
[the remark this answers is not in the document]
LefterisJP: Which does nothing? Are you sure?

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def test_update_spam_assets(data_updater: RotkiDataUpdater) -> None:
+    """
+    Test that spam assets for different chains have been correctly populated
+    """
+    # consume warnings from other tests
+    data_updater.msg_aggregator.consume_warnings()
+
+    # check that spam assets doesn't exist before
```
LefterisJP, line 66:
```suggestion
    # check that spam assets don't exist before
```

rotkehlchen/tests/fixtures/rotkehlchen.py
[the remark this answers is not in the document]
yabirgb: yes, it has the information that mock provides but does nothing

rotkehlchen/db/updates.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class UpdateType(Enum):
+    SPAM_ASSETS = 'spam_assets'
+    RPC_NODES = 'rpc_nodes'
+    CONTRACTS = 'contracts'
+
+    def to_db_key(self) -> str:
+        if self == UpdateType.SPAM_ASSETS:
+            return DATA_UPDATES_DETECT_KEY
```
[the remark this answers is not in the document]
yabirgb: Yes, it makes more sense one for all of them

rotkehlchen/db/updates.py
```diff
@@ -0,0 +1,113 @@
...
+                log.warning(f'Failed to update spam assets due to {str(e)}')
+                break
+            try:
+                if len(updates['assets']) != 0:
+                    log.info(f'Applying update for spam assets from {version}. {len(updates)} tokens to add')  # noqa: E501
+                    with GlobalDBHandler().conn.critical_section():
+                        # Use a critical section to avoid another greenlet adding spam assets at
+                        # the same time
+                        update_spam_assets(db=self.user_db, assets_info=updates['assets'])
+            except KeyError as e:
```
LefterisJP, line 87: nitpick: When I ask to handle the key errors, if it's just one place, keep it only there and if it's just one key you can also just do:

```python
updated_assets = updates.get('assets')
if updated_assets is None:
    log.error(f'Remote update {file_url} does not contain the assets key')
    continue
```

Keep the areas under `try` as small as possible. It's cleaner and shows the potential errors surface clearly

rotkehlchen/db/updates.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def check_for_updates(self) -> None:
+        """Retrieve the information about the latest available update"""
+        remote_information = self._get_remote_info_json()
+
+        # Get latest applied versions
+        with self.user_db.conn.read_ctx() as cursor:
+            cursor.execute(
+                'SELECT value FROM settings WHERE name=?',
+                (UpdateType.SPAM_ASSETS.serialize(),),
```
LefterisJP, line 105: nope. Man ... think a bit more. Don't you see what this will do? This will add a setting with name `spam assets` in the DB? Really? You think this is a good idea?

Just override serialize and don't diretly use `SerializableEnumValueMixin`. I suggested it only as a way to see how an enum can be serialized. 

I will leave a comment in the enum itself to show you what you should do.

rotkehlchen/db/updates.py
```diff
@@ -0,0 +1,113 @@
...
+log = RotkehlchenLogsAdapter(logger)
+
+LAST_DATA_UPDATES_KEY = 'last_data_updates_ts'
+SPAM_ASSETS_URL = 'https://raw.githubusercontent.com/rotki/data/{branch}/updates/spam_assets/assets_v{version}.json'  # noqa: E501
+
+
+class UpdateType(SerializableEnumValueMixin):
+    SPAM_ASSETS = 'spam_assets'
+    RPC_NODES = 'rpc_nodes'
+    CONTRACTS = 'contracts'
```
LefterisJP, line 30: As I explained later this is not gonna work. Create a proper class and use it properly below

```suggestion
class UpdateType(Enum):
    SPAM_ASSETS = 'spam_assets'
    RPC_NODES = 'rpc_nodes'
    CONTRACTS = 'contracts'
    
    def serialize(self) -> str
        """Serializes the update type for the DB and API"""
        return f'{self.value}_last_update_ts'
        
    @classmethod
    def deserialize(cls: type[T], value: str) -> T:
    """Deserialize string from api/DB to UpdateType
    May raise:
    - Deserialization error if value is not a valid UpdateType
    """
    try:
        return cls(value[:-15])  # length of the last_update_ts suffix
    except ValueError as e:
        raise DeserializationError(f'Failed to deserialize UpdateTypevalue {value}') from e
```

rotkehlchen/constants/timing.py
```diff
@@ -21,3 +21,4 @@
 ENS_UPDATE_INTERVAL = DAY_IN_SECONDS
 
 ETH_PROTOCOLS_CACHE_REFRESH = DAY_IN_SECONDS * 3
+DATA_UPDATES_REDETECT = DAY_IN_SECONDS
```
LefterisJP, line 24: As mentioned to Alexey in the other PR let's keep common naming across such variables

```suggestion
DATA_UPDATES_REFRESH = DAY_IN_SECONDS
```

rotkehlchen/tasks/manager.py
```diff
@@ -586,6 +593,21 @@ def _maybe_update_yearn_vaults(self) -> Optional[list[gevent.Greenlet]]:
 
         return None
 
+    def _maybe_check_data_updates(self) -> Optional[list[gevent.Greenlet]]:
+        """
+        Function that schedules the data update task if either there is no data update
+        cache yet or this cache is older than `DATA_UPDATES_REDETECT`
```
LefterisJP, line 599: I asked you to change this name so don't forget to change here

rotkehlchen/tasks/utils.py
```diff
@@ -0,0 +1,24 @@
...
+from rotkehlchen.constants.timing import DATA_UPDATES_REDETECT
+from rotkehlchen.db.updates import LAST_DATA_UPDATES_KEY
+from rotkehlchen.serialization.deserialize import deserialize_timestamp
+from rotkehlchen.utils.misc import ts_now
+
+if TYPE_CHECKING:
+    from rotkehlchen.db.dbhandler import DBHandler
+
+
+def should_check_data_updates(database: 'DBHandler') -> bool:
```
LefterisJP, line 11: Both you and Alexey have the same file here, with the same thing, but just checkign a different constant. This should be abstracted by one of you. Guess whoever handles the conflicts.

LefterisJP (commented):
More comments unfortunately

rotkehlchen/db/updates.py
yabirgb: okey but we don't store the timestamp, is the version

rotkehlchen/db/updates.py
```diff
@@ -0,0 +1,129 @@
...
+from rotkehlchen.utils.network import query_file
+
+if TYPE_CHECKING:
+    from rotkehlchen.db.dbhandler import DBHandler
+    from rotkehlchen.user_messages import MessagesAggregator
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+LAST_DATA_UPDATES_KEY = 'last_data_updates_ts'
```
LefterisJP, line 24: Shouldn't this be also part of update type then with different `ts` for each?

LefterisJP (approved):
I will merge it as is to unblock you. If/When we need multiple last_update_ts per update we can do it and will need a tiny migration.

PR #5598, opened 2023-02-13
Compound borrows

yabirgb:
This PR introduces decoders for:

- Repay events in compound
- Borrow events in compound
- Collect comp token when repaying
- Changes default log level for VCR

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/compound/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/compound/decoder.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Optional
+from typing import Any, Optional, cast
 
 from rotkehlchen.accounting.structures.base import HistoryBaseEntry
 from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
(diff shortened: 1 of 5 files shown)


rotkehlchen/chain/ethereum/modules/compound/decoder.py
```diff
@@ -122,6 +124,49 @@ def _decode_redeem(
...
+    def _decode_borrow_and_repay(
+            self,
+            tx_log: EvmTxReceiptLog,
+            decoded_events: list[HistoryBaseEntry],
+            compound_token: EvmToken,
+    ) -> tuple[Optional[HistoryBaseEntry], list[ActionItem]]:
+        """
+        Decode borrow and repayments for compound tokens
+        """
+        underlying_token = get_crypto_asset_by_symbol(
```
nebolax, line 136: Won't it be better to check `underlying_tokens` of the compound token?

nebolax (commented):
Also "Changes default **debug level** for VCR" confused me :stuck_out_tongue: 

It looked like the log level should have been changed to debug, but in fact it is changed to critical :)

rotkehlchen/chain/ethereum/modules/compound/decoder.py
yabirgb: The reason is that for example the CUSDC token (0x39AA39c021dfbaE8faC545936693aC917d5E7563) doesn't have USDC as underlying token in our database and we use the symbol in the token to know what token is the underlying one. We can change it but I would do in another PR. Also not 100% sure if they have a registry or something to keep it up to date

nebolax: Yeah, okay, checking underlying tokens is just cleaner, but the current approach should be okay too I guess.

The only question is whether cETH can appear here? If yes, then the cast to evm token that is happening below will be wrong.

rotkehlchen/tests/external_apis/test_coingecko.py
```diff
@@ -44,7 +44,7 @@ def test_asset_data(session_coingecko):
         identifier='yearn-finance',
         symbol='yfi',
         name='yearn.finance',
-        image_url='https://assets.coingecko.com/coins/images/11849/small/yfi-192x192.png?1598325330',  # noqa: E501
+        image_url='https://assets.coingecko.com/coins/images/11849/small/yearn-finance-yfi.png?1676174746',  # noqa: E501
```
nebolax, line 47: This is already fixed

LefterisJP: yes but it's in bugfixes and this is a develop PR. It's either this or merge bugfixes to develop, which I want to do but well ... takes time

LefterisJP (approved):
LGTM. I assume the events are decoded as they should be, since you tested it and added tests

Issue #5587, opened 2023-02-09
Make pagination and filtering part of the route/history


kelsos:
## Abstract

We should enable vue-router support for pagination and filtering in tables around the application. 

We can begin with the following:

- Evm Transactions
- Trades
- Asset Movements
- Ledger Actions

How would this would look like?
## Motivation

It adds navigation history so that the user can go back to the last filtered state even after navigating to another part of the application.

## Specification

When the user changes a page/filter/sorting this should be added as a query parameter on the URL.

e.g. Assuming we are using the following parameters:
- **p** for page
- **f** for filter
- **s** for sort by
- **o** for order (asc|desc)

> **Note**: For the ones that support multiple parameters see if we can somehow do an array or something comma separated.

The URL would look like the following:

```
http://localhost:8080/#/history/trades?p=1&f=base%3DETH&s=amount&o=desc
```

Changing page, order, or filter should update the query params, and when navigating for the first time to the route the parameters should update the state.

If the user goes to page two for example and the back button is pressed this should get the user back to page one.

> **Note**: Maybe it makes sense to enable the back button in the toolbar if you are not on the first page without the defaults?

> **Note**: You can test this behavior if you have a mouse that has side buttons for browser navigation

If the page does not have data the user should be redirected back to the first page.

lukicenturi:
Shouldn't we save the applied filter somewhere in store / localStorage?
Save the filter in route works for back button, but it won't work if the user click the tab directly here (?)
![image](https://user-images.githubusercontent.com/26648140/218040742-ee58aef7-1e92-4402-8d78-99910a29f5c5.png)

kelsos:
> Shouldn't we save the applied filter somewhere in store / localStorage?
> Save the filter in route works for back button, but it won't work if the user click the tab directly here (?)
> ![image](https://user-images.githubusercontent.com/26648140/218040742-ee58aef7-1e92-4402-8d78-99910a29f5c5.png)

No, we should not save the last used filter in any storage, this is not what this issue is about. This issue is precisely about making the filters part of the URL history. 

When navigating to the page using either the tab or navigation menu you should always get the defaults, (first page, not filtering and default sorting)

PR #5358, opened 2023-01-08
feat: remove defi subgraph tx view

lukicenturi:
Closes #(issue_number)

PR #5599, opened 2023-02-13
feat: store and read filters from routes for history submenus

lukicenturi:
Part of #5587

## Checklist


kelsos:
I started looking into the PR but it would make sense to finish merging #5358 first.

I will need to re-evaluate our approach in the history pages. I have been thinking a bit and some things we do, don't really make sense anymore, so I need to check if we can actually simplify things.

kelsos:
@lukicenturi it seems that there is a problem with one of the imports can you fix it?

kelsos (approved):
@lukicenturi ok so let's merge this and start applying some changes

1. Remove the global state for the history section and replace with local per page state.
2. Split the actions for data refresh/data fetch into two separate parts.
3. Make sure that backwards/forward history works properly in the other history sections.
   - Make sure that only user actions call the `router.push` (see userAction ref in TransactionContent)

kelsos:
@lukicenturi maybe we can split the work between the two of us to change trades/ledger actions and asset movements

PR #5601, opened 2023-02-13
Periodically query ilk registry to add new makerdao vaults collaterals automatically

### LINKED ISSUE #5576: Ability to support WSTETH Maker Vaults
## Abstract

When moving an ETH vault to a WSTETH vault on Maker, it no longer appears in the app. After a message in the discord group it appears the app does not currently support new vault types.

The requirement is to be able to support other vault types, specifically WSTETH

## Motivation

I would like this to be implemented as it allows any rotki user who has makerdao vaults to see the full details within the defi section of the rotki app as currently these are not visible to users.

LefterisJP:
Closes #5576

Instead of adding the requested vault as a constant went a tiny (which ended up being days of work) bit further and:

- Added GlobalDB data migrations that do not modify the schema
- Added a new migration, which adds the ilk registry as a contract and creates the ilk (collateral type) cache in the global DB (plus does a small cleanup).
- Added code to query etherscan for ABI of verified contracts and their deployed block.
- Added code to query the ilk registry and save all vault ilks collaterals and join adapters.
- Adapted the code we have in makerdao to not have any constant mapppings but instead query the DB whenever it needs to learn something. This saves **a lot of** memory since the constant mappings were in memory and contained contract ABIs.
- Upgraded built-in globalDB to start with the first migration applied and also having ran the ilk registry query and cache update as of 13/02/2023

Some things that will not work even with this change and will need further work:
- If a collateral type has a join contract adapter that is not compatible with GemJoin we can't query it and it's skipped. In order to do that properly we would need to understand how to query other join adapters such as CropJoin, D3MHub etc.
- Now that we query the ABI of contracts it can happen (and it has for collateral type `CRVV1ETHSTETH-A`) that the ABI is for a proxy contract. We need to figure out a way to realize ABI is of proxy contract and ask etherscan again for the implementation contract's ABI. Simple way would be to do it all the time,and more complicated way would be to have a cache of proxy to implementation and ABI mappings?


LefterisJP:
So I think the two failed tests here are due to etherscan so I am going to merge.

Issue #5569, opened 2023-02-06
Make EVM address detection user-initiatable and/or periodic


LefterisJP:
## Abstract

This came up through a conversation with @A2be in discord.

In 1.27.0 we added a data migration that scanned all of the user's ETH addresses and checked the EVM chains supported at the time (Optimism and Avalanche) for activity in those addresses. If the addresses had activity they were also added as tracked addresses in the respective chain.

We should make this an action that can be initiated by the user and/or periodically done action.

## Motivation

It's useful! It should not just happen when a user upgrades to 1.27.0. It would be cool for the user to be able to ask rotki to re-detect EVM addresses with activity or even for rotki to automatically let the user know via the same websocket messages we already added. Something like: "Hey seems 0xsomethingsomething has activity in Optimism now. Adding it for tracking" happening automagically.

## Specification

It should not be too hard.

- [ ] Abstract functionality out of the migration and make it for all supported evm chains.
- [ ] Add it as an api endpoint.
- [ ] Also make it happen periodically.
- [ ] Make sure frontend can process and shows those websocket messages too.

PR #5603, opened 2023-02-13
Detect evm

nebolax:
Backend part of #5569 

## Checklist


rotkehlchen/chain/aggregator.py
```diff
@@ -1339,19 +1359,57 @@ def add_accounts_to_all_evm(
...
+        # Distinguish between contracts and EOAs
+        for account in accounts:
+            if self.is_contract(account, SupportedBlockchain.ETHEREUM):
+                added_chains = self.track_evm_address(account, [SupportedBlockchain.ETHEREUM])
+                if len(added_chains) == 1:  # Is always either 1 or 0 since is only for ethereum
+                    added_accounts.append((SupportedBlockchain.ETHEREUM, account))
+            else:
+                added_accounts += self.check_chains_and_add_accounts(
+                    account=account,
+                    chains=list(typing.get_args(SUPPORTED_EVM_CHAINS)),
```
yabirgb, line 1372: say that I already have the account in ethereum and optimism... would it try to check them again? I understand that it is what will happen. Maybe you can exclude the chains where this address has been already added?

rotkehlchen/chain/aggregator.py
```diff
@@ -1339,19 +1359,57 @@ def add_accounts_to_all_evm(
...
+
+        added_accounts: list[tuple[SUPPORTED_EVM_CHAINS, ChecksumEvmAddress]] = []
+        for account, chains in current_accounts.items():
+            if progress_handler is not None:
+                progress_handler.new_step(f'Checking {account} EVM chain activity')
+
+            if self.is_contract(account, chains[0]):
+                continue  # do not check contracts
+
+            chains_to_check = list(set(typing.get_args(SUPPORTED_EVM_CHAINS)) - set(chains))
```
yabirgb, line 1396: is a bit of a nitpick but you can move `set(typing.get_args(SUPPORTED_EVM_CHAINS)` out of the for loop

rotkehlchen/chain/aggregator.py
```diff
@@ -1286,50 +1286,70 @@ def get_evm_manager(
...
+            added_chains.append(chain)
+
+        return added_chains
+
+    def check_chains_and_add_accounts(
+            self,
+            account: ChecksumEvmAddress,
+            chains: list[SUPPORTED_EVM_CHAINS],
+    ) -> list[tuple[SUPPORTED_EVM_CHAINS, ChecksumEvmAddress]]:
+        active_chains = self.check_single_address_activity(
```
yabirgb, line 1342: missing docstring

rotkehlchen/chain/aggregator.py
```diff
@@ -1286,50 +1286,70 @@ def get_evm_manager(
...
-                elif chain == SupportedBlockchain.OPTIMISM and chain_manager.node_inquirer.etherscan.has_activity(account) is False:  # noqa: E501
-                    continue
-                # else can only be ethereum so add it unconditionally
-                filtered_accounts.append((chain, account))
+    def track_evm_address(
+            self,
+            address: ChecksumEvmAddress,
+            chains: list[SUPPORTED_EVM_CHAINS],
+    ) -> list[SUPPORTED_EVM_CHAINS]:
+        added_chains = []
```
yabirgb, line 1321: missing  docstring

rotkehlchen/chain/aggregator.py
```diff
@@ -1286,50 +1286,70 @@ def get_evm_manager(
...
+            address=account,
+            chains=chains,
+        )
+        added_chains = self.track_evm_address(account, active_chains)
+
+        added_accounts = []
+        for chain in added_chains:
+            added_accounts.append((chain, account))
+
+        return added_accounts
```
yabirgb, line 1352:
```suggestion
        return [(chain, account) for chain in added_chains]
```

rotkehlchen/chain/aggregator.py
```diff
@@ -1286,50 +1286,70 @@ def get_evm_manager(
...
+    def check_chains_and_add_accounts(
+            self,
+            account: ChecksumEvmAddress,
+            chains: list[SUPPORTED_EVM_CHAINS],
+    ) -> list[tuple[SUPPORTED_EVM_CHAINS, ChecksumEvmAddress]]:
+        active_chains = self.check_single_address_activity(
+            address=account,
+            chains=chains,
+        )
+        added_chains = self.track_evm_address(account, active_chains)
```
yabirgb, line 1346: added_chains was confusing to me since I wasn't sure what to expect here. Maybe you can use `new_tracked_chains`?

rotkehlchen/data_migrations/migrations/migration_8.py
```diff
@@ -30,36 +27,7 @@ def data_migration_8(rotki: 'Rotkehlchen', progress_handler: 'MigrationProgressH
...
-        with rotki.data.db.user_write() as write_cursor:
-            for chain, account in to_add_accounts:  # add them to the DB
-                rotki.data.db.add_blockchain_accounts(
-                    write_cursor=write_cursor,
-                    account_data=[BlockchainAccountData(
-                        chain=chain,
-                        address=account,
-                    )],  # not duplicating label and tags as it's chain specific
-                )
+        to_add_accounts = chains_aggregator.detect_evm_accounts(progress_handler)
```
yabirgb, line 30:
```suggestion
        new_tracked_accounts = chains_aggregator.detect_evm_accounts(progress_handler)
```

rotkehlchen/chain/aggregator.py
```diff
@@ -1339,19 +1359,57 @@ def add_accounts_to_all_evm(
...
+            progress_handler.new_step('Potentially write migrated addresses to the DB')
+
+        with self.database.user_write() as write_cursor:
+            self.database.add_blockchain_accounts(
+                write_cursor=write_cursor,
+                account_data=[
+                    BlockchainAccountData(chain, account)
+                    for chain, account in added_accounts
+                ],
+            )
```
yabirgb, line 1412: keep the comment since it helps later to remember why things were done

```suggestion
        with self.database.user_write() as write_cursor:
            self.database.add_blockchain_accounts(
                write_cursor=write_cursor,
                account_data=[
                    BlockchainAccountData(chain, account)
                    for chain, account in added_accounts
                ],    # not duplicating label and tags as it's chain specific
            )
```

rotkehlchen/chain/aggregator.py
```diff
@@ -1339,19 +1359,57 @@ def add_accounts_to_all_evm(
...
-        added_accounts = []
-        for chain, account in filtered_accounts:
-            try:
-                self.modify_blockchain_accounts(
-                    blockchain=chain,
-                    accounts=[account],
-                    append_or_remove='append',
+        added_accounts: list[tuple[SUPPORTED_EVM_CHAINS, ChecksumEvmAddress]] = []
+        # Distinguish between contracts and EOAs
+        for account in accounts:
```
yabirgb, line 1364: in the docstring what is the filter you mention? Also move the comment about contracts and ethereum from where you took this logic

rotkehlchen/tests/unit/test_chains_aggregator.py
```diff
@@ -26,42 +27,81 @@ def test_module_deactivation(blockchain):
         assert module_name not in blockchain.eth_modules
 
 
-def test_filter_active_evm_addresses(blockchain):
+@pytest.mark.parametrize('ethereum_accounts', [[]])
+def test_detect_evm_accounts(blockchain: 'ChainsAggregator'):
+    eth_addy_contract = make_evm_address()
```
yabirgb, line 32: Can you add a docstring explaining what is getting tested?

```suggestion
def test_detect_evm_accounts(blockchain: 'ChainsAggregator') -> None:
    eth_addy_contract = make_evm_address()
```

rotkehlchen/tests/unit/test_chains_aggregator.py
```diff
@@ -26,42 +27,81 @@ def test_module_deactivation(blockchain):
...
+                chain=SupportedBlockchain.OPTIMISM,
+                address=account,
+            ))
+        for account in raw_accounts.avax:
+            accounts_in_db.append(BlockchainAccountData(
+                chain=SupportedBlockchain.AVALANCHE,
+                address=account,
+            ))
+
+    assert set(accounts_in_db) == set(expected_accounts_data)
```
yabirgb, line 107: is the set needed? The db result should be always in the same order right? if it is because they are mixed can you assert that there is the same number of results and set didn't remove any duplicate that might have been introduced?

rotkehlchen/tests/utils/blockchain.py
```diff
@@ -603,15 +603,16 @@ def set_web3_in_inquirer(ethereum_inquirer: 'EthereumInquirer', web3: Web3) -> N
...
         chains_aggregator: 'ChainsAggregator',
-        contract_addresses: list[ChecksumEvmAddress],
+        eth_contract_addresses: list[ChecksumEvmAddress],
+        ethereum_addresses: list[ChecksumEvmAddress],
         avalanche_addresses: list[ChecksumEvmAddress],
         optimism_addresses: list[ChecksumEvmAddress],
 ) -> 'ExitStack':
     def mock_ethereum_get_code(account):
-        if account in contract_addresses:
+        if account in eth_contract_addresses:
```
yabirgb, line 615: why renaming to `eth_contract_addresses`?

rotkehlchen/tests/unit/test_chains_aggregator.py
```diff
@@ -26,42 +27,81 @@ def test_module_deactivation(blockchain):
...
+    ]
 
-    contract_addy = make_evm_address()
-    all_addy = make_evm_address()
-    optimism_addy = make_evm_address()
-    avalanche_addy = make_evm_address()
+    blockchain.modify_blockchain_accounts(
+        blockchain=SupportedBlockchain.ETHEREUM,
+        accounts=[eth_addy_contract],
+        append_or_remove='append',
```
yabirgb, line 48: can you also test and address in ethereum that is not a contract? since ethereum is a case that has different paths in the logic

rotkehlchen/tests/unit/test_chains_aggregator.py
```diff
@@ -26,42 +27,81 @@ def test_module_deactivation(blockchain):
...
+    )
+    blockchain.modify_blockchain_accounts(
+        blockchain=SupportedBlockchain.OPTIMISM,
+        accounts=[addy_eoa],
+        append_or_remove='append',
+    )
```
yabirgb, line 54: Also can you add to the test the case for an address that is already present in all the tracked chains?

yabirgb (commented):
Thanks! A few comments but important:

- Update the docs
- Check the failing test

rotkehlchen/chain/aggregator.py
nebolax: Makes sense, changed that

rotkehlchen/tests/unit/test_chains_aggregator.py
nebolax: They are mixed because one is all initial addresses and then the ones that were detected, while the other is eth addresses, op addresses, avax addresses. Will add a length assertion

rotkehlchen/tests/utils/blockchain.py
nebolax: To add more sense to the name. Because it is actually for eth contracts only, not for contracts on all chains

rotkehlchen/tests/unit/test_chains_aggregator.py
nebolax: Makes sense, thanks

rotkehlchen/chain/aggregator.py
```diff
@@ -1286,72 +1287,157 @@ def get_evm_manager(
...
+                    BlockchainAccountData(chain, account)
+                    for chain, account in added_accounts
+                ],  # not duplicating label and tags as it's chain specific
+            )
+
+        with self.database.user_write() as cursor:
+            cursor.execute(
+                'INSERT OR REPLACE INTO multisettings (name, value) VALUES (?, ?)',
+                (EVM_ACCOUNTS_DETECT_KEY, str(ts_now())),
+            )
```
yabirgb, line 1440: I made the same mistake, if you look at this table name and value are not primary keys so this will keep inserting rows with the same name instead of keeping only one. We said to use the settings table for now and then we can create a new table for this and clean the settings table

docs/api.rst
```diff
@@ -9233,6 +9233,46 @@ Adding EVM accounts to all EVM chains
...
+            "AVAX": ["0x78b0AD50E768D2376C6BA7de33F426ecE4e03e0B"],
+            "OPTIMISM": ["0x78b0AD50E768D2376C6BA7de33F426ecE4e03e0B"]
+        },
+        "message": ""
+      }
+
+   :resjson list result: A mapping containing the evm chain keys and which addresses were added for each chain.
+   :statuscode 200: Accounts successfully added
+   :statuscode 400: Provided JSON is in some way malformed.
+   :statuscode 409: User is not logged in.
```
yabirgb, line 9272: It can happen also if there is an issue with the node (`EthSyncError`)? The other case with the tags I believe doesn't happen here

docs/api.rst
```diff
@@ -9233,6 +9233,46 @@ Adding EVM accounts to all EVM chains
...
+   :statuscode 500: Internal rotki error
+
```
yabirgb, line 9274: This is missing bad gateway (502) that happens when there is a remote error

docs/api.rst
```diff
@@ -9233,6 +9233,46 @@ Adding EVM accounts to all EVM chains
...
+   .. http:example:: curl wget httpie python-requests
+
+      POST /api/1/blockchains/evm/accounts HTTP/1.1
+      Host: localhost:5042
+
+      {}
+
+      .. note::
+        This endpoint can also be queried asynchronously by using ``"async_query": true``
```
yabirgb, line 9252: You have written the note inside the http request 

image

rotkehlchen/chain/aggregator.py
```diff
@@ -1286,72 +1287,157 @@ def get_evm_manager(
...
+            if self.is_contract(account, SupportedBlockchain.ETHEREUM):
+                added_chains = self.track_evm_address(account, [SupportedBlockchain.ETHEREUM])
+                if len(added_chains) == 1:  # Is always either 1 or 0 since is only for ethereum
+                    added_accounts.append((SupportedBlockchain.ETHEREUM, account))
+            else:
+                chains_to_check = []
+                for chain in typing.get_args(SUPPORTED_EVM_CHAINS):
+                    if account in self.accounts.get(chain):
+                        continue
+                    chains_to_check.append(chain)
```
yabirgb, line 1380:
```suggestion
                    if account not in self.accounts.get(chain):
                        chains_to_check.append(chain)
```

rotkehlchen/tests/unit/test_chains_aggregator.py
```diff
@@ -26,42 +27,102 @@ def test_module_deactivation(blockchain):
...
+    everywhere_addy = make_evm_address()
+
+    initial_accounts_data = []
+    addies_to_add = [
+        (SupportedBlockchain.ETHEREUM, eth_addy_contract),
+        (SupportedBlockchain.OPTIMISM, addy_eoa_1),
+        (SupportedBlockchain.ETHEREUM, addy_eoa_2),
+        (SupportedBlockchain.ETHEREUM, everywhere_addy),
+        (SupportedBlockchain.OPTIMISM, everywhere_addy),
+        (SupportedBlockchain.AVALANCHE, everywhere_addy),
```
yabirgb, line 58: I was thinking to loop every evm chain here but this test will need adjustment anyway everytime we add a new evm chain

rotkehlchen/chain/ethereum/utils.py
```diff
@@ -163,3 +169,18 @@ def should_update_protocol_cache(cache_key: GeneralCacheType) -> bool:
         key_parts=[cache_key],
     )
     return ts_now() - last_update_ts >= ETH_PROTOCOLS_CACHE_REFRESH
+
+
+def should_detect_evm_accounts(database: 'DBHandler') -> bool:
+    """Checks if the last time we detected EVM accounts is far enough to trigger
+    the process of querying it again.
+    """
```
yabirgb, line 177: This function is in a strange file. Maybe it makes more sense to move it to a utils file in the tasks folder and the `EVM_ACCOUNTS_DETECT_KEY` to a constant there?

rotkehlchen/chain/aggregator.py
```diff
@@ -1286,72 +1287,157 @@ def get_evm_manager(
...
+        return active_chains
+
+    def track_evm_address(
+            self,
+            address: ChecksumEvmAddress,
+            chains: list[SUPPORTED_EVM_CHAINS],
+    ) -> list[SUPPORTED_EVM_CHAINS]:
+        """
+        Tried to track address for the chains provided. If the address is already tracked on a
+        chain, skips this chain. Returns a list of chains where the address was added successfully.
```
yabirgb, line 1324:
```suggestion
        Track address for the chains provided. If the address is already tracked on a
        chain, skips this chain. Returns a list of chains where the address was added successfully.
```

yabirgb (commented):
Thanks! A few more comments

docs/api.rst
nebolax: lol. thanks

docs/api.rst
nebolax: Yes, thanks

rotkehlchen/chain/aggregator.py
nebolax: Thanks for pointing it out!

rotkehlchen/chain/ethereum/utils.py
nebolax: Makes sense, thanks. I put it there because saw that `should_update_curve_pools` is there, but storing in the task manager directory perhaps makes more sense

rotkehlchen/tasks/manager.py
```diff
@@ -586,6 +589,19 @@ def _maybe_update_yearn_vaults(self) -> Optional[list[gevent.Greenlet]]:
 
         return None
 
+    def _maybe_detect_evm_accounts(self) -> Optional[list[gevent.Greenlet]]:
+        """Function that schedules EVM accounts detection task if either there is no EVM accounts
+        cache yet or this cache has expired (i.e. it's been more than a week since last update)."""
+        if should_detect_evm_accounts(self.database) is False:
+            return None
```
yabirgb, line 596: Sorry I didn't notice before as I just noticed while working in my branch. This hasn't been included in the potential tasks list

yabirgb (commented):
Just one issue that I found

rotkehlchen/tasks/manager.py
nebolax: Ah wow, thanks

rotkehlchen/tests/unit/test_chains_aggregator.py
```diff
@@ -26,42 +27,102 @@ def test_module_deactivation(blockchain):
...
 
 
-def test_filter_active_evm_addresses(blockchain):
+@pytest.mark.parametrize('ethereum_accounts', [[]])
+def test_detect_evm_accounts(blockchain: 'ChainsAggregator') -> None:
+    """
+    Tests that the detection of EVM accounts activity in chains where they are not tracked yet
+    works as expected.
+    """
+    # Is a contract in ethereum mainnet and should not be added anywhere else despite it has
```
LefterisJP, line 36:
```suggestion
    # Is a contract in ethereum mainnet and should not be added anywhere else despite it having
```

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -67,6 +67,7 @@ def fixture_task_manager(
         chains_aggregator=blockchain,
         exchange_manager=exchange_manager,
         deactivate_premium=lambda: None,
+        detect_evm_accounts=lambda: None,
```
LefterisJP, line 70: can go away
```suggestion
```

docs/api.rst
```diff
@@ -9233,6 +9233,46 @@ Adding EVM accounts to all EVM chains
...
+
+   **Example Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+        "result": {
```
LefterisJP, line 9261: I am not sure you read the issue correctly. Returning the accounts is one thing, which may or may not be needed but the important thing was the websocket messages.

LefterisJP (commented):
So I think there is a big misunderstanding about the implementation here.

In the issue it mentions to abstract the functionality of the migration. It also mentions websockets. Specifically. This entire functionality depends on them.

You on the other hand seem to have kept the completely out of the logic you moved around and left them only in the migration.

This won't work and this is not what the issue was asking.

There is two things we need.

1. At manual redetection, you return the detected chains. I am not even sure this is needed but let's ask @rotki/frontend. Is it? Wouldn't it make more sense for you guys to let the task run and just read the websockets?
2. At the periodic redetection there is no way for the frontend (or anyone) to know when it ran, and what it has found and added to the DB if anything. This is why the issue insisted on websockets.

Ahh ... and also add a changelog for this!

rotkehlchen/chain/aggregator.py
```diff
@@ -1286,72 +1287,156 @@ def get_evm_manager(
...
+            for account in chain_accounts:
+                current_accounts[account].append(chain)
+
+        all_evm_chains = set(typing.get_args(SUPPORTED_EVM_CHAINS))
+        added_accounts: list[tuple[SUPPORTED_EVM_CHAINS, ChecksumEvmAddress]] = []
+        for account, chains in current_accounts.items():
+            if progress_handler is not None:
+                progress_handler.new_step(f'Checking {account} EVM chain activity')
+
+            if self.is_contract(account, chains[0]):
```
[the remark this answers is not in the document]
nebolax: Will also add it as a comment, but no, not assuming chain[0] being ETH.

I rather assume that if an address is a contract in one chain, it will be a contract in all other chains as well.

And thus checking only the first chain.

rotkehlchen/constants/timing.py
```diff
@@ -21,3 +21,4 @@
 ENS_UPDATE_INTERVAL = DAY_IN_SECONDS
 
 ETH_PROTOCOLS_CACHE_REFRESH = DAY_IN_SECONDS * 3
+EVM_ACCOUNTS_REDETECT = DAY_IN_SECONDS
```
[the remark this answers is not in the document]
nebolax: Makes sense, thanks

nebolax:
> Ahh ... and also add a changelog for this!

And of course I forgot about it!

Will add in the next iteration, after a review.

rotkehlchen/tasks/utils.py
```diff
@@ -22,3 +25,18 @@ def should_check_data_updates(database: 'DBHandler') -> bool:
 
     last_update_ts = deserialize_timestamp(timestamp_in_db)
     return ts_now() - last_update_ts >= DATA_UPDATES_REFRESH
+
+
+def should_detect_evm_accounts(database: 'DBHandler') -> bool:
```
LefterisJP, line 30: nope as I said to both you and @yabirgb you should combine those two functions. They are identical. The only thing missing is the arguments: `key_name`, `refresh_period`

docs/changelog.rst
```diff
@@ -3,6 +3,7 @@ Changelog
 =========
 
 * :release:`1.28.0`
+* :feature:`5569` Make it possible for the users to initiate evm accounts detection and add a periodic task to do it automatically.
```
LefterisJP, line 6: A user does not understand what this means. This is addressed to a user, not a dev.

```suggestion
* :feature:`5569` It is now possible to detect when an evm address is used in other chains and start tracking it. Also a background task runs periodically that performs this check.
```

docs/websockets.rst
```diff
@@ -125,22 +125,28 @@ The messages sent by rotki when a user is logging in and a db upgrade is happeni
...
-EVM Addresses Migrations
-============================
+EVM Accounts Detection
+=======================
+
+If there are new evm accounts detected (due to a data migration, user-initiated detection, periodic detection, etc.) then the frontend will be notified of any new EVM accounts that were detected. This is done via the following message.
 
-At the user DB migrations when a new evm chain is introduced rotki will do a migration that will add any addresses used in mainnet to the new evm chain. At the same time it needs to notify the frontend that new evm chain/ evm address combination was added at migration so the frontend can do extra actions such as detecting tokens. The message is simple and just contains the list of migrated addresses if any.
+- ``"address"``: Address of a account.
+- ``"evm_chain"``: Chain of a account.
```
LefterisJP, line 134: Looking at other messsage docs in this same file these are without the `""`and they come **after** the message. Please be consistent.

docs/websockets.rst
```diff
@@ -125,22 +125,28 @@ The messages sent by rotki when a user is logging in and a db upgrade is happeni
...
     {
-        "type": "evm_address_migration",
-        "data": [{
-            "evm_chain": "optimism",
-            "address": "0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12"
-        }, {
-            "evm_chain": "optimism",
-            "address": "0xFeebabE6b0418eC13b30aAdF129F5DcDd4f70CeA"
-        }]
+        "type": "new_evm_accounts_detected",
```
LefterisJP, line 139: rename: evm_accounts_detection

Keeping the names consistent with the previous one.

rotkehlchen/api/websockets/typedefs.py
```diff
@@ -15,7 +15,7 @@ class WSMessageType(Enum):
     PREMIUM_STATUS_UPDATE = auto()
     DB_UPGRADE_STATUS = auto()
     # Used for evm address migration after new chain integration
-    EVM_ADDRESS_MIGRATION = auto()
+    NEW_EVM_ACCOUNTS_DETECTED = auto()
```
LefterisJP, line 18: re consistency as per the previous comment

```suggestion
    EVM_ACCOUNTS_DETECTION = auto()
```

rotkehlchen/chain/aggregator.py
```diff
@@ -1286,72 +1289,164 @@ def get_evm_manager(
...
+    def detect_evm_accounts(
+            self,
+            progress_handler: Optional['ProgressUpdater'] = None,
+    ) -> list[tuple[SUPPORTED_EVM_CHAINS, ChecksumEvmAddress]]:
+        """
+        Detects user's EVM accounts on different chains and adds them to the tracked accounts.
+        1. Iterates through already added addresses
+        2. Checks whether the address is the chain that it is currently added to
+        3. If it is not, checks whether the address is active in a chain
+        4. If it is active, adds it to the tracked accounts on that chain
```
LefterisJP, line 1399: From bullet point 2 and on it does not make sense.

> Checks whether the address is the chain that it is currently added to 
???

 I am rewriting how I understand it.

```suggestion
        2. For each address, assuming it's not a contract, checks which chains it's already in
        3. Get the rest of the chains, and check activity. If active in any of them it tracks the address for that chain.
```

rotkehlchen/tasks/manager.py
```diff
@@ -608,6 +612,21 @@ def _maybe_check_data_updates(self) -> Optional[list[gevent.Greenlet]]:
             method=self.data_updater.check_for_updates,
         )]
 
+    def _maybe_detect_evm_accounts(self) -> Optional[list[gevent.Greenlet]]:
+        """
+        Function that schedules the EVM accounts detection task if there has been more than
+        EVM_ACCOUNTS_DETECTION_REFRESH since the last time it ran.
```
LefterisJP, line 618:
```suggestion
        EVM_ACCOUNTS_DETECTION_REFRESH second since the last time it ran.
```

rotkehlchen/tasks/utils.py
```diff
@@ -1,24 +1,22 @@
...
+
 from rotkehlchen.serialization.deserialize import deserialize_timestamp
 from rotkehlchen.utils.misc import ts_now
 
 if TYPE_CHECKING:
     from rotkehlchen.db.dbhandler import DBHandler
 
 
-def should_check_data_updates(database: 'DBHandler') -> bool:
+def should_run_periodic_task(database: 'DBHandler', key_name: str, refresh_period: int) -> bool:
```
LefterisJP, line 10: you can limit the type at `key_name`. We know all of them so you can use `Literal`

LefterisJP (commented):
Some comments that need addressing. We need to remind to the frontend that the websocket message for the migration changes name.

docs/websockets.rst
nebolax: I took this format from `EVM Token Detection` section. So I adjusted that section too.

rotkehlchen/chain/aggregator.py
nebolax: Thanks

docs/websockets.rst
LefterisJP: you have not renamed this here yet

docs/websockets.rst
```diff
@@ -125,39 +125,43 @@ The messages sent by rotki when a user is logging in and a db upgrade is happeni
...
+            },
+            {
+                "address": "0x4bBa290826C253BD854121346c370a9886d1bC26",
+                "evm_chain": "avalanche"
+            }
+        ]
     }
 
 
+- ``address``: Address of a account.
```
LefterisJP, line 150:
```suggestion
- ``address``: Address of an account.
```

docs/websockets.rst
```diff
@@ -125,39 +125,43 @@ The messages sent by rotki when a user is logging in and a db upgrade is happeni
...
+- ``evm_chain``: Chain of a account.
```
LefterisJP, line 151:
```suggestion
- ``evm_chain``: Chain of an account.
```

rotkehlchen/db/updates.py
```diff
@@ -21,7 +21,8 @@
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
-LAST_DATA_UPDATES_KEY = 'last_data_updates_ts'
+LAST_DATA_UPDATES_KEY_TYPE = Literal['last_data_updates_ts']
```
LefterisJP, line 24: no no you don't need to do that. use the `Final` modifie. It makes it constant

rotkehlchen/tasks/utils.py
```diff
@@ -1,24 +1,28 @@
...
-def should_check_data_updates(database: 'DBHandler') -> bool:
+def should_run_periodic_task(
+        database: 'DBHandler',
+        key_name: Literal[LAST_DATA_UPDATES_KEY_TYPE, LAST_EVM_ACCOUNTS_DETECT_KEY_TYPE],
+        refresh_period: int,
+) -> bool:
     """
-    Checks if the last time we checked data updates is far enough to trigger
-    the process of querying it again.
+    Checks if there has enough time elapsed since the last run of a periodic task to run it again.
```
LefterisJP, line 18:
```suggestion
    Checks if enough time has elapsed since the last run of a periodic task in order to run it again.
```

LefterisJP (approved):
LGTM. I think both test failures are due to etherscan failures.  PRobably could also check for etherscan errors in the test itself

PR #5605, opened 2023-02-14
Use branching for vcr

yabirgb:
This PR splits VCR tests in two branches:

- develop
- bugfixes

develop is in sync with develop at rotki/rotki and bugfixes with the bugfixes branch in the same repo.

By default it will try to checkout the branch with the same name. Say that you are working in the branch called `workon-4444`, the script will try to checkout branch `workon-4444` in the cassettes cache. If it fails it will default to bugfixes branch and this behavior can be modified by the usage of the `DEFAULT_VCR_BRANCH` env var

### DIFF
--- a/rotkehlchen/tests/conftest.py
+++ b/rotkehlchen/tests/conftest.py
@@ -10,6 +10,7 @@ from enum import auto
 from http import HTTPStatus
 from json import JSONDecodeError
 from pathlib import Path
+from subprocess import PIPE, Popen
 from typing import TYPE_CHECKING, Any, Optional
 
 import py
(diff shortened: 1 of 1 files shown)


rotkehlchen/tests/conftest.py
```diff
@@ -241,7 +242,19 @@ def vcr_cassette_dir(request: pytest.FixtureRequest) -> str:
             log.debug('VCR setup completed')
         else:
             log.debug('Pulling information for VCR cassettes')
-            os.popen(f'cd {tests_caching_dir} && git pull').read()
+            current_branch = os.popen('git rev-parse --abbrev-ref HEAD').read().rstrip('\n')
+            checkout_proc = Popen(f'cd {tests_caching_dir} && git checkout {current_branch}', shell=True, stdout=PIPE, stderr=PIPE)  # noqa: E501
```
LefterisJP, line 246: Shouldn't it try to first fetch/pull from remote? It could be it has the branch but the local does not

LefterisJP, line 246: you should make this behavior also configurable. So with maybe another env variable to set the currently checking out vcr branch?

rotkehlchen/tests/conftest.py
```diff
@@ -241,7 +242,19 @@ def vcr_cassette_dir(request: pytest.FixtureRequest) -> str:
...
+            _, stderr = checkout_proc.communicate(timeout=10)
+
+            if (
+                len(stderr) != 0 and
+                b'Already on' not in stderr and
+                b'Switched to branch' not in stderr
+            ):
+                default_branch = os.environ.get('DEFAULT_VCR_BRANCH', 'bugfixes')
```
LefterisJP, line 254: Defaulting to develop always would probably be best as I think the majority of PRs in the long run are against develop.

Also ... couldn't you somehow find the "parent" branch of a given branch? That would be the perfect solution to choose here automatically. https://stackoverflow.com/questions/3161204/how-to-find-the-nearest-parent-of-a-git-branch

I tried the solution listed there but got the `cannot handle more than 26 refs` so I did not look further.

LefterisJP (commented):
Guess should be fine.

rotkehlchen/tests/conftest.py
yabirgb: I saw that approach but for me it failed also with a similar error. I believe that it fails for the people using the tool then is not good

LefterisJP (approved):
LGTM. We probably need the docs to change to mention these too

PR #5608, opened 2023-02-15
Feat: add the ability to save and restore filters

### LINKED ISSUE #5588: Add the ability to save and restore filters
## Abstract

We want to introduce the ability to save and quickly restore filters in the table filter. 

## Motivation

Allows the user to quickly apply often-used filters without having to insert the details every time.

## Specification

Filters should be saved per location (trades/deposits & withdrawals/evm transactions/ledger actions) see and identify any other places where it would make sense.

- [ ] We need a button to save a filter
- [ ] A menu to manage (delete) or apply a filter

As for where to persist the values, one option is local storage which has the problem of making the filters local only to the application/browser window etc.

Another would be to save them in the user database in order to make them portable.
You move your user database and the filters follow you.

If we want to do this without any work on the backend being done, the only place for this is the frontend settings JSON. 

If we go this way we should add a limit to the max number of saved filters per location to avoid this growing too large and creating problems (maybe 10 max entries per location). 

@LefterisJP if you have any thoughts on this chime in.

lukicenturi:
Closes #5588

### DIFF
--- a/frontend/app/src/components/history/ClosedTrades.vue
+++ b/frontend/app/src/components/history/ClosedTrades.vue
@@ -24,6 +24,7 @@ import { type Collection } from '@/types/collection';
 import { defaultCollectionState } from '@/utils/collection';
 import { assert } from '@/utils/assertions';
 import { defaultOptions } from '@/utils/history';
+import { SavedFilterLocation } from '@/types/filtering';
 
 const props = withDefaults(
   defineProps<{
@@ -330,6 +331,7 @@ onMounted(async () => {
     newExternalTrade();
     await router.replace({ query: {} });
   } else {
+    await fetchData();
     await refreshTrades();
   }
 });
@@ -464,6 +466,7 @@ watch(loading, async (isLoading, wasLoading) => {
               <table-filter
                 :matches="filters"
                 :matchers="matchers"
+                :location="SavedFilterLocation.HISTORY_TRADES"
                 @update:matches="setFilter($event)"
               />
             </div>
--- a/frontend/app/src/components/history/transactions/TransactionContent.vue
+++ b/frontend/app/src/components/history/transactions/TransactionContent.vue
@@ -23,8 +23,6 @@ import { Section } from '@/types/status';
 import { TaskType } from '@/types/task-type';
 import { defaultCollectionState, getCollectionData } from '@/utils/collection';
 import { IgnoreActionType } from '@/types/history/ignored';
-import EvmChainIcon from '@/components/helper/display/icons/EvmChainIcon.vue';
-import AdaptiveWrapper from '@/components/display/AdaptiveWrapper.vue';
 import { type TablePagination } from '@/types/pagination';
 import {
   type LocationQuery,
@@ -35,6 +33,7 @@ import { type Collection } from '@/types/collection';
 import { type ActionStatus } from '@/types/action';
 import { assert } from '@/utils/assertions';
 import { defaultOptions } from '@/utils/history';
+import { SavedFilterLocation } from '@/types/filtering';
 
 const props = withDefaults(
   defineProps<{
@@ -496,7 +495,10 @@ const showDeleteConfirmation = () => {
 const { txEvmChains, getEvmChainName, getChain } = useSupportedChains();
 const txChains = useArrayMap(txEvmChains, x => x.id);
 
-onMounted(async () => refreshTransactions());
+onMounted(async () => {
+  await fetchData();
+  await refreshTransactions();
+});
 
 watch(loading, async (isLoading, wasLoading) => {
   if (!isLoading && wasLoading) {
@@ -564,6 +566,7 @@ watch(loading, async (isLoading, wasLoading) => {
               <table-filter
                 :matches="filters"
                 :matchers="matchers"
+                :location="SavedFilterLocation.HISTORY_TRANSACTIONS"
                 @update:matches="setFilter($event)"
               />
             </div>
--- a/frontend/app/src/types/filtering.ts
+++ b/frontend/app/src/types/filtering.ts
@@ -1,6 +1,7 @@
 import { type ComputedRef, type Ref } from 'vue';
 import { type AssetInfo } from '@rotki/common/lib/data';
-import { type AssetInfoWithId, type AssetsWithId } from '@/types/asset';
+import { z } from 'zod';
+import { AssetInfoWithId, type AssetsWithId } from '@/types/asset';
 import { convertFromTimestamp, convertToTimestamp } from '@/utils/date';
 import { type DateFormat } from '@/types/date-format';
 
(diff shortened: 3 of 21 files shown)


kelsos:
@lukicenturi now that we merged the work for the history this PR needs to be adjusted and rebased on top of develop since it has a lot of conflicts.

frontend/app/src/types/filtering.ts
```diff
@@ -37,12 +38,21 @@ export type MatchedKeyword<T extends string> = {
...
+  value: AssetInfoWithId.or(z.string())
+});
+
+export type Suggestion = z.infer<typeof Suggestion>;
+
+export enum SavedFilterLocation {
+  HISTORY_TRADES = 'HISTORY-TRADES',
+  HISTORY_DEPOSITS_WITHDRAWALS = 'HISTORY-DEPOSITS-WITHDRAWALS',
+  HISTORY_TRANSACTIONS = 'HISTORY-TRANSACTIONS',
+  HISTORY_LEDGER_ACTIONS = 'HISTORY-LEDGER-ACTIONS'
```
kelsos, line 55: Are these values in the enum going to be serialized to JSON?

lukicenturi: hmm wdym serialized to JSON?

kelsos: image

kelsos: Btw for the route history only the `key` and `value` is stored, why do we need to serialize everything here?

kelsos: I need to think a bit about it but generally, our approach here should be to store the least possible amount of information that can help us restore the filter.

frontend/app/src/components/history/ClosedTrades.vue
```diff
@@ -155,7 +156,10 @@ const {
   state: trades,
   execute
 } = useAsyncState<Collection<TradeEntry>>(
-  args => fetchTrades(args),
+  async args => {
+    if (!args) return defaultCollectionState<TradeEntry>();
```
kelsos, line 160: do we still have a case where args is empty? do you know when this happens?

frontend/app/src/components/history/transactions/TransactionContent.vue
```diff
@@ -495,7 +497,12 @@ const showDeleteConfirmation = () => {
 const { txEvmChains, getEvmChainName, getChain } = useSupportedChains();
 const txChains = useArrayMap(txEvmChains, x => x.id);
 
-onMounted(async () => refreshTransactions());
+onMounted(async () => {
+  await fetchData();
+  if (await refreshTransactions()) {
+    await fetchData();
+  }
```
kelsos, line 504: Shouldn't loading finishing cover this?https://github.com/rotki/rotki/blob/4637dcbd727aba1740f6fd217f7722f2f4c02888/frontend/app/src/components/history/transactions/TransactionContent.vue#L500

frontend/app/src/types/filtering.ts
lukicenturi: okay, let me see how I can reduce the information stored there

frontend/app/src/components/history/transactions/TransactionContent.vue
lukicenturi: Yeah, my bad

frontend/app/src/components/history/ClosedTrades.vue
lukicenturi: so for the first call, the args is always `undefined`.
I console something inside the `fetchData` but it's not called first.
So I think the useAsyncState, that initiate the first call.

frontend/app/src/types/filtering.ts
lukicenturi: > image

I thought I could get away from the conversion by using kebab case

lukicenturi: > 

Yeah I think we can store it like that, the ones with multiple usages will be stored as array instead single string (for the value)

frontend/app/src/components/history/ClosedTrades.vue
kelsos: https://github.com/vueuse/vueuse/blob/941392cd60c187e757bdf3f1b7445f072b96db0e/packages/core/useAsyncState/index.ts#L29

Ok it seems that by default execute is immediate and it happens automatically.
But instead of doing this we can make it `immediate: false` and call it manually

kelsos: In general, it is always good to check the documentation, and if that is not listing everything also the source code.

lukicenturi:
> Another thing, how do you think the buttons would look like if we added them next to the search fields instead of the dropdown?
> 
> I would have expected the thing to be more easily accessible. You should be able to quickly save/restore the saved items.

Yeah I can try to arrange them next to the search fields

lukicenturi:
> > Another thing, how do you think the buttons would look like if we added them next to the search fields instead of the dropdown?
> > I would have expected the thing to be more easily accessible. You should be able to quickly save/restore the saved items.
> 
> Yeah I can try to arrange them next to the search fields

I think all are updated, can you check again? @kelsos

kelsos:
@lukicenturi can we somehow make the schema parsing for this to fall back to the defaults if the saved filter schema is not valid?

I have the old filters stored from the last testing and this is actually making the login fail...

lukicenturi:
> @lukicenturi can we somehow make the schema parsing for this to fall back to the defaults if the saved filter schema is not valid?
> 
> I have the old filters stored from the last testing and this is actually making the login fail...

yeah2 it will be make sense

kelsos:
> > @lukicenturi can we somehow make the schema parsing for this to fall back to the defaults if the saved filter schema is not valid?
> > I have the old filters stored from the last testing and this is actually making the login fail...
> 
> yeah2 it will be make sense

We can maybe use a combination of `safeParse` and `transform` for this key?

Also another comment can we use a [popover menu](https://v2.vuetifyjs.com/en/components/menus/#popover-menu) instead of a dialog for the stored filters?

lukicenturi:
> 

I think this works for this purpose (?) https://github.com/colinhacks/zod#catch
Okay moving to popover...

lukicenturi:
> > 
> 
> I think this works for this purpose (?) https://github.com/colinhacks/zod#catch Okay moving to popover...



> > > @lukicenturi can we somehow make the schema parsing for this to fall back to the defaults if the saved filter schema is not valid?
> > > I have the old filters stored from the last testing and this is actually making the login fail...
> > 
> > 
> > yeah2 it will be make sense
> 
> We can maybe use a combination of `safeParse` and `transform` for this key?
> 
> Also another comment can we use a [popover menu](https://v2.vuetifyjs.com/en/components/menus/#popover-menu) instead of a dialog for the stored filters?


I think this works for this purpose (?) https://github.com/colinhacks/zod#catch
Okay moving to popover...

kelsos:
> > I think this works for this purpose (?) https://github.com/colinhacks/zod#catch Okay moving to popover...
> 
> > > > @lukicenturi can we somehow make the schema parsing for this to fall back to the defaults if the saved filter schema is not valid?
> > > > I have the old filters stored from the last testing and this is actually making the login fail...
> > > 
> > > 
> > > yeah2 it will be make sense
> > 
> > 
> > We can maybe use a combination of `safeParse` and `transform` for this key?
> > Also another comment can we use a [popover menu](https://v2.vuetifyjs.com/en/components/menus/#popover-menu) instead of a dialog for the stored filters?
> 
> I think this works for this purpose (?) https://github.com/colinhacks/zod#catch Okay moving to popover...

Great catch and fallback to the default value for the filters.

lukicenturi:
@kelsos updated to popover
image

kelsos (approved):
lgtm

PR #5616, opened 2023-02-15
Fix genesis transactions decoding

nebolax:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/evm/constants.py
+++ b/rotkehlchen/chain/evm/constants.py
@@ -5,4 +5,24 @@ from .types import string_to_evm_address
 MAX_BLOCKTIME_CACHE = 250  # 55 mins with 13 secs avg block time
 ZERO_ADDRESS = string_to_evm_address('0x0000000000000000000000000000000000000000')
 ETH_SPECIAL_ADDRESS = string_to_evm_address('0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE')
-GENESIS_HASH = deserialize_evm_tx_hash('0x' + '0' * 64)  # hash for transactions in genesis block
+ZERO_32_BYTES_HEX = '0x' + '0' * 64
+GENESIS_HASH = deserialize_evm_tx_hash(ZERO_32_BYTES_HEX)  # hash for transactions in genesis block
+
+# Fake receipt with values taken from ethereum mainnet, to emulate a receipt for the
+# genesis transactions
+FAKE_GENESIS_TX_RECEIPT = {
+    'blockHash': ZERO_32_BYTES_HEX,
+    'blockNumber': 0,
+    'contractAddress': None,
+    'cumulativeGasUsed': 21000,
+    'effectiveGasPrice': 50000000000000,
+    'from': ZERO_ADDRESS,
+    'gasUsed': 21000,
+    'logs': [],
+    'logsBloom': '0x' + '0' * 512,
+    'root': ZERO_32_BYTES_HEX,
+    'to': ZERO_ADDRESS,
+    'transactionHash': ZERO_32_BYTES_HEX,
+    'transactionIndex': 0,
+    'type': '0x0',
+}
--- a/rotkehlchen/chain/evm/node_inquirer.py
+++ b/rotkehlchen/chain/evm/node_inquirer.py
@@ -27,6 +27,7 @@ from web3.types import BlockIdentifier, FilterParams
 from rotkehlchen.chain.constants import DEFAULT_EVM_RPC_TIMEOUT
 from rotkehlchen.chain.ethereum.constants import DEFAULT_TOKEN_DECIMALS
 from rotkehlchen.chain.ethereum.utils import MULTICALL_CHUNKS
+from rotkehlchen.chain.evm.constants import FAKE_GENESIS_TX_RECEIPT, GENESIS_HASH
 from rotkehlchen.chain.evm.contracts import EvmContract, EvmContracts
 from rotkehlchen.chain.evm.proxies_inquirer import EvmProxiesInquirer
 from rotkehlchen.chain.evm.types import NodeName, WeightedNode
@@ -631,6 +632,8 @@ class EvmNodeInquirer(metaclass=ABCMeta):
             web3: Optional[Web3],
             tx_hash: EVMTxHash,
     ) -> dict[str, Any]:
+        if tx_hash == GENESIS_HASH:
+            return FAKE_GENESIS_TX_RECEIPT
         if web3 is None:
             tx_receipt = self.etherscan.get_transaction_receipt(tx_hash)
             try:
--- a/rotkehlchen/chain/evm/transactions.py
+++ b/rotkehlchen/chain/evm/transactions.py
@@ -8,12 +8,13 @@ from gevent.lock import Semaphore
 from pysqlcipher3 import dbapi2 as sqlcipher
 
 from rotkehlchen.api.websockets.typedefs import TransactionStatusStep, WSMessageType
+from rotkehlchen.chain.evm.constants import GENESIS_HASH
 from rotkehlchen.chain.evm.types import EvmAccount
 from rotkehlchen.chain.structures import TimestampOrBlockRange
 from rotkehlchen.db.evmtx import DBEvmTx
 from rotkehlchen.db.filtering import EvmTransactionsFilterQuery
 from rotkehlchen.db.ranges import DBQueryRanges
-from rotkehlchen.errors.misc import RemoteError
+from rotkehlchen.errors.misc import InputError, RemoteError
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.types import ChecksumEvmAddress, EVMTxHash, Timestamp, deserialize_evm_tx_hash
 from rotkehlchen.utils.misc import ts_now
@@ -200,7 +201,6 @@ class EvmTransactions(metaclass=ABCMeta):  # noqa: B024
                 start_ts=start_ts,
                 end_ts=end_ts,
             )
-
         for query_start_ts, query_end_ts in ranges_to_query:
             log.debug(f'Querying {self.evm_inquirer.chain_name} transactions for {address} -> {query_start_ts} - {query_end_ts}')  # noqa: E501
             try:
@@ -455,16 +455,44 @@ class EvmTransactions(metaclass=ABCMeta):  # noqa: B024
                 has_premium=True,  # we don't need any limiting here
             )
 
-        if len(result) == 0:
+        if tx_hash == GENESIS_HASH:
+            # For each tracked account, query to see if it had any transactions in the
+            # genesis block. We check this even if there already is a 0x0..0 transaction in the db
+            # in order to be sure that genesis transactions are queried for all accounts.
+            with self.database.conn.read_ctx() as cursor:
+                accounts_data = self.database.get_blockchain_account_data(
+                    cursor=cursor,
+                    blockchain=self.evm_inquirer.chain_id.to_blockchain(),
+                )
+                for data in accounts_data:
+                    self._get_transactions_for_range(
+                        address=data.address,
+                        start_ts=Timestamp(0),
+                        end_ts=Timestamp(0),
+                    )
+
+            # Check whether the genesis tx was added
+            with self.database.conn.read_ctx() as cursor:
+                added_tx = dbevmtx.get_evm_transactions(
+                    cursor=cursor,
+                    filter_=EvmTransactionsFilterQuery.make(tx_hash=GENESIS_HASH, chain_id=self.evm_inquirer.chain_id),  # noqa: E501
+                    has_premium=True,  # we don't need any limiting here
+                )
+            if len(added_tx) == 0:
+                raise InputError(
+                    f'There is no tracked {str(self.evm_inquirer.chain_id)} address that '
+                    f'would have a genesis transaction',
+                )
+        elif len(result) == 0:  # normal functionality
             transaction = self.evm_inquirer.get_transaction_by_hash(tx_hash)
             with self.database.user_write() as write_cursor:
-                dbevmtx.add_evm_transactions(write_cursor, [transaction], relevant_address=None)
+                dbevmtx.add_evm_transactions(write_cursor, [transaction], relevant_address=None)  # noqa: E501
             period = TimestampOrBlockRange(
                 range_type='blocks',
                 from_value=transaction.block_number,
                 to_value=transaction.block_number,
             )
-            if transaction.to_address is not None:  # internal transactions only through contracts
+            if transaction.to_address is not None:  # internal transactions only through contracts  # noqa: E501
                 self._query_and_save_internal_transactions_for_range(
                     address=None,  # get all internal transactions for the parent hash
                     period=period,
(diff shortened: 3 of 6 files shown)


rotkehlchen/chain/evm/constants.py
```diff
@@ -6,3 +6,20 @@
 ZERO_ADDRESS = string_to_evm_address('0x0000000000000000000000000000000000000000')
 ETH_SPECIAL_ADDRESS = string_to_evm_address('0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE')
 GENESIS_HASH = deserialize_evm_tx_hash('0x' + '0' * 64)  # hash for transactions in genesis block
+
+EMPTY_RECEIPT = {
+    'blockHash': '0x' + '0' * 64,
```
yabirgb, line 11: can you maybe store this in a variable and use it here, in genesis hash and below?

yabirgb (commented):
Mostly questions about the approach

rotkehlchen/chain/evm/constants.py
```diff
@@ -6,3 +6,20 @@
...
+    'blockNumber': 0,
+    'contractAddress': None,
+    'cumulativeGasUsed': 21000,
+    'effectiveGasPrice': 50000000000000,
```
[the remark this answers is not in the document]
nebolax: Hm, yeah, okay

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -631,6 +632,8 @@ def _get_transaction_receipt(
             web3: Optional[Web3],
             tx_hash: EVMTxHash,
     ) -> dict[str, Any]:
+        if tx_hash == GENESIS_HASH:
+            return EMPTY_RECEIPT
```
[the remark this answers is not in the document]
nebolax: It just returns a not found transaction / receipt

rotkehlchen/chain/evm/transactions.py
```diff
@@ -456,19 +457,45 @@ def get_or_query_transaction_receipt(self, tx_hash: EVMTxHash) -> 'EvmTxReceipt'
...
+                    added_tx = dbevmtx.get_evm_transactions(
+                        cursor=cursor,
+                        filter_=EvmTransactionsFilterQuery.make(tx_hash=GENESIS_HASH, chain_id=self.evm_inquirer.chain_id),  # noqa: E501
+                        has_premium=True,  # we don't need any limiting here
+                    )
+                if len(added_tx) == 0:
+                    raise InputError(
+                        f'There is no tracked {str(self.evm_inquirer.chain_id)} address that '
+                        f'would have a genesis transaction',
+                    )
```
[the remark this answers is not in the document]
nebolax: This can happen if there are no tracked addresses with genesis transactions, but someone attempts to add a genesis (`0x0..0`) tx. I think that raising an error is better because normally this should not happen.

rotkehlchen/chain/evm/constants.py
```diff
@@ -5,4 +5,22 @@
 MAX_BLOCKTIME_CACHE = 250  # 55 mins with 13 secs avg block time
 ZERO_ADDRESS = string_to_evm_address('0x0000000000000000000000000000000000000000')
 ETH_SPECIAL_ADDRESS = string_to_evm_address('0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE')
-GENESIS_HASH = deserialize_evm_tx_hash('0x' + '0' * 64)  # hash for transactions in genesis block
+EMPTY_64_HEX = '0x' + '0' * 64
```
LefterisJP, line 8: this is an empty 32 byte hash. not 64.

```suggestion
ZERO_32_BYTES_HEX = '0x' + '0' * 64
```

rotkehlchen/chain/evm/constants.py
```diff
@@ -5,4 +5,22 @@
...
+GENESIS_HASH = deserialize_evm_tx_hash(EMPTY_64_HEX)  # hash for transactions in genesis block
+
+EMPTY_RECEIPT = {  # Uses ethereum receipt structure
```
LefterisJP, line 11: This is not just an empty receipt. It's specifically made to be fake and tied to the genesis block

```suggestion

# Fake receipt with values taken from ethereum mainnet, to emulate a receipt for the genesis transactions
FAKE_GENESIS_TX_RECEIPT = {
```

rotkehlchen/chain/evm/transactions.py
```diff
@@ -456,19 +457,45 @@ def get_or_query_transaction_receipt(self, tx_hash: EVMTxHash) -> 'EvmTxReceipt'
...
-            )
-            if transaction.to_address is not None:  # internal transactions only through contracts
-                self._query_and_save_internal_transactions_for_range(
-                    address=None,  # get all internal transactions for the parent hash
-                    period=period,
+            if tx_hash == GENESIS_HASH:
+                # Query genesis transactions
+                with self.database.conn.read_ctx() as cursor:
+                    accounts_data = self.database.get_blockchain_account_data(
+                        cursor,
```
LefterisJP, line 464: kwargs

rotkehlchen/chain/evm/transactions.py
```diff
@@ -456,19 +457,45 @@ def get_or_query_transaction_receipt(self, tx_hash: EVMTxHash) -> 'EvmTxReceipt'
...
-                range_type='blocks',
-                from_value=transaction.block_number,
-                to_value=transaction.block_number,
-            )
-            if transaction.to_address is not None:  # internal transactions only through contracts
-                self._query_and_save_internal_transactions_for_range(
-                    address=None,  # get all internal transactions for the parent hash
-                    period=period,
+            if tx_hash == GENESIS_HASH:
+                # Query genesis transactions
```
LefterisJP, line 461:
```suggestion
                # For each tracked account, query to see if it had any transactions in the genesis block
```

rotkehlchen/chain/evm/transactions.py
```diff
@@ -456,19 +457,45 @@ def get_or_query_transaction_receipt(self, tx_hash: EVMTxHash) -> 'EvmTxReceipt'
...
+                with self.database.conn.read_ctx() as cursor:
+                    accounts_data = self.database.get_blockchain_account_data(
+                        cursor,
+                        self.evm_inquirer.chain_id.to_blockchain(),
+                    )
+                for data in accounts_data:
```
LefterisJP, line 467: How often can we come here?

Every time the `tx_hash == GENESIS_HASH` we will enter here, if none of the accounts ever have any transactions in the genesis block.

Shouldn't you also check if the account you are iterating has had the block range queried? Cause if it had, then you should not requery etherscan for it every time.

rotkehlchen/chain/evm/transactions.py
```diff
@@ -456,19 +457,45 @@ def get_or_query_transaction_receipt(self, tx_hash: EVMTxHash) -> 'EvmTxReceipt'
...
+                    self._query_and_save_transactions_for_range(
+                        address=data.address,
+                        period=TimestampOrBlockRange(range_type='blocks', from_value=0, to_value=0),  # noqa: E501
+                    )
+
+                # Check whether the genesis tx was added
+                with self.database.conn.read_ctx() as cursor:
+                    added_tx = dbevmtx.get_evm_transactions(
```
LefterisJP, line 475: Hmmm why do this check? It makes no sense.

Also .... what happens if you do write the genesis hash in the DB and then you add another address later, that also has a genesis transaction?

rotkehlchen/chain/evm/transactions.py
```diff
@@ -456,19 +457,45 @@ def get_or_query_transaction_receipt(self, tx_hash: EVMTxHash) -> 'EvmTxReceipt'
...
+                        cursor=cursor,
+                        filter_=EvmTransactionsFilterQuery.make(tx_hash=GENESIS_HASH, chain_id=self.evm_inquirer.chain_id),  # noqa: E501
+                        has_premium=True,  # we don't need any limiting here
+                    )
+                if len(added_tx) == 0:
+                    raise InputError(
+                        f'There is no tracked {str(self.evm_inquirer.chain_id)} address that '
+                        f'would have a genesis transaction',
+                    )
+            else:
```
LefterisJP, line 485: The else is what was there before, right?

```suggestion
            else:  # normal functionality
```

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -489,3 +496,61 @@ def test_eth_deposit(
             extra_data=None,
         ),
     ]
+
+
+@pytest.mark.parametrize('ethereum_accounts', [[
+    '0xA1E4380A3B1f749673E270229993eE55F35663b4',
+    '0x756F45E3FA69347A9A973A725E3C98bC4db0b5a0',
+]])
+def test_genesis_transaction(database, ethereum_inquirer, ethereum_accounts):
```
LefterisJP, line 505: this needs vcr

rotkehlchen/chain/evm/transactions.py
nebolax: > Hmmm why do this check? It makes no sense.

This check tells us whether any of the tracked addresses had a genesis transaction.

> Also .... what happens if you do write the genesis hash in the DB and then you add another address later, that also has a genesis transaction?

You will have to pass through the classic process of querying transactions (the one that happens on the Evm Transactions page in the app) and get the transactions queried for that address.

rotkehlchen/chain/evm/transactions.py
nebolax: Right

rotkehlchen/chain/evm/transactions.py
LefterisJP: > This check tells us whether any of the tracked addresses had a genesis transaction.

And? Why do you need this here? I don't; see it used

nebolax: As you can see in the check below ```python
if len(added_tx) == 0:
                    raise InputError(
                        f'There is no tracked {str(self.evm_inquirer.chain_id)} address that '
                        f'would have a genesis transaction',
                    )
```

if case if there will are no genesis transactions found, an input error is thrown

Comment (nebolax) [reply rotkehlchen/chain/evm/transactions.py:475]: In case of "normal" transactions if the transaction is not found, a remote error is raised.

So for genesis transactions we also need some error to be thrown.

Comment (LefterisJP) [reply rotkehlchen/chain/evm/transactions.py:475]: When can this happen? Can you describe the flow here? This function is called and handles remote error only at `decode_transaction_hashes`. So when can it get here with the genesis hash and no tracked address having a transaction?

Comment (nebolax) [reply rotkehlchen/chain/evm/transactions.py:475]: `decode_transaction_hashes` method is exposed via API, so anything can be passed there at any time.

However in the app it probably won't happen since for redecoding frontend sends only already existing transactions.

But if we consider only backend then this zero genesis hash can be sent to the backend at any time.

File: rotkehlchen/chain/evm/constants.py [hunk 13]
```diff
@@ -5,4 +5,24 @@
 MAX_BLOCKTIME_CACHE = 250  # 55 mins with 13 secs avg block time
 ZERO_ADDRESS = string_to_evm_address('0x0000000000000000000000000000000000000000')
 ETH_SPECIAL_ADDRESS = string_to_evm_address('0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE')
-GENESIS_HASH = deserialize_evm_tx_hash('0x' + '0' * 64)  # hash for transactions in genesis block
+ZERO_32_BYTES_HEX = '0x' + '0' * 64
+GENESIS_HASH = deserialize_evm_tx_hash(ZERO_32_BYTES_HEX)  # hash for transactions in genesis block
+
+# Fake receipt with values taken from ethereum mainnet, to emulate a receipt for the
+# genesis transactions
+FAKE_GENESIS_TX_RECEIPT = {  # Uses ethereum receipt structure
```
LefterisJP, line 13: Please apply [my comment](https://github.com/rotki/rotki/pull/5616#discussion_r1108802070) directly. If you notice this comment is now redundant
```suggestion
FAKE_GENESIS_TX_RECEIPT = {
```

rotkehlchen/chain/evm/transactions.py
```diff
@@ -456,19 +457,52 @@ def get_or_query_transaction_receipt(self, tx_hash: EVMTxHash) -> 'EvmTxReceipt'
...
+                # genesis block
+                with self.database.conn.read_ctx() as cursor:
+                    accounts_data = self.database.get_blockchain_account_data(
+                        cursor=cursor,
+                        blockchain=self.evm_inquirer.chain_id.to_blockchain(),
+                    )
+                    for data in accounts_data:
+                        queried_range = self.database.get_used_query_range(
+                            cursor=cursor,
+                            name=self.evm_inquirer.chain_id.name + 'txs_' + data.address,
```
LefterisJP, line 471: This is wrong. Notice how it's done in the rest of this file

```python
        location_string = f'{self.evm_inquirer.blockchain.to_range_prefix("txs")}_{address}'
```

The result is not the same. Which also means this may not be well tested. Can you add a test here that will fail with your current implementation and reveal the mistake?

LefterisJP (commented):
Left some comments and for a serious bug in the implementation.

What's more this is broken since the relevant tests fail.

```
rotkehlchen/tests/unit/decoders/test_main.py::test_genesis_transaction
rotkehlchen/tests/unit/decoders/test_main.py::test_genesis_transaction_no_address
```

rotkehlchen/chain/evm/transactions.py
nebolax: Fixed

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -1,21 +1,37 @@
...
+    ChainID,
+    EvmTransaction,
+    Location,
+    Timestamp,
+    TimestampMS,
+    deserialize_evm_tx_hash,
+)
 from rotkehlchen.utils.hexbytes import hexstring_to_bytes
 
+# Have to use a constant instead of make_evm_address() because vcr doesn't work otherwise.
```
LefterisJP, line 32: why doesn't VCR work otherwise? I suppose you mean because you need it repeatable, right?
```suggestion
# Have to use a constant so the recorded (VCR) tests work properly when repeated
```

rotkehlchen/tests/utils/ethereum.py
```diff
@@ -365,21 +365,32 @@ def get_decoded_events_of_transaction(
         evm_inquirer: EthereumInquirer,
         database: DBHandler,
         tx_hash: EVMTxHash,
+        transactions: Optional[EvmTransactions] = None,
 ) -> tuple[list[HistoryBaseEntry], EVMTransactionDecoder]:
     """A convenience function to ask get transaction, receipt and decoded event for a tx_hash
 
+    It also accepts `transactions` in case the caller whants to apply some mocks (like call_count)
+    on that object.
```
LefterisJP, line 373:
```suggestion
    It also optionally accepts `transactions` in case the caller wants to apply
    some mocks (like call_count) on that object.
```

LefterisJP (approved):
LGTM, @nebolax ! Thanks. Left two comments whcih should be fixed in another PR. They are typos/comments only

PR #5617, opened 2023-02-16
Refactor node querying logic

prettyirrelevant:
## Checklist
This PR does the following:

- Upon node connection, its capabalities are determined i.e whether it's pruned, archive node or not.
- Refactor typing of node methods e.g `get_transaction_by_hash`, `get_transaction_receipt`
- Handle cases where a tx hash does not exist on-chain in evm txn addition by hash
- Add and refactor tests to handle these changes.

Also, it fixes [`rotkehlchen/tests/conftest.py`](https://github.com/rotki/rotki/blob/2de468cae60907d1cfa685c4c98d9a0c5d891610/rotkehlchen/tests/conftest.py#L258) issue with the `caching_dir` filepath on Mac

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/yearn/vaults.py
+++ b/rotkehlchen/chain/ethereum/modules/yearn/vaults.py
@@ -1,3 +1,4 @@
+import logging
 from typing import TYPE_CHECKING, Any, NamedTuple, Optional
 
 from gevent.lock import Semaphore
@@ -78,6 +79,7 @@ from rotkehlchen.errors.asset import UnknownAsset
 from rotkehlchen.fval import FVal
 from rotkehlchen.history.price import query_usd_price_zero_if_error
 from rotkehlchen.inquirer import Inquirer
+from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.premium.premium import Premium
 from rotkehlchen.types import (
     YEARN_VAULTS_V2_PROTOCOL,
@@ -101,6 +103,9 @@ if TYPE_CHECKING:
     )
     from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
     from rotkehlchen.db.dbhandler import DBHandler
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
 
 
 class YearnVaultHistory(NamedTuple):
@@ -472,16 +477,16 @@ class YearnVaults(EthereumModule):
             tx_receipt = self.ethereum.get_transaction_receipt(tx_hash)
             deposit_index = deposit_event['logIndex']
             mint_amount = None
-            for log in tx_receipt['logs']:
+            for tx_log in tx_receipt['logs']:
                 found_event = (
-                    log['topics'][0] == '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef' and  # noqa: E501
-                    log['topics'][1] == address_to_bytes32(ZERO_ADDRESS) and
-                    log['topics'][2] == address_to_bytes32(address)
+                    tx_log['topics'][0] == '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef' and  # noqa: E501
+                    tx_log['topics'][1] == address_to_bytes32(ZERO_ADDRESS) and
+                    tx_log['topics'][2] == address_to_bytes32(address)
                 )
                 if found_event:
                     # found the mint log
                     mint_amount = token_normalized_value(
-                        token_amount=hexstr_to_int(log['data']),
+                        token_amount=hexstr_to_int(tx_log['data']),
                         token=vault.token,
                     )
 
@@ -558,16 +563,16 @@ class YearnVaults(EthereumModule):
             tx_receipt = self.ethereum.get_transaction_receipt(tx_hash)
             withdraw_index = withdraw_event['logIndex']
             burn_amount = None
-            for log in tx_receipt['logs']:
+            for tx_log in tx_receipt['logs']:
                 found_event = (
-                    log['topics'][0] == '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef' and  # noqa: E501
-                    log['topics'][1] == address_to_bytes32(address) and
-                    log['topics'][2] == address_to_bytes32(ZERO_ADDRESS)
+                    tx_log['topics'][0] == '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef' and  # noqa: E501
+                    tx_log['topics'][1] == address_to_bytes32(address) and
+                    tx_log['topics'][2] == address_to_bytes32(ZERO_ADDRESS)
                 )
                 if found_event:
                     # found the burn log
                     burn_amount = token_normalized_value(
-                        token_amount=hexstr_to_int(log['data']),
+                        token_amount=hexstr_to_int(tx_log['data']),
                         token=vault.token,
                     )
 
--- a/rotkehlchen/chain/ethereum/node_inquirer.py
+++ b/rotkehlchen/chain/ethereum/node_inquirer.py
@@ -10,9 +10,16 @@ from ens.main import ENS_MAINNET_ADDR
 from ens.utils import is_none_or_zero_address, normal_name_to_hash, normalize_name
 from eth_typing import BlockNumber, HexStr
 from web3 import Web3
+from web3.exceptions import TransactionNotFound
 
 from rotkehlchen.chain.constants import DEFAULT_EVM_RPC_TIMEOUT
-from rotkehlchen.chain.ethereum.constants import ETHEREUM_ETHERSCAN_NODE
+from rotkehlchen.chain.ethereum.constants import (
+    ARCHIVE_NODE_CHECK_ADDRESS,
+    ARCHIVE_NODE_CHECK_BLOCK,
+    ARCHIVE_NODE_CHECK_EXPECTED_BALANCE,
+    ETHEREUM_ETHERSCAN_NODE,
+    PRUNED_NODE_CHECK_TX_HASH,
+)
 from rotkehlchen.chain.ethereum.graph import Graph
 from rotkehlchen.chain.ethereum.modules.curve.pools_cache import (
     ensure_curve_tokens_existence,
@@ -22,10 +29,13 @@ from rotkehlchen.chain.ethereum.modules.curve.pools_cache import (
 )
 from rotkehlchen.chain.evm.contracts import EvmContracts
 from rotkehlchen.chain.evm.node_inquirer import WEB3_LOGQUERY_BLOCK_RANGE, EvmNodeInquirer
-from rotkehlchen.chain.evm.types import string_to_evm_address
-from rotkehlchen.errors.misc import InputError, RemoteError, UnableToDecryptRemoteData
+from rotkehlchen.errors.misc import (
+    BlockchainQueryError,
+    InputError,
+    RemoteError,
+    UnableToDecryptRemoteData,
+)
 from rotkehlchen.errors.serialization import DeserializationError
-from rotkehlchen.fval import FVal
 from rotkehlchen.globaldb.handler import GlobalDBHandler
 from rotkehlchen.greenlets.manager import GreenletManager
 from rotkehlchen.logging import RotkehlchenLogsAdapter
@@ -281,17 +291,26 @@ class EthereumInquirer(EvmNodeInquirer, LockableQueryMixIn):
 
         return BlockNumber(block_number)
 
-    def have_archive(self, requery: bool = False) -> bool:
-        if self.queried_archive_connection and requery is False:
-            return self.archive_connection
+    def _is_pruned(self, web3: Web3) -> bool:
+        try:
+            tx = web3.eth.get_transaction(PRUNED_NODE_CHECK_TX_HASH)  # type: ignore
+        except (
+            requests.exceptions.RequestException,
+            TransactionNotFound,
+            BlockchainQueryError,
+            KeyError,
+        ):
+            tx = None
 
+        return tx is None
+
+    def _have_archive(self, web3: Web3) -> bool:
         balance = self.get_historical_balance(
-            address=string_to_evm_address('0x50532e4Be195D1dE0c2E6DfA46D9ec0a4Fee6861'),
-            block_number=87042,
+            address=ARCHIVE_NODE_CHECK_ADDRESS,
+            block_number=ARCHIVE_NODE_CHECK_BLOCK,
+            web3=web3,
         )
-        self.archive_connection = balance is not None and balance == FVal('5.1063307')
-        self.queried_archive_connection = True
-        return self.archive_connection
+        return balance == ARCHIVE_NODE_CHECK_EXPECTED_BALANCE
 
     def _get_blocknumber_by_time_from_subgraph(self, ts: Timestamp) -> int:
         """Queries Ethereum Blocks Subgraph for closest block at or before given timestamp"""
--- a/rotkehlchen/chain/evm/node_inquirer.py
+++ b/rotkehlchen/chain/evm/node_inquirer.py
@@ -2,9 +2,9 @@ import json
 import logging
 import random
 from abc import ABCMeta, abstractmethod
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 from itertools import zip_longest
-from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Union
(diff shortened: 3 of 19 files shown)


rotkehlchen/chain/ethereum/node_inquirer.py
```diff
@@ -281,17 +287,27 @@ def query_highest_block(self) -> BlockNumber:
...
             block_number=87042,
+            web3=web3,
         )
-        self.archive_connection = balance is not None and balance == FVal('5.1063307')
-        self.queried_archive_connection = True
-        return self.archive_connection
+        is_archive = balance == FVal('5.1063307')
+
+        try:
+            tx = web3.eth.get_transaction(HexStr('0x5c504ed432cb51138bcf09aa5e8a410dd4a1e204ef84bfed1be16dfba1b22060'))  # noqa: E501
```
nebolax, line 299: I would move this tx hash and  the address used above to constants

rotkehlchen/chain/ethereum/node_inquirer.py
```diff
@@ -281,17 +287,27 @@ def query_highest_block(self) -> BlockNumber:
...
+            block_no = tx['blockNumber']
+        except (
+            requests.exceptions.RequestException,
+            TransactionNotFound,
+            BlockchainQueryError,
+            KeyError,
+        ):
+            block_no = None
+        is_pruned = block_no != 46147
```
nebolax, line 308: Why is there a need in checking block number? Won't it be sufficient to just check that the transaction was found?

rotkehlchen/chain/optimism/node_inquirer.py
```diff
@@ -57,8 +62,27 @@ def query_highest_block(self) -> BlockNumber:
         log.debug('Optimism highest block result', block=block_number)
         return BlockNumber(block_number)
 
-    def have_archive(self, requery: bool = False) -> bool:
-        return False
+    def determine_capabilities(self, web3: Web3) -> tuple[bool, bool]:
```
nebolax, line 65: Looks like this code for determining capabilities could be abstracted.

rotkehlchen/externalapis/etherscan.py
```diff
@@ -133,7 +144,7 @@ def _query(
             action: str,
             options: Optional[dict[str, Any]] = None,
             timeout: Optional[tuple[int, int]] = None,
-    ) -> Union[list[dict[str, Any]], str, list[EvmTransaction], dict[str, Any]]:
+    ) -> Union[list[dict[str, Any]], str, list[EvmTransaction], dict[str, Any], None]:
```
nebolax, line 147:
```suggestion
    ) -> Optional[Union[list[dict[str, Any]], str, list[EvmTransaction], dict[str, Any]]]:
```

rotkehlchen/externalapis/etherscan.py
```diff
@@ -200,6 +211,9 @@ def _query(
             try:
                 result = json_ret.get('result', None)
                 if result is None:
+                    if action in ('eth_getTransactionByHash', 'eth_getTransactionReceipt'):
```
nebolax, line 214: Hmm why to silence the error in these cases?

rotkehlchen/chain/ethereum/modules/yearn/vaults.py
```diff
@@ -556,18 +568,24 @@ def _get_vault_withdraw_events(
             )
             tx_hash = deserialize_evm_tx_hash(withdraw_event['transactionHash'])
             tx_receipt = self.ethereum.get_transaction_receipt(tx_hash)
+            if tx_receipt is None:
```
nebolax, line 571: In `_query` you added a check so that for receipts and transactions pruned nodes are not queried. So is there a need to handle an empty receipt here?

prettyirrelevant: An empty receipt is possible since a `tx_hash` might be valid but does not exist on-chain.

rotkehlchen/chain/ethereum/node_inquirer.py
prettyirrelevant: ah yes, thanks

rotkehlchen/chain/ethereum/node_inquirer.py
prettyirrelevant: I am not so familiar with the nitty-gritty of ethereum but I guess you're right

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -444,25 +456,33 @@ def _query(self, method: Callable, call_order: Sequence[WeightedNode], **kwargs:
...
-                TransactionNotFound,
                 BlockNotFound,
                 BadResponseFormat,
                 ValueError,  # Yabir saw this happen with mew node for unavailable method at node. Since it's generic we should replace if web3 implements https://github.com/ethereum/web3.py/issues/2448  # noqa: E501
             ) as e:
-                log.warning(f'Failed to query {node} for {str(method)} due to {str(e)}')
+                log.warning(f'Failed to query {node_info} for {str(method)} due to {str(e)}')
                 # Catch all possible errors here and just try next node call
                 continue
+            except TransactionNotFound:
```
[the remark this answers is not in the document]
prettyirrelevant: When a transaction is added by its hash, if the hash is valid but isn't present on-chain, the user sees a `RemoteError` saying.... **"we tried all nodes but couldn't make the query. Check logs"** which is the same error you'd get if all nodes are down.

...which isn't so helpful to the user. So yeah

rotkehlchen/externalapis/etherscan.py
prettyirrelevant: Because if a transaction does not exist on-chain, Etherscan returns `{'result': null}` as a response

rotkehlchen/chain/evm/transactions.py
```diff
@@ -567,11 +602,17 @@ def add_transaction_by_hash(
...
         - DeserializationError if there's an issue deserializing a value.
+        - InputError if the tx_hash or its receipt is not found on the blockchain.
         - pysqlcipher3.dbapi2.IntegrityError if the tx_hash is present in the db or address is not tracked.
         """  # noqa: E501
         dbevmtx = DBEvmTx(self.database)
         transaction = self.evm_inquirer.get_transaction_by_hash(tx_hash)
+        if transaction is None:
+            raise InputError('Transaction not found on chain.')
         receipt_data = self.evm_inquirer.get_transaction_receipt(tx_hash)
+        if receipt_data is None:
```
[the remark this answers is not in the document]
prettyirrelevant: Since it's a user input... also checked its usage in the codebase.
Looks more appropriate

rotkehlchen/externalapis/etherscan.py
nebolax: Yeah, but should then a Remote Error / Input Error be raised?

rotkehlchen/chain/ethereum/modules/yearn/vaults.py
nebolax: Yeah, I mean that this tx hash is retrieved from chain, so it is not something user input, and thus I would say we can rely on it's existence.

But anyway, I think having an extra check won't hurt.

rotkehlchen/chain/evm/node_inquirer.py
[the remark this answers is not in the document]
nebolax: Mmm aha, yeah, makes sense :+1:

rotkehlchen/chain/evm/transactions.py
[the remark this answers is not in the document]
nebolax: Yes, makes sense :+1:

One thing though -- receipt always exists if transaction exists, doesn't it? So I would say that there is no need in checking for receipt being None.

But is not something serious.

rotkehlchen/externalapis/etherscan.py
prettyirrelevant: `RemoteError` would be caught `self._query()` of `EVMNodeInquirer` which is generic

rotkehlchen/tests/unit/test_ethereum_inquirer.py
```diff
@@ -304,3 +307,88 @@ def test_get_blocknumber_by_time_subgraph(ethereum_inquirer):
 def test_get_blocknumber_by_time_etherscan(ethereum_inquirer):
     """Queries etherscan for known block times"""
     _test_get_blocknumber_by_time(ethereum_inquirer, True)
+
+
+@pytest.mark.parametrize(*ETHEREUM_NODES_PARAMETERS_WITH_PRUNED_AND_NOT_ARCHIVED)
```
LefterisJP, line 312: for this and the other tests in here if they do remote connections, they should use vcr believe.

rotkehlchen/tests/unit/test_evm_transactions.py
```diff
@@ -1,4 +1,5 @@
 from unittest.mock import patch
+
```
LefterisJP, line 2: why this space here? If not changing something don't touch it :smile:

rotkehlchen/tests/unit/test_optimism_inquirer.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+
+from rotkehlchen.tests.utils.ethereum import wait_until_all_nodes_connected
+from rotkehlchen.tests.utils.optimism import OPTIMISM_NODES_PARAMETERS_WITH_PRUNED_AND_NOT_ARCHIVED
+
+
+@pytest.mark.parametrize(*OPTIMISM_NODES_PARAMETERS_WITH_PRUNED_AND_NOT_ARCHIVED)
```
LefterisJP, line 7: same comment about vcr in this file

rotkehlchen/tests/utils/ethereum.py
```diff
@@ -48,6 +48,17 @@
 ])
 ALCHEMY_TEST = 'https://eth-mainnet.alchemyapi.io/v2/ga1GtB7R26UgzjextaVpbaWZ49nSi2zt'
 
+PRUNED_AND_NOT_ARCHIVED_NODE = WeightedNode(
+    node_info=NodeName(
+        name='Public Node',
+        endpoint='https://ethereum.publicnode.com',
```
LefterisJP, line 54: nice so you found a public pruned node?

LefterisJP (commented):
Hey so approach looks okay but have some comments:

### Failed Tests

Related tests seem to fail.

`rotkehlchen/tests/unit/test_ethereum_inquirer.py::test_use_open_nodes`
`rotkehlchen/tests/unit/test_ethereum_inquirer.py::test_ethereum_nodes_prune_and_archive_status`

### Different version of the get functions

So as discussed in the PR we need a different version of those get_transaction and get_receipt functions. Either the old ones should be renamed to `maybe_get_xxx` and then create the `get_xxx` version which also internally does an assert and returns the non optional type. Or ... you can create a `get_and_check_xxx` function ... but hm ... I think using the `maybe_get_xxx` approach is cleaner.

### Tests

I see you added lots of tests that test the functionality. you should use vcr there to mock this, as it's really not something we should be querying remotes in the tests for. Either with vcr or a simple mock of the responses. But perhaps vcr is cleaner.

rotkehlchen/tests/utils/ethereum.py
prettyirrelevant: yeah, for Ethereum

rotkehlchen/chain/ethereum/node_inquirer.py
```diff
@@ -281,17 +291,26 @@ def query_highest_block(self) -> BlockNumber:
 
         return BlockNumber(block_number)
 
-    def have_archive(self, requery: bool = False) -> bool:
-        if self.queried_archive_connection and requery is False:
-            return self.archive_connection
-
+    def _is_pruned(self, web3: Web3) -> bool:
+        try:
+            tx = web3.eth.get_transaction(HexStr(PRUNED_NODE_CHECK_TX_HASH.hex()))  # pylint: disable=no-member  # noqa: E501
```
LefterisJP, line 296: this is not needed. The typing in web3.eth.get_transaction is wrong. What you are doing here is you are turning a hexbytes, to hexstr and then pass it in, but hexbytes is accepted.

Just pass it as it and add a type ignore and an explanation as here: https://github.com/rotki/rotki/blob/c6bded6e4e4b44aea7771f6de8a90d8b87c508c7/rotkehlchen/chain/evm/node_inquirer.py#L690

Should work.

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -668,26 +693,43 @@ def _get_transaction_receipt(
...
             tx_hash=tx_hash,
         )
 
+    def get_transaction_receipt(
+            self,
+            tx_hash: EVMTxHash,
+            call_order: Optional[Sequence[WeightedNode]] = None,
+    ) -> dict[str, Any]:
+        """Retrieves the transaction receipt for the tx_hash provided.
+        This method assumes the tx_hash is present on-chain.
```
LefterisJP, line 713:
```suggestion
        This method assumes the tx_hash is present on-chain and we are connected to at least 1 node that can retrieve it.
```

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -703,17 +745,32 @@ def _get_transaction_by_hash(
...
             tx_hash=tx_hash,
         )
 
+    def get_transaction_by_hash(
+            self,
+            tx_hash: EVMTxHash,
+            call_order: Optional[Sequence[WeightedNode]] = None,
+    ) -> EvmTransaction:
+        """Retrieves information about a transaction from its hash.
+        This method assumes the tx_hash is present on-chain.
```
LefterisJP, line 765:
```suggestion
        This method assumes the tx_hash is present on-chain and we are connected to at least one node that can retrieve it.
```

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -1103,6 +1160,18 @@ def _process_contract_info(
 
         return decoded_contract_info
 
+    def determine_capabilities(self, web3: Web3) -> tuple[bool, bool]:
+        """This method checks for the capabilities of a rpc node. This includes:
```
LefterisJP, line 1164:
```suggestion
        """This method checks for the capabilities of an rpc node. This includes:
```

rotkehlchen/chain/optimism/node_inquirer.py
```diff
@@ -57,8 +68,26 @@ def query_highest_block(self) -> BlockNumber:
         log.debug('Optimism highest block result', block=block_number)
         return BlockNumber(block_number)
 
-    def have_archive(self, requery: bool = False) -> bool:
-        return False
+    def _is_pruned(self, web3: Web3) -> bool:
+        try:
+            tx = web3.eth.get_transaction(HexStr(PRUNED_NODE_CHECK_TX_HASH.hex()))  # pylint: disable=no-member  # noqa: E501
```
LefterisJP, line 73: same comment on the typing here as in ethereum/node_inquirer.py

rotkehlchen/tests/conftest.py
```diff
@@ -235,7 +235,7 @@ def vcr_cassette_dir(request: pytest.FixtureRequest) -> str:
 
         # verify that the cassettes are present locally and up to date
         if (tests_caching_dir / '.git').exists() is False:
-            cmd = f'git clone https://github.com/rotki/test-caching.git {tests_caching_dir}'
+            cmd = f'git clone https://github.com/rotki/test-caching.git "{tests_caching_dir}"'
```
LefterisJP, line 238: Can you move this change in a different commit and make it to bugfixes? We need it in oth places if it fixes something.

LefterisJP (commented):
Some more comments and 1 related test fail: `rotkehlchen/tests/api/test_evm_transactions.py::test_evm_transaction_hash_addition`

LefterisJP (approved):
LGTM. The failed test is due to:

```
18/02/2023 15:28:44 UTC] WARNING rotkehlchen.chain.evm.node_inquirer Greenlet-1: Failed to query NodeName(name='etherscan', endpoint='', owned=False, blockchain=<SupportedBlockchain.ETHEREUM: 'ETH'>) for <bound method EvmNodeInquirer._get_transaction_receipt of <rotkehlchen.chain.ethereum.node_inquirer.EthereumInquirer object at 0x7f208b541b80>> due to ethereum Etherscan API request https://api.etherscan.io/api?module=proxy&action=eth_getTransactionReceipt&txhash=0x692f9a6083e905bdeca4f0293f3473d7a287260547f8cbccc38c5cb01591fcda&apikey=8JT7WQBB2VQP5C3416Y8X3S8GBA3CVZKP4 failed with HTTP status code 503 and text The service is unavailable.
```

One thing I thought of is you probably need to re-add the "not found" error code in the add transaction by hash endpoint since it's now implemented.

PR #5620, opened 2023-02-16
fix: per field error from endpoint not showed

lukicenturi:
Closes #(issue_number)

So for the errors per field we show it like this
`$v.field.$errors.map(e => e.$message)`

I think in previous `vuelidate` version, it also pick up the external error from endpoint that we passed through `$externalResults`.
But in current version we use, it no longer picked up. So I have to merge `$errors` and `$externalResults` and show them.

We can move all to use `getErrorMessages` but I think we can do that later in develop instead, in this PR I only repair the ones that have `$externalResults`.

### DIFF
--- a/frontend/app/src/components/accounts/blockchain/Eth2Input.vue
+++ b/frontend/app/src/components/accounts/blockchain/Eth2Input.vue
@@ -8,6 +8,7 @@ import {
   minValue,
   requiredUnless
 } from '@vuelidate/validators';
+import isEmpty from 'lodash/isEmpty';
 import { type Eth2Validator } from '@/types/balances';
 import { type ValidationErrors } from '@/types/api/errors';
 
@@ -20,6 +21,7 @@ const props = defineProps<{
 const emit = defineEmits<{
   (e: 'update:validator', validator: Eth2Validator | null): void;
 }>();
+
 const { validator, errorMessages } = toRefs(props);
 const validatorIndex = ref('');
 const publicKey = ref('');
@@ -61,6 +63,12 @@ const v$ = useVuelidate(
     $externalResults: errorMessages
   }
 );
+
+watch(errorMessages, errors => {
+  if (!isEmpty(errors)) {
+    get(v$).$validate();
+  }
+});
 
 onMounted(() => updateProperties(validator.value));
 
--- a/frontend/app/src/components/history/ExternalTradeForm.vue
+++ b/frontend/app/src/components/history/ExternalTradeForm.vue
@@ -72,13 +72,13 @@ const baseSymbol = assetSymbol(base);
 const quoteSymbol = assetSymbol(quote);
 
 const rules = {
-  base: {
+  baseAsset: {
     required: helpers.withMessage(
       t('external_trade_form.validation.non_empty_base').toString(),
       required
     )
   },
-  quote: {
+  quoteAsset: {
     required: helpers.withMessage(
       t('external_trade_form.validation.non_empty_quote').toString(),
       required
@@ -119,8 +119,8 @@ const rules = {
 const v$ = useVuelidate(
   rules,
   {
-    base,
-    quote,
+    baseAsset: base,
+    quoteAsset: quote,
     amount,
     rate,
     quoteAmount,
@@ -231,6 +231,8 @@ const save = async (): Promise<boolean> => {
       errorMessages,
       convertKeys(deserializeApiErrorMessage(result.message) ?? {}, true, false)
     );
+
+    await get(v$).$validate();
   }
 
   return false;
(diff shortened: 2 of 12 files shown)


kelsos:
Apparently, they did add an extra step between the alphas and the release, which is shown in the example below:

https://vuelidate-next.netlify.app/advanced_usage.html#external-results-with-composition-api

I did a quick check and if you call `await v.value.$validate()` in the method that sets the external errors, then the $externalResults error correctly appears in the `field.$errors`.

Can you adjust the code so that is uses vuelidate correctly?

kelsos:
Created a poc/check here https://github.com/kelsos/vuelidate-external-results-example

kelsos:
@lukicenturi can we keep refactoring on `bugfixes` PRs to a minimum, please? 

In general, we should try to only do the minimum necessary to fix a bug when a PR is targeting `bugfixes`.

For example `:error-messages="getErrorMessages(...)"` change can go to `develop` but it is not needed to fix the PR.

Can we reduce the number of changes in this PR to only what is needed for the fix?

frontend/app/src/components/history/ExternalTradeForm.vue
```diff
@@ -377,29 +398,31 @@ onMounted(() => {
...
+                  v-model="quoteAsset"
                   required
                   outlined
                   data-cy="quote-asset"
-                  :error-messages="v$.quote.$errors.map(e => e.$message)"
+                  :error-messages="getErrorMessages(v$.quoteAsset)"
                   :hint="t('external_trade_form.quote_asset.hint')"
                   :label="t('external_trade_form.quote_asset.label')"
-                  @focus="delete errorMessages['quoteAsset']"
+                  @focus="removeError('quoteAsset')"
```
kelsos, line 424: What happens if we don't explicitly do this?

lukicenturi:
> @lukicenturi can we keep refactoring on `bugfixes` PRs to a minimum, please?
> 
> In general, we should try to only do the minimum necessary to fix a bug when a PR is targeting `bugfixes`.
> 
> For example `:error-messages="getErrorMessages(...)"` change can go to `develop` but it is not needed to fix the PR.
> 
> Can we reduce the number of changes in this PR to only what is needed for the fix?
> 
> Also, another problem is that we don't test these cases can we test them somehow?

Okay, my bad, I will revert unnecessary changes

frontend/app/src/components/history/ExternalTradeForm.vue
lukicenturi: If we use the old way (delete errorMessage), it doesn't change the $error variable. I think it because it doesn't trigger the reactivity.

lukicenturi:
> > @lukicenturi can we keep refactoring on `bugfixes` PRs to a minimum, please?
> > In general, we should try to only do the minimum necessary to fix a bug when a PR is targeting `bugfixes`.
> > For example `:error-messages="getErrorMessages(...)"` change can go to `develop` but it is not needed to fix the PR.
> > Can we reduce the number of changes in this PR to only what is needed for the fix?
> > Also, another problem is that we don't test these cases can we test them somehow?
> 
> Okay, my bad, I will revert unnecessary changes

@kelsos updated

kelsos:
> > > @lukicenturi can we keep refactoring on `bugfixes` PRs to a minimum, please?
> > > In general, we should try to only do the minimum necessary to fix a bug when a PR is targeting `bugfixes`.
> > > For example `:error-messages="getErrorMessages(...)"` change can go to `develop` but it is not needed to fix the PR.
> > > Can we reduce the number of changes in this PR to only what is needed for the fix?
> > > Also, another problem is that we don't test these cases can we test them somehow?
> > 
> > 
> > Okay, my bad, I will revert unnecessary changes
> 
> @kelsos updated, should i add the test here or in develop?

If we keep the modifications to production code to a minimum we can also add them here

frontend/app/src/components/history/ExternalTradeForm.vue
kelsos: ok let me think about it a bit, it's because this has to update the property outside.

frontend/app/src/components/accounts/blockchain/Eth2Input.vue
```diff
@@ -102,6 +121,7 @@ watch(
         :disabled="disabled"
         :label="tc('eth2_input.public_key')"
         :error-messages="v$.publicKey.$errors.map(e => e.$message)"
+        @focus="removeError('publicKey')"
```
kelsos, line 124: @lukicenturi question can you remove this remove errors and the event handler for this,

Then go to the ValidatorAccountForm comment the code and change it so that it sets errorMessages?

```ts
set(errorMessages, { "ownershipPercentage": ["i have an error"] }
```

and play a bit in the form. Add any validator index press save get the error on ownershipPercentage, then go edit the ownership percentage and see if that clears the externalError and allows you to save again.

kelsos: I did a check and it seems that it still works.

kelsos (approved):
lgtm

PR #5621, opened 2023-02-16
Fix events reshuffling

nebolax:
Removes the possibility of the appearance of duplicated indices when reshuffling events.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,6 +9,7 @@ Changelog
 * :bug:`-` Fixes bug where navigation bar on the left didn't expand by default.
 * :bug:`-` Fixes curve deposits and withdrawals accounting.
 * :bug:`5561` Fixes average cost basis calculation.
+* :bug:`-` Fixes bug when some decoded events in transactions would disappear. 
 
 * :release:`1.27.0 <2023-02-03>`
 * :feature:`5015` EVM assets across multiple chains will now appear together in the dashboard, with an option to break them down into their per-chain holding.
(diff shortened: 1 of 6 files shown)


rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -489,3 +491,25 @@ def test_eth_deposit(
...
+    """
+    Tests that `maybe_reshuffle_events` works correctly.
+    Especially tests that there are no duplicated indices produced.
+    """
+    event_a = make_ethereum_event(1)
+    event_b = make_ethereum_event(2)
+    event_c = make_ethereum_event(3)
+
+    events = [event_a, event_b, event_c]
+    assert events == [event_a, event_b, event_c]  # sanity check
```
prettyirrelevant, line 506: You assigned the variable above yourself.

The `sequence_index` check is acceptable but this one seems like a tautology.

docs/changelog.rst
```diff
@@ -9,6 +9,7 @@ Changelog
 * :bug:`-` Fixes bug where navigation bar on the left didn't expand by default.
 * :bug:`-` Fixes curve deposits and withdrawals accounting.
 * :bug:`5561` Fixes average cost basis calculation.
+* :bug:`-` Fixes bug when some decoded events in transactions would disappear.
```
LefterisJP, line 12:
```suggestion
* :bug:`-` Fixes a bug when some decoded events in transactions would disappear. 
```

LefterisJP (commented):
It's in general okay, but I need to see a decoding test that fails when this change is not in. Basically the scenario you saw that lead you to this (perhaps as unit test also and simplified -- just need the scenario)

LefterisJP (approved):
LGTM. @nebolax explained that there were duplicated indices in `rotkehlchen/tests/unit/decoders/test_curve.py::test_deposit_multiple_tokens` which this change changes and fixes.

PR #5624, opened 2023-02-17
feat: scramble numbers with custom multiplier

lukicenturi:
Closes #(issue_number)

So I add the option to fill the multiplier of the scrambling.
I disable the scramble for the prices, because if it's not disabled, then the observer can calculate the `constant` used by comparing it with the real price.

image

### DIFF
--- a/frontend/app/src/components/AssetBalances.vue
+++ b/frontend/app/src/components/AssetBalances.vue
@@ -99,6 +99,7 @@ const sortItems = getSortItems(asset => get(assetInfo(asset)));
     <template #item.usdPrice="{ item }">
       <amount-display
         v-if="item.usdPrice && item.usdPrice.gte(0)"
+        no-scramble
         show-currency="symbol"
         :price-asset="item.asset"
         :price-of-asset="item.usdPrice"
--- a/frontend/app/src/components/display/AmountDisplay.vue
+++ b/frontend/app/src/components/display/AmountDisplay.vue
@@ -54,7 +54,8 @@ const props = defineProps({
   },
 
   // This prop to give color to the text based on the value
-  pnl: { required: false, type: Boolean, default: false }
+  pnl: { required: false, type: Boolean, default: false },
+  noScramble: { required: false, type: Boolean, default: false }
 });
 
 const {
@@ -66,7 +67,8 @@ const {
   priceOfAsset,
   priceAsset,
   integer,
-  forceCurrency
+  forceCurrency,
+  noScramble
 } = toRefs(props);
 
 const {
@@ -75,7 +77,7 @@ const {
   floatingPrecision
 } = storeToRefs(useGeneralSettingsStore());
 
-const { scrambleData, shouldShowAmount } = storeToRefs(
+const { scrambleData, shouldShowAmount, scrambleMultiplier } = storeToRefs(
   useSessionSettingsStore()
 );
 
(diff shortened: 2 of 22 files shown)


frontend/app/src/components/AssetBalances.vue
```diff
@@ -99,6 +99,7 @@ const sortItems = getSortItems(asset => get(assetInfo(asset)));
     <template #item.usdPrice="{ item }">
       <amount-display
         v-if="item.usdPrice && item.usdPrice.gte(0)"
+        no-scramble
```
kelsos, line 102: What does no-scramble do?

frontend/app/src/components/display/PercentageDisplay.vue
```diff
@@ -26,12 +26,10 @@ const props = defineProps({
...
 
 const { assetPadding, value } = toRefs(props);
-const { shouldShowPercentage, scrambleData } = storeToRefs(
-  useSessionSettingsStore()
-);
+const { shouldShowPercentage } = storeToRefs(useSessionSettingsStore());
 
 const displayValue = computed<string>(() => {
-  if (get(scrambleData) || !get(shouldShowPercentage)) {
+  if (!get(shouldShowPercentage)) {
```
kelsos, line 32: can we also use the multiplier for the percentages?

frontend/app/src/components/display/AmountDisplay.vue
```diff
@@ -115,44 +117,47 @@ const convertFiat = (
...
-
-  // If there is no `sourceCurrency`, it means that no fiat currency, or the unit is asset not fiat, hence we should just show the `value` passed.
-  // If `forceCurrency` is true, we should also just return the value.
-  if (!get(sourceCurrency) || get(forceCurrency)) return get(value);
-
-  const sourceCurrencyVal = get(sourceCurrency)!;
-  const currentCurrencyVal = get(currentCurrency);
-  const priceAssetVal = get(priceAsset);
-  const isCurrentCurrencyVal = get(isCurrentCurrency);
+const realValueBeforeScrambled: ComputedRef<BigNumber> = computed<BigNumber>(
```
kelsos, line 120: nitpick: can we rename it to `internalValue`?

frontend/app/src/components/AssetBalances.vue
lukicenturi: to disable scramble for some prices value, as I said in issue description

frontend/app/src/components/display/PercentageDisplay.vue
lukicenturi: hmm should we? since the value is multiplied by same constant, the percentage shouldn't really matter (should be the same). Plus if we scramble it, the value will not add up.

frontend/app/src/components/AssetBalances.vue
kelsos: Is that to disable the scrambling for the price column?

lukicenturi: yes, but manually by prop

kelsos:
@LefterisJP can you check if this works they way you would like?

kelsos (approved):
lgtm on my side, but let's wait for @LefterisJP to verify that it does what he expects before merging

LefterisJP (commented):
Looking at the scrambler, not related to the numbers but the hex addresses, can we make it smarter?

It seems to scramble all appearances differently. But if it's the same address it should be scrambled to the same one.



For example (1), (2) and (3) above should be the same.


Also as asked in discord is the constant multiplier when activating the setting randomly chosen each time?

lukicenturi:
> Looking at the scrambler, not related to the numbers but the hex addresses, can we make it smarter?
> 
> It seems to scramble all appearances differently. But if it's the same address it should be scrambled to the same one.
> 
> 
> 
> For example (1), (2) and (3) above should be the same.
> 
> Also as asked in discord is the constant multiplier when activating the setting randomly chosen each time?

Hmm yeah I think we can calculate the scrambled hex using the multiplier too. I will update them

lukicenturi (commented):
LGTM

PR #5633, opened 2023-02-18
Move the installation from setup tools to pyproject

yabirgb:
- Replace setup.py with pyproject
- Modify bump script
- Edit CI to get the proper version


LefterisJP (commented):
What does this do? Why is it needed? What am I supposed to check? Code does not have any obvious flaws.

But what do we want to achieve by this and what do you need me to check?

yabirgb:
> What does this do? 

This PR moves the install process from using the `setup.py` file to use `pyproject.toml`

> Why is it needed?

TL;DR: a dependency has issues building in the legacy mode that we need with the current setup.

After introducing the `pyproject.toml` file to the project we had to introduce the `--no-se-pep517` to disable the behaviour that `pip` has during the build process. This is using the `pyproject.toml` file and considering `setup.py` as legacy https://peps.python.org/pep-0517/ .

The error that appear in the builds is related to the usage of this flag and how it behaves with the macos build when it tries to build dependencies because the wheel file is not available.

Since we can move the project and use the `pyproject.toml` to specify how to install rotki this is what this PR does. The `setup.py` as for what I've seen is considered legacy after that PEP https://pip.pypa.io/en/stable/reference/build-system/setup-py/

> What am I supposed to check?

The things that need to be checked by you is the `pyproject.toml` field that should be correctly replaced and if you have any other comment or consideration to go against this change.

> What are we achieving 

Update the project to move away from the now considered legacy `setup.py` and be compatible with deps using the pyproject.toml and need to be built while they don't have a setup.py file

LefterisJP:
>  pyproject.toml field that should be correctly replaced

I don't know what should be contained there nor do I know the intricacies of pyproject. If it works it works. If not you will have to fix it.

pyproject.toml
```diff
@@ -1,3 +1,45 @@
...
+rotkehlchen = "rotkehlchen.__main__:main"
+
+[tool.setuptools.dynamic]
+dependencies = {file = ["requirements.txt"]}
+
+[tool.setuptools_scm]
+fallback_version = "1.27.0"
+
+[tool.setuptools.package-data]
+rotkehlchen = ["*", "chain/ethereum/modules/dxdaomesa/data/contracts.json"]
```
LefterisJP, line 41: What about the stuff under /data?

LefterisJP (commented):
Small question. Also have you generated binaries and tested they work?

LefterisJP (approved):
Since you say you tested it and it work, I will merge and we can fix stuff if we find

PR #5640, opened 2023-02-20
Refactor async endpoints

nebolax:


rotkehlchen/api/v1/resources.py
```diff
@@ -1697,6 +1712,10 @@ def post(
             file: FileStorage,
             **kwargs: Any,
     ) -> Response:
+        _, tmpfilepath = tempfile.mkstemp()
+        file.save(tmpfilepath)
+        self.rest_api.import_tmp_files[file] = Path(tmpfilepath)
+
```
yabirgb, line 1718: Questions about this:

1. Why moving it here?
2. if you already know the filepath why not pass the filepath instead of the file? In the lofic seems that you are using the filepath and not the file itself
3. I would change this to use the TempFile library to avoid forgetting about this files and remove the import_tmp_files if this is possible. I don't see why not right now

rotkehlchen/rotkehlchen.py
```diff
@@ -193,8 +193,8 @@ def maybe_kill_running_tx_query_tasks(
             for greenlet in self.api_task_greenlets:
                 is_evm_tx_greenlet = (
                     len(greenlet.args) >= 1 and
-                    isinstance(greenlet.args[0], MethodType) and
-                    greenlet.args[0].__func__.__qualname__ == 'RestAPI._get_evm_transactions'
+                    isinstance(greenlet.args[0], FunctionType) and
+                    greenlet.args[0].__qualname__ == 'RestAPI.get_evm_transactions'
```
yabirgb, line 197: do we have tests to verify that this change doesn't break the logic?

rotkehlchen/api/rest.py
```diff
@@ -247,6 +247,34 @@ def api_response(
     return response
 
 
+def asynchronous() -> Callable:
```
yabirgb, line 250: Probably is fine but for me `asynchronous` is a word that could mean many things in the context of the code. @LefterisJP might have another view but what do you think about using a name like `async_call`?

rotkehlchen/api/rest.py
```diff
@@ -407,7 +435,13 @@ def query_tasks_outcome(self, task_id: Optional[int]) -> Response:
         }
         return api_response(result=result_dict, status_code=HTTPStatus.NOT_FOUND)
 
-    def _get_exchange_rates(self, given_currencies: list[AssetWithOracles]) -> dict[str, Any]:
+    @asynchronous()
+    def get_exchange_rates(self, given_currencies: list[AssetWithOracles]) -> dict[str, Any]:
+        if len(given_currencies) == 0:
+            return wrap_in_fail_result(
+                message='Empty list of currencies provided',
+                status_code=HTTPStatus.BAD_REQUEST,
```
yabirgb, line 443: Is this missing from the docs? Add it if needed

rotkehlchen/api/v1/resources.py
```diff
@@ -363,7 +364,7 @@ def put(
             self,
             settings: ModifiableDBSettings,
     ) -> Response:
-        return self.rest_api.set_settings(settings)
+        return self.rest_api.set_settings(settings=settings)
```
yabirgb, line 367: why do all of them need to set the kwargs? Or you just did it? If is only one argument there is no place to confusion and is fine afaik

rotkehlchen/api/rest.py
```diff
@@ -4861,33 +3970,6 @@ def _detect_evm_tokens(
...
-    def detect_evm_tokens(
-            self,
-            async_query: bool,
-            only_cache: bool,
-            addresses: Optional[list[ChecksumEvmAddress]],
-            blockchain: SUPPORTED_EVM_CHAINS,
-    ) -> Response:
-        manager = self.rotkehlchen.chains_aggregator.get_chain_manager(blockchain)
-        if addresses is None:
-            addresses = self.rotkehlchen.chains_aggregator.accounts.get(blockchain)
```
yabirgb, line 4873: This is part of the logic that is missing in the method that you made async. You have to be careful with those. Also can you make sure that this is properly tested if is not?

rotkehlchen/api/rest.py
```diff
@@ -4106,52 +3315,18 @@ def _get_avalanche_transactions(
...
-            from_timestamp=from_timestamp,
-            to_timestamp=to_timestamp,
-        )
-
-        result = response['result']
-        if len(result) == 0:
-            return api_response(
-                wrap_in_fail_result('Not found.'),
-                status_code=HTTPStatus.NOT_FOUND,
-            )
```
yabirgb, line 4135: This is missing from the new version if I'm correct

yabirgb (commented):
Additional things:

- In line 2213 there is `noqa` rule that is not needed, can you remove it? (`add_eth2_validator`)

rotkehlchen/api/rest.py
```diff
@@ -2110,64 +1808,49 @@ def _refresh_evm_accounts(self) -> dict[str, Any]:
...
+    def add_single_blockchain_accounts(
             self,
             chain: SUPPORTED_SUBSTRATE_CHAINS,
             account_data: list[SingleBlockchainAccountData[SubstrateAddress]],
     ) -> dict[str, Any]:
         ...
 
     @overload
-    def _add_single_blockchain_accounts(
+    @asynchronous()
```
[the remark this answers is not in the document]
nebolax: I changed it a bit, but now `async_query` argument has to be specified in the overloads and the return type of overloads became `Response`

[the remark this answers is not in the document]
yabirgb: Yes, I don't mean that. What I mean if I'm correct is that you don't need the `@asynchronous` decorator in the methods that have the `@overload` since the definition is empty. You can remove it and leave it only in the method that has the actual behaviour

rotkehlchen/api/v1/resources.py
nebolax: > Why moving it here?

Previously there were 2 methods. `import_data` and `_import_data`. `import_data` (the parent function) had this logic of making a temporary file and then it was asynchronously calling `_import_data`. Now that there is only one function `import_data` which is asynchronous, this logic of creating a temporary file has to be executed before calling `import_data`. I thought that a good place to do it would be `resources.py`

As for whether to pass the path instead of the file directly -- looks reasonable, but I am not sure. This code was made some time ago by @LefterisJP and I don't remember the reason for choosing the current approach.

rotkehlchen/rotkehlchen.py
nebolax: Yes, there is a test.

https://github.com/rotki/rotki/blob/e3a0c09c83eae4cb8ed28d676b7d60cc4107b3bd/rotkehlchen/tests/api/blockchain/test_evm.py#L299

rotkehlchen/api/rest.py
nebolax: Maybe, yeah. If I rename it to `async_api_call`, will that be fine from your perspective?

rotkehlchen/api/rest.py
nebolax: Do you mean the api .rst docs file? If you mean that one, then this is mentioned there https://github.com/rotki/rotki/blob/e3a0c09c83eae4cb8ed28d676b7d60cc4107b3bd/docs/api.rst#L1347

rotkehlchen/api/v1/resources.py
nebolax: Just to have more consistent style. Async endpoints now have to use kwargs, so I thought that to stay in line it will make sense to have all the endpoints use kwargs.

rotkehlchen/api/rest.py
nebolax: Why do you think that it is missing? Isn't it present here https://github.com/rotki/rotki/blob/ae7b18368fc4cbc0b45a94808d102287d6371b52/rotkehlchen/api/v1/resources.py#L2702 ?

rotkehlchen/api/rest.py
yabirgb: yes, for me at least is clearer with that name what the wrapper is doing

rotkehlchen/api/rest.py
yabirgb: Didn't expect it to me moved to the resource file. Okey then

rotkehlchen/api/rest.py
nebolax: Why do you think that it is missing? Isn't it handled here https://github.com/rotki/rotki/blob/ae7b18368fc4cbc0b45a94808d102287d6371b52/rotkehlchen/api/rest.py#L3301 ?

yabirgb: That checks if response is None but not if `len(response) == 0`. In the removed code `[]` also returns status code NOT FOUND

rotkehlchen/api/v1/resources.py
yabirgb: But then is not style, is indeed needed or otherwise the call will fail. Mmmmm this looks like a very easy to write bugs :sweat:

nebolax: This should not cause bugs since if the arguments will be passed as args, the `rest.py`'s method will not even be executed, but rather will explicitly fail.

The `asynchronous` decorator accepts only kwargs, so if the arguments will be passed as args, the `rest.py`'s method will not receive enough arguments for the execution.

rotkehlchen/api/rest.py
nebolax: Previously async wrapper function `get_avalanche_transactions` was reading the result of the function with real logic `_get_avalanche_transactions` and doing some modifications of the status codes on top of it (e.g. getting the `result` and forcing status code to `NOT_FOUND` if length of the result is 0).

Now, since the status codes and messages of sync and async paths of execution became the same, there is now more consistency. Which is good.

Considering the fact that avax transactions endpoint currently is not used by the frontend, this change doesn't break anything, but just gives more consistency for the future, so I think it is a good change.

LefterisJP: No this is not sufficient, yabir is absolutely right? What do you think? That hiding bugs for the future is okay so long as things are "consistent"? Absolutely not. Whether or not something is used plays no role at all, because things change and when they do this is a serious bug which will bite us.

Fortunately it's very easy to solve. The only reason an empty list can be returned there is due to the implementation of what used to be `_get_avalanche_transactions` and what is now `get_avalanche_transactions` in this same file. Don't you see it?

Can't easily link since you are editing it in this PR but look here:

```python
        except RemoteError as e:
            return {
                'result': [],
                'message': f'{str(e)}',
                'status_code': HTTPStatus.BAD_GATEWAY,
            }
        if response is None:
            return {
                'result': [],
                'message': 'Not found.',
                'status_code': HTTPStatus.NOT_FOUND,
            }
```

It returns empty list in both error cases, which is an error in itself :). You know what to do.

rotkehlchen/api/rest.py
```diff
@@ -4262,21 +3415,18 @@ def add_manual_latest_price(
...
         except InputError as e:
-            return api_response(
-                result=wrap_in_fail_result(message=str(e)),
-                status_code=HTTPStatus.CONFLICT,
-            )
+            return wrap_in_fail_result(message=str(e), status_code=HTTPStatus.CONFLICT)
         Inquirer().remove_cache_prices_for_asset(pairs_to_invalidate)
 
-        return api_response(result=OK_RESULT)
+        return OK_RESULT
```
LefterisJP, line 3421: Okay question. Why is this and the other manual latest price related endpoints changing from api response to dict?

rotkehlchen/api/rest.py
LefterisJP: you were right not to expect it to move there. It should not have. Refer to my comment in import data.

rotkehlchen/api/v1/resources.py
```diff
@@ -1353,6 +1364,10 @@ def put(self, async_query: bool, filepath: Path) -> Response:
         @require_loggedin_user()
         @use_kwargs(upload_schema, location='form_and_file')
         def patch(self, async_query: bool, filepath: FileStorage) -> Response:
+            _, tmpfilepath = tempfile.mkstemp()
```
LefterisJP, line 1367: note my comment in the other import data endpoint. This logic shoul dnot be here

rotkehlchen/api/v1/resources.py
```diff
@@ -2676,11 +2699,15 @@ def post(
             only_cache: bool,
             addresses: Optional[list[ChecksumEvmAddress]],
     ) -> Response:
+        manager = self.rest_api.rotkehlchen.chains_aggregator.get_chain_manager(blockchain)
```
LefterisJP, line 2702: as per my other comments let's keep logic out of resources.py

LefterisJP (commented):
Left some comments

rotkehlchen/api/rest.py
nebolax: Hmm looks like the return type should have not changed. Probably I just automatically added the `@async_api_call()` decorator and because of that changed the return type.

In fact this function doesn't get called asynchronously, so probably the decorator should not be here and the return type should be `Response`.

LefterisJP (commented):
needs rebase

Issue #5570, opened 2023-02-06
Allow to filter trades by ignored/not ignored


yabirgb:
## Abstract

Allow users to filter ignored/un ignored trades in the trade page

## Motivation

This is useful when doing changes to the PnL report to check what is ignored and what is not

PR #5641, opened 2023-02-20
Add support for including/excluding ignored trades in api response

prettyirrelevant:
Resolves backend aspect of #5570


rotkehlchen/api/v1/schemas.py
```diff
@@ -453,18 +451,27 @@ def make_trades_query(
...
 
-        if self.treat_eth2_as_eth is True and data['base_asset'] == A_ETH:
+        if treat_eth2_as_eth is True and data['base_asset'] == A_ETH:
             base_assets = (A_ETH, A_ETH2)
-        elif self.treat_eth2_as_eth is True and data['quote_asset'] == A_ETH:
+        elif treat_eth2_as_eth is True and data['quote_asset'] == A_ETH:
             quote_assets = (A_ETH, A_ETH2)
 
+        ignored_trades = None
+        if not data['include_ignored_trades']:
```
nebolax, line 470: To make the reader sure that the `None` / `False` case is not missed here

```suggestion
        if not data['include_ignored_trades']:  # either False or not specified (None)
```

rotkehlchen/db/filtering.py
```diff
@@ -645,6 +646,15 @@ def make(
         if location is not None:
             filter_query.location_filter = DBLocationFilter(and_op=True, location=location)
             filters.append(filter_query.location_filter)
+        if identifiers_to_ignore is not None:
+            filters.append(
+                DBMultiStringFilter(
```
nebolax, line 651: I would say that `DBIgnoreValuesFilter` fits better here

rotkehlchen/api/v1/schemas.py
prettyirrelevant: It can only be a boolean in this case, since a default is specified.

nebolax: :+1: Then I would change it to `is False`

LefterisJP (approved):
LGTM. Just the one comment and the rebase

LefterisJP (approved):
all green. @rotki/frontend should pick it up

PR #5642, opened 2023-02-21
feat: manual EVM account detection in frontend

### LINKED ISSUE #5569: Make EVM address detection user-initiatable and/or periodic

lukicenturi:
Closes #5569

## Checklist


frontend/app/src/pages/balances/blockchain/index.vue
```diff
@@ -197,6 +204,26 @@ const { createAccount, editAccount } = useAccountDialog();
...
+            @click="detectEvmAccounts"
+          >
+            {{ tc('blockchain_balances.evm_detection.title') }}
+          </v-btn>
+        </template>
+        <span>
+          {{ tc('blockchain_balances.evm_detection.tooltip') }}
+        </span>
+      </v-tooltip>
+    </div>
```
kelsos, line 225: This can be a self-contained component right? can you extract it into a new component?

kelsos (approved):
lgtm @lukicenturi thank you

PR #5643, opened 2023-02-21
Updates start-dev script to allow both localhost and ip access

kelsos:
- To fix a cors problem with when running dev:web

Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/scripts/start-dev.js
+++ b/frontend/scripts/start-dev.js
@@ -91,7 +91,7 @@ pids[commonProcesses.pid] = COMMON;
 if (noElectron) {
   logger.info('Starting python backend');
   const backendProcess = spawn(
-    'python -m rotkehlchen --rest-api-port 4242 --websockets-api-port 4244 --api-cors http://localhost:8080 --loglevel debug --max-size-in-mb-all-logs 120',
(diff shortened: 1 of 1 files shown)


frontend/scripts/start-dev.js
```diff
@@ -91,7 +91,7 @@ pids[commonProcesses.pid] = COMMON;
 if (noElectron) {
   logger.info('Starting python backend');
   const backendProcess = spawn(
-    'python -m rotkehlchen --rest-api-port 4242 --websockets-api-port 4244 --api-cors http://localhost:8080 --loglevel debug --max-size-in-mb-all-logs 120',
+    'python -m rotkehlchen --rest-api-port 4242 --api-cors http://localhost:8080,http://127.0.0.1:8080 --loglevel debug --max-size-in-mb-all-logs 120',
```
lukicenturi, line 94: why do we remove the `--websockets-api-port` ?

kelsos: > why do we remove the `--websockets-api-port` ?

Because it is not needed, initially, we had two different ports for WebSockets and API. This is not the case anymore.

If you see ws connects to `4242:/ws` and the rest API handles the upgrade to a WebSocket connection.

PR #5654, opened 2023-02-23
Fees settings

### LINKED ISSUE #5592: Make the per-event fees accounting calculations clear.
## Abstract

At the moment all events are standalone and when processed into accounting to calculate cost basis and PnL do not group together. For example a simple trade with a fee(Buy 1 Ξ for 10 € with 0.5 € fee) would be split into three separate events:
- In event for 1 Ξ
- Out event for 10 €
- Out (Fee) event for 0.5 €

And then they are processed during accounting one by one, but the price is figured out by combining the in/out events and the fees.

For most case this should be fine but there is some accounting cost basis calculations when fee should also be taken into account in the specific trade like ACB: https://github.com/rotki/rotki/issues/5561#issuecomment-1423338938

## Specification

Think if any special handling should be made in the fees of swaps/trades or other places so they work properly with all accounting methods instead of just being generaly deducted as a loss.

At the moment the places fees are processed are:

1. Processing of a `Trade` data structure for accounting: https://github.com/rotki/rotki/blob/a1a43b69966ceb153a53dbda3446e3ff9fc42090/rotkehlchen/exchanges/data_structures.py#L349
2. Processing of a `MargingPosition` data structure for accounting: https://github.com/rotki/rotki/blob/a1a43b69966ceb153a53dbda3446e3ff9fc42090/rotkehlchen/exchanges/data_structures.py#L572
3. Processing of `EVMTransaction` swap events: https://github.com/rotki/rotki/blob/a1a43b69966ceb153a53dbda3446e3ff9fc42090/rotkehlchen/accounting/transactions.py#L103

We need to think how fees should be taken into account.

At the moment, the approach seems to be a bit weirdly mixed. The fee in most of the places is just as an extra deducted loss by the fee out event. But in swaps in the above linked locations, only in a specific string of if conditions when calculating the price of the `out_event` during `get_prices_for_swap()` it takes the fee into account and incorporates it into the price: https://github.com/rotki/rotki/blob/a1a43b69966ceb153a53dbda3446e3ff9fc42090/rotkehlchen/accounting/pot.py#L382

This approach is convoluted and can double count fees. The fee_out event should always be there, since PnL aside we also keep track of amounts of each asset owned, and fees out reduces the amount of the fee asset.

The two different approaches should probably (need thinking) be:
1. The fee_out event should be fully taxable, and as such count as "loss" due to fees. (This is what happens right now in most cases)
2. OR the fee_out event should just be there but not count as loss. Just count amount reduction (so taxable=False). Instead the fee should be incorporated into either the buy or sell event price calculation as seen above in `get_prices_for_swap()`

nebolax:
Closes #5592

## Checklist

### DIFF
--- a/rotkehlchen/accounting/pot.py
+++ b/rotkehlchen/accounting/pot.py
@@ -324,23 +324,22 @@ class AccountingPot(CustomizableDateMixin):
             asset_in: Asset,
             amount_out: FVal,
             asset_out: Asset,
-            fee: Optional[FVal],
-            fee_asset: Optional[Asset],
+            fee_info: Optional[tuple[FVal, Asset]],
     ) -> Optional[tuple[Price, Price]]:
         """
         Calculates the prices for assets going in and out of a swap/trade.
 
-        The rules are:
-        - For the asset_in we get the equivalent rate from asset_out + fee if any.
-        If there is no price found for fee_currency we ignore it.
-        If there is no price for asset_out then we switch to using the asset_in price itself.
-        If neither of the 2 assets can have their price known, we bail.
+        The algorithm is:
+        1. Query oracles for prices of asset_out and asset_in.
+        2.1 If either of the assets is fiat -- use its amount and price for calculations.
+        2.2. If neither of the assets is fiat -- use `out_price` if `out_price` is known,
+        otherwise `in_price`.
+        3.1 If `fee_info` is provided, fee is included in the price of one of the assets.
+        3.2. If `asset_out` is fiat -- fee is added to `calculated_in_price`.
+        3.3. If `asset_in` is fiat -- fee is subtracted from `calculated_out_price`.
+        3.4. Otherwise fee is added to the price of the asset that was bought.
 
-        - For the asset_out we get the equivalent rate from asset_in.
-        if there is no price found for asset_in then we switch to using the asset_out price.
-        If neither of the 2 assets can have their price known we bail.
-
-        Returns (out_price, in_price) or None if it can't find proper prices
+        Returns (calculated_out_price, calculated_in_price) or None if it can't find proper prices.
         """
         if ZERO in (amount_in, amount_out):
             log.error(
@@ -356,16 +355,6 @@ class AccountingPot(CustomizableDateMixin):
         except (PriceQueryUnsupportedAsset, NoPriceForGivenTimestamp, RemoteError):
             out_price = None
 
-        fee_price = None
-        if fee is not None and fee_asset is not None and fee != ZERO:
-            # also checking fee_asset != None due to https://github.com/rotki/rotki/issues/4172
-            try:
-                fee_price = self.get_rate_in_profit_currency(
-                    asset=fee_asset,
-                    timestamp=timestamp,
-                )
-            except (PriceQueryUnsupportedAsset, NoPriceForGivenTimestamp, RemoteError):
-                fee_price = None
         try:
             in_price = self.get_rate_in_profit_currency(
                 asset=asset_in,
@@ -378,20 +367,53 @@ class AccountingPot(CustomizableDateMixin):
             if e.rate_limited is True and out_price is None:
                 raise  # in_price = out_price = None -> notify user
 
-        if out_price is None and in_price is None:
+        if fee_info is not None:
+            try:
+                fee_price = self.get_rate_in_profit_currency(
+                    asset=fee_info[1],
+                    timestamp=timestamp,
+                )
+            except (PriceQueryUnsupportedAsset, RemoteError):
+                fee_price = None
+
+        # Determine whether to use `out_price` or `in_price` for calculations
+        price_to_use: Literal['in', 'out']
+        if asset_out.is_fiat() and asset_out is not None:
+            price_to_use = 'out'  # Use `out_price` if `asset_out` is fiat
+        elif asset_in.is_fiat() and asset_in is not None:
+            price_to_use = 'in'  # Use `in_price` if `asset_in` is fiat
+        elif out_price is not None:
+            price_to_use = 'out'  # Prefer `out_price` over `in_price`
+        elif in_price is not None:
+            price_to_use = 'in'
+        else:  # Can't proceed if there is no price known
             return None
 
-        if out_price is not None:
-            paid = amount_out * out_price
-            if fee_price is not None:
-                paid += fee_price * fee  # type: ignore # fee should exist here
-            calculated_in = Price(paid / amount_in)
+        if price_to_use == 'in':
+            total_paid = amount_in * in_price  # type: ignore[operator]  # in_price is not None
         else:
-            calculated_in = in_price  # type: ignore # in_price should exist here
+            total_paid = amount_out * out_price  # type: ignore[operator]  # out_price is not None
 
-        if in_price is not None:
-            calculated_out = Price((amount_in * in_price) / amount_out)
-        else:
-            calculated_out = out_price  # type: ignore # out_price should exist here
+        if asset_in.is_fiat():
+            if fee_info is not None and fee_price is not None:
+                total_paid -= fee_price * fee_info[0]  # Subtract fee from cost basis
 
-        return (calculated_out, calculated_in)
+            calculated_out_price = Price(total_paid / amount_out)
+
+            if price_to_use == 'in':
+                calculated_in_price = in_price
+            else:
+                calculated_in_price = Price((amount_out * out_price) / amount_in)  # type: ignore[operator]  # out_price is not None  # noqa: E501
+
+        else:  # if asset_out is fiat or both assets are crypto or both are fiat
+            if fee_info is not None and fee_price is not None:
+                total_paid += fee_price * fee_info[0]  # Add fee to cost basis
+
+            calculated_in_price = Price(total_paid / amount_in)
+
+            if price_to_use == 'out':
+                calculated_out_price = out_price  # type: ignore[assignment]  # out_price is not None  # noqa: E501
+            else:
+                calculated_out_price = Price((amount_in * in_price) / amount_out)  # type: ignore[operator]  # in_price is not None  # noqa: E501
+
+        return (calculated_out_price, calculated_in_price)  # type: ignore[return-value]  # calculated_in_price is not None  # noqa: E501
--- a/rotkehlchen/exchanges/data_structures.py
+++ b/rotkehlchen/exchanges/data_structures.py
@@ -368,21 +368,26 @@ class Trade(AccountingEventMixin):
         else:  # settlement buy/sell only in poloniex. Should properly process when margin
             return 1  # trades are implemented
 
+        group_id = self.identifier
+        taxable = asset_in.is_fiat() or accounting.settings.include_crypto2crypto
+
+        should_calculate_fee = self.fee is not None and self.fee_currency is not None and self.fee != ZERO  # noqa: E501
+
+        fee_info_for_cost_basis = None
+        if should_calculate_fee and accounting.settings.include_fees_in_cost_basis is True:
+            fee_info_for_cost_basis = (self.fee, self.fee_currency)
+
         prices = accounting.get_prices_for_swap(
             timestamp=self.timestamp,
             amount_in=amount_in,
             asset_in=asset_in,
             amount_out=amount_out,
             asset_out=asset_out,
-            fee=self.fee,
-            fee_asset=self.fee_currency,
+            fee_info=fee_info_for_cost_basis,  # type: ignore[arg-type]  # mypy doesn't see that `should_calculate_fee` condition makes sure that the values are not None.  # noqa: E501
         )
         if prices is None:
             log.debug(f'Skipping {self} at accounting due to inability to find a price')
             return 1
-
-        group_id = self.identifier
-        taxable = asset_in.is_fiat() or accounting.settings.include_crypto2crypto
 
         _, trade_taxable_amount = accounting.add_spend(
             event_type=AccountingEventType.TRADE,
@@ -408,8 +413,7 @@ class Trade(AccountingEventMixin):
             extra_data={'group_id': group_id},
         )
 
-        if self.fee is not None and self.fee_currency is not None and self.fee != ZERO:
-            # also checking fee_asset != None due to https://github.com/rotki/rotki/issues/4172
+        if should_calculate_fee:  # include fee as a standalone event
             fee_price = None
             if self.fee_currency == accounting.profit_currency:
                 fee_price = Price(ONE)
@@ -418,19 +422,28 @@ class Trade(AccountingEventMixin):
             elif self.fee_currency == asset_out:
                 fee_price = prices[0]
 
+            if accounting.settings.include_fees_in_cost_basis:
+                # If fee is included in cost basis, we just reduce the amount of fee asset owned
+                fee_taxable = False
+                fee_taxable_amount_ratio = ONE
+            else:
+                # Otherwise we make it a normal spend event
+                fee_taxable = True
+                fee_taxable_amount_ratio = trade_taxable_amount / amount_out
+
             accounting.add_spend(
                 event_type=AccountingEventType.FEE,
                 notes=notes + 'Fee',
                 location=self.location,
                 timestamp=self.timestamp,
-                asset=self.fee_currency,
-                amount=self.fee,
-                taxable=True,
+                asset=self.fee_currency,  # type: ignore[arg-type]  # mypy doesn't see that `calculate_fee` condition makes sure that the values are not None.  # noqa: E501
+                amount=self.fee,  # type: ignore[arg-type]  # mypy doesn't see that `calculate_fee` condition makes sure that the values are not None.  # noqa: E501
+                taxable=fee_taxable,
                 given_price=fee_price,
                 # By setting the taxable amount ratio we determine how much of the fee
                 # spending should be a taxable spend and how much free.
-                taxable_amount_ratio=trade_taxable_amount / amount_out,
-                count_cost_basis_pnl=accounting.settings.include_crypto2crypto,
+                taxable_amount_ratio=fee_taxable_amount_ratio,
+                count_cost_basis_pnl=True,
                 count_entire_amount_spend=True,
                 extra_data={'group_id': group_id},
             )
@@ -590,7 +603,7 @@ class MarginPosition(AccountingEventMixin):
             amount=amount,
             taxable=True,
         )
-        if self.fee != ZERO:
+        if self.fee != ZERO:  # Fee is not included in the asset price here since it is not a swap/trade event.  # noqa: E501
             accounting.add_spend(
                 event_type=AccountingEventType.FEE,
                 notes='Margin position. Fee',
--- a/rotkehlchen/tests/api/test_history.py
+++ b/rotkehlchen/tests/api/test_history.py
@@ -49,11 +49,20 @@ from rotkehlchen.utils.misc import ts_now
     'start_ts,end_ts',
     [(0, 1601040361), (1539713237, 1539713238)],
 )
+@pytest.mark.parametrize('db_settings', [
+    {'include_fees_in_cost_basis': True},
+    {'include_fees_in_cost_basis': False},
+])
 def test_query_history(rotkehlchen_api_server_with_exchanges, start_ts, end_ts):
     """Test that the history processing REST API endpoint works. Similar to test_history.py
 
     Both a test for full and limited time range.
+    Also tests different values of `include_fees_in_cost_basis`.
     """
+    rotki = rotkehlchen_api_server_with_exchanges.rest_api.rotkehlchen
+    with rotki.data.db.conn.read_ctx() as cursor:
+        fees_in_cost_basis = rotki.data.db.get_settings(cursor).include_fees_in_cost_basis
+
     report_id, report_result, events_result = query_api_create_and_get_report(
         server=rotkehlchen_api_server_with_exchanges,
         start_ts=start_ts,
@@ -78,13 +87,18 @@ def test_query_history(rotkehlchen_api_server_with_exchanges, start_ts, end_ts):
         assert overview[str(AccountingEventType.ASSET_MOVEMENT)] is not None
         assert overview[str(AccountingEventType.MARGIN_POSITION)] is not None
         assert overview[str(AccountingEventType.TRANSACTION_EVENT)] is not None
-    else:
+        assert overview[str(AccountingEventType.FEE)] is not None
+    elif fees_in_cost_basis is True:  # start_ts is 1539713237
+        assert len(overview) == 1
+        # Fees events are not taxable in this case, so they should not be included in total pnls.
+        assert str(AccountingEventType.FEE) not in overview
+    else:  # start_ts is 1539713237 and fees_in_cost_basis is False
         assert len(overview) == 2
+        assert overview[str(AccountingEventType.FEE)] is not None
     assert overview[str(AccountingEventType.TRADE)] is not None
-    assert overview[str(AccountingEventType.FEE)] is not None
 
     settings = report['settings']
-    assert len(settings) == 8
+    assert len(settings) == 9
     assert settings['profit_currency'] == 'EUR'
     assert settings['account_for_assets_movements'] is True
     assert settings['calculate_past_cost_basis'] is True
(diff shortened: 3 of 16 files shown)


rotkehlchen/accounting/pot.py
```diff
@@ -349,49 +362,70 @@ def get_prices_for_swap(
...
-            paid = amount_out * out_price
-            if fee_price is not None:
-                paid += fee_price * fee  # type: ignore # fee should exist here
-            calculated_in = Price(paid / amount_in)
+        # Calculate in price and optionally include a fee in this price
+        if oracle_out_price is not None:
+            total_paid = amount_out * oracle_out_price  # Use calculated_out_price
         else:
-            calculated_in = in_price  # type: ignore # in_price should exist here
+            total_paid = amount_in * oracle_in_price  # type: ignore  # in_price should exist here  # noqa: E501
```
prettyirrelevant, line 403: `oracle_in_price` can most definitely not exist here. The conditional on L396 only checks **if both are None**

rotkehlchen/accounting/pot.py
```diff
@@ -349,49 +362,70 @@ def get_prices_for_swap(
...
+
+        if fee_info is not None and oracle_fee_price is not None:
```
prettyirrelevant, line 405: there's a check above for `fee_info is not None`, so this might be redundant here. 
Unless you had other ideas or saw something during implementation.

rotkehlchen/tests/unit/accounting/test_basic.py
```diff
@@ -79,8 +79,8 @@ def test_selling_crypto_bought_with_crypto(accountant, google_service):
     assert sells[0].rate.is_close(FVal('386.03406326'))
 
     expected_pnls = PnlTotals({
-        AccountingEventType.TRADE: PNL(taxable=FVal('74.3118704999540625'), free=ZERO),
-        AccountingEventType.FEE: PNL(taxable=FVal('-0.419658351381311222'), free=ZERO),
+        AccountingEventType.TRADE: PNL(taxable=FVal('74.3564936249540625'), free=ZERO),
```
prettyirrelevant, line 82: Question. Why is there a disparity in values?

prettyirrelevant (commented):
Relevant tests failed.
`test_margin_events_affect_gained_lost_amount[mocked_price_queries0]`
`test_fees_in_received_asset[mocked_price_queries0]`

rotkehlchen/accounting/pot.py
nebolax: Yes. So when the execution comes to this block
```python
if oracle_out_price is not None:
    total_paid = amount_out * oracle_out_price
else:
    total_paid = amount_in * oracle_in_price 
```

we know for sure that at least one of `oracle_out_price` or `oracle_in_price` exists.

So if we come to this `else` block (line 403) we know that `oracle_out_price` is `None` and thus  `oracle_in_price` is NOT `None`.

rotkehlchen/accounting/pot.py
nebolax: Hmm well yeah, this is mostly to satisfy mypy. Maybe can be changed to `type: ignore` instead

prettyirrelevant (commented):
Might be an overkill but why not create a `FeeCalculator` mixin, define a method `should_calculate_fee` and use where needed. What do you think?

Also, relevant tests failed.

nebolax:
> Might be an overkill but why not create a FeeCalculator mixin, define a method should_calculate_fee and use where needed. What do you think?

I would say that it is an overkill. Especially considering that `pot` is not a member of `AccountingEventMixin`. And also considering that currently fee is calculated in only a single place I think that a mixin is not needed.

rotkehlchen/exchanges/data_structures.py
```diff
@@ -368,22 +368,28 @@ def process(
...
+        group_id = self.identifier
+        taxable = asset_in.is_fiat() or accounting.settings.include_crypto2crypto
+
+        calculate_fee = self.fee is not None and self.fee_currency is not None and self.fee != ZERO
+
+        if calculate_fee is False or accounting.include_fees_in_cost_basis() is False:
+            fee_info_for_cost_basis = None
+        else:
+            fee_info_for_cost_basis = (self.fee, self.fee_currency)
+
```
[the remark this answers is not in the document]
nebolax: No, can't be done like this because `calculate_fee` is reused below

rotkehlchen/accounting/pot.py
```diff
@@ -349,49 +357,53 @@ def get_prices_for_swap(
...
-        if out_price is not None:
-            paid = amount_out * out_price
-            if fee_price is not None:
-                paid += fee_price * fee  # type: ignore # fee should exist here
-            calculated_in = Price(paid / amount_in)
+        if oracle_out_price is not None:
+            total_paid = amount_out * oracle_out_price  # Use calculated_out_price
         else:
-            calculated_in = in_price  # type: ignore # in_price should exist here
+            total_paid = amount_in * oracle_in_price  # type: ignore  # in_price should exist here  # noqa: E501
```
LefterisJP, line 397: Here and in other parts in this code you should rename in_price to oracle_in_price and out price
```suggestion
            total_paid = amount_in * oracle_in_price  # type: ignore  # oracle_in_price should exist here  # noqa: E501
```

But why add the `oracle_` prefix everywhere? It just makes it all longer.

rotkehlchen/accounting/pot.py
LefterisJP: If there is already a check and mypy does not see it add type ignore with an explaining comment.

rotkehlchen/accounting/pot.py
```diff
@@ -324,11 +331,12 @@ def get_prices_for_swap(
             asset_in: Asset,
             amount_out: FVal,
             asset_out: Asset,
-            fee: Optional[FVal],
-            fee_asset: Optional[Asset],
+            fee_info: Optional[tuple[FVal, Asset]],
     ) -> Optional[tuple[Price, Price]]:
         """
         Calculates the prices for assets going in and out of a swap/trade.
+        If fee is provided, includes it in the buy cost basis. If fee should be treated as a
```
LefterisJP, line 338:
```suggestion
        If fee_info is provided, it is included in the buy cost basis. If fee should be treated as a
```

rotkehlchen/exchanges/data_structures.py
```diff
@@ -408,8 +414,7 @@ def process(
             extra_data={'group_id': group_id},
         )
 
-        if self.fee is not None and self.fee_currency is not None and self.fee != ZERO:
-            # also checking fee_asset != None due to https://github.com/rotki/rotki/issues/4172
+        if should_calculate_fee and accounting.include_fees_in_cost_basis() is False:
```
LefterisJP, line 417:
```suggestion
        if should_calculate_fee and accounting.include_fees_in_cost_basis() is False:  # include fee as a standalone event
```

rotkehlchen/exchanges/data_structures.py
```diff
@@ -591,6 +596,7 @@ def process(
             taxable=True,
         )
         if self.fee != ZERO:
+            # Fee is never included in the asset price here since it is not a swap/trade event.
```
LefterisJP, line 599:
```suggestion
        if self.fee != ZERO:  # Fee is not included in the asset price here since it is not a swap/trade event.
```

rotkehlchen/tests/unit/accounting/test_basic.py
```diff
@@ -32,7 +32,7 @@
         {'cost_basis_method': CostBasisMethod.ACB},
         PnlTotals({
             AccountingEventType.TRADE: PNL(taxable=FVal('551.2588760825750541683933333'), free=ZERO),  # noqa: E501
-            AccountingEventType.FEE: PNL(taxable=FVal('-0.2640127597017956302424120028'), free=ZERO),  # noqa: E501
+            AccountingEventType.FEE: PNL(taxable=ZERO, free=ZERO),
```
LefterisJP, line 35: Doesn't the trade one also change when fees are taken into account in cost basis? This is weird and sounds wrong

rotkehlchen/tests/unit/accounting/test_basic.py
```diff
@@ -130,11 +130,11 @@ def test_buy_event_creation(accountant):
     assert len(buys) == 2
     assert buys[0].amount == FVal(5)
     assert buys[0].timestamp == 1476979735
-    assert buys[0].rate.is_close('578.6438412')  # (578.505 * 5 + 0.0012 * 578.505)/5
+    assert buys[0].rate.is_close('578.505')  # pure price, fee is counted as a separate event
```
LefterisJP, line 133: As I said this should not be like that. Should be configurable and probably set as default for all to be part of the price.

rotkehlchen/tests/unit/test_cost_basis.py
```diff
@@ -1047,3 +1060,80 @@ def test_taxable_acquisition(accountant):
...
+    [{'cost_basis_method': CostBasisMethod.ACB}, [ZERO, ZERO, FVal(3495), ZERO, ZERO, ZERO, FVal(-6), ZERO]],  # noqa: E501
+])
+def test_fees(accountant: 'Accountant', expected_pnls: list[FVal]):
+    """
+    Tests that fees are properly either calculated as standalone events or included in the price.
+    Values for the example are taken from https://github.com/rotki/rotki/issues/5561#issuecomment-1423338938.
+
+    Note: pnls are $10 higher than in the issue since examples in the issue say that transaction
+    costs should be subtracted when selling, but in fact transaction costs are always about
+    buy cost basis https://discord.com/channels/657906918408585217/683245846070165504/1077994503849386054.
```
LefterisJP, line 1082: What? That sounds confusing.

1. Do not write things that are intended for the reviewer as part of the docstring. Comment on the PR itself. This won't stay in the docstring after the review.
2. What are you talking aout? Which example? That comment is huge and has different examples written by me based on two different examples from two different pdfs.

```suggestion
    Note: pnls are $10 higher than in the issue since examples in the issue say that transaction
    costs should be subtracted when selling, but in fact transaction costs are always about
    buy cost basis https://discord.com/channels/657906918408585217/683245846070165504/1077994503849386054.
```

rotkehlchen/tests/unit/test_cost_basis.py
```diff
@@ -1047,3 +1060,80 @@ def test_taxable_acquisition(accountant):
...
+            fee=Fee(FVal(10)),
+            fee_currency=A_EUR,
+            notes='Trade 4',
+        ),
+    ]
+    accounting_history_process(
+        accountant=accountant,
+        start_ts=Timestamp(0),
+        end_ts=Timestamp(1677593077),
+        history_list=list(history),  # converting to list to satisfy mypy
```
LefterisJP, line 1135: what? Why do you even need to declare the sequence type in a test in the first place? Don't add unecessary stuff in the tests

rotkehlchen/tests/unit/test_cost_basis.py
```diff
@@ -1047,3 +1060,80 @@ def test_taxable_acquisition(accountant):
     assert pot.processed_events[0].free_amount == ZERO
     assert pot.processed_events[0].pnl.taxable == ONE
     assert pot.processed_events[0].pnl.free == ZERO
+
+
+@pytest.mark.parametrize('force_no_price_found_for', [[
```
LefterisJP, line 1065: Why not just use simple mocked prices?

LefterisJP (commented):
Left many comments, most major of which is that this should be a customizable setting and should most probably be set to always go in cost basis by default.

rotkehlchen/tests/unit/test_cost_basis.py
nebolax: Because we essentially don't need some prices for certain timestamps, but if there is no mocked price provided and it is not explicitly specified to raise an error, then there is an assertion error raised.

I thought that telling to raise an error is better than providing mocked prices because it shows that we in fact don't need these prices.

LefterisJP: No it's not. It's confusing to the reader, it's a much more complicated fixture and is not used as much in the code. And never used for this purpose. Please use the same mocked prices logic we use elsewhere.

rotkehlchen/tests/unit/test_cost_basis.py
nebolax: I meant Canada example

rotkehlchen/tests/unit/accounting/test_basic.py
nebolax: There is no change because of the events that were used for this test.

The event with a fee is #3 in `history1` list. The only event after #3 is event #4.

Then because of:
1. Amount spent by event #4 is quite low.
2. FIFO is used

Only acquisition from event #2 is used and thus the fee change in event #3 is not reflected in the pnl result.

rotkehlchen/tests/api/test_history.py
```diff
@@ -95,7 +96,7 @@ def test_query_history(rotkehlchen_api_server_with_exchanges, start_ts, end_ts):
     assert settings['eth_staking_taxable_after_withdrawal_enabled'] is False
 
     assert events_result['entries_limit'] == FREE_PNL_EVENTS_LIMIT
-    entries_length = 43 if start_ts == 0 else 40
+    entries_length = 35 if start_ts == 0 else 32
```
LefterisJP, line 99: Is this reduction due to the fees being included in the cost basis and thus not having standalone events? Couldn't this be made configurable here to also see standalone event approach still works?

rotkehlchen/tests/unit/accounting/test_prefork_assets.py
```diff
@@ -14,6 +14,7 @@
 
 @pytest.mark.parametrize('mocked_price_queries', [prices])
 @pytest.mark.parametrize('ignored_assets', [[A_ETC], []])
+@pytest.mark.parametrize('db_settings', [{'include_fees_in_cost_basis': False}])
```
LefterisJP, line 17: Why making it false here and below?

rotkehlchen/tests/unit/accounting/test_prefork_assets.py
```diff
@@ -140,18 +142,19 @@ def test_buying_selling_btc_before_bchfork(accountant, google_service):
     assert len(buys) == 1
     assert buys[0].remaining_amount == amount_bch
     assert buys[0].timestamp == 1491593374
-    assert buys[0].rate.is_close('1128.98961538')
+    assert buys[0].rate.is_close('1128.905')
```
LefterisJP, line 145: Why did this and the other numbers in the test change? With inlclude in cost basis being False, the result should be same as before?

The change is super super small, so sounds tricky. What changed? Was it doing some double counting of fees before?

rotkehlchen/tests/unit/accounting/test_prefork_assets.py
```diff
@@ -259,8 +262,8 @@ def test_buying_selling_bch_before_bsvfork(accountant, google_service):
     assert accountant.pots[0].cost_basis.get_calculated_asset_amount(A_BSV) == amount_bsv
     expected_pnls = PnlTotals({
         AccountingEventType.TRADE: PNL(
-            taxable=FVal('13877.57646153846153846153846'),
-            free=FVal('-464.4416923076923076923076920'),
+            taxable=FVal('13877.898'),
```
LefterisJP, line 265: exact same question here

rotkehlchen/tests/unit/accounting/test_settings.py
```diff
@@ -45,20 +45,23 @@
...
     'include_crypto2crypto': False,
+    'include_fees_in_cost_basis': False,
 }])
 def test_nocrypto2crypto(accountant, google_service):
     accounting_history_process(accountant, 1436979735, 1519693374, history5)
     no_message_errors(accountant.msg_aggregator)
     expected_pnls = PnlTotals({
-        AccountingEventType.TRADE: PNL(taxable=ZERO, free=FVal('264693.433642820')),
-        AccountingEventType.FEE: PNL(taxable=FVal('-1.172677782721563021308995038'), free=ZERO),
+        AccountingEventType.TRADE: PNL(taxable=ZERO, free=FVal('264693.43364282')),
```
LefterisJP, line 54: exact same question on the very small number change here ... and basically every other place in this file. Won't comment on all of them.

rotkehlchen/tests/unit/test_cost_basis.py
```diff
@@ -1047,3 +1059,87 @@ def test_taxable_acquisition(accountant):
...
+    },
+}])
+@pytest.mark.parametrize(('db_settings', 'expected_pnls'), [
+    (
+        {'cost_basis_method': CostBasisMethod.FIFO, 'include_fees_in_cost_basis': False},
+        [ZERO, ZERO, FVal(-10), FVal(3500), ZERO, FVal(-10), ZERO, ZERO, FVal(-10), FVal(1600), ZERO, FVal(-10)],  # noqa: E501
+    ),
+    (
+        {'cost_basis_method': CostBasisMethod.ACB, 'include_fees_in_cost_basis': True},
+        [ZERO, ZERO, FVal(3495), ZERO, ZERO, ZERO, FVal(-6), ZERO],
```
LefterisJP, line 1081: Continuing the conversation from before. 

> costs should be subtracted when selling, but in fact transaction costs are always about
    buy cost basis 

This does not make sense. I think you mean you want to keep the fees to add them to the buy cost basis. This I am not sure about. Perhaps it should be configurable? Perhaps it's an overkill ... but in this example ... it shoudl not be part of the buy cost basis.

Since the "buy" part of the sell here is fiat. And for fiat we do not count cost basis. Fiat is the profit currency. So with what you wrote the 10$ fee is lost here. The results for this example should be the same as my comment issue that you linked, not with $10 more. It's also same as here: https://www.adjustedcostbase.ca/blog/how-to-calculate-adjusted-cost-base-acb-and-capital-gains/


Now .... there is a valid question about what to do if it was something like sell ETH for BTC. There .. I think we can go with the buy cost basis thing. So in short, I believe:
1. This example should match the issue you linked.
2. You need to configure the implementation to act differently depending on if the sell is for FIAT or not.
    a. If it's for FIAT, then you should add the fee as per this example
    b. If it's for something that also gets its own cost basis, then add it to its cost basis.

And maybe... (but let's leave that for later) this behaviour can also be customized.

nebolax: Makes sense

rotkehlchen/tests/unit/accounting/test_prefork_assets.py
nebolax: I made `include_fees_in_cost_basis` `False` in some tests to see that the previous approach with having fees as standalone events still works.

rotkehlchen/tests/api/test_history.py
nebolax: Yes, is because now there are no fee events.

I'll check, but it should be easy to parametrize this test.

rotkehlchen/tests/unit/accounting/test_misc.py
```diff
@@ -125,8 +125,7 @@ def test_fees_count_in_cost_basis(accountant, google_service):
     )
 
     expected_pnls = PnlTotals({
-        AccountingEventType.TRADE: PNL(taxable=FVal('1550.055'), free=ZERO),
-        AccountingEventType.FEE: PNL(taxable=FVal('-300.630'), free=ZERO),
+        AccountingEventType.TRADE: PNL(taxable=FVal('319.395'), free=ZERO),
```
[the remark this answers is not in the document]
nebolax: Here the difference is caused by this fee https://github.com/rotki/rotki/blob/30309d115687c78a2ba0d9d713338d4f297f910b/rotkehlchen/tests/unit/accounting/test_misc.py#L104

In the previous PR status fee was always added to buy cost basis. Now I changed to take into account cases when one of the assets is fiat.

So previously `0.5` ETH * `1800` EUR /ETH = `+900` EUR to cost basis.

But since in the event that I linked ETH is sold for EUR, now it is `0.5` ETH * `1800` EUR /ETH = `-900` EUR to cost basis.

So in total there is a `1800` EUR difference in cost basis due to the fee, which caused this huge number change

rotkehlchen/tests/unit/accounting/test_basic.py
```diff
@@ -44,7 +44,23 @@ def test_simple_accounting(accountant, google_service, expected_pnls):
 
 
 @pytest.mark.parametrize('mocked_price_queries', [prices])
-def test_selling_crypto_bought_with_crypto(accountant, google_service):
+@pytest.mark.parametrize(('db_settings', 'expected_pnl_totals'), [
+    (
+        {'include_fees_in_cost_basis': True},
+        PnlTotals({
+            AccountingEventType.TRADE: PNL(taxable=FVal('40.58809373210132103047999998'), free=ZERO),  # noqa: E501
```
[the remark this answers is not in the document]
nebolax: Here the difference is because I changed the prices calculation to be more consistent.

So previously:

1. Here https://github.com/rotki/rotki/blob/bugfixes/rotkehlchen/accounting/pot.py#L384 `out_price` was preferred
2. But here https://github.com/rotki/rotki/blob/bugfixes/rotkehlchen/accounting/pot.py#L392 `in_price` was preferred

Now what happened is that:
1. I changed the calculation so that `in_price` is checked first everywhere
2. In this test `rate` here https://github.com/rotki/rotki/blob/bugfixes/rotkehlchen/tests/unit/accounting/test_basic.py#L56 is quite different from the mocked price here https://github.com/rotki/rotki/blob/f87130989c512d60b05aabde96889bdc22dccd75/rotkehlchen/tests/utils/history.py#L72. So when I adjusted the calculation logic, the difference ended up being quite large.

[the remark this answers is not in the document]
LefterisJP: Hmmm what do you mean by "is preferred"? I do not follow this explanation. The links you gave me to bugfixes before compute `calculated_in` depending on existence and value of `out_price`, and `calculated_out` depending on existence and value of `in_price`.

LefterisJP (commented):
I need two things to continue here:

1. A very detailed explanation of the logic change you did and explanation of what you want to achieve.
2. I need you to carefully go through each test touched and check and see if numbers add up, if comments should be edited and follow them with the logic as I did with the one with 0.5 ETH fee, to detect if stuff make sense.

rotkehlchen/tests/unit/accounting/test_misc.py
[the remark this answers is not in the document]
nebolax: Yes, makes sense

rotkehlchen/tests/unit/accounting/test_basic.py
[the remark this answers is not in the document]
nebolax: Discussed and settled on the following priority:

- If asset_out or asset_in is fiat --> use price of the fiat asset.
- If neither of the assets is fiat --> use out_price if known, otherwise in_price.

rotkehlchen/accounting/pot.py
```diff
@@ -378,20 +369,53 @@ def get_prices_for_swap(
...
+            try:
+                fee_price = self.get_rate_in_profit_currency(
+                    asset=fee_info[1],
+                    timestamp=timestamp,
+                )
+            except (PriceQueryUnsupportedAsset, RemoteError):
+                fee_price = None
+
+        # Determine whether to use `out_price` or `in_price` for calculations
+        price_to_use: Literal['in', 'out']
```
LefterisJP, line 382: nitpick: Not sure if it's worth using an enum instead of raw strings here.

LefterisJP (commented):
All right looking good. Left some comments.

Did not check all tests in too much detail due to lack of time.

Please address the small stuff I asked and remove the TMP comments.

rotkehlchen/accounting/pot.py
nebolax: Hm, I would say it will be an overkill since this variable is only for local usage in this function.

Afaik approach with such literals is widely used in our codebase

https://github.com/rotki/rotki/blob/5433c04b162749b3a19a006eabeeeea6d805c6d6/rotkehlchen/premium/sync.py#L223

https://github.com/rotki/rotki/blob/b75036a755ec8e130b51a900f7edfc67f2e0ee3a/rotkehlchen/api/rest.py#L1686

https://github.com/rotki/rotki/blob/58ca7de234284bdc434828f1f2c3e32f1db6f330/rotkehlchen/chain/ethereum/modules/compound/compound.py#L275

etc.

LefterisJP: Yeah sure but we also have a lot of enums in tthe codebase, even for small functions. And if something is widely used does not always make it correct.

The benefit of an enum is smaller memory footprint. But as you said it's only used in here so you setup an enum for a function. But anyway this is super minor.

LefterisJP (approved):
LGTM

PR #5661, opened 2023-02-24
Redecode / delete genesis tx events when needed

nebolax:


rotkehlchen/externalapis/etherscan.py
```diff
@@ -372,6 +374,19 @@ def get_transactions(
                                 transactions=[internal_tx],
                                 relevant_address=None,  # can't know the address here
                             )
+
+                        dbevents = DBHistoryEvents(self.db)  # type: ignore[arg-type]
```
yabirgb, line 378: instead of having this multiple ignore types for the db why not do a `assert self.db is not None` at the start of the method?

rotkehlchen/externalapis/etherscan.py
```diff
@@ -372,6 +374,19 @@ def get_transactions(
...
+                        with self.db.user_write() as write_cursor:  # type: ignore[union-attr]
+                            # Redecode genesis tx events
+                            dbevents.delete_events_by_tx_hash(
+                                write_cursor=write_cursor,
+                                tx_hashes=[GENESIS_HASH],
+                                chain_id=self.chain.to_chain_id(),  # type: ignore[arg-type]
```
yabirgb, line 384: nitpick: I'm always afraid of this type ignores because they assume conditions that can change in the future. Maybe you can add a check for the case where chain is in `EVM_CHAIN_IDS_WITH_TRANSACTIONS_TYPE`.

Checking it seems that etherscan is not called with `SUPPORTED_EVM_CHAINS` but with `EVM_CHAIN_IDS_WITH_TRANSACTIONS` but even after changing the type there you will have another issue because the type of `to_chain_id` is not compatible with `EVM_CHAIN_IDS_WITH_TRANSACTIONS_TYPE`. Just leave it as it is then

rotkehlchen/db/history_events.py
```diff
@@ -168,6 +170,16 @@ def delete_events_by_tx_hash(
             bindings = [(x,) for x in tx_hashes]
         write_cursor.executemany(querystr, bindings)
 
+    def delete_genesis_tx_events(
+            self,
+            write_cursor: 'DBCursor',
+            address: ChecksumEvmAddress,
+    ) -> None:
+        write_cursor.execute(
```
yabirgb, line 178: This function is called only in one place for what I'm seeing and the code executed is just a db query. Don't create a new function for this and just move the code to the place where is used

rotkehlchen/db/evmtx.py
```diff
@@ -484,6 +488,14 @@ def delete_transactions(
             tx_hashes=tx_hashes,
             chain_id=chain_id,  # type: ignore[arg-type] # comes from SUPPORTED_EVM_CHAINS
         )
+        dbevents.delete_genesis_tx_events(write_cursor, address)
+        genesis_hash_events_count = dbevents.get_history_events_count(
+            cursor=write_cursor,
+            query_filter=HistoryEventFilterQuery.make(event_identifiers=[GENESIS_HASH]),
+        )
```
yabirgb, line 495: Here you are are creating a filter and then calling a function just to do a 

```
SELECT COUNT(*) FROM history_events WHERE event_identifier=?
```

I would move it to a simple sql query since you avoid:
1. Creating the filter
2. The extra function call
3. Prepare of the filter
4. Query execution (this is the only thing you do)

The filters are good when there is a complex query involved but here I believe is not needed

rotkehlchen/db/evmtx.py
```diff
@@ -484,6 +488,14 @@ def delete_transactions(
             tx_hashes=tx_hashes,
             chain_id=chain_id,  # type: ignore[arg-type] # comes from SUPPORTED_EVM_CHAINS
         )
+        dbevents.delete_genesis_tx_events(write_cursor, address)
```
yabirgb, line 491: Can you add a comment saying that this is needed for the case where multiple addresses tracked were involved in the genesis tx?

rotkehlchen/db/evmtx.py
yabirgb: After thinking a bit you don't need to execute this query. Correct me if I'm wrong but if there are other events related to the genesis tx it means that the `.rowcount` of the previous query `dbevents.delete_genesis_tx_events(write_cursor, address)` (if you move it to a simple sql query) is > 0 because that is the case were nothing was deleted for the genesis tx in `delete_events_by_tx_hash`.

If you add a comment for this would be nice

yabirgb (commented):
Looks okey. If we had this issue it means that there is no test covering this case so you should add one that fails before this change and works fine after it.

rotkehlchen/externalapis/etherscan.py
nebolax: Can be also done that way

rotkehlchen/db/evmtx.py
nebolax: I can change it to a simple query.

`.rowcount` returns the number of rows modified, but doesn't say anything about the remaining events

yabirgb: Yes but since you have deleted all the events related to transaction touched by that address and not any other if after deleting the events of this address you still delete some events it means that there is another address having some events related to the geneisis transaction

nebolax: Sorry, I still don't get how can we judge about the remaining number of decoded events in the genesis tx by checking whether current address had any events in the genesis tx.

rotkehlchen/tests/unit/test_evm_tx_decoding.py
```diff
@@ -201,3 +205,66 @@ def test_query_and_decode_transactions_works_with_different_chains(
...
+    """
+    Checks that if an address had a genesis transacion:
+    1. The decoded event gets deleted when the address is removed
+    2. Genesis tx gets removed if it was the last tracked address with a genesis tx
+    """
+    dbevmtx = DBEvmTx(database)
+    dbevents = DBHistoryEvents(database)
+    genesis_address_1, genesis_address_2, no_genesis_address = ethereum_accounts
+
+    def get_events() -> list[HistoryBaseEntry]:
```
yabirgb, line 229: nitpick: I would call this `get_genesis_events`

rotkehlchen/tests/unit/test_evm_tx_decoding.py
```diff
@@ -201,3 +205,66 @@ def test_query_and_decode_transactions_works_with_different_chains(
...
+        with database.conn.read_ctx() as cursor:
+            events = dbevents.get_history_events(
+                cursor=cursor,
+                filter_query=HistoryEventFilterQuery.make(event_identifiers=[GENESIS_HASH]),
+                has_premium=True,
+            )
+
+        return events
+
+    def delete_address(address: ChecksumEvmAddress) -> None:
```
yabirgb, line 239: You are not deleting the address but the transactions related to this address. The name is confusing

rotkehlchen/tests/unit/test_evm_tx_decoding.py
```diff
@@ -201,3 +205,66 @@ def test_query_and_decode_transactions_works_with_different_chains(
...
+    assert get_events() == [], 'There should be no events at this point'
+
+    with database.conn.read_ctx() as cursor:
+        genesis_tx = dbevmtx.get_evm_transactions(
+            cursor=cursor,
+            filter_=EvmTransactionsFilterQuery.make(tx_hash=GENESIS_HASH),
+            has_premium=True,
+        )
+
+    assert len(genesis_tx) == 0, 'Genesis transactio should have been deleted'
```
yabirgb, line 270:
```suggestion
    assert len(genesis_tx) == 0, 'Genesis transaction should have been deleted'
```

rotkehlchen/tests/unit/test_evm_tx_decoding.py
```diff
@@ -201,3 +205,66 @@ def test_query_and_decode_transactions_works_with_different_chains(
...
+            dbevmtx.delete_transactions(
+                write_cursor=write_cursor,
+                address=address,
+                chain=SupportedBlockchain.ETHEREUM,
+            )
+
+    ethereum_transaction_decoder.decode_transaction_hashes(
+        ignore_cache=True,
+        tx_hashes=[GENESIS_HASH],
+    )
```
yabirgb, line 250: This is querying etherscan for the transactions. If is not now it needs to be mocked using vcr

yabirgb (commented):
Only minor comments about the tests

rotkehlchen/db/evmtx.py
```diff
@@ -477,13 +477,32 @@ def delete_transactions(
...
+            ).fetchone()[0] != 0
+            if genesis_tx_exists is False:
+                return
 
         dbevents.delete_events_by_tx_hash(
             write_cursor=write_cursor,
             tx_hashes=tx_hashes,
             chain_id=chain_id,  # type: ignore[arg-type] # comes from SUPPORTED_EVM_CHAINS
         )
+        write_cursor.execute(  # delete genesis tx devents related to the provided address
```
LefterisJP, line 494:
```suggestion
        write_cursor.execute(  # delete genesis tx events related to the provided address
```

rotkehlchen/externalapis/etherscan.py
```diff
@@ -366,12 +369,25 @@ def get_transactions(
...
+                        with self.db.user_write() as cursor:
                             dbtx.add_evm_internal_transactions(
                                 write_cursor=cursor,
                                 transactions=[internal_tx],
                                 relevant_address=None,  # can't know the address here
                             )
+
+                        dbevents = DBHistoryEvents(self.db)
+                        with self.db.user_write() as write_cursor:
+                            # Redecode genesis tx events
```
LefterisJP, line 381:
```suggestion
                            # Delete decoded genesis events so they can be later redecoded.
```

rotkehlchen/tests/unit/test_evm_tx_decoding.py
LefterisJP: Yes. Change the name here .
```suggestion
    def delete_transactions_for_address(address: ChecksumEvmAddress) -> None:
```

rotkehlchen/tests/unit/test_evm_tx_decoding.py
LefterisJP: yes + 1

rotkehlchen/tests/unit/test_evm_tx_decoding.py
LefterisJP: yes + 1

LefterisJP (commented):
Lots of tests failed

LefterisJP (approved):
LGTM. Only failed test is timeout

PR #5663, opened 2023-02-24
Fix wrong method for accounting in curve

yabirgb:
The wrong accounting method was being used here


rotkehlchen/chain/ethereum/modules/curve/accountant.py
```diff
@@ -77,7 +77,7 @@ def event_settings(self, pot: 'AccountingPot') -> dict[str, TxEventSettings]:  #
                 taxable=False,
                 count_entire_amount_spend=False,
                 count_cost_basis_pnl=False,
-                method='spend',
+                method='acquisition',
```
LefterisJP, line 80: That means there is no test for this. And it should be written.

LefterisJP (approved):
LGTM. Tests failed due to the cassete setup being misunderstood by all

PR #5670, opened 2023-02-26
Add statistics to liquity

yabirgb:
This PR adds a new endpoint allowing to query stats from staking in the liquity protocol. Also adds a new decoding rule for events related to the stability pool


docs/api.rst
```diff
@@ -7529,6 +7529,166 @@ Getting Liquity stability pool infomration
...
+              "total_usd_gains_stacking": "0.0",
+              "total_deposited_stability_pool": "1519146.7290263602",
+              "total_withdrawn_stability_pool": "914454.5094041774",
+              "total_deposited_stability_pool_usd_value": "0.0",
+              "total_withdrawn_stability_pool_usd_value": "0.0",
+              "staking_gains": [
+                {
+                  "asset": "ETH",
+                  "amount": "0.18236022449762773",
+                  "usd_value": "0.0"
```
nebolax, line 7612: Why is usd value 0 here?

docs/api.rst
```diff
@@ -7529,6 +7529,166 @@ Getting Liquity stability pool infomration
...
+
+   :resjson object result: A mapping with the keys ``global_stats`` and ``by_address``.
+   :resjson object global_stats: Stats aggregating the information for all the addresses tracked in the liquity module.
+   :resjson object global_stats: Breakdown by tracked address of the stats.
+   :resjson string total_usd_gains_stability_pool: Sum of all the gains valuated at the moment of the event for the stability pool.
+   :resjson string total_usd_gains_stacking: Sum of all the gains valuated at the moment of the event for liquity staking.
+   :resjson string total_deposited_stability_pool: Total amount of LUSD deposited in the stability pool.
+   :resjson string total_withdrawn_stability_pool: Total amount of LUSD withdrawn from the stability pool.
+   :resjson string total_deposited_stability_pool_usd_value: Sum of the USD value deposited in the stability pool at the time of the events.
+   :resjson string total_withdrawn_stability_pool_usd_value: Sum of the USD value withdrawn in the stability pool at the time of the events.
```
nebolax, line 7683:
```suggestion
   :resjson string total_withdrawn_stability_pool_usd_value: Sum of the USD value withdrawn from the stability pool at the time of the events.
```

rotkehlchen/api/rest.py
```diff
@@ -4499,7 +4513,7 @@ def _query_kraken_staking_events(
                 ),
                 'received': [
                     {
-                        'asset': entry[0].identifier,
```
nebolax, line 4502: Why this change?

rotkehlchen/chain/ethereum/modules/liquity/decoder.py
```diff
@@ -125,22 +125,27 @@ def _decode_stability_pool_event(
...
+                    ) or (
+                        event.asset == self.lqty and
+                        event.balance.amount == collected_lqty and
+                        tx_log.topics[0] == STABILITY_POOL_LQTY_PAID
+                    )
+                ):
+                    event.event_type = HistoryEventType.STAKING
+                    event.event_subtype = HistoryEventSubType.REWARD
+                    event.counterparty = CPT_LIQUITY
+                    event.notes = f"Collect {event.balance.amount} {event.asset.resolve_to_crypto_asset().symbol} from liquity's stability pool"  # noqa: E501
```
nebolax, line 143: I would use `event.asset.symbol_or_name()` here

rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -18,19 +21,48 @@
...
 
 MIN_COLL_RATE = '1.1'
 
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
 
+QUERY_STAKING_EVENTS = """
+WHERE event_identifier IN
+(SELECT A.event_identifier FROM history_events AS A JOIN history_events AS B ON A.event_identifier = B.event_identifier WHERE
```
nebolax, line 41: Why to join history_events to history_events?

nebolax, line 41: Won't it be possible to use a db filter here?

rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -18,19 +21,48 @@
...
+"""  # noqa: E501
+BINDINGS_STAKING_EVENTS = [
+    CPT_LIQUITY, A_LQTY.identifier, A_LUSD.identifier, HistoryEventSubType.DEPOSIT_ASSET.serialize(),  # noqa: E501
+    HistoryEventType.STAKING.serialize(), HistoryEventType.STAKING.serialize(),
+    HistoryEventSubType.REWARD.serialize(),
+]
+# stability pool rewards
+QUERY_STABILITY_POOL_EVENTS = """
+WHERE event_identifier IN (
+    SELECT A.event_identifier FROM history_events AS A JOIN history_events AS B ON
```
nebolax, line 54: Same questions here

rotkehlchen/tests/api/test_liquity.py
```diff
@@ -226,3 +226,47 @@ def mocked_request_dict(url, *_args, **_kwargs):
...
+@pytest.mark.parametrize('start_with_valid_premium', [True])
+@pytest.mark.parametrize('should_mock_current_price_queries', [True])
+@pytest.mark.parametrize('add_account_to_db', [[False]])
+@pytest.mark.parametrize('ethereum_accounts', [
+    [
+        '0xF662f831361c8Ab48d807f7753eb3d641be25d24',
+        '0xbB8311c7bAD518f0D8f907Cad26c5CcC85a06dC4',
+    ],
+])
+def test_staking_stats(rotkehlchen_api_server, ethereum_accounts):
```
nebolax, line 244: Shouldn't this test be vcr'ed ?

nebolax (commented):
Some comments / questions

docs/api.rst
yabirgb: I took the example for the database and it didn't have the prices queried yet. For the api example it shouldn't be a problem

rotkehlchen/api/rest.py
yabirgb: I saw that we didn't need to create an Asset object since we were taking directly the identifiers. This saves us from creating useless objects

rotkehlchen/chain/ethereum/modules/liquity/trove.py
yabirgb: Because there are multiple rewards collected in the same transaction but we don't have an easy way to differenciate what are rewards from the stability pool and what are rewards from staking. The only difference is that you are rewarded LQTY from the stability pool and LUSD from staking. This is why I search for this special assets and then get the events in the transactions containing them

yabirgb: I believe the filter will over complicated it since is a query that we won't reuse and it has many special conditions

rotkehlchen/tests/api/test_liquity.py
yabirgb: No external query is made. If you look at it is a prepared database that already contains the events and what it does is verify that the results are what I've verified by manually checking them. There are multiple events and adding them manually would be slow

docs/api.rst
nebolax: Alright

rotkehlchen/chain/ethereum/modules/liquity/trove.py
nebolax: Hmmm okay. Looks weird, but probably makes sense

rotkehlchen/tests/api/test_liquity.py
nebolax: Yes, makes sense :+1:

rotkehlchen/chain/ethereum/modules/liquity/decoder.py
```diff
@@ -125,22 +125,27 @@ def _decode_stability_pool_event(
...
+                    ) or (
+                        event.asset == self.lqty and
+                        event.balance.amount == collected_lqty and
+                        tx_log.topics[0] == STABILITY_POOL_LQTY_PAID
+                    )
+                ):
+                    event.event_type = HistoryEventType.STAKING
+                    event.event_subtype = HistoryEventSubType.REWARD
+                    event.counterparty = CPT_LIQUITY
+                    event.notes = f"Collect {event.balance.amount} {event.asset.symbol_or_name()} from liquity's stability pool"  # noqa: E501
```
LefterisJP, line 143: Why symbol_or_name?

rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -256,3 +297,101 @@ def liquity_staking_balances(
...
+                    'asset': entry[0],
+                    'amount': entry[1],
+                    'usd_value': entry[2],
+                } for entry in stability_rewards_breakdown
+            ],
+        }
+
+    def get_stats(self, addresses: list[ChecksumEvmAddress]) -> dict[str, Any]:
+        """
+        Query staking information for the liquity module related to both the LQTY statking
```
LefterisJP, line 359:
```suggestion
        Query staking information for the liquity module related to both the LQTY staking
```

LefterisJP (commented):
This will also require frontend changes, right? And as such changelog should stay until they are done?

rotkehlchen/chain/ethereum/modules/liquity/decoder.py
yabirgb: In this case it should always use the symbol but alexey suggested it and I saw it cleaner than do the resolve and then call symbol, is what happens in that function

rotkehlchen/db/history_events.py
```diff
@@ -332,12 +332,12 @@ def get_history_events_count(self, cursor: 'DBCursor', query_filter: HistoryEven
     def get_value_stats(
             self,
             cursor: 'DBCursor',
-            query_filter: HistoryEventFilterQuery,
-    ) -> tuple[FVal, list[tuple[Asset, FVal, FVal]]]:
+            query_filters: str,
+            bindings: list[Any],
+    ) -> tuple[FVal, list[tuple[str, FVal, FVal]]]:
```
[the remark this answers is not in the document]
yabirgb: I edited this because it is a function that we also use for the kraken stats here https://github.com/rotki/rotki/blob/40577eba978a6c988157b4637e780d89caa90a6e/rotkehlchen/api/rest.py#L3645

The reason to pass the query and binding instead of the filter is because the liquity query is complicated and creating a filter for it would just complicate things. With the refactor of the history events this filter would be much easier to do and I can add a TODO to simplify this then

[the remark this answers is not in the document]
yabirgb: I also disliked this and this is why I proposed the refactor to the history events. I prefer to make keep this change like this for now and revert it later (is really easy) once the filters can be properly used for this because it unblocks the frontend to work in this issue

LefterisJP:
I have asked for label changes when it's sent back to me. Please remember this

rotkehlchen/db/history_events.py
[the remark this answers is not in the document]
LefterisJP: All right. Fair enough.

LefterisJP (approved):
LGTM

PR #5673, opened 2023-02-26
ETC, BCH, BSV are now fully ignored in accounting if set

LefterisJP:
There was a bug where ETC, BCH and BSV prefork acquisition events (all after-fork were properly ignored) were still taken into account if any of those assets was set to ignored for accounting.

This is now fixed

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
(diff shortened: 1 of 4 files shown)


LefterisJP:
Only frontend failed, all relevant backend ones passed. But the frontend failure needs further looking into as it seems like a backend error but unrelated to this PR.

Issue #1505, opened 2020-09-22
Take Binance Lending/Savings into account for profit calculation


LefterisJP:
## Abstract

After https://github.com/rotki/rotki/pull/1503 we can now show the lending/savings balances in binance.

But users probably want a bit more detailed breakdown.

When they click on the binance balances component it should differentiate between which assets are just sitting in the account and which are in savings/lending.

We can achieve that by using this [endpoint](https://binance-docs.github.io/apidocs/spot/en/#savings-endpoints).

Also take the interest history into account in the tax profit/loss report.

## Task

- [ ] Add a new API endpoint for binance savings details (or somehow include it in the current balances one? -- need thinking -- probably new endpoint)
- [ ] In the UI, when the user clicks on the binance balances components balances should be split into what's in the normal exchange and what's in savings. For those in savings we should show how much they are earning in interest. And also how much they have earned from savings. Using [this](https://binance-docs.github.io/apidocs/spot/en/#get-interest-history-user_data-2) endpoint.
- [ ] Take the binance savings historical earnings into account in the tax profit/loss report.

Tomson124:
+1

PR #5681, opened 2023-02-27
Add Binance Savings endpoint & PnL support

prettyirrelevant:
Backend aspect of #1505

## Checklist
- [x] Add endpoint for breakdown of Binance Savings into balances & interest accrued.
- [x] Take interests history into PnL calculations.
- [x] Write tests.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -12194,3 +12194,86 @@ Add EVM Transaction By Hash
    :statuscode 409: No user is currently logged in.
    :statuscode 500: Internal rotki error.
    :statuscode 502: An external service used in the query such as etherscan could not be reached or returned unexpected response.
+
+
+Get Binance Savings Interests History
+=======================================
+
+.. http:get:: /api/(version)/(location)/savings
+
+   Doing a GET on this endpoint will return all history events relating to interest payments for the specified location.
+   .. note::
+      This endpoint can also be queried asynchronously by using ``"async_query": true``.
+
+   **Example Request**:
+
+   .. http:example:: curl wget httpie python-requests
+
+      GET /api/1/binance/savings HTTP/1.1
+      Host: localhost:5042
+      Content-Type: application/json;charset=UTF-8
+
+      {"from_timestamp": 1451606400, "to_timestamp": 1571663098, "only_cache": false}
+
+   :reqjson bool async_query: Boolean denoting whether this is an asynchronous query or not.
+   :reqjson int limit: Optional. This signifies the limit of records to return as per the `sql spec <https://www.sqlite.org/lang_select.html#limitoffset>`__.
+   :reqjson int offset: This signifies the offset from which to start the return of records per the `sql spec <https://www.sqlite.org/lang_select.html#limitoffset>`__.
+   :reqjson list[string] order_by_attributes: Optional. This is the list of attributes of the history by which to order the results. If none is given 'timestamp' is assumed. Valid values are: ['timestamp', 'location', 'amount'].
+   :reqjson list[bool] ascending: Optional. False by default. Defines the order by which results are returned depending on the chosen order by attribute.
+   :reqjson int from_timestamp: The timestamp from which to query. Can be missing in which case we query from 0.
+   :reqjson int to_timestamp: The timestamp until which to query. Can be missing in which case we query until now.
+   :reqjson bool only_cache: Optional.If this is true then the equivalent exchange/location is not queried, but only what is already in the DB is returned.
+   :reqjson string location: This represents the exchange's name. Can only be one of `"binance"` or `"binanceus"`.
+
+   **Example Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+    {
+        "result": {
+            "events": [
+                {
+                    "timestamp": 1587233562,
+                    "location": "binance",
+                    "asset": "eip155:1/erc20:0x6B175474E89094C44Da98b954EedeAC495271d0F",
+                    "amount": "0.00987654",
+                    "usd_value": "0.05432097"
+                },
+                {
+                    "timestamp": 1577233578,
+                    "location": "binance",
+                    "asset": "eip155:1/erc20:0x4Fabb145d64652a948d72533023f6E7A623C7C53",
+                    "amount": "0.00006408",
+                    "usd_value": "0.00070488"
+                }
+            ],
+            "entries_found": 2,
+            "entries_limit": 100,
+            "entries_total": 2,
+            "total_usd_value": "0.05502585",
+            "assets": ["eip155:1/erc20:0x4Fabb145d64652a948d72533023f6E7A623C7C53", "eip155:1/erc20:0x6B175474E89094C44Da98b954EedeAC495271d0F"],
+            "received": [{"asset": "eip155:1/erc20:0x4Fabb145d64652a948d72533023f6E7A623C7C53", "amount": "0.00006408", "usd_value": "0.00070488"}, {"asset": "eip155:1/erc20:0x6B175474E89094C44Da98b954EedeAC495271d0F", "amount": "0.00987654", "usd_value": "0.05432097"}]
+        },
+        "message": ""
+    }
+
+   :resjsonarr int timestamp: The timestamp at which the event occurred.
+   :resjsonarr string location: A valid location at which the event happened.
+   :resjsonarr string amount: The amount related to the event.
+   :resjsonarr string asset: Asset involved in the event.
+   :resjsonarr string message: It won't be empty if the query to external services fails for some reason.
+   :resjson int entries_found: The number of entries found for the current filter. Ignores pagination.
+   :resjson int entries_limit: The limit of entries if free version. -1 for premium.
+   :resjson int entries_total: The number of total entries ignoring all filters.
+   :resjsonarr string total_usd_value: Sum of the USD value for the assets received computed at the time of acquisition of each event.
+   :resjson list[string] assets: Assets involved in events ignoring all filters.
+   :resjson list[object] received: Assets received with the total amount received for each asset and the aggregated USD value at time of acquisition.
+
+   :statuscode 200: The balances were returned successfully.
+   :statuscode 400: Invalid location provided.
+   :statuscode 409: No user is currently logged in.
+   :statuscode 500: Internal rotki error.
+   :statuscode 502: An external service used in the query such as binance could not be reached or returned unexpected response.
--- a/rotkehlchen/accounting/structures/base.py
+++ b/rotkehlchen/accounting/structures/base.py
@@ -359,10 +359,16 @@ class StakingEvent:
 
     def serialize(self) -> dict[str, Any]:
         data = {
-            'event_type': self.event_type.serialize(),
             'asset': self.asset.identifier,
             'timestamp': self.timestamp,
             'location': str(self.location),
         }
+        # binance has only one type of event, so serializing it is not needed.
+        if not (
+            self.location == Location.BINANCE and
+            self.event_type == HistoryEventSubType.INTEREST_PAYMENT
+        ):
+            data['event_type'] = self.event_type.serialize()
+
         balance = abs(self.balance).serialize()
         return {**data, **balance}
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -3594,91 +3594,18 @@ class RestAPI():
             query_filter: HistoryEventFilterQuery,
             value_filter: HistoryEventFilterQuery,
     ) -> dict[str, Any]:
-        history_events_db = DBHistoryEvents(self.rotkehlchen.data.db)
-        table_filter = HistoryEventFilterQuery.make(
+        return self._get_exchange_staking_or_savings_history(
+            only_cache=only_cache,
             location=Location.KRAKEN,
-            event_types=[
-                HistoryEventType.STAKING,
-            ],
+            query_filter=query_filter,
+            value_filter=value_filter,
+            event_types=[HistoryEventType.STAKING],
             exclude_subtypes=[
                 HistoryEventSubType.RECEIVE_WRAPPED,
                 HistoryEventSubType.RETURN_WRAPPED,
             ],
+            event_subtypes=None,
         )
-
-        message = ''
-        entries_limit = -1
-        if self.rotkehlchen.premium is None:
-            entries_limit = FREE_HISTORY_EVENTS_LIMIT
-
-        exchanges_list = self.rotkehlchen.exchange_manager.connected_exchanges.get(
-            Location.KRAKEN,
-        )
-        if exchanges_list is None:
-            return wrap_in_fail_result(
-                message='There is no kraken account added.',
-                status_code=HTTPStatus.CONFLICT,
-            )
-
-        # After 3865 we have a recurring task that queries for missing prices but
-        # we make sure that the returned values have their correct value calculated
-        task_manager = self.rotkehlchen.task_manager
-        if task_manager is not None:
-            try:
-                entries = task_manager.get_base_entries_missing_prices(query_filter)
-                task_manager.query_missing_prices_of_base_entries(
-                    entries_missing_prices=entries,
-                )
-            except sqlcipher.OperationalError as e:  # pylint: disable=no-member
-                return wrap_in_fail_result(
-                    message=f'Database query error retrieving misssing prices {str(e)}',
-                    status_code=HTTPStatus.CONFLICT,
-                )
-
-        # Query events from database
-        with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
-            events_raw, entries_found = self.rotkehlchen.events_historian.query_history_events(
-                cursor=cursor,
-                filter_query=query_filter,
-                only_cache=only_cache,
-            )
-            events = []
-            for event in events_raw:
-                try:
-                    staking_event = StakingEvent.from_history_base_entry(event)
-                except DeserializationError as e:
-                    log.warning(f'Could not deserialize staking event: {event} due to {str(e)}')
-                    continue
-                events.append(staking_event)
-
-            entries_total = history_events_db.get_history_events_count(cursor=cursor, query_filter=table_filter)  # noqa: E501
-            value_query_filters, value_bindings = value_filter.prepare(with_pagination=False, with_order=False)  # noqa: E501
-            usd_value, amounts = history_events_db.get_value_stats(
-                cursor=cursor,
-                query_filters=value_query_filters,
-                bindings=value_bindings,
-            )
-
-            result = {
-                'events': events,
-                'entries_found': entries_found,
-                'entries_limit': entries_limit,
-                'entries_total': entries_total,
-                'total_usd_value': usd_value,
-                'assets': history_events_db.get_entries_assets_history_events(
-                    cursor=cursor,
-                    query_filter=table_filter,
-                ),
-                'received': [
-                    {
-                        'asset': entry[0],
-                        'amount': entry[1],
-                        'usd_value': entry[2],
-                    } for entry in amounts
-                ],
-            }
-
-        return {'result': result, 'message': message, 'status_code': HTTPStatus.OK}
 
     def get_user_added_assets(self, path: Optional[Path]) -> Response:
         """
@@ -4155,3 +4082,114 @@ class RestAPI():
             )
         evm_manager.transactions_decoder.decode_transaction(transaction=transaction, tx_receipt=tx_receipt)  # noqa: E501
         return OK_RESULT
+
+    def _get_exchange_staking_or_savings_history(
+            self,
+            only_cache: bool,
+            location: Literal[Location.KRAKEN, Location.BINANCE, Location.BINANCEUS],
+            event_types: list[HistoryEventType],
+            query_filter: HistoryEventFilterQuery,
+            value_filter: HistoryEventFilterQuery,
+            event_subtypes: Optional[list[HistoryEventSubType]] = None,
+            exclude_subtypes: Optional[list[HistoryEventSubType]] = None,
+    ) -> dict[str, Any]:
+        """Query exchanges for either staking or savings history.
+
+        If `only_cache` is False, only data stored in the database is returned. Otherwise,
+        the latest data is fetched from the exchange.
+        """
+        history_events_db = DBHistoryEvents(self.rotkehlchen.data.db)
+        table_filter = HistoryEventFilterQuery.make(
+            location=location,
+            event_types=event_types,
+            event_subtypes=event_subtypes,
+            exclude_subtypes=exclude_subtypes,
+        )
+
+        message = ''
+        entries_limit = -1
+        if self.rotkehlchen.premium is None:
+            entries_limit = FREE_HISTORY_EVENTS_LIMIT
+
+        exchanges_list = self.rotkehlchen.exchange_manager.connected_exchanges.get(
+            location,
+        )
+        if exchanges_list is None:
+            return wrap_in_fail_result(
+                message=f'There is no {location.name} account added.',
+                status_code=HTTPStatus.CONFLICT,
+            )
+
+        # query events from db and remote data(if `only_cache` is false).
+        with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
+            try:
+                events_raw, entries_found = self.rotkehlchen.events_historian.query_history_events(  # noqa: E501
+                    cursor=cursor,
+                    location=location,
+                    filter_query=query_filter,
+                    only_cache=only_cache,
+                    task_manager=self.rotkehlchen.task_manager,
+                )
+            except sqlcipher.OperationalError as e:  # pylint: disable=no-member
+                return wrap_in_fail_result(
+                    message=f'Database query error retrieving missing prices {str(e)}',
+                    status_code=HTTPStatus.CONFLICT,
+                )
+
+            events = []
+            for event in events_raw:
+                try:
+                    staking_event = StakingEvent.from_history_base_entry(event)
+                except DeserializationError as e:
+                    log.warning(f'Could not deserialize staking event: {event} due to {str(e)}')  # noqa: E501
+                    continue
+                events.append(staking_event)
+
+            entries_total = history_events_db.get_history_events_count(
+                cursor=cursor,
+                query_filter=table_filter,
+            )
+            value_query_filters, value_bindings = value_filter.prepare(with_pagination=False, with_order=False)  # noqa: E501
+            usd_value, amounts = history_events_db.get_value_stats(
+                cursor=cursor,
+                query_filters=value_query_filters,
+                bindings=value_bindings,
+            )
+            result = {
+                'events': events,
+                'entries_found': entries_found,
+                'entries_limit': entries_limit,
+                'entries_total': entries_total,
+                'total_usd_value': usd_value,
+                'assets': history_events_db.get_entries_assets_history_events(
+                    cursor=cursor,
+                    query_filter=table_filter,
+                ),
+                'received': [
+                    {
+                        'asset': entry[0],
+                        'amount': entry[1],
+                        'usd_value': entry[2],
+                    } for entry in amounts
+                ],
+            }
+
+        return {'result': result, 'message': message, 'status_code': HTTPStatus.OK}
+
+    @async_api_call()
+    def get_binance_savings_history(
+            self,
+            only_cache: bool,
+            location: Literal[Location.BINANCE, Location.BINANCEUS],
+            query_filter: HistoryEventFilterQuery,
+            value_filter: HistoryEventFilterQuery,
+    ) -> dict[str, Any]:
+        """Returns a summary of Binance savings events that match the filter provided."""
+        return self._get_exchange_staking_or_savings_history(
+            only_cache=only_cache,
+            location=location,
+            event_types=[HistoryEventType.RECEIVE],
+            event_subtypes=[HistoryEventSubType.INTEREST_PAYMENT],
+            query_filter=query_filter,
+            value_filter=value_filter,
+        )
--- a/rotkehlchen/api/v1/resources.py
+++ b/rotkehlchen/api/v1/resources.py
@@ -45,6 +45,7 @@ from rotkehlchen.api.v1.schemas import (
     BaseXpubSchema,
     BinanceMarketsSchema,
     BinanceMarketsUserSchema,
+    BinanceSavingsSchema,
     BlockchainAccountsDeleteSchema,
     BlockchainAccountsGetSchema,
     BlockchainAccountsPatchSchema,
@@ -2419,6 +2420,28 @@ class BinanceUserMarkets(BaseMethodView):
     @use_kwargs(get_schema, location='json_and_query_and_view_args')
     def get(self, name: str, location: Location) -> Response:
         return self.rest_api.get_user_binance_pairs(name=name, location=location)
+
+
+class BinanceSavingsResource(BaseMethodView):
+
+    post_schema = BinanceSavingsSchema()
+
+    @use_kwargs(post_schema, location='json_and_query_and_view_args')
+    def post(
+            self,
+            async_query: bool,
+            only_cache: bool,
+            location: Literal[Location.BINANCE, Location.BINANCEUS],
+            query_filter: 'HistoryEventFilterQuery',
+            value_filter: 'HistoryEventFilterQuery',
+    ) -> Response:
+        return self.rest_api.get_binance_savings_history(
+            async_query=async_query,
+            location=location,
+            only_cache=only_cache,
+            query_filter=query_filter,
+            value_filter=value_filter,
+        )
 
 
 class AvalancheTransactionsResource(BaseMethodView):
--- a/rotkehlchen/api/v1/schemas.py
+++ b/rotkehlchen/api/v1/schemas.py
@@ -491,7 +491,7 @@ class TradesQuerySchema(
         }
 
 
-class StakingQuerySchema(
+class BaseStakingQuerySchema(
     AsyncQueryArgumentSchema,
     OnlyCacheQuerySchema,
     DBPaginationSchema,
@@ -500,6 +500,66 @@ class StakingQuerySchema(
     from_timestamp = TimestampField(load_default=Timestamp(0))
     to_timestamp = TimestampField(load_default=ts_now)
     asset = AssetField(expected_type=AssetWithOracles, load_default=None)
+
+    def _get_assets_list(
+            self,
+            data: dict[str, Any],
+    ) -> Optional[tuple['AssetWithOracles', ...]]:
+        return (data['asset'],) if data['asset'] is not None else None
+
+    def _make_query(
+            self,
+            location: Location,
+            data: dict[str, Any],
+            event_types: list[HistoryEventType],
+            value_event_subtypes: list[HistoryEventSubType],
+            query_event_subtypes: Optional[list[HistoryEventSubType]] = None,
+            exclude_event_subtypes: Optional[list[HistoryEventSubType]] = None,
+    ) -> dict[str, Any]:
+        if data['order_by_attributes'] is not None:
+            attributes = []
+            for order_by_attribute in data['order_by_attributes']:
+                if order_by_attribute == 'event_type':
+                    attributes.append('subtype')
+                else:
+                    attributes.append(order_by_attribute)
+            data['order_by_attributes'] = attributes
+
+        asset_list = self._get_assets_list(data)
+        query_filter = HistoryEventFilterQuery.make(
+            order_by_rules=create_order_by_rules_list(data),
+            limit=data['limit'],
+            offset=data['offset'],
+            from_ts=data['from_timestamp'],
+            to_ts=data['to_timestamp'],
+            location=location,
+            event_types=event_types,
+            event_subtypes=query_event_subtypes,
+            exclude_subtypes=exclude_event_subtypes,
+            assets=asset_list,
+        )
+
+        value_filter = HistoryEventFilterQuery.make(
+            limit=data['limit'],
+            offset=data['offset'],
+            from_ts=data['from_timestamp'],
+            to_ts=data['to_timestamp'],
+            location=location,
+            event_types=event_types,
+            event_subtypes=value_event_subtypes,
+            order_by_rules=None,
+            assets=asset_list,
+        )
+
+        return {
+            'async_query': data['async_query'],
+            'only_cache': data['only_cache'],
+            'query_filter': query_filter,
+            'value_filter': value_filter,
+        }
+
+
+class StakingQuerySchema(BaseStakingQuerySchema):
     event_subtypes = fields.List(
         SerializableEnumField(enum_class=HistoryEventSubType),
         load_default=None,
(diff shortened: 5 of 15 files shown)


rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +591,72 @@ def _query_lending_balances(
...
+        interests: defaultdict[AssetWithOracles, Balance] = defaultdict(Balance)
+        for lending_type in ('DAILY', 'ACTIVITY', 'CUSTOMIZED_FIXED'):
+            data = self._api_query_list_within_time_delta(
+                api_type='sapi',
+                start_ts=Timestamp(0),
+                time_delta=BINANCE_LENDING_INTEREST_HISTORY_INTERVAL_TS,
+                end_ts=ts_now(),
+                method='lending/union/interestHistory',
+                additional_options={'lendingType': lending_type, 'size': 100},
+            )
```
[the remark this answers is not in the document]
LefterisJP: Is there no "current balances" from interest, or gained so far? If not why not just save it as history events in the DB and query only the newest ones? You would need to process them for PnL anyway no?

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +591,72 @@ def _query_lending_balances(
...
+            for entry in data:
+                try:
+                    interest_paid = deserialize_asset_amount_force_positive(entry['interest'])
+                    if interest_paid == ZERO:
+                        continue
+
+                    asset = asset_from_binance(entry['asset'])
+                except UnsupportedAsset as e:
```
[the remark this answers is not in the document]
prettyirrelevant: `deserialise_asset_amount_force_positive` raises DeserializationError

rotkehlchen/exchanges/binance.py
```diff
@@ -1292,6 +1361,46 @@ def query_online_deposits_withdrawals(
...
+        except UnknownAsset as e:
+            self.msg_aggregator.add_warning(
+                f'Found binance lending interest history with unknown asset '
+                f'{e.identifier}. Ignoring it.',
+            )
+        except UnsupportedAsset as e:
+            self.msg_aggregator.add_warning(
+                f'Found binance lending interest history with unsupported asset '
+                f'{e.identifier}. Ignoring it.',
+            )
```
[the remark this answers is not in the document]
prettyirrelevant: I don't think I agree with this, if asset_from_binance raises `DeserializationError`, `deserialize_timestamp_from_intms` raises the same error, same for `deserialize_asset_amount_force_positive`

Moving the try catch will only lead to duplication of code for the same thing. If you're seeing something different, do point me out to it?

rotkehlchen/exchanges/binance.py
```diff
@@ -61,6 +62,7 @@
 # https://www.binance.com/en/support/articles/115000599831-Binance-Exchange-Launched-Date-Set
 BINANCE_LAUNCH_TS = Timestamp(1500001200)
 API_TIME_INTERVAL_CONSTRAINT_TS = Timestamp(7776000)  # 90 days
+BINANCE_LENDING_INTEREST_HISTORY_INTERVAL_TS = Timestamp(2592000)  # 30 days
```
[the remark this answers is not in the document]
prettyirrelevant: Point me out if I'm wrong but interval is a timestamp given that it's a duration -- timedelta in python terms.
Or did you mean something else?

rotkehlchen/exchanges/binance.py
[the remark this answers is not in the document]
yabirgb: I mean that `deserialize_asset_amount_force_positive` doesn't raise UnsupportedAsset or UnknownAsset so it can be in a different try/except statement

rotkehlchen/exchanges/binance.py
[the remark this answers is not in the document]
yabirgb: Timestamp is a type that we have defined and semantically is a moment of time in seconds. Since is a type over int it won't have implications code wise but for me at least is strange to see a difference in seconds as a Timestamp since it doesn't mean a moment in time

rotkehlchen/exchanges/binance.py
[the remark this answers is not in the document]
yabirgb: Yes, those can stay together. What I mean is `asset_from_binance` doesn't raise any deserialization error. Is good to have the try statements as simple as possible

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,140 @@ def _query_lending_balances(
...
+            )
+
+            filter_query = HistoryEventFilterQuery.make(
+                from_ts=Timestamp(0),
+                to_ts=now,
+                event_types=[HistoryEventType.RECEIVE],
+                event_subtypes=[HistoryEventSubType.REWARD],
+                location=Location.BINANCE,
+                location_label=self.name,
+            )
```
yabirgb, line 729: It doesn't matter much but since this doesn't use the filter it can be created outside the with statement

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,140 @@ def _query_lending_balances(
...
+        now = ts_now()
+        with self.db.conn.read_ctx() as cursor:
+            self._query_lending_interests_history(
+                cursor=cursor,
+                start_ts=BINANCE_LAUNCH_TS,
+                end_ts=now,
+            )
+
+            filter_query = HistoryEventFilterQuery.make(
+                from_ts=Timestamp(0),
```
yabirgb, line 723: Since you have it, this can be rewritten just to make it clearer to


```suggestion
                from_ts=BINANCE_LAUNCH_TS,
```

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,140 @@ def _query_lending_balances(
...
+                            location=self.location,
+                            location_label=self.name,
+                            asset=asset_from_binance(entry['asset']),
+                            balance=Balance(
+                                amount=interest_received,
+                                usd_value=ZERO,
+                            ),
+                            notes=f'Interest paid from {entry["lendingType"]} {entry["productName"]} savings',  # noqa: E501
+                            event_type=HistoryEventType.RECEIVE,
+                            event_subtype=HistoryEventSubType.REWARD,
```
yabirgb, line 665: What do you think of adding a new type of subtype `INTEREST PAYMENT`?

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,140 @@ def _query_lending_balances(
...
+        for query_start_ts, query_end_ts in ranges_to_query:
+            events = []
+            for lending_type in ('DAILY', 'ACTIVITY', 'CUSTOMIZED_FIXED'):
+                response = self._api_query_list_within_time_delta(
+                    api_type='sapi',
+                    start_ts=query_start_ts,
+                    time_delta=BINANCE_LENDING_INTEREST_HISTORY_INTERVAL_TS,
+                    end_ts=query_end_ts,
+                    method='lending/union/interestHistory',
+                    additional_options={'lendingType': lending_type, 'size': 100},
```
yabirgb, line 638: how was this size chosen?

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,140 @@ def _query_lending_balances(
...
+        range_query_name = f'{self.location}_history_events_{self.name}'
+        ranges_to_query = ranges.get_location_query_ranges(
+            cursor=cursor,
+            location_string=range_query_name,
+            start_ts=start_ts,
+            end_ts=end_ts,
+        )
+        for query_start_ts, query_end_ts in ranges_to_query:
+            events = []
+            for lending_type in ('DAILY', 'ACTIVITY', 'CUSTOMIZED_FIXED'):
```
yabirgb, line 631: what is the difference between this 3 types?

rotkehlchen/tests/api/test_exchanges.py
```diff
@@ -1245,3 +1245,99 @@ def test_binance_query_pairs(rotkehlchen_api_server_with_exchanges):
...
+
+    interests_history_call_count = 0
+    lending_balances_call_count = 0
+
+    def mock_api_query(api_type, method, options):  # pylint: disable=unused-argument
+        nonlocal interests_history_call_count
+        nonlocal lending_balances_call_count
+
+        if 'lending/union/account' in method:
+            if lending_balances_call_count >= 1:
```
yabirgb, line 1273: have you checked that the 3 types of lending types are correctly tracked?

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,140 @@ def _query_lending_balances(
...
+                    additional_options={'lendingType': lending_type, 'size': 100},
+                )
+                for idx, entry in enumerate(response):
+                    try:
+                        interest_received = deserialize_asset_amount_force_positive(entry['interest'])  # noqa: E501
+                        if interest_received == ZERO:
+                            continue
+
+                        event = HistoryBaseEntry(
+                            event_identifier=HistoryBaseEntry.deserialize_event_identifier(str(uuid4())),  # noqa: E501
```
yabirgb, line 647: I would move the event creation out of the try where you have all the try/except just because it makes clearer and easier to read what is happening and what failed. If any of this deserialization fails it will point to this first line instead of the line that failed I believe

yabirgb (commented):
A few comments. Also this is still missing what we commented about including them in the balances too right? Not sure if the interest earned are returned or not in the balances but the deposit amounts shouldn't right?

rotkehlchen/exchanges/binance.py
prettyirrelevant: It represents the various lending types supported.

rotkehlchen/tests/api/test_exchanges.py
prettyirrelevant: yeah

rotkehlchen/exchanges/binance.py
prettyirrelevant: I basically chose the max size i.e. 100

Plus since this is used for pagination as a result of `current` query parameter, it'll reduce number of calls

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+            response = self._api_query_list_within_time_delta(
+                api_type=api_type,
+                start_ts=start_ts,
+                end_ts=end_ts,
+                time_delta=time_delta,
+                method=method,
+                additional_options=additional_options,
+            )
+            # we get an empty response, and `current` is at least 1
+            if len(response) == 0 and current_page >= 1:
```
yabirgb, line 635: Isn't it always true that `current_page >= 1`? I believe you can remove this condition

```suggestion
            if len(response) == 0:
```

either you wanted to set current page to 0 or this condition is wrong. Also what happens when you don't have any events? does this for loop stop (I mean if you change the condition)?

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+            self,
+            start_ts: Timestamp,
+            end_ts: Timestamp,
+            time_delta: int,
+            api_type: BINANCE_API_TYPE,
+            method: Literal['lending/union/interestHistory'],
+            additional_options: Optional[dict] = None,
+    ) -> list[dict[str, Any]]:
+        """Request via `_api_query_list_within_time_delta` using `current` param for pagination"""
+        data = []
```
yabirgb, line 620: This can raise:

```
         - RemoteError
         - BinancePermissionError
```

make sure to state it in the docs of the functions that use this private method or here too and catch it where needed

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+
+        "DAILY" -> flexible savings.
+        "CUSTOMIZED_FIXED" -> fixed savings.
+        "ACTIVITY" -> fixed savings but as a result of token partners.
+
+        Lending Interest History Documentation:
+        https://binance-docs.github.io/apidocs/spot/en/#get-interest-history-user_data-2
+
+        May raise:
+        - RemoteError
```
yabirgb, line 660: This can also raise `BinancePermissionError`

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+
+                        event_identifier = HistoryBaseEntry.deserialize_event_identifier(str(uuid4()))  # noqa: E501
+                        timestamp = TimestampMS(
+                            deserialize_fval(
+                                value=entry['time'],
+                                name='time',
+                                location='binance lending interest processing',
+                            ).to_int(exact=False),
+                        )
+                        asset = asset_from_binance(entry['asset'])
```
yabirgb, line 695: This is the only line in this try except that can raise `UnsupportedAsset, UnknownAsset`, please move it to a different try/except

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+                        sequence_index=idx,
+                        timestamp=timestamp,
+                        location=self.location,
+                        location_label=self.name,
+                        asset=asset,
+                        balance=Balance(
+                            amount=interest_received,
+                            usd_value=ZERO,
+                        ),
+                        notes=f'Interest paid from {entry["lendingType"]} {entry["productName"]} savings',  # noqa: E501
```
yabirgb, line 727: This can raise keyerror

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+                    events.append(event)
+
+            if len(events) != 0:
+                with self.db.user_write() as write_cursor:
+                    try:
+                        self.history_events_db.add_history_events(write_cursor, events)
+                    except InputError as e:
+                        self.msg_aggregator.add_error(
+                            f'Failed to save Binance lending interest history {query_start_ts} '
+                            f'to {query_end_ts} in database. {str(e)}',
```
yabirgb, line 740:
```suggestion
                            f'Failed to save Binance lending interest history from timestamp {query_start_ts} '
                            f'to {query_end_ts} in the database. {str(e)}',
```

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+        - RemoteError
+        """
+        now = ts_now()
+        balances: defaultdict[AssetWithOracles, Balance] = defaultdict(Balance)
+        interests: defaultdict[AssetWithOracles, Balance] = defaultdict(Balance)
+        filter_query = HistoryEventFilterQuery.make(
+            from_ts=BINANCE_LAUNCH_TS,
+            to_ts=now,
+            event_types=[HistoryEventType.RECEIVE],
+            event_subtypes=[HistoryEventSubType.REWARD],
```
yabirgb, line 764: I would create a different subtype like interest payment because you are creating a very special subset and if for example we track in the future airdrops from binance then you will have an issue querying them. What do you think?

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+                    except DeserializationError as e:
+                        self.msg_aggregator.add_error(
+                            f'Error at deserializing {self.name} asset. {str(e)}. '
+                            f'Ignoring its lending interest history query.',
+                        )
+                        continue
+
+                    event = HistoryBaseEntry(
+                        event_identifier=event_identifier,
+                        sequence_index=idx,
```
yabirgb, line 718: Since the event identifier is different for each one you can use 0 as sequence index for all. Sequence index is used when multiple historyBaseEntry share the same event identifier

docs/api.rst
```diff
@@ -12032,3 +12032,55 @@ Add EVM Transaction By Hash
...
+
+   .. http:example:: curl wget httpie python-requests
+
+      GET /api/1/binance/savings HTTP/1.1
+      Host: localhost:5042
+      Content-Type: application/json;charset=UTF-8
+
+      {"async_query": true}
+
+   :reqjson bool async_query: Boolean denoting whether this is an asynchronous query or not.
```
yabirgb, line 12059: Include the location argument and the description. The note above should be an entry here instead

docs/api.rst
```diff
@@ -12032,3 +12032,55 @@ Add EVM Transaction By Hash
...
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+          "result": {
+            "savings_balance": {
+                "BTC": {"amount": "0.004", "usd_value": "30303.303"}
+            },
+            "interest_gained": {
+                "BTC": {"amount": "0.000000004", "usd_value": "3.303"}
```
yabirgb, line 12074: usd_value is the value at time of acquisition right? Add this to the response description and state it

docs/api.rst
```diff
@@ -12032,3 +12032,55 @@ Add EVM Transaction By Hash
...
+            }
+          },
+          "message": ""
+      }
+
+   :resjson object result: Returns an object where ``"savings_balance"`` represents the total amount in the savings and ``"interest_gained"`` represents the interest accrued for token(s) deposited.
+
+   :statuscode 200: The balance were returned successfully.
```
yabirgb, line 12082:
```suggestion
   :statuscode 200: The balance was returned successfully.
```

rotkehlchen/api/rest.py
```diff
@@ -4140,3 +4140,46 @@ def add_evm_transaction_by_hash(
...
+        exchanges_list = self.rotkehlchen.exchange_manager.connected_exchanges.get(location)
+        if exchanges_list is None:
+            return wrap_in_fail_result(
+                message=f'There is no {location.name} account connected.',
+                status_code=HTTPStatus.CONFLICT,
+            )
+
+        balances: dict[str, defaultdict[AssetWithOracles, Balance]] = {
+            'savings_balance': defaultdict(Balance),
+            'interest_gained': defaultdict(Balance),
```
yabirgb, line 4162: why does this need to be a defaultdict? Can't it be

```suggestion
            'savings_balance': {},
            'interest_gained': {},
```

?

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+        "ACTIVITY" -> fixed savings but as a result of token partners.
+
+        Lending Interest History Documentation:
+        https://binance-docs.github.io/apidocs/spot/en/#get-interest-history-user_data-2
+
+        May raise:
+        - RemoteError
+        """
+        ranges = DBQueryRanges(self.db)
+        range_query_name = f'{self.location}_history_events_{self.name}'
```
yabirgb, line 663: I would use a different name here since history_events is a wide tag.

```suggestion
        range_query_name = f'{self.location}_lending_history_{self.name}'
```

Also you need to make sure that when deleting the information from binance and binance us this query range also gets deleted. It happens in the dbhandler file. I believe it does because it searches by self.location but can you double check it?

rotkehlchen/exchanges/binance.py
yabirgb: Can you state that the timedelta is because

```
The time between startTime and endTime cannot be longer than 30 days.
```

https://binance-docs.github.io/apidocs/spot/en/#get-interest-history-user_data-2

Is the kind of thing that you look later and forget why is like this.

And okey, the 100 is because is stated in the docs that is the maxium

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+                        if interest_received == ZERO:
+                            continue
+
+                        event_identifier = HistoryBaseEntry.deserialize_event_identifier(str(uuid4()))  # noqa: E501
+                        timestamp = TimestampMS(
+                            deserialize_fval(
+                                value=entry['time'],
+                                name='time',
+                                location='binance lending interest processing',
+                            ).to_int(exact=False),
```
yabirgb, line 693: Can you use `deserialize_timestamp`. If you check the code it does the same thing

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+                        self.msg_aggregator.add_error(
+                            f'Failed to save Binance lending interest history {query_start_ts} '
+                            f'to {query_end_ts} in database. {str(e)}',
+                        )
+
+            with self.db.user_write() as write_cursor:
+                ranges.update_used_query_range(
+                    write_cursor=write_cursor,
+                    location_string=range_query_name,
+                    queried_ranges=[(start_ts, end_ts)] + ranges_to_query,
```
yabirgb, line 747: Question about this. I've seen that we to the same in other places but isn't this equivalent to 

```
queried_ranges=[(start_ts, end_ts)] 
```
? What is the difference?

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+            self._query_lending_interests_history(
+                cursor=cursor,
+                start_ts=BINANCE_LAUNCH_TS,
+                end_ts=now,
+            )
+            events = self.history_events_db.get_history_events(
+                cursor=cursor,
+                filter_query=filter_query,
+                has_premium=True,
+            )
```
yabirgb, line 778: since you only need the asset and the balance why not just do a `sum(cast(amout AS REAL)), sum(cast(usd_value AS REAL)) .. group by asset` and the filter you just built? I believe is the only information you are using right?

rotkehlchen/tests/api/test_exchanges.py
yabirgb: can you also assert that all the 3 valid lending types were queried? Even you could fake a different asset per response to validate it and would be better

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+    def _api_query_list_paginated_within_time_delta(
+            self,
+            start_ts: Timestamp,
+            end_ts: Timestamp,
+            time_delta: int,
+            api_type: BINANCE_API_TYPE,
+            method: Literal['lending/union/interestHistory'],
+            additional_options: Optional[dict] = None,
+    ) -> list[dict[str, Any]]:
+        """Request via `_api_query_list_within_time_delta` using `current` param for pagination"""
```
yabirgb, line 619: This docstring is not very useful since I had to look at the code to understand what it was doing. Can you maybe edit it to

```suggestion
        """Query binance lending endpoint for a time range iterating in periods of `time_delta` and using pagination"""
```

or any other info that adds value

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +607,184 @@ def _query_lending_balances(
...
+            )
+            events = self.history_events_db.get_history_events(
+                cursor=cursor,
+                filter_query=filter_query,
+                has_premium=True,
+            )
+
+        for event in events:
+            # asset resolution exceptions cannot be raised here since the asset(s)
+            # was added using `asset_from_binance`.
```
yabirgb, line 782: It can happen that a user deletes the asset from the globaldb after saving it

yabirgb (commented):
Thanks Isaac. Still some things need improvement

rotkehlchen/exchanges/binance.py
prettyirrelevant: I think I shouldn't have included the `ranges_to_query` since I've not queried it.

rotkehlchen/exchanges/binance.py
prettyirrelevant: Yeah, you're right. The `and` case isn't necessary here

rotkehlchen/serialization/deserialize.py
```diff
@@ -205,6 +206,20 @@ def deserialize_timestamp_from_intms(time: int) -> Timestamp:
     return Timestamp(int(time / 1000))
 
 
+def deserialize_timestamp_ms(time: int) -> TimestampMS:
+    """Deserializes a timestamp in milliseconds from an integer.
+
+    May raise:
+    - DeserializationError
```
yabirgb, line 213: This function is doing nothing. The linter would warn you that you are passing something that is not int to TimestampMS. If anything I believe the tipe of time should be `Any` and then the function might make sense
but then I would check if a conversion is possible. Basically the same as `deserialize_timestamp`.

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +603,223 @@ def _query_lending_balances(
...
+                    additional_options={'lendingType': lending_type, 'size': 100},
+                )
+                for entry in response:
+                    try:
+                        interest_received = deserialize_asset_amount_force_positive(entry['interest'])  # noqa: E501
+                        if interest_received == ZERO:
+                            continue
+
+                        event_identifier = HistoryBaseEntry.deserialize_event_identifier(str(uuid4()))  # noqa: E501
+                        timestamp = deserialize_timestamp_ms(entry['time'])
```
yabirgb, line 694: But this change is not doing the same as before. Before you had a conversion using exact true and now you are just taking the same value. Is this what you wanted?

yabirgb (commented):
Still some questions about the changes you made

rotkehlchen/exchanges/binance.py
prettyirrelevant: Yeah, it was a silly copy pasta.
Binance returns a timestamp in milliseconds.

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +603,223 @@ def _query_lending_balances(
...
+                            to_asset=A_USD,
+                            timestamp=Timestamp(int(timestamp / 1000)),
+                        )
+                        usd_value = usd_price * interest_received
+                    except NoPriceForGivenTimestamp as e:
+                        self.msg_aggregator.add_error(
+                            f'Could not find USD price of {asset} at {timestamp}. {str(e)}'
+                            f'Ignoring its lending interest history.',
+                        )
+                        continue
```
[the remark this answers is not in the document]
prettyirrelevant: Okay got it.

yabirgb (commented):
I believe you should change the approach here

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +610,241 @@ def _query_lending_balances(
 
         return balances
 
+    def _api_query_list_paginated_within_time_delta(
+            self,
+            start_ts: Timestamp,
+            end_ts: Timestamp,
+            time_delta: int,
```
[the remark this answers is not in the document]
prettyirrelevant: okay, done.

rotkehlchen/tests/api/test_exchanges.py
```diff
@@ -1245,3 +1245,131 @@ def test_binance_query_pairs(rotkehlchen_api_server_with_exchanges):
...
+                        'asset': 'USDT',
+                        'interest': '0.00687654',
+                        'lendingType': 'CUSTOMIZED_FIXED',
+                        'productName': 'USDT',
+                        'time': 1577233562000,
+                    },
+                ]
+            elif options['lendingType'] == 'ACTIVITY':
+                if interest_activity_call_count >= 1:
+                    return []
```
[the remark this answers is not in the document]
prettyirrelevant: done that.

rotkehlchen/exchanges/binance.py
```diff
@@ -1205,21 +1407,36 @@ def _api_query_list_within_time_delta(
...
+                result = self.api_query_list(api_type, method, options=options)
+                results.extend(result)
+
             # Case stop requesting
             if to_ts >= end_ts:
                 break
 
             from_ts = to_ts + 1
             to_ts = min(to_ts + offset, end_ts)
+            current_page += 1
```
yabirgb, line 1439: As we discussed this is wrong

rotkehlchen/tests/api/test_exchanges.py
```diff
@@ -1245,3 +1245,142 @@ def test_binance_query_pairs(rotkehlchen_api_server_with_exchanges):
...
+    assert_error_response(response, 'is not one of binance,binanceus')
+
+    lending_balances_call_count = 0
+    interest_daily_call_count = 0
+    interest_customized_fixed_call_count = 0
+    interest_activity_call_count = 0
+
+    def mock_api_query(api_type, method, options):  # pylint: disable=unused-argument
+        nonlocal interest_daily_call_count, \
+            interest_customized_fixed_call_count, \
```
yabirgb, line 1273: since this is big maybe you can move to an auxiliar function in a utils file? It makes harder to read the test

yabirgb (commented):
The logic is wrong and needs fixes

rotkehlchen/exchanges/binance.py
```diff
@@ -1205,15 +1407,29 @@ def _api_query_list_within_time_delta(
...
             }
             if additional_options:
                 options.update(additional_options)
-            result = self.api_query_list(api_type, method, options=options)
-            results.extend(result)
+                if method == 'lending/union/interestHistory':
+                    options['current'] = current_page
+
+            if method == 'lending/union/interestHistory':
+                while True:
```
LefterisJP, line 1422: This is dangerous. We should not have an unprotected while True in the code. There should be some kidn of limit and auto-breaking.

docs/api.rst
```diff
@@ -12032,3 +12032,55 @@ Add EVM Transaction By Hash
    :statuscode 409: No user is currently logged in.
    :statuscode 500: Internal rotki error.
    :statuscode 502: An external service used in the query such as etherscan could not be reached or returned unexpected response.
+
+
+Get Binance Savings Balance Details
+=======================================
+
+.. http:get:: /api/(version)/(location)/savings
```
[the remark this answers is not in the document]
LefterisJP: Lending balances **are already** part of the normal binance balances. Isn't this the same thing? I guess this endpoint is more about the history afaics. So it should be either renamed or just the title of this section renamed.

LefterisJP (commented):
###General comments

#### Generalize

Can't we generalize? Can't we count this as a "staking" page for Binance? Like we have for Kraken? In the end it's the same thing. A bunch of events that give you rewards over time.

Then we can use this also in the frontend in the exact same way we use kraken staking events.

#### Cache/no cache

As mentioned inside the PR review I think you should have the same treatment of first querying DB cache, then refreshing from latest binance if needed and then showing all.

To do that we need to give the `cache_only` option so the frontend can do two queries as with other views. First one only hitting the DB cache and showing results immediately and then a refresh one.

###General questions

1. Are these "savings" supporrted in both binance and binance US? I would expect the US one to be a lot more strict with these stuff.

2. How is this counted in PnL? Just due to the existence of the events? Is the historical price of the amount always taken properly?

rotkehlchen/exchanges/binance.py
prettyirrelevant: There's auto-breaking two lines below where it checks if the result is an empty list or not

rotkehlchen/exchanges/binance.py
```diff
@@ -202,6 +217,7 @@ def __init__(
         self.msg_aggregator = msg_aggregator
         self.offset_ms = 0
         self.selected_pairs = binance_selected_trade_pairs
+        self.history_events_db = DBHistoryEvents(self.db)
```
LefterisJP, line 220: you don't need to do this. This is just a wrapper that is initialized ony when needed. No need to create it and keep it in memory for the entire duration of the lifetime of the object

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +605,121 @@ def _query_lending_balances(
 
         return balances
 
+    def query_lending_interests_history(
+            self,
+            cursor: 'DBCursor',
+            start_ts: Timestamp,
+            end_ts: Timestamp,
+    ) -> bool:
+        """Queries Binance lending interest history, transforms it into `HistoryBaseEntry` objects and save it in the database.
```
LefterisJP, line 614:
```suggestion
        """Queries Binance lending interest history, transforms it into `HistoryBaseEntry` objects and saves it in the database.
```

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +605,121 @@ def _query_lending_balances(
...
+            events = []
+            for lending_type in ('DAILY', 'ACTIVITY', 'CUSTOMIZED_FIXED'):
+                try:
+                    response = self._api_query_list_within_time_delta(
+                        api_type='sapi',
+                        start_ts=query_start_ts,
+                        time_delta=BINANCE_LENDING_INTEREST_HISTORY_INTERVAL_TS,
+                        end_ts=query_end_ts,
+                        method='lending/union/interestHistory',
+                        # size determines the length of the data returned,
```
LefterisJP, line 645: In my previous comment I said to move this to a constant and add the comment in the constant's declaration

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +605,121 @@ def _query_lending_balances(
...
+                    )
+                    return True
+
+                for entry in response:
+                    try:
+                        interest_received = deserialize_asset_amount(entry['interest'])  # noqa: E501
+                        if interest_received == ZERO:
+                            continue
+
+                        timestamp = TimestampMS(deserialize_timestamp_from_intms(entry['time']) * 1000)  # noqa: E501
```
LefterisJP, line 662: What are you doing here? This is completely wrong.

`entry['time']` is a timestam in MS. You are passing it to `deserialize_timestamp_from_intms` which takes it and divides it by `1000` to get the normal unix timestamp ... and then you multiply it again by 1000? Wtf? Just cast it normally and directly.

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +605,121 @@ def _query_lending_balances(
...
+                    event = HistoryBaseEntry(
+                        event_identifier=uuid4().bytes,
+                        sequence_index=0,  # since event_identifier is always different
+                        timestamp=timestamp,
+                        location=self.location,
+                        location_label=self.name,  # the name of the CEX instance
+                        asset=asset,
+                        balance=Balance(
+                            amount=interest_received,
+                            usd_value=ZERO,
```
LefterisJP, line 696: so you put it as zero here and leaveit up to the periodic task to query?

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +605,121 @@ def _query_lending_balances(
...
+            end_ts: Timestamp,
+    ) -> bool:
+        """Queries Binance lending interest history, transforms it into `HistoryBaseEntry` objects and save it in the database.
+
+        "DAILY" -> flexible savings.
+        "CUSTOMIZED_FIXED" -> fixed savings.
+        "ACTIVITY" -> fixed savings but as a result of token partners.
+
+        Lending Interest History Documentation:
+        https://binance-docs.github.io/apidocs/spot/en/#get-interest-history-user_data-2
```
LefterisJP, line 621: So I am quite pissed here. You are ignoring my questions/comments. I wrote here why was it written that the restriction is 30 days: https://github.com/rotki/rotki/pull/5681#discussion_r1126745716, when this restriction is not cheked anywhere.

You ignored it and just removed the comment and replaced it with a link to the docs. So get your shit together and answer the PR comments and **do not make me waste my time**.

1. Keep the link to the docs.
2. Re-add in the docstring the restriction that `end_ts` - `start_ts` need to be less than or equal to 30 days.
3. Explain to me **where is this restriction respected and enforced**? It's not anywhere and it should be in this function. This is right now directly called by the api, so it can have any value. So people wil query range [0 - now], get only last 30 days, save therange [0 - now] as queried in the DB and lose all older data.

rotkehlchen/exchanges/binance.py
LefterisJP: This is not auto breaking dude wtf? This depends on external input. If somehow the binance api is returning us the same page for whatever reason this is an infinite loop.

rotkehlchen/api/v1/schemas.py
```diff
@@ -2923,3 +2926,72 @@ def validate_schema(
...
+
+    def __init__(
+            self,
+            treat_eth2_as_eth: bool,
+    ) -> None:
+        super().__init__()
+        self.treat_eth2_as_eth = treat_eth2_as_eth
+
+    @post_load
+    def make_binance_savings_query(
```
LefterisJP, line 2945: What in the hell are you doign here? Did you copy paste the code of the base class's post-load? Why?

Why is all this needed? You don't need all this logic. And even if you did then jsut call the base class. Don't just copy paste.

rotkehlchen/api/rest.py
```diff
@@ -4155,3 +4082,125 @@ def add_evm_transaction_by_hash(
...
+            events: list[Any] = []
+            if location == Location.KRAKEN:
+                for event in events_raw:
+                    try:
+                        staking_event = StakingEvent.from_history_base_entry(event)
+                    except DeserializationError as e:
+                        log.warning(f'Could not deserialize staking event: {event} due to {str(e)}')  # noqa: E501
+                        continue
+                    events.append(staking_event)
+            else:
```
LefterisJP, line 4155: Why not count them ast staking events here? It's the same thing. Just keep it simple and a single interface

rotkehlchen/api/v1/resources.py
```diff
@@ -2421,6 +2422,33 @@ def get(self, name: str, location: Location) -> Response:
         return self.rest_api.get_user_binance_pairs(name=name, location=location)
 
 
+class BinanceSavingsResource(BaseMethodView):
+
+    def make_post_schema(self) -> BinanceSavingsSchema:
+        with self.rest_api.rotkehlchen.data.db.conn.read_ctx() as cursor:
+            settings = self.rest_api.rotkehlchen.data.db.get_settings(cursor)
+        return BinanceSavingsSchema(
+            treat_eth2_as_eth=settings.treat_eth2_as_eth,
```
LefterisJP, line 2431: As I said before why is this needed here? Just cause you are extending the superclass which needs it? In that case then this should be a big hint for you. That class inheritance scheme is wrong.

LefterisJP (commented):
Pretty dissapointed with how sloppy some things are. Needs work.

rotkehlchen/exchanges/binance.py
prettyirrelevant: The restriction is enforced using `_api_query_list_within_time_delta()`. 

It uses a while to ensure that if the difference exceeds 30days e.g 34 days, it queries the first 30 days and subsequently queries the 4 days left. It is achieved using a while loop.

rotkehlchen/exchanges/binance.py
prettyirrelevant: Initially, yeah. However, since this request is meant to be async, I'll just make the usd_value request here

rotkehlchen/api/v1/resources.py
prettyirrelevant: The use case of the inheritance is to handle `eth2 == eth` which is valid imo. Unless I'm not seeing the right thing, I don't think it's wrong.

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +611,140 @@ def _query_lending_balances(
...
+                    )
+                    return True
+
+                for entry in response:
+                    try:
+                        interest_received = deserialize_asset_amount(entry['interest'])  # noqa: E501
+                        if interest_received == ZERO:
+                            continue
+
+                        timestamp = TimestampMS(deserialize_timestamp(entry['time']))
```
LefterisJP, line 672: I don't think this is required again. As far as I can see timestamps in binance are `int` right? So you can just do this, I believe:
```suggestion
                        timestamp = TimestampMS(entry['time'])
```

rotkehlchen/exchanges/binance.py
```diff
@@ -589,6 +611,140 @@ def _query_lending_balances(
...
+                            f'Found {error_type} {self.name} asset {e.identifier}. '
+                            f'Ignoring its lending interest history query.',
+                        )
+                        continue
+
+                    try:
+                        usd_price = PriceHistorian().query_historical_price(
+                            from_asset=asset,
+                            to_asset=A_USD,
+                            timestamp=Timestamp(int(timestamp / 1000)),
```
LefterisJP, line 701: can probably use `ts_ms_to_sec` instead here

rotkehlchen/exchanges/binance.py
```diff
@@ -1205,15 +1365,34 @@ def _api_query_list_within_time_delta(
...
             }
             if additional_options:
                 options.update(additional_options)
-            result = self.api_query_list(api_type, method, options=options)
-            results.extend(result)
+                if method == 'lending/union/interestHistory':
+                    options['current'] = current_page
+
+            if method == 'lending/union/interestHistory':
+                last_queried_entry: tuple[str, int] = ('', 0)
```
LefterisJP, line 1380: add a comment here that this check is to just protect against misbehaving binance giving us an infinite loop

rotkehlchen/accounting/structures/base.py
```diff
@@ -359,10 +359,15 @@ def from_history_base_entry(cls, event: HistoryBaseEntry) -> 'StakingEvent':
 
     def serialize(self) -> dict[str, Any]:
         data = {
-            'event_type': self.event_type.serialize(),
             'asset': self.asset.identifier,
             'timestamp': self.timestamp,
             'location': str(self.location),
         }
+        if not (
```
LefterisJP, line 366: Would need a comment here why this is happening. Explan the endpoint only returns one type of event so this should not be serialized

rotkehlchen/api/v1/schemas.py
```diff
@@ -577,6 +559,50 @@ def make_staking_query(
...
+    )
+
+    def __init__(
+            self,
+            treat_eth2_as_eth: bool,
+    ) -> None:
+        super().__init__()
+        self.treat_eth2_as_eth = treat_eth2_as_eth
+
+    def _get_assets_list(
```
LefterisJP, line 575: See? This is so much better! Clean and nice.

LefterisJP (approved):
LGTM. Merging.

PR #5682, opened 2023-02-27
Changes AmountDisplay to support historic prices

kelsos:
Closes #(issue_number)

## Checklist


kelsos:
Still need to add a couple of tests but @LefterisJP if you find some time have a look at how it works for transactions for you.

frontend/app/src/components/display/amount/AmountDisplay.vue
```diff
@@ -0,0 +1,404 @@
...
+
+const isCurrentCurrency = isAssetPriceInCurrentCurrency(priceAsset);
+
+const { findCurrency } = useCurrencies();
+
+const historicExchangeRate = asyncComputed(
+  async () => {
+    if (!isDefined(sourceCurrency)) {
+      return One.negated();
+    }
```
lukicenturi, line 103: `
    if (get(sourceCurrency) === get(currentCurrency)) {
      return One;
    }
`

I think it makes sense to put this here (?)
If the currency is the same, we don't need to call the historic price.

frontend/app/src/components/display/amount/AmountDisplay.vue
```diff
@@ -0,0 +1,404 @@
...
+      const historicRate = get(historicExchangeRate);
+      if (historicRate.isPositive()) {
+        calculatedValue = get(value).multipliedBy(historicRate);
+      } else {
+        calculatedValue = Zero;
+      }
+    }
+
+    return calculatedValue;
+  }
```
lukicenturi, line 173: I guess we can return early here.

```
// If `sourceCurrency` and `currentCurrency` is not equal, we should convert the value
  if (sourceCurrencyVal !== currentCurrencyVal) {
    if (get(timestamp) > 0) {
      const historicRate = get(historicExchangeRate);
      if (historicRate.isPositive()) {
        return get(value).multipliedBy(historicRate);
      }
      return Zero;
    }

    return get(latestFiatValue);
  }
```

lukicenturi (changes requested):

frontend/app/src/components/display/amount/AmountDisplay.vue
kelsos: This is intentionally, to reduce the number of return statements in the function.

frontend/app/src/components/display/amount/AmountDisplay.vue
kelsos: Ok, the intent is not clearly communicated here.

The commented code is supposed to be a no-op for the compiler. I will turn it into an assertion instead, to that we don't change the signature of the `getHistoricPrice`.

Because the `historicPrice` `asyncComputed` is lazy it will only be evaluated when called. and when it is called the check has already occurred that the source and current currency are not the same.

Calling `getHistoricPrice` with the same asset here is a mistake it should not be called with the same from and to. We already do such checks before in the code.

I will turn both into assertions.

PR #5686, opened 2023-02-28
Tweak address scramble algorithm

lukicenturi:
Should fix 
image

The reason there are many duplicate characters, is that we multiplied the index with constant `10000`, which this value has `16` (num of possible characters) as one of its factor.

This is updated version, looks better
image


kelsos (approved):
lgtm @lukicenturi thanks

PR #5690, opened 2023-02-28
feat: add unit tests for pinia (store/blockchain/accounts)

lukicenturi:
Closes #(issue_number)

## Checklist

- [x] Unit test for "store/blockchain/accounts/*"
- [x] Remove all external calls during unit test


frontend/app/tests/unit/specs/store/blockchain/accounts/btc.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        }
+      ]
+    };
+
+    store.update(Blockchain.BTC, data);
+
+    expect(get(btc)).toEqual(data);
+  });
+
+  test('removeTag', () => {
```
kelsos, line 46: Can you make it so that the tests don't share state?

frontend/app/tests/unit/specs/store/blockchain/accounts/chains.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+      }
+    ];
+
+    store.update(Blockchain.OPTIMISM, accounts);
+
+    expect(get(optimism)).toEqual(accounts);
+  });
+
+  test('removeTag', () => {
+    const newData: GeneralAccountData = {
```
kelsos, line 32: same here make sure that tests don't share state

kelsos (commented):
In general it looks good, only one comment can you check that each test doesn't depend on state from the previous one.

Sharing state can lead to debugging headaches in the future

kelsos (approved):
let's merge it

PR #5698, opened 2023-02-28
Mention that PUT for evm accounts can be called async

yabirgb:
Add missing note that @kelsos saw in the docs


LefterisJP (approved):
LGTM

PR #5699, opened 2023-02-28
Fix deserialization of chain from remote spam asset data

LefterisJP:


rotkehlchen/assets/spam_assets.py
```diff
@@ -32,8 +33,12 @@ def _save_or_update_spam_assets(
     tokens_to_ignore = set()
     # Try to add custom list
     for info in assets_info:
-        chain_id = info.get('chain')
-        chain = ChainID.deserialize(chain_id) if chain_id is not None else ChainID.ETHEREUM
+        chain_name = info.get('chain')  # we use name since in thefuture we may have non-EVM chains
```
yabirgb, line 36:
```suggestion
        chain_name = info.get('chain')  # we use name since in the future we may have non-EVM chains
```

LefterisJP: aaargh

LefterisJP: will only fix if something fails though :smile:

yabirgb (approved):
Since all is green we merged and leave the wording for another PR

PR #5701, opened 2023-02-28
Ensure correct behaviour of periodic tasks

yabirgb:
Make the tasks that read from the user db the last queried timestamp work as expected. Also make sure to record the last queried timestamp for data updates

### DIFF
--- a/rotkehlchen/tests/unit/test_tasks_manager.py
+++ b/rotkehlchen/tests/unit/test_tasks_manager.py
@@ -6,12 +6,14 @@ import pytest
 
 from rotkehlchen.chain.bitcoin.hdkey import HDKey
 from rotkehlchen.chain.bitcoin.xpub import XpubData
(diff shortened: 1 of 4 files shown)


prettyirrelevant (approved):
LGTM

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -390,3 +393,40 @@ def simple_task():
             rotki.task_manager._maybe_update_snapshot_balances,
             simple_task,
         }
+
+
+def test_should_run_periodic_task(database: 'DBHandler') -> None:
+    """
+    Check that should_run_periodic_task correctly reads the settings when they have been
+    set and where the database doesn't have them yet.
```
LefterisJP, line 401:
```suggestion
    set and when the database doesn't have them yet.
```

LefterisJP (approved):
LGTM. Small wording nitpick to fix later

PR #5704, opened 2023-02-28
Claiming COMP after comptroller refill correct decoding

LefterisJP:
This fixes a bug where there was COMP reward accumulated in the controll but failed to be paid out in an older transaction due to (probably) being out of COMP.

Code is here:
https://etherscan.io/address/0xBafE01ff935C7305907c33BF824352eE5979B526#code

After being refilled the Distributed[Supplier/Borrower] events won't match the transfers.

This commit fixes that bug by forcing the transfer search even if compdelta is zero, as accumulation of rewards can have happend in a previous transaction.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` Ethereum transactions claiming COMP after comptroller's COMP ran out and has been refilled will now be decoded correctly as COMP rewards.
 * :bug:`-` Fixed an edge case where removing an EVM account multiple times in a row, while a transactions querying task ran, would result in an error.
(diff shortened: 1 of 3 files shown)


LefterisJP:
I have not VCRed the test yet, on purpose. As I wrote in discord, that single test (and I guess all other tests?!?) made about 90-100 etherscan transactions according to the test logs, which scares me and I would like to see why and see if there is something serious to fix.

LefterisJP:
Merging this since it's green. The test is not VCRed yet since I figured out something fundamentally wrong with one of our queries and will need to fix and this may affect all recorded cassettes. I hope VCR is smart enough to  just ignore missign calls

PR #5706, opened 2023-03-01
Remotely update default list of RPC nodes

### LINKED ISSUE #5236: Remotely updatable default rpc nodes
## Abstract

In rotki we have a default list of open ethereum rpc nodes that we use for querying various stuff. The problem is that many of these nodes don't live long and at some point they shut down. When this happens rotki becomes slower for people who have never changed that list.

We need to do something about this.

## Motivation

Be able to change default rpc nodes without making new releases.

## Specification

Make the per chain default nodes configuration as something that can be imported from a remote repo just like the assets are and have them pulled by rotki.

---

Similar to #5532

prettyirrelevant:
Closes #5236 

PR that adds `rpc_nodes` to the rotki/data repo.
https://github.com/rotki/data/pull/9


rotkehlchen/db/updates.py
```diff
@@ -80,6 +82,45 @@ def _get_remote_info_json(self) -> dict[str, Any]:
...
+            if updated_rpc_nodes is None:
+                log.error(f'Remote update {file_url} does not contain rpc_nodes key')
+                break
+
+            log.info(f'Applying update for rpc nodes from v{version}')
+            with (
+                GlobalDBHandler().conn.read_ctx() as cursor,
+                GlobalDBHandler().conn.write_ctx() as write_cursor,
+            ):
+                cursor.execute('DELETE FROM rpc_nodes')
```
yabirgb, line 108: this should have been a delete cursor. **Important** This is very dangerous because if the user had custom rpc nodes you will delete them. Also you will delete the nodes for all the chains so that means that you will always need a full list of all the nodes that need to exists, then... why do you need the versioning?

What I would do is in the update file keep a list of nodes to add and a list of nodes to deprecate if they are present in the database taking the endpoint as identifier

rotkehlchen/globaldb/schema.py
```diff
@@ -280,6 +280,18 @@
...
+CREATE TABLE IF NOT EXISTS rpc_nodes (
+    identifier INTEGER NOT NULL PRIMARY KEY,
+    name TEXT NOT NULL,
+    endpoint TEXT NOT NULL,
+    owned INTEGER NOT NULL CHECK (owned IN (0, 1)),
+    active INTEGER NOT NULL CHECK (active IN (0, 1)),
+    weight TEXT NOT NULL,
+    blockchain TEXT NOT NULL
+);
+"""
```
yabirgb, line 293: Why are you adding this to the globaldb? We have this already in the user db

rotkehlchen/db/updates.py
```diff
@@ -80,6 +82,45 @@ def _get_remote_info_json(self) -> dict[str, Any]:
...
+
+            updated_rpc_nodes = updates.get('rpc_nodes')
+            if updated_rpc_nodes is None:
+                log.error(f'Remote update {file_url} does not contain rpc_nodes key')
+                break
+
+            log.info(f'Applying update for rpc nodes from v{version}')
+            with (
+                GlobalDBHandler().conn.read_ctx() as cursor,
+                GlobalDBHandler().conn.write_ctx() as write_cursor,
```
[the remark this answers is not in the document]
yabirgb: Why are you storing them in the globaldb and the user db? This should stay in the user's db

rotkehlchen/db/updates.py
```diff
@@ -80,6 +82,45 @@ def _get_remote_info_json(self) -> dict[str, Any]:
...
+            ):
+                cursor.execute('DELETE FROM rpc_nodes')
+                write_cursor.executemany(
+                    'INSERT OR IGNORE INTO rpc_nodes(name, endpoint, owned, active, weight, blockchain) '  # noqa: E501
+                    'VALUES (?, ?, ?, ?, ?, ?)',
+                    [
+                        (node['name'], node['endpoint'], node['owned'], node['active'], str(FVal(node['weight'])), node['blockchain'])  # noqa: E501
+                        for node in updated_rpc_nodes
+                    ],
```
yabirgb, line 115: This can raise InputError I believe (please check it) if we make a mistake with one of the updates and the values don't have the correct type, please handle it and any possible keyError.

rotkehlchen/globaldb/upgrades/manager.py
```diff
@@ -28,6 +29,10 @@
         from_version=3,
         function=migrate_to_v4,
     ),
+    UpgradeRecord(
+        from_version=4,
+        function=migrate_to_v5,
+    ),
```
yabirgb, line 35: This shouldn't be in the globaldb. It should be store in the user's db in the table we have for rpc nodes.

rotkehlchen/chain/evm/nodes.py
```diff
@@ -1,27 +1,24 @@
...
 
     This is supposed to run for a clean new DB
     """
-    dir_path = Path(__file__).resolve().parent.parent.parent
-    with open(dir_path / 'data' / 'nodes.json') as f:
-        nodes_info = json.loads(f.read())
+    # get rpc nodes from global db.
+    nodes = globaldb_cursor.execute(
+        'SELECT name, endpoint, owned, active, weight, blockchain FROM rpc_nodes',
+    ).fetchall()
```
yabirgb, line 18: following this approach you are duplicating information and if you notice you have the same information both in the user's db and in the globaldb. Don't do this, just make the updates in the user's db

yabirgb (commented):
The approach needs to change. You have duplicated information and I would add a way to also remove nodes that are no longer active or we know don't perform well.

rotkehlchen/db/updates.py
```diff
@@ -117,20 +175,36 @@ def check_for_updates(self) -> None:
...
-            last_version = cursor.fetchone()
-        local_spam_assets_version = int(last_version[0]) if last_version is not None else 0
 
         # update spam assets
         latest_spam_assets_version = remote_information[UpdateType.SPAM_ASSETS.value]['latest']
         if local_spam_assets_version < latest_spam_assets_version:
             self.update_spam_assets(local_spam_assets_version, latest_spam_assets_version)
 
+        latest_rpc_nodes_version = remote_information[UpdateType.RPC_NODES.value]['latest']
+        if local_spam_assets_version < latest_rpc_nodes_version:
```
yabirgb, line 193:
```suggestion
        if local_rpc_nodes_version < latest_rpc_nodes_version:
```

yabirgb (commented):
Needs a few changes but seems that a test for spam assets broke

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -9,15 +9,15 @@
...
+from rotkehlchen.types import SPAM_PROTOCOL, ChainID, EvmTokenKind, SupportedBlockchain
 
 ETHEREUM_SPAM_ASSET_ADDRESS = make_evm_address()
 OPTIMISM_SPAM_ASSET_ADDRESS = make_evm_address()
 
 
 def mock_github_data_response(url, timeout):  # pylint: disable=unused-argument
     if 'info' in url:
-        return MockResponse(200, '{"spam_assets":{"latest":1}}')
+        return MockResponse(200, '{"spam_assets":{"latest":1}, "rpc_nodes": {"latest":1}}')
```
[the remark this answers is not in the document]
prettyirrelevant: I don't think I understand you here.

rotkehlchen/db/updates.py
```diff
@@ -10,20 +10,23 @@
...
-    from rotkehlchen.db.dbhandler import DBHandler
+    from rotkehlchen.db.dbhandler import DBCursor, DBHandler
     from rotkehlchen.user_messages import MessagesAggregator
 
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
 LAST_DATA_UPDATES_KEY: Final = 'last_data_updates_ts'
 SPAM_ASSETS_URL = 'https://raw.githubusercontent.com/rotki/data/{branch}/updates/spam_assets/assets_v{version}.json'  # noqa: E501
+RPC_NODES_URL = 'https://raw.githubusercontent.com/rotki/data/{branch}/updates/rpc_nodes/v{version}.json'  # noqa: E501
```
LefterisJP, line 29: Hey @prettyirrelevant @yabirgb, here I see a discrepancy in naming that should be addressed.

Why do spam assets have the prefix `assets_` for the filename? They are always going to be the same under the directory `spam_assets` no?

Also @prettyirrelevant why is there no PR in the data repo for the `rpc_nodes` stuff?

rotkehlchen/db/updates.py
```diff
@@ -81,6 +84,64 @@ def _get_remote_info_json(self) -> dict[str, Any]:
...
+            try:
+                updates = query_file(file_url, is_json=True)
+            except RemoteError as e:
+                log.warning(f'Failed to update rpc nodes due to {str(e)}')
+                break
+
+            updated_rpc_nodes = updates.get('rpc_nodes')
+            if updated_rpc_nodes is None:
+                log.error(f'Remote update {file_url} does not contain rpc_nodes key')
+                break
```
LefterisJP, line 103: All these, from line 92 to 104 is common between this and `update_spam_assets`. Abstract it away please.

rotkehlchen/db/updates.py
```diff
@@ -81,6 +84,64 @@ def _get_remote_info_json(self) -> dict[str, Any]:
...
+                ]
+            except (KeyError, DeserializationError) as e:
+                msg = f'missing key {str(e)}' if isinstance(e, KeyError) else str(e)
+                log.error(
+                    f'Unable to update rpc nodes to '
+                    f'{version} {updated_rpc_nodes} due to {msg}',
+                )
+                continue
+
+            log.info(
```
LefterisJP, line 122: No no no no. This is totally wrong. @yabirgb @prettyirrelevant did you even read the issue?

This is to change the **default** nodes. Not the user's nodes. Essentially what `populate_rpc_nodes_in_database()` does by reading `nodes.json`. That `nodes.json` should change.

And as an extra you can check the DB of the user, as you do here and **if and only if** a node is completely missing add it, and **if and only if** a node you want to delete should be deleted. But that should only happen for the last version. We don't need to run through all versions of an rpc node update. It's not a DB schema.

rotkehlchen/db/updates.py
```diff
@@ -117,20 +178,36 @@ def check_for_updates(self) -> None:
...
+                update_type=UpdateType.SPAM_ASSETS,
+            )
+            local_rpc_nodes_version = self._check_for_last_version(
+                cursor=cursor,
+                update_type=UpdateType.RPC_NODES,
             )
-            last_version = cursor.fetchone()
-        local_spam_assets_version = int(last_version[0]) if last_version is not None else 0
 
         # update spam assets
```
LefterisJP, line 190:
```suggestion
        # Update all remote data
```

LefterisJP (commented):
Hmm no guys I don't think that's it.

rotkehlchen/db/updates.py
prettyirrelevant: > Also @prettyirrelevant why is there no PR in the data repo for the rpc_nodes stuff?

No reason, was only waiting till  when the PR almost gets merged.

rotkehlchen/chain/evm/nodes.py
```diff
@@ -1,27 +1,24 @@
...
     from rotkehlchen.db.drivers.gevent import DBCursor
 
 
-def populate_rpc_nodes_in_database(write_cursor: 'DBCursor') -> None:
-    """Populates the rpc nodes in the datatabase from nodes.json
+def populate_rpc_nodes_in_database(
+        db_write_cursor: 'DBCursor',
+        globaldb_cursor: 'DBCursor',
+) -> None:
+    """Populates the rpc nodes in the user database from global database.
```
LefterisJP, line 12:
```suggestion
    """Populates the rpc nodes in the user database from the global DB.
```

rotkehlchen/db/updates.py
```diff
@@ -88,49 +90,111 @@ def update_spam_assets(self, local_version: int, to_version: int) -> None:
...
-            file_url = SPAM_ASSETS_URL.format(branch=self.branch, version=version)
-            try:
-                updates = query_file(file_url, is_json=True)
-            except RemoteError as e:
-                log.warning(f'Failed to update spam assets due to {str(e)}')
-                break
-            updated_assets = updates.get('assets')
-            if updated_assets is None:
-                log.error(f'Remote update {file_url} does not contain the assets key')
+            ok, updated_assets = self._get_file_content(
```
LefterisJP, line 93: nitpick. Too short a var name. 

```suggestion
            success, updated_assets = self._get_file_content(
```

rotkehlchen/db/updates.py
```diff
@@ -88,49 +90,111 @@ def update_spam_assets(self, local_version: int, to_version: int) -> None:
...
                     (UpdateType.SPAM_ASSETS.serialize(), version),
                 )
 
+    def update_rpc_nodes(self, local_version: int, to_version: int) -> None:
+        """
+        Check the updates in the inclusive range [local_version + 1, to_version] and update the
+        rpc nodes for those versions.
+        """
+        for version in range(local_version + 1, to_version + 1):
+            ok, updated_rpc_nodes = self._get_file_content(
```
LefterisJP, line 118: same here

rotkehlchen/db/updates.py
```diff
@@ -88,49 +90,111 @@ def update_spam_assets(self, local_version: int, to_version: int) -> None:
...
+                file_url=RPC_NODES_URL.format(branch=self.branch, version=version),
+                update_type=UpdateType.RPC_NODES,
+            )
+            if ok is False:
+                continue
+
+            log.info(f'Applying update for rpc nodes from v{version}')
+            with GlobalDBHandler().conn.write_ctx() as write_cursor:
+                write_cursor.execute('DELETE FROM rpc_nodes')
```
LefterisJP, line 127: explain that this is only the default nodes that change

```suggestion
            with GlobalDBHandler().conn.write_ctx() as write_cursor:
                # Update the default nodes, stores in the global DB
                write_cursor.execute('DELETE FROM rpc_nodes')
```

rotkehlchen/db/updates.py
```diff
@@ -88,49 +90,111 @@ def update_spam_assets(self, local_version: int, to_version: int) -> None:
...
                     'INSERT OR REPLACE INTO settings(name, value) VALUES (?, ?)',
                     (UpdateType.SPAM_ASSETS.serialize(), version),
                 )
 
+    def update_rpc_nodes(self, local_version: int, to_version: int) -> None:
+        """
+        Check the updates in the inclusive range [local_version + 1, to_version] and update the
+        rpc nodes for those versions.
+        """
+        for version in range(local_version + 1, to_version + 1):
```
LefterisJP, line 117: As I said in the last review, does it not make sense to just jump to last available update and just apply that here? Since rpc node updates are not cumulative and they do not build on one another.

rotkehlchen/db/updates.py
```diff
@@ -88,49 +90,111 @@ def update_spam_assets(self, local_version: int, to_version: int) -> None:
...
+
+    @staticmethod
+    def _check_for_last_version(cursor: 'DBCursor', update_type: UpdateType) -> int:
+        """This method checks the database for the last local version of `update_type`"""
+        found_version = cursor.execute(
+            'SELECT value FROM settings WHERE name=?',
+            (update_type.serialize(),),
+        ).fetchone()
+
+        return int(found_version[0]) if found_version is not None else 0
```
LefterisJP, line 181:
```suggestion
        return int(found_version[0]) if found_version is not None else 0
```

rotkehlchen/db/updates.py
```diff
@@ -88,49 +90,111 @@ def update_spam_assets(self, local_version: int, to_version: int) -> None:
...
+
+    @staticmethod
+    def _get_file_content(
+            file_url: str,
+            update_type: UpdateType,
+    ) -> tuple[bool, list[dict[str, Any]]]:
+        """Retrieves the content of data to be updated."""
```
LefterisJP, line 188:
```suggestion
        """Retrieves the content of data to be updated. Returns True/False for succes along with the content"""
```

rotkehlchen/globaldb/schema.py
```diff
@@ -10,6 +10,18 @@
 );
 """
 
+DB_CREATE_RPC_NODES = """
+CREATE TABLE IF NOT EXISTS rpc_nodes (
```
LefterisJP, line 14: name it default rpc nodes and also write in a comment that these are the defaults for each new user

rotkehlchen/globaldb/upgrades/v4_v5.py
```diff
@@ -0,0 +1,65 @@
...
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+def _create_new_tables(cursor: 'DBCursor') -> None:
+    log.debug('Enter _create_new_tables')
+
+    cursor.execute(
+        """
+        CREATE TABLE IF NOT EXISTS rpc_nodes (
```
LefterisJP, line 21: mind the name as I said before

rotkehlchen/globaldb/utils.py
```diff
@@ -7,7 +7,7 @@
 # Whenever you upgrade the global DB make sure to:
 # 1. Go to assets repo and tweak the min/max schema of the updates
 # 2. Tweak ASSETS_FILE_IMPORT_ACCEPTED_GLOBALDB_VERSIONS
-GLOBAL_DB_VERSION = 4
+GLOBAL_DB_VERSION = 5
```
LefterisJP, line 10: Make sure that this is also still fine for all assets in the update. Seems it is?
https://github.com/rotki/assets/blob/master/updates/info.json

But if we do another upgrade we will need to move the max_assets version for the latest updates.

rotkehlchen/tests/unit/globaldb/test_upgrades.py
```diff
@@ -289,6 +290,55 @@ def test_upgrade_v3_v4(globaldb):
         assert cursor.fetchone()[0] == YEARN_V1_ASSETS_IN_V3
 
 
+@pytest.mark.parametrize('globaldb_upgrades', [[]])
+@pytest.mark.parametrize('custom_globaldb', ['v4_global.db'])
```
LefterisJP, line 294: how did you generate v4_global.db for the test?

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -90,16 +113,39 @@ def test_no_update_performed(data_updater: RotkiDataUpdater) -> None:
...
         patch.object(data_updater, 'update_spam_assets') as spam_assets,
+        patch.object(data_updater, 'update_rpc_nodes') as rpc_nodes,
     ):
         data_updater.check_for_updates()
         assert spam_assets.call_count == 0
+        assert rpc_nodes.call_count == 0
+
+
+def test_update_rpc_nodes(data_updater: RotkiDataUpdater) -> None:
+    """Test that rpc nodes are for different blockchain are updated correctly."""
```
LefterisJP, line 137:
```suggestion
    """Test that rpc nodes for different blockchains are updated correctly."""
```

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -90,16 +113,39 @@ def test_no_update_performed(data_updater: RotkiDataUpdater) -> None:
...
+    data_updater.msg_aggregator.consume_warnings()
```
LefterisJP, line 138: Why consume warnings?

rotkehlchen/db/updates.py
```diff
@@ -88,49 +90,111 @@ def update_spam_assets(self, local_version: int, to_version: int) -> None:
...
+                write_cursor.executemany(
+                    'INSERT OR IGNORE INTO rpc_nodes(name, endpoint, owned, active, weight, blockchain) '  # noqa: E501
+                    'VALUES (?, ?, ?, ?, ?, ?)',
+                    [
+                        (node['name'], node['endpoint'], node['owned'], node['active'], str(FVal(node['weight'])), node['blockchain'])  # noqa: E501
+                        for node in updated_rpc_nodes
+                    ],
+                )
+
+            with self.user_db.conn.write_ctx() as write_cursor:
```
LefterisJP, line 137: And as we said in the last PR review, apart from updating the defaults it would make sense to do the following with the user DB:

Check if the endpoint/chain combination exists for the user. If not add it and update the weights.
Removing an old endpoint may be trickier. Not sure if it would make much sense. But perhaps check the previous defaults, compare to the new defaults and see if something is missing. If this missing node is present in the user DB remove it. But not totally sure.

Finally make sure this user rpc nodes editing is in a different function.

LefterisJP (commented):
Left some comments.

When you change the shipped global.db we need to know **exactly** what the diff is. What did you add?

I suppose only the new table? If yes, remember that after changing what I asked you will need to regenerate the global DB. The only good way to do this is run it through the upgrade. Do not do manual stuff otherwise we may end up with broken stuff in the shipped global DB.

rotkehlchen/tests/unit/test_data_updates.py
prettyirrelevant: oh i copied the template from that of `test_update_spam_assets`

prettyirrelevant:
> When you change the shipped global.db we need to know **exactly** what the diff is. What did you add?

```sh
(.venv) ➜  rotki git:(5236-remotely-update-rpc-nodes) ✗ sqldiff  ~/Downloads/global.db rotkehlchen/data/global.db
CREATE TABLE default_rpc_nodes (
    name TEXT NOT NULL,
    endpoint TEXT NOT NULL,
    owned INTEGER NOT NULL CHECK (owned IN (0, 1)),
    active INTEGER NOT NULL CHECK (active IN (0, 1)),
    weight TEXT NOT NULL,
    blockchain TEXT NOT NULL
);
INSERT INTO default_rpc_nodes(identifier,name,endpoint,owned,active,weight,blockchain) VALUES(1,'etherscan','',0,1,'0.30','ETH');
INSERT INTO default_rpc_nodes(identifier,name,endpoint,owned,active,weight,blockchain) VALUES(2,'cloudflare','https://cloudflare-eth.com/',0,1,'0.20','ETH');
INSERT INTO default_rpc_nodes(identifier,name,endpoint,owned,active,weight,blockchain) VALUES(3,'flashbots','https://rpc.flashbots.net/',0,1,'0.20','ETH');
INSERT INTO default_rpc_nodes(identifier,name,endpoint,owned,active,weight,blockchain) VALUES(4,'ankr','https://rpc.ankr.com/eth',0,1,'0.20','ETH');
INSERT INTO default_rpc_nodes(identifier,name,endpoint,owned,active,weight,blockchain) VALUES(5,'myetherwallet','https://nodes.mewapi.io/rpc/eth',0,1,'0.10','ETH');
INSERT INTO default_rpc_nodes(identifier,name,endpoint,owned,active,weight,blockchain) VALUES(6,'optimism etherscan','',0,1,'0.40','OPTIMISM');
INSERT INTO default_rpc_nodes(identifier,name,endpoint,owned,active,weight,blockchain) VALUES(7,'optimism official','https://mainnet.optimism.io',0,1,'0.20','OPTIMISM');
INSERT INTO default_rpc_nodes(identifier,name,endpoint,owned,active,weight,blockchain) VALUES(8,'optimism blastapi','https://optimism-mainnet.public.blastapi.io',0,1,'0.20','OPTIMISM');
INSERT INTO default_rpc_nodes(identifier,name,endpoint,owned,active,weight,blockchain) VALUES(9,'optimism ankr','https://rpc.ankr.com/optimism',0,1,'0.10','OPTIMISM');
INSERT INTO default_rpc_nodes(identifier,name,endpoint,owned,active,weight,blockchain) VALUES(10,'optimism 1rpc','https://1rpc.io/op',0,1,'0.10','OPTIMISM');
UPDATE settings SET value='5' WHERE rowid=29;
```

rotkehlchen/globaldb/utils.py
prettyirrelevant: > https://github.com/rotki/assets/blob/master/updates/info.json

Yeah, it is

rotkehlchen/db/updates.py
```diff
@@ -88,49 +90,175 @@ def update_spam_assets(self, local_version: int, to_version: int) -> None:
...
+                node['endpoint'],
+                node['owned'],
+                node['active'],
+                str(FVal(node['weight'])),
+                node['blockchain'],
+            )
+            for node in updated_rpc_nodes
+        ]
+        with GlobalDBHandler().conn.write_ctx() as write_cursor:
+            # Update the default nodes, stores in the global DB
```
LefterisJP, line 137:
```suggestion
            # Update the default nodes, stored in the global DB
```

rotkehlchen/db/updates.py
```diff
@@ -88,49 +90,175 @@ def update_spam_assets(self, local_version: int, to_version: int) -> None:
...
+    ) -> None:
+        """Updates the user nodes using the default nodes from the global db.
+
+        This function does the following:
+        1. Compares the previous default rpc nodes to the new ones, the difference is deleted from the user db.
+        2. Adds the new default rpc nodes to the user db.
+
+        indexes 1 & 2 -> endpoint of node
+        indexes 5 & 6 -> blockchain of node
+        """  # noqa: E501
```
LefterisJP, line 231:
```suggestion
        1. Compares the previous default rpc nodes to the new ones,
        and the difference is deleted from the user db.
        2. Adds the new default rpc nodes to the user db.

        indexes 1 & 2 -> endpoint of node
        indexes 5 & 6 -> blockchain of node
        """
```

rotkehlchen/db/updates.py
```diff
@@ -88,49 +90,175 @@ def update_spam_assets(self, local_version: int, to_version: int) -> None:
...
+        # checking if it's not already present in the user db.
+        nodes_to_add = [
+            node_to_add
+            for node_to_add in new_default_nodes
+            if (node_to_add[1], node_to_add[5]) not in user_rpc_nodes
+        ]
+        with self.user_db.user_write() as write_cursor:
+            log.debug(f'Adding {nodes_to_add} nodes to the user database...')
+            write_cursor.executemany(
+                'INSERT OR REPLACE INTO rpc_nodes(name, endpoint, owned, active, weight, blockchain) '  # noqa: E501
```
LefterisJP, line 261: the replace here is useless. Since the primary key is the identifier.
```suggestion
                'INSERT INTO rpc_nodes(name, endpoint, owned, active, weight, blockchain) '  # noqa: E501
```

rotkehlchen/globaldb/upgrades/v4_v5.py
```diff
@@ -0,0 +1,65 @@
...
+    with open(root_dir / 'data' / 'nodes.json') as f:
+        nodes_info = json.loads(f.read())
+
+    nodes_tuples = [
+        (node['name'], node['endpoint'], False, True, str(FVal(node['weight'])), node['blockchain'])  # noqa: E501
+        for node in nodes_info
+    ]
+    log.debug(nodes_tuples)
+    cursor.executemany(
+        'INSERT OR IGNORE INTO default_rpc_nodes(name, endpoint, owned, active, weight, blockchain) '  # noqa: E501
```
LefterisJP, line 48: Isn't ignore no-op anyway? The identifier is the primary key

```suggestion
        'INSERT INTO default_rpc_nodes(name, endpoint, owned, active, weight, blockchain) '  # noqa: E501
```

rotkehlchen/globaldb/upgrades/v4_v5.py
```diff
@@ -0,0 +1,65 @@
...
+        'VALUES (?, ?, ?, ?, ?, ?)',
+        nodes_tuples,
+    )
+
+    log.debug('Exit _populate_rpc_nodes')
+
+
+def migrate_to_v5(connection: 'DBConnection') -> None:
+    """This globalDB upgrade does the following:
+    - Adds an `rpc_nodes` table.
```
LefterisJP, line 58:
```suggestion
    - Adds the `default_rpc_nodes` table.
```

rotkehlchen/tests/unit/test_data_updates.py
LefterisJP: so if it's not needed don't do it

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -90,16 +113,55 @@ def test_no_update_performed(data_updater: RotkiDataUpdater) -> None:
...
+
+    # check the db state after updating
+    with GlobalDBHandler().conn.read_ctx() as cursor:
+        cursor.execute('SELECT COUNT(*) FROM default_rpc_nodes')
+        assert cursor.fetchone()[0] == 2
+
+    # check that 2 nodes are present in the user db
+    # 9 nodes were deleted since the updated rpc nodes data
+    # did not contain them.
+    with data_updater.user_db.conn.read_ctx() as cursor:
```
LefterisJP, line 161: can you also add a custom user node to the user DB and make sure this is not deleted? I think from the code it should not but need to have it in the test

LefterisJP (approved):
LGTM

PR #5707, opened 2023-03-01
Reduce etherscan queries

LefterisJP:
### [Dont query all block internal transactions, but all parent hash ones](https://github.com/rotki/rotki/commit/7fd2272ec01101e2f27a1c4a8f810f4e163b79cb) 

This can save **a lot** of etherscan queries for when decoding
transactions of a block with many internal transactions.

Before, in order to find internal transactions of a transaction we
used to query for the same block's internal transactions.

That added a lot of extra queries, as a block can have many many
transactions, each with their own internal transactions, and all were
pulled, since an internal transaction's parent is also pulled when the
internal is pulled.

This is easily solved since etherscan has recently added a query for internal
transactions of a transaction hash.

### [At deserialize_transaction remember receipt data to not query again](https://github.com/rotki/rotki/commit/d76aa3d8ab97cef6f0e368a23f66326171c199b0) 

This avoids 1 extra receipt query per transaction that got queried and
used receipt data query in order to figure out gas used


LefterisJP:
finally all green. Merging.

PR #5710, opened 2023-03-01
Change how history events are created from kraken

yabirgb:
- This PR changes the logic handling the deserialization for kraken events to make it store positive amounts and properly use the direction defined by the event types

Notes:

- This would need updates in the database but since we already made one change in develop for this we would need to combine them.

The issue appeared after 1.27.1

We already have tests that cover the correct creation of the trades, movements and different events

### DIFF
--- a/rotkehlchen/accounting/cost_basis/base.py
+++ b/rotkehlchen/accounting/cost_basis/base.py
@@ -166,7 +166,7 @@ class BaseCostBasisMethod(metaclass=ABCMeta):
         """
         # this is a temporary assertion to test that new accounting tools work properly.
         # Written on 06.06.2022 and can be removed after a couple of months if everything goes well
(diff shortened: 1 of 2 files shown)


LefterisJP (commented):
What are you fixing here and why did this suddenly appear? If this is/was a problem a test should exist that should fail without this change.

I still don't understand exactly what problem it is you are solving. The only change that happened here was properly deserialize the subtype null ones.

yabirgb:
@LefterisJP The problem it is solving is the presence of negative amounts in the history events for kraken that was creating issues in the PnL report because it assumes that all the amounts from history events are positive. Will create an scenario to reproduce it and improve this PR

yabirgb:
Closing this PR as I couldn't reproduce the error. Will reuse the code if I get to reproduce it

PR #5714, opened 2023-03-02
feat: allow addition of evm tx

### LINKED ISSUE #5582: Allow addition of EVM transaction by hash and association with an address
## Abstract

rotki is trying to pull as many relevant transactions for your addresses as possible. It's possible though that some transactions that involve your address can be missed. Especially if there is no value or token transfer directly from or to the address. We need to provide a way for users to correct this.

## Motivation

Find and decode transactions related to an address that rotki does not automatically find. Most notable example are the liquidation transactions for many protocols which are not detected since they contain no identifiable information about the user's address. But if the user has been liquidated they know it, and with this solution they can at least fix it by requesting the transaction directly by hash.

https://github.com/rotki/rotki/issues/5250

## Specification

- New endpoint to add a transaction by: `evm chain`, `tx_hash` and `associated_address`
- In the backend this will simply pull the transaction and associate it with the `associated_address`, get its receipt and decode all logs.
- In the frontend in the UI, not sure where the best place to place it would be. Probaly somewhere in the EVM transactions?

lukicenturi:
Closes #5582

## Checklist


frontend/app/src/components/history/ledger-actions/LedgerActionContent.vue
```diff
@@ -140,6 +136,10 @@ const fetchData = async (): Promise<void> => {
   await execute(0, pageParams);
 };
 
+const dialogTitle: Ref<string> = ref('');
+const dialogSubtitle: Ref<string> = ref('');
+const openDialog: Ref<boolean> = ref(false);
+const editableItem: Ref<LedgerActionEntry | null> = ref(null);
```
kelsos, line 142: Can you move these back to where they were before, please?

frontend/app/src/components/history/trades/ClosedTrades.vue
```diff
@@ -169,6 +163,10 @@ const {
   }
 );
 
+const dialogTitle: Ref<string> = ref('');
+const dialogSubtitle: Ref<string> = ref('');
+const openDialog: Ref<boolean> = ref(false);
+const editableItem: Ref<TradeEntry | null> = ref(null);
```
kelsos, line 169: Same here please move them back

frontend/app/src/components/history/trades/ClosedTrades.vue
```diff
@@ -249,6 +247,8 @@ const deleteTradeHandler = async () => {
   await fetchData();
 };
 
+const valid: Ref<boolean> = ref(false);
+const form: Ref<InstanceType<typeof ExternalTradeForm> | null> = ref(null);
```
kelsos, line 251: Also here move them back

frontend/app/src/components/history/transactions/TransactionContent.vue
```diff
@@ -172,6 +168,10 @@ const addEvent = (item: EthTransactionEntry) => {
   set(openDialog, true);
 };
 
+const dialogTitle: Ref<string> = ref('');
+const openDialog: Ref<boolean> = ref(false);
+const editableItem: Ref<EthTransactionEventEntry | null> = ref(null);
+const valid: Ref<boolean> = ref(false);
```
kelsos, line 174: Also here

frontend/app/src/components/history/transactions/TransactionContent.vue
```diff
@@ -234,6 +234,8 @@ const deleteEventHandler = async () => {
   set(confirmationPrimaryAction, '');
 };
 
+const form = ref<InstanceType<typeof TransactionEventForm> | null>(null);
+
```
kelsos, line 238: Same here

frontend/app/src/components/history/transactions/TransactionContent.vue
```diff
@@ -728,6 +770,18 @@ watch(loading, async (isLoading, wasLoading) => {
...
+      :display="openAddTxDialog"
+      :title="tc('transactions.dialog.add_tx')"
+      :primary-action="tc('common.actions.save')"
+      :action-disabled="!txFormValid || txForm?.loading"
+      :loading="loading || txForm?.loading"
+      @confirm="confirmSaveTx()"
+      @cancel="clearAddTxDialog()"
+    >
+      <transaction-form ref="txForm" v-model="txFormValid" />
+    </big-dialog>
```
kelsos, line 784: How much do we actually need from the `TransactionContent` and how much can go to a new component?

frontend/app/src/components/history/transactions/events/TransactionEventForm.vue
```diff
@@ -27,26 +26,23 @@ import { convertFromTimestamp, convertToTimestamp } from '@/utils/date';
...
+const props = withDefaults(
+  defineProps<{
+    value?: boolean;
+    edit?: EthTransactionEvent | null;
+    transaction?: EthTransactionEntry | null;
+  }>(),
+  {
+    value: false,
+    edit: () => null,
+    transaction: () => null
```
kelsos, line 38: Why `() => null` instead of `null`?

For arrays and objects, we use it because otherwise, they will pass a reference. But null is a special case and we can treat it like a primitive.

lukicenturi: I remember it threw an error in the past, but I don't see it anymore 😅, will change it

frontend/app/src/components/history/transactions/TransactionContent.vue
lukicenturi: maybe we can move the dialog, like how i did it with ledger action

kelsos: Yes it would make more sense to keep the minimum footprint in TransactionContent

frontend/app/src/services/utils.ts
```diff
@@ -105,7 +105,7 @@ export function validWithSessionAndExternalService(status: number): boolean {
  * @return The validity of the status code
  */
 export function validStatus(status: number): boolean {
-  return isValid([200, 400, 409], status);
+  return isValid([200, 400, 404, 409], status);
```
kelsos, line 108: The problem with this approach is that 404 is now a valid status code in a number of APIs were it should not be a valid status.

This means that instead of that it will try to parse this response instead of going directly to the error.

If we don't have a validator that has the combination of these statuses, create a new one instead. And check the rest documentation to see that we follow the same pattern as the one mentioned in the response statuses.

kelsos: If the codes are exactly the same the sure, it won't be a problem. 
The problem usually is if a code that is considered valid is not actually valid for the endpoint.

kelsos (approved):
lgtm @lukicenturi

PR #5715, opened 2023-03-02
feat: warn on unknown event type-subtype combination

### LINKED ISSUE #5703: Warn on unknown event type/subtype combination when adding/editing history events in EVM transactions
Rotki does allow the users to pick any event type / subtype combination which can result in Transaction event type "unknown" but there is no error or warning presented to the user. 

This can be improved by:

- Simple: give error when transaction type unknown ir just do not allow to save it.
- Medium complex: Only allow to select subtypes which match event type by shrinking the subtype selection list based on event type
- Premium solution: 90% of defi events I found in my history are decoded as 3-4 events:

1. Gas fee
2. Action A
3. Action B
4. Approval (optional)

Usually Action A and Action B are linked which rotki could consider to improve UX by proposing event type for Action B based on Action A.

- If Action A is set as Trade / Spend then Action B is Trade / Receive
- If Action A is Deposit / Deposit Asset then Action B is Received / Receive Wrapped
- If Action A is Withdrawal / Remove Deposit then Action B is Received / Return Wrapped

What I don't know if this can be used as a general procedure to decode not yet implemented protocols.

lukicenturi:
Closes #5703

## Checklist

image

### DIFF
--- a/frontend/app/src/components/history/TransactionEventForm.vue
+++ b/frontend/app/src/components/history/TransactionEventForm.vue
@@ -3,7 +3,7 @@ import { type BigNumber } from '@rotki/common';
 import useVuelidate from '@vuelidate/core';
 import { helpers, required } from '@vuelidate/validators';
 import dayjs from 'dayjs';
-import { type PropType } from 'vue';
+import { type ComputedRef, type PropType } from 'vue';
 import {
   HistoryEventSubType,
   type HistoryEventType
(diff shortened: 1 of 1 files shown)


kelsos:
Can't we hide the invalid options instead, rather than warning.

LefterisJP:
Can you also perhaps have a talk with our frontend designer about UI/UX here? We need to hide all non-essential and invalid data combinations.

And later help them by showing them what do combinations mean etc.
This started as a direct schema editing view, but this needs to change

lovetacirupeca:
Here Raquel 🙋🏻‍♀️
If I understand, the issue is we are showing some options that the user can't select. If it's the case, those options should be hidden from them.

kelsos (approved):
@lukicenturi lgtm, so if there is an invalid combination they will get some empty result?

lukicenturi:
> @lukicenturi lgtm, so if there is an invalid combination they will get some empty result?

yeah the sub event type will be empty when editing

PR #5720, opened 2023-03-03
feat: remove inheritAttrs from AmountInput.vue

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/.eslintrc.js
+++ b/frontend/app/.eslintrc.js
@@ -37,13 +37,6 @@ module.exports = {
     {
       files: ['*.ts'],
       extends: ['@rotki/eslint-config-ts']
-    },
-    {
-      // Disabled because it moves the inheritAttrs export to the setup script
-      files: ['AmountInput.vue'],
-      rules: {
-        'import/first': 'off'
-      }
     }
   ],
   settings: {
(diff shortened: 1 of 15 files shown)


frontend/app/src/components/inputs/AmountInput.vue
```diff
@@ -1,9 +1,3 @@
-<script lang="ts">
-export default {
-  inheritAttrs: false
-};
-</script>
-
```
lukicenturi, line 6: So I think we can remove this completely since we bind it manually too
`v-bind="useAttrs()"`

I remember why I set the `inheritAttrs: false` though, it was to prevent prop `type` to be passed to the `v-text-field`, but now I try to pass the type, it doesn't cause an error anymore, so I think it is safe to remove this.

kelsos:
@lukicenturi can you also update the eslint config? if we remove the option we don't need to exclude the file anymore

frontend/app/.eslintrc.js
```diff
@@ -33,19 +33,6 @@ module.exports = {
       }
     ]
   },
-  overrides: [
-    {
-      files: ['*.ts'],
-      extends: ['@rotki/eslint-config-ts']
-    },
```
kelsos, line 40: This we have to keep otherwise it messes lint :'(

kelsos (approved):
lgtm @lukicenturi thank you

Issue #5709, opened 2023-03-01
Special accounting for LP contracts


yabirgb:
## Problem description

Currently we don't make a special handling for pools where an asset is deposited and a token representing a participation in the pool is given to the user. What we do is consider them like a swap when the protocol is known like the case of curve pools or yearn vaults.

This generates issues like not having a clear representation of the profit/loss made after a withdrawal from such pools and the users actually having to go and match them.

## Solution

We want to improve the accounting of such contracts and have a clearer vision of what is happening on such events. To achieve this we base the improvement on the changes made in #5689 where we introduce the `contract` field.

Using this contract we can filter the events that happened withing a specific contract and this allows us to track the profits made in a certain pool.

Since for example curve pools allow to deposit one or multiple tokens and withdraw in one or multiple tokens too we would need to track the PnL at least initially using the profit currency and not the underlying tokens.

## Specification

1. We will allow users when viewing EVM transactions to mark a certain deposit to a contract, as a pool deposit thus also marking the target evm address as 'pool' (or a better name) meaning that interaction with that exchanging tokens are subject to this new accounting method. Also the curve pool, yearn vaults and others that we already know will be used here.
2. We will introduce a new `TxAccountingTreatment` that will handle the special case of `pool contracts`. This new setting will search the related events using the `event identifier` and if we are depositing it will track the deposits made to the contract and when withdrawing (i.e. returning the wrapping token) it will track the PnL accrued in the pool. As we said this should be calculated using the profit currency

## Future improvements

- This will allow to track performance in certain pools and this can be used to create statistics. Right now this change will be tight to the accounting part so is out of the scope for this issue.
- Contracts like curve gauges are special cases since they use virtual balances and don't emit a LP token balance. Performance of those would be scope for another issue.

LefterisJP:
This heavily overlaps with #9457. So moved the unique parts of it there and closing it.

Issue #2203, opened 2021-01-29
Move to backend module information from hardcoded values


kelsos:
## Description

Endpoints have been added to provide names for modules along with the ids. 

### Tasks
- [ ] Introduce a simple id to icon mapping with some default fallback
- [ ] Simplify code

LefterisJP:
Damn this issue hurts. Since then a lot more stuff have been added that need to be provided by the backend to the frontend and not be duplicating data between both places. Moving it again.

LefterisJP:
I am closing this. Some mapping have moved. For other specific ones we either move them when we find them or make separate issues for them.

PR #5721, opened 2023-03-03
Correctly decode token lefts when depositing in balancer v1

### LINKED ISSUE #5697: Balancer decoding: not deposited assets declared as "Receive wrapped"
When Depositing into balancer the not deposited and received assets are declared as "received wrapped" even they are not wrapped tokens.
See screenshot:
image

I can provide details in discord. PM me.

yabirgb:
Seems that balancer V1 was returning some tokens to the users if they were not used during the deposit (I guess due to changes in the conditions on the pool but not sure) and this was a special case that the decoder didn't know about.

Closes #5697

This PR also changes the type when depositing from staking to deposit


rotkehlchen/chain/ethereum/modules/balancer/decoder.py
```diff
@@ -249,14 +249,47 @@ def _maybe_enrich_balancer_v1_events(
...
+            ):
+                deposited_assets.append(event.asset)
+            elif (
+                event.counterparty == CPT_BALANCER_V1 and
+                event.event_type == HistoryEventType.DEPOSIT and
+                event.event_subtype == HistoryEventSubType.RECEIVE_WRAPPED and
+                event.asset in deposited_assets
+            ):
+                # in this case we got refunded one of the assets deposited
+                event.event_subtype = HistoryEventSubType.REMOVE_ASSET
```
prettyirrelevant, line 287: I'm not so sure about using `REMOVE_ASSET` here. 
Also, shouldn't you change the `event_type` from `DEPOSIT` to `RECEIVE`?

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
```diff
@@ -249,14 +249,47 @@ def _maybe_enrich_balancer_v1_events(
...
+            transaction: EvmTransaction,  # pylint: disable=unused-argument
+            decoded_events: list[HistoryBaseEntry],
+            all_logs: list[EvmTxReceiptLog],  # pylint: disable=unused-argument
+    ) -> list[HistoryBaseEntry]:
+        """
+        It can happen that after sending tokens to the DSProxy in balancer V1 the amount of tokens
+        required for the deposit is lower than the amount sent and then those tokens are returned
+        to the DSProxy and then to the user.
+        """
+        deposited_assets = []
```
prettyirrelevant, line 271:
```suggestion
        deposited_assets = set()
```

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
```diff
@@ -249,14 +249,47 @@ def _maybe_enrich_balancer_v1_events(
...
+
+        for event in decoded_events:
+            if (
+                event.counterparty == CPT_BALANCER_V1 and
+                event.event_type == HistoryEventType.DEPOSIT and
+                event.event_subtype == HistoryEventSubType.DEPOSIT_ASSET
+            ):
+                deposited_assets.append(event.asset)
```
prettyirrelevant, line 279:
```suggestion
                deposited_assets.add(event.asset)
```

prettyirrelevant (commented):
minor comments

rotkehlchen/chain/ethereum/modules/balancer/accountant.py
```diff
@@ -0,0 +1,45 @@
...
+                accountant_cb=self._process_deposit_or_withdrawal,
+            ),
+            get_tx_event_type_identifier(HistoryEventType.TRADE, HistoryEventSubType.SPEND, CPT_BALANCER_V2): TxEventSettings(  # noqa: E501
+                taxable=True,
+                count_entire_amount_spend=False,
+                count_cost_basis_pnl=True,
+                method='spend',
+                accounting_treatment=TxAccountingTreatment.SWAP,
+            ),
+        }
```
LefterisJP, line 45: Where is the treatment for all the other transactions? Such as the ones you touched? DEPOSIT/DEPOSIT_ASSET and WITHDRAWAL/REMOVE_ASSET? Is it via the superclass you just added?

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
```diff
@@ -272,3 +392,6 @@ def enricher_rules(self) -> list[Callable]:
 
     def counterparties(self) -> list[str]:
         return [CPT_BALANCER_V1, CPT_BALANCER_V2]
+
+    def post_decoding_rules(self) -> list[tuple[int, Callable]]:
+        return [(1, self._check_refunds_v1), (0, self._check_deposits_withdrawals_v1)]
```
LefterisJP, line 397: Change the order here to not confuse the reader. The higher the number is the later the function is ran. And explain why you are giving this priority. Why first deposits/withdrawalsand then refunds.

```suggestion
        return [(0, self._check_deposits_withdrawals_v1), (1, self._check_refunds_v1), ]
```

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
```diff
@@ -249,14 +249,134 @@ def _maybe_enrich_balancer_v1_events(
...
+            transaction: EvmTransaction,  # pylint: disable=unused-argument
+            decoded_events: list[HistoryBaseEntry],
+            all_logs: list[EvmTxReceiptLog],  # pylint: disable=unused-argument
+    ) -> list[HistoryBaseEntry]:
+        """
+        It can happen that after sending tokens to the DSProxy in balancer V1 the amount of tokens
+        required for the deposit is lower than the amount sent and then those tokens are returned
+        to the DSProxy and then to the user.
+        """
+        deposited_assets = set()
```
LefterisJP, line 270: Can you add a transaction type constraint here? Otherwise this will run for all transactions.

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
LefterisJP: It's confusing, I agree. You guys seem to pick new types/subtypes combinations for everything depending on how you feel on a given day. This is confusing as heck and does not scale. Will bite us in the ass and does not let us generalize for common types of events between protocols.

We don't have a refund type though so not sure what the right approach here is either.

We need to generalize all the combinations and write rules for them.

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
```diff
@@ -249,14 +249,134 @@ def _maybe_enrich_balancer_v1_events(
...
+                )
+            ):
+                deposit_events.append(event)
+            elif (
+                event.event_type == HistoryEventType.WITHDRAWAL and
+                event.event_subtype == HistoryEventSubType.REMOVE_ASSET
+            ):
+                withdrawal_events.append(event)
+
+        if balancer_v1_event is False:
```
LefterisJP, line 337: What happens if there is multiple receive, return or withdrawal events in a transaction? Your approach will have the last one only at this point.

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
```diff
@@ -249,14 +249,134 @@ def _maybe_enrich_balancer_v1_events(
...
+            self,
+            transaction: EvmTransaction,  # pylint: disable=unused-argument
+            decoded_events: list[HistoryBaseEntry],
+            all_logs: list[EvmTxReceiptLog],  # pylint: disable=unused-argument
+    ) -> list[HistoryBaseEntry]:
+        """
+        Check for accounting in v1 that the deposits/withdrawals events have the needed information
+        to process them during accounting.
+        """
+        receive_event, return_event = None, None
```
LefterisJP, line 302: add constraing for the type of transaction this should run for. Otherwise this will run as post decoding for all transactions

rotkehlchen/chain/evm/accounting/interfaces.py
```diff
@@ -36,3 +46,68 @@ def event_settings(self, pot: 'AccountingPot') -> dict[str, 'TxEventSettings']:
     def reset(self) -> None:  # pylint: disable=no-self-use
         """Subclasses may implement this to reset state between accounting runs"""
         return None
+
+
+class DepositableAccountantInterface(ModuleAccountantInterface):
+    """
+    Interface for protocols that allow to deposit multiple tokens in exchange for a token
+    representing the position in the pool. Examples are:
+    - Curve
```
LefterisJP, line 55: Shouldn't this be used in a lot more places? Isn't this kind of related to https://github.com/rotki/rotki/issues/5709?

rotkehlchen/chain/evm/accounting/interfaces.py
```diff
@@ -36,3 +46,68 @@ def event_settings(self, pot: 'AccountingPot') -> dict[str, 'TxEventSettings']:
...
+            # Pool token is received, which means it is a deposit
+            events_to_consume = event.extra_data.get('deposit_events_num', None) if event.extra_data is not None else None  # noqa: E501
+            method = 'spend'
+        else:  # Withdrawal
+            events_to_consume = event.extra_data.get('withdrawal_events_num', None) if event.extra_data is not None else None  # noqa: E501
+            method = 'acquisition'
+
+        if events_to_consume is None:
+            log.debug(
+                f'Could not find the number of events to consume for a curve deposit/withdrawal '
```
LefterisJP, line 86: You need the name here, so this needs a constructor .. or you can use the counterparty as you did below

```suggestion
                f'Could not find the number of events to consume for a {self.name} deposit/withdrawal '
```

rotkehlchen/tests/unit/accounting/evm/test_balancer.py
```diff
@@ -0,0 +1,219 @@
...
+        ProcessedAccountingEvent(
+            type=AccountingEventType.TRANSACTION_EVENT,
+            notes='Refunded 6578.947368421052624578 mUSD after depositing in Balancer V1 pool',
+            location=Location.ETHEREUM,
+            timestamp=TIMESTAMP_1_SECS,
+            asset=Asset('eip155:1/erc20:0xe2f2a5C287993345a840Db3B0845fbC70f5935a5'),
+            free_amount=ZERO,
+            taxable_amount=FVal('6578.947368421052624578'),
+            price=Price(ONE),
+            pnl=PNL(taxable=ZERO, free=ZERO),  # Deposits are not taxable
```
LefterisJP, line 185:
```suggestion
            pnl=PNL(taxable=ZERO, free=ZERO),  # Withdrawals/refunds are not taxable
```

rotkehlchen/tests/unit/accounting/evm/test_balancer.py
```diff
@@ -0,0 +1,219 @@
...
+        ProcessedAccountingEvent(
+            type=AccountingEventType.TRANSACTION_EVENT,
+            notes='Refunded 6571.085163 USDC after depositing in Balancer V1 pool',
+            location=Location.ETHEREUM,
+            timestamp=TIMESTAMP_1_SECS,
+            asset=A_USDC,
+            free_amount=ZERO,
+            taxable_amount=FVal('6571.085163'),
+            price=Price(ONE),
+            pnl=PNL(taxable=ZERO, free=ZERO),  # Deposits are not taxable
```
LefterisJP, line 199: as above

rotkehlchen/tests/unit/accounting/evm/test_balancer.py
```diff
@@ -0,0 +1,219 @@
...
+        ProcessedAccountingEvent(
+            type=AccountingEventType.TRANSACTION_EVENT,
+            notes='Receive 1675.495956074927519908 BPT from a Balancer v1 pool',
+            location=Location.ETHEREUM,
+            timestamp=TIMESTAMP_1_SECS,
+            asset=Asset('eip155:1/erc20:0x72Cd8f4504941Bf8c5a21d1Fd83A96499FD71d2C'),
+            free_amount=FVal('1675.495956074927519908'),
+            taxable_amount=ZERO,
+            price=Price(ONE),
+            pnl=PNL(taxable=ZERO, free=ZERO),  # Deposits are not taxable
```
LefterisJP, line 213: Withdrawals are not taxable as above.

But then .... comes the question. What about profit or loss. You can withdraw more/less than you deposited in which case profit or loss is realized. This needs to be accounted for.

LefterisJP (commented):
Some comments. Also this makes it obvious to me we need (related to https://github.com/rotki/rotki/issues/2203) to maintain a list of all valid Type/Subtypes in our docs and with examples of what they represent.

As I wrote in the review you guys seem to use random combinations all over the decoders and this is unsustainable and it kills the generalized approach the events give us.

I need you (or someone else from the backend) to create this list, along with examples where each combination is valid. Then to go through all decoders and make sure the events match.

This is the only way to settle on an accepted set of valid combinations.

rotkehlchen/chain/ethereum/modules/balancer/accountant.py
yabirgb: Yes, what the superclass does is group the events as we do with curve

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
yabirgb: Actually this is wrong and this is why I sorted like this. The comment in the interface is wrong. https://github.com/rotki/rotki/blob/6a2a706c295e4a307c3b1f91baf431f52f5dc457/rotkehlchen/chain/evm/decoding/decoder.py#L142 `sort` by default sorts in ascending order and by doing reverse 1 has higher priority than 0

```
>>> a = [0,1,2]
>>> a.sort(reverse=True)
>>> a
[2, 1, 0]
```

rotkehlchen/tests/unit/accounting/evm/test_balancer.py
yabirgb: This is the event for receiving the pool token, not the withdrawal

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
yabirgb: Not sure if there is a multiple deposit case in the wild but then this is a flaw that we also have in the curve decoder.

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
yabirgb: There is no constraint because the balancer v1 events execute against a DSProxy for each user. I'm not sure if there is any other constraint that can be used as filter. What I do is exit early if there is no balancer v1 events

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
yabirgb: I have changed it so 0 is the highest priority

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
yabirgb: I have handled this case now

rotkehlchen/tests/unit/accounting/evm/test_balancer.py
yabirgb: This is the case of only a deposit where you get a refund. For the case of withdrawing a token as we said will be handled in a different PR only with that change

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
yabirgb: As discussed in discord the types here follow the curve ones now and I'll create a list + enforce the types in the decoders

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -139,7 +139,7 @@ def __init__(
         rules = self._recursively_initialize_decoders(self.chain_modules_root)
         self.rules += rules
         # Sort post decoding rules by priority (which is the first element of the tuple)
-        self.rules.post_decoding_rules.sort(key=lambda x: x[0], reverse=True)
+        self.rules.post_decoding_rules.sort(key=lambda x: x[0])
```
LefterisJP, line 142: Wait ... isn't this change breaking all the existing priorities for post-decoding rules?. @nebolax put it there for a reason, no?

rotkehlchen/chain/ethereum/modules/balancer/v1/decoder.py
```diff
@@ -0,0 +1,274 @@
...
+    ) -> list[HistoryBaseEntry]:
+        """
+        It can happen that after sending tokens to the DSProxy in balancer V1 the amount of tokens
+        required for the deposit is lower than the amount sent and then those tokens are returned
+        to the DSProxy and then to the user.
+        """
+        deposited_assets = set()
+
+        for event in decoded_events:
+            if event.counterparty != CPT_BALANCER_V1:
```
LefterisJP, line 155: No this is not good enough. So we need more work on the post-decoding rules. I think we need to tag transactions with which protocols they decoded and then only go in the post-decoding rules if they match. Otherwise this system won't scale.

rotkehlchen/chain/ethereum/modules/balancer/v1/decoder.py
```diff
@@ -0,0 +1,274 @@
...
+        Check for accounting in v1 that the deposits/withdrawals events have the needed information
+        to process them during accounting.
+        """
+        related_events: list[HistoryBaseEntry] = []
+        related_events_map: dict[HistoryBaseEntry, list[HistoryBaseEntry]] = {}
+        # last event is only tracked in the case of exiting a pool and contains the event
+        # sending the BPT token
+        last_event = None
+        for event in decoded_events:
+            if event.counterparty != CPT_BALANCER_V1:
```
LefterisJP, line 191: same comment here

rotkehlchen/chain/ethereum/modules/curve/accountant.py
```diff
@@ -88,3 +36,7 @@ def event_settings(self, pot: 'AccountingPot') -> dict[str, TxEventSettings]:  #
                 accountant_cb=self._process_deposit_or_withdrawal,
             ),
         }
+
+    @property
+    def name(self) -> str:
```
LefterisJP, line 41: In the end why do this? Didn't we say that the counterparty is good enough or am I missing something?

rotkehlchen/chain/evm/accounting/interfaces.py
```diff
@@ -36,3 +46,74 @@ def event_settings(self, pot: 'AccountingPot') -> dict[str, 'TxEventSettings']:
...
+                f'Could not find the number of events to consume for a {self.name} '
+                f'deposit/withdrawal transaction {event.serialized_event_identifier}',
+            )
+            return
+
+        # Consume the events
+        for idx in range(events_to_consume):
+            next_event = next(other_events, None)
+            if next_event is None:
+                log.debug(f'Could not consume event nr. {idx} for {event.counterparty} deposit/withdrawal')  # noqa: E501
```
LefterisJP, line 95: This is what I mean. Is it gonna be `event.coutnerparty` or `self.name`?

rotkehlchen/chain/ethereum/modules/balancer/decoder.py
LefterisJP: have you handled this in both cases?

rotkehlchen/tests/unit/accounting/evm/test_balancer.py
LefterisJP: Make sure that you know who is assigned to do it and when. Let's please not "forget that we had to do that"

rotkehlchen/chain/evm/decoding/decoder.py
nebolax: Currently I think it won't break anything since currently all post-decoding rules have priority of `0`. So I would say it is just more a convenience question for the future -- whether decoders with lower or with higher priority should run first.

yabirgb: I didn't change it for convenience. It was to match what the docstring was saying https://github.com/rotki/rotki/blob/6a2a706c295e4a307c3b1f91baf431f52f5dc457/rotkehlchen/chain/evm/decoding/interfaces.py#L58

As @nebolax said I checked that this change doesn't affect any other decoder

rotkehlchen/chain/ethereum/modules/balancer/v1/decoder.py
yabirgb: I have added a new commit that modifies how the rules are used. Now this logic is only executed when there is a balancer event detected (as it should)

rotkehlchen/chain/ethereum/decoding/decoder.py
```diff
@@ -64,20 +64,20 @@ def _maybe_enrich_transfers(
             decoded_events: list[HistoryBaseEntry],
             action_items: list[ActionItem],  # pylint: disable=unused-argument
             all_logs: list[EvmTxReceiptLog],  # pylint: disable=unused-argument
-    ) -> tuple[Optional[HistoryBaseEntry], list[ActionItem]]:
+    ) -> tuple[Optional[HistoryBaseEntry], list[ActionItem], Optional[str]]:
```
LefterisJP, line 67: add a docstring explaining what this function does and what it returns

rotkehlchen/chain/ethereum/decoding/decoder.py
```diff
@@ -150,26 +150,29 @@ def _enrich_protocol_tranfers(  # pylint: disable=no-self-use
...
-            except (UnknownAsset, WrongAssetType) as e:
-                log.error(
-                    f'Failed to enrich transfer due to unknown asset {event.asset}. {str(e)}',
-                )
-                # Don't try other rules since all of them will fail to resolve the asset
-                break
-
-            if transfer_enriched:
-                break
+    ) -> Optional[str]:
```
LefterisJP, line 153: Same here. Need a docstring

rotkehlchen/chain/ethereum/modules/balancer/v1/decoder.py
```diff
@@ -0,0 +1,289 @@
...
+            decoded_events: list[HistoryBaseEntry],
+            all_logs: list[EvmTxReceiptLog],  # pylint: disable=unused-argument
+            counterparties: set[str],
+    ) -> list[HistoryBaseEntry]:
+        """
+        It can happen that after sending tokens to the DSProxy in balancer V1 the amount of tokens
+        required for the deposit is lower than the amount sent and then those tokens are returned
+        to the DSProxy and then to the user.
+        """
+        if CPT_BALANCER_V1 not in counterparties:
```
LefterisJP, line 153: is this check needed now? Won't it only run if a balancer even is detected?

rotkehlchen/chain/ethereum/modules/balancer/v1/decoder.py
```diff
@@ -0,0 +1,289 @@
...
+        # sending the BPT token
+        last_event = None
+        for event in decoded_events:
+            if event.counterparty != CPT_BALANCER_V1:
+                continue
+
+            # When joining a pool first we spend the assets and then we receive the BPT token.
+            # In the case of exiting the pool first we return the BPT token and then we receive
+            # the assets.
+            # To handle properly the case of multiple events happening in the same transaction we
```
LefterisJP, line 204:
```suggestion
            # To handle the case of multiple events happening in the same transaction properly we
```

rotkehlchen/chain/ethereum/modules/balancer/v1/decoder.py
```diff
@@ -0,0 +1,289 @@
...
+                    related_events_map[last_event] = related_events
+                    last_event = None
+                    related_events = []
+
+                related_events.append(event)
+            elif (
+                event.event_type == HistoryEventType.WITHDRAWAL and
+                event.event_subtype == HistoryEventSubType.REMOVE_ASSET
+            ):
+                related_events.append(event)
```
LefterisJP, line 245: For such convoluted if/elif leave empty lines, otherwise it's really hard to read

```suggestion
                    last_event = None
            elif (
                event.event_type == HistoryEventType.SPEND and
                event.event_subtype == HistoryEventSubType.RETURN_WRAPPED
            ):
                related_events = []
                last_event = event

            elif (
                (
                    event.event_type == HistoryEventType.DEPOSIT and
                    event.event_subtype == HistoryEventSubType.DEPOSIT_ASSET
                ) or
                (
                    event.event_type == HistoryEventType.RECEIVE and
                    event.event_subtype == HistoryEventSubType.REMOVE_ASSET
                )
            ):
                # In the case that we are handling a join that comes after an exit we have to
                # check if there is a change in the direction of the assets (from spending them)
                # to receiving them
                if last_event is not None:
                    # save the exit event and reset the related events
                    related_events_map[last_event] = related_events
                    last_event = None
                    related_events = []

                related_events.append(event)

            elif (
                event.event_type == HistoryEventType.WITHDRAWAL and
                event.event_subtype == HistoryEventSubType.REMOVE_ASSET
            ):
                related_events.append(event)
```

rotkehlchen/chain/ethereum/modules/balancer/v1/decoder.py
```diff
@@ -0,0 +1,289 @@
...
+                ) or
+                (
+                    event.event_type == HistoryEventType.RECEIVE and
+                    event.event_subtype == HistoryEventSubType.REMOVE_ASSET
+                )
+            ):
+                # In the case that we are handling a join that comes after an exit we have to
+                # check if there is a change in the direction of the assets (from spending them)
+                # to receiving them
+                if last_event is not None:
```
LefterisJP, line 234: here if `last_event` is None is it fine to add `event` to `related_events`? I don't really understand the logic here ... it's really hard to read

rotkehlchen/chain/evm/decoding/structures.py
```diff
@@ -22,3 +23,30 @@ class ActionItem(NamedTuple):
...
+    Define the different types of post decoding rules. There are some rules that only
+    need to be executed when certain decoders have been used, this is the case of balancer
+    post decoding rules for example but there are others that execute always and is by checking
+    the transaction address how we choose to run them or not (happens with uniswap).
+
+    This data class defines the field `common` for decoding rules that need to be executed
+    always and the field `with_counterparties` that will be later filtered to extract from it
+    the needed rules.
+    """
+    common: list[tuple[int, Callable]]
```
LefterisJP, line 40: Why common rules? I don't think they should ever exist. Post-decoding is inherently protocol specific. There should never be common post decoding rules.

rotkehlchen/chain/ethereum/modules/uniswap/utils.py
```diff
@@ -239,7 +239,7 @@ def find_uniswap_v2_lp_price(
             f'Couldnt retrieve non zero price information for tokens {token0}, {token1} '
             f'with result {token0_price}, {token1_price}',
         )
-    numerator = (token0_supply * token0_price + token1_supply * token1_price)
+    numerator = token0_supply * token0_price + token1_supply * token1_price
```
LefterisJP, line 242: think this is already done in develop. So this will conflict

rotkehlchen/chain/ethereum/modules/uniswap/v2/common.py
```diff
@@ -89,6 +89,7 @@ def decode_uniswap_v2_like_swap(
         if event.asset == A_ETH and event.event_type == HistoryEventType.RECEIVE:
             received_eth += event.balance.amount
 
+    counterparty_found = False
```
LefterisJP, line 92: Why this variable? Why not just have `counterparty` as `None` and then just populate it if found?

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -557,5 +558,8 @@ def enricher_rules(self) -> list[Callable]:
     def counterparties(self) -> list[str]:
         return [CPT_UNISWAP_V3]
 
-    def post_decoding_rules(self) -> list[tuple[int, Callable]]:
-        return [(0, self._routers_post_decoding)]
+    def post_decoding_rules(self) -> PostDecodingrules:
+        return PostDecodingrules(
+            common=[(0, self._routers_post_decoding)],
```
LefterisJP, line 563: No as I said the "common" makes no sense. Post decoding rules should only exist if the decoder runs.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -187,10 +187,16 @@ def _recursively_initialize_decoders(
...
+                    first_counterparty = self.decoders[class_name].counterparties()[0]
+                    for address, decoders_by_address in decoders.items():
+                        decoders_with_counterparty[address] = (decoders_by_address, first_counterparty)  # noqa: E501
+
+                    results.address_mappings.update(decoders_with_counterparty)
                     results.event_rules.extend(self.decoders[class_name].decoding_rules())
-                    results.token_enricher_rules.extend(self.decoders[class_name].enricher_rules())
-                    results.post_decoding_rules.extend(self.decoders[class_name].post_decoding_rules())  # noqa: E501
+                    results.token_enricher_rules.extend([(self.decoders[class_name].enricher_rules(), first_counterparty)])  # noqa: E501
+                    results.post_decoding_rules += self.decoders[class_name].post_decoding_rules()  # noqa: E501
```
LefterisJP, line 199: Okay so here I don't understand what you are doing. You need to add some comments.

1. addresses_to_decoders is not the decoders. A decoder is the class. Name the variable accordingly. If you meana single decoding function, that is a decoding rule.
2. Why `first_counterparty`? What is the significance? A decoder can have multiple counterparties. I don't understand what you are doing here.
3. Wow I really don't get what you are doing here in those lines. Are you trying to map counterparties to post-decoding rules? Why not just dothis as part of each module's `post_decoding_rules()` function? IT's the responsiility of the module to give you the map. Not trying to find it out post-facto by some obscure logic.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -244,35 +256,45 @@ def decode_by_address_rules(
...
     def run_all_post_decoding_rules(
             self,
             transaction: EvmTransaction,
             decoded_events: list[HistoryBaseEntry],
             all_logs: list[EvmTxReceiptLog],
+            counterparties: set[str],
     ) -> list[HistoryBaseEntry]:
         """
-        Runs all post-decoding rules from self.rules.post_decoding_rules.
+        Runs only the required post decoding rules for this transaction
```
LefterisJP, line 285: Keep as is. This is used also in the other lines below

```suggestion
        Runs only the required post-decoding rules for this transaction
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -244,35 +256,45 @@ def decode_by_address_rules(
...
         The post-decoding rules list consists of tuples (priority, rule) and must be sorted by
         priority in ascending order. The higher the priority number the later the rule is run.
         """
-        for (_, rule) in self.rules.post_decoding_rules:
-            decoded_events = rule(transaction=transaction, decoded_events=decoded_events, all_logs=all_logs)  # noqa: E501
+        # first include the rules that need to be executed always
+        rules = self.rules.post_decoding_rules.common
```
LefterisJP, line 290: there should never be such rules. The entire point of this change is to avoid their existence.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -654,13 +685,13 @@ def _maybe_decode_erc20_721_transfer(
...
         if transfer is None:
-            return None, []
+            return None, [], None
 
         for idx, action_item in enumerate(action_items):
             if action_item.asset == found_token and action_item.amount == transfer.balance.amount and action_item.from_event_type == transfer.event_type and action_item.from_event_subtype == transfer.event_subtype:  # noqa: E501
                 if action_item.action == 'skip':
                     action_items.pop(idx)
-                    return None, []
+                    return None, [], None
```
LefterisJP, line 694: General comment. This is why I did not want you to do this change like this. You said it's not many places when I told you that you would need to add counterparty everywhere. Yet .... here you are. Doing exactly what I todl you I wanted to avoid.

Can't scale decoders like this by having to add a new argument or return argument to all functions involved in the call chain. The code becomes unmaintenable.

The solution here is to have a context class that is passed as argument and as return. Something like `DecodersArgContext` and `DecodersReturnContext`.

You simply change the context class in one place and don't need to re-touch every single call function in the decoder call stack ever again.

rotkehlchen/tests/unit/accounting/evm/test_balancer.py
LefterisJP: You have not answered this @yabirgb

yabirgb: There was nothing to answer. I got note of it. Can add an emoji next time

LefterisJP: 1. Emoji
2. Either make an issue or a sticky note or a pin in discord. Something to remind you, otherwise you **will** forget

rotkehlchen/chain/ethereum/modules/balancer/v1/decoder.py
yabirgb: Yes, in this case last event is not `None` when you are handling an exit and then a deposit. Adding of `event` to `related_events` is needed because you are handling a different part of the transaction, the deposit.

So the if statement makes sure that we finish any previous handling and the the append is starting the new one

rotkehlchen/chain/ethereum/modules/balancer/v1/decoder.py
yabirgb: As we discussed in discord I'll revert the changes and update in a follow up

LefterisJP (approved):
LGTM. Can't say I get all the logic 100% but now it's fine. 

You will need to add a changelog entry though and this can be done in the other PR.

Tests that failed seem to all be coingecko timeout/connection problems

PR #5726, opened 2023-03-03
Refactor HistoryBaseEntry & create a new class dedicated for evm events

### LINKED ISSUE #5689: Refactor HistoryBaseEntry
At the moment the HistoryBaseEntry contain special fields that are used only in EVM events. https://github.com/rotki/rotki/blob/d4bd4d3ded2583761e990a02a92ac5327007520b/rotkehlchen/accounting/structures/base.py#L89 . This is the case of the counterparty field for example.

We want to use this class and create statistics from them and do other complex operations. This would require the introduction of more complex fields to classify the history events but those won't be needed for example in kraken events or other types of base history event which is supposed to be the least common denominator of an event.

While developing statistics for liquity we saw that the current approach for counterparty doesn't allow us to filter by product or contract and this is something that would have made the queries easier. In addition there will be cases for certain protocols where we won't be able to query the needed information with the data we currently have.

The HistoryBaseEntry contains the `extra_data` field but since it is json if we want fast SQL queries we would need a text search and this is slower than doing a filter in a column.

## Proposed solution

We should create a new table that could be used with `JOIN` statements and would allow to consume extra information for EVM transactions while avoiding unused fields in other kind of events like exchange events.

Currently the evm transactions use location `ETHEREUM` and `OPTIMISM` and this can be used as criteria to apply the join

The columns that can be removed from the history events table are:

- counterparty
- extra_data is filled at the moment only in evm events

The new table let's call it `evm_events_info` should have

- [string] identifier: the same identifier as in history_events so we can join by this column.
- [string] protocol:  used to mark the protocol, for example curve or liquity
- [string] product: for example if we are interacting with a pool, staking contract or others. This will help when filtering the events adding easier granularity to the searches.
- [string] address: This would be the address of the contract. This would help to filter by older versions or limit searches to certain subsets of contracts. For example this would help filtering interactions with curve gauges.
- [string(json)] the extra data column from the history_events

### After this is done

- [ ] After this issue also do #5716

nebolax:
Closes #5689

Currently is in progress since there is a lot to do.

My, I think, biggest question right now is how to organize events filters? I implemented something, but I don't really like it and I don't have anything better in mind right now.

## TODO when this PR gets merged:

- [ ] In @LefterisJP's refactor of evm transactions resource don't forget to change: `/home/nebolax/p/rotki/rotki/rotkehlchen/tests/api/test_ethereum_transactions.py::test_query_transactions_check_decoded_events`, `/home/nebolax/p/rotki/rotki/rotkehlchen/tests/api/test_ethereum_transactions.py::test_events_filter_params`, `/home/nebolax/p/rotki/rotki/rotkehlchen/tests/api/test_ethereum_transactions.py::test_ignored_assets`
- [ ] Add an upgrade step that would drop the decoded events (since they all need redecoding)


rotkehlchen/db/history_events.py
```diff
@@ -31,9 +32,25 @@
...
+
+HISTORY_BASE_ENTRY_SELECT_COUNT = 'SELECT COUNT(*) FROM history_events '
+
+EVM_EVENT_INSERT = """INSERT OR IGNORE INTO evm_events_info(identifier, protocol, product,
+address, extra_data) VALUES (?, ?, ?, ?, ?);"""
+
+EVM_EVENT_SELECT = """SELECT A.identifier, A.event_identifier, A.sequence_index, A.timestamp,
+A.location, A.location_label, A.asset, A.amount, A.usd_value, A.notes, A.type, A.subtype,
+B.protocol, B.product, B.address, B.extra_data
+ FROM history_events as A INNER JOIN evm_events_info as B ON A.identifier=B.identifier """
```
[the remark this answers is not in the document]
yabirgb: What do you think? Lefteris had concerns about performance but based on the swap events that was a complex query that we were using before. Here we are just doing a join. Do you think it will improve the work on filters?

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -31,25 +39,64 @@ def _update_history_events_schema(write_cursor: 'DBCursor', conn: 'DBConnection'
...
+
+    tuples_to_insert = []
+    for (identifier, counterparty, extra_data) in dropped_data:
+        protocol = None
+        address = None
+        if counterparty is not None:
+            if counterparty.startswith('0x'):
+                address = counterparty
+            else:
+                protocol = counterparty
```
yabirgb, line 91: wouldn't it be better to just re-decode everything? Otherwise you are missing information. I would do this for customized events. There we have to think a bit more

yabirgb (commented):
Some comments/questions asking also your opinion

rotkehlchen/db/history_events.py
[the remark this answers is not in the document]
nebolax: A view is a helper sql script in the db, right?

If yes, then hmm... Does this `EVM_EVENT_SELECT` constant and others look badly-readable to you?

rotkehlchen/db/history_events.py
```diff
@@ -31,9 +32,25 @@
...
+
+
+EVM_EVENT_SELECT_COUNT = """SELECT COUNT(A.identifier) FROM history_events as A INNER JOIN
+ evm_events_info as B ON A.identifier=B.identifier """
```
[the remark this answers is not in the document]
nebolax: Hmm yeah, this makes sense...

rotkehlchen/db/upgrades/v36_v37.py
nebolax: Yes, is true that we lose information in the current implementation.

My thought here was that with redecoding we will only populate `address` field, which is currently needed for decoding purposes and doesn't make any change to the UI / accounting / etc.

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,128 @@
...
+            'extra_data': self.extra_data,
+        }
+
+    def serialize_without_extra_data(self) -> dict[str, Any]:
+        result = self.serialize()
+        result.pop('extra_data')
+        return result
+
+    @classmethod
+    def deserialize_evm_event_from_db(cls, entry: EVM_EVENT_DB_TUPLE_READ) -> 'EvmEvent':
```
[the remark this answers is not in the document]
nebolax: The problem is that if we use `deserialize_from_db` as a name here, it will break inheritance rules (the method in `BaseHistoryEntry` and this one would have different input arguments types).

rotkehlchen/db/history_events.py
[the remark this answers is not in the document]
yabirgb: Yes. I propose this thinking in the filters and how you would need to use them later, to avoid repeating all the time the join as this operation is not consuming and can go in a view. If you think is not such a big deal we can just go with joining everywhere

yabirgb (commented):
Thanks. Left many comments and things that should be different also some missing things.

Regarding the tests I fixed the decoders but there are some for aave and compound that were failing and I believe is not related to the missing address field. Can you also check those. Also there are some issues with the mypy linter

rotkehlchen/accounting/structures/base.py
```diff
@@ -331,7 +281,7 @@ def process(
             )
             return 1
         # else it's a decoded transaction event and should be processed there
-        return accounting.transactions.process(self, events_iterator)
+        return accounting.transactions.process(self, events_iterator)  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
nebolax: `self` here is a `HistoryBaseEntry` while `transacions.process` expects an `EvmEvent`.

Can't use `cast` here since `cast` requires the type to be provided. And we can't import `EvmEvent` here because we will get recursive imports.

rotkehlchen/accounting/structures/base.py
```diff
@@ -68,21 +65,17 @@
...
 
-ALL_DETAILS_KEYS = {
-    SUB_SWAPS_DETAILS,
-    LIQUITY_STAKING_DETAILS,
-}
+def determine_event_type(
+        location: Location,
+        event_identifier: bytes,
+) -> Literal['base_entry', 'evm_event']:
+    if location == Location.KRAKEN or event_identifier.startswith(b'rotki_events'):
```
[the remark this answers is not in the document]
nebolax: Hmmm if I change this condition to check whether location is `Location.ETHEREUM` or `Location.OPTIMISM`, do you think it will be okay?

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,147 @@
...
+    LIQUITY_STAKING_DETAILS,
+}
+
+
+def get_tx_event_type_identifier(event_type: HistoryEventType, event_subtype: HistoryEventSubType, protocol: str) -> str:  # noqa: E501
+    return str(event_type) + '__' + str(event_subtype) + '__' + protocol
+
+
+@dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=False)
+class EvmEvent(HistoryBaseEntry):
```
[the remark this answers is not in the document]
nebolax: Hmmm makes sense, but if I understand correctly the only thing that we lose is the ability to have mandatory fields in `EvmEvent`.

Right now all the fields are optional there, so I would suggest keeping it as dataclasses for now and change later if will need it. What do you think?

rotkehlchen/accounting/structures/base.py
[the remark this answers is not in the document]
yabirgb: those didn't exist when that code was written but I believe is the best thing to split the logic here

rotkehlchen/accounting/structures/evm_event.py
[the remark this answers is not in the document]
yabirgb: if you make a class you make the fields mandatory just by not setting a default value in the init. As I mentioned the address shouldn't be optional if I'm not missing something

rotkehlchen/accounting/structures/base.py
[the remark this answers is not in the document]
yabirgb: okey but add it as a comment to the code. Is what we do always that we ignore a type

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,147 @@
...
+    protocol: Optional[str] = None
+
+    # For example if we are interacting with a pool, staking contract or others.
+    # This will help when filtering the events adding easier granularity to the searches.
+    product: Optional[str] = None
+
+    # If we are working with evm information this would be the address of the contract.
+    # This would help to filter by older versions or limit searches to certain subsets of
+    # contracts. For example this would help filtering interactions with curve gauges.
+    address: Optional[ChecksumEvmAddress] = None
```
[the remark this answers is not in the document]
nebolax: Sorry, by

> Here it is always EVM

do you mean to rename the variable to `evm_address` ?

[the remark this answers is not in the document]
nebolax: As for `None` -- yes, probably address is not nullable here.

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,147 @@
...
+    Optional[str],  # protocol
+    Optional[str],  # product
+    Optional[str],  # address
+    Optional[str],  # extra_data
+]
+
+EVM_EVENT_FIELDS_COUNT = len(typing.get_args(EVM_EVENT_FIELDS))
+
+
+EVM_EVENT_DB_TUPLE_READ = tuple[
```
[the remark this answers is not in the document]
nebolax: Yeah, I also tried, but couldn't find a way to do this. However this can probably be achieved if instead of making `EVM_EVENT_DB_TUPLE_READ` a long tuple I make it a tuple of 2 subtuples -- one for history base entry and one for evem event.

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,147 @@
...
+        result = self.serialize()
+        result.pop('extra_data')
+        return result
+
+    @classmethod
+    def deserialize_evm_event_from_db(cls, entry: EVM_EVENT_DB_TUPLE_READ) -> 'EvmEvent':
+        history_base_entry_data: HISTORY_EVENT_DB_TUPLE_READ = entry[:-EVM_EVENT_FIELDS_COUNT]  # type: ignore[assignment]  # noqa: E501
+        evm_event_data: EVM_EVENT_FIELDS = entry[-EVM_EVENT_FIELDS_COUNT:]  # type: ignore[assignment]  # noqa: E501
+        instance: 'EvmEvent' = super().deserialize_from_db(history_base_entry_data)  # type: ignore[assignment]  # noqa: E501
+        instance.protocol = evm_event_data[0]
```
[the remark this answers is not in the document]
nebolax: Yes, cast is probably cleaner.

Note though: from the type checking perspective `cast` does not have any benefits. Same as with `type: ignore` it just silently believes the developer.

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,147 @@
...
+    str,            # subtype
+    Optional[str],  # protocol
+    Optional[str],  # product
+    Optional[str],  # address
+    Optional[str],  # extra_data
+]
+
+
+SUB_SWAPS_DETAILS = 'sub_swaps'
+LIQUITY_STAKING_DETAILS = 'liquity_staking'
```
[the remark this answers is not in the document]
nebolax: okay

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,147 @@
...
+            return None
+
+        details = {k: v for k, v in self.extra_data.items() if k in ALL_DETAILS_KEYS}
+        return details if len(details) > 0 else None
+
+    @classmethod
+    def deserialize_evm_event(cls, data: dict[str, Any]) -> 'EvmEvent':
+        instance: 'EvmEvent' = super().deserialize(data)  # type: ignore[assignment]
+        instance.protocol = deserialize_optional(data['protocol'], str)
+        instance.product = deserialize_optional(data['product'], str)
```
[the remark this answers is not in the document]
nebolax: Yeah, sure, makes sense, I will changed it. But same comment as before -- it is probably cleaner, but from the type-checking perspective doesn't have any benefits.

rotkehlchen/accounting/transactions.py
```diff
@@ -3,7 +3,7 @@
 from typing import TYPE_CHECKING
 
 from rotkehlchen.accounting.mixins.event import AccountingEventMixin, AccountingEventType
-from rotkehlchen.accounting.structures.base import HistoryBaseEntry
+from rotkehlchen.accounting.structures.evm_event import EvmEvent
```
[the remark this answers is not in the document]
nebolax: There is also an `isinstnace` check, which requires a normal import

rotkehlchen/api/rest.py
```diff
@@ -36,7 +36,8 @@
 from rotkehlchen.accounting.debugimporter.json import DebugHistoryImporter
 from rotkehlchen.accounting.ledger_actions import LedgerAction
 from rotkehlchen.accounting.structures.balance import Balance, BalanceType
-from rotkehlchen.accounting.structures.base import HistoryBaseEntry, StakingEvent
+from rotkehlchen.accounting.structures.base import StakingEvent
+from rotkehlchen.accounting.structures.evm_event import EvmEvent
```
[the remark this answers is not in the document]
nebolax: Okay

rotkehlchen/api/rest.py
```diff
@@ -4111,11 +4113,11 @@ def get_custom_asset_types(self) -> Response:
 
     def get_event_details(self, identifier: int) -> Response:
         dbevents = DBHistoryEvents(self.rotkehlchen.data.db)
-        event = dbevents.get_history_event_by_identifier(identifier=identifier)
+        event = dbevents.get_evm_event_by_identifier(identifier=identifier)
         if event is None:
             return api_response(wrap_in_fail_result('No event found'), status_code=HTTPStatus.NOT_FOUND)  # noqa: E501
 
-        details = event.get_details()
+        details = event.get_details()  # pylint: disable=no-member
```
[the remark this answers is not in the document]
nebolax: Added a comment, but this actually a place about which I don't know why does pylint complain :D

rotkehlchen/chain/ethereum/modules/convex/decoder.py
```diff
@@ -1,11 +1,8 @@
 import logging
 from typing import Any, Callable, Optional
 
-from rotkehlchen.accounting.structures.base import (
-    HistoryBaseEntry,
-    HistoryEventSubType,
-    HistoryEventType,
-)
+from rotkehlchen.accounting.structures.base import HistoryEventSubType, HistoryEventType
+from rotkehlchen.accounting.structures.evm_event import EvmEvent
```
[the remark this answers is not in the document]
nebolax: I get it. But why though? Imo moving an import to type checking is needed only when there is a risk of a circular import. Or am I wrong?

rotkehlchen/db/filtering.py
```diff
@@ -980,7 +980,7 @@ def make(
                     values=event_identifiers,
                 ),
             )
-        if protocols is not None:
+        if protocols is not None and cls == EvmEventFilterQuery:
             filters.append(
                 DBProtocolsFilter(and_op=True, protocols=protocols),
             )
```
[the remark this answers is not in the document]
nebolax: Yes, perhaps calling super's make is better. What made me confused about this all filtering is that if we call `HistoryEventFilterQuery`'s make, it would require `EvmEventFilterQuery`'s make to accept the entrie list of all the `HistoryEventFilterQuery` make's arguments to then just pass them along.

-----------

However, choosing from two, I agree with you and think that the current approach is way more ugly. Will change.

rotkehlchen/db/history_events.py
```diff
@@ -89,13 +109,16 @@ def add_history_events(
             )
 
     def edit_history_event(self, event: HistoryBaseEntry) -> tuple[bool, str]:
-        """Edit a history entry to the DB. Returns the edited entry"""
+        """Edit a history entry in the DB.
```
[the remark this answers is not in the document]
nebolax: There is `ON DELETE CASCADE` property in sqlite, so deletion will be handled automatically.

rotkehlchen/db/history_events.py
```diff
@@ -31,9 +32,22 @@
...
+ sequence_index, timestamp, location, location_label, asset, amount, usd_value, notes,
+ type, subtype) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?);"""
+
+EVM_EVENT_INSERT = """INSERT OR IGNORE INTO evm_events_info(identifier, protocol, product,
+address, extra_data) VALUES (?, ?, ?, ?, ?);"""
+
+HISTORY_BASE_ENTRY_FIELDS = """history_events.identifier, history_events.event_identifier,
+ history_events.sequence_index, history_events.timestamp, history_events.location,
+ history_events.location_label, history_events.asset, history_events.amount,
+ history_events.usd_value, history_events.notes, history_events.type, history_events.subtype"""
```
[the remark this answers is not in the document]
nebolax: I added the table name here just to be more explicit. In case we will add a column with one of these names to another table later. To don't get ambiguous columns.

But perhaps for now the table name can be removed.

-----

As for the small tab (it is a single whitespace) -- it is to have a uniform format of the query. After commas we have whitespaces, but when there is a new line the whitespace needs to be at the beginning of the new line (can't be at the end of the previous line because then linters complain due to a trailing whitespace).

rotkehlchen/tests/api/test_ethereum_transactions.py
```diff
@@ -1008,6 +1008,7 @@ def mocked_request_dict(url, *_args, **_kwargs):
         assert result['entries_total'] == 2
 
 
+@pytest.mark.skip()  # TODO: unskip after Lefteris' refactor of this endpoint
```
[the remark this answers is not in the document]
nebolax: Okay. Added to the PR's description.

rotkehlchen/accounting/structures/evm_event.py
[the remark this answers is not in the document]
yabirgb: the difference is that with type ignore you completly ignore the type from where it used in places where this code is executed

rotkehlchen/accounting/structures/evm_event.py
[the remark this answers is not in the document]
yabirgb: no, I meant that it is always an evm transaction, is not like in history events where this field could be empty. In evm transactions you always have an address you are interacting with

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,147 @@
...
+    Optional[str],  # product
+    Optional[str],  # address
+    Optional[str],  # extra_data
+]
+
+EVM_EVENT_FIELDS_COUNT = len(typing.get_args(EVM_EVENT_FIELDS))
+
+
+EVM_EVENT_DB_TUPLE_READ = tuple[
+    int,            # identifier
```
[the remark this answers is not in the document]
nebolax: That PR changed only the tuple for writing, no?

This tuple is for reading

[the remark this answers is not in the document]
yabirgb: yes the onw fro writting. True this one is for reading, but letting you know in case that when rebasing it gets lost or needs adjustment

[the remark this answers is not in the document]
nebolax: Yep, alright, thanks

rotkehlchen/db/upgrades/v36_v37.py
nebolax: Okay

rotkehlchen/api/rest.py
```diff
@@ -4038,11 +4040,11 @@ def get_custom_asset_types(self) -> Response:
 
     def get_event_details(self, identifier: int) -> Response:
         dbevents = DBHistoryEvents(self.rotkehlchen.data.db)
-        event = dbevents.get_history_event_by_identifier(identifier=identifier)
+        event = dbevents.get_evm_event_by_identifier(identifier=identifier)
         if event is None:
             return api_response(wrap_in_fail_result('No event found'), status_code=HTTPStatus.NOT_FOUND)  # noqa: E501
 
-        details = event.get_details()
+        details = event.get_details()  # pylint: disable=no-member  # not sure why pylint doesn't see `get_details` here. The type is `EvmEvent`.  # noqa: E501
```
yabirgb, line 4047: I've run the linter and seems that this is not needed. I installed the deps and removed this comment and linter passed

rotkehlchen/chain/ethereum/modules/aave/v2/decoder.py
```diff
@@ -49,14 +49,15 @@ def _decode_collateral_events(
             balance=Balance(),
             location_label=user,
             notes=f'{"Enable" if tx_log.topics[0] == ENABLE_COLLATERAL else "Disable"} {token.symbol} as collateral on AAVE v2',  # noqa: E501
-            counterparty=CPT_AAVE_V2,
+            protocol=CPT_AAVE_V2,
+            address=cast(ChecksumEvmAddress, transaction.to_address),  # transaction.to_address is not None here  # noqa: E501
```
yabirgb, line 53: what is the case where this is None? Contracts deployments is one irc but not sure what other can have it

rotkehlchen/chain/ethereum/modules/ens/decoder.py
```diff
@@ -202,18 +202,19 @@ def _decode_ens_registry_with_fallback_event(
                 balance=Balance(),
                 location_label=transaction.from_address,
                 notes=notes,
-                counterparty=CPT_ENS,
+                protocol=CPT_ENS,
+                address=cast(ChecksumEvmAddress, transaction.to_address),  # transaction.to_address is not None here  # noqa: E501
```
yabirgb, line 206: Is ugly that you had to do this in so many places but right now I can't think of a better solution :thinking:

rotkehlchen/chain/ethereum/modules/makerdao/decoder.py
```diff
@@ -357,17 +357,18 @@ def _decode_vault_creation(
             asset=A_ETH,
             balance=Balance(),
             notes=notes,
-            counterparty='makerdao vault',
+            protocol=CPT_VAULT,
+            address=cast(ChecksumEvmAddress, transaction.to_address),  # transaction.to_address is not None here  # noqa: E501
```
yabirgb, line 361: btw why was this introduced? i don't recall it before

rotkehlchen/db/history_events.py
```diff
@@ -108,6 +126,12 @@ def edit_history_event(self, event: HistoryBaseEntry) -> tuple[bool, str]:
                 )
                 return False, msg
 
+            if isinstance(event, EvmEvent):
+                cursor.execute(
+                    'UPDATE evm_events_info SET protocol=?, product=?, address=? WHERE identifier=?',    # noqa: E501
+                    (*event.serialize_evm_event_for_db_without_extra_data(), event.identifier),
+                )
```
yabirgb, line 133: shouldn't this catch integrity error as we do when updating `history_events` ?

rotkehlchen/db/history_events.py
```diff
@@ -227,17 +254,27 @@ def get_history_events(
...
+            bindings.insert(0, FREE_HISTORY_EVENTS_LIMIT)
+
+        cursor.execute(base_query + query, bindings)
         output = []
         for entry in cursor:
             try:
-                deserialized = HistoryBaseEntry.deserialize_from_db(entry)
+                # Deserialize event depending on its type
+                deserialized: Union[HistoryBaseEntry, EvmEvent]
+                if determine_event_type(Location.deserialize_from_db(entry[4])) == 'evm_event':
```
yabirgb, line 274: I believe you had a discussion already with lefteris about the usage of constants like this but don't know what was the output of that.

Here I believe that is better to use enums for different reasons:

1. The constants are used in different places in the code and they are not imported but written as string.
2. We  will introduce more probably in the future and this complicates things
3. Is easier to navigate and search docs if this is a enum instead of a string constant

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -42,25 +42,46 @@ def _update_history_events_schema(write_cursor: 'DBCursor', conn: 'DBConnection'
 
         for entry in read_cursor:
             if entry[11] is None:
-                new_entries.append(entry[:11] + ('none',) + entry[12:])
+                new_entries.append(entry[:11] + ('none',))
             else:
-                new_entries.append(entry)
+                new_entries.append(entry[:12])
```
yabirgb, line 47: this [:12] looks prone to error. Aren't you using a length constant somewhere else already?

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -42,25 +42,46 @@ def _update_history_events_schema(write_cursor: 'DBCursor', conn: 'DBConnection'
 
         for entry in read_cursor:
             if entry[11] is None:
-                new_entries.append(entry[:11] + ('none',) + entry[12:])
+                new_entries.append(entry[:11] + ('none',))
```
yabirgb, line 45: Could you write in a comment what this is doing? I came here and I have 0 context about what was happening here and I still don't know 100%

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -42,25 +42,46 @@ def _update_history_events_schema(write_cursor: 'DBCursor', conn: 'DBConnection'
...
             else:
-                new_entries.append(entry)
+                new_entries.append(entry[:12])
 
-    write_cursor.executemany('INSERT INTO history_events_copy VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ? ,? ,? ,?)', new_entries)  # noqa: E501
+    write_cursor.executemany('INSERT INTO history_events_copy VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)', new_entries)  # noqa: E501
```
yabirgb, line 49: As I said before I would copy only the non evm events since they would need to be redecoded (the evm ones) that means modifying other tables too. If you want I believe is fair to leave that as a TODO for a smaller PR

yabirgb (commented):
Some things need change and tests didn't run. Please fix them and make sure that everything runs okey

rotkehlchen/chain/ethereum/modules/aave/v2/decoder.py
nebolax: No other case afaik. Only contracts deployment.

rotkehlchen/chain/ethereum/modules/ens/decoder.py
nebolax: Me too

rotkehlchen/db/upgrades/v36_v37.py
nebolax: Yes, forgot about this thing. Makes sense

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def get_tx_event_type_identifier(event_type: HistoryEventType, event_subtype: HistoryEventSubType, protocol: str) -> str:  # noqa: E501
+    return str(event_type) + '__' + str(event_subtype) + '__' + protocol
+
+
+class EvmEvent(HistoryBaseEntry):
+    def __init__(
+            self,
+            *,  # enforce kwargs for this constructor
```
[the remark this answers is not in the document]
nebolax: If to allow args -- then how to sort the arguments in the constructor? 

Since `address` is always required it should go before optional fields like `location_label`, `identifier`, etc., which belong to `HistoryBaseEntry`.

And then I think it will be quite easy to make a mistake by using a wrong order of arguments. 

We don't use args for such construstors anyway -- so do you still think that enforcing kwargs is a bad idea here?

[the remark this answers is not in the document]
yabirgb: I believe we never use the arguments without kwargs but I believe is fine to reorder them, they are different classes and if you use kwargs the order doesn't matter, the position argument is used if no kwargs are provided afaik

rotkehlchen/db/history_events.py
nebolax: Okay, will change to an enum.

rotkehlchen/db/history_events.py
nebolax: No since there are no constraints on the values (like unique, primary key, etc.) touched by this statement.

rotkehlchen/db/upgrades/v36_v37.py
nebolax: There is a docstring, but sure, I'll add a few comments

rotkehlchen/db/upgrades/v36_v37.py
nebolax: Using a length constant is less safe imo since the constant will be changing over time when we add / remove fields from history base entry. 12 is the number of needed fields at the point when the upgrade is happening.

yabirgb: Pleases write is a contant with a comment

nebolax: Ok

rotkehlchen/accounting/structures/base.py
```diff
@@ -353,7 +278,7 @@ def process(
             )
             return 1
         # else it's a decoded transaction event and should be processed there
-        return accounting.transactions.process(self, events_iterator)
+        return accounting.transactions.process(self, events_iterator)  # type: ignore[arg-type]  # self has type HistoryBaseEntry here, while transactions.process expects EvmEvent  # noqa: E501
```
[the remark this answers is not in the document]
nebolax: Yes, makes sense, thanks

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -468,7 +468,8 @@ def _maybe_decode_internal_transactions(
             if amount == ZERO:
                 continue
 
-            event_type, location_label, counterparty, verb = direction_result
+            event_type, location_label, address, protocol, verb = direction_result
+            counterparty = protocol or address  # protocol if found, otherwise address
```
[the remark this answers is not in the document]
yabirgb: and also, why duplicate information here? if counterparty is none you don't need to set its value to address, we already have that info in address

yabirgb (commented):
Some things still need to change. Also rebase the cassettes branch and don't forget to rebase this PR too. It is easier since you will have conflicts unfortunately

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -424,9 +424,9 @@ def get_or_decode_transaction_events(
                     (transaction.tx_hash, serialized_chain_id, HISTORY_MAPPING_STATE_DECODED),
                 )
                 if cursor.fetchone()[0] != 0:  # already decoded and in the DB
-                    events = self.dbevents.get_history_events(
+                    events: list[EvmEvent] = self.dbevents.get_history_events(  # type: ignore[assignment]  # noqa: E501
                         cursor=cursor,
-                        filter_query=HistoryEventFilterQuery.make(
+                        filter_query=EvmEventFilterQuery.make(
```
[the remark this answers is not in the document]
nebolax: No. I also tried to make an overload, but didn't manage to.

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,266 @@
...
+        return accounting.transactions.process(self, events_iterator)
+
+    def __eq__(self, other: Any) -> bool:
+        """Performs a comparison.
+
+        If `other` is a history base entry but not an evm event then compares only fields of
+        history base entry. Otherwise compates evm event fields too.
+        """
+        if isinstance(other, HistoryBaseEntry) is False:
+            return False  # Can't comapre with a non-history-base-entry
```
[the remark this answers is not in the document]
nebolax: Hmm probably we don't compare it like that indeed

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,266 @@
...
+        details = {k: v for k, v in self.extra_data.items() if k in ALL_DETAILS_KEYS}
+        return details if len(details) > 0 else None
+
+    @classmethod
+    def deserialize_evm_event(cls, data: dict[str, Any]) -> 'EvmEvent':
+        instance = super().deserialize(data)
+        # since we call super the type for the linter is HistoryBaseEntry, but in fact is EvmEvent
+        instance = cast(EvmEvent, instance)
+        if 'address' not in data or data['address'] is None:
+            raise InputError('When deserializing EvmEvent address must be given')
```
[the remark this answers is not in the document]
nebolax: Yes, forgot to change

rotkehlchen/accounting/structures/base.py
```diff
@@ -119,26 +104,9 @@ class HistoryBaseEntry(AccountingEventMixin):
     # currently we use to identify the exchange name assigned by the user.
     location_label: Optional[str] = None
     notes: Optional[str] = None
-    # identifier for counterparty.
-    # For a send it's the target
```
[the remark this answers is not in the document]
nebolax: Hmm... does `HistoryBaseEntry` need to be a normal class? What is the problem with keeping it a dataclass?

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1269,8 +1269,6 @@ def test_upgrade_db_35_to_36(user_data_dir):  # pylint: disable=unused-argument
         ('0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12', 'ETH', 'lefteris GTC'),
         ('0xc37b40ABdB939635068d3c5f13E7faF686F03B65', None, 'yabir everywhere'),
     ]
-    cursor.execute('SELECT extra_data FROM history_events WHERE counterparty="liquity"')
-    res = cursor.fetchall()
```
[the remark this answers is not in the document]
nebolax: Ah, I deleted it because this was a useless query. `res` was not compared with anything.

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -216,7 +217,7 @@ def test_add_edit_delete_entries(rotkehlchen_api_server: 'APIServer'):
     # check that the extra data information hasn't been overwritten
     with db.db.conn.read_ctx() as cursor:
         cursor.execute(
-            'SELECT extra_data FROM history_events WHERE identifier=?',
+            'SELECT extra_data FROM evm_events_info WHERE identifier=?',
```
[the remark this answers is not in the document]
nebolax: :thinking: `extra_data` is now stored in `evm_events_info`, not in `history_events`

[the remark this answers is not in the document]
yabirgb: True, my fault. Was late ans read a count instead of a select

rotkehlchen/tests/db/test_db_upgrades.py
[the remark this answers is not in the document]
yabirgb: Don't just delete it. Check what needs to be checked instead. If it was not needed is fine but if someone forgot to check something please do

rotkehlchen/accounting/structures/base.py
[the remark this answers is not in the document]
yabirgb: The problem I see is that the dataclass has a special behavior and it is aware of its field. The problem is in the subclass  where you make modifications breaking how the dataclass is built and I fear that this introduces bugs. Plus what property are you using from the dataclass? Only the str and the eq that I recall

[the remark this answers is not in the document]
nebolax: Okay, will change

rotkehlchen/tests/db/test_db_upgrades.py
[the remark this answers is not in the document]
nebolax: Not sure whether there should be a check because the idea of this part of the upgrade is to remove the events for redecoding.

However, just in case, I added a check that `res` is an empty list here.

rotkehlchen/chain/evm/decoding/decoder.py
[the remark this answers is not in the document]
nebolax: > and also, why duplicate information here? if counterparty is none you don't need to set its value to address, we already have that info in address

It became a duplicate after renaming `protocol` -> `counterparty`.

Previously it was `counterparty = protocol or address` and this `counterparty` variable was used only in notes.

rotkehlchen/chain/ethereum/modules/compound/decoder.py
```diff
@@ -252,13 +252,12 @@ def decode_comp_claim(
...
         # decoded evens iteration only at the end but can't think of a good way to avoid
         # the possibility of checking all such events
-
         supplier_address = hex_or_bytes_to_address(tx_log.topics[2])
         if not self.base.is_tracked(supplier_address):
             return None, []
 
         for event in decoded_events:
-            if event.event_type == HistoryEventType.RECEIVE and event.location_label == supplier_address and event.asset == A_COMP and event.counterparty == COMPTROLLER_PROXY_ADDRESS:  # noqa: E501
+            if event.event_type == HistoryEventType.RECEIVE and event.event_subtype == HistoryEventSubType.NONE and event.location_label == supplier_address and event.asset == A_COMP and event.address == COMPTROLLER_PROXY_ADDRESS:  # noqa: E501
```
yabirgb, line 260: why adding this subtype check here?

rotkehlchen/chain/ethereum/modules/uniswap/utils.py
```diff
@@ -309,7 +309,7 @@ def decode_basic_uniswap_info(
             # 2.4 SWAP_SIGNATURE event
             # etc.
             # So if we are at SWAP_SIGNATURE № 2 then all events that are before SWAP_SIGNATURE № 1
-            # should have already been decoded, have counterparty set and have type Trade.
+            # should have already been decoded, have protocol set and have type Trade.
```
yabirgb, line 312: This is my fault, should be counterparty here

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -696,7 +700,7 @@ def _maybe_decode_erc20_721_transfer(
                 action_items.pop(idx)
                 break  # found an action item and acted on it
 
-        # Add additional information to transfers for different protocols
+        # Add additional information to transfers for different counterpartys
```
yabirgb, line 703: This is also my fault + typo

rotkehlchen/data_migrations/migrations/migration_9.py
```diff
@@ -24,7 +24,21 @@ def _reset_curve_decoded_events(write_cursor: 'DBCursor') -> None:
...
     """
-    write_cursor.execute('SELECT DISTINCT tx_hash from evm_transactions LEFT JOIN history_events ON history_events.event_identifier=tx_hash WHERE counterparty IS ?', (CPT_CURVE,))  # noqa: E501
+    db_version = write_cursor.execute('SELECT value FROM settings WHERE name="version"').fetchone()
+
+    query = (
+        'SELECT DISTINCT tx_hash from evm_transactions LEFT JOIN history_events '
+        'ON history_events.event_identifier=tx_hash '
+    )
+    if db_version is None or int(db_version[0]) < 37:
+        query += 'WHERE counterparty IS ?'
```
yabirgb, line 34: Data migrations happen after db upgrade so I'm not sure why this distinction here. You should assume that the schema is the latest one

rotkehlchen/tests/exchanges/test_coinbasepro.py
```diff
@@ -69,7 +69,7 @@
 "profile_id": "75da88c5-05bf-4f54-bc85-5c775bd68254"
 }},{ETH_ACCOUNTS}]"""
 
-INVALID_ACCOUNTS_RESPONSE = '[{"id": "71452118-efc7-4cc4-8780-a5e22d4baa53",,,}]'
+INVALID_ACCOUNTS_RESPONSE = '[{"id": "71452118-efc7-4cc4-8780-a5e22d4baa53",,}]'
```
yabirgb, line 72: Why this change?

rotkehlchen/tests/unit/accounting/test_default_settings.py
```diff
@@ -180,7 +180,7 @@ def test_accounting_receive_settings(
     assert accounting_pot.pnls.free == ZERO
 
 
-@pytest.mark.parametrize(('event_type', 'event_subtype', 'is_taxable', 'counterparty', 'gas_taxable', 'include_crypto2crypto'), [  # noqa: E501
+@pytest.mark.parametrize(('event_type', 'event_subtype', 'is_taxable', 'protocol', 'gas_taxable', 'include_crypto2crypto'), [  # noqa: E501
```
yabirgb, line 183: This is also my fault. I believe it makes sense to change back to counterparty

rotkehlchen/tests/unit/accounting/test_settings.py
```diff
@@ -105,7 +106,7 @@ def test_include_gas_costs(accountant, google_service):
             fee=None,
             fee_currency=None,
             link=None,
-        ), HistoryBaseEntry(
+        ), EvmEvent(
```
yabirgb, line 109: this should use the evm event deserialize event identifier

rotkehlchen/tests/utils/history.py
```diff
@@ -817,9 +817,9 @@ def check_result_of_history_creation(
             assert asset_movements[12].category == AssetMovementCategory.WITHDRAWAL
             assert asset_movements[12].asset == A_BTC
 
-        tx_events = [x for x in events if isinstance(x, HistoryBaseEntry) and x.serialized_event_identifier.startswith('0x')]  # noqa: E501
+        tx_events = [x for x in events if isinstance(x, EvmEvent) and x.serialized_event_identifier.startswith('0x')]  # noqa: E501
         gas_in_eth = FVal('14.36963')
-        assert len(tx_events) == 6
+        assert len(tx_events) in {6, 7}  # can be 7 if there is extra unicode-test event provided
```
yabirgb, line 822: Is it random?

rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -32,7 +32,7 @@ def addresses_to_decoders(self) -> dict[ChecksumEvmAddress, tuple[Any, ...]]:
     @abstractmethod
     def counterparties(self) -> list[str]:
         """
-        Subclasses implement this to specify which counterparty values are introduced by the module
+        Subclasses implement this to specify which protocol values are introduced by the module
```
yabirgb, line 35: this is also my fault, can you change it?

yabirgb (commented):
A few things are odd and would need to change

rotkehlchen/chain/ethereum/modules/compound/decoder.py
nebolax: That is the neat part.

There is a `counterparty` check. Previously if there was a match, counterparty (which is initially an address) was overwritten by a protocol label and thus counterparty was not matched in the next calls to this function.

Now, since we don't modify the counterparty anymore, for a particular event this condition will always succeed. And to prevent that I added a subtype check to check whether an event was decoded or not.

rotkehlchen/tests/utils/history.py
nebolax: No, is not random. I thought that it will be hard to provide an argument here and check on a specific value, but it ended up not being easy.

Now changed this. Thanks for pointing out.

rotkehlchen/chain/evm/decoding/base.py
```diff
@@ -141,14 +146,15 @@ def decode_erc20_721_transfer(
             return None
 
         extra_data = None
-        event_type, location_label, counterparty, verb = direction_result
+        event_type, location_label, address, counterparty, verb = direction_result
+        notes_counterparty = counterparty or address  # counterparty if found, otherwise address
```
LefterisJP, line 150: Good naming is better than comments

```suggestion
        counterparty_or_address = counterparty or address
```

rotkehlchen/chain/evm/decoding/decoder.py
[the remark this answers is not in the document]
LefterisJP: Why? Overload should work here as I mentioned in the other part that you added the type ignore. Learn how to use typing and don't ask me how to do it. It's possible as this is what overload is for and we also use it in the rest of the code

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -468,7 +468,8 @@ def _maybe_decode_internal_transactions(
             if amount == ZERO:
                 continue
 
-            event_type, location_label, counterparty, verb = direction_result
+            event_type, location_label, address, counterparty, verb = direction_result
+            notes_counterparty = counterparty or address  # counterparty if found, otherwise address  # noqa: E501
```
LefterisJP, line 472: same as before
```suggestion
            counterparty_or_address = counterparty or address
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -478,18 +479,20 @@ def _maybe_decode_internal_transactions(
...
+    def _get_eth_transfer_event(self, tx: EvmTransaction) -> Optional[EvmEvent]:
         direction_result = self.base.decode_direction(
             from_address=tx.from_address,
             to_address=tx.to_address,
         )
         if direction_result is None:
             return None
-        event_type, location_label, counterparty, verb = direction_result
+        event_type, location_label, address, counterparty, verb = direction_result
+        notes_counterparty = counterparty or address  # counterparty if found, otherwise address
```
LefterisJP, line 495: same as before

```suggestion
        counterparty_or_address = counterparty or address
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -696,7 +700,7 @@ def _maybe_decode_erc20_721_transfer(
                 action_items.pop(idx)
                 break  # found an action item and acted on it
 
-        # Add additional information to transfers for different protocols
+        # Add additional information to transfers for different counterparties
```
LefterisJP, line 703: Why counterparties here? The function is named `_enrich_protocol_transfers`

rotkehlchen/data_migrations/migrations/migration_9.py
```diff
@@ -24,7 +24,12 @@ def _reset_curve_decoded_events(write_cursor: 'DBCursor') -> None:
     The code is taken from `delete_events_by_tx_hash` right before 1.27 release and is
     modified to delete only transactions that have a decoded curve event.
     """
-    write_cursor.execute('SELECT DISTINCT tx_hash from evm_transactions LEFT JOIN history_events ON history_events.event_identifier=tx_hash WHERE counterparty IS ?', (CPT_CURVE,))  # noqa: E501
+    query = (
+        'SELECT DISTINCT tx_hash from evm_transactions LEFT JOIN history_events '
+        'ON history_events.event_identifier=tx_hash INNER JOIN evm_events_info ON '
+        'history_events.identifier=evm_events_info.identifier WHERE counterparty IS ?'
+    )
+    write_cursor.execute(query, (CPT_CURVE,))
```
LefterisJP, line 32: no need to put it in its own variable if you still split it in multiple lines. Just do it directly as before but also splitting in multiple lines from inside `cursor.execute()`

rotkehlchen/db/filtering.py
```diff
@@ -998,15 +993,64 @@ def make(
...
+            )
+
+        return filter_query
+
+
 @dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=False)
-class DBProtocolsFilter(DBFilter):
-    """Protocols should be not empty. Otherwise, it can break the query's logic"""
-    protocols: list[str]
+class DBPCounterpartyFilter(DBFilter):
```
LefterisJP, line 1046: This is a terrible name. What is `P` doing in there? Write a full name, for whatever this filter is supposed to be

rotkehlchen/db/history_events.py
```diff
@@ -31,9 +36,20 @@
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
-HISTORY_INSERT = """INSERT OR IGNORE INTO history_events(event_identifier, sequence_index,
-timestamp, location, location_label, asset, amount, usd_value, notes,
-type, subtype, counterparty, extra_data) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?);"""
+HISTORY_BASE_ENTRY_INSERT = """INSERT OR IGNORE INTO history_events(event_identifier,
```
LefterisJP, line 39: If something is not used in other/multiple places, please don't add it as a variale in the top. At least in this file it's used only once.It's pretty hard to read the code having to jump around the various definitions to see what a cursor.execute() does.

rotkehlchen/db/history_events.py
```diff
@@ -226,18 +257,27 @@ def get_history_events(
...
+            join_query = EVM_EVENT_JOIN
         else:
-            query = 'SELECT * FROM (SELECT * from history_events ORDER BY timestamp DESC, sequence_index ASC LIMIT ?) ' + query  # noqa: E501
-            cursor.execute(query, [FREE_HISTORY_EVENTS_LIMIT] + bindings)
+            join_query = ALL_EVENTS_DATA_JOIN
+
+        base_query = f'SELECT {HISTORY_BASE_ENTRY_FIELDS}, {EVM_EVENT_FIELDS} {join_query}'
+        if has_premium is False:
+            base_query = f'SELECT * FROM ({base_query} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?) '  # noqa: E501
+            bindings.insert(0, FREE_HISTORY_EVENTS_LIMIT)
```
LefterisJP, line 269: no need to assign to `base_query` twice when `has_premium` is False

```suggestion
        if has_premium is True:
            base_query = f'SELECT {HISTORY_BASE_ENTRY_FIELDS}, {EVM_EVENT_FIELDS} {join_query}'
        else:
            base_query = f'SELECT * FROM ({base_query} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?) '  # noqa: E501
            bindings.insert(0, FREE_HISTORY_EVENTS_LIMIT)
```

rotkehlchen/db/schema.py
```diff
@@ -632,13 +632,25 @@
...
     UNIQUE(event_identifier, sequence_index)
 );
 """
 
+
+# Table that extends history_events table and stores data specific to evm events.
+DB_CREATE_EVM_EVENTS_INFO = """
+CREATE TABLE IF NOT EXISTS evm_events_info(
+    identifier INTEGER PRIMARY KEY,
+    counterparty TEXT,
```
LefterisJP, line 645: Shouldn't this be protocol? It's really confusing that these two terms are used interchangably

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -42,25 +45,46 @@ def _update_history_events_schema(write_cursor: 'DBCursor', conn: 'DBConnection'
...
+    write_cursor.execute("""
+        CREATE TABLE IF NOT EXISTS evm_events_info(
+            identifier INTEGER PRIMARY KEY,
+            counterparty TEXT,
+            product TEXT,
+            address TEXT,
+            extra_data TEXT,
+            FOREIGN KEY(identifier) REFERENCES history_events(identifier) ON UPDATE CASCADE ON DELETE CASCADE
+        );
+    """)  # noqa: E501
```
LefterisJP, line 72: Ehm ... @nebolax @yabirgb where is the "and moves there values previously stored in history_events part" ?

rotkehlchen/tests/api/test_ethereum_transactions.py
```diff
@@ -1261,6 +1262,7 @@ def query_transactions(rotki):
         assert dbevents.get_history_events(cursor, HistoryEventFilterQuery.make(), True) == []
 
 
+@pytest.mark.skip()  # Unskip after Leftris' changes to the evm trasnactions endpoint
```
LefterisJP, line 1265: Please in all those tests add a comment fine but don't make typos like "Leftris'"

I need all those greppable. And the field to use is this:
```suggestion
@pytest.mark.skip('LEFTERIS TODO')  # Unskip after Lefteris' changes to the evm transactions endpoint
```

Add this `LEFTERIS TODO` in all

rotkehlchen/tests/data/data_migration_9.db
LefterisJP, line 1: Why add a new file here? What is this? Whenever you add such a binary file please explain in the PR description or somewhere in the PR how it was generated

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1322,9 +1323,9 @@ def test_upgrade_db_36_to_37(user_data_dir):  # pylint: disable=unused-argument
     new_history_events = cursor.execute('SELECT * FROM history_events;').fetchall()
     for idx, entry in enumerate(old_history_events):
         if entry[11] is None:
-            assert entry[0:11] + ('none',) + entry[12:] == new_history_events[idx]
+            assert entry[0:11] + ('none',) == new_history_events[idx]
         else:
-            assert entry == new_history_events[idx]
```
LefterisJP, line 1327: you completely ignored to check that base history events table and evm events info table are properly populated in this test. If you had ... you would have seen that this does not happen

rotkehlchen/tests/unit/decoders/test_weth.py
```diff
@@ -160,21 +156,19 @@ def test_weth_interaction_with_protocols_deposit(database, ethereum_inquirer):
     https://etherscan.io/tx/0xab0dec3785632c567365c48ea1fd1178f0998773136a555912625d2668ef53e9
     """
     tx_hex = '0xab0dec3785632c567365c48ea1fd1178f0998773136a555912625d2668ef53e9'
+    timesatmp = TimestampMS(1666595591000)
```
LefterisJP, line 159:
```suggestion
    timestamp = TimestampMS(1666595591000)
```

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,269 @@
...
+    """This is a class for storing evm events data and it extends HistoryBaseEntry.
+
+    It adds the following fields:
+
+    1. counterparty: Optional[str] -- Used to mark the protocol name, for example curve or liquity.
+
+    2. product: Optional[str] -- For example if we are interacting with a pool, staking contract
+    or others. This will help when filtering the events adding easier granularity to the searches.
+
+    3. Optional[address]: ChecksumEvmAddress -- If we are working with evm information this would
```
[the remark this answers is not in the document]
yabirgb: Yes when there is no contract interaction this is set to None.

For proxy interactions we can see how to handle but in general it depends on us how to trace those. First we need to know that is a proxy contract (and we need something that works for a wide range of cases, I believe it doesn't scale to handle this per decoder) and in that case there would be some interactions with the proxy but most of them will happen with proxy<->pool. In that case my opinion is that we would need to create virtual events. But this is a bigger problem that what to set the value address in a event to. 

Indeed is a problem that we have now

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+class EvmEvent(HistoryBaseEntry):
+    """This is a class for storing evm events data and it extends HistoryBaseEntry.
+
+    It adds the following fields:
+
+    1. counterparty: Optional[str] -- Used to mark the protocol name, for example curve or liquity.
+
+    2. product: Optional[str] -- For example if we are interacting with a pool, staking contract
```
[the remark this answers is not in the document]
yabirgb: We haven't written a set of values to be put here yet. We have just created the struct. The first use case that I have in mind is to label the stability pool vs staking LQTY in the liquity protocol. As for the approach we can go with a enum value per protocol. What do you think?

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,269 @@
...
+                    self.counterparty == other.counterparty,
+                    self.product == other.product,
+                    self.address == other.address,
+                    self.extra_data == other.extra_data,
+                ])
+            )
+        )
+
+    def __repr__(self) -> str:
+        fields = self._history_base_entry_repr_fields() + [
```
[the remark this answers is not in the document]
nebolax: `super`'s repr would return `HistoryBaseEntry(a=...,b=...)`, so we won't be able to easily add more fields and turn it into `EvmEven(a=...,b=...,c=...)`

rotkehlchen/api/v1/schemas.py
```diff
@@ -620,7 +620,9 @@ class HistoryBaseEntrySchema(Schema):
         load_default=HistoryEventSubType.NONE,
         allow_none=True,
     )
-    counterparty = fields.String(required=False)
+    counterparty = fields.String(load_default=None)
```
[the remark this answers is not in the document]
nebolax: Yes, I made it `protocol`, but later, since not only protocols, but also exchange names like `kraken` can be here Yabir said to rename it to counterprty.

rotkehlchen/tests/data/data_migration_9.db
nebolax: Hmm it is not a new file, but an upgraded version of the previous file (with the new db schema)

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,269 @@
...
+        Serialize information for the database excluding the extra data field.
+        This is used when writting information edited by the user
+        """
+        return (
+            self.counterparty,
+            self.product,
+            self.address,
+        )
+
+    def serialize_evm_event_for_db(self) -> EVM_EVENT_FIELDS:
```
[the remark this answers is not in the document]
LefterisJP: It depends which functio calls it @nebolax. Please understand how inheritance works.

`BasehistoryEntry.serialize_for_db()` it's the base stuff.
`EvmEvent.serialize_for_db()` it's the other one

rotkehlchen/api/v1/schemas.py
LefterisJP, line 623: As discussed here: https://github.com/rotki/rotki/pull/5726/files#r1138462066

We need to figure out if this should be counterparty or protocol. Is the only reason it's counterparty since it can also be an exchange name for deposits/withdrawal EVM transactions?

rotkehlchen/chain/ethereum/modules/uniswap/v2/common.py
```diff
@@ -32,14 +32,14 @@
...
         transaction: EvmTransaction,
         counterparty: str,
         database: 'DBHandler',
         ethereum_inquirer: 'EthereumInquirer',
-        notify_user: Callable[[HistoryBaseEntry, str], None],
-) -> tuple[Optional[HistoryBaseEntry], list[ActionItem]]:
-    """Common logic for decoding uniswap v2 like protocols (uniswap and sushiswap atm)
+        notify_user: Callable[['EvmEvent', str], None],
+) -> tuple[Optional['EvmEvent'], list[ActionItem]]:
+    """Common logic for decoding uniswap v2 like counterpartys (uniswap and sushiswap atm)
```
LefterisJP, line 42: Leave it as it was. Even if the variable ends up staying coutnerparty this is for protocols.

```suggestion
    """Common logic for decoding uniswap v2 like protocols(uniswap and sushiswap atm)
```

rotkehlchen/chain/evm/decoding/base.py
```diff
@@ -76,9 +76,14 @@ def decode_direction(
             self,
             from_address: ChecksumEvmAddress,
             to_address: Optional[ChecksumEvmAddress],
-    ) -> Optional[tuple[HistoryEventType, Optional[str], str, str]]:
+    ) -> Optional[tuple[HistoryEventType, Optional[str], ChecksumEvmAddress, str, str]]:
         """Depending on addresses, if they are tracked by the user or not, if they
-        are an exchange address etc. determine the type of event to classify the transfer as"""
+        are an exchange address etc. determine the type of event to classify the transfer as.
+
+        Returns event type, loation label, address, counterparty and verb.
```
LefterisJP, line 83: Forgot to fix this. I mentioned here: https://github.com/rotki/rotki/pull/5726/files#r1138539132

```suggestion
        Returns event type, location label, address, counterparty and verb.
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -437,9 +437,9 @@ def get_or_decode_transaction_events(
                     (transaction.tx_hash, serialized_chain_id, HISTORY_MAPPING_STATE_DECODED),
                 )
                 if cursor.fetchone()[0] != 0:  # already decoded and in the DB
-                    events = self.dbevents.get_history_events(
+                    events: list['EvmEvent'] = self.dbevents.get_history_events(  # type: ignore[assignment]  # noqa: E501
```
LefterisJP, line 440: The type:ignore seems to have still not been addressed here as mentioned in this comment: https://github.com/rotki/rotki/pull/5726#discussion_r1138543311

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -709,7 +713,7 @@ def _maybe_decode_erc20_721_transfer(
                 action_items.pop(idx)
                 break  # found an action item and acted on it
 
-        # Add additional information to transfers for different protocols
+        # Add additional information to transfers for different counterparties
```
LefterisJP, line 716: As I mentioned here: https://github.com/rotki/rotki/pull/5726/files#r1138545263

Use the word protocols, irrespective of the variable name
```suggestion
        # Add additional information to transfers for different protocols
```

rotkehlchen/db/history_events.py
```diff
@@ -200,44 +225,75 @@ def get_customized_event_identifiers(
...
     def get_history_events(
             self,
             cursor: 'DBCursor',
             filter_query: HistoryEventFilterQuery,
             has_premium: bool,
     ) -> list[HistoryBaseEntry]:
+        ...
+
+    @overload
+    def get_history_events(  # type: ignore[misc]  # getting from mypy "Overloaded function signature 2 will never be matched: signature 1's parameter type(s) are the same or broader"  # noqa: E501
```
LefterisJP, line 257: Just write that as a comment in the PR not in the code. I will look at this, I guess since this is what I commented in the other parts too.

rotkehlchen/db/upgrades/v36_v37.py
LefterisJP: Then for the love of god add a comment here. And force a redecoding.

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -31,36 +36,58 @@ def _update_history_events_schema(write_cursor: 'DBCursor', conn: 'DBConnection'
...
```
LefterisJP, line 75: As mentioned here: https://github.com/rotki/rotki/pull/5726/files#r1138584425

The author needs to specify here that the data **are not** carried over and need to be redecoded. Otherwise every other reader will see a bug here.

Also what happens to custom events? Would they not need to be carried over? Custom events **must never** be lost.

Also when/where is the redecoding done? Just let it run periodically? wouldn't you need to also clear all decoded events too?

rotkehlchen/tests/data/data_migration_9.db
LefterisJP, line 1: So the only new thing here is the new schema?

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1322,9 +1322,9 @@ def test_upgrade_db_36_to_37(user_data_dir):  # pylint: disable=unused-argument
...
+            assert entry[:12] == new_history_events[idx]
```
LefterisJP, line 1327: I still think the DB upgrade is broken here.

You either move over all stuff (which is the preferred methodology) or you say you remove everyhing, **except** the custom events, and then expect them to be redecoded.

In both case you need a test.

Case 1: Test that data got migrated correctly. Including custom events
Case 2: Test that data got deleted properly, but that custom events ... and custom evm events have been preserved

rotkehlchen/tests/data/data_migration_9.db
nebolax: Yes

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+class EvmEvent(HistoryBaseEntry):
+    """This is a class for storing evm events data and it extends HistoryBaseEntry.
+
+    It adds the following fields:
+
+    1. counterparty: Optional[str] -- Used to mark the protocol name, for example curve or liquity.
+
+    2. product: Optional[str] -- For example if we are interacting with a pool, staking contract
```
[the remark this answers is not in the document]
nebolax: Yeah, sounds like it makes sense. And then we can easily expose it to frontend via an api endpoint similar to recently done "all chain ids and names" if will be needed.

PR #5736, opened 2023-03-07
decode curve liquidity additions/removal via zap depositer contracts

### LINKED ISSUE #5696: Decode deposits to Curve.fi: (Deposit USDN)
Deposit to curve pools should decode and I guess this is a pool not a gauge (which is still WIP).

Deposit to Curve.fi: Deposit USDN are not decoded.
https://etherscan.io/address/0x094d12e5b541784701FD8d65F11fc0598FBC6332

Request: Decode it.

image

prettyirrelevant:
Closes #5696

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/curve/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/curve/decoder.py
@@ -1,6 +1,6 @@
 import logging
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, Any, Callable, Optional
+from typing import TYPE_CHECKING, Any, Callable, Literal, Optional
 from rotkehlchen.accounting.structures.evm_event import EvmProduct
 
 from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
@@ -36,15 +36,59 @@ if TYPE_CHECKING:
     from rotkehlchen.user_messages import MessagesAggregator
 
 
-ADD_LIQUIDITY = b'B?d\x95\xa0\x8f\xc6RB\\\xf4\xed\r\x1f\x9e7\xe5q\xd9\xb9R\x9b\x1c\x1c#\xcc\xe7\x80\xb2\xe7\xdf\r'  # noqa: E501
-ADD_LIQUIDITY_2_ASSETS = b'&\xf5Z\x85\x08\x1d$\x97N\x85\xc6\xc0\x00E\xd0\xf0E9\x91\xe9Xs\xf5+\xff\r!\xaf@y\xa7h'  # noqa: E501
-ADD_LIQUIDITY_4_ASSETS = b'?\x19\x15w^\x0c\x9a8\xa5z{\xb7\xf1\xf9\x00_Ho\xb9\x04\xe1\xf8J\xa2\x156MVs\x19\xa5\x8d'  # noqa: E501
-REMOVE_LIQUIDITY = b"Z\xd0V\xf2\xe2\x8a\x8c\xec# \x15@k\x846h\xc1\xe3l\xdaY\x81'\xec;\x8cY\xb8\xc7's\xa0"  # noqa: E501
-REMOVE_ONE = b'\x9e\x96\xdd;\x99z*%~\xecM\xf9\xbbn\xafbn m\xf5\xf5C\xbd\x966\x82\xd1C0\x0b\xe3\x10'  # noqa: E501
-REMOVE_LIQUIDITY_3_ASSETS = b'\xa4\x9dL\xf0&V\xae\xbf\x8cw\x1fZ\x85\x85c\x8a*\x15\xeel\x97\xcfr\x05\xd4 \x8e\xd7\xc1\xdf%-'  # noqa: E501
-REMOVE_LIQUIDITY_4_ASSETS = b'\x98x\xca7^\x10o*C\xc3\xb5\x99\xfcbEh\x13\x1cL\x9aK\xa6j\x14V7\x15v;\xe9\xd5\x9d'  # noqa: E501
+ADD_LIQUIDITY_EVENTS = {
+    b'B?d\x95\xa0\x8f\xc6RB\\\xf4\xed\r\x1f\x9e7\xe5q\xd9\xb9R\x9b\x1c\x1c#\xcc\xe7\x80\xb2\xe7\xdf\r',  # ADD_LIQUIDITY  # noqa: E501
+    b'&\xf5Z\x85\x08\x1d$\x97N\x85\xc6\xc0\x00E\xd0\xf0E9\x91\xe9Xs\xf5+\xff\r!\xaf@y\xa7h',  # ADD_LIQUIDITY_2_ASSETS  # noqa: E501
+    b'?\x19\x15w^\x0c\x9a8\xa5z{\xb7\xf1\xf9\x00_Ho\xb9\x04\xe1\xf8J\xa2\x156MVs\x19\xa5\x8d',  # ADD_LIQUIDITY_4_ASSETS  # noqa: E501
+}
 REMOVE_LIQUIDITY_IMBALANCE = b'\xb9d\xb7/s\xf5\xef[\xf0\xfd\xc5Y\xb2\xfa\xb9\xa7\xb1*9\xe4x\x17\xa5G\xf1\xf0\xae\xe4\x7f\xeb\xd6\x02'  # noqa: E501
-CURVE_Y_DEPOSIT = string_to_evm_address('0xbBC81d23Ea2c3ec7e56D39296F0cbB648873a5d3')
+REMOVE_LIQUIDITY_EVENTS = {
+    REMOVE_LIQUIDITY_IMBALANCE,
+    b'|68T\xcc\xf7\x96#A\x1f\x89\x95\xb3b\xbc\xe5\xed\xdf\xf1\x8c\x92~\xdco]\xbb\xb5\xe0X\x19\xa8,',  # REMOVE_LIQUIDITY  # noqa: E501,
+    b'\x9e\x96\xdd;\x99z*%~\xecM\xf9\xbbn\xafbn m\xf5\xf5C\xbd\x966\x82\xd1C0\x0b\xe3\x10',  # REMOVE_ONE  # noqa: E501
+    b"Z\xd0V\xf2\xe2\x8a\x8c\xec# \x15@k\x846h\xc1\xe3l\xdaY\x81'\xec;\x8cY\xb8\xc7's\xa0",  # REMOVE_LIQUIDITY  # noqa: E501,
+    b'\xa4\x9dL\xf0&V\xae\xbf\x8cw\x1fZ\x85\x85c\x8a*\x15\xeel\x97\xcfr\x05\xd4 \x8e\xd7\xc1\xdf%-',  # REMOVE_LIQUIDITY_3_ASSETS  # noqa: E501
+    b'\x98x\xca7^\x10o*C\xc3\xb5\x99\xfcbEh\x13\x1cL\x9aK\xa6j\x14V7\x15v;\xe9\xd5\x9d',  # REMOVE_LIQUIDITY_4_ASSETS  # noqa: E501
+}
+# Deposit contracts are retrieved from the links below Deposit<pool>:
+# https://curve.readthedocs.io/ref-addresses.html#base-pools
+# https://curve.readthedocs.io/ref-addresses.html#metapools
+#
+# The duplicates were found via Etherscan:
+# https://etherscan.io/find-similar-contracts?a=0x35796DAc54f144DFBAD1441Ec7C32313A7c29F39
+CURVE_DEPOSIT_CONTRACTS = {
+    string_to_evm_address('0x094d12e5b541784701FD8d65F11fc0598FBC6332'),  # curve usdn
+    string_to_evm_address('0x35796DAc54f144DFBAD1441Ec7C32313A7c29F39'),  # curve usdn duplicate
+    string_to_evm_address('0xB0a0716841F2Fc03fbA72A891B8Bb13584F52F2d'),  # curve ust
+    string_to_evm_address('0x3c8cAee4E09296800f8D29A68Fa3837e2dae4940'),  # curve usdp
+    string_to_evm_address('0xF1f85a74AD6c64315F85af52d3d46bF715236ADc'),  # curve usdk
+    string_to_evm_address('0x6600e98b71dabfD4A8Cac03b302B0189Adb86Afb'),  # curve usdk duplicate
+    string_to_evm_address('0xBE175115BF33E12348ff77CcfEE4726866A0Fbd5'),  # curve rsv
+    string_to_evm_address('0x803A2B40c5a9BB2B86DD630B274Fa2A9202874C2'),  # curve musd
+    string_to_evm_address('0x78CF256256C8089d68Cde634Cf7cDEFb39286470'),  # curve musd duplicate
+    string_to_evm_address('0x1de7f0866e2c4adAC7b457c58Cc25c8688CDa1f2'),  # curve linkusd
+    string_to_evm_address('0xF6bDc2619FFDA72c537Cd9605e0A274Dc48cB1C9'),  # curve linkusd duplicate  # noqa: E501
+    string_to_evm_address('0x09672362833d8f703D5395ef3252D4Bfa51c15ca'),  # curve husd
+    string_to_evm_address('0x0a53FaDa2d943057C47A301D25a4D9b3B8e01e8E'),  # curve husd duplicate
+    string_to_evm_address('0x64448B78561690B70E17CBE8029a3e5c1bB7136e'),  # curve gusd
+    string_to_evm_address('0x0aE274c98c0415C0651AF8cF52b010136E4a0082'),  # curve gusd duplicate
+    string_to_evm_address('0x61E10659fe3aa93d036d099405224E4Ac24996d0'),  # curve dusd
+    string_to_evm_address('0xbBC81d23Ea2c3ec7e56D39296F0cbB648873a5d3'),  # curve y
+    string_to_evm_address('0xb6c057591E073249F2D9D88Ba59a46CFC9B59EdB'),  # curve busd
+    string_to_evm_address('0xeB21209ae4C2c9FF2a86ACA31E123764A3B6Bc06'),  # curve compound
+    string_to_evm_address('0xA50cCc70b6a011CffDdf45057E39679379187287'),  # curve pax
+    string_to_evm_address('0xFCBa3E75865d2d561BE8D220616520c171F12851'),  # curve susd v2
+    string_to_evm_address('0x331aF2E331bd619DefAa5DAc6c038f53FCF9F785'),  # curve zap
+    string_to_evm_address('0xac795D2c97e60DF6a99ff1c814727302fD747a80'),  # curve usdt
+    string_to_evm_address('0xC45b2EEe6e09cA176Ca3bB5f7eEe7C47bF93c756'),  # curve bbtc
+    string_to_evm_address('0xd5BCf53e2C81e1991570f33Fa881c49EEa570C8D'),  # curve obtc
+    string_to_evm_address('0x11F419AdAbbFF8d595E7d5b223eee3863Bb3902C'),  # curve pbtc
+    string_to_evm_address('0xaa82ca713D94bBA7A89CEAB55314F9EfFEdDc78c'),  # curve tbtc
+    string_to_evm_address('0xA79828DF1850E8a3A3064576f380D90aECDD3359'),  # curve 3pool
+    string_to_evm_address('0x08780fb7E580e492c1935bEe4fA5920b94AA95Da'),  # curve fraxusdc
+    string_to_evm_address('0x7abdbaf29929e7f8621b757d2a7c04d78d633834'),  # curve sbtc
+    string_to_evm_address('0xA2d40Edbf76C6C0701BA8899e2d059798eBa628e'),  # curve sbtc2
+}
 
 GAUGE_DEPOSIT = b'\xe1\xff\xfc\xc4\x92=\x04\xb5Y\xf4\xd2\x9a\x8b\xfcl\xda\x04\xeb[\r<F\x07Q\xc2@,\\\\\xc9\x10\x9c'  # noqa: E501
 GAUGE_WITHDRAW = b'\x88N\xda\xd9\xceo\xa2D\r\x8aT\xcc\x124\x90\xeb\x96\xd2v\x84y\xd4\x9f\xf9\xc76a%\xa9BCd'  # noqa: E501
@@ -102,7 +146,7 @@ class CurveDecoder(DecoderInterface, ReloadableDecoderMixin):
             transaction: EvmTransaction,
             tx_log: EvmTxReceiptLog,
             decoded_events: list['EvmEvent'],
-            user_address: ChecksumEvmAddress,
+            user_or_contract_address: ChecksumEvmAddress,
     ) -> DecodingOutput:
         """Decode information related to withdrawing assets from curve pools"""
         withdrawal_events: list['EvmEvent'] = []
@@ -118,7 +162,7 @@ class CurveDecoder(DecoderInterface, ReloadableDecoderMixin):
                 event.event_type == HistoryEventType.RECEIVE and
                 event.event_subtype == HistoryEventSubType.NONE and
                 event.asset == A_ETH and
-                event.location_label == user_address
+                event.location_label == user_or_contract_address
             ):
                 event.event_type = HistoryEventType.WITHDRAWAL
                 event.event_subtype = HistoryEventSubType.REMOVE_ASSET
@@ -130,7 +174,8 @@ class CurveDecoder(DecoderInterface, ReloadableDecoderMixin):
                 event.event_subtype == HistoryEventSubType.NONE and
                 event.location_label == transaction.from_address and
                 (
-                    user_address == event.location_label or
+                    user_or_contract_address == event.location_label or
+                    user_or_contract_address in CURVE_DEPOSIT_CONTRACTS or
                     tx_log.topics[0] == REMOVE_LIQUIDITY_IMBALANCE
                 )
             ):
@@ -143,13 +188,13 @@ class CurveDecoder(DecoderInterface, ReloadableDecoderMixin):
                 event.event_type == HistoryEventType.RECEIVE and
                 event.event_subtype == HistoryEventSubType.NONE and
                 event.location_label == transaction.from_address and
-                user_address == event.location_label and
+                user_or_contract_address == event.location_label and
                 tx_log.address in self.curve_pools
             ):
                 event.event_type = HistoryEventType.WITHDRAWAL
                 event.event_subtype = HistoryEventSubType.REMOVE_ASSET
                 event.counterparty = CPT_CURVE
-                event.notes = f'Remove {event.balance.amount} {crypto_asset.symbol} from the curve pool {tx_log.address}'  # noqa: E501
+                event.notes = f'Remove {event.balance.amount} {crypto_asset.symbol} from {tx_log.address} curve pool'  # noqa: E501
                 withdrawal_events.append(event)
 
         # Make sure that the order is the following:
@@ -158,18 +203,24 @@ class CurveDecoder(DecoderInterface, ReloadableDecoderMixin):
         # 3. Withdrawal 2
         # etc.
         if return_event is None or len(withdrawal_events) == 0:
+            # for deposit zap contracts, this is handled during post decoding
+            if user_or_contract_address in CURVE_DEPOSIT_CONTRACTS and return_event is not None:
+                return DecodingOutput(matched_counterparty=CPT_CURVE)
+
             log.debug(
-                f'Expected to see a receive pool token event and deposit events for a curve pool, '
-                f'but have not found them. Tx_hash: {transaction.tx_hash.hex()} '
-                f'User address: {user_address}',
+                f'Expected to see a return pool token event and '
+                f'withdrawal events for a curve pool, but have not found them. '
+                f'Tx_hash: {transaction.tx_hash.hex()} '
+                f'User address: {user_or_contract_address}',
             )
             return DEFAULT_DECODING_OUTPUT
 
-        return_event.extra_data = {'withdrawal_events_num': len(withdrawal_events)}  # for accounting  # noqa: E501
-        previous_event = return_event
-        for event in withdrawal_events:
-            maybe_reshuffle_events(previous_event, event, decoded_events)
-            previous_event = event
+        self._sort_events(
+            action_type='removal',
+            return_or_receive_event=return_event,
+            withdrawal_or_deposit_events=withdrawal_events,
+            all_events=decoded_events,
+        )
         return DEFAULT_DECODING_OUTPUT
 
     def _decode_curve_deposit_events(
@@ -177,7 +228,7 @@ class CurveDecoder(DecoderInterface, ReloadableDecoderMixin):
             transaction: EvmTransaction,
             tx_log: EvmTxReceiptLog,
             decoded_events: list['EvmEvent'],
-            user_address: ChecksumEvmAddress,
+            user_or_contract_address: ChecksumEvmAddress,
     ) -> DecodingOutput:
         """Decode information related to depositing assets in curve pools"""
         deposit_events: list['EvmEvent'] = []
@@ -193,37 +244,17 @@ class CurveDecoder(DecoderInterface, ReloadableDecoderMixin):
                 event.event_type == HistoryEventType.SPEND and
                 event.event_subtype == HistoryEventSubType.NONE and
                 event.asset == A_ETH and
-                event.location_label == user_address
+                event.location_label == user_or_contract_address
             ):
                 event.event_type = HistoryEventType.DEPOSIT
                 event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
                 event.counterparty = CPT_CURVE
                 event.notes = f'Deposit {event.balance.amount} {crypto_asset.symbol} in curve pool'  # noqa: E501
-                deposit_events.append(event)
-            elif (  # deposit give asset
-                (
-                    event.event_type == HistoryEventType.SPEND and
-                    event.event_subtype == HistoryEventSubType.NONE and
-                    event.location_label == user_address and
-                    tx_log.address in self.curve_pools
-                ) or
-                (
-                    tx_log.topics[0] == ADD_LIQUIDITY_4_ASSETS and
-                    user_address == CURVE_Y_DEPOSIT and
-                    event.asset != A_ETH
-                )
-            ):
-                event.event_type = HistoryEventType.DEPOSIT
-                event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
-                event.counterparty = CPT_CURVE
-                event.notes = f'Deposit {event.balance.amount} {crypto_asset.symbol} in curve pool'  # noqa: E501
-                if tx_log.address in self.curve_pools:
-                    event.notes += f' {tx_log.address}'
                 deposit_events.append(event)
             elif (  # Deposit receive pool token
                 event.event_type == HistoryEventType.RECEIVE and
                 event.event_subtype == HistoryEventSubType.NONE and
-                event.location_label == user_address and
+                event.location_label == user_or_contract_address and
                 tx_log.address in self.curve_pools
             ):
                 event.event_type = HistoryEventType.RECEIVE
@@ -234,7 +265,10 @@ class CurveDecoder(DecoderInterface, ReloadableDecoderMixin):
             elif (  # deposit give asset
                 event.event_type == HistoryEventType.SPEND and
                 event.event_subtype == HistoryEventSubType.NONE and
-                event.location_label == user_address and
+                (
+                    event.location_label == user_or_contract_address or
+                    user_or_contract_address in CURVE_DEPOSIT_CONTRACTS
+                ) and
                 tx_log.address in self.curve_pools
             ):
                 event.event_type = HistoryEventType.DEPOSIT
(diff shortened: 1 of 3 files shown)


rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -285,6 +291,61 @@ def _maybe_enrich_curve_transfers(
...
+            event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
+            event.counterparty = CPT_CURVE
+            event.notes = f'Deposit {event.balance.amount} {crypto_asset.symbol} in curve pool'
+            return True
+
+        if (
+            event.event_type == HistoryEventType.RECEIVE and
+            event.event_subtype == HistoryEventSubType.NONE and
+            to_address == transaction.from_address and
+            transaction.input_data.startswith(CURVE_DEPOSIT_ADD_LIQUIDITY_METHOD)
```
nebolax, line 313: Hmmmm I would not rely on such input_data checks.

What if these add / remove liquidity method are called as internal transactions?

nebolax, line 313: Also question. Are these depositing contracts the only way to deposit to pools like USDN?

If deposits can also be made by other contracts then this check with the executed method won't work there since the methods might have different names and signatures.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -285,6 +291,61 @@ def _maybe_enrich_curve_transfers(
...
+        ):
+            crypto_asset = event.asset.resolve_to_crypto_asset()
+            event.event_type = HistoryEventType.RECEIVE
+            event.event_subtype = HistoryEventSubType.RECEIVE_WRAPPED
+            event.counterparty = CPT_CURVE
+            event.notes = f'Receive {event.balance.amount} {crypto_asset.symbol} for depositing in the curve pool'  # noqa: E501
```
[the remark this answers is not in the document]
prettyirrelevant: The pool cannot be known since only transfer txns happen.
## Deposits
-> add token e.g. USDT
<- receive wrapped token e.g. usdnCrv

## Withdrawals
-> return wrapped token e.g. usdnCrv
<- receive one or more tokens deposited e.g. dai, usdt, usdc

**All transfer events**

rotkehlchen/chain/ethereum/modules/curve/decoder.py
prettyirrelevant: > Also question. Are these depositing contracts the only way to deposit to pools like USDN?

For USDN, yeah.

> If deposits can also be made by other contracts then this check with the executed method won't work there since the methods might have different names and signatures.


Other contracts use the same factory contract, the only difference is the contract variables at deployment i.e. the pool, base_pool and token

image

prettyirrelevant: Internal transactions? That can only happen for ETH transactions right?

nebolax: No-no. Example: https://etherscan.io/tx/0xecc7aedd140923d7760476ed84c00318c943fddddef33111f6bcf957268bcee1/advanced

As you can see here https://ethtx.info/mainnet/0xecc7aedd140923d7760476ed84c00318c943fddddef33111f6bcf957268bcee1/ it calls `RemoveLiquidityOne` method underneath 
image

nebolax: > > Also question. Are these depositing contracts the only way to deposit to pools like USDN?
>
> For USDN, yeah.

I just checked etherscan. This doesn't seem to be true. Example: https://etherscan.io/tx/0xca0fa8063e9ca48db77107c1c67ed3e7051eb275ff0ae003517ac0707ef413d3. 

Since this is not true, this concern is still relevant.
> If deposits to, say, USDN pool, can also be made by other contracts then this check with the executed method won't work there since the methods might have different names and signatures.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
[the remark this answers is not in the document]
nebolax: Can't the emitter of the add / remove liquidity event be checked? 
image

(took the screenshot from here https://etherscan.io/tx/0x6c28df56ae4a7f784577273f72402a9b6640024327ee952fdde72c9cfdf08da5#eventlog)

nebolax (commented):
Btw, just noticed. Related curve tests have failed.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -257,35 +259,155 @@ def _decode_curve_events(  # pylint: disable=no-self-use
...
+            log.debug(
+                f'Expected to see a return pool token event and withdrawal events '
+                f'for a curve pool, but have not found them. '
+                f'Tx_hash: {transaction.tx_hash.hex()}',
+            )
+            return decoded_events
+
+        # is this really needed?
+        pool_address = None
+        for tx_log in all_logs:
```
nebolax, line 383: What if a transaction has multiple remove liquidity events?

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -257,35 +259,155 @@ def _decode_curve_events(  # pylint: disable=no-self-use
...
+
+        # exit early if previously decoded deposit events aren't present.
+        if len(deposit_events) == 0:
+            log.debug(
+                f'Expected to see a receive pool token event and deposit events for a curve pool, '
+                f'but have not found them. Tx_hash: {transaction.tx_hash.hex()}',
+            )
+            return decoded_events
+
+        # search for the receipt of the wrapped token after deposit
```
nebolax, line 316: iirc we tend to don't use word `receipt`. And use `receive` everywhere instead.

nebolax: iirc we tend to don't use word `receipt`. And use `receive` instead.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -257,35 +259,155 @@ def _decode_curve_events(  # pylint: disable=no-self-use
...
+        if receive_event_idx is None:
+            log.debug(
+                f'Expected to see a receive pool token event and deposit events for a curve pool, '
+                f'but have not found them. Tx_hash: {transaction.tx_hash.hex()}',
+            )
+            return decoded_events
+
+        decoded_events[receive_event_idx].extra_data = {'deposit_events_num': len(deposit_events)}  # for accounting  # noqa: E501
+        previous_event = decoded_events[receive_event_idx]
+        for event in deposit_events:
```
nebolax, line 342: If I am not mistaken then the logic from `_handle_zap_contracts_liquidity_addition`, `_handle_zap_contracts_liquidity_removal`, `_decode_curve_deposit_events` and `_decode_curve_remove_events` can be abstracted.

nebolax (commented):
Some more comments

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -27,12 +26,35 @@
...
 REMOVE_LIQUIDITY_IMBALANCE = b'\xb9d\xb7/s\xf5\xef[\xf0\xfd\xc5Y\xb2\xfa\xb9\xa7\xb1*9\xe4x\x17\xa5G\xf1\xf0\xae\xe4\x7f\xeb\xd6\x02'  # noqa: E501
-CURVE_Y_DEPOSIT = string_to_evm_address('0xbBC81d23Ea2c3ec7e56D39296F0cbB648873a5d3')
+REMOVE_LIQUIDITY_EVENTS = {
+    REMOVE_LIQUIDITY_IMBALANCE,
+    b'\x9e\x96\xdd;\x99z*%~\xecM\xf9\xbbn\xafbn m\xf5\xf5C\xbd\x966\x82\xd1C0\x0b\xe3\x10',  # REMOVE_ONE  # noqa: E501
+    b"Z\xd0V\xf2\xe2\x8a\x8c\xec# \x15@k\x846h\xc1\xe3l\xdaY\x81'\xec;\x8cY\xb8\xc7's\xa0",  # REMOVE_LIQUIDITY  # noqa: E501,
+    b'\xa4\x9dL\xf0&V\xae\xbf\x8cw\x1fZ\x85\x85c\x8a*\x15\xeel\x97\xcfr\x05\xd4 \x8e\xd7\xc1\xdf%-',  # REMOVE_LIQUIDITY_3_ASSETS  # noqa: E501
+    b'\x98x\xca7^\x10o*C\xc3\xb5\x99\xfcbEh\x13\x1cL\x9aK\xa6j\x14V7\x15v;\xe9\xd5\x9d',  # REMOVE_LIQUIDITY_4_ASSETS  # noqa: E501
+}
+CURVE_DEPOSIT_CONTRACTS = {
```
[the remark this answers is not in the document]
prettyirrelevant: > How did you get this list of deposit contracts?

https://etherscan.io/find-similar-contracts?a=0x094d12e5b541784701FD8d65F11fc0598FBC6332

> Did you make sure that these are the only deposit contracts?

Yeah, used etherscan fuzzy search.
image
image


> How frequently can new such contracts pop up? Will it worth putting them in an automatically updatable cache?

The last time such contract popped up was Feb, 2021 so I don' think an automatically updateable cache is needed.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -115,11 +138,12 @@ def _decode_curve_remove_events(
         # 3. Withdrawal 2
         # etc.
         if return_event is None or len(withdrawal_events) == 0:
-            log.debug(
-                f'Expected to see a receive pool token event and deposit events for a curve pool, '
-                f'but have not found them. Tx_hash: {transaction.tx_hash.hex()} '
-                f'User address: {user_address}',
-            )
+            if user_address not in CURVE_DEPOSIT_CONTRACTS:
```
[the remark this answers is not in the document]
prettyirrelevant: It is to prevent a false positive in the logs since the behaviour of not seeing receive pool event & deposit events for a deposit contract is expected.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -157,26 +181,6 @@ def _decode_curve_deposit_events(
                 event.counterparty = CPT_CURVE
                 event.notes = f'Deposit {event.balance.amount} {crypto_asset.symbol} in curve pool'  # noqa: E501
                 deposit_events.append(event)
-            elif (  # deposit give asset
```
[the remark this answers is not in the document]
prettyirrelevant: It's not needed since the `curve y-deposit` contract addition & removal of liquidity is now being handled at the post decoding level.

`Curve Y-Deposit` contract is similar to `Curve USDN deposit`

rotkehlchen/chain/ethereum/modules/curve/decoder.py
prettyirrelevant: that's the reason for my uncertainty since multiple remove liquidity events can occur

rotkehlchen/chain/ethereum/modules/curve/decoder.py
prettyirrelevant: The only piece of code that is common to them is
```python
        decoded_events[return_event_idx].extra_data = {'withdrawal_events_num': len(withdrawal_events)}  # for accounting  # noqa: E501
        previous_event = decoded_events[return_event_idx]
        for event in withdrawal_events:
            maybe_reshuffle_events(previous_event, event, decoded_events)
            previous_event = event
```

Or are you talking about something else?

rotkehlchen/chain/ethereum/modules/curve/decoder.py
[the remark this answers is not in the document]
nebolax: Makes sense

rotkehlchen/chain/ethereum/modules/curve/decoder.py
nebolax: Yeah, at least this. Also in general labeling deposit / withdraw events and receive / return wrapped tokens functionality looks same to me.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -228,64 +233,181 @@ def _decode_curve_events(  # pylint: disable=no-self-use
...
+
+        This is required because certain events are not available when `AddLiquidity` or `RemoveLiquidity` occurs.
+        An example of such is when liquidity is removed, the transfer event of the token removed comes after the
+        `RemoveLiquidity` event.
+        """  # noqa: E501
+        liquidity_action = None
+        for tx_log in all_logs:
+            if (
+                tx_log.topics[0] in REMOVE_LIQUIDITY_EVENTS and
+                transaction.to_address in CURVE_DEPOSIT_CONTRACTS
```
nebolax, line 271: Checking `transaction.to_address` doesn't make it more reliable than checking input data :stuck_out_tongue:

Will have same problems.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -257,35 +259,155 @@ def _decode_curve_events(  # pylint: disable=no-self-use
...
+        Curve zap contracts such as e.g. Curve Deposit USDN, etc.
+        """
+        return_event_idx, withdrawal_events = None, []
+        for idx, event in enumerate(decoded_events):
+            if (
+                event.event_type == HistoryEventType.SPEND and
+                event.event_subtype == HistoryEventSubType.RETURN_WRAPPED and
+                event.counterparty == CPT_CURVE
+            ):
+                # exit early since transaction was decoded properly.
```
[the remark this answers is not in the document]
nebolax: But isn't it guaranteed that if it is a zap deposit contract then it will not be decoded correctly?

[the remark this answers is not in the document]
nebolax: Also if I am not mistaken then currently if a transaction has multiple deposits / withdrawals to pools then if at least one of the deposits / withdrawals is decoded correctly the post decoding rules won't run. Am I correct? If yes, then this looks like something that needs to change.

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -808,3 +809,149 @@ def test_deposit_multiple_tokens(ethereum_transaction_decoder, ethereum_accounts
...
+            event_type=HistoryEventType.DEPOSIT,
+            event_subtype=HistoryEventSubType.DEPOSIT_ASSET,
+            asset=A_USDC,
+            balance=Balance(FVal('761.396655')),
+            location_label=user_address,
+            counterparty=CPT_CURVE,
+            notes='Deposit 761.396655 USDC in curve pool 0xbEbc44782C7dB0a1A60Cb6fe97d0b483032FF1C7',  # noqa: E501
+        ),
+    ]
+    assert events == expected_events
```
nebolax, line 866: Looks like this test failed

rotkehlchen/chain/ethereum/modules/curve/decoder.py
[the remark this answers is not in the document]
prettyirrelevant: > But isn't it guaranteed that if it is a zap deposit contract then it will not be decoded correctly?

Yeah, until it gets to the post-decoding process.

> Also if I am not mistaken then currently if a transaction has multiple deposits / withdrawals to pools then if at least one of the deposits / withdrawals is decoded correctly the post decoding rules won't run. Am I correct? If yes, then this looks like something that needs to change.

Not sure I totally understand you here but I'd describe what I understood...
## 1.
>If a transaction has multiple deposits/withdrawals to pools

If this happens, then I expect(not entirely sure) the already existing decoder to handle it. The post decoding only exists to handle decoding liquidity addition/removal via the **Zap depositer contracts**.

Also, looking at the transactions page of the contracts(Curve Deposit: USDN), they all happen to call a single function(you're either removing liquidity or adding it). I didn't come across a transaction where an aggregation(multicall) of method calls occur(scanning through the ~25 pages of transactions).

Sorry but if you could point me to a sample txn where such scenario occur, to avoid a revisit on this at a later point I'd really appreciate it.

## 2.
> then if at least one of the deposits / withdrawals is decoded correctly the post decoding rules won't run. Am I correct?

Yeah, you're correct that if a combination of `AddLiquidity` & `RemoveLiquidity` events in a single txn for Zap depositer contracts occur. It'd be an unexpected behaviour.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
prettyirrelevant: The point of checking the `transaction.to_address` is to prevent the post decoding rule from running for every possible transaction.

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -808,3 +808,148 @@ def test_deposit_multiple_tokens(ethereum_transaction_decoder, ethereum_accounts
         ),
     ]
     assert events == expected_events
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('ethereum_accounts', [['0xdE206bC0Fde2eF5C8BB6A1d552a64F82A2407Be4']])
+def test_curve_usdn_add_liquidity(ethereum_transaction_decoder, ethereum_accounts):
```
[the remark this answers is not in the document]
prettyirrelevant: there's a test for curve y contract also.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
prettyirrelevant: > Also in general labeling deposit / withdraw events and receive / return wrapped tokens functionality looks same to me

Yeah to an extent. But combining/abstracting in this case will result in it being harder to read but the mention I pointed out above has been abstracted.

If you feel that's not the case, for sure I'd abstract the said logics.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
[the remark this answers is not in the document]
nebolax: Yeah, I see. I think we then can consider a limitation the fact that it won't work good with multiple deposits / withdrawals in one transaction.

Example transaction with multiple deposits btw https://etherscan.io/tx/0xc2c0c5e2941795132bc06b7933b130471d6d147a6ae9f0a7beadca4532fb93b1/

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -228,64 +235,190 @@ def _decode_curve_events(  # pylint: disable=no-self-use
...
+            all_logs: list[EvmTxReceiptLog],  # pylint: disable=unused-argument
+    ) -> list[HistoryBaseEntry]:
+        """This method handles post decoding for curve events made through the deposit contracts.
+
+        This is required because certain events are not available when `AddLiquidity` or `RemoveLiquidity` occurs.
+        An example of such is when liquidity is removed, the transfer event of the token removed comes after the
+        `RemoveLiquidity` event.
+        """  # noqa: E501
+        # only run this post decoding rule for transactions interacting the contract.
+        if transaction.to_address not in CURVE_DEPOSIT_CONTRACTS:
```
nebolax, line 270: As said before -- this doesn't sound good to me because if a transaction was made via a smart contract wallet, or via some aggregator, or there is basically any intermediate address between the caller and the curve deposit contract then these post decoding rules won't run.

But leaving this question open for @LefterisJP

rotkehlchen/chain/ethereum/modules/curve/decoder.py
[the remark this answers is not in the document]
LefterisJP: Interesting. Nice approach @prettyirrelevant.

Can you please also do the following:

1. Write how you got the list (as you did in the PR comment), in a comment above the list in the code?
2.Ask in zapper's or whoever makes them chat/discord/twitter if they have maybe a list we can query instead of doing it manually?

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -59,7 +76,7 @@ def _decode_curve_remove_events(
             transaction: EvmTransaction,
             tx_log: EvmTxReceiptLog,
             decoded_events: list[HistoryBaseEntry],
-            user_address: ChecksumEvmAddress,
+            maybe_user_address: ChecksumEvmAddress,
     ) -> tuple[Optional[HistoryBaseEntry], list[ActionItem]]:
         """Decode information related to withdrawing assets from curve pools"""
```
LefterisJP, line 81: Can you add a comment in this function to explain why it's `maybe_` prefixed? So it is users address or ... what? In which case a better name may be `user_or_pool_address`? Or I am not understanding the logic here?

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -115,26 +133,29 @@ def _decode_curve_remove_events(
         # 3. Withdrawal 2
         # etc.
         if return_event is None or len(withdrawal_events) == 0:
-            log.debug(
-                f'Expected to see a receive pool token event and deposit events for a curve pool, '
-                f'but have not found them. Tx_hash: {transaction.tx_hash.hex()} '
-                f'User address: {user_address}',
-            )
+            if maybe_user_address not in CURVE_DEPOSIT_CONTRACTS:
```
LefterisJP, line 136: Why this extra if? So if it's in curve deposits why it is fine to just `return None, []`?

There is no comment here or in the docstring of the function explaining the logic.
I guess it moves somewhere else? You need to write where/why

rotkehlchen/chain/ethereum/modules/curve/decoder.py
LefterisJP: Yes this won't work. We can't check only for direct contract interaction. You also asked me in discord and I said this is not fine.

Also keep in mind the refactoring of post-decoding rules that @yabirgb is going to do in develop so that post decoding only runs for a transaction if the transaction has decoded an event for the given protocol

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -228,64 +235,190 @@ def _decode_curve_events(  # pylint: disable=no-self-use
...
+        An example of such is when liquidity is removed, the transfer event of the token removed comes after the
+        `RemoveLiquidity` event.
+        """  # noqa: E501
+        # only run this post decoding rule for transactions interacting the contract.
+        if transaction.to_address not in CURVE_DEPOSIT_CONTRACTS:
+            return decoded_events
+
+        liquidity_action = None
+        for tx_log in all_logs:
+            if tx_log.topics[0] in REMOVE_LIQUIDITY_EVENTS:
```
LefterisJP, line 275: What if there is a single transaction that has multiple liquidity addition and removal for different pools? This logic will detect the first and skip the rest.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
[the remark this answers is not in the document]
LefterisJP: Alexey is right and I think this is the same thing as I mentioned in `_handle_post_decoding` function. I don't want to put more limitations when we know about them. It should work for all cases, so this needs to be changed.

Ofcourse there can be multiple deposits and withdrawals in different pools in one transaction. That's the whole point of a Zap contract. To combine stuff. Lots of crazy combinations in a single transaction exist out there.

you need to write decoders that are future-proof and work for all of that. not break easily.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -228,64 +235,190 @@ def _decode_curve_events(  # pylint: disable=no-self-use
...
-            event.event_subtype = HistoryEventSubType.REMOVE_ASSET
-            event.counterparty = CPT_CURVE
-            event.notes = f'Receive {event.balance.amount} {crypto_asset.symbol} from the curve pool {CURVE_Y_DEPOSIT}'  # noqa: E501
-            return True
-        return False
+        if action_type == 'addition':
+            return_or_receive_event.extra_data = {
+                'deposit_events_num': len(withdrawal_or_deposit_events),
+            }
+        else:  # can only removal
```
LefterisJP, line 412:
```suggestion
        else:  # can only be removal
```

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -808,3 +809,149 @@ def test_deposit_multiple_tokens(ethereum_transaction_decoder, ethereum_accounts
...
+            notes='Deposit 761.396655 USDC in curve pool 0xbEbc44782C7dB0a1A60Cb6fe97d0b483032FF1C7',  # noqa: E501
+        ),
+    ]
+    assert events == expected_events
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('ethereum_accounts', [['0x6d84264A7bD2Cffa4A117BA2350403b3A9866949']])
+def test_curve_usdn_remove_liquidity(ethereum_transaction_decoder, ethereum_accounts):
+    """Check that removing liquidity from a curve pool using the USDN contract is properly decoded."""  # noqa: E501
```
LefterisJP, line 872: Are there tests for other zaps?

LefterisJP (commented):
So yeah left some comments. And the commentts that @nebolax made stand and were not addressed so I am not sure why this PR ended up for review by me.

IF you know something is wrng, and ask me in discord and tell you that yes it needs to be addressed then address it. Do not send me the PR for review.

Once addressed I would like you to review again @nebolax

rotkehlchen/tests/unit/decoders/test_curve.py
prettyirrelevant: there's an existing test for curve y deposit(a zap)

nebolax (commented):
Looked at the new approach. Currently you made it to iterate all transaction logs for each transaction, right?

If I understood correctly then this is not gonna work in the current state, however if you manage to integrate it with the refactor of post decoding rules and make it so that these rules run only for curve-zap-affected transactions, then I suppose it should be fine.

Currently this PR is targeting bugfixes, so to use those new rules I suppose you will have to move it to develop.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -48,28 +53,208 @@
...
-    string_to_evm_address('0xFCBa3E75865d2d561BE8D220616520c171F12851'),  # curve susd v2
-    string_to_evm_address('0x331aF2E331bd619DefAa5DAc6c038f53FCF9F785'),  # curve zap
-    string_to_evm_address('0xac795D2c97e60DF6a99ff1c814727302fD747a80'),  # curve usdt
-    string_to_evm_address('0x071c661B4DeefB59E2a3DdB20Db036821eeE8F4b'),  # curve bbtc
-    string_to_evm_address('0xd5BCf53e2C81e1991570f33Fa881c49EEa570C8D'),  # curve obtc
-    string_to_evm_address('0x11F419AdAbbFF8d595E7d5b223eee3863Bb3902C'),  # curve pbtc
-    string_to_evm_address('0xaa82ca713D94bBA7A89CEAB55314F9EfFEdDc78c'),  # curve tbtc
+    string_to_evm_address('0x094d12e5b541784701FD8d65F11fc0598FBC6332'): (  # curve usdn
+        {
+            string_to_evm_address('0x6B175474E89094C44Da98b954EedeAC495271d0F'),  # dai
```
nebolax, line 60: I think it should be sufficient to map only to lp token / address of the pool. And coins can be found in the curve cache, can't they?

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -157,9 +343,39 @@ def _decode_curve_remove_events(
...
+            # so we loop through all the logs searching for the ones that match.
+            if user_or_pool_address in CURVE_DEPOSIT_CONTRACTS and return_event is not None:
+                assets_to_amount: dict[Asset, int] = defaultdict(int)
+                for log_entry in all_logs:
+                    if (
+                        log_entry.topics[0] == ERC20_OR_ERC721_TRANSFER and
+                        hex_or_bytes_to_address(log_entry.topics[1]) == user_or_pool_address and
+                        log_entry.address in CURVE_DEPOSIT_CONTRACTS[user_or_pool_address][0]
+                    ):
+                        assets_to_amount[Asset(ethaddress_to_identifier(log_entry.address))] = hex_or_bytes_to_int(log_entry.data[:32])  # noqa: E501
```
nebolax, line 356: Afaik we try to avoid using `ethaddress_to_identifier` in favor of `evm_address_identifier`

nebolax (commented):
I played with a few addresses (`0xd829B67e1D4d24C5947031efcddA28388612E9b2`, `0xBb4153B55A59cC8BdE72550B0Cf16781b08Ef7b0`) to test the functionality and it looks like some events are not being decoded currently :( (maybe they are not related to zap decoding, but I think it will be good to check them) 

Also during testing I found event type `RemoveLiquidityOne` (and probably there is a corresponsing event type for deposits) that is currently is not supported in our decoder https://etherscan.io/tx/0x362ccc6bb40929ce6375a929cd2e3db6984934cbb9f7281d77d33c25643aa2cd.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
prettyirrelevant: Some deposit zap contracts don't have a pool/base_pool so the curve_pool cache does not work here.

e.g. `0xA50cCc70b6a011CffDdf45057E39679379187287` `0xbBC81d23Ea2c3ec7e56D39296F0cbB648873a5d3` `0xb6c057591E073249F2D9D88Ba59a46CFC9B59EdB` to name a few

nebolax: How can it be like that? Are you sure that these really curve deposit contracts if they don't interact with curve pools?

Btw, I checked and it seems like they do. Examples:
1. `0xA50cCc70b6a011CffDdf45057E39679379187287`: https://etherscan.io/tx/0x53ef827562d9283943f688e97ace1f5816ee12b3cd9c9be73023360e22c901ba#eventlog
2. `0xbBC81d23Ea2c3ec7e56D39296F0cbB648873a5d3`: https://etherscan.io/tx/0xd8eefcd63ea90ff98245a9a9ac5177b8b15a49e9270e8bfa9e832adb3757aab0#eventlog
3. `0xb6c057591E073249F2D9D88Ba59a46CFC9B59EdB`: https://etherscan.io/tx/0x025e996b1b77dede53436e4c97bcf098a47d8699e7ee1e644f3b33bd14d17f6b#eventlog

They all 3 interact with curve pools, or do I misunderstand something?

prettyirrelevant: The logs only show transfers of **coins**. No interaction with pools. Neither does the contract define them.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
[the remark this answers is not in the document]
prettyirrelevant: > Ask in zapper's or whoever makes them chat/discord/twitter if they have maybe a list we can query instead of doing it manually?

The list they provided was for just a subset(like 4). Also, there's no API nor registry contract for this

prettyirrelevant:
> Ofcourse there can be multiple deposits and withdrawals in different pools in one transaction. That's the whole point of a Zap contract. To combine stuff. Lots of crazy combinations in a single transaction exist out there.
> 
> you need to write decoders that are future-proof and work for all of that. not break easily.

From transactions I've encountered, it's either a deposit or withdrawal. If multiple are combined, then its via a protocol -- we'd have the support the said protocol e.g. [PowerPool Finance](https://powerpool.finance/), etc.
For Curve, it is primarily for adding tokens to a pool or removing tokens.

@nebolax. For the action items approach, certain limitations exist.
1. When sorting deposit events and the return events, there was no way to avoid using a nested loop because certain events like **approval** & **revoke** can occur between deposit events.

2. For 3Pool Zap contract(0xA79828DF1850E8a3A3064576f380D90aECDD3359), can interact with multiple pools and as such does not have a defined pool token.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -430,26 +461,141 @@ def _decode_curve_gauge_events(self, context: DecoderContext) -> DecodingOutput:
...
+                event.event_type == HistoryEventType.RECEIVE and
+                event.event_subtype == HistoryEventSubType.NONE and
+                event.location_label == transaction.from_address and
+                event.address in CURVE_DEPOSIT_CONTRACTS
+            ):
+                crypto_asset = event.asset.resolve_to_crypto_asset()
+                event.event_type = HistoryEventType.WITHDRAWAL
+                event.event_subtype = HistoryEventSubType.REMOVE_ASSET
+                event.counterparty = CPT_CURVE
+                event.notes = f'Remove {event.balance.amount} {crypto_asset.symbol} from the curve pool'  # noqa: E501
```
LefterisJP, line 574: same comment for name/address of pool here. And it should be "a curve pool" if it was unknown --- but it can't be unknown

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -461,6 +607,32 @@ def _maybe_enrich_curve_transfers(self, context: EnricherContext) -> TransferEnr
...
+    ) -> list['EvmEvent']:
+        """
+        This method sorts the events in the expected order and adds
+        extra info to the in/out event depending on the action_type.
+        """
+        if action_type == 'addition':
+            return_or_receive_event.extra_data = {
+                'deposit_events_num': len(withdrawal_or_deposit_events),
+            }
+        else:  # can only removal
```
LefterisJP, line 625: I already asked you to fix it and you marked this as resolved but did not do it ... 3 weeks ago!!!

https://github.com/rotki/rotki/pull/5736/files#r1139452060
```suggestion
        else:  # can only be removal
```

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -730,7 +731,7 @@ def test_curve_remove_imbalanced(database, ethereum_transaction_decoder):
             asset=A_USDC,
             balance=Balance(amount=FVal('665.267705'), usd_value=ZERO),
             location_label=location_label,
-            notes='Receive 665.267705 USDC from the curve pool 0xbBC81d23Ea2c3ec7e56D39296F0cbB648873a5d3',  # noqa: E501
+            notes='Remove 665.267705 USDC from the curve pool',
```
LefterisJP, line 734: yeah see? The information is there. Want it back. Why remove it?

prettyirrelevant: > yeah see? The information is there. Want it back. Why remove it?

For this test, the address passed isn't a pool. It's the zap contract(Curve Y Deposit). Added it back but with the correct metapool.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -430,26 +461,141 @@ def _decode_curve_gauge_events(self, context: DecoderContext) -> DecodingOutput:
...
+                    event.address in CURVE_DEPOSIT_CONTRACTS or
+                    event.address == ZERO_ADDRESS
+                )
+            ):
+                receive_event_idx = idx
+                crypto_asset = event.asset.resolve_to_crypto_asset()
+                event.event_type = HistoryEventType.RECEIVE
+                event.event_subtype = HistoryEventSubType.RECEIVE_WRAPPED
+                event.counterparty = CPT_CURVE
+                event.notes = f'Receive {event.balance.amount} {crypto_asset.symbol} after depositing in curve pool'  # noqa: E501
```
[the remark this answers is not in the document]
prettyirrelevant: https://etherscan.io/tx/0xf7c6764b832069785eeee22a078f4cb3c92149c25eb0bdc6bba36ebd1598c255#eventlog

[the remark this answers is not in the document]
prettyirrelevant: > Don't you have either a name or an address of the curve pool here?

For liquidity, it's isn't always accurate for the pool token received.

Why?
In certain transactions, more than one `AddLiquidity` event can be emitted -- one for the metapool, the other for the base pool.

The base pool is emitted first and is sufficient to get all needed data. Meanwhile, the second event, is for the metapool , whose `tx_log.address` is the pool_token received.
https://etherscan.io/tx/0xf7c6764b832069785eeee22a078f4cb3c92149c25eb0bdc6bba36ebd1598c255#eventlog

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -430,26 +461,143 @@ def _decode_curve_gauge_events(self, context: DecoderContext) -> DecodingOutput:
...
```
LefterisJP, line 530: As I also wrote to you in discord and got no response I am doing it here too.

https://discord.com/channels/657906918408585217/1080783409640648765/1092506473491333140

I don't believe there is no way to find the right data here. But I am just tired of this PR being open for so long so if you absolutely can't find it someone else may do as an additional issue in the future.

```suggestion
                event.notes = f'Receive {event.balance.amount} {crypto_asset.symbol} after depositing in a curve pool'  # noqa: E501
```

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -430,26 +461,143 @@ def _decode_curve_gauge_events(self, context: DecoderContext) -> DecodingOutput:
...
+                event.event_type == HistoryEventType.RECEIVE and
+                event.event_subtype == HistoryEventSubType.NONE and
+                event.location_label == transaction.from_address and
+                event.address in CURVE_DEPOSIT_CONTRACTS
+            ):
+                crypto_asset = event.asset.resolve_to_crypto_asset()
+                event.event_type = HistoryEventType.WITHDRAWAL
+                event.event_subtype = HistoryEventSubType.REMOVE_ASSET
+                event.counterparty = CPT_CURVE
+                event.notes = f'Remove {event.balance.amount} {crypto_asset.symbol} from the curve pool {pool_address}'  # noqa: E501
```
LefterisJP, line 576: Do it just like the rest of the file. Be consistent.

```suggestion
                event.notes = f'Remove {event.balance.amount} {crypto_asset.symbol} from {pool_address} curve pool'  # noqa: E501
```

LefterisJP (commented):
Really tired of this PR being open for so long

rotkehlchen/chain/ethereum/modules/curve/decoder.py
prettyirrelevant: The rest of the file doesn't follow the format you suggested. Unless, you meant switching everything to what you suggested

LefterisJP: The gauges do. "The curve pool X" is not correct English. It's either:

1. from curve pool {name}
or
2. from {name} curve pool

And (2) is much more natural when you use it in a sentence.

LefterisJP (approved):
LGTM with the caveat we mentioned about the name of the pool in the Zap receive wrapped case. Let's see if we can find a way to handle that.

PR #5740, opened 2023-03-07
Handle nightlies for VCR tests

yabirgb:
When running nightlies the `GITHUB_HEAD_REF` is empty and then the nightlies fail to execute because a git command gets malformed. This PR handles all this cases

### DIFF
--- a/rotkehlchen/tests/conftest.py
+++ b/rotkehlchen/tests/conftest.py
@@ -230,6 +230,10 @@ def fixture_vcr_base_dir() -> Path:
     if 'CI' in os.environ:
         current_branch = os.environ.get('GITHUB_HEAD_REF')  # get branch from github actions
         root_dir = Path(os.environ['CASSETTES_DIR'])
+        if current_branch is None or (current_branch is not None and current_branch == ''):
+            # This is needed when the job doesn't happen due to a PR for example is executed
+            # in a cron job inside github
+            current_branch = os.environ.get('GITHUB_REF_NAME')
     else:
         current_branch = os.environ.get('VCR_BRANCH')
         root_dir = default_data_directory().parent / 'test-caching'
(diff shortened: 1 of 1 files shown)


yabirgb:
Test for this executed here https://github.com/yabirgb/rotki/actions/runs/4357533009

rotkehlchen/tests/conftest.py
```diff
@@ -230,6 +230,10 @@ def fixture_vcr_base_dir() -> Path:
     if 'CI' in os.environ:
         current_branch = os.environ.get('GITHUB_HEAD_REF')  # get branch from github actions
         root_dir = Path(os.environ['CASSETTES_DIR'])
+        if current_branch is None or (current_branch is not None and len(current_branch) == 0):
```
LefterisJP, line 233:
```suggestion
        if current_branch is None or (current_branch is not None and len(current_branch) == 0):
```

Why the 2nd part? `current_branch is not None and len(current_branch) == 0`? When does this happen?

rotkehlchen/tests/conftest.py
```diff
@@ -246,13 +250,26 @@ def fixture_vcr_base_dir() -> Path:
     log.debug(f'At VCR setup, {current_branch=} {root_dir=}')
 
     checkout_proc = Popen(f'cd "{root_dir}" && git fetch origin && git checkout {current_branch}', shell=True, stdout=PIPE, stderr=PIPE)  # noqa: E501
-    _, stderr = checkout_proc.communicate(timeout=SUBPROCESS_TIMEOUT)
+
+    try_fallback_branch = False
```
LefterisJP, line 254: Why this logic with `try_fallback_branch` and not just add one more `stderr` check? It only times out? Why? When does it happen?

rotkehlchen/tests/conftest.py
```diff
@@ -246,13 +250,26 @@ def fixture_vcr_base_dir() -> Path:
...
-        b'Switched to' not in stderr
+        try_fallback_branch is True or
+        (
+            len(stderr) != 0 and
+            b'Already on' not in stderr and
+            b'Switched to' not in stderr
+        )
     ):
         default_branch = os.environ.get('GITHUB_BASE_REF', os.environ.get('DEFAULT_VCR_BRANCH', 'develop'))  # noqa: E501
+        if len(default_branch) == 0 and 'CI' in os.environ:
```
LefterisJP, line 270: When is `len(default_branch) == 0`? If someone sets `GITHUB_BASE_REF` and/or `DEFAULT_VCR_BRANCH` to `''`? When can that happen?

rotkehlchen/tests/conftest.py
yabirgb: `GITHUB_HEAD_REF` is set only in pull requests and in the nighlies or runs that are not triggered by a PR the variable value is `''`. This is why I set the second part of the condition

rotkehlchen/tests/conftest.py
yabirgb: In the case of the `test` branch that we use to trigger the same work as in the nightlies I found that the command was timing out instead of failing directly. This case is the case where the `current_branch` doesn't exists. I tried to reproduce it locally and works fine so I believe is some special case in the CI

rotkehlchen/tests/conftest.py
yabirgb: `GITHUB_BASE_REF` is set to `''` for jobs that don't run from a pull requests and in the CI we don't set `DEFAULT_VCR_BRANCH`. If someone set this variables to `''` locally is the user faults but in the case of the CI we need to protect against this case

rotkehlchen/tests/conftest.py
LefterisJP: This sounds wrong. You should try to figure out why it is timing out.

rotkehlchen/tests/conftest.py
LefterisJP: Okay so does that mean that when any venv is not set it's `''`? That can't be right. So Github CI must specifically set it to the empty string. If so can you state this in a comment? And instead of doing `len(x) == 0` compare it to the empty string to make it obvious what you are doing?

rotkehlchen/tests/conftest.py
LefterisJP: again same as above. Comment and comparison to empty string. The length check is really realy confusing here.

rotkehlchen/tests/conftest.py
yabirgb: No no, is github setting it to `''`. Is not that is not set. Okey will assert that ` == ''` but made it like this because we always check len instead of `==''`

LefterisJP: no we only do that if it's a list not a string or worse yet here ... an object of any type. `len(x)` make you think it's a list.

yabirgb: Also is stated in a comment just below this line

LefterisJP (approved):
LGTM

PR #5748, opened 2023-03-09
fix: wrong price transaction event

### LINKED ISSUE #5733: Fix different value used in tx amount and pnl report
Issue 1: For the transaction event form 

So if the main currency is other than USD (e.g EUR),
it first will get the historic price of the EUR (based on the timestamp set there)
and then when saving, the EUR value is converted back to USD with current exchange rate 

This is wrong, it should use historic price too when saving.

Issue 2: The value used in PNL report is not coming from EVM tx

So for this, we need to add an action in each PNL report to edit the historical price, and then the user can recalculate the PNL report.

lukicenturi:
Closes #5733 

## Checklist

- [x] Fix different value used in `transaction event form`, `transaction event table`, and `pnl report events` in any currency (USD or other)
- [x] Add option to add historic price in pnl report events, so user can update the price and regenerate the pnl report.


lukicenturi:
@kelsos so for the `<AmountInput>` component, I had to take it from develop, because I need the functionality to show the price at the given timestamp

lukicenturi:
So for the transaction view table, for the fiat price below the amount of asset, I don't use the value returned from the endpoint, but calculate it by the historic price of the asset. This way we can keep it consistent between tx view and, the form, and pnl events.

PR #5753, opened 2023-03-10
Robust price query for yearn vaults and other known protocol tokens

LefterisJP:
Our logic was really flawed. Two things fixed.

### [More robust price query for yearn vault tokens](https://github.com/rotki/rotki/commit/09f4e3f4ff2981717fb89007d9dc73f0c543f39f) 

If a yearn vault is for some reason missing the underlying token from
the global DB we query the chain for it and fill it in so that the
price query works properly. Also save it in the DB, so that next time the on-chain query is skipped.

### [If known protocol token chain price query fails, fallback to oracles](https://github.com/rotki/rotki/commit/7812001228a0028a05dcbb846d76dfa4cf157730) 

The code before exited immediately after the on-chain query failed,
thinking that the external oracles will not be able to give an answer.

But that's misguided. Both coingecko and defillama have lots of those
special tokens and even the ability to query by address (which we
already utilize at least for defillama).

A user tried to query for example [yvDAI](https://www.coingecko.com/en/coins/yvdai) and it failed due to missing underlying token. Then they edited coingecko id and it did not work. Neither did defillama. The problem was that if on-chain query failed we gave up.

This is now fixed.

### DIFF
--- a/rotkehlchen/inquirer.py
+++ b/rotkehlchen/inquirer.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Iterable, NamedTuple, Optional, Sequence, Union
 
 from rotkehlchen.assets.asset import Asset, EvmToken, FiatAsset, UnderlyingToken
+from rotkehlchen.assets.utils import TokenSeenAt, get_or_create_evm_token
 from rotkehlchen.chain.ethereum.defi.price import handle_defi_price_query
 from rotkehlchen.chain.ethereum.utils import token_normalized_value_decimals
 from rotkehlchen.chain.evm.contracts import EvmContract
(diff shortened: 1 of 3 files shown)


rotkehlchen/inquirer.py
```diff
@@ -890,12 +888,43 @@ def find_yearn_price(
...
+        if maybe_underlying_tokens is None or len(maybe_underlying_tokens) != 1:
+            # underlying token not recorded in the DB. Ask the chain
+            try:
+                remote_underlying_token = contract.call(ethereum.node_inquirer, 'token')
+            except (RemoteError, BlockchainQueryError) as e:
+                log.error(f'Failed to query underlying token method in Yearn v2 Vault. {str(e)}')
+                return None
+
+            try:
+                underlying_token_address = deserialize_evm_address(remote_underlying_token)
```
yabirgb, line 907: Shouldn't this also do a get_or_create evm token? It might be the case that the underlying token is not in the globaldb for some reason

rotkehlchen/inquirer.py
```diff
@@ -890,12 +888,43 @@ def find_yearn_price(
...
+                            token_kind=EvmTokenKind.ERC20,  # this may be a guess here
+                            weight=ONE,  # all yearn vaults have single underlying
+                        )],
+                    chain_id=ChainID.ETHEREUM,
+                )
+        else:
+            underlying_token_address = maybe_underlying_tokens[0].address
 
-        underlying_token = EvmToken(ethaddress_to_identifier(maybe_underlying_token[0].address))
+        underlying_token = EvmToken(ethaddress_to_identifier(underlying_token_address))
```
yabirgb, line 927: this assumes that the underlying token exists in the globaldb, is that guaranteed somewhere? I don't see it right now

yabirgb (commented):
Thanks! This indeed is a much robust approach to the problem. Is something I didn't consider because we were adding them manually but there are cases as you mentioned deleting the underlying token where this logic breaks.

Left a small issue that I believe can break the flow

yabirgb (approved):
Thanks. LGTM

PR #5756, opened 2023-03-10
Rewrite manual historical prices on addition

### LINKED ISSUE #5741: Adding historic price with same `timestamp, toAsset, and fromAsset` doesn't replace the price of old entry.
## Problem Definition

For the add historic price endpoint, adding the new entry with `timestamp`, `toAsset`, and `fromAsset` that already exist, doesn't update the price of the entry.

Can we make it, so that the new entry with the same payload (except the price), updates the already existing entry price?
I see that `adding the latest price` uses this mechanism.

(Why not just use the edit endpoint?) Sometimes the front end cannot make sure if the entry already exists or not.
In doing so, we need to fetch all saved historical prices. By implementing the mechanism I said in the paragraph above, we can prevent unnecessary fetching for all saved historical prices.

This is needed in order to finish #5733, so can this be put on `bugfixes`?

nebolax:
Closes #5741

## Checklist


rotkehlchen/api/rest.py
```diff
@@ -4012,6 +4012,10 @@ def add_manual_price(  # pylint: disable=no-self-use
         price: Price,
         timestamp: Timestamp,
     ) -> Response:
+        """
+        Add a manual historical price.
+        If price for the specified assets, source type and timestamp already exists -- replaces it.
```
prettyirrelevant, line 4017:
```suggestion
        If the price for the specified pair and timestamp already exists, it is replaced.
```

PR #5758, opened 2023-03-12
Sunday package bumpity bump

LefterisJP:


requirements_dev.txt
```diff
@@ -15,7 +15,7 @@ flaky==3.7.0
 
 
 # To test google spreadsheet uploading
-google-api-python-client==2.65.0
+google-api-python-client==2.80.0
 google-auth-httplib2==0.1.0
-google-api-core==1.32.0  # temporarily due to protobuf conflictwith web3
-google-auth==1.25.0  # temporarily due to protobuf conflictwith web3
+google-api-core==2.11.0  # temporarily due to protobuf conflictwith web3
+google-auth==2.16.2  # temporarily due to protobuf conflictwith web3
```
yabirgb, line 21: Guess we can remove the comments then?

rotkehlchen/externalapis/etherscan.py
```diff
@@ -355,7 +355,7 @@ def get_transactions(
                 options['txHash'] = period_or_hash.hex()
                 parent_tx_hash = period_or_hash
 
-        transactions: Union[Sequence[EvmTransaction], Sequence[EvmInternalTransaction]] = []  # noqa: E501
+        transactions: Union[list[EvmTransaction], list[EvmInternalTransaction]] = []  # type: ignore  # noqa: E501
```
yabirgb, line 358: why this change?

LefterisJP: because this was the least type ignores to make it work. list[None] != list[X] in mypy

requirements_dev.txt
LefterisJP: in another PR I guess

PR #5759, opened 2023-03-13
Remove `TestEnvironment` form tests discovery

nebolax:
When a class name starts with `Test` pytest considers it a class that contains tests.

`TestEnvironment` should not be treated like that, so I renamed it to `EnvironmentForTests`.

----------------------------------

Also question: do you think such PR is better to be merged to bugfixes or develop?

1. I think that develop is better because this warning was seen in nightlies and nightlies run on develop (so by merging into develop the warning would immediately go away)
2. I think that bugfixes is better because by merging into bugfixes first we will not have merge conflicts later when will be meging bugfixes -> develop.

What do you think?


rotkehlchen/tests/conftest.py
```diff
@@ -34,7 +34,7 @@
 SUBPROCESS_TIMEOUT = 30
 
 
-class TestEnvironment(SerializableEnumMixin):
+class EnvironmentForTests(SerializableEnumMixin):
```
prettyirrelevant, line 37: nitpick: Rather than doing a renaming...  you could just do this
```suggestion
class TestEnvironment(SerializableEnumMixin):
    __test__ = False
```

https://github.com/pytest-dev/pytest/pull/2099

prettyirrelevant:
> Also question №2 : do you think such PR is better to be merged to bugfixes or develop?

A pattern I've seen is merging to bugfixes and cherry-picking the commit to develop iirc. So I'm tilting in favour of option 2

LefterisJP:
No no please just do bugfixes. And we frequently merge bugfixes to develop anyway.

rotkehlchen/tests/conftest.py
nebolax: Didn't want to do it this was because declaring a `__test__` parameter in an enum looks strange to me, but okay.

rotkehlchen/tests/conftest.py
```diff
@@ -35,6 +35,8 @@
 
 
 class TestEnvironment(SerializableEnumMixin):
+    __test__ = False  # tell pytest to don't try to collect this class
```
prettyirrelevant, line 38:
```suggestion
    __test__ = False  # tell pytest not to collect this class
```

nebolax: What is the benefit of this change?

prettyirrelevant: just typo correction. That's all

LefterisJP (commented):
Fix the typo @prettyirrelevant pointed out.

> It will be better to instead of renaming TestEnvironment to just completely disable tests-classes discovery for pytest?

Sure that's also fine. we don't use test classes. But no preference here

LefterisJP (approved):
LGTM

PR #5765, opened 2023-03-14
add `proxies` to liquity stability & staking responses

prettyirrelevant:
Unblock @lukicenturi

test-caching PR: https://github.com/rotki/test-caching/tree/add-proxy-to-liquity-pools

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -7439,15 +7439,48 @@ Getting Liquity staked amount
       {
           "result": {
             "0x063c26fF1592688B73d8e2A18BA4C23654e2792E": {
-                "asset": "eip155:1/erc20:0x6DEA81C8171D0bA574754EF6F8b412F2Ed88c54D"
-                "amount": "177.02",
-                "usd_value": "265.530"
+                "balances": {
+                  "staked": {
+                    "asset": "ETH",
+                    "amount": "43.180853032438783295",
+                    "usd_value": "43.180853032438783295",
+                  },
+                  "lusd_rewards": {
+                    "asset": "eip155:1/erc20:0x6DEA81C8171D0bA574754EF6F8b412F2Ed88c54D",
+                    "amount": "94477.70111867384658505",
+                    "usd_value": "94477.70111867384658505",
+                  },
+                  "eth_rewards": {
+                    "asset": "eip155:1/erc20:0x5f98805A4E8be255a32880FDeC7F6728C6568bA0",
+                    "amount": "10211401.723115634393264567",
+                    "usd_value": "10211401.723115634393264567",
+                  }
+                },
+                "proxies": {
+                    "0x063c26fF1592688B73d8e2A18BA4C23654e2792E": {
+                      "staked": {
+                        "asset": "ETH",
+                        "amount": "43.180853032438783295",
+                        "usd_value": "43.180853032438783295",
+                      },
+                      "lusd_rewards": {
+                        "asset": "eip155:1/erc20:0x6DEA81C8171D0bA574754EF6F8b412F2Ed88c54D",
+                        "amount": "94477.70111867384658505",
+                        "usd_value": "94477.70111867384658505",
+                      },
+                      "eth_rewards": {
+                        "asset": "eip155:1/erc20:0x5f98805A4E8be255a32880FDeC7F6728C6568bA0",
+                        "amount": "10211401.723115634393264567",
+                        "usd_value": "10211401.723115634393264567",
+                      }
+                }
             }
           },
           "message": ""
       }
 
-   :resjson object result: A mapping of the amount and value of LQTY staked in the protocol.
+   :resjson object balances: A mapping of the category to the amount & value of assets staked in the protocol.
+   :resjson object proxies: A mapping of proxy addresses to the amount and value of assets staked in the protocol.
 
    :statuscode 200: Liquity staking information successfully queried.
    :statuscode 409: User is not logged in or Liquity module is not activated.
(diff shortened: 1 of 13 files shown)


rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -226,32 +240,49 @@ def _query_deposits_and_rewards(
...
             contract=self.staking_contract,
             addresses=addresses,
             methods=('stakes', 'getPendingLUSDGain', 'getPendingETHGain'),
             keys=('staked', 'lusd_rewards', 'eth_rewards'),
             assets=(A_LQTY, A_LUSD, A_ETH),
         )
+        staking_balances_with_proxy = {}
+        for addr, balances in staking_balances.items():
+            staking_balances_with_proxy[addr] = LiquityStakingBalance(
+                staked=balances['staked'],
```
yabirgb, line 283: Do you need to create the class instance?  From a quick check you can avoid creating a new object and also accessing the other values and just add the `proxy_to` information. Mentioning it because this looks like unnecessary access to memory and creating objects with extra calls with minimum advantages. If you want to use the types maybe you can use

 ```
class StabilityPoolBalance(TypedDict):
    gains: AssetBalance
    rewards: AssetBalance
    deposited: AssetBalance
```

and then

```
class StabilityPoolBalanceInfo(StabilityPoolBalance):
    proxy_to: Optional[CheckSumAddress]
```

and then `_query_deposits_and_rewards` returns those two depending on the arguments and you add the field 

```
output[addr]['proxy_to'] = self.ethereum.proxies_inquirer.proxy_to_address.get(addr)
```

My concern is that you are adding extra things that complicated the logic

Comment (yabirgb) [reply rotkehlchen/chain/ethereum/modules/liquity/trove.py:283]: Also I mention it because

```python
from typing import TypedDict


class A(TypedDict):
    name: str

def prone() -> A:
    return {
        'name': 's'
    }
```

is valid

yabirgb (commented):
Not sure about the requirements that luki had. For example if he tries to filter the deposited balances by an address or the proxy address will it work or does he filter them in the frontend?

docs/api.rst
```diff
@@ -7434,15 +7434,29 @@ Getting Liquity staked amount
...
+              "lusd_rewards": {
+                "asset": "eip155:1/erc20:0x6DEA81C8171D0bA574754EF6F8b412F2Ed88c54D",
+                "amount": "94477.70111867384658505",
+                "usd_value": "94477.70111867384658505",
+              },
+              "eth_rewards": {
+                "asset": "eip155:1/erc20:0x5f98805A4E8be255a32880FDeC7F6728C6568bA0",
+                "amount": "10211401.723115634393264567",
+                "usd_value": "10211401.723115634393264567",
+              },
```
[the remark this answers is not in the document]
prettyirrelevant: yeah, the docs contained just the balance dict.

rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -226,13 +226,20 @@ def _query_deposits_and_rewards(
                     usd_value=asset_price * amount,
                 ),
             )
+            # `proxy_to` is not accessed from any other location.
+            data[current_address]['proxy_to'] = self.ethereum.proxies_inquirer.proxy_to_address.get(current_address)  # type: ignore  # noqa: E501
```
yabirgb, line 230: But this is a really bad solution. You are ignoring the type and that should let you know that something is wrong. This function returns a mapping of address to assetbalance and you are breaking that constraint with this change. Basically you made useless all the typing for functions that call this one.

What you can do and I find better is in the two functions that use this one (`get_stability_pool_balances` and `liquity_staking_balances`) change the type to 

```python
dict[ChecksumEvmAddress, dict[str, Any]]
```

and in there do:

```python
for address in addresses:
    staking_data['proxy_to`] = self.ethereum.proxies_inquirer.proxy_to_address.get(current_address)
```

Why I find this better than your first approach:

1. You are not initializing new objects that you don't use
2. You are not reading and settings properties that you don't need

Why I see it better than this approach:

1. You don't break the types everywhere

rotkehlchen/tests/api/test_liquity.py
```diff
@@ -270,3 +270,21 @@ def test_staking_stats(rotkehlchen_api_server, ethereum_accounts):
...
+def test_proxy_info_is_shown(rotkehlchen_api_server, ethereum_accounts):
+    """Check that `proxy_to` is not None for proxy addresses."""
+    user_address = ethereum_accounts[0]
+    response = requests.get(api_url_for(
+        rotkehlchen_api_server,
+        'liquitystabilitypoolresource',
+    ), json={'async_query': False})
+
+    result = assert_proper_response_with_result(response)
+    assert not result[user_address]['proxy_to']  # since this address is the proxy owner
```
yabirgb, line 289: Can you check instead

```suggestion
    assert result[user_address]['proxy_to'] is None # since this address is the proxy owner
```

rotkehlchen/tests/api/test_liquity.py
```diff
@@ -270,3 +270,21 @@ def test_staking_stats(rotkehlchen_api_server, ethereum_accounts):
...
+    assert result['0x33EAfDB72b69BFBe6b5911eDCbab41011e63C523']['proxy_to'] == user_address
```
yabirgb, line 290: btw what is better. `proxy_to` or `proxy_of`? Proxy to sounds weird to me but not sure 100% if is wrong

rotkehlchen/tests/api/test_liquity.py
```diff
@@ -270,3 +270,21 @@ def test_staking_stats(rotkehlchen_api_server, ethereum_accounts):
...
     assert address_1_data['stability_pool_gains'][1]['amount'] == '4240.34942308358'
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('ethereum_accounts', [['0xbB8311c7bAD518f0D8f907Cad26c5CcC85a06dC4']])
+@pytest.mark.parametrize('ethereum_modules', [['liquity']])
+@pytest.mark.parametrize('start_with_valid_premium', [True])
+@pytest.mark.parametrize('should_mock_current_price_queries', [True])
+def test_proxy_info_is_shown(rotkehlchen_api_server, ethereum_accounts):
+    """Check that `proxy_to` is not None for proxy addresses."""
```
yabirgb, line 281:
```suggestion
def test_proxy_info_is_shown(rotkehlchen_api_server, ethereum_accounts):
    """Check that information about proxies is added to the responses for liquity endpoints"""
```

rotkehlchen/tests/api/test_liquity.py
```diff
@@ -270,3 +270,21 @@ def test_staking_stats(rotkehlchen_api_server, ethereum_accounts):
...
+    user_address = ethereum_accounts[0]
+    response = requests.get(api_url_for(
+        rotkehlchen_api_server,
+        'liquitystabilitypoolresource',
```
yabirgb, line 285: shouldn't you check also the other endpoint?

rotkehlchen/tests/api/test_liquity.py
```diff
@@ -270,3 +270,21 @@ def test_staking_stats(rotkehlchen_api_server, ethereum_accounts):
     assert address_0_data['stability_pool_gains'][0]['amount'] == '1.7820064710306824'
     assert address_1_data['stability_pool_gains'][0]['amount'] == '12.294706987216218'
     assert address_1_data['stability_pool_gains'][1]['amount'] == '4240.34942308358'
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('ethereum_accounts', [['0xbB8311c7bAD518f0D8f907Cad26c5CcC85a06dC4']])
```
yabirgb, line 276: I told you that another test would probably fail. have you checked it? https://github.com/rotki/rotki/actions/runs/4416177427/jobs/7740251558#step:10:1320

yabirgb (commented):
A few things left

rotkehlchen/chain/ethereum/modules/liquity/trove.py
prettyirrelevant: > dict[ChecksumEvmAddress, dict[str, Any]]

This is no different from adding a `type: ignore` and as a result the assertions made above as regards rendering typing redundant also holds.

The only +ve to this is that wherever the function is called, you won't expect the key-value pairs to have the same types.

yabirgb: Yes, for the linter is no different but for the reader now it won't expect it to be a mapping for balances

rotkehlchen/tests/api/test_liquity.py
prettyirrelevant: I'd just go with `proxy_of`

LefterisJP (commented):
As discussed in discord not reviewing this due to failed tests. Changed the label

prettyirrelevant:
Faiiling tests are fixed in bugfixes

rotkehlchen/chain/aggregator.py
```diff
@@ -1014,18 +1014,28 @@ def _add_eth_protocol_balances(self, eth_balances: DefaultDict[ChecksumEvmAddres
...
-                if staked_balance.amount > ZERO:
-                    eth_balances[address].assets[A_LUSD] += staked_balance
+                pool_balance = staked_info['balances']['deposited'].balance
+                if pool_balance.amount > ZERO:
+                    eth_balances[address].assets[A_LUSD] += pool_balance
+
+                for proxy_address, proxy_balance in staked_info['proxies'].items():
+                    pool_balance = proxy_balance['deposited'].balance
+                    if pool_balance.amount > ZERO:
+                        eth_balances[proxy_address].assets[A_LUSD] += pool_balance
```
LefterisJP, line 1038: Hmmm are you sure this is the correct approach? Here and in line 1024? Afaik, the way we do it with proxies **at the moment** is to add the balances directly to the address that owns the proxy. Otherwise, the way you do it right now, there is gonna be eth balances for an address you don't track and I am not sure this is gonna work with the way the frontend works.

rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -194,7 +199,12 @@ def _query_deposits_and_rewards(
 
         # the structure of the queried data is:
         # staked address 1, reward 1 of address 1, reward 2 of address 1, staked address 2, reward 1 of address 2, ...  # noqa: E501
-        data: DefaultDict[ChecksumEvmAddress, dict[str, AssetBalance]] = defaultdict(dict)
+        data: DefaultDict[ChecksumEvmAddress, LiquityBalanceWithProxy] = defaultdict(
+            lambda: {
+                'balances': defaultdict(AssetBalance),  # type: ignore[arg-type]
```
LefterisJP, line 204: Why the type ignore here? Due to defaultdict and not dict?

LefterisJP (commented):
A question on the approach here.

rotkehlchen/chain/aggregator.py
prettyirrelevant: While iterating over the PR, I initially went with adding the proxies balances under the owner's address... can't recollect why I changed to this.

Will switch to using the owner's address. Thanks

LefterisJP (commented):
Tests failed

LefterisJP (approved):
LGTM! Thank you Isaac

PR #5766, opened 2023-03-14
Fix uniswap v2 swap decoding edge case

### LINKED ISSUE #5722: Approval event in swaps
## Problem Definition

An approval event in between of the spend and the receive event in a uniswap v2 swap, which causes wrong numbers in PnL.

## Spec

- [x] Fix the particular case reported by Sebastian.
image

- [x] Add a test for that case
- [ ] Check other swaps decoders and make sure that are no approval events can appear there in between spend and receive. 

Operating system: 
Rotki version: 1.27.1

prettyirrelevant:
Fix bug in #5722

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`5572` Fix a detection error that caused some uniswap v2 trade edge cases not to be accounted for properly in pnl report.
 * :feature:`5592` Either include fees in the bought assets price or treat them as standalone events depending on the accounting settings.
 * :feature:`-` For known protocols like yearn, curve, uniswap etc, if the on-chain price query fails, external oracles will still be queried in case something is found there.
 * :bug:`-` Yearn vault price queries will now work more robustly. If the underlying token is not in rotki's DB it will be queried from the chain.
--- a/rotkehlchen/tests/unit/decoders/test_uniswapv2.py
+++ b/rotkehlchen/tests/unit/decoders/test_uniswapv2.py
@@ -28,6 +28,7 @@ from rotkehlchen.utils.hexbytes import hexstring_to_bytes
 ADDY_1 = string_to_evm_address('0x3CAdf2cA458376a6a5feA2EF3612346037D5A787')
 ADDY_2 = string_to_evm_address('0xCC917Ab28544c80E2f0e8efFbd22551A3cB096bE')
 ADDY_3 = string_to_evm_address('0x65fc65C639467423Bf19801a59FCfd62f0F29777')
+ADDY_4 = string_to_evm_address('0x43e141534d718D72552De1B606a5FCBc72256cD7')
 
 
 @pytest.mark.parametrize('ethereum_accounts', [[ADDY_1]])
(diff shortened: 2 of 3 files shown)


rotkehlchen/chain/ethereum/modules/uniswap/v2/decoder.py
```diff
@@ -1,6 +1,7 @@
 from typing import Callable, Optional
 
 from rotkehlchen.accounting.structures.base import HistoryBaseEntry
+from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
```
LefterisJP, line 4: > Now, you might begin to wonder, "why doesn't a bugfix contain a test?"... A test actually does exist however, the user doesn't want to be doxxed and multiple attempts at finding transactions that match this edge case have been futile.
> 
> As a result of this, whoever will be reviewing this PR I'll share a diff that contains the test. Sorry for any incoveniences
> 
> TLDR: Closes #5722

Please use the data you have and add the test in the codebase without using VCR and hard-coding the transaction and its log events, but changing hashes and addresses.

I won't even look at a PR for a bugfix without a test.

rotkehlchen/chain/ethereum/modules/uniswap/v2/decoder.py
```diff
@@ -144,6 +146,42 @@ def _maybe_enrich_lp_tokens_transfers(
             lp_token_symbol=UNISWAP_PROTOCOL,
         )
 
+    def _fix_order_of_events_for_swaps(
+            self,
+            transaction: EvmTransaction,  # pylint: disable=unused-argument
+            decoded_events: list[HistoryBaseEntry],
+            all_logs: list[EvmTxReceiptLog],  # pylint: disable=unused-argument
+    ) -> list[HistoryBaseEntry]:
+        # look for the swap send & receive
```
yabirgb, line 155: Please add a docstring to the function

rotkehlchen/tests/unit/decoders/test_uniswapv2.py
```diff
@@ -747,3 +748,196 @@ def test_uniswap_v2_remove_liquidity(database, ethereum_inquirer, eth_transactio
...
+
+@pytest.mark.parametrize('ethereum_accounts', [[ADDY_4]])
+def test_uniswap_v2_swap_events_order(
+        database,
+        ethereum_inquirer,
+        eth_transactions,
+        ethereum_accounts,
+):
+    """Check that the order of swap events are consecutive."""
+    tx_hash = '0x75b4d35b14f435c60e0fb9a82a70457a213c91ef829c034c743c66bec2dcd40f'
```
yabirgb, line 761: Can you state that the transaction hash here has been randomly selected to protect the transaction of the user? We will at some point look at this transaction, see that it doesn't match and wonder why

rotkehlchen/chain/ethereum/modules/uniswap/v2/decoder.py
```diff
@@ -144,6 +146,42 @@ def _maybe_enrich_lp_tokens_transfers(
...
+        send_event_idx, receive_event_idx = None, None
+        for idx, event in enumerate(decoded_events):
+            if (
+                event.event_type == HistoryEventType.TRADE and
+                event.event_subtype == HistoryEventSubType.SPEND and
+                event.counterparty == CPT_UNISWAP_V2
```
yabirgb, line 161: if this happened with uniswap v2 can also happen with sushiswap

rotkehlchen/chain/ethereum/modules/uniswap/v2/decoder.py
```diff
@@ -157,5 +195,10 @@ def enricher_rules(self) -> list[Callable]:
             self._maybe_enrich_lp_tokens_transfers,
         ]
 
+    def post_decoding_rules(self) -> list[tuple[int, Callable]]:
+        return [
+            (0, self._fix_order_of_events_for_swaps),
```
yabirgb, line 200: As has been commented in other PR this will execute for all the transactions and is costly. We will refactor the post decoding rule to gain context of the transaction and not execute them always

rotkehlchen/tests/unit/decoders/test_uniswapv2.py
```diff
@@ -747,3 +748,196 @@ def test_uniswap_v2_remove_liquidity(database, ethereum_inquirer, eth_transactio
...
+
+
+@pytest.mark.parametrize('ethereum_accounts', [[ADDY_4]])
+def test_uniswap_v2_swap_events_order(
+        database,
+        ethereum_inquirer,
+        eth_transactions,
+        ethereum_accounts,
+):
+    """Check that the order of swap events are consecutive."""
```
yabirgb, line 760: Please add a description of the test case because we will read it again and won't know why this test it here. Mention that it handles the case of an approval in between of the trade

rotkehlchen/chain/ethereum/modules/uniswap/v2/decoder.py
```diff
@@ -144,6 +146,42 @@ def _maybe_enrich_lp_tokens_transfers(
...
+        if (
+            send_event_idx is not None and
+            receive_event_idx is not None and
+            receive_event_idx != send_event_idx + 1
+        ):
+            maybe_reshuffle_events(
+                out_event=decoded_events[send_event_idx],
+                in_event=decoded_events[receive_event_idx],
+                events_list=decoded_events,
+            )
```
yabirgb, line 181: This looks to me that could happen in other cases too. You have the transaction, is it using a router or something special? if that is the case maybe uniswap v3 could also have this

yabirgb (commented):
Some comments and also this is missing the changelog entry

rotkehlchen/chain/ethereum/modules/uniswap/v2/decoder.py
prettyirrelevant: It was via the router.

rotkehlchen/chain/ethereum/modules/sushiswap/decoder.py
```diff
@@ -112,6 +124,11 @@ def decoding_rules(self) -> list[Callable]:
             self._maybe_decode_v2_liquidity_addition_and_removal,
         ]
 
+    def post_decoding_rules(self) -> list[tuple[int, Callable]]:
+        return [
+            (0, self._fix_order_of_events_for_swaps),
+        ]
```
yabirgb, line 130: There is no need to add this post decoding rule for all three. You can check in the function that if the counterparty is any of the three we said make sure that the swaps are consecutive. This way a transactions will loop thrice trough this logic for no reason

rotkehlchen/chain/ethereum/modules/uniswap/v2/decoder.py
yabirgb: if is a thing of the router isn't possible to filter the transactions for which this function runs using the router address? We have `UNISWAP_ROUTERS`... isn't possible to do

```
if transaction.to in UNISWAP_ROUTERS:
    return
```

?

This will prevent the logic from executing for every transaction

rotkehlchen/chain/ethereum/modules/sushiswap/decoder.py
```diff
@@ -104,6 +105,17 @@ def _maybe_enrich_lp_tokens_transfers(
             lp_token_symbol=SUSHISWAP_PROTOCOL,
         )
 
+    def _fix_order_of_events_for_swaps(
+            self,
+            transaction: EvmTransaction,  # pylint: disable=unused-argument
```
yabirgb, line 110: Adding it as a comment since you forgot. This needs a changelog entry

yabirgb (commented):
Few comments in case there are room for improvement

rotkehlchen/chain/ethereum/modules/uniswap/v2/decoder.py
prettyirrelevant: `UNISWAP_ROUTERS` is for AUTO routers but this is via normal uniswap v2 routers

yabirgb:
The failing test is due the test-caching repo branch not being up to date

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`5572` Trade events in Uniswap V2 swaps now occur in a consecutive order.
```
LefterisJP, line 5: This is not a good description as it tells nothing to the user and also implies all events were wrong.

```suggestion
* :bug:`5572` Fix a detection error that caused some uniswap v2 trade edge cases not to be accounted for properly in pnl report.
```

LefterisJP (approved):
LGTM

PR #5769, opened 2023-03-15
Updates build runner to ubuntu 20.04

### LINKED ISSUE #4746: Prepare for ubuntu 18.04 runner deprecation
## Abstract

The ubuntu 18.04 runner is going to be deprecated (https://github.com/actions/runner-images/issues/6002)

The suggestion is to move to a newer version of the runner. We can easily do that for our regular CI but for packaging, a newer version also means a newer glibc version that would affect anyone running older distro versions.

## Possible approaches

- Use official python builds from python.org, similar to how we got around the mac situation.
- Use docker for the backend build. As with the wheels build, we use the [manylinux](https://github.com/pypa/manylinux) docker images to create compatible binaries.

kelsos:
Closes #4746 

## Checklist

> **Note** We might need to cherry-pick this to bugfixes after the end of March

The deprecation is in two weeks and they have at least two more 24h brownouts planned

- March 21, 00.00 UTC – March 22, 00.00 UTC
- March 28, 00.00 UTC – March 29, 00.00 UTC

### DIFF
--- a/.github/workflows/rotki_dev_builds.yml
+++ b/.github/workflows/rotki_dev_builds.yml
@@ -20,7 +20,7 @@ jobs:
     env:
       CI: true
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


LefterisJP (approved):
This looks good to me, because Kelsos told me to merge. I hold no responsibility for what happens during the nightly runs in the CI. Blame Kelsos. It's all gonna be his fault.

PR #5775, opened 2023-03-16
Fix two failing prices tests

prettyirrelevant:
This PR fixes the two tests by mocking the coingecko API responses.

1. `rotkehlchen/tests/api/test_current_assets_price.py::test_inquirer_oracles_affect_manual_price`
2. `rotkehlchen/tests/api/test_current_assets_price.py::test_manual_current_prices_loop`

### DIFF
--- a/rotkehlchen/tests/api/test_current_assets_price.py
+++ b/rotkehlchen/tests/api/test_current_assets_price.py
@@ -282,10 +282,15 @@ def test_remove_manual_current_price(rotkehlchen_api_server):
     assert (A_BTC, A_ETH) not in Inquirer()._cached_current_price
 
 
+@pytest.mark.vcr()
 @pytest.mark.parametrize('should_mock_current_price_queries', [False])
 def test_manual_current_prices_loop(inquirer):
     """Check that if we got a loop of manual current prices
-    (e.g. 1 ETH costs 2 BTC and 1 BTC costs 5 ETH), it is handled properly."""
+    (e.g. 1 ETH costs 2 BTC and 1 BTC costs 5 ETH), it is handled properly.
+
+    This test is mocked because we were seeing cases of tests failing due to
+    an exception from coingecko API.
+    """
     GlobalDBHandler().add_manual_latest_price(
         from_asset=A_ETH,
         to_asset=A_USD,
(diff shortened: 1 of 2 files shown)


rotkehlchen/tests/unit/decoders/test_compound.py
```diff
@@ -16,7 +16,7 @@
 ADDY3 = '0xb99CC7e10Fe0Acc68C50C7829F473d81e23249cc'
 
 
-@pytest.mark.vcr()
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
```
nebolax, line 19: I am not sure if the apikey is the source of problem here. Are you sure about it?

rotkehlchen/tests/api/test_current_assets_price.py
```diff
@@ -282,10 +282,15 @@ def test_remove_manual_current_price(rotkehlchen_api_server):
     assert (A_BTC, A_ETH) not in Inquirer()._cached_current_price
 
 
+@pytest.mark.vcr()
 @pytest.mark.parametrize('should_mock_current_price_queries', [False])
 def test_manual_current_prices_loop(inquirer):
     """Check that if we got a loop of manual current prices
```
nebolax, line 288: For this test did you also check that the warnings are correct?

For example from this output from https://github.com/rotki/rotki/actions/runs/4435118738/jobs/7781949206

```
FAILED rotkehlchen/tests/api/test_current_assets_price.py::test_manual_current_prices_loop[False] - AssertionError: assert 2 == 1
 +  where 2 = len(['Was not able to find price from ETH(Ethereum) to EUR(Euro) since your manual latest prices form a loop. For now, oth...(United States Dollar) to EUR(Euro) since your manual latest prices form a loop. For now, other oracles will be used.'])
```

It looks like there are 2 warnings about loops, which I think is not correct. I think there should be one warning about loops and one warning about inability to query price.

rotkehlchen/tests/unit/decoders/test_compound.py
prettyirrelevant: Initially this change wasn't part of the PR. But after running the CI, the this particular test failed... ran it locally and it also failed.

Then checked the same test in develop and it was having this `filter_query_parameters=['apikey']`

### Develop
https://github.com/prettyirrelevant/rotki/blob/eb8de3d4f0b93251eda8ea7ec4a0ce73fcc77864/rotkehlchen/tests/unit/decoders/test_compound.py#L29

### Bugfixes
https://github.com/prettyirrelevant/rotki/blob/a61f978d3a2e2780c8bd3b88e59f5a92cc259260/rotkehlchen/tests/unit/decoders/test_compound.py#L19

nebolax: Ouff alright then. How did this discrepancy appear thought? :thinking:

@yabirgb pinging you since you seem to be the author of the compound tests

rotkehlchen/tests/api/test_current_assets_price.py
prettyirrelevant: So the reason for the two warnings is:
We call 
```python
price = inquirer.find_price(
        from_asset=A_ETH.resolve_to_asset_with_oracles(),
        to_asset=A_EUR.resolve_to_asset_with_oracles(),
)
```
with the following oracles = [Manual, Coingecko, Defillama, Cryptocompare]
---> tries the first oracle(manual), enters recursion & adds the first warning...
---> up next coingecko, `RemoteError` occurs, next oracle...
---> tries defillama, it works we get `ETH -> USD`... from the defillama's `query_current_price` method, `Inquirer().find_usd_price()` is called to get `USD->EUR`
--------> calls the first oracle(manual) again, enters recursion & adds the second warning...

LefterisJP (approved):
LGTM

PR #5777, opened 2023-03-16
Cracefully shutdown docker containers

### LINKED ISSUE #5757: Make sure that Docker image can properly shutdown
## Problem Definition

While trying to set up the proxy I noticed that the proxy didn't exit cleanly. 
Instead, it exited with (code=137). 

This might be because there is an infinite loop (intentional) monitoring the subprocesses, and this loop only exits if one of the subprocesses stops running.

This will cause the script to exit with code 1.

https://github.com/rotki/rotki/blob/53d36bc1f5aa0ff64fcfaf21f0ba6ca005a032eb/packaging/docker/entrypoint.py#L173

## Requirements

- Handle `SIGINT`/`SIGTERM` normally and exit cleanly (code=0)
- Make sure that if the subprocesses die we exit (code=1)

nebolax:
Closes #5757

Gracefully shutdown docker containers on `SIGINT`, `SIGTERM` and `SIGQUIT`.

Changes made:

1. Changed `CMD ["sh", "-c", "/opt/rotki/entrypoint.py"]` to `CMD["/opt/rotki/entrypoint.py"]` because `sh -c` doesn't propagate signals. Actually not sure what was the reason for using `sh -c`? If I am not mistaken, `CMD` command in Docker runs the same shell.
2. In addition to `SIGINT` and `SIGTERM` mentioned in the issue also added handling of `SIGQUIT` because for some reason in our Dockerfile default stop signal (the signal sent on `docker stop`) is `SIGQUIT` instead of common `SIGTERM`. I tried to find the step at which this override happens, but couldn't find it quickly. The other approach for handling `SIGQUIT` on `docker stop` though would be to explicitly set `STOPSIGNAL SIGTERM` in Dockerfile, but I am not sure whether this is better than handling it in python code.
3. Added a handler in `entrypoint.py` to handle mentioned signals and shut down gracefully, killing `rotki` and `nginx` before exiting.


kelsos:
I tested the PR I don't see anything broken, (environment variables still work fine) and the container now exits normally when stop it and without a delay

packaging/docker/entrypoint.py
```diff
@@ -170,6 +171,29 @@ def load_config() -> List[str]:
     logger.error('Failed to start nginx')
     exit(1)
 
+def graceful_exit(signal, frame):
+    logger.info(f'Received signal {signal}. Exiting gracefully')
+    if rotki.poll() is None:
+        rotki.send_signal(SIGTERM)
```
LefterisJP, line 177: Why work with `send_signal` and not use `process.terminate`? For unix it's the same (according to docs), but for windows it uses windows specific code as those signals are unix specific.

Though since this is docker, it would only run in unix, right?

Also after `send_signal()` shouldn't you wait somewhere that the signal is handled? Or does `send_signal()` block?

nebolax: Yes, I checked the underlying code and is SIGTERM is provided it calls `.terminate`, so using `.terminate` probably makes more sense.

> Though since this is docker, it would only run in unix, right?

Yeah, this is also true, but I changed it to `.terminate` anyways.

> Also after send_signal() shouldn't you wait somewhere that the signal is handled? Or does send_signal() block?

Thanks for attracting attention to this! Originally I did a quick check and decided that send_signal is synchronous, but now checked once again and see that it is not synchronous. Will fix this.

packaging/docker/entrypoint.py
```diff
@@ -170,6 +171,30 @@ def load_config() -> List[str]:
...
 
+
+def terminate_process(process_name: str, process: subprocess.Popen) -> None:
+    logger.info(f'Terminating {process_name}')
+    if process.poll() is not None:
+        logger.error(f'{process_name} was not running. This means that some error occurred.')
+        exit(1)
+
+    process.terminate()
+    process.wait()  # wait untill the process terminates
```
LefterisJP, line 182:
```suggestion
    process.wait()  # wait until the process terminates
```

LefterisJP (approved):
LGTM

PR #5778, opened 2023-03-17
feat: binance saving breakdown

### LINKED ISSUE #1505: Take Binance Lending/Savings into account for profit calculation

lukicenturi:
Closes #1505

## Checklist

image


.pnpm-debug.log
kelsos, line 1: we can update .gitignore to add this file to ignore I guess 😄

kelsos:
@lukicenturi also a question does it have elevation in the UI in the screenshot? if so can we maybe make it flat?

lukicenturi:
> @lukicenturi also a question does it have elevation in the UI in the screenshot? if so can we maybe make it flat?

oh yeah didn't notice that, I think we can make it flat

kelsos:
Ok so for this PR I guess what we need to do is update the `events` => `entries` also for kraken staking (and maybe can you unify the mapper)

frontend/app/src/pages/balances/exchange/index.vue
```diff
@@ -172,7 +227,11 @@ const { t, tc } = useI18n();
 </template>
 
 <style scoped lang="scss">
-:deep(.v-tabs-bar) {
+/* stylelint-disable selector-class-pattern */
+
+:deep(.exchange-balances__tabs) {
```
kelsos, line 232: This is our own class right? why use deep? is it in another component?

frontend/app/src/store/balances/exchanges.ts
```diff
@@ -341,6 +352,90 @@ export const useExchangeBalancesStore = defineStore(
...
+      setStatus(Status.REFRESHING);
+
+      try {
+        const locations = get(connectedExchanges)
+          .map(({ location }) => location)
+          .filter(uniqueStrings)
+          .filter(
+            location =>
+              location === SupportedExchange.BINANCE ||
+              location === SupportedExchange.BINANCEUS
```
kelsos, line 424: nitpick: do you think we can do a [b, us].includes(location) instead?

frontend/app/src/types/collection.ts
```diff
@@ -1,4 +1,12 @@
-import { type BigNumber } from '@rotki/common';
+import { type BigNumber, NumericString } from '@rotki/common';
+import { z } from 'zod';
+
+export const CollectionGenericType = z.object({
```
kelsos, line 4: nitpick: since generic usually means something else in the language context maybe CollectionCommonFields would be less confusing what do you think?

frontend/app/src/store/balances/exchanges.ts
```diff
@@ -341,6 +352,90 @@ export const useExchangeBalancesStore = defineStore(
...
+        ascending: [false],
+        orderByAttributes: ['timestamp'],
+        onlyCache: false,
+        location
+      };
+
+      const { taskId } = await getExchangeSavingsTask(defaults);
+      const taskMeta = {
+        title: 'test1',
+        message: 'test2'
```
kelsos, line 381: 😅

frontend/app/src/store/balances/exchanges.ts
```diff
@@ -341,6 +352,90 @@ export const useExchangeBalancesStore = defineStore(
...
+          taskId,
+          taskType,
+          taskMeta,
+          true
+        );
+        return true;
+      } catch {
+        notify({
+          title: 'notify1',
+          message: 'notify2',
```
kelsos, line 395: 😅

frontend/app/src/utils/collection.ts
```diff
@@ -94,3 +98,15 @@ export const defaultOptions = <T extends Object>(
...
+  response: ExchangeSavingsCollectionResponse
+): ExchangeSavingsCollection => {
+  return {
+    found: response.entriesFound,
+    limit: response.entriesLimit,
+    total: response.entriesTotal,
+    data: response.entries,
+    ...response
+  };
+};
```
kelsos, line 112: Any reason we can't merge this with the other function above?
It should be possible to just use a single function for all cases

frontend/app/src/store/balances/exchanges.ts
lukicenturi: hardest part of making PR's is write correct sentences in English 😛

frontend/app/src/pages/balances/exchange/index.vue
```diff
@@ -60,15 +67,41 @@ const openExchangeDetails = async () => {
...
+
+onMounted(() => {
+  refreshExchangeSavings();
+});
+
+const isBinance = computed(() => {
+  const exchangeVal = get(exchange);
+  return (
+    exchangeVal === SupportedExchange.BINANCE ||
+    exchangeVal === SupportedExchange.BINANCEUS
```
kelsos, line 91: nitpick: we could also do the [].includes thing here

frontend/app/src/utils/collection.ts
kelsos: btw if you merge them maybe do a `const { entriesFound, entriesLimit, entriesTotal, entries, ...rest } = x`

and then return `{ found: entriesFound, ..., ...rest }`

because this was the response there would still pass all the properties of the original.

frontend/app/src/pages/balances/exchange/index.vue
lukicenturi: I'm targeting the tab component. Anyway it's moved now

frontend/app/src/utils/collection.ts
```diff
@@ -1,18 +1,24 @@
...
     limit: response.entriesLimit,
     total: response.entriesTotal,
-    totalUsdValue: response.totalUsdValue
+    ...omit(response, [
+      'entries',
+      'entriesFound',
+      'entriesLimit',
+      'entriesTotal'
+    ])
   };
```
kelsos, line 22: I am gonna adjust this a bit and make sure that there is an explicit return type.

kelsos (approved):
thanks @lukicenturi lgtm

PR #5787, opened 2023-03-20
feat: expose scramble multiplier to premium components

### LINKED ISSUE #5780: Change statistic page scrambler algorithm
## Problem Definition

For the statistics page, we can use the scramble to not show the real value,
But the way it works now is by randomizing the value.
Instead, we can show it better by multiplying the value with the random multiplier we have

lukicenturi:
Closes #5780

## Checklist

### DIFF
--- a/frontend/common/src/defi/balancer/index.ts
+++ b/frontend/common/src/defi/balancer/index.ts
@@ -1,6 +1,5 @@
-import { type default as BigNumber } from 'bignumber.js';
 import { z } from 'zod';
-import { Balance, NumericString } from '../../index';
+import { Balance, type BigNumber, NumericString } from '../../index';
 import { XswapPool } from '../xswap';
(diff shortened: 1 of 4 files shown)


frontend/common/src/defi/balancer/index.ts
```diff
@@ -1,4 +1,4 @@
-import { type default as BigNumber } from 'bignumber.js';
+import { type BigNumber } from 'bignumber.js';
```
kelsos, line 1: This seems like accidental do you think we can change this to import from the index re-export?

kelsos (commented):
Can you also bump the premium API version too?

While we will have a major version update for the next release
It would be still good to note this as an API version update

kelsos (approved):
lgtm @lukicenturi thanks

PR #5790, opened 2023-03-20
Query and cache ENS avatars

### LINKED ISSUE #5622: Show ENS avatars for addresses instead of blockies where possible
## Abstract

At the moment in rotki we are showing the blockies icons for each evm address. Also if it's a primary ENS name we show its name.

Fos7Zd3X0AAzv4r

This is nice but it would be a much more personal touch and give a social spin to rotki if where both an avatar and a primary ens name is set, then instead of the blockies to show the avatar.

### Backend

- [ ] Fetch avatars for evm addresses that have primary ens name and avatars set. We should probably have a local cache of those in a directory to not have to make remote network queries all the time. We assume a user's tracked accounts, and all the accounts it interacts with will have very limited amount with avatars set. That cache should be invalidated every `X` hours since avatars can change
- [ ] Add the avatar existence information in the api, perhaps in the exact same endpoing as the reverse ens name check. But we shoudl discuss with @rotki/frontend about this

### Frontend

- [ ] When an evm account appears, if it has a primary ENS name (reverse resolution) set AND an avatar set, then instead of the blockie, show that avatar.

nebolax:
Closes #5622

## Checklist


prettyirrelevant (commented):
Thinking out loud here @nebolax.
Why not go with the same approach icons follow currently -- the use of md5 of the avatar as an etag.

Also might be an overkill but why not have a periodic task query the `ens_mappings` table and check if avatars have a new ens or not?

What do you think about this approach?

rotkehlchen/chain/ethereum/node_inquirer.py
```diff
@@ -96,6 +96,18 @@ def __init__(
         self.etherscan = cast(EthereumEtherscan, self.etherscan)
         self.ens_reverse_records = self.contracts.contract('ENS_REVERSE_RECORDS')
 
+    def ens_avatar_lookup(self, ens_name: str) -> Optional[str]:
+        """
+        Checks whether given ens name has an ens avatar set.
+        If there is avatar set -- returns the url of the avatar.
+        """
```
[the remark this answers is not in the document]
nebolax: Thanks. Will add a note about the remote error, however I don't think we should return `None` in case of a remote error :thinking:

If we do it, then a remote error will be equal to non-existence of an avatar, which is not correct.

nebolax:
> Also might be an overkill but why not have a periodic task query the ens_mappings table and check if avatars have a new ens or not?

Maybe ... but not sure. I would say that this would be useful if many ens avatars would be expiring at the same moment (and thus we could re-query avatars of many ens names at the same time), but I don't think this is our case, so not sure about it.

nebolax:
> Why not go with the same approach icons follow currently -- the use of md5 of the avatar as an etag.

Yes, this probably makes sense, will do it, thanks.

rotkehlchen/api/v1/resources.py
```diff
@@ -2835,3 +2836,18 @@ def put(
...
+
+    @require_loggedin_user()
+    @use_kwargs(get_schema, location='json_and_query')
+    def get(self, ens_name: str) -> Response:
+        # Process the if-match and if-none-match headers so that comparison with etag can be done
+        match_header = flask_request.headers.get('If-Match', None)
+        if not match_header:
+            match_header = flask_request.headers.get('If-None-Match', None)
+        if match_header:
+            match_header = match_header[1:-1]  # remove enclosing quotes
```
prettyirrelevant, line 2852: since this block of code is repeated twice... why not just make it a decorator and reuse in both locations?
What do you think?

rotkehlchen/api/rest.py
```diff
@@ -4197,3 +4201,61 @@ def get_binance_savings_history(
...
         )
+
+    def get_ens_avatar(self, ens_name: str, match_header: Optional[str]) -> Response:
+        """
+        Searches for ENS avatar of the given `ens_name`.
+        If found returns the avatar, otherwise returns 404 status code.
+
+        Also supports etag mechanism that helps with caching on the API caller side.
+        """
+        avatar_path = self.rotkehlchen.data_dir / 'avatars' / f'{ens_name}.png'
```
prettyirrelevant, line 4212: Since there's a `DBEns` class, why not move this logic into that class?
That way you'd only have to return the bytes data of the image.

Also, the return of an image is also repeated twice...
I'm proposing something like...
```python
return api_image_response(image_retrieval_func)
```

The `image_retrieval_func` can return `tuple[Optional[bytes], str]` where the `str` is the error message.
You might have better ideas here

rotkehlchen/api/rest.py
```diff
@@ -4197,3 +4201,61 @@ def get_binance_savings_history(
...
+        if last_update is None or ts_now() - last_update > ENS_AVATARS_REFRESH:
+            # If avatar for this ens name has never been checked or the avatar has expired
+            # then we try to download.
+            try_download_ens_avatar(
+                eth_inquirer=self.rotkehlchen.chains_aggregator.ethereum.node_inquirer,
+                avatars_dir=Path(self.rotkehlchen.data_dir) / 'avatars',
+                ens_name=ens_name,
+            )
+
+        if avatar_path.exists() is False:  # No avatar found, return 404
```
prettyirrelevant, line 4244: nitpick: maybe use `.is_file()` instead
```suggestion
        if avatar_path.is_file() is False:  # No avatar found, return 404
```

rotkehlchen/api/rest.py
```diff
@@ -4197,3 +4201,61 @@ def get_binance_savings_history(
...
+            return api_response(
+                result={'result': None, 'message': f'Ens avatar for {ens_name} was not found'},
+                status_code=HTTPStatus.NOT_FOUND,
+            )
+        else:  # There is an avatar!
```
prettyirrelevant, line 4249: This else check is not needed here since there's a return above

rotkehlchen/chain/ethereum/node_inquirer.py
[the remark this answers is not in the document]
prettyirrelevant: You could log an error...
However, not handling `RemoteError` here is synonymous to returning `None`.

Or maybe handle the `RemoteError` where the method is used.

rotkehlchen/chain/ethereum/utils.py
```diff
@@ -166,3 +173,26 @@ def should_update_protocol_cache(cache_key: GeneralCacheType, *args: str) -> boo
...
+    2. If it does -- downloads the avatar and saves it to the provided `dir`
+    3. Updates last avatar check timestamp for the given ens name
+    """
+    dbens = DBEns(eth_inquirer.database)
+    avatar_url = eth_inquirer.ens_avatar_lookup(ens_name)
+    dbens.set_last_avatar_update(ens_name=ens_name, timestamp=ts_now())
+    if avatar_url is None:
+        return  # Avatar is not set
+
+    avatar = requests.get(avatar_url, timeout=10).content
```
prettyirrelevant, line 195: this can raise `RequestException`

rotkehlchen/api/v1/fields.py
```diff
@@ -864,3 +864,20 @@ def _deserialize(
...
+class EnsNameField(fields.String):
+    def _deserialize(
+            self,
+            value: str,
+            attr: Optional[str],  # pylint: disable=unused-argument
+            data: Optional[Mapping[str, Any]],
+            **_kwargs: Any,
+    ) -> str:
+        if not isinstance(value, str):
+            raise ValidationError('Ens name has to be a string')
```
prettyirrelevant, line 878: Since you're inheriting from `fields.String` I don't think this check is needed.

prettyirrelevant (commented):
more comments

also this test failed.
```sh
rotkehlchen/tests/unit/test_evm_tx_decoding.py::test_tx_decode[ethtxs.db] - rotkehlchen.errors.misc.DBUpgradeError: Failed at database upgrade from version 36 to 37: no such table: ens_mappings
```

docs/api.rst
```diff
@@ -12334,3 +12334,34 @@ Get Binance Savings Interests History
...
+      *Example Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: image/png
+
+      :statuscode 200: The avatar was successfully found and returned.
+      :statuscode 404: There is no avatar set for the given ens name.
+      :statuscode 409: Avatar was requested for an ens name that is not currently in the database.
```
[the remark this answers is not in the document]
prettyirrelevant: yeah, you did. Thanks

rotkehlchen/api/rest.py
nebolax: About DBEns -- I would not put the logic for reading a file there since I think it will be good to keep only db-related logic there.

As for `api_image_response` -- yeah, I think on how to abstract it.

rotkehlchen/chain/ethereum/node_inquirer.py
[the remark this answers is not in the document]
nebolax: Now is handled

rotkehlchen/api/v1/schemas.py
```diff
@@ -2992,3 +2993,7 @@ def make_binance_savings_query(
         )
         query_dict.update({'location': location})
         return query_dict
+
+
+class EnsAvatarsSchema(Schema):
+    ens_name = EnsNameField()
```
prettyirrelevant, line 2999: Aren't you missing a required=True?

docs/api.rst
```diff
@@ -12334,3 +12334,35 @@ Get Binance Savings Interests History
...
    :statuscode 500: Internal rotki error.
    :statuscode 502: An external service used in the query such as binance could not be reached or returned unexpected response.
+
+
+Get avatar for an ens name
+============================
+
+.. http:get:: /api/(version)/avatars/ens/<ens_name>
+
+   Doing a GET on this endpoint will return the avatar that is currently set for the provided ens name. For successfull responses returns `ETag` header and for requests accepts `If-Match` and `If-None-Match` headers to allow caching on the client side.
```
prettyirrelevant, line 12344:
```suggestion
   Doing a GET on this endpoint will return the avatar that is currently set for the provided ens name. If successful, responses include an `ETag` header for caching.
```

rotkehlchen/api/rest.py
```diff
@@ -280,6 +285,49 @@ def inner(rest_api: 'RestAPI', async_query: bool = False, **kwargs: Any) -> Resp
     return wrapper
 
 
+def check_cached_image(image_path: Path, match_header: Optional[str]) -> Optional[Response]:
```
prettyirrelevant, line 288: I think the name of this function conflicts with what it does. I might be wrong though.

rotkehlchen/api/rest.py
```diff
@@ -280,6 +285,49 @@ def inner(rest_api: 'RestAPI', async_query: bool = False, **kwargs: Any) -> Resp
...
+    """
+    Check whether md5 hash of the file at `image_path` matches the value in `match_header`.
```
prettyirrelevant, line 290:
```suggestion
    Checks whether the md5 hash of the file at `image_path` matches the value in `match_header`.
```

rotkehlchen/api/rest.py
```diff
@@ -280,6 +285,49 @@ def inner(rest_api: 'RestAPI', async_query: bool = False, **kwargs: Any) -> Resp
...
+                {'mimetype': 'image/png', 'Content-Type': 'image/png'},
+            ),
+        )
+
+    return None
+
+
+def maybe_return_image(image_path: Optional[Path]) -> Response:
+    """
+    Checks whether file at `image_path` exists and if it does -- returns this file.
```
prettyirrelevant, line 309:
```suggestion
    """Checks whether the file at `image_path` exists.
    
    Returns a response with the image if it exists, otherwise a NOT FOUND response.
    """
```

rotkehlchen/api/rest.py
```diff
@@ -280,6 +285,49 @@ def inner(rest_api: 'RestAPI', async_query: bool = False, **kwargs: Any) -> Resp
...
+    Otherwise returns a NOT_FOUND response.
+    """
+    if image_path is None or image_path.is_file() is False:
+        return make_response(
+            (
+                b'',
+                HTTPStatus.NOT_FOUND, {'mimetype': 'image/png', 'Content-Type': 'image/png'}),
```
prettyirrelevant, line 316:
```suggestion
            (
                b'',
                HTTPStatus.NOT_FOUND, 
                {'mimetype': 'image/png', 'Content-Type': 'image/png'},
            ),
```

rotkehlchen/api/rest.py
```diff
@@ -4197,3 +4229,41 @@ def get_binance_savings_history(
             query_filter=query_filter,
             value_filter=value_filter,
         )
+
+    def get_ens_avatar(self, ens_name: str, match_header: Optional[str]) -> Response:
+        """
+        Searches for ENS avatar of the given `ens_name`.
```
prettyirrelevant, line 4235:
```suggestion
        Searches for the ENS avatar of the given `ens_name`.
```

rotkehlchen/api/rest.py
```diff
@@ -4197,3 +4229,41 @@ def get_binance_savings_history(
...
+        If found returns the avatar, otherwise returns 404 status code.
```
prettyirrelevant, line 4236: This is a wrong assertion because you return 409 right?
Also, isn't it best you only return either 200, 304 or 404?

I don't think 409 is needed. Thoughts?

rotkehlchen/api/rest.py
```diff
@@ -4197,3 +4229,41 @@ def get_binance_savings_history(
...
+
+        Also supports etag mechanism that helps with caching on the API caller side.
```
prettyirrelevant, line 4238:
```suggestion
        Also supports etag mechanism that helps with caching on the client side.
```

rotkehlchen/api/rest.py
nebolax: If there will be no 409, then how to handle errors? Telling the client that there is no avatar when there just a network error occurred will be wrong.

Or if the client requests avatar for an ens name that is not in our database -- we can't just say "no avatar found" because it is a logical error.

rotkehlchen/api/v1/resources.py
```diff
@@ -2835,3 +2840,12 @@ def put(
             tx_hash=tx_hash,
             associated_address=associated_address,
         )
+
+
+class EnsAvatarsResource(BaseMethodView):
+    get_schema = EnsAvatarsSchema()
+
+    @require_loggedin_user()
+    @use_kwargs(get_schema, location='json_and_query')
```
prettyirrelevant, line 2849:
```suggestion
    @use_kwargs(get_schema, location='view_args')
```

rotkehlchen/api/v1/schemas.py
prettyirrelevant: Hmmm okay.

rotkehlchen/tests/unit/test_evm_names.py
```diff
@@ -105,3 +112,41 @@ def test_uses_sources_only_when_needed(evm_address, database: 'DBHandler'):
...
+        )
+        assert dbens.get_last_avatar_update('random.ens.name.eth') == ts_now(), 'Last update timestamp should have been set'  # noqa: E501
+        assert list(tempdir.iterdir()) == []
+        try_download_ens_avatar(
+            eth_inquirer=ethereum_inquirer,
+            avatars_dir=tempdir,
+            ens_name='nebolax.eth',  # an avatar should be downloaded
+        )
+        assert dbens.get_last_avatar_update('nebolax.eth') == ts_now(), 'Last update timestamp should have been set'  # noqa: E501
+        assert list(tempdir.iterdir()) == [Path(tempdir / 'nebolax.eth.png')]
```
prettyirrelevant, line 149:
```suggestion
        assert list(tempdir.iterdir()) == [tempdir / 'nebolax.eth.png']
```

rotkehlchen/api/rest.py
prettyirrelevant: This error is not user facing, is it?
The error message returned is not used by the frontend. If it's not 200/304 then it falls back to the blocky image.

If this was a json response, yeah that's very useful but it's just blob that's returned here.

rotkehlchen/api/rest.py
```diff
@@ -4197,3 +4230,41 @@ def get_binance_savings_history(
...
+                result={'result': None, 'message': str(e)},
+                status_code=HTTPStatus.CONFLICT,
+            )
+        if last_update is None or ts_now() - last_update > ENS_AVATARS_REFRESH:
+            # If avatar for this ens name has never been checked or the avatar has expired
+            # then we try to download.
+            try:
+                try_download_ens_avatar(
+                    eth_inquirer=self.rotkehlchen.chains_aggregator.ethereum.node_inquirer,
+                    avatars_dir=Path(self.rotkehlchen.data_dir) / 'avatars',
```
prettyirrelevant, line 4261: why not make this a constant and reuse it where necessary?

prettyirrelevant (commented):
fewer comments @nebolax

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -52,15 +52,34 @@ def _update_history_events_schema(write_cursor: 'DBCursor', conn: 'DBConnection'
...
+    write_cursor.execute('ALTER TABLE ens_mappings_copy RENAME TO ens_mappings')
+    log.debug('Exit _update_ens_mappings_schema')
+
+
 def upgrade_v36_to_v37(db: 'DBHandler', progress_handler: 'DBUpgradeProgressHandler') -> None:
     """Upgrades the DB from v36 to v37. This was in v1.28.0 release.
 
         - Replace null history event subtype
     """
     log.debug('Entered userdb v36->v36 upgrade')
```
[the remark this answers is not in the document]
nebolax: What typo?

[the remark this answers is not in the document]
prettyirrelevant: v36->v37 instead of v36->v36. Not related to your pr but it's nice to correct I suppose

rotkehlchen/api/rest.py
nebolax: I would ask @LefterisJP since I think that the same way as we notify users about remote errors when querying transactions or something like that we should also display such errors here.

rotkehlchen/chain/ethereum/utils.py
```diff
@@ -166,3 +174,38 @@ def should_update_protocol_cache(cache_key: GeneralCacheType, *args: str) -> boo
...
+
+def try_download_ens_avatar(
+        eth_inquirer: 'EthereumInquirer',
+        avatars_dir: Path,
+        ens_name: str,
+) -> None:
+    """
+    Handles ens avatar downloading.
+    1. Checks whether given ens name has avatar set
+    2. If it does -- downloads the avatar and saves it in `avatars_dir`
```
LefterisJP, line 187: bad habit to use `--` instead of a comma

```suggestion
    2. If it does, downloads the avatar and saves it in `avatars_dir`
```

rotkehlchen/chain/ethereum/utils.py
```diff
@@ -166,3 +174,38 @@ def should_update_protocol_cache(cache_key: GeneralCacheType, *args: str) -> boo
...
+    - RequestsError if got avatar url, but failed to download it
+    """
+    dbens = DBEns(eth_inquirer.database)
+    avatar_url = eth_inquirer.ens_avatar_lookup(ens_name)
+    dbens.set_last_avatar_update(ens_name=ens_name, timestamp=ts_now())
+    if avatar_url is None:
+        return  # Avatar is not set
+
+    try:
+        avatar = requests.get(avatar_url, timeout=10).content
```
LefterisJP, line 201: why a hard coded 10 here? Why not use our usual constant timeout? This is what we do with coingecko icons.

rotkehlchen/chain/ethereum/utils.py
```diff
@@ -166,3 +174,38 @@ def should_update_protocol_cache(cache_key: GeneralCacheType, *args: str) -> boo
...
+) -> None:
+    """
+    Handles ens avatar downloading.
+    1. Checks whether given ens name has avatar set
+    2. If it does -- downloads the avatar and saves it in `avatars_dir`
+    3. Updates last avatar check timestamp for the given ens name
+
+    May raise:
+    - RemoteError if failed to query chain
+    - RequestsError if got avatar url, but failed to download it
```
LefterisJP, line 192: What error is this?

rotkehlchen/chain/ethereum/utils.py
```diff
@@ -166,3 +174,38 @@ def should_update_protocol_cache(cache_key: GeneralCacheType, *args: str) -> boo
...
+    dbens = DBEns(eth_inquirer.database)
+    avatar_url = eth_inquirer.ens_avatar_lookup(ens_name)
+    dbens.set_last_avatar_update(ens_name=ens_name, timestamp=ts_now())
+    if avatar_url is None:
+        return  # Avatar is not set
+
+    try:
+        avatar = requests.get(avatar_url, timeout=10).content
+    except RequestException as e:
+        log.info(
```
LefterisJP, line 203: Wtf? Why info!??
```suggestion
        log.error(
```

rotkehlchen/chain/ethereum/utils.py
```diff
@@ -166,3 +174,38 @@ def should_update_protocol_cache(cache_key: GeneralCacheType, *args: str) -> boo
...
+            f'Got {str(e)} during querying ens avatar for {ens_name}. '
+            f'Considering the url broken and the avatar non-existent.',
```
LefterisJP, line 205: Short and to the point. The logs are for us to read and we will know when that happened.

So here if this happens avatar won't be attempted to be fetched for another day, right?

```suggestion
            f'Got error {str(e)} during querying ens avatar for {ens_name}. '
```

rotkehlchen/db/ens.py
```diff
@@ -86,3 +87,28 @@ def update_values(
...
+        - InputError if given `ens_name` is not in `ens_mappings` table
+        """
+        with self.db.conn.read_ctx() as cursor:
+            result = cursor.execute(
+                'SELECT last_avatar_update FROM ens_mappings WHERE ens_name = ?',
+                (ens_name,),
+            ).fetchone()
+
+        if result is None:
+            raise InputError('Provided ens name is not being tracked')
```
LefterisJP, line 104: Though this most probably won't happen in production as frontend calls this only, give more info

```suggestion
            raise InputError(f'ens name {ens_name} is not being tracked')
```

rotkehlchen/db/ens.py
```diff
@@ -86,3 +87,28 @@ def update_values(
...
+
+    def get_last_avatar_update(self, ens_name: str) -> Timestamp:
+        """
+        Returns the timestamp when the avatar for the given ens name was updated last time.
+        May raise:
+        - InputError if given `ens_name` is not in `ens_mappings` table
+        """
+        with self.db.conn.read_ctx() as cursor:
+            result = cursor.execute(
+                'SELECT last_avatar_update FROM ens_mappings WHERE ens_name = ?',
```
LefterisJP, line 99: I know it's incosistent in the code but we want to go for this, if only for less space

```suggestion
                'SELECT last_avatar_update FROM ens_mappings WHERE ens_name=?',
```

rotkehlchen/db/ens.py
```diff
@@ -86,3 +87,28 @@ def update_values(
...
+        if result is None:
+            raise InputError('Provided ens name is not being tracked')
+
+        return Timestamp(result[0])
+
+    def set_last_avatar_update(self, ens_name: str, timestamp: Timestamp) -> None:
+        """Sets timestamp of the last update of the avatar for the given ens name"""
+        with self.db.conn.write_ctx() as cursor:
+            cursor.execute(
+                'UPDATE ens_mappings SET last_avatar_update = ? WHERE ens_name = ?',
```
LefterisJP, line 112:
```suggestion
                'UPDATE ens_mappings SET last_avatar_update=? WHERE ens_name=?',
```

rotkehlchen/db/ens.py
```diff
@@ -86,3 +87,28 @@ def update_values(
...
+                'SELECT last_avatar_update FROM ens_mappings WHERE ens_name = ?',
+                (ens_name,),
+            ).fetchone()
+
+        if result is None:
+            raise InputError('Provided ens name is not being tracked')
+
+        return Timestamp(result[0])
+
+    def set_last_avatar_update(self, ens_name: str, timestamp: Timestamp) -> None:
```
LefterisJP, line 108: I am pretty sure this is an unneeded function.Delete it and move its code to the caller.

rotkehlchen/tests/data/example_ens_avatar.png
LefterisJP, line 1: awwwww

rotkehlchen/tests/unit/test_evm_names.py
```diff
@@ -105,3 +112,41 @@ def test_uses_sources_only_when_needed(evm_address, database: 'DBHandler'):
         )],
     )
     assert names == [], 'No names should have been returned since the blockchain was None'
+
+
+@freeze_time('2023-03-20')  # freesing time just to make sure comparisons of timestamps won't fail
```
LefterisJP, line 117:
```suggestion
@freeze_time('2023-03-20')  # freezing time just to make sure comparisons of timestamps won't fail
```

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1335,6 +1335,8 @@ def test_upgrade_db_36_to_37(user_data_dir):  # pylint: disable=unused-argument
     ]
     assert all(x in old_history_events for x in kraken_events)
     assert all(x in old_history_events for x in custom_events)
+    old_ens_mappings = cursor.execute('SELECT * FROM ens_mappings').fetchall()
```
LefterisJP, line 1338: Question: I don't see you editing the test DB. Did it already have ENS mappings? I guess yes.

rotkehlchen/tests/data/ethtxs.db
LefterisJP, line 1: What is this file and why has it been touched by this PR? `test_evm_tx_decoding.py::test_tx_decode` has not been touched in the PR's code.

rotkehlchen/tests/unit/test_evm_names.py
```diff
@@ -105,3 +112,41 @@ def test_uses_sources_only_when_needed(evm_address, database: 'DBHandler'):
...
+def test_download_ens_avatar(ethereum_inquirer):
```
LefterisJP, line 118: Can't you also vcr this? I think it should be VCRed as it will be making network calls.

And `random.ens.name.eth` can at some point in the future add an avatar.
And `nebolax.eth` can remove his :wink:

rotkehlchen/tests/unit/test_evm_names.py
```diff
@@ -105,3 +112,41 @@ def test_uses_sources_only_when_needed(evm_address, database: 'DBHandler'):
...
+        assert dbens.get_last_avatar_update('random.ens.name.eth') == ts_now(), 'Last update timestamp should have been set'  # noqa: E501
+        assert list(tempdir.iterdir()) == []
+        try_download_ens_avatar(
+            eth_inquirer=ethereum_inquirer,
+            avatars_dir=tempdir,
+            ens_name='nebolax.eth',  # an avatar should be downloaded
+        )
+        assert dbens.get_last_avatar_update('nebolax.eth') == ts_now(), 'Last update timestamp should have been set'  # noqa: E501
+        assert list(tempdir.iterdir()) == [tempdir / 'nebolax.eth.png']
+        downlaoded_hash = file_md5(tempdir / 'nebolax.eth.png')
```
LefterisJP, line 150:
```suggestion
        downloaded_hash = file_md5(tempdir / 'nebolax.eth.png')
```

LefterisJP (commented):
Lots of comments unfortunately for the size of the PR.

rotkehlchen/tests/db/test_db_upgrades.py
nebolax: Yes, it did

rotkehlchen/tests/data/ethtxs.db
nebolax: This file was somehow upgrade incorrectly in the past and it had no `ens_mappings` table while it should have had. And thus the new upgrade step that I created was failing here. I changed the file to add that table.

rotkehlchen/api/rest.py
nebolax: I checked the icons endpoint and it never returns 409. So will probably make the avatars endpoint have the same interface and return 404 in all bad situations.

rotkehlchen/chain/ethereum/utils.py
nebolax: `RequestsError` is a general error coming from `requests` module. Expected it to be potentially raised here if there is an error with downloading an avatar, but now added handling for it inside this function, so will remove it from the docstring.

rotkehlchen/chain/ethereum/utils.py
nebolax: > So here if this happens avatar won't be attempted to be fetched for another day, right?

Yes. Since it can easily be not a temporary network error, but a broken link set in the ens app I think that retrying to download the avatar every time is too much.

rotkehlchen/tests/data/ethtxs.db
LefterisJP: These files do not ever get updated with each DB upgrade. So it's probably just an old DB file

rotkehlchen/tests/unit/test_evm_names.py
nebolax: Yes, makes sense. I thought that vcr doesn't support binary requests/responses, but it seems like it does.

rotkehlchen/api/rest.py
```diff
@@ -4200,3 +4191,39 @@ def get_all_evm_chains(self) -> Response:
                 for chain in ChainID
             ]),
         )
+
+    def get_ens_avatar(self, ens_name: str, match_header: Optional[str]) -> Response:
+        """
+        Searches for the ENS avatar of the given `ens_name`.
+        If found returns the avatar, otherwise returns 404 status code.
```
LefterisJP, line 4198: you will need to update the docstring
```suggestion
        If found returns the a response with the avatar, otherwise a 404 empty response if the avatar is not found or a 409 empty response if there is any error
```

rotkehlchen/tests/conftest.py
```diff
@@ -174,7 +174,7 @@ def get_cassette_dir(request: pytest.FixtureRequest) -> Path:
 
 def is_etherscan_rate_limited(response: dict[str, Any]) -> bool:
     result = False
-    with suppress(JSONDecodeError, KeyError):
+    with suppress(JSONDecodeError, KeyError, UnicodeDecodeError):
```
LefterisJP, line 177: Why this change? What prompted it?

rotkehlchen/chain/ethereum/utils.py
LefterisJP: that's not what info is for though. Info is really for stuff that should appear only in normal application run. Like ... "starting api", "running in Windows", etc.

If you put log level INFO, you should have very very limited logs. This is not an info log, this is an error log. An incorrect avatar link set by the user is also an error.

rotkehlchen/tests/conftest.py
nebolax: For responses that contain bytes, the code under this `suppress` was failing because it was trying to decode a bytes image as a utf-8 string

LefterisJP: That's not the right solution though.

Look at where this function is called. All responses that go in this function **should be json**.

If this function is called for a bytes response it's an error. I think what you should do is modify the if check in

```python
def vcr_fixture
```

to not go in that function if it's bytes.

LefterisJP (approved):
I left one comment where logic is broken in conftest. But this can go in a separate PR/commit and will merge since all is green and frontend can proceed

PR #5797, opened 2023-03-21
Refactor pagination and filters on history pages

tewshi:
There exists common codes for components on the history pages, used in managing pagination and filtering, refactors these common code into a composable for reuse.

## Checklist

- [x] The PR modified the frontend, and updated the components on history pages, to create a composable from reusable code
- [x] replace parts of these components with the composable
- [x] contains unit test
- [x] requires dependency update

### DIFF
--- a/frontend/app/package.json
+++ b/frontend/app/package.json
@@ -96,6 +96,7 @@
     "eslint": "8.34.0",
     "eslint-plugin-cypress": "2.12.1",
     "flush-promises": "1.0.2",
+    "msw": "^1.2.0",
     "postcss": "8.4.21",
     "postcss-html": "1.5.0",
     "postcss-scss": "4.0.6",
--- a/frontend/app/package.json
+++ b/frontend/app/package.json
@@ -96,7 +96,7 @@
     "eslint": "8.34.0",
     "eslint-plugin-cypress": "2.12.1",
     "flush-promises": "1.0.2",
-    "msw": "^1.2.0",
+    "msw": "1.2.1",
     "postcss": "8.4.21",
     "postcss-html": "1.5.0",
     "postcss-scss": "4.0.6",
--- a/frontend/app/src/components/history/trades/ClosedTrades.vue
+++ b/frontend/app/src/components/history/trades/ClosedTrades.vue
@@ -286,7 +286,7 @@ watch(loading, async (isLoading, wasLoading) => {
         data-cy="closed-trades__add-trade"
         @click="newExternalTrade()"
       >
-        <v-icon> mdi-plus</v-icon>
+        <v-icon> mdi-plus </v-icon>
       </v-btn>
       <template #title>
         <refresh-button
@@ -318,7 +318,7 @@ watch(loading, async (isLoading, wasLoading) => {
                     :disabled="selected.length === 0"
                     @click="massDelete"
                   >
-                    <v-icon> mdi-delete-outline</v-icon>
+                    <v-icon> mdi-delete-outline </v-icon>
                   </v-btn>
                 </v-col>
               </v-row>
--- a/frontend/app/src/components/history/trades/ClosedTrades.vue
+++ b/frontend/app/src/components/history/trades/ClosedTrades.vue
@@ -132,6 +132,7 @@ const {
   setPage,
   setOptions,
   setFilter,
+  applyRouteFilter,
   fetchData
 } = useHistoryPaginationFilter<
   Trade,
@@ -253,6 +254,10 @@ const getItemClass = (item: TradeEntry) =>
   item.ignoredInAccounting ? 'darken-row' : '';
 
 const pageRoute = Routes.HISTORY_TRADES;
+
+onBeforeMount(() => {
+  applyRouteFilter();
+});
 
 onMounted(async () => {
   const query = get(route).query;
--- a/frontend/app/src/components/history/trades/ClosedTrades.vue
+++ b/frontend/app/src/components/history/trades/ClosedTrades.vue
@@ -1,9 +1,7 @@
 <script setup lang="ts">
-import dropRight from 'lodash/dropRight';
-import { type ComputedRef, type Ref, type UnwrapRef } from 'vue';
+import { type Ref, type UnwrapRef } from 'vue';
 import { type DataTableHeader } from 'vuetify';
 import isEqual from 'lodash/isEqual';
-import { type MaybeRef } from '@vueuse/core';
 import Fragment from '@/components/helper/Fragment';
 import { Routes } from '@/router/routes';
 import { type TradeLocation } from '@/types/history/trade/location';
@@ -19,8 +17,7 @@ import {
   type LocationQuery,
   RouterPaginationOptionsSchema
 } from '@/types/route';
-import { type Collection } from '@/types/collection';
-import { defaultCollectionState, defaultOptions } from '@/utils/collection';
+import { defaultOptions } from '@/utils/collection';
 import { assert } from '@/utils/assertions';
 import { SavedFilterLocation } from '@/types/filtering';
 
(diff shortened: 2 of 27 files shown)


frontend/app/src/composables/history/filter-paginate.ts
```diff
@@ -0,0 +1,201 @@
...
+      | TradesFilters
+      | LedgerActionsFilters
+      | AssetMovementFilters
+  ): void;
+
+  RouteFilterSchema: ZodSchema;
+}
+
+export const useHistoryPaginationFilter = <T extends Object, U, V>(
+  locationOverview: Ref,
```
lukicenturi, line 34: use MaybeRef is better here

tewshi: Okay, will update it

frontend/app/src/components/history/trades/ClosedTrades.vue
```diff
@@ -440,7 +319,7 @@ watch(loading, async (isLoading, wasLoading) => {
                     :disabled="selected.length === 0"
                     @click="massDelete"
                   >
-                    <v-icon> mdi-delete-outline </v-icon>
+                    <v-icon> mdi-delete-outline</v-icon>
```
kelsos, line 322: This doesn't seem like an intentional change

tewshi: reverted.

frontend/app/src/composables/history/filter-paginate.ts
```diff
@@ -0,0 +1,201 @@
...
+    selected,
+    openDialog,
+    editableItem,
+    itemsToDelete,
+    confirmationMessage,
+    expanded,
+    isLoading,
+    state,
+    filters,
+    matchers,
```
kelsos, line 194: Is it possible to make it so that filters is not `Ref<any>` and matches not `ComputedRef<any>`

tewshi: Sure, checking.

emphasis on `any` right?

frontend/app/src/components/history/transactions/TransactionContent.vue
```diff
@@ -68,10 +56,57 @@ const {
   mainPage
 } = toRefs(props);
 
+const userAction = ref(false);
```
kelsos, line 59: Do we still need this external one?

tewshi: missed this one... will remove and get from composable

tewshi: used in template:
```html
@input="userAction = true"
```

kelsos (commented):
Pretty much what I had in mind @tewshi 

One two things, as I wrote if we can make sure that the filter/matcher is not any after passing through the composable

And the second thing is if you can add some unit tests for the new composable.

tewshi:
> Pretty much what I had in mind @tewshi
> 
> Only two things, as I wrote if we can make sure that the filter/matcher is not any after passing through the composable
> 
> And the second thing is if you can add some unit tests for the new composable.

Alrighty, working on the comments... and the test

kelsos:
> > Pretty much what I had in mind @tewshi
> > Only two things, as I wrote if we can make sure that the filter/matcher is not any after passing through the composable
> > And the second thing is if you can add some unit tests for the new composable.
> 
> Alrighty, working on the comments... and the test

Great, thank you

frontend/app/src/composables/history/filter-paginate.ts
```diff
@@ -142,15 +136,11 @@ export const useHistoryPaginationFilter = <T extends Object, U, V>(
...
 
-  const setFilter = (newFilter: UnwrapRef<typeof filters>) => {
+  const setFilter = (newFilter: UnwrapRef<Ref<W>>) => {
     set(userAction, true);
     updateFilter(newFilter);
   };
 
-  onBeforeMount(() => {
-    applyRouteFilter();
-  });
```
kelsos: What is the problem with having it inside the composable?

tewshi: it is causing an issue in the test environment...

on a second thought, I could mock the `onBeforeMount` composable.

frontend/app/src/composables/history/filter-paginate.ts
```diff
@@ -76,7 +70,7 @@ export const useHistoryPaginationFilter = <T extends Object, U, V>(
       offset,
       orderByAttributes: sortBy?.length > 0 ? sortBy : ['timestamp'],
       ascending: sortBy?.length > 0 ? sortDesc.map(bool => !bool) : [false]
-    };
+    } as U;
```
kelsos, line 73: What is the error here? Is this because of the extra params?

frontend/app/src/composables/history/filter-paginate.ts
kelsos: Yes, if it is only for testing then maybe we can mock it. 
Because just copying it in every component would be really inconvenient

frontend/app/src/composables/history/filter-paginate.ts
tewshi: Type 'W & { limit: number; offset: number; orderByAttributes: (keyof T)[] | string[]; ascending: boolean[]; }' is not assignable to type 'U'.     'U' could be instantiated with an arbitrary type which could be unrelated to 'W & { limit: number; offset: number; orderByAttributes: (keyof T)[] | string[]; ascending: boolean[]; }'.

kelsos: Ok, can you add a TODO there to figure out later? No need to spend more time on it right now.

It's preferable to not have casting, but figuring out type issues is usually a time-sink.

tewshi: Okay, will add todo to it, thanks.

frontend/app/package.json
```diff
@@ -96,6 +96,7 @@
     "eslint": "8.34.0",
     "eslint-plugin-cypress": "2.12.1",
     "flush-promises": "1.0.2",
+    "msw": "^1.2.0",
```
kelsos, line 99: can you do exact version here?

tewshi: yes sure, sorry I missed this

kelsos:
@tewshi can you rebase it on top of develop? it needs a rebase

kelsos:
@tewshi I am doing some manual testing and at least on the `Evm transactions` page, I can see for example that the `protocol` filter doesn't work. It doesn't seem to have any effect.

In the evm transactions also the Address filter doesn't seem to be properly reactive.
The URL updates on back and forth button press,  but there is no change in the filter component (the address remains visible), and the data does not reload.

Can you figure out why and along with the fix add some tests to make sure that these work fine?

kelsos:
They other sections seem to work fine, so the problem seems with transactions since this is a specialized case

tewshi:
> @tewshi I am doing some manual testing and at least on the `Evm transactions` page, I can see for example that the `protocol` filter doesn't work. It doesn't seem to have any effect.
> 
> In the evm transactions also the Address filter doesn't seem to be properly reactive. The URL updates on back and forth button press, but there is no change in the filter component (the address remains visible), and the data does not reload.
> 
> Can you figure out why and along with the fix add some tests to make sure that these work fine?

I'll go through and check, I think this is due to the fact that I didn't have enough data to present on the EVM UI...

I'll see how I can populate the data and test, and fix.

kelsos:
> > @tewshi I am doing some manual testing and at least on the `Evm transactions` page, I can see for example that the `protocol` filter doesn't work. It doesn't seem to have any effect.
> > In the evm transactions also the Address filter doesn't seem to be properly reactive. The URL updates on back and forth button press, but there is no change in the filter component (the address remains visible), and the data does not reload.
> > Can you figure out why and along with the fix add some tests to make sure that these work fine?
> 
> I'll go through and check, I think this is due to the fact that I didn't have enough data to present on the EVM UI...
> 
> I'll see how I can populate the data and test, and fix.

You can create a test db, and you could check the backend tests for some addresses.

tewshi:
> > > @tewshi I am doing some manual testing and at least on the `Evm transactions` page, I can see for example that the `protocol` filter doesn't work. It doesn't seem to have any effect.
> > > In the evm transactions also the Address filter doesn't seem to be properly reactive. The URL updates on back and forth button press, but there is no change in the filter component (the address remains visible), and the data does not reload.
> > > Can you figure out why and along with the fix add some tests to make sure that these work fine?
> > 
> > 
> > I'll go through and check, I think this is due to the fact that I didn't have enough data to present on the EVM UI...
> > I'll see how I can populate the data and test, and fix.
> 
> You can create a test db, and you could check the backend tests for some addresses.

Updated @kelsos

kelsos:
https://github.com/rotki/rotki/blob/5b20530262a1a3a7b8031e4e4f530391b6dbd07f/frontend/app/src/utils/data.ts#L33

The utility is here maybe we can use it in the composable to make sure that empty arrays e.g. don't overwrite values.

kelsos (approved):
@tewshi like we discussed, I am merging this and we just need to make sure that the account filter reacts properly to navigation back/forward and also we need to migrate the rest of the places where this logic is used.

PR #5798, opened 2023-03-22
Show ENS avatar frontend

### LINKED ISSUE #5622: Show ENS avatars for addresses instead of blockies where possible

lukicenturi:
Closes #5622 

image

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`5798` Users will now be able to see the avatar of the ENS.
 * :feature:`1505` Users will now be able to see breakdown of Binance/Binance US exchange balances, which are in account balances, and which are in savings/lending. The interest history also will be taken into account in the PnL report.
 * :feature:`5639` Cowswap transactions are now decoded properly.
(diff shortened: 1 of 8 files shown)


kelsos (approved):
looks good @lukicenturi thanks :)

PR #5805, opened 2023-03-23
feat: remove stores data when it goes out of scopes

### LINKED ISSUE #5752: Identify scope for pinia stores and remove stores when out of scope
## Abstract

All our stores get initialized when the user navigates to a specific page in the application, and while they get reset on logout, they are staying in memory for the rest of the application lifecycle, even if the user never navigates back to the application section. 

### Example
- reports, is required only when the user navigates to `/reports/*`

## Motivation

Reduce memory footprint during runtime by unloading unused stores.

## Specification
- Identify which of the stores have specific scopes
- Make sure that the store gets disposed when the user is out of the scope

lukicenturi:
Closes #5752

## Checklist

- [x] Clear reports store when go out from '/reports'
- [x] Clear staking store when go out from their routes,
- [x] Clear defi store when go out from '/defi' (except for uniswap, sushiswap, balancer, and liquity, since they are used in other places too)


lukicenturi:
Some endpoint doesn't support `only_cache` I will talk to backend guys

kelsos:
Also other than that is there anything else remaining?

lukicenturi:
> Also other than that is there anything else remaining?

I think no more for now.

kelsos (approved):
lgtm

PR #5807, opened 2023-03-23
Decode curve trades

### LINKED ISSUE #5677: Decode curve trades
At the moment we provide decoders for curve pool events but not for trades.

## Task

Add decoders for trades made using the curve contracts and add test for them

Decode trades made via curve swap router https://etherscan.io/address/0x99a58482bd75cbab83b27ec03ca68ff489b5788f

nebolax:
Closes #5677

## Checklist


rotkehlchen/chain/ethereum/modules/curve/pools_cache.py
```diff
@@ -203,11 +228,12 @@ def save_curve_pools_to_cache(
             key_parts=[GeneralCacheType.CURVE_POOL_ADDRESS, lp_token_address],
             values=[pool_address],
         )
-        globaldb_set_general_cache_values(
-            write_cursor=write_cursor,
-            key_parts=[GeneralCacheType.CURVE_POOL_TOKENS, pool_address],
-            values=coins,
-        )
+        for i, coin in enumerate(coins):
+            globaldb_set_general_cache_values(
```
yabirgb, line 232: I believe we tend to avoid single letter variables, can you rename to `index` or `coin_index`?

rotkehlchen/globaldb/cache.py
```diff
@@ -122,3 +122,12 @@ def globaldb_delete_general_cache(
...
     else:
         write_cursor.execute('DELETE FROM general_cache WHERE key=? AND value=?', (cache_key, value))  # noqa: E501
+
+
+def globaldb_delete_general_cache_like(
+        write_cursor: DBCursor,
+        key_parts: Iterable[Union[str, GeneralCacheType]],
+) -> None:
+    """Function to delete globaldb cache. Deletes all the values that start with key."""
+    cache_key = _compute_cache_key(key_parts)  # note: cache key should not contain `.` or `%`
```
yabirgb, line 132: This looks like a bug waiting to happen. Can you add an assertion here so if that happens at least we know?

rotkehlchen/tests/unit/test_general_cache.py
```diff
@@ -115,10 +116,7 @@ def mock_call_contract(contract, node_inquirer, method_name, **kwargs):
             )[0]
             lp_tokens_to_pools_in_cache[lp_token_addr] = pool_addr
 
-            pool_coins = globaldb_get_general_cache_values(
-                cursor=cursor,
-                key_parts=[GeneralCacheType.CURVE_POOL_TOKENS, pool_addr],
-            )
+            pool_coins = read_curve_pool_tokens(cursor=cursor, pool_address=pool_addr)
             pool_coins_in_cache[pool_addr] = pool_coins
```
yabirgb, line 120: You don't use it after right? Can just

```suggestion
            pool_coins_in_cache[pool_addr] = read_curve_pool_tokens(cursor=cursor, pool_address=pool_addr)
```

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -792,3 +792,231 @@ def test_deposit_multiple_tokens(ethereum_transaction_decoder, ethereum_accounts
...
+            address=string_to_evm_address('0x8038C01A0390a8c547446a0b2c18fc9aEFEcc10c'),
+        ),
+    ]
+    assert events == expected_events
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('ethereum_accounts', [['0x3Da232a0c0A5C59918D7B5fF77bf1c8Fc93aeE1B']])
+def test_curve_swap_router(ethereum_transaction_decoder, ethereum_accounts):
+    """Check the case for a pool where multiple deposit events appear in the transaction"""
```
yabirgb, line 971: What deposits? I believe the docstring is wrong

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -792,3 +792,231 @@ def test_deposit_multiple_tokens(ethereum_transaction_decoder, ethereum_accounts
...
+            address=string_to_evm_address('0xA96A65c051bF88B4095Ee1f2451C2A9d43F53Ae2'),
+        ),
+    ]
+    assert events == expected_events
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('ethereum_accounts', [['0x38abab9766e0b27d2912718a884292b8E7eb2803']])
+def test_curve_trade_exchange_underlying(ethereum_transaction_decoder, ethereum_accounts):
+    """Check the case for a pool where multiple deposit events appear in the transaction"""
```
yabirgb, line 914: Please adjust the docstring to match what you are testing

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -221,6 +229,128 @@ def _decode_curve_deposit_events(
...
+            bought_token_id = hex_or_bytes_to_int(tx_log.data[64:96])
+            raw_bought_amount = hex_or_bytes_to_int(tx_log.data[96:128])
+            if (
+                tx_log.topics[0] == TOKEN_EXCHANGE and
+                pool_address in self.curve_pools and
+                len(self.curve_pools[pool_address]) > max(sold_token_id, bought_token_id)
+            ):
+                sold_token_address = self.curve_pools[pool_address][sold_token_id]
+                bought_token_address = self.curve_pools[pool_address][bought_token_id]
+        else:  # EXCHANGE_MULTIPLE
```
yabirgb, line 274: what else can it be? I fear that this matches other topics (other methods in the contract) and you end up having index errors. Can't you get the topics for the other cases and add a else statement catching any unexpected method?

yabirgb (commented):
A few comments with things to improve

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -221,6 +229,128 @@ def _decode_curve_deposit_events(
...
+        Then match and label events.
+        Then reshuffle events to make sure that spend and receive are consecutive.
+        """
+
+        # These are nullable because in case a curve pool is not stored in our cache or if it
+        # is a swap in a metapool (TOKEN_EXCHANGE_UNDERLYING) we will skip token check.
+        sold_token_address: Optional[ChecksumEvmAddress] = None
+        bought_token_address: Optional[ChecksumEvmAddress] = None
+
+        pool_or_router_address: ChecksumEvmAddress
```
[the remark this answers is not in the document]
nebolax: Alright

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -221,6 +229,128 @@ def _decode_curve_deposit_events(
...
+        receive_event: Optional['EvmEvent'] = None
+        for event in decoded_events:
+            if event.address != pool_or_router_address:
+                continue
+
+            crypto_asset = event.asset.resolve_to_crypto_asset()
+            if (
+                event.location_label == spender_address and
+                event.event_type == HistoryEventType.SPEND and
+                event.balance.amount == asset_normalized_value(amount=raw_sold_amount, asset=crypto_asset) and  # noqa: E501
```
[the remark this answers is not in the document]
nebolax: Can't do it since we don't know the asset for sure. In some cases (for example if a curve pool is not stored in out cache for some reason) there will be no asset known.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -221,6 +229,128 @@ def _decode_curve_deposit_events(
...
+                (sold_asset is None or event.asset == sold_asset)
+            ):
+                event.event_type = HistoryEventType.TRADE
+                event.event_subtype = HistoryEventSubType.SPEND
+                event.notes = f'Swap {event.balance.amount} {crypto_asset.symbol} in curve'
+                event.counterparty = CPT_CURVE
+                spend_event = spend_event
```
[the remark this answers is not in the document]
nebolax: Ouch. Thanks

rotkehlchen/globaldb/cache.py
nebolax: yes, will add

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -221,6 +229,128 @@ def _decode_curve_deposit_events(
...
+            # 1. Token in
+            # 2..X Pools that were used
+            # X+1. Token out
+            # X+2..9 (unused items) Zero addresses
+            # Here we read only addresses of token in and token out.
+            sold_token_address = hex_or_bytes_to_address(tx_log.data[:32])
+            for i in range(1, 9):  # Starting from 1 because at 0 is `sold_token_address`
+                address = hex_or_bytes_to_address(tx_log.data[32 * i:32 * (i + 1)])
+                if address == ZERO_ADDRESS:
+                    break
```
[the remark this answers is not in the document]
nebolax: There always is token in and token out. As said in the comment above, the structure is:
```
# 1. Token in
# 2..X Pools that were used
# X+1. Token out
# X+2..9 (unused items) Zero addresses
```

So the minimum of "actual" (non-zero) addresses is three. Token in, Pool, Token out.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -221,6 +229,128 @@ def _decode_curve_deposit_events(
...
+            raw_bought_amount = hex_or_bytes_to_int(tx_log.data[-32:])
+            # Curve swap router logs route (a list of addresses) that was used. Route consists of
+            # at max 9 items. The structure is:
+            # 1. Token in
+            # 2..X Pools that were used
+            # X+1. Token out
+            # X+2..9 (unused items) Zero addresses
+            # Here we read only addresses of token in and token out.
+            sold_token_address = hex_or_bytes_to_address(tx_log.data[:32])
+            for i in range(1, 9):  # Starting from 1 because at 0 is `sold_token_address`
```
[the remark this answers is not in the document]
nebolax: Removed `at max`. Yes, is always 9.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -221,6 +229,128 @@ def _decode_curve_deposit_events(
...
+            spender_address = owner
+            receiver_address = owner
+            sold_token_id = hex_or_bytes_to_int(tx_log.data[:32])
+            raw_sold_amount = hex_or_bytes_to_int(tx_log.data[32:64])
+            bought_token_id = hex_or_bytes_to_int(tx_log.data[64:96])
+            raw_bought_amount = hex_or_bytes_to_int(tx_log.data[96:128])
+            if (
+                tx_log.topics[0] == TOKEN_EXCHANGE and
+                pool_address in self.curve_pools and
+                len(self.curve_pools[pool_address]) > max(sold_token_id, bought_token_id)
```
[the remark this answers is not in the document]
nebolax: Will add a comment, but this is to make sure that the pool and the tokens of the pool are cached so that we don't get an index error further down the line.

rotkehlchen/globaldb/cache.py
```diff
@@ -122,3 +122,18 @@ def globaldb_delete_general_cache(
...
+        write_cursor: DBCursor,
+        key_parts: Iterable[Union[str, GeneralCacheType]],
+) -> None:
+    """Function to delete globaldb cache. Deletes all the values that start with key."""
+    condition = all(
+        isinstance(part, GeneralCacheType) or ('%' not in part and '.' not in part)
+        for part in key_parts
+    )
+    message = 'Parts of cache key that is used with LIKE should not contain "%" or "."'
+    assert condition, message
```
yabirgb, line 137: The  issue with moving this to a variable is that that the check doesn't get removed when building for production and you are adding extra logic. It should be inside the assert otherwise you are making the check always

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -228,6 +236,125 @@ def _decode_curve_deposit_events(
...
+            swapping_contract = pool_address
+            # When a single pool is used, spender and receiver is always the same
+            spender_address = receiver_address = hex_or_bytes_to_address(context.tx_log.topics[1])
+            sold_token_id = hex_or_bytes_to_int(context.tx_log.data[:32])
+            raw_sold_amount = hex_or_bytes_to_int(context.tx_log.data[32:64])
+            bought_token_id = hex_or_bytes_to_int(context.tx_log.data[64:96])
+            raw_bought_amount = hex_or_bytes_to_int(context.tx_log.data[96:128])
+            if (
+                # Can only do this check for "normal" exchanges (not exchange_unederlying) since
+                #
```
yabirgb, line 271: something in the comment is missing here

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -228,6 +236,125 @@ def _decode_curve_deposit_events(
...
+                context.tx_log.topics[0] == TOKEN_EXCHANGE and  # Perform this check only for `TOKEN_EXCHANGE`  # noqa: E501
+                pool_address in self.curve_pools and  # Make sure that the pool is cached
+                len(self.curve_pools[pool_address]) > max(sold_token_id, bought_token_id)  # Make sure that tokens of the pool are cached  # noqa: E501
```
yabirgb, line 274: this is because the indices are 0 indexed and then always holds `>` instead of `>=`?

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -306,7 +440,7 @@ def reload_data(self) -> Optional[Mapping[ChecksumEvmAddress, tuple[Any, ...]]]:
         """
         self.ethereum.assure_curve_protocol_cache_is_queried()
         new_curve_pools = read_curve_pools()
-        curve_pools_diff = new_curve_pools - self.curve_pools
+        curve_pools_diff = new_curve_pools.keys() - self.curve_pools.keys()
         if curve_pools_diff == set():
```
yabirgb, line 444: not your code but isn't better here to check

```python
 if len(curve_pools_diff) == 0
```

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -228,6 +236,125 @@ def _decode_curve_deposit_events(
...
+        if bought_token_address is not None:
+            if bought_token_address == ETH_SPECIAL_ADDRESS:
+                bought_asset = A_ETH
+            else:
+                bought_asset = Asset(evm_address_to_identifier(
+                    address=bought_token_address,
+                    chain_id=self.evm_inquirer.chain_id,
+                    token_type=EvmTokenKind.ERC20,
+                ))
+
```
yabirgb, line 319: to avoid repeating the code this could be a `_function` in this file. Then is

```
sold_asset = _read_curve_asset(sold_token_address)
bought_asset = _read_curve_asset(bought_token_address)
```

yabirgb (commented):
Thanks!  A few things to improve

rotkehlchen/chain/ethereum/modules/curve/decoder.py
nebolax: yes

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -228,6 +236,125 @@ def _decode_curve_deposit_events(
...
+                event.event_subtype = HistoryEventSubType.RECEIVE
+                event.notes = f'Receive {event.balance.amount} {crypto_asset.symbol} as the result of a swap in curve'  # noqa: E501
+                event.counterparty = CPT_CURVE
+                receive_event = event
+
+        if spend_event is not None and receive_event is not None:
+            # Just to make sure that spend and receive events are consecutive
+            maybe_reshuffle_events(spend_event, receive_event, context.decoded_events)
+        else:
+            log.error(f'Did not find spend and receive events for a curve swap. {spend_event=} {receive_event=}')  # noqa: E501
```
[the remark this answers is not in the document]
nebolax: tx hash is contained in events in `even_identifier` field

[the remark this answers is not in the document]
yabirgb: but does it show completely? I recall that sometimes the field is not shown enterely at least in the tests. Don't know if is the case in the logs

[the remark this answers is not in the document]
nebolax: Yes, it does. I just did a quick check and this is the output that I got:
```
EvmEvent(self.event_identifier=HexBytes('0xd561728d989c4d8a25ca6708051cdb265dbc455927bb8c355083b790101487e9'),self.sequence_index=0,self.timestamp=1679550275000,self.location=<Location.ETHEREUM: 38>,self.event_type=<HistoryEventType.SPEND: 5>,self.event_subtype=<HistoryEventSubType.FEE: 3>,self.asset=<Asset identifier:ETH>,self.balance=Balance(amount=FVal(0.003261945529483024), usd_value=FVal(0)),self.location_label='0x3Da232a0c0A5C59918D7B5fF77bf1c8Fc93aeE1B',self.notes='Burned 0.003261945529483024 ETH for gas',self.identifier=None,self.counterparty='gas',self.product=None,self.address=None,self.extra_data=None)
```

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -41,35 +44,67 @@
...
+    Reads tokens for a particular curve pool. Tokens are stored with their indexes to make sure
+    that the order of coins in pool contract and in our cache is the same. This functions reads
+    and returns tokens in sorted order.
+    """
+    tokens_data = globaldb_get_general_cache_keys_and_values_like(
+        cursor=cursor,
+        key_parts=[GeneralCacheType.CURVE_POOL_TOKENS, pool_address],
+    )
+    found_tokens: list[tuple[int, ChecksumEvmAddress]] = []
+    base_key_length = len(compute_cache_key([GeneralCacheType.CURVE_POOL_TOKENS, pool_address]))
```
yabirgb, line 71: nitpick: truth is that this is a constant of 

```
len(GeneralCacheType.CURVE_POOL_TOKENS) + len(any ethereum_address)
```

yabirgb (commented):
Alexey commented in discord that the change to the globaldb is adding the index key to the curve pools

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -231,6 +257,108 @@ def _decode_curve_deposit_events(
...
+            receiver_address = hex_or_bytes_to_address(context.tx_log.topics[2])
+            raw_sold_amount = hex_or_bytes_to_int(context.tx_log.data[-64:-32])
+            raw_bought_amount = hex_or_bytes_to_int(context.tx_log.data[-32:])
+            # Curve swap router logs route (a list of addresses) that was used. Route consists of
+            # 9 items. The structure is:
+            # 1. Token in
+            # 2..X Pools that were used
+            # X+1. Token out
+            # X+2..9 (unused items) Zero addresses
+            # Here we read only addresses of token in and token out.
```
LefterisJP, line 314: This explanation does not make sense.
```suggestion
            # 9 items. The structure is:
            # 1. Token in
            # 2..X Pools that were used
            # X+1. Token out
            # X+2..9 (unused items) Zero addresses
            # Here we read only addresses of token in and token out.
```

Please write a proper comment explaining what each address is at each positiion.

You write 1, 2...X,, then X+1, x+2 ...9. This makes absolute no sense and I can't understand the code here.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -348,8 +484,9 @@ def reload_data(self) -> Optional[Mapping[ChecksumEvmAddress, tuple[Any, ...]]]:
         otherwise `None` is returned.
         """
         self.ethereum.assure_curve_protocol_cache_is_queried()
+
```
LefterisJP, line 487:
```suggestion
```

rotkehlchen/globaldb/cache.py
```diff
@@ -58,18 +58,38 @@ def globaldb_get_general_cache_values(
...
 
+def globaldb_get_general_cache_keys_and_values_like(
+        cursor: DBCursor,
+        key_parts: Iterable[Union[str, GeneralCacheType]],
+) -> list[tuple[str, str]]:
+    """
+    Function to read globaldb cache.
+    Returns all pairs key-value where key starts with the provided `key_parts`.
+    """
+    assert all(
```
LefterisJP, line 74: We never check validity of arguments like that with assertions. Add these in the docstring.

rotkehlchen/globaldb/cache.py
```diff
@@ -107,8 +127,22 @@ def globaldb_delete_general_cache(
...
     else:
         write_cursor.execute('DELETE FROM general_cache WHERE key=? AND value=?', (cache_key, value))  # noqa: E501
+
+
+def globaldb_delete_general_cache_like(
+        write_cursor: DBCursor,
+        key_parts: Iterable[Union[str, GeneralCacheType]],
+) -> None:
+    """Function to delete globaldb cache. Deletes all the values that start with key."""
+    assert all(
```
LefterisJP, line 142: same here. Why did you think you need to do this? If we were doing this all over the code then each function should be ful of assertions at the start

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -45,10 +48,33 @@
...
+CURVE_SWAP_ROUTER = string_to_evm_address('0x99a58482BD75cbab83b27EC03CA68fF489b5788f')
+
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
 
+def _read_curve_asset(
+        asset_address: Optional[ChecksumEvmAddress],
+        chain_id: ChainID,
+) -> Optional[Asset]:
```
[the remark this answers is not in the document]
nebolax: Technically it can, but there is no need for this, since the asset that is returned from here is used only for comparison with `event.asset` when iterating over `decoded_events`.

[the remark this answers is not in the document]
LefterisJP: Yes but isn't it better to be more specific? Since you already use the `Asset` constructor and the `evm_address_to_identifier` functions this should not be any more work and would be more explicit as to what type of asset this function returns.

I would only consider keeping it `Asset` if there is any more (considerable) work that is done with other constructors.

Okay .. just checked the code. It's not much work but it's potentially more DB calls. It's fine then. **leave it as is**
but also add a comment explaining why it's asset. That you don't use any other fields.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -231,6 +257,109 @@ def _decode_curve_deposit_events(
...
+            receiver_address = hex_or_bytes_to_address(context.tx_log.topics[2])
+            raw_sold_amount = hex_or_bytes_to_int(context.tx_log.data[-64:-32])
+            raw_bought_amount = hex_or_bytes_to_int(context.tx_log.data[-32:])
+            # Curve swap router logs route (a list of addresses) that was used. Route consists of
+            # 9 elements. Consider X a number of pools that was used. Then the structure can be
+            # described in the following way:
+            # At 0 index: Address of token in
+            # From 1 to X indices: Addresses of pools that were used
+            # At 1+X index: Address of token out
+            # From 2+X to 8 indices: Unused elements (zero addresses)
```
LefterisJP, line 314: The yoda usage here is confusing

```suggestion
            # At X + 1 index: Address of token out
            # From X + 2 to X + 8 indices: Unused elements (zero addresses)
```

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -231,6 +257,109 @@ def _decode_curve_deposit_events(
...
+            # Here we read only addresses of token in and token out.
+            sold_token_address = hex_or_bytes_to_address(context.tx_log.data[:32])
+            for i in range(1, 9):  # Starting from 1 because at 0 is `sold_token_address`
```
LefterisJP, line 317: What do you mean `sold_token_address`? Above you write: `At 0 index: Address of token in`. Which of the two is it?

rotkehlchen/globaldb/cache.py
```diff
@@ -58,18 +58,34 @@ def globaldb_get_general_cache_values(
...
 
 
+def globaldb_get_general_cache_keys_and_values_like(
+        cursor: DBCursor,
+        key_parts: Iterable[Union[str, GeneralCacheType]],
+) -> list[tuple[str, str]]:
+    """
+    Function to read globaldb cache.
+    Returns all pairs key-value where key starts with the provided `key_parts`.
+    `key_parts` should not contain neither `%` nor `.` symbols.
```
LefterisJP, line 73:
```suggestion

    `key_parts` should contain neither the `%` nor the `.` symbol.
```

rotkehlchen/globaldb/cache.py
```diff
@@ -107,8 +123,20 @@ def globaldb_delete_general_cache(
...
         write_cursor.execute('DELETE FROM general_cache WHERE key=? AND value=?', (cache_key, value))  # noqa: E501
+
+
+def globaldb_delete_general_cache_like(
+        write_cursor: DBCursor,
+        key_parts: Iterable[Union[str, GeneralCacheType]],
+) -> None:
+    """
+    Function to delete globaldb cache. Deletes all the values that start with key.
+    `key_parts` should not contain neither `%` nor `.` symbols.
```
LefterisJP, line 139:
```suggestion

    `key_parts` should contain neither the `%` nor the `.` symbol.
```

LefterisJP (approved):
I think it's mostly small nitpick and questions. Could move to other PR.

rotkehlchen/chain/ethereum/modules/curve/decoder.py
nebolax: Can be a confusion, yes. `Token in` and `sold token` are the same in the terminology here.

So it's like `token in` / `sold token` goes in the router / is spent and `token out` / `received_token` comes from the router / is bought.

LefterisJP: Then remove the token_in part, or explain it's token going in the router when editing the comment.

LefterisJP:
All the extra work requested shoudl be done in a susequent PR. Do not forget.

PR #5808, opened 2023-03-23
Fix get_value_stats

nebolax:


yabirgb (approved):

rotkehlchen/tests/data/liquity_stats.db
LefterisJP, line 1: What new data are in this DB? Is it

> Changed schema of history_events table to the new (latest) schema.

LefterisJP (commented):
small q

Issue #5795, opened 2023-03-21
Clear images cache


nebolax:
## Abstract

We should allow users to clear cache of icons and avatars -- in case icons or avatars change and they want to refresh them before the default refresh timeout passes.

## Specification

 - [ ] On backend: create an endpoint to clear cache that would accept arguments: `icons` / `avatars`, icon/avatar `identifier` (if empty -- delete all icons/avatars).
 - [ ] On frontend: somewhere provide users buttons to clear icons and avatars cache.

Issue #4841, opened 2022-09-15
Change the data folder structure so that app data and user accounts are not under the same directory


kelsos:
## Abstract

We need to change the directory structure in the application data folder so that application-specific data such as airdrops/global_data etc are not in the same directory as the user directories.

PR #5811, opened 2023-03-24
feat: add cache invalidation for icons & avatars

prettyirrelevant:
Backend aspect of #5795

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -12408,3 +12408,47 @@ Get avatar for an ens name
    :statuscode 400: Invalid ens name provided (it doesn't end with `.eth`).
    :statuscode 404: There is no avatar set for the given ens name.
    :statuscode 409: Avatar was requested for an ens name that is not currently in the database or we couldn't query the blockchain.
+
+
+Clear Icons/Avatars Cache
+=================================
+
+.. http:post:: /api/(version)/cache/<cache_type>/clear
+
+   Doing a POST on this endpoint will clear the cache of avatars/icons depending on the ``cache_type``. Valid options of ``cache_type`` are ``icons`` & ``avatars``.
+
+   **Example Request**
+
+    .. http:example:: curl wget httpie python-requests
+
+      POST /api/1/cache/icons/clear HTTP/1.1
+      Host: localhost:5042
+      Content-Type: application/json;charset=UTF-8
+
+      {
+         "icons": ["ETH", "BTC"]
+      }
+
+    .. http:example:: curl wget httpie python-requests
+
+      POST /api/1/cache/avatars/clear HTTP/1.1
+      Host: localhost:5042
+      Content-Type: application/json;charset=UTF-8
+
+      {
+         "avatars": ["rotki.eth", "nebolax.eth"]
+      }
+
+   :reqjsonarr optional[string] icons: An array of the icons to be cleared from the cache. All icons are deleted in the cache if ``null``. Only works with ``cache_type=icons``.
+   :reqjsonarr optional[string] avatars: An array of the avatars to be cleared from the cache. All avatars are deleted in the cache if ``null``. Only works with ``cache_type=avatars``.
+
+   **Example Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+   :statuscode 200: The cache items were cleared successfully.
+   :statuscode 400: Entries provided do not match the ``cache_type`` specified. Invalid ``cache_type`` provided.
+   :statuscode 500: Internal rotki error
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -4234,3 +4234,41 @@ class RestAPI():
                 )
 
         return maybe_create_image_response(image_path=avatar_path)
+
+    def clear_icons_cache(self, icons: Optional[list[Asset]]) -> Response:
+        """Clears cache entries for the specified icons.
+
+        If no icons are provided, the icons cache is cleared entirely.
+        """
+        icons_dir = self.rotkehlchen.icon_manager.icons_dir
+        if icons is None:
+            for entry in icons_dir.iterdir():
+                if entry.is_file() and entry.parent != icons_dir / 'avatars':
+                    entry.unlink()
+
+            return api_response(OK_RESULT)
+
+        for icon in icons:
+            self.rotkehlchen.icon_manager.delete_icon(icon)
+
+        return api_response(OK_RESULT)
+
+    def clear_avatars_cache(self, avatars: Optional[list[str]]) -> Response:
+        """Clears cache entries for the specified avatars of ENS names.
+
+        If no avatars are provided, the avatars cache is cleared entirely.
+        """
+        avatars_dir = self.rotkehlchen.icon_manager.icons_dir / 'avatars'
+        if avatars is None:
+            for entry in avatars_dir.iterdir():
+                if entry.is_file():
+                    entry.unlink()
+
+            return api_response(OK_RESULT)
+
+        avatars_to_delete = [avatars_dir / f'{avatar_name}.png' for avatar_name in avatars]
+        for avatar in avatars_to_delete:
+            if avatar.is_file():
+                avatar.unlink()
+
+        return api_response(OK_RESULT)
--- a/rotkehlchen/api/server.py
+++ b/rotkehlchen/api/server.py
@@ -49,6 +49,7 @@ from rotkehlchen.api.v1.resources import (
     BlockchainBalancesResource,
     BlockchainsAccountsResource,
     BTCXpubResource,
+    ClearCacheResource,
     CompoundBalancesResource,
     CompoundHistoryResource,
     ConfigurationsResource,
@@ -314,6 +315,7 @@ URLS_V1: URLS = [
         'per_timestamp_db_snapshots_resource',
     ),
     ('/notes', UserNotesResource),
+    ('/cache/<string:cache_type>/clear', ClearCacheResource),
 ]
 
 logger = logging.getLogger(__name__)
--- a/rotkehlchen/api/v1/resources.py
+++ b/rotkehlchen/api/v1/resources.py
@@ -50,6 +50,9 @@ from rotkehlchen.api.v1.schemas import (
     BlockchainAccountsPatchSchema,
     BlockchainAccountsPutSchema,
     BlockchainBalanceQuerySchema,
+    ClearAvatarsCacheSchema,
+    ClearCacheSchema,
+    ClearIconsCacheSchema,
     CryptoAssetSchema,
     CurrentAssetsPriceSchema,
     CustomAssetsQuerySchema,
@@ -2858,3 +2861,22 @@ class EnsAvatarsResource(BaseMethodView):
     @use_kwargs(get_schema, location='view_args')
     def get(self, ens_name: str) -> Response:
         return self.rest_api.get_ens_avatar(ens_name=ens_name, match_header=get_match_header())
+
+
+class ClearCacheResource(BaseMethodView):
+    post_schema = ClearCacheSchema()
+    icons_schema = ClearIconsCacheSchema()
+    avatars_schema = ClearAvatarsCacheSchema()
+
+    @require_loggedin_user()
+    @use_kwargs(post_schema, location='view_args')
+    def post(self, cache_type: Literal['icons', 'avatars']) -> Response:
+        req_body = flask_request.get_json(force=True, silent=True)
+        req_body = {} if req_body is None else req_body
+        if cache_type == 'icons':
+            data = self.icons_schema.load(req_body)
+            return self.rest_api.clear_icons_cache(data['icons'])
+
+        # can only be avatars
+        data = self.avatars_schema.load(req_body)
+        return self.rest_api.clear_avatars_cache(data['avatars'])
--- a/rotkehlchen/api/v1/schemas.py
+++ b/rotkehlchen/api/v1/schemas.py
@@ -3,9 +3,10 @@ from enum import auto
 from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, get_args
 from uuid import uuid4
 
+import marshmallow
 import webargs
 from eth_utils import to_checksum_address
-from marshmallow import Schema, fields, post_load, validates_schema
+from marshmallow import Schema, fields, post_load, validate, validates_schema
 from marshmallow.exceptions import ValidationError
 
 from rotkehlchen.accounting.ledger_actions import LedgerAction, LedgerActionType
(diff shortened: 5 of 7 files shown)


rotkehlchen/api/rest.py
```diff
@@ -4234,3 +4234,41 @@ def get_ens_avatar(self, ens_name: str, match_header: Optional[str]) -> Response
...
+
+                if entry.is_file():
+                    entry.unlink()
+
+            return api_response(OK_RESULT)
+
+        if cache_type == 'icons':
+            for identifier in entries:
+                self.rotkehlchen.icon_manager.delete_icon(Asset(identifier))
+            return api_response(OK_RESULT)
```
nebolax, line 4267: Question. Shouldn't an error response be returned if no icon/avatar was found for the requested identifier?

rotkehlchen/api/rest.py
```diff
@@ -4234,3 +4234,41 @@ def get_ens_avatar(self, ens_name: str, match_header: Optional[str]) -> Response
...
+
+        avatars_to_delete = [cache_dir / f'{i}.png' for i in entries]
```
nebolax, line 4269:
```suggestion
        avatars_to_delete = [cache_dir / f'{ens_name}.png' for ens_name in entries]
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -98,6 +98,7 @@
 from rotkehlchen.utils.misc import create_order_by_rules_list, ts_now
 from rotkehlchen.utils.mixins.serializableenum import SerializableEnumMixin
 
+from ...errors.asset import UnknownAsset
```
nebolax, line 101: I think this should better be an absolute import

rotkehlchen/api/v1/schemas.py
```diff
@@ -2939,7 +2940,7 @@ def __init__(self, db: 'DBHandler') -> None:
         self.db = db
 
     @validates_schema
-    def validate_schema(
+    def validate_schema(  # pylint: disable=no-self-use
```
nebolax, line 2943: Why this change?

rotkehlchen/icons.py
```diff
@@ -91,6 +91,9 @@ def __init__(self, data_dir: Path, coingecko: Coingecko) -> None:
         self.custom_icons_dir.mkdir(parents=True, exist_ok=True)
         self.failed_asset_ids: set[str] = set()
 
+        avatars_dir = self.icons_dir / 'avatars'
```
nebolax, line 94: Why is this needed? Existence of avatars directory is already ensured (at the moment of querying an avatar)

rotkehlchen/tests/api/test_misc.py
```diff
@@ -409,3 +409,84 @@ def test_query_all_chain_ids(rotkehlchen_api_server):
     response = requests.get(api_url_for(rotkehlchen_api_server, 'allevmchainsresource'))
     result = assert_proper_response_with_result(response)
     assert result == [{'id': chain.value, 'name': chain.to_name()} for chain in ChainID]
+
+
+def test_cache_deletion(rotkehlchen_api_server):
```
nebolax, line 414:
```suggestion
def test_images_cache_deletion(rotkehlchen_api_server):
```

nebolax, line 414: Also a small docstring won't hurt

rotkehlchen/tests/api/test_misc.py
```diff
@@ -409,3 +409,84 @@ def test_query_all_chain_ids(rotkehlchen_api_server):
...
+    icons_dir = rotkehlchen_api_server.rest_api.rotkehlchen.icon_manager.icons_dir
+    avatars_dir = icons_dir / 'avatars'
```
nebolax, line 416: Maybe move `avatars_dir` to `icon_manager`? So that it can be accessed as a property too.

rotkehlchen/api/rest.py
prettyirrelevant: I don't think that should be the case since the frontend doesn't know if an avatar/icon exists in the first instance

rotkehlchen/api/v1/schemas.py
prettyirrelevant: yeah, a mistake from my ide during an auto-import

rotkehlchen/tests/api/test_misc.py
prettyirrelevant: hmm okay.

rotkehlchen/api/rest.py
```diff
@@ -4234,3 +4234,41 @@ def get_ens_avatar(self, ens_name: str, match_header: Optional[str]) -> Response
...
+        """Removes cached items from the filesystem."""
+        if cache_type == 'icons':
+            cache_dir = self.rotkehlchen.icon_manager.icons_dir
+        else:  # can only be avatars
+            cache_dir = self.rotkehlchen.icon_manager.icons_dir / 'avatars'
+
+        # delete every file in the cache directory.
+        if entries is None or len(entries) == 0:
+            for entry in cache_dir.iterdir():
+                # don't delete avatars dir when clearing icons cache
```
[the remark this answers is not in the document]
prettyirrelevant: Yeah, this is cool. Was surprised avatars were placed in icons directory. So a data migration.

rotkehlchen/tests/api/test_misc.py
prettyirrelevant: But that has nothing to do with icons, so I'm not so sure about that.

rotkehlchen/api/rest.py
nebolax: Yes, probably makes sense. But pinging @rotki/frontend just to make sure.

rotkehlchen/tests/api/test_misc.py
nebolax: Yes, has nothing to do with icons indeed, however our avatars logic is already tied to icons logic. So I would ask @LefterisJP about this

docs/api.rst
```diff
@@ -12408,3 +12408,36 @@ Get avatar for an ens name
    :statuscode 400: Invalid ens name provided (it doesn't end with `.eth`).
    :statuscode 404: There is no avatar set for the given ens name.
    :statuscode 409: Avatar was requested for an ens name that is not currently in the database or we couldn't query the blockchain.
+
+
+Clear Icons/Avatars Cache
+============================
+
+.. http:post:: /api/(version)/cache/images/<cache_type>/clear
```
LefterisJP, line 12416: If you are adding a new `/cache` endpoint perhaps it would make sense to move other cache/data cleaning there?

docs/api.rst
```diff
@@ -12408,3 +12408,36 @@ Get avatar for an ens name
...
+   **Example Request**
+
+    .. http:example:: curl wget httpie python-requests
+
+      POST /api/1/cache/icons/clear HTTP/1.1
+      Host: localhost:5042
+      Content-Type: application/json;charset=UTF-8
+
+      {
+         "entries": ["ETH", "BTC"]
```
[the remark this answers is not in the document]
LefterisJP: no it's possible and you can keep the validation. You just make the sole argument be a list. But this is also fine. No need to change.

docs/api.rst
```diff
@@ -12408,3 +12408,36 @@ Get avatar for an ens name
...
+   **Example Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+   :statuscode 200: The cache items were cleared successfully.
+   :statuscode 400: Entries provided do not match the ``cache_type`` specified. Invalid ``cache_type`` provided.
+    :statuscode 500: Internal rotki error
```
LefterisJP, line 12443: At least in github preview this seems like it has an extra space?

```suggestion
   :statuscode 500: Internal rotki error
```

rotkehlchen/api/rest.py
```diff
@@ -4196,7 +4196,7 @@ def get_ens_avatar(self, ens_name: str, match_header: Optional[str]) -> Response
 
         Also supports etag mechanism that helps with caching on the client side.
         """
-        avatars_dir = self.rotkehlchen.data_dir / 'icons' / 'avatars'
+        avatars_dir = self.rotkehlchen.data_dir / 'images' / 'avatars'
```
LefterisJP, line 4199: What are you doing? Why are you renaming this? Undo. The `icons` directory already exists in all the user's systems. You are not going to create another one. If we ever wanted to do changes there we would need a migration.

rotkehlchen/api/rest.py
LefterisJP: For these stuff ask in discord. Seeing a ping here may take time. I also think it's fine if there is nothing special returned in case missing. Result is the same. No file for this identifier.

rotkehlchen/api/server.py
```diff
@@ -314,6 +315,7 @@
         'per_timestamp_db_snapshots_resource',
     ),
     ('/notes', UserNotesResource),
+    ('/cache/<string:cache_type>/clear', ClearCacheResource),
```
LefterisJP, line 318: This looks different to the endpoint you described in api.rst. There you wrote:  

```
http:post:: /api/(version)/cache/images/<cache_type>/clear
```

```suggestion
    ('/cache/images/<string:cache_type>/clear', ClearImagesCacheResource),
```

So which one is it? Do you need the images or not? I would guess not, since we probably can keep one such endpoint for all types of caches?

rotkehlchen/data_migrations/migrations/migration_10.py
```diff
@@ -0,0 +1,29 @@
+import logging
+import shutil
+from typing import TYPE_CHECKING
+
+from rotkehlchen.logging import RotkehlchenLogsAdapter
```
LefterisJP, line 5: Is this needed? Why can't you have `avatars` under `icons` for now?

I would prefer to do these kind of changes all together and also solve https://github.com/rotki/rotki/issues/4841 while we are at it.

rotkehlchen/tests/api/test_misc.py
```diff
@@ -409,3 +409,85 @@ def test_query_all_chain_ids(rotkehlchen_api_server):
     response = requests.get(api_url_for(rotkehlchen_api_server, 'allevmchainsresource'))
     result = assert_proper_response_with_result(response)
     assert result == [{'id': chain.value, 'name': chain.to_name()} for chain in ChainID]
+
+
+def test_images_cache_deletion(rotkehlchen_api_server):
```
LefterisJP, line 414: Set a temporary directory for this test since you are going to be creating and deleting files in that directory. 
```suggestion
@pytest.mark.parametrize('use_clean_caching_directory', [True])
def test_images_cache_deletion(rotkehlchen_api_server):
```

rotkehlchen/tests/api/test_misc.py
```diff
@@ -409,3 +409,85 @@ def test_query_all_chain_ids(rotkehlchen_api_server):
...
+            'clearcacheresource',
+            cache_type='icons',
+        ),
+    )
+    assert_proper_response(response)
+    assert len([i for i in icons_dir.iterdir() if i.is_file()]) == 0
+    assert len([i for i in avatars_dir.iterdir() if i.is_file()]) == 1
+
+    # test avatars cache
+    # populate avatars dir
```
LefterisJP, line 461:
```suggestion
    # populate avatars dir to test the cache deletion
```

rotkehlchen/tests/api/test_misc.py
```diff
@@ -409,3 +409,85 @@ def test_query_all_chain_ids(rotkehlchen_api_server):
...
+    with open(f'{icons_dir}/ETH_small.png', 'wb') as f:
+        f.write(b'')
+
+    with open(f'{icons_dir}/BTC_small.png', 'wb') as f:
+        f.write(b'')
+
+    with open(f'{icons_dir}/AVAX_small.png', 'wb') as f:
+        f.write(b'')
+
+    # make sure this is not deleted.
```
LefterisJP, line 430:
```suggestion
    # also add an avatar, to make sure it's not deleted when icons are
```

LefterisJP (commented):
1. Many many tests are failing. Please don't send me PRs when this is the case.
2. Left some comments. Important: I don't think we need the new images directory yet. There is more stuff (linked an issue) we need to change at some point.

docs/api.rst
```diff
@@ -12408,3 +12372,40 @@ Get avatar for an ens name
...
+
+      POST /api/1/cache/icons/clear HTTP/1.1
+      Host: localhost:5042
+      Content-Type: application/json;charset=UTF-8
+
+      {
+         "icons": ["ETH", "BTC"]
+      }
+
+   :reqjsonarr optional[string] icons: An array of the ass to be cleared from the cache. All items are deleted in the cache if ``null``. Only works with ``cache_type=icons``.
```
LefterisJP, line 12396: ass?

```suggestion
   :reqjsonarr optional[string] icons: An array of the ??? to be cleared from the cache. All items are deleted in the cache if ``null``. Only works with ``cache_type=icons``.
```

docs/api.rst
```diff
@@ -2101,42 +2101,6 @@ Request creation of oracle price cache
    :statuscode 500: Internal rotki error
    :statuscode 502: The oracle could not be queried due to an error on their side.
 
-Delete an oracle price cache
```
LefterisJP, line 2104: This makes no sense. I did not mean to do it now. I meant to do it perhaps at some other point. This endpoint already exists and is used with other verbs. This change you are doing does not make sense then as is. Please ask before goign and doing stuff like that.

rotkehlchen/api/v1/resources.py
```diff
@@ -2858,3 +2847,30 @@ class EnsAvatarsResource(BaseMethodView):
     @use_kwargs(get_schema, location='view_args')
     def get(self, ens_name: str) -> Response:
         return self.rest_api.get_ens_avatar(ens_name=ens_name, match_header=get_match_header())
+
+
+class ClearCacheResource(BaseMethodView):
+    post_schema = ClearCacheSchema()
+    icons_schema = ClearIconsCacheSchema()
+    avatars_schema = ClearAvatarsCacheSchema()
+    price_schema = NamedOracleCacheSchema()
```
LefterisJP, line 2856: Undo this entire change please. It's not a good place while keeping the previous endpoint for price oracle caches for everything else

LefterisJP (commented):
No no no. What you did by putting that price oracle cache in the new endpoint makes no sense.

**PLEASE** ask in discord either your colleagues or me when you try to do something like that.

You are wasting your time and mine by doing work that is not needed.

Why is it not a good idea?

1. You do not remove any endpoint. The endpoint for price oracle cache still exists and is accessed by other verbs. So you are force-moving a certain sub functionality only.
2. The schema verification becomes an ugly code salad. Instead of having the verification per endpoint nicely and clean you start some dynamic magic.

rotkehlchen/api/rest.py
kelsos: We don't need any error code if the icon was missing in the first place. (Anything other than 200).

The only case IMO that would make sense to get an error is if the deletion fails, just to notify the user that the action was not successful.

rotkehlchen/api/v1/schemas.py
```diff
@@ -2997,3 +2998,32 @@ def make_binance_savings_query(
 
 class EnsAvatarsSchema(Schema):
     ens_name = fields.String(required=True, validate=lambda x: x.endswith('.eth'))
+
+
+class ClearCacheSchema(Schema):
+    cache_type = fields.String(
+        required=True,
+        validate=[validate.OneOf({'icons', 'avatars', 'prices'})],
```
LefterisJP, line 3006:
```suggestion
        validate=[validate.OneOf({'icons', 'avatars')],
```

PR #5816, opened 2023-03-27
feat: setting to render NFTs image

### LINKED ISSUE #3975: Rendering images hosted by a malicious 3rd party can lead to ip leaks
## Problem Definition

The blog post author [here](https://medium.com/@alxlpsc/critical-privacy-vulnerability-getting-exposed-by-metamask-693c63c2ce94) describes a scenario where a malicious actor creates an NFT with an image link that points to an image hosted by a malicious server. That NFT is specially prepared and "donated" to a specific ethereum address.

When the owner of that addressr (or anybody else) goes to render that image from a wallet, NFT viewer, aggregator then the ip is logged. If that rendering is done client side, as for example in metamask or rotki then the logged ip would be the ip of the user. The malicious server can then match ip + timestamp + ethereum address.

That is not a good thing to leak and is quite sensitive info as also explained in the blogpost.

## Task

The way to address this would be:  (if anyone has other ideas all ears!)

1. Only render ipfs links by default. Since that scenario won't work with ipfs. Any other hosting methodology that we can have an exception for?
2. For all others, explicitly ask the user if they trust that url. Put it in some kind of whitelist. Have that list editable by the user at all times. At the popup prompt we should also give them an option to go and set (3)
3. Have a catch-all setting to just blindly trust all 3rd party images. Should be off by default. At the place where this setting is set clearly explain the dangers.

lukicenturi:
Closes #3975

image
image


frontend/app/src/types/user.ts
```diff
@@ -9,12 +9,12 @@ import { LedgerActionEnum } from '@/types/history/ledger-action/ledger-actions-t
...
-import { snakeCaseTransformer } from '@/services/axios-tranformers';
+import { camelCaseTransformer } from '@/services/axios-tranformers';
 
 const OtherSettings = z.object({
   krakenAccountType: KrakenAccountType.optional(),
   frontendSettings: z.string().transform(arg => {
-    const data = arg ? snakeCaseTransformer(JSON.parse(arg)) : {};
+    const data = arg ? camelCaseTransformer(JSON.parse(arg)) : {};
     return FrontendSettings.parse(data);
   }),
```
kelsos, line 19: hmmm, don't we have some unit tests that cover this scenario?

frontend/app/src/plugins/vuetify.ts
```diff
@@ -87,7 +87,6 @@ export default new Vuetify({
         graphFade: '#ffffff',
         error: '#f03a47',
         success: '#06D6A0',
-        border: '#0000001f',
```
kelsos, line 90: wasn't this added recently, why is it removed again?

frontend/app/src/store/assets/nft.ts
```diff
@@ -0,0 +1,79 @@
+import { type ActionResult } from '@rotki/common/lib/data';
+import { getDomain } from 'tldts';
```
kelsos, line 2: Is this the only function we use from the `tldts` library? any reason why we need a new dependency and why we can't use a regex for the same purpose?

frontend/app/src/components/nft/NftGalleryItem.vue
```diff
@@ -11,43 +13,133 @@ const props = defineProps({
...
+
+const imageUrlSource: ComputedRef<string | null> = computed(
+  () => get(item).imageUrl
+);
+
+watch(
+  imageUrlSource,
+  image => {
+    if (image) {
+      addToListImagesDomain(image);
```
kelsos, line 31: Why do we need this?

frontend/app/src/components/settings/general/nft/NftImageRenderingSetting.vue
```diff
@@ -0,0 +1,80 @@
...
+        v-model="renderAllNftImages"
+        :success-messages="success"
+        :error-messages="error"
+        @change="update"
+      >
+        <v-radio label="Allow all images" :value="true" />
+        <v-radio
+          label="Only allow images from whitelisted domains"
+          :value="false"
+        />
```
kelsos, line 55: Can you extract the text here to the resource files too?

frontend/app/src/types/user.ts
lukicenturi: the test we have doesn't test this part :(
I'll update the test

frontend/app/src/store/assets/nft.ts
lukicenturi: regex doesn't work because some url can have subdomain, and have more than 1 part suffix (.co.id)
the worst case looks like this 
Input: `https://subdomain.domain.co.id`
Output: `domain.co.id`

If we want to make it right, we need to list all possible suffixes `.co.id`, `.co.uk`.

The best way we can do with regex is only get this output:
`subdomain.domain.co.id`

The problem is if we have multiple image with same domain but different subdomain, we need to allow it separately.

frontend/app/src/components/nft/NftGalleryItem.vue
kelsos: Can the user add a domain that they know, which doesn't already appear in the gallery?

For this in general I would turn tracking and auto-suggestion off and allow the users to input anything that checks out as a valid domain regardless.

For specifics from the screenshot, I can see that you already added a click to trust so that would be enough. 

Also, maybe for the whitelist mode, we should add a confirmation step, just to avoid accidentally whitelisting domains, since this might have implications

frontend/app/src/store/assets/nft.ts
kelsos: How does the library do it? is it overly complicated?

Maybe we could do it ourselves? It feels a bit weird importing a huge library just for a method.

frontend/app/src/components/nft/NftGalleryItem.vue
lukicenturi: Yes they can add the domain, outside from the given list.
For the confirmation, yeah I think it makes sense.

kelsos:
@lukicenturi what is remaining for this? is it ready to re-check?

frontend/app/src/locales/en.json
```diff
@@ -3369,7 +3388,7 @@
       "sync": "Allow data sync with rotki Server"
     },
     "delete_confirmation": {
-      "message": "Are you sure you want to delete the rotki premium keys for your account? If you want to re-enable premium you will have to enter your keys again.",
+      "message": "Are you sure you want to delete the rotki premium keys for your account? If you want to re-enable premium you will have to enter your keys again. \n asdfasdfsa \n asdfasfd \n",
```
kelsos, line 3391: 😄 I think we should remove this 😛

frontend/app/src/components/nft/NftGalleryItem.vue
```diff
@@ -11,41 +13,145 @@ const props = defineProps({
...
```
kelsos, line 31: @lukicenturi what I meant with the original comment was to remove the automated domain tracking completely, and instead have either the whitelisting from the gallery item or manually add a domain from the settings.

Also for the settings component the confirmation feels a bit weird.
The chip is added int he component and only when I try to leave do I get the confirmation. Can we change that?

kelsos: Also, we need to do something about videos, this is how they appear
image

frontend/app/src/data/psl.ts
kelsos, line 1: Any chance we could keep this as a `json` file?

frontend/app/src/locales/en.json
lukicenturi: I'm so ashamed 🥲🥲🥲🥲

frontend/app/src/data/psl.ts
lukicenturi: Hmm but then I need to recreate the `Set` every time I need it (?) 🧐

lukicenturi:
> Another question is do you think it would be possible to make it so that you can whitelist a domain from the image component in the nonfungible balances?

Yeah it can be worked on

frontend/app/src/components/nft/NftGalleryItem.vue
lukicenturi: > @lukicenturi what I meant with the original comment was to remove the automated domain tracking completely, and instead have either the whitelisting from the gallery item or manually add a domain from the settings.
> 
> Also for the settings component the confirmation feels a bit weird. The chip is added in the component and only when I try to leave do I get the confirmation. Can we change that?

Okay I get it,
for the confirmation I think onButtonClick will be better than onBlur

frontend/app/src/data/psl.ts
kelsos: what about importing the array from the JSON file and passing it to set?

frontend/app/src/components/nft/NftGallery.vue
```diff
@@ -148,7 +147,7 @@ const collections = computed(() => {
     .filter(uniqueStrings);
 });
 
-const { fetchNfts: nftFetch } = useNft();
+const { fetchNfts: nftFetch } = useNftsStore();
```
lukicenturi, line 150: wait this should be composable instead

frontend/app/src/components/nft/NftGalleryItem.vue
```diff
@@ -11,41 +13,147 @@ const props = defineProps({
...
+                v-if="isMediaVideo"
+                controls
+                width="auto"
+                :src="imageUrl"
+                :style="
+                  renderImage
+                    ? {
+                        'background-color': `#${item.backgroundColor}`
+                      }
+                    : {}
```
kelsos, line 113: this style is the same as the one bellow right? any chance we can move it to a computed prop?

kelsos (commented):
Ok, so I left a small comment, but overall it is good.

Also, I am thinking, if it would make sense for the table to either show the domain as part of the tooltip or even make it so that you can copy it, what do you think?

For the gallery, you have the whitelist button which makes things easier, but for the table, while you have access to the settings you still don't know the domain.

kelsos:
> > Ok, so I left a small comment, but overall it is good.
> > Also, I am thinking, if it would make sense for the table to either show the domain as part of the tooltip or even make it so that you can copy it, what do you think?
> > For the gallery, you have the whitelist button which makes things easier, but for the table, while you have access to the settings you still don't know the domain.
> 
> So for the table, I make the image clickable, to add it to the whitelisted domains

As long as you have to confirm first that should be fine :)

lukicenturi:
> > > Ok, so I left a small comment, but overall it is good.
> > > Also, I am thinking, if it would make sense for the table to either show the domain as part of the tooltip or even make it so that you can copy it, what do you think?
> > > For the gallery, you have the whitelist button which makes things easier, but for the table, while you have access to the settings you still don't know the domain.
> > 
> > 
> > So for the table, I make the image clickable, to add it to the whitelisted domains
> 
> As long as you have to confirm first that should be fine :)

It's updated now

PR #5820, opened 2023-03-28
added a cache table to globaldb for next upgrade

### LINKED ISSUE #5631: Create a key-value cache for global DB
## Abstract

At the moment the `general_cache` table in the global DB is the only cache table. Its purpose is for multiple key-value combinations which are not unique. So it's an association cache. The primary key **is not** the key but the key-value pair. Allowing for multiple values per key. So it's more like a multu-value cache.

We also have cases like for the ilk_registry, and other, where the cache should really be unique per key but we just use the same table due to lack of another one.

The problem there is that by using the wrong type table for a unique-key cache is we have to clean it if we use general cache before. If we don't we will end up having multiple values per key, with new ones not replacing old one, but just adding on top growing the DB size.

## Task

Add a global DB upgrade (should not affect assets, so it should be included as okay in the asset updates) that adds a new table and moves the cache data that should be unique per key there.

For example:

```
CURVE_POOL_TOKENS0x9f4A88da14F2b6DBc785C1Db3511A53B8F342bde|0x8b921e618dD3Fa5a199b0a8B7901f5530D74EF27|1676727187
CURVE_POOL_TOKENS0x9f4A88da14F2b6DBc785C1Db3511A53B8F342bde|0x6c3F90f043a72FA612cbac8115EE7e52BDe6E490|1676727187
CURVE_POOL_ADDRESS0x04c90C198b2eFF55716079bc06d7CCc4aa4d7512|0x04c90C198b2eFF55716079bc06d7CCc4aa4d7512|1676727187
```

`CURVE_POOL_TOKENS` makes sense to use the multivalue cache as it has one mapping per token, showing which tokens a pool hs.

But `CURVE_POOL_ADDRESS` does not really need to do that. It's just a list of all curve pool addresses.

Juyotal:
Closes #5631 

## Checklist

- [ ] Added a global DB upgrade that adds a new cache table and moves the cache data that should be unique per key there.


docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`5798` Added a global DB upgrade that adds a new cache table and moves the cache data that should be unique per key there.
```
yabirgb, line 5: We only add to the changelog changes that are useful to know for the user. In this case is a internal refactor and is not relevant for the user. Please remove it

rotkehlchen/globaldb/cache.py
```diff
@@ -7,6 +7,12 @@
 from rotkehlchen.utils.misc import ts_now
 
 
+UNIQUE_CACHE = 'unique_cache'
+GENERAL_CACHE = 'general_cache'
+UNIQUE_CACHE_KEYS = ['TEST_KEY']
+# should be ["CURVE_POOL_ADDRESS", "MAKERDAO_VAULT_ILK", "CURVE_GAUGE_ADDRESS"]
```
yabirgb, line 13: why having test_key here? shouldn't it take the values in the comment? Edit this and adjust where needed

rotkehlchen/globaldb/cache.py
```diff
@@ -29,37 +35,47 @@ def globaldb_set_general_cache_values_at_ts(
     """Function to update cache in globaldb. Inserts all values paired with the cache key.
     If any entry exists, overwrites it."""
     cache_key = _compute_cache_key(key_parts)
+    unique_cache = any(cache_key.startswith(key_part) for key_part in UNIQUE_CACHE_KEYS)
+    if unique_cache:
+        cache = UNIQUE_CACHE
+    else:
+        cache = GENERAL_CACHE
```
yabirgb, line 42: since you check `startswith` it means that `GeneralCacheType` is the first part of the cache so you can change it and make `UNIQUE_CACHE_KEYS` a `set[GeneralCacheType]` and check `unique_cache = key_parts[0] in UNIQUE_CACHE_KEYS`

rotkehlchen/globaldb/cache.py
```diff
@@ -29,37 +35,47 @@ def globaldb_set_general_cache_values_at_ts(
...
     tuples = [(cache_key, value, timestamp) for value in values]
     write_cursor.executemany(
-        'INSERT OR REPLACE INTO general_cache '
+        f'INSERT OR REPLACE INTO {cache} '
```
yabirgb, line 45: can you change `cache` to `cache_table`?

rotkehlchen/globaldb/cache.py
```diff
@@ -89,8 +105,13 @@ def globaldb_get_general_cache_last_queried_ts_by_key(
     element returns Timestamp(0)
     """
     cache_key = _compute_cache_key(key_parts)
+    unique_cache = any(cache_key.startswith(key_part) for key_part in UNIQUE_CACHE_KEYS)
+    if unique_cache:
+        cache = UNIQUE_CACHE
+    else:
+        cache = GENERAL_CACHE
```
yabirgb, line 112: why not add this logic to `_compute_cache_key` and avoid repeating it in each function? I'm thinking in changing the return type to `cache_key, cache_type` and avoid repeating this in every function that uses it

rotkehlchen/globaldb/cache.py
```diff
@@ -100,15 +121,21 @@ def globaldb_get_general_cache_last_queried_ts_by_key(
...
-def globaldb_delete_general_cache(
+def globaldb_delete_cache(
         write_cursor: DBCursor,
         key_parts: Iterable[Union[str, GeneralCacheType]],
         value: Optional[str] = None,
 ) -> None:
-    """Deletes from cache in globaldb. If value is None deletes all entries with the provided
-    cache key. Otherwise, deletes only single entry `key - value`."""
+    """Deletes from cache in globaldb. In the case of the general cache,
+    If value is None deletes all entries with the provided cache key.
```
yabirgb, line 130: general cache shouldn't contain anymore multivalues right? it can be deleted from the docstring

rotkehlchen/tests/unit/globaldb/test_upgrades.py
```diff
@@ -328,19 +339,24 @@ def test_upgrade_v4_v5(globaldb):
...
     assert globaldb.get_setting_value('version', None) == 5
-
+    # insert some values in unique_cache table.
+    with globaldb.conn.write_ctx() as write_cursor:
+        # check that values are replaced and not appended
+        globaldb_set_cache_values(
+            write_cursor=write_cursor,
+            key_parts=['TEST_KEY', '123'],
+            values=['klm'],
+        )
```
yabirgb, line 350: we should have already tests for this function in other places. I believe is better to move this part of the test to that file. Also I'm not sure why you are using `TEST_KEY`, that value is not something that should go to the project

rotkehlchen/tests/unit/globaldb/test_upgrades.py
```diff
@@ -328,19 +339,24 @@ def test_upgrade_v4_v5(globaldb):
...
     with globaldb.conn.read_ctx() as cursor:
+        # check again existence of default_rpc_nodes table
         cursor.execute(
             'SELECT COUNT(*) FROM sqlite_master WHERE type="table" and name=?',
             ('default_rpc_nodes',),
         )
         assert cursor.fetchone()[0] == 1
-
```
yabirgb, line 340: restore this deleted lines

rotkehlchen/tests/unit/globaldb/test_upgrades.py
```diff
@@ -353,7 +369,18 @@ def test_upgrade_v4_v5(globaldb):
...
+        cursor.execute(
+            'SELECT COUNT(*) FROM sqlite_master WHERE type="table" and name=?',
+            ('unique_cache',),
+        )
+        assert cursor.fetchone()[0] == 1
+        value = globaldb_get_cache_values(
+            cursor=cursor,
+            key_parts=['TEST_KEY', '123'],
+        )
+        assert value == ['klm']
```
yabirgb, line 382: As before I wouldn't test this here. We had a function here because we are making changes to the curve cache but is a different topic

rotkehlchen/globaldb/schema.py
```diff
@@ -273,6 +273,15 @@
 );
 """
 
+DB_CREATE_UNIQUE_CACHE = """
+CREATE TABLE IF NOT EXISTS unique_cache (
+    key TEXT NOT NULL PRIMARY KEY,
+    value TEXT NOT NULL,
```
yabirgb, line 279: I would add also a special test that checks the behavior with the primary key

rotkehlchen/globaldb/cache.py
yabirgb, line 45: This can now raise a new error if there is a conflict with the primary key. Is something that needs to be handled in the 3 keys where it can happen. It should be `sqlcipher.IntegrityError`

rotkehlchen/tests/unit/globaldb/test_upgrades.py
```diff
@@ -309,17 +313,24 @@ def test_upgrade_v4_v5(globaldb):
...
         cursor.execute(
             'SELECT COUNT(*) FROM sqlite_master WHERE type="table" and name=?',
             ('default_rpc_nodes',),
         )
         assert cursor.fetchone()[0] == 0
-        last_queried_ts = globaldb_get_general_cache_last_queried_ts_by_key(
+        # check existence of unique_cache table
+        cursor.execute(
+            'SELECT COUNT(*) FROM sqlite_master WHERE type="table" and name=?',
+            ('unique_cache',),
```
yabirgb, line 325: You have a constant for this, use it
```suggestion
            (UNIQUE_CACHE,),
```

yabirgb (commented):
Thank you for your contribution! I left some comments that need to be addressed. Also there was a recent change in this part of the code that would affect. You will need to rebase first and address the possible conflicts

PR #5821, opened 2023-03-28
Query balances in curve/convex gauges

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/accounting/structures/base.py
+++ b/rotkehlchen/accounting/structures/base.py
@@ -325,7 +325,7 @@ class HistoryEvent(HistoryBaseEntry):
         )
 
     def __repr__(self) -> str:
-        return f'HistoryEvent({",".join(self._history_base_entry_repr_fields())})'
+        return f'HistoryEvent({", ".join(self._history_base_entry_repr_fields())})'
 
     def serialize_for_db(self) -> tuple[HISTORY_EVENT_DB_TUPLE_WRITE]:
         return (self._serialize_base_tuple_for_db(HistoryBaseEntryType.BASE_ENTRY),)
--- a/rotkehlchen/chain/ethereum/modules/convex/balances.py
+++ b/rotkehlchen/chain/ethereum/modules/convex/balances.py
@@ -1,0 +1,56 @@
+from collections import defaultdict
+from typing import TYPE_CHECKING
+
+from rotkehlchen.accounting.structures.balance import Balance
+from rotkehlchen.accounting.structures.evm_event import EvmProduct
+from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.assets.asset import EvmToken
+from rotkehlchen.chain.ethereum.interfaces.balances import ProtocolWithBalance
+from rotkehlchen.chain.ethereum.modules.curve.balances import query_gauges_balances
+from rotkehlchen.chain.evm.tokens import get_chunk_size_call_order
+from rotkehlchen.db.dbhandler import DBHandler
+from rotkehlchen.types import ChecksumEvmAddress
+
+from .constants import CPT_CONVEX
+
+if TYPE_CHECKING:
+    from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
+
+
+class ConvexBalances(ProtocolWithBalance):
+    """Query staked balances in the Convex gauges"""
+
+    def __init__(self, database: DBHandler, evm_inquirer: 'EvmNodeInquirer'):
+        super().__init__(database=database, evm_inquirer=evm_inquirer, counterparty=CPT_CONVEX)
+
+    def query_balances(self) -> dict[ChecksumEvmAddress, dict[EvmToken, Balance]]:
+        balances: dict[ChecksumEvmAddress, dict[EvmToken, Balance]] = {}
+        gauges_to_token: dict[ChecksumEvmAddress, EvmToken] = {}
+
+        # query addresses and gauges where they interacted
+        addresses_with_deposits = self.addresses_with_deposits(
+            product=EvmProduct.CONVEX_GAUGE,
+            deposit_events={(HistoryEventType.DEPOSIT, HistoryEventSubType.NONE)},
+        )
+
+        # get details to query balances on chain
+        chunk_size, call_order = get_chunk_size_call_order(self.evm_inquirer)
+        for address, events in addresses_with_deposits.items():
+            balances[address] = defaultdict(Balance)
+            # Create a mapping of gauge to its token
+            for event in events:
+                if event.address is None:
+                    continue
+                if event.extra_data is None or (gauge_address := event.extra_data.get('gauge_address')) is None:  # noqa: E501
+                    continue
+                gauges_to_token[gauge_address] = event.asset.resolve_to_evm_token()
+
+            balances[address] = query_gauges_balances(
+                user_address=address,
+                gauges_to_token=gauges_to_token,
+                call_order=call_order,
+                chunk_size=chunk_size,
+                balances_contract=self._get_staking_contract_balances,
+            )
+
+        return balances
--- a/rotkehlchen/chain/ethereum/modules/convex/constants.py
+++ b/rotkehlchen/chain/ethereum/modules/convex/constants.py
@@ -1,6 +1,7 @@
+from typing import Final
 from rotkehlchen.chain.evm.types import string_to_evm_address
 
-CPT_CONVEX = 'convex'
+CPT_CONVEX: Final = 'convex'
 
 BOOSTER = string_to_evm_address('0xF403C135812408BFbE8713b5A23a04b3D48AAE31')
 CVX_LOCKER = string_to_evm_address('0xD18140b4B819b895A3dba5442F959fA44994AF50')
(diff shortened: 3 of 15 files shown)


rotkehlchen/chain/ethereum/interfaces/balance.py
```diff
@@ -0,0 +1,51 @@
...
+    from rotkehlchen.accounting.structures.evm_event import EvmProduct
+
+
+class ProtocolWithBalance:
+
+    def __init__(self, counterparty: str, database: 'DBHandler'):
+        self.counterparty = counterparty
+        self.event_db = DBHistoryEvents(database)
+
+    def addresses_with_deposits(
```
nebolax, line 22: Hmmm I am not so sure if this needs to be in a class if it is the only function here. Maybe it can just be a standalone function?

rotkehlchen/chain/evm/tokens.py
```diff
@@ -155,6 +155,41 @@ def _get_token_balances(
             balances[token] += normalized_balance
         return balances
 
+    def get_staking_contract_balances(
```
nebolax, line 158: Questions.

1. Will it also be used for convex, or only for curve?
2. If only for curve, then maybe move contents of this function to the curve-specific place?
3. If for both, then maybe move the function to `ProtocolWithBalance` class or somewhere like that? `evm/tokens.py` sounds too generic to me to put staking-specific functionality here. I might be wrong though.

yabirgb: Is used both by convex and curve. I've moved it to a private method

rotkehlchen/chain/ethereum/interfaces/balance.py
yabirgb: I added the other method of the interface. Having here an interface is good for future protocols that we can add imo

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -797,7 +797,7 @@ def test_deposit_multiple_tokens(ethereum_transaction_decoder, ethereum_accounts
 @pytest.mark.vcr()
 @pytest.mark.parametrize('ethereum_accounts', [['0xd289986c25Ae3f4644949e25bC369e9d8e0caeaD']])
 def test_gauge_deposit(ethereum_transaction_decoder, ethereum_accounts):
-    tx_hex = deserialize_evm_tx_hash('0x5ae70d68241d85feac65c90e4546154e232dba9fecad9036bcec10082acc9d46 ')  # noqa: E501
+    tx_hex = deserialize_evm_tx_hash('0x5ae70d68241d85feac65c90e4546154e232dba9fecad9036bcec10082acc9d46')  # noqa: E501
```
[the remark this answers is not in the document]
yabirgb: The failing tests were because I didn't have the branch of cassettes updated. It is now

rotkehlchen/chain/ethereum/interfaces/balance.py
nebolax: Yes, makes sense

rotkehlchen/chain/ethereum/modules/convex/decoder.py
```diff
@@ -71,19 +75,42 @@ def _decode_convex_events(self, context: DecoderContext) -> DecodingOutput:
...
                         event.notes = f'Withdraw {event.balance.amount} {crypto_asset.symbol} from convex {CONVEX_POOLS[context.tx_log.address]} pool'  # noqa: E501
+                        event.product = EvmProduct.CONVEX_GAUGE
                     else:
                         event.notes = f'Withdraw {event.balance.amount} {crypto_asset.symbol} from convex'  # noqa: E501
-                    event.counterparty = CPT_CONVEX
+                        if (
+                            isinstance(crypto_asset, EvmToken) and
+                            crypto_asset.protocol == CURVE_POOL_PROTOCOL
+                        ):
+                            event.product = EvmProduct.CONVEX_GAUGE
```
nebolax, line 113: I would adjust notes here

rotkehlchen/chain/ethereum/modules/convex/balances.py
```diff
@@ -0,0 +1,67 @@
...
+
+                # Now map the gauge to the underlying token
+                for lp_token, balance in gauges_balances.items():
+                    lp_token_price = Inquirer().find_usd_price(lp_token)
+                    balances[address][lp_token] += Balance(
+                        amount=balance,
+                        usd_value=lp_token_price * balance,
+                    )
+
+        return balances
```
nebolax, line 67: This part (from line ~36 to line 67) looks like it can be abstracted, can't it?

rotkehlchen/chain/ethereum/modules/curve/balances.py
```diff
@@ -0,0 +1,89 @@
...
+from .constants import CPT_CURVE
+
+if TYPE_CHECKING:
+    from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
+
+
+def query_gauges_balances(
+        user_address: ChecksumEvmAddress,
+        gauges_to_token: dict[ChecksumEvmAddress, EvmToken],
+        call_order: list[WeightedNode],
```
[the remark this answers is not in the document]
yabirgb: This is executed inside a for loop for each address tracked so to avoid executing that logic I calculate it once and reuse it for each address tracked

rotkehlchen/chain/ethereum/modules/convex/decoder.py
LefterisJP: Not even a pool address? @yabirgb that sounds weird. Where from do you withdraw this amount? What is this `else` clause?

rotkehlchen/chain/ethereum/modules/curve/balances.py
```diff
@@ -0,0 +1,89 @@
...
+            lp_token_price = Inquirer().find_usd_price(lp_token)
+            balances[lp_token] = Balance(
+                amount=balance,
+                usd_value=lp_token_price * balance,
+            )
+
+    return balances
+
+
+class CurveBalances(ProtocolWithBalance):
```
LefterisJP, line 55: just a short docstring of the protocol and function of this class

rotkehlchen/chain/ethereum/modules/curve/balances.py
```diff
@@ -0,0 +1,89 @@
...
+        # query addresses and gauges where they interacted
+        addresses_with_deposits = self.addresses_with_deposits(
+            product=EvmProduct.CURVE_GAUGE,
+            deposit_events={(HistoryEventType.DEPOSIT, HistoryEventSubType.DEPOSIT_ASSET)},
+        )
+
+        # get details to query balances on chain
+        chunk_size, call_order = get_chunk_size_call_order(self.evm_inquirer)
+
+        for address, events in addresses_with_deposits.items():
```
LefterisJP, line 73:
```suggestion
        # get details to query balances on chain
        chunk_size, call_order = get_chunk_size_call_order(self.evm_inquirer)
        for address, events in addresses_with_deposits.items():
```

rotkehlchen/chain/ethereum/modules/curve/balances.py
```diff
@@ -0,0 +1,89 @@
...
+            balances[address] = defaultdict(Balance)
+            # Create a mapping of gauge to its token
```
LefterisJP, line 75:
```suggestion
            # Create a mapping of a gauge to its token
```

rotkehlchen/chain/ethereum/modules/curve/balances.py
```diff
@@ -0,0 +1,89 @@
...
+def query_gauges_balances(
+        user_address: ChecksumEvmAddress,
+        gauges_to_token: dict[ChecksumEvmAddress, EvmToken],
+        call_order: list[WeightedNode],
+        chunk_size: int,
+        balances_contract: Callable,
+) -> dict[EvmToken, Balance]:
+    """
+    Query the set of gauges in gauges_to_token and return the balances for each
+    lp token desposited in the gauge.
```
LefterisJP, line 31:
```suggestion
    lp token desposited in all gauges.
```

rotkehlchen/chain/evm/tokens.py
```diff
@@ -97,6 +97,18 @@ def generate_multicall_chunks(
     return multicall_chunks
 
 
+def get_chunk_size_call_order(evm_inquirer: 'EvmNodeInquirer') -> tuple[int, list[WeightedNode]]:
```
LefterisJP, line 100: Add a docstring. Also move the `# skipping etherscan because chunk size is too big for etherscan` in that docstring

rotkehlchen/db/filtering.py
```diff
@@ -281,6 +282,14 @@ def prepare(self) -> tuple[list[str], list[Any]]:
         return ['location=?'], [self.location.serialize_for_db()]
 
 
+@dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=False)
+class DBProductFilter(DBFilter):
```
LefterisJP, line 286: Do you need to create a new filter? Can't you use `DBEqualsFilter`? I don't think we should be creating new filters for everything but try to consolidate

LefterisJP (commented):
Thanks. Left a review

rotkehlchen/accounting/structures/base.py
```diff
@@ -325,7 +325,7 @@ def __init__(
         )
 
     def __repr__(self) -> str:
-        return f'HistoryEvent({",".join(self._history_base_entry_repr_fields())})'
+        return f'HistoryEvent({", ".join(self._history_base_entry_repr_fields())})'
```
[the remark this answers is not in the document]
yabirgb: is because when you show an evm using the repr the fields are compated in the form `field_value,self.other_field=` and is hard to read. This spacing shows them more like the dataclass does it

rotkehlchen/chain/ethereum/modules/convex/constants.py
```diff
@@ -162,6 +162,40 @@
     string_to_evm_address('0xdEbc7B51043b4D0dd4f3310C68A12382b8843ceE'): 'cvxCrvFrax-f',
     string_to_evm_address('0xb235205E1096E0Ad221Fb7621a2E2cbaB875bE75'): 'pETH-ETH-f',
     string_to_evm_address('0x17962aB30c7F291b125A0A38d0ad220ab22F8a5B'): 'agEURFRAXB-f',
+    string_to_evm_address('0x1A3c8B2F89B1C2593fa46C30ADA0b4E3D0133fF8'): 'CNCETH-f',
```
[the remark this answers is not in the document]
yabirgb: We already query them from a registry. I'm not sure why those aren't in the globaldb cache. @nebolax can maybe answer this

rotkehlchen/chain/ethereum/modules/convex/decoder.py
yabirgb: In the case of the deposit I had in the tests that is for `steCRV` (eth/steth) pool it falls under the else statement. The reason is because there was a migration at some point irc and the old transactions are recorded with the old pools. I guess that we should keep this history (probably by looking at older versions of the registry) and then this would fall under the if statement. There is also CVX deposits in this part of the if/else statement

LefterisJP: Then you should really add this explanation at the else. And yes we could in the future look for older pools or just hard-code them?

rotkehlchen/chain/evm/tokens.py
```diff
@@ -97,6 +97,23 @@ def generate_multicall_chunks(
...
+    """
+    Return the max number of tokens that can be queried in a single call depending wheter we
```
LefterisJP, line 102:
```suggestion
    Return the max number of tokens that can be queried in a single call depending on whether we
```

rotkehlchen/chain/evm/tokens.py
```diff
@@ -97,6 +97,23 @@ def generate_multicall_chunks(
...
+    have a web3 node connected or we are going to use etherscan.
```
LefterisJP, line 103:
```suggestion
    have a web3 node connected or if we are going to use etherscan.
```

LefterisJP (approved):
LGTM. I think the 2-3 small comments can be moved to another PR. Please remember them.

Issue #5819, opened 2023-03-28
Automatically merge vcr cassettes


nebolax:
## Abstract

Currently we tend to often forget to merge vcr cassettes branch when a pr is merged in the main repo. This can be fixed by setting up an automatic cassettes merge.

## Specification

I browsed internet and this seems to be possible to do. We should make a github action that would run on pr merge event, check whether there is a corresponding cassettes branch, and if there is one, merge it.

kelsos:
Yes, this should be doable with a custom action. I have already mentioned this to @LefterisJP before.

Is the only thing required from the custom action to auto-merge?

nebolax:
> Is the only thing required from the custom action to auto-merge?

One awesome feature that I can think of would be to automatically rebase unmerged branches when something is merged to develop/bugfixes. But not sure whether it will be worth doing it.

@yabirgb @prettyirrelevant I don't think we have any other thing that could be needed here, right?

prettyirrelevant:
> One awesome feature that I can think of would be to automatically rebase unmerged branches when something is merged to develop/bugfixes. But not sure whether it will be worth doing it.

Not sure it's worth doing since conflicts might exists in certain cases. 


> I don't think we have any other thing that could be needed here, right?

I think that's sufficient. Keeping it simple ;)

kelsos:
Just for clarification here, the PR branch name and the cassette branch name would match?

nebolax:
Yes, they are always the same

kelsos:
OK I will create the repo and bootstrap the action later today then.

kelsos:
I created the repo [here](https://github.com/rotki/action-cassette-deck) I am gonna setup the initial stuff and then start building the action

LefterisJP:
This is done

PR #5827, opened 2023-03-29
ci: introduces cassette deck workflow

kelsos:
Maybe for #5819

## Checklist

I am not closing the issue until we make sure that this works fine.


LefterisJP (approved):
This looks good to me because I want what it's supposed to do. The autopsy will show as my friend says.

PR #5828, opened 2023-03-29
Utilize curve api and metaregistry for curve cache

### LINKED ISSUE #5809: Add more information for curve and convex pools and gauge addresses
## Abstract

Show names for curve pools, curve gauges and other addresses of protocols where we can give users more information.

In case of curve pools / gauges we should generate the name from symbols of tokens that are used for a pool.

nebolax:
Closes #5809

## Checklist


rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -205,171 +213,158 @@ def save_curve_data_to_cache(
...
-    """
-    registry_pool_count = registry_contract.call(
+def query_curve_data_from_api() -> list[CurvePoolData]:
+    log.debug('Querying curve pools from curve api')
+    all_api_pools = []
+    for api_url in CURVE_API_URLS:
+        try:
+            response = requests.get(api_url, timeout=DEFAULT_TIMEOUT_TUPLE).json()
+        except RequestException as e:
+            raise CurveApiException(f'Could not query curve api due to: {str(e)}') from e
```
prettyirrelevant, line 247: Your usages(here & below) of `CurveApiException` is simply what `RemoteError` is meant for. 
I don't see a reason to create a new exception class for this.

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -205,171 +213,158 @@ def save_curve_data_to_cache(
...
+
+        if response['success'] is False:
```
prettyirrelevant, line 249: Asides this, shouldn't you check the status code? Or their API documentation suggests doing only this check?

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -205,171 +213,158 @@ def save_curve_data_to_cache(
...
+            coins.append(to_checksum_address(coin_data['address']))
+        underlying_coins = None
+        if api_pool_data['isMetaPool'] is False and 'underlyingCoins' in api_pool_data:
+            underlying_coins = []
+            for underlying_coin_data in api_pool_data['underlyingCoins']:
+                underlying_coins.append(to_checksum_address(underlying_coin_data['address']))
+
+        all_processed_pools.append(CurvePoolData(
+            pool_address=to_checksum_address(api_pool_data['address']),
+            pool_name=api_pool_data['name'],
```
prettyirrelevant, line 267: I might be overreaching but I think you should catch `KeyError` error and raise a `RemoteError` saying you got an invalid data or something of such if it ever happens.

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -205,171 +213,158 @@ def save_curve_data_to_cache(
...
-    return pools_to_gauges
+    try:
+        pools_data = query_curve_data_from_api()
+    except CurveApiException as e:
+        log.error(f'Could not query curve api due to: {str(e)}. Will query metaregistry on chain')
+        try:
+            pools_data = query_curve_data_from_chain(ethereum)
+        except RemoteError as err:
+            log.error(f'Could not query chain for curve pools due to: {str(err)}')
+            pools_data = []  # Setting pools_data to an empty list so that it is re-queried later
```
prettyirrelevant, line 368: This could have been initialized at the top. Say before the try/except
minor detail though

rotkehlchen/chain/ethereum/node_inquirer.py
```diff
@@ -264,28 +261,10 @@ def assure_curve_protocol_cache_is_queried(self) -> bool:
...
-            address=registry_address,
-            abi=self.contracts.abi('CURVE_REGISTRY'),
-            deployed_block=0,  # deployment_block is not used and the contract is dynamic
-        )
-        pools = query_curve_registry_pools(ethereum=self, registry_contract=registry_contract)  # noqa: E501
-        pools.update(query_curve_meta_pools(ethereum=self, curve_address_provider=curve_address_provider))  # noqa: E501
-        updated_pools = ensure_curve_tokens_existence(ethereum_inquirer=self, pools_mapping=pools)
-        pools_addresses = [pool_data[0] for pool_data in pools.values()]
-        gauges = query_curve_gauges(ethereum=self, registry_contract=registry_contract, known_pools=pools_addresses)  # noqa: E501
+        all_pools = query_curve_data(ethereum=self)
```
prettyirrelevant, line 264: since you return an empty list in the case of both the API & on-chain calls failing, shouldn't you check for that here?

The docstring above mentions **"returns true if the cache was modified"** but an empty list isn't modifying the cache.

rotkehlchen/tests/unit/test_general_cache.py
```diff
@@ -150,11 +174,11 @@ def mock_call_contract(contract, node_inquirer, method_name, **kwargs):
 
     # Check that the token was created
     token = GlobalDBHandler().get_evm_token(
-        address='0xD71eCFF9342A5Ced620049e616c5035F1dB98620',
```
prettyirrelevant, line 153: any reason why you changed the asset here?  DAI is always present in the DB iirc.

rotkehlchen/tests/unit/test_general_cache.py
```diff
@@ -109,8 +112,29 @@ def mock_call_contract(contract, node_inquirer, method_name, **kwargs):
         new=mock_call_contract,
     )
 
+    def mock_requests_get(url, timeout):  # pylint: disable=unused-argument
+        if url in CURVE_API_URLS:
+            if use_curve_api:
+                if url == CURVE_API_URLS[0]:
```
prettyirrelevant, line 118: you can return early here and above...
```py
if url != CURVE_API_URLS[0]:
   return MockResponse()
```

prettyirrelevant (commented):
left some comments

rotkehlchen/chain/ethereum/node_inquirer.py
nebolax: Makes sense

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -56,8 +63,27 @@
 # Using any random address here, since length of all addresses is the same
 BASE_POOL_TOKENS_KEY_LENGTH = len(compute_cache_key([GeneralCacheType.CURVE_POOL_TOKENS, ZERO_ADDRESS]))  # noqa: E501
 
+# list of pools that we know contain bad tokens
+IGNORED_CURVE_POOLS = {'0x066B6e1E93FA7dcd3F0Eb7f8baC7D5A747CE0BF9'}
```
prettyirrelevant, line 67: I don't see this being used anywhere

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -241,172 +262,196 @@ def save_curve_data_to_cache(
...
+    log.debug('Querying curve pools from curve api')
+    all_api_pools = []
+    for api_url in CURVE_API_URLS:
+        try:
+            raw_response = requests.get(api_url, timeout=DEFAULT_TIMEOUT_TUPLE)
+            if raw_response.status_code != 200:
+                raise RemoteError(f'Curve api returned unexpected status code {raw_response.status_code}')  # noqa: E501
+            response_json = raw_response.json()
+        except RequestException as e:
+            raise RemoteError(f'Could not query curve api due to: {str(e)}') from e
```
prettyirrelevant, line 323: There's a utility function for this `request_get()`. You can use that instead I guess

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -241,172 +262,196 @@ def save_curve_data_to_cache(
...
+
+        if response_json['success'] is False:
+            raise RemoteError('Curve api returned success=False. Bailing.')
+
+        try:
+            all_api_pools.extend(response_json['data']['poolData'])
```
prettyirrelevant, line 329: do you need to do this though? I mean creating a new variable for validation?
You can easily put the for loop in a try/except and return an empty list if keyError is raised.

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -241,172 +262,196 @@ def save_curve_data_to_cache(
...
+    for pool in all_pools:
+        addresbook_entries = [AddressbookEntry(
+            address=pool.pool_address,
+            name=pool.pool_name,
+            blockchain=SupportedBlockchain.ETHEREUM,
+        )]
+        if pool.gauge_address is not None:
+            addresbook_entries.append(AddressbookEntry(
+                address=pool.gauge_address,
+                name=f'Gauge for {pool.pool_name}',
```
prettyirrelevant, line 275: question. what do you think of adding curve prefix here?
minor detail though
```suggestion
                name=f'Curve gauge for {pool.pool_name}',
```

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -241,172 +262,196 @@ def save_curve_data_to_cache(
...
+                blockchain=SupportedBlockchain.ETHEREUM,
+            ))
+        try:
+            db_addressbook.add_addressbook_entries(
+                cursor=write_cursor,
+                book_type=AddressbookType.GLOBAL,
+                entries=addresbook_entries,
+            )
```
prettyirrelevant, line 283: looking at the usage of `add_addressbook_entries()`...

If a known entry is first encountered, it raises an error at that point and does not bother to add other entries irrespective of whether they're known or not.

prettyirrelevant (commented):
Some comments. Also, relevant tests failed.

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
nebolax: Thanks! Fixed.

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
nebolax: Makes sense, thanks!

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
nebolax: Not sure that I got the question. Here I get list of pools, combine all pools from all endpoints in a single list and then iterate them all at once. What do you suggest to change here?

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
nebolax: Yes, probably will make sense to better differentiate with convex staking

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
nebolax: Yes, thanks for pointing it out, will add a parameter to `add_addressbook_entries`

rotkehlchen/tests/data/curve-api-mock.json
```diff
@@ -0,0 +1,160 @@
+{
```
prettyirrelevant, line 1: minor but this can easily be stored as a JSON string assigned to a variable

rotkehlchen/db/addressbook.py
```diff
@@ -112,7 +125,8 @@ def add_addressbook_entries(
                         )
 
                     write_cursor.execute(
-                        'INSERT INTO address_book (address, name, blockchain) VALUES (?, ?, ?)',
+                        f'INSERT {"OR IGNORE " if exists_ok else ""}INTO address_book '
```
prettyirrelevant, line 128: This is ugly imo. Why not just do this outside?
```python
query = ''
if exists_ok is True:
    query += ''
else:
    query += ''

File: rotkehlchen/db/addressbook.py [hunk 15]
```diff
@@ -92,13 +100,18 @@ def add_addressbook_entries(
...
         """
         Add every entry of entries to the address book table.
         If blockchain is None then make sure that the same address doesn't appear in combination
         with other blockchain values.
+
+        If cursor is given then it is used, otherwise a cursor that corresponds to book_type
+        is generated.
         """
-        with self.write_ctx(book_type) as write_cursor:
+        with self.write_ctx(book_type, given_cursor=cursor) as write_cursor:
```
prettyirrelevant, line 114: Is this needed? Or is having nested cursors a bad thing?

prettyirrelevant (commented):
Some comments. Also a relevant test fails.

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -241,172 +257,189 @@ def save_curve_data_to_cache(
...
+            key_parts=[GeneralCacheType.CURVE_POOL_ADDRESS, pool.lp_token_address],
+            values=[pool.pool_address],
         )
-        if pool_address in gauges:
+        for i, coin in enumerate(pool.coins):
             globaldb_set_general_cache_values(
                 write_cursor=write_cursor,
-                key_parts=[GeneralCacheType.CURVE_GAUGE_ADDRESS, pool_address],
-                values=[gauges[pool_address]],
+                key_parts=[GeneralCacheType.CURVE_POOL_TOKENS, pool.pool_address, str(i)],
```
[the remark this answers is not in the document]
nebolax: The order of coins doesn't change. Index is not for uniqueness, but for keeping coins sorted.

rotkehlchen/db/addressbook.py
```diff
@@ -36,7 +36,15 @@ def read_ctx(self, book_type: AddressbookType) -> Generator['DBCursor', None, No
             yield cursor
 
     @contextmanager
-    def write_ctx(self, book_type: AddressbookType) -> Generator['DBCursor', None, None]:
+    def write_ctx(
+            self,
+            book_type: AddressbookType,
+            given_cursor: Optional['DBCursor'] = None,
```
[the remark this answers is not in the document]
nebolax: This is for situations when some cursor already exists and has to be reused. Since a lock is used for cursors https://github.com/rotki/rotki/blob/ee11e2046badefcbc9578c05e641f410dd03d151/rotkehlchen/db/drivers/gevent.py#L358 we can't have 2 cursors instantiated at the same time.

As for why to pass the cursor here -- it looked logical to me to give to `write_ctx` context manager all the data that is needed to determine which cursor to use.

rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -106,6 +107,7 @@ def test_get_addressbook(
     assert AddressbookEntry.deserialize(data=result[0]) == KELSOS_BOOK_ENTRY
 
 
+@pytest.mark.parametrize('custom_globaldb', ['with_empty_addressbook.db'])
```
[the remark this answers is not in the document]
nebolax: `use_clean_caching_directory` has a different effect. With `use_clean_caching_directory` globaldb that is stored in `rotkehlchen/data/global.db` that has addressbook populated will still be used, and we don't need that.

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -241,172 +257,189 @@ def save_curve_data_to_cache(
...
+        )
         globaldb_set_general_cache_values(
             write_cursor=write_cursor,
-            key_parts=[GeneralCacheType.CURVE_POOL_ADDRESS, lp_token_address],
-            values=[pool_address],
+            key_parts=[GeneralCacheType.CURVE_POOL_ADDRESS, pool.lp_token_address],
+            values=[pool.pool_address],
         )
-        if pool_address in gauges:
+        for i, coin in enumerate(pool.coins):
```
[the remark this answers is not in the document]
nebolax: Currently we don't use them anywhere

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -241,172 +257,189 @@ def save_curve_data_to_cache(
...
-        curve_address_provider: EvmContract,
-) -> CURVE_POOLS_MAPPING_TYPE:
-    """Query pools from curve metapool factory and return their mapping
-    May raise:
-    - RemoteError if failed to query etherscan or node
-    - NotERC20Conformant if failed to query info while calling get_or_create_evm_token
-    """
-    factory_address_result = curve_address_provider.call(
+def query_curve_data_from_api() -> list[CurvePoolData]:
+    log.debug('Querying curve pools from curve api')
```
prettyirrelevant, line 304: `request_get()` logs the endpoint called so I'm not sure this is needed.

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -241,172 +257,189 @@ def save_curve_data_to_cache(
...
+    all_api_pools = []
+    for api_url in CURVE_API_URLS:
+        response_json: dict = request_get(api_url)  # type: ignore[assignment]  # response is a dictionary for sure  # noqa: E501
```
prettyirrelevant, line 307:
```suggestion
        response_json = request_get_dict(api_url)
```

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -241,172 +257,189 @@ def save_curve_data_to_cache(
...
+    for api_pool_data in all_api_pools:
+        try:
+            pool_address = to_checksum_address(api_pool_data['address'])
+            if pool_address in IGNORED_CURVE_POOLS:
+                continue
+            coins = []
+            for coin_data in api_pool_data['coins']:
+                coins.append(to_checksum_address(coin_data['address']))
+            underlying_coins = None
+            if api_pool_data['isMetaPool'] is False and 'unde   rlyingCoins' in api_pool_data:
```
prettyirrelevant, line 326:
```suggestion
            if api_pool_data['isMetaPool'] is False and 'underlyingCoins' in api_pool_data:
```

rotkehlchen/db/addressbook.py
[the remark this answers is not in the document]
prettyirrelevant: > we can't have 2 cursors instantiated at the same time.

Ohhh, got it.

prettyirrelevant (commented):
I think you need to rebase your PR on latest develop.
Some comments

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -215,172 +254,190 @@ def save_curve_data_to_cache(
...
-            if checksumed_coin_addr == ZERO_ADDRESS:
-                # curve returns array of fixed length. So if current pool utilizes fewer
-                # coins than the array's length, first values in the result are the real
-                # coins and all addresses after are just 0x00..00
+        pool_name = metaregistry.call(
+            node_inquirer=ethereum,
+            method_name='get_pool_name',
+            arguments=[pool_address],
+        )
+        gauge_address = to_checksum_address(metaregistry.call(
```
LefterisJP, line 378: In all of this function you can't call `to_checksum_address` you need to do deserialize_evm_address and you need to be checking for error

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -215,172 +254,190 @@ def save_curve_data_to_cache(
...
+        for raw_underlying_coin_address in underlying_coins_raw:
+            if raw_underlying_coin_address == ZERO_ADDRESS:
+                break
+            underlying_coins.append(to_checksum_address(raw_underlying_coin_address))
 
-        # for metapools pool addr is the lp token as well
-        pools_mapping[pool_addr] = (pool_addr, pool_coins, None)
+        all_pools.append(CurvePoolData(
+            pool_address=pool_address,
+            pool_name=f'{pool_name} curve pool',
```
LefterisJP, line 411: Do you put the suffix ` name` next to every name you write?

```suggestion
            pool_name=f'{pool_name}',
```

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -215,172 +254,190 @@ def save_curve_data_to_cache(
...
-        if gauge_address != ZERO_ADDRESS:
-            pools_to_gauges[pool_address] = to_checksum_address(gauge_address)
+def query_curve_data(ethereum: 'EthereumInquirer') -> Optional[list[CurvePoolData]]:
+    """Query curve lp tokens, curve pools and curve gauges.
+    First tries to find data via curve api and if fails to do so, queries the chain (metaregistry).
 
-    return pools_to_gauges
+    Returns list of pools if either api or chain query was successful, otherwise None.
+
+    There is a known issue that curve api and metaregistry return different pool names.
```
LefterisJP, line 427: Different how? Add an example here.

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -215,172 +254,190 @@ def save_curve_data_to_cache(
...
+    TODO: think of how to make pool names uniform.
+
+    May raise:
+    - RemoteError if failed to query etherscan or node
+    """
+    pools_data = None
+    try:
+        pools_data = query_curve_data_from_api()
+    except (RemoteError, UnableToDecryptRemoteData) as e:
```
LefterisJP, line 436: as I wrote ... there is more errors that can be raised from there

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -215,172 +254,190 @@ def save_curve_data_to_cache(
...
+        log.error(f'Could not query curve api due to: {str(e)}. Will query metaregistry on chain')
+        try:
+            pools_data = query_curve_data_from_chain(ethereum)
+        except RemoteError as err:
```
LefterisJP, line 440: same here

rotkehlchen/data/eth_abi.json
```diff
@@ -12,5 +12,6 @@
...
...view","type":"function"},{"inputs":[{"internalType":"uint256","name":"tokenId","type":"uint256"}],"name":"tokenURI","outputs":[{"internalType":"string","name":"","type":"string"}],"stateMutability":"view","type":"function"},{"inputs":[],"name":"totalSupply","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address","name":"from","type":"address"},{"internalType":"address","name":"to","type":"address"},{"internalType":"uint256","name":"tokenId","type":"uint256"}],"name":"transferFrom","outputs":[],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"uint256","name":"amount0Owed","type":"uint256"},{"internalType":"uint256","name":"amount1Owed","type":"uint256"},{"internalType":"bytes","name":"data","type":"bytes"}],"name":"uniswapV3MintCallback","outputs":[],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"uint256","name":"amountMinimum","type":"uint256"},{"internalType":"address","name":"recipient","type":"address"}],"name":"unwrapWETH9","outputs":[],"stateMutability":"payable","type":"function"},{"stateMutability":"payable","type":"receive"}]
```
LefterisJP, line 15: What are you doing here? eth_abi json file is no longer utilized. ABI is in the global DB.

rotkehlchen/data/global.db
LefterisJP, line 1: I need you to write what has been added to the global.db. As I said best is to show a comparison with the sqldiff tool

rotkehlchen/db/addressbook.py
```diff
@@ -36,7 +36,15 @@ def read_ctx(self, book_type: AddressbookType) -> Generator['DBCursor', None, No
...
+    ) -> Generator['DBCursor', None, None]:
+        if given_cursor is not None:
```
LefterisJP, line 44: This makes no sense and is a hack. If you need to do that then you are not using the write_ctx correctly somewhere.

As for your comment:

> As for why to pass the cursor here -- it looked logical to me to give to write_ctx context manager all the data that is needed to determine which cursor to use.

Why does it look logical? It's not logical. It's hiding functionality and makes it absurdly easy to make mistakes. You should never have to pass a cursor here. Or in any other of the `write_cursor()` functions. Remove this entire thing.

You are corrupting a function with a very specific purpose by adding some logic which does not fit in there and is just to avoid adding logic in the proper place. The single caller where you need it. And thus introduce hidden fucntionality and if checks to every other caller.

rotkehlchen/db/addressbook.py
```diff
@@ -92,13 +100,18 @@ def add_addressbook_entries(
...
             entries: list[AddressbookEntry],
+            cursor: Optional['DBCursor'] = None,
+            exists_ok: bool = False,
     ) -> None:
         """
         Add every entry of entries to the address book table.
         If blockchain is None then make sure that the same address doesn't appear in combination
         with other blockchain values.
+
+        If cursor is given then it is used, otherwise a cursor that corresponds to book_type
```
LefterisJP, line 111: We **never** do that. Anywhere in the code. At all. Ever. Didn't you wonder why?

Because it does not work. It does not scale. You either pass a cursor always or you never pass a cursor. Delete this optionality hack.

rotkehlchen/db/addressbook.py
```diff
@@ -111,10 +124,11 @@ def add_addressbook_entries(
                             (entry.address,),
                         )
 
-                    write_cursor.execute(
-                        'INSERT INTO address_book (address, name, blockchain) VALUES (?, ?, ?)',
-                        entry.serialize_for_db(),
-                    )
+                    query = 'INSERT '
+                    if exists_ok is True:
```
LefterisJP, line 128: This is yet another hack. Maaaaan wtf?

1. If you wanted to do this move the entire thing out of the for loop and call it `query_prefix`. It's the same for all entries in the loop.
2. I guess this is only needed for you now since the logic you use in curve pools deletes everything and rewrites it all over again. Every. Single. Time. It runs. Doesn't this look like a problem to you? And instead of adding this hack here perhaps find a way for it not to happen instead and not have to go through so much logic for nothing?

rotkehlchen/globaldb/upgrades/v4_v5.py
```diff
@@ -87,6 +87,20 @@ def _reset_curve_cache(write_cursor: 'DBCursor') -> None:
     )
 
 
+def _add_contract_abis(write_cursor: 'DBCursor', root_dir: Path) -> None:
+    """
+    Adds the following contract abis to the globaldb:
+    - Curve Metaregistry
+    """
+    with open(root_dir / 'data' / 'eth_abi.json') as f:
```
LefterisJP, line 95: No. Don't add it there. Just add it to the upgrade directly. That file is no longer used and was used only in an old DB upgrade.

rotkehlchen/tests/api/test_assets.py
```diff
@@ -306,7 +306,7 @@ def test_get_all_assets(rotkehlchen_api_server):
     assert 'entries_total' in result
     assert 'entries_limit' in result
     for entry in result['entries']:
-        assert 'Uniswap' in entry['name']
+        assert 'uniswap' in entry['name'].casefold()
```
LefterisJP, line 309: why did this change? Also we use.lower() in the codebase not casefold for this.

rotkehlchen/tests/data/with_empty_addressbook.db
LefterisJP, line 1: Is this the same as globalDB with the address book empty? If yes then don't do this. Why do you think this is a good idea instead of doing a fixture in the test that for the specific tests just cleans the addressbook?

Do you realize that with what you are doing, every time we add a global DB migration, or god forbid an upgrade we will have to apply it to this file in the tests?

rotkehlchen/tests/unit/globaldb/test_upgrades.py
```diff
@@ -230,7 +230,7 @@ def test_upgrade_v3_v4(globaldb):
         groups = [MAKERDAO_ABI_GROUP_1, MAKERDAO_ABI_GROUP_2, MAKERDAO_ABI_GROUP_3, YEARN_ABI_GROUP_1, YEARN_ABI_GROUP_2, YEARN_ABI_GROUP_3, YEARN_ABI_GROUP_4]  # noqa: E501
         cursor.execute('SELECT COUNT(*) FROM contract_abi')
         assert cursor.fetchone()[0] == (  # len(eth_abi) + contracts_length - uniswap_NFT_MANAGER - 3 optimism contracts that share ABI with mainnet - len(7 abi groups) + 7  # noqa: E501
-            15 +
+            16 +
```
LefterisJP, line 233: nope. This is since you added the abi in the wrong file

rotkehlchen/tests/unit/test_assets.py
```diff
@@ -317,10 +317,9 @@ def test_case_does_not_matter_for_asset_constructor():
     assert a1.identifier == 'BTC'
     assert a2.identifier == 'BTC'
 
-    # We use a token that is only in ethereum (redacted)
-    a3 = symbol_to_ethereum_token('BTRFLY')
-    a4 = symbol_to_ethereum_token('BtRfLy')
-    assert a3.identifier == a4.identifier == strethaddress_to_identifier('0xC0d4Ceb216B3BA9C3701B291766fDCbA977ceC3A')  # noqa: E501
+    a3 = symbol_to_ethereum_token('DAI', chain_id=ChainID.ETHEREUM)
```
LefterisJP, line 320: This looks wrong. `symbol_to_ethereum_token` with a `chain_id` argument? :raised_eyebrow:

rotkehlchen/tests/unit/test_general_cache.py
```diff
@@ -36,29 +48,61 @@
...
+        name='Curve gauge for aave curve pool',
+        blockchain=SupportedBlockchain.ETHEREUM,
+    ),
+]
+
 
 @pytest.mark.parametrize('have_decoders', [True])
-def test_curve_cache(rotkehlchen_instance):
+@pytest.mark.parametrize('use_curve_api', [True, False])
+def test_curve_cache(rotkehlchen_instance, use_curve_api):
```
LefterisJP, line 101: Okay so why do you do all thse magic mocking functions inside this test and keep an entire json file mock that you hand-crafted instead of just using vcr? We have the system in place now. Keep things simple. Don't waste your time and  write tons of mocking code like this ...

LefterisJP (commented):
**A lot of comments**. Not close to ready.

Some generic problems I notice:

- Not good documentation
- Almost complete absence of error checking
- You don't seem to understand, **again**, what logging is for. And what `log.info` means.
- Very poor engineering decisions all over the PR. Like really a lot of them.

Something we already talked about:

The way you use the cache for all these is terribly ugly. This cache was supposed to be a temporary thing and it has grown to a monstrosity. If multiple protocols use pools, which they do, and we end up keeping similar data for all of them, which we will. We should have a generic pool table with all this information.

Deleting and rewriting all pools each time you query pools is just weird.

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -215,172 +254,190 @@ def save_curve_data_to_cache(
...
+def query_curve_data_from_api() -> list[CurvePoolData]:
+    """Query all curve information(lp tokens, pools, gagues, pool coins) from curve api"""
+    all_api_pools = []
+    for api_url in CURVE_API_URLS:
+        response_json: dict = request_get_dict(api_url)
+        if response_json['success'] is False:
+            raise RemoteError('Curve api returned success=False. Bailing.')
+
+        try:
+            all_api_pools.extend(response_json['data']['poolData'])
```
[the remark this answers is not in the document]
nebolax: Yes

rotkehlchen/tests/api/test_assets.py
nebolax: Because some assets were added to the globaldb along with the pre-queried general cache. And some of those assets also match the uniswap query used in this test, however the case of the letters in the names of the new assets is different from what was expected here before.

rotkehlchen/tests/unit/test_assets.py
nebolax: Removed the chain_id parameter for now since it was not used anywhere anyway

rotkehlchen/tests/unit/test_general_cache.py
nebolax: Hand-mocking is needed here to limit the number of pools used. By hand-mocking I limit the number of pools to 2. If to mock only with vcr then all curve pools will be queried and it will be difficult to test it.

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -215,172 +254,190 @@ def save_curve_data_to_cache(
...
+            book_type=AddressbookType.GLOBAL,
+            entries=addresbook_entries,
+            exists_ok=True,
+        )
         globaldb_set_general_cache_values(
             write_cursor=write_cursor,
-            key_parts=[GeneralCacheType.CURVE_POOL_ADDRESS, lp_token_address],
-            values=[pool_address],
+            key_parts=[GeneralCacheType.CURVE_LP_TOKENS],
+            values=[pool.lp_token_address],  # keys of pools_mapping are lp tokens
```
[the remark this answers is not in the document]
nebolax: Yes. The key is just "CURVE_LP_TOKENS" and the value is an address of a single lp token

rotkehlchen/tests/unit/test_general_cache.py
LefterisJP: It's not. You can limit the number of pools to 2 with VCR and a patch to the list of pools to query. It's really simple actually.

rotkehlchen/data/global.db
nebolax: Full diff: https://gist.github.com/nebolax/cd91ee6a4b1bde9602e117d0b9fd5e2e

The changes are:
- Populate addressbook for curve pools
- More pools were detected, so add new lp tokens, pool addresses, etc.

rotkehlchen/tests/unit/test_general_cache.py
nebolax: It is easy to patch method queries to the chain this way, but I don't see how this can simplify mocking curve api since curve api endpoints just return a huge list of pools.

LefterisJP: So you get a list of pools like this: `pools`. And then you want only 2 of those? Can't you do `pools[:2]`?

LefterisJP: I mean to find a way to write this in a simple way using VCR and minimal mocking. I am not going to implement it for you I am working on something else. My answer is direct to your problem of "returning a huge list of pools while you need 2"

rotkehlchen/db/addressbook.py
```diff
@@ -90,36 +90,42 @@ def get_addressbook_entries(
...
-                    # in the case of given blockchain being None delete any other entry for that
-                    # address since they are rendered redundant
-                    if entry.blockchain is None:
-                        write_cursor.execute(
-                            'DELETE FROM address_book where address=? AND blockchain IS NOT NULL',
-                            (entry.address,),
-                        )
 
+        If cursor is given then it is used, otherwise a cursor that corresponds to book_type
+        is generated.
```
LefterisJP, line 103: You resolved my comment, made the cursor a permanent argument but the place where my comment was did you read it? There is a reason I left it on this line.


```suggestion
```

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -171,216 +216,236 @@ def ensure_curve_tokens_existence(
...
+            addresbook_entries.append(AddressbookEntry(
+                address=pool.gauge_address,
+                name=f'Curve gauge for {pool.pool_name}',
+                blockchain=SupportedBlockchain.ETHEREUM,
+            ))
+        db_addressbook.add_addressbook_entries(
             write_cursor=write_cursor,
-            key_parts=[GeneralCacheType.CURVE_POOL_TOKENS, pool_addr],
+            entries=addresbook_entries,
+            exists_ok=True,
```
LefterisJP, line 249: Is this really needed now that you do the new pools?

rotkehlchen/db/addressbook.py
```diff
@@ -90,36 +90,42 @@ def get_addressbook_entries(
 
     def add_addressbook_entries(
             self,
-            book_type: AddressbookType,
+            write_cursor: 'DBCursor',
             entries: list[AddressbookEntry],
+            exists_ok: bool = False,
```
LefterisJP, line 95: I [told you](https://github.com/rotki/rotki/pull/5828#discussion_r1156376139) to remove this argument in the previous review unless it's absolutely needed. And I don't think it is from the way you use it now. You only should call this function for new pools.

rotkehlchen/globaldb/upgrades/v4_v5.py
```diff
@@ -62,29 +54,8 @@ def _populate_rpc_nodes(cursor: 'DBCursor', root_dir: Path) -> None:
...
-        write_cursor=write_cursor,
-        key_parts=[GeneralCacheType.CURVE_LP_TOKENS],
-        values=lp_tokens,
-        timestamp=refreshing_timestamp,
-    )
-    globaldb_delete_general_cache_like(
-        write_cursor=write_cursor,
-        key_parts=[GeneralCacheType.CURVE_POOL_TOKENS],
-    )
+    """Resets curve cache to query gauges and update format of the lp tokens"""
```
LefterisJP, line 57: Since you are touching this file I notice it has multiple things missing.

1. The `migrate_to_v5` should have at its docstring the version where it's introduced. For v4->v5 this is 1.28.
2. At the end of `migrate_to_v5` the `Exit` debug log is missing

rotkehlchen/tests/unit/test_general_cache.py
```diff
@@ -110,8 +147,31 @@ def mock_call_contract(contract, node_inquirer, method_name, **kwargs):
         new=mock_call_contract,
     )
 
+    requests_get = requests.get
+
+    def mock_requests_get(url, timeout):  # pylint: disable=unused-argument
```
LefterisJP, line 152: you need a docstring here to explain what you mock and why, especially considering this is a vcred test

rotkehlchen/tests/unit/test_general_cache.py
```diff
@@ -110,8 +147,31 @@ def mock_call_contract(contract, node_inquirer, method_name, **kwargs):
...
+        response_json = requests_get(url).json()
+        response_json['data']['poolData'] = response_json['data']['poolData'][:2]
+        return MockResponse(status_code=200, text=json.dumps(response_json))
+
+    requests_patch = patch(
+        'requests.get',
+        side_effect=mock_requests_get,
+    )
+
+    constant_assets_patch = patch('rotkehlchen.constants.assets.CONSTANT_ASSETS', [])
```
LefterisJP, line 171: Why do you need this patch? Can you add a small comment?

LefterisJP (commented):
Some more work needed

rotkehlchen/chain/ethereum/modules/curve/curve_cache.py
```diff
@@ -171,216 +216,236 @@ def ensure_curve_tokens_existence(
...
+    Query all curve information(lp tokens, pools, gagues, pool coins) from metaregistry.
+
     May raise:
-    - RemoteError if failed to query etherscan or node
-    - NotERC20Conformant if failed to query info while calling get_or_create_evm_token
+    - RemoteError if failed to query chain
     """
-    factory_address_result = curve_address_provider.call(
+    address_provider = ethereum.contracts.contract('CURVE_ADDRESS_PROVIDER')
+    metaregistry_address = deserialize_evm_address(address_provider.call(
```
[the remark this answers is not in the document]
nebolax: Yes, forgot here, thx

rotkehlchen/tests/unit/test_general_cache.py
```diff
@@ -110,8 +147,42 @@ def mock_call_contract(contract, node_inquirer, method_name, **kwargs):
...
+
+    requests_patch = patch(
+        'requests.get',
+        side_effect=mock_requests_get,
+    )
+
+    # Since DAI (which is a constant asset) was deleted at the beginning of this test and DAI is
+    # used in one of the tested curve pools, we disable constant assets here (otherwise the code
+    # will attempt to load DAI from packaged db, won't find it there and will raise an error).
+    constant_assets_patch = patch('rotkehlchen.constants.assets.CONSTANT_ASSETS', [])
```
LefterisJP, line 182: This seems like a hack complicating the test. Please if you want to delete an asset delete any non-constant asset.

LefterisJP (approved):
LGTM

PR #5834, opened 2023-03-30
ci: updates permissions for merge action

kelsos:
Closes #(issue_number)

## Checklist

### DIFF
--- a/.github/workflows/rotki_cassette_merge.yml
+++ b/.github/workflows/rotki_cassette_merge.yml
@@ -8,6 +8,9 @@ on:
 jobs:
   merge:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      content: write
     steps:
       - uses: rotki/action-cassette-deck@v1
         with:


.github/workflows/rotki_cassette_merge.yml
```diff
@@ -8,6 +8,9 @@ on:
 jobs:
```
nebolax, line 8: Small question. It currently has

```
on:
  pull_request:
    types: [ closed ]
```

Does it mean that the cassettes branch will be merged even if PR in the main repo is closed without merging?

kelsos: I will start with a counter-question. Why would we merge a PR in the second repo if the PR in the main repo was closed, without being merged?

There is no `pull_request` merged event. The `merged` status is just a property in a closed pull request. So in both cases of closing and merging, there is a single event fired which is `closed`. 

This only means that the action will trigger on every closed pull request regardless of the way it was closed since the trigger is the same in both cases.

So to answer your question no it does not mean that the branch in the secondary repo will be merged in a closed PR.

The action checks the event context and verifies that the PR was actually merged and if not it just completes.

https://github.com/rotki/action-cassette-deck/blob/934990a444efd50ef35db4479c8dab0e0aa95b30/src/main.ts#L103

kelsos:
https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#contents

https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#pull-requests

PR #5835, opened 2023-03-30
Add mappings of possible types and subtypes in evm events

yabirgb:
This PR adds a mapping of counterparties to their possible combinations of event type and subtype.


docs/api.rst
```diff
@@ -12475,3 +12475,111 @@ Clear Icons/Avatars Cache
    :statuscode 200: The cache items were cleared successfully.
    :statuscode 400: Entries provided do not match the ``cache_type`` specified. Invalid ``cache_type`` provided.
    :statuscode 500: Internal rotki error
+
+
+Event Mappings
+================
+
+.. http:post:: /api/(version)/history/events/type_mappings
```
nebolax, line 12483:
```suggestion
.. http:get:: /api/(version)/history/events/type_mappings
```

docs/api.rst
```diff
@@ -12475,3 +12475,111 @@ Clear Icons/Avatars Cache
...
+
+  Doing a GET on this endpoint will return a mapping of history events types and subtypes to the frontend's ``TransactionEventType`` representation.
```
nebolax, line 12485: I would also mention here `event_type_details` part

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -794,3 +798,22 @@ def _is_non_conformant_erc721(address: ChecksumEvmAddress) -> bool:
     def _address_is_exchange(address: ChecksumEvmAddress) -> Optional[str]:
         """Takes an address and returns if it's an exchange in the given chain
         and the counterparty to use if it is."""
+
+    def _check_correct_types(self, decoded_events: list['EvmEvent']) -> bool:
```
nebolax, line 802: Maybe put the entire method definition under `__debug__` ?

So something like
```python
if __debug__:
    def _check_correct_types(self, decoded_events: list['EvmEvent']) -> bool:
        unexpected_types = set()
        for event in decoded_events:
            if event.counterparty is None:
                continue
```

Also afaik `if __debug__` should be used without `is ...` in order for the code to be removed in production (or at least we use it like this everywhere).

nebolax (commented):
Left a few comments, but also have a general question. Currently frontend is supposed to get valid pairs event type-subtype by looking at `FRONTEND_MAPPING`, right?

So a suggestion here. Maybe it will be better to give to the frontend `events_types_tuples` for this purpose so that they can also take counterparty into account and thus make the validation more accurate?

rotkehlchen/chain/evm/decoding/decoder.py
yabirgb: I didn't know the debug rule thing, thanks! The function call should be removed since it has an assert in front and loading the function in the class I believe is not such a big deal. Having the `if __debug__ return` thing is just to avoid calling it without noticing in prod

yabirgb:
@nebolax this is just a first step that tries to achieve two things:

1. Collect all the types used in the decoders so we can then make them consistent across modules
2. Give the frontend the mappings to avoid duplicated code between frontend and backend.

About your suggestion I believe we have an issue for that that is providing the info in the new endpoint + the counterparty and returning it but I would do as a separte PR as this is big already

nebolax:
Alright then

docs/api.rst
```diff
@@ -12475,3 +12475,111 @@ Clear Icons/Avatars Cache
...
+                    "payback debt": "repay",
+                    "return wrapped": "send",
+                    "donate": "donate",
+                    "liquidate": "liquidate",
+                    "none": "send"
+                }
+            },
+            "event_type_details":{
+              "gas":{
+                  "label": "gas_fee",
```
LefterisJP, line 12516: Why would you have `_` in a human readable label?

docs/api.rst
```diff
@@ -12475,3 +12475,111 @@ Clear Icons/Avatars Cache
...
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+        "result":{
+            "mappings":{
+                "spend":{
+                    "fee": "gas",
+                    "payback debt": "repay",
```
LefterisJP, line 12507: Is there any central place where we have all "Transaction" types? The value part of this mapping? So "donate", "liquidate" etc.?

docs/api.rst
```diff
@@ -12475,3 +12475,111 @@ Clear Icons/Avatars Cache
...
+  :resjson string icon: Icon to be used by the frontend for this event type.
+  :resjson string color[optional]: Optional color to apply to the icon
+
+  :statuscode 200: Information was correctly generated
+  :statuscode 500: Internal rotki error
+
+Counterparties
+================
+
+.. http:post:: /api/(version)/history/events/counterparties
```
LefterisJP, line 12546: I assume this is copy pasta

```suggestion
.. http:get:: /api/(version)/history/events/counterparties
```

rotkehlchen/chain/evm/frontend_structures/events_mappings.py
```diff
@@ -0,0 +1,170 @@
+from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.chain.evm.frontend_structures.types import EventDetails, TransactionEventType
```
LefterisJP, line 2: This approach is wrong.

Each decoder should add to these mappings on its own and those should be added dynamically as decoders are created. Much like counterparties and accounting rules are created.

We are creating a modular system not a hard-coded mapping ...

LefterisJP (commented):
This approach is wrong.

Each decoder should add to these mappings on its own and those should be added dynamically as decoders are created. Much like counterparties and accounting rules are created.

We are creating a modular system not a hard-coded mapping ...

rotkehlchen/chain/evm/frontend_structures/protocols.py
```diff
@@ -0,0 +1,224 @@
+from rotkehlchen.chain.ethereum.decoding.constants import CPT_GNOSIS_CHAIN
```
LefterisJP, line 1: As I said elsewhere this should move to the decoders themselves. So this is againt not a frontend specific thing and should cease to exist as a file.

rotkehlchen/chain/evm/frontend_structures/types.py
```diff
@@ -0,0 +1,46 @@
+from enum import auto
```
LefterisJP, line 1: As mentioned before. Concept of `frontend_structures` is mistaken. This file should not be under this directory.

rotkehlchen/chain/evm/frontend_structures/types.py
```diff
@@ -0,0 +1,46 @@
...
+    image: str
+
+
+class EventDetails(NamedTuple):
+    label: str
+    icon: str
+    color: Optional[Literal['green', 'red']] = None
+
+
+class TransactionEventType(SerializableEnumMixin):
```
LefterisJP, line 19: And tell frontend to also change their shit. But also why duplicate? Can't we just export this enum to them? They should not duplicate this data.

```suggestion
class EventCategory(SerializableEnumMixin):
```

rotkehlchen/chain/evm/frontend_structures/types.py
```diff
@@ -0,0 +1,46 @@
+from enum import auto
+from typing import Literal, NamedTuple, Optional
+
+from rotkehlchen.utils.mixins.serializableenum import SerializableEnumMixin
+
+
+class ProtocolDetails(NamedTuple):
```
LefterisJP, line 7: This should be CoutnerpartyDetails, yes we had this conversation before. Kraken is also a counterparty and it's not a protocol. And it should be populated from the decoders themselves. Or ... well returned by the decoders which use them. Not all be in one big constant like you have it now.

rotkehlchen/serialization/serialize.py
```diff
@@ -153,6 +159,7 @@ def _process_entry(entry: Any) -> Union[str, list[Any], dict[str, Any], Any]:
             WeightedNode,
     )):
         return process_result(entry.serialize())
+    from rotkehlchen.chain.evm.frontend_structures.protocols import ProtocolDetails  # isort:skip  # noqa: E501  # pylint: disable=import-outside-toplevel
```
LefterisJP, line 162: wtf is this import doing here?!

rotkehlchen/tests/api/test_misc.py
```diff
@@ -518,3 +518,26 @@ def test_cache_deletion(rotkehlchen_api_server):
...
+    Test that the information for event types mappings and counterparties is correctly
+    generated for the frontend.
+    """
+    response = requests.get(
+        api_url_for(
+            rotkehlchen_api_server,
+            'typesmappingsresource',
+        ),
+    )
+    assert_proper_response(response)
```
LefterisJP, line 535: you are testing nothing here. Please test at least something or delete the test. This is only testing both endpoints exist.

docs/api.rst
```diff
@@ -12475,3 +12475,127 @@ Clear Icons/Avatars Cache
...
+          "global_mappings":{
+            "spend":{
+              "fee":"gas",
+              "return wrapped":"send",
+              "donate":"donate",
+              "none":"send"
+            }
+          },
+          "per_protocol_mappings":{
+            "ethereum":{
```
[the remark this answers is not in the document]
yabirgb: Also because the same counterparty could appear in different chines with different types

LefterisJP (commented):
Okay some more comments. Mostly two things I guess:

1. CounterpartyDetails can be declared as constants more times than you do right now
2. Some things need to be renamed for consistency and remove outdated comments about duplication

rotkehlchen/chain/ethereum/modules/airdrops/decoder.py
```diff
@@ -260,5 +270,41 @@ def addresses_to_decoders(self) -> dict[ChecksumEvmAddress, tuple[Any, ...]]:
             ELFI_LOCKING: (self._decode_elfi_claim,),
         }
 
-    def counterparties(self) -> list[str]:
-        return ETHEREUM_AIRDROPS_LIST
+    def counterparties(self) -> list[CounterpartyDetails]:
+        return [
+            CounterpartyDetails(
```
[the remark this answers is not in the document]
yabirgb: For the uniswap case we have each version and the airdrop has its own too since it doesn't belong to any specific version. That is why it is not reused

LefterisJP (approved):
LGTM. We probably need to re-think the counterparty details for very similar entries.

Also @rotki/frontend need to adjust the frontend after this.

PR #5839, opened 2023-03-31
Fix curve inquirer

nebolax:
The issue that this PR fixes was that method in our Inquirer that queries price for a curve lp token used old curve pool tokens schema (so unindexed/unordered tokens). This was not working.

Also to solve circular import that popped up after the fix I moved `read_curve_data` function to globaldb/cache.py


yabirgb (approved):

yabirgb:
We didn't see this error before because is the combination of two PRs. The one that changes the curve cache and the one that deleted the old entries from the globaldb

PR #5842, opened 2023-03-31
reset last_avatar_update column when avatars cache is cleared

prettyirrelevant:
## Description
- Whenever the cache of an avatar is cleared, `last_avatar_update` for the particular ENS name is set to `Timestamp(0)`.
- If the entire cache is cleared, `last_avatar_update` is set to `Timestamp(0)` for all ENS names present in the DB.

- `last_avatar_update` is now non-nullable and a default of 0 is set.
- A follow up from a previous PR is also included.


rotkehlchen/api/rest.py
```diff
@@ -4253,11 +4253,20 @@ def clear_avatars_cache(self, avatars: Optional[list[str]]) -> Response:
                 if entry.is_file():
                     entry.unlink()
 
+            with self.rotkehlchen.data.db.user_write() as delete_cursor:
+                delete_cursor.execute('UPDATE ens_mappings SET last_avatar_update=?', (Timestamp(0),))  # noqa: E501
```
LefterisJP, line 4257: Isn't it better to set it to null? What was the thinking of having it as null before @nebolax ?

rotkehlchen/tests/api/test_misc.py
```diff
@@ -418,6 +420,20 @@ def test_cache_deletion(rotkehlchen_api_server):
     avatars_dir = icons_dir / 'avatars'
     avatars_dir.mkdir(exist_ok=True)
 
+    dbens = DBEns(rotkehlchen_api_server.rest_api.rotkehlchen.data.db)
+    with dbens.db.user_write() as write_cursor:
+        dbens.add_ens_mapping(
```
LefterisJP, line 425: why is this needed in the test?

prettyirrelevant: Querying an avatar that's not present in the ens_mappings will raise an error.

rotkehlchen/api/rest.py
nebolax: I used null to be more explicit

prettyirrelevant: Alright then. Will set it to NULL

rotkehlchen/tests/api/test_misc.py
LefterisJP: then you should have added a comment explaining this is why you are doing it.

LefterisJP (approved):
LGTM but not sure why the 2 tests failed.

LefterisJP:
merging as discussed the tests are fixed elsewhere

PR #5844, opened 2023-03-31
Refactor pagination filters for nfts balances

tewshi:
## Checklist

- [x] extends the pagination/filter composable to work with components without filters
- [x] The PR modified the frontend, and updated the nfts balances component to use pagination/filter composable
- [x] adds unit test for the composable's use case with nfts
- [x] fixes reactivity of accounts on route change in evm transactions


kelsos (commented):
So I left some initial comments, I am gonna checkout and test too and I will add if there is more

frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
```diff
@@ -179,125 +173,51 @@ const massIgnore = async (ignored: boolean) => {
...
+  isLoading,
+  fetchData,
+  options,
+  setPage,
+  setOptions
+} = useHistoryPaginationFilter<
+  NonFungibleBalance,
+  NonFungibleBalancesRequestPayload,
+  NonFungibleBalance,
+  undefined,
```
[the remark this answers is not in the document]
tewshi: I just found out I could set default for generics... fantastic.

frontend/app/src/composables/history/filter-paginate.ts
```diff
@@ -14,33 +16,41 @@ import { nonEmptyProperties } from '@/utils/data';
...
   locationOverview: MaybeRef<string | null>,
-  mainPage: Ref<boolean>,
+  mainPage: MaybeRef<boolean>,
   filterSchema: () => FilterSchema<W, X>,
   fetchAssetData: (payload: MaybeRef<U>) => Promise<Collection<V>>,
   options: {
     onUpdateFilters?: (query: LocationQuery) => void;
     extraParams?: ComputedRef<LocationQuery>;
     customPageParams?: ComputedRef<Partial<U>>;
+    paginationOrderBy?: keyof T;
```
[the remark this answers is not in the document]
tewshi: sorry, I refactored it into the object below, and forgot to remove this, removing it now.

frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
```diff
@@ -179,125 +173,51 @@ const massIgnore = async (ignored: boolean) => {
...
+  undefined
+>(
+  null,
+  true,
+  () => ({
+    filters: ref(),
+    matchers: undefined,
+    RouteFilterSchema: undefined,
+    updateFilter: undefined
```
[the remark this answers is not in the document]
tewshi: looking into it again, I did try to avoid these values on Fri...

frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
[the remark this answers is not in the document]
kelsos: Yes it's nice

frontend/app/src/composables/history/filter-paginate.ts
```diff
@@ -166,9 +190,12 @@ export const useHistoryPaginationFilter = <
...
       // Route should only be updated on user action otherwise it messes with
       // forward navigation.
-      await router.push({
-        query: getQuery()
-      });
+      const query = getQuery();
+      if (isEqual(route.query, query)) {
+        // prevent pushing same route
+        return;
+      }
```
kelsos, line 197: in which case does this happen?

frontend/app/src/composables/history/filter-paginate.ts
```diff
@@ -16,7 +18,7 @@ interface FilterSchema<F, M> {
   filters: Ref<F>;
   matchers: ComputedRef<M[]>;
 
-  updateFilter(filter: F): void;
+  updateFilter: (filter: F) => void;
 
   RouteFilterSchema: ZodSchema;
```
kelsos, line 23: Hmm here and in the empty one this is not marked as optional but when called in the composable it is.

frontend/app/src/composables/history/filter-paginate.ts
tewshi: while testing, I noticed sometimes, vue drops warning about pushing to the same route, that was why I added this extra check, but I can remove it.

kelsos: not related to hot reloading thought right?

kelsos: If this covers an actual production case then it's ok

tewshi: Ok

tewshi: on a second thought, maybe it is on hot reload...
so, I'll remove the check again, try to make a change, save and see if it happens on hot reload.

frontend/app/src/composables/history/filter-paginate.ts
tewshi: `RouteFilterSchema` right? 

yeah, I'll remove the null check since in every instance it will always be there...
I initially made them optional.

frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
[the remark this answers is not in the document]
tewshi: done

frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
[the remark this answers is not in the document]
tewshi: done, made an empty filter that returns defaults for these fields.

frontend/app/src/composables/history/filter-paginate.ts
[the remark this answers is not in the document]
tewshi: done

frontend/app/src/composables/history/filter-paginate.ts
tewshi: done

kelsos:
Question @tewshi can you rebase on develop so that it picks the fix for running cypress on chrome. Code looks ok to me so we need a green PR and it will be good to merge.

kelsos:
So the cause for the failed tests is the following

image

The fix should be simple. We just have to add these two in the optimize deps: 

https://github.com/rotki/rotki/blob/ed1a8bf659d969adec72bcec2ae626c57867e19f/frontend/app/vite.config.ts#L47

If we don't vite will hot-reload and cause a test to fail

kelsos (approved):
lgtm @tewshi thanks

PR #5845, opened 2023-03-31
Remotely update contracts data in globaldb

### LINKED ISSUE #5594: Remotely update EVM contracts information in the globaldb
## Abstract

Allow to update the set of contracts abi's as we do for updating the list of assets

## Motivation

At the moment if we want to add new contracts to the user's globaldb we have to make a globaldb migration and add them. We can avoid migrations by using an update sysmtem

Related issues are

- #5532
- #5236

nebolax:
Closes #5594

## Checklist

### DIFF
--- a/rotkehlchen/db/updates.py
+++ b/rotkehlchen/db/updates.py
@@ -2,7 +2,7 @@ import json
 import logging
 import sys
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Final
+from typing import TYPE_CHECKING, Any, Final, Literal, Union, overload
 
 import requests
 
@@ -26,6 +26,7 @@ log = RotkehlchenLogsAdapter(logger)
 LAST_DATA_UPDATES_KEY: Final = 'last_data_updates_ts'
 SPAM_ASSETS_URL = 'https://raw.githubusercontent.com/rotki/data/{branch}/updates/spam_assets/v{version}.json'  # noqa: E501
 RPC_NODES_URL = 'https://raw.githubusercontent.com/rotki/data/{branch}/updates/rpc_nodes/v{version}.json'  # noqa: E501
+CONTRACTS_URL = 'https://raw.githubusercontent.com/rotki/data/{branch}/updates/contracts/v{version}.json'  # noqa: E501
 
 
 class UpdateType(Enum):
@@ -83,13 +84,13 @@ class RotkiDataUpdater:
 
         return json_data
 
-    def update_spam_assets(self, local_version: int, to_version: int) -> None:
+    def update_spam_assets(self, from_version: int, to_version: int) -> None:
         """
         Check the updates in the inclusive range [local_version + 1, to_version] and update the
         spam assets for those versions. Assets are also added to the globaldb if they don't exist
         locally
         """
-        for version in range(local_version + 1, to_version + 1):
+        for version in range(from_version + 1, to_version + 1):
             success, updated_assets = self._get_file_content(
                 file_url=SPAM_ASSETS_URL.format(branch=self.branch, version=version),
                 update_type=UpdateType.SPAM_ASSETS,
@@ -109,7 +110,7 @@ class RotkiDataUpdater:
                     (UpdateType.SPAM_ASSETS.serialize(), version),
                 )
 
-    def update_rpc_nodes(self, to_version: int) -> None:
+    def update_rpc_nodes(self, from_version: int, to_version: int) -> None:  # pylint: disable=unused-argument  # from_version is here to have the same format as other updating functions  # noqa: E501
         """Checks the update for the latest version and update the default rpc nodes in the global db.
 
         It also updates the user db with these default nodes.
@@ -154,29 +155,89 @@ class RotkiDataUpdater:
             new_default_nodes=new_default_nodes,
         )
 
+    def update_contracts(self, from_version: int, to_version: int) -> None:
+        """
+        Updates evm contracts and ABIs in globaldb.
+        If an ABI already exists in the db, it is reused.
+        If a contract (address + chain id) exists, it is replaced by the remote counterpart.
+        """
+        for version in range(from_version + 1, to_version + 1):
+            success, updated_contracts = self._get_file_content(
+                file_url=CONTRACTS_URL.format(branch=self.branch, version=version),
+                update_type=UpdateType.CONTRACTS,
+            )
+            if success is False:
+                return
+
+            log.info(f'Applying update for contracts to v{version}')
+            remote_id_to_local_id = {}
+            with GlobalDBHandler().conn.read_ctx() as cursor:
+                for single_abi_data in updated_contracts['abis_data']:
+                    serialized_abi = json.dumps(single_abi_data['value'], separators=(',', ':'))
+
+                    # check if the abi is already present in the database
+                    existing_abi_id = cursor.execute(
+                        'SELECT id FROM contract_abi WHERE value=?',
+                        (serialized_abi,),
+                    ).fetchone()
+
+                    if existing_abi_id is not None:
+                        abi_id = existing_abi_id[0]
+                    else:
+                        with GlobalDBHandler().conn.write_ctx() as write_cursor:
+                            write_cursor.execute(
+                                'INSERT INTO contract_abi(name, value) VALUES(?, ?)',
+                                (single_abi_data['name'], serialized_abi),
+                            )
+                            abi_id = write_cursor.lastrowid
+                    # store a mapping of the virtual id to the id in the user's globaldb
+                    remote_id_to_local_id[single_abi_data['id']] = abi_id
+
+            new_contracts_data = []
+            for single_contract_data in updated_contracts['contracts_data']:
+                if single_contract_data['abi'] not in remote_id_to_local_id:
+                    self.msg_aggregator.add_error(
+                        f'ABI with id {single_contract_data["abi"]} was missing in a contracts '
+                        f'update. Please report it to the rotki team.',
+                    )
+                    continue
+                new_contracts_data.append((
+                    single_contract_data['address'],
+                    single_contract_data['chain_id'],
+                    single_contract_data['name'],
+                    remote_id_to_local_id[single_contract_data['abi']],
+                    single_contract_data['deployed_block'],
+                ))
+
+            with GlobalDBHandler().conn.write_ctx() as write_cursor:
+                write_cursor.executemany(
+                    'INSERT OR REPLACE INTO contract_data(address, chain_id, name, abi, deployed_block) VALUES(?, ?, ?, ?, ?)',  # noqa: E501
+                    new_contracts_data,
+                )
+
+            with self.user_db.conn.write_ctx() as write_cursor:
+                write_cursor.execute(
+                    'INSERT OR REPLACE INTO settings(name, value) VALUES (?, ?)',
+                    (UpdateType.CONTRACTS.serialize(), version),
+                )
+
     def check_for_updates(self) -> None:
         """Retrieve the information about the latest available update"""
         remote_information = self._get_remote_info_json()
 
-        # Get latest applied versions
-        with self.user_db.conn.read_ctx() as cursor:
-            local_spam_assets_version = self._check_for_last_version(
-                cursor=cursor,
-                update_type=UpdateType.SPAM_ASSETS,
-            )
-            local_rpc_nodes_version = self._check_for_last_version(
-                cursor=cursor,
-                update_type=UpdateType.RPC_NODES,
-            )
+        for update_type in UpdateType:
+            # Get latest applied version
+            with self.user_db.conn.read_ctx() as cursor:
+                local_version = self._check_for_last_version(
+                    cursor=cursor,
+                    update_type=update_type,
+                )
 
-        # Update all remote data
-        latest_spam_assets_version = remote_information[UpdateType.SPAM_ASSETS.value]['latest']
-        if local_spam_assets_version < latest_spam_assets_version:
-            self.update_spam_assets(local_spam_assets_version, latest_spam_assets_version)
-
-        latest_rpc_nodes_version = remote_information[UpdateType.RPC_NODES.value]['latest']
-        if local_rpc_nodes_version < latest_rpc_nodes_version:
-            self.update_rpc_nodes(latest_rpc_nodes_version)
+            # Update all remote data
+            latest_version = remote_information[update_type.value]['latest']
+            if local_version < latest_version:
+                update_function = getattr(self, f'update_{update_type.value}')
+                update_function(from_version=local_version, to_version=latest_version)
 
         with self.user_db.user_write() as cursor:
             cursor.execute(  # remember last time data updates were detected
@@ -194,10 +255,26 @@ class RotkiDataUpdater:
         return int(found_version[0]) if found_version is not None else 0
 
     @staticmethod
+    @overload
+    def _get_file_content(
+            file_url: str,
+            update_type: Literal[UpdateType.SPAM_ASSETS, UpdateType.RPC_NODES],
+    ) -> tuple[bool, list[dict[str, Any]]]:
+        ...
+
+    @staticmethod
+    @overload
+    def _get_file_content(
+            file_url: str,
+            update_type: Literal[UpdateType.CONTRACTS],
+    ) -> tuple[bool, dict[str, Any]]:
+        ...
+
+    @staticmethod
     def _get_file_content(
             file_url: str,
             update_type: UpdateType,
-    ) -> tuple[bool, list[dict[str, Any]]]:
+    ) -> tuple[bool, Union[list[dict[str, Any]], dict[str, Any]]]:
         """Retrieves the content of data to be updated.
 
         Returns True/False for success along with the content.
--- a/rotkehlchen/tests/unit/test_data_updates.py
+++ b/rotkehlchen/tests/unit/test_data_updates.py
@@ -16,59 +16,85 @@ ETHEREUM_SPAM_ASSET_ADDRESS = make_evm_address()
 OPTIMISM_SPAM_ASSET_ADDRESS = make_evm_address()
 
 
-def mock_github_data_response(url, timeout):  # pylint: disable=unused-argument
-    if 'info' in url:
-        return MockResponse(200, '{"spam_assets":{"latest":1},"rpc_nodes":{"latest":1}}')
-    if 'spam_assets/v' in url:
-        data = {
-            'spam_assets': [
-                {
-                    'address': ETHEREUM_SPAM_ASSET_ADDRESS,
-                    'name': '$ ClaimUniLP.com',
-                    'symbol': '$ ClaimUniLP.com - Visit to claim',
-                    'decimals': 18,
+ABI_DATA = [
+    {'id': 1, 'name': 'ABI1', 'value': '[{"inputs":[{"internalType":"address","name":"_factory","type":"address"}]'},  # noqa: E501
+    {'id': 2, 'name': 'ABI2', 'value': '[{"name":"BasePoolAdded","inputs":[{"name":"base_pool","type":"address","indexed":false}],"anonymous":false,"type":"event"}]'},  # noqa: E501
+]
+CONTRACT_DATA = [
+    {'address': '0x4bBa290826C253BD854121346c370a9886d1bC26', 'chain_id': 1, 'name': 'CONTRACT1_ETH', 'abi': 1, 'deployed_block': 1000},  # noqa: E501
+    {'address': '0xc37b40ABdB939635068d3c5f13E7faF686F03B65', 'chain_id': 10, 'name': 'CONTRACT1_OP', 'abi': 1, 'deployed_block': 20},  # noqa: E501
+    {'address': '0x19e4057A38a730be37c4DA690b103267AAE1d75d', 'chain_id': 1, 'name': None, 'abi': 2, 'deployed_block': None},  # noqa: E501
+]
+
+
+def make_mock_github_data_response(target: UpdateType):
+    """Creates a mocking function for github requests."""
+    def mock_github_data_response(url, timeout):  # pylint: disable=unused-argument
+        if 'info' in url:
+            data = {
+                'spam_assets': {'latest': 1 if target == UpdateType.SPAM_ASSETS else 0},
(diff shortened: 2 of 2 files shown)


rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -172,3 +187,37 @@ def test_update_rpc_nodes(data_updater: RotkiDataUpdater) -> None:
...
+            contract_data.get('name'),
+            contract_data['abi'],
+            contract_data.get('deployed_block'),
+        ))
+
+    with GlobalDBHandler().conn.read_ctx() as cursor:
+        # Assert state after the update
+        abis_data_in_db = cursor.execute('SELECT id, value, name FROM contract_abi')
+        assert set(abis_data_in_db) == set(expected_abis_data)
+        contracts_data_in_db = cursor.execute('SELECT address, chain_id, name, abi, deployed_block FROM contract_data').fetchall()  # noqa: E501
```
yabirgb, line 222: the fetchall here is not needed

rotkehlchen/db/updates.py
```diff
@@ -154,6 +155,53 @@ def update_rpc_nodes(self, to_version: int) -> None:
             new_default_nodes=new_default_nodes,
         )
 
+    def update_contracts(self, to_version: int) -> None:
+        """Updates evm contracts and ABIs in globaldb"""
+        updated_contracts: dict
+        success, updated_contracts = self._get_file_content(  # type: ignore[assignment]  # updated_contracts is a dictionary for sure  # noqa: E501
+            file_url=CONTRACTS_URL.format(branch=self.branch, version=to_version),
+            update_type=UpdateType.CONTRACTS,
+        )
```
yabirgb, line 164: this is ugly because is breaking the types. `_get_file_content` expects to return a list of dicts. You should overload it if is not the case and adjust properly the type

rotkehlchen/db/updates.py
```diff
@@ -154,6 +155,53 @@ def update_rpc_nodes(self, to_version: int) -> None:
...
+            ))
+
+        new_contracts_data = []
+        for single_contract_data in updated_contracts['contracts_data']:
+            new_contracts_data.append((
+                single_contract_data['address'],
+                single_contract_data['chain_id'],
+                single_contract_data.get('name'),
+                single_contract_data['abi'],
+                single_contract_data.get('deployed_block'),
```
yabirgb, line 184: Shouldn't the schema be consistent? I would provide all the fields in the json data and set them to null there

rotkehlchen/db/updates.py
```diff
@@ -154,6 +155,53 @@ def update_rpc_nodes(self, to_version: int) -> None:
...
+            ))
+
+        with GlobalDBHandler().conn.write_ctx() as write_cursor:
+            write_cursor.execute('DELETE FROM contract_data')
+            write_cursor.execute('DELETE FROM contract_abi')
```
yabirgb, line 189: why delete what we have? If you do incremental versions you won't delete anything otherwise what was provided in v1 will be lost in v2.

The good thing to go with versioning here is that you avoid downloading huge files

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -172,3 +187,37 @@ def test_update_rpc_nodes(data_updater: RotkiDataUpdater) -> None:
...
+        expected_abis_data.append((abi_data['id'], abi_data['value'], abi_data['name']))
+
+    expected_contracts_data = []
+    for contract_data in MOCK_CONTRACTS_DATA:
+        expected_contracts_data.append((
+            contract_data['address'],
+            contract_data['chain_id'],
+            contract_data.get('name'),
+            contract_data['abi'],
+            contract_data.get('deployed_block'),
```
yabirgb, line 215: same comment as before

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -172,3 +187,37 @@ def test_update_rpc_nodes(data_updater: RotkiDataUpdater) -> None:
...
+
+
+def test_update_contracts(data_updater: RotkiDataUpdater) -> None:
+    """Tests functionality for remotely updating contracts"""
+    with GlobalDBHandler().conn.read_ctx() as cursor:
+        # Check state before the update
+        cursor.execute('SELECT COUNT(*) FROM contract_abi')
+        assert cursor.fetchone()[0] == 73
+        cursor.execute('SELECT COUNT(*) FROM contract_data')
+        assert cursor.fetchone()[0] == 132
```
yabirgb, line 199: you should state in a comment where those numbers come from. In fact we will have a problem because if in the next update those change due to the contracts being in the globaldb this test will fail. With the assets what we do is ship them already in the packaged globaldb

yabirgb (commented):
Thanks, needs adjustment

rotkehlchen/db/updates.py
nebolax: Okay

rotkehlchen/db/updates.py
nebolax: Mmm yes it makes sense, however the problem is that contracts reference abi ids. And to make sure that the referenced abi id are the needed ones I think it will be needed to provide all abis and all contracts here.

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -16,52 +17,82 @@
...
+                    {
+                        'address': OPTIMISM_SPAM_ASSET_ADDRESS,
+                        'name': 'spooky-v3.xyz',
+                        'symbol': 'Visit https://spooky-v3.xyz and claim rewards',
+                        'decimals': 18,
+                        'chain': 'optimism',
+                    },
+                ],
+            }
+            return MockResponse(200, json.dumps(data))
```
yabirgb, line 61: can you set `data = None` and avoid repeating this with something like

```python
if data is None:
     raise AssertionError(f'Unexpected url {url} called')

return  MockResponse(200, json.dumps(data))
```

rotkehlchen/db/updates.py
```diff
@@ -154,6 +155,51 @@ def update_rpc_nodes(self, to_version: int) -> None:
...
+
+        with GlobalDBHandler().conn.write_ctx() as write_cursor:
+            write_cursor.executemany(
+                'INSERT INTO contract_abi(id, value, name) VALUES(?, ?, ?)',
+                new_abi_data,
+            )
+            write_cursor.executemany(
+                'INSERT INTO contract_data(address, chain_id, name, abi, deployed_block) VALUES(?, ?, ?, ?, ?)',  # noqa: E501
+                new_contracts_data,
+            )
```
yabirgb, line 195: This can raise foreign key error. Can you catch it? I believe it should rollback

rotkehlchen/db/updates.py
```diff
@@ -154,6 +155,51 @@ def update_rpc_nodes(self, to_version: int) -> None:
...
+                single_contract_data['address'],
+                single_contract_data['chain_id'],
+                single_contract_data.get('name'),
+                single_contract_data['abi'],
+                single_contract_data.get('deployed_block'),
+            ))
+
+        with GlobalDBHandler().conn.write_ctx() as write_cursor:
+            write_cursor.executemany(
+                'INSERT INTO contract_abi(id, value, name) VALUES(?, ?, ?)',
```
yabirgb, line 189: Can you change this for a insert or replace? Is what we do in other cases and I believe is best to avoid errors here if for some reason the id exists. Also helps to updated it if we need

rotkehlchen/db/updates.py
```diff
@@ -154,6 +155,51 @@ def update_rpc_nodes(self, to_version: int) -> None:
             new_default_nodes=new_default_nodes,
         )
 
+    def update_contracts(self, to_version: int) -> None:
+        """Updates evm contracts and ABIs in globaldb"""
+        updated_contracts: dict
```
yabirgb, line 160: I believe there is no need now for this

```suggestion
```

rotkehlchen/db/updates.py
```diff
@@ -178,6 +228,10 @@ def check_for_updates(self) -> None:
         if local_rpc_nodes_version < latest_rpc_nodes_version:
             self.update_rpc_nodes(latest_rpc_nodes_version)
 
+        latest_contracts_version = remote_information[UpdateType.CONTRACTS.value]['latest']
+        if local_contracts_version < latest_contracts_version:
+            self.update_contracts(latest_contracts_version)
```
yabirgb, line 233: This is repeating over the other differnt  types. Why not move this to a function that iterates UpdateType and with an if clause you choose what function to execute?

yabirgb (commented):
A few comments. Check also the failing tests since there are a few. From your question I had doubts about what happens for code that expects some contracts to exist but they don't because the update hasn't happened yet. In this case follow the discord thread https://discord.com/channels/657906918408585217/1092386945235177492/1092443894052048979

rotkehlchen/db/updates.py
```diff
@@ -154,29 +156,72 @@ def update_rpc_nodes(self, to_version: int) -> None:
...
+                local_version = self._check_for_last_version(
+                    cursor=cursor,
+                    update_type=update_type,
+                )
 
-        latest_rpc_nodes_version = remote_information[UpdateType.RPC_NODES.value]['latest']
-        if local_rpc_nodes_version < latest_rpc_nodes_version:
-            self.update_rpc_nodes(latest_rpc_nodes_version)
+            # Update all remote data
+            latest_version = remote_information[update_type.value]['latest']
```
yabirgb, line 221: In this case you also need to add a key in the dict https://github.com/rotki/data/blob/develop/updates/info.json if we don't have contracts to add maybe you can do a `.get(` instead and if is None just not execute the logic

rotkehlchen/db/updates.py
```diff
@@ -154,29 +156,72 @@ def update_rpc_nodes(self, to_version: int) -> None:
...
+                    write_cursor.executemany(
+                        'INSERT OR REPLACE INTO contract_abi(id, value, name) VALUES(?, ?, ?)',
+                        new_abi_data,
+                    )
+                    write_cursor.executemany(
+                        'INSERT INTO contract_data(address, chain_id, name, abi, deployed_block) VALUES(?, ?, ?, ?, ?)',  # noqa: E501
+                        new_contracts_data,
+                    )
+            except IntegrityError as e:
+                log.error(f'Got error during updating contracts info: {str(e)}')
```
yabirgb, line 199: Should we let the user know in this case? contracts is a bit more critical part

rotkehlchen/db/updates.py
```diff
@@ -154,29 +156,72 @@ def update_rpc_nodes(self, to_version: int) -> None:
...
+                    single_contract_data['abi'],
+                    single_contract_data.get('deployed_block'),
+                ))
+
+            try:
+                with GlobalDBHandler().conn.write_ctx() as write_cursor:
+                    write_cursor.executemany(
+                        'INSERT OR REPLACE INTO contract_abi(id, value, name) VALUES(?, ?, ?)',
+                        new_abi_data,
+                    )
```
yabirgb, line 193: This one doesn't need to be in the try/exceptment but I believe is here so it can be rolled back in the same transaction right?

yabirgb (commented):
One comment and two questions

rotkehlchen/db/updates.py
nebolax: Yes

rotkehlchen/chain/evm/contracts.py
```diff
@@ -131,6 +135,115 @@ def decode_event(
...
+        return None
+
+    if isinstance(abi_data, list) is False:
+        log.error(f'Abi should be a list. But we got {abi_data}')
+        return None
+
+    with GlobalDBHandler().conn.write_ctx() as write_cursor:
+        write_cursor.execute(
+            'INSERT INTO contract_abi(name, value) VALUES(?, ?)',
+            (name, json.dumps(abi_data, separators=(',', ':'))),
```
yabirgb, line 159: why are the separators needed here?

rotkehlchen/chain/evm/contracts.py
```diff
@@ -131,6 +135,115 @@ def decode_event(
...
+        )
+
+    return abi_data  # type: ignore[return-value]  # abi_data is a list here for sure
```
yabirgb, line 162: I'm completely against this ignore types. Can you maybe overload it or in `request_get_dict` we use assertion, imo that is enough here

rotkehlchen/chain/evm/contracts.py
```diff
@@ -131,6 +135,115 @@ def decode_event(
...
+            f'not found them. The contract data was: {contract_data}',
+        )
+        return None
+
+    if contract_data['chain_id'] != target_chain_id.value:
+        log.error(
+            f'Expected contract {name} to have chain id {target_chain_id.value}, '
+            f'but it had chain id {contract_data["chain_id"]}',
+        )
+        return None
```
yabirgb, line 199: what about contracts deployed in different chains with the same name? should maybe `chain_id` be a list

rotkehlchen/chain/evm/contracts.py
```diff
@@ -131,6 +135,115 @@ def decode_event(
...
+    with GlobalDBHandler().conn.write_ctx() as write_cursor:
+        existing_contract_name = write_cursor.execute(
+            'SELECT name FROM contract_data WHERE address=? AND chain_id=?',
+            (contract_data['address'], contract_data['chain_id']),
+        ).fetchone()
+
+        if existing_contract_name is not None:
+            log.error(
+                f'Contract with address {contract_data["address"]} and chain id {contract_data["chain_id"]} '  # noqa: E501
+                f'already exists in the globaldb. Local name: {existing_contract_name[0]}, remote name: {name}',  # noqa: E501
```
yabirgb, line 210: if you are splitting it anyway why not do it in a way to avoid the noqa rules

rotkehlchen/tests/unit/test_evm_contracts.py
```diff
@@ -1,9 +1,40 @@
...
+    'outputs': [{
+        'name': '',
+        'type': 'string',
+    }],
+    'payable': False,
+    'type': 'function',
+}]
+
+TEST_CONTRACT_NAME = 'SOME_CONTRACT_THAT_EXISTS_ON_REMOTE'
+
```
yabirgb, line 29: Do we leave this space between constants? I believe we don't in other places of the code

```suggestion
```

rotkehlchen/tests/unit/test_evm_contracts.py
```diff
@@ -33,3 +64,71 @@ def test_evm_abi_data(globaldb):
...
+def test_get_remote_abi(ethereum_inquirer: 'EthereumInquirer'):
+    """Tests that pulling an ABI from rotki data repository work"""
+    def requests_mock(url, timeout):
+        if url == f'https://raw.githubusercontent.com/rotki/data/develop/contracts/abi/{TEST_ABI_NAME}.json':  # noqa: E501
+            return MockResponse(status_code=200, text=json.dumps(TEST_ABI))
+
+        raise AssertionError(f'Unexpected url {url} was called')
+
+    with patch('requests.get', side_effect=requests_mock):
+        assert ethereum_inquirer.contracts.abi(TEST_ABI_NAME) == TEST_ABI  # type: ignore[call-overload]  # noqa: E501
```
yabirgb, line 78: why this type ignore?

rotkehlchen/chain/evm/contracts.py
```diff
@@ -215,17 +328,18 @@ def abi_or_none(self, name: str) -> Optional[list[dict[str, Any]]]:
...
+            return try_download_abi_from_remote(name)
+        try:
+            abi_data = json.loads(result[0])
+        except json.decoder.JSONDecodeError as e:
+            log.error(
+                f'Failed to decode {name} abi {result[0]} from DB as json due to {str(e)}',
+            )
+            return None
+
+        return abi_data
```
yabirgb, line 342: this changes is just because the cursor  can be closed earlier right?

rotkehlchen/tests/unit/test_evm_contracts.py
```diff
@@ -33,3 +64,71 @@ def test_evm_abi_data(globaldb):
...
+    def requests_mock(url, timeout):
+        if url == f'https://raw.githubusercontent.com/rotki/data/develop/contracts/contract/{TEST_CONTRACT_NAME}.json':  # noqa: E501
+            return MockResponse(status_code=200, text=json.dumps(TEST_CONTRACT))
+
+        raise AssertionError(f'Unexpected url {url} was called')
+
+    expected_contract = EvmContract(
+        address=TEST_CONTRACT['address'],  # type: ignore[arg-type]
+        abi=TEST_CONTRACT['abi'],  # type: ignore[arg-type]
+        deployed_block=TEST_CONTRACT['deployed_block'],  # type: ignore[arg-type]
```
yabirgb, line 114: why are this type ignore needed?

rotkehlchen/tests/unit/test_evm_contracts.py
```diff
@@ -33,3 +64,71 @@ def test_evm_abi_data(globaldb):
...
+        ).fetchone()
+        assert contract_data_in_db == (
+            TEST_CONTRACT['address'],
+            TEST_CONTRACT['chain_id'],
+            TEST_CONTRACT_NAME,
+            json.dumps(TEST_CONTRACT['abi'], separators=(',', ':')),
+            TEST_CONTRACT['deployed_block'],
+        )
+        expected_abi_count = initial_abi_count if reuse_abi is True else initial_abi_count + 1
+        assert cursor.execute('SELECT COUNT(*) FROM contract_abi').fetchone()[0] == expected_abi_count  # noqa: E501
```
yabirgb, line 134: can you also check that querying a contract that exists doesn't trigger a remote call? to prevent that scenario if we make changes in the future

yabirgb (commented):
Looks okey now. Just a few comments regarding things that maybe can be improved. I'm happy because it looks much better than the first iteration

rotkehlchen/chain/evm/contracts.py
nebolax: Yes

rotkehlchen/chain/evm/contracts.py
nebolax: Good question.

> should maybe chain_id be a list

I wouldn't do this since `try_download_contract_from_remote` is always called for a particular chain id.

However I think it will make sense in the contract url/filename. So that a contract url is computed from name + chain id. What do you think?

rotkehlchen/chain/evm/contracts.py
nebolax: Afaik this is just the common format abis. Default json separators are `, ` and `: ` and for abis `,` and `:` are used to make it more compact.

rotkehlchen/chain/evm/contracts.py
nebolax: I actually made an `isinstance` check that it is a list a few lines above. I am confused why mypy thinks that the type here is not a list only. I'll try to solve it.

rotkehlchen/tests/unit/test_evm_contracts.py
nebolax: Okay

rotkehlchen/tests/unit/test_evm_contracts.py
nebolax: `.abi` function for agument `name` supports only values from literal `ETHEREUM_KNOWN_ABI` and `TEST_ABI_NAME` is not a part of that literal.

rotkehlchen/chain/evm/contracts.py
nebolax: Yes

rotkehlchen/chain/evm/contracts.py
yabirgb: Then we would need to mantain duplicated information right? or do you have something in mind to avoid that?

rotkehlchen/tests/unit/test_evm_contracts.py
nebolax: Okay, I figured it out

rotkehlchen/chain/evm/contracts.py
nebolax: No-no, where will the duplicate be? We can just have the chain id in the url. No need to store it in the json data. Or am I missing something?

rotkehlchen/chain/evm/contracts.py
```diff
@@ -131,6 +136,105 @@ def decode_event(
...
+        write_cursor.execute(
+            'INSERT INTO contract_abi(name, value) VALUES(?, ?)',
+            (name, json.dumps(abi_data, separators=(',', ':'))),
+        )
+
+    return abi_data
+
+
+class ContractSchema(Schema):
+    address = fields.String(required=True)  # not using EvmAddressField since it would cause circular imports  # noqa: E501
```
yabirgb, line 163: you can move this schema to `serialization/schemas.py` and should be fine I guess

yabirgb (commented):
Just one small comment to do proper validation

LefterisJP (commented):
Okay so the approach here is totally wrong. Please use your brain a bit more and read the issue descriptions. And if in doubt, which you should have very well had here, ask me for the overall approach.

The issue, just like all the other issues about remotely updating data .... is about remotely and periodically **updating** data.

For example, the spam assets. A periodic query runs to check if we need to pull new data from the remote. This is work that has been already done!!! Look here: https://github.com/rotki/rotki/blob/develop/rotkehlchen/db/updates.py

I mean there is even an `UpdateType.CONTRACTS` that's waiting to be used and has not been used. All these issues are linked and were supposed to be done in the same manner.

- What needs to be done: Periodically update the contract data, including abi data in the Database.
- What you did: Every time you query the DB for a contract by name and it is missing ... you also query the remote??? How did you come up with that? And in what universe is that (a) good / performant / smart and (b) completing what's described in the issue?

As mentioned inside the review contents the remote should have all data by address + chain id. Unambiguous. Not names ... names are irrelevant. And we currently only use them in the codebase as variable names as we transitioned from the keys of the json files to those variable names.

Note: Did not check the tests as the whole thing is off.

LefterisJP (approved):
Looks good to me. have some non-blocking requests but they can be done in a subsequent PR. Will merge as this is small. Please check the review's 3 requests.

PR #5853, opened 2023-04-03
Periodically query beaconchain for the produced blocks

LefterisJP:


rotkehlchen/tasks/manager.py
```diff
@@ -577,6 +579,27 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
                 )]
         return None
 
+    def _maybe_query_produced_blocks(self) -> Optional[list[gevent.Greenlet]]:
+        """Schedules the blocks production query if enough time has passed"""
+        with self.database.conn.read_ctx() as cursor:
+            result = self.database.get_used_query_range(cursor, LAST_PRODUCED_BLOCKS_QUERY_TS)
+            if result is not None and ts_now() - result[1] < 1:
```
yabirgb, line 586: Isn't `ts_now() - result[1] < 1` query if it was queried less than a second ago? Shouldn't we be using greater space times?

yabirgb (commented):
Question

rotkehlchen/tasks/manager.py
LefterisJP: I am not in a good state of mind today. I was supposed to correct this before pushing.

PR #5856, opened 2023-04-03
Add colibri base code

yabirgb:
Closes #(issue_number)

## Checklist


Dockerfile
```diff
@@ -17,9 +18,12 @@ ARG PYINSTALLER_VERSION=v5.7.0
 RUN python3 -m venv /opt/venv
 ENV PATH="/opt/venv/bin:$PATH"
 
-RUN if [ "$TARGETARCH" != "amd64" ]; then curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y; fi
+WORKDIR /app
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 ENV PATH="/root/.cargo/bin:${PATH}"
-RUN if [ "$TARGETARCH" != "amd64" ]; then rustup default nightly-2021-03-24; fi
+RUN rustup default nightly-2021-03-24
```
kelsos, line 24: while it is not related to the PR directly do you think we can update from this old nightly?

yabirgb: did you pin it for some reason?

kelsos: Wasn't me
image

It was an external contributor that made the aarch64 docker contribution.

But it was around two years ago so it was probably the last one around that time and whatever feature was missing, it should have landed to `stable` after two years.

package.py
```diff
@@ -988,6 +995,23 @@ def build(self) -> None:
...
+        )
+        if build_ret_code != 0:
+            logger.error('packaging failed')
+            sys.exit(1)
+
+        backend_binary = colibri_directory / 'release' / 'colibri'
+        ret_code = subprocess.call(f'{backend_binary}', shell=True)
+        if ret_code != 0:
+            logger.error('colibri binary check failed')
+            sys.exit(1)
```
kelsos, line 1013: For macOS we need to build for two different targets and then merge with lipo

`aarch64-apple-darwin` and `x86_64-apple-darwin`

for the merging, I think we can use this

https://github.com/rotki/rotki/blob/ea75e4aeec8d79d0b5e2e8463ddb9666fc095f1a/package.py#L430

I am not sure how we can use `rustup` from the action to support both the architectures targets

colibri/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+cargo-features = ["edition2021"]
```
kelsos, line 1: This gives a warning that is not needed anymore if you check the CI

kelsos: Ok, so this was only needed because we are using a really old compiler version for the docker image. Let's update the docker image to use the stable toolchain instead.

.github/workflows/rotki_dev_builds.yml
```diff
@@ -58,6 +58,10 @@ jobs:
           key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
           restore-keys: |
             ${{ runner.os }}-pnpm-store-
+      - name: Setup rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
```
kelsos, line 64: @yabirgb can you find an alternative to this action? 

I checked the warnings about set-output/node12 in the logs and then went to the repo, and the last update was a bit more than 3 years ago, which makes it unmaintained. This essentially means that in the near future it will break. 

If there is no official action, we should do the setup by a script.

kelsos (approved):
image

frontend/app/.electron-builder.config.js
```diff
@@ -14,12 +23,9 @@ if (buildDebPackage) {
...
   appId: "com.rotki.app",
   directories: {
-    output: 'build',
-    buildResources: 'buildResources',
+    output: "build",
+    buildResources: "buildResources"
   },
-  files: [
-    'dist/**',
-  ],
```
lukicenturi, line 22: this

PR #5861, opened 2023-04-04
Refactor pagination filters for binance savings

tewshi:
## Checklist

- [x] extends the pagination/filter composable to allow passing default collection state function
- [x]  The PR modified the frontend, and updated the binance savings balance component to use pagination/filter composable
- [x]  adds unit test for the composable's use case with binance savings

### DIFF
--- a/frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
+++ b/frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
@@ -1,16 +1,10 @@
 <script setup lang="ts">
 import { type PropType, type Ref } from 'vue';
 import { type DataTableHeader } from 'vuetify';
-import { type Collection } from '@/types/collection';
-import { defaultCollectionState } from '@/utils/collection';
-import { useEmptyFilter } from '@/composables/filters';
 import { type ActionStatus } from '@/types/action';
 import { type IgnoredAssetsHandlingType } from '@/types/asset';
 import { type Module } from '@/types/modules';
-import {
-  type NonFungibleBalance,
-  type NonFungibleBalancesRequestPayload
-} from '@/types/nfbalances';
+import { type NonFungibleBalance } from '@/types/nfbalances';
 import { type ManualPriceFormPayload } from '@/types/prices';
 import { Section } from '@/types/status';
 import { assert } from '@/utils/assertions';
@@ -187,17 +181,11 @@ const {
   options,
   setPage,
   setOptions
-} = useHistoryPaginationFilter<
-  NonFungibleBalance,
-  NonFungibleBalancesRequestPayload,
-  NonFungibleBalance,
-  Collection<NonFungibleBalance>
->(
+} = useHistoryPaginationFilter<NonFungibleBalance>(
   null,
   true,
   useEmptyFilter,
   fetchNonFungibleBalances,
-  defaultCollectionState,
   {
     onUpdateFilters(query) {
       set(ignoredAssetsHandling, query.ignoredAssetsHandling || 'exclude');
--- a/frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
+++ b/frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
@@ -1,6 +1,8 @@
 <script setup lang="ts">
 import { type PropType, type Ref } from 'vue';
 import { type DataTableHeader } from 'vuetify';
+import { type Collection } from '@/types/collection';
+import { defaultCollectionState } from '@/utils/collection';
 import { useEmptyFilter } from '@/composables/filters';
 import { type ActionStatus } from '@/types/action';
 import { type IgnoredAssetsHandlingType } from '@/types/asset';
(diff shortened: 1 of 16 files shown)


frontend/app/src/composables/history/filter-paginate.ts
```diff
@@ -27,24 +27,27 @@ interface FilterSchema<F, M> {
...
  * @param {MaybeRef<string | null>} locationOverview
  * @param {MaybeRef<boolean>} mainPage
  * @param {() => FilterSchema<W, X>} filterSchema
  * @param {(payload: MaybeRef<U>) => Promise<Collection<V>>} fetchAssetData
+ * @param {() => S} defaultCollection
  * @param {{onUpdateFilters?: (query: LocationQuery) => void, extraParams?: ComputedRef<LocationQuery>, customPageParams?: ComputedRef<Partial<U>>, defaultSortBy?: {pagination?: keyof T, pageParams?: (keyof T)[], pageParamsAsc?: boolean[]}}} options
  */
 export const useHistoryPaginationFilter = <
   T extends Object,
   U,
```
kelsos, line 40: So I am checking a few things and I think it might be safe to assume that `U extends PaginationRequestPayload<T>` can you check if this breaks anything or not?

Maybe we can make it default to `PaginationRequestPayload<T>`


https://github.com/rotki/rotki/blob/3d879ca26c97b7ae40364c724031848ee698faa2/frontend/app/src/types/common.ts#L1

frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
```diff
@@ -188,18 +190,26 @@ const {
...
-  onUpdateFilters(query) {
-    set(ignoredAssetsHandling, query.ignoredAssetsHandling || 'exclude');
-  },
-  extraParams,
-  defaultSortBy: {
-    pagination: 'name',
-    pageParams: ['name'],
-    pageParamsAsc: [true]
+  NonFungibleBalance,
+  Collection<NonFungibleBalance>
```
kelsos, line 194: In this case, do we need to specify it here explicitly? since we are using a default Generic value?

Also maybe for this case we would use a `V = T` default and at least when we have a default Collection and no transformation it would get us rid of the two extra types in the generic

frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
```diff
@@ -188,18 +190,26 @@ const {
...
+>(
+  null,
+  true,
+  useEmptyFilter,
+  fetchNonFungibleBalances,
+  defaultCollectionState,
```
kelsos, line 200: Wouldn't it make sense to move the collection state to the options instead and use the default internally only if there is no external state?

kelsos (commented):
thanks @tewshi I left you some comments let me know what you think.

frontend/app/src/composables/history/filter-paginate.ts
tewshi: checking it.

frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
tewshi: this I didn't consider, I'd refactor it now.

frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
tewshi: I did initially, but was causing type errors, but I'll check again.

frontend/app/src/composables/history/filter-paginate.ts
tewshi: yes it does, but everywhere, there's additional props added to U, setting it as default doesn't break anything, it only means I have to find a way to either swap U with V, or make V have a default value (because Required type parameters may not follow optional type parameters)

tewshi: done, since we're also setting `V = T`

frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
tewshi: this will require one typecasting... 

```typescript
  const getCollectionDefault = (): S => {
    if (defaultCollection) {
      return defaultCollection();
    }

    return defaultCollectionState<V>() as S;
  };
```

been trying to avoid it, it's pressing my neck. 😛

kelsos: > this will require one typecasting...
> 
> ```ts
>   const getCollectionDefault = (): S => {
>     if (defaultCollection) {
>       return defaultCollection();
>     }
> 
>     return defaultCollectionState<V>() as S;
>   };
> ```
> 
> been trying to avoid it, it's pressing my neck. 😛

Yes, it is a bit weird, do what you can and I will also have a look. The type system is really powerful but some times it can be a timesink to figure out how to make it work properly without force.

tewshi: Ok, will push in a bit.

PR #5862, opened 2023-04-04
Use function scope bittrex in most tests

LefterisJP:
This is to make sure that session scope bittrex which uses a session MessagesAggregator is not used.


LefterisJP:
Merging since frontend failure can't be related. This PR only touched tests

PR #5866, opened 2023-04-05
Remote global address book updates

### LINKED ISSUE #5625: Provide remote updates for the global address book
## Abstract

We have a global name registry (address book) in the global DB. This is editable by the user. But unlike other global data such as assets, spam, assets, rpc nodes etc. we do not push data updates for it yet.

## Task

The approach should be same as in #5532 

Related issues: 

- #5594 
- #5532 
- #5236

nebolax:
Closes #5625

## Checklist

### DIFF
--- a/rotkehlchen/db/updates.py
+++ b/rotkehlchen/db/updates.py
@@ -96,7 +96,7 @@ class RotkiDataUpdater:
 
     def update_spam_assets(self, from_version: int, to_version: int) -> None:
         """
-        Check the updates in the inclusive range [local_version + 1, to_version] and update the
+        Check the updates in the inclusive range [from_version + 1, to_version] and update the
         spam assets for those versions. Assets are also added to the globaldb if they don't exist
         locally
         """
@@ -120,7 +120,7 @@ class RotkiDataUpdater:
                     (UpdateType.SPAM_ASSETS.serialize(), version),
                 )
 
-    def update_rpc_nodes(self, from_version: int, to_version: int) -> None:  # pylint: disable=unused-argument  # from_version is here to have the same format as other updating functions  # noqa: E501
+    def update_rpc_nodes(self, from_version: int, to_version: int) -> None:  # pylint: disable=unused-argument  # noqa: E501
         """Checks the update for the latest version and update the default rpc nodes in the global db.
 
         It also updates the user db with these default nodes.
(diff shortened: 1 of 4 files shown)


rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -172,3 +208,41 @@ def test_update_rpc_nodes(data_updater: RotkiDataUpdater) -> None:
...
+    db_addressbook = DBAddressbook(data_updater.user_db)
+    # check state of the address book before updating
+    with GlobalDBHandler().conn.read_ctx() as cursor:
+        cursor.execute('SELECT COUNT(*) FROM address_book')
+        assert cursor.fetchone()[0] == 0
+
+    # Add some initial entries. "Local name 1" doesn't conflict with any of the remote names
+    # and should stay as is after the update application. "Local name 2" conflicts with
+    # "Remote name 3", but since it is local, it should be preferred over the remote entry
+    # and should also stay in the db as is.
```
prettyirrelevant, line 224:
```suggestion
    # populate the `addressbook` table and see that existing local entries are preferred to their remote counterpart
```

prettyirrelevant (commented):
A question.

If an update contains a wrong name for an address, and a new update is created(containing the right name for that entry), your current implementation won't handle that right?

Should that be an expected behaviour?

nebolax:
> A question.
> If an update contains a wrong name for an address, and a new update is created(containing the right name for that entry), your current implementation won't handle that right?
> Should that be an expected behaviour?

I would say that it is kinda overkill~ish. I would ask @LefterisJP whether this is needed.

LefterisJP:
> > A question.
> > If an update contains a wrong name for an address, and a new update is created(containing the right name for that entry), your current implementation won't handle that right?
> > Should that be an expected behaviour?
> 
> I would say that it is kinda overkill~ish. I would ask @LefterisJP whether this is needed.

We should be able to update names. That's an important point of the existence of remote data. Say if an address is compromised. First it's a Kraken address. Then something happens and it's a hacked compromised address. We would update our remote data and this should be pulled downstream.

rotkehlchen/db/updates.py
```diff
@@ -154,6 +162,47 @@ def update_rpc_nodes(self, to_version: int) -> None:
...
+                    ),
+                )
+                if existing_name is None:
+                    entries_to_add.append(entry)
+                else:
+                    log.debug(
+                        f'Global address book entry for address {entry.address} and blockchain '
+                        f'{entry.blockchain} already exists (probably was input by the user). '
+                        f'Update for this entry will not be applied.',
+                    )
```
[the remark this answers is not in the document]
nebolax: In the other PR (curve api pr) Lefteris said that we should not add things like `exists_ok` flag to `add_addresbook_entries` method, so I did it like this here. I'll also add the check here whether to force insert remote name, and for this it will probably be needed to get current local names.

prettyirrelevant (commented):
some comments

docs/changelog.rst
```diff
@@ -3,6 +3,7 @@ Changelog
 =========
 
 * :feature:`5816` The NFT images will not be automatically rendered now. It is made so to prevent a known security issue, that may result in leakage of your privacy (read https://medium.com/@alxlpsc/critical-privacy-vulnerability-getting-exposed-by-metamask-693c63c2ce94 ). You can add domains you trust to the whitelisted domain in the NFT setting.
+* :feature:`5625` Global address book will now receive automatic updates regularly i.e. known addresses will be associated with names.
```
LefterisJP, line 6: Perhaps let's not advertise this. As discussed this will only have some of our small updates to our own remote data and not the usage of other sources. So such a changelog probably may create expectations

rotkehlchen/db/updates.py
```diff
@@ -154,6 +163,70 @@ def update_rpc_nodes(self, to_version: int) -> None:
...
+                existing_name = db_addressbook.get_addressbook_entry_name(
+                    book_type=AddressbookType.GLOBAL,
+                    chain_address=OptionalChainAddress(
+                        address=address,
+                        blockchain=blockchain,
+                    ),
+                )
+                if existing_name is None:
+                    entries_to_add.append(entry)
+                elif raw_entry['old_name'] is not None and existing_name == raw_entry['old_name']:
```
LefterisJP, line 206: If for some reason we miss it you were gonna get a key error here

```suggestion
                elif raw_entry.get('old_name') is not None and existing_name == raw_entry['old_name']:
```

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -5,21 +5,71 @@
...
+        address=ADDRESSBOOK_ADDRESS_2,
+        blockchain=SupportedBlockchain.ETHEREUM,
+    ),
+]
+SERIALIZED_REMOTE_ADDRESSBOOK = [
+    {
+        'address': ADDRESSBOOK_ADDRESS_1,
+        'blockchain': 'ETH',
+        'new_name': 'Remote name that doesnt conflict with anything',
+        'old_name': None,
```
LefterisJP, line 48: I would advise putting `old_name` as optional. A lot less data to serialize for only an edge case. Only add `old_name` in the remote if it's needed. It's easy to guard against this in the app code as I showed you in the other comment and this really helps us as makes the size of the data in the remote and that needs to be serialized over the network smaller.

LefterisJP (commented):
This needs a rebase and conflicts to be solved.

Also some more comments to act on

LefterisJP (approved):
LGTM

PR #5868, opened 2023-04-05
fix: asset select abort wrong controller

lukicenturi:
Closes #(issue_number)

Fix a very rare case where the wrong abortController is called.

### DIFF
--- a/frontend/app/src/components/inputs/AssetSelect.vue
+++ b/frontend/app/src/components/inputs/AssetSelect.vue
@@ -106,7 +106,7 @@ const searchAssets = async (
   }
 };
 
-const pending: Record<string, AbortController> = {};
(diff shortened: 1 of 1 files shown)


frontend/app/src/components/inputs/AssetSelect.vue
```diff
@@ -118,18 +118,18 @@ watch(search, search => {
 
 watchThrottled(
```
lukicenturi, line 119: so for the watchThrottled, for example we type `BTC`
it runs twice for 
(search = "B", old = "") and
(search = "BTC", old = "BT")

so when it's the turn for `search=BTC` it tried to abort controller for `BT` which doesn't exist, instead it should abort controller for `B`

kelsos (approved):
lgtm

PR #5874, opened 2023-04-06
Refactor pagination filters for managed assets

tewshi:
## Checklist

- [x] The PR modified the frontend, and updated the managed assets component to use pagination/filter composable
- [x] use a more generic name for the composable as it is no longer used only for history
- [x] refactor props definition, use more modern Vue style which has good ts check support


kelsos (commented):
The code is ok I don't have any comments there, but I noticed something when navigating back in the managed assets. 

The page does not update properly so let's say I was 31-40, when I go to the next page using the buttons (41-50) if I press back the page will change but the display will remain 41-50. Instead of going back to 31-40 as it should. 

@tewshi can you figure it out and maybe also add some cypress test for this?

tewshi:
> The code is ok I don't have any comments there, but I noticed something when navigating back in the managed assets.
> 
> The page does not update properly so let's say I was 31-40, when I go to the next page using the buttons (41-50) if I press back the page will change but the display will remain 41-50. Instead of going back to 31-40 as it should.
> 
> @tewshi can you figure it out and maybe also add some cypress test for this?

Checking and fixing

PR #5875, opened 2023-04-07
Catch one more case for hop bridging

nebolax:
Catch one more case for HOP bridging

### DIFF
--- a/rotkehlchen/chain/optimism/modules/hop/decoder.py
+++ b/rotkehlchen/chain/optimism/modules/hop/decoder.py
@@ -20,6 +20,7 @@ if TYPE_CHECKING:
     from rotkehlchen.user_messages import MessagesAggregator
 
 ETH_BRIDGE = string_to_evm_address('0x83f6244Bd87662118d96D9a6D44f09dffF14b30E')
+ETH_WRAPPER = string_to_evm_address('0x86cA30bEF97fB651b8d866D45503684b90cb3312')
 
 TRANSFER_FROM_L1_COMPLETED = b'2\tX\x17i0\x80N\xb6l#C\xc74?\xc06}\xc1bIY\x0c\x0f\x19W\x83\xbe\xe1\x99\xd0\x94'  # noqa: E501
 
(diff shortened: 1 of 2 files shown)


rotkehlchen/chain/optimism/modules/hop/decoder.py
```diff
@@ -41,24 +43,41 @@ def __init__(  # pylint: disable=super-init-not-called
...
+            for event in context.decoded_events:
+                if event.event_type == HistoryEventType.RECEIVE and event.event_subtype == HistoryEventSubType.NONE and recipient == event.location_label and event.asset == A_ETH:  # noqa: E501
+                    event.event_type = HistoryEventType.WITHDRAWAL
+                    event.event_subtype = HistoryEventSubType.BRIDGE
+                    event.counterparty = CPT_HOP
+                    event.notes = f'Bridge {event.balance.amount} ETH via hop protocol'
+                    event.extra_data = {'sent_amount': str(heth_amount)}
+                    break
+        else:
+            # No event for a particular withdrawal is emitted, so we can't verify the recipient,
```
LefterisJP, line 66: how come no event is emitted. Isn't there some wrapping event?

rotkehlchen/tests/unit/decoders/test_hop.py
```diff
@@ -85,3 +85,35 @@ def test_hop_optimism_eth_receive(database, optimism_inquirer):
             extra_data={'sent_amount': '0.2'},
         )]
     assert expected_events == events
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('optimism_accounts', [['0x4bBa290826C253BD854121346c370a9886d1bC26']])
+def test_hop_optimism_eth_receive_2(database, optimism_inquirer, optimism_accounts):
```
LefterisJP, line 92: Also describe in the docstring what's the difference with the previous one.

```suggestion
def test_hop_optimism_eth_receive_no_event(database, optimism_inquirer, optimism_accounts):
```


Can you tell me also what did you do differently? Why is the `TransferFromL1Completed` event missing?

LefterisJP (approved):
As discussed in discord this can probably be done better with enricher but let's consider this a small improvement that can be iterated on.

Issue #5873, opened 2023-04-06
Enhance address book


nebolax:
## Abstract

Enhance address book and use external apis to fetch names of addresses.

## Specification

- [ ] Add source column to global (only global) address book that would help differentiate entries added manually by the user, entries added by some background tasks (e.g. curve cache querying), entries added from an external api, etc.
- [ ] Display the names grouped by source in frontend. This is important so that users don't lose manually added entries in the huge list of all entries.
- [ ] Make an order (probably hard-coded for now) of global address book sources that will be used when searching for a name of an address.
- [ ] Add a list of external apis, query them when an address is missing a name in the current address book, and cache the results.

PR #5876, opened 2023-04-07
Enhance address book

nebolax:
Backend part of #5873


rotkehlchen/db/filtering.py
```diff
@@ -1537,3 +1545,132 @@ def make(
...
+            ))
+
+        filter_query.filters = filters
+        return filter_query
+
+    def get_book_type(self) -> AddressbookType:
+        return AddressbookType.USER
+
+    def get_base_query(self) -> str:
+        return 'SELECT address, name, blockchain FROM address_book '
```
yabirgb, line 1636: this is not the right place for those since you are mixing concerns. One thing is the filters and other is the structures that is related to. Those can be functions that check the class type and return the proper value.

rotkehlchen/db/filtering.py
```diff
@@ -1537,3 +1545,132 @@ def make(
...
+            ))
+        filters.extend(filter_query.filters)
+        filter_query.filters = filters
+        return filter_query
+
+    def get_book_type(self) -> AddressbookType:
+        return AddressbookType.GLOBAL
+
+    def get_base_query(self) -> str:
+        return 'SELECT address, name, blockchain, source FROM address_book '
```
yabirgb, line 1676: same comment as before

rotkehlchen/db/filtering.py
```diff
@@ -1537,3 +1545,132 @@ def make(
...
+        filters: list[DBFilter] = []
+        if sources is not None:
+            filters.append(DBMultiIntFilter(
+                and_op=True,
+                column='source',
+                operator='IN',
+                values=[source.serialize_for_db() for source in sources],
+            ))
+        filters.extend(filter_query.filters)
+        filter_query.filters = filters
```
yabirgb, line 1669: why not do the append directly to `filter_query.filters`

rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -9,60 +9,96 @@
...
-KELSOS_BOOK_ENTRY = AddressbookEntry(
-    name='new kelsos',
-    address=KELSOS_ADDR,
-    blockchain=None,
-)
+
+
+def make_kelsos_book_entry(book_type: AddressbookType) -> NamedAddressbookEntry:
+    if book_type == AddressbookType.USER:
+        return UserAddressbookEntry(
```
yabirgb, line 49: add a docstring explaining what those do. Here and below

rotkehlchen/tests/utils/factories.py
```diff
@@ -159,24 +163,59 @@ def generate_tx_entries_response(
...
+
+def make_addressbook_entry(
+        book_type: AddressbookType,
+        address: ChecksumEvmAddress,
+        name: str,
+        blockchain: Optional[SupportedBlockchain],
+) -> NamedAddressbookEntry:
+    if book_type == AddressbookType.USER:
+        return UserAddressbookEntry(
+            address=address,
```
yabirgb, line 184: This is the same as `make_kelsos_book_entry` so you can import it there

yabirgb: add a docstring please

rotkehlchen/types.py
```diff
@@ -646,7 +647,8 @@ class CostBasisMethod(SerializableEnumMixin):
     ACB = auto()
 
 
-class AddressbookEntry(NamedTuple):
+@dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=True, frozen=False)
+class NamedAddressbookEntry():
```
yabirgb, line 651: This dataclass is not supposed to be instantiated right? is there a way to mark it in addition to having it in the docstring? Not sure if just adding metaclass as parent class avoids that

rotkehlchen/types.py
```diff
@@ -658,23 +660,90 @@ def serialize(self) -> dict[str, Optional[str]]:
...
+        - DeserializationError if blockchain is not known
+        """
+        return cls(
+            address=data['address'],
+            name=data['name'],
+            blockchain=SupportedBlockchain.deserialize(data['blockchain']) if data['blockchain'] is not None else None,  # noqa: E501
+        )
+
+    def __str__(self) -> str:
+        return f'Named addressbook entry with name "{self.name}", address "{self.address}" and blockchain {self.blockchain.value if self.blockchain is not None else None}'  # noqa: E501
```
yabirgb, line 688: since is the same string maybe you could have this in a method called `_base_str` and then in the subclass call 

```
return super()._base_str.format(book_name='your book name')
```

rotkehlchen/chain/evm/names.py
```diff
@@ -166,7 +174,7 @@ def _private_addressbook_address_to_name(
     """
     db_addressbook = DBAddressbook(db)
     return db_addressbook.get_addressbook_entry_name(
-        book_type=AddressbookType.PRIVATE,
+        book_type=AddressbookType.USER,
```
yabirgb, line 177: is there really a need to rename from user to private?

rotkehlchen/chain/evm/names.py
```diff
@@ -24,6 +25,13 @@
     from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
 
 
+GLOBAL_ADDRESSBOOK_PRIORITIES = {
+    GlobalAddressbookSource.MANUAL: 0,
+    GlobalAddressbookSource.CURVE_CACHE: 1,
+    GlobalAddressbookSource.WALLET_LABELS_XYZ: 2,
```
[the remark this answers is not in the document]
yabirgb: Can you also specify in a comment if 0 is highest priority or lowest? I know is sorting from lower to higher later but is good to have it here to avoid misunderstandings later

rotkehlchen/db/addressbook.py
```diff
@@ -179,20 +168,49 @@ def delete_addressbook_entries(
...
+        if len(result) == 0:
+            return None
+        elif book_type == AddressbookType.USER:
+            return result[0][0]  # There is always a single entry for user address book
+        else:  # global address book
+            deserialized_result = [
+                (name, GlobalAddressbookSource.deserialize_from_db(source))
+                for name, source in result
+            ]
+            deserialized_result.sort(key=lambda entry: priorities[entry[1]])  # type: ignore  # for global address book priorities is not None  # noqa: E501
```
yabirgb, line 215: why is the type not properly detected if you have the function overloaded? Maybe is because you need to add it as None to the other definition?

yabirgb (commented):
This is a really big PR, we should try to make smaller changes if possible. Left some things that need fixing and questions

LefterisJP:
Closing the PR. Too old, out of sync and issue not a priority.

PR #5877, opened 2023-04-09
Remove decoded events from tx api. Enrich history events api

LefterisJP:
Group history events query via api for event ids and make sure that api can handle all the responsibilities of querying history events.

Also adjust all related tests querying events as they would be queried by the frontend instead of using the transactions query API.


LefterisJP:
Working on fixing failing tests here

LefterisJP:
All tests green. Merging this.

PR #5881, opened 2023-04-11
fix: e2e for ignored assets

lukicenturi:
Closes #(issue_number)

## Checklist


tewshi (approved):
lgtm

lukicenturi:
> on a second thought, if the endpoint will still be used, why not just remove the call to the function in `frontend/app/src/composables/session/load.ts` and everywhere it is imported... so that the api function can still remain.
> 
> but it's fine if we want to take it out completely for now

hmm it won't be used anymore. It will be replaced by new endpoint.

PR #5883, opened 2023-04-11
Update bitpanda mappings

yabirgb:
When checking the issue that the user reported about bitpanda saw that we were missing mappings. It seems that their endpoint doesn't provide all the available assets.


LefterisJP (approved):
LGTM

PR #5884, opened 2023-04-11
Move ignored_in_accounting flag to evm events for API responses

yabirgb:
Closes #(issue_number)

## Checklist


docs/api.rst
```diff
@@ -4917,8 +4917,9 @@ Dealing with History Events
                       "timestamp": 1642802807
                   },
                   "customized": false,
+                  "ignored_in_accounting": false,
                   "has_details": false,
-		  "grouped_events_num": 1
+                  "grouped_events_num": 1
```
yabirgb, line 4922: this change is because the line was using tab instead of spaces for indentation

rotkehlchen/api/rest.py
```diff
@@ -3570,13 +3559,29 @@ def get_history_events(
                 cursor=cursor,
                 chain_id=chain_id,  # type: ignore
             )
+            ignored_ids_mapping = self.rotkehlchen.data.db.get_ignored_action_ids(
+                cursor=cursor,
+                action_type=ActionType.EVM_TRANSACTION,
```
nebolax, line 3564: Shouldn't it be checking for all events and not only evm transactions?

yabirgb: At the moment `ActionType` can only be

```
class ActionType(DBEnumMixIn):
    TRADE = 1
    ASSET_MOVEMENT = 2
    EVM_TRANSACTION = 3
    LEDGER_ACTION = 4
```

and the only history events that we currently ignore are the evm transactions. Ledger actions, trades and asset movements are not displayed in this view yet

docs/api.rst
LefterisJP: Thanks. I think most of these mistakes is due to some fucked up settings in my editor for rst

docs/api.rst
```diff
@@ -4945,7 +4947,7 @@ Dealing with History Events
...
+   :resjson list decoded_events: A list of history events. Each event is an object comprised of the event entry and a boolean denoting if the event has been customized by the user or not. Each entry also has a `has_details` flag. If `has_details` is true, then it is possible to call /history/events/details endpoint to retrieve some extra information about the event. Also each entry has a `customized` flag denoting if the event has been customized/added by the user. Finally if `group_by_event_ids` is true, each entry contains `grouped_events_num` which is an integer with the amount of events under the common event identifier. The consumer has to query this endpoint again with `group_by_event_ids` set to false and with the `event_identifiers` filter set to the identifier of the events having more than 1 event. Finally `ignored_in_accounting` is set to `true` when the user has marked as ignored this event.
```
LefterisJP, line 4950:
```suggestion
   :resjson list decoded_events: A list of history events. Each event is an object comprised of the event entry and a boolean denoting if the event has been customized by the user or not. Each entry also has a `has_details` flag. If `has_details` is true, then it is possible to call /history/events/details endpoint to retrieve some extra information about the event. Also each entry has a `customized` flag denoting if the event has been customized/added by the user. Finally if `group_by_event_ids` is true, each entry contains `grouped_events_num` which is an integer with the amount of events under the common event identifier. The consumer has to query this endpoint again with `group_by_event_ids` set to false and with the `event_identifiers` filter set to the identifier of the events having more than 1 event. Finally `ignored_in_accounting` is set to `true` when the user has marked this event as ignored
```

rotkehlchen/api/rest.py
LefterisJP: In a subsequent PR or release we probably need to change/remove this mapping. As we would need to ignore specific events. And since all of the above will become events the mapping makes no sense anymore.

rotkehlchen/api/rest.py
```diff
@@ -3570,13 +3559,29 @@ def get_history_events(
...
+            )
 
+        # mypy fails to correctly detect the type here. It works fine now with pylance
```
LefterisJP, line 3567: What is pylance? We use mypy in the codebase. Please don't add confusing comments. And you are adding a cast below. Ofc it will work with a cast. Just remove the comment if you want to keep the cast

Or what would make more sense would be to overload `get_all_history_events_and_limit_info` depending on the value of the boolean to return proper result.

But still not sure if it will work in this case as the value for `group_by_event_ids` is not a constant and I believe mypy would not follow the logic enough to see that the same argument is checked in the if above.

So I think I would keep the comment and add type ignores. So that if at some point mypy gets smarter and we add the overload we automatically get pinged that this is detected and fixed. Otherwise this cast would stay here forever

```suggestion
        # mypy fails to correctly detect the type here. It works fine now with pylance
```

rotkehlchen/db/history_events.py
```diff
@@ -356,6 +356,19 @@ def get_all_history_events(
...
+            self,
+            cursor: 'DBCursor',
+            filter_query: HistoryBaseEntryFilterQuery,
+            has_premium: bool,
+            group_by_event_ids: bool,
+    ) -> Union[list[tuple[int, HistoryBaseEntry]], list[HistoryBaseEntry]]:
+        """
+        This fallback is needed due to
+        https://github.com/python/mypy/issues/6113#issuecomment-869828434
+        """
```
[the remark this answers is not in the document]
LefterisJP: Ah Isee you did it. Yes it's not. The reason is that mypy does not understand the variable is final. It can change at any point. Perhaps adding Final to the variable before putting it as an argument would work? But sitll probably not and this is a mypy thing to improve

rotkehlchen/db/history_events.py
```diff
@@ -399,19 +412,52 @@ def get_all_history_events(
...
+        ...
+
+    @overload
+    def get_all_history_events_and_limit_info(
+            self,
+            cursor: 'DBCursor',
+            filter_query: HistoryBaseEntryFilterQuery,
+            has_premium: bool,
+            group_by_event_ids: bool,
+    ) -> tuple[Union[list[tuple[int, HistoryBaseEntry]], list[HistoryBaseEntry]], int]:
```
LefterisJP, line 442: Where did `EvmEvent` go?

rotkehlchen/tests/data/expected/test_evm_transactions-test_query_transactions.json
```diff
@@ -1 +1 @@
...
...ae000000330000000000000000000000000000000000000000000000000000000000000003000000000000000000000000000000000000000000000000000000000000001ba36db1f4217e5d2fe99209742d5bbad549cb745940fd9ddc1ecfae06b1013207446649ed77f231e073e6a075082d5c2ace08858c50baa4071a28ad217216d22a0000000000000000000000000000000000000000000000000090943930da640039","nonce":0}},{"entry":{"tx_hash":"0xe32037154199c9b363aacbd6ce853fd4633cc62fdeb34a9305656ee280f4f6ef","evm_chain":"ethereum","timestamp":1671929963,"block_number":16258333,"from_address":"0x56Eddb7aa87536c09CCc2793473599fD21A8b17F","to_address":"0xb8553D9ee35dd23BB96fbd679E651B929821969B","value":"61577330000000000","gas":"207128","gas_price":"12391657059","gas_used":"21000","input_data":"0x","nonce":4031412}},{"entry":{"tx_hash":"0xfeabe8233f374eed505960f4bf8364a9d9398d7e1320ff7df425934134450231","evm_chain":"ethereum","timestamp":1671926939,"block_number":16258082,"from_address":"0x3794Da19a22CeDe7976f9d3181F2CCD58f6408a5","to_address":"0xb8553D9ee35dd23BB96fbd679E651B929821969B","value":"20000000000000000","gas":"21000","gas_price":"10625802284","gas_used":"21000","input_data":"0x","nonce":5}}],"entries_found":49,"entries_total":18,"entries_limit":100}
```
LefterisJP, line 1: Why was a change here needed? Perhaps it would make more sense to VCR the test instead?

yabirgb: This changed because I also removed the flag from the ethereum transactions endpoint as discussed in discord. I believe it doesn't make sense to use VCR here since we are comparing the local output agaisnt the expected one in this file

PR #5885, opened 2023-04-12
Refactor history events

### LINKED ISSUE #5846: New EVM Transactions View in 1.28, stepping stone for unified history
## Abstract

This is part of https://github.com/rotki/rotki/issues/2469 made for 1.28.0.

In 1.28 we will change how EVM transactions view work to make it be based on history events instead of transactions. This is a stepping stone towards going to a fully unified history view which will include everything and be easily filterable and searchable.

For now what will also be displayed here in addition to EVM transactions is the ethereum block production events and the staking skimming and withdrawal events.

## Specification

1. When the user enters the view the history events will be queried for the current page and shown. User shoud be able to navigate those as these are the events in the DB.
2. Then the frontend will, like before query the EVM transactions but with only_cache=False. This is only to refresh the transactions and get the latest ones.
3. Once that is done it should call the endpoint that decodes undecoded EVM transactions. Once decoding is done the page should refresh. Ideally we should do this decoding once each address has finished querying.

The history events query will happen in 2 steps.
First step is the page creation query. The backend here will use the `GROUP BY event_identifier` and `COUNT(*)`trick I mentioned sometime in discord to get the events paginated based on their event identifier. This will group events that belong to the same transaction (and for the future other events).

This first query will only return 1 event per event identifier though, which is why we also will be returning the `COUNT(*)` of the query. 

If it's > 1 (probably endpoint should be like `query_more` or something -- unless frontend would benefit from knowing the amount of extra events) then frontend will perform a query for all events with that `event_identifier` to get all events and then group them together.

lukicenturi:
Closes #5846
## Checklist


kelsos (commented):
So code comments not many.

I am trying to remove the existing in the big account and try it but so far I have been unlucky and I only managed to get the backend blocking and not responding.

frontend/app/src/components/history/transactions/TransactionContent.vue
```diff
@@ -471,21 +473,29 @@ const addTransactionHash = () => {
               <table-filter
                 :matches="filters"
                 :matchers="matchers"
-                :location="SavedFilterLocation.HISTORY_TRANSACTIONS"
+                :location="SavedFilterLocation.HISTORY_EVENTS"
```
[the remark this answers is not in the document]
lukicenturi: I change only the enum, but keep the value of the enum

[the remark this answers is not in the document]
kelsos: > I change only the enum, but keep the value of the enum, so it's fine. And also this feature hasn't released yet (just introduced in this release, so should I change the value of the enum too? 🧐)

yes if it is only develop then we can move it I guess.

frontend/app/src/components/history/transactions/events/TransactionEvents.vue
```diff
@@ -68,31 +67,60 @@ const headers: DataTableHeader[] = [
...
+const events: Ref<HistoryEventEntry[]> = ref([get(eventGroupHeader)]);
+
+const fetchEvents = async () => {
+  set(evaluating, true);
+  const { data } = await fetchHistoryEvents({
+    limit: -1,
+    offset: 0,
+    eventIdentifiers: get(eventGroupHeader).eventIdentifier,
+    groupByEventIds: false
+  });
```
kelsos, line 87: What does this do? get all the events for the identifier?

frontend/app/src/composables/history/events/index.ts
```diff
@@ -144,32 +154,80 @@ export const useTransactions = () => {
...
+      ]);
+
+      const taskMeta = {
+        title: t('actions.transactions_events.task.title').toString(),
+        description: tc(
+          'actions.transactions_events.task.description',
+          2,
+          account
+        ),
+        numericKeys: [],
```
kelsos, line 183: do we still use numeric keys anywhere?

frontend/app/src/components/history/transactions/events/TransactionEvents.vue
lukicenturi: yes all events on each event group header

frontend/app/src/composables/history/events/index.ts
lukicenturi: no I think, this one is leftover, will remove

frontend/app/src/components/history/transactions/events/TransactionEventAsset.vue
```diff
@@ -5,12 +5,12 @@ import {
   TransactionEventType
 } from '@rotki/common/lib/history/tx-events';
 import { getEventType } from '@/utils/history';
-import { type EthTransactionEventEntry } from '@/types/history/tx';
+import { type HistoryEventEntry } from '@/types/history/tx';
```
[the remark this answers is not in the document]
lukicenturi: will move it in next PR, since this will cause a lot of conflicts if I do it now

Issue #5611, opened 2023-02-15
Convert etherscan api key missing warning to separate websocket message


LefterisJP:
## Abstract

At the moment we the etherscan key missing as a legacy warning which is easily missed by the user. If the user does not see it or does not take action on it then they are going to have a terribly slow experience in rotki as we do depend on etherscan for querying transactions.

Multiple times, users have come in rotki and said that it's too slow querying only to find out they never added an etherscan key.

## Motivation

Improve user experience.

### Backend
- [x] Create a new websocket message type as per the specification above
- [x] Replace the etherscan and optimistic etherscan api keys missing messages with usage of the new message

### Frontend
- [ ] In the frontend catch this message and direct the user to take action with as much hand-holding as possible. Perhaps in a menu with steps?
    - Step 1: Point them to the proper etherscan signup page to create the account and get the api key. Include link in our docs where we explain it?
    - Tell them when they have the key to press "Next" ...
    - Take them to the place where it should be input (or perhaps have it input from the same dialog?)

yabirgb:
Issue completed in https://github.com/rotki/rotki/pull/5897

PR #5889, opened 2023-04-12
Move etherscan missing api key to websockets

nebolax:
Backend aspect of #5611

## Checklist

### DIFF
--- a/docs/websockets.rst
+++ b/docs/websockets.rst
@@ -181,3 +181,23 @@ This also contains two optional, mutually excluse keys. If one exists the other 
 
 - ``seen_tx_hash``: A transaction hash in the same chain as the token in which the token was first seen.
 - ``seen_description``: A description of the action in which the token was first seen and added to the DB. For example, querying curve pools, querying yearn pools etc.
+
+
+Missing API Key
+=======================
+
+Having API keys for some services (such as etherscan) is crucial for rotki to work. So sometimes we might specifically prompt the user to add an API key for a service. In that case we send a websocket message to the frontend.
+
+::
+
+    {
+        "type": "missing_api_key",
+        "data": {
+            "service": "etherscan",
+            "location": "optimism"
+        }
+    }
+
+
+- ``service``: Service for which an API key is needed.
+- ``location``: For services like etherscan that require different API keys for different locations this is the location for which the API key is needed. If a service does not require a location then this key will be missing.
--- a/rotkehlchen/externalapis/etherscan.py
+++ b/rotkehlchen/externalapis/etherscan.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING, Any, Literal, Optional, Union, overload
 
 import gevent
 import requests
+from rotkehlchen.api.websockets.typedefs import WSMessageType
 
 from rotkehlchen.chain.evm.constants import GENESIS_HASH, ZERO_ADDRESS
 from rotkehlchen.chain.structures import TimestampOrBlockRange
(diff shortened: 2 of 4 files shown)


docs/websockets.rst
```diff
@@ -181,3 +181,23 @@ This also contains two optional, mutually excluse keys. If one exists the other
...
+
+Having API keys for some services (such as etherscan) is crucial for rotki to work. So sometimes we might specifically prompt the user to add an API key for a service. In that case we send a websocket message to the frontend.
+
+::
+
+    {
+        "type": "missing_api_key",
+        "data": {
+            "service": "etherscan",
+            "blockchain": "optimism"
```
yabirgb, line 197: We discussed in discord and the frontend doesn't store a enum like we do for `ExternalService` so for them is better to have this split in two fields

docs/websockets.rst
```diff
@@ -181,3 +181,23 @@ This also contains two optional, mutually excluse keys. If one exists the other
...
+        }
+    }
+
+
+- ``service``: Service for which an API key is needed.
+- ``blockchain``: For services like etherscan that require different API keys for different blockchains this is the blockchain for which the API key is needed.
```
yabirgb, line 203: For other services like beaconchain or opensea what would be the default value? I guess is missing what we do. Can you state it also in the docs?

rotkehlchen/externalapis/etherscan.py
```diff
@@ -187,12 +188,9 @@ def _query(
...
             if not self.warning_given:
-                self.msg_aggregator.add_warning(
-                    f'You do not have an {self.chain} Etherscan API key configured. rotki '
-                    f'etherscan queries will still work but will be very slow. '
-                    f'If you are not using your own ethereum node, it is recommended '
-                    f'to go to https://{self.base_url}/register, create an API '
-                    f'key and then input it in the external service credentials setting of rotki',
+                self.msg_aggregator.add_message(
+                    message_type=WSMessageType.MISSING_API_KEY,
+                    data={'service': 'etherscan', 'blockchain': self.chain.serialize()},
```
yabirgb, line 193: not 100% sure if is fine to have `etherscan` hardcoded here or is better to use `ExternalLocation.ETHERSCAN`  along the blockcahin field. I would do that since in the other places like beacon chain and opensea I believe is good to use the same constant. What do you think?

yabirgb (commented):
request + question

rotkehlchen/externalapis/etherscan.py
nebolax: Can be like that, yes

yabirgb (approved):
LGTM

yabirgb:
We have discussed the spec with the frontend and this is what better fits them

docs/websockets.rst
```diff
@@ -181,3 +181,23 @@ This also contains two optional, mutually excluse keys. If one exists the other
...
+
+Having API keys for some services (such as etherscan) is crucial for rotki to work. So sometimes we might specifically prompt the user to add an API key for a service. In that case we send a websocket message to the frontend.
+
+::
+
+    {
+        "type": "missing_api_key",
+        "data": {
+            "service": "etherscan",
+            "location": "optimism"
```
LefterisJP, line 197: Don't we also need the url of where to sign up for the key? Otherwise the frontend will need to maintain a mapping of location to URL

LefterisJP (approved):
As discussed in the standup we don't add the url in this endpoint since we need the url in the view even if that notification is not sent

PR #5895, opened 2023-04-13
Keep old passphrase if new credentials are invalid

nebolax:
This PR does 2 things:
1. Previously when editing an exchange there was a remote call (to validate new api credentials) under an open write cursor (which is bad because it blocks write access for a long time). This PR fixes it.
2. Previously when editing an exchange if there was a change in passphrase and new api credentials were incorrect then an inconsistency occurred. In database old passphrase was kept (which is what has to happen), but in the exchange object new (invalid) passphrase was set. This PR fixes it.

### DIFF
--- a/rotkehlchen/exchanges/exchange.py
+++ b/rotkehlchen/exchanges/exchange.py
@@ -19,6 +19,7 @@ from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.types import (
     ApiKey,
     ApiSecret,
+    ExchangeAuthCredentials,
     ExchangeLocationID,
     Location,
     T_ApiKey,
--- a/rotkehlchen/exchanges/exchange.py
+++ b/rotkehlchen/exchanges/exchange.py
@@ -69,27 +69,45 @@ class ExchangeInterface(CacheableMixIn, LockableQueryMixIn):
         set_user_agent(self.session)
         log.info(f'Initialized {str(location)} exchange {name}')
 
+    def reset_to_db_credentials(self) -> None:
+        """Resets the exchange credentials to the ones saved in the DB"""
+        with self.db.conn.read_ctx() as cursor:
+            credentials_in_db = self.db.get_exchange_credentials(
+                cursor=cursor,
+                location=self.location,
+                name=self.name,
+            )
+        assert self.location in credentials_in_db and len(credentials_in_db[self.location]) == 1
+        credentials = credentials_in_db[self.location][0]
+        ftx_subaccount = self.db.get_ftx_subaccount(self.name) if self.location in (Location.FTX, Location.FTXUS) else None  # noqa: E501
+        self.edit_exchange_credentials(ExchangeAuthCredentials(
+            api_key=credentials.api_key,
+            api_secret=credentials.api_secret,
+            passphrase=credentials.passphrase,
+            ftx_subaccount=ftx_subaccount,
+        ))
+
+    def reset_to_db_extras(self) -> None:
+        """Resets the exchange extras to the ones saved in the DB"""
+        extras = self.db.get_exchange_credentials_extras(location=self.location, name=self.name)
+        self.edit_exchange_extras(extras)
+
     def location_id(self) -> ExchangeLocationID:
         """Returns unique location identifier for this exchange object (name + location)"""
         return ExchangeLocationID(name=self.name, location=self.location)
 
-    def edit_exchange_credentials(
-            self,
-            api_key: Optional[ApiKey],
-            api_secret: Optional[ApiSecret],
-            passphrase: Optional[str],
-    ) -> bool:
+    def edit_exchange_credentials(self, credentials: ExchangeAuthCredentials) -> bool:
         """Edits the exchange object with new credentials given from the API
         Returns true if an edit happened and false otherwise.
 
         Needs to be implemented for each subclass
         """
-        if api_key is not None:
-            self.api_key = api_key
-        if api_secret is not None:
-            self.secret = api_secret
+        if credentials.api_key is not None:
+            self.api_key = credentials.api_key
+        if credentials.api_secret is not None:
+            self.secret = credentials.api_secret
 
-        return api_key is not None or api_secret is not None or passphrase is not None
+        return credentials.api_key is not None or credentials.api_secret is not None or credentials.passphrase is not None  # noqa: E501
 
     def edit_exchange(
             self,
@@ -136,6 +154,8 @@ class ExchangeInterface(CacheableMixIn, LockableQueryMixIn):
         if name is not None:
             self.name = name
 
+    def edit_exchange_extras(self, extras: dict) -> tuple[bool, str]:  # pylint: disable=unused-argument  # noqa: E501
+        """Subclasses may implement this method to accept extra properties"""
         return True, ''
 
     def query_balances(self, **kwargs: Any) -> ExchangeQueryBalances:
--- a/rotkehlchen/exchanges/exchange.py
+++ b/rotkehlchen/exchanges/exchange.py
@@ -109,51 +109,6 @@ class ExchangeInterface(CacheableMixIn, LockableQueryMixIn):
 
         return credentials.api_key is not None or credentials.api_secret is not None or credentials.passphrase is not None  # noqa: E501
 
-    def edit_exchange(
-            self,
-            name: Optional[str],
-            api_key: Optional[ApiKey],
-            api_secret: Optional[ApiSecret],
-            **kwargs: Any,
-    ) -> tuple[bool, str]:
-        """Edits the exchange object with new info given from the API
-
-        Returns False and error message in case of problems
-
-        If extra exchange info should be edited this needs to also be implemented by the subclass.
-        """
-        passphrase = kwargs.get('passphrase')
-        old_passphrase = None
-        if passphrase is not None:  # backup old passphrase
-            with self.db.conn.read_ctx() as cursor:
-                mapping = self.db.get_exchange_credentials(cursor, name=self.name, location=self.location)  # noqa: E501
-            credentials = mapping.get(self.location)
-            if not credentials or len(credentials) == 0 or credentials[0].passphrase is None:
-                old_passphrase = None  # should not happen, unless passphrase is optional
-                log.warning(
-                    f'When updating the passphrase for {self.name} {str(self.location)} '
-                    f'exchange, could not find an old passphrase to restore.',
-                )
-            else:
-                old_passphrase = credentials[0].passphrase
-
-        old_api_key = self.api_key
-        old_api_secret = self.secret
-        changed = self.edit_exchange_credentials(api_key, api_secret, passphrase)
-        if changed is True:
-            try:
-                success, message = self.validate_api_key()
-            except Exception as e:  # pylint: disable=broad-except
-                success = False
-                message = str(e)
-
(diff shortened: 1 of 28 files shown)


rotkehlchen/tests/unit/test_exchanges.py
```diff
@@ -54,3 +56,111 @@ def test_exchanges_filtering(database, exchange_manager, function_scope_messages
...
+    api_secret=ApiSecret(b'api-secret-3'),
+    passphrase='passphrase-3',
+)
+
+
+def test_change_credentials(rotkehlchen_api_server) -> None:
+    """
+    Test that chaning exchange credentials works as expected and if incorrect credentials
+    were passed then the old credentials are restored.
+    """
```
yabirgb, line 90: Does this code fail before the change?

rotkehlchen/tests/unit/test_exchanges.py
```diff
@@ -54,3 +56,111 @@ def test_exchanges_filtering(database, exchange_manager, function_scope_messages
...
+    api_key=ApiKey('api-key-3'),
+    api_secret=ApiSecret(b'api-secret-3'),
+    passphrase='passphrase-3',
+)
+
+
+def test_change_credentials(rotkehlchen_api_server) -> None:
+    """
+    Test that chaning exchange credentials works as expected and if incorrect credentials
+    were passed then the old credentials are restored.
```
yabirgb, line 89:
```suggestion
    were provided then the old credentials are restored.
```

yabirgb (commented):
I'm not sure about the approach taken in this PR. The logic in `edit_exchange` should handle what you are trying to fix but is not done in the case of an exception. Not sure how you reproduced this error but would be nice to confirm why didn't work for you

rotkehlchen/exchanges/manager.py
```diff
@@ -88,38 +88,57 @@ def edit_exchange(
...
-                api_key=api_key,
-                api_secret=api_secret,
-                passphrase=passphrase,
-                kraken_account_type=kraken_account_type,
-                binance_selected_trade_pairs=binance_selected_trade_pairs,
-                ftx_subaccount=ftx_subaccount,
+        old_credentials = old_db_credentials[location][0]
+        success, msg = exchangeobj.edit_exchange(  # Edit the exchange object
+            name=new_name,
+            api_key=api_key,
```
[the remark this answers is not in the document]
nebolax: > Way not edit the code of edit_exchange at ... to restore credentials?

It already works this way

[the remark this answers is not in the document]
nebolax: > Can you add the comment mentioning that this is done like this to validate the keys?

To what do you refer by "this is done like this" ? There is no special step that only validates the keys. The line where you left the comment on actually edits the exchange object

rotkehlchen/tests/unit/test_exchanges.py
nebolax: Talking about it in discord

nebolax: (and yes, it does :) )

rotkehlchen/exchanges/manager.py
```diff
@@ -82,58 +90,96 @@ def edit_exchange(
...
+                    )
+                return False, msg
 
-            return True, ''
+        return True, ''
 
-    def delete_exchange(self, name: str, location: Location) -> None:
+    def delete_exchange(self, name: str, location: Location) -> tuple[bool, str]:
         """Deletes exchange only from the manager. Not from the DB"""
-        exchanges_list = self.connected_exchanges.get(location)
```
yabirgb, line 126: docstring is outdated

rotkehlchen/exchanges/manager.py
```diff
@@ -82,58 +90,96 @@ def edit_exchange(
...
+                return False, f'{str(location)} exchange {name} is not registered'
+
+            if len(exchanges_list) == 1:
+                self.connected_exchanges.pop(location)
+            else:
+                self.connected_exchanges[location] = [x for x in exchanges_list if x.name != name]
+
+            with self.database.user_write() as write_cursor:  # Also remove it from the db
+                self.database.remove_exchange(write_cursor=write_cursor, name=name, location=location)  # noqa: E501
+                if self.connected_exchanges.get(location) is None:
```
yabirgb, line 176:
```suggestion
            last_in_location = len(exchanges_list) == 1
            if last_in_location is True:
                self.connected_exchanges.pop(location)
            else:
                self.connected_exchanges[location] = [x for x in exchanges_list if x.name != name]

            with self.database.user_write() as write_cursor:  # Also remove it from the db
                self.database.remove_exchange(write_cursor=write_cursor, name=name, location=location)  # noqa: E501
                if last_in_location is True:
```

yabirgb (commented):
Just two small comments. I've checked the code and seems that there are no dead locks and tried the app to verify that it works fine.

rotkehlchen/db/dbhandler.py
```diff
@@ -3330,3 +3330,28 @@ def get_blockchain_account_label(self, chain_address: ChainAddress) -> Optional[
...
         return None if result is None else result[0]
+
+    def get_all_exchange_settings(
+            self,
+            name: str,
+            location: Location,
+    ) -> Optional[tuple[ExchangeApiCredentials, dict[str, Any]]]:
+        """
+        For a particular exchange returns its credentials and extra settings.
+        If can't find the credentials, None is returned
```
LefterisJP, line 3341:
```suggestion
        If it can't find the credentials, None is returned
```

LefterisJP (commented):
Okay so I don't understand why this PR makes so many changes to fix what you said in (2).

1. Should be super small and easy to adjust.

As for (2) you have made the approach very complicated. I see restoring DB to previous values and other weird approaches, adding lots of complicated code. This can't be the right approach to solve this sorry.

The DB should be written once. After everytihng succeeds. Not before. And generally this PR looks complicated and big for what it's supposed to solve.

rotkehlchen/exchanges/manager.py
```diff
@@ -82,58 +88,88 @@ def edit_exchange(
...
+
+            if credentials_are_valid is False:
+                exchangeobj.reset_to_db_credentials()
+                return False, f'New credentials are invalid. {msg}'
+
+        # Then edit extra properties
+        success, msg = exchangeobj.edit_exchange_extras({
+            KRAKEN_ACCOUNT_TYPE_KEY: kraken_account_type,
+            BINANCE_MARKETS_KEY: binance_selected_trade_pairs,
+        })
```
yabirgb, line 118: this shouldn't execute for all the exchanges, only for those that require modification of the extra data.

Also this is the old design too but if feels broken. Now is executed always because you don't know what exchanges should read from this properties (in this case kraken, binance and binance US) but it breaks the abstraction somehow and not the manager needs to know about the exchanges.

I would not touch anything and leave it as it is now but I don't like the design here

rotkehlchen/exchanges/manager.py
```diff
@@ -82,58 +88,88 @@ def edit_exchange(
...
             self.connected_exchanges.pop(location)
-            return
-
-        self.connected_exchanges[location] = [x for x in exchanges_list if x.name != name]
+        else:
+            self.connected_exchanges[location] = [x for x in exchanges_list if x.name != name]
+        with self.database.user_write() as write_cursor:  # Also remove it from the db
+            self.database.remove_exchange(write_cursor=write_cursor, name=name, location=location)  # noqa: E501
+            if last_in_location is True:
+                self.database.delete_used_query_range_for_exchange(write_cursor=write_cursor, location=location)  # noqa: E501
```
yabirgb, line 168: I'm not sure why we do it like this. When deleting an exchange seems that we don't remove the used query ranges for that exchange until all the exchanges for that location have been removed. Maybe we can improve this in another PR? Take note of this please

rotkehlchen/types.py
```diff
@@ -603,6 +603,13 @@ class AssetMovementCategory(DBEnumMixIn):
     WITHDRAWAL = 2
 
 
+class ExchangeAuthCredentials(NamedTuple):
+    api_key: Optional[ApiKey]
```
yabirgb, line 607: can you add a doctring explaining where this used and if any of the fields need to be populated or can all be none

rotkehlchen/exchanges/manager.py
```diff
@@ -82,58 +88,88 @@ def edit_exchange(
...
+                    api_secret=api_secret,
+                    passphrase=passphrase,
+                    kraken_account_type=kraken_account_type,
+                    binance_selected_trade_pairs=binance_selected_trade_pairs,
+                    ftx_subaccount=ftx_subaccount,
+                )
+        except InputError as e:
+            exchangeobj.reset_to_db_credentials()  # DB is already rolled back at this point
+            exchangeobj.reset_to_db_extras()
+            return False, f'Coudnt update exchange properties in the DB. {str(e)}'
```
yabirgb, line 140: There is a typo but also 

```suggestion
            return False, f"Couldn't update exchange properties in the DB. {str(e)}"
```

yabirgb (commented):
A few things. Also seems that a relevant test failed

rotkehlchen/exchanges/independentreserve.py
```diff
@@ -247,15 +248,10 @@ def __init__(
...
-            api_secret: Optional[ApiSecret],
-            passphrase: Optional[str],
-    ) -> bool:
-        changed = super().edit_exchange_credentials(api_key, api_secret, passphrase)
-        if api_key is not None:
-            self.session.headers.update({'x-api-key': api_key})
+    def edit_exchange_credentials(self, credentials: ExchangeAuthCredentials) -> bool:
+        changed = super().edit_exchange_credentials(credentials)
+        if credentials.api_key is not None:
+            self.session.headers.update({'x-api-key': credentials.api_key})
```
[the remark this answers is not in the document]
nebolax: Not sure that I got the question, but if `credentials.api_key` will be None, then current `x-api-key` will stay untouched, which sounds correct.

rotkehlchen/exchanges/exchange.py
```diff
@@ -69,73 +70,52 @@ def __init__(
...
-
-            if success is False:
-                self.edit_exchange_credentials(old_api_key, old_api_secret, old_passphrase)
-                return False, message
-
-        if name is not None:
-            self.name = name
+    def edit_exchange_name(self, name: str) -> None:
+        """Edits the exchange object with new name"""
+        self.name = name
```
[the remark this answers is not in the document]
nebolax: Hmm looks a bit less clean to me, but okay

LefterisJP (commented):
Some small notes

rotkehlchen/exchanges/exchange.py
```diff
@@ -69,73 +70,48 @@ def __init__(
...
 
+    def reset_to_db_credentials(self) -> None:
+        """Resets the exchange credentials to the ones saved in the DB"""
+        with self.db.conn.read_ctx() as cursor:
+            credentials_in_db = self.db.get_exchange_credentials(
+                cursor=cursor,
+                location=self.location,
+                name=self.name,
+            )
+        assert self.location in credentials_in_db and len(credentials_in_db[self.location]) == 1
```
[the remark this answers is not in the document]
nebolax: Not really needed. Will remove

LefterisJP (approved):
LGTM.

PR #5897, opened 2023-04-14
Handle missing api key notifications

tewshi:
## Checklist

- [x] The PR modified the frontend, and updated the notifications component to handle missing api key notifications
- [x] use css module composable in components within the folder
- [x] prioritises notifications by date, then by notifications with actions, so actionable notifications stay on top, then legacy notifications have lowest priority, they go down the list.

### DIFF
--- a/frontend/app/src/components/status/notifications/MissingKeyNotification.vue
+++ b/frontend/app/src/components/status/notifications/MissingKeyNotification.vue
@@ -1,0 +1,21 @@
+<script setup lang="ts">
+import { type I18nParam } from '@rotki/common/lib/messages';
+
+defineProps<{
+  params: I18nParam;
+}>();
+</script>
+
+<template>
+  <i18n :path="params.message" :plural="params.choice" tag="div">
+    <template #service>
+      {{ params.props.service }}
+    </template>
+    <template #location>
+      {{ params.props.location }}
+    </template>
+    <template #url>
+      <base-external-link :href="params.props.url" :text="params.props.url" />
+    </template>
+  </i18n>
+</template>
--- a/frontend/app/src/components/status/notifications/Notification.vue
+++ b/frontend/app/src/components/status/notifications/Notification.vue
@@ -10,9 +10,11 @@ const props = withDefaults(
   { popup: false }
 );
 
-const emit = defineEmits(['dismiss']);
+const emit = defineEmits<{ (e: 'dismiss', id: number): void }>();
 
+const css = useCssModule();
 const { t } = useI18n();
+const { copy: copyToClipboard } = useClipboard();
 
 const { notification } = toRefs(props);
 const dismiss = (id: number) => {
@@ -45,7 +47,17 @@ const color = computed(() => {
 const date = computed(() => dayjs(get(notification).date).format('LLL'));
 
 const copy = async () => {
-  await navigator.clipboard.writeText(get(notification).message);
+  const { message, i18nParam } = get(notification);
+  let messageText = message;
+
+  if (i18nParam) {
+    messageText = t(i18nParam.message, {
+      service: i18nParam.props.service,
+      location: i18nParam.props.location,
+      url: i18nParam.props.url
+    }).toString();
+  }
+  await copyToClipboard(messageText);
 };
 
 const { fontStyle } = useTheme();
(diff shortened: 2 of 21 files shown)


frontend/common/src/messages/index.ts
```diff
@@ -1,3 +1,4 @@
+import { type SocketMessageType } from '@/types/websocket-messages';
```
kelsos, line 1: ok, so this is the reason for the failure this type is not actually available in the common library since it comes from the app itself.

You must move the type from the app to the common if you want to use it.

tewshi: alrighty, checking... thanks.

tewshi: it'll take a lot to move the definition of SocketMessageType to the common lib

frontend/app/src/composables/message-handling.ts
```diff
@@ -77,7 +83,8 @@ export const useMessageHandling = () => {
               'notification_messages.premium.inactive.network_problem_message'
             ),
         display: true,
-        severity: Severity.ERROR
+        severity: Severity.ERROR,
+        type: SocketMessageType.PREMIUM_STATUS_UPDATE
```
kelsos, line 87: Question, why not make a priority enum instead of exposing the type? 

it could be something like: 

```typescript
enum Priority {
  BULK,
  NORMAL,
  HIGH,
  ACTION
}
```

And since it is an enum we can just order by the value then?

frontend/app/src/components/status/notifications/Notification.vue
```diff
@@ -96,9 +112,33 @@ const action = async (notification: NotificationData) => {
...
+            </template>
+            <template #location>
+              {{ notification.i18nParam.props.location }}
+            </template>
+            <template #url>
+              <a :href="notification.i18nParam.props.url" target="_blank">{{
+                notification.i18nParam.props.url
+              }}</a>
+            </template>
+          </i18n>
```
kelsos, line 137: Do you think we can move it to a small subcomponent?

kelsos, line 137: Btw wouldn't it be more clear semantically if we kept `message` only for text messages and used a key property in `i18n.param` for the key?

kelsos (changes requested):
So along with the comments above, it would be nice to add some test for the prioritization of messages.

Also a small request would be to capitalize Ethereum Optimism in the title

frontend/app/src/components/status/notifications/Notification.vue
```diff
@@ -44,7 +46,21 @@ const color = computed(() => {
...
 const date = computed(() => dayjs(get(notification).date).format('LLL'));
 
+const isMissingKeyNotification = computed(
+  () => get(notification).type === SocketMessageType.MISSING_API_KEY
+);
+
 const copy = async () => {
+  if (get(isMissingKeyNotification)) {
+    const { message, i18nParam } = get(notification);
+    return await navigator.clipboard.writeText(
```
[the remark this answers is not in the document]
tewshi: cool, will do.

frontend/app/src/components/status/notifications/Notification.vue
tewshi: yeah, you're right

tewshi: sure, will do

frontend/app/src/composables/message-handling.ts
tewshi: done

frontend/app/src/components/status/notifications/Notification.vue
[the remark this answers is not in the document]
tewshi: done

frontend/app/src/components/status/notifications/Notification.vue
tewshi: done

tewshi: done

tewshi:
> So along with the comments above, it would be nice to add some test for the prioritization of messages.
> 
> Also a small request would be to capitalize Ethereum Optimism in the title

done, writing test

frontend/app/src/components/status/notifications/MissingKeyNotification.vue
```diff
@@ -0,0 +1,21 @@
...
+<template>
+  <i18n :path="params.message" :plural="params.choice" tag="div">
+    <template #service>
+      {{ params.props.service }}
+    </template>
+    <template #location>
+      {{ params.props.location }}
+    </template>
+    <template #url>
+      <base-external-link :href="params.props.url" :text="params.props.url" />
```
kelsos, line 18: @tewshi I changed this here to use the component that is designed to handle electron too

frontend/app/src/components/status/notifications/Notification.vue
```diff
@@ -45,7 +47,17 @@ const color = computed(() => {
...
+  let messageText = message;
+
+  if (i18nParam) {
+    messageText = t(i18nParam.message, {
+      service: i18nParam.props.service,
+      location: i18nParam.props.location,
+      url: i18nParam.props.url
+    }).toString();
+  }
+  await copyToClipboard(messageText);
```
kelsos, line 60: And also did a small change here

kelsos (approved):
Looks great @tewshi thank you very much :)

PR #5900, opened 2023-04-15
Some upgrades

LefterisJP:


LefterisJP:
The failed test is the defilamma orale due to moving price

PR #5903, opened 2023-04-17
ethereum withdrawals - WIP

LefterisJP:


LefterisJP:
`rotkehlchen/tests/unit/test_ethereum_airdrops.py::test_check_airdrops` is the only failing test and it passes locally. The failure is connection timeout when fetching a gist.

LefterisJP:
There is probably more stuff to do but will have to see as tests progress and test with real data. All green means I am merging for now.

rotkehlchen/chain/ethereum/modules/eth2/eth2.py
```diff
@@ -320,6 +327,67 @@ def _query_services_for_validator_daily_stats(
...
+            now = ts_now()
+            self.validator_withdrawals_queried += 1
+            self.last_withdrawals_query_ts = now
+            for entry in new_data:
+                withdrawal = EthWithdrawalEvent(
+                    validator_index=validator_index,
+                    timestamp=ts_sec_to_ms(entry[0]),
+                    balance=Balance(amount=entry[2]),
+                    withdrawal_address=entry[1],
+                    is_exit=entry[0] >= exit_epoch[validator_index],
```
yabirgb, line 375: Should we prevent against index error here? It can happen if beaconchain returns bad data for some reason I guess

rotkehlchen/db/eth2.py
```diff
@@ -19,12 +19,37 @@
...
+            LEFT JOIN history_events E ON
+            E.identifier = S.identifier WHERE ? - (SELECT MAX(timestamp) FROM history_events WHERE identifier=S.identifier) > ? AND S.validator_index IS NOT NULL
+            UNION
+            SELECT DISTINCT V2.validator_index, 0 FROM eth2_validators V2 WHERE
+            V2.validator_index NOT IN (SELECT validator_index FROM eth_staking_events_info)
+        """  # noqa: E501
+        cursor = self.db.conn.cursor()
+        result = cursor.execute(
+            query_str, (up_to_tsms, WITHDRAWALS_RECHECK_PERIOD),
+        )
```
yabirgb, line 50: This fits in one line

rotkehlchen/chain/ethereum/modules/eth2/eth2.py
LefterisJP: That's not possible. This is already done in the function that returns entry

rotkehlchen/db/eth2.py
LefterisJP: Heh ... this is what you see? Check the line above.

PR #5907, opened 2023-04-18
Change liquity staking queries

yabirgb:
To improve consistency in the API and performance in the frontend we agreed to change the returned value for the staking endpoints in liquity to use `null` instead of `{}` for the empty case.

This PR:

- Makes nullable the API response for staked amounts in liquity
- Makes a little refactor in the aggregator to avoid duplicated code

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/liquity/trove.py
+++ b/rotkehlchen/chain/ethereum/modules/liquity/trove.py
@@ -1,14 +1,15 @@
 import logging
 from collections import defaultdict
-from typing import TYPE_CHECKING, Any, NamedTuple, Optional, TypedDict
+from typing import TYPE_CHECKING, Any, NamedTuple, Optional, TypedDict, cast
 
 from gevent.lock import Semaphore
 
-from rotkehlchen.accounting.structures.balance import AssetBalance, Balance
+from rotkehlchen.accounting.structures.balance import AssetBalance, Balance, BalanceSheet
 from rotkehlchen.chain.ethereum.defi.defisaver_proxy import HasDSProxy
 from rotkehlchen.chain.ethereum.utils import token_normalized_value_decimals
 from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.constants.assets import A_ETH, A_LQTY, A_LUSD
+from rotkehlchen.constants.misc import ZERO
 from rotkehlchen.errors.misc import BlockchainQueryError, RemoteError
 from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.fval import FVal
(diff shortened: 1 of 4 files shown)


rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -282,3 +283,39 @@ def liquity_staking_balances(
...
+            balances=balances,
+            new_balances=liquity_staked,
+            token=A_LQTY,
+        )
+        lusd_stability_pool = self.get_stability_pool_balances(addresses=queried_addresses)
+        self._add_addr_and_proxy_balances(
+            balances=balances,
+            new_balances=lusd_stability_pool,
+            token=A_LUSD,
+        )
```
nebolax: Seems like it's missing a newline?

rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -233,15 +232,25 @@ def _query_deposits_and_rewards(
             )
             proxy_owner = self.ethereum.proxies_inquirer.proxy_to_address.get(current_address)
             if proxy_owner is not None:
-                data[proxy_owner]['proxies'][current_address][key] = AssetBalance(
+                if data[proxy_owner]['proxies'] is None:
+                    # mypy fails to detect that the defaultdict is returning the
+                    # correct type for the not None case
+                    data[proxy_owner]['proxies'] = defaultdict(lambda: defaultdict(AssetBalance))  # type: ignore[arg-type]  # noqa: E501
```
nebolax, line 238: The problem here and below is not the None case, but that `defaultdict` needs a callable that doesn't accept any arguments, and here `AssetBalance` is given, which requires `asset` and `balance` arguments to be initialized.

I think that right now it is broken since if at some place there will be access to `data[proxy_owner]['proxies'][some_address].some_property_of_AssetBalance` and `data[proxy_owner]['proxies'][some_address]` won't be defined, it will result in an error.

rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -284,38 +293,45 @@ def liquity_staking_balances(
             assets=(A_LQTY, A_LUSD, A_ETH),
         )
 
+    @staticmethod
```
nebolax, line 296: Why there was no issue before if it was defined as an instance method and there was no `self` in the list of arguments?

rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -284,38 +293,45 @@ def liquity_staking_balances(
...
         )
         lusd_stability_pool = self.get_stability_pool_balances(addresses=queried_addresses)
         self._add_addr_and_proxy_balances(
             balances=balances,
             new_balances=lusd_stability_pool,
             token=A_LUSD,
-        )
\ No newline at end of file
+            key='deposited',
+        )
```
nebolax, line 337: Ah, okay, the newline was added in the other commit :stuck_out_tongue:

nebolax (commented):
Is the test failure related? Seem like it can be

yabirgb:
Closing since the CI has broken in this PR

PR #5909, opened 2023-04-18
Add locations endpoint and mappings for kraken

yabirgb:
This PR:

- Adds a new default mapping: STAKING + FEE that can appear in kraken
- adds a new endpoint to report to the frontend the possible valid locations. I tough avoid making a db query and get the list of used locations but then the frontend would need to be constantly querying it so opted for this approach where they are hardcoded


docs/api.rst
```diff
@@ -12459,3 +12459,37 @@ Getting all available counterparties
 
   :statuscode 200: Information was correctly generated
   :statuscode 500: Internal rotki error
+
+Getting valid locations for history events
```
LefterisJP, line 12463: What is this endpoint? What do you mean by valid location for history events? There is no such thing. Please **ask me** or read the code before proceeding to make PRs.

I already told you in our call all locations are valid. There already is events from a lot of different locations imported by the generic CSV importer. We discussed this.

rotkehlchen/types.py
```diff
@@ -594,7 +594,8 @@ def to_chain_id(self) -> int:
 
 
 EVM_LOCATIONS_TYPE_ = Literal[Location.ETHEREUM, Location.OPTIMISM]
-EVM_LOCATIONS: tuple[EVM_LOCATIONS_TYPE_, ...] = typing.get_args(EVM_LOCATIONS_TYPE_)
+EVM_LOCATIONS: tuple[Location, ...] = typing.get_args(EVM_LOCATIONS_TYPE_)
+HISTORY_EVENTS_SUPPORTED_LOCATIONS = EVM_LOCATIONS + (Location.KRAKEN,)
```
LefterisJP, line 598: no ....

rotkehlchen/api/rest.py
```diff
@@ -4317,3 +4317,12 @@ def get_evm_counterparties_details(self) -> Response:
             )),
             status_code=HTTPStatus.OK,
         )
+
+    def get_history_event_locations(self) -> Response:
```
LefterisJP, line 4321: Remove this. This function makes no sense. Just do a process_result() location directly in resource and don't use the misleading "events" related name

LefterisJP (approved):
LGTM

PR #5911, opened 2023-04-18
Simplify and improve balance aggregation for liquity

yabirgb:
To improve consistency in the API and performance in the frontend we agreed to change the returned value for the staking endpoints in liquity to use `null` instead of `{}` for the empty case.

This PR:

- Makes nullable the API response for staked amounts in liquity
- Makes a little refactor in the aggregator to avoid duplicated code

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -7427,8 +7427,8 @@ Getting Liquity staked amount
           "message": ""
       }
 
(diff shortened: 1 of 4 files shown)


yabirgb:
Closing in favor of https://github.com/rotki/rotki/pull/5912 since github bugged and the branch doesn't detect properly the test-caching repo

PR #5917, opened 2023-04-20
feat: adjust liquity staking schema

lukicenturi:
Closes #(issue_number)

## Checklist


frontend/app/.eslintrc-auto-import.json
kelsos, line 1: I think a `pnpm install` is missing here

frontend/common/src/liquity/index.ts
```diff
@@ -35,8 +35,8 @@ export type LiquityStakingDetailProxyEntries = z.infer<
 >;
 
 export const LiquityStakingDetail = z.object({
-  balances: LiquityStakingDetailEntry,
-  proxies: LiquityStakingDetailProxyEntries
+  balances: LiquityStakingDetailEntry.nullish(),
+  proxies: LiquityStakingDetailProxyEntries.nullish()
```
kelsos, line 39: They should be `nullable` instead of `nullish` since they will either have a value or be null they won't be undefined.

kelsos (approved):

PR #5921, opened 2023-04-20
Handle ETH.S automatic conversions to ETH in Kraken

yabirgb:
ETH.S is a special asset in Kraken and rewards don't compound like the other staking assets do. What they seem to be doing is creating a virtual event that adds the ETH to your account and substract the ETH.S amount

image

This PR transforms those events into informational events so no information is lost in our side.

Note: The users might already have those events in their DB since they can be queried by 1.27.1. What we can do is delete all the kraken events or create a data migrations to fix the bad events


rotkehlchen/exchanges/kraken.py
```diff
@@ -153,7 +153,14 @@ def history_event_from_kraken(
                 else:
                     event_subtype = HistoryEventSubType.RECEIVE
             elif event_type == HistoryEventType.STAKING:
-                event_subtype = HistoryEventSubType.REWARD
+                # in the case of ETH.S after the merge rewards don't compound unlike what happens
+                # with other assets and a virtual event with negative amount is created
+                if asset == A_ETH2 and raw_amount < ZERO:
+                    event_type = HistoryEventType.INFORMATIONAL
+                    notes = 'Automatic virtual conversion of staked ETH rewards to ETH'
+                    raw_amount = AssetAmount(abs(raw_amount))
```
nebolax, line 161: nitpick: It could be `raw_amount = AssetAmount(-raw_amount)` to be a bit more clear

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -158,13 +161,41 @@ def _update_ens_mappings_schema(write_cursor: 'DBCursor') -> None:
...
+def _fix_kraken_eth2_events(write_cursor: 'DBCursor') -> None:
+    """Fix kraken events with negative amounts related to staking ETH after the merge"""
+    log.debug('Enter _fix_kraken_eth2_events')
+    write_cursor.execute(
+        'SELECT identifier, amount, usd_value FROM history_events WHERE location="B" AND '
+        'asset=? AND type="staking" AND subtype="reward" AND CAST(amount AS REAL) < 0',
+        (A_ETH2.identifier,),
+    )
+    update_tuples = []
+    for event_row in write_cursor:
```
nebolax, line 173: Not sure what's better, but this can also be done in a single sql query I guess

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1400,6 +1409,7 @@ def test_upgrade_db_36_to_37(user_data_dir):  # pylint: disable=unused-argument
         ('OPTIMISMtokentxs_0xc37b40ABdB939635068d3c5f13E7faF686F03B65', 0, 1679312760),
         ('ETHtokentxs_0xc37b40ABdB939635068d3c5f13E7faF686F03B65', 0, 1679312760),
     ]
+    cursor.execute('SELECT * FROM history_events WHERE identifier=?', (176,))
```
nebolax, line 1412: This looks like it's missing a check

rotkehlchen/db/upgrades/v36_v37.py
nebolax: The problematic point can be I guess in checking whether `amount` is less than zero since `REAL` in sqlite is not super-precise, but this can be checked with sqlites's method like amount.startswith('-')

nebolax: It will help in case there will be many kraken events to modify, but if you think my proposal is too ugly, let's keep the code as it is right now

yabirgb: I did went with something similar but the problem is that we have to change the sign in two column and this I believe can be an issue since I'm not sure how good it is and if we will loose pricesion. That is why I opted for this approach instead

rotkehlchen/exchanges/kraken.py
```diff
@@ -153,7 +153,14 @@ def history_event_from_kraken(
                 else:
                     event_subtype = HistoryEventSubType.RECEIVE
             elif event_type == HistoryEventType.STAKING:
-                event_subtype = HistoryEventSubType.REWARD
+                # in the case of ETH.S after the merge rewards don't compound unlike what happens
```
LefterisJP, line 156:
```suggestion
                # in the case of ETH.S after the activation of withdrawals rewards no longer compound unlike what happens
```

rotkehlchen/exchanges/kraken.py
```diff
@@ -153,7 +153,14 @@ def history_event_from_kraken(
...
+                # with other assets and a virtual event with negative amount is created
+                if asset == A_ETH2 and raw_amount < ZERO:
```
LefterisJP, line 158: what about events before withdrawals activation? Would we have any negative events then?

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -158,13 +161,42 @@ def _update_ens_mappings_schema(write_cursor: 'DBCursor') -> None:
...
+        update_tuples.append(
+            (
+                HistoryEventType.INFORMATIONAL.serialize(),
```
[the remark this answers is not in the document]
yabirgb: We are avoiding:

1. Negative amounts in the events that shouldn't exists
2. The type was marked as receive reward so it was taxable and it having negative amount I believe breaks the logic

rotkehlchen/exchanges/kraken.py
yabirgb: Sorry I wrote merge but is the withdrawals activation what triggered it. Will fix the comment. Before the withdrawals the events are correct

LefterisJP (approved):
LGTM

PR #5922, opened 2023-04-21
fix: kraken staking issues

lukicenturi:
Closes #(issue_number)

## Checklist


frontend/app/src/composables/filters/events.ts
```diff
@@ -33,7 +33,10 @@ export type Matcher = SearchMatcher<
 >;
 export type Filters = MatchedKeyword<HistoryEventFilterValueKeys>;
 
-export const useHistoryEventFilter = (disableProtocols: boolean) => {
+export const useHistoryEventFilter = (
+  disableProtocols: boolean,
+  disableLocations = false
```
kelsos, line 38:
```suggestion
  disabled: { protocols?: boolean, locations?: boolean } 
```

Can we maybe change the syntax to this? 

The check should look like `if(!disabled?.protocols)` etc

kelsos (commented):
I left you a couple of comments @lukicenturi

frontend/app/src/components/history/events/HistoryEventsView.vue
```diff
@@ -171,6 +178,10 @@ const {
 
       const accounts = get(usedAccounts);
 
+      if (isDefined(location)) {
```
kelsos, line 181: I already mentioned in the previous comment that for this to work we have to make the initial value for the prop undefined (or null) for this to work.

if the default value for location is `''` then this would always be true because `''` is a defined value.

lukicenturi: sorry my bad, should be updated now

PR #5923, opened 2023-04-21
Add details information to locations endpoint

yabirgb:
It was needed by the frontend to remove the mappings for locations in the frontend and that meant also to remove the hardocded images that they had.


rotkehlchen/types.py
```diff
@@ -597,6 +598,24 @@ def to_chain_id(self) -> int:
 EVM_LOCATIONS: tuple[EVM_LOCATIONS_TYPE_, ...] = typing.get_args(EVM_LOCATIONS_TYPE_)
 
 
+class LocationDetails(NamedTuple):
```
LefterisJP, line 601: Add a docstring here explaining what this structure is

rotkehlchen/types.py
```diff
@@ -597,6 +598,24 @@ def to_chain_id(self) -> int:
...
+    display_name: Optional[str] = None
+    icon: Optional[str] = None
+    image: Optional[str] = None
+
+    def serialize(self) -> dict[str, str]:
```
LefterisJP, line 606: you are not using asdict here only to have optional stuff not appear?

rotkehlchen/types.py
```diff
@@ -597,6 +598,24 @@ def to_chain_id(self) -> int:
...
+        data = {}
+        if self.display_name is not None:
+            data = {'display_name': self.display_name}
+
+        if self.icon is not None:
+            return data | {'icon': self.icon}
+        if self.image is not None:
+            return data | {'image': self.image}
+
+        raise InputError('Location details does not have either icon or image')
```
LefterisJP, line 616:
```suggestion
        raise InputError('Location details has neither an icon nor an image')
```

rotkehlchen/detail_mappings.py
```diff
@@ -0,0 +1,54 @@
+from rotkehlchen.types import Location, LocationDetails
```
LefterisJP, line 1: This should not be in the root of the project. Put it under `/constants`. Also name can be more specific. location_details.py

rotkehlchen/types.py
yabirgb: Yes, the logic using asdict would be similar but using pop with if checks for None

LefterisJP (approved):
LGTM

PR #5924, opened 2023-04-21
Allow to trigger a remote query of history events for exchanges

yabirgb:
Currently the only way to query kraken history events is by navigating to trades/deposits or do a report. This PR introduces a new endpoint and ws messages to allow querying kraken remote events and present them to the user

### DIFF
--- a/docs/websockets.rst
+++ b/docs/websockets.rst
@@ -201,3 +201,43 @@ Having API keys for some services (such as etherscan) is crucial for rotki to wo
 
 - ``service``: Service for which an API key is needed.
 - ``location``: For services like etherscan that require different API keys for different locations this is the location for which the API key is needed. If a service does not require a location then this key will be missing.
+
+
+Query new history events
+=========================
+
+In addition to history from evm transactions we need to query events from exchanges that support this kind of events. The backend will emit a ws message as the following when a query start for such exchange.
+
+::
+
+    {
+        "type": "querying_events_status",
+        "data": {
+            "status": "querying_events_started",
+            "location": "kraken",
+            "name": "My kraken exchange"
+        }
+    }
+
+
+- ``status``: Can be either `querying_events_started`, `querying_events_finished`, `querying_events_status_update`. Each pair of events is triggered per exchange instance.
+- ``location``: Exchange location triggering the event.
+- ``name``: Name of the exchange instance that is being queried.
+
+Finally the backend provides more granular information to know what interval of time is getting queried.
+
+::
+
+    {
+        "type": "querying_events_status",
+        "data": {
+            "status": "history_event_status",
+            "location": "kraken",
+            "name": "My kraken exchange",
+            'period': [0, 1682100570]
+        }
+    }
+
+- ``location``: Exchange location triggering the event.
+- ``name``: Name of the exchange instance that is being queried.
+- ``period``: Time range that is being queried.
--- a/docs/websockets.rst
+++ b/docs/websockets.rst
@@ -206,7 +206,7 @@ Having API keys for some services (such as etherscan) is crucial for rotki to wo
 Query new history events
 =========================
 
-In addition to history from evm transactions we need to query events from exchanges that support this kind of events. The backend will emit a ws message as the following when a query start for such exchange.
+In addition to history from evm transactions we need to query events from exchanges, eth staking events etc. The backend will emit a ws message as the following when such a query starts.
 
 ::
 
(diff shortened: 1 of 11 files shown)


docs/websockets.rst
```diff
@@ -201,3 +201,43 @@ Having API keys for some services (such as etherscan) is crucial for rotki to wo
 
 - ``service``: Service for which an API key is needed.
 - ``location``: For services like etherscan that require different API keys for different locations this is the location for which the API key is needed. If a service does not require a location then this key will be missing.
+
+
+Query new history events
+=========================
+
+In addition to history from evm transactions we need to query events from exchanges that support this kind of events. The backend will emit a ws message as the following when a query start for such exchange.
```
LefterisJP, line 209:
```suggestion
In addition to history from evm transactions we need to query events from exchanges, eth staking events etc. The backend will emit a ws message as the following when such a query starts.
```

docs/websockets.rst
```diff
@@ -201,3 +201,43 @@ Having API keys for some services (such as etherscan) is crucial for rotki to wo
...
+
+::
+
+    {
+        "type": "querying_events_status",
+        "data": {
+            "status": "querying_events_started",
+            "location": "kraken",
```
LefterisJP, line 217: Should it be location or also even type? Since location alone won't tell us much for eth withdrawals or eth block production

docs/websockets.rst
```diff
@@ -201,3 +201,43 @@ Having API keys for some services (such as etherscan) is crucial for rotki to wo
...
+            "name": "My kraken exchange"
+        }
+    }
+
+
+- ``status``: Can be either `querying_events_started`, `querying_events_finished`, `querying_events_status_update`. Each pair of events is triggered per exchange instance.
```
LefterisJP, line 223: Here in all of these the description seems to be specific to exchanges. I think we need to change this to make it generic enough for when adding WS progress for other event type queries.

rotkehlchen/types.py
```diff
@@ -829,3 +829,10 @@ class OracleSource(SerializableEnumMixin):
         dict['HistoryEventSubType', 'EventCategory'],
     ],
 ]
+
+
+class HistoryEventQueryType(SerializableEnumMixin):
+    """Locations to """
```
LefterisJP, line 835: Dude focus ...
```suggestion
    """Locations to query for history events"""
```

rotkehlchen/api/rest.py
```diff
@@ -3525,22 +3526,29 @@ def get_associated_locations(self) -> Response:
...
-            name: Literal['eth_withdrawals', 'block_productions'],
-    ) -> Response:
+    @async_api_call()
+    def query_online_events(self, query_type: HistoryEventQueryType) -> dict[str, Any]:
         """Queries the specified event type for any new events and saves them in the DB"""
+        if query_type == HistoryEventQueryType.EXCHANGES:
+            self.rotkehlchen.exchange_manager.query_history_events()
+            return OK_RESULT
+
+        # else we query eth2 events
```
LefterisJP, line 3536:
```suggestion
        # else we query eth staking events
```

rotkehlchen/exchanges/exchange.py
```diff
@@ -244,6 +244,10 @@ def query_online_income_loss_expense(
             'query_online_income_loss_expense should only be implemented by subclasses',
         )
 
+    def query_history_events(self) -> None:
+        """Query and store in the database history events from the current exchange instance"""
```
LefterisJP, line 248:
```suggestion
        """Query history events from the current exchange instance and store them in the database"""
```

rotkehlchen/exchanges/kraken.py
```diff
@@ -295,6 +296,7 @@ def __init__(
         self.call_counter = 0
         self.last_query_ts = 0
         self.history_events_db = DBHistoryEvents(self.db)
+        self.supports_history_events = True
```
LefterisJP, line 299: What is this? Why add a new variable? Just implement or not the function you added

rotkehlchen/exchanges/kraken.py
```diff
@@ -1149,3 +1160,27 @@ def query_kraken_ledgers(
...
+        self.msg_aggregator.add_message(
+            message_type=WSMessageType.HISTORY_EVENTS_STATUS,
+            data={
+                'status': str(HistoryEventsStep.QUERYING_EVENTS_STARTED),
+                'location': str(self.location),
+                'name': self.name,
+            },
+        )
+        with self.db.conn.read_ctx() as cursor:
+            self.query_kraken_ledgers(
```
LefterisJP, line 1174: This will properly query only not queried parts of the events, right? It won't query entire history every time I hope ... otherwise this won't work for anybody. Add a comment to make it obvious.

LefterisJP (commented):
Left some comments

rotkehlchen/exchanges/kraken.py
yabirgb: Yes, not all the events but only the parts that need it

rotkehlchen/exchanges/kraken.py
```diff
@@ -1149,3 +1164,34 @@ def query_kraken_ledgers(
                 return True  # we had errors so stop any further queries and quit
 
         return False  # no errors
+
+    def query_history_events(self) -> None:
+        """Query missing exchange history events.
+        This function will ask to query the full history but only the ranges that we haven't
+        queried yet will be queried.
+        """
```
LefterisJP, line 1172: You already specified this inside the function.

```suggestion
```

LefterisJP (approved):
LGTM

PR #5925, opened 2023-04-23
Adjust nfts logic

nebolax:
There was a problem that the order of addresses to query in nfts module was randomized, which led to different remote queries from test to test, and this led to a vcr issue since not all queries were covered in the cassette.

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/nft/nfts.py
+++ b/rotkehlchen/chain/ethereum/modules/nft/nfts.py
@@ -209,7 +209,7 @@ class Nfts(EthereumModule, CacheableMixIn, LockableQueryMixIn):
         with self.db.conn.read_ctx() as cursor:
             accounts = self.db.get_blockchain_accounts(cursor=cursor)
         # Be sure that the only addresses queried already exist in the database. Fix for #4456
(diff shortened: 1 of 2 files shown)


rotkehlchen/chain/ethereum/modules/nft/nfts.py
```diff
@@ -209,7 +209,7 @@ def query_balances(
         with self.db.conn.read_ctx() as cursor:
             accounts = self.db.get_blockchain_accounts(cursor=cursor)
         # Be sure that the only addresses queried already exist in the database. Fix for #4456
-        queried_addresses = list(set(accounts.eth) & set(addresses))
+        queried_addresses = sorted(set(accounts.eth) & set(addresses))  # Sorting for consistency in tests  # noqa: E501
```
LefterisJP, line 212: We never ever change production to accomodate tests. **NEVER**.

But it's needed to have reproducible queries from API endpoints. So this is why you do this.

```suggestion
        queried_addresses = sorted(set(accounts.eth) & set(addresses))  # Sorting to remove randomness
```

LefterisJP (approved):
nitpick , but serious one in the comment

PR #5926, opened 2023-04-23
Add an endpoint to query online sources of all event types

LefterisJP:


LefterisJP:
After making this I noticed @yabirgb has a Draft PR in a similar fashion but for exchange events: https://github.com/rotki/rotki/pull/5924/

It needs to be adjusted to use the endpoint introduced in this PR. We will need to have a generic way for the frontend to query for new online events of any type. At least for withdrawals and for block production they happen periodically in the background but querying on demand should be possible like this PR does.

docs/api.rst
```diff
@@ -5093,6 +5093,50 @@ Dealing with History Events
...
+.. http:post:: /api/(version)/history/events/query
+
+   Doing a POST on this endpoint will query latest online events for the given event type and save them in the DB
+
+   **Example Request**:
+
+   .. http:example:: curl wget httpie python-requests
+
+      POST /api/1/history/events/pull HTTP/1.1
+      Host: localhost:5042
```
yabirgb, line 5109: Url differ in the example from the one in the description

```suggestion
      POST /api/1/history/events/query HTTP/1.1
      Host: localhost:5042
```

docs/api.rst
```diff
@@ -5093,6 +5093,50 @@ Dealing with History Events
...
+
+   :reqjson bool async_query: Boolean denoting whether this is an asynchronous query or not
+   :reqjson string name: The name of the type of events to query for. Valid values are: ``"eth_withdrawals"``, ``"block_productions"``
+
+    **Example Response**
+
+    .. sourcecode:: http
+
+        HTTP/1.1 200 OK
+        Content-Type: application/zip
```
yabirgb, line 5126: I believe is json isntead of zip. Copy paste?

rotkehlchen/api/rest.py
```diff
@@ -3525,6 +3525,23 @@ def get_associated_locations(self) -> Response:
...
         )
 
+    def query_online_events(
+            self,
+            name: Literal['eth_withdrawals', 'block_productions'],
+    ) -> Response:
+        """Queries the specified event type for any new events and saves them in the DB"""
+        eth2 = self.rotkehlchen.chains_aggregator.get_module('eth2')
+        if eth2 is None:
+            return api_response(wrap_in_fail_result('eth2 module is not active'), status_code=HTTPStatus.CONFLICT)  # noqa: E501
```
yabirgb, line 3535: okey will add my logic to this section of the code

rotkehlchen/api/v1/schemas.py
```diff
@@ -333,6 +333,12 @@ def validate_schema(
             )
 
 
+class EventsOnlineQuerySchema(AsyncQueryArgumentSchema):
+    name = fields.String(
+        validate=webargs.validate.OneOf(choices=('eth_withdrawals', 'block_productions')),
```
yabirgb, line 338: Shouldn't this literals be in a list to avoid repeating them in the `resources.py` file too?

yabirgb (commented):
A few things in the docs need adjustment

docs/api.rst
LefterisJP: yes

docs/api.rst
LefterisJP: ah fuck I missed this one. Started with pull but then saw query in yours and liked it more so I adjusted.

LefterisJP:
Will merge this for now and do the rest in another PR I am working on

PR #5927, opened 2023-04-24
feat: update location selector

lukicenturi:
Closes #(issue_number)

## Checklist


frontend/app/src/components/display/ExchangeDisplay.vue
```diff
@@ -1,30 +1,30 @@
...
+const location: ComputedRef<TradeLocationData | undefined> = computed(() =>
+  getLocation(exchange)
 );
 
 const name = computed<string>(
   () => get(location)?.name ?? toSentenceCase(get(exchange))
 );
 
-const icon = computed<string>(() => get(location)?.icon ?? '');
+const image = computed<string>(() => get(location)?.image ?? '');
```
kelsos, line 21: Would it be possible to use the `useRefMap` composable in these two cases?

frontend/app/src/components/history/events/HistoryEventNote.vue
```diff
@@ -34,27 +36,32 @@ interface NoteFormat {
...
+const {
+  notes,
+  amount,
+  asset: assetId,
+  noTxHash,
+  validatorIndex
+} = toRefs(props);
+
+const formatNotes: ComputedRef<NoteFormat[]> = computed(() => {
   const { assetSymbol } = useAssetInfoRetrieval();
```
kelsos, line 51: Can we move this outside of the computed function?

frontend/app/src/components/history/events/HistoryEventNote.vue
```diff
@@ -34,27 +36,32 @@ interface NoteFormat {
...
 
-  const asset = get(assetSymbol(assetId));
+  const asset = get(assetSymbol(get(assetId)));
```
kelsos, line 53: `assetSymbol` gets `MaybeRef` right? Is there a problem with the typing? we should not have to do `get` in the parameter

frontend/app/src/composables/api/history/index.ts
```diff
@@ -31,8 +32,23 @@ export const useHistoryApi = () => {
     return handleResponse(response);
   };
 
+  interface AllLocationResponse {
+    locations: Record<string, Omit<ActionDataEntry, 'identifier'>>;
+  }
```
kelsos, line 37: Can we put the interface at the top-level instead of inside the method?

frontend/app/src/types/history/events/index.ts
```diff
@@ -87,7 +87,9 @@ export const HistoryEvent = z.object({
   notes: z.string().nullish(),
   counterparty: z.string().nullish(),
   product: z.string().nullish(),
-  address: z.string().nullish()
+  address: z.string().nullish(),
+  isExit: z.boolean().nullish(),
+  validatorIndex: z.number().nullish()
```
kelsos, line 92: For this as soon as there is a way to distinguish between the types let's separate them. Also, we need to make sure that we don't use `nullish`. In most cases, `nullish` is overly broad, since the API is either using one or the other. The correct approach should be either `nullable` or `optional`.

frontend/app/src/pages/locations/[identifier].vue
```diff
@@ -5,16 +5,16 @@ defineOptions({
...
 const { identifier } = toRefs(props);
 
-const { getLocation } = useLocationInfo();
+const { getLocation } = useLocations();
 
-const location = computed<TradeLocationData>(() =>
-  getLocation(get(identifier))
+const location: ComputedRef<TradeLocationData> = computed(() =>
+  getLocation(identifier)
 );
```
Comment (kelsos) [inline frontend/app/src/pages/locations/[identifier].vue:18]: Same as the previous comments, expose the wrapped version from `useLocation`

frontend/app/src/composables/session/load.ts
```diff
@@ -1,11 +1,13 @@
 import { Section, Status } from '@/types/status';
+import { useLocations } from '@/composables/locations';
```
kelsos, line 2: The import here should be redundant.

frontend/app/src/composables/locations.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    assert(!!exchange, 'location should not be falsy');
+    return exchange.name;
+  };
+
+  const { getAssetImageUrl } = useAssetIcon();
+  const { getChainInfoById } = useSupportedChains();
+
+  const getLocation = (identifier: MaybeRef<string>): TradeLocationData => {
+    const id = get(identifier);
```
kelsos, line 63: Can we expose two versions of this to avoid doing the wrapping in computed in each component?

kelsos (approved):
lgtm thanks @lukicenturi

PR #5929, opened 2023-04-24
Adding block production events for block reward and MEV reward

LefterisJP:
1. Added the block production reward and MEV reward event as another type of ETH staking event
2. Various smaller adjustment fixes
3. Removed unused blocks table
4. Added a transaction hash field to the evm_events_info to replace the usage of event_identifier as the way we store transaction hash.
5. Since bytes are no longer needed due to (4) turned event_identifier to a string
6. Add the logic that will detect which transaction events are MEV reward transfers
7. Call it periodically


LefterisJP:
All green. Woohoo! :partying_face:

PR #5935, opened 2023-04-26
fix: some blockchain balances refreshing issue

lukicenturi:
Closes #(issue_number)

## Checklist

This fix some problems Alexey had 
1. The refresh icon on the `<account-balances />` does not do token detection. Only balances refresh.
2. Refresh of tokens + balances might abort here https://github.com/rotki/rotki/blob/bfc6da4f3c159bed7809fd6f725ca6fd67fd296b/frontend/app/src/composables/blockchain/balances/index.ts#L32 if the status is loading() for some reason.
3. Fix the tokens detection issue, where the refresh icon on the image is not properly disabled when there is a token detection/balances refresh process already running.

### DIFF
--- a/frontend/app/src/composables/blockchain/balances/index.ts
+++ b/frontend/app/src/composables/blockchain/balances/index.ts
@@ -23,13 +23,14 @@ export const useBlockchainBalances = () => {
 
   const fetch = async (
     blockchain: Blockchain,
-    ignoreCache = false
+    ignoreCache = false,
+    force = false
   ): Promise<void> => {
     const { loading, setStatus, resetStatus, isFirstLoad } = useStatusUpdater(
       chainSection[blockchain]
     );
 
-    if (loading()) {
+    if (loading() && !force) {
       return;
     }
 
@@ -72,7 +73,8 @@ export const useBlockchainBalances = () => {
   const fetchBlockchainBalances = async (
     payload: BlockchainBalancePayload = {
       ignoreCache: false
-    }
+    },
+    force = false
   ): Promise<void> => {
     const { blockchain, ignoreCache } = payload;
 
@@ -84,7 +86,9 @@ export const useBlockchainBalances = () => {
     }
 
     try {
-      await Promise.allSettled(chains.map(chain => fetch(chain, ignoreCache)));
+      await Promise.allSettled(
+        chains.map(chain => fetch(chain, ignoreCache, force))
+      );
     } catch (e: any) {
       logger.error(e);
       const message = tc('actions.balances.blockchain.error.description', 0, {
--- a/frontend/app/src/composables/blockchain/balances/index.ts
+++ b/frontend/app/src/composables/blockchain/balances/index.ts
@@ -94,12 +94,9 @@ export const useBlockchainBalances = () => {
   ): Promise<void> => {
     const { blockchain, ignoreCache } = payload;
 
-    const chains: Blockchain[] = [];
-    if (!blockchain) {
-      chains.push(...Object.values(Blockchain));
-    } else {
-      chains.push(blockchain);
-    }
+    const chains: Blockchain[] = blockchain
+      ? [blockchain]
+      : Object.values(Blockchain);
 
     try {
       await Promise.allSettled(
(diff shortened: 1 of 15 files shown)


nebolax (commented):
The functionality looks good. About (2) did you find anything?

frontend/app/src/composables/blockchain/balances/index.ts
```diff
@@ -23,13 +23,14 @@ export const useBlockchainBalances = () => {
...
-    ignoreCache = false
+    ignoreCache = false,
+    force = false
   ): Promise<void> => {
     const { loading, setStatus, resetStatus, isFirstLoad } = useStatusUpdater(
       chainSection[blockchain]
     );
 
-    if (loading()) {
+    if (loading() && !force) {
```
kelsos, line 33: In which case would you need to force refresh balances when they are already refreshing?

lukicenturi: For token detection.
So it guaranteed that after token detection, the balance refreshing should always happens.

In the past there is a case that 
- balance refreshing is running, then token detection is called
- when token detection is done, it should called the `balance refreshing`, but canceled because the previous `balance refreshing` is still running.

Although in the UI I try to make sure that when you call balance refreshing, the token detection button is disabled, and vice versa, but I think this force mechanism is nice to have (?)

kelsos: @rotki/backend what happens if the frontend schedules a second balance refresh while one is already running?

@lukicenturi let's see if this is a problem or not for the backend, but from the frontend perspective, rather than forcing a second one while one is running I think either canceling the first one before starting the second or queueing the second one would be a better approach.

frontend/app/src/store/blockchain/tokens.ts
```diff
@@ -39,6 +39,18 @@ export const useBlockchainTokensStore = defineStore('blockchain/tokens', () => {
...
+    const states = {
+      [Blockchain.ETH]: ethTokens,
+      [Blockchain.OPTIMISM]: optimismTokens
+    };
+
+    set(states[chain], {
+      ...get(states[chain]),
+      ...data
+    });
+  };
```
kelsos, line 52: Rather than doing this do you think we can do the same thing we do in blockchain balances? and use a record instead of two individual refs?

frontend/app/src/composables/blockchain/balances/index.ts
```diff
@@ -69,10 +65,32 @@ export const useBlockchainBalances = () => {
...
+    const loading = isLoading(chainSection[blockchain]);
+
+    const call = () => handleFetch(blockchain, ignoreCache);
+
+    if (get(loading)) {
+      if (periodic) {
+        return;
+      }
+
+      await until(loading).toBe(false);
```
lukicenturi, line 83: @kelsos, by assuming that all manual refresh all disabled, when there is balance refresh running, this part will queued once for manual refresh (by click the refresh button, or triggered by detect tokens)

kelsos: Can you maybe also add a test for this? just to be sure?

lukicenturi: absolutely

frontend/app/src/store/blockchain/tokens.ts
```diff
@@ -39,6 +46,17 @@ export const useBlockchainTokensStore = defineStore('blockchain/tokens', () => {
     );
   };
 
+  const setState = (chain: TokenChains, data: EvmTokensRecord) => {
+    const tokensVal = { ...get(tokensState) };
+    set(tokensState, {
+      ...tokensVal,
+      [chain]: {
+        ...tokensVal[chain],
+        ...data
```
kelsos, line 55: Do we really need this? doesn't data have all the detected tokens?

lukicenturi: so the setState happens per address call, so the data returned there is data on 1 address only

kelsos (approved):
lgtm @lukicenturi thanks

PR #5943, opened 2023-04-27
feat: online history events query

lukicenturi:
Closes #(issue_number)

## Checklist


lukicenturi:
1. One comment is to remove the event group id for kraken events, and instead just show it as `Kraken event may be enough`. But I think need to wait for the design from Raquel.

frontend/app/.eslintrc.js
```diff
@@ -17,7 +17,9 @@ module.exports = {
           '/backend_mappings.*/',
           '/notification_messages.missing_api_key.*/',
           '/premium_components.*/',
-          '/transactions.query_status.*/'
+          '/transactions.query_status.*/',
+          '/transactions.query_status_events.*/',
+          '/transactions.events.headers.*/'
```
kelsos, line 22: We need to figure out something for this, just ignoring everything would make this hard to maintain.

frontend/app/src/components/history/events/HistoryEventType.vue
```diff
@@ -11,74 +18,57 @@ const props = defineProps<{
...
-  useHistoryEventMappings();
-
+const { getEventTypeData } = useHistoryEventMappings();
 const attrs = getEventTypeData(event);
-const counterparty = getEventCounterpartyData(event);
 
-const isTx: ComputedRef<boolean> = computed(() =>
-  isValidTxHash(get(event).eventIdentifier)
-);
+const eventAsEvmEvent = event as Ref<EvmHistoryEvent>;
```
lukicenturi, line 24: Tbh I'm still confused how to do it better without casting

frontend/app/src/components/history/events/HistoryEventsIdentifier.vue
```diff
@@ -0,0 +1,91 @@
...
+
+const translationKey: ComputedRef<string> = computed(
+  () => `transactions.events.headers.${get(event).entryType.replace(/ /g, '_')}`
+);
+
+const { getChain } = useSupportedChains();
+
+const blockEvent = event as Ref<EthBlockEvent>;
+const withdrawEvent = event as Ref<EthWithdrawalEvent>;
+const evmEvent = event as Ref<EvmHistoryEvent>;
```
lukicenturi, line 24: Tbh I'm still confused how to do it better without casting

frontend/app/src/utils/history/events.ts
```diff
@@ -0,0 +1,32 @@
+import { type MaybeRef } from '@vueuse/core';
+import { HistoryEventEntryType } from '@rotki/common/lib/history/events';
+
+export const isEvmEvent = (type: MaybeRef<HistoryEventEntryType>) =>
+  get(type) === HistoryEventEntryType.EVM_EVENT;
+
+export const isEvmEventRef = (
+  type: MaybeRef<HistoryEventEntryType>
+): ComputedRef<boolean> => computed(() => isEvmEvent(type));
```
kelsos, line 9: To resolve the casting issue just change the first to a type guard:

```typescript
export const isEvmEvent = (
  type: MaybeRef<HistoryEventEntryType>
): type is MaybeRef<EvmHistoryEvent> =>
  get(type) === HistoryEventEntryType.EVM_EVENT;
```

And make the second to return either object if the type guard is true or `undefined` if false so that isEvmEventRef returns `ComputedRef<EvmHistoryEvent | undefined>` you can do a v-if on the undefined/defined and narrow the types in the checks and you don't have to cast or lose types.

kelsos (commented):
I am still going through but here are some initial comments

frontend/app/src/store/history/events-query-status.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    const resetQueryStatus = (): void => {
+      set(queryStatus, {});
+    };
+
+    const isStatusFinished = (item: HistoryEventsQueryData): boolean =>
+      item.status === HistoryEventsQueryStatus.QUERYING_EVENTS_FINISHED;
+
+    const isAllFinished: ComputedRef<boolean> = computed(() => {
+      const queryStatusVal = get(queryStatus);
```
kelsos, line 32: nit: can we do`statuses`

frontend/app/src/composables/history/events/tx-query-status.ts
```diff
@@ -1,11 +1,45 @@
...
+    if (chains.length === 0) {
+      return statuses;
+    }
+
+    return statuses.filter(({ evmChain }) =>
+      chains.includes(getChain(evmChain))
+    );
+  });
+
+  const sortedQueryStatus: ComputedRef<EvmTransactionQueryData[]> = computed(
```
kelsos, line 30: there should be a useArraySort util in vue use can we use it? :)

frontend/app/src/composables/history/events/events-query-status.ts
```diff
@@ -0,0 +1,61 @@
+import { type ComputedRef } from 'vue';
+import { type MaybeRef } from '@vueuse/core';
+import { type HistoryEventsQueryData } from '@/types/websocket-messages';
+
+export const useEventsQueryStatus = (locations: MaybeRef<string[]> = []) => {
```
kelsos, line 5: Can you extract a common core with this and the tx-query-status they have similarities

frontend/app/src/components/settings/explorers/Explorers.vue
```diff
@@ -11,21 +11,30 @@ const { explorers } = storeToRefs(store);
...
 
-const txUrl = computed<string>(() => {
+const addressUrl = computed<string>(() => {
   const setting = get(explorers)[get(selection)];
-  return setting?.transaction ?? explorerUrls[get(selection)].transaction;
+  return setting?.address ?? explorerUrls[get(selection)].address;
 });
 
-const addressUrl = computed<string>(() => {
+const txUrl: ComputedRef<string | null> = computed(() => {
```
kelsos, line 21: where we do multiple `get(selection)` in the same composable do you think you can extract it to `const chain = get(selection)`

frontend/app/src/components/settings/explorers/Explorers.vue
```diff
@@ -76,6 +85,26 @@ const saveTransaction = async (newTransaction?: string) => {
...
+    delete updated.block;
+  }
+
+  await store.updateSetting({
+    explorers: {
+      ...get(explorers),
+      [get(selection)]: updated
+    }
+  });
+};
```
kelsos, line 106: do you think we can somehow unify these 3 methods by making a new function that take the new value and the key? they seem almost the same with only the update key being different

frontend/app/src/components/display/EnsAvatar.vue
```diff
@@ -26,13 +26,20 @@ const avatarUrl: ComputedRef<string | null> = computed(() => {
...
-      <v-img :src="getBlockie(address)" />
-      <v-img v-if="avatarUrl" :class="css.avatar" :src="avatarUrl" />
+  <v-lazy :class="css.wrapper">
+    <div>
+      <v-img v-if="!success" :src="getBlockie(address)" />
+      <v-img
+        v-if="avatarUrl"
+        :class="css.avatar"
+        :src="avatarUrl"
+        @load="success = true"
```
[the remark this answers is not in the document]
lukicenturi: yes, it's true

frontend/app/src/utils/history/events.ts
lukicenturi: Oh yeah this is a cool idea, will try it

frontend/app/src/components/history/events/HistoryEventsView.vue
```diff
@@ -148,10 +156,13 @@ const {
...
-    useHistoryEventFilter({
-      protocols: get(protocols).length > 0,
-      locations: !!get(location)
-    }),
+    useHistoryEventFilter(
+      {
+        protocols: get(protocols).length > 0,
+        locations: !!get(location)
+      },
+      get(entryTypes)
```
kelsos, line 164: Why not pass the ref directly to the composable?

frontend/app/src/components/history/events/HistoryEventsView.vue
```diff
@@ -565,81 +591,82 @@ const fetchDataAndRefreshEvents = async (
...
-                      </v-list-item-icon>
-                      <v-list-item-content>
-                        {{ tc('transactions.actions.redecode_events') }}
-                      </v-list-item-content>
-                    </v-list-item>
-                  </v-list>
-                </v-menu>
-              </div>
+              <v-lazy>
+                <div class="d-flex align-center">
```
kelsos: do you think it would be possible to extract the menu to a new component

frontend/app/src/components/display/EnsAvatar.vue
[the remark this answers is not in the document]
kelsos: Does it make sense to calculate and cache a blockie to only show it while the avatar is loading though? What do you think?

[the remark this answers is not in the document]
kelsos: Aren't 2 and 3 the same? So what I mean wouldn't it make more sense to create and cache the blockie only if the avatar fails to load?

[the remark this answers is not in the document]
kelsos: We can take this to a new PR btw no need to do it here

kelsos (approved):
lgtm @lukicenturi thank you.

I did some minor adjustments here and there (extracted some utilities too)

PR #5944, opened 2023-04-27
Small fixes backend

yabirgb:
- Introduces two new mappings and icons for MEV rewards and block production
- Removes a pending task from the rest.py file
- Fixes an issue in the steps for accounting where we were overwriting a variable before using it


rotkehlchen/history/events.py
```diff
@@ -355,8 +354,8 @@ def new_step_cb(state_name: str) -> None:
                 fail_callback=fail_history_cb,
                 new_step_data=(new_step_cb, exchange.name),
             )
-            step = self._increase_progress(step, total_steps)
-            step = step_before + STEPS_PER_CEX  # Make sure remote failures don't throw steps off
+            # each exchange instance executes STEPS_PER_CEX steps for total_steps
+            step = self._increase_progress(step, total_steps, step_by=STEPS_PER_CEX)
```
nebolax, line 358: Why a new argument is needed? Why not do it this way?

```suggestion
            step = self._increase_progress(step + STEPS_PER_CEX, total_steps)
```

rotkehlchen/api/rest.py
```diff
@@ -3568,11 +3570,11 @@ def get_history_events(
                 group_by='event_identifier' if group_by_event_ids else None,
             )
             location = filter_query.location
-            chain_id = None
-            if location == Location.ETHEREUM:  # todo: this should prolly become a function
-                chain_id = ChainID.ETHEREUM
-            elif location == Location.OPTIMISM:
-                chain_id = ChainID.OPTIMISM
+            if location in EVM_LOCATIONS:
```
nebolax, line 3573: `to_chain_id`'s documentation says "Dealing directly with ints since it's only used in one place". Since it is now used in more than one place, and here `ChainID` class is needed, maybe it will make sense to modify `to_chain_id` to return `ChainID` ?

rotkehlchen/api/rest.py
```diff
@@ -3568,11 +3570,11 @@ def get_history_events(
...
+                chain_id = ChainID(Location.to_chain_id(location))
+            else:
+                chain_id = None
```
nebolax, line 3576: iirc we prefer this format in such places

```suggestion
            chain_id = None
            if location in EVM_LOCATIONS:
                chain_id = ChainID(Location.to_chain_id(location)) 
```

or even

```suggestion
            chain_id = ChainID(Location.to_chain_id(location)) if location in EVM_LOCATIONS else None
```

rotkehlchen/accounting/constants.py
```diff
@@ -1,2 +1,183 @@
...
+        HistoryEventSubType.RECEIVE_WRAPPED: EventCategory.RECEIVE,
+        HistoryEventSubType.REMOVE_ASSET: EventCategory.WITHDRAW,
+        HistoryEventSubType.RETURN_WRAPPED: EventCategory.SEND,
+        HistoryEventSubType.FEE: EventCategory.FEE,
+        HistoryEventSubType.MEV_REWARD: EventCategory.MEV,
+        HistoryEventSubType.BLOCK_PRODUCTION: EventCategory.CREATE_BLOCK,
+    },
+}
+
+EVENT_CATEGORY_DETAILS = {
```
nebolax, line 70: What is the reason for moving this mapping?

yabirgb: Doesn't contain only evm default events but also defaults for eth staking and also we have some events used by exchanges

rotkehlchen/history/events.py
yabirgb: In `_increase_progress` the `step` variable will be modified and the new value will be `step + STEPS_PER_CEX + 1`. To achieve what you want then the correct value would be `step + STEPS_PER_CEX - 1` and this starts to be ugly. `step_by` seems a more redeable argument for me. What I missed is a docstring explaining what it does

rotkehlchen/api/rest.py
yabirgb: Is not used anymore in only one place and the integer value is the thing used, not the chainId instance. Also the comment states that is there to avoid the import issues so I will keep it. Change it to a single line assignation

rotkehlchen/history/events.py
nebolax: Hmmm yeah, probably makes sense :+1:

rotkehlchen/chain/evm/decoding/types.py
```diff
@@ -45,3 +45,5 @@ class EventCategory(SerializableEnumNameMixin):
     CANCEL_ORDER = auto()
     REFUND = auto()
     FEE = auto()
+    MEV = auto()
```
LefterisJP, line 48: please name this MEV_REWARD. MEV is way too generic. We may have different type of MEV events in the frontend in the future.
```suggestion
    MEV_REWARD = auto()
```

Question. Renaming those is it easy in the future? Is it just a frontend thing? So can easily just rename here and in frontend and we are good?

rotkehlchen/history/events.py
```diff
@@ -355,8 +354,8 @@ def new_step_cb(state_name: str) -> None:
                 fail_callback=fail_history_cb,
                 new_step_data=(new_step_cb, exchange.name),
             )
-            step = self._increase_progress(step, total_steps)
-            step = step_before + STEPS_PER_CEX  # Make sure remote failures don't throw steps off
+            # each exchange instance executes STEPS_PER_CEX steps for total_steps
```
LefterisJP, line 357: What do you mean by the "for total steps" part. It does not make sense in English. Do you mean something like?

```suggestion
            # each exchange instance executes STEPS_PER_CEX steps out of the total_steps
```

or something completely different?

LefterisJP (commented):
small questions

rotkehlchen/history/events.py
yabirgb: It meant towards the count of total_steps. But yes

LefterisJP: Writing `towards the count of total_steps` is also understandable.

LefterisJP (approved):
LGTM

PR #5946, opened 2023-04-27
Expose validator_indices querying filter for history events

LefterisJP:


LefterisJP:
Frontend test failure is unrelated

PR #5948, opened 2023-04-27
Catch RemoteError at check_single_address_activity

LefterisJP:
If a RemoteError happens at check address activity do not stop the entire task,but just skip that chain


LefterisJP:
Merging. Errors are substrate interface timeouts for polkadot. Passed locally.

PR #5949, opened 2023-04-27
Remove some migrations, fix some edge cases

LefterisJP:
- Remove un-needed data migrations
- Fix substrate interface edge case
- Add stracktrace log printing to the error case of migrations


LefterisJP:
all green. Merging.

PR #5950, opened 2023-04-27
Add eth2 staking stats

yabirgb:
Add stats about total eth earned


rotkehlchen/db/eth2.py
```diff
@@ -215,3 +218,69 @@ def delete_validator(self, validator_index: Optional[int], public_key: Optional[
...
+        with self.db.conn.read_ctx() as cursor:
+            # query eth withdrawn
+            filter_query.filters = withdrawals_filters
+            query, bindings = filter_query.prepare(with_pagination=False)
+            cursor.execute(base_query + query, bindings)
+            if (amount := cursor.fetchone()[0]) is not None:
+                withdrawn_amount = FVal(amount)
+
+            # query blocks produced and mev rewards
+            filter_query.filters = rewards_filters
```
nebolax, line 280: Yeah, as said above, this assignment modification of `filters` looks ugly to me

rotkehlchen/tests/unit/test_eth2.py
```diff
@@ -939,3 +943,97 @@ def test_combine_block_with_tx_events(eth2, database):
         event_identifier=EthBlockEvent.form_event_identifier(block_number),
     )
     assert modified_event == events[2]  # pylint: disable=unsubscriptable-object
+
+
+def test_eth_validator_stat_calculation(database):
+    """Test that filter works correctly to obtains stats for eth staking events"""
```
nebolax, line 949:
```suggestion
    """Test that filter works correctly and obtains stats for eth staking events"""
```

nebolax (commented):
Some comments

rotkehlchen/api/v1/resources.py
```diff
@@ -1821,12 +1821,18 @@ def patch(self, validator_index: int, ownership_percentage: FVal) -> Response:
...
+    post_schema = HistoryEventSchema()
 
     @require_premium_user(active_check=False)
     @use_kwargs(get_schema, location='json_and_query')
     def get(self, async_query: bool) -> Response:
         return self.rest_api.get_eth2_stake_details(async_query=async_query)
 
+    @require_premium_user(active_check=False)
+    @use_kwargs(get_schema, location='json_and_query')
+    def post(self, filter_query: 'HistoryEventFilterQuery', group_by_event_ids: bool) -> Response:  # pylint: disable=unused-argument  # noqa: E501
```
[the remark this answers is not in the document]
yabirgb: Is due to the other schema but I'm adding a new one and will remove this

rotkehlchen/chain/aggregator.py
```diff
@@ -1094,7 +1094,10 @@ def get_eth2_staking_details(self) -> list['ValidatorDetails']:
         eth2 = self.get_module('eth2')
         if eth2 is None:
             raise ModuleInactive('Cant query eth2 staking details since eth2 module is not active')
-        return eth2.get_details(addresses=self.queried_addresses_for_module('eth2'))
+
+        return eth2.get_details(
```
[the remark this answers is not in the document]
yabirgb: I tried different approaches and forgot to undo this change

rotkehlchen/db/filtering.py
```diff
@@ -1143,6 +1144,25 @@ def make(
                 operator='IN',
             ))
 
+        if location_labels_or_validator_indices:
```
nebolax, line 1147: Why not to split it into 2 filters? One for location labels and the other one for validator indices

rotkehlchen/db/eth2.py
```diff
@@ -192,3 +193,27 @@ def delete_validator(self, validator_index: Optional[int], public_key: Optional[
...
+
+    def get_validators_profit(
+            self,
+            withdrawals_filter_query: EthStakingEventFilterQuery,
+            execution_filter_query: EthStakingEventFilterQuery,
+    ) -> list[FVal]:
+        """Query profit as sum of withdrawn ETH and rewards for block production and mev"""
+        filters = (withdrawals_filter_query, execution_filter_query)
+        result = [ZERO] * len(filters)
+        base_query = (
```
nebolax, line 205: This could go to a constant

rotkehlchen/db/eth2.py
```diff
@@ -192,3 +193,27 @@ def delete_validator(self, validator_index: Optional[int], public_key: Optional[
                     f'Tried to delete eth2 validator with {field} '
                     f'{input_tuple[0]} from the DB but it did not exist',
                 )
+
+    def get_validators_profit(
+            self,
+            withdrawals_filter_query: EthStakingEventFilterQuery,
+            execution_filter_query: EthStakingEventFilterQuery,
+    ) -> list[FVal]:
```
nebolax, line 201: As said in the comment below, I think it will be better to be explicit here and just return a tuple of 2 values.

In the place where this function is used, it is used in format `withdrawn_amount, execution_layer_rewards = dbeth2.get_validators_profit(`, so the values are unpacked anyway.

rotkehlchen/api/v1/schemas.py
```diff
@@ -3127,3 +3127,79 @@ class ClearAvatarsCacheSchema(Schema):
...
+        base_arguments = {
+            'from_ts': data['from_timestamp'],
+            'to_ts': data['to_timestamp'],
+        }
+        validator_indices = data['validator_indices']
+
+        # for mev rewards if we filter by address is possible to miss the block production
+        # since the locaiton label is the builder address and not the user address. To avoid this
+        # we use the location label to get the validator index and then use it to add extra
+        # validator indices to not miss any event.
```
nebolax, line 3158: I am not sure why does it have to happen here and can't happen in the main query (the one that would happen in `dbeth2.get_validators_profit` function)?

rotkehlchen/api/v1/schemas.py
```diff
@@ -3127,3 +3127,79 @@ class ClearAvatarsCacheSchema(Schema):
...
+            if validator_indices is not None and len(extra_indices) != 0:
+                validator_indices.extend(extra_indices)
+
+        if None in (location_labels, validator_indices):
+            common_arguments = base_arguments | {
+                'location_labels': location_labels,
+                'validator_indices': validator_indices,
+            }
+        else:
+            common_arguments = base_arguments | {
```
nebolax, line 3183: As said in the comment in `filtering.py`, I think it will be much cleaner to handle this condition directly in the filtering class. So always pass `location_labels` and `validator_indices`, and if they are both not None, make the nested `or` filter.

rotkehlchen/db/filtering.py
yabirgb: The reason to have this extra filter is because as for how filters work right now if you provide at the same time `location_labels` and `validator_indices` the filter will do `... AND location_label IN location_labels AND validator_index IN validator_indices AND ...`

What you really want is by filtering by some validator index and some address is `... AND (location_label IN location_labels OR validator_index IN validator_indices) AND...`. Notice the difference from `AND` to `OR`. For this we need to use a nested filter and this is why I have this extra logic here

rotkehlchen/db/eth2.py
```diff
@@ -192,3 +193,27 @@ def delete_validator(self, validator_index: Optional[int], public_key: Optional[
...
+        result = [ZERO] * len(filters)
+        base_query = (
+            'SELECT SUM(CAST(amount AS REAL)) FROM history_events JOIN '
+            'eth_staking_events_info ON eth_staking_events_info.identifier='
+            'history_events.identifier '
+        )
+
+        with self.db.conn.read_ctx() as cursor:
+            # query eth withdrawn
+            for idx, filter_query in enumerate(filters):
```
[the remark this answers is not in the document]
yabirgb: I had it in two queries but ended up repeating code. will restore that if is clearer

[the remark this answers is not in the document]
nebolax: Can wait for @LefterisJP's opinion on this. As for repeating code maybe it will be possible to create some auxiliary functions?

rotkehlchen/db/filtering.py
nebolax: Yeah, I get it. My suggestion, as said here https://github.com/rotki/rotki/pull/5950#discussion_r1182266551, is to handle this condition in this file, in `filtering.py`

rotkehlchen/api/v1/schemas.py
yabirgb: We discussed about this in the discord thread https://discord.com/channels/657906918408585217/1102958606061686954/1102959668776677376 And I prefer to not move the logic to the filter itself. The reason is that you would still need an argument to tell if this two fields should be queried with an `AND` or an `OR`, that adds a db query inside the filter and this is bad imo, and we are abstracting in a way that maybe is not needed anywhere else. If we endup needing it we can then move to the filter

docs/api.rst
```diff
@@ -8581,24 +8581,33 @@ Getting Loopring balances
...
       Host: localhost:5042
 
-   :reqjson bool async_query: Boolean denoting whether this is an asynchronous query or not
+      {'validator_indices': [1111]}
+
+      :reqjson bool async_query: Boolean denoting whether this is an asynchronous query or not.
+      :reqjson bool only_cache: If true then only the stats are calculated and the details are empty.
+      :reqjson int from_timestamp: The timestamp from which to start querying. Default is 0.
+      :reqjson int to_timestamp: The timestamp until which to query. Default is now.
+      :reqjson list[string] location_labels: A list of location labels to optionally filter by. Location label is a string field that allows to provide more information about the location. When we use this structure in blockchains, it is used to specify the user address. For exchange events it's the exchange name assigned by the user.
```
LefterisJP, line 8608: Do not copy paste from other docs. Give more specific information about how location label is treated **here**

docs/api.rst
```diff
@@ -8581,24 +8581,33 @@ Getting Loopring balances
...
+      :reqjson list[int] validator_indices: An optional list of validator indices to filter by. This makes it an EthStakingevent query
```
LefterisJP, line 8609: Do not copy paste from other docs. Give more specific information about how location label is treated **here**

```suggestion
      :reqjson list[int] validator_indices: An optional list of validator indices to filter by.
```

docs/api.rst
```diff
@@ -8638,11 +8650,14 @@ Getting Eth2 Staking details
...
               "performance_1y": {"amount": "0", "usd_value": "0"},
               "performance_total": {"amount": "42.5", "usd_value": "43500"}
-          }],
+          }]
+        },
         "message": "",
       }
 
-   :resjson result list: The result of the Eth2 staking details for all of the user's accounts. It's a list of details per validator. Important thing to note here is that if all performance entries are 0 then this means that the validator is not active yet and is still waiting in the deposit queue.
+   :resjson withdrawn_consensus_layer_rewards string: Amount of ETH collected as withdrawals from the beacon chain.
```
LefterisJP, line 8658:
```suggestion
   :resjson withdrawn_consensus_layer_rewards string: Amount of ETH collected as withdrawals from the beacon chain for the given filter.
```

docs/api.rst
```diff
@@ -8638,11 +8650,14 @@ Getting Eth2 Staking details
...
+   :resjson execution_layer_rewards string: Amount of ETH earned as part of MEV rewards and new blocks created.
```
LefterisJP, line 8659:
```suggestion
   :resjson execution_layer_rewards string: Amount of ETH earned as part of MEV rewards and new blocks created for the given filter.
```

docs/api.rst
```diff
@@ -8638,11 +8650,14 @@ Getting Eth2 Staking details
...
+   :resjson details list: The result of the Eth2 staking details for all of the user's accounts. It's a list of details per validator. Important thing to note here is that if all performance entries are 0 then this means that the validator is not active yet and is still waiting in the deposit queue.
```
LefterisJP, line 8660: Isn't this also for the given filter?

rotkehlchen/api/v1/schemas.py
```diff
@@ -3127,3 +3128,75 @@ class ClearAvatarsCacheSchema(Schema):
         fields.String(required=True, validate=lambda x: x.endswith('.eth')),
         load_default=None,
     )
+
+
+class EthStakingHistoryStats(AsyncQueryArgumentSchema, OnlyCacheQuerySchema):
+    """Schema for quering history events"""
```
LefterisJP, line 3134:
```suggestion
    """Schema for quering ethereum staking history stats"""
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -3127,3 +3128,75 @@ class ClearAvatarsCacheSchema(Schema):
...
+            **_kwargs: Any,
+    ) -> dict[str, Any]:
+        base_arguments = {
+            'from_ts': data['from_timestamp'],
+            'to_ts': data['to_timestamp'],
+        }
+        validator_indices = data['validator_indices']
+
+        # for mev rewards if we filter by address is possible to miss the block production
+        # since the locaiton label is the builder address and not the user address. To avoid this
```
LefterisJP, line 3157:
```suggestion
        # since the location label is the builder address and not the user address. To avoid this
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -3127,3 +3128,75 @@ class ClearAvatarsCacheSchema(Schema):
...
+        # we use the location label to get the validator index and then use it to add extra
```
LefterisJP, line 3158: Your logic for doing this does not make much sense though.

If the location label is the builder address then you are not interested in that query, no? You should **not** sum up these events if they are not tracked. The amount of ETH that goes in the block production to the builder is not something I am getting. So it should not be part of this

rotkehlchen/db/filtering.py
```diff
@@ -1112,6 +1112,7 @@ def make(
             entry_types: Optional[list[HistoryBaseEntryType]] = None,
             exclude_ignored_assets: bool = False,
             validator_indices: Optional[list[int]] = None,
+            location_labels_or_validator_indices: Optional[tuple[list[str], list[int]]] = None,
```
LefterisJP, line 1115: Hmm okay this approach is quite bad.

We do not have such logic in the filters. Yes indeed there is no "OR"ing filters but this is not the way to go. This is a very ugly hack.

Why do you need to or those? As I wrote in the api.rest file I don't think this is needed at all

rotkehlchen/db/eth2.py
```diff
@@ -20,6 +21,10 @@
 log = RotkehlchenLogsAdapter(logger)
 
 WITHDRAWALS_RECHECK_PERIOD = 3 * 3600 * 1000  # 3 hours in milliseconds
+ETH_STAKING_EVENTS = (
```
LefterisJP, line 24: Isn't this duplicating almost all of `ETH_STAKING_EVENT_JOIN` which is defined in `db/filtering.py`?
Please do not do that.

LefterisJP (commented):
I don't think this is a good approach. Wrote few comments

docs/api.rst
yabirgb: It is now

rotkehlchen/chain/ethereum/modules/eth2/eth2.py
```diff
@@ -179,6 +179,7 @@ def get_balances(
     def get_details(
             self,
             addresses: list[ChecksumEvmAddress],
+            validator_indices: Optional[set[int]] = None,
```
LefterisJP, line 182: Why do you need to call and modify this? We discussed in the very beginning of this feature that we can't have the beaconchain api code run each time the filter changes. This needs to be split into different code.

And filtering should get out of this function

LefterisJP (commented):
more stuff

rotkehlchen/api/v1/schemas.py
```diff
@@ -3130,3 +3130,55 @@ class ClearAvatarsCacheSchema(Schema):
...
+        return {
+            'withdrawals_filter_query': withdrawals_filter_query,
+            'execution_filter_query': execution_filter_query,
+        }
+
+
+class EthStakingHistoryStatsDetails(AsyncIgnoreCacheQueryArgumentSchema):
+    """Schema for querying ethereum staking history details"""
+    location_labels = DelimitedOrNormalList(EvmAddressField, load_default=None)
+    validator_indices = DelimitedOrNormalList(fields.Integer(), load_default=None)
```
LefterisJP, line 3184: Probably nitpick but you could abstract the 

```python
    location_labels = DelimitedOrNormalList(EvmAddressField, load_default=None)
    validator_indices = DelimitedOrNormalList(fields.Integer(), load_default=None)
```

to its own schema interface and add it to both `EthStakingHistoryStatsDetails` and `EthStakingHistoryStatsProfit`

rotkehlchen/tests/unit/test_eth2.py
```diff
@@ -640,6 +641,119 @@ def test_get_validators_to_query_for_withdrawals(database):
...
+    dbevents = DBHistoryEvents(database)
+    dbeth2 = DBEth2(database)
+    vindex1 = 45555
+    vindex1_address = string_to_evm_address('0x0fdAe061cAE1Ad4Af83b27A96ba5496ca992139b')
+    vindex2 = 114543
+    vindex2_address = string_to_evm_address('0xF4fEae08C1Fa864B64024238E33Bfb4A3Ea7741d')
+    mev_builder_address = string_to_evm_address('0x690B9A9E9aa1C9dB991C7721a92d351Db4FaC990')
+    block_number = 15824493
+    tx_hash = deserialize_evm_tx_hash('0x8d0969db1e536969ba2e29abf8e8945e4304d49ae14523b66cbe9be5d52df804')  # noqa: E501
+    # tx_hash_2 doesn't exists. Only for test
```
LefterisJP, line 657:
```suggestion
    # tx_hash_2 doesn't exist. Only here for testing
```

rotkehlchen/tests/api/test_eth2.py
```diff
@@ -41,6 +41,14 @@
 @pytest.mark.parametrize('ethereum_modules', [['eth2']])
 def test_query_eth2_deposits_details_and_stats(rotkehlchen_api_server, ethereum_accounts):
     """This test uses real data and queries the eth2 details, deposits and daily stats"""
+    # first check that the endpoint for profit can be correctly queried
```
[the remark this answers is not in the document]
yabirgb: yes. I test the logic in a unit test but wanted to make sure that the endpoint worked correctly and added the call to this test that also tests the details (PUT) of the same endpoint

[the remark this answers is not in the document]
LefterisJP: Yes but the result is not tested at all. For existence of right keys etc. I guess unit test is fine, but it has to test a realistic scenario. Most of the times VCR sounds like a good choice for these no?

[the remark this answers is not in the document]
yabirgb: I tried VCR for this test and it was still slow that is why I didn't change it. I will check also the keys but the value is covered in the unit test for different values

LefterisJP (approved):
LGTM

PR #5954, opened 2023-04-28
Decode multiple ETH2 deposits

nebolax:
This PR makes transactions that perform multiple eth2 deposits (such as [this one](https://etherscan.io/tx/0x819fe4a07894cf044f5d8c63e5c1e2294e068d05bf91d9cfc3e7ae3e60528ae5)) properly decoded.

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/eth2/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/eth2/decoder.py
@@ -2,6 +2,7 @@ import logging
 from typing import Any
 
 from eth_utils import encode_hex
+from rotkehlchen.accounting.structures.balance import Balance
 
 from rotkehlchen.accounting.structures.eth2 import EthDepositEvent
 from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
@@ -10,6 +11,7 @@ from rotkehlchen.chain.ethereum.modules.curve.decoder import DEFAULT_DECODING_OU
 from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface
 from rotkehlchen.chain.evm.decoding.structures import DecoderContext, DecodingOutput
 from rotkehlchen.chain.evm.decoding.types import CounterpartyDetails, EventCategory
+from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.constants.assets import A_ETH
 from rotkehlchen.errors.misc import RemoteError
 from rotkehlchen.externalapis.beaconchain import BeaconChain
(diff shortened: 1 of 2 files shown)


LefterisJP (commented):
This also needs:

1. Rebase on top of the develop eth deposit changes.
2. Removing the frontend file changes

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
```diff
@@ -65,23 +66,27 @@ def _decode_eth2_deposit_event(self, context: DecoderContext) -> DecodingOutput:
                 event.event_type == HistoryEventType.SPEND and
                 event.event_subtype == HistoryEventSubType.NONE and
                 event.asset == A_ETH and
-                event.balance.amount == amount
+                event.balance.amount >= amount
```
yabirgb, line 69: for this contract it is the only method you can call. In this case I believe you can use

```
event.balance.amount == amount * len(context.all_logs)
```

have you checked it?

rotkehlchen/tests/unit/decoders/test_eth2.py
```diff
@@ -59,3 +59,53 @@ def test_deposit(database, ethereum_inquirer, ethereum_accounts):
...
+            asset=A_ETH,
+            balance=Balance(amount=FVal('0.010956672')),
+            location_label=user_address,
+            notes='Burned 0.010956672 ETH for gas',
+            counterparty=CPT_GAS,
+            address=None,
+        ), EthDepositEvent(
+            tx_hash=evmhash,
+            validator_index=55752,
+            sequence_index=1,
```
yabirgb, line 91: mmmm I believe the sequence index shouldn't be the same for all since it can break the logic. The assumption afail is that the event + sequence index is uniqe

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
```diff
@@ -65,23 +66,27 @@ def _decode_eth2_deposit_event(self, context: DecoderContext) -> DecodingOutput:
...
-            balance=old_event.balance,
-            depositor=old_event.location_label,  # type: ignore  # it's an address
-            extra_data=extra_data,  # only used if validator index is unknown
+                eth_deposit_event = EthDepositEvent(
+                    tx_hash=context.transaction.tx_hash,
+                    validator_index=validator_index,
+                    sequence_index=event.sequence_index,
+                    timestamp=event.timestamp,
+                    balance=Balance(amount=amount),
+                    depositor=event.location_label,  # type: ignore  # it's an address
```
yabirgb, line 77: I would suggest against using the type ignores. I would use `string_to_evm_address` and add to the condition that `location_label is not None` or do an assertion

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
nebolax: I would not do this due to 2 reasons:
1. It would limit us to a particular contract method. Others may have different events emitted.
2. Doing the check that you suggested will be possible only for the first deposit event. Since after each deposit event processed `event.balance.amount` is reduced, the amount changes. And trying to guess the right amount I think is not a good idea since it would also require to rely on the structure of logs emitted by contracts, as mentioned in (1).

nebolax: I would not do this due to 2 reasons:
1. It would limit us to a particular contract method. Others may have different events emitted.
2. Doing the check that you suggested will be possible only for the first deposit event. Since after each deposit event processed `event.balance.amount` is reduced, the amount changes. And trying to guess the right amount I think is not a good idea since it would also require to rely on the structure of logs emitted by contracts, as mentioned in (1).

rotkehlchen/tests/unit/decoders/test_eth2.py
nebolax: Yes, probably makes sense. Will fix

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
nebolax: This is the code written by Lefteris, but yeah, I can implement the suggestion

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
yabirgb: later I saw the play you make with the amounts. For 1. you are already limitating it since is a rule for one address. Only this contract will process this logic and this contract has only this method

rotkehlchen/tests/unit/decoders/test_eth2.py
```diff
@@ -53,9 +53,59 @@ def test_deposit(database, ethereum_inquirer, ethereum_accounts):
...
             depositor=user_address,
         ),
     ]
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('ethereum_accounts', [['0x3e5fd0244e13d82fC230f3Fc610bcd76b3c8217C']])
+def test_multiple_deposits(database, ethereum_inquirer, ethereum_accounts):
+    evmhash = deserialize_evm_tx_hash('0x819fe4a07894cf044f5d8c63e5c1e2294e068d05bf91d9cfc3e7ae3e60528ae5')  # noqa: E501
+    user_address = ethereum_accounts[0]
```
LefterisJP, line 68: it's VCRed anyway but just fyi, if you read the decoder you could have saved 3 (as many as the deposits) calls to beaconchain, by pre-adding the validators like I did here: https://github.com/rotki/rotki/blob/f49c115efa640cf8fed8266ed93c17e312fb873d/rotkehlchen/tests/integration/test_eth2.py#L26-L42

As the decoder, if you look at the code, first tries to match pubkey to validator by checking the DB

LefterisJP (approved):
Small comment in the test, plus needs a changelog entry. But will merge, and changelog can be added later.

PR #5956, opened 2023-04-28
fix: hides upgrade progress on error

kelsos:
Closes #(issue_number)

## Checklist

## Note

Previously on login error if the upgrade was running the UI would appear as being stuck to the upgrade progress.


frontend/app/src/pages/user/login/index.vue
```diff
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { type ComputedRef } from 'vue';
```
lukicenturi, line 2: we don't need this import anymore, do we?

lukicenturi (approved):
LGTM, just small comment, then we can merge

PR #5957, opened 2023-04-28
Add EthDepositEvent

LefterisJP:
Did not add any API/integration tests. Just two small unit tests which test the differences between this and other events. We should make this a habit to have more unit tests when possible.


LefterisJP:
all green. Merging.

PR #5960, opened 2023-04-28
Fix upgrade

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/db/drivers/gevent.py
+++ b/rotkehlchen/db/drivers/gevent.py
@@ -506,11 +506,11 @@ class DBConnection:
         db_name = self.connection_type.name.lower()
         with self.read_ctx() as cursor:
             cursor.execute('SELECT name, sql FROM sqlite_master WHERE type="table"')
-            tables_data_from_db = {}
+            tables_data_from_db: dict[str, tuple[str, str]] = {}
             for (name, raw_script) in cursor:
-                minimized_script = raw_script.replace(' ', '').replace('\n', '').replace('\'', '"')
+                minimized_script = raw_script.replace(' ', '').replace('\n', '').replace('\'', '"').lower()  # noqa: E501
                 table_properties = re.findall(
-                    pattern=r'CREATETABLE.*?\((.+)\)',
+                    pattern=r'createtable.*?\((.+)\)',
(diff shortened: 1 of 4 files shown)


rotkehlchen/db/drivers/gevent.py
```diff
@@ -506,7 +506,7 @@ def schema_sanity_check(self) -> None:
         db_name = self.connection_type.name.lower()
         with self.read_ctx() as cursor:
             cursor.execute('SELECT name, sql FROM sqlite_master WHERE type="table"')
-            tables_data_from_db = {}
+            tables_data_from_db: dict[str, tuple[str, str]] = {}
             for (name, raw_script) in cursor:
                 minimized_script = raw_script.replace(' ', '').replace('\n', '').replace('\'', '"')
```
nebolax, line 511: Hmm if to do case-insensitive comparisons, then maybe it will be better to have entire `minimized_schema.py` files lowercase and do
```suggestion
                minimized_script = raw_script.replace(' ', '').replace('\n', '').replace('\'', '"').lower()
```

here ? What do you think?

rotkehlchen/tests/db/test_db.py
```diff
@@ -1517,10 +1518,16 @@ def test_fresh_db_adds_version(user_data_dir, sql_vm_instructions_cb):
     assert int(query[0][0]) == ROTKEHLCHEN_DB_VERSION
 
 
-def test_db_schema_sanity_check(database, caplog):
+def test_db_schema_sanity_check(database: 'DBHandler', caplog) -> None:
     connection = database.conn
     # by default should run without problems
     connection.schema_sanity_check()
+    # verify that a the difference in text being upper or lower case doesn't affect the check
```
LefterisJP, line 1525:
```suggestion
    # verify that the text being upper or lower case doesn't affect the check
```

LefterisJP (approved):
LGTM but test addition has a typo and needs to be rewritten.

Also failed test, though unrelated seems like it can have a logic error if this can happen

```
>       assert result[account]['last_update_timestamp'] >= cur_time
E       assert 1682702464 >= 1682702465
```

Please keep those in mind and fix them in a subsequent PR.

PR #5962, opened 2023-04-28
Smol rpc lama fix. Hopefully temporary

LefterisJP:
It seems llama rpc is using some kind of load balancer and we sometimes hit a pruned node and sometimes a non-pruned one.

This can create problems as we expect a transaction hash to exist if we think rpc is non-pruned but next call we get an error.

With this, hopefully temporary, commit we always mark the llama rpc as pruned so we don't rely on them for anything historical.


yabirgb:
image

LefterisJP:
all tests passed here. Merging.

PR #5963, opened 2023-04-29
Add covalent api keys

yabirgb:
Adds a new set of api keys for covalent

### DIFF
--- a/rotkehlchen/externalapis/covalent.py
+++ b/rotkehlchen/externalapis/covalent.py
@@ -1,4 +1,5 @@
 import logging
+import random
 from json.decoder import JSONDecodeError
 from typing import Any, Optional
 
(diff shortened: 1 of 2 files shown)


rotkehlchen/externalapis/covalent.py
```diff
@@ -25,7 +26,11 @@
 CONST_RETRY = 1
 DATE_FORMAT_COVALENT = '%Y-%m-%dT%H:%M:%SZ'
 DEFAULT_API = 'covalent'
-KEY = 'ckey_cb70cd52c833489ea522f104e1a'
+COVALENT_KEYS = (
+    'ckey_cb70cd52c833489ea522f104e1a',
```
LefterisJP, line 30: shouldn't the used up key be removed? Otherwise at 1/3 of the cases we go the same problem.

Also please make sure to not use it in the CI at all. Or use different set of keys for testing. To be totally safe you can do 3 keys in __debug__ and 3 keys in production

Ofc as discussed this is just a temporary solution

rotkehlchen/externalapis/covalent.py
```diff
@@ -108,7 +112,8 @@ def _query(
         query_str += f'/{module}/'
 
         own_key = self._get_api_key()
-        query_str += f'?key={own_key if own_key else KEY}'
+        covalent_key = random.choice(COVALENT_KEYS)
+        query_str += f'?key={own_key if own_key else covalent_key}'
```
LefterisJP, line 116: no need to always run the randomization in every query

```suggestion
        if own_key is None:
            query_str += f'?key={random.choice(COVALENT_KEYS)}'
        else:
            query_str += f'?key={own_key}'
```

LefterisJP (approved):
LGTM

PR #5964, opened 2023-04-30
Periodically check for validator activation and update deposit event

LefterisJP:
It's possible that an ETH staking deposit event is decoded while the validator is waiting in the activation queue. In that case the validator has no validator index yet.

This commit is adding a periodic check (and tests for it) that if a deposit has no validator index and only a publickey, we ask beaconchain if it has been activated and if yes, the deposit data is updated and the full validator data (index + pubkey) is written in the DB.


LefterisJP:
all green. merging!

PR #5979, opened 2023-05-02
Curve aave withdrawals

nebolax:
This PR makes sure that withdrawals from curve pools with aave-wrapped tokens are decoded correctly.

Interesting thing is that incorrect decoding happened only for aave withdrawals. I manually checked and depositing into aave pool, depositing into compound pool and withdrawing from compound pool works correctly.

So this PR in the end touches only curve aave withdrawals, and the code diff ended up being quite small.

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/curve/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/curve/decoder.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Any, Callable, Literal, Optional
 from rotkehlchen.accounting.structures.evm_event import EvmProduct
 from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.assets.asset import Asset
+from rotkehlchen.chain.ethereum.modules.aave.constants import CPT_AAVE_V1, CPT_AAVE_V2
 from rotkehlchen.chain.ethereum.utils import asset_normalized_value
 from rotkehlchen.chain.evm.constants import ETH_SPECIAL_ADDRESS, ZERO_ADDRESS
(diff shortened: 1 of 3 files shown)


rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -183,26 +187,39 @@ def _decode_curve_remove_events(
...
+                    event.event_type == HistoryEventType.RECEIVE and
+                    event.event_subtype == HistoryEventSubType.NONE and
+                    event.location_label == user_or_contract_address
+                ):
+                    event.event_type = HistoryEventType.WITHDRAWAL
+                    event.event_subtype = HistoryEventSubType.REMOVE_ASSET
+                    event.counterparty = CPT_CURVE
+                    event.notes = notes
+                    withdrawal_events.append(event)
+                elif (  # Is already decoded by aave
```
LefterisJP, line 213: Shouldn't this check only be done if the pool is one of the aave pools in curve?

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -1461,3 +1461,64 @@ def test_3pool_remove_liquidity(ethereum_transaction_decoder, ethereum_accounts)
         ),
     ]
     assert events == expected_events
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('ethereum_accounts', [['0x60b0f1919cf4ee46d1A8D63428276512814de570']])
+def test_remove_from_aave_pool(ethereum_transaction_decoder, ethereum_accounts):
+    """
+    Test that if liquidity is removed from a pool with a(aave) tokens,
```
LefterisJP, line 1470: Please fix this typo in a subsequent PR

```suggestion
    Test that if liquidity is removed from a pool with aave atokens,
```

LefterisJP (approved):
lgtm

PR #5980, opened 2023-05-02
Curve zap logic tweak

nebolax:
Previously for certain curve zap transactions there was an issue that decoding logic was called twice. So the outcome was correct, but there were unneeded extra calls to our decoding methods. This PR fixes it.

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/curve/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/curve/decoder.py
@@ -200,7 +200,7 @@ class CurveDecoder(DecoderInterface, ReloadableDecoderMixin):
                 event.counterparty = CPT_CURVE
                 event.notes = f'Return {event.balance.amount} {crypto_asset.symbol}'
                 return_event = event
-            if (  # Withdraw receive asset
+            elif (  # Withdraw receive asset
                 tx_log.address in self.curve_pools and
                 transaction.from_address == user_or_contract_address
             ):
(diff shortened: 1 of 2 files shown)


rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -283,13 +294,22 @@ def _decode_curve_deposit_events(
                 event.counterparty = CPT_CURVE
                 event.notes = f'Deposit {event.balance.amount} {crypto_asset.symbol} in curve pool {tx_log.address}'  # noqa: E501
                 deposit_events.append(event)
+            elif (
+                event.event_type == HistoryEventType.DEPOSIT and
+                event.event_subtype == HistoryEventSubType.DEPOSIT_ASSET and
+                event.counterparty == CPT_CURVE
+            ):
+                already_processed = True
```
yabirgb, line 303: This is incorrect. What happens is that we iterate multiple times over the decoded events. This happens because we try to decode events every time they match a decoding rule in 

```
decoding_output = self.decode_by_address_rules(tx_log, transaction, events, tx_receipt.logs, action_items)  # noqa: E501
```

So here this function is first called with for example 2 events and then called again with 3 events in`decoded_events`. You can check in `test_3pool_add_liquidity`. 

if you exit early as you are doing then you will miss the `_sort_events` logic that I believe is needed in some cases. You can check it yourself.

I'm not really sure what guided you to add this piece of code here

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -494,6 +514,7 @@ def _handle_post_decoding(
                     transaction=transaction,
                     decoded_events=decoded_events,
                 )
+                break  # We only need to handle zap contracts once
```
yabirgb, line 517: this is valid as you don't need to iterate over all the events for what is getting executed

rotkehlchen/chain/ethereum/modules/curve/decoder.py
yabirgb: What I'm not sure is why no test failed after your change. I believe as they are there is no deposit + zap or the conditions that require the logic happening after

rotkehlchen/chain/ethereum/modules/curve/decoder.py
nebolax: Do you ask for any change or is it just a comment?

rotkehlchen/chain/ethereum/modules/curve/decoder.py
nebolax: > if you exit early as you are doing then you will miss the _sort_events logic that I believe is needed in some cases. You can check it yourself.

If you check closer, 

```python
if already_processed is True:
```

this condition happens only inside

```python
if receive_event is None or len(deposit_events) == 0:
```

so it does not affect sorting.

-------------

The only thing that it may early-exit before is `return DecodingOutput(matched_counterparty=CPT_CURVE)`, which could disable the trigger of the needed post decoding rule for zaps. But all is fine here since `already_processed` can be `True` only if it is the second call this curve decoding method during processing a transaction, and in this case, if it is a zap call, `return DecodingOutput(matched_counterparty=CPT_CURVE)` will happen during the first call and the zaps logic will be handled properly.

nebolax: I can also move this check

```python
if already_processed is True:
```

after
```python
if user_or_contract_address in CURVE_DEPOSIT_CONTRACTS and len(deposit_events) > 0:
                return DecodingOutput(matched_counterparty=CPT_CURVE)
```

to be 100% sure that the post decoding rule trigger is not missed.

LefterisJP (approved):
LGTM

PR #5981, opened 2023-05-02
Fix withdrawals query logic and change some default settings post-withdrawals

LefterisJP:


LefterisJP:
Finally all green. Merging.

PR #5982, opened 2023-05-03
feat: improve ens avatar and blockie

lukicenturi:
Closes #(issue_number)

## Checklist


frontend/app/src/components/display/EnsAvatar.vue
```diff
@@ -28,17 +28,23 @@ const { getBlockie } = useBlockie();
...
 
 <template>
   <v-lazy :class="css.wrapper">
     <div>
-      <v-img v-if="!success" :src="getBlockie(address)" />
+      <v-img v-if="!avatarUrl || failed" :src="getBlockie(address)" />
+      <v-img
+        v-if="avatarUrl && !success && !failed"
+        :src="getBlockie(address, false)"
+      />
```
kelsos, line 41: I didn't mean that. I meant that it would be preferable to have a placeholder until loading either completes or fails.

If it completes successfully nothing extra gets done.
If it fails it generates and caches the blockie and it shows it.
If it is loading it shows a placeholder.

This way the blocky will always get generated to display for a couple of seconds which is even worse than before, because now we generated it every single time since now it is not cached.

lukicenturi: Okay my bad for misunderstood, updated now

kelsos (approved):

PR #5987, opened 2023-05-03
Reduces number of loading icons in history and extracts common query status components

kelsos:
Closes #(issue_number)

## Checklist


frontend/app/src/components/history/LocationDisplay.vue
```diff
@@ -23,18 +23,22 @@ const { identifier, detailPath } = toRefs(props);
 const { locationData } = useLocations();
 const location = locationData(identifier);
 
+const name = computed(() => get(location)?.name ?? '')
```
tewshi, line 26: @kelsos

frontend/app/src/components/history/LocationIcon.vue
```diff
@@ -1,11 +1,10 @@
...
 import { type TradeLocationData } from '@/types/history/trade/location';
 
 const props = defineProps({
   item: {
     required: true,
-    type: Object as PropType<TradeLocationData>
+    type: [Object, String] as PropType<TradeLocationData | string | null>
   },
   horizontal: { required: false, type: Boolean, default: false },
   icon: { required: false, type: Boolean, default: false },
```
tewshi, line 10: Thought we're moving to the defineProps<{...}>() pattern

frontend/app/src/composables/history/events/events-query-status.ts
```diff
@@ -3,8 +3,8 @@ import { type HistoryEventsQueryData } from '@/types/websocket-messages';
```
lukicenturi, line 3: I must be forgot to do this. Can you move this file to inside `/composables/history/events/query-status/`

frontend/app/src/composables/history/events/tx-query-status.ts
```diff
@@ -10,8 +10,8 @@ export const useTransactionQueryStatus = (
 ) => {
```
lukicenturi, line 10: this file too

frontend/app/src/components/history/LocationDisplay.vue
kelsos: Yes I saw the linting failure :)

We need some pre-commit / lint-staged support. I tried that last week but it is a bit more complicated with our setup here.

tewshi (approved):
lgtm!

frontend/app/src/composables/locations.ts
```diff
@@ -56,37 +23,35 @@ export const useLocations = createSharedComposable(() => {
...
-        identifier: assetId,
-        image: getAssetImageUrl(assetId),
-        detailPath: `${Routes.ACCOUNTS_BALANCES_BLOCKCHAIN}#blockchain-balances-${id}`
-      };
-    }
-
-    const locationFound = get(tradeLocations).find(
-      location => location.identifier === id
-    );
+  const locationData = (
```
lukicenturi, line 26: Btw I wonder why the base function here is the one that returns computed (`locationData`)
and the `getLocationData` need to call that function
Why not the other way around? Since we can reduce the computed call. Is it some ref issue?

lukicenturi: Okay I think I got it, if I do the other way, the `getChainInfoById` won't be reactive, since it's not inside computed directly

Issue #5975, opened 2023-05-02
Websocket from backend to trigger query balances in frontend


lukicenturi:
## Problem Definition

The issue is that there are some balances that are not queried like tokens, knowing beforehand the token that needs to be queried
What we do is to use the decoded events to say "you deposited tokens in this contract so we have to query it in order to get your deposit balance". This happens for curve and convex atm.
The problem is that in a fresh start since the events are not decoded those balances won’t appear

## Solution proposed
If at some point by navigating to the history events or by the background task that we run, we detect any of those contracts, then we should send a message to the frontend, and the frontend need to `query balances` again.

LefterisJP:
This is in develop and will be in 1.28

PR #5990, opened 2023-05-04
Autorefresh balances

nebolax:
Backend aspect of #5975.

Adds logic on backend that makes sure that if a new curve gauge / convex staking event was decoded, a corresponding websocket message is sent to the frontend.

### DIFF
--- a/docs/websockets.rst
+++ b/docs/websockets.rst
@@ -246,3 +246,23 @@ Finally the backend provides more granular information to know what interval of 
 - ``location``(Optional): When the same ``event_type`` can be queried in multiple locations this helps to differentiate them.
 - ``name``(Optional): If multiple appearences of the same location are possible it will differentiate each one of them.
 - ``period``: Time range that is being queried.
+
+
+Request a refresh of balances
+============================
+
+If at some point backend detects that balances need to be refreshed, it will send this message to the frontend.
+
+::
+
+    {
+        "type": "refresh_balances",
+        "data": {
+            "type": "blockchain_balances",
+            "blockchain": "optimism"
+        }
+    }
+
+
+- ``type``: Balances section that needs a refresh. Valid values are: ``blockchain_balances``.
+- ``blockchain``: Returned only for section: ``blockchain_balances``. The blockchain for which balances need to be refreshed. Valid values are: ``optimism``, ``eth``.
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -4084,7 +4084,7 @@ class RestAPI():
         """
         evm_manager = self.rotkehlchen.chains_aggregator.get_evm_manager(evm_chain)
         try:
-            transaction, tx_receipt = evm_manager.transactions.add_transaction_by_hash(
+            evm_manager.transactions.add_transaction_by_hash(
                 tx_hash=tx_hash,
                 associated_address=associated_address,
             )
@@ -4103,7 +4103,11 @@ class RestAPI():
                 ),
                 status_code=status_code,
             )
-        evm_manager.transactions_decoder.decode_transaction(transaction=transaction, tx_receipt=tx_receipt)  # noqa: E501
+
+        evm_manager.transactions_decoder.decode_transaction_hashes(
+            ignore_cache=True,
+            tx_hashes=[tx_hash],
+        )
         return OK_RESULT
 
     def _get_exchange_staking_or_savings_history(
--- a/rotkehlchen/chain/ethereum/decoding/decoder.py
+++ b/rotkehlchen/chain/ethereum/decoding/decoder.py
@@ -157,10 +157,7 @@ class EthereumTransactionDecoder(EVMTransactionDecoder):
 
     # -- methods that need to be implemented by child classes --
 
-    def _enrich_protocol_tranfers(
-            self,
-            context: EnricherContext,
-    ) -> TransferEnrichmentOutput:
+    def _enrich_protocol_tranfers(self, context: EnricherContext) -> TransferEnrichmentOutput:
         for enrich_call in self.rules.token_enricher_rules:
             try:
                 transfer_enrich: TransferEnrichmentOutput = enrich_call(context)
--- a/rotkehlchen/chain/ethereum/modules/convex/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/convex/decoder.py
@@ -23,7 +23,6 @@ from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
 from rotkehlchen.chain.evm.decoding.constants import ERC20_OR_ERC721_TRANSFER
 from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface
 from rotkehlchen.chain.evm.decoding.structures import (
-    DEFAULT_DECODING_OUTPUT,
     FAILED_ENRICHMENT_OUTPUT,
     DecoderContext,
     DecodingOutput,
@@ -48,6 +47,7 @@ class ConvexDecoder(DecoderInterface):
     def _decode_convex_events(self, context: DecoderContext) -> DecodingOutput:
         amount_raw = hex_or_bytes_to_int(context.tx_log.data[0:32])
         interacted_address = hex_or_bytes_to_address(context.tx_log.topics[1])
+        found_event_modifying_balances = False
 
         for event in context.decoded_events:
             try:
@@ -66,6 +66,7 @@ class ConvexDecoder(DecoderInterface):
                 event.event_type == HistoryEventType.SPEND and
                 event.event_subtype == HistoryEventSubType.NONE
             ):
+                found_event_modifying_balances = True
                 if event.address == ZERO_ADDRESS:
                     event.event_subtype = HistoryEventSubType.RETURN_WRAPPED
                     event.counterparty = CPT_CONVEX
@@ -106,6 +107,7 @@ class ConvexDecoder(DecoderInterface):
                 if context.tx_log.topics[0] in WITHDRAWAL_TOPICS:
                     event.event_type = HistoryEventType.WITHDRAWAL
                     event.counterparty = CPT_CONVEX
+                    found_event_modifying_balances = True
                     if context.tx_log.address in CONVEX_POOLS:
                         event.notes = f'Withdraw {event.balance.amount} {crypto_asset.symbol} from convex {CONVEX_POOLS[context.tx_log.address]} pool'  # noqa: E501
                         event.product = EvmProduct.GAUGE
@@ -119,11 +121,12 @@ class ConvexDecoder(DecoderInterface):
                 elif context.tx_log.topics[0] in REWARD_TOPICS:
                     event.event_subtype = HistoryEventSubType.REWARD
                     event.counterparty = CPT_CONVEX
+                    found_event_modifying_balances = True
                     if context.tx_log.address in CONVEX_POOLS:
                         event.notes = f'Claim {event.balance.amount} {crypto_asset.symbol} reward from convex {CONVEX_POOLS[context.tx_log.address]} pool'  # noqa: E501
                     else:
                         event.notes = f'Claim {event.balance.amount} {crypto_asset.symbol} reward from convex'  # noqa: E501
-        return DEFAULT_DECODING_OUTPUT
+        return DecodingOutput(refresh_balances=found_event_modifying_balances)
 
     @staticmethod
     def _maybe_enrich_convex_transfers(context: EnricherContext) -> TransferEnrichmentOutput:
--- a/rotkehlchen/chain/ethereum/modules/curve/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/curve/decoder.py
@@ -508,6 +508,7 @@ class CurveDecoder(DecoderInterface, ReloadableDecoderMixin):
         provider = hex_or_bytes_to_address(context.tx_log.topics[1])
         gauge_address = context.tx_log.address
         raw_amount = hex_or_bytes_to_int(context.tx_log.data)
+        found_event_modifying_balances = False
         for event in context.decoded_events:
             crypto_asset = event.asset.resolve_to_crypto_asset()
             if (
@@ -517,6 +518,7 @@ class CurveDecoder(DecoderInterface, ReloadableDecoderMixin):
             ):
                 event.counterparty = CPT_CURVE
                 event.product = EvmProduct.GAUGE
+                found_event_modifying_balances = True
                 if context.tx_log.topics[0] == GAUGE_DEPOSIT:
                     event.event_type = HistoryEventType.DEPOSIT
                     event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
@@ -526,7 +528,7 @@ class CurveDecoder(DecoderInterface, ReloadableDecoderMixin):
                     event.event_subtype = HistoryEventSubType.REMOVE_ASSET
                     event.notes = f'Withdraw {event.balance.amount} {crypto_asset.symbol} from {gauge_address} curve gauge'  # noqa: E501
 
-        return DEFAULT_DECODING_OUTPUT
+        return DecodingOutput(refresh_balances=found_event_modifying_balances)
 
     def _handle_post_decoding(
             self,
--- a/rotkehlchen/chain/evm/decoding/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/decoder.py
@@ -15,6 +15,7 @@ from rotkehlchen.accounting.structures.types import (
     HistoryEventSubType,
     HistoryEventType,
 )
+from rotkehlchen.api.websockets.typedefs import WSMessageType
 from rotkehlchen.assets.asset import AssetWithOracles, EvmToken
 from rotkehlchen.assets.utils import TokenSeenAt, get_or_create_evm_token
 from rotkehlchen.chain.ethereum.utils import token_normalized_value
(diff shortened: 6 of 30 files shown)


docs/websockets.rst
```diff
@@ -245,4 +245,24 @@ Finally the backend provides more granular information to know what interval of
...
+If at some point backend detects that balances need to be refreshed, it will send this message to the frontend.
+
+::
+
+    {
+        "type": "refresh_balances",
+        "data": {
+            "section": "blockchain_balances",
+            "blockchain": "optimism"
+        }
```
yabirgb, line 263: we discussed this in the call this morning. Should we add the address that needs to be refreshed now in the message?

docs/websockets.rst
```diff
@@ -245,4 +245,24 @@ Finally the backend provides more granular information to know what interval of
...
+    }
+
+
+- ``section``: Balances section that needs a refresh.
```
yabirgb, line 267: list the possible values for this field. Currently one but better keep a list in the docs

docs/websockets.rst
```diff
@@ -245,4 +245,24 @@ Finally the backend provides more granular information to know what interval of
...
+- ``blockchain``: Returned only for section: ``blockchain_balances``. The blockchain for which balances need to be refreshed.
```
yabirgb, line 268: same, can you list the values. Requesting it because in some places we use `eth` and in other `ethereum`. To remove this ambiguity

rotkehlchen/api/rest.py
```diff
@@ -2882,6 +2885,10 @@ def decode_evm_transactions(
         else:  # no break in the for loop, success
             result = True
 
+        for evm_chain, refresh_balances in refresh_balances_dict.items():
+            self.rotkehlchen.chains_aggregator.get_evm_manager(evm_chain).transactions_decoder.post_process(
+                refresh_balances=refresh_balances,
+            )
```
yabirgb, line 2891:
```suggestion
            chain_manager.transactions_decoder.post_process(
                refresh_balances=refresh_balances,
            )
```

rotkehlchen/chain/ethereum/modules/convex/decoder.py
```diff
@@ -48,6 +47,7 @@ class ConvexDecoder(DecoderInterface):
     def _decode_convex_events(self, context: DecoderContext) -> DecodingOutput:
         amount_raw = hex_or_bytes_to_int(context.tx_log.data[0:32])
         interacted_address = hex_or_bytes_to_address(context.tx_log.topics[1])
+        found_relevant_event = False
```
yabirgb, line 50: add a comment or lets use a better name. What about `found_event_modifying_balances`

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -476,6 +476,7 @@ def _decode_curve_gauge_events(self, context: DecoderContext) -> DecodingOutput:
         provider = hex_or_bytes_to_address(context.tx_log.topics[1])
         gauge_address = context.tx_log.address
         raw_amount = hex_or_bytes_to_int(context.tx_log.data)
+        found_relevant_event = False
```
yabirgb, line 479: same comment as before. We should use good names for the variables since someone that is not you or me will see this name and won't know for what this variable has found relevant events

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -437,13 +444,17 @@ def get_and_decode_undecoded_transactions(
                 limit=limit,
                 addresses=addresses,
             )
-            self.decode_transaction_hashes(ignore_cache=False, tx_hashes=hashes)
+            _, refresh_balances = self.decode_transaction_hashes(
+                ignore_cache=False,
+                tx_hashes=hashes,
+            )
+            self.post_process(refresh_balances=refresh_balances)
```
yabirgb, line 451: If I read correctly the code the core of the decoding process always happens in `decode_transaction_hashes` (this function in the case of the task manager `get_and_decode_undecoded_transactions` and `decode_transaction_hashes` for `rest.py`.

So instead of further propagating the `refresh_balances` why not set it to `False` (you already do this) at the beggining of `decode_transaction_hashes` and before exiting `decode_transaction_hashes` you just execute

```
self.post_process(refresh_balances=refresh_balances)
```

Why this:

1. the decoding logic and what is needed to be executed is contained in the decoder class. Right now the `rest.py` file is required to execute logic that it shouldn't know about
2. You don't propagate values outside of where they are needed

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -479,20 +491,23 @@ def decode_transaction_hashes(
...
                 tx_receipt=receipt,
                 ignore_cache=ignore_cache,
-            ))
+            )
+            events.extend(new_events)
+            if new_refresh_balances is True:
+                refresh_balances = True
 
-        return events
+        return events, refresh_balances
```
yabirgb, line 503: Check my previous comment. I believe you don't need to return here

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -800,12 +815,33 @@ def _maybe_decode_erc20_721_transfer(
...
+        decoded. Currently may only send a websocket message to the frontend to refresh balances.
+
+        The caller of decode_transactions should call this method after a batch of transactions
+        has been decoded.
+        """
+        if refresh_balances is True:
+            self.msg_aggregator.add_message(
+                message_type=WSMessageType.REFRESH_BALANCES,
+                data={
+                    'section': 'blockchain_balances',
```
yabirgb, line 836: Shouldn't this be an enum serialized?

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -517,7 +520,7 @@ def test_curve_remove_liquidity_with_internal(database, ethereum_transaction_dec
     with database.user_write() as cursor:
         dbevmtx.add_evm_transactions(cursor, [transaction], relevant_address=None)
         dbevmtx.add_evm_internal_transactions(cursor, [internal_tx], relevant_address=location_label)  # noqa: E501
-    events = ethereum_transaction_decoder.decode_transaction(
+    events, _ = ethereum_transaction_decoder.decode_transaction(
```
yabirgb, line 523: is there any test where the value for the refresh is true?

rotkehlchen/tests/unit/decoders/test_convex.py
```diff
@@ -284,7 +284,7 @@ def test_cvxcrv_get_reward(database, ethereum_inquirer, eth_transactions):
         ethereum_inquirer=ethereum_inquirer,
         transactions=eth_transactions,
     )
-    events = decoder.decode_transaction(transaction=transaction, tx_receipt=receipt)
+    events, _ = decoder.decode_transaction(transaction=transaction, tx_receipt=receipt)
```
yabirgb, line 287: shouldn't there be a test for which the value is true? assert it

rotkehlchen/tests/unit/decoders/test_balancer.py
```diff
@@ -230,7 +230,7 @@ def test_balancer_v1_join(database, ethereum_manager, eth_transactions):
             ethereum_inquirer=ethereum_manager.node_inquirer,
             transactions=eth_transactions,
         )
-    events = decoder.decode_transaction(transaction=transaction, tx_receipt=receipt)
+    events, _ = decoder.decode_transaction(transaction=transaction, tx_receipt=receipt)
```
yabirgb, line 233: I fear that it will happen as with the decoding output that we end up adding return arguments and having to refactor all the code. Why not to create a namedtuple and isntead of doing this around the code just

```suggestion
   events = decoder.decode_transaction(transaction=transaction, tx_receipt=receipt).events
```

rotkehlchen/chain/ethereum/decoding/decoder.py
```diff
@@ -166,12 +170,12 @@ def _enrich_protocol_tranfers(self, context: EnricherContext) -> Optional[str]:
                     f'{context.event.asset}. {str(e)}',
                 )
                 # Don't try other rules since all of them will fail to resolve the asset
-                return None
+                return DEFAULT_ENRICHMENT_OUTPUT
 
-            if transfer_enrich.matched_counterparty is not None:
-                return transfer_enrich.matched_counterparty
+            if transfer_enrich != DEFAULT_ENRICHMENT_OUTPUT:
+                return transfer_enrich
```
yabirgb, line 176: but this doesn't fix the logic unless you change the `transfer_enrich` in the modules right? it will continue to loop even if the transaction was enriched.

Is either to set the counterparty (lefteris wanted to avoid it) or add a flag `enriched=True` and break if True

yabirgb (commented):
A few comments

docs/websockets.rst
nebolax: iirc Lefteris proposed this, but in the end we agreed to not add it, since right now it is not useful anyway

rotkehlchen/api/rest.py
nebolax: No, this won't work. It will take the last used chain aggregator, and it will not respect the evm chain

rotkehlchen/chain/ethereum/modules/convex/decoder.py
nebolax: okay, will change the name

rotkehlchen/chain/evm/decoding/decoder.py
nebolax: Yes, probably it makes sense. I agree that the way it is currently made is bad.

The reason why I did it the way I did it is because things like `decode_transaction` are also called separately, not only in `decode_transaction_hashes`, so was a bit tricky there to automatically handle post processing. But I think I have an idea about it.

rotkehlchen/chain/evm/decoding/decoder.py
nebolax: Could be, but I would not overkill it since atm it's a single place where this ws message is sent

rotkehlchen/tests/unit/decoders/test_curve.py
nebolax: Check below

rotkehlchen/tests/unit/decoders/test_balancer.py
nebolax: It would make sense, but I think not right now, since atm it's only 2 returned values. For when it grows -- yes, sure, will make sense to put it in a separate class.

rotkehlchen/chain/ethereum/decoding/decoder.py
nebolax: This change I made to handle the case if `refresh_balances` is true. As for fixing extra running enrichers I was not sure if this needs fixing. But yeah, a flag can be added

rotkehlchen/api/rest.py
```diff
@@ -4119,7 +4119,11 @@ def add_evm_transaction_by_hash(
                 ),
                 status_code=status_code,
             )
-        evm_manager.transactions_decoder.decode_transaction(transaction=transaction, tx_receipt=tx_receipt)  # noqa: E501
+
+        evm_manager.transactions_decoder.decode_transaction_hashes(
+            ignore_cache=True,
```
yabirgb, line 4124: What happened before? We were doing nothing if it was already decoded?

rotkehlchen/tests/unit/decoders/test_convex.py
```diff
@@ -88,8 +90,18 @@ def test_convex_pools(ethereum_inquirer):
 
 
 @pytest.mark.vcr()
+@pytest.mark.parametrize('have_decoders', [True])
+@pytest.mark.parametrize('legacy_messages_via_websockets', [True])
 @pytest.mark.parametrize('ethereum_accounts', [[string_to_evm_address('0xC960338B529e0353F570f62093Fd362B8FB55f0B')]])  # noqa: E501
-def test_booster_deposit(database, ethereum_inquirer, ethereum_accounts):
+def test_booster_deposit(
+        rotkehlchen_api_server: 'APIServer',
```
yabirgb, line 97: why do you need the server here?

rotkehlchen/api/rest.py
nebolax: No, it was calling `decode_transaction` that always performed a detection

rotkehlchen/tests/unit/decoders/test_convex.py
nebolax: Due to this https://github.com/rotki/rotki/blob/2ed321bc04b9d1ed15e8f5797cff62fb6a136001/rotkehlchen/tests/utils/database.py#L76.

websocket connection feature needs `rotkehlchen_api_server` fixture, so we can't use a separate database fixture

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -348,21 +350,24 @@ def run_all_post_decoding_rules(
         assert self._check_correct_types(decoded_events)
         return decoded_events
 
-    def decode_transaction(
+    def _decode_transaction(
```
LefterisJP, line 353: Why are you doing this? Why add the underscore here and make this a "private" function, and force such a big change in lines in an already big PR?

This function is called mostly in tests but over 80 times. What do we gain by you adding this underscore at this point, in this PR?

It's just adding another 80 lines for me to check.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -348,21 +350,24 @@ def run_all_post_decoding_rules(
...
             self,
             transaction: EvmTransaction,
             tx_receipt: EvmTxReceipt,
-    ) -> list['EvmEvent']:
+    ) -> tuple[list['EvmEvent'], bool]:
```
LefterisJP, line 357: you are adding a new return value to the function. What is it for? Why have you not changed the docstring?

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -479,20 +488,24 @@ def decode_transaction_hashes(
...
         return events
 
-    def get_or_decode_transaction_events(
+    def _get_or_decode_transaction_events(
             self,
             transaction: EvmTransaction,
             tx_receipt: EvmTxReceipt,
             ignore_cache: bool,
-    ) -> list['EvmEvent']:
+    ) -> tuple[list['EvmEvent'], bool]:
```
LefterisJP, line 508: Same here. Explain what this new return value does in the docstring

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -800,12 +812,33 @@ def _maybe_decode_erc20_721_transfer(
...
+        decoded. Currently may only send a websocket message to the frontend to refresh balances.
+
+        The caller of decode_transactions should call this method after a batch of transactions
+        has been decoded.
+        """
+        if refresh_balances is True:
+            self.msg_aggregator.add_message(
+                message_type=WSMessageType.REFRESH_BALANCES,
+                data={
+                    'section': 'blockchain_balances',
```
LefterisJP, line 833: Shouldn't this be type?

```suggestion
                    'type': 'blockchain_balances',
```

rotkehlchen/chain/evm/decoding/structures.py
```diff
@@ -67,8 +68,11 @@ class TransferEnrichmentOutput(NamedTuple):
     matched_counterparty is optionally set if needed for enrichment rules
     and is used in post-decoding rules like in the case of balancer
     """
+    enriched: bool = False
     matched_counterparty: Optional[str] = None
+    refresh_balances: bool = False
```
LefterisJP, line 73: In the docstring above you need to explain what the two new arguments are doing here. Separate the docstring in bullet points to explain each argument

rotkehlchen/chain/evm/decoding/structures.py
```diff
@@ -67,8 +68,11 @@ class TransferEnrichmentOutput(NamedTuple):
     matched_counterparty is optionally set if needed for enrichment rules
     and is used in post-decoding rules like in the case of balancer
     """
+    enriched: bool = False
```
LefterisJP, line 71: This argument makes little sense. I already wrote about it in discord: https://discord.com/channels/657906918408585217/1080783409640648765/1104155914367418429

If the only reason for the existence of this is to save a few characters when typing undo this change.

> so instead of having to do always TransferEnrichmentOutput(matched_counterparty=MY_COUNTERPARTY) we return now SUCCESSFUL_ENRICHMENT_OUTPUT

This is not a good reason to add a whole new argument to a context that is passed around in memory across the entrie decoder stack, multiplied by an ever increasing number of decoders. This should be kept as lean as possible.

rotkehlchen/tests/unit/decoders/test_convex.py
LefterisJP: Then this is not a unit test. You pointed at a test assertion. This tells me this is fixable and you need to find a way to  make this into a light unit test.

LefterisJP (commented):
Okay so this PR is very hard to review. Already left enough comments and will stop the review here.

1. Split it into 2 PRs. Yabir already hinted at that, but he should have insisted. It's very hard to review.
2. You have a lot of unecessary changes adding size in an already big review. Don't complicate things by adding an `_` in already existing functions forcing you to make changes in functions in over 80 lines across all tests.
3. I think the approach with adding a boolean in the EnrichmentOutput is not needed, assuming I understood the reason correctly.

Generally please keep things easy to review.

rotkehlchen/chain/evm/decoding/decoder.py
nebolax: I guess you have already marked the files that are touched by this change as "Viewed", so in order to not reset that "Viewed" status I will keep it as it is right now.

rotkehlchen/chain/evm/decoding/structures.py
```diff
@@ -53,21 +53,31 @@ class EnricherContext(DecoderBasicContext):
...
+    added to the list of other decoded events.
+    - action_items is a list of actions to be performed later automatically or to be passed
+    in further decoding methods.
+    - matched_counterparty is optionally set if needed for decoder rules that matched
     and is used in post-decoding rules like in the case of balancer
     """
     event: Optional['EvmEvent'] = None
     action_items: list[ActionItem] = []
     matched_counterparty: Optional[str] = None
+    refresh_balances: bool = False
```
LefterisJP, line 67: You did not add the explanation for refresh balances in this docstring. Only on the next one. Please don't make me have to spell out every needed change.

rotkehlchen/chain/evm/decoding/decoder.py
LefterisJP: Not really. As I said I stopped reviewing the PR before due to the size of the irrelevant diff. But whatever I am not gonna let this PR block the frontend and the release anymore.

Please, this is something to keep in mind for future PRs. Keep things short and simple. And do not add any unecessary changes like this in a PR.

rotkehlchen/tests/fixtures/messages.py
```diff
@@ -1,13 +1,43 @@
...
+
+    def broadcast(  # pylint: disable=unused-argument
+            self,
+            message_type: 'WSMessageType',
+            to_send_data: Union[dict[str, Any], list[Any]],
+            **kwargs: Any,
+    ) -> None:
+        self.messages.append(MockedWsMessage(type=message_type, data=to_send_data))
+
+    def pop_message(self) -> Optional[MockedWsMessage]:
```
LefterisJP, line 25: I would add a function to clear all messages too. Since you added it also as a session scope fixture, it will probably be needed for those tests that use it in a session scope. Otherwise it will have leftover messages from other tests whenever it's used in a test

rotkehlchen/tests/unit/decoders/test_convex.py
```diff
@@ -99,6 +106,17 @@ def test_booster_deposit(database, ethereum_inquirer, ethereum_accounts):
         database=database,
         tx_hash=evmhash,
     )
+    mocked_notifier = database.msg_aggregator.rotki_notifier
+    mocked_notifier.pop_message()  # detected stETH token
```
LefterisJP, line 110: test for those. Why just add comments? If this is what it's supposed to be then you should test for them.

LefterisJP (commented):
Small comments. But a lot of tests failed.

PR #5991, opened 2023-05-04
Add hidden flag to returned events

LefterisJP:
- Potentially adds a hidden key to each returned event if it should be hidden by the frontend's UI
- All flags for the history events are now optional. Since they were all booleans if they exist, they are set to true. If missing they are false and as such we reduce amount of data moving between backend and frontend.

### DIFF
--- a/rotkehlchen/accounting/structures/evm_event.py
+++ b/rotkehlchen/accounting/structures/evm_event.py
@@ -162,14 +162,17 @@ class EvmEvent(HistoryBaseEntry):
             self,
             customized_event_ids: list[int],
             ignored_ids_mapping: dict[ActionType, set[str]],
+            hidden_event_ids: list[int],
             grouped_events_num: Optional[int] = None,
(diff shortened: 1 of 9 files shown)


rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -162,14 +162,17 @@ def serialize_for_api(
...
     ) -> dict[str, Any]:
         result = super().serialize_for_api(
             customized_event_ids=customized_event_ids,
             ignored_ids_mapping=ignored_ids_mapping,
+            hidden_event_ids=hidden_event_ids,
             grouped_events_num=grouped_events_num,
         )
-        result['has_details'] = self.has_details()
+        if self.has_details():
+            result['has_details'] = True
```
yabirgb, line 175: Should we?

```suggestion
        if self.has_details() is True:
            result['has_details'] = True
```

PR #5992, opened 2023-05-04
Add exchange mappings and fix kraken values

yabirgb:
- Adds a new set of mappings for exchanges
- Changes the kraken values to use positive amounts to make the usage consistent with evm events and staking events
- Changes sushiswap icon as requested by luki
- Fix instant swaps in exchanges


rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -185,33 +187,131 @@ def _update_ens_mappings_schema(write_cursor: 'DBCursor') -> None:
...
+    write_cursor.execute(
+        'SELECT identifier, amount, usd_value FROM history_events WHERE location="B" AND '
+        'type="withdrawal"',
+    )
+    for event_row in write_cursor:
+        amount = FVal(event_row[1])
+        usd_value = FVal(event_row[2])
+        update_tuples.append(
+            (
+                str(-amount) if amount != ZERO else str(amount),
```
nebolax, line 262: Same q for zeros. Why do the need special treatment?

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -185,33 +187,131 @@ def _update_ens_mappings_schema(write_cursor: 'DBCursor') -> None:
...
+
+    if len(update_tuples) != 0:
+        write_cursor.executemany(
+            'UPDATE history_events SET type=?, subtype=?, amount=?, usd_value=? WHERE identifier=?',  # noqa: E501
+            update_tuples,
+        )
+
+    log.debug('Fixing kraken withdrawals')
+    update_tuples = []
+    write_cursor.execute(
```
nebolax, line 253: Wouldn't it be possible to do this change for withdrawals in a single sql query? So check if the amount string and usd value string start with a minus, and if they do, remove the minus.

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -185,33 +187,131 @@ def _update_ens_mappings_schema(write_cursor: 'DBCursor') -> None:
...
+        'SELECT identifier, event_identifier, type, amount, usd_value FROM history_events '
+        'WHERE location="B" AND type IN ("spend", "receive") AND subtype="none"',
+    )
+    for row in write_cursor:
+        grouped_events[row[1]].append(row)
+
+    for groups in grouped_events.values():
+        if len(groups) != 2:
+            continue
```
nebolax, line 285: Doesn't this mean an error?

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -185,33 +187,131 @@ def _update_ens_mappings_schema(write_cursor: 'DBCursor') -> None:
...
+    update_tuples = []
+    grouped_events: dict[str, list[Any]] = defaultdict(list)
+    write_cursor.execute(
+        'SELECT identifier, event_identifier, type, amount, usd_value FROM history_events '
+        'WHERE location="B" AND type IN ("spend", "receive") AND subtype="none"',
+    )
+    for row in write_cursor:
+        grouped_events[row[1]].append(row)
+
+    for groups in grouped_events.values():
```
nebolax, line 283: I would rename `groups` to something like `events`, since it is not a list of groups here, is it?

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -185,33 +187,131 @@ def _update_ens_mappings_schema(write_cursor: 'DBCursor') -> None:
...
+        if len(groups) != 2:
+            continue
+
+        for event in groups:
+            event_type = HistoryEventType.TRADE.serialize()
```
nebolax, line 288: What is the reason for doing this here? Why not to do it when appending to `update_tuples`?

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -185,33 +187,131 @@ def _update_ens_mappings_schema(write_cursor: 'DBCursor') -> None:
...
+        write_cursor.executemany(
+            'UPDATE history_events SET amount=?, usd_value=? WHERE identifier=?',
+            update_tuples,
+        )
+
+    log.debug('Fixing kraken instant swaps')
+    update_tuples = []
+    grouped_events: dict[str, list[Any]] = defaultdict(list)
+    write_cursor.execute(
+        'SELECT identifier, event_identifier, type, amount, usd_value FROM history_events '
```
nebolax, line 277: Same question about the guarantees on the order of the events, as for the trades upgrade part.

rotkehlchen/exchanges/manager.py
```diff
@@ -307,3 +308,13 @@ def get_user_binance_pairs(self, name: str, location: Location) -> list[str]:
     def query_history_events(self) -> None:
         for exchange in self.iterate_exchanges():
             exchange.query_history_events()
+
+    def get_exchange_mappings(self) -> LocationEventMappingType:
+        """Collect from the different exchanges the event mappings"""
+        mappings: LocationEventMappingType = {}
+        for location, exchanges in self.connected_exchanges.items():
+            for exchange in exchanges:
+                if (exchange_mapping := exchange.get_event_mappings()) != {}:
```
nebolax, line 317: nitpick: I think doing `if len(exchange_mapping := exchange.get_event_mappings()) != 0:` is a bit safer

rotkehlchen/tests/api/test_misc.py
```diff
@@ -521,26 +521,28 @@ def test_cache_deletion(rotkehlchen_api_server):
...
-            rotkehlchen_api_server,
+            rotkehlchen_api_server_with_exchanges,
             'typesmappingsresource',
         ),
     )
     result = assert_proper_response_with_result(response)
     assert 'per_protocol_mappings' in result
     assert 'global_mappings' in result
     assert 'event_category_details' in result
+    assert 'kraken' in result['exchange_mappings'] and len(result['exchange_mappings']) == 1
```
nebolax, line 541: This could be checked in a single condition with `result['exchange_mappings'].keys() == {'kraken'}`

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1407,15 +1410,24 @@ def test_upgrade_db_36_to_37(user_data_dir):  # pylint: disable=unused-argument
...
+        if entry[5] != 'B':
+            # we only check the kraken events here
             continue
-
-        if entry[11] is None and entry[0]:
-            assert (entry[0], 0, *entry[1:11]) + ('none',) in new_history_events
-        else:
-            assert (entry[0], 0, *entry[1:12]) in new_history_events
+        # for kraken events we check also values and subtypes
+        assert entry in updated_kraken_events
```
nebolax, line 1430: I would compute sets of expected `updated_kraken_events` and the actual entries, and then compare the sets, to make sure that no event was lost

rotkehlchen/tests/data/v36_rotkehlchen.db
nebolax, line 1: DB diff please :stuck_out_tongue:

rotkehlchen/exchanges/kraken.py
```diff
@@ -225,6 +239,19 @@ def history_event_from_kraken(
                 f'Failed to read ledger event from kraken {raw_event} due to {msg}',
             )
             return [], False
+
+    for event_set in receive_spend_events.values():
+        if len(event_set) == 2:
```
nebolax, line 244: What if len(event_set) is not 2? This means an error, right? I guess we should warn the user / log it.

rotkehlchen/exchanges/kraken.py
```diff
@@ -225,6 +239,19 @@ def history_event_from_kraken(
...
+            for event in event_set:
+                if event.event_type == HistoryEventType.RECEIVE:
+                    event.event_subtype = HistoryEventSubType.RECEIVE
+                else:
+                    event.event_subtype = HistoryEventSubType.SPEND
+                    event.balance.amount = AssetAmount(-event.balance.amount)
```
nebolax, line 250: Why to use `AssetAmount`? Shouldn't this be just an FVal?

rotkehlchen/exchanges/kraken.py
```diff
@@ -225,6 +239,19 @@ def history_event_from_kraken(
                 f'Failed to read ledger event from kraken {raw_event} due to {msg}',
             )
             return [], False
+
+    for event_set in receive_spend_events.values():
+        if len(event_set) == 2:
+            for event in event_set:
```
nebolax, line 245: If it is guaranteed here that `spend` event goes first and `receive` event goes after, then I would access `event_set` by index instead of looping, since the order is known beforehand.

If there are no guarantees on the order, then I guess there is currently a bug, and we have to make sure that the spend event goes first.

I would solve this by doing
```python
spend_event, receive_event = event_set
if spend_event.event_type != HistoryEventSubType.SPEND:
    spend_event, receive_event = receive_event, spend_event

spend_event.balance.amount = -spend_event.balance.amount
spend_event.event_type = HistoryEventType.TRADE
receive_event.event_type = HistoryEventType.TRADE
spend_event.event_subtype = HistoryEventSubType.SPEND
receive_event.event_subtype = HistoryEventSubType.RECEIVE
```

Does this make sense to you?

rotkehlchen/exchanges/kraken.py
```diff
@@ -758,9 +785,6 @@ def query_online_deposits_withdrawals(
                 fee = Fee(ZERO)
 
             amount = movement.balance.amount
-            if movement.event_type == HistoryEventType.WITHDRAWAL:
```
nebolax, line 761: What is the reason for removing this?

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -185,33 +187,131 @@ def _update_ens_mappings_schema(write_cursor: 'DBCursor') -> None:
...
+
+    log.debug('Fixing kraken trades')
+    update_tuples = []
+    write_cursor.execute(
+        'SELECT identifier, amount, usd_value FROM history_events WHERE location="B" AND '
+        'type="trade" AND subtype="none"',
+    )
+    for event_row in write_cursor:
+        asset_amount = FVal(event_row[1])
+        usd_value = FVal(event_row[2])
```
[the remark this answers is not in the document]
yabirgb: I don't want to do it like this because It would mean to trust the kraken data since the sign comes directly from them and we don't filter instead of trusting the types that is something that we deserialize

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -185,33 +187,131 @@ def _update_ens_mappings_schema(write_cursor: 'DBCursor') -> None:
...
     for event_row in write_cursor:
+        amount = FVal(event_row[1])
+        usd_value = FVal(event_row[2])
         update_tuples.append(
             (
                 HistoryEventType.INFORMATIONAL.serialize(),
                 HistoryEventSubType.NONE.serialize(),
-                str(-FVal(event_row[1])),
-                str(-FVal(event_row[2])),
+                str(-amount) if amount != ZERO else str(amount),
```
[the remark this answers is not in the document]
yabirgb: No, before in the database there were negative vamous for the column. The check is to avoid writting `"-0"` to the database

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -185,33 +187,131 @@ def _update_ens_mappings_schema(write_cursor: 'DBCursor') -> None:
...
         )
-    log.debug('Exit _fix_kraken_eth2_events')
+
+    log.debug('Fixing kraken trades')
+    update_tuples = []
+    write_cursor.execute(
+        'SELECT identifier, amount, usd_value FROM history_events WHERE location="B" AND '
+        'type="trade" AND subtype="none"',
+    )
+    for event_row in write_cursor:
```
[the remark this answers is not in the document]
yabirgb: They will be grouped but the order in the database doesn't matter. We order always by sequence index

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -185,33 +187,131 @@ def _update_ens_mappings_schema(write_cursor: 'DBCursor') -> None:
...
+            ),
+        )
+
+    if len(update_tuples) != 0:
+        write_cursor.executemany(
+            'UPDATE history_events SET type=?, subtype=?, amount=?, usd_value=? WHERE identifier=?',  # noqa: E501
+            update_tuples,
+        )
+
+    log.debug('Fixing kraken withdrawals')
```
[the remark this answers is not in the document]
yabirgb: We always work with positive amounts in the events. That is the reason for this PR

rotkehlchen/db/upgrades/v36_v37.py
yabirgb: To avoid writting `"-0"` in the database. I believe it will deserialize well but I want to avoid this case

rotkehlchen/db/upgrades/v36_v37.py
yabirgb: Yes but since those are strings I wanted to avoid messing with them at sql level. Probably is faster as you suggest but then you have to build a more complex query

rotkehlchen/db/upgrades/v36_v37.py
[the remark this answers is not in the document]
yabirgb: I just noticed that since I use FVal it shouldn't happen the case of `-0`

rotkehlchen/db/upgrades/v36_v37.py
yabirgb: yes, is the group of events belonging to the same identifier

rotkehlchen/db/upgrades/v36_v37.py
yabirgb: The order doesn't matter as for how the events are inserted in the database. We have the sequence index for that

rotkehlchen/exchanges/kraken.py
yabirgb: as explained before is not an error

rotkehlchen/exchanges/kraken.py
yabirgb: I got it from another line but seems that is not needed

rotkehlchen/exchanges/kraken.py
yabirgb: Commmented already in discord. We use another approach to build the trades since the logic is more complicated

rotkehlchen/exchanges/kraken.py
yabirgb: The amount is already positive after the changes So when building the withdrawal event we don't need to make the absolute value

yabirgb: The amount is already positive after the changes So when building the withdrawal event we don't need to make the absolute value

rotkehlchen/db/upgrades/v36_v37.py
nebolax: Then 2 other questions.

Are those other types of actions correctly presented in out db? No fixing is needed for them?

Is it guaranteed that if there are 2 events then it is an instant swap and not some other type of action?

I would add an explanation of these in a comment.

rotkehlchen/exchanges/kraken.py
nebolax: I would add a comment that no action is needed for those other events

rotkehlchen/db/upgrades/v36_v37.py
yabirgb: for the groups they are built using the event id. This is unique per group and has all the parts involved in a "kraken event" If you have only one event there is surely not a trade as it is how the representation for events work in kraken. For instant swaps using spend/receive they are always 2 events

rotkehlchen/tests/db/test_db_upgrades.py
LefterisJP: Why did you not do this suggestion?
Either do what Alexey suggests or add a counter and check the counter equals length. Otherwise test can miss events.

LefterisJP (commented):
Left some comments

rotkehlchen/db/upgrades/v36_v37.py
```diff
@@ -205,13 +214,105 @@ def _fix_kraken_eth2_events(write_cursor: 'DBCursor') -> None:
...
+                HistoryEventType.TRADE.serialize(),
+                event_subtype.serialize(),
+                str(asset_amount),
+                str(usd_value),
+                event_row[0],
+            ),
+        )
+
+    if len(update_tuples) != 0:
+        write_cursor.executemany(
```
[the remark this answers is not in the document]
yabirgb: Your suggestion here would be to delete the row and do the insert again? I was not aware that the update was slower than the delete + insert

rotkehlchen/exchanges/kraken.py
```diff
@@ -225,6 +243,19 @@ def history_event_from_kraken(
                 f'Failed to read ledger event from kraken {raw_event} due to {msg}',
             )
             return [], False
+
+    for event_set in receive_spend_events.values():
```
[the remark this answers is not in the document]
yabirgb: Yes, amount is saved negative for cases that we don't know yet. What I did was to fix the values displayed in the history events both for ethereum staking and instant swaps

rotkehlchen/db/upgrades/v36_v37.py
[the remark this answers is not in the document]
LefterisJP: as I said in discord and also in my first comment we will have to see with my data, to see size that is touched. But yes update + executemany is much much slower than delete + Reinsert.

example: https://stackoverflow.com/questions/72492738/why-does-the-sqlite-function-executemany-run-slow-with-updating-large-amount-of

rotkehlchen/exchanges/kraken.py
```diff
@@ -1177,3 +1205,14 @@ def query_history_events(self) -> None:
                 'name': self.name,
             },
         )
+
+    @staticmethod
+    def get_event_mappings() -> EventMappingType:
```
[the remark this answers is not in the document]
yabirgb: yes the staking ones where in the default mapping. As for the others they should be covered except `ADJUSTMENT` that I've added too

rotkehlchen/tests/db/test_db_upgrades.py
yabirgb: I was waiting to clarify the other issue and alexey pinging for a review before closing everything

rotkehlchen/db/upgrades/v36_v37.py
yabirgb: Done

LefterisJP (approved):
LGTM. I will ignore the frontend test as it's probably flakiness again.

PR #5994, opened 2023-05-04
Handle error on manage assets form

tewshi:
## Checklist

- [x] The PR modified the frontend, and updated the managed assets component to handle the validation errors from backend
- [x] includes end to end test for add managed asset

### DIFF
--- a/frontend/app/src/components/asset-manager/ManagedAssetForm.vue
+++ b/frontend/app/src/components/asset-manager/ManagedAssetForm.vue
@@ -5,6 +5,8 @@ import {
   type SupportedAsset,
   type UnderlyingToken
 } from '@rotki/common/lib/data';
+import useVuelidate from '@vuelidate/core';
+import { required } from '@vuelidate/validators';
 import omit from 'lodash/omit';
 import { type ComputedRef, type Ref } from 'vue';
 import Fragment from '@/components/helper/Fragment';
@@ -20,7 +22,6 @@ function time(t: string): number | undefined {
 
 const props = withDefaults(
   defineProps<{
-    value: boolean;
     edit?: SupportedAsset | null;
     saving?: boolean;
   }>(),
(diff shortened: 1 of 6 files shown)


frontend/app/src/components/asset-manager/ManagedAssetForm.vue
```diff
@@ -79,6 +80,16 @@ const {
...
+const v$ = useVuelidate(
+  {
+    asset_type: { required }
+  },
+  {
+    asset_type: computed(() => get(assetType))
+  },
+  { $autoDirty: true }
+);
+
```
lukicenturi, line 92:
```
const v$ = useVuelidate(
  {
    assetType: { required }
  },
  {
    assetType
  },
  { $autoDirty: true }
);
```

frontend/app/src/components/asset-manager/ManagedAssetForm.vue
```diff
@@ -79,6 +80,16 @@ const {
...
 watch(address, async () => {
   const sanitizedAddress = sanitizeAddress(get(address));
   if (get(address) !== sanitizedAddress) {
```
lukicenturi, line 95: after the `decimals`, `name`, and `symbol` filled, after we search by the address, 
the error messages for these fields need to be cleared.

tewshi: okay, so those 3 are also updated on auto fetching the address...

just noticed that even though `decimals` is set after fetching address, it still says the decimals field cannot be null

frontend/app/src/components/asset-manager/ManagedAssetForm.vue
```diff
@@ -97,10 +116,16 @@ watch(address, async () => {
...
+  set(decimals, newDecimals ?? get(decimals));
   set(name, newName || get(name));
   set(symbol, newSymbol || get(symbol));
   set(fetching, false);
+  clearFieldErrors(['decimals', 'name', 'symbol']);
+});
+
+watch(assetType, () => {
+  // clearing these errors because these fields are disabled for non evm assets
+  clearFieldErrors(['evmChain', 'tokenKind']);
```
kelsos, line 128: Out of curiosity why not clear all the errors when changing asset type?

Since each asset type has a different set of required fields we should be able to assume that any previous validation error is irrelevant.

tewshi: okay, that'll do...

tewshi: running tests, to be sure it's all good.

kelsos (approved):
lgtm @tewshi thank you

PR #5995, opened 2023-05-05
fix: disable validator form when editing

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/accounts/management/types/ValidatorAccountForm.vue
+++ b/frontend/app/src/components/accounts/management/types/ValidatorAccountForm.vue
@@ -99,7 +99,7 @@ onMounted(() => {
   <v-form v-model="valid">
     <eth2-input
       :validator="validator"
-      :disabled="loading"
+      :disabled="loading || !!accountToEdit"
       :error-messages="errorMessages"
       @update:validator="validator = $event"
     />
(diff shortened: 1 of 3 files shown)


frontend/app/tests/e2e/pages/history-page/index.ts
```diff
@@ -5,6 +5,7 @@ export class HistoryPage {
     cy.get('.navigation__history').click();
     cy.get('.navigation__history-trades').scrollIntoView();
     cy.get('.navigation__history-trades').should('be.visible');
+    cy.get('.navigation__history-trades').trigger('mouseover');
```
lukicenturi, line 8: So I want to fix the flakiness here that happen often,
What I see from the video is, the `button` is not being clicked here.

I look into https://stackoverflow.com/questions/51254946/cypress-does-not-always-executes-click-on-element and https://github.com/cypress-io/cypress/issues/2928

It seems this line helps.
Maybe we can consider to move to `playwright` after the release, cypress is too bad.
Also I read somewhere https://www.reddit.com/r/QualityAssurance/comments/yqhg55/cypress_vs_playwright_some_comparison_elements/ that cypress will be soon become paid tool.

kelsos (approved):
lgtm

PR #6000, opened 2023-05-05
Cleanup identifiers usage in /assets/all

nebolax:
Updates the docs, removes unneeded code and makes sure that if `identifiers` list is provided, it has the highest priority in `POST /assets/all` enpoint

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -1254,30 +1254,14 @@ class RestAPI():
         # else
         return api_response(OK_RESULT, status_code=HTTPStatus.OK)
 
-    def query_list_of_all_assets(
-            self,
-            filter_query: AssetsFilterQuery,
-            identifiers: Optional[list[str]],
-    ) -> Response:
-        """
-        If a set of identifiers is provided returns the information for those identifiers otherwise
-        returns all supported assets with pagination.
-        """
+    def query_list_of_all_assets(self, filter_query: AssetsFilterQuery) -> Response:
+        """Query assets using the provided filter_query and return them in a paginated format"""
         assets, assets_found = GlobalDBHandler().retrieve_assets(filter_query=filter_query)
         with GlobalDBHandler().conn.read_ctx() as cursor:
             assets_total = self.rotkehlchen.data.db.get_entries_count(
                 cursor=cursor,
                 entries_table='assets',
             )
-
-        if identifiers is not None and len(identifiers) != assets_found:
-            not_found = []
-            found_assets = {asset['identifier'] for asset in assets}
-            for asset in identifiers:
-                if asset not in found_assets:
-                    not_found.append(asset)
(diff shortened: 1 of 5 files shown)


rotkehlchen/api/v1/schemas.py
```diff
@@ -2042,11 +2042,10 @@ def make_assets_post_query(
             elif data['ignored_assets_handling'] == IgnoredAssetsHandling.SHOW_ONLY:
                 ignored_assets_filter_params = ('IN', self.db.get_ignored_asset_ids(cursor))
 
-            if data['show_user_owned_assets_only'] is True:
+            identifiers = data['identifiers']
+            if data['show_user_owned_assets_only'] is True and identifiers is None:
                 globaldb_read_cursor.execute('SELECT asset_id FROM user_owned_assets;')
                 identifiers = [entry[0] for entry in globaldb_read_cursor]
-            else:
-                identifiers = data['identifiers']
```
yabirgb, line 2049: why this change?

rotkehlchen/api/rest.py
```diff
@@ -1270,15 +1266,6 @@ def query_list_of_all_assets(
...
             )
 
-        if identifiers is not None and len(identifiers) != assets_found:
-            not_found = []
-            found_assets = {asset['identifier'] for asset in assets}
-            for asset in identifiers:
-                if asset not in found_assets:
-                    not_found.append(asset)
-            msg = f"Queried identifiers {','.join(not_found)} are not present in the database"
-            return api_response(wrap_in_fail_result(msg), status_code=HTTPStatus.CONFLICT)
```
yabirgb, line 1280: then you need to update the docs but why this change?

yabirgb (commented):
I'm not sure about the background for the changes in this PR. If you could explain would be nice

rotkehlchen/api/rest.py
nebolax: This part of the code looked a bit strange to me. I asked Luki and he said that this check is not used / relied upon anywhere anyway, so I removed the unused code.

rotkehlchen/api/v1/schemas.py
yabirgb: Yes, that is what I would expect from the condition you wrote

rotkehlchen/tests/api/test_assets.py
```diff
@@ -470,22 +470,6 @@ def test_get_all_assets(rotkehlchen_api_server):
...
     assert result['entries'][2]['identifier'] == A_BTC
 
-    # ask for a non existent asset
-    response = requests.post(
-        api_url_for(
-            rotkehlchen_api_server,
-            'allassetsresource',
-        ),
-        json={
-            'identifiers': ['my_life'],
```
LefterisJP, line 480: Hahahah @yabirgb nicely done. Too bad @nebolax is removing it

LefterisJP (approved):
LGTM. Error is just blockchain.info timeout

PR #6001, opened 2023-05-05
fix: historic price cache for events view

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/price-manager/historic/HistoricPriceTable.vue
+++ b/frontend/app/src/components/price-manager/historic/HistoricPriceTable.vue
@@ -1,9 +1,4 @@
 <script setup lang="ts">
-import {
-  NotificationCategory,
-  type NotificationPayload,
-  Severity
-} from '@rotki/common/lib/messages';
 import { type DataTableHeader } from 'vuetify';
 import { type HistoricalPrice } from '@/types/prices';
 
@@ -19,14 +14,13 @@ const props = withDefaults(
 
 const emit = defineEmits<{
   (e: 'edit', item: HistoricalPrice): void;
-  (e: 'refresh', modified: boolean | HistoricalPrice): void;
+  (e: 'delete', item: HistoricalPrice): void;
+  (e: 'refresh'): void;
 }>();
 
 const { items } = toRefs(props);
 
-const { notify } = useNotificationsStore();
 const { tc } = useI18n();
-const { deleteHistoricalPrice } = useAssetPricesApi();
 
 const headers = computed<DataTableHeader[]>(() => [
   {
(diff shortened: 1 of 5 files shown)


frontend/app/src/components/price-manager/historic/HistoricPriceTable.vue
```diff
@@ -64,11 +62,17 @@ const headers = computed<DataTableHeader[]>(() => [
   }
 ]);
 
+const edit = (item: HistoricalPrice) => {
+  emit('edit', item);
+};
+const refresh = (modified: boolean | HistoricalPrice = false) =>
+  emit('refresh', modified);
+
 const deletePrice = async (item: HistoricalPrice) => {
```
kelsos, line 71: Since all the other actions are emitting events and are handled to the parent component wouldn't it make sense to also have this moved to?

frontend/app/src/pages/price-manager/historic/index.vue
```diff
@@ -26,11 +32,60 @@ const valid = ref(false);
...
+    };
+    notify(notification);
+  } finally {
+    set(loading, false);
+  }
+};
+
+const { resetHistoricalPricesData } = useHistoricCachePriceStore();
+
+const refresh = async (modified: boolean | HistoricalPrice = false) => {
```
kelsos, line 67: Why do that instead of making modified optional?

lukicenturi: so there are 3 state of refresh
1. modified = false, refresh without modify anything from `useHistoricCachePriceStore`
2. modified = true, remove cache from all data we have on historic price table
3. modified = Historical Price, do no 2 + delete cache for specific key (happens when we delete historic price, where the item is not exist anymore on the table, but still need to be cleared)

for no 1, yes we can make it optional

frontend/app/src/components/price-manager/historic/HistoricPriceTable.vue
lukicenturi: Hmm, I'm not sure I follow, this only call the emit, the refresh happens on parent component🧐

frontend/app/src/pages/price-manager/historic/index.vue
lukicenturi: like this
```
const refresh = async (modified?: true | HistoricalPrice) => {
```

lukicenturi: > Can you use two parameters here instead of this unholy union 😄

Just saw the edited version, yeah I will make it 2 parameters

kelsos: Yes I was reading the code and saw that it was 3 states so I updated it :)

frontend/app/src/components/price-manager/historic/HistoricPriceTable.vue
kelsos: it was about doing `emit('delete', item)` and moving the delete call to the parent too, since now it is the only call to API in this component. Just to have them in proximity.

Is there something that stops us from this? Do you think there is a reason why we should not do this?

lukicenturi: oh damn I was looking at the wrong line, I thought you highlight the refresh function, okay2, I got it

kelsos (approved):
lgtm @lukicenturi

PR #6006, opened 2023-05-05
feat: update eth2 staking stats

kelsos:
[skip ci]

Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/locales/en.json
+++ b/frontend/app/src/locales/en.json
@@ -1451,6 +1451,7 @@
     "price_in_symbol": "Price in {symbol}",
     "priority": "Priority",
     "protocol": "Protocol",
+    "rewards": "@:common.rewards",
     "select_address": "Select address(es)",
     "total": "Total",
     "tx": "Tx",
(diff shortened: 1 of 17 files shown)


frontend/app/src/locales/en.json
```diff
@@ -1451,6 +1451,7 @@
     "price_in_symbol": "Price in {symbol}",
     "priority": "Priority",
     "protocol": "Protocol",
+    "rewards": "@:common.rewards",
```
lukicenturi, line 1454:
```
"rewards": "Rewards",
```

lukicenturi:
image

from beginning, there's no `dailystats` that called with `only_cache=false`.
So the table always empty.
image

frontend/app/src/locales/en.json
kelsos: I did mass replace everything here, didn't I?

lukicenturi:
It seems that this endpoint `<POST> /blockchains/ETH2/stake/details` is always error when called with empty payload `({})`

image

But if we send the validator_index, it works
image

Not sure if this should be fixed in backend or frontend.

lukicenturi:
> image
> 
> from beginning, there's no `dailystats` that called with `only_cache=false`. So the table always empty. image

You did the `only_cache=False`, but I think you forget to set the state from the task call.
I need to call once again with `only_cache=True` to see the table filled.

lukicenturi (approved):
LGTM!

PR #6007, opened 2023-05-05
Configure web3.py middlewares

yabirgb:
We found that web3.py was making one extra rpc call after every eth_call. After investigating we discovered that it was the on-chain validation in the web3.py middleware. Since we don't need it for our usage atm we decided to remove this middleware and save rpc calls.

We have also removed middlewares that we don't need like gas estimation and ens resolution

### DIFF
--- a/rotkehlchen/chain/evm/node_inquirer.py
+++ b/rotkehlchen/chain/evm/node_inquirer.py
@@ -3,6 +3,7 @@ import logging
 import random
 from abc import ABCMeta, abstractmethod
 from collections.abc import Callable, Sequence
+from contextlib import suppress
 from itertools import zip_longest
 from typing import TYPE_CHECKING, Any, Literal, Optional, Union
 from urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -352,6 +353,14 @@ def _init_web3(self, node: NodeName) -> tuple[Web3, str]:
         )
         ens = ENS(provider) if self.chain_id == ChainID.ETHEREUM else None
         web3 = Web3(provider, ens=ens)
+        with suppress(ValueError):
+            # https://github.com/ethereum/web3.py/blob/bba87a283d802bbebbfe3f8c7dc47560c7a08583/web3/middleware/validation.py#L137-L142  # noqa: E501
+            # validation middleware makes an un-needed for us chain ID validation causing 1 extra rpc call per eth_call # noqa: E501
+            web3.middleware_onion.remove('validation')
+            # We do not need to automatically estimate gas
+            web3.middleware_onion.remove('gas_price_strategy')
```
LefterisJP, line 361: We also don't need `gas_estimate`

Remember these were all of them:

```
OrderedDict([('gas_estimate', <function buffered_gas_estimate_middleware at 0x7f0ed7f02ca0>),                                                                                                                                                  
             ('abi', <function construct_web3_formatting_middleware.<locals>.formatter_middleware at 0x7f0ed7f02430>),                                                                                                                         
             ('validation', <function construct_web3_formatting_middleware.<locals>.formatter_middleware at 0x7f0ed7e2e430>),                                                                                                                  
             ('pythonic', <function construct_web3_formatting_middleware.<locals>.formatter_middleware at 0x7f0ed7e27280>),                                                                                                                    
             ('attrdict', <function attrdict_middleware at 0x7f0ed7f02b80>),                                                                                                                                                                                ('name_to_address', <function construct_web3_formatting_middleware.<locals>.formatter_middleware at 0x7f0ed2abea60>),                                                                                                             
             ('gas_price_strategy', <function gas_price_strategy_middleware at 0x7f0ed7e21790>),                                                                                                                                               
             ('request_param_normalizer', <function construct_web3_formatting_middleware.<locals>.formatter_middleware at 0x7f0ed7e21ca0>)])  
```

Not sure about all the rest

yabirgb: Added it. From the list it seems that `attrdict` but I would need to check if we use it somewhere in the app and the others seems to be needed

LefterisJP: For `attrdict` I think it's the automagic conversion of a dictionary to their own `AttributeDict` structures such as `Transaction`. But not sure.

LefterisJP (approved):
ZE PR LOOKS GUT ZU MIR

LefterisJP:
> I would suggest to cache the chain Id value instead of removing the validation, something simple like [KarpatkeyDAO/defi-protocols#8](https://github.com/KarpatkeyDAO/defi-protocols/pull/8) would do it.

What is the point of caching the chain ID call? What do you gain by just doing it periodically?

For the way rotki uses eth_call, which is to query state and not to actively modify it, chain ID rugging us in the endpoint will be noticd by other failures. There really is no need for this check. It's useless. I think that web3.py people were really overzealous in putting it there even for apps that actively interact with the chain.

PR #6008, opened 2023-05-06
Upgrade production requirements

LefterisJP:


LefterisJP:
Will merge. The defillama price test fails due to market moves. We should really just change this test

PR #6015, opened 2023-05-08
Avoid sending ws messages from kraken always when querying events

yabirgb:
When querying internally trades and deposits we were querying the history events and this was sending an unexpected ws message that could lead to inconsistent states in the UI.

### DIFF
--- a/rotkehlchen/exchanges/kraken.py
+++ b/rotkehlchen/exchanges/kraken.py
@@ -1084,12 +1084,16 @@ class Kraken(ExchangeInterface, ExchangeWithExtras):
             cursor: 'DBCursor',
             start_ts: Timestamp,
             end_ts: Timestamp,
+            notify_events: bool = False,
     ) -> bool:
         """
         Query Kraken's ledger to retrieve events and transform them to our internal representation
         of history events. Internally we look for the query range that needs to be queried in the
         range (start_ts, end_ts) to avoid double quering the kraken API when this method is called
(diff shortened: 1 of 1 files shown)


rotkehlchen/exchanges/kraken.py
```diff
@@ -1084,12 +1084,16 @@ def query_kraken_ledgers(
...
             start_ts: Timestamp,
             end_ts: Timestamp,
+            notify_events: bool = False,
     ) -> bool:
         """
         Query Kraken's ledger to retrieve events and transform them to our internal representation
         of history events. Internally we look for the query range that needs to be queried in the
         range (start_ts, end_ts) to avoid double quering the kraken API when this method is called
         for deposits/withdrawals and trades. The events queried are then stored in the database.
+        If notify_events is True we send websocket messages to notify the current interval of
```
LefterisJP, line 1094:
```suggestion
        
        If notify_events is True we send websocket messages to notify the current interval of
```

rotkehlchen/exchanges/kraken.py
```diff
@@ -1084,12 +1084,16 @@ def query_kraken_ledgers(
...
+        time being queried. By default is set to False to avoid sending unwanted message when this
```
LefterisJP, line 1095:
```suggestion
        time being queried. By default it is set to False to avoid sending unwanted message when this
```

LefterisJP (approved):
LGTM. Small nitpick to be fixed in a subsequent PR

PR #6021, opened 2023-05-09
Protect with a lock contract query in packaged db

yabirgb:
It could happen that two different requests tried to access at the same time the packaged db to copy a contract and then the task would fail because the packaged db would be already attached by one of them.

This was easily seen while visiting a page with multiple ens avatars when the contract was still not copied.

Notes:

1. I didn't use a lock at funtion level to avoid blocking contracts queries since the problem was only when querying from the packaged db
2. In the case of two greenlets executing the copy it will be a no-op for the second since we do `get_or_write_abi`
and a `INSERT OR IGNORE`

### DIFF
--- a/rotkehlchen/chain/evm/contracts.py
+++ b/rotkehlchen/chain/evm/contracts.py
@@ -15,6 +15,7 @@ from typing import (
 )
 
 from eth_typing.abi import Decodable
+from gevent.lock import Semaphore
 from web3 import Web3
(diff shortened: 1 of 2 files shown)


LefterisJP (commented):
minor comments. Also had asked you some niticks from yesterday's PR which I don't see done @yabirgb

LefterisJP (approved):
LGTM

PR #6023, opened 2023-05-09
chore: update dependencies

lukicenturi:
Closes #(issue_number)

## Checklist


lukicenturi:
> @lukicenturi do we absolutely need this?
> 
> Most of them are minor, but I would rather not touch the dependencies while we are already testing unless it is absolutely necessary.

Alright then, I think we can move it after the release.
But I think we need to upgrade the `chart.js`

kelsos:
> > @lukicenturi do we absolutely need this?
> > Most of them are minor, but I would rather not touch the dependencies while we are already testing unless it is absolutely necessary.
> 
> Alright then, I think we can move it after the release. But I think we need to upgrade the `chart.js`

yes, single problematic dependency is ok :)

PR #6028, opened 2023-05-09
Improve icons query

yabirgb:
This commit does:
1. Exit earlier with fewer checks if a custom icon is used
2. Spawn a greenlet to query the icon from coingecko
3. Notify with a 202 that the request has been processed

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -156,7 +156,11 @@ from rotkehlchen.globaldb.handler import GlobalDBHandler
 from rotkehlchen.globaldb.updates import ASSETS_VERSION_KEY
 from rotkehlchen.history.price import PriceHistorian
 from rotkehlchen.history.types import NOT_EXPOSED_SOURCES, HistoricalPrice, HistoricalPriceOracle
-from rotkehlchen.icons import check_if_image_is_cached, maybe_create_image_response
+from rotkehlchen.icons import (
+    check_if_image_is_cached,
+    create_image_response,
+    maybe_create_image_response,
+)
 from rotkehlchen.inquirer import CurrentPriceOracle, Inquirer
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.premium.premium import PremiumCredentials
@@ -1371,8 +1375,7 @@ class RestAPI():
             status_code=HTTPStatus.OK,
         )
 
-    @staticmethod
-    def edit_user_asset(data: dict[str, Any]) -> Response:
+    def edit_user_asset(self, data: dict[str, Any]) -> Response:
         try:
             GlobalDBHandler().edit_user_asset(data)
         except InputError as e:
@@ -1380,6 +1383,8 @@ class RestAPI():
 
         # Also clear the in-memory cache of the asset resolver to requery DB
         AssetResolver().assets_cache.remove(data['identifier'])
+        # clear the icon cache in case the coingecko id was edited
+        self.rotkehlchen.icon_manager.failed_asset_ids.remove(data['identifier'])
         return api_response(OK_RESULT, status_code=HTTPStatus.OK)
 
     def delete_asset(self, identifier: str) -> Response:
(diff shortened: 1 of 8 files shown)


rotkehlchen/icons.py
```diff
@@ -167,10 +176,8 @@ def get_icon(
         If not, all icons of the asset are queried from coingecko and cached
         locally before the requested data are returned.
         """
-        # First search custom icons
-        custom_icon_path = self.custom_iconfile_path(asset)
-        if custom_icon_path is not None:
-            return custom_icon_path
```
yabirgb, line 173: I removed this since is duplicated logic from `asset_icon_path`

yabirgb, line 173: I removed this since is duplicated logic from `asset_icon_path`

rotkehlchen/api/rest.py
```diff
@@ -2921,8 +2927,17 @@ def get_asset_icon(
             )
             if response is not None:
                 return response
+            else:
+                return maybe_create_image_response(image_path=icon_path)
```
LefterisJP, line 2931: What are you doing here? Did you not already just check for the same icon path above?

rotkehlchen/icons.py
```diff
@@ -127,7 +137,7 @@ def query_coingecko_for_icon(self, asset: Asset, coingecko_id: str) -> bool:
         # we only keep delisted asset coingecko mappings since historical prices
         # can still be queried.
         if asset.identifier in DELISTED_ASSETS:
-            self.failed_asset_ids.add(asset.identifier)
+            self.failed_asset_ids.add(asset.identifier, asset.identifier)
```
LefterisJP, line 140: This is useless. Change it also everywhere else you use it. you are making a cache with double the needed size. You are caching the asset identifier both as the key and the value in the cache. For the value you don't need anything. Just a  boolean would be enough.

In fact this cache you are using is a key-value mapping cache. You don't need this. You need a set LRU cache.

rotkehlchen/icons.py
LefterisJP: Do we have a test for cusom icon paths? Make sure we do if we don't already

rotkehlchen/icons.py
```diff
@@ -167,10 +177,8 @@ def get_icon(
         If not, all icons of the asset are queried from coingecko and cached
         locally before the requested data are returned.
         """
-        # First search custom icons
-        custom_icon_path = self.custom_iconfile_path(asset)
-        if custom_icon_path is not None:
-            return custom_icon_path
+        if self.failed_asset_ids.get(asset.identifier) is not None:
```
LefterisJP, line 180: you added a return value argument. Please edit the docstring to explain what this function returns.

rotkehlchen/api/rest.py
```diff
@@ -2921,8 +2927,17 @@ def get_asset_icon(
...
 
-        image_path = self.rotkehlchen.icon_manager.get_icon(asset)
+        image_path, processed = self.rotkehlchen.icon_manager.get_icon(asset)
```
LefterisJP, line 2933: processed is not really the correct name to use here.

```suggestion
        image_path, launched_remote_query = self.rotkehlchen.icon_manager.get_icon(asset)
```

rotkehlchen/api/rest.py
yabirgb: here is creating a the response but is true that adds a check that was already done. I created a new function to avoid that. Another system call saved

rotkehlchen/icons.py
yabirgb: yes, is in `test_upload_custom_icon`

rotkehlchen/utils/data_structures.py
```diff
@@ -33,3 +33,32 @@ def remove(self, key: str) -> None:
     def clear(self) -> None:
         """Delete all entries in the cache"""
         self.cache.clear()
+
+
+class LRUSetCache(Generic[RT]):
+    """
+    LRU cache that works like a set.
+    Internally it uses an OrderedDict to keep the order that maps to None.
```
LefterisJP, line 41: I don't get it. Why did you use ordered dict internally? I don't understand what `"to keep the order that maps to None."` means

You probably mean: In order to be able to keep the order of insertion so that the LRU mechanism can function. Right?

Can you rephrase?

LefterisJP (approved):
LGTM. I left a nitpick comment on a docstring, but other than that, all green so I am merging.

PR #6029, opened 2023-05-10
chore: upgrade chart.js version

lukicenturi:
https://github.com/rotki/rotki/pull/6023#issuecomment-1541311042

Only update chart.js version, since current version is kinda buggy.


kelsos (approved):
lgtm

PR #6031, opened 2023-05-10
fix: tag deletion issue

lukicenturi:
Fix small issue with tag deletion, when you edit account, and remove the tag from the tag-manager, you cannot save the account


kelsos (approved):
lgtm, @lukicenturi thank you

PR #6036, opened 2023-05-10
feat: use backend mappings for profit loss events type

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/types/reports.ts
+++ b/frontend/app/src/types/reports.ts
@@ -3,10 +3,6 @@ import { z } from 'zod';
 import { type Quarter } from '@/types/frontend-settings';
 import { BaseAccountingSettings } from '@/types/user';
 
-export enum ProfitLossEventTypeEnum {
-  TRANSACTION_EVENT = 'transaction event'
-}
-
 export const ProfitLossOverviewItem = z.object({
   taxable: NumericString,
   free: NumericString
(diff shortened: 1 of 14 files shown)


frontend/app/src/types/reports.ts
```diff
@@ -4,15 +4,6 @@ import { type Quarter } from '@/types/frontend-settings';
...
-  TRADE = 'trade',
-  FEE = 'fee',
-  ASSET_MOVEMENT = 'asset movement',
-  MARGIN_POSITION = 'margin position',
-  LOAN = 'loan',
-  PREFORK_ACQUISITION = 'prefork acquisition',
-  LEDGER_ACTION = 'ledger action',
-  STAKING = 'staking',
-  HISTORY_BASE_ENTRY = 'history base entry',
   TRANSACTION_EVENT = 'transaction event'
```
lukicenturi, line 7: I think we better do this, keep the item that we will need to check with, other than do the assertion with the string directly.

Maybe we can do this for hardcoded eventType and eventSubType too

kelsos: We certainly should not. The reason we move things to the backend is to avoid hardcoding them in the frontend.

I know we cannot always avoid it (special views for events), but we should not use a value more than once. Having an enum encourages using it more than once. If we have to hardcode a value that is supplied by the mapping more than once then we should re-evaluate our approach

kelsos (approved):
lgtm @lukicenturi thanks

PR #6040, opened 2023-05-10
Withdrawals remote error

LefterisJP:
- When querying for withdrawals, exchange events or block production the API should now handle errors properly
- Do not silently ignore EVM filters if counterparty is eth2


LefterisJP:
Finnaly after 7 pushes tests ran and they are green.

PR #6045, opened 2023-05-11
fix: small chart bugs

lukicenturi:
Fix the issue where the main chart when all point is deleted, the chart become empty.


kelsos (approved):
lgtm @lukicenturi thank you

PR #6047, opened 2023-05-11
Withdrawals to query for will no longer duplicate validator indices

LefterisJP:
The SQL query had an issue and provided as many validators to query
for a single validator as there were withdrawals, which greatly
increased the number of times we queried a single validator's withdrawals


LefterisJP:
all green. Le merge sportiff

PR #6049, opened 2023-05-11
Update nft schema

yabirgb:
This PR flatterns the NFT balances response to avoid grouping by address the response

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/nft/nfts.py
+++ b/rotkehlchen/chain/ethereum/modules/nft/nfts.py
@@ -1,5 +1,4 @@
 import logging
-from collections import defaultdict
 from typing import TYPE_CHECKING, Any, Optional
 
 from pysqlcipher3 import dbapi2 as sqlcipher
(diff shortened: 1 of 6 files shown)


rotkehlchen/chain/ethereum/modules/nft/nfts.py
```diff
@@ -163,14 +162,14 @@ def get_all_info(
...
-        entries = defaultdict(list)
+        entries: list[dict[str, Any]] = []
         query, bindings = filter_query.prepare()
         total_usd_value = ZERO
         with self.db.conn.read_ctx() as cursor:
             cursor.execute(NFT_INFO_SQL_QUERY + query, bindings)
             for db_entry in cursor:
                 row_data = _deserialize_nft_from_db(entry=db_entry)
-                entries[db_entry[5]].append(row_data)
+                entries.append(row_data)
```
LefterisJP, line 172: IF you don't use one part of the data(owner_address), remove it from `NFT_INFO_SQL_QUERY`.

In fact remove this silly variable. Why put it in a variable when it's not even used anywhee. Just spell it out here. Much easier to read the code. Generally remove any data we don't use from the query

LefterisJP (approved):
LGTM

PR #6058, opened 2023-05-12
Use lock for packaged db connections

nebolax:
Closes #(issue_number)

## Checklist


rotkehlchen/globaldb/handler.py
```diff
@@ -1550,77 +1554,75 @@ def hard_reset_assets_list(
...
-                read_cursor.execute(detach_database)
-            log.error(f'Failed to restore assets in globaldb due to {str(e)}')
-            return False, 'Failed to restore assets. Read logs to get more information.'
-
         with GlobalDBHandler().conn.read_ctx() as read_cursor:
-            read_cursor.execute(detach_database)
+            # First check that the operation can be made. If the difference is not the
+            # empty set the operation is dangerous and the user should be notified.
+            with user_db.user_write() as user_db_cursor:
+                diff_ids = GlobalDBHandler().get_user_added_assets(
```
nebolax, line 1567: There was an issue with `get_user_added_assets`. Problem is that `get_user_added_assets` uses packaged_db_lock and this call to `get_user_added_assets` was also wrapped in packaged_db_lock, which resulted in a problem because when `get_user_added_assets` was called, the lock was already acquired. That's why I moved this call out of the try-except statement.

rotkehlchen/globaldb/handler.py
```diff
@@ -1550,77 +1554,75 @@ def hard_reset_assets_list(
...
+                    query = read_cursor.execute('SELECT value from clean_db.settings WHERE name="version";')  # noqa: E501
+                    version = query.fetchone()
+                    if version is None or int(version[0]) != globaldb_get_setting_value(read_cursor, 'version', GLOBAL_DB_VERSION):  # noqa: E501
+                        msg = (
+                            'Failed to restore assets. Global database is not '
+                            'updated to the latest version'
+                        )
+                        return False, msg
+
+                with GlobalDBHandler().conn.write_ctx() as write_cursor:
```
yabirgb, line 1591: We had this double cursor I believe because the write cursor blocks and we were doing some reads. Since you added another with statement maybe we could use a single

```python
with (
    GlobalDBHandler().packaged_db_lock
    GlobalDBHandler().conn.write_ctx() as write_cursor
):
```

to remove indentation from this part of the code as I don't believe is such a critical thing to execute two more sql queries with a write cursor. You would be able to remove the cursor from the `finally` statement

rotkehlchen/globaldb/handler.py
```diff
@@ -1634,54 +1636,56 @@ def soft_reset_assets_list() -> tuple[bool, str]:
...
+                        read_cursor.execute(detach_database)
+                        msg = (
+                            'Failed to restore assets. Global database is not '
+                            'updated to the latest version'
+                        )
+                        return False, msg
+
+                    # Get the list of ids that we will restore
+                    query = read_cursor.execute('SELECT identifier from clean_db.assets;')
+                    shipped_asset_ids = set(query.fetchall())
```
yabirgb, line 1656: you didn't write it but I believe is possible to do just `set(query)` isn't it?

rotkehlchen/globaldb/handler.py
```diff
@@ -1634,54 +1636,56 @@ def soft_reset_assets_list() -> tuple[bool, str]:
...
+                    asset_ids = ', '.join([f'"{id[0]}"' for id in shipped_asset_ids])
+                    query = read_cursor.execute('SELECT id FROM clean_db.asset_collections')
+                    shipped_collection_ids = set(query.fetchall())
```
yabirgb, line 1659: same here

yabirgb (commented):
Just a few small comments

rotkehlchen/globaldb/handler.py
nebolax: Should be

rotkehlchen/globaldb/handler.py
nebolax: If we move `write_cursor` out of the try-except scope, then no revert will happen in case of an error. So I would not do that

rotkehlchen/globaldb/handler.py
```diff
@@ -1550,77 +1554,75 @@ def hard_reset_assets_list(
...
+                        user_db_cursor.switch_foreign_keys('ON')
+
+                with user_db.conn.read_ctx() as cursor:
+                    # Update the owned assets table
+                    user_db.update_owned_assets_in_globaldb(cursor)
+
+            except sqlite3.Error as e:
+                log.error(f'Failed to restore assets in globaldb due to {str(e)}')
+                return False, 'Failed to restore assets. Read logs to get more information.'
+            finally:  # Finally is executed even if there is a return in try or except blocks
```
LefterisJP, line 1623: no reason to teach python to the reader

```suggestion
            finally:  # on the way out always detach the DB
```

rotkehlchen/globaldb/handler.py
```diff
@@ -1634,54 +1636,56 @@ def soft_reset_assets_list() -> tuple[bool, str]:
...
-                if version is None or int(version[0]) != globaldb_get_setting_value(read_cursor, 'version', GLOBAL_DB_VERSION):  # noqa: E501
+        with GlobalDBHandler().packaged_db_lock:
+            try:
+                with GlobalDBHandler().conn.read_ctx() as read_cursor:
+                    read_cursor.execute(f'ATTACH DATABASE "{builtin_database}" AS clean_db;')
+                    # Check that versions match
+                    query = read_cursor.execute('SELECT value from clean_db.settings WHERE name="version";')  # noqa: E501
+                    version = query.fetchone()
+                    if version is None or int(version[0]) != globaldb_get_setting_value(read_cursor, 'version', GLOBAL_DB_VERSION):  # noqa: E501
+                        read_cursor.execute(detach_database)
```
LefterisJP, line 1647: Do the same thing you did above

1. Delete the `detach_database` variable as it's just used in 1 place.
2. Move it to the finally and just use the literal string.

```suggestion
```

rotkehlchen/globaldb/handler.py
```diff
@@ -1634,54 +1636,56 @@ def soft_reset_assets_list() -> tuple[bool, str]:
...
+                    write_cursor.execute('INSERT INTO evm_tokens SELECT * FROM clean_db.evm_tokens;')  # noqa: E501
+                    write_cursor.execute('INSERT INTO underlying_tokens_list SELECT * FROM clean_db.underlying_tokens_list;')  # noqa: E501
+                    write_cursor.execute('INSERT INTO common_asset_details SELECT * FROM clean_db.common_asset_details;')  # noqa: E501
+                    write_cursor.execute('INSERT INTO asset_collections SELECT * FROM clean_db.asset_collections')  # noqa: E501
+                    write_cursor.execute('INSERT INTO multiasset_mappings SELECT * FROM clean_db.multiasset_mappings')  # noqa: E501
+                    # TODO: think about how to implement multiassets insertion
+                    write_cursor.switch_foreign_keys('ON')
+            except sqlite3.Error as e:
+                log.error(f'Failed to restore assets in globaldb due to {str(e)}')
+                with GlobalDBHandler().conn.read_ctx() as read_cursor:
```
LefterisJP, line 1682: as I said above, you can do same you did before and move detachment at finally

LefterisJP (approved):
LGTM. Thank you Alexey

PR #6059, opened 2023-05-12
filtering field with exclude and include behaviour was added

dimyG:
* dataclass is used to store the deserialized data
* currently the field is ussed for the entry_types filtering argument
* docs for entry_types were updated
* some test cases added
* fix entry_types usage

Closes #(issue_number)

## Checklist


rotkehlchen/api/v1/fields.py
```diff
@@ -66,6 +67,51 @@
...
+    ) -> IncludeExcludeFilterData:
+
+        if 'behaviour' not in value:
+            # If for any reason the behaviour key is not given, we use the default behaviour.
+            value['behaviour'] = 'include'
+
+        values = value['values']
+        behaviour = value['behaviour']
+
+        if behaviour not in ['include', 'exclude']:
```
LefterisJP, line 104: tuples. This is immutable

```suggestion
        if behaviour not in ('include', 'exclude'):
```

LefterisJP, line 104: tuples. This is immutable

```suggestion
        if behaviour not in ('include', 'exclude'):
```

rotkehlchen/api/v1/fields.py
```diff
@@ -66,6 +67,51 @@
...
+            raise ValidationError("Behaviour must be either 'include' or 'exclude'.")
+
+        deserialized_values = DelimitedOrNormalList(self.values_field).deserialize(values)
```
LefterisJP, line 107: This can fail by raising a DeserializationError if you check how it's implemented. You need to handle it.
```suggestion
        try:
            deserialized_values = DelimitedOrNormalList(self.values_field).deserialize(values)
        except DeserializationError as e:
            raise ValidationError(message=str(e), field_name='values') from e
```

rotkehlchen/api/v1/fields.py
```diff
@@ -66,6 +67,51 @@
...
+
+        if 'behaviour' not in value:
+            # If for any reason the behaviour key is not given, we use the default behaviour.
+            value['behaviour'] = 'include'
+
+        values = value['values']
+        behaviour = value['behaviour']
+
+        if behaviour not in ['include', 'exclude']:
+            raise ValidationError("Behaviour must be either 'include' or 'exclude'.")
```
LefterisJP, line 105: ValidationErrors need to specify the field if you know it.

```suggestion
            raise ValidationError(message="Behaviour must be either 'include' or 'exclude'.", field='behaviour')
```

rotkehlchen/api/v1/fields.py
```diff
@@ -66,6 +67,51 @@
...
+
+        deserialized_values = DelimitedOrNormalList(self.values_field).deserialize(values)
+        deserialized_behaviour = cast(Literal['IN', 'NOT IN'], 'IN' if behaviour == 'include' else 'NOT IN')  # noqa: E501
```
LefterisJP, line 108: cast is one way but annoying and does not help or tell us anything. In this case I would accept a type:ignore[XXX] where XXX is the name of the violation, which I assume is it does not see that it's Literal for some reason? But why?

```suggestion
        deserialized_behaviour = 'IN' if behaviour == 'include' else 'NOT IN'  # type: ignore[XXX]
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -527,7 +528,7 @@ def _make_query(
             event_subtypes=query_event_subtypes,
             exclude_subtypes=exclude_event_subtypes,
             assets=asset_list,
-            entry_types=[HistoryBaseEntryType.HISTORY_EVENT],
+            entry_types=IncludeExcludeFilterData(values=[HistoryBaseEntryType.HISTORY_EVENT], behaviour='IN'),  # noqa: E501
```
LefterisJP, line 531: Here and everywhere else, since you have put a default, just use it.

```suggestion
            entry_types=IncludeExcludeFilterData(values=[HistoryBaseEntryType.HISTORY_EVENT]),  # noqa: E501
```

rotkehlchen/db/filtering.py
```diff
@@ -911,11 +912,13 @@ def make(
                     ),
                 )
         if entry_types is not None:
+            values = entry_types.values
+            behaviour = entry_types.behaviour
```
LefterisJP, line 916:
```suggestion
```

rotkehlchen/db/filtering.py
```diff
@@ -911,11 +912,13 @@ def make(
...
             filters.append(DBMultiIntegerFilter(
                 and_op=True,
                 column='entry_type',
-                values=[x.value for x in entry_types],
-                operator='IN',
+                values=[x.value for x in values],
+                operator=behaviour,
```
LefterisJP, line 921: Do not create temporary variable for no reason

```suggestion
                values=[x.value for x in entry_type.values],
                operator=entry_types.behaviour,
```

rotkehlchen/api/v1/types.py
```diff
@@ -11,3 +13,9 @@ class EvmTransactionDecodingApiData(TypedDict):
 class EvmPendingTransactionDecodingApiData(TypedDict):
     evm_chain: SUPPORTED_CHAIN_IDS
     addresses: Optional[list[ChecksumEvmAddress]]
+
+
+@dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=True)
+class IncludeExcludeFilterData:
+    values: list[HistoryBaseEntryType]
+    behaviour: Literal['IN', 'NOT IN'] = 'IN'
```
LefterisJP, line 21: Can you rename this here as `operator`? For the API leave it behaviour, but when translating to `IncludeExcludeFilterData` use this as it's the same we use in fitlersand it's also shorter.

rotkehlchen/tests/api/test_eth2.py
```diff
@@ -961,3 +961,41 @@ def test_query_combined_mev_reward_and_block_production_events(rotkehlchen_api_s
...
+            rotkehlchen_api_server,
+            'historyeventresource',
+        ),
+        json={'entry_types': entry_types_exclude_arg},
+    )
+    result = assert_proper_response_with_result(response)
+    assert len(result['entries']) == 1
+    for outter_entry in result['entries']:
+        entry = outter_entry['entry']
+        assert entry['entry_type'] != 'eth block event'
```
LefterisJP, line 1001: typo

```suggestion
    for outer_entry in result['entries']:
        entry = outer_entry['entry']
        assert entry['entry_type'] != 'eth block event'
```

rotkehlchen/tests/db/test_history_events.py
```diff
@@ -126,12 +117,34 @@ def test_read_write_events_from_db(database):
...
+        7: (make_evm_tx_hash(), TimestampMS(7), 7, 'compound', EvmProduct.POOL, '0x19222290DD7278Aa3Ddd389Cc1E1d165CC4BAf34'),  # noqa: E501
+    }
+
+    add_history_events_to_db(db, history_data)
+    add_evm_events_to_db(db, evm_data)
+
+    # args for creating filter queries
+    filter_query_args = [
+        (HistoryEventFilterQuery, None),
+        (HistoryEventFilterQuery, IncludeExcludeFilterData(values=[HistoryBaseEntryType.HISTORY_EVENT], behaviour='IN')),  # noqa: E501
```
LefterisJP, line 142: again here as I said, use the fact you got the default argument set for 'IN'

rotkehlchen/api/v1/fields.py
```diff
@@ -66,6 +67,56 @@
...
+        if 'behaviour' not in value:
+            # If for any reason the behaviour key is not given, we use the default behaviour.
+            value['behaviour'] = 'include'
+
+        values = value['values']
+        behaviour = value['behaviour']
+
+        if behaviour not in ('include', 'exclude'):
+            raise ValidationError(message="Behaviour must be either 'include' or 'exclude'.",
+                                  field_name='behaviour')
```
LefterisJP, line 106: style: When splitting in multiple lines do it like this:

```suggestion
            raise ValidationError(
                message="Behaviour must be either 'include' or 'exclude'.",
                field_name='behaviour',
            )
```

rotkehlchen/api/v1/fields.py
```diff
@@ -66,6 +67,56 @@
...
+            data: Any,
+            **kwargs: Any,
+    ) -> IncludeExcludeFilterData:
+
+        if 'behaviour' not in value:
+            # If for any reason the behaviour key is not given, we use the default behaviour.
+            value['behaviour'] = 'include'
+
+        values = value['values']
+        behaviour = value['behaviour']
```
LefterisJP, line 102: shouldn't you check that 'behaviour' is given as a key? Either with an `in` or a try/except KeyError?

dimyG: I check if it is not given, and add the default value if not given:
```
        if 'behaviour' not in value:
            # If for any reason the behaviour key is not given, we use the default behaviour.
            value['behaviour'] = 'include'
```

LefterisJP: Ah sorry Dimitris. I meant the other one. `value["values"]`

LefterisJP (approved):
LGTM.

PR #6060, opened 2023-05-12
feat: update NFT sort column

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
+++ b/frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
@@ -4,7 +4,11 @@ import { type DataTableHeader } from 'vuetify';
 import { type ActionStatus } from '@/types/action';
 import { type IgnoredAssetsHandlingType } from '@/types/asset';
 import { type Module } from '@/types/modules';
-import { type NonFungibleBalance } from '@/types/nfbalances';
+import {
+  type NonFungibleBalance,
+  type NonFungibleBalanceWithLastPrice,
+  type NonFungibleBalancesRequestPayload
+} from '@/types/nfbalances';
 import { type ManualPriceFormPayload } from '@/types/prices';
 import { Section } from '@/types/status';
 
@@ -54,10 +58,9 @@ const tableHeaders = computed<DataTableHeader[]>(() => [
   },
   {
     text: tc('common.price_in_symbol', 0, { symbol: get(currencySymbol) }),
-    value: 'usdPrice',
+    value: 'lastPrice',
     align: 'end',
-    class: 'text-no-wrap',
-    sortable: false
+    class: 'text-no-wrap'
   },
   {
     text: tc('non_fungible_balances.column.custom_price'),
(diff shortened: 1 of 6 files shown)


frontend/app/src/components/dashboard/NftBalanceTable.vue
```diff
@@ -31,19 +34,17 @@ const {
...
-  {
-    extraParams,
-    defaultSortBy: {
-      key: 'name',
-      ascending: [true]
-    }
+} = usePaginationFilters<
+  NonFungibleBalance,
+  NonFungibleBalancesRequestPayload,
+  NonFungibleBalance & { lastPrice?: any }
```
kelsos, line 40: Don't we know the type of `lastPrice`?

kelsos, line 40: Since we are re-using this in multiple places maybe it would make sense to just extract it?

frontend/app/src/components/accounts/balances/NonFungibleBalances.vue
```diff
@@ -284,7 +284,7 @@ const showDeleteConfirmation = (item: NonFungibleBalance) => {
             />
             <span v-else>-</span>
           </template>
-          <template #item.usdPrice="{ item }">
+          <template #item.lastPrice="{ item }">
```
kelsos, line 287: How come we were not using `lastPrice`?

frontend/app/src/components/dashboard/NftBalanceTable.vue
```diff
@@ -177,6 +177,8 @@ watch(loading, async (isLoading, wasLoading) => {
           :loading="isLoading"
           :options="options"
           :server-items-length="itemLength"
+          multi-sort
+          :must-sort="false"
```
kelsos, line 181: Why disable must-sort is there any specific reason?

frontend/app/src/composables/filter-paginate.ts
```diff
@@ -12,6 +12,7 @@ import {
   type LocationQuery,
   RouterPaginationOptionsSchema
 } from '@/types/route';
+import { getUpdatedKey } from '@/services/axios-tranformers';
```
kelsos, line 15: At some point, we should maybe move some of the functions since this is clearly not used only in the Axios transformers. Don't do it now but we need to keep that in mind.

frontend/app/src/components/dashboard/NftBalanceTable.vue
lukicenturi: so it's true that we can sort by `lastPrice`, but that field isn't in the response

frontend/app/src/components/dashboard/NftBalanceTable.vue
lukicenturi: So if we dont put it, dont we can remove all sort.
Actually I'm thinking now, we can set the default value of `mustSort` in the `<data-table />` to false

kelsos: > So if we dont put it, dont we can remove all sort. Actually, I'm thinking now, we can set the default value of `mustSort` in the `<data-table />` to false

I understand what the option itself does. If you check vuetify's documentation, By default `must-sort` is set to `false`, so it should be pretty obvious that this was not by accident.

We decided to enforce sorting by default all across the application at some point. This has been a deliberate decision.

My question was, about what made this case special and why did you want to introduce the ability to disable sorting in this specific case. Just so that by default it falls back to the backend's default?

Please revert this change if your only reason is to use the backend's default, otherwise please let me know, and we can see if it is a valid reason.

One of the reasons we removed the ability to have a default (neutral) sorting was that this was confusing to people and it was not clear how the data were sorted.

We are not changing the default across the application. This is a terrible idea.

lukicenturi: Okay, I just realized that we intended to set the `mustSort` to true by default.

I just understand the pain point here if we use `mustSort=false`.
I revert it, and also remove this in the other Non Fungible Balances in `Asset & Balances`

kelsos (approved):
@lukicenturi I added some documentation and a warning in the setup function in case we accidentally do that.

I think it's ok if the properties are not reactive and only checked on setup because we generally don't use these properties other than as flags from what I have seen.

PR #6062, opened 2023-05-12
ENS avatars with NFT images will now render fine

LefterisJP:
If an ENS avatar is set as an NFT we first try the ENS app metadata as it's only one call.

If that fails we fallback to opensea, to ask the image url and then query that url to get the image.


rotkehlchen/chain/ethereum/utils.py
```diff
@@ -208,11 +212,25 @@ def try_download_ens_avatar(
     if avatar_url == '':
         return  # Avatar is not set
 
-    try:
-        avatar = requests.get(avatar_url, timeout=DEFAULT_TIMEOUT_TUPLE).content
-    except RequestException as e:
-        log.error(f'Got error {str(e)} during querying ens avatar for {ens_name}.')
-        return
+    avatar = None
+    if avatar_url.startswith('eip155'):  # an NFT is set. Let's try first ENS app's own metadata
```
yabirgb, line 216: We have `EVM_CHAIN_DIRECTIVE`, maybe use it?

```suggestion
    if avatar_url.startswith(EVM_CHAIN_DIRECTIVE):  # an NFT is set. Let's try first ENS app's own metadata
```

rotkehlchen/chain/ethereum/utils.py
```diff
@@ -208,11 +212,25 @@ def try_download_ens_avatar(
...
+            avatar_url = opensea.get_nft_image(avatar_url)
+            if avatar_url is None:
+                return  # no luck with opensea
+            # proceed to query the new avatar url
+
+    if avatar is None:  # we have not populated it via an NFT query above yet
+        try:
+            avatar = requests.get(avatar_url, timeout=DEFAULT_TIMEOUT_TUPLE).content
+        except RequestException as e:
+            log.error(f'Got error {str(e)} during querying ens avatar {avatar_url} for {ens_name}.')  # noqa: E501
```
yabirgb, line 232: question, is during querying correct?

```suggestion
            log.error(f'Got error {str(e)} while querying ens avatar {avatar_url} for {ens_name}.')  # noqa: E501
```

rotkehlchen/chain/ethereum/utils.py
```diff
@@ -208,11 +212,25 @@ def try_download_ens_avatar(
...
 
-    try:
-        avatar = requests.get(avatar_url, timeout=DEFAULT_TIMEOUT_TUPLE).content
-    except RequestException as e:
-        log.error(f'Got error {str(e)} during querying ens avatar for {ens_name}.')
-        return
+    avatar = None
+    if avatar_url.startswith('eip155'):  # an NFT is set. Let's try first ENS app's own metadata
+        try:
+            avatar = requests.get(f'{ENS_METADATA_URL}/avatar/{ens_name}', timeout=DEFAULT_TIMEOUT_TUPLE).content  # noqa: E501
```
yabirgb, line 218: Should you check that the response has status code 200 or at least no error and if it has headers that the content is `image/`?  Content for a page can also be the html text and I see this happening if hitting a cdn or nginx errors...

yabirgb (commented):
A few comments/questions

rotkehlchen/chain/ethereum/utils.py
LefterisJP: yes it's correct

rotkehlchen/chain/ethereum/utils.py
LefterisJP: This is not done at all wherever we query for images. Check the production code right now. But sure this is something that can be added.

rotkehlchen/chain/ethereum/utils.py
LefterisJP: But to be honest while does sound better. I think during woudl be more matched to something like: "During the query

PR #6063, opened 2023-05-12
fix: delete exchanges cache endpoint call

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/composables/api/balances/exchanges.ts
+++ b/frontend/app/src/composables/api/balances/exchanges.ts
@@ -128,14 +128,15 @@ export const useExchangeApi = () => {
   };
 
   const deleteExchangeData = async (
-    name: SupportedExchange | '' = ''
+    name?: SupportedExchange
   ): Promise<boolean> => {
-    const response = await api.instance.delete<ActionResult<boolean>>(
-      `/exchanges/data/${name}`,
-      {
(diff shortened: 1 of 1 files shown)


frontend/app/src/composables/api/balances/exchanges.ts
```diff
@@ -130,12 +130,13 @@ export const useExchangeApi = () => {
...
-        validateStatus: validStatus
-      }
-    );
+    let url = `/exchanges/data`;
+    if (name) {
+      url += `/${name}`;
+    }
+    const response = await api.instance.delete<ActionResult<boolean>>(url, {
+      validateStatus: validStatus
+    });
```
tewshi, line 139: from the previous code;

when `name  = ''` 
then the url will be `/exchanges/data/`

else it will be `/exchanges/data/$name`

the new code is adding `/$name` if is it not `''`

was having a trailing slash on the url problematic? (`/exchanges/data/` vs `/exchanges/data`)

lukicenturi: yes the trailing slash is problematic, it throw 404 with trailing slash

tewshi (approved):
lgtm!

kelsos:
I will do the change and merge.

PR #6066, opened 2023-05-12
Make sure that usd value is populated before returning liquity statistics

yabirgb:
The `usd_value` of the history events entries was not getting queried if it was 0 before running the calculation and it could lead to a wrong value getting shown.

in the test db what I did was to fix the schema and make sure that events are decoded


rotkehlchen/api/rest.py
```diff
@@ -2720,6 +2721,15 @@ def get_liquity_stability_pool_positions(self) -> dict[str, Any]:
     @async_api_call()
     def get_liquity_stats(self) -> dict[str, Any]:
         liquity_addresses = self.rotkehlchen.chains_aggregator.queried_addresses_for_module('liquity')  # noqa: E501
+        # make sure that all the entries that need it have the usd value queried
+        task_manager = self.rotkehlchen.task_manager
```
LefterisJP, line 2725: Isn't this potentially really heavy? If there is a lot of events?

rotkehlchen/tests/api/test_liquity.py
```diff
@@ -249,6 +249,12 @@ def test_staking_stats(rotkehlchen_api_server, ethereum_accounts):
     information and that the stats combining all the data are consistent with the
     information for each tracked address
     """
+    response = requests.post(
```
LefterisJP, line 252: why add the query at start?

yabirgb: To make sure that no event is left undecoded. For example if in a db upgrade we delete the decoded events

rotkehlchen/api/rest.py
yabirgb: is a price query that only affects the liquity events. In the best scenario nothing extra is queried, only the database. In the worst scenario you need the usd value of all the events (only liquity) but otherwise you get the wrong price

rotkehlchen/tests/api/test_liquity.py
LefterisJP: Then please add a comment there in your next PR.

LefterisJP (approved):

LefterisJP (approved):

PR #6067, opened 2023-05-12
Fix packaged_db is locked error

LefterisJP:
Unfortunately the previous approach here:
https://github.com/rotki/rotki/pull/6058 was not tested and as such was not a solution.

The problem happens only at detaching from the packaged DB and only when there is an active transaction for the same connection.

https://stackoverflow.com/questions/56243770/sqlite3-detach-database-produces-database-is-locked-error

So the approach I followed was to guard with the transaction lock when we try to detach a DB.

Also reproduced in a test.

### DIFF
--- a/rotkehlchen/tests/unit/test_evm_contracts.py
+++ b/rotkehlchen/tests/unit/test_evm_contracts.py
@@ -1,9 +1,10 @@
 import json
 
+import gevent
 from eth_utils import is_checksum_address
+
 from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
 from rotkehlchen.globaldb.handler import GlobalDBHandler
-
 from rotkehlchen.types import ChainID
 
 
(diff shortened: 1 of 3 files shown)


rotkehlchen/tests/unit/test_evm_contracts.py
```diff
@@ -36,11 +37,18 @@ def test_evm_abi_data(globaldb):
             abis_set.add(entry[1])
 
 
+def keep_open_globaldb_write():
+    """Silly way to keep a transaction open in another greenlet"""
+    with GlobalDBHandler().conn.write_ctx():
+        gevent.sleep(1.5)
+        gevent.sleep(1.5)
```
yabirgb, line 44: why do you need two of them? To check that it changes and when coming back it keeps working?

LefterisJP: yes otherwise it can end before the contract.contracts() thing ends. Or at least that was my thinking. Is not really accurate science.

LefterisJP:
Okay so this... may have done it

PR #6068, opened 2023-05-12
Improves eth2 functionality

kelsos:
Closes #(issue_number)

## Checklist


frontend/app/src/components/display/BalanceDisplay.vue
```diff
@@ -38,6 +40,31 @@ const usdValue = useValueOrDefault(
...
+const valueInCurrency = computed(() => {
+  if (!get(calculateValue)) {
+    return get(usdValue);
+  }
+
+  const owned = get(amount);
+  const ethPrice = assetPrice('ETH');
+
+  if (isDefined(ethPrice)) {
+    return owned.multipliedBy(get(ethPrice));
```
lukicenturi, line 63: Why it hardcoded to ETH and not taken from prop `asset`?

kelsos: That is a good suggestion, thanks :)

lukicenturi:
image

Now I'm thinking, will it be better if we use the `validatorIndex` here instead of the `publicKey`, since I think that's is more readable for users.

For consistency also, the history events and daily stats below show `validatorIndex`.

kelsos:
> image
> 
> Now I'm thinking, will it be better if we use the `validatorIndex` here instead of the `publicKey`, since I think that's is more readable for users.
> 
> For consistency also, the history events and daily stats below show `validatorIndex`.

I can change that in the components it should not be hard, let me see.

PR #6069, opened 2023-05-12
Some fixes

LefterisJP:
- Kraken is a valid counterparty for Ethereum transactions
- Add critical section to packaged DB access
- Add old ENS resolver 2 to the global DB to allow avatar resolution for names that have not been upgraded


LefterisJP:
All green finally. Merging.

Issue #5432, opened 2023-01-18
Flaky sqlite failure - Error binding parameter 0 - probably unsupported type


LefterisJP:
## Problem Definition

During initial blockchain balance query, the ETH query task just died and no ETH balances were shown. It was a shock for a bit but then I realized something was wrong by looking at the logs. Happened first time and has never been seen before.

Restarting,it all went well. The logs of what happened are below.

## Logs

```
[19/01/2023 00:17:48 CET] DEBUG rotkehlchen.api.server Greenlet-8: start rotki api GET /api/1/balances/blockchains/ETH view_args={'blockchain': 'ETH'}, query_string=b'async_query=true'
[19/01/2023 00:17:48 CET] DEBUG rotkehlchen.api.server Greenlet-8: end rotki api GET /api/1/balances/blockchains/ETH view_args={'blockchain': 'ETH'}, query_string=b'async_query=true', status_code=200, result={'result': {'task_id': 7}, 'message': ''}
[19/01/2023 00:17:48 CET] DEBUG rotkehlchen.api.rest Greenlet-24: Async task with task id 7 started
...
...
[19/01/2023 00:18:10 CET] DEBUG rotkehlchen.chain.ethereum.manager Greenlet-7: Querying ethereum chain for ETH balance eth_addresses=["redacted--was about 20 addresses"]
[19/01/2023 00:18:10 CET] DEBUG web3.RequestManager Making request. Method: eth_call
[19/01/2023 00:18:10 CET] DEBUG web3.RequestManager Making request. Method: eth_chainId
[19/01/2023 00:18:10 CET] DEBUG web3.providers.HTTPProvider Making request HTTP. URI: https://eth-mainnet.alchemyapi.io/v2/REDACTEDKEY, Method: eth_chainId
[19/01/2023 00:18:10 CET] ERROR rotkehlchen.api.rest Greenlet with id 140602543744272: Greenlet for task 7 dies with exception: Error binding parameter 0 - probably unsupported type..
Exception Name: <class 'pysqlcipher3.dbapi2.InterfaceError'>
Exception Info: Error binding parameter 0 - probably unsupported type.
Traceback:
   File "src/gevent/greenlet.py", line 908, in gevent._gevent_cgreenlet.Greenlet.run
  File "rotkehlchen/api/rest.py", line 298, in _do_query_async
  File "rotkehlchen/api/rest.py", line 664, in _query_blockchain_balances
[...]

PR #6071, opened 2023-05-13
Various changes centered around fixing 5432

LefterisJP:
### BlockchainAccounts lists are now tuples -- immutable

Also using Sequence as type in all relevant functions, which is an
immutable sequence. Matches either tuple or list or any other sequence
used immutably. As these should not be modified at any point apart from when reading from the DB or adding/removing an address to/from the DB.

This was all done in order to try and reproduce and find the root cause for this: https://github.com/rotki/rotki/issues/5432
But apart from liquity module where there was addresses being added, but on a carefully made copy I did not see anything else where mypy had warnings.

### Evm chain bug

[Fix bug where eth_accounts were always used irrespective of chain id](https://github.com/rotki/rotki/commit/c623743df9e4254ff3a234a58e26af0004f6f927)

### Reproduce and "fix"

Finally reproduced 5432 reliably in a test and provided a bad fix. For more info read the issue.

### DIFF
--- a/rotkehlchen/chain/evm/proxies_inquirer.py
+++ b/rotkehlchen/chain/evm/proxies_inquirer.py
@@ -1,5 +1,6 @@
 
 import logging
+from collections.abc import Sequence
 from typing import TYPE_CHECKING, Optional
 
 from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
@@ -60,7 +61,7 @@ class EvmProxiesInquirer():
 
     def get_accounts_proxy(
             self,
-            addresses: list[ChecksumEvmAddress],
+            addresses: Sequence[ChecksumEvmAddress],
     ) -> dict[ChecksumEvmAddress, ChecksumEvmAddress]:
         """
         Returns DSProxy if it exists for a list of addresses using only one call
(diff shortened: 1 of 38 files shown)


LefterisJP:
Oh goody. Tests failed here and there. Let's see if this makes sense to continue looking into.

LefterisJP:
I am gonna merge this, though there is still a known issue with `database packaged_db is locked` but this PR already does too many things.

rotkehlchen/chain/evm/proxies_inquirer.py
```diff
@@ -110,8 +111,7 @@ def get_accounts_having_proxy(self) -> dict[ChecksumEvmAddress, ChecksumEvmAddre
 
         with self.node_inquirer.database.conn.read_ctx() as cursor:
             accounts = self.node_inquirer.database.get_blockchain_accounts(cursor)
-        eth_accounts = accounts.eth
-        mapping = self.get_accounts_proxy(eth_accounts)
+        mapping = self.get_accounts_proxy(accounts.get(self.node_inquirer.blockchain))
```
LefterisJP, line 114: This was another bug I found. We always were taking eth accounts here irrespective of the evm chain type

PR #6074, opened 2023-05-15
Use packaged DB connection instead of attach/detach

LefterisJP:
Use packaged DB connection instead of attach/detach to get packaged DB contract data.

I only left the attach/detach in the hard/soft assets_reset functionality as there it combines queries in both DBs. It can probably be rewritten and we can then also get rid of the packaged db lock but I don't want to do such changes right before the release


LefterisJP:
all green. Merging.

PR #6076, opened 2023-05-15
Do not query eth2 daily stats for exited validators

LefterisJP:


LefterisJP:
Will merge. All failed tests were timeout related and passed locally when running again.

PR #6078, opened 2023-05-15
Allow custom assets to be imported and improve assets manipulation methods

### LINKED ISSUE #6073: Cannot import custom asset
For the import asset functionality, we cannot do the import if I have custom asset in the csv
endpoint:  `<PUT> /api/1/assets/user`
errors: 

```Provided file does not have the expected format. {'assets': {0: {'asset_type': ['Missing data for required field.'], 'cryptocompare': ['Missing data for required field.'], 'forked': ['Missing data for required field.'], 'coingecko': ['Missing data for required field.'], 'symbol': ['Missing data for required field.'], 'swapped_for': ['Missing data for required field.'], 'started': ['Missing data for required field.'], 'custom_asset_type': ['Unknown field.'], 'notes': ['Unknown field.']}}}```
Example of CSV
[assets.zip](https://github.com/rotki/rotki/files/11474689/assets.zip)

### PR DESCRIPTION
**Pull request by nebolax**
Closes #6073 

## Checklist

### CODE REVIEW

File: rotkehlchen/globaldb/handler.py [hunk 1]
```diff
@@ -254,69 +256,57 @@ def add_setting_value(name: str, value: Any) -> None:
...
+                if asset.asset_type == AssetType.CUSTOM_ASSET:
+                    write_cursor.execute(
+                        'INSERT INTO custom_assets(identifier, type, notes) VALUES(?, ?, ?)',
+                        cast(CustomAsset, asset).serialize_for_db(),
+                    )
+                else:
+                    asset = cast(AssetWithOracles, asset)
+                    if asset.is_evm_token():
+                        GlobalDBHandler.add_evm_token_data(write_cursor, cast(EvmToken, asset))
+
```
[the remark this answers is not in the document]
nebolax: Not sure if it makes much sense. `EvmToken` is a subclass of `AssetWithOracles`, so there is nothing wrong

rotkehlchen/globaldb/handler.py
```diff
@@ -254,69 +256,57 @@ def add_setting_value(name: str, value: Any) -> None:
...
+                    forked = None
+                    started = None
+                    swapped_for = None
+                    if asset.is_crypto():
+                        asset = cast(CryptoAsset, asset)
+                        forked = asset.forked.identifier if asset.forked else None
+                        started = asset.started
+                        swapped_for = asset.swapped_for.identifier if asset.swapped_for else None
```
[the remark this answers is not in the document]
nebolax: Rewrote it a bit. Not exactly the way that you proposed, but I think I made it more readable

rotkehlchen/api/v1/resources.py
```diff
@@ -755,12 +757,12 @@ def post(self, filter_query: AssetsFilterQuery) -> Response:
 
     @require_loggedin_user()
     @resource_parser.use_kwargs(make_add_schema, location='json')
-    def put(self, asset_type: AssetType, **kwargs: Any) -> Response:
-        return self.rest_api.add_user_asset(asset_type, **kwargs)
+    def put(self, crypto_asset: CryptoAsset) -> Response:
+        return self.rest_api.add_user_asset(crypto_asset)
```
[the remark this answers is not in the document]
nebolax: I'll add a docstring, yes. But I think the endpoint should also support editing of fiat assets (but this would be better to handle in another PR)

rotkehlchen/globaldb/handler.py
```diff
@@ -254,69 +256,58 @@ def add_setting_value(name: str, value: Any) -> None:
...
+                    write_cursor.execute(
+                        'INSERT INTO custom_assets(identifier, type, notes) VALUES(?, ?, ?)',
+                        cast(CustomAsset, asset).serialize_for_db(),
+                    )
+                    return
+
+                asset = cast(AssetWithOracles, asset)
+                forked, started, swapped_for = None, None, None
+                if asset.is_crypto():
+                    if asset.is_evm_token():
```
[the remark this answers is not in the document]
nebolax: I made it like this since I remember Lefteris mentioning in some PR that if a function's / variable's name clearly states that it is a boolean, then we should use it without `is True` / `is False`.

yabirgb (commented):
A few relevant tests failed. Please check them

rotkehlchen/api/v1/resources.py
```diff
@@ -727,6 +729,10 @@ def delete(self, files: list[Path]) -> Response:
 
 
 class AllAssetsResource(BaseMethodView):
+    """
+    Supports querying of all assets and modification of crypto assets / evm tokens.
+    TODO: Support editing of fiat assets.
```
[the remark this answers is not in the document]
nebolax: Yes, this should be done of course. But I don't think that this should be done in this PR. This PR is already big and it's main purpose is to allow import of custom assets, which is different from editing assets via an endpoint

rotkehlchen/serialization/schemas.py
```diff
@@ -1,100 +1,236 @@
...
+            data: dict[str, Any],
+            **_kwargs: Any,
+    ) -> None:
+        given_underlying_tokens = data.get('underlying_tokens', None)
+        if given_underlying_tokens is not None:
+            if given_underlying_tokens == []:
+                raise ValidationError(
+                    f'Gave an empty list for underlying tokens of {data["address"]}. '
+                    f'If you need to specify no underlying tokens give a null value',
+                )
```
LefterisJP, line 169: We can get rid of this by using `DelimitedOrNormalList` as the field for `underlying_tokens`

```suggestion
```

rotkehlchen/serialization/schemas.py
```diff
@@ -1,100 +1,236 @@
...
+        strict=True,
+        validate=webargs.validate.Range(
+            min=0,
+            max=18,
+            error='Evm token decimals should range from 0 to 18',
+        ),
+        required=True,
+    )
+    protocol = fields.String(load_default=None)
+    underlying_tokens = fields.List(fields.Nested(UnderlyingTokenInfoSchema), load_default=None)
```
LefterisJP, line 145: pseudocode - did not test

```suggestion
    underlying_tokens = DelimitedOrNormalLIst(fields.Nested(UnderlyingTokenInfoSchema), load_default=None)
```

rotkehlchen/api/v1/resources.py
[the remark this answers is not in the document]
LefterisJP: Is there an issue for this? Or will you do it right after since you noted it? This should not be forgotten but yeah no need to add it in the same PR

rotkehlchen/globaldb/assets_management.py
```diff
@@ -48,53 +46,28 @@ def import_assets_from_file(
...
-                asset_id=asset_data['identifier'],
-                asset_type=asset_type,
-                data=asset_data['extra_information'],
+    for asset in data['assets']:
+        with suppress(UnknownAsset):
+            # we expect that the asset is not in the database and this check raises UnknownAsset
+            asset.check_existence()
+            msg_aggregator.add_warning(
+                f'Tried to import existing asset {asset.identifier} with '
+                f'name {asset.name}',
```
LefterisJP, line 55: This is really hard to read code due to the way you use suppress.

Some questions:

1. Is it really important here to show a pop-up warning in the UI for each asset that already exists in the DB? What is the use case for this?
2. Clearer comment. And perhaps we need a boolean version of this. This code here is [using exceptions for control flow](https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why) in a not so straightforward way. I think a boolean function will be much better here.

```suggestion
        with suppress(UnknownAsset):
            # check if asset already exists in the DB and if yes skip importing it
            asset.check_existence(query_packaged_db=False)
            msg_aggregator.add_warning(
                f'Tried to import existing asset {asset.identifier} with '
                f'name {asset.name}',
```

rotkehlchen/tests/data/exported_assets.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"version":"3","assets":[{"asset_type":"custom asset","identifier":"e7fa04fa-2473-49c1-898c-09332d8ee3e9","name":"AAPL","custom_asset_type":"Stock","notes":"Apple Stock"},{"identifier":"b3a59173-7275-42e2-b3b1-4f5e4abd7878","name":"MIMATIC","symbol":"MAI","asset_type":"other","started":null,"forked":null,"swapped_for":null,"cryptocompare":"","coingecko":"mimatic"},{"name":"Staked SDT","symbol":"xSDT","asset_type":"evm token","started":null,"forked":null,"swapped_for":null,"cryptocompare":"","coingecko":"stake-dao","address":"0xaC14864ce5A98aF3248Ffbf549441b04421247D3","evm_chain":"ethereum","token_kind":"erc20","decimals":18,"protocol":null,"underlying_tokens":null},{"name":"1INCH Token (Staked)","symbol":"st1INCH","asset_type":"evm token","started":null,"forked":null,"swapped_for":null,"cryptocompare":"","coingecko":null,"address":"0xA0446D8804611944F1B527eCD37d7dcbE442caba","evm_chain":"ethereum","token_kind":"erc20","decimals":18,"protocol":null,"underlying_tokens":[{"address":"0x111111111117dC0aa78b770fA6A738034120C302","weight":"100","token_kind":"erc20"}]}]}
```
LefterisJP, line 1: why did you remove assets from here?

LefterisJP (commented):
Left few comments.

LefterisJP (commented):
May also need to rebase on develop

rotkehlchen/api/v1/resources.py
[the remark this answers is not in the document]
nebolax: No issue since it's small. Will do it right after

rotkehlchen/tests/data/exported_assets.json
nebolax: Ouch. I probably should not have done it. There was a problem with importing this asset in the test with the changes made in this PR and I thought that it's something specific to this asset. But it turns out that there is a problem in the test, which I will fix

rotkehlchen/globaldb/assets_management.py
nebolax: > Is it really important here to show a pop-up warning in the UI for each asset that already exists in the DB? What is the use case for this?

I think that normally all assets should be imported, and if they are not, it means a problem? For example a user might expect to see certain details of an asset after importing, but if asset with this identifier already exists, then currently existing details will be kept, and it would result in unexpected behaviour for the user.

nebolax: Also used a boolean method for this. Good idea

rotkehlchen/tests/data/exported_assets.json
```diff
@@ -1 +1 @@
...
...for":null,"cryptocompare":"","coingecko":null,"address":"0xFEEf77d3f69374f66429C91d732A244f074bdf74","evm_chain":"ethereum","token_kind":"erc20","decimals":18,"protocol":null,"underlying_tokens":[{"address":"0x3432B6A60D23Ca0dFCa7761B7ab56459D9C964D0","weight":"100","token_kind":"erc20"}]},{"identifier":"b3a59173-7275-42e2-b3b1-4f5e4abd7878","name":"MIMATIC","symbol":"MAI","asset_type":"other","started":null,"forked":null,"swapped_for":null,"cryptocompare":"","coingecko":"mimatic"},{"name":"Staked SDT","symbol":"xSDT","asset_type":"evm token","started":null,"forked":null,"swapped_for":null,"cryptocompare":"","coingecko":"stake-dao","address":"0xaC14864ce5A98aF3248Ffbf549441b04421247D3","evm_chain":"ethereum","token_kind":"erc20","decimals":18,"protocol":null,"underlying_tokens":null},{"name":"1INCH Token (Staked)","symbol":"st1INCH","asset_type":"evm token","started":null,"forked":null,"swapped_for":null,"cryptocompare":"","coingecko":null,"address":"0xA0446D8804611944F1B527eCD37d7dcbE442caba","evm_chain":"ethereum","token_kind":"erc20","decimals":18,"protocol":null,"underlying_tokens":[{"address":"0x111111111117dC0aa78b770fA6A738034120C302","weight":"100","token_kind":"erc20"}]}]}
```
LefterisJP, line 1: This is still modified in the PR.

LefterisJP (approved):
LGTM. Failing tests are due to the graph blackout. We need to get rid of this dependency.

PR #6081, opened 2023-05-15
Notify on adding empty EVM accounts

tewshi:
## Checklist

- [x] The PR modified the frontend, and updated the add evm callback to notify users when they have entered an address with no activity
- [x] Fixes issue when a user is viewing a specific blockchain account and clicks on add, should set the default blockchain to the currently visible one.

### DIFF
--- a/frontend/app/src/pages/balances/blockchain/index.vue
+++ b/frontend/app/src/pages/balances/blockchain/index.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { Blockchain } from '@rotki/common/lib/blockchain';
+import pickBy from 'lodash/pickBy';
 import { type ComputedRef, type Ref } from 'vue';
-import { type BlockchainAccountWithBalance } from '@/types/blockchain/accounts';
 
 type Intersections = {
   [key in Blockchain]: boolean;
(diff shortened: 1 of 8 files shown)


frontend/app/src/pages/balances/blockchain/index.vue
```diff
@@ -67,47 +55,16 @@ const { ksmAccounts, dotAccounts, avaxAccounts, optimismAccounts } =
...
+  // pick only intersections that are visible (at least 50%)
+  const activeObservers = pickBy(intersect, e => e);
 
-  for (const current in Blockchain) {
-    if (intersect[current as Blockchain]) {
+  for (const current in activeObservers) {
+    if (activeObservers[current]) {
       currentContext = current as Blockchain;
     }
   }
```
lukicenturi, line 70: Since the `activeObservers` are filtered with value=true above, you don't need this if anymore, don't you.
Also in this case you know the keys of `activeObservers`, but you will only used the last key of it, isn't it?

If that's the case, will this code become more efficient?
```
const context: ComputedRef<Blockchain> = computed(() => {
  const intersect = get(intersections);
  // pick only intersections that are visible (at least 50%)
  const activeObservers = Object.keys(pickBy(intersect, e => e));
  if (activeObservers.length > 0) {
    return activeObservers[activeObservers.length - 1] as Blockchain;
  }

  return Blockchain.ETH;
});
```

lukicenturi (approved):
LGTM!
Functionaity also OK

kelsos (approved):
lgtm @tewshi thank you :)

PR #6082, opened 2023-05-15
Hacky import of custom assets

### LINKED ISSUE #6073: Cannot import custom asset

nebolax:
Closes #6073 

## Checklist


rotkehlchen/globaldb/handler.py
```diff
@@ -770,11 +790,18 @@ def check_asset_exists(
...
+            query = cursor.execute(
+                'SELECT identifier from assets WHERE type=? AND name=?;',
+                (AssetType.CUSTOM_ASSET.serialize_for_db(), name),
+            )
+        else:
+            query = cursor.execute(
+                'SELECT A.identifier from assets AS A JOIN common_asset_details as C '
+                'ON A.identifier=C.identifier WHERE A.type=? AND A.name=? AND C.symbol=?;',
+                (asset_type.serialize_for_db(), name, symbol),
+            )
```
yabirgb, line 803: If you check where this is used at the moment of importing assets we can just use `add_custom_asset` that has `_raise_if_custom_asset_exists`

rotkehlchen/globaldb/handler.py
```diff
@@ -257,14 +258,33 @@ def add_setting_value(name: str, value: Any) -> None:
...
+                    )
+                    write_cursor.execute(
+                        'INSERT INTO custom_assets(identifier, notes, type) VALUES(?, ?, ?)',
+                        (asset_id, asset.notes, asset.custom_asset_type),
+                    )
+            except sqlite3.IntegrityError as e:
+                raise InputError(
+                    f'Failed to add asset {asset_id} into the assets table due to {e!s}',
+                ) from e
+            return
```
yabirgb, line 286: This already exists at `add_custom_asset`

rotkehlchen/globaldb/handler.py
```diff
@@ -257,14 +258,33 @@ def add_setting_value(name: str, value: Any) -> None:
...
+            data: Union[EvmToken, CustomAsset, dict[str, Any]],
     ) -> None:
         """
         Add an asset in the DB. Either an ethereum token or a user asset.
 
         If it's a user asset the data should be typed. As given in by marshmallow.
 
         May raise InputError in case of error, meaning asset exists or some constraint hit"""
+        if asset_type == AssetType.CUSTOM_ASSET:
+            # TODO: change later to a normal (non-hacky) approach
```
yabirgb, line 270: You are writing again code that we already have at `import_assets_from_file`

rotkehlchen/tests/data/exported_assets.json
```diff
@@ -1 +1,86 @@
...
+{
+    "version": "3",
+    "assets": [
+        {
+            "identifier": "e7fa04fa-2473-49c1-898c-09332d8ee3e9",
+            "asset_type": "custom asset",
+            "name": "AAPL",
+            "custom_asset_type": "Stock",
```
yabirgb, line 8: This file should be minified

yabirgb (commented):
The PR contains duplicated code

rotkehlchen/globaldb/handler.py
nebolax: I would not use `add_custom_asset` since it requires user db handler, and here we only have global db handler

rotkehlchen/globaldb/handler.py
nebolax: Same reply as above

rotkehlchen/tests/data/exported_assets.json
nebolax: Would it make much sense? It is not a huge file and having it the way it is right now makes it more readable.

rotkehlchen/globaldb/handler.py
yabirgb: But you shouldn't use it here. Instead of modifying this functions you should modify `import_assets_from_file` and it will be cleaner. Please check that function

rotkehlchen/globaldb/handler.py
yabirgb: You didn't read this comment

nebolax: What exact code do we have there? I don't get

rotkehlchen/globaldb/assets_management.py
```diff
@@ -53,26 +54,31 @@ def import_assets_from_file(
...
-                f'name {asset_data["name"]}',
-            )
-            continue
-
         try:
+            if asset_type == AssetType.CUSTOM_ASSET:
+                custom_assets_db = DBCustomAssets(db_handler)
+                custom_assets_db.add_custom_asset(asset_data['extra_information'])
+                continue
+
```
yabirgb, line 62: this try statement is big and contains things that shouldn't be here. I guess that the only reason is to avoid duplication of the error message. I would have break it in two but since we are changing this part of the code maybe it makes sense to split later

yabirgb (commented):
I believe it has been broken since a few releases. It should have a changelog entry but can be done in a different PR

rotkehlchen/globaldb/assets_management.py
LefterisJP: yes seems like it. We should not be adding stuff in a big try. But if it will change again anyway very soon it's not critical

rotkehlchen/tests/data/exported_assets.json
```diff
@@ -1 +1 @@
...
...s": [{"address": "0x3432B6A60D23Ca0dFCa7761B7ab56459D9C964D0","weight": "100","token_kind": "erc20"}]},{"identifier": "b3a59173-7275-42e2-b3b1-4f5e4abd7878","name": "MIMATIC","symbol": "MAI","asset_type": "other","started": null,"forked": null,"swapped_for": null,"cryptocompare": "","coingecko": "mimatic"},{"identifier": "eip155:1/erc20:0xaC14864ce5A98aF3248Ffbf549441b04421247D3","name": "Staked SDT","symbol": "xSDT","asset_type": "evm token","started": null,"forked": null,"swapped_for": null,"cryptocompare": "","coingecko": "stake-dao","address": "0xaC14864ce5A98aF3248Ffbf549441b04421247D3","evm_chain": "ethereum","token_kind": "erc20","decimals": 18,"protocol": null,"underlying_tokens": null},{"identifier": "eip155:1/erc20:0xA0446D8804611944F1B527eCD37d7dcbE442caba","name": "1INCH Token (Staked)","symbol": "st1INCH","asset_type": "evm token","started": null,"forked": null,"swapped_for": null,"cryptocompare": "","coingecko": null,"address": "0xA0446D8804611944F1B527eCD37d7dcbE442caba","evm_chain": "ethereum","token_kind": "erc20","decimals": 18,"protocol": null,"underlying_tokens": [{"address": "0x111111111117dC0aa78b770fA6A738034120C302","weight": "100","token_kind": "erc20"}]}]}
```
LefterisJP, line 1: This is not minified. It has spaces after every `:`

LefterisJP (approved):
lgtm

PR #6090, opened 2023-05-16
feat: adds etherescan link for erc20 tokens on asset breakdown

kelsos:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/pages/assets/[identifier].vue
+++ b/frontend/app/src/pages/assets/[identifier].vue
@@ -29,10 +29,20 @@ const toggleIgnoreAsset = async () => {
 
 const premium = usePremium();
 
-const { assetName, assetSymbol, assetInfo } = useAssetInfoRetrieval();
+const { assetName, assetSymbol, assetInfo, tokenAddress } =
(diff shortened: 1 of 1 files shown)


frontend/app/src/pages/assets/[identifier].vue
```diff
@@ -2,6 +2,7 @@
 import { type RawLocation } from 'vue-router';
 import { type ComputedRef } from 'vue';
 import { type AssetBalanceWithPrice } from '@rotki/common';
+import { useSupportedChains } from '@/composables/info/chains';
```
Comment (lukicenturi) [inline frontend/app/src/pages/assets/[identifier].vue:5]: remove the import

Comment (kelsos) [reply frontend/app/src/pages/assets/[identifier].vue:5]: 😄 damn we all have the same problem with Webstorm, I wonder how much effort it would be to add a linter check for this

tewshi (approved):
nos show stopper:
I usually like the flow of handle bad cases first then finally return good case.

so, 
```typescript
if (!evmChain) {
  return undefined;
}

return getChain(evmChain);
```

overall lgtm!

PR #6091, opened 2023-05-16
Allow editing of optimism etherscan

nebolax:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/api/v1/schemas.py
+++ b/rotkehlchen/api/v1/schemas.py
@@ -39,6 +39,7 @@ from rotkehlchen.chain.ethereum.modules.eth2.constants import CPT_ETH2
 from rotkehlchen.chain.ethereum.modules.nft.structures import NftLpHandling
 from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
 from rotkehlchen.chain.evm.types import EvmAccount
+from rotkehlchen.chain.optimism.constants import OPTIMISM_ETHERSCAN_NODE_NAME
 from rotkehlchen.chain.substrate.types import SubstrateAddress, SubstratePublicKey
 from rotkehlchen.chain.substrate.utils import (
     get_substrate_address_from_public_key,
@@ -2854,6 +2855,10 @@ class RpcAddNodeSchema(Schema):
 
 
 class RpcNodeEditSchema(RpcAddNodeSchema):
+    def __init__(self, dbhandler: 'DBHandler') -> None:
+        super().__init__()
+        self.dbhandler = dbhandler
+
     name = fields.String(
         required=True,
         validate=webargs.validate.NoneOf(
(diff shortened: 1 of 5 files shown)


rotkehlchen/api/v1/schemas.py
```diff
@@ -2869,13 +2873,17 @@ def validate_schema(
             data: dict[str, Any],
             **_kwargs: Any,
     ) -> None:
-        if data['identifier'] == 1 and data['name'] != ETHEREUM_ETHERSCAN_NODE_NAME:
+        etherscan_node_name = self.dbhandler.check_if_is_etherscan_node(data['identifier'])
+        if etherscan_node_name is not None and data['name'] != etherscan_node_name:
```
yabirgb, line 2877: here becomes difficult to read due the name. You expect it to be a boolean the output but is `None` or string.

Why not make a function that returns a set of the etherscan names (query by lengh in sqlite or in python) and 

```suggestion
        etherscan_rpc_names = self.dbhandler.get_etherscan_rpc_names(data['identifier'])
        if etherscan_node_name is not None and data['name'] not in etherscan_rpc_names
```

nebolax: In this case `etherscan_rpc_names` will be a list/set with always at max 1 element. This looks even more confusing to me. I have some other idea in mind on how to make this more readable and I will try to implement it.

rotkehlchen/api/v1/schemas.py
```diff
@@ -2869,18 +2874,25 @@ def validate_schema(
...
-            )
-        # verify that if the node is not etherscan the endpoint field has valid information
-        if data['identifier'] != 1 and len(data['endpoint'].strip()) == 0:
+        if len(data['endpoint'].strip()) != 0:
+            # if it is not etherscan (if the endpoint is not empty), skip.
+            # We only apply validation rules to etherscan modification requests.
+            return
+
+        is_etherscan = self.dbhandler.is_etherscan_node(data['identifier'])
+        if is_etherscan is False:
```
yabirgb, line 2883: this is

```suggestion
        if self.dbhandler.is_etherscan_node(data['identifier']) is False:
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -2869,18 +2874,25 @@ def validate_schema(
...
-                field_name='name',
-            )
-        # verify that if the node is not etherscan the endpoint field has valid information
-        if data['identifier'] != 1 and len(data['endpoint'].strip()) == 0:
+        if len(data['endpoint'].strip()) != 0:
+            # if it is not etherscan (if the endpoint is not empty), skip.
+            # We only apply validation rules to etherscan modification requests.
+            return
+
+        is_etherscan = self.dbhandler.is_etherscan_node(data['identifier'])
```
yabirgb, line 2882: this is a different check. By doing this you could allow to replace the etherscan endpoint. I believe it wouldn't have any effect but I'm not sure of the side effects that it have. before there was an extra check for the identifier

rotkehlchen/api/v1/schemas.py
```diff
@@ -2869,18 +2874,25 @@ def validate_schema(
...
+        if is_etherscan is False:
+            # expected it to be etherscan, but since it's not, then the endpoint can't be empty
             raise ValidationError(
```
yabirgb, line 2885: this makes the code confusing since it points the reader to a non existent case to say something different

```suggestion
            # since it's not etherscan then the endpoint can't be empty
            raise ValidationError(
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -2869,18 +2874,25 @@ def validate_schema(
...
                 message="endpoint can't be empty",
                 field_name='endpoint',
             )
 
+        if data['name'] not in (ETHEREUM_ETHERSCAN_NODE_NAME, OPTIMISM_ETHERSCAN_NODE_NAME):
```
yabirgb, line 2890:
```suggestion
        # else
        if data['name'] not in (ETHEREUM_ETHERSCAN_NODE_NAME, OPTIMISM_ETHERSCAN_NODE_NAME):
```

rotkehlchen/db/dbhandler.py
```diff
@@ -3104,6 +3104,17 @@ def rebalance_rpc_nodes_weights(
             new_weights,
         )
 
+    def is_etherscan_node(self, node_identifier: int) -> bool:
+        """
+        Checks if a a given node is an etherscan node (etehreum, optimism, etc).
+        Returns node name if it is an etherscan node and None otherwise.
```
yabirgb, line 3110: needs adjustment

yabirgb (commented):
Still not there

rotkehlchen/tests/api/test_misc.py
```diff
@@ -349,6 +349,18 @@ def test_manage_ethereum_nodes(rotkehlchen_api_server):
         },
     )
     assert_proper_response_with_result(response)
+    response = requests.patch(
```
LefterisJP, line 352: What does this test for? That the request does not fail? Then add a comment.

LefterisJP (commented):
small comments. I can probably merge and handle them at changelog addition PR.

rotkehlchen/tests/api/test_misc.py
yabirgb: Yes. testing that the request doesn't fail as is an allowed change. Changing the etherscan optimism rpc node

PR #6092, opened 2023-05-16
More eth2 shit

LefterisJP:
#### [Remove all, all zero daily stats entries from exited validators](https://github.com/rotki/rotki/commit/c7c1c85b44c61ae796e4e739d497e6d31685661e) 

This should no longer happen after 1.28 but exits were possible before
1.28, so users can have gathered many zero daily stats rows if they
have exited their validators.


rotkehlchen/db/eth2.py
```diff
@@ -126,18 +126,21 @@ def get_validator_daily_stats_and_limit_info(
...
          - How many are the total entries found for the filter
+         - Sum of ETH gained/lost for the filter
         """
         stats = self.get_validator_daily_stats(cursor, filter_query=filter_query)
         query, bindings = filter_query.prepare(with_pagination=False)
-        query = 'SELECT COUNT(*) from eth2_daily_staking_details ' + query
-        count = cursor.execute(query, bindings).fetchone()[0]
-        return stats, count
+        # TODO: A weakness of this query is that it does not take into account the
+        # ownership proportion of the validator in the PnL here
```
yabirgb, line 140: should this go into the docs or an issue? I believe it will affect the users that track profits for rocket eth validators

LefterisJP: Issue. I want to make one

PR #6104, opened 2023-05-17
Correctly set version for data migrations

yabirgb:
Fix an issue counting the last applied data migration version


LefterisJP (approved):
LGTM

PR #6105, opened 2023-05-17
Three last minute backend changes

LefterisJP:
#### [Handle data update if info can't be reached](https://github.com/rotki/rotki/commit/cdd8832a21d6cf512a6034ed408da277ad1ec2f0) 

Do not let it get into an endless update loop

#### [History events for free users should now respect limit](https://github.com/rotki/rotki/commit/f6e521966e1c221aa2ae19bef5e501dfa30da36d) 

- The limit is now applied on the grouped events
- The limit does not take into account the ignored/spam assets as that
  would be both unfair and confusing

The limiting logic has become a bit ugly and we should improve in the future.

### DIFF
--- a/rotkehlchen/db/filtering.py
+++ b/rotkehlchen/db/filtering.py
@@ -292,7 +292,20 @@ class DBFilterQuery():
             with_pagination: bool = True,
             with_order: bool = True,
             with_group_by: bool = False,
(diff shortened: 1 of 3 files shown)


yabirgb (approved):
LGTM

PR #6120, opened 2023-05-18
refactor: make exclusion filter more generic

### LINKED ISSUE #6079: Make exclusion filter more generic and not tied to the history filters
## Abstract
Follow up from https://github.com/rotki/rotki/pull/6075

Code in question: 
https://github.com/rotki/rotki/blob/7066b4c6debff1ffda9d166a2e83a317d0819645/frontend/app/src/composables/filters/events.ts#L210

## Motivation

We should be able to make exclusion more generic so that we can use it in other places in the future.

The filters should not need special code to handle exclusion, anything should be handled in the table filter probably. 

## Specification

- Make the mechanism generic and available for all filters

lukicenturi:
Closes #6079

## Checklist


kelsos (approved):
lgtm

PR #6121, opened 2023-05-18
Add decoder for compunding CRV in convex

yabirgb:
The current decoded didn't support compunding events and was raising a key error. This PR covers the correct decoding of the event

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* :bug:`-` Users will now be able to decode compounding transactions for Convex gauges.
+
 * :release:`1.28.0 <2023-05-17>`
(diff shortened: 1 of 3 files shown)


docs/changelog.rst
```diff
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* :bug:`-` Users will now be able to decode compounding transactions for Convex gauges.
```
LefterisJP, line 5: Isn't this a feature? Every decoder addition we make is not a bugfix. It's still fine to put it in bugifxes branch though not really needed as we should keep it for bugs.

```suggestion
* :feature:`-` Users will now be able to decode compounding transactions for Convex gauges.
```

LefterisJP (approved):
LGTM with a small nitpick which I want you to address in a subsequent PR. The failing test is unrelated.

PR #6123, opened 2023-05-19
Base polygon POS code

nebolax:
This PR introduces basic polygon code. It essentially copies the code that we have for optimism (node inquirer, chain manager, etc.) and adjusts it for polygon.


rotkehlchen/types.py
```diff
@@ -212,7 +213,7 @@ class ChainID(Enum):
     OPTIMISM = 10
     BINANCE = 56
     GNOSIS = 100
-    MATIC = 137
+    POLYGON_POS = 137
```
yabirgb, line 216: Do we use the string version? I see that we do it when serializing assets so it might be an issue to change this without caution. Have you checked the places where this might create an issue?

rotkehlchen/chain/polygon_pos/etherscan.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(
+            self,
+            database: 'DBHandler',
+            msg_aggregator: 'MessagesAggregator',
+    ) -> None:
+        super().__init__(
+            database=database,
+            msg_aggregator=msg_aggregator,
+            chain=SupportedBlockchain.POLYGON_POS,
+            base_url='polygonscan.com',
```
yabirgb, line 22: have you checked if the api endpoints provide everything we need?

rotkehlchen/api/rest.py
```diff
@@ -4377,6 +4379,7 @@ def get_types_mappings(self) -> Response:
             'per_protocol_mappings': {
                 'ethereum': self.rotkehlchen.chains_aggregator.ethereum.transactions_decoder.events_types_tuples,  # noqa: E501
                 'optimism': self.rotkehlchen.chains_aggregator.optimism.transactions_decoder.events_types_tuples,  # noqa: E501
+                'polygon_pos': self.rotkehlchen.chains_aggregator.polygon_pos.transactions_decoder.events_types_tuples,  # noqa: E501
```
yabirgb, line 4382: maybe we should be consistent here and use `SupportedBlockchain` so is constant

rotkehlchen/constants/assets.py
```diff
@@ -30,7 +30,8 @@
 A_USDT = Asset('eip155:1/erc20:0xdAC17F958D2ee523a2206206994597C13D831ec7')
 A_USDC = Asset('eip155:1/erc20:0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48')
 A_TUSD = Asset('eip155:1/erc20:0x0000000000085d4780B73119b644AE5ecd22b376')
-A_MATIC = Asset('eip155:1/erc20:0x7D1AfA7B718fb893dB30A3aBc0Cfc608AaCfeBB0')
+A_MATIC_ETH = Asset('eip155:1/erc20:0x7D1AfA7B718fb893dB30A3aBc0Cfc608AaCfeBB0')
+A_MATIC_POLYGON_POS = Asset('eip155:137/erc20:0x0000000000000000000000000000000000001010')
```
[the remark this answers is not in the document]
nebolax: Perhaps makes sense, yes

rotkehlchen/constants/location_details.py
```diff
@@ -51,4 +51,5 @@
     Location.OKX: LocationDetails(image='okx.svg'),
     Location.ETHEREUM: LocationDetails(image='ethereum.svg'),
     Location.OPTIMISM: LocationDetails(image='optimism.svg'),
+    Location.POLYGON_POS: LocationDetails(image='polygon_pos.svg'),
```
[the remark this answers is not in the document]
nebolax: Yes, I wanted to coordinate with frontend on this later

rotkehlchen/tests/utils/database.py
```diff
@@ -43,6 +43,10 @@ def maybe_include_etherscan_key(db: DBHandler, include_etherscan_key: bool) -> N
         service=ExternalService.OPTIMISM_ETHERSCAN,
         api_key=ApiKey('IK3GCCMPQXTRIEGBAXQ9DUFWIUR524K3MW'),
     )])
+    db.add_external_service_credentials([ExternalServiceApiCredentials(
+        service=ExternalService.POLYGON_POS_ETHERSCAN,
+        api_key=ApiKey('1M4TM28QKJHED9QPDWXFCBEX5CK5ID3ESG'),
```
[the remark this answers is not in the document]
nebolax: Okay

rotkehlchen/types.py
nebolax: From what I checked it has no issues since the only usage of this constant was in assets mappings

rotkehlchen/chain/polygon_pos/node_inquirer.py
```diff
@@ -0,0 +1,105 @@
...
+            greenlet_manager=greenlet_manager,
+            database=database,
+            etherscan=etherscan,
+            blockchain=SupportedBlockchain.POLYGON_POS,
+            etherscan_node=POLYGON_POS_ETHERSCAN_NODE,
+            etherscan_node_name=POLYGON_POS_ETHERSCAN_NODE_NAME,
+            contracts=contracts,
+            connect_at_start=connect_at_start,
+            rpc_timeout=rpc_timeout,
+            contract_multicall=...,  # type: ignore[arg-type]  # TODO: find out and add later
```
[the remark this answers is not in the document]
nebolax: I would prefer to handle these contracts later when we introduce specific logic for querying balances and other stuff. The contract that Pablo found doesn't match the contract deployed in Ethereum mainnet

rotkehlchen/chain/polygon_pos/constants.py
```diff
@@ -0,0 +1,26 @@
+from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
+from rotkehlchen.chain.evm.types import NodeName, WeightedNode
+from rotkehlchen.constants.misc import ONE, ZERO
+from rotkehlchen.types import SupportedBlockchain, Timestamp, deserialize_evm_tx_hash
+
+ARCHIVE_NODE_CHECK_ADDRESS = ZERO_ADDRESS  # TODO: Find out later
+ARCHIVE_NODE_CHECK_BLOCK = 0  # TODO: Find out later
+ARCHIVE_NODE_CHECK_EXPECTED_BALANCE = ZERO  # TODO: Find out later
+
+PRUNED_NODE_CHECK_TX_HASH = deserialize_evm_tx_hash('0x' + '0' * 64)  # TODO: Find out later
```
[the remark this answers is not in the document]
nebolax: Can also happen this way, yes

rotkehlchen/db/drivers/gevent.py
```diff
@@ -37,6 +37,7 @@
 import logging
 
 logger: 'RotkehlchenLogger' = logging.getLogger(__name__)  # type: ignore
+logger.trace = lambda msg: None  # type: ignore
```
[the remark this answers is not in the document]
nebolax: oh, yes, thanks. was a temporary mock

rotkehlchen/chain/aggregator.py
```diff
@@ -1252,7 +1269,7 @@ def get_loopring_balances(self) -> dict[CryptoAsset, Balance]:
 
     def get_chain_manager(
             self,
-            blockchain: Literal[SupportedBlockchain.ETHEREUM, SupportedBlockchain.KUSAMA, SupportedBlockchain.POLKADOT, SupportedBlockchain.AVALANCHE, SupportedBlockchain.OPTIMISM],  # noqa: E501
+            blockchain: Literal[SupportedBlockchain.ETHEREUM, SupportedBlockchain.KUSAMA, SupportedBlockchain.POLKADOT, SupportedBlockchain.AVALANCHE, SupportedBlockchain.OPTIMISM, SupportedBlockchain.POLYGON_POS],  # noqa: E501
```
LefterisJP, line 1272: Wuld make sense to make this into a constant no?Or isn't it already one?

rotkehlchen/chain/evm/contracts.py
```diff
@@ -263,6 +263,10 @@ def abi(self: 'EvmContracts[Literal[ChainID.ETHEREUM]]', name: 'ETHEREUM_KNOWN_A
     def abi(self: 'EvmContracts[Literal[ChainID.OPTIMISM]]', name: Literal['to', 'do', 'CTOKEN']) -> list[dict[str, Any]]:  # noqa: E501
         ...
 
+    @overload
+    def abi(self: 'EvmContracts[Literal[ChainID.POLYGON_POS]]', name: Literal['']) -> list[dict[str, Any]]:  # noqa: E501
```
LefterisJP, line 267: Can you also correct the Optimism one? It has a sneaky 'TODO'. And CTOKEN seems wrong. IF there are no abis we ask from there just use empty string literal there too.

But we probably should have a list per chain? Shouldn't we?

rotkehlchen/chain/evm/types.py
```diff
@@ -27,7 +27,7 @@ class NodeName(NamedTuple):
     name: str
     endpoint: str
     owned: bool
-    blockchain: Literal[SupportedBlockchain.ETHEREUM, SupportedBlockchain.KUSAMA, SupportedBlockchain.POLKADOT, SupportedBlockchain.AVALANCHE, SupportedBlockchain.OPTIMISM]  # noqa: E501
+    blockchain: Literal[SupportedBlockchain.ETHEREUM, SupportedBlockchain.KUSAMA, SupportedBlockchain.POLKADOT, SupportedBlockchain.AVALANCHE, SupportedBlockchain.OPTIMISM, SupportedBlockchain.POLYGON_POS]  # noqa: E501
```
LefterisJP, line 30: This should also be a constant and probably same as the one I mentioned before?

rotkehlchen/chain/polygon_pos/node_inquirer.py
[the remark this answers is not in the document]
LefterisJP: We need the multicall2 if I recall correctly

rotkehlchen/chain/polygon_pos/node_inquirer.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # -- Implementation of EvmNodeInquirer base methods --
+
+    def query_highest_block(self) -> BlockNumber:
+        block_number = self.etherscan.get_latest_block_number()
+        log.debug('Polygon POS highest block result', block=block_number)
+        return BlockNumber(block_number)
+
+    def _is_pruned(self, web3: Web3) -> bool:
+        return False  # TODO: find out a rule for determining pruned nodes in polygon POS
```
LefterisJP, line 64: Both of these _is_pruned and _have_archive should be easy to do. Why not add them? At this point so far it just seems a skeleton PR with lots of TODOs.

The problem is that any future reviews will become **really** hard like this. These 2 TODOs are trivial to implement. Just do them.

rotkehlchen/chain/polygon_pos/tokens.py
```diff
@@ -0,0 +1,19 @@
...
+    from .node_inquirer import PolygonPOSInquirer
+
+
+class PolygonPOSTokens(EvmTokens):
+
+    def __init__(self, database: 'DBHandler', polygon_pos_inquirer: 'PolygonPOSInquirer') -> None:
+        super().__init__(database=database, evm_inquirer=polygon_pos_inquirer)
+
+    # -- methods that need to be implemented per chain
+    def _get_token_exceptions(self) -> list[ChecksumEvmAddress]:
```
LefterisJP, line 18: Hmmmm again here you don't do anything and you don't even leave a TODO.

This we will just forget it even needed to do something.

```python
    def _get_token_exceptions(self) -> list[ChecksumEvmAddress]:
        exceptions = [A_OPTIMISM_ETH.resolve_to_evm_token().evm_address]
            ignored_asset_ids = self.db.get_ignored_asset_ids(cursor=cursor)

        # TODO: Shouldn't this query be filtered in the DB?
        for asset_id in ignored_asset_ids:  # don't query for the ignored tokens
            if (evm_details := asset_id_is_evm_token(asset_id)) is not None and evm_details[0] == ChainID.OPTIMISM:  # noqa: E501
                exceptions.append(evm_details[1])

        return exceptions
```

As you can see above the code for this in optimism is also simple. And it's even a sort of a copy of the mainnet ones. So I think you should even abstract it.

rotkehlchen/constants/location_details.py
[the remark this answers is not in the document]
LefterisJP: But again here there is no TODO. 

Hmm okay no this is not gonna work. I can't review it like this and be sure we won't miss or forget anything.

LefterisJP (commented):
Did not finish the review but don't like the approach of just adding this skeleton with TODOs in some places and forgetting in others.

rotkehlchen/chain/evm/contracts.py
nebolax: Yeah, we have a list for ethereum. But for optimism and polygon there are no abis used atm, so I guess a list does not make sense there yet

rotkehlchen/chain/polygon_pos/tokens.py
nebolax: Makes sense

rotkehlchen/chain/polygon_pos/node_inquirer.py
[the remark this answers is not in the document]
nebolax: Okay I found multicall2.

https://polygonscan.com/address/0x275617327c958bd06b5d6b871e7f491d76113dd8#code

rotkehlchen/chain/polygon_pos/node_inquirer.py
```diff
@@ -0,0 +1,105 @@
...
+        self.etherscan = cast(PolygonPOSEtherscan, self.etherscan)
+
+    # -- Implementation of EvmNodeInquirer base methods --
+
+    def query_highest_block(self) -> BlockNumber:
+        block_number = self.etherscan.get_latest_block_number()
+        log.debug('Polygon POS highest block result', block=block_number)
+        return BlockNumber(block_number)
+
+    def _is_pruned(self, web3: Web3) -> bool:
```
LefterisJP, line 73: Hmm I know it's like this now but wouldn't all these be better abstracted if all you need to provide is the tx hash here and the block , address and expected balance in archive? Since we seem to repeat functionality here and in all node_inquirers.

rotkehlchen/rotkehlchen.py
```diff
@@ -353,6 +355,13 @@ def unlock_user(
             connect_at_start=optimism_nodes,
         )
         optimism_manager = OptimismManager(optimism_inquirer)
+        polygon_pos_nodes = self.data.db.get_rpc_nodes(blockchain=SupportedBlockchain.POLYGON_POS, only_active=True)  # noqa: E501
```
LefterisJP, line 358: You need to add a POLYGON_TODO to populate those nodes too. Most probably at the DB upgrade. And we will need to also update our default nodes.

rotkehlchen/tests/data/v37_rotkehlchen.db
LefterisJP, line 1: how did you create this?

rotkehlchen/tests/fixtures/blockchain.py
```diff
@@ -186,6 +188,12 @@ def fixture_optimism_mock_data():
     return {}
 
 
+@pytest.fixture(name='polygon_pos_mock_data')
```
LefterisJP, line 191: With the way we have switched to VCR we probably don't need all these. Both for optimism and polygon.

nebolax: Hmm but I see it still being used in certain ethereum tests. Maybe we could handle this change later in a separate PR since it's not something polygon-specific?

rotkehlchen/data/global.db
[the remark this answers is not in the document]
nebolax: Only new eth scan and multicall2 contracts

```sql
INSERT INTO contract_data(rowid,address,chain_id,abi,deployed_block) VALUES(138,'0x275617327c958bD06b5D6b871E7f491D76113dd8',137,52,16253032);
INSERT INTO contract_data(rowid,address,chain_id,abi,deployed_block) VALUES(139,'0x2aB513B211C801673758D1C32815605B5289ad29',137,33,43047682);
```

rotkehlchen/chain/polygon_pos/tokens.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class PolygonPOSTokens(EvmTokens):
+
+    def __init__(self, database: 'DBHandler', polygon_pos_inquirer: 'PolygonPOSInquirer') -> None:
+        super().__init__(database=database, evm_inquirer=polygon_pos_inquirer)
+
+    # -- methods that need to be implemented per chain
+    def _per_chain_token_exceptions(self) -> list[ChecksumEvmAddress]:
+        return [A_POLYGON_POS_MATIC.resolve_to_evm_token().evm_address]
```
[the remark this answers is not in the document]
nebolax: Yes. If you query `balanceOf` of this token https://polygonscan.com/address/0x0000000000000000000000000000000000001010#readContract#F6 for some address, it will return the amount of MATIC owned by that address on polygon.

rotkehlchen/tests/fixtures/blockchain.py
LefterisJP: yes can definitely move later

LefterisJP: it is used in some tests, yes but it's all old tests before vcr. No new tests like that should be written. So you definitely don't need to add these fixtures for polygon

rotkehlchen/data/global.db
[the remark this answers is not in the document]
LefterisJP: You will also need to add those new contracts in the upgrade. (perhaps we need a global DB upgrade then?)

Otherwise upgrading users won't get these contracts, right? Or will they get them from the packaged DB? In which case it's fine to not do anything.

rotkehlchen/chain/polygon_pos/node_inquirer.py
nebolax: Makes sense

rotkehlchen/chain/evm/tokens.py
```diff
@@ -371,9 +371,21 @@ def query_tokens_for_addresses(
...
+        with self.db.conn.read_ctx() as cursor:
+            ignored_asset_ids = self.db.get_ignored_asset_ids(cursor=cursor)
+
+        # TODO: Shouldn't this query be filtered in the DB?
+        exceptions = []
+        for asset_id in ignored_asset_ids:  # don't query for the ignored tokens
+            if (evm_details := asset_id_is_evm_token(asset_id)) is not None and evm_details[0] == self.evm_inquirer.chain_id:  # noqa: E501
+                exceptions.append(evm_details[1])
+
+        return list(set(exceptions + self._per_chain_token_exceptions()))
```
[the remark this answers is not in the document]
nebolax: Makes sense, thanks

rotkehlchen/chain/evm/proxies_inquirer.py
```diff
@@ -50,6 +54,9 @@ def get_account_proxy(self, address: ChecksumEvmAddress) -> Optional[ChecksumEvm
         - BlockchainQueryError if an evm node is used and the contract call
         queries fail for some reason
         """
+        if self.dsproxy_registry is None:
```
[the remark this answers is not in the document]
nebolax: Totally makes sense

rotkehlchen/chain/ethereum/node_inquirer.py
```diff
@@ -97,6 +94,7 @@ def __init__(
         self.blocks_subgraph = Graph('https://api.thegraph.com/subgraphs/name/blocklytics/ethereum-blocks')  # noqa: E501
         self.etherscan = cast(EthereumEtherscan, self.etherscan)
         self.ens_reverse_records = self.contracts.contract(string_to_evm_address('0x3671aE578E63FdF66ad4F3E12CC0c0d71Ac7510C'))  # noqa: E501
+        self.proxies_inquirer: EvmProxiesInquirer  # To specify that it's not None
```
LefterisJP, line 97: That's still not the way to do classes.

rotkehlchen/chain/evm/tokens.py
```diff
@@ -231,7 +231,9 @@ def _compute_detected_tokens_info(self, addresses: Sequence[ChecksumEvmAddress])
         It generates a dictionary where key is an address and value is
         either a list of tokens or the timestamp of the last tokens query for that address.
         """
-        proxies_mapping = self.evm_inquirer.proxies_inquirer.get_accounts_having_proxy()
+        proxies_mapping = {}
+        if self.evm_inquirer.proxies_inquirer is not None:
```
LefterisJP, line 235: with the other approach this would be an isinstance() on the evm inquirer

LefterisJP (commented):
All right so some more comments left but most are centered around the DSProxy registry.

1. Keep in mind the DSProxy is not the only proxy registry out there. There are more. So we may need to have a more generic approach in the future.

2. Looking at how you did it, it's not terrible or anything but for the reasons I stated in the various comments here and there that it may make more sense to do the following:
    - Introduce two new abstract classes. `EvmNodeInquirerWithDSProxy` and `EVMTransactionDecoderWithDSProxy`. They should extend the old ones and have all proxy registry functionality embedded in them.
    - Mainnet and Optimism classes should inherit from them and should function as before.
    - Polygon should inherit from the base one and as such it will not have any ds proxy functionality.
    - Any place that works on the base class should check if they need to do proxy work with `isinstance()` checks.

Does that make sense?

rotkehlchen/chain/aggregator.py
```diff
@@ -889,6 +894,18 @@ def query_optimism_balances(
...
+    @cache_response_timewise()
+    def query_polygon_pos_balances(
+            self,  # pylint: disable=unused-argument
+            # Kwargs here is so linters don't complain when the "magic" ignore_cache kwarg is given
+            **kwargs: Any,
+    ) -> None:
+        """
+        Queries all the polygon pos balances and populates the state.
+        Same potential exceptions as ethereum
+        """
```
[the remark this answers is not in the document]
nebolax: Ouch

rotkehlchen/chain/ethereum/defi/defisaver_proxy.py
```diff
@@ -20,7 +20,7 @@ class HasDSProxy(EthereumModule):
 
     def __init__(
             self,
-            ethereum_inquirer: 'EvmNodeInquirer',
+            ethereum_inquirer: 'EthereumInquirer',
```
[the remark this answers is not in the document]
LefterisJP: As this is also initiated in Optimism this does not make sense. Please put it as it was.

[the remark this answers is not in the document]
nebolax: Ok

rotkehlchen/chain/aggregator.py
LefterisJP, line 907: At this point I have to wonder if you are messing with me @nebolax.

This is the 3rd time I ask you to add a TODO here.

https://github.com/rotki/rotki/pull/6123#discussion_r1202793174

You can follow the links all to the first comment.
```suggestion
        POLYGON_TODO
        """
```

rotkehlchen/chain/ethereum/defi/defisaver_proxy.py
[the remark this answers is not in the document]
LefterisJP: Okay so ... why is it is still changed?

rotkehlchen/chain/evm/decoding/base.py
```diff
@@ -298,3 +297,30 @@ def make_event_next_index(
             address=address,
             extra_data=extra_data,
         )
+
+
+class BaseDecoderToolsWithDSProxy(BaseDecoderTools):
+    """A class that keeps a common state and offers some common decoding functionality"""
```
LefterisJP, line 303:
```suggestion
    """Like BaseDecoderTools but with DSProxy evm inquirers"""
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -145,12 +146,21 @@ def __init__(
...
         self.dbevmtx = DBEvmTx(self.database)
         self.dbevents = DBHistoryEvents(self.database)
-        self.base = BaseDecoderTools(
-            database=database,
-            evm_inquirer=self.evm_inquirer,
-            is_non_conformant_erc721_fn=self._is_non_conformant_erc721,
-            address_is_exchange_fn=self._address_is_exchange,
-        )
+        self.base: BaseDecoderTools
+        if isinstance(self.evm_inquirer, EvmNodeInquirerWithDSProxy):
```
LefterisJP, line 150: Since you have a different class `EVMTransactionDecoderWithDSProxy` you do not need this if in here as this goes again against the entire point of using classes.

Pass the BaseDecoder tools as an argument and initialize the proper type at the appropriate chain. So ethereum would use BaseDecoderToolsWithDSProxy, Polygon without

rotkehlchen/chain/aggregator.py
nebolax: Found out what's happened. I actually added POLYGON_TODO there after both the first and the second comments. But when I later ran `make format`, ruff removed it.

The way I was added the todo is by adding the following comment `# POLYGON_TODO: implement`, so the function would look like
```python
@protect_with_lock()
    @cache_response_timewise()
    def query_polygon_pos_balances(
            self,  # pylint: disable=unused-argument
            # Kwargs here is so linters don't complain when the "magic" ignore_cache kwarg is given
            **kwargs: Any,
    ) -> None:
        """
        Queries all the polygon pos balances and populates the state.
        Same potential exceptions as ethereum
        """
        # POLYGON_TODO: implement
```

When I later ran ruff, it considered the commented out todo line as commented out code (which is probably valid because a line of code like `some_variable: TypeOfTheVariable` is a valid line of code), and if it is run with `--fix` (which happens in `make format`), it automatically removes that commented out code.

-------------

I now added it in the docstring and it's fine.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -145,12 +147,11 @@ def __init__(
...
-            database=database,
-            evm_inquirer=self.evm_inquirer,
-            is_non_conformant_erc721_fn=self._is_non_conformant_erc721,
-            address_is_exchange_fn=self._address_is_exchange,
-        )
+        self.base: BaseDecoderTools
+        self.base = base_tools
+        # if isinstance(self.evm_inquirer, EvmNodeInquirerWithDSProxy):
+        #     self.base = BaseDecoderToolsWithDSProxy(
+        #     self.base = BaseDecoderTools(
```
LefterisJP, line 154: Dude

```suggestion
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -145,12 +147,8 @@ def __init__(
...
         self.dbevmtx = DBEvmTx(self.database)
         self.dbevents = DBHistoryEvents(self.database)
-        self.base = BaseDecoderTools(
-            database=database,
-            evm_inquirer=self.evm_inquirer,
-            is_non_conformant_erc721_fn=self._is_non_conformant_erc721,
-            address_is_exchange_fn=self._address_is_exchange,
-        )
+        self.base: BaseDecoderTools
+        self.base = base_tools
```
LefterisJP, line 151: This is probably not needed as it's exactly the same as the one you assign to

```suggestion
        self.base = base_tools
```

LefterisJP (approved):
LGTM with one note. Please handle this note in another PR, either you or someone else.

PR #6124, opened 2023-05-19
chore: major dependency updates ts5, electron 24

kelsos:
Closes #(issue_number)

## Checklist


lukicenturi:
@kelsos can you run it with node 18?
I got error on the dev-proxy with node 18, I think it's because `http-proxy-middleware` hasn't supported node 18 yet.

Everything is fine with node 16.

kelsos:
> @kelsos can you run it with node 18? I got error on the dev-proxy with node 18, I think it's because `http-proxy-middleware` hasn't supported node 18 yet.
> 
> Everything is fine with node 16.

Let me see I might have had the proxy disabled

kelsos:
> > @kelsos can you run it with node 18? I got error on the dev-proxy with node 18, I think it's because `http-proxy-middleware` hasn't supported node 18 yet.
> > Everything is fine with node 16.
> 
> Let me see I might have had the proxy disabled it might be related to the e2e changes I had to do where I had to specify the listening address as either `127.0.0.1` or `localhost`

So that's it, I am not sure what changed to resolution in newer node, but unless you specify 127.0.0.1 which is the backend's listening IP it won't accept connections.

It could be that the changed something and localhost uses the IPv6 instead. That would explain the problem.

lukicenturi (approved):
LGTM! works well now, thanks

PR #6127, opened 2023-05-22
feat: update validation UX

lukicenturi:
Closes #(issue_number)

## Checklist


lukicenturi:
@kelsos I create a small implementation example for the `<external-trade-form />`. If it's good, I will update the rest.

frontend/app/src/components/accounts/manual-balances/ManualBalancesForm.vue
```diff
@@ -177,18 +171,10 @@ const openCustomAssetForm = async () => {
...
-  set(customAssetFormSaving, true);
-
-  set(customAssetFormSaving, true);
-  const identifier = await get(customAssetFormRef)?.save();
-
-  if (identifier) {
-    set(showCustomAssetForm, false);
-    set(asset, identifier);
+const customAssetSaved = async (assetId: string) => {
+  if (assetId) {
```
kelsos, line 175: Do we emit falsy values from the components?

frontend/app/src/components/asset-manager/CustomAssetFormDialog.vue
```diff
@@ -0,0 +1,81 @@
...
+    return;
+  }
+  set(saving, true);
+  const assetId = await get(form).save();
+  if (assetId) {
+    clearDialog();
+    emit('saved', assetId);
+  }
+  set(saving, false);
+};
```
kelsos, line 52: Is there any specific reason why we can't move the `save` function back to the form itself?

frontend/app/src/components/accounts/manual-balances/ManualBalancesForm.vue
lukicenturi: no, my bad, the if is unnecessary.

frontend/app/src/components/asset-manager/CustomAssetFormDialog.vue
lukicenturi: You mean this whole function `save` right?

Because the button to `save` is in the `<big-dialog>`, not in the `form`, 
and we modify the `saving` state here (to disable the button)

lukicenturi:
Sorry I accidentally force pushed, the last thing updated only to resolve the 1 comment.

frontend/app/src/components/asset-manager/CustomAssetFormDialog.vue
kelsos: do you think we can follow a similar approach like here https://github.com/rotki/rotki/blob/201662111d866c1a27ef0dcbd38c26a070a06188/frontend/app/src/composables/accounts/dialog.ts#L11

We can extract the commonly shared logic to a composable, make it shared so that they both share the same instance, and remove the ref in the parent. Otherwise, the changes are in the right direction.

lukicenturi: Ah okay, I think we can do it better that way, will do it.

lukicenturi:
@kelsos I don't think we can use `createSharedComposable` since there are some parts that we put a form inside a form, but we need the state between those 2 forms separated.
So I decided to use vue `provide` and `inject` instead.
So the idea is provide the state on the parent, and the children can use those state too.

kelsos:
> @kelsos I don't think we can use `createSharedComposable` since there are some parts that we put a form inside a form, but we need the state between those 2 forms separated. So I decided to use vue `provide` and `inject` instead. So the idea is provide the state on the parent, and the children can use those state too.

I am not sure why the `createSharedComposable` can't work. These are not the same forms right? so each subform would have a different sharedComposable.

Do we use the same form inside the same form?

frontend/app/src/composables/form.ts
```diff
@@ -0,0 +1,94 @@
...
+  const formComposable = useForm<T>();
+  provide(CONTEXT, formComposable);
+  return formComposable;
+};
+
+export const getInjectedForm = <T>(): InjectedForm<T> => {
+  const form = inject<InjectedForm<T>>(CONTEXT);
+  assert(form, "form isn't provided by parent component");
+  return form;
+};
```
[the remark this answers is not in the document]
kelsos: This will work by wrapping `useForm` in a `createSharedComposable` for each different cache.

frontend/app/src/components/accounts/manual-balances/ManualBalancesForm.vue
```diff
@@ -159,11 +188,13 @@ watch(label, label => {
   get(v$).label.$validate();
 });
 
-const customAssetFormRef: Ref<InstanceType<typeof CustomAssetForm> | null> =
-  ref(null);
-const showCustomAssetForm: Ref<boolean> = ref(false);
-const customAssetFormValid: Ref<boolean> = ref(false);
-const customAssetFormSaving: Ref<boolean> = ref(false);
+const { setOpenDialog, setPostSubmitFunc } = useCustomAssetForm();
```
kelsos, line 191: Now it's more clear what each thing is. Thank you :)

kelsos (approved):
@lukicenturi I am approving since this is mostly ok.

There is one thing we need to take care but can be done in a separate PR.

image

The save button is enabled when adding a validator and clicking does not validate, resulting in a stuck dialog.

Can you make a PR to fix that?

Also, we now have some dialogs with a button enabled and some without, is it too much to make sure that all of them are consistent (and maybe only disable when loading, or running tasks)

PR #6128, opened 2023-05-22
feat(electron): make window initial size proportional

### LINKED ISSUE #5920: frontend(electron): make the window initial size proportional to screen
## Abstract

Make sure that the initial window size when the application is opened for the first time is relative to the screen site.

We can keep the current default value as the minimum (base) and then calculate the percentage this default has on a basic 15" laptop screen `1366 x 768` screen resolution.

Then based on the screen size make the window that percentage of the active resolution assuming that it is bigger than `1366 x 768`.

## Motivation

The initial window might appear too small when opened on a retina or 4k screen.

## Specification

- Make sure that the initial screen size is proportional to the screen
- Make sure that when re-opening the app the window restores to the last known size (`electron-window-state` should take care of it just make sure it does)

lukicenturi:
Closes #5920

## Checklist

### DIFF
--- a/frontend/app/src/background.ts
+++ b/frontend/app/src/background.ts
@@ -1,5 +1,5 @@
 import path from 'node:path';
-import { BrowserWindow, Menu, MenuItem, app, protocol } from 'electron';
+import { BrowserWindow, Menu, MenuItem, app, protocol, screen } from 'electron';
 import windowStateKeeper from 'electron-window-state';
 import { ipcSetup } from '@/electron-main/ipc-setup';
 import { getUserMenu } from '@/electron-main/menu';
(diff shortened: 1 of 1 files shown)


lukicenturi:
to reset the saved size, go to `/Users/{username}/Library/Application Support/rotki` and delete `rm -rf window-state.json`

kelsos:
@lukicenturi could you read the linked issue again? 

What I describe in the issue is not making the window full-screen for every screen size.

lukicenturi:
> @lukicenturi could you read the linked issue again?
> 
> What I describe in the issue is not making the window full-screen for every screen size.

This should be updated now

kelsos (approved):
lgtm @lukicenturi thanks

PR #6129, opened 2023-05-22
Remove Aave subgraph from the code. Part 1

yabirgb:
What this PR does:

- Remove subgraph queries (not used) from liquity module
- Remove all code related to subgraphs (Aave)
- Remove the `/blockchains/ETH/modules/aave/history` endpoint (needs to be removed in the frontend too)
- Removes related tests from the code

What still needs to be done:

- [ ] Do a DB upgrade removing the db events and queried ranges. Was not done here because there is another PR that needs a db upgrade and also want to fetch real data for this (currently subgraph is down)
- [ ] Add decoders for liquidations
- [ ] Create a stats endpoint for total lost/gained in liquidations and total lost/earned by borrowing


nebolax:
There are also subgraphs used in:
1. Uniswap
2. Sushiswap
3. Compound
4. Balancer
5. Yearn

6. This method
```python
def _get_blocknumber_by_time_from_subgraph(self, ts: Timestamp) -> int:
        """Queries Ethereum Blocks Subgraph for closest block at or before given timestamp"""
        response = self.blocks_subgraph.query(
            f"""
            {{
                blocks(
                    first: 1, orderBy: timestamp, orderDirection: desc,
                    where: {{timestamp_lte: "{ts}"}}
                ) {{
                    id
                    number
                    timestamp
                }}
            }}
            """,
        )
        try:
            result = int(response['blocks'][0]['number'])
        except (IndexError, KeyError) as e:
            raise RemoteError(
                f'Got unexpected ethereum blocks subgraph response: {response}',
            ) from e
        else:
            return result
```

Shouldn't all these (as well as base thegraph code in `rotkehlchen/chain/ethereum/graph.py`) be removed too?

rotkehlchen/chain/ethereum/modules/aave/aave.py
```diff
@@ -73,21 +59,6 @@ def __init__(
...
-            )
-        except RemoteError as e:
-            self.msg_aggregator.add_error(
-                f'Could not initialize the Aave subgraph due to {e!s}. '
-                f' All aave historical queries are not functioning until this is fixed. '
-                f'Probably will get fixed with time. If not report it to rotkis support channel ',
-            )
-            raise ModuleInitializationFailure('Aave subgraph remote error') from e
-
         self.history_lock = Semaphore()
```
LefterisJP, line 62: Remove this lock. It's not used anywhere now.

```suggestion
```

LefterisJP (approved):
LGTM. The unused lock can be removed in another PR.

PR #6130, opened 2023-05-22
Eliminate circular imports in decoder/accountant module import

LefterisJP:
In the past we dynamically imported each decoder/accountant by the module automatically.

This created a serious problem with circular imports when importing other things from the same module/subdirectory such as constants.

To solve this we now do not automatically import anything from __init__.py but instead search for either a decoder.py or an accountant.py module underneath the package and if it exists we find the decoder class in there and import it.


LefterisJP:
Will merge. Only 1 test failed due to presumably graph problems.

PR #6131, opened 2023-05-22
User approval to resume from user db backup

dimyG:
Addresses issue #5221 

* If there is an ongoing upgrade and the user tries to login he is prompted to resume from the latest user db backup. In that case, the frontend should show two options to the user: 1. Resume from backup  2. Abort login.
* a resume_from_backup argument was added to user login endpoint
* a test case was added

Closes #(issue_number)

## Checklist

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -186,11 +186,13 @@ Handling user creation, sign-in, log-out and querying
 
       {
           "password": "supersecurepassword",
-          "sync_approval": "unknown"
+          "sync_approval": "unknown",
+          "resume_from_backup": false
       }
 
    :reqjson string password: The password that unlocks the account
    :reqjson bool sync_approval: A string denoting if the user approved an initial syncing of data from premium. Valid values are ``"unknown"``, ``"yes"`` and ``"no"``. Should always be ``"unknown"`` at first and only if the user approves should a login with approval as ``"yes`` be sent. If he does not approve a login with approval as ``"no"`` should be sent. If there is the possibility of data sync from the premium server and this is ``"unknown"`` the login will fail with an appropriate error asking the consumer of the api to set it to ``"yes"`` or ``"no"``.
+   :reqjson bool resume_from_backup: An optional boolean denoting if the user approved a resume from backup. This is used to handle the case where the encrypted user database is in a semi upgraded state during user login. If not given, the value defaults to ``false`` so if a semi upgraded database exists during login, the consumer of the api will receive a response with a status '300 Multiple Choices', an explanatory message and an empty result. If the value is ``true`` then the latest backup will be used and the login will proceed as usual.
 
    **Example Response**:
 
@@ -235,7 +237,11 @@ Handling user creation, sign-in, log-out and querying
 
    :resjson object result: For successful requests, result contains the currently connected exchanges,and the user's settings. For details on the user settings refer to the `Getting or modifying settings`_ section.
    :statuscode 200: Logged in successfully
-   :statuscode 300: Possibility of syncing exists and the login was sent with sync_approval set to ``"unknown"``. Consumer of api must resend with ``"yes"`` or ``"no"``. In this case the result will contain an object with a payload for the message under the ``result`` key and the message under the ``message`` key. The payload has the following keys: ``local_size``, ``remote_size``, ``local_last_modified``, ``remote_last_modified``.
+   :statuscode 300: This would be the response status in the following two cases:
+
+        - There is an unfinished upgrade to the encrypted user database, and the login was sent with resume_from_backup set to ``false``. The consumer of the api can resend with resume_from_backup set to ``true``, so that the user will login using the latest backup of the encrypted database. In this case the response will contain an empty ``result`` key and an explanatory ``message`` on the message key.
+        - Possibility of syncing exists and the login was sent with sync_approval set to ``"unknown"``. Consumer of api must resend with ``"yes"`` or ``"no"``. In this case the result will contain an object with a payload for the message under the ``result`` key and the message under the ``message`` key. The payload has the following keys: ``local_size``, ``remote_size``, ``local_last_modified``, ``remote_last_modified``.
+
    :statuscode 400: Provided JSON is in some way malformed
    :statuscode 401: Provided password is wrong for the user or some other authentication error.
    :statuscode 409: Another user is already logged in. User does not exist. There was a fatal error during the upgrade of the DB. Permission error while trying to access the directory where rotki saves data.
--- a/rotkehlchen/db/dbhandler.py
+++ b/rotkehlchen/db/dbhandler.py
@@ -91,7 +91,11 @@ from rotkehlchen.db.utils import (
     replace_tag_mappings,
     str_to_bool,
 )
-from rotkehlchen.errors.api import AuthenticationError, IncorrectApiKeyFormat
+from rotkehlchen.errors.api import (
+    AuthenticationError,
+    IncorrectApiKeyFormat,
+    RotkehlchenPermissionError,
+)
 from rotkehlchen.errors.asset import UnknownAsset, UnsupportedAsset
 from rotkehlchen.errors.misc import (
     DBUpgradeError,
@@ -180,6 +184,7 @@ class DBHandler:
             msg_aggregator: MessagesAggregator,
             initial_settings: Optional[ModifiableDBSettings],
             sql_vm_instructions_cb: int,
+            resume_from_backup: bool,
     ):
         """Database constructor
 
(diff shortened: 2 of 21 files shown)


rotkehlchen/db/dbhandler.py
```diff
@@ -209,6 +214,7 @@ def __init__(
         self.get_or_create_evm_token_lock = Semaphore()
         self.password = password
         self._connect()
+        self.resume_from_backup = resume_from_backup  # before calling _check_unfinished_upgrades
```
nebolax, line 217: What is the reason for storing this as a variable? Isn't it used just a single time?

rotkehlchen/rotkehlchen.py
```diff
@@ -265,10 +266,11 @@ def unlock_user(
             sync_approval=sync_approval,
             sync_database=sync_database,
             initial_settings=initial_settings,
+            resume_from_backup=resume_from_backup,
         )
 
         # unlock or create the DB
-        self.user_directory = self.data.unlock(user, password, create_new, initial_settings)
+        self.user_directory = self.data.unlock(user, password, create_new, initial_settings, resume_from_backup)  # noqa: E501
```
nebolax, line 273: kwargs could be added here and it this call could be split in multiple lines

nebolax (commented):
Have you checked whether the same functionality can be achieved with globaldb upgrades?

dimyG:
> Have you checked whether the same functionality can be achieved with globaldb upgrades?

Probably it can. The difference is that the globaldb is initialized at startup and it is outside of the api. There is no api call for that so the approval should come as a backend initialization parameter (similar to sqlite_instructions). Not quite sure though yet, how we can properly catch an exception and return something to the frontend because as I mentioned were are outside of the api.

rotkehlchen/errors/api.py
```diff
@@ -7,7 +7,7 @@ class RotkehlchenPermissionError(Exception):
     The payload contains information to be shown to the user by the frontend so
     they can decide what to do
     """
-    def __init__(self, error_message: str, payload: Optional[dict[str, Any]]) -> None:
+    def __init__(self, error_message: str, payload: Optional[dict[str, Any]] = None) -> None:
```
LefterisJP, line 10: Why do you need to add the default value here? Explicit is better than implicit. What does this have to do with the goal of your PR?

rotkehlchen/tests/api/test_users.py
```diff
@@ -631,6 +631,46 @@ def test_user_login(rotkehlchen_api_server, username, db_password, data_dir):
...
+    data = {'password': db_password, 'sync_approval': 'unknown', 'async_query': True, 'resume_from_backup': False}  # noqa: E501
+    response = requests.post(
+        api_url_for(rotkehlchen_api_server, 'usersbynameresource', name=username),
+        json=data,
+    )
+    task_id = assert_ok_async_response(response)
+    response_data = wait_for_async_task(rotkehlchen_api_server, task_id)
+    assert_error_async_response(
+        response_data=response_data,
+        contained_in_msg='Resume from a backup or quit login',
```
LefterisJP, line 666: you will need to edit this after changing the message as I asked above

rotkehlchen/tests/api/test_users.py
```diff
@@ -631,6 +631,46 @@ def test_user_login(rotkehlchen_api_server, username, db_password, data_dir):
...
+        status_code=HTTPStatus.MULTIPLE_CHOICES,
+        result_exists=True,
+    )
+    # Notice: Any login after this point will require a user db backup to be resumed from.
+    # If you need to login without a backup write your test case before the point at which
+    # the user db upgrade status is set to ongoing.
```
LefterisJP, line 672: Why add this comment and not just resume from a backup normally? This also should be part ofthe test otherwise the test is incomplete.

rotkehlchen/tests/fixtures/rotkehlchen.py
```diff
@@ -165,11 +166,23 @@ def mock_get_settings(_cursor) -> DBSettings:
 
     original_unlock = rotki.data.unlock
 
-    def augmented_unlock(user, password, create_new, initial_settings):
+    def augmented_unlock(
+            username: str,
+            password: str,
+            create_new: bool,
+            initial_settings: Optional[ModifiableDBSettings] = None,
+            resume_from_backup: bool = True,
```
LefterisJP, line 174: same thing as before for this argument

LefterisJP (commented):
How is the user prompted in this case? Shouldn't the frontend also make any changes here so that they:

(1) show when there is a problem and we need it to relogin with resume_from_backup = True
(2) Actually call login again with resume_from_backup=Tre

dimyG:
> 

Yes, frontend changes are required. I'm thinking it like this: 

- The user logs in as usual.
- The frontend sends a request to the login endpoint with resume_from_backup=False (or not sending it at all, it defaults to False)
- If everything is ok and no ongoing upgrade is happening, user logs in normally. 
- If there is an ongoing upgrade the login endpoint returns the message "`The encrypted database is in a half-upgraded state. Either resume from a backup or solve the issue manually.`" with an http status code 300 (multiple choices). 
- In that case, the front end should read this and show a message to the user while providing two buttons. One is "Resume from latest backup" and one is "Abort login" (or something like that). 
- If the user presses the "Resume from latest backup" the frontend sends a normal request to the login endpoint but with the resume_from_backup=True. The user logins with the backup user db. 
- If the user presses "Abort login" no request is made to the backend. He is just shown a message to fix manually or something and he is back to the normal login view.

rotkehlchen/db/dbhandler.py
```diff
@@ -233,7 +238,16 @@ def _check_unfinished_upgrades(self) -> None:
         if ongoing_upgrade_from_version is None:
             return  # We are all good
 
-        # Otherwise replace the db with a backup and reconnect
+        # if there is an unfinished upgrade, check user approval to resume from backup
+        if resume_from_backup is False:
+            raise RotkehlchenPermissionError(
+                error_message='The encrypted database is in a half-upgraded state. '
+                              'Either resume from a backup or solve the issue manually.',
```
LefterisJP, line 245: It's a pity we don't have a way of enforcing this via a linter but this is not how we handle multiline strings. Please do this instead:

Also please notice all the changes my diff had in the last review.

`half -> semi`

```suggestion
                error_message=(
                    'The encrypted database is in a semi upgraded state. '
                    'Either resume from a backup or solve the issue manually.'
                ),
```

or you can also just make it all ine one line if its' not too huge and add `#noqa: E501` at the end

rotkehlchen/tests/api/test_users.py
```diff
@@ -631,6 +634,81 @@ def test_user_login(rotkehlchen_api_server, username, db_password, data_dir):
...
+    response = requests.post(
+        api_url_for(rotkehlchen_api_server, 'usersbynameresource', name=username),
+        json=data,
+    )
+    # And make sure it works
+    task_id = assert_ok_async_response(response)
+    result = wait_for_async_task_with_result(rotkehlchen_api_server, task_id)
+    check_proper_unlock_result(result)
+    assert rotki.user_is_logged_in is True
+    users_data = check_user_status(rotkehlchen_api_server)
```
LefterisJP, line 699: here you should also check tha this is the DB with the `VALUES("is_backup", "Yes")` in its settings. You actually went into the trouble of doing it to identify the DB properly so I suppose you just forgot to check it here?

LefterisJP (commented):
Okay You will also need to:

1. Edit the api docs to explain the new way login should work and what the new arguments do and what their possible valies are
2. Coordinate with the frontend to let them know what they need to do to complete the issue from their side. Your suggestion [here](https://github.com/rotki/rotki/pull/6131#issuecomment-1559837107) makes sense to me.

docs/api.rst
```diff
@@ -235,7 +237,11 @@ Handling user creation, sign-in, log-out and querying
...
+        - Possibility of syncing exists and the login was sent with sync_approval set to ``"unknown"``. Consumer of api must resend with ``"yes"`` or ``"no"``. In this case the result will contain an object with a payload for the message under the ``result`` key and the message under the ``message`` key. The payload has the following keys: ``local_size``, ``remote_size``, ``local_last_modified``, ``remote_last_modified``.
+
    :statuscode 400: Provided JSON is in some way malformed
    :statuscode 401: Provided password is wrong for the user or some other authentication error.
    :statuscode 409: Another user is already logged in. User does not exist. There was a fatal error during the upgrade of the DB. Permission error while trying to access the directory where rotki saves data.
```
LefterisJP, line 247: you will need to add the possible statuscode of 300 here too

PR #6139, opened 2023-05-23
Add decoding for StakeDAO bribe reward claims

LefterisJP:


LefterisJP:
Will merge. Only test failure is due to 502 from blockchain.info

PR #6140, opened 2023-05-24
fix: shorten alias name

lukicenturi:
Shorten alias name for very long ENS name, for example `0x65aac64335bcae4573254794682f6bbd596ecaa3.eth`


kelsos (approved):
lgtm

PR #6143, opened 2023-05-24
Allow filtering by contract address and product

yabirgb:
This PR:

- Adds a new endpoint to retrieve the available products and the counterparties using them
- Allows to filter by the address field for evm events


docs/api.rst
```diff
@@ -12466,6 +12467,54 @@ Getting all available counterparties
...
+  **Example Response**
+
+  .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+        "result":{
+          "mappings":{
```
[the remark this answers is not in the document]
yabirgb: Luki said that there is no special requirement and atm we don't have any restriction that would make us consider a split per chain on the products. Will leave as it is now

rotkehlchen/api/rest.py
```diff
@@ -4353,6 +4354,22 @@ def get_evm_counterparties_details(self) -> Response:
             status_code=HTTPStatus.OK,
         )
 
+    def get_evm_products(self) -> Response:
+        """
+        Collect the mappings of counterparties to the products they list
+        """
+        ethereum_counterparties = self.rotkehlchen.chains_aggregator.ethereum.transactions_decoder.possible_products  # noqa: E501
```
nebolax, line 4361: One idea. Instead having a variable per chain we could iterate over all evm transaction decoders in a loop. We already have a constant with evm chains that have a transactions decoder.

LefterisJP: This is not just a good to have idea. This is a must. Otherwise we will end up forgetting to add new stuff here each time a new evm chain is added. i.e. just now with polygon.

So in short: Iterate over all evm chains from the chain aggregator, and get their decoders and then get the possible products.

rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -85,6 +86,11 @@ def possible_events(self) -> DecoderEventMappingType:
         """Return the possible event types and subtypes used in the specific decoder"""
         return {}
 
+    @staticmethod
+    def possible_products() -> dict[str, list[EvmProduct]]:
+        """Returns the possible evmproducts associated to a counterparty"""
```
LefterisJP, line 91:
```suggestion
        """Returns a mapping of counterparty to possible evmproducts associated with it for the decoder."""
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -165,6 +166,8 @@ def __init__(
         self.decoders: dict[str, 'DecoderInterface'] = {}
         # store the mapping of possible counterparties to the allowed types and subtypes in events
         self.events_types_tuples: DecoderEventMappingType = {}
+        # store the mapping of possible counterparties to their evm products
+        self.possible_products: dict[str, list[EvmProduct]] = {}
```
LefterisJP, line 170: Why store those?

I would say to just calculate them on the fly. The api will be called only once and then they will be stored in the frontend. So keeping them forever in memory here makes no sense.

rotkehlchen/db/filtering.py
```diff
@@ -1092,7 +1093,15 @@ def make(
             filter_query.filters.append(DBMultiBytesFilter(
                 and_op=True,
                 column='tx_hash',
-                values=tx_hashes,  # type: ignore  # EVMTxHash is equal to bytes
+                values=cast(list[bytes], tx_hashes),  # EVMTxHash is equal to bytes
```
LefterisJP, line 1096: keep the tx ignore. As cast is kind of silly and is an actual function also in production: https://docs.python.org/3/library/typing.html#typing.cast
```suggestion
                values=tx_hashes,  # type: ignore # EVMTxHash is equal to bytes
```

I mean it's super minimal, but why put it here when comment + type: ignore achieve the same thing. I don't want us to add extra code just to just appease mypy

rotkehlchen/db/filtering.py
```diff
@@ -1092,7 +1093,15 @@ def make(
...
+                operator='IN',
+            ))
+
+        if addresses is not None:
+            filter_query.filters.append(DBMultiStringFilter(
+                and_op=True,
+                column='address',
+                values=cast(list[str], addresses),  # ChecksumEvmAddress is equal to str
```
LefterisJP, line 1104: same here please

rotkehlchen/tests/api/test_ethereum_transactions.py
```diff
@@ -1330,11 +1332,12 @@ def test_events_filter_params(rotkehlchen_api_server, ethereum_accounts, start_w
     tx2 = make_ethereum_transaction(tx_hash=b'2', timestamp=2)
     tx3 = make_ethereum_transaction(tx_hash=b'3', timestamp=3)
     tx4 = make_ethereum_transaction(tx_hash=b'4', timestamp=4)
-    event1 = make_ethereum_event(tx_hash=b'1', index=1, asset=A_ETH, timestamp=TimestampMS(1), location_label=ethereum_accounts[0])  # noqa: E501
+    test_contract_address = string_to_evm_address('0xc37b40ABdB939635068d3c5f13E7faF686F03B65')
```
LefterisJP, line 1335: why not use make_evm_address from factories here? Is there any reason to have this specific address?

rotkehlchen/tests/api/test_misc.py
```diff
@@ -433,6 +436,7 @@ def test_events_mappings(rotkehlchen_api_server_with_exchanges):
     Test different mappings and information that we provide for rendering events information
     - Test that the structure for types mappings is correctly generated
     - Test that the valid locations are correctly provided to the frontend
+    - Test the the products are correctly returned
```
LefterisJP, line 439:
```suggestion
    - Test that the products are correctly returned
```

LefterisJP (commented):
Left some comments. Also coordinate with the frontend on the work they need to do.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -224,25 +225,38 @@ def _recursively_initialize_decoders(
...
+    def get_decoders_products(self) -> dict[str, list[EvmProduct]]:
+        """Get the list of possible products"""
+        possible_products: dict[str, list[EvmProduct]] = {}
+        for decoder in self.decoders.values():
+            possible_products |= decoder.possible_products()
+
+        return possible_products
+
+    def get_decoders_event_types(self) -> DecoderEventMappingType:
+        """Get the mappings of counterparties to their possible even type
```
LefterisJP, line 244:
```suggestion
        """Get the mappings of counterparties to their possible event type
```

LefterisJP (approved):
LGTM with one nitpick but tests still fail, so please look at it.

LefterisJP:
Merging. The uniswap subgraph has an indexing error.

PR #6146, opened 2023-05-24
Remove compound graph queries

yabirgb:
This PR removes the subgraph queries from compound.

TODOS:

- [ ] Add compound liquidations
- [ ] Notify frontend to remove calls to `/blockchains/ETH/modules/compound/history` (they shouldn't be calling it but just to confirm)
- [ ] Remove the endpoint from the docs. I forgot it here but is a docs update only change


LefterisJP (approved):
LGTM

PR #6152, opened 2023-05-26
feat: ask user approval from user db backup

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/account-management/LoginForm.vue
+++ b/frontend/app/src/components/account-management/LoginForm.vue
@@ -1,18 +1,13 @@
 <script setup lang="ts">
 import useVuelidate from '@vuelidate/core';
 import { helpers, required, requiredIf } from '@vuelidate/validators';
-import {
-  type LoginCredentials,
-  type SyncApproval,
-  type SyncConflict
-} from '@/types/login';
+import { type LoginCredentials, type SyncApproval } from '@/types/login';
 
 const { t } = useI18n();
 
 const props = withDefaults(
   defineProps<{
     loading: boolean;
-    syncConflict: SyncConflict;
     errors?: string[];
   }>(),
   {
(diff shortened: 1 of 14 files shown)


lukicenturi:
e2e test failed, we need the changes for the assets first

LefterisJP:
> e2e test failed, we need the changes for the assets endpoint first

Seems to be green now. Did you ask kelsos for review?

frontend/app/src/components/account-management/LoginForm.vue
```diff
@@ -434,19 +438,68 @@ const login = async (syncApproval: SyncApproval = 'unknown') => {
...
                   </v-btn>
                 </v-col>
               </v-row>
             </v-alert>
           </transition>
 
+          <transition>
+            <v-alert
+              v-if="!!halfUpgradeConflict.message"
+              class="animate half__upgrade-error mt-8"
```
kelsos, line 465: should this be `half-upgrade-error` also maybe instead of half it would make sense to use `partial`

frontend/app/src/store/session/index.ts
```diff
@@ -53,9 +59,8 @@ export const useSessionStore = defineStore('session', () => {
       return { success: true };
     } catch (e: any) {
       logger.error(e);
-      if (e instanceof SyncConflictError) {
-        set(syncConflict, { message: e.message, payload: e.payload });
-        return { success: false, message: '' };
+      if (e instanceof SyncConflictError || e instanceof HalfUpgradeError) {
+        return handleLoginError(e);
       }
```
kelsos, line 64: This way we do multiple instance of right? can't we make a handler that returns `ActionStatus | null` and only do the instance of inside the method?

kelsos (commented):
So some initial comments

Btw is there an easy way to test this?

frontend/app/src/store/tasks/index.ts
```diff
@@ -57,7 +59,11 @@ export const useTaskStore = defineStore('tasks', () => {
 
   const registerHandler = <R, M extends TaskMeta>(
     task: TaskType,
-    handlerImpl: (actionResult: ActionResult<R>, meta: M) => void,
+    handlerImpl: (
+      actionResult: ActionResult<R>,
+      meta: M,
+      error?: Error
```
kelsos, line 65: Any way to use an enriched `ActionResult` internally that also includes the error? instead of passing it as a parameter?

PR #6154, opened 2023-05-26
Reduce user DB size by removing unused internaltxs data

LefterisJP:
The block number and the timestamp are both data that are neither used nor are they needed as they are also part of the parent transaction entry.


LefterisJP:
The two failed tests here also fail in develop and only in CI: https://github.com/rotki/rotki/pull/6155

So they seem to be unrelated but not sure why. Will merge this and then try to see why this happens.

Issue #1217, opened 2020-07-25
Bitcoin.tax import


LefterisJP:
## Abstract

This was requested by @davbre in [this](https://github.com/rotki/rotki/issues/1212#issue-665581857) issue where he provided feedback. He mentioned that bitcoin.tax has a CSV export and it would be nice to provide an import functionality from it.

## Issues with importing data from external source

In this section please describe how you propose to address the problem you described earlier. Try to include example use cases that showcase the value of your proposal.

1. Generally importing data from other services via CSV is hard since the other services seem to change their export format often. This has caused us a lot of headaches with the cointracking.info imports.
2. If the data format is incomplete and missing required data then we have to extrapolate data and that is always bad for the user and would be worse than simply importing your data directly from an exchange with an API key or from the blockchain.

## Task

1. Think on the approach and if it makes sense to implement this feature
2. If this feature is implemented find someone who uses BitcoinTax to provide some sample data (obviously edited so no personal information leak) so that the import functionality can be implemented.

LefterisJP:
The user above has given an example. Implementing the import should be straightforward.

PR #6157, opened 2023-05-28
Bitcoin tax import support

dimyG:
* It reads trades csv type
* It reads income and spending csv types
* Creates rotki history events from them
* All csv file types are imported from the same interface 

Addresses #1217

## Checklist


rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,337 @@
...
+            self,
+            cursor: DBCursor,
+            csv_row: dict[str, Any],
+            csv_type: str,
+            timestamp_format: str = '%Y-%m-%d %H:%M:%S %z',
+    ) -> None:
+        """
+        Consume the file containing trades from BitcoinTax.
+        - UnknownAsset
+        - DeserializationError
```
nebolax, line 248: I would state in the docstring that this and all other relevant functions in this file can raise `UnsupportedCSVEntry` errors

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,337 @@
...
+                sequence_index=0,
+                timestamp=timestamp,
+                location=location,
+                asset=receive_asset,
+                balance=Balance(
+                    amount=receive_amount,
+                    usd_value=ZERO,
+                ),
+                notes=memo,
+                event_type=event_type,
```
nebolax, line 130: Since event type is always RECEIVE, it would be better to use it directly here instead of making a variable

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,337 @@
...
+                event_subtype=HistoryEventSubType.RECEIVE,
```
nebolax, line 131: As said in other places, for receive events we use RECEIVE + NONE types

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,337 @@
...
+            sequence_index=0,
+            timestamp=timestamp,
+            location=location,
+            asset=spend_asset,
+            balance=Balance(
+                amount=spend_amount,
+                usd_value=ZERO,
+            ),
+            notes=memo,
+            event_type=event_type,
```
nebolax, line 148: event_type is always TRADE here, isn't it?

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,337 @@
...
+            balance=Balance(
+                amount=receive_amount,
+                usd_value=ZERO,
+            ),
+            notes=memo,
+            event_type=event_type,
+            event_subtype=HistoryEventSubType.RECEIVE,
+        )
+        events.extend([spend_event, receive_event])
+        if fee_amount != ZERO and fee_asset is not None:
```
nebolax, line 166: This probably needs some special accounting treatment to be defined in order to respect `should_calculate_fee` setting

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,337 @@
...
+            quote_asset: AssetWithOracles,
+            quote_asset_amount: AssetAmount,
+            fee_asset: Optional[AssetWithOracles],
+            fee_amount: Fee,
+            memo: str,
+    ) -> None:
+        if action not in ('BUY', 'SELL', 'SWAP'):
+            raise UnsupportedCSVEntry(f'Unsupported entry {action}. Data: {csv_row}')
+
+        events: list[HistoryBaseEntry] = []
```
nebolax, line 87: Why to create this intermediary list and not use `add_history_events` directly?

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,337 @@
...
+                    f'Ignoring renamed crypto asset {receive_asset} in BitcoinTax CSV SWAP action')  # noqa: E501
+                return  # skip this csv row
+
+            receive_amount = quote_asset_amount
+            receive_event = HistoryEvent(
+                event_identifier=event_identifier,
+                sequence_index=0,
+                timestamp=timestamp,
+                location=location,
+                asset=receive_asset,
```
nebolax, line 124: If it is a forked asset and rotki knows that it was forked, then why to create a receive event? We have `handle_prefork_asset_acquisitions`

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,337 @@
...
+            location: Location,
+            action: str,
+            asset: AssetWithOracles,
+            asset_amount: AssetAmount,
+            fee_asset: Optional[AssetWithOracles],
+            fee_amount: Fee,
+            memo: str,
+    ) -> None:
+        try:
+            event_type, event_subtype = self.ACTION_TO_HISTORY_EVENT_TYPE[action]
```
nebolax, line 199: Using exceptions in places where the logic is known beforehand is an antipattern. Using `if action not in self.ACTION_TO_HISTORY_EVENT_TYPE: raise ...` will be better here

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,337 @@
...
+        except KeyError as e:
+            raise UnsupportedCSVEntry(
+                f'Unsupported entry {action}. Data: {csv_row}') from e
+
+        events: list[HistoryBaseEntry] = []
```
nebolax, line 204: Same question about using a separate `events` list

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,337 @@
...
+            location='BitcoinTax',
+        ) * 1000)
+
+        try:
+            location = Location.deserialize(csv_row['Account'])
+        except DeserializationError:
+            location = self.csv_value_to_location(csv_row['Account'])
+
+        asset_resolver = LOCATION_TO_ASSET_MAPPING.get(location, asset_from_common_names)
+        base_asset = asset_resolver(csv_row['Symbol'])
```
nebolax, line 267: I think it will make sense to use `AssetBalance` here in order to reduce the number of variables used.

Additionally for fees I think it will make sense to do the check `fee_amount != ZERO and fee_asset is not None` here and use `None` if there is no fee.

rotkehlchen/data_import/manager.py
```diff
@@ -29,6 +30,7 @@ class DataImportSource(SerializableEnumNameMixin):
     BINANCE = 9
     ROTKI_TRADES = 10
     ROTKI_EVENTS = 11
+    BITCOIN_TAX = 12
```
nebolax, line 33: api docs probably need a small adjustment

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,337 @@
...
+                event_identifier=event_identifier,
+                sequence_index=0,
+                timestamp=timestamp,
+                location=location,
+                asset=receive_asset,
+                balance=Balance(
+                    amount=receive_amount,
+                    usd_value=ZERO,
+                ),
+                notes=memo,
```
nebolax, line 129: memo is just an identifier, right? I am not sure if users want to see it in the notes. But no matter whether to keep it in the notes or not, notes should be more descriptive. Like "Spend `amount` `asset`, as imported from bitcoin tax"

nebolax (commented):
I am not sure how do we handle accounting of imported events? `HistoryEvent.process` works only with Kraken events, so I guess that some adjustments need to be made in order to support the event types and subtypes used by bitcoin tax importer.

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,337 @@
...
+from rotkehlchen.types import AssetAmount, Fee, Location, TimestampMS
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+class BitcoinTaxImporter(BaseExchangeImporter):
+
+    ACTION_TO_HISTORY_EVENT_TYPE = {
+        'INCOME': (HistoryEventType.RECEIVE, HistoryEventSubType.RECEIVE),
```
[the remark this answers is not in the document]
dimyG: For all receive events? 
```
        'INCOME': (HistoryEventType.RECEIVE, HistoryEventSubType.RECEIVE),
        'MINING': (HistoryEventType.RECEIVE, HistoryEventSubType.REWARD),
        'GIFTIN': (HistoryEventType.RECEIVE, HistoryEventSubType.NONE),
        'BORROW': (HistoryEventType.RECEIVE, HistoryEventSubType.GENERATE_DEBT),
```

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,337 @@
...
+        # use a deterministic event_identifier to avoid duplicate events in case of reimport
+        event_identifier = f'rotki_events_{self.hash_csv_row(csv_row)}'
+        timestamp = TimestampMS(deserialize_timestamp_from_date(
+            date=csv_row['Date'],
+            formatstr=timestamp_format,
+            location='BitcoinTax',
+        ) * 1000)
+
+        try:
+            location = Location.deserialize(csv_row['Account'])
```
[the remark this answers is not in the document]
dimyG: The Location enum contains all supported exchanges already plus some other locations like equities, realestate etc. so maybe it's better to use it instead?

rotkehlchen/data_import/importers/bitcoin_tax.py
dimyG: As I saw the `should_calculate_fee` is specific to the `Trade `dataclass while this one here is a HistoryEvent. But I'm not familiar with accounting so not sure about this.

rotkehlchen/data_import/importers/bitcoin_tax.py
dimyG: memo is an optional description for an entry (trade, income, expense) in bitcoin tax that has been created by the user, so I think that if we are to store it we should not change it but store it as it is in the csv so that the user can identify it if he needs to. So, I guess either store it in notes as it is or not at all?

dimyG:
> I am not sure how do we handle accounting of imported events? `HistoryEvent.process` works only with Kraken events, so I guess that some adjustments need to be made in order to support the event types and subtypes used by bitcoin tax importer.

I'm not yet familiar with accounting, but I guess that if some events created from bitcoin tax import are from kraken, they should be processed as regular kraken history events. In case of bitcoin tax these are the types currently: 

```
    ACTION_TO_HISTORY_EVENT_TYPE = {
        'INCOME': (HistoryEventType.RECEIVE, HistoryEventSubType.NONE),
        'MINING': (HistoryEventType.RECEIVE, HistoryEventSubType.REWARD),
        'GIFTIN': (HistoryEventType.RECEIVE, HistoryEventSubType.AIRDROP),
        'BORROW': (HistoryEventType.RECEIVE, HistoryEventSubType.GENERATE_DEBT),
        'SPEND': (HistoryEventType.SPEND, HistoryEventSubType.NONE),
        'DONATION': (HistoryEventType.SPEND, HistoryEventSubType.DONATE),
        'GIFT': (HistoryEventType.SPEND, HistoryEventSubType.NONE),
        'STOLEN': (HistoryEventType.SPEND, HistoryEventSubType.NONE),
        'LOST': (HistoryEventType.SPEND, HistoryEventSubType.NONE),
        'REPAY': (HistoryEventType.SPEND, HistoryEventSubType.PAYBACK_DEBT),
    }
```

and 

```
HistoryEventType.TRADE, HistoryEventSubType.SPEND
HistoryEventType.TRADE, HistoryEventSubType.RECEIVE
HistoryEventType.TRADE, HistoryEventSubType.FEE
```

Are these handled already or not? 
In any case probably this is something that requires it's own PR

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,303 @@
...
+from rotkehlchen.utils.misc import ts_sec_to_ms
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+class BitcoinTaxImporter(BaseExchangeImporter):
+    CSVType = Literal['trades', 'other_events']
+
+    ACTION_TO_HISTORY_EVENT_TYPE = {
```
nebolax, line 32: Sorry, just noticed, but we tend to defined constants at the module level and not in classes. Same with static methods

rotkehlchen/data_import/importers/bitcoin_tax.py
nebolax: Lefteris talked about forks that are exchange-specific. But here you added
```python
if receive_asset.forked is None:
                log.info(
                    f'Ignoring renamed crypto asset {receive_asset} in BitcoinTax CSV SWAP action')  # noqa: E501
                return  # skip this csv row
```

which means that by reaching the moment where the recieve even is being created, the asset is already the one that we have forked handling for, which means that there is no need to create a new event.

rotkehlchen/data_import/importers/bitcoin_tax.py
nebolax: Ah, if memo is some sort of descriptive text, then it's fine. I would do then something like `f'Imported from bitcoin tax: {memo}'`

rotkehlchen/data_import/importers/bitcoin_tax.py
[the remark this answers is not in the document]
nebolax: I am not familiar with possible locations of bitcoin tax. But if it's for crypto only then I guess locations like `equities` should not be present in the csv.

Also have you checked that all possible bitcoin tax locations are properly deserialized by rotki? Don't we need some mapping?

rotkehlchen/data_import/importers/bitcoin_tax.py
nebolax: > should_calculate_fee is specific to the Trade

It should not be specific to `Trade` dataclass since a collection of history events (2 or 3 depending on fee) can represent exactly same action as `Trade` class represents. And we are actually moving right now from `Trade`, `LedgerAction`, etc. dataclasses towards storing everything in `HistoryEvent`s.

What I just realized though is that `HistoryEvent.process` works only with kraken staking events. And I don't know how/where the rest kraken events are processed. @yabirgb do you know?

nebolax (commented):
> Are these handled already or not?

We have some default handling logic for evm events, but I don't remember anything generic for pure HistoryEvents, so I think that these accounting rules are something that should be done, but I also might have forgotten something and we might already have them implemented somewhere.

rotkehlchen/data_import/importers/bitcoin_tax.py
yabirgb: Yes, that method needs adjustment. In fact the imported events I believe they won't be correctly processed. We need more tests for accounting

rotkehlchen/data_import/importers/bitcoin_tax.py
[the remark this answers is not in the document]
dimyG: - There is a mapping containing only coinbase pro. The others are ok. 
- The location in bitcoin tax can be altered manually by the user, and the user can create entries manually and add whatever location name he wants, so it's not only for exchanges. I think that using the Location enum is more appropriate in this case.

rotkehlchen/data_import/importers/bitcoin_tax.py
dimyG: As I see in the code, the `handle_prefork_asset_acquisitions` creates `ProcessedAccountingEvent` events directly. But it doesn't create history events. Do we need to create these history events? If we need, then maybe the most suitable place to create them would be in  handle_prefork_asset_acquisitions (and not in the import). So I think that indeed, I should not create any event if a bitcoin_tax event action is swap (which means a fork).

nebolax: We should not create any history events during pnl generation. Also a history event for a fork is not needed since forks are handled fully automatically in accounting.

The only event kind that I think could make sense is an informational event, but we don't create such informational fork events in any other place, so I would better not create it.

rotkehlchen/data_import/importers/bitcoin_tax.py
nebolax: Created an issue for it https://github.com/rotki/rotki/issues/6162 and checking it

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,278 @@
...
+                timestamp=timestamp,
+                location=location,
+                action=action,
+                base_asset_balance=base_asset_balance,
+                quote_asset_balance=quote_asset_balance,
+                fee_asset_balance=fee_asset_balance,
+                memo=memo,
+            )
+            return
+        self._consume_income_spending_event(
```
LefterisJP, line 239:
```suggestion
        # else
        self._consume_income_spending_event(
```

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,278 @@
...
+                    self.db.msg_aggregator.add_warning(
+                        f'Deserialization error during BitcoinTax CSV import. '
+                        f'{e!s}. Ignoring entry',
+                    )
+                    continue
+                except UnsupportedCSVEntry as e:
+                    self.db.msg_aggregator.add_warning(str(e))
+                    continue
+                except KeyError as e:
+                    raise InputError(f'Could not find key {e!s} in csv row {row!s}') from e
```
LefterisJP, line 278: There is nothing afterwards so all these `continue` are not needed, are they?

```suggestion
                except UnknownAsset as e:
                    self.db.msg_aggregator.add_warning(
                        f'During BitcoinTax CSV import found action with unknown '
                        f'asset {e.identifier}. Ignoring entry',
                    )
                except DeserializationError as e:
                    self.db.msg_aggregator.add_warning(
                        f'Deserialization error during BitcoinTax CSV import. '
                        f'{e!s}. Ignoring entry',
                    )
                except UnsupportedCSVEntry as e:
                    self.db.msg_aggregator.add_warning(str(e))
                except KeyError as e:
                    raise InputError(f'Could not find key {e!s} in csv row {row!s}') from e
```

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -1664,3 +1665,146 @@ def assert_is_equal_history_event(
     expected_dict = expected.serialize()
     expected_dict.pop('event_identifier')
     assert actual_dict == expected_dict
+
+
+def assert_bitcoin_tax_trades_import_results(rotki: Rotkehlchen, csv_file_name: str):
```
LefterisJP, line 1670: Use literal types here for the filename as you know the possible values

LefterisJP (commented):
Left few comments.

Also:

1. Rebase on develop to fix the test failures.
2. Coordinate what's needed with the frontend to know what they need to do or if we need something on our side
3. Add changelog either here or in the frontend PR and make sure issue is closed afterwards

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,278 @@
...
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+CSVType = Literal['trades', 'other_events']
+
+ACTION_TO_HISTORY_EVENT_TYPE = {
+    'INCOME': (HistoryEventType.RECEIVE, HistoryEventSubType.NONE),
+    'MINING': (HistoryEventType.RECEIVE, HistoryEventSubType.REWARD),
+    'GIFTIN': (HistoryEventType.RECEIVE, HistoryEventSubType.AIRDROP),
```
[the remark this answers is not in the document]
dimyG: Yes, it is supposed to mean gift-in, as for a gift that you have received.

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,285 @@
...
+    deserialize_asset_amount,
+    deserialize_timestamp_from_date,
+)
+from rotkehlchen.types import Fee, Location, TimestampMS
+from rotkehlchen.utils.misc import ts_sec_to_ms
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+CSVType = Literal['trades', 'other_events']
```
LefterisJP, line 28: It's okay to do this, but also keep in mind that for a type in just 2 places it's probaly not important to avoid the duplication. Could just put it as is.

rotkehlchen/data_import/importers/bitcoin_tax.py
```diff
@@ -0,0 +1,285 @@
...
+                event_identifier=event_identifier,
+                timestamp=timestamp,
+                location=location,
+                action=action,
+                base_asset_balance=base_asset_balance,
+                quote_asset_balance=quote_asset_balance,
+                fee_asset_balance=fee_asset_balance,
+                memo=memo,
+            )
+        else:
```
LefterisJP, line 248: That's not what I said. Please read my comment here: https://github.com/rotki/rotki/pull/6157#discussion_r1210189320

I specifically asked to just add a comment with an else since it makes the code more readable. Having the implicit else without any newline space or anything is hard to read. We do this a lot in the code.

LefterisJP (approved):
LGTM. Let some nitpicks which can be handled next time. Let's see at the 90 mins mark the timeout shoudlo happen and am curious to see if it will have any data. Probably not.

PR #6168, opened 2023-05-30
Temporary fix for history events accounting

nebolax:
It's an ugly hotfix that should be reverted later

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,7 @@ Changelog
 * :bug:`-` ENS avatar fetching for each name's avatar should now work for all name resolvers. Even older or custom ones.
 * :bug:`-` Users will now be able to decode compounding transactions for Convex gauges.
 * :bug:`-` Users will no longer see double counting of balances in curve gauges.
+* :bug:`-` Events that were imported via a csv file will now be properly accounted for in PnL reports.
 * :feature:`-` The PnL report page will load faster if there is many old reports in the DB.
 * :feature:`-` Convex staking and Curve gauge balances will no longer need a manual balances referesh to be detected.
 
--- a/rotkehlchen/accounting/structures/base.py
+++ b/rotkehlchen/accounting/structures/base.py
@@ -397,6 +397,12 @@ class HistoryEvent(HistoryBaseEntry):
     def should_ignore(self, ignored_ids_mapping: dict[ActionType, set[str]]) -> bool:
         return False  # TODO: How do we ignore general history events? Not possible yet, I think
 
+    def get_type_identifier(self, include_counterparty: bool = True) -> str:  # pylint: disable=unused-argument  # noqa: E501
+        """
+        Computes the identifier from event type and event subtype.
+        """
+        return str(self.event_type) + '__' + str(self.event_subtype)
+
     def process(
             self,
             accounting: 'AccountingPot',
@@ -427,7 +433,11 @@ class HistoryEvent(HistoryBaseEntry):
             )
             return 1
 
-        return 1
+        # TODO: change to a proper solution. Right now all generic history events are processed as
+        # if they were evm events. In order to apply the correct accounting rules for HistoryEvents
+        # we use the default accounting rules that are chosen only using the type and the subtype
+        # of the event. But we should properly abstract accounting rules in the future.
+        return accounting.transactions.process(self, events_iterator)
 
 
 @dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=False)
--- a/rotkehlchen/accounting/transactions.py
+++ b/rotkehlchen/accounting/transactions.py
@@ -1,8 +1,9 @@
 import logging
 from collections.abc import Iterator
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Union
 
 from rotkehlchen.accounting.mixins.event import AccountingEventMixin, AccountingEventType
+from rotkehlchen.accounting.structures.base import HistoryBaseEntry, HistoryEvent
 from rotkehlchen.accounting.structures.evm_event import EvmEvent
 from rotkehlchen.chain.evm.accounting.structures import TxAccountingTreatment, TxEventSettings
 from rotkehlchen.logging import RotkehlchenLogsAdapter
@@ -16,19 +17,19 @@ logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
 
-def evm_events_iterator(
+def next_events_iterator(
         events_iterator: Iterator[AccountingEventMixin],
-        associated_event: EvmEvent,
-) -> Iterator[EvmEvent]:
+        associated_event: Union[EvmEvent, HistoryEvent],
+) -> Iterator[Union[EvmEvent, HistoryEvent],]:
     """
     Takes an iterator of accounting events and transforms it into a history base entries iterator.
     Takes associated event as an argument to be able to log it in case of errors.
     """
     for event in events_iterator:
-        if not isinstance(event, EvmEvent):
+        if not isinstance(event, (EvmEvent, HistoryEvent)):  # TODO accounting: undo. Is a temporary hack  # noqa: E501
             log.error(
-                f'At accounting for tx_event {associated_event.notes} with hash '
-                f'{associated_event.tx_hash.hex()} we expected to take an additional '
+                f'At accounting for tx_event {associated_event.notes} with identifier '
+                f'{associated_event.event_identifier} we expected to take an additional '
                 f'event but found a non history base entry event',
             )
             return
@@ -53,7 +54,7 @@ class TransactionsAccountant():
 
     def process(
             self,
-            event: EvmEvent,
+            event: Union[HistoryEvent, EvmEvent],
             events_iterator: Iterator[AccountingEventMixin],
     ) -> int:
         """Process a transaction event and return amount of actions consumed from the iterator"""
@@ -78,12 +79,12 @@ class TransactionsAccountant():
         if event_settings.accountant_cb is not None:
             event_settings.accountant_cb(
                 pot=self.pot,
-                event=event,
-                other_events=evm_events_iterator(events_iterator, event),
+                event=event,  # type: ignore[arg-type]
+                other_events=next_events_iterator(events_iterator, event),  # type: ignore[arg-type]  # next_events_iterator will always be yielding EvmEvents here  # noqa: E501
             )
 
         if event_settings.accounting_treatment == TxAccountingTreatment.SWAP:
-            in_event = next(evm_events_iterator(events_iterator, event), None)
+            in_event = next(next_events_iterator(events_iterator, event), None)
             if in_event is None:
                 log.error(
                     f'Tried to process accounting swap but could not find the in '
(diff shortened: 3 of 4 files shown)


rotkehlchen/accounting/transactions.py
```diff
@@ -75,15 +76,15 @@ def process(
                 return 1
 
         # if there is any module specific accountant functionality call it
-        if event_settings.accountant_cb is not None:
+        if event_settings.accountant_cb is not None and isinstance(event, EvmEvent):
```
yabirgb, line 79: Isn't `accountant_cb is None == True` for anything else than `EvmEvent`?

rotkehlchen/accounting/transactions.py
```diff
@@ -75,15 +76,15 @@ def process(
...
             event_settings.accountant_cb(
                 pot=self.pot,
                 event=event,
                 other_events=evm_events_iterator(events_iterator, event),
             )
 
         if event_settings.accounting_treatment == TxAccountingTreatment.SWAP:
-            in_event = next(evm_events_iterator(events_iterator, event), None)
+            in_event = next(evm_events_iterator(events_iterator, event), None)  # type: ignore[arg-type] # noqa: E501
```
yabirgb, line 87: why not change the type in `evm_events_iterator`?

rotkehlchen/accounting/transactions.py
```diff
@@ -108,15 +109,17 @@ def process(
             taxable=event_settings.taxable,
             count_entire_amount_spend=event_settings.count_entire_amount_spend,
             count_cost_basis_pnl=event_settings.count_cost_basis_pnl,
-            extra_data={'tx_hash': event.tx_hash.hex()},
+            extra_data={
+                'tx_hash': event.tx_hash.hex() if isinstance(event, EvmEvent) else event.event_identifier,  # TODO: undo. Is a temporary hack  # noqa: E501
```
yabirgb, line 113: Same with the todo. Unless you mark it in a special way will be hard for you to find them later

rotkehlchen/tests/unit/accounting/test_default_settings.py
```diff
@@ -170,7 +190,7 @@ def test_accounting_receive_settings(
         ),
         cost_basis=None,
         index=0,
-        extra_data={'tx_hash': EXAMPLE_TX_HASH_HEX},
+        extra_data={'tx_hash': expected_tx_hash},
```
yabirgb, line 193: something is off, shouldn't the key be `event_identifier` here if it is `history_event`? Mentioning due to `_gain_one_ether`

rotkehlchen/accounting/transactions.py
```diff
@@ -25,15 +26,15 @@ def evm_events_iterator(
     Takes associated event as an argument to be able to log it in case of errors.
     """
     for event in events_iterator:
-        if not isinstance(event, EvmEvent):
+        if not isinstance(event, EvmEvent) and not isinstance(event, HistoryEvent):  # TODO: undo. Is a temporary hack  # noqa: E501
             log.error(
                 f'At accounting for tx_event {associated_event.notes} with hash '
                 f'{associated_event.tx_hash.hex()} we expected to take an additional '
```
yabirgb, line 32: Won't this line fail for history events?

rotkehlchen/accounting/transactions.py
```diff
@@ -25,15 +26,15 @@ def evm_events_iterator(
...
                 f'event but found a non history base entry event',
             )
             return
 
-        yield event
+        yield event  # type: ignore[misc]
```
yabirgb, line 37: Maybe you can rename this function? Is used in two places and even being this a hack is really confusing to follow if things are incorrect due to naming

yabirgb (commented):
A few things:

1. I would mark the TODOs with a special tag so you can find them faster
2. Even when this is a temp hack I would do things a bit better and rename some functions and variables, otherwise is hard to follow
3. For non evm events you are using the default rules right? Are they documented in the docs?
4. have you checked if the events get correctly exported in the csv?
5. What happens with the eth2 staking events (the third child class)? Are they properly processed?

nebolax:
> For non evm events you are using the default rules right? Are they documented in the docs?

Yes

> have you checked if the events get correctly exported in the csv?

I don't get. How is this related to the issue that this PR solves?

> What happens with the eth2 staking events (the third child class)? Are they properly processed?

Yes, they have their own `process` method.

yabirgb:
> I don't get. How is this related to the issue that this PR solves?

They were not correctly processed before right? Were they in the exported csv after generating the report?

nebolax:
> They were not correctly processed before right? Were they in the exported csv after generating the report?

They were not getting into the pot. No `ProcessedAccountingEvent` was generated for them.

rotkehlchen/accounting/transactions.py
```diff
@@ -25,15 +26,15 @@ def evm_events_iterator(
     Takes associated event as an argument to be able to log it in case of errors.
     """
     for event in events_iterator:
-        if not isinstance(event, EvmEvent):
+        if not isinstance(event, EvmEvent) and not isinstance(event, HistoryEvent):  # TODO: undo. Is a temporary hack  # noqa: E501
```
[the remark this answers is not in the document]
nebolax: > but since HistoryBaseEntry > EvmEvent isn't it enough to do (unless you explecitly want to restrict the type and avoid the eth staking events):

Yeah, I used HistoryEvent + EvmEvent to be more explicit

rotkehlchen/accounting/structures/base.py
```diff
@@ -427,7 +433,11 @@ def process(
             )
             return 1
 
-        return 1
+        # TODO: change to a proper solution. Right now all generic history events are processed as
+        # if they were evm events. This works since default accounting rules identify events only
+        # by event type + subtype and don't pay attention to counterparty, but we should properly
```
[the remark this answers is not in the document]
nebolax: There is
```
This works since default accounting rules identify events only
        # by event type + subtype and don't pay attention to counterparty
```

that I think should make this clear. If to rephrase it, how would you do it?

rotkehlchen/accounting/transactions.py
nebolax: Yes, but for the typing sake in the body of this `if`, I would keep this check.

rotkehlchen/accounting/structures/base.py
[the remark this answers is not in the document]
yabirgb:
```suggestion
        # if they were evm events. In order to apply the correct accounting rules for historyEvents we use
        # the default accounting rules that are chosen only using the type and the subtype of the event.
```

rotkehlchen/tests/unit/accounting/test_default_settings.py
nebolax: Yeah, you can see above that there is a condition
```python
if entry_type == 'history_event':
        expected_tx_hash = f'rotki_events_{EXAMPLE_TX_HASH_HEX}'
    else:  # entry_type == 'evm_event'
        expected_tx_hash = EXAMPLE_TX_HASH_HEX
```

rotkehlchen/accounting/transactions.py
```diff
@@ -143,7 +147,7 @@ def _process_tx_swap(
             given_price=prices[0],
             count_entire_amount_spend=False,
             extra_data={
-                'tx_hash': out_event.tx_hash.hex(),
+                'tx_hash': base_group_id,  # TODO: undo. Is a temporary hack
```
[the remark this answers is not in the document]
nebolax: It is

rotkehlchen/accounting/transactions.py
nebolax: :thinking: where and why? What will be wrong?

nebolax: Ah, you mean `{associated_event.tx_hash.hex()}`? Yeah, probably it will. Fixed it

rotkehlchen/accounting/transactions.py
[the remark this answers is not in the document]
yabirgb: I mean the key on the `extra_data` attribute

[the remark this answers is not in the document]
nebolax: What I did right now is I made the functionality the way it was at the moment of rotki 1.27.1 version. Not sure how event identifier will be useful for the frontend and for the user. Though I get that putting event identifier in tx_hash field is also strange, but this is the way it was before and since this PR is just a temporary fix, I would say that it will be better to stay with what we had before.

rotkehlchen/tests/unit/accounting/test_default_settings.py
yabirgb: Is related to the other comment, I meant that the key shouldn't be `tx_hash`

docs/changelog.rst
```diff
@@ -5,6 +5,7 @@ Changelog
 * :bug:`-` ENS avatar fetching for each name's avatar should now work for all name resolvers. Even older or custom ones.
 * :bug:`-` Users will now be able to decode compounding transactions for Convex gauges.
 * :bug:`-` Users will no longer see double counting of balances in curve gauges.
+* :bug:`-` Events that were imported via a csv file will now be properly accounted in PnL reports.
```
LefterisJP, line 8: nitpick

```suggestion
* :bug:`-` Events that were imported via a csv file will now be properly accounted for in PnL reports.
```

rotkehlchen/accounting/transactions.py
LefterisJP: We should not add runtime if checks to satisfy mypy type checks.

rotkehlchen/tests/unit/accounting/test_default_settings.py
```diff
@@ -96,10 +97,20 @@ def _gain_one_ether(
...
 ) -> None:
     """Helper function to gain 1 ETH, so that spending events have something to spend"""
-    eth_gain_event = EvmEvent(
-        tx_hash=EXAMPLE_EVM_HASH,
+    event_class: type[Union[HistoryEvent, EvmEvent]]
+    kwargs: dict[str, Any]
+    if entry_type == 'history_event':
+        event_class = HistoryEvent
+        kwargs = {'event_identifier': f'rotki_events_{EXAMPLE_EVM_HASH.hex()}'}  # pylint: disable=no-member  # noqa: E501
+    else:  # entry_type == 'evm_event'
```
LefterisJP, line 108:
```suggestion
    else:  # can only be evm_event
```

rotkehlchen/tests/unit/accounting/test_default_settings.py
```diff
@@ -143,18 +154,27 @@ def test_accounting_no_settings(accounting_pot: 'AccountingPot'):
...
     """Test that the default accounting settings for receiving are correct"""
     _gain_one_ether(
         transactions=accounting_pot.transactions,
         event_type=event_type,
         event_subtype=event_subtype,
+        entry_type=entry_type,
     )
+    if entry_type == 'history_event':
+        expected_tx_hash = f'rotki_events_{EXAMPLE_TX_HASH_HEX}'
+    else:  # entry_type == 'evm_event'
```
LefterisJP, line 175:
```suggestion
    else:  # can only be evm event
```

LefterisJP (commented):
Some small nitpick comments which I can also handle in my eth2 test fix PR. But also the question I asked in discord about tx_hash

LefterisJP (approved):
LGTM

PR #6173, opened 2023-05-30
Add polygon pos events, balances, api input, and notification handler

tewshi:
## Checklist


frontend/app/src/components/history/events/HistoryEventsList.vue
```diff
@@ -180,7 +182,7 @@ watch(
                   <v-lazy>
                     <history-event-type
                       :event="item"
-                      :chain="getChain(item.location)"
+                      :chain="getChain(toSnakeCase(item.location))"
```
kelsos, line 185: Do we really need it to do for `.location`? the values should not be affected by our conversion.

The only places would be where it is used as the key

frontend/app/src/utils/text.ts
```diff
@@ -40,6 +40,42 @@ export const toCapitalCase = (string: string): string =>
...
+ * @param {'capitalize' | 'sentence' | 'uppercase' | 'lowercase'} transform
+ * @return {string} - String converted to human-readable case
+ * @example
+ * toHumanReadable('POLYGON_POS', 'sentence'); // Polygon Pos
+ * @example
+ * toHumanReadable('POLYGON_POS'); // POLYGON POS
+ * @example
+ * toHumanReadable('polygon_pos', 'uppercase'); // POLYGON POS
+ */
+export const toHumanReadable = (
```
kelsos, line 55: Do we have a test for this utility? if not can you add a test also?

frontend/app/src/pages/settings/api-keys/external/index.vue
```diff
@@ -58,10 +65,12 @@ const updateKeys = ({
   beaconchain,
   loopring,
   opensea,
-  optimismEtherscan
+  optimismEtherscan,
+  polygonPosEtherscan
 }: ExternalServiceKeys) => {
   evmEtherscanTabs.ethereum.value = etherscan?.apiKey || '';
   evmEtherscanTabs.optimism.value = optimismEtherscan?.apiKey || '';
+  evmEtherscanTabs.polygon_pos.value = polygonPosEtherscan?.apiKey || '';
```
kelsos, line 73: Do you think it would be easy to make sure we don't have snake_case in the frontend code for consistency?

kelsos (commented):
@tewshi I was trying odd only a polygon address and it seems there if failed because of the difference between snake/camelcase.

To replicate just uncheck the add all chains and select polygon from the dropdown.

frontend/app/src/utils/text.ts
tewshi: I was trying not to get the PR too long, but I can add a test for this.

frontend/app/src/components/history/events/HistoryEventsList.vue
tewshi: yes, because it doesn't pick the correct url for the chain on the hashlink component (takes the default which is etherscan)

and in this object, location comes in as **polygon pos**

image

tewshi: I did ask on Discord, but Lef said this value is from the enum, so I have to handle on FE

frontend/app/src/pages/settings/api-keys/external/index.vue
tewshi: checking, perhaps i'd just have to use the Blockchain enum here.

frontend/app/src/components/history/events/HistoryEventsList.vue
tewshi: think to avoid future issues, we should call it in `getChain`
since, for single name chains (eth, avax, ...) `toSnakeCase` does not affect and combined names will always be snake case...

frontend/app/src/pages/settings/api-keys/external/index.vue
```diff
@@ -151,9 +164,11 @@ const confirm = async (service: ExternalServiceName) => {
   set(loading, false);
 };
 
+const getName = (chain: string) => get(getChainName(chain as Blockchain));
```
kelsos, line 167: Btw we don't have to do anything for this PR but I think we should have a different approach with this.

@rotki/frontend 

We can use plain functions and [reactify](https://vueuse.org/shared/reactify/) when needed.

This way we can also avoid that thing here.

Also useful is this [createUnrefFn](https://vueuse.org/core/createUnrefFn/) which makes it so that you can pass refs to a normal non-reactive function.

So the approach should be to use plain functions and use either of the utilities above when needed.

frontend/app/tests/unit/specs/utils/text.spec.ts
```diff
@@ -0,0 +1,16 @@
+describe('utils/text', () => {
```
kelsos, line 1: Great, thanks :)

kelsos (approved):
Thanks, @tewshi, it looks good. I did only a couple of small changes related to the native token, to fix them so that the tables show the proper asset in the amount column.

PR #6176, opened 2023-05-31
feat: update date time picker

lukicenturi:
Closes #(issue_number)

## Checklist


kelsos:
I did a quick check @lukicenturi and it is in the right direction.

I noticed a couple of things though.

First, we must also replace `cleave.js` in optimizeDeps in `vite.config.ts` with `imask` to avoid the optimizing deps situation that causes tests to fail.

For the external trades form, we might need to adjust the size because it is too small now.
image

For the date input is there any chance we can show the mask (from the docs this should be possible, but I am not 100% sure about handling the seconds being an optional situation)?

image

Finally can you also add some tests for the date input to make sure that it works as expected?

kelsos (approved):
lgtm @lukicenturi we need a small follow-up but it can go in a new PR since it was already in develop.

the date component with seconds required shows a `[[HH:MM]:ss]` error that indicates the you can either `HH:MM` or `HH:MM:ss` or skip it completely. 

If we are enforcing seconds then it can be only `[HH:MM:ss]` the first pattern is only if we can skip seconds.

Can you make a followup PR? and if possible add some component tests there.

PR #6180, opened 2023-05-31
Don't skip HistoryEvents in accounting

nebolax:
Proper solution for history events accounting.

This PR doesn't add a changelog since it has already been added in the hotfix pr that was against bugfixes.


rotkehlchen/chain/evm/accounting/structures.py
```diff
@@ -27,11 +45,41 @@ class TxAccountingTreatment(Enum):
...
+        self.count_entire_amount_spend = count_entire_amount_spend
+        self.count_cost_basis_pnl = count_cost_basis_pnl
+        self.method = method
+        self.accounting_treatment = accounting_treatment
+        self.accountant_cb = accountant_cb
+
+
+class TxEventSettings(HistoryBaseEntrySettings):
+    """Settings for an evm event generated by a decoder"""
+    def __init__(
```
yabirgb, line 69: Why is this one child of the other? and why do you need two classes? aren't they doing the same?

rotkehlchen/accounting/history_base_entries.py
```diff
@@ -0,0 +1,257 @@
...
+        timestamp = event.get_timestamp_in_sec()
+        type_identifier = event.get_type_identifier()
+        event_settings = self.tx_event_settings.get(type_identifier, None)
+        if event_settings is None:
+            event_settings = self.tx_event_settings.get(
+                event.get_type_identifier(include_counterparty=False),
+            )
+            if (  # For swaps we have a default treatment but for the rest we bail
+                event_settings is None or
+                event_settings.accounting_treatment != TxAccountingTreatment.SWAP
```
yabirgb, line 162: Looking again I believe the only way to get here is by a mapping that doesn't exist but I'm still not sure why the check `event_settings.accounting_treatment != TxAccountingTreatment.SWAP`

rotkehlchen/accounting/history_base_entries.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def _process_swap(
+            self,
+            timestamp: Timestamp,
+            out_event: HistoryBaseEntry,
+            in_event: HistoryBaseEntry,
+            event_settings: HistoryBaseEntrySettings,
+            general_extra_data: dict[str, Any],
+    ) -> int:
+        prices = self.pot.get_prices_for_swap(
```
yabirgb, line 220: can you add a docstring?

rotkehlchen/accounting/history_base_entries.py
```diff
@@ -0,0 +1,257 @@
...
+                f'At accounting for event {associated_event.notes} with identifier '
+                f'{associated_event.event_identifier} we expected to take an additional '
+                f'event but found a non history base entry event',
+            )
+            return
+
+        yield event
+
+
+class HistoryBaseEntriesAccountant():
```
yabirgb, line 129: nitpick: in python3 is not needed

```suggestion
class HistoryBaseEntriesAccountant:
```

rotkehlchen/accounting/structures/base.py
```diff
@@ -427,7 +431,7 @@ def process(
             )
             return 1
 
-        return 1
+        return accounting.history_base_entries.process(event=self, events_iterator=events_iterator)
```
yabirgb, line 434: this is the key change in the PR right?

rotkehlchen/accounting/history_base_entries.py
```diff
@@ -0,0 +1,257 @@
...
+    def process(
+            self,
+            event: HistoryBaseEntry,
+            events_iterator: Iterator[AccountingEventMixin],
+    ) -> int:
+        """Process a history base entry and return number of actions consumed from the iterator"""
+        timestamp = event.get_timestamp_in_sec()
+        type_identifier = event.get_type_identifier()
+        event_settings = self.tx_event_settings.get(type_identifier, None)
+        if event_settings is None:
```
yabirgb, line 156: shouldn't this check happen only if is an evm event?

yabirgb (commented):
A few questions about how the code is designed. Also don't forget to update develop and rebase this PR since you will have conflicts

rotkehlchen/accounting/structures/base.py
nebolax: Yep

rotkehlchen/accounting/history_base_entries.py
nebolax: Yes, thanks

rotkehlchen/accounting/history_base_entries.py
```diff
@@ -0,0 +1,257 @@
...
+        if event_settings.accountant_cb is not None:
+            event_settings.accountant_cb(
+                pot=self.pot,
+                event=event,
+                other_events=history_base_entries_iterator(events_iterator, event),
+            )
+
+        general_extra_data = {}
+        if isinstance(event, EvmEvent):
+            general_extra_data['tx_hash'] = event.tx_hash.hex()
```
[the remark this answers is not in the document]
nebolax: I removed it in bugfixes, but at the time when this PR was made, the bugfixes fix was not merged in develop. I will rebase and adjust it

rotkehlchen/chain/evm/accounting/structures.py
nebolax: Yes, they are doing the same, I made 2 classes for typing purposes. Accounting settings that are defined in evm modules expect the event and the iterator to be for evm events, so I made a special class for this.

rotkehlchen/accounting/history_base_entries.py
nebolax: Hmm I also don't get it tbh. iirc it was added as part of some hotfix that was specific to swaps, but I think that it will make sense to do it for all evm events.

rotkehlchen/accounting/history_base_entries.py
```diff
@@ -0,0 +1,261 @@
...
+        self.evm_accounting_aggregators = evm_accounting_aggregators
+        self.pot = pot
+        self.tx_event_settings: dict[str, HistoryBaseEntrySettings] = {}
+
+    def reset(self) -> None:
+        self.evm_accounting_aggregators.reset()
+        self.tx_event_settings = (  # Using | operator is fine since keys are unique
+            self.evm_accounting_aggregators.get_accounting_settings(self.pot) |  # type: ignore[assignment]  # noqa: E501
+            make_default_accounting_settings(self.pot)
+        )
```
yabirgb, line 145: As it is now it will work but having this two together opens the door to an error:

1. Developer makes a mistake and the key doesn't have the counterparty for evm events in the settings rules
2. The rule gets imported and it has a callback
3. The callback gets called in a history event

Would it make sense to to split this in a mapping having by key the type or two mappings and have a function that takes the event and checks depending on the type in the correct set of rules? Basically to avoid calling the callback of an evm event in other type of event

rotkehlchen/chain/evm/accounting/structures.py
```diff
@@ -27,11 +32,28 @@ class TxAccountingTreatment(Enum):
...
+        self.accounting_treatment = accounting_treatment
+        self.accountant_cb = accountant_cb
+
+
+class HistoryBaseEntrySettings(BaseEventSettings[HistoryBaseEntry]):
+    ...
+
+
+class TxEventSettings(BaseEventSettings[EvmEvent]):
+    ...
```
yabirgb, line 59: After messing with the types I believe to have found the reason for having two classes here since there are rules that can be applied to any event and others that can be applied only to evm events. 

Still the problem appears in the 

```
self.evm_accounting_aggregators.get_accounting_settings(self.pot) |  # type: ignore[assignment]  # noqa: E501
```
that basically breaks the type checking. What do you think of my other suggestion?

yabirgb (commented):
Two comments for a specific part of the code

rotkehlchen/accounting/history_base_entries.py
nebolax: Hmm yeah, it could make sense indeed

rotkehlchen/accounting/history_base_entries.py
```diff
@@ -0,0 +1,269 @@
...
+    )
+    renew_key = str(HistoryEventType.RENEW) + '__' + str(HistoryEventSubType.NONE)
+    result[renew_key] = BaseEventSettins(
+        taxable=True,
+        count_entire_amount_spend=True,
+        count_cost_basis_pnl=True,
+        method='spend',
+    )
+    swap_key = str(HistoryEventType.TRADE) + '__' + str(HistoryEventSubType.SPEND)
+    result[swap_key] = BaseEventSettins(
```
yabirgb, line 84:
```suggestion
    result[swap_key] = BaseEventSettings(
```

rotkehlchen/accounting/history_base_entries.py
```diff
@@ -0,0 +1,269 @@
...
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+def make_default_accounting_settings(pot: 'AccountingPot') -> dict[str, BaseEventSettins]:
+    """
+    Returns accounting settings for events that can come from various decoders and thus don't have
+    any particular protocol. These settings also allow users to customize events in the UI.
+    Users are supposed to apply these settings in the history view.
```
LefterisJP, line 31: apply these settings in the history view? What is this supposed to mean? This is moved from the old docstring so perhaps it's speaking in the future tense. I think it refers to customization of how they should be handled in accounting, somethign which is not yet possible. The issue is here: https://github.com/rotki/rotki/issues/4341

So perhaps write:

```suggestion
    Users are supposed to apply these settings in the history view once they become customizable: https://github.com/rotki/rotki/issues/4341
```

rotkehlchen/accounting/history_base_entries.py
```diff
@@ -0,0 +1,269 @@
...
+class HistoryBaseEntriesAccountant:
+
+    def __init__(
+            self,
+            evm_accounting_aggregators: 'EVMAccountingAggregators',
+            pot: 'AccountingPot',
+    ) -> None:
+        self.evm_accounting_aggregators = evm_accounting_aggregators
+        self.pot = pot
+        self.tx_event_settings: dict[str, BaseEventSettins] = {}
```
LefterisJP, line 142: This is not just for transactions anymore

```suggestion
        self.event_settings: dict[str, BaseEventSettins] = {}
```

rotkehlchen/accounting/pot.py
```diff
@@ -55,7 +55,7 @@ def __init__(
         )
         self.pnls = PnlTotals()
         self.processed_events: list[ProcessedAccountingEvent] = []
-        self.transactions = TransactionsAccountant(
+        self.history_base_entries = HistoryBaseEntriesAccountant(
```
LefterisJP, line 58: Name is confusing.

```suggestion
        self.events_accountant = HistoryBaseEntriesAccountant(
```

rotkehlchen/chain/evm/accounting/structures.py
```diff
@@ -27,11 +28,39 @@ class TxAccountingTreatment(Enum):
...
 
-class TxEventSettings(NamedTuple):
-    """Settings for an event generated by a decoder"""
-    taxable: bool
-    count_entire_amount_spend: bool
-    count_cost_basis_pnl: bool
-    method: Literal['acquisition', 'spend']
-    accounting_treatment: Optional[TxAccountingTreatment] = None
-    accountant_cb: Optional[AccountantCallback] = None
+class BaseEventSettins:
```
LefterisJP, line 31: As discussed this needs to be named properly.
```suggestion
class BaseEventSettings:
```

rotkehlchen/chain/evm/accounting/structures.py
```diff
@@ -27,11 +28,39 @@ class TxAccountingTreatment(Enum):
...
+    ):
+        self.taxable = taxable
+        self.count_entire_amount_spend = count_entire_amount_spend
+        self.count_cost_basis_pnl = count_cost_basis_pnl
+        self.method = method
+        self.accounting_treatment = accounting_treatment
+
+
+class TxEventSettings(BaseEventSettins):
+    """Settings for history base entry accounting"""
```
LefterisJP, line 49: copy pasta

```suggestion
    """Accounting settings for EVM transaction events"""
```

rotkehlchen/tests/unit/accounting/test_default_settings.py
```diff
@@ -94,7 +94,7 @@ def fixture_accounting_pot(accountant, airdrops_taxable, gas_taxable, include_cr
 
 
 def _gain_one_ether(
-        transactions: 'TransactionsAccountant',
+        history_base_entries: 'HistoryBaseEntriesAccountant',
```
LefterisJP, line 97: Adjust names as explained before. here and everywhere else

```suggestion
        events_accountant: 'HistoryBaseEntriesAccountant',
```

LefterisJP (approved):
perfect. LGTM

PR #6181, opened 2023-05-31
Dont connect to evm nodes if no account is tracked for the chain

LefterisJP:


rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -215,12 +214,27 @@ def __init__(
...
+        tracked accounts in the DB. Otherwise when we have none.
+        """
+        if self.connected_to_any_web3() or self.greenlet_manager.has_task(f'Attempt connection to {self.chain_name} node'):  # noqa: E501
+            return
+
+        with self.database.conn.read_ctx() as cursor:
+            accounts = self.database.get_blockchain_accounts(cursor)
+
+        tracked_accounts_num = len(accounts.get(self.blockchain))
+        if (tracked_accounts_num != 0 and when_tracked_accounts) or (not when_tracked_accounts and tracked_accounts_num == 0):  # noqa: E501
```
yabirgb, line 235: A bit of a nitpick but wouldn't it be better to use here

```suggestion
        if (tracked_accounts_num != 0 and when_tracked_accounts) or (when_tracked_accounts is False and tracked_accounts_num == 0):  # noqa: E501
```

for me it reads better used to read the rest of the code

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -215,12 +214,27 @@ def __init__(
...
+        If `when_tracked_accounts` is True then it will connect when we have some
+        tracked accounts in the DB. Otherwise when we have none.
+        """
+        if self.connected_to_any_web3() or self.greenlet_manager.has_task(f'Attempt connection to {self.chain_name} node'):  # noqa: E501
+            return
+
+        with self.database.conn.read_ctx() as cursor:
+            accounts = self.database.get_blockchain_accounts(cursor)
+
+        tracked_accounts_num = len(accounts.get(self.blockchain))
```
yabirgb, line 234: if I don't have any address in polygon then `get_blockchain_accounts` doesn't return polygon if I udnerstood correctly and then the get here will return `None`? or I missed something

yabirgb: Also that function uses `get_key` instead of returning the `SupportedBlockchain` enum value

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -215,12 +214,27 @@ def __init__(
...
+        if (tracked_accounts_num != 0 and when_tracked_accounts) or (not when_tracked_accounts and tracked_accounts_num == 0):  # noqa: E501
+            rpc_nodes = self.database.get_rpc_nodes(blockchain=self.blockchain, only_active=True)
+            self.connect_to_multiple_nodes(rpc_nodes)
```
yabirgb, line 237: Could you add a logging statement here? I have found it useful to know what nodes are being tried to connect in the past when reading the users' logs

yabirgb (commented):
A few comments/questions

rotkehlchen/chain/evm/node_inquirer.py
LefterisJP: There is logging inside the function. For each node it connects to. So anything more is redundant

rotkehlchen/chain/evm/node_inquirer.py
LefterisJP: as mentioned before if it's a boolean which is made to sound like a boolean it's okay. I specifically made a name that would read well here

rotkehlchen/chain/evm/node_inquirer.py
yabirgb: Ahg the get key nvm. Inside is overloaded

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -465,7 +479,7 @@ def connect_to_multiple_nodes(self, nodes: Sequence[WeightedNode]) -> None:
             if weighted_node.node_info.name == self.etherscan_node_name:
                 continue
 
-            task_name = f'Attempt connection to {weighted_node.node_info.name!s} {self.chain_name} node'  # noqa: E501
+            task_name = f'Attempt connection to {self.chain_name} node {weighted_node.node_info.name!s}'  # noqa: E501
```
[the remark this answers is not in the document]
LefterisJP: yes I was torn here to abstract into a function so it's the same in both. This is what I would do instead of commenting if anything

LefterisJP:
I will see if I can address any nitpick in other PR. But since no problematic comment found I will merge, okay @yabirgb ?

yabirgb (approved):
LGTM

PR #6182, opened 2023-06-01
refactor(frontend): adjust form validation for several forms

lukicenturi:
Closes #(issue_number)

## Checklist

Adjust forms for 
- [x] HistoryEventsForm
- [x] AddTxHashForm
- [x] LedgerActionForm
- [x] UserNotesForm


kelsos:
@lukicenturi this is ready for review, right? can you rebase on top of `develop` first?

frontend/app/src/components/notes/UserNotesList.vue
```diff
@@ -324,17 +331,17 @@ onMounted(async () => {
...
       "
-      :action-disabled="!valid"
-      @confirm="save()"
+      :action-disabled="submitting || !valid"
+      @confirm="trySubmit()"
       @cancel="resetForm()"
     >
-      <user-note-form v-model="form" @valid="valid = $event" />
+      <user-note-form v-model="form" />
     </big-dialog>
```
kelsos, line 345: Do you think you can also extract this to a `UserFormDialog` component?

frontend/app/src/components/history/events/HistoryEventFormDialog.vue
```diff
@@ -58,19 +29,17 @@ const title: ComputedRef<string> = computed(() =>
...
   <big-dialog
-    :display="value"
+    :display="openDialog"
     :title="title"
     :primary-action="t('common.actions.save')"
-    :action-disabled="loading || !valid"
-    :loading="loading"
-    @confirm="confirmSave()"
-    @cancel="clearDialog()"
+    :action-disabled="loading || submitting || !valid"
```
kelsos, line 35: Don't we also want to remove the validity check from there and do it on submit?

frontend/app/src/components/history/events/tx/TransactionFormDialog.vue
```diff
@@ -0,0 +1,29 @@
...
+
+const { t } = useI18n();
+</script>
+
+<template>
+  <big-dialog
+    :display="openDialog"
+    :title="t('transactions.dialog.add_tx')"
+    :primary-action="t('common.actions.save')"
+    :action-disabled="loading || submitting || !valid"
```
kelsos, line 22: Also here if we are validating when the button is pressed, we should remove the validity check.

Action should only be disabled when loading/submitting only

frontend/app/src/components/history/ledger-actions/LedgerActionFormDialog.vue
```diff
@@ -1,81 +1,49 @@
...
-    :display="value"
+    :display="openDialog"
     :title="title"
     :subtitle="subtitle"
     :primary-action="t('common.actions.save')"
-    :action-disabled="loading || !valid"
-    :loading="loading"
-    @confirm="confirmSave()"
-    @cancel="clearDialog()"
+    :action-disabled="loading || submitting || !valid"
```
kelsos, line 42: As before if we validate on submit and submit bails out if the rules are invalid then we don't need the `!valid`

lukicenturi:
@kelsos so for the valid, it will always be `true` from the start. 
If then user click the submit button, when there is fields with failed validation, it will be `false`, in that case I think it's makes sense to disable the button, until user repair the fields

kelsos (approved):
lgtm thanks @lukicenturi

PR #6183, opened 2023-06-01
For kelsos

LefterisJP:


rotkehlchen/types.py
```diff
@@ -418,6 +418,15 @@ def is_bitcoin(self) -> bool:
     def is_substrate(self) -> bool:
         return self in get_args(SUPPORTED_SUBSTRATE_CHAINS)
 
+    def get_native_token_id(self) -> str:
+        """Returns the string identifier of the native token for the chain"""
+        if self == SupportedBlockchain.OPTIMISM:
+            return 'ETH'
+        if self == SupportedBlockchain.POLYGON_POS:
+            return 'eip155:137/erc20:0x0000000000000000000000000000000000001010'
+
```
yabirgb, line 427: Should it use here

```suggestion
        if self == SupportedBlockchain.POLYGON_POS:
            return A_POLYGON_POS_MATIC.identifier

```

LefterisJP: no this is not possible. This is a very ugly and impossible to fix cyclic import. KEep in mind the file is types.py which is one of the most imported files in the project. 

We should take out huge classes like this from there.

LefterisJP:
Okay so again here we have timeouts for no apparent reason. I document them in this comment 

### API tests


https://gist.github.com/LefterisJP/c79f539de6c679466bdc6d5e3ecb8212

### Other backend tests


https://gist.github.com/LefterisJP/c03bc32b1c8b4acac508164ee5bfb52b

### Notes

With a quick look I really can't say what's wrong, just like before. Any ideas are welcome. Perhaps we need to reduce the timeout per job in any case.

LefterisJP:
All tests are green again finally. I think that if we get full test job failure  in more places we should just apply what I did in [this](https://github.com/rotki/rotki/pull/6183/commits/3e803acb6dab8285dc6c3ae68bd7a84da23fa9ba) commit to places where the timeout happens due to running out of connections and having no timeout in the request.

We can find it by either running tests locally or putting the `-s` argument to pytest in the CI to print every log to stdout (this can slow things down and even kill some tests)

LefterisJP:
Just for history's sake adding the explanation I put in discord also here:

But taking a step back. What's happening is this:
We use requests directly in all our tests. To call our own api endpoints. Requests itself uses a connection pool by urllib3. For each url, requests creates a pool of default max size 10 connections.

All of our requests in tests do not use timeouts. As you may recall there is no global way to add a timeout to requests, apart from patching or subclassing. Connections are cleared/closed from the pool periodically(but not without some external action it seems). We also never close the connections manually ourselves.

So what this results to is that we may run out of connections for a url from the default requests connection pool and since there is no timeout the test does not fail, but it hangs waiting for that connection. Since nothing cleans up the old connection. So it waits forever (until the CI job is killed).

The solution here can be seen in that [commit](https://github.com/rotki/rotki/pull/6183/commits/3e803acb6dab8285dc6c3ae68bd7a84da23fa9ba) up there.
1. Add a new custom session that is loaded in the tests as a fixture on demand.That fixture has a built-in timeout and also 10x the pool size (so 100).
2. Use it in the tests around where this happened instead of pure requests (using it everywhere would not only be a lot of work to change all tests but also slow us down and increase memory consumption)
3. use a with context manager for the request so that connection is also closed right after using it (also doing this all the time presumably may slow things down)

What I can't answer is why did we hit this now. I can only determine this started when we upgraded:
```
requests==2.31.0
urllib3==1.26.16
```

2.31.0 was a security release and changelog says nothing else.

PR #6184, opened 2023-06-01
Asset chart to zero

### LINKED ISSUE #2822: Continue "asset amount and value" graph after the amount goes to zero
There are a few assets in my Rotki that I used to have, but I now don't have. When looking "asset amount and value" graph it ends at the last non-zero amount, and it states under the graph that the data is available until the date I sold the asset. It would make sense to me if the graph continued on until the current date, but the amount / value goes to zero.

dimyG:
Closes #2822

## Checklist

### DIFF
--- a/rotkehlchen/db/dbhandler.py
+++ b/rotkehlchen/db/dbhandler.py
@@ -2233,6 +2233,101 @@ class DBHandler:
                 data.append(total)
         return times_int, data
 
+    @staticmethod
+    def _infer_zero_timed_balances(
+            cursor: 'DBCursor',
+            balances: list[SingleDBAssetBalance],
+            from_ts: Optional[Timestamp] = None,
+            to_ts: Optional[Timestamp] = None,
+    ) -> list[SingleDBAssetBalance]:
+        """
+        Given a list of asset specific timed balances, infers the missing zero timed balances
+        for the asset. We add 0 balances on the start and end of a period of 0 balances.
+        It addresses this issue: https://github.com/rotki/rotki/issues/2822
+
+        Example
+        We have the following timed balances for ETH (value, time):
+        (1, 1), (1, 2), (2, 3), (5, 7), (5, 12)
+        The timestamps of all timed balances in the DB are:
+        (1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12)
+        So we need to infer the following zero timed balances:
+        (0, 4), (0, 6), (0, 8), (0, 11)
+
+        Keep in mind that in a case like this (1, 1), (1, 2), (5, 4) we will infer (0, 3)
+        despite the fact that it is not strictly needed by the front end.
+        """
+        if len(balances) == 0:
+            return []
+
+        if from_ts is None:
+            from_ts = Timestamp(0)
+        if to_ts is None:
+            to_ts = ts_now()
+
+        cursor.execute(
+            'SELECT COUNT(*) FROM timed_balances WHERE timestamp BETWEEN ? AND ?',
+            (from_ts, to_ts),
+        )
+        num_all_balances = cursor.fetchone()[0]
+
+        if len(balances) == num_all_balances:
+            return []
+
+        asset_timestamps = [b.time for b in balances]
+        cursor.execute(
+            'SELECT timestamp, category FROM timed_balances WHERE timestamp BETWEEN ? AND ? '
+            'ORDER BY timestamp ASC',
+            (from_ts, to_ts),
+        )
+        all_timestamps, all_categories = zip(*cursor)
+        # dicts maintain insertion order in python 3.7+
+        timestamps_have_asset_balance = dict.fromkeys(all_timestamps, False)
+        timestamps_with_asset_balance = dict.fromkeys(asset_timestamps, True)
+        timestamps_have_asset_balance.update(timestamps_with_asset_balance)
+        prev_has_asset_balance = timestamps_have_asset_balance[all_timestamps[0]]  # the value of the first timestamp  # noqa: E501
+        prev_timestamp = all_timestamps[0]
+        inferred_balances: list[SingleDBAssetBalance] = []
+        is_zero_period_open = False
+        last_asset_category = all_categories[-1]  # just a placeholder value, no need to calculate the actual value here  # noqa: E501
+        for idx, (timestamp, has_asset_balance) in enumerate(timestamps_have_asset_balance.items()):  # noqa: E501
+            if idx == len(timestamps_have_asset_balance) - 1 and has_asset_balance is False:
+                # If there is no balance for the last timestamp add a zero balance.
+                inferred_balances.append(
+                    SingleDBAssetBalance(
+                        time=timestamp,
+                        amount=ZERO,
+                        usd_value=ZERO,
+                        category=last_asset_category,
+                    ),
+                )
+            elif has_asset_balance is False and prev_has_asset_balance is True:
+                # add the start of a zero balance period
+                inferred_balances.append(
+                    SingleDBAssetBalance(
+                        time=timestamp,
+                        amount=ZERO,
+                        usd_value=ZERO,
+                        category=BalanceType.deserialize_from_db(all_categories[idx - 1]),  # noqa: E501  # the category of the previous timed_balance of the asset
+                    ),
+                )
+                is_zero_period_open = True
+            elif has_asset_balance is True and prev_has_asset_balance is False and is_zero_period_open is True:  # noqa: E501
+                # add the end of a zero balance period
+                inferred_balances.append(
+                    SingleDBAssetBalance(
+                        time=prev_timestamp,
+                        amount=ZERO,
+                        usd_value=ZERO,
+                        category=inferred_balances[-1].category,  # noqa: E501  # the category of the asset at the start of the zero balance period
+                    ),
+                )
+                is_zero_period_open = False
+                last_asset_category = inferred_balances[-1].category
+            elif has_asset_balance is True:
+                last_asset_category = BalanceType.deserialize_from_db(all_categories[idx])  # noqa: E501
+            prev_has_asset_balance, prev_timestamp = has_asset_balance, timestamp
+        return inferred_balances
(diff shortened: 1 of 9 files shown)


rotkehlchen/db/dbhandler.py
```diff
@@ -2233,6 +2233,66 @@ def get_netvalue_data(
...
+            'ORDER BY timestamp ASC',
+            (from_ts, to_ts),
+        ).fetchall()
+        all_timestamps = [e[0] for e in entries]
+        all_categories = [BalanceType.deserialize_from_db(e[1]) for e in entries]
+        # dicts maintain insertion order in python 3.7+
+        timestamps_dict = dict.fromkeys(all_timestamps, False)
+        asset_dict = dict.fromkeys(asset_timestamps, True)
+        timestamps_dict.update(asset_dict)
+        prev_from_asset = timestamps_dict[all_timestamps[0]]  # the value of the first timestamp
```
yabirgb, line 2281: I believe it was a bit harder to follow due to the names used for the variables? not sure if for you is the same. What do you think of

```suggestion
        timestamps_had_balance = dict.fromkeys(all_timestamps, False)
        timestamps_with_asset_balance = dict.fromkeys(asset_timestamps, True)
        timestamps_had_balance.update(asset_dict)
        prev_from_asset = timestamps_had_balance[all_timestamps[0]]  # the value of the first timestamp
```

rotkehlchen/db/dbhandler.py
```diff
@@ -2233,6 +2233,66 @@ def get_netvalue_data(
...
+        inferred_balances = []
+        for idx, (timestamp, from_asset) in enumerate(timestamps_dict.items()):
```
yabirgb, line 2283: `from_asset` is a boolean here right? Maybe rename it to `had_balance`?

rotkehlchen/db/dbhandler.py
```diff
@@ -2233,6 +2233,66 @@ def get_netvalue_data(
...
+            if not from_asset and prev_from_asset:
```
yabirgb, line 2284:
```suggestion
            if from_asset  is False and prev_from_asset is True:
```

rotkehlchen/db/dbhandler.py
```diff
@@ -2297,6 +2357,9 @@ def query_timed_balances(
                     ),
                 )
 
+        inferred_balances = self._infer_zero_timed_balances(cursor, balances, from_ts, to_ts)
+        balances.extend(inferred_balances)
```
yabirgb, line 2361: question: this doesn't guarantee that they are sorted by timestamp rigth? is that a problem?

rotkehlchen/db/dbhandler.py
```diff
@@ -2233,6 +2233,66 @@ def get_netvalue_data(
...
+        (1, 2, 3, 4, 5, 6, 7, 8, 9)
+        So we need to infer the following zero timed balances:
+        (0, 4), (0, 7)
+        """
+        if len(balances) == 0:
+            return []
+
+        # else
+        cursor.execute(
+            'SELECT COUNT(*) FROM timed_balances WHERE timestamp BETWEEN ? AND ?',
```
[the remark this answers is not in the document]
dimyG: Oops, you are right.

rotkehlchen/db/dbhandler.py
```diff
@@ -2233,6 +2234,95 @@ def get_netvalue_data(
...
+        for idx, (timestamp, has_asset_balance) in enumerate(timestamps_have_asset_balance.items()):  # noqa: E501
+            if idx == len(timestamps_have_asset_balance) - 1 and has_asset_balance is False:
+                # If there is no balance for the last timestamp add a zero balance.
+                inferred_balances.append(
+                    SingleDBAssetBalance(
+                        time=timestamp,
+                        amount=ZERO,
+                        usd_value=ZERO,
+                        # the category of the asset at the start of the zero balance period
+                        category=inferred_balances[-1].category,
```
yabirgb, line 2298: Can't this raise error if you have two snapshots and in the second one you are missing the asset? Wouldn't `inferred_balances` be empty?

rotkehlchen/db/dbhandler.py
```diff
@@ -2297,6 +2387,9 @@ def query_timed_balances(
                     ),
                 )
 
+        inferred_balances = self._infer_zero_timed_balances(cursor, balances, from_ts, to_ts)
+        balances = list(heapq.merge(balances, inferred_balances, key=lambda x: x.time))
```
yabirgb, line 2391: why use a heapq. Isn't the time complexity `O(NlogN)` in this case with n the number of elements? same as doing `(balances + inferred_balances) .sort()`? also it adds an additional iteration to transform into the list. Just asking if is better or not than a simple sort

rotkehlchen/tests/api/test_statistics.py
```diff
@@ -209,6 +209,67 @@ def test_query_statistics_asset_balance_errors(rotkehlchen_api_server, rest_api_
     )
 
 
+@pytest.mark.parametrize('start_with_valid_premium', [True])
+def test_statistics_asset_balance_inferred_zeros(rotkehlchen_api_server):
+    rotki = rotkehlchen_api_server.rest_api.rotkehlchen
+    formatstr = '%Y-%m-%dT%H:%M:%SZ'
```
yabirgb, line 215: can you add a small docstring explaining what is being tested here? it helps later when navigating the tests

yabirgb (commented):
Just two smaller questions

rotkehlchen/db/dbhandler.py
```diff
@@ -2233,6 +2233,105 @@ def get_netvalue_data(
...
+                )
+            elif has_asset_balance is False and prev_has_asset_balance is True:
+                # add the start of a zero balance period
+                inferred_balances.append(
+                    SingleDBAssetBalance(
+                        time=timestamp,
+                        amount=ZERO,
+                        usd_value=ZERO,
+                        # the category of the previous timed_balance of the asset
+                        category=BalanceType.deserialize_from_db(all_categories[idx - 1]),
```
yabirgb, line 2318: In this cases I believe lefteris prefers

```suggestion
                        category=BalanceType.deserialize_from_db(all_categories[idx - 1]),  # noqa: E501  # the category of the previous timed_balance of the asset
```

yabirgb (commented):
Just two minor things but I believe it makes sense to change if left has comments or in a different PR

rotkehlchen/db/dbhandler.py
```diff
@@ -2233,6 +2233,99 @@ def get_netvalue_data(
...
+        # dicts maintain insertion order in python 3.7+
+        timestamps_have_asset_balance = dict.fromkeys(all_timestamps, False)
+        timestamps_with_asset_balance = dict.fromkeys(asset_timestamps, True)
+        timestamps_have_asset_balance.update(timestamps_with_asset_balance)
+        prev_has_asset_balance = timestamps_have_asset_balance[all_timestamps[0]]  # the value of the first timestamp  # noqa: E501
+        prev_timestamp = all_timestamps[0]
+        inferred_balances: list[SingleDBAssetBalance] = []
+        is_zero_period_open = False
+        last_asset_category = all_categories[-1]  # just a placeholder value, no need to calculate the actual value here  # noqa: E501
+        for idx, (timestamp, has_asset_balance) in enumerate(timestamps_have_asset_balance.items()):  # noqa: E501
```
LefterisJP, line 2290: This iterates over all of the balances returned from the previous sql query.

This can be a long iteration if the graph has been going on for 5 years, like some of our first users.

Why not limit this from the sqlite query itself?

And you seem to be getting the specific asset balances as an argument to this function called `balances`. I think it should be possible to do this entire logic in sqlite itself. Otherwise you are loading potentially a ton of data in memory here and iterating over all of them at each and every api endoint call. So each time the graph moves etc.

rotkehlchen/db/dbhandler.py
```diff
@@ -2233,6 +2233,100 @@ def get_netvalue_data(
...
+
+        Example
+        We have the following timed balances for ETH (value, time):
+        (1, 1), (1, 2), (2, 3), (5, 7), (5, 12)
+        The timestamps of all timed balances in the DB are:
+        (1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12)
+        So we need to infer the following zero timed balances:
+        (0, 4), (0, 6), (0, 8), (0, 11)
+
+        Have in mind that in a case like this (1, 1), (1, 2), (5, 4) we will infer (0, 3)
```
LefterisJP, line 2255: nitpick

```suggestion
        Keep in mind that in a case like this (1, 1), (1, 2), (5, 4) we will infer (0, 3)
```

rotkehlchen/db/dbhandler.py
```diff
@@ -2233,6 +2233,100 @@ def get_netvalue_data(
...
+    @staticmethod
+    def _infer_zero_timed_balances(
+            cursor: 'DBCursor',
+            balances: list[SingleDBAssetBalance],
+            from_ts: Optional[Timestamp] = None,
+            to_ts: Optional[Timestamp] = None,
+    ) -> list[SingleDBAssetBalance]:
+        """
+        Given a list of asset specific timed balances, infers the missing zero timed balances
+        for the asset. We add 0 balances on the start and end of a period of 0 balances.
```
LefterisJP, line 2245: Maybe also add a link to the issue this addresses

LefterisJP (commented):
Some tests failed and I also left some comments

rotkehlchen/db/dbhandler.py
LefterisJP: As discussed extensively in discord both python and sqlite approach turned out to be pretty similar. @dimyG implemented it and run the appropriate timing tests: https://discord.com/channels/657906918408585217/1080783409640648765/1117864559185768548

So we ended up going only with python approach and as a configurable setting that is off by default due to slowing down graphs quite a bit.

LefterisJP (approved):
LGTM. Thank you Dimitris.

PR #6191, opened 2023-06-02
correctly pick the explorer url for the different evm events

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/accounting/structures/processed_event.py
+++ b/rotkehlchen/accounting/structures/processed_event.py
@@ -1,7 +1,8 @@
 import builtins
 import json
 from dataclasses import dataclass, field
-from typing import Any, Callable, Optional, TypeVar
+from enum import Enum, auto
+from typing import Any, Callable, Literal, Optional, TypeVar, overload
 
 from rotkehlchen.accounting.cost_basis import CostBasisInfo
 from rotkehlchen.accounting.mixins.event import AccountingEventType
@@ -17,6 +18,12 @@ from rotkehlchen.types import Location, Price, Timestamp
 from rotkehlchen.utils.serialization import rlk_jsondumps
 
 T = TypeVar('T', bound='ProcessedAccountingEvent')
+
+
+class AccountingEventExportType(Enum):
+    API = auto()
+    CSV = auto()
+    DB = auto()
 
 
 @dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=False)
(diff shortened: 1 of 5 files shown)


rotkehlchen/tests/utils/pnl_report.py
```diff
@@ -38,30 +39,30 @@ def query_api_create_and_get_report(
...
             'limit': events_limit,
             'order_by_attributes': ['timestamp'],
             'ascending': [events_ascending_timestamp],
         },
-    )
-    events_result = assert_proper_response_with_result(response)
-    return report_id, report_result, events_result
+    ) as response:
+        events_result = assert_proper_response_with_result(response)
+        return report_id, report_result, events_result
```
LefterisJP, line 68: This does not need to be under the context manager

```suggestion
    return report_id, report_result, events_result
```

rotkehlchen/accounting/structures/processed_event.py
```diff
@@ -56,13 +63,15 @@ def to_string(self, ts_converter: Callable[[Timestamp], str]) -> str:
     def to_exported_dict(
             self,
             ts_converter: Callable[[Timestamp], str],
-            eth_explorer: Optional[str],
-            for_api: bool,
+            export_type: AccountingEventExportType,
```
LefterisJP, line 66: Hmm but this still does not show the connection I meant needs to be typechecked. Please add two typing overloads here showing that if the export_type == AccountingEventExportType.CSV then the evm_explorer needs to be given.

rotkehlchen/accounting/structures/processed_event.py
```diff
@@ -77,22 +86,22 @@ def to_exported_dict(
...
             taxable_basis = free_basis = ''
             if self.cost_basis is not None:
                 taxable_basis, free_basis = self.cost_basis.to_string(ts_converter)
             exported_dict['cost_basis_taxable'] = taxable_basis
             exported_dict['cost_basis_free'] = free_basis
             exported_dict['asset'] = str(self.asset)
-            if tx_hash:
-                exported_dict['notes'] = f'{eth_explorer}{tx_hash}  ->  {self.notes}'
-        else:
+            if tx_hash is not None and evm_explorer is not None:
```
LefterisJP, line 96: Shouldn't it always be given?

```suggestion
            if tx_hash is not None:
```

LefterisJP (approved):
LGTM but need a small change. Tell me if you want it here or in another PR. Herre would make sense imo as it's part of this.

rotkehlchen/accounting/structures/processed_event.py
```diff
@@ -53,16 +60,35 @@ def to_string(self, ts_converter: Callable[[Timestamp], str]) -> str:
 
         return desc
 
+    @overload
+    def to_exported_dict(
+            self,
+            ts_converter: Callable[[Timestamp], str],
+            export_type: Literal[AccountingEventExportType.CSV],
+            evm_explorer: Optional[str],
```
LefterisJP, line 68: Why is this optional here? Shouldn't it be always given?

PR #6199, opened 2023-06-03
fix(frontend): account label removal issue

### LINKED ISSUE #6189: Removing a label does not update the account display name
## Problem Definition

1. Have an account with a label, such as: "My metamask".
2. Pull all history and events and see it displayed everywhere with the name "My metamask"
3. Go to accounts & balances and edit the account to remove the label. You do this by putting empty string as label.
4. Press save.
5. Now navigate anywhere in the app, either accounts & balances or history and unfortunately you still see the old label everywhere. It's not updated.
6. It only updates if the app is restarted.

It's important to note that t**his bug only applies to removing the label. If you change the string to anything else but the label still exists then it updates properly throughout the app.**

## Task

Fix it.

lukicenturi:
Closes #6189

## Checklist


kelsos (approved):
lgtm @lukicenturi thanks

PR #6222, opened 2023-06-07
feat: floating data table header

lukicenturi:
Closes #(issue_number)

## Checklist


frontend/app/package.json
```diff
@@ -112,6 +112,7 @@
     "unplugin-auto-import": "0.16.4",
     "unplugin-vue-components": "0.25.0",
     "unplugin-vue-define-options": "1.3.7",
+    "vh-sticky-table-header": "1.5.0",
```
kelsos, line 115: How much work is to implement this functionality as part of the application?

kelsos: In general, we avoid dependencies such as this, it is a library that is small, doesn't have traction and a single person maintains it.

We will eventually run into similar issues as with the `vue-cli-electron-builder` thing. I would prefer if we don't add this kind of dependency.

kelsos: If this is a leftover from trial an error then just remove it 😄

frontend/app/src/types/frontend-settings.ts
```diff
@@ -171,7 +171,7 @@ export const FrontendSettings = z.object({
     RefreshPeriod.default(-1)
   ),
   explorers: ExplorersSettings.default({}),
-  itemsPerPage: z.number().positive().int().default(10),
+  itemsPerPage: z.number().positive().int().default(100),
```
kelsos, line 174: is that change only for testing?

frontend/app/package.json
lukicenturi: was playing around, I need to modify a lot of the algorithms😛

lukicenturi:
not sure why, but force push doesn't update the UI

PR #6156, opened 2023-05-27
Upgrade to python 3.10

LefterisJP:
135937807-fd3e0fd2-a31a-47a4-90c6-b0bb1d0704d4

PR #6224, opened 2023-06-07
Add polygon PoS to rotkehlchen.spec

LefterisJP:
Moving the commit from https://github.com/rotki/rotki/pull/6156 since we discussed we should not force 3.10 in upcoming release

### DIFF
--- a/rotkehlchen.spec
+++ b/rotkehlchen.spec
@@ -83,9 +83,11 @@ for exchange_name in SUPPORTED_EXCHANGES:
         continue
     hiddenimports.append(f'rotkehlchen.exchanges.{exchange_name}')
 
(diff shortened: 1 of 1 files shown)


yabirgb (commented):
For the failing test I recall reading that yesterday it was reported to be slow the action of migrating the archive nodes

LefterisJP:
> For the failing test I recall reading that yesterday it was reported to be slow the action of migrating the archive nodes

It passed locally.

LefterisJP:
Tihis test `rotkehlchen/tests/unit/test_optimism_inquirer.py::test_optimism_nodes_prune_and_archive_status` seems to fail often in CI since bedrock but passes locally.

PR #6230, opened 2023-06-08
fix: cannot set backend setting to default value in frontend

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/settings/OnboardingSettings.vue
+++ b/frontend/app/src/components/settings/OnboardingSettings.vue
@@ -8,27 +8,19 @@ import {
   required
 } from '@vuelidate/validators';
 import { type Ref } from 'vue';
+import isEqual from 'lodash/isEqual';
 import { type BackendOptions } from '@/electron-main/ipc';
 import { type Writeable } from '@/types';
 import { LogLevel } from '@/utils/log-level';
-import {
-  type BackendConfiguration,
-  type DefaultBackendArguments
-} from '@/types/backend';
+import { type BackendConfiguration } from '@/types/backend';
 
 const emit = defineEmits<{
   (e: 'dismiss'): void;
 }>();
 
-const { dataDirectory } = storeToRefs(useMainStore());
+const { dataDirectory, defaultBackendArguments } = storeToRefs(useMainStore());
 
-const userDataDirectory: Ref<string> = ref('');
-const userLogDirectory: Ref<string> = ref('');
-const logFromOtherModules: Ref<boolean> = ref(false);
-const maxLogSize: Ref<number> = ref(0);
-const sqliteInstructions: Ref<number> = ref(0);
-const maxLogFiles: Ref<number> = ref(0);
-const formValid: Ref<boolean> = ref(false);
+const valid: Ref<boolean> = ref(false);
 
 const { backendSettings } = useSettingsApi();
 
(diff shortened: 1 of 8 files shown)


kelsos (commented):
Would it makes sense to also have a unittest for the composable it self that checks they the correct setting is returned when merged? If you don't have anything for example or if you have a file for etc?

frontend/app/src/components/settings/OnboardingSettings.vue
```diff
@@ -38,74 +30,78 @@ const configuration: Ref<BackendConfiguration> = asyncComputed(() =>
...
   options
 } = useBackendManagement(loaded);
 
+const logLevel: Ref<LogLevel> = ref<LogLevel>(get(defaultLogLevel));
+const userDataDirectory: Ref<string> = ref('');
+const userLogDirectory: Ref<string> = ref('');
+const logFromOtherModules: Ref<boolean> = ref(false);
+const maxLogSize: Ref<number> = ref(0);
+const sqliteInstructions: Ref<number> = ref(0);
+const maxLogFiles: Ref<number> = ref(0);
```
kelsos, line 86: Is there any reason why we can't leave reference declarations at the top of the file?

frontend/app/src/components/settings/OnboardingSettings.vue
```diff
@@ -116,71 +112,66 @@ const resetDefaultArguments = (field: 'files' | 'size' | 'instructions') => {
...
+
+  return !isEqual(form, get(initialOptions));
 });
 
 const { openDirectory } = useInterop();
 
 const nonNegativeNumberRules = {
   required: helpers.withMessage(
-    tc('backend_settings.errors.non_empty'),
+    t('backend_settings.errors.non_empty').toString(),
```
kelsos, line 170: This might create some conflict when merging to develop can you undo it?

frontend/app/tests/unit/specs/components/settings/onboarding-settings.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+      },
+      dataDirectory: '/Users/home/rotki/develop_data',
+      logLevel: 'DEBUG',
+      acceptDockerRisk: false,
+      backendDefaultArguments: {
+        maxLogfilesNum: 3,
+        maxSizeInMbAllLogs: 300,
+        sqliteInstructions: 5000
+      }
+    }),
```
kelsos, line 54: does this mock an API response? do you think it would be possible to use `msw` instead?

frontend/app/tests/unit/specs/components/settings/onboarding-settings.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+    await useMainStore().getInfo();
+  });
+
+  beforeEach(async () => {
+    wrapper = createWrapper();
+    await wrapper.vm.$nextTick();
+  });
+
+  describe('standard settings', () => {
+    test('use default value from info api or electron config, save button should be disabled', async () => {
```
kelsos, line 102: After merging on develop we should change this too to follow the new approach in form validation.

frontend/app/tests/unit/specs/components/settings/onboarding-settings.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+      expect(logLevelInput.value).toBe('debug');
+
+      expect(
+        wrapper
+          .find('[data-cy=onboarding-setting__submit-button]')
+          .attributes('disabled')
+      ).toBe('disabled');
+    });
+
+    test('should save the setting', async () => {
```
kelsos, line 124: We also need a test for the bug we are fixing where you change the loglevel from the default and then you can't change it back from the default.

kelsos (approved):

PR #6231, opened 2023-06-08
Decode aave suply of ETH correctly

yabirgb:
Aave has a special contract for when a wrap of ETH is needed. If the user used this contract the values in the deposit event were not the expected ones and this caused the transaction to not be properly decoded


rotkehlchen/chain/ethereum/modules/aave/v2/decoder.py
```diff
@@ -74,9 +78,11 @@ def _decode_deposit(
     ) -> None:
         """Decode aave v2 deposit event"""
         user = hex_or_bytes_to_address(tx_log.data[:32])
-        on_behalf_of = hex_or_bytes_to_address(tx_log.data[64:96]) if hex_or_bytes_to_str(tx_log.data[64:96]) != '' else None  # noqa: E501
+        on_behalf_of = hex_or_bytes_to_address(tx_log.topics[2]) if hex_or_bytes_to_str(tx_log.topics[2]) != '' else None  # noqa: E501
```
LefterisJP, line 81: Why did you change this? Was it wrong before?

yabirgb: yes, I have checked and there is no third element in the data, there are 3 topics and in the data only one address and one amount

LefterisJP (approved):
LGTM.

Don't forget what I mentioned about cassetes repo

PR #6233, opened 2023-06-08
Fix some tests that were failing on windows

### LINKED ISSUE #5168: Figure out and fix slow windows CI test runs and enable windows backend tests
## Enabling pytest on Windows.

The reusable workflow for the backend is windows ready, however, multiple tries yesterday proved that running on windows is extremely slow.

While trying different options and arguments the tests manage to get from 16%-20% of the test suite in 1h10 minutes.

For this reason the windows nightly was removed.

## TODO

To activate windows we need to figure out:
1. If this is a CI situation or also happens in a Windows PC installation.
2. Figure out what causes the tests to be extremely slow and fix it.

> **Note**! Someone should still check the full test suite on a Windows machine since there were a number of errors that didn't appear on Linux/macOS (a lot of them related to the filesystem/temporary files)

_Originally posted by @kelsos in https://github.com/rotki/rotki/issues/5167#issuecomment-1326204677_

IanMichaelHarper:
Partially closes #5168 by fixing some of the failing tests. 

My initial changes fixed most of the tests that failed when I ran them locally on my windows machine, but then I noticed not nearly as many tests fail on the CI in a windows env when no changes are applied. I relied less on my personal machine and more on the CI in Github Actions, but then ran into the curious case where as I fixed tests, other tests started to fail. I now think the Summary in Github Actions isn't telling me the full story and maybe more tests are failing but it is not showing them to me in the Summary. Maybe my personal windows machine is in fact accurate. Nevertheless I thought I'd open a PR with my changes so far to fix some of the tests that fail on windows and get some feedback.

It seems some tests failed when run together but passed individually, so I added `@pytest.mark.parametrize('use_clean_caching_directory', [True])` to these tests.

Windows doesn't seem to like `Path`'s `/` for directory concatenation, so I replaced with `os.path.join()` instead. 

Most other tests were failing for various different reasons.

Also added some typing.


rotkehlchen/tests/api/test_liquity.py
```diff
@@ -275,7 +277,10 @@ def test_staking_stats(rotkehlchen_api_server, ethereum_accounts):
     assert len(global_stats['staking_gains']) == 3
     assert len(global_stats['stability_pool_gains']) == 2
     assert FVal(global_stats['stability_pool_gains'][0]['amount']).is_close(FVal(address_0_data['stability_pool_gains'][0]['amount']) + FVal(address_1_data['stability_pool_gains'][0]['amount']), max_diff='1e-8')  # noqa: E501
-    assert address_0_data['stability_pool_gains'][0]['amount'] == '1.7820064710306824'
+    assert math.isclose(
+        float(address_0_data['stability_pool_gains'][0]['amount']), 1.7820064710306824,
+        rel_tol=1e-15,
+    )
```
IanMichaelHarper, line 283: Windows yields `1.7820064710306822`. I'd put that down to floating point precision.

yabirgb:
Thank you very much @IanMichaelHarper for the PR! We will check it, most probably one of our developers using windows. We fixed recently an issue in our tests also related to a setup that was leaving files open. Did you notice any improvement in windows with the latest commits in develop?

IanMichaelHarper:
> Thank you very much @IanMichaelHarper for the PR! We will check it, most probably one of our developers using windows. We fixed recently an issue in our tests also related to a setup that was leaving files open. Did you notice any improvement in windows with the latest commits in develop?

I hadn't checked this week but I didn't notice much of an improvement last weekend

PR #6234, opened 2023-06-08
Fix edge case in uniswap liquidity withdrawal

yabirgb:
There are two types of withdrawals in uniswap LPs, one where eth is withdrawn and another where WETH is withdrawn. As it was before the WETH withdrawal was not properly decoded since it was looking for ETH instead of ETH due to a condition in the assert searched.

The solution proposed is to look for eth and weth in the events and once found create the correct transform instruction

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,6 +3,7 @@ Changelog
 =========
 
 * :bug:`-` Fix issue for ledger actions with zero rate, where the rate isn't used in the PnL report.
+* :bug:`-` Events where the user withdraws WETH from a Uniswap V2 liquidity position will be now correctly decoded.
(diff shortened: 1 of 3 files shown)


docs/changelog.rst
```diff
@@ -3,6 +3,7 @@ Changelog
 =========
 
 * :bug:`-` Fix issue for ledger actions with zero rate, where the rate isn't used in the PnL report.
+* :bug:`-` Events where the user withdraws WETH from a Uniswap V2 liquidity position will be now correctly decoded.
```
LefterisJP, line 6: wording nitpick

```suggestion
* :bug:`-` Events where the user withdraws WETH from a Uniswap V2 liquidity position will now be correctly decoded.
```

LefterisJP (approved):
tiny nitpick typo. Will merge. also remember same thing for cassetes repo here

PR #6244, opened 2023-06-12
fix: also take perProtocolMappings into subtype options

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/tests/unit/fixtures/history-event-counterparties.json
+++ b/frontend/app/tests/unit/fixtures/history-event-counterparties.json
@@ -1,0 +1,40 @@
+{
+  "result": [
+    {
+      "identifier": "gitcoin",
+      "label": "Gitcoin",
+      "image": "gitcoin.svg"
+    },
+    {
+      "identifier": "1inch-v1",
+      "label": "1inch",
+      "image": "1inch.svg"
+    },
+    {
+      "identifier": "convex",
+      "label": "Convex",
+      "image": "convex.jpeg"
+    },
+    {
+      "identifier": "curve",
+      "label": "Curve.fi",
+      "image": "curve.png"
+    },
+    {
+      "identifier": "yearn-v2",
+      "label": "Yearn",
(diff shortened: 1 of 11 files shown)


kelsos:
@lukicenturi can you provide a small explanation of how to test this change?

What can I select and what should I expect?

Event better is it possible to add some test for this?

lukicenturi:
> @lukicenturi can you provide a small explanation of how to test this change?
> 
> What can I select and what should I expect?
> 
> Event better is it possible to add some test for this?

For a brief explanation, if you go to history events form
Then you set the eventType to `deposit` you will see 4 `eventSubTypes` options.
If then you choose protocol `yearn-v2` you will see  additional `receive wrapped` in the `eventSubTypes` options.

It's because there are eventSubtypes that only exists on `perProtocolMappings` but not in `globalMappings`.
Will add the test

kelsos:
> > @lukicenturi can you provide a small explanation of how to test this change?
> > What can I select and what should I expect?
> > Event better is it possible to add some test for this?
> 
> For a brief explanation, if you go to history events form Then you set the eventType to `deposit` you will see 4 `eventSubTypes` options. If then you choose protocol `yearn-v2` you will see additional `receive wrapped` in the `eventSubTypes` options.
> 
> It's because there are eventSubtypes that only exists on `perProtocolMappings` but not in `globalMappings`. Will add the test

Great thank you I will give it a try then to see how things work, thanks for the details.

frontend/app/tests/unit/fixtures/history-event-counterparties.json
```diff
@@ -0,0 +1,220 @@
+{
+  "result": [
```
kelsos, line 2: Do you think we can reduce the content of the fixtures to the bare minimum that is required to test the functionality? This looks a bit like the full response from the API, we certainly don't need everything.

kelsos (approved):
looks good @lukicenturi thank you

Issue #6243, opened 2023-06-12
frontend: add a transaction (evm) reset button


kelsos:
## Abstract

Add a button in the overflow menu to reset a transaction. The button will delete any customized events the user might have and re-decode the specific transaction.

## Motivation

There is no way to reset a single event you can only ignore.
Deleting and re-decoding can work but it is not obvious.

## Specification

- [ ] Add a reset button that deletes any customized evm transaction and re-decodes events for this transaction.

PR #6245, opened 2023-06-12
Add force_delete for history event deletion

LefterisJP:
Backend part of https://github.com/rotki/rotki/issues/6243. To enable that we create a special flag in history event deletion where we ignore the check for the last history event in a transaction.


LefterisJP:
The only test that failed is not relevant to the PR. So will merge.

PR #6246, opened 2023-06-12
feat: reset EVM transaction events

### LINKED ISSUE #6243: frontend: add a transaction (evm) reset button

lukicenturi:
Closes #6243

## Checklist

### DIFF
--- a/frontend/app/src/components/history/events/HistoryEventFormDialog.vue
+++ b/frontend/app/src/components/history/events/HistoryEventFormDialog.vue
@@ -32,7 +32,7 @@ const title: ComputedRef<string> = computed(() =>
     :display="openDialog"
     :title="title"
     :primary-action="t('common.actions.save')"
-    :action-disabled="loading"
+    :action-disabled="loading && !editableItem"
     :loading="submitting"
     @confirm="trySubmit()"
     @cancel="closeDialog()"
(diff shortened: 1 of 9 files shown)


kelsos:
@lukicenturi for this as I wrote in discord

- We need to make sure that editing is possible even when loading (unless it's saving)
- We need to only disable reset only when we re-decode.

lukicenturi:
> @lukicenturi for this as I wrote in discord
> 
> * We need to make sure that editing is possible even when loading (unless it's saving)
> * We need to only disable reset only when we re-decode.

This comment has been addressed now on the last commit

kelsos:
> > @lukicenturi for this as I wrote in discord
> > 
> > * We need to make sure that editing is possible even when loading (unless it's saving)
> > * We need to only disable reset only when we re-decode.
> 
> This comment has been addressed now on the last commit

Thanks, I am gonna check then.

kelsos:
We need a changelog entry for this btw

PR #6261, opened 2023-06-15
Fix kyber decoder for trades

yabirgb:
The decoder was using the wrong tx log topic to filter trandes and it was failing to decode the transaction because it matched againt an event that didn't have the expected fields.

This was not noticed in tests because I didn't use all the logs and just some of them. By mocking the two tests it is now properly tested.

Also fixed an issue with uniswap front running the kyber decoder


rotkehlchen/chain/ethereum/modules/kyber/decoder.py
```diff
@@ -91,7 +94,7 @@ def _maybe_update_events_legacy_contrats(
 class KyberDecoder(DecoderInterface):
 
     def _decode_legacy_trade(self, context: DecoderContext) -> DecodingOutput:
-        if context.tx_log.topics[0] == KYBER_TRADE_LEGACY:
+        if context.tx_log.topics[0] != KYBER_TRADE:
```
LefterisJP, line 97: why this change? Add a comment

LefterisJP (approved):
LGTM. Hopefully this solves kelsos' problem. Left a nitpick for later

PR #6267, opened 2023-06-16
Specify transaction must exist and if yes query more nodes

LefterisJP:
Since some open rpc nodes are load balancing between pruned and non-pruned nodes it's possible to use a node and have the transaction exist and then have it not exist or its receipt missing in the same function.

This break stuff, so we add extra logic so that a non-existing transaction when it should exist is seen as error and continues to query other nodes


LefterisJP:
all green. mergimentation.

PR #6268, opened 2023-06-16
Allow for tasks that can raise error when logout

yabirgb:
There are task that can be killed during the logout but don't acquire the lock used for the tasks in the logout so can be kill while executing a query to the database and this would raise an unhandled exception. In those cases don't raise an error to the frontend but just silently log it.

Closes #(issue_number)

## Checklist


LefterisJP (approved):
LGTM!

PR #6270, opened 2023-06-16
Change order of things during logout

kelsos:
Closes #(issue_number)

## Checklist

## Notes
What as happening is that during logout the state was cleared first and then the navigation happened. 

Because some of the elements where still visible they would trigger a refresh of state after the logout. (e.g. fetching assets mappings)

### DIFF
--- a/frontend/app/src/store/plugins.ts
+++ b/frontend/app/src/store/plugins.ts
@@ -13,6 +13,11 @@ export const StoreResetPlugin = ({ store }: PiniaPluginContext): void => {
       state = cloneDeep(initialState);
     }
     store.$patch(toReactive(state));
+    // If a store has its own reset function (e.g. to cancel actions)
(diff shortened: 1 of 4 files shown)


frontend/app/src/store/plugins.ts
```diff
@@ -13,6 +13,11 @@ export const StoreResetPlugin = ({ store }: PiniaPluginContext): void => {
       state = cloneDeep(initialState);
     }
     store.$patch(toReactive(state));
+    // If a store has its own reset function (e.g. to cancel actions)
+    // Then we will also call this to properly cleanup.
+    if (store.reset && typeof store.reset === 'function') {
```
tewshi, line 18: `typeof store.reset === 'function'` will be enough here, yeah?

tewshi (approved):
other than the little nitpick, lgtm!

PR #6273, opened 2023-06-17
Added Stale bot

Narcos101:
This PR:

- Adds a stale bot to manage issues

### DIFF
--- a/.github/workflows/stale-bot.yml
+++ b/.github/workflows/stale-bot.yml
@@ -1,0 +1,49 @@
+name: Triage Stale Issues
+
+on:
+  schedule:
+  - cron: "0 * * * *"
+  workflow_dispatch:
+
+jobs:
+  issue-cleanup:
+    runs-on: ubuntu-latest
+    name: Triage Stale Issues
+    steps:
+    - uses: aws-actions/stale-issue-cleanup@v4
+      with:
+        issue-types: issues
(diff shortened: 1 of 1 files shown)


LefterisJP:
Ah  hey @Narcos101 for stale bot I don't think it would work well for us as we do have issues which we do eventually plan to address years later and we keep them as reminders of things we want to work on.

Narcos101:
Hey @LefterisJP I think such things should better be discussed on the Discussions page. Anyways if it doesn't work for you. I am ready to contribute another workflow for welcoming new contributors to the repository. Any thoughts on that?

LefterisJP:
> Hey @LefterisJP I think such things should better be discussed on the Discussions page. 

Discussions are not as easy to track and use in milestones and this is our current workflow anyway so I don't think we will want to change it, at least at this point.

> Anyways if it doesn't work for you. I am ready to contribute another workflow for welcoming new contributors to the repository. Any thoughts on that?

Yes. Anything is better than what we have right now. The biggest problem with rotki and new contributors is how to find easy stuff to work on. It has turned out many times that most of the issues we had tagged as "good first issue" were actually not that easy.

It's hard to find out small things that are easy to pick up for a newcomer and also to explain how the codebase works. We can discus it in discord if you want!

Thank you.

PR #6276, opened 2023-06-18
Optimism gas

### LINKED ISSUE #6033: Optimism gas fees are not taking L1 into account
## Problem Definition

The optimism gas fees seem to not be calculated correctly by rotki.

rotki seems to calculate smaller fees than what etherscan shows. As an example let's take transaction: https://optimistic.etherscan.io/tx/0x8743e91d59f79135d0d81854c2825ccee9873d87b2d60945ab8a8712c50151ee

Etherscan
2023-05-10-130333_672x216_scrot

rotki
2023-05-10-130605_884x345_scrot

So it seems we are not taking L1 gas fees at all into the calculation. We need to fix this.

## Task

Figure out the exact formula and if more data need to also be pulled for each transaction in order for the calclation to be correct. Then plan out a way to implement it.

0xGusMcCrae:
Closes #6033

Summary of changes:

- Adds an optional l1_fee attribute to EvmTransaction. For Optimism transactions, this value will be pulled from etherscan as part of the transaction receipt and added to the EvmTransaction.  Optimistic etherscan computes this value itself, so it should still be accurate for the changes that came with the bedrock upgrade. At the least, the values shown in rotki match those shown on etherscan with the calculation I've got it using.
- The node_inquirer was altered so that it always uses etherscan for optimistic transactions since web3.py doesn't have support for the additional values needed for the l1 fee calculation
- The EvmTransaction database was altered to accomodate the new l1_fee value. This will be NULL for non-optimistic transactions.
- Updates to test files to reflect the correct gas fees and new EvmTransaction table/database structure.

Note: 
I'm not sure if way I've got it coded currently is how you'd prefer it. Should I override the necessary functions in the optimism specific node_enquirer, decoder etc? Or is it fine having conditonals checking for optimistic txs in the base EVM versions? 

For test_evm_tx_decoding.py, I needed to add the l1_fee column to the ethtxs.db sample test data. I didn't have the password to the database, so I added `cursor.execute('ALTER TABLE evm_transactions ADD COLUMN l1_fee TEXT')` at line 118 to get it to pass (https://github.com/0xGusMcCrae/rotki/blob/67951ff82371ce88bfbd25ae372c120a5c1af19f/rotkehlchen/tests/unit/test_evm_tx_decoding.py#L118). If I can get the password to edit the actual database file, I can probably fix it so this isn't necessary.


docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` Fix gas fee calculation for Optimism transactions to include L1 fees.
```
LefterisJP, line 5: add the issue number

```suggestion
* :bug:`6033` Fix gas fee calculation for Optimism transactions to include L1 fees.
```

LefterisJP (commented):
Hey @0xGusMcCrae thanks a lot for your PR trying to tackle a quite difficult issue. :muscle: 

As you can see from my comments quite a few changes in the approach are needed.

And as I told you in the issue this is quite complicated as a problem.

But it's really cool to see you delve into the code and the problem. Very happy to see the attempts and thinking.

So about the "right" approach here. It's a big question mark and I am not totally sure myself as I have not put any really serious thinking here.

But perhaps things to consider would be:

1. Since this l1fee a different value, optional only to optimism (maybe also arbitrum?) it would make sense to split it off? Both in the DB in a different table which would connect with a foreign key relation to the evm transactions for optimism?
2. Perhaps turn EvmTransaction into a frozen dataclass with inheritance and also create OptimismTransaction ... or RollupTransaction (if the same kind of thinking also applied to Arbitrum) which just adds the extra field on top.
3. The tricky part and question is what to do for the l1 fee. As you said, only etherscan has it, but we can't limit our queries only to it. The open rpc nodes are invaluable. We should not stop querying them for optimism. Perhaps the fee can be added if it's an etherscan query and set to `None` otherwise. Then create a periodic task to run (in the same spirit as the query of missing prices: https://github.com/rotki/rotki/blob/90fa68b412dc9be211491caba5e165a0870b2919/rotkehlchen/tasks/manager.py#L394) and find missing L1 fees from optimism transactions, query them, add them to the DB and trigger a redecoding of the specific events tied to those transactions (to calculate the gas correctly).
4. Do the other stuff I pointed out in the code.

This is definitely not an easy first issue to tackle but you seem to be getting the hang of it!
Hope my blurbing ideas at 01:40 at night make sense :smile:

0xGusMcCrae:
Thanks for the feedback. I'll get started implementing the changes and try to figure something out for the node inquiries. I'll pop into the discord if I have any questions.

Closing this PR as I think I'm going to start on a fresh branch rather than undo the current changes.

PR #6277, opened 2023-06-18
cowswap trade fees

LefterisJP:
Display fees in cowswap trades and take them into account in accounting


LefterisJP:
Probably lots of tests will fail, especially in decoding due to my changes to the way events are shuffled. Will have to address them

LefterisJP:
Well this is all green. nice. Cowswap fees go in then.

PR #6279, opened 2023-06-19
Make the total lost amount always positive

yabirgb:
After 9189ce73c237dd14bce043e05f4fd7f14ab70752 the aave eth events got correctly decoded and this uncovered an issue in the aave stats.In the test the total lost amount was getting displayed as negative instead of positive.

It didn't appear before because it came from a liquidation event that was being correctly accounted and the debt event that existed before was getting cancelled.


LefterisJP (approved):
I hereby approve of this PR by the powers vested upon me by the settings of this repository.
giphy (13)

Issue #6282, opened 2023-06-19
The backend query task died unexpectedly: 'upload_ts'


PabloCastellano:
## Problem Definition

Yesterday I played a bit with rotki PnL reports, renewed my subscription and created different accounts in rotki to test different APIs and data import methods.

Today I can't log in to the account that I created yesterday and worked fine until now.

![image](https://github.com/rotki/rotki/assets/73274/10e3feec-4afd-4e61-bed8-3366f5da34df)

## Logs

```
[19/06/2023 15:25:52 CEST] ERROR rotkehlchen.api.rest Greenlet with id 140171512169920: Greenlet for task 0 dies with exception: 'upload_ts'.
Exception Name: <class 'KeyError'>
Exception Info: 'upload_ts'
Traceback:
   File "src/gevent/greenlet.py", line 908, in gevent._gevent_cgreenlet.Greenlet.run
  File "rotkehlchen/api/rest.py", line 362, in _do_query_async
  File "rotkehlchen/api/rest.py", line 1132, in user_login
  File "rotkehlchen/rotkehlchen.py", line 295, in unlock_user
  File "rotkehlchen/premium/sync.py", line 285, in try_premium_at_start
  File "rotkehlchen/premium/premium.py", line 373, in premium_create_and_verify
  File "rotkehlchen/premium/premium.py", line 174, in is_active
  File "rotkehlchen/premium/premium.py", line 300, in query_last_data_metadata
```

### System Description

Operating system: Linux
Rotki version: v1.28.0

LefterisJP:
Thanks for finding this out @PabloCastellano !

This is a problem with our server that started right about now. We are looking at it. It affects all premium users and we will solve it asap. Once done will close the issue.

The issue though seems to have uncovered a bad handling of the error there. Should have at least let you login properly.

LefterisJP:
Okay so all premium users should be able to log in fine again. :sweat_smile: 

Also opened a PR to not fail in such an ugly way if the rotki server returns unexpected (but valid) response again:

PR #6284, opened 2023-06-19
Better error handling if rotki server gives unexpected response

LefterisJP:
Partially address what is seen at
https://github.com/rotki/rotki/issues/6282 if the server ever responds with unexpected results again.

### DIFF
--- a/rotkehlchen/premium/premium.py
+++ b/rotkehlchen/premium/premium.py
@@ -29,7 +29,6 @@ log = RotkehlchenLogsAdapter(logger)
 
 HANDLABLE_STATUS_CODES = [
     HTTPStatus.OK,
-    HTTPStatus.NOT_FOUND,
     HTTPStatus.UNAUTHORIZED,
     HTTPStatus.BAD_REQUEST,
 ]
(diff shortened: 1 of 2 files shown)


rotkehlchen/premium/premium.py
```diff
@@ -296,12 +296,19 @@ def query_last_data_metadata(self) -> RemoteMetadata:
             raise RemoteError(msg) from e
 
         result = _process_dict_response(response)
```
yabirgb, line 298: should we check the status code here? The server returned a 404 that is allowed in this function but doesn't raise any error

rotkehlchen/premium/premium.py
```diff
@@ -296,12 +296,19 @@ def query_last_data_metadata(self) -> RemoteMetadata:
...
+        try:
+            metadata = RemoteMetadata(
+                upload_ts=Timestamp(result['upload_ts']),
+                last_modify_ts=Timestamp(result['last_modify_ts']),
+                data_hash=result['data_hash'],
+                data_size=result['data_size'],
+            )
+        except KeyError as e:
+            msg = f'Problem connecting to rotki server. last_data_metadata response missing {e!s} key'  # noqa: E501
+            log.error(f'{msg}. Response was {result}')
```
yabirgb, line 309: Also here the error is silent for the user. The only thing that will be noticed is that the account doesn't have the premium status

LefterisJP: > Also here the error is silent for the user. The only thing that will be noticed is that the account doesn't have the premium status

that's how all these errors are so far handled by rotki no? All failures to connec to rotki server.

rotkehlchen/premium/premium.py
LefterisJP: ah good point. Will add it.

yabirgb (commented):
test need adjustment

yabirgb (approved):
LGTM

PR #6286, opened 2023-06-20
feat: update labeled address display

lukicenturi:
- Fix the ellipsis issue where it's truncated when the space is still to many.
- Keep showing the ENS in the tooltip if additional label is added.
- Simplify the code to do the truncation.

### DIFF
--- a/frontend/app/src/filters.ts
+++ b/frontend/app/src/filters.ts
@@ -20,12 +20,14 @@ export function truncateAddress(address: string, truncLength = 4): string {
     ? 4
     : 0;
 
-  if (address.length <= truncLength * 2 + startPadding) {
+  const length = address.length;
+
+  if (length <= truncLength * 2 + startPadding) {
     return address;
   }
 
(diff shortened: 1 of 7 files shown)


frontend/app/src/filters.ts
```diff
@@ -20,12 +20,14 @@ export function truncateAddress(address: string, truncLength = 4): string {
...
+  if (length <= truncLength * 2 + startPadding) {
     return address;
   }
 
-  return `${address.slice(0, truncLength + startPadding)}...${address!.slice(
-    address!.length - truncLength,
-    address!.length
+  return `${address.slice(0, truncLength + startPadding)}...${address.slice(
+    length - truncLength,
+    length
```
kelsos, line 31: After the release, we must remove the `filters.t`s and move the utilities under `src/utils/.

This is a leftover from before we started the migration.

lukicenturi: I think we can also do that after this PR, just moving the file right

kelsos (approved):
lgtm

PR #6296, opened 2023-06-21
Update spam assets in data migration

yabirgb:
Update the spam assets before detecting new addresses to avoid adding addresses without meaningful transactions


LefterisJP (approved):
LGTM. The failed tests are again unfortunately optimism problems and uniswap subgraph problems

PR #6301, opened 2023-06-21
Bump many packages and change code due to the bumps

LefterisJP:
- Both linting packages here forced code changes
- But also bumped some production packages that should be bumped before release

### DIFF
--- a/rotkehlchen/chain/ethereum/graph.py
+++ b/rotkehlchen/chain/ethereum/graph.py
@@ -56,7 +56,7 @@ def get_common_params(
     return param_types, param_values
 
 
-class Graph():
+class Graph:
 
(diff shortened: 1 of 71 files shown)


LefterisJP:
All green here. Too green to not just merge.

rotkehlchen/chain/ethereum/graph.py
```diff
@@ -56,7 +56,7 @@ def get_common_params(
     return param_types, param_values
 
 
-class Graph():
+class Graph:
```
yabirgb, line 59: Thank god you removed all this empty parenthesis :pray:

LefterisJP: Just a new ruff rule. Consistency is great!

PR #6303, opened 2023-06-22
feat: group evm chain for statistic to false

lukicenturi:
Closes #(issue_number)

Fix this issue where current value is not used for the graph (only happen for asset that have multiple chains)
image


kelsos (approved):
lgtm

PR #6304, opened 2023-06-22
Add polygon_pos connect at fixture start

LefterisJP:


LefterisJP:
> patch_no_op_unlock

seems to just set them all to []

PR #6305, opened 2023-06-22
Fix error adding multiple evm tokens with the same name and symbol

yabirgb:
We weren't checking for evm tokens existence against the identifier but the combination of name and symbol that can be the same.

Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -51,6 +51,7 @@ from rotkehlchen.assets.asset import (
     AssetWithNameAndType,
     AssetWithOracles,
     CustomAsset,
+    EvmToken,
     FiatAsset,
 )
 from rotkehlchen.assets.resolver import AssetResolver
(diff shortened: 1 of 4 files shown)


rotkehlchen/globaldb/handler.py
```diff
@@ -754,11 +754,7 @@ def get_evm_token_identifier(
         return result[0][0]
 
     @staticmethod
-    def check_asset_exists(
-            asset_type: AssetType,
-            name: str,
-            symbol: str,
-    ) -> Optional[list[str]]:
+    def check_asset_exists(asset: AssetWithOracles) -> Optional[list[str]]:
```
LefterisJP, line 757: please also adjust the docstring

rotkehlchen/globaldb/handler.py
```diff
@@ -767,11 +763,17 @@ def check_asset_exists(
...
-            (asset_type.serialize_for_db(), name, symbol),
-        )
+        if isinstance(asset, EvmToken):
+            query = cursor.execute(
+                'SELECT identifier FROM evm_tokens WHERE identifier=?',
+                (asset.identifier,),
+            )
+        else:
+            query = cursor.execute(
+                'SELECT A.identifier from assets AS A JOIN common_asset_details as C '
```
LefterisJP, line 773: Hmmm shouldn't this also check by ID? I don't get why check name/symbol here. What if there is multiple assets with same name and symol and type?

yabirgb: Answered in discord. Is because for evm tokens we know 100% by the address/chain/token kind if the asset exists just by checking the identifier. For other assets is not possible to do so with the user provided by the user

rotkehlchen/api/rest.py
```diff
@@ -1354,11 +1355,15 @@ def add_user_asset(self, asset: AssetWithOracles) -> Response:
...
         # There is no good way to figure out if an asset already exists in the DB
         # Best approximation we can do is this.
-        identifiers = globaldb.check_asset_exists(
-            asset_type=asset.asset_type,
-            name=asset.name,
-            symbol=asset.symbol,
-        )
+        if isinstance(asset, EvmToken):
+            try:
+                asset.check_existence()  # for evm token we know the uniqueness of the identifier
```
LefterisJP, line 1360: nitpick

```suggestion
                asset.check_existence()  # the EVM tokens identifier is unique
```

LefterisJP (approved):
LGTM. Left a nitpick to fix at another place

PR #6307, opened 2023-06-22
passing submitUsageAnalytics in create account payload

Jaspreet-singh-1032:
## Fixes
- Fixes the issue while creating a new account on step 3 the system was not considering the `submit_usage_analytics` value selected by the user and was always saving it as true.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`6307` Fix issue create account always saving submit_usage_analytics as true.
 * :feature:`2822` In the asset graph, users will see another setting `Infer zero timed balances` which when activated will show the periods when users weren't holding the asset.
 * :feature:`-` Transactions changing the content hash of an ENS name will now be properly decoded.
 * :feature:`5255` Pnl report assets now have an etherscan link to make it easier to identify pool assets.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,7 +2,7 @@
 Changelog
 =========
 
-* :bug:`6307` Fix issue create account always saving submit_usage_analytics as true.
+* :bug:`-` Fix issue create account always saving submit_usage_analytics as true.
 * :feature:`2822` In the asset graph, users will see another setting `Infer zero timed balances` which when activated will show the periods when users weren't holding the asset.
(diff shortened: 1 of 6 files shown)


frontend/app/src/composables/api/session/users.ts
```diff
@@ -73,7 +73,7 @@ export const useUsersApi = () => {
         premiumApiKey: premiumSetup?.apiKey,
         premiumApiSecret: premiumSetup?.apiSecret,
         initialSettings: {
```
kelsos, line 75: if they have the same schema maybe it would make sense to just pas

frontend/app/tests/unit/specs/composables/user/account.spec.ts
```diff
@@ -82,7 +82,8 @@ describe('composables::user/account', () => {
       });
 
       await createNewAccount({
-        credentials: { username: 'test', password: '1234' }
+        credentials: { username: 'test', password: '1234' },
+        initialSettings: { submitUsageAnalytics: true }
```
kelsos, line 86: hey @Jaspreet-singh-1032 do you think it would be possible to change the default here to false and verify that false is properly passed through? since that was the bug?

kelsos (commented):
Thanks @Jaspreet-singh-1032 the fix works. I left you two minor comments let me know if you have any questions.

Also can you please also add an entry to the `docs/changelog.rst` file?

The entry should look like
```
* :bug:`-` Fix an issue...
```

LefterisJP (commented):
Hey @Jaspreet-singh-1032 since this is a frontend only PR I will leave you in the capable hands of @kelsos but can you also specify in `http:put:: /api/(version)/users` docs in api.rst that the default value of `submit_usage_analytics` is True if not given?

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`6307` Fix issue create account always saving submit_usage_analytics as true.
```
LefterisJP, line 5: Bug numbers are for issues, not PRs.

```suggestion
* :bug:`-` Fix issue create account always saving submit_usage_analytics as true.
```

kelsos (approved):
Thanks for the contribution @Jaspreet-singh-1032 lgtm 😃

PR #6309, opened 2023-06-22
Generalize "fix" for 5432 after new occurence

LefterisJP:
As I wrote in
https://github.com/rotki/rotki/issues/5432#issuecomment-1603396022 the issue popped up again in a different place.

This means that this driver bug is more generic and can hit us in many different places.

So I am moving the "fix" to the cursor itself.

We really need to find a proper fix for this


LefterisJP:
The only failing tests were uniswap and `rotkehlchen/tests/unit/test_eth2.py::test_scrape_validator_withdrawals` which was just a hiccup

PR #6312, opened 2023-06-23
Do not run background tasks until first snapshot is taken

LefterisJP:
This is so that we can keep the login process lean and not start any background tasks while we are still in DB upgrades, migration or asset updates which will end up killin them and restarting the app.


LefterisJP:
all green

PR #6314, opened 2023-06-23
Do not use sql vars in reset history events for v37->v38 upgrade

LefterisJP:
This avoids too many sql variables error for more than a given number of transactions


yabirgb (commented):
Something seems broken in the tests

LefterisJP:
> Something seems broken in the tests

Weird.

> rotkehlchen/tests/api/test_liquity.py::test_staking_stats

complaining about DB upgrade

PR #6316, opened 2023-06-23
Decode an 1inchv4 swap via uniswap v3

dimyG:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/oneinch/v4/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/oneinch/v4/decoder.py
@@ -1,0 +1,111 @@
+from typing import TYPE_CHECKING, Callable
+
+from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.chain.evm.decoding.oneinch.decoder import OneinchCommonDecoder
+from rotkehlchen.chain.evm.decoding.structures import (
+    DEFAULT_DECODING_OUTPUT,
+    DecoderContext,
+    DecodingOutput,
+)
+from rotkehlchen.chain.evm.decoding.utils import maybe_reshuffle_events
+from rotkehlchen.chain.evm.structures import EvmTxReceiptLog
+from rotkehlchen.types import ChecksumEvmAddress, EvmTransaction
+
+if TYPE_CHECKING:
+    from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
+    from rotkehlchen.user_messages import MessagesAggregator
+    from rotkehlchen.chain.evm.decoding.base import BaseDecoderTools
+    from rotkehlchen.accounting.structures.evm_event import EvmEvent
+
+from rotkehlchen.chain.ethereum.modules.balancer.v2.decoder import (
+    V2_SWAP as BALANCER_V2_SWAP_SIGNATURE,
+)
+from rotkehlchen.chain.ethereum.modules.uniswap.v2.decoder import (
+    SWAP_SIGNATURE as UNISWAP_V2_SWAP_SIGNATURE,
+)
+from rotkehlchen.chain.ethereum.modules.uniswap.v3.decoder import (
+    SWAP_SIGNATURE as UNISWAP_V3_SWAP_SIGNATURE,
+)
+
+from ..constants import CPT_ONEINCH_V4
+from .constants import ONEINCH_V4_MAINNET_ROUTER
+
+
+class Oneinchv4Decoder(OneinchCommonDecoder):
+
+    def __init__(
+            self,
+            evm_inquirer: 'EvmNodeInquirer',
+            base_tools: 'BaseDecoderTools',
+            msg_aggregator: 'MessagesAggregator',
+    ) -> None:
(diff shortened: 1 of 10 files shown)


rotkehlchen/chain/ethereum/modules/oneinch/v4/decoder.py
```diff
@@ -0,0 +1,85 @@
...
+        aggregator decoder. This function updates the spend and receive events so that the swap
+        is properly decoded. (The decoding is done with this post decoding rule and not from
+        addresses_to_decoders mapping because the 1inch v4 router is not in the addresses of the
+        logs of the transaction, and consequently it can't be mapped.)
+        """
+        for tx_log in all_logs:
+            if tx_log.topics[0] == UNISWAP_V3_SWAP_SIGNATURE:
+                sender = transaction.from_address
+                for event in decoded_events:
+                    if event.event_type == HistoryEventType.RECEIVE and event.location_label == sender:  # maybe amount and asset as well?  # noqa: E501
```
LefterisJP, line 63: why not match on amount and asset?

rotkehlchen/tests/unit/decoders/test_1inch.py
```diff
@@ -163,6 +168,65 @@ def test_1inchv2_swap_for_eth(database, ethereum_inquirer):
     assert expected_events == events
 
 
+@pytest.mark.parametrize('ethereum_accounts', [['0x312419eEC9C4632155904D9440dc1EeeafFBb280']])
+def test_1inchv4_swap_on_uniswapv3(database, ethereum_inquirer):
```
LefterisJP, line 172: I would need @yabirgb to also check this fix work for the thing he wanted to fix.

rotkehlchen/tests/unit/decoders/test_1inch.py
```diff
@@ -163,6 +168,65 @@ def test_1inchv2_swap_for_eth(database, ethereum_inquirer):
...
+            event_type=HistoryEventType.SPEND,
+            event_subtype=HistoryEventSubType.FEE,
+            asset=A_ETH,
+            balance=Balance(amount=FVal('0.005753512813468296')),
+            location_label=user_address,
+            notes='Burned 0.005753512813468296 ETH for gas',
+            counterparty=CPT_GAS,
+        ), EvmEvent(
+            tx_hash=tx_hash,
+            sequence_index=1,
```
LefterisJP, line 202: The order is wrong here. Swaps **always** have the out event first and then then in event. Otherwise accounting and other assumptions we make in the way we analyze events break down.

LefterisJP (commented):
LEft some comments. Also need 1inch v4 normal swaps to be decoded.

I think you just did this special case swap and not the normal swap events:
https://etherscan.io/txs?a=0x1111111254fb6c44bAC0beD2854e76F90643097d



Also  would need a changelog entry, mentioning that 1inch v4 events should now be decoded properly

rotkehlchen/tests/unit/decoders/test_1inch.py
yabirgb: yes, I got my transaction

rotkehlchen/chain/ethereum/modules/oneinch/v4/decoder.py
dimyG: The asset is not part of the Swap event and the amount of the Swap event is not the amount of the complete swap we want to show to the user. The trasnaction might have many smaller swap events and the total amount is the total swap amount we show to teh user (also I think there are some fees by 1inch). In all transactions that I tested, this approach works fine.

rotkehlchen/chain/ethereum/modules/oneinch/v4/decoder.py
```diff
@@ -0,0 +1,111 @@
...
+        It handles the case where a swap is done via 1inch v4 on any of the known pool contracts.
+        In that case there are 3 simple decoded events already created: Fee, spend and receive.
+        If Spend and receive events are not decoded by the swap decoder (for example from uniswap
+        v3 decoder), they should be decoded by this aggregator decoder.
+        The decoding is done with this post decoding rule and not from
+        addresses_to_decoders mapping because the 1inch v4 router is not in the addresses of the
+        logs of the transaction, and consequently it can't be extracted and mapped.
+        """
+        for tx_log in all_logs:
+            if tx_log.topics[0] in self.swapped_signatures:
```
LefterisJP, line 71: Hmm then this is not needed. Why are you runnign this rule? There is no point. Just to change counterparty and notes from uniswa/sushiswap to 1inch?

This makes no sense. Those transactions can be left decoded as they were. What you need to do is decode transactions that are not decoded properly at all and are using the aggregator part of 1inch.

LefterisJP (commented):
There is more that is not decoded properly. e.g. eth transfers: https://etherscan.io/tx/0x7097e7e9ef2b8bb096ed98950875b4512a833d41ceb3246903e06b61665cd5cd

LefterisJP (approved):
LGTM. If we find more discrepancies with 1inch v4 wecan fix em at that point

PR #6317, opened 2023-06-23
Filter ignored assets without using extra sql queries and variables

LefterisJP:


LefterisJP:
Would also probably be good to add a few unit tests for this.

Both with/without any ignored assets.
Both for exclude/include them in a query.

rotkehlchen/assets/utils.py
```diff
@@ -290,3 +292,15 @@ def symbol_to_evm_token(symbol: str) -> EvmToken:
...
     return maybe_asset.resolve_to_evm_token()
+
+
+class IgnoredAssetsHandling(SerializableEnumNameMixin):
+    NONE = auto()
+    EXCLUDE = auto()
+    SHOW_ONLY = auto()
+
+    def operator(self) -> Literal['IN', 'NOT IN']:
+        """Should only call this if you have narrowed it between exclude and show only"""
```
LefterisJP, line 303:
```suggestion
        """Caller should make sure this is narrowed between exclude and show only"""
```

LefterisJP:
Great. Seems all is green. Merging then.

PR #6319, opened 2023-06-25
Ruff to 0.0.275 and required changes

LefterisJP:
## [Activate and solve RUF013](https://github.com/rotki/rotki/commit/6060a671c61be365464dcb42b19955bc72c64e2e) 

Get rid of places where builting typing module attributes were
imported through rotkehlchen.types

## [Activate RUF012 and fix mutability in some class vars](https://github.com/rotki/rotki/commit/bf9303c48f29b43db9075badc424a74b7a7978ab) 

- NamedTuples having mutable default args is a bug waiting to happen,
so those were changed to dataclasses
- Various other changes that had to be done around that mutability concern


LefterisJP:
all green here. MERGEMENTATION

PR #6321, opened 2023-06-26
Properly decode ENS name transfers

LefterisJP:
This adds a small optional dependency in a subgraph for turning ENS token ID (which is the hash of the name) to the name. If there is any problem or if subgraph goes down we just don't show the specific name.

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/ens/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/ens/decoder.py
@@ -7,8 +7,11 @@ from eth_utils import to_checksum_address
 
 from rotkehlchen.accounting.structures.balance import Balance
 from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.assets.utils import TokenSeenAt, get_or_create_evm_token
 from rotkehlchen.chain.ethereum.abi import decode_event_data_abi_str
+from rotkehlchen.chain.ethereum.graph import Graph
 from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
+from rotkehlchen.chain.evm.decoding.constants import ERC20_OR_ERC721_TRANSFER
 from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface
 from rotkehlchen.chain.evm.decoding.structures import (
     DEFAULT_DECODING_OUTPUT,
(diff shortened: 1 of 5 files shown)


rotkehlchen/chain/ethereum/modules/ens/decoder.py
```diff
@@ -169,6 +174,63 @@ def _decode_name_renewed(self, context: DecoderContext) -> DecodingOutput:
...
+        if transfer_event is None:
+            log.error(f'Could not decode an ERC721 transfer for an ENS name transfer: {context.transaction.tx_hash.hex()}')  # noqa: E501
+            return DEFAULT_DECODING_OUTPUT
+
+        label_hash = hex(transfer_event.extra_data['token_id'])  # type: ignore[index]  # noqa: E501  # ERC721 transfer always has extra data
+        try:
+            result = self.graph.query(
+                querystr=f'query{{domains(first:1, where:{{labelhash:"{label_hash}"}}){{labelName}}}}')  # noqa: E501
+            name_to_show = result['domains'][0]['labelName'] + '.eth '
+        except (RemoteError, KeyError) as e:
```
yabirgb, line 204: Should this also catch index error?

rotkehlchen/chain/ethereum/modules/ens/decoder.py
```diff
@@ -169,6 +174,63 @@ def _decode_name_renewed(self, context: DecoderContext) -> DecodingOutput:
...
+            from_text = f'from {transfer_event.address} '
+            to_text = f'to {transfer_event.location_label}'
+        else:  # can only be ...
+            verb = 'Transfer'
+            if transfer_event.location_label != context.transaction.from_address:
+                from_text = f'from {transfer_event.location_label} '
+            to_text = f'to {to_address}'
+
+        transfer_event.counterparty = CPT_ENS
+        transfer_event.notes = f'{verb} ENS name {name_to_show}{from_text}{to_text}'
```
yabirgb, line 231: does the name to show have an extra whitespace at the end?

rotkehlchen/chain/ethereum/modules/ens/decoder.py
LefterisJP: yes good point

rotkehlchen/chain/ethereum/modules/ens/decoder.py
yabirgb: missed it since it has the '.eth ' after the graph response

PR #6322, opened 2023-06-26
Balancer uniswap removal

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/balancer/balancer.py
+++ b/rotkehlchen/chain/ethereum/modules/balancer/balancer.py
@@ -1,7 +1,5 @@
-import datetime
 import logging
 from collections import defaultdict
-from contextlib import suppress
 from operator import add, sub
 from typing import TYPE_CHECKING, Literal, Optional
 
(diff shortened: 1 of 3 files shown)


rotkehlchen/chain/ethereum/modules/balancer/balancer.py
```diff
@@ -815,33 +802,6 @@ def _get_protocol_balance_graph(
...
-            timestamp: Timestamp,
-    ) -> Price:
-        if token.has_oracle():
-            usd_price = query_usd_price_or_use_default(
-                asset=token,
-                time=timestamp,
-                default_value=ZERO,
-                location=str(Location.BALANCER),
-            )
-        else:
```
LefterisJP, line 830: what about removing this suppression part? Why is it okay?

LefterisJP (commented):
Just left a question. Also did not remove the tools I asked bit that's fine

rotkehlchen/chain/ethereum/modules/balancer/balancer.py
yabirgb: in that case it will use the defillama oracle or otherwise will default to 0. To query historical prices we would need archive logic to do it
the supress logic was there to go and find a price if coingecko/cryptocompare didn't have any

LefterisJP (approved):
LGTM

PR #6329, opened 2023-06-27
Add timestamp to yearn vaults

yabirgb:
Before we weren't adding the timestamp to the yearn vaults and this had consecuences in the calculation of the ROI. Now the tokens are edited/added with timestamp


LefterisJP (approved):
LGTM 👍

Like the test change from 140 to 2

PR #6332, opened 2023-06-27
Fix the test for querying makerdao vaults

dimyG:
* Only check the first vault so that if a new vault is added by the user the test still runs normally.
* Use vcr

Closes #(issue_number)

## Checklist


LefterisJP (commented):
1. If you add VCR you can remove the nightly requirement
2. Only 1 test was touched. The other 2 are stilll failing


https://github.com/rotki/rotki/actions/runs/5384920361/jobs/9773391244

LefterisJP (approved):
LGTM. Can't pull VCR repor from this connection but since PR tests passed I will merge.

PR #6340, opened 2023-06-28
fix: history events redecode

lukicenturi:
Closes #(issue_number)
1. Different behaviour per view. 
 	a. Certainly in all all non-history views limit to ethereum chain. 
2. A confirmation for sure as this is a very long-running task and we should explain to the user what it does. (Redecodes all events except the customized ones).
3. Remove liquity extra data from history events view, since it's not consistent


kelsos (approved):
lgtm

PR #6351, opened 2023-06-29
fix: contained card issue

lukicenturi:
Closes #(issue_number)

Since the `contained` cards are always inside the bottom-sheet, we can hardcode the height of `card-text` to `calc(90vh - total_height_of_other_elements)`

`useElementBounding` won't help, since we use the `top` of the `card-text` for the calculation, the value is not valid anymore (not the value that we expect) if it's broken from the start.


frontend/app/src/components/helper/Card.vue
```diff
@@ -26,27 +26,21 @@ const props = withDefaults(
...
+const { height: actionsHeight } = useElementBounding(actionsRef);
 
-  return {
-    height: `calc(100vh - ${diff}px)`
-  };
+const otherHeights = computed(() => {
+  const subTitleHeightVal = get(subTitleHeight) ?? 0;
+  const totalHeight =
+    (get(titleHeight) ?? 0) +
+    (subTitleHeightVal > 0 ? subTitleHeightVal - 16 : subTitleHeightVal) +
```
kelsos, line 41: Is 16 the padding?

frontend/app/src/components/helper/Card.vue
```diff
@@ -26,27 +26,21 @@ const props = withDefaults(
 );
 
 const { contained } = toRefs(props);
-const body = ref<HTMLDivElement | null>(null);
-const actions = ref<HTMLDivElement | null>(null);
-const top = ref(206);
+const titleRef = ref<HTMLDivElement | null>(null);
+const subTitleRef = ref<HTMLDivElement | null>(null);
+const actionsRef = ref<HTMLDivElement | null>(null);
```
kelsos, line 31: small nit maybe we can just do `ref<HTMLDivElement>()` which will result into `Ref<HTMLDivElement | undefined>` being the type and we get rid of the needed initialization.

Most of this initialization to null was a problem with normal options properties anyway before `Ref` where undefined was not reactive, which is not a problem anymore if with refs.

frontend/app/src/components/helper/Card.vue
lukicenturi: > Is 16 the padding?

So -16 is margin-top of the subtitle (if exist), so need to be substracted

lukicenturi: > if so do you think we can use `getComputedStyle` on the element to dynamically calculate it?

yes let me try

lukicenturi:
@kelsos change of plan, instead of find out what's the margin of the subtitle is, 
I calculate the `top part height` using `(bodyTop - cardTop)`

frontend/app/src/components/status/update/AssetUpdateMessage.vue
```diff
@@ -106,9 +106,8 @@ const { t } = useI18n();
         </v-col>
       </v-col>
     </v-row>
-    <template #options>
+    <template v-if="headless" #options>
```
lukicenturi, line 109: this is just small fix to improve the padding

kelsos (commented):
I tested it with the onboarding settings that behave in the same.
The code looks fine too.

Can we add a changelog entry @lukicenturi

kelsos (approved):

PR #6354, opened 2023-06-29
fix(packaging): exit 1 when notarization is unsuccessful

kelsos:
Closes #(issue_number)

## Checklist

## Note
This is to make sure that nightlies fail if we ever ran into the same situation when the terms have to be accepted before proceeding with the build

https://github.com/rotki/rotki/actions/runs/5400861329

There are two attempts here, on before and one after accepting the terms on the site to verify that the changes work as expected.

### DIFF
--- a/frontend/app/scripts/notarize.js
+++ b/frontend/app/scripts/notarize.js
@@ -1,7 +1,7 @@
 require('dotenv').config();
 const { notarize } = require('@electron/notarize');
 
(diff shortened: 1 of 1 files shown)


lukicenturi (commented):
LGTM!

lukicenturi (approved):
LGTM!

kelsos:
With this we will at least know that we have to accept the license before starting the build.

PR #6356, opened 2023-06-29
fix 1inchv4 eth weth swap

dimyG:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/oneinch/v4/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/oneinch/v4/decoder.py
@@ -68,9 +68,14 @@ class Oneinchv4Decoder(OneinchCommonDecoder):
         addresses_to_decoders mapping because the 1inch v4 router is not in the addresses of the
         logs of the transaction, and consequently it can't be extracted and mapped.
         """
-        tx_has_swap = False
-        for counter, tx_log in enumerate(all_logs):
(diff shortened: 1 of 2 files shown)


rotkehlchen/chain/ethereum/modules/oneinch/v4/decoder.py
```diff
@@ -80,8 +80,13 @@ def _handle_post_decoding(
                 )
                 self._decode_swapped(context)
                 tx_has_swap = True
-            elif counter == len(all_logs) - 1 and tx_has_swap is False and tx_log.topics[0] in [WETH_DEPOSIT_TOPIC, WETH_WITHDRAW_TOPIC]:  # noqa: E501
-                # If the last log is a WETH deposit or withdraw and no swap event was found,
+
+        if tx_has_swap:
+            return decoded_events
+
+        for tx_log in all_logs:
```
LefterisJP, line 87: Wait why are you doing a double iteration here? This does not sound correct

LefterisJP: Doing an iteration of all logs in a post-decoding handler is already the worst thing we can do. It decreases the performance of decoding across the entire app. But doing it twice is bad. Find out another way

LefterisJP (approved):
Lgtm

PR #6357, opened 2023-06-29
Run preview and no type-check

tewshi:
## Checklist


.github/workflows/task_e2e_tests.yml
```diff
@@ -93,4 +93,4 @@ jobs:
         if: always()
         with:
           name: videos-${{ runner.os }}
-          path: ./frontend/app/tests/e2e/videos
\ No newline at end of file
+          path: ./frontend/app/tests/e2e/videos
```
kelsos, line 96: Is the change here whitespace related?

tewshi: let me remove it, so, I made a change on the file and reverted, but prettier had already added the space

kelsos (approved):
Thanks @tewshi after some tries I managed to also run headless locally without errors.

I downgraded to an older version of cypress. I think we should give it a try upgrading once 12.17 is out.

I also enabled the experimental memory management and reduced the number of test in memory, (we might need to adjust this for development).

PR #6359, opened 2023-06-29
Add L1 fees to optimism gas calculations

0xGusMcCrae:
Addressed part of https://github.com/rotki/rotki/issues/6033

(Next PR will need to handlethe DB upgrade/migration)

Re-attempt of https://github.com/rotki/rotki/pull/6276, rebased on top of 1.29 release

Summary of changes:

- Create new OptimismTransaction class to handle additional l1_fee attribute on top of EvmTransaction.
- Updated optimism's decoder, node_inquirer, etherscan, transactions with function overrides for optimism-specific functionality
- Created new DB table optimism_transactions to handle l1_fee.
- Created DBOptimismTx, which inherits DBEvmTx in order to help query the evm_transactions table in conjunction with the optimism_transactions table.

The majority of changes were overrides of the base EVM functions in the optimism classes in order to accommodate new optimism-specific function names (i.e. using DBOptimismTx instead of DBEvmTx in a function body). Very little new code was written.

Notes:

- web3.py/RPC node inquiries are supported in this PR (in addition to etherscan). I'm not sure how I came to the conclusion that web3.py didn't support l1_fees, but it does.
- For now, rather than inheriting OptimismTransaction, I simply made OptimismTransaction a ~copy of EvmTransaction with the new l1_fee. For some reason, it kept giving me unexpected parameter l1_fee for EvmTransaction.new when I tried inheriting EvmTransaction.


0xGusMcCrae:
I'm still working on tracking down an error that's causing `/tests/unit/decoders/test_main.py` to fail. It appears that the `EvmEvent` being returned have assets as `AssetWithSymbol`, which causes an assertion error when comparing with `EvmEvent` that have regular `Asset`.

    self.asset=<Asset identifier:eip155:1/erc20:0xdAC17F958D2ee523a2206206994597C13D831ec7 name:Tether symbol:USDT>
vs 
            
    self.asset=<Asset identifier:eip155:1/erc20:0xdAC17F958D2ee523a2206206994597C13D831ec7>

Will try to get this sorted out ASAP. Everything else is working correctly though AFAIK

LefterisJP:
Hey @0xGusMcCrae thanks will look when you tell me it all works fine.

One thing though, please do not close and reopen PRs. Instead just force push on your branch. It will rewrite the PR completely.

0xGusMcCrae:
Ok @LefterisJP I've got it working and ready for review.

rotkehlchen/db/optimismtx.py
```diff
@@ -0,0 +1,107 @@
...
+        """Returns a list of optimism transactions optionally filtered by
+        the given filter query
+
+        This function can raise:
+        - pysqlcipher3.dbapi2.OperationalError if the SQL query fails due to invalid
+        filtering arguments.
+        """
+        query, bindings = filter_.prepare()
+        if has_premium:
+            query = 'SELECT DISTINCT evm_transactions.tx_hash, evm_transactions.chain_id, timestamp, block_number, from_address, to_address, value, gas, gas_price, gas_used, input_data, nonce, l1_fee FROM (SELECT * FROM evm_transactions WHERE chain_id = 10) AS evm_transactions LEFT JOIN optimism_transactions on evm_transactions.tx_hash = optimism_transactions.tx_hash ' + query  # noqa: E501
```
LefterisJP, line 74: you are duplicating queries here. This is not how you join. You are making an unneeded subquery. What you should write is something more like.

```
SELECT DISTINCT EVM.tx_hash, 10, EVM.timestamp, EVM.block_number, EVM.from_address, EVM.to_address, EMV.value, EVM.gas, EVM.gas_price, EVM.gas_used, EVM.input_data, EVM.nonce, OP.l1_fee FROM evm_transactions AS EVM LEFT JOIN optimism_transactions AS OP ON EVM.tx_hash=OP.tx_hash
```

rotkehlchen/db/schema.py
```diff
@@ -437,6 +437,16 @@
 );
 """
 
+DB_CREATE_OPTIMISM_TRANSACTIONS = """
+CREATE TABLE IF NOT EXISTS optimism_transactions (
+    tx_hash BLOB NOT NULL,
+    chain_id INTEGER NOT NULL,
```
LefterisJP, line 443: You do not need the chain id. It's optimism (value is 10)

```suggestion
```

LefterisJP (commented):
Hey @0xGusMcCrae 

So few things.

This PR as you also said duplicates a lot of code. It's 1.1k lines while it could just be a hundred.

So can you please:

1. Do all of the requested changes and simplify and reduce the code?
2. (optional) squash all commits to one. I can do that myself before merging
3. **DO NOT** close this PR again and open another one. I spent a lot of time reviewing it and as such do not want to lose the review. 

If for whatever reason you can't, let me know. I can do it myself pretty fast as what needs to be done is clear to me after doing this review, thanks to the work you have put in.

LefterisJP (commented):
Also from the failing tests make sure to fix `rotkehlchen/tests/db/test_db_upgrades.py::test_latest_upgrade_adds_remove_tables`

It fails since it does not take into account for the new table you added.

0xGusMcCrae:
I'll get to work on the fixes. Will try to get it done by the end of the weekend.

As you can probably tell, I'm not very experienced, so I really appreciate you taking the time to work with me and give me feedback on the proper way to handle these things.

LefterisJP:
> As you can probably tell, I'm not very experienced, so I really appreciate you taking the time to work with me and give me feedback on the proper way to handle these things.

My pleasure! No worries. Let's keep on trying!

rotkehlchen/db/filtering.py
```diff
@@ -209,6 +209,17 @@ def prepare(self) -> tuple[list[str], list[Any]]:
         return ['tx_hash=?'], [self.tx_hash]
 
 
+@dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=False)
+class DBOptimismTransactionHashFilter(DBFilter):
```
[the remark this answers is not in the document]
0xGusMcCrae: Should I edit the base chain/TransactionHash filter classes for these? I get an ambiguous column name error for the `WHERE (tx_hash = ?)  and (chain_id = ?)` part of the query if I don't edit the filters to specify whether it's the EVM or OP tx_hash/chain_id. 

    statement = 'SELECT DISTINCT EVM.tx_hash, 10, EVM.timestamp, EVM.block_number, EVM.from_address, 
    EVM.to_address, EVM.value, EVM.ga...T JOIN optimism_transactions AS OP ON EVM.tx_hash=OP.tx_hash WHERE 
    (tx_hash=?) AND (chain_id=?) ORDER BY timestamp ASC'
    bindings = ([b'\x06=E\x91\x0f)\xe0\x95JR\xae\xe3\x9f\xeb\xba\x9b\xe7\x84\xd4\x9a\xf7X\x8aY\r\xc2\xfd}\x15kFe', 
    10],)
    
        def execute(self, statement: str, *bindings: Sequence) -> 'DBCursor':
            if __debug__:
                logger.trace(f'EXECUTE {statement}')
            try:
    >           self._cursor.execute(statement, *bindings)
    E           pysqlcipher3.dbapi2.OperationalError: ambiguous column name: tx_hash

I can edit the base filters to always specify `evm_transaction.tx_hash = ?` and the same for chain_id and it'll work without creating the new Optimism___Filter classes, but I don't know if that's acceptable or if it will cause problems for any other chains (although it shouldn't since OP is currently the only chain that deviates from the base EVM query).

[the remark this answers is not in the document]
LefterisJP: yes to fix the ambiquity you should just specify the table

rotkehlchen/db/schema.py
0xGusMcCrae: pysqlcypher3/sqlite3 requires that all columns of the referenced table's primary key be in the foreign key, right? I can't eliminate the chain_id without getting a foreign key error:

      rotkehlchen/tests/unit/test_evm_tx_decoding.py:98: in _add_transactions_to_db
          dboptimismtx.add_evm_transactions(cursor, [transaction_opt], relevant_address=ethereum_accounts[0])  # noqa: E501
      rotkehlchen/db/optimismtx.py:45: in add_evm_transactions
          self.db.write_tuples(
      rotkehlchen/db/dbhandler.py:1537: in write_tuples
          write_cursor.executemany(query, tuples)
      _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
      
      self = <rotkehlchen.db.drivers.gevent.DBCursor object at 0x7fc9e631f520>
      statement = '\n            INSERT INTO optimism_transactions(\n              tx_hash,\n              l1_fee)\n            VALUES (?, ?)\n        '
      bindings = ([(HexBytes('0x063d45910f29e0954a52aee39febba9be784d49af7588a590dc2fd7d156b4665'), 455063072063200)],)
      
          def executemany(self, statement: str, *bindings: Sequence[Sequence]) -> 'DBCursor':
              if __debug__:
                  logger.trace(f'EXECUTEMANY {statement}')
      >       self._cursor.executemany(statement, *bindings)
      E       pysqlcipher3.dbapi2.OperationalError: foreign key mismatch - "optimism_transactions" referencing "evm_transactions"


This is the last thing I need to figure out before I'm ready to do some final linting and push changes

LefterisJP: > pysqlcypher3/sqlite3 requires that all columns of the referenced table's primary key be in the foreign key, right? I can't eliminate the chain_id without getting a foreign key error:
> 
> ```
>   rotkehlchen/tests/unit/test_evm_tx_decoding.py:98: in _add_transactions_to_db
>       dboptimismtx.add_evm_transactions(cursor, [transaction_opt], relevant_address=ethereum_accounts[0])  # noqa: E501
>   rotkehlchen/db/optimismtx.py:45: in add_evm_transactions
>       self.db.write_tuples(
>   rotkehlchen/db/dbhandler.py:1537: in write_tuples
>       write_cursor.executemany(query, tuples)
>   _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
>   
>   self = <rotkehlchen.db.drivers.gevent.DBCursor object at 0x7fc9e631f520>
>   statement = '\n            INSERT INTO optimism_transactions(\n              tx_hash,\n              l1_fee)\n            VALUES (?, ?)\n        '
>   bindings = ([(HexBytes('0x063d45910f29e0954a52aee39febba9be784d49af7588a590dc2fd7d156b4665'), 455063072063200)],)
>   
>       def executemany(self, statement: str, *bindings: Sequence[Sequence]) -> 'DBCursor':
>           if __debug__:
>               logger.trace(f'EXECUTEMANY {statement}')
>   >       self._cursor.executemany(statement, *bindings)
>   E       pysqlcipher3.dbapi2.OperationalError: foreign key mismatch - "optimism_transactions" referencing "evm_transactions"
> ```
> 
> This is the last thing I need to figure out before I'm ready to do some final linting and push changes

Hmm this is an ugly dependecy. But then make it a computed column so that it does not take space. Much like we do with the nfts table and the blockchain column.

```
    blockchain TEXT GENERATED ALWAYS AS ('ETH') VIRTUAL,
```

Note to use single quote in the sqlite query itself.

https://www.sqlite.org/gencol.html

rotkehlchen/db/dbhandler.py
```diff
@@ -1537,7 +1537,13 @@ def write_tuples(
             write_cursor.executemany(query, tuples)
             if relevant_address is not None:
                 # chain_id should always be the same so perform lookup out of the loop
-                blockchain = ChainID(tuples[0][1]).to_blockchain().value
+                try:
+                    blockchain = ChainID(tuples[0][1]).to_blockchain().value
```
LefterisJP, line 1541: This is hacky here. Plus we may need to have similar logic for other chains (maybe arbitrum?) and then this logic will break.

Any such transformations should happen at the caller at dbevmtx.py.

rotkehlchen/db/dbhandler.py
```diff
@@ -1552,7 +1558,13 @@ def write_tuples(
                 try:
                     write_cursor.execute(query, entry)
                     if relevant_address is not None:
-                        blockchain = ChainID(entry[1]).to_blockchain().value
+                        try:
+                            blockchain = ChainID(entry[1]).to_blockchain().value
```
LefterisJP, line 1562: same comment as above

rotkehlchen/db/dbhandler.py
```diff
@@ -1567,7 +1579,13 @@ def write_tuples(
                         # Also if we have transactions of one account sending to the
                         # other and both accounts are being tracked.
                         if relevant_address is not None:
-                            blockchain = ChainID(entry[1]).to_blockchain().value
+                            try:
+                                blockchain = ChainID(entry[1]).to_blockchain().value
```
LefterisJP, line 1583: same as above

rotkehlchen/db/evmtx.py
```diff
@@ -175,29 +175,16 @@ def get_evm_transactions(
         """
         query, bindings = filter_.prepare()
         if has_premium:
-            query = 'SELECT DISTINCT evm_transactions.tx_hash, evm_transactions.chain_id, timestamp, block_number, from_address, to_address, value, gas, gas_price, gas_used, input_data, nonce FROM evm_transactions ' + query  # noqa: E501
+            query = self._get_evm_transaction_query(query, has_premium)
```
LefterisJP, line 178: you are duplicating the if check here. Both before the function and then you do it also in the function.

The way around this is easy. Rename to `_form_evm_transaction_dbquery` and make it return a tuple of (str, list[tuple]) which means it will requrns the query string and its bindings.

Then you can avoid the external if.

rotkehlchen/db/evmtx.py
```diff
@@ -594,3 +581,27 @@ def get_or_create_genesis_transaction(
                     relevant_address=account,
                 )
         return tx
+
+    def _get_evm_transaction_query(self, query: str, has_premium: bool) -> str:
+        """Return queries for the evm_transaction database table"""
```
LefterisJP, line 586: Make the changes I asked above to this function, along with adjusting the docstring after the changes.

rotkehlchen/db/evmtx.py
```diff
@@ -594,3 +581,27 @@ def get_or_create_genesis_transaction(
...
+        if has_premium:
+            return 'SELECT DISTINCT evm_transactions.tx_hash, evm_transactions.chain_id, timestamp, block_number, from_address, to_address, value, gas, gas_price, gas_used, input_data, nonce FROM evm_transactions ' + query  # noqa: E501
+        else:
+            return 'SELECT DISTINCT evm_transactions.tx_hash, evm_transactions.chain_id, timestamp, block_number, from_address, to_address, value, gas, gas_price, gas_used, input_data, nonce FROM (SELECT * from evm_transactions ORDER BY timestamp DESC LIMIT ?) AS evm_transactions ' + query  # noqa: E501
+
+    def _build_evm_transaction(self, result: Any) -> EvmTransaction:
```
LefterisJP, line 592: Do not use the `Any` type, it's effectively no type.

```suggestion
    def _build_evm_transaction(self, result: tuple[Union[str, int]]) -> EvmTransaction:
```

rotkehlchen/db/optimismtx.py
```diff
@@ -0,0 +1,74 @@
...
+              tx_hash,
+              l1_fee)
+            VALUES (?, ?)
+        """
+        self.db.write_tuples(
+            write_cursor=write_cursor,
+            tuple_type='evm_transaction',
+            query=query,
+            tuples=tx_tuples,
+            relevant_address=relevant_address,
```
LefterisJP, line 50: Remove relevant address from here as the logic concerning it should already be handled when you add the main data.
The only point of using the `write_tuples` function and not doing just `executemany` directly is to repeat with a for loop if one transaction already exists in the DB.

rotkehlchen/db/optimismtx.py
```diff
@@ -0,0 +1,74 @@
...
+        if has_premium:
+            return 'SELECT DISTINCT evm_transactions.tx_hash, evm_transactions.chain_id, evm_transactions.timestamp, evm_transactions.block_number, evm_transactions.from_address, evm_transactions.to_address, evm_transactions.value, evm_transactions.gas, evm_transactions.gas_price, evm_transactions.gas_used, evm_transactions.input_data, evm_transactions.nonce, OP.l1_fee FROM evm_transactions LEFT JOIN optimism_transactions AS OP ON evm_transactions.tx_hash=OP.tx_hash ' + query  # noqa: E501
+        else:
+            return 'SELECT DISTINCT evm_transactions.tx_hash, evm_transactions.chain_id, evm_transactions.timestamp, evm_transactions.block_number, evm_transactions.from_address, evm_transactions.to_address, evm_transactions.value, evm_transactions.gas, evm_transactions.gas_price, evm_transactions.gas_used, evm_transactions.input_data, evm_transactions.nonce, OP.l1_fee FROM (SELECT * FROM evm_transactions ORDER BY timestamp DESC LIMIT ?) AS evm_transactions LEFT JOIN optimism_transactions AS OP ON evm_transactions.tx_hash=OP.tx_hash ' + query  # noqa: E501
+
+    def _build_evm_transaction(self, result: Any) -> OptimismTransaction:
```
LefterisJP, line 59: same here for the type of the function as mentioned before

rotkehlchen/db/schema.py
```diff
@@ -437,6 +437,16 @@
 );
 """
 
+DB_CREATE_OPTIMISM_TRANSACTIONS = """
```
LefterisJP, line 440: Can you please also add a step in the DB upgrade where new tables are created and add this exact same table creation there?

Or just add a TODO so that I or me do it later.

WE will also need a data migration to update all existing optimism transactions to conform to the new schema. As it can't happen during the upgrade since it's too many network calls.

But that should really be a separate PR.

rotkehlchen/serialization/deserialize.py
```diff
@@ -533,6 +535,17 @@ def deserialize_evm_transaction(
     ...
 
 
+@overload
+def deserialize_evm_transaction(  # type: ignore[misc]
+        data: dict[str, Any],
+        internal: Literal[False],
+        chain_id: ChainID,
```
LefterisJP, line 542: This should also be specified

```suggestion
        chain_id: Literal[ChainID.OPTIMISM],
```

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1661,7 +1661,7 @@ def test_latest_upgrade_adds_remove_tables(user_data_dir):
     missing_views = views_before - views_after_upgrade
     assert missing_tables == removed_tables
     assert missing_views == removed_views
-    assert tables_after_creation - tables_after_upgrade == set()
+    assert tables_after_creation - tables_after_upgrade == {'optimism_transactions'}
```
LefterisJP, line 1664: this should not be like this. Which is why I mentioned to add the addition at the upgrade

LefterisJP (commented):
Hey @0xGusMcCrae much better now. Still left some comments on things that need to be done, and some TODOs to add for later/other PRs.

0xGusMcCrae:
Got it. These shouldn't be too difficult. I'll try to get it done later today.

rotkehlchen/db/dbhandler.py
0xGusMcCrae: The problem I was trying to solve with these actually went away once I took out the relevant_address from optimismtx's call to write_tuples, so I simply restored to the original

rotkehlchen/db/schema.py
0xGusMcCrae: I wasn't 100% sure what I needed to do here, I just added the TODO

LefterisJP: Oh it's simple. So these commands in db/schema.py only run for the new creation of the DB. Which would happen for a new user. For all existing users, when there is any sort of modification of the schema, and a new DB table is one such modification, it needs to be added to the DB upgrade.

The latest DB upgrade is this: https://github.com/rotki/rotki/blob/develop/rotkehlchen/db/upgrades/v38_v39.py . It will be included in our next release from the develop branch (1.30.0).
And you should just add a new step with a new function to create new tables and insert this one you made there.

But just the TODO is fine also since I am touching all this code in my PR anyway and would need to change it again so it's fine.

But just letting you know so you know what I meant. One such create_new_tables step is here for the v36->v37 DB upgrade: https://github.com/rotki/rotki/blob/63d49f2501114376d5a9f727b083b280fa092f4e/rotkehlchen/db/upgrades/v36_v37.py#L166

rotkehlchen/chain/optimism/decoding/decoder.py
```diff
@@ -43,6 +47,10 @@ def __init__(
                 address_is_exchange_fn=self._address_is_exchange,
             ),
         )
+        self.dbevmtx = DBOptimismTx(database)
```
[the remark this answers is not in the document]
LefterisJP: That is because the superclass `EVMTransactionDecoder` already initializes this here: https://github.com/rotki/rotki/blob/63d49f2501114376d5a9f727b083b280fa092f4e/rotkehlchen/chain/evm/decoding/decoder.py#L148

But for the optimism decoder you need a different tx class. You are overriding it here. It's not totally wrong, but it would be cleaner to parametrize the constructor to initialize the different class.

LefterisJP (approved):
Okay so @0xGusMcCrae I am going to approve and merge this in develop with some TODOs left.

1. ~~https://github.com/rotki/rotki/pull/6359#discussion_r1254059689 do this parametrization at the constructor~~ DONE
2. https://github.com/rotki/rotki/pull/6359#discussion_r1252850465 save the receipt in the DB if it's retrieved here. -- This may be harder to do. May skip. Not sure would have to investigate
3. ~~Add the new table to the DB upgrade: https://github.com/rotki/rotki/pull/6359#discussion_r1252891305 and adjust the test here: https://github.com/rotki/rotki/pull/6359#discussion_r1252894045~~ DONE
4. ~~Adjust types in _build_evm_transaction. Too many type ignores, probably can improve.~~
5. Migrate the existing optimism data in a migration. This is gonna be fun. 
6. For when arbitrum comes (which is planned for this release, so maybe do it now) makes sense to abstract this a bit more. And rename the table and class to RollupEvmTransaction. And adjust code accordingly.

But will merge since it's all green and the rest of the stuff can happen in subsequent PRs!

PR #6367, opened 2023-07-02
docs: removes conda setup from installation guide

kelsos:
Closes #(issue_number)

## Checklist

## Reasoning
It isn't obvious to the people that want to contribute and since no one on the team uses it, it is untested.

The guide should be based on a setup we actually use, otherwise, we can't help people set things up.


tewshi (approved):
lgtm

PR #6370, opened 2023-07-03
Introduce v39 userdb upgrade and reduce some history event identifier size

LefterisJP:
Also add VACUUM as last part of the user DB upgrade


LefterisJP (commented):
All tests seemed to pass now. So I am gonna merge this one now as is.

PR #6377, opened 2023-07-05
Reduce DB size by using integer primary key instead of composite hash + chain id

### LINKED ISSUE #6372: Save DB size by making EVM transactions primary key to be an integer identifier
## Abstract

The DB tables that deal with EVM transactions duplicate the primary key to every table to connect them.

```python
DB_CREATE_EVM_TRANSACTIONS = """
CREATE TABLE IF NOT EXISTS evm_transactions (
    tx_hash BLOB NOT NULL,
    chain_id INTEGER NOT NULL,
    timestamp INTEGER NOT NULL,
    block_number INTEGER NOT NULL,
    from_address TEXT NOT NULL,
    to_address TEXT,
    value TEXT NOT NULL,
    gas TEXT NOT NULL,
    gas_price TEXT NOT NULL,
    gas_used TEXT NOT NULL,
    input_data BLOB NOT NULL,
    nonce INTEGER NOT NULL,
    PRIMARY KEY(tx_hash, chain_id)
);
"""

# from/to address/value is also in the primary key of the internal transactions since
# trace_id, which is returned by etherscan does not guarantee uniqueness. Example:
# https://api.etherscan.io/api?module=account&action=txlistinternal&sort=asc&startBlock=16779092&endBlock=16779092
DB_CREATE_EVM_INTERNAL_TRANSACTIONS = """
CREATE TABLE IF NOT EXISTS evm_internal_transactions (
    parent_tx_hash BLOB NOT NULL,
    chain_id INTEGER NOT NULL,
    trace_id INTEGER NOT NULL,
    from_address TEXT NOT NULL,
    to_address TEXT,
    value TEXT NOT NULL,
    FOREIGN KEY(parent_tx_hash, chain_id) REFERENCES evm_transactions(tx_hash, chain_id) ON DELETE CASCADE ON UPDATE CASCADE,
    PRIMARY KEY(parent_tx_hash, chain_id, trace_id, from_address, to_address, value)
);
"""  # noqa: E501

DB_CREATE_EVMTX_RECEIPTS = """
CREATE TABLE IF NOT EXISTS evmtx_receipts (
    tx_hash BLOB NOT NULL,
    chain_id INTEGER NOT NULL,
    contract_address TEXT, /* can be null */
    status INTEGER NOT NULL CHECK (status IN (0, 1)),
    type INTEGER NOT NULL,
    FOREIGN KEY(tx_hash, chain_id) REFERENCES evm_transactions(tx_hash, chain_id) ON DELETE CASCADE ON UPDATE CASCADE,
    PRIMARY KEY(tx_hash, chain_id)
);
"""  # noqa: E501

DB_CREATE_EVMTX_RECEIPT_LOGS = """
CREATE TABLE IF NOT EXISTS evmtx_receipt_logs (
    tx_hash BLOB NOT NULL,
    chain_id INTEGER NOT NULL,
    log_index INTEGER NOT NULL,
    data BLOB NOT NULL,
    address TEXT NOT NULL,
    removed INTEGER NOT NULL CHECK (removed IN (0, 1)),
    FOREIGN KEY(tx_hash, chain_id) REFERENCES evmtx_receipts(tx_hash, chain_id) ON DELETE CASCADE ON UPDATE CASCADE,
    PRIMARY KEY(tx_hash, chain_id, log_index)
);
"""  # noqa: E501

DB_CREATE_EVMTX_RECEIPT_LOG_TOPICS = """
CREATE TABLE IF NOT EXISTS evmtx_receipt_log_topics (
    tx_hash BLOB NOT NULL,
    chain_id INTEGER NOT NULL,
    log_index INTEGER NOT NULL,
    topic BLOB NOT NULL,
    topic_index INTEGER NOT NULL,
    FOREIGN KEY(tx_hash, chain_id, log_index) REFERENCES evmtx_receipt_logs(tx_hash, chain_id, log_index) ON DELETE CASCADE ON UPDATE CASCADE,
    PRIMARY KEY(tx_hash, chain_id, log_index, topic_index)
);
"""  # noqa: E501
```

The DB tables that deal with EVM transactions duplicate the primary key to every table. The primary key is composite, which makes it worse. We have 5 tables. That is 32 + 4 bytes (at best case) duplicated across 5 tables. So 36 * 5 = 180 bytes per entry versus 4 * 5 = 20 bytes per entry.

And that gets worse in reality since each receipt can have multiple logs (10-100) and each log has 1-4 topics. So this blows up.

This is from one of my accounts:

```
sqlite> SELECT COUNT(*) from evm_transactions;
3274
sqlite> SELECT COUNT(*) from evm_internal_transactions;
198
sqlite> SELECT COUNT(*) from evmtx_receipts;
3274
sqlite> SELECT COUNT(*) from evmtx_receipt_logs;
172026
sqlite> SELECT COUNT(*) from evmtx_receipt_log_topics ;
510256
```

So the duplicated data for linking the tables are 198 internal_transaction, 3274 receipts, 176026 logs, 510256 log topics.

3274 + 198 + 172026 + 510256 = 685754

The difference for each entry calculated above is 180-20 = 160

So in total for this DB 685754 * 160 = 109720640 bytes.
Which should be ...  **109.72064 MB**. 

This is A LOT. For comparison the DB table is 142 MB after `VACUUM`. Wonder if I made a mistake in the calculation.

## Motivation

The DB size is growing exponentially, especially the more EVM chains we add. We need to find ways to reduce it, but also not introduce too much additional complexity.

The above calculations though seem to point to an extreme gain in size, so I think doing this is a no-brainer.

## Challenges

The challenge here is that some queries may become more complex, needing a join to the parent table to get tx_hash and chain id. But that should be fine. Also we will need to adjust the schema to explain the combination of tx_hash and chain_id should be unique though they are not the primary key.

## Specification

- Change the schema of `evm_transaction` to have an identifier integer primary key. Also make the combination of chain_id and tx_hash need to be uniqe.
- All other 4 table schemas should just foreign key to evm_transaction via identifier and no longer duplicate tx_hash and chain id.
- Adjust the code all around to adapt it for changes, especially in sqlite queries.

LefterisJP:
Closes #6372 

Still needs some work on adjustments to the code, but just wanna run the tests.


LefterisJP:
Even when this is ready should try to wait with merging it until https://github.com/rotki/rotki/pull/6359 is ready, so that I can rebase my work on top and not force the external dev rebase their work on these changes.

LefterisJP:
Hmm ... all ... green? Maybe merge? Still got multiple things that need to work on top of this, but will make in other PRs.

PR #6380, opened 2023-07-06
First optimism L1 fee PR followups

LefterisJP:
Some follows up from https://github.com/rotki/rotki/pull/6359#pullrequestreview-1516015137

### DIFF
--- a/rotkehlchen/db/evmtx.py
+++ b/rotkehlchen/db/evmtx.py
@@ -1,5 +1,5 @@
 import logging
-from typing import TYPE_CHECKING, Any, Optional, Union, get_args
+from typing import TYPE_CHECKING, Any, Optional, get_args
 
 from rotkehlchen.chain.ethereum.constants import ETHEREUM_GENESIS
 from rotkehlchen.chain.evm.constants import GENESIS_HASH, ZERO_ADDRESS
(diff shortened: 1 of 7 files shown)


rotkehlchen/db/evmtx.py
```diff
@@ -585,25 +585,25 @@ def _form_evm_transaction_dbquery(self, query: str, bindings: list[Any], has_pre
...
-                [FREE_ETH_TX_LIMIT] + bindings,
-            )
+        # else
+        return (
+            'SELECT DISTINCT evm_transactions.tx_hash, evm_transactions.chain_id, timestamp, block_number, from_address, to_address, value, gas, gas_price, gas_used, input_data, nonce FROM (SELECT * from evm_transactions ORDER BY timestamp DESC LIMIT ?) AS evm_transactions ' + query,  # noqa: E501
+            [FREE_ETH_TX_LIMIT] + bindings,
+        )
 
-    def _build_evm_transaction(self, result: tuple[Union[str, int]]) -> EvmTransaction:
+    def _build_evm_transaction(self, result: tuple[Any, ...]) -> EvmTransaction:
```
yabirgb, line 594: Since we know the fields... isn't it worth to have the tuple type with the type for each field and use it here?

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1644,6 +1645,7 @@ def test_upgrade_db_38_to_39(user_data_dir):  # pylint: disable=unused-argument
     )
     cursor = db.conn.cursor()
 
+    assert cursor.execute("SELECT COUNT(*) FROM sqlite_master WHERE type='table' AND name='optimism_transactions';").fetchone()[0] == 1  # noqa: E501
     assert cursor.execute('SELECT COUNT(*) FROM history_events').fetchone()[0] == events_num_before
```
yabirgb, line 1649: I believe the new table is missing in `test_latest_upgrade_adds_remove_tables`

rotkehlchen/db/evmtx.py
LefterisJP: > Since we know the fields... isn't it worth to have the tuple type with the type for each field and use it here?

no it really isn't. Would need to create a new special tuple type for this just to typecheck this one call. And it's meaningless since cursor.execute() returns tuple[Any, ...]

PR #6386, opened 2023-07-08
Add transaction decoders for Safe multisig actions

LefterisJP:
This commit also adds a new kind of decoding action. One that depends on input data. Depending on the 4bytes signature of the input data specific topics may be looked for matching during decoding.


rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -397,8 +434,29 @@ def _decode_transaction(
...
+                    action_items=action_items,
+                )
+                try:  # run specific decoder if the 4bytes signature + topic match
+                    result = input_rule(context)
+                except (DeserializationError, ConversionError, UnknownAsset) as e:
+                    log.error(f'Decoding log {tx_log} of {transaction} failed due to {e!s}')
+                    result = DEFAULT_DECODING_OUTPUT
+
+                if result.event:
+                    events.append(result.event)
```
yabirgb, line 458: should thsi do a continue? otherwise the decoding by address logic is executed

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -397,8 +434,29 @@ def _decode_transaction(
...
+                    tx_log=tx_log,
+                    transaction=transaction,
+                    decoded_events=events,
+                    all_logs=tx_receipt.logs,
+                    action_items=action_items,
+                )
+                try:  # run specific decoder if the 4bytes signature + topic match
+                    result = input_rule(context)
+                except (DeserializationError, ConversionError, UnknownAsset) as e:
+                    log.error(f'Decoding log {tx_log} of {transaction} failed due to {e!s}')
```
yabirgb, line 454: maybe is worth adding to this message that is using the input data rules?

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -397,8 +434,29 @@ def _decode_transaction(
...
+                    result = DEFAULT_DECODING_OUTPUT
+
+                if result.event:
+                    events.append(result.event)
+
             decoding_output = self.decode_by_address_rules(tx_log, transaction, events, tx_receipt.logs, action_items)  # noqa: E501
```
yabirgb, line 460: Maybe not in this PR but `decode_by_address_rules` builds the same `DecoderContext` so maybe is worth to change the arguments here to be only one of type `DecoderContext` and then it can be reused

yabirgb (approved):

rotkehlchen/chain/evm/decoding/decoder.py
LefterisJP: hmm yes I think it should continue here

rotkehlchen/chain/evm/decoding/decoder.py
LefterisJP: thought about that but not sure if it's worth doing. Will see.

rotkehlchen/chain/evm/decoding/decoder.py
LefterisJP: sure

rotkehlchen/chain/evm/decoding/decoder.py
LefterisJP: did it. It was trivial and reduced duplication

LefterisJP:
All green and did the requested changes. Merging.

PR #6387, opened 2023-07-09
Upgrade ruff to 0.0277 and fix PERF401 and PERF402

LefterisJP:


LefterisJP:
Green finally. This was harder to do without an autofix than I expected. Also error prone for the one nested comprehension.

PR #6389, opened 2023-07-11
Add migration 11, to populate optimism fees DB table

LefterisJP:


rotkehlchen/data_migrations/migrations/migration_11.py
```diff
@@ -0,0 +1,45 @@
...
+    """Since we now need 1 extra column per optimism transaction we need to repull all
+    optimism transactions to store the l1 gas fee"""
+    with rotki.data.db.conn.read_ctx() as cursor:
+        cursor.execute('SELECT identifier, tx_hash FROM evm_transactions WHERE chain_id=10')
+        txhash_to_id = {txhash: identifier for identifier, txhash in cursor}
+
+    progress_handler.set_total_steps(len(txhash_to_id))
+
+    db_tuples = []
+    optimism_manager = rotki.chains_aggregator.get_evm_manager(ChainID.OPTIMISM)
```
yabirgb, line 25: It happened in previous releases that if you pull a database that needs to be updated the `chains_aggregator` attribute at rotki is not set yet when the update is executed

rotkehlchen/data_migrations/migrations/migration_11.py
```diff
@@ -0,0 +1,45 @@
...
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+def _get_optimism_transaction_fees(rotki: 'Rotkehlchen', progress_handler: 'MigrationProgressHandler') -> None:  # noqa: E501
+    """Since we now need 1 extra column per optimism transaction we need to repull all
+    optimism transactions to store the l1 gas fee"""
+    with rotki.data.db.conn.read_ctx() as cursor:
+        cursor.execute('SELECT identifier, tx_hash FROM evm_transactions WHERE chain_id=10')
+        txhash_to_id = {txhash: identifier for identifier, txhash in cursor}
```
yabirgb, line 20: this can be rewritten as 

```suggestion
        cursor.execute('SELECT tx_hash, identifier FROM evm_transactions WHERE chain_id=10')
        txhash_to_id = dict(cursor)
```

if I'm not wrong

rotkehlchen/data_migrations/migrations/migration_11.py
```diff
@@ -0,0 +1,45 @@
...
+
+    progress_handler.set_total_steps(len(txhash_to_id))
+
+    db_tuples = []
+    optimism_manager = rotki.chains_aggregator.get_evm_manager(ChainID.OPTIMISM)
+    assert optimism_manager, 'should exist at this point'
+    for txhash, txid in txhash_to_id.items():
+        progress_handler.new_step(f'Fetching optimism transaction {txhash.hex()}')
+        transaction, _ = optimism_manager.node_inquirer.get_transaction_by_hash(tx_hash=txhash)
```
yabirgb, line 29: This can raise remoteError. It shouldn't but I believe is better to protect against any possible remote error

yabirgb (commented):
Some considerations

rotkehlchen/data_migrations/migrations/migration_11.py
LefterisJP: > It happened in previous releases that if you pull a database that needs to be updated the `chains_aggregator` attribute at rotki is not set yet when the update is executed

What did we do then? I don't see this in any other migration that's left.

yabirgb (approved):
LGTM

PR #6390, opened 2023-07-11
Cached settings singleton class

dimyG:
* A singleton class that stores settings values in memory is implemented.
* Settings can be read from this class instead of hitting the database.
* This is especially useful in places where the dbhandler is not available.

Closes #(6041)

## Checklist


rotkehlchen/chain/ethereum/modules/yearn/utils.py
```diff
@@ -47,7 +47,7 @@ def query_yearn_vaults(db: 'DBHandler', ethereum_inquirer: 'EthereumInquirer') -
     - RemoteError
     """
     try:
-        response = requests.get(YEARN_OLD_API, timeout=DEFAULT_TIMEOUT_TUPLE)
+        response = requests.get(YEARN_OLD_API, timeout=CachedSettings().get_timeout_tuple())
```
LefterisJP, line 50: Out of all these I think the other thing we need is to have a default version of `requests.get()` overriden by using ourdefault timeout ... once. Not having to import this everywhere.

rotkehlchen/chain/substrate/manager.py
```diff
@@ -26,6 +25,7 @@
 from rotkehlchen.user_messages import MessagesAggregator
 from rotkehlchen.utils.serialization import jsonloads_dict
 
+from ...db.settings import CachedSettings
```
LefterisJP, line 28: don't do relative imports like this unless it's meant to be part of the same module/directory.

```suggestion
from rotkehlchen.db.settings import CachedSettings
```

rotkehlchen/db/drivers/gevent.py
```diff
@@ -370,6 +371,7 @@ def write_ctx(self, commit_ts: bool = False) -> Generator['DBCursor', None, None
                         'INSERT OR REPLACE INTO settings(name, value) VALUES(?, ?)',
                         ('last_write_ts', str(ts_now())),
                     )
+                    settings.CachedSettings().update_entry('last_write_ts', ts_now())
```
LefterisJP, line 374: this is incorrect. You end up using different timestamp here and different in the line above. You should save it in a variable first and use it in both places.

Also ... instead of doing that, replace with a comment, and say that this particular settings is not cached knowingly.

This is a critical section code. Adding even one more function call can have very ugly and detrimental effects in the entire codebase as everything calls this .

rotkehlchen/db/settings.py
```diff
@@ -306,3 +317,96 @@ def serialize_db_setting(
         else:
             value = [x.serialize() for x in value]
     return value
+
+
+@dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=False)
+class DBSettingsUnfrozen:
```
LefterisJP, line 323: No no no. Why add this? Just use one class and have it unfrozen. This is silly duplication.

rotkehlchen/db/settings.py
```diff
@@ -306,3 +317,96 @@ def serialize_db_setting(
...
+            return CachedSettings.__instance
+
+        CachedSettings.__instance = super().__new__(cls)
+        return CachedSettings.__instance
+
+    def reset(self) -> None:
+        self._settings = DBSettingsUnfrozen()
+
+    def update_entry(self, attr: str, value: Any) -> None:
+        if value is None:
```
LefterisJP, line 389: This is very bad python logic. You are throwing the type checker out of the window with the `Any` type and instead do a dynamic if check every time this is called. Use proper types here and lose the dynamic checks.

rotkehlchen/db/settings.py
```diff
@@ -306,3 +317,96 @@ def serialize_db_setting(
...
+            return
+        setattr(self._settings, attr, value)
+
+    def update_entries(self, settings: ModifiableDBSettings) -> None:
+        for attr in settings._fields:
+            self.update_entry(attr, getattr(settings, attr))
+
+    def get_entry(self, attr: str) -> Any:
+        if attr not in [f.name for f in fields(self._settings)]:
```
LefterisJP, line 398: same here. Use types. These are not user errors. These are development errors. They should be found by the type checker.

We don't check arguments of functions with if checks dynamically. That is what linters are for.

rotkehlchen/db/settings.py
```diff
@@ -306,3 +317,96 @@ def serialize_db_setting(
...
+            raise AttributeError(f'Attribute {attr} not found in cached settings dataclass.')
+        return getattr(self._settings, attr)
+
+    def get_settings(self) -> Optional[DBSettingsUnfrozen]:
+        return self._settings
+
+    # commonly used settings with their own get function
+    def get_timeout_tuple(self) -> tuple[int, int]:
+        conx_timeout = self.get_entry('connect_timeout')
```
LefterisJP, line 407:
```suggestion
        conn_timeout = self.get_entry('connect_timeout')
```

rotkehlchen/db/drivers/gevent.py
```diff
@@ -15,6 +15,7 @@
 import gevent
 from pysqlcipher3 import dbapi2 as sqlcipher
 
+from rotkehlchen.db import settings
```
LefterisJP, line 18: Though I asked you to remove this, never import stuff like this. You are importing the entire module here. You only needed CachedSettings.

rotkehlchen/db/updates.py
```diff
@@ -7,7 +7,7 @@
 import requests
 
 from rotkehlchen.assets.spam_assets import update_spam_assets
-from rotkehlchen.constants.timing import DEFAULT_TIMEOUT_TUPLE
+from rotkehlchen.db import settings
```
LefterisJP, line 10: Don't do that. Where in the code have you ever seen us import entire modules like this. Import only what you need.
```suggestion
from rotkehlchen.db.settings import CachedSettings
```

rotkehlchen/db/updates.py
```diff
@@ -83,7 +83,7 @@ def _get_remote_info_json(self) -> dict[str, Any]:
         """
         url = f'https://raw.githubusercontent.com/rotki/data/{self.branch}/updates/info.json'
         try:
-            response = requests.get(url=url, timeout=DEFAULT_TIMEOUT_TUPLE)
+            response = requests.get(url=url, timeout=settings.CachedSettings().get_timeout_tuple())
```
LefterisJP, line 86:
```suggestion
            response = requests.get(url=url, timeout=CachedSettings().get_timeout_tuple())
```

rotkehlchen/externalapis/etherscan.py
```diff
@@ -627,7 +623,7 @@ def get_logs(
             module='logs',
             action='getLogs',
             options=options,
-            timeout=(DEFAULT_CONNECT_TIMEOUT, DEFAULT_READ_TIMEOUT * 2),
+            timeout=(CachedSettings().get_timeout_tuple()[0], CachedSettings().get_timeout_tuple()[1] * 2),  # noqa: E501
```
LefterisJP, line 626: you are asking for the same thing twice. Put it in a variable and reuse it in here.

rotkehlchen/externalapis/github.py
```diff
@@ -18,7 +17,9 @@ def _query(self, path: str) -> dict:
         - RemoteError if there is a problem querying Github
         """
         try:
-            response = requests.get(url=f'{self.prefix}{path}', timeout=DEFAULT_TIMEOUT_TUPLE)
+            from rotkehlchen.db.settings import \
+                CachedSettings  # avoid circular import # pylint: disable=import-outside-toplevel
+            response = requests.get(url=f'{self.prefix}{path}', timeout=CachedSettings().get_timeout_tuple())  # noqa: E501
```
LefterisJP, line 22: This is not a good solution. Find a different way to do this. We already seem to do this in line 34, so something is wrong with this file.

What's more we never ever do what you did about splitting the line with `\`. Please don't do arbitrary stuff like this.

Generally these new CachedSettings should be in their own file with no other imports at all. To avoid all such import errors.

rotkehlchen/externalapis/xratescom.py
```diff
@@ -4,7 +4,7 @@
 from bs4 import BeautifulSoup, SoupStrainer
 
 from rotkehlchen.assets.asset import FiatAsset
-from rotkehlchen.constants.timing import DEFAULT_TIMEOUT_TUPLE
+from rotkehlchen.db import settings
```
LefterisJP, line 7: please don't do this ...these arbitrary import differences are annoying. Be consistent.

rotkehlchen/globaldb/updates.py
```diff
@@ -24,6 +23,7 @@
 from rotkehlchen.utils.misc import is_production
 from rotkehlchen.utils.network import query_file
 
+from ..db.settings import CachedSettings
```
LefterisJP, line 26: Again here .... don't use relative imports like this, unless it's the same module.

rotkehlchen/icons.py
```diff
@@ -13,7 +13,7 @@
 
 from rotkehlchen.assets.asset import Asset
 from rotkehlchen.assets.types import AssetType
-from rotkehlchen.constants.timing import DEFAULT_TIMEOUT_TUPLE
+from rotkehlchen.db import settings
```
LefterisJP, line 16: Again ... be consistent.

rotkehlchen/tests/api/test_settings.py
```diff
@@ -25,6 +34,59 @@
...
+    assert json_data['result']['connect_timeout'] == 45
+    assert json_data['result']['read_timeout'] == 45
+    assert json_data['result']['submit_usage_analytics'] is True
+
+    # Now make sure that the cached settings are also updated
+    assert CachedSettings().get_query_retry_limit() == 3
+    assert CachedSettings().get_timeout_tuple() == (45, 45)
+    assert CachedSettings().get_entry('submit_usage_analytics') is True
+
+    # logout the user and make sure cached settings are reset to default values
```
LefterisJP, line 77:
```suggestion
    # log the user out and make sure cached settings are reset to default values
```

rotkehlchen/utils/network.py
```diff
@@ -7,7 +7,7 @@
 import requests
 
 from rotkehlchen.constants import GLOBAL_REQUESTS_TIMEOUT
-from rotkehlchen.constants.timing import DEFAULT_TIMEOUT_TUPLE, QUERY_RETRY_TIMES
+from rotkehlchen.db import settings
```
LefterisJP, line 10: again with the incosistency

```suggestion
from rotkehlchen.db.settings import CachedSettings
```

LefterisJP (commented):
Lots of comments and stuff that need to be addressed.

Please be consistent and more focused when developing.

There is also places that can be changed I suppose later and use the CachedSettings instead of passing the DB handleronly for settings (if there is any such place) which we could change, but this can happen in later PR.s

rotkehlchen/db/settings.py
dimyG: Yes, it's very bad. Since I'm not familiar with the entire codebase i don't know if the DBSettings dataclass has a reason to be frozen (as it is now). If not, then I can just mark it as frozen=False and use that. I tried to avoid the duplication with inheritance but in datclasses you can't have unforzen child and frozen parent class. I tried a different approach too but didn't work as expected. So  Ileft this duplicaiton and see how to handle it better after dicsussion. The best way is to make teh DBSettings frozen=False.

LefterisJP: please be asking question instead of making assumptions. Thigns will go much faster.

rotkehlchen/db/updates.py
dimyG: I do the import like this, to solve circular dependency issues. If you do  "from rotkehlchen.db.settings import CachedSettings" then you get circular dependency error. If I'm not mistaken the reason is that in the latter case, python evaluate the whole module on import (runs the top level code etc.) to find CachedSettings and this causes the circular dependency..

rotkehlchen/externalapis/github.py
dimyG: ok, isort does this splitting automatically. I whould isort:skip this line

rotkehlchen/chain/substrate/manager.py
dimyG: Ahh, this was done by the Pycharm automatically when I moved the CachedSettings

LefterisJP: Please don't use any external configurations apart from our own configs for all linting programs we use. This is what make lint does. This is how we guarantee consistency across the codebase.

rotkehlchen/db/settings.py
dimyG: So, should I make the DBSettings unfrozen?

rotkehlchen/db/drivers/gevent.py
LefterisJP: Should be fine. Can leave a note explaining why, and if we ever want to add it we can but I don't see it as a requirments. As this is read in only a single place, but we would be writingto the cache with extra logic in every single DB write.

rotkehlchen/externalapis/github.py
dimyG: I noticed that if you remove the "as" in the import and just add a long comment, isort will split the line like in my case. For example copy paste my import.

rotkehlchen/externalapis/coingecko.py
```diff
@@ -9,7 +9,8 @@
 from rotkehlchen.assets.asset import Asset, AssetWithOracles
 from rotkehlchen.constants.misc import ZERO_PRICE
 from rotkehlchen.constants.resolver import evm_address_to_identifier, strethaddress_to_identifier
-from rotkehlchen.constants.timing import DAY_IN_SECONDS, DEFAULT_TIMEOUT_TUPLE
+from rotkehlchen.constants.timing import DAY_IN_SECONDS
+from rotkehlchen.db import settings
```
LefterisJP, line 13: don't do this.

rotkehlchen/externalapis/github.py
LefterisJP: we never ever do this. Adjust your editor to not do this. Also I think that if you move it to its own file as adviced then you can fix the circular import and remove this hack from here

rotkehlchen/icons.py
LefterisJP: Again here you ignored my comment.

rotkehlchen/utils/network.py
LefterisJP: Same thing here. You ignored my comment.

LefterisJP (commented):
Please do not ignore my comments in a PR.

LefterisJP (commented):
Btw just checking out the current branch and doing `python -m rotkehlchen` fails with circular import.

You need to restructure the code. It's not working even like this.

LefterisJP (commented):
All test failures, were due to optimism etherscan being down. Works now again so I restarted

LefterisJP (approved):
All green. So will merge.

PR #6391, opened 2023-07-12
fix signal linting issues on windows

dimyG:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/server.py
+++ b/rotkehlchen/server.py
@@ -57,7 +57,7 @@ class RotkehlchenServer:
         hub.threadpool_size = 2
         hub.threadpool.maxsize = 2
         if os.name != 'nt':
-            gevent.hub.signal(signal.SIGQUIT, self.shutdown)
+            gevent.hub.signal(signal.SIGQUIT, self.shutdown)  # type: ignore[attr-defined,unused-ignore]  # pylint: disable=no-member  # noqa: E501  # linters don't understand the os.name check
         gevent.hub.signal(signal.SIGINT, self.shutdown)
         gevent.hub.signal(signal.SIGTERM, self.shutdown)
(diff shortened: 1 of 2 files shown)


rotkehlchen/server.py
```diff
@@ -57,7 +57,7 @@ def main(self) -> None:
         hub.threadpool_size = 2
         hub.threadpool.maxsize = 2
         if os.name != 'nt':
-            gevent.hub.signal(signal.SIGQUIT, self.shutdown)
+            gevent.hub.signal(signal.SIGQUIT, self.shutdown)  # type: ignore[attr-defined,unused-ignore]  # pylint: disable=no-member  # noqa: E501  # windows doesn't have SIGQUIT
```
LefterisJP, line 60: This is why the if check is there.


```suggestion
            gevent.hub.signal(signal.SIGQUIT, self.shutdown)  # type: ignore[attr-defined,unused-ignore]  # pylint: disable=no-member  # noqa: E501  # linters don't understand the os.name check
```

tools/profiling/timer.py
```diff
@@ -4,8 +4,8 @@
 
 from .constants import INTERVAL_SECONDS
 
-TIMER = signal.ITIMER_PROF
-TIMER_SIGNAL = signal.SIGPROF
+TIMER = signal.ITIMER_PROF  # type: ignore[attr-defined,unused-ignore]  # pylint: disable=no-member  # noqa: E501  # windows doesn't have ITIMER_PROF
```
LefterisJP, line 7: Put those under an os.name check then same as the one below and explain in the comment why and in the type ignore that it's just the linters not undestanding the os.name check

LefterisJP (approved):
LGTM

PR #6398, opened 2023-07-17
Modify some tests so they pass on windows

IanMichaelHarper:
Partially closes https://github.com/rotki/rotki/issues/5168 by modifying some of the tests so they pass on windows.

Unfortunately these changes don't make the tests run any faster on windows and the test suite still takes almost twice as long to run on windows as it does on linux.

## Changes
- It seems some tests failed when run together but passed individually, so I added @pytest.mark.parametrize('use_clean_caching_directory', [True]) to these tests.

- Windows doesn't seem to like Path's / for directory concatenation, so I replaced with os.path.join() instead.

- os.chmod was used in some tests to change directory permissions, but this has no effect on windows, so I mocked them

- Some new cassettes are needed so https://github.com/rotki/test-caching/pull/35 will need to be merged for the tests on this PR to pass
- Most other tests were failing for various different reasons.
- Also added some typing.


rotkehlchen/tests/api/test_liquity.py
```diff
@@ -252,7 +254,10 @@ def test_staking_stats(rotkehlchen_api_server, ethereum_accounts):
     assert len(global_stats['staking_gains']) == 3
     assert len(global_stats['stability_pool_gains']) == 2
     assert FVal(global_stats['stability_pool_gains'][0]['amount']).is_close(FVal(address_0_data['stability_pool_gains'][0]['amount']) + FVal(address_1_data['stability_pool_gains'][0]['amount']), max_diff='1e-8')  # noqa: E501
-    assert address_0_data['stability_pool_gains'][0]['amount'] == '1.7820064710306824'
+    assert math.isclose(
```
LefterisJP, line 257: huh that's interesting. Any idea why this is different for windows only?

What you should do instead btw is use the `Fval` functionality we got which you can see in action in the line right above as I think `address_0_data['stability_pool_gains'][0]['amount']` is already an `FVal`

rotkehlchen/tests/api/test_user_assets.py
```diff
@@ -889,7 +895,7 @@ def test_exporting_user_assets_list(rotkehlchen_api_server, globaldb, with_custo
         coingecko='YAB',
         cryptocompare='YAB',
     ))
-    with tempfile.TemporaryDirectory() as path:
+    with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as path:
```
LefterisJP, line 898: Interesting. I see what this does from the docs (https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory). Can you maybe explain why in a small comment next to this, why this is needed for Windows tests? As the reader of the code will end up wondering in the future.

rotkehlchen/tests/api/test_user_assets.py
```diff
@@ -952,13 +958,17 @@ def test_exporting_user_assets_list(rotkehlchen_api_server, globaldb, with_custo
...
+def test_importing_user_assets_list(
+        rotkehlchen_api_server: APIServer,
+        method: str,
+        file_type: str,
+):
     """Test that the endpoint for importing user assets works correctly"""
     dir_path = Path(__file__).resolve().parent.parent
     if file_type == 'zip':
-        filepath = dir_path / 'data' / 'exported_assets.zip'
+        filepath = os.path.join(dir_path, 'data', 'exported_assets.zip')
```
LefterisJP, line 969: This should not be needed. Why exactly is it faiing here? `dir_path` is a `Path` and as such should be OS agnostic: https://docs.python.org/3/library/pathlib.html

Something else must be wrong on this one.

What's more, `os.path.join` already returns a string and you stringify it again later down (or well the previous code did).

But generally we need to figure out why this is wrong here. It should work.

rotkehlchen/tests/api/test_users.py
```diff
@@ -144,9 +150,16 @@ def test_user_creation_with_no_analytics(rotkehlchen_api_server, data_dir):
 
 @pytest.mark.parametrize('use_clean_caching_directory', [True])
 @pytest.mark.parametrize('start_with_logged_in_user', [False])
-def test_user_creation_permission_error(rotkehlchen_api_server, data_dir):
+@mock.patch.object(
```
LefterisJP, line 153: nice!

rotkehlchen/tests/db/test_locations.py
```diff
@@ -19,6 +21,7 @@
 )
 
 
+@pytest.mark.parametrize('use_clean_caching_directory', [True])
```
LefterisJP, line 24: Okay so looking at this being put everywhere I am starting to suspect there is some sort of fixture we have that works fine and implcitly clears something in Unix, but does not work in Windows. Can't explain the need to add this everywhere otherwise.

rotkehlchen/tests/unit/test_assets.py
```diff
@@ -749,7 +749,7 @@ def test_symbol_or_name(database):
 def test_load_from_packaged_db(globaldb: GlobalDBHandler):
     """Test that connecting to the packaged globaldb doesn't try to write into it."""
     packaged_db_path = Path(__file__).resolve().parent.parent.parent / 'data' / 'global.db'
-    with TemporaryDirectory() as tmpdirname:
+    with TemporaryDirectory(ignore_cleanup_errors=True) as tmpdirname:
```
LefterisJP, line 752: same here for explaining why this is needed

rotkehlchen/tests/unit/test_cost_basis.py
```diff
@@ -533,15 +535,17 @@ def test_accounting_lifo_order(accountant):
     cost_basis.reset(DBSettings(cost_basis_method=CostBasisMethod.LIFO))
     asset_events = cost_basis.get_events(asset)
     # first we do a simple test that from 2 events the second is used
+    d = dt.datetime(2023, 4, 15, tzinfo=dt.timezone.utc)
```
LefterisJP, line 538: Why is this change needed? Why would timestamp 1, 2 etc. not work?

rotkehlchen/tests/unit/test_price_historian.py
```diff
@@ -144,9 +146,11 @@ def test_token_to_fiat_no_price_found_exception(fake_price_historian):
     """
     price_historian = fake_price_historian
 
+    d = dt.datetime(2023, 4, 15, tzinfo=dt.timezone.utc)
+    unix_timestamp = time.mktime(d.timetuple())
```
LefterisJP, line 150: same question here for timestamp

LefterisJP (commented):
Hey @IanMichaelHarper thanks a lot for the PR!

I see  two tests failed due to VCR cassettes:

```
FAILED rotkehlchen/tests/unit/test_general_cache.py::test_curve_cache[True-True-True] - vcr.errors.CannotOverwriteExistingCassetteException: Can't overwrite existing cassette ('/home/runner/work/rotki/rotki/test-caching/cassettes/unit/test_general_cache/test_curve_cache[True-True-True].yaml') in your current record mode ('none').
No match for the request (<Request (GET) https://api.curve.fi/api/getPools/ethereum/main>) was found.
No similar requests, that have not been played, found.
FAILED rotkehlchen/tests/unit/test_general_cache.py::test_curve_cache[True-False-True] - vcr.errors.CannotOverwriteExistingCassetteException: Can't overwrite existing cassette ('/home/runner/work/rotki/rotki/test-caching/cassettes/unit/test_general_cache/test_curve_cache[True-False-True].yaml') in your current record mode ('none').
No match for the request (<Request (GET) https://api.etherscan.io/api?module=proxy&action=eth_call&to=0x0000000022D53366457F9d5E68Ec105046FC4383&data=0x493f4f740000000000000000000000000000000000000000000000000000000000000007&apikey=8JT7WQBB2VQP5C3416Y8X3S8GBA3CVZKP4>) was found.
```

I will need to push the branch properly to the VCR repo, something which external contributors can't do. It's explained here: https://rotki.readthedocs.io/en/latest/contribute.html#syncing-with-the-cassettes-repository

I will do it and rebase this PR perhaps after merging the one with the DNS fix on windows.

rotkehlchen/tests/api/test_eth2.py
```diff
@@ -297,7 +297,7 @@ def test_query_eth2_deposits_details_and_stats(rotkehlchen_api_server, ethereum_
 def test_eth2_add_eth1_account(rotkehlchen_api_server):
     """This test uses real data and tests that adding an ETH1 address with
     ETH2 deposits properly detects validators"""
-    new_account = '0xa966B0eabCD717fa28Bd165F1cE160E7057FA369'
+    new_account = '0xa966B0eabCD717fa28Bd165F1cE160E7057FA369'  # exited
```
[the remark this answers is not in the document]
IanMichaelHarper: No, it fails. I stopped looking into it after I saw it was skipped in the CI since it is a slow test and the ultimate goal was to re-enable the windows nightly tests in the CI. I can fix it as part of this PR but maybe it would be better to leave it for another PR? A bunch of the tests that are skipped in the CI fail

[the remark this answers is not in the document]
LefterisJP: Oh had not even noticed it's skipped in CI. Yes ofc, if something is skipped in CI ignore for now.

rotkehlchen/tests/api/test_liquity.py
IanMichaelHarper: > huh that's interesting. Any idea why this is different for windows only?

Not sure, but it's off by `2e-16` which I would put down to floating point precision differences between OSs

> What you should do instead btw is use the Fval functionality we got which you can see in action in the line right above as I think address_0_data['stability_pool_gains'][0]['amount'] is already an FVal

Oh cool, I didn't see that. Neat, thanks!

rotkehlchen/tests/unit/test_cost_basis.py
IanMichaelHarper: Because it doesn't have a timezone. Here's the stacktrace:
```
rotkehlchen\tests\unit\test_cost_basis.py:575:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
rotkehlchen\accounting\cost_basis\base.py:218: in calculate_spend_cost_basis
    time=timestamp_to_date(acquisition_event.timestamp),
rotkehlchen\utils\mixins\customizable_date.py:24: in timestamp_to_date
    return timestamp_to_date(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

ts = 2, formatstr = '%d/%m/%Y %H:%M:%S %Z', treat_as_local = True

    def timestamp_to_date(
            ts: Timestamp,
            formatstr: str = '%d/%m/%Y %H:%M:%S',
            treat_as_local: bool = False,
    ) -> str:
        """Transforms a timestamp to a datestring depending on given formatstr and UTC/local choice"""
        if treat_as_local is False:
            date = datetime.datetime.fromtimestamp(ts, tz=datetime.timezone.utc).strftime(formatstr)
        else:  # localtime
            date = datetime.datetime.fromtimestamp(
                ts,  # ignore below is due to: https://github.com/pjknkda/flake8-datetimez/issues/11
>               tz=datetime.datetime.fromtimestamp(ts).astimezone().tzinfo,  # noqa: DTZ006
            ).strftime(formatstr)
E           OSError: [Errno 22] Invalid argument
```

This is the behaviour on windows:
```
>>> dt.datetime.fromtimestamp(1).astimezone()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OSError: [Errno 22] Invalid argument
```

LefterisJP: Hmm I am still not sure what the error is. 

So `dt.datetime.fromtimestamp(1)` works right? It's just that the `astimezone()` doesn't? And somehow by adding `d = dt.datetime(2023, 4, 15, tzinfo=dt.timezone.utc)` you initialize timezone to utc? What does `dt.datetime.fromtimestamp(1)` return?

IanMichaelHarper: > So dt.datetime.fromtimestamp(1) works right? It's just that the astimezone() doesn't?

Yes

> And somehow by adding d = dt.datetime(2023, 4, 15, tzinfo=dt.timezone.utc) you initialize timezone to utc?

Yes

>What does dt.datetime.fromtimestamp(1) return?

`datetime.datetime(1970, 1, 1, 1, 0, 1)`

LefterisJP: Okay so astimezone() from a timestamp without a timezone returns an error. The docs do not mention any such thing. 
https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone

Without an arg it should take localtime which is what we want.

Only relevant issue I can find is this: https://github.com/python/cpython/issues/80940

So on the naive timezone object, the `astimezone` calls `time.localtime()` and perhaps calls it with a negative argument? Could be a timezone problem that the time becomes a pre-epoch time for you if your timezone is west of UTC?

Seems like the problem is deeper than just the test and the real fix would need to be done in the function, but need to first understand why/what's happening.

IanMichaelHarper: >Could be a timezone problem that the time becomes a pre-epoch time for you if your timezone is west of UTC?
Not the case as I'm in UTC+2

It's very strange, it actually fails for all timestamps up until 86400
```
>>> dt.datetime.fromtimestamp(86400).astimezone()
datetime.datetime(1970, 1, 2, 1, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=3600), 'W. Europe Standard Time'))
>>> dt.datetime.fromtimestamp(86399).astimezone()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OSError: [Errno 22] Invalid argument
```

LefterisJP: It seems to be then exatctly like this bug: https://github.com/python/cpython/issues/73283
But in a different place. I don't see this bug reported so perhaps it would need to be reported. Is this in python 3.10 and 3.11? I guess then we found a bug in python in Windows and would need to report it.

IanMichaelHarper: 3.9, 3.10 and 3.11 it fails for me on windows 11

LefterisJP: Okay so then 2 questions.

1. Should I report a bug in python? Or do you wanna do it? Pointing to the previous bug and explaining that we hit the same thing with astimezone()

2. Production-wise ... what should be done here? Adding two extra ifchecks in all the `timestamp_to_date` calls seems silly. Perhaps just adjust test to use timestamp 86400 and up? And then add a comment pointing to the bug and leaving the handling for the future? Rotki in its current form won't handle any such timestamps so close to the EPOCH. What do you think?

IanMichaelHarper: > Should I report a bug in python? Or do you wanna do it? Pointing to the previous bug and explaining that we hit the same thing with astimezone()

I'm happy to let you claim the glory of finding a bug in python if you want ;) otherwise I can do it if you want to focus on other things.

> Perhaps just adjust test to use timestamp 86400 and up?

I think that makes the makes the most sense. If this is a bonafide bug in python then it will eventually get fixed upstream, and like you said no one is going to be using any dates that close to EPOCH, so I don't think there's any danger to production here.

rotkehlchen/tests/api/test_user_assets.py
IanMichaelHarper: You're right. This one actually passes. I think this is a remnant from my early iterations where changing this (somehow) fixed the test, but it passes now without this change so I'll change it back.

rotkehlchen/tests/unit/test_cost_basis.py
LefterisJP: > happy to let you claim the glory of finding a bug in python if you want ;)

Made an issue here: https://github.com/python/cpython/issues/107078

> I think that makes the makes the most sense. If this is a bonafide bug in python then it will eventually get fixed upstream, and like you said no one is going to be using any dates that close to EPOCH, so I don't think there's any danger to production here.

All right so when you make any change please do that then in both places you added the timestamp fix.

rotkehlchen/tests/unit/test_cost_basis.py
```diff
@@ -535,13 +535,13 @@ def test_accounting_lifo_order(accountant):
     # first we do a simple test that from 2 events the second is used
     event1 = AssetAcquisitionEvent(
         amount=ONE,
-        timestamp=1,
+        timestamp=1614556800,  # 01/03/2021, changed from 1 for windows. See https://github.com/rotki/rotki/pull/6398#discussion_r1271323846  # noqa: E501
```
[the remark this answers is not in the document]
IanMichaelHarper: Of course. Can't believe I didn't think of that at the time. Thanks!

PR #6401, opened 2023-07-17
Fix gemini file typo

LefterisJP:


LefterisJP:
test passed locally (but seems flaky -- but also unrelated) so will merge for now.

PR #6408, opened 2023-07-22
Fix slow windows tests and re-enable windows nightly tests

### LINKED ISSUE #5168: Figure out and fix slow windows CI test runs and enable windows backend tests

IanMichaelHarper:
Work on #5168 together with https://github.com/rotki/rotki/pull/6398 and https://github.com/rotki/test-caching/pull/35

## Problem
After profiling the tests it became clear that `requests` calls were the culprit. `requests` calls took 4 seconds longer on windows than it did on linux. 

## Solution
Turns out this is a DNS issue as `requests` use DNS in the case when it's not given an IP Address. Changing `localhost` to `127.0.0.1` when creating the api server for the tests fixes the issue.

It appears that the slow connections on Windows were not directly due to name resolution, but were caused by `urllib3` (the transport library used by `requests`) attempting an IPv6 connection first, which is refused since the server is listening on IPv4 only. For Windows only, however, there is an unavoidable 1 second timeout on `socket.connect` for refused TCP connections, while the OS 'helpfully' retries the connection up to 3 times. This explains the 4 second time difference I saw when I profiled the tests.

A local run on my windows 11 machine confirms the tests runs in almost the same time as linux or macos with this change, therefore I've re-enabled the windows nightly.

### DIFF
--- a/.github/workflows/rotki_nightly.yml
+++ b/.github/workflows/rotki_nightly.yml
@@ -24,6 +24,14 @@ jobs:
       test_environment: nightly
     secrets: inherit
 
+  windows:
+    uses: ./.github/workflows/task_backend_tests.yml
+    if: ${{ always() }}
+    with:
+      os: windows-latest
+      test_environment: nightly
+    secrets: inherit
+
   e2e:
     uses: ./.github/workflows/task_e2e_tests.yml
     if: ${{ always() }}
--- a/rotkehlchen/tests/api/test_statistics.py
+++ b/rotkehlchen/tests/api/test_statistics.py
@@ -7,6 +7,7 @@ import pytest
 import requests
 
 from rotkehlchen.accounting.structures.balance import BalanceType
+from rotkehlchen.api.server import APIServer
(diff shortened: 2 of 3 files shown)


LefterisJP (approved):
LGTM. The failed tests are fixed in develop by me.

PR #6411, opened 2023-07-24
Fix bug with simultaneous filter asset searches

LefterisJP:
Doing multiple simultaneous filter asset searches no longer results in Global DB locked error.

### DIFF
--- a/rotkehlchen/db/search_assets.py
+++ b/rotkehlchen/db/search_assets.py
@@ -52,51 +52,53 @@ def _search_only_assets_levenstein(
     resolved_eth = A_ETH.resolve_to_crypto_asset()
     globaldb = GlobalDBHandler()
     treat_eth2_as_eth = db.get_settings(cursor).treat_eth2_as_eth
-    cursor.execute(
-        f'ATTACH DATABASE "{globaldb.filepath()!s}" AS globaldb KEY "";',
-    )
-    try:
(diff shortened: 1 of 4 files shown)


rotkehlchen/db/search_assets.py
```diff
@@ -52,51 +52,53 @@ def _search_only_assets_levenstein(
...
-                        'identifier': resolved_eth.identifier,
-                        'name': resolved_eth.name,
-                        'symbol': resolved_eth.symbol,
-                        'asset_type': AssetType.OWN_CHAIN.serialize(),
-                    }))
-                    found_eth = True
-                continue
+    with db.conn.critical_section():  # needed due to ATTACH. Must not context switch out of this
+        cursor.execute(
+            f'ATTACH DATABASE "{globaldb.filepath()!s}" AS globaldb KEY "";',
```
yabirgb, line 57: is the key needed here? since the globaldb is not encrypted

yabirgb (approved):

rotkehlchen/db/search_assets.py
LefterisJP: Absolutely yes. As this is an sqlcipher connection and we are attaching a normal sqlite DB to it. https://stackoverflow.com/questions/19742498/unable-to-attach-with-sqlcipher-encrypted-database-to-another-one

PR #6413, opened 2023-07-25
Fix remaining windows nightly tests

IanMichaelHarper:
## Problem
`test_try_premium_at_start_new_account_different_password_than_remote_db` and `test_user_creation_with_invalid_premium_credentials` fail on windows with the following error:
```
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_data_dir604\\testuser\\rotkehlchen.db'
``` 
## Solution
In both cases `shutil.move()` was trying to create a backup of the DB but the pre-existing connection did not appear to get closed. Therefore I've added lines to explicitly close the connection before making the backup.

### DIFF
--- a/rotkehlchen/premium/sync.py
+++ b/rotkehlchen/premium/sync.py
@@ -115,6 +115,8 @@ class PremiumSyncManager:
         try:
             self.data.decompress_and_decrypt_db(result['data'])
         except UnableToDecryptRemoteData as e:
+            self.data.db.disconnect()
+            self.data.db.disconnect(conn_attribute='conn_transient')
             raise PremiumAuthenticationError(
                 'The given password can not unlock the database that was retrieved  from '
                 'the server. Make sure to use the same password as when the account was created.',
@@ -254,8 +256,10 @@ class PremiumSyncManager:
         premium API keys and we failed. But a directory was created. Remove it.
         But create a backup of it in case something went really wrong
         and the directory contained data we did not want to lose"""
+        user_data_dir = self.data.user_data_dir
+        self.data.db.logout()  # wipes self.data.user_data_dir
         shutil.move(
-            self.data.user_data_dir,  # type: ignore
+            user_data_dir,  # type: ignore
             self.data.data_directory / f'auto_backup_{username}_{ts_now()}',
         )
         raise PremiumAuthenticationError(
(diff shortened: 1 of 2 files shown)


rotkehlchen/premium/sync.py
```diff
@@ -115,6 +115,8 @@ def _sync_data_from_server_and_replace_local(self, perform_migrations: bool) ->
         try:
             self.data.decompress_and_decrypt_db(result['data'])
         except UnableToDecryptRemoteData as e:
+            self.data.db.disconnect()
```
LefterisJP, line 118: just to be explicit since you call it 2 times

```suggestion
            self.data.db.disconnect(conn_attribute='conn')
```

Also since this already raises `PremiumAuthenticationError` shouldn't the cleanup happen in `_abort_new_syncing_premium_user` which seems to be called wherever a `PremiumAuthenticationError` is handled?

rotkehlchen/premium/sync.py
```diff
@@ -254,8 +256,10 @@ def _abort_new_syncing_premium_user(
         premium API keys and we failed. But a directory was created. Remove it.
         But create a backup of it in case something went really wrong
         and the directory contained data we did not want to lose"""
+        user_data_dir = self.data.user_data_dir
+        self.data.db.logout()  # wipes self.data.user_data_dir
```
LefterisJP, line 260: Hmmm it already seems to happen here. So perhaps with this you don't need the explicit disconnects above?
```suggestion
        self.data.db.logout()  # wipes self.data.user_data_dir, so store it
```

rotkehlchen/premium/sync.py
IanMichaelHarper: You're right.

rotkehlchen/premium/sync.py
IanMichaelHarper: Yes you're correct. Doing this fixes one of the tests, but unfortunately doesn't fix the other. I must have ran the same test twice locally before pushing thinking it was the two failing ones. I am going to have to re-think this one as it's not as simple as I first thought. I am thinking of removing `_abort_new_syncing_premium_user` altogether...

This is actually a genuine bug in production though. I was able to reproduce it and got the same error:
image

PR #6415, opened 2023-07-25
Fix premium account creation bug on Windows

IanMichaelHarper:
## Problem
On Windows, when trying to create a new premium account with a random base64 string for the API secret, the backend throws an unexpected error
image
## Cause
On windows the database connection is not properly closed before trying to back it up for the new user.
## Solution
Close the DB connection before backing up the new user's DB.
## Tests
After the fix I performed the same steps I did to re-produce the bug, and now we get the expected error message: `Could not verify keys for the new account. rotki API key was rejected by server`.

Also logged in as a non-premium user and navigated around... added a bitcoin address to check balances... things seem to work fine.

`test_try_premium_at_start_new_account_different_password_than_remote_db` and `test_user_creation_with_invalid_premium_credentials` now pass with this change.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` Errors creating a premium account with an invalid premium key in Windows are now handled gracefully
 * :bug:`-` Doing multiple simultaneous filter asset searches no longer results in Global DB locked error.
 * :bug:`-` Removing an evm address will no longer affect metadata such as detected tokens of the address if it is also tracked for another evm chain.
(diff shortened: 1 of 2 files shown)


LefterisJP (approved):
LGTM. Can you also add a changelog entry by amending the commit?

Something like:

"Errors creating a premium account with an invalid premium key in Windows are now handled gracefully"

IanMichaelHarper:
by amending the commit? Does the changelog get auto-generated from commits somehow? If so, I'm not familiar. Do I just need to put the `fix:` prefix to my commit message?

LefterisJP (approved):
LGTM. Only failure is due to a timeout.

PR #6416, opened 2023-07-26
fix test_iso8601ts_to_timestamp() in windows ci

IanMichaelHarper:
Maybe github actions runs windows images in a different region? Is there a way to find out?

### DIFF
--- a/rotkehlchen/tests/unit/test_utils.py
+++ b/rotkehlchen/tests/unit/test_utils.py
@@ -1,5 +1,5 @@
 import json
-import time
+from datetime import datetime, timezone
 from json.decoder import JSONDecodeError
 from unittest.mock import patch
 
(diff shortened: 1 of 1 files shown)


LefterisJP:
> Maybe github actions runs windows images in a different region? Is there a way to find out?

Probably checking the output when building the image for clues. Also worst case adding some log/print to make sure.

But in any case all tests should be independent of timezone. If timezone specific then they are wrong (unless the test mocks a specific timezone for some reason)

IanMichaelHarper:
Windows CI doesn't support DST: 
image

So in the Windows CI `time.daylight == 0`, but the docs say not to use `time.altzone` in that case
image

The test is indeed timezone specific: https://github.com/rotki/rotki/blob/528a126aede41960d8be850723ef17097575c2fa/rotkehlchen/tests/unit/test_utils.py#L63

So perhaps there should be a separate section of this test dedicated to testing how `iso8601ts_to_timestamp()` behaves under different timezones and whether they support DST or not. In this section the timezone and DST could be explicitly set rather than take it from the runtime environment. What do you think?

IanMichaelHarper:
@LefterisJP I updated the PR with changes which I think are most suitable for the test in its current form. I just used `datetime` instead of `time` to get the `utc_offset` since that's what `iso8601ts_to_timestamp()` uses to interpret date strings without timezone information. 

With these change the test now passes in github actions for linux, windows and macos.

LefterisJP (approved):
LGTM

PR #6419, opened 2023-07-26
Improve error messages for server actions

yabirgb:
image

### DIFF
--- a/rotkehlchen/premium/sync.py
+++ b/rotkehlchen/premium/sync.py
@@ -136,7 +136,14 @@ class PremiumSyncManager:
 
         return True
 
-    def maybe_upload_data_to_server(self, force_upload: bool = False) -> bool:
+    def maybe_upload_data_to_server(
+            self,
+            force_upload: bool = False,
+    ) -> tuple[bool, Optional[str]]:
+        """
+        Returns a boolean value denoting whether we can upload the DB to the server.
+        In case of error we also return a message to provide information to the user.
+        """
         with self.upload_lock:
             assert self.premium is not None, 'caller should make sure premium exists'
             log.debug('Starting maybe_upload_data_to_server')
@@ -144,7 +151,7 @@ class PremiumSyncManager:
                 metadata = self.premium.query_last_data_metadata()
             except (RemoteError, PremiumAuthenticationError) as e:
                 log.debug('upload to server -- fetching metadata error', error=str(e))
-                return False
+                return False, str(e)
 
             with self.data.db.conn.read_ctx() as cursor:
                 our_last_write_ts = self.data.db.get_setting(cursor=cursor, name='last_write_ts')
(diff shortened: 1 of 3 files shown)


rotkehlchen/premium/sync.py
```diff
@@ -136,15 +136,22 @@ def check_if_should_sync(self, force_upload: bool) -> bool:
 
         return True
 
-    def maybe_upload_data_to_server(self, force_upload: bool = False) -> bool:
+    def maybe_upload_data_to_server(
+            self,
+            force_upload: bool = False,
+    ) -> tuple[bool, Optional[str]]:
+        """
+        Returns a boolean value meaning if we can upload or not the db to the server.
```
LefterisJP, line 144:
```suggestion
        Returns a boolean value denoting whether we can upload the DB to the server.
```

rotkehlchen/premium/sync.py
```diff
@@ -136,15 +136,22 @@ def check_if_should_sync(self, force_upload: bool) -> bool:
...
+        In case of error we also return a message to provide to the user with information.
```
LefterisJP, line 145:
```suggestion
        In case of error we also return a message to provide information to the user.
```

rotkehlchen/premium/sync.py
```diff
@@ -154,7 +161,7 @@ def maybe_upload_data_to_server(self, force_upload: bool = False) -> bool:
                     f'upload to server stopped -- remote db({metadata.last_modify_ts}) '
                     f'more recent than local({our_last_write_ts})',
                 )
-                return False
+                return False, 'Remote db is more recent than local'
```
LefterisJP, line 164: Consistency with the other messages

```suggestion
                return False, 'Remote database is more recent than local'
```

LefterisJP (commented):
Some nitpicks. Also guess these should not have any new tests/affected or should we add some?

rotkehlchen/tests/integration/test_premium.py
```diff
@@ -517,3 +518,38 @@ def test_upload_data_to_server_big_db(rotkehlchen_instance):
...
+    patched_put = patch.object(
+        rotkehlchen_instance.premium.session,
+        'post',
+        side_effect=mock_error_upload_data_to_server,
+    )
+    patched_get = create_patched_requests_get_for_premium(
+        session=rotkehlchen_instance.premium.session,
+        metadata_last_modify_ts=0,
+        metadata_data_hash=remote_hash,
+        # Smaller Remote DB size
```
LefterisJP, line 547: What's this comment for? Why is it relevant?

yabirgb: because if the database size is bigger in the remote then the upload will fail with a different error

LefterisJP: okay then write that in the comment please next time. Leave a note somewhere to adjust this comment.

LefterisJP (approved):
Will merge but please remember to adjust that comment.

PR #6420, opened 2023-07-26
Querying asset existence with case insensitive ID now normalizes it

LefterisJP:
When resolving an asset via `resolve()` the returned resolved asset, has the normalized identifier. But when checking for asset existence with `check_existence()` the asset can be seen to be in the DB but its identifier will not be normalized.

This can result to all the following assets moving around in our codebase: Asset('ETH'), Asset('eth'), Asset('eTh'), Asset ('etH') etc. While the normalized identifier is 'ETH'.

This hit us recently in the icons query, where the frontend was sending us 'eth' instead of 'ETH', and this was seen as a valid asset, but its icon filepath was not found in Linux since it expected "$ICON_DIR/eth_small.png" but did not find it. What's interesting is that this works fine for Mac and Windows since filepaths there are case insensitive.

To solve it:

- Added a new type `ResolvedType` which is just used for typing in functions to signify we know an asset that has been checked to be in the DB.
- Modified the existence check of an asset to check for the identifier in the DB and return the normalized one. Then set it to the returned `Asset` instance
- Wrote test for the case that hit us with the icons and case insensitive identifiers

### DIFF
--- a/rotkehlchen/api/v1/resources.py
+++ b/rotkehlchen/api/v1/resources.py
@@ -153,7 +153,13 @@ from rotkehlchen.api.v1.schemas import (
     XpubAddSchema,
     XpubPatchSchema,
 )
-from rotkehlchen.assets.asset import Asset, AssetWithNameAndType, AssetWithOracles, CustomAsset
+from rotkehlchen.assets.asset import (
+    Asset,
+    AssetWithNameAndType,
+    AssetWithOracles,
+    CustomAsset,
+    ResolvedAsset,
+)
 from rotkehlchen.assets.types import AssetType
 from rotkehlchen.balances.manual import ManuallyTrackedBalance
 from rotkehlchen.chain.accounts import SingleBlockchainAccountData
@@ -2195,7 +2201,7 @@ class AssetIconFileResource(BaseMethodView):
     get_schema = SingleAssetIdentifierSchema()
 
     @use_kwargs(get_schema, location='query')
-    def get(self, asset: Asset) -> Response:
+    def get(self, asset: ResolvedAsset) -> Response:
         return self.rest_api.get_asset_icon(asset=asset, match_header=get_match_header())
 
 
--- a/rotkehlchen/assets/resolver.py
+++ b/rotkehlchen/assets/resolver.py
@@ -8,7 +8,6 @@ from rotkehlchen.utils.data_structures import LRUCacheWithRemove
 
 if TYPE_CHECKING:
     from rotkehlchen.assets.asset import (
-        Asset,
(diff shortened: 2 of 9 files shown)


rotkehlchen/api/v1/resources.py
```diff
@@ -153,7 +153,13 @@
...
     XpubPatchSchema,
 )
-from rotkehlchen.assets.asset import Asset, AssetWithNameAndType, AssetWithOracles, CustomAsset
+from rotkehlchen.assets.asset import (
+    Asset,
+    AssetWithNameAndType,
+    AssetWithOracles,
+    CustomAsset,
+    ResolvedAsset,
+)
```
yabirgb, line 162: nitpick: all this can be moved under if TYPE_CHECKING since they are just used for typing

rotkehlchen/assets/resolver.py
```diff
@@ -111,6 +114,36 @@ def get_asset_type(identifier: str, query_packaged_db: bool = True) -> AssetType
...
+
+        For example if 'eTh' is given here then 'ETH' should be returned.
+
+        May raise:
+        - UnknownAsset: If asset identifier does not exist.
+        """
+        # TODO: This is ugly here but is here to avoid a cyclic import in the Assets file
+        # Couldn't find a reorg that solves this cyclic import
+        from rotkehlchen.constants.assets import CONSTANT_ASSETS  # pylint: disable=import-outside-toplevel  # isort:skip  # noqa: E501
+        from rotkehlchen.globaldb.handler import GlobalDBHandler   # pylint: disable=import-outside-toplevel  # isort:skip  # noqa: E501
```
yabirgb, line 129: We should look into this. I'm not sure if there is a performance hit due to having the import here

rotkehlchen/globaldb/handler.py
```diff
@@ -1858,6 +1858,30 @@ def get_asset_type(identifier: str, use_packaged_db: bool = False) -> AssetType:
...
+        with connection.read_ctx() as cursor:
+            normalized_id = cursor.execute(
+                'SELECT identifier FROM assets WHERE identifier=?',
+                (identifier,),
+            ).fetchone()
+
+        if normalized_id is None:  # should not happen
+            raise UnknownAsset(identifier)
+
+        return normalized_id
```
yabirgb, line 1883: this returns a tuple, it should be `normalized_id[0]`

rotkehlchen/assets/resolver.py
LefterisJP: yes. And it's not just this function. Many functions in this part. We really need to solve this dependency.

PR #6425, opened 2023-07-27
Optimism governance votes that include params should be decoded properly

LefterisJP:


LefterisJP:
all green. Under the bell tower, under my skin ... mergin.

Issue #6169, opened 2023-05-30
Handle known kraken edge cases with negative amounts properly


LefterisJP:
## Problem Definition

Some times the kraken API for its ledger events returns weird combinations of events that rotki does not handle properly. We should identify the ones we know that can happen and handle them properly

## Task
The known cases are below

### Failed Withdrawals

As seen here: https://twitter.com/LefterisJP/status/1663530022416547841

Figure out what should be handled better. Perhaps combining the info with the withdrawals endpoint you can see more information (including that it's failed)

The way it's displayed now is not terrible, but we need to make sure that it's a no-op in accounting. As it most probably is not. If we can also display it in a better way that would be nice.

Whoever implements it this I can provide test data I suppose.

### Edge cases in the new earn/staking

Some earn/staking events after kraken's revamp appear wrongly in the history events, with negative amounts.

![2023-09-21_18-34](https://github.com/rotki/rotki/assets/1658405/9dd1f637-8756-4b50-a2dc-393496f4db55)

This needs to be fixed.

### A user reported negative amounts in trades

~~This will need more investigation.~~ Can't reproduce or get data. I think he misunderstood as he mentioned tiny amounts which sound like the kraken internal adjustments I am fixing.

LefterisJP:
Putting this in 1.31.0, let's see if we can do at least part of it

Issue #6097, opened 2023-05-16
Move the trades fully in history events view


yabirgb:
This is a sub issue for #2469 

## Changes backend
The trades as a concept needs to be replaced in the code by history events:

- [ ] Migrate all trades in the database to history events
- [ ] Remove endpoints that make use of trades
- [ ] Remove accounting logic for trades
- [ ] Allow to edit/create/delete history events that replace the trades

## Changes frontend

TBD

yabirgb:
Closing as it shipped in 1.39.0

Issue #6096, opened 2023-05-16
Replace ledger actions with history events


yabirgb:
This is a sub issue for #2469 

The ledger actions need to be replaced in the code by history events:

- [ ] Migrate all ledger actions to history events
- [ ] Remove endpoints that make use of ledger actions
- [ ] Remove accounting logic for ledger actions
- [ ] Allow to edit/create/delete history events that replace the ledger actions

LefterisJP:
The last bullet point ofthis issue is not directly related and is moved to a note for the same release: https://github.com/orgs/rotki/projects/11?pane=issue&itemId=40691321

PR #6426, opened 2023-07-27
feat: support BitMEX wallet history imports

IanMichaelHarper:
# Bitmex Wallet History CSV Import
The lack of this feature is one of the reasons I never used Rotki for my own taxes, as I had traded on BitMEX a bunch in the past but then got locked out of my account and only had the `walletHistory.csv` file. 

I am not sure if the file I have is the "official format" for the file. I can't remember if I got it via download or just copy and pasting into an excel sheet, but it looks sufficiently similar to what is used in the [bitmex api test](https://github.com/rotki/rotki/blob/99eafffe371c9cc68eb2f6de70a675ff30807fc9/rotkehlchen/tests/exchanges/test_bitmex.py#L127-L133) to base this functionality off. It is missing a few of the fields that exist in that example string (such as `currency` and `transactID`) so this may need to be iterated on based on demands from users and if they have data in a different format. For now, all rows are assumed to be denominated in satoshis and only `Deposit`, `Withdrawal` and `RealisedPNL` events are supported.

Importing the `walletHistory.csv` file allows `RealisedPNL` events to show up on `History Events` tab as profit from a margin trade. A profit is marked as a `Receive` whilst a loss is marked as a `Spend`. Both are shown as positive values (should the `Spend`s be negative?). Deposits and withdrawals show up on their tab but the `Exch. Trades` tab is not populated (as these are not really exchange trades, but profit or loss from an exchange trade. Is is more akin to income/capital gains in that regard). the `margin_positions` table is populated in the DB and is used for the Profit and Loss Report. 
## Testing
I did a bunch of user acceptance testing, importing different files with different date formats with and without custom date format field checked. Seems to work fine. History events look correct, Deposits and Withdrawals look correct. Nothing in Exch. Trades. PnL Report is successfully generated. Values look realistic. No double counting between History Events and Margin Positions. Everything looks good except sometimes the taxable PnL sometimes shows 0 for some entries (see image below). I can't figure it out, maybe someone could point me in the right direction there?
image

## Checklist

- [x] The PR modified the frontend, backend and updated the [user guide](https://github.com/rotki/rotki/blob/develop/docs/usage_guide.rst) to reflect the changes.


IanMichaelHarper:
Turns out the issue with taxable PnL being 0 for some rows is because I didn't have a cost basis for them. After manually adding a large buy before the trade period, taxable PnL is correct for all rows.

LefterisJP:
> Turns out the issue with taxable PnL being 0 for some rows is because I didn't have a cost basis for them. After manually adding a large buy before the trade period, taxable PnL is correct for all rows.

So is this reviewable now? Also seems to need a rebase already, with a conflict in changelog :sweat_smile:

IanMichaelHarper:
> > Turns out the issue with taxable PnL being 0 for some rows is because I didn't have a cost basis for them. After manually adding a large buy before the trade period, taxable PnL is correct for all rows.
> 
> So is this reviewable now? Also seems to need a rebase already, with a conflict in changelog 😅

Yes It's ready for review. Will I rebase now?

LefterisJP:
> > > Turns out the issue with taxable PnL being 0 for some rows is because I didn't have a cost basis for them. After manually adding a large buy before the trade period, taxable PnL is correct for all rows.
> > 
> > 
> > So is this reviewable now? Also seems to need a rebase already, with a conflict in changelog sweat_smile
> 
> Yes It's ready for review. Will I rebase now?

Yeah please do it, since I have not looked at this yet anyway, to run tests again and be mergable

IanMichaelHarper:
> > > > Turns out the issue with taxable PnL being 0 for some rows is because I didn't have a cost basis for them. After manually adding a large buy before the trade period, taxable PnL is correct for all rows.
> > > 
> > > 
> > > So is this reviewable now? Also seems to need a rebase already, with a conflict in changelog sweat_smile
> > 
> > 
> > Yes It's ready for review. Will I rebase now?
> 
> Yeah please do it, since I have not looked at this yet anyway, to run tests again and be mergable

Rebased. Ready for review now.

rotkehlchen/data_import/importers/bitmex.py
```diff
@@ -0,0 +1,166 @@
...
+            event_type=HistoryEventType.MARGIN,
+            event_subtype=history_event_subtype,
+        )
+
+        return margin_position, history_event
+
+    @staticmethod
+    def _consume_deposits_or_withdrawals(
+            csv_row: dict[str, Any], timestamp_format: str = '%m/%d/%Y, %H:%M:%S %p',
+    ) -> AssetMovement:
```
LefterisJP, line 108: Can you add a small comment with possible exceptions this can raise? It's the way we try to trace exception handling in the code

rotkehlchen/data_import/importers/bitmex.py
```diff
@@ -0,0 +1,166 @@
...
+        - InputError if a column we need is missing
+        """
+        history_events = []
+        with open(filepath, encoding='utf-8-sig') as csvfile:
+            data = csv.DictReader(csvfile)
+            for row in data:
+                try:
+                    if row['transactType'] == 'RealisedPNL':
+                        margin_position, history_event = self._consume_realised_pnl(row, **kwargs)
+                        self.add_margin_trade(cursor, margin_position)
```
LefterisJP, line 148: So as I wrote in discord this is something that can't work. You should only keep the `MarginPosition` here. As we are duplicating the event otherwise.

We have open issues to break down everything into `HistoryEvent` eventually.

https://github.com/rotki/rotki/issues/6097
https://github.com/rotki/rotki/issues/6096

And possibly we need more. We also need for Margin positions. So when these would be completed we would duplicate any imported info.

So for now if you want to import margin positions, import them simply as MarginPosition.

You could in a next step:

1. Create an issue to turn `MarginPosition` into HistoryEvent.
2. Then implement it by:
    a. Removing MarginPosition
    b. Adding a user DB upgrade to turn all existing MarginPosition to HistoryEvent
    c. Replace all usages of MarginPosition with `HistoryEvent`
    d. Make sure accounting/PnL works fine for them

rotkehlchen/serialization/deserialize.py
```diff
@@ -44,7 +44,7 @@ def deserialize_fee(fee: Optional[str]) -> Fee:
 
     Can throw DeserializationError if the fee is not as expected
     """
-    if fee is None:
+    if fee is None or fee == 'null':
```
LefterisJP, line 47: Hmm I don't like this. The string here should only be a number. Can you handle the extra `'null'` case in the single source case where you encountered it?

LefterisJP (commented):
Few comments

rotkehlchen/data_import/importers/bitmex.py
```diff
@@ -0,0 +1,166 @@
...
+            'Processing Bitmex Realised PnL',
+            timestamp=close_time,
+            profit_loss=realised_pnl,
+            fee=fee,
+            notes=notes,
+        )
+
+        event_identifier = f'BMEX_{hash_csv_row(csv_row)}'
+        usd_price = PriceHistorian().query_historical_price(A_BTC, A_USD, close_time)
+        abs_amount = realised_pnl.__abs__()  # pylint: disable=no-member
```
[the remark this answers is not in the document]
IanMichaelHarper: Ah cool it does. I'll change it.

IanMichaelHarper:
@LefterisJP I updated the PR with changes from your comments. While I was working on them I realised an `AssetMovement` was being added for for a `Canceled` withdrawal, so I just added [this line](https://github.com/IanMichaelHarper/rotki/blob/792ef5c901624cb2a47f6a6ea06372114478ff0e/rotkehlchen/data_import/importers/bitmex.py#L132) to skip it. Do we want to show cancelled/failed events in general? If so, maybe work here could be paired with #6169 in terms of the design of how to show incompleted events.

lukicenturi:
Frontend part are okay

rotkehlchen/accounting/structures/types.py
```diff
@@ -67,6 +67,7 @@ class HistoryEventType(SerializableEnumNameMixin):
     INFORMATIONAL = auto()
     MIGRATE = auto()
     RENEW = auto()
+    MARGIN = auto()
```
LefterisJP, line 70: Can you remove this then since it's not used? Let's maybe add it when we need it. Not sure yet how we would like to represent it. On the type or subtype.

LefterisJP (commented):
Looks good, just needs to remove a now unneeded change.

> While I was working on them I realised an AssetMovement was being added for for a Canceled withdrawal, so I just added [this line](https://github.com/IanMichaelHarper/rotki/blob/792ef5c901624cb2a47f6a6ea06372114478ff0e/rotkehlchen/data_import/importers/bitmex.py#L132) to skip it. Do we want to show cancelled/failed events in general? If so, maybe work here could be paired with https://github.com/rotki/rotki/issues/6169 in terms of the design of how to show incompleted events.

The issues are different. One is in the API of kraken and the way we query it. Since we query only historical ledger entries and not the specific withdrawal info we don't get the "Canceled" information. So it actually shows very wrong.

While here you already see it's canceled from the CSV status. I don't think you need to let the user know anything. There is no need to display or take canceled events into account. Is there?

rotkehlchen/data_import/importers/bitmex.py
```diff
@@ -0,0 +1,146 @@
...
+    deserialize_timestamp_from_date,
+)
+from rotkehlchen.types import AssetAmount, AssetMovementCategory, Fee, Location
+from rotkehlchen.utils.misc import satoshis_to_btc
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+class BitMEXWalletHistoryColnames(Enum):
```
LefterisJP, line 31: Is this Enum used anywhere?

LefterisJP (approved):
LGTM.

Test that failed is `rotkehlchen/tests/unit/test_inquirer.py::test_saddle_oracle[` which is being fixed  by yabir.

But I want 2 things:

1. I suspect there is an unused Enum being introduced by the PR. Am I blind or is it not being used anywhere?
2. Generally keep commits limited/combined if they target the same thing. So here both commits would need to be squashed into one.

rotkehlchen/data_import/importers/bitmex.py
IanMichaelHarper: I added it after typing out all the column names when accessing `csv_row` but I never ended up using it. Do you think it's better to use something like this or just access columns of `csv_row` by typing them out directly?

LefterisJP: > I added it after typing out all the column names when accessing `csv_row` but I never ended up using it. Do you think it's better to use something like this or just access columns of `csv_row` by typing them out directly?

Hmmm if it's used in multiple places maybe this is not bad, but using them directly if they are just used in the parser is fine. No need to create in memory structures like this enum which will persists for the entire lifetime of the app.

IanMichaelHarper: ok, thanks. I wasn't sure so good to know. I'll remove it, squash my commits and push again

LefterisJP (approved):
LGTM

PR #6436, opened 2023-07-30
Use location asset mapping on generic csv import

IanMichaelHarper:
When doing a generic import of all of my trades, I get hundreds of unknown asset warnings for assets that Rotki does indeed support. This change improves the import by doing the following:
- If Location is one of the supported locations, use the asset mappings from that location
- Exclude spam tokens from list of results that might come from searching the DB for assets with the imported data symbol.

This change allows me to import my trades without skipping any rows due to unsupported asset warnings.

### DIFF
--- a/rotkehlchen/assets/exchanges_mappings/ftx.py
+++ b/rotkehlchen/assets/exchanges_mappings/ftx.py
@@ -1,0 +1,125 @@
+from rotkehlchen.assets.exchanges_mappings.common import COMMON_ASSETS_MAPPINGS
+from rotkehlchen.constants.resolver import evm_address_to_identifier, strethaddress_to_identifier
+from rotkehlchen.types import ChainID, EvmTokenKind
+
+WORLD_TO_FTX = COMMON_ASSETS_MAPPINGS | {
+    strethaddress_to_identifier('0x50D1c9771902476076eCFc8B2A83Ad6b9355a4c9'): 'FTT',
+    strethaddress_to_identifier('0xcca0c9c383076649604eE31b20248BC04FdF61cA'): 'ASD',
+    'SOL-2': 'SOL',
+    'COIN': 'COIN',
+    # SLP is smooth love potion
+    strethaddress_to_identifier('0xCC8Fa225D80b9c7D42F96e9570156c65D6cAAa25'): 'SLP',
+    'MER-2': 'MER',
+    strethaddress_to_identifier('0x476c5E26a75bd202a9683ffD34359C0CC15be0fF'): 'SRM',
+    strethaddress_to_identifier('0xF57e7e7C23978C3cAEC3C3548E3D615c346e79fF'): 'IMX',
+    'GENE': 'GENE',
+    strethaddress_to_identifier('0xC581b735A1688071A1746c968e0798D642EDE491'): 'EURT',
+    'LUNA-2': 'LUNA',
+    strethaddress_to_identifier('0x3392D8A60B77F8d3eAa4FB58F09d835bD31ADD29'): 'INDI',
+    strethaddress_to_identifier('0xAf5191B0De278C7286d6C7CC6ab6BB8A73bA2Cd6'): 'STG',
+    strethaddress_to_identifier('0x5c147e74D63B1D31AA3Fd78Eb229B65161983B2b'): 'WFLOW',
+    strethaddress_to_identifier('0x27702a26126e0B3702af63Ee09aC4d1A084EF628'): 'ALEPH',
+    evm_address_to_identifier('0x0f2D719407FdBeFF09D87557AbB7232601FD9F29', ChainID.ETHEREUM, EvmTokenKind.ERC20): 'SYN',  # noqa: E501
+    evm_address_to_identifier('0x5A98FcBEA516Cf06857215779Fd812CA3beF1B32', ChainID.ETHEREUM, EvmTokenKind.ERC20): 'LDO',  # noqa: E501
+    evm_address_to_identifier('0x08d967bb0134F2d07f7cfb6E246680c53927DD30', ChainID.ETHEREUM, EvmTokenKind.ERC20): 'MATH',  # noqa: E501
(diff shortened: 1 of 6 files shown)


rotkehlchen/data_import/importers/rotki_events.py
```diff
@@ -50,8 +50,10 @@ def _consume_rotki_event(
         location = Location.deserialize(csv_row['Location'])
         timestamp = TimestampMS(deserialize_timestamp(csv_row['Timestamp']))
         fee = Fee(deserialize_asset_amount(csv_row['Fee'])) if csv_row['Fee'] else Fee(ZERO)  # noqa: E501
+        asset_resolver = LOCATION_TO_ASSET_MAPPING.get(location, asset_from_common_identifier)
+        asset = asset_resolver(csv_row['Currency'])
```
LefterisJP, line 54: naming nitpick, since at least for me (and the way we use classes in the code) mapping, sounds like a dict mapping, while resolver like a class

```suggestion
        asset_mapping = LOCATION_TO_ASSET_MAPPING.get(location, asset_from_common_identifier)
        asset = asset_mapping(csv_row['Currency'])
```

rotkehlchen/data_import/importers/rotki_events.py
```diff
@@ -102,15 +104,12 @@ def _import_csv(self, cursor: DBCursor, filepath: Path, **kwargs: Any) -> None:
                         f'During rotki generic events CSV import, found action with unknown '
                         f'asset {e.identifier}. Ignoring entry',
                     )
-                    continue
```
LefterisJP, line 105: This is probably fine but not future-proof. I can see someone adding extra logic to the end of the function and adding a bug that will only hit when an exception hits.

Can you add  comment in the end of the function, that if we ever add more logic to add continue to all exception handlers?

rotkehlchen/data_import/importers/rotki_trades.py
```diff
@@ -34,13 +35,15 @@ def _consume_rotki_trades(
...
         """
         fee = Fee(deserialize_asset_amount(csv_row['Fee'])) if csv_row['Fee'] else None  # noqa: E501
-        fee_currency = (
-            symbol_to_asset_or_token(csv_row['Fee Currency'])
-            if csv_row['Fee Currency'] and fee is not None else None
-        )
         amount_sold = deserialize_asset_amount(csv_row['Sell Amount'])
         amount_bought = deserialize_asset_amount(csv_row['Buy Amount'])
         location = Location.deserialize(csv_row['Location'])
+        asset_resolver = LOCATION_TO_ASSET_MAPPING.get(location, asset_from_common_identifier)
```
LefterisJP, line 41: same anming comments here. May also be an overkill but since the lnes of code here are identical to `rotki_events` do you think you could abstract it into a helper function?

rotkehlchen/data_import/importers/rotki_trades.py
```diff
@@ -71,18 +74,15 @@ def _import_csv(self, cursor: DBCursor, filepath: Path, **kwargs: Any) -> None:
...
                         f'Deserialization error during rotki generic trades CSV import. '
                         f'{e!s}. Ignoring entry',
                     )
-                    continue
                 except DivisionByZero:
                     self.db.msg_aggregator.add_warning(
                         f'During rotki generic trades import, csv row {row!s} has zero '
                         f'amount bought. Ignoring entry',
                     )
-                    continue
```
LefterisJP, line 86: same commet on future-proofing here

LefterisJP (commented):
Left a review.

Failing tests are 503 from independent reserve.
And `rotkehlchen/tests/unit/test_arbitrum_one_inquirer.py::test_arbitrum_one_nodes_prune_and_archive_status` maybe points to a node status having changed?

rotkehlchen/assets/exchanges_mappings/ftx.py
```diff
@@ -0,0 +1,125 @@
+from rotkehlchen.assets.exchanges_mappings.common import COMMON_ASSETS_MAPPINGS
+from rotkehlchen.constants.resolver import evm_address_to_identifier, strethaddress_to_identifier
+from rotkehlchen.types import ChainID, EvmTokenKind
+
+WORLD_TO_FTX = COMMON_ASSETS_MAPPINGS | {
```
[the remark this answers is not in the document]
IanMichaelHarper: >We probaly used to have these before we removed FTX. Is it from there?

Yes I took it from the git history.

I just checked a handful of them and all of them were in the `common_asset_details` table in the globalDB. I wouldn't have thought they'd disappear unless a migration was run to remove them from the globalDB?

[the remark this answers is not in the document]
LefterisJP: > I just checked a handful of them and all of them were in the common_asset_details table in the globalDB. I wouldn't have thought they'd disappear unless a migration was run to remove them from the globalDB?

Yes no worries, if it's the one we had before they should be there. We don't remove assets.

rotkehlchen/data_import/importers/rotki_trades.py
IanMichaelHarper: I think that was overkill, but I updated updated the PR with it along with the other changes. To make the function common enough I had to replace ` else Fee(ZERO)` with `else None` on [this line](https://github.com/rotki/rotki/blob/7a55f9bdb7e42e883faa4a5a8acc82a1845e7ac3/rotkehlchen/data_import/importers/rotki_events.py#L52) but `else Fee(ZERO)` didn't make sense anyway because [it's only used if it's non-zero](https://github.com/rotki/rotki/blob/7a55f9bdb7e42e883faa4a5a8acc82a1845e7ac3/rotkehlchen/data_import/importers/rotki_events.py#L72), so I changed that line too to just `if fee:`

LefterisJP: Great. Thanks. It looks okay now!

LefterisJP (approved):
LGTM!

PR #6440, opened 2023-07-31
Autodetect some spam tokens

yabirgb:
There are common patterns around spam tokens symbols that can be used to filter many of them. This commit introduces a simple logic to detect those and early classify tokens as spam


rotkehlchen/assets/utils.py
```diff
@@ -238,6 +261,14 @@ def get_or_create_evm_token(
 
                 with userdb.user_write() as cursor:
                     userdb.add_asset_identifiers(cursor, [evm_token.identifier])
+            # at this point the asset is in the database
+            if is_spam_token:
+                with userdb.user_write() as write_cursor:
+                    if write_cursor.execute(
```
LefterisJP, line 267: Why do we need this if?

1. Has it not just been added to the DB? So either it happens at addition (at which case this is not needed) or this is always true. Perhaps we need a more well defined, "Add new token to the user and global DB" function.
2. I see that `add_to_ignored_assets` does not do INSERT OR IGNORE in which case this if is needed. Can't we change it? This extra logic is silly and not sure if we have it elsewhere too.

rotkehlchen/assets/utils.py
```diff
@@ -122,6 +129,20 @@ def _edit_token_and_clean_cache(
         GlobalDBHandler().edit_evm_token(evm_token)
 
 
+def test_if_is_spam_token(symbol: Optional[str]) -> bool:
+    """Makes basic checks to test if a token could be spam or not"""
+    if symbol is None:
+        return False
+    if (
```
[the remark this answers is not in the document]
yabirgb: yes but in the name I haven't seen such good patterns. They usually add this to the symbol because is what is rendered in almost every platform first and bigger

rotkehlchen/assets/utils.py
yabirgb: I have moved it to only addition and also changed the insert to insert or ignore

LefterisJP (approved):
LGTM

PR #6441, opened 2023-07-31
Remove saddle oracle

yabirgb:
Removes saddle oracle since it has no liquidity in almost all the pools


rotkehlchen/db/upgrades/v34_v35.py
```diff
@@ -373,7 +373,7 @@ def _add_defillama_to_oracles(cursor: 'DBCursor', setting_name: Literal['current
     oracle_list = default_oracles
     if price_oracles is not None:
         oracles = json.loads(price_oracles[0])
-        oracle_list = [oracle_cls.deserialize(oracle) for oracle in oracles]
+        oracle_list = [oracle_cls.deserialize(oracle) for oracle in oracles if oracle != 'saddle']  # noqa: E501  # saddle got removed in 1.30 but could be present in all versions
```
LefterisJP, line 376: Why? It's totally fine for saddle to be there in v35

rotkehlchen/db/upgrades/v38_v39.py
```diff
@@ -261,6 +262,20 @@ def _add_arbitrum_one_location_and_nodes(write_cursor: 'DBCursor') -> None:
     log.debug('Exit _add_arbitrum_one_location_and_nodes')
 
 
+def _remove_saddle_excahnge(write_cursor: 'DBCursor') -> None:
```
LefterisJP, line 265: excahnge -> exchange 

... but better yet. Oracle

```suggestion
def _remove_saddle_oracle(write_cursor: 'DBCursor') -> None:
```

Here and below

rotkehlchen/db/upgrades/v38_v39.py
```diff
@@ -261,6 +262,20 @@ def _add_arbitrum_one_location_and_nodes(write_cursor: 'DBCursor') -> None:
...
+    log.debug('Enter _remove_saddle_excahnge')
+    write_cursor.execute('SELECT value FROM settings WHERE name="current_price_oracles"')
```
LefterisJP, line 267: it was not in historical ones, right?

rotkehlchen/db/upgrades/v38_v39.py
```diff
@@ -261,6 +262,20 @@ def _add_arbitrum_one_location_and_nodes(write_cursor: 'DBCursor') -> None:
...
+    if (data := write_cursor.fetchone()) is None:
+        return  # oracles not configured
+    oracles: list[str] = json.loads(data[0])
```
LefterisJP, line 270: It should never happen here, but please since this is an upgrade and is critical code can you please handle json decoding error? 

In case it happens, log a critical error and proceed to remove the setting. I think this would be the only sensible option?

rotkehlchen/db/upgrades/v38_v39.py
```diff
@@ -261,6 +262,20 @@ def _add_arbitrum_one_location_and_nodes(write_cursor: 'DBCursor') -> None:
...
+    new_oracles = [oracle for oracle in oracles if oracle != 'saddle']
+    write_cursor.execute(
+        'INSERT OR REPLACE INTO settings(name, value) VALUES(?, ?)',
+        ('current_price_oracles', json.dumps(new_oracles)),
+    )
+    log.debug('Exit _remove_saddle_excahnge')
```
LefterisJP, line 276: change name in these logs too

rotkehlchen/oracles/structures.py
```diff
@@ -9,7 +9,6 @@ class CurrentPriceOracle(OracleSource):
     CRYPTOCOMPARE = auto()
     UNISWAPV2 = auto()
     UNISWAPV3 = auto()
-    SADDLE = auto()
     MANUALCURRENT = auto()
```
LefterisJP, line 12: Hmm ... I did not check but are we **absolutely 100% sure** that nothing depends on the order of these enums and that the value is just used in runtime only. They are not saved in the DB ad numerical values right?

Cause you are removing one and moving all the ones after it down by `-1` here

rotkehlchen/tests/data/v38_rotkehlchen.db
LefterisJP, line 1: Can you please give the diff here? I guess you added customized oracle order to include saddle for the d upgrade test, but would just like to see the sqldiff to be sure.

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -999,7 +1000,7 @@ def try_insert_mapping(cur):
         oracles_after_upgrade = cursor.execute(
             'SELECT value FROM settings WHERE name="current_price_oracles"',
         ).fetchone()[0]
-        assert oracles_after_upgrade == '["manualcurrent", "cryptocompare", "coingecko", "defillama", "uniswapv2", "uniswapv3", "saddle"]'  # noqa: E501
+        assert oracles_after_upgrade == '["manualcurrent", "cryptocompare", "coingecko", "defillama", "uniswapv2", "uniswapv3"]'  # noqa: E501
```
LefterisJP, line 1003: as I said in the post there is no need to remove saddle from upgrades before v38->v39

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1681,6 +1682,11 @@ def get_hashchainlog(entry):
 
     assert cursor.execute('SELECT MAX(seq) FROM location').fetchone()[0] == 40
 
+    # get settings and get that saddle is present
```
LefterisJP, line 1685:
```suggestion
    # get settings and confirm saddle is present before the upgrade
```

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1776,6 +1782,17 @@ def get_hashchainlog(entry):
         (SupportedBlockchain.ARBITRUM_ONE.value,),
     ).fetchall()) == len(DEFAULT_ARBITRUM_ONE_NODES_AT_V39)
 
+    # get current oracles and check that it deserializes correctly
```
LefterisJP, line 1785:
```suggestion
    # get current oracles and check that saddle was removed and all others remain as they were.
```

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1681,6 +1682,11 @@ def get_hashchainlog(entry):
 
     assert cursor.execute('SELECT MAX(seq) FROM location').fetchone()[0] == 40
 
+    # get settings and get that saddle is present
+    cursor.execute('SELECT value FROM settings WHERE name="current_price_oracles"')
+    assert 'saddle' in cursor.fetchone()[0]
```
LefterisJP, line 1687: Can you instead save the value of the old oracles list? And then later in the after-upgrade check, test that the list is same but excludes saddle?

rotkehlchen/db/upgrades/v34_v35.py
yabirgb: when we call deserialize saddle is not a valid option anymore

rotkehlchen/oracles/structures.py
LefterisJP: > Yes, I checked and is not needed. But this enum is the reason to remove saddle also in the old db upgrades. The reason is that the settings call deserialize on this class and that can lead to deserialization error if saddle is not removed or handled

That just means the old upgrade is built wrong, no? It should not rely on functions that can change

rotkehlchen/db/upgrades/v38_v39.py
yabirgb: No, only current

rotkehlchen/tests/data/v38_rotkehlchen.db
yabirgb:
```
INSERT INTO "settings" ("name","value") VALUES ('current_price_oracles','["manualcurrent", "defillama", "cryptocompare", "coingecko", "uniswapv2", "uniswapv3", "saddle"]');
```

LefterisJP (approved):
LGTM

Issue #6444, opened 2023-08-01
Show asset graph in asset collection view


LefterisJP:
## Problem Definition

Normally in the asset view we show the asset graph. But if the asset is part of a multi-chain (or other) collection the asset view shows the breakdown nicely, but does not show the graph

![2023-08-01_12-09](https://github.com/rotki/rotki/assets/1658405/9f20eb67-7592-4d02-ab29-16c7693cdc45)

We should definitely show a graph there too. The reason we currently don't is that it's not possible to ask for the graph of an asset collection from the backend. So this is a frontend and backend problem.

But it's solvable and we should have it implemented.

## Task

- **Backend:** The asset graph data endpoint should accept an asset collection. If a collection is given then the asset data of collection members should be summed up for the timestamps where multiple assets of the collection have snapshot value and returned.
- **Frontend:** If the asset view is of a collection, adjust the code accordingly to make it work after the backend work.

PR #6447, opened 2023-08-02
Backend for statistics balance snapshots of collections

LefterisJP:
This is the backend part of #6444


LefterisJP:
All green. Will merge.

PR #6458, opened 2023-08-05
add setuptools-scm in requirements_dev.txt

chamalis:
used in package.py


requirements_dev.txt
```diff
@@ -13,6 +13,7 @@ pytest-vcr==1.0.2
 vcrpy==4.3.1
 freezegun==1.2.2
 flaky==3.7.0
+setuptools-scm==7.1.0  # needed by package.py
```
LefterisJP, line 16: hey hello, thanks `requirements_dev` is only for development. For packaging we also need these 3 packages:

```
packaging wheel setuptools-scm
```

Which we install during the packaging build job (if you check the github actions) only for that vm, or locally if we build locally.

What may make sense in that case is another requirements_packaging.txt to specify those in one common place?

What do you think @kelsos ?

chamalis: Hi, thanks for your comment.
My bad, I missed the workflows ...:(

kelsos: > hey hello, thanks `requirements_dev` is only for development. For packaging we also need these 3 packages:
> 
> ```
> packaging wheel setuptools-scm
> ```
> 
> Which we install during the packaging build job (if you check the github actions) only for that vm, or locally if we build locally.
> 
> What may make sense in that case is another requirements_packaging.txt to specify those in one common place?
> 
> What do you think @kelsos ?

Yes having a requirements_packaging.txt would make sense.

LefterisJP:
Hey so @chamalis would you like to do this?

Essentially create a new requirements file called `requirements_packaging.txt ` and also call it from inside the github actions packaging code instead of the direct pip install.

chamalis:
Sure, making a new PR, right?

LefterisJP:
> Sure, making a new PR, right?

In general we prefer to just use the same PR. No need to close and reopen.

chamalis:
My bad. I accidentally closed it by pushing a hard reset in my fork for the wrong commit, and don''t have permission to reopen

PR #6461, opened 2023-08-07
Fix debug export of PnL report

yabirgb:
For the debug export we were retrieving the ignored events that is a mapping of str to set and set is not json serializable. Since we weren't converting to a serializable object the report couldn't be exported.

### DIFF
--- a/rotkehlchen/tests/api/test_history.py
+++ b/rotkehlchen/tests/api/test_history.py
@@ -2,6 +2,7 @@ import random
 from contextlib import ExitStack
 from http import HTTPStatus
 from pathlib import Path
+from typing import TYPE_CHECKING
 from unittest.mock import MagicMock, patch
 
 import pytest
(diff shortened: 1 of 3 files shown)


rotkehlchen/tests/api/test_history.py
```diff
@@ -337,9 +344,19 @@ def test_query_pnl_report_events_pagination_filtering(
             assert x['timestamp'] >= events[idx + 1]['timestamp']
 
 
-@pytest.mark.parametrize('ethereum_accounts', [])
-def test_history_debug_export(rotkehlchen_api_server):
+@pytest.mark.parametrize('ethereum_accounts', [[]])
+@pytest.mark.parametrize('have_decoders', [[True]])
+def test_history_debug_export(rotkehlchen_api_server: 'APIServer') -> None:
```
LefterisJP, line 349: So this test was not running. Cool. We need to check if more tests are skipped due to this or any other weird unexpected conditions.

rotkehlchen/tests/api/test_history.py
```diff
@@ -337,9 +344,19 @@ def test_query_pnl_report_events_pagination_filtering(
...
     """Check that the format of the data exported matches the expected type."""
+    tx_id = '10' + str(make_evm_tx_hash())
```
LefterisJP, line 351: Why are you ignoring a random optimism transaction? Perhaps add a comment next to this that you want to check it later?

yabirgb: The set that couldn't be json serialised came from the ignored events. What I did was create a event, add it to the list of ignored ones so when querying the endpoint the key is populated with such set

PR #6466, opened 2023-08-07
Filter for customized history events

### LINKED ISSUE #6439: Enable filter to only show manually modified history events
## Abstract

I would like to have a filter in the history events that allows me to only view manually modified transactions. This will enable to review and possibly redecode these transactions.

## Motivation

There are always EVM transactions that can not be properly decoded by rotki. For correct PNL report these transactions can be manually modified.
Having a filter to only show modified entries is usefuel:
1. I can manually review all modified transactions to make sure I did not do a mistake
2. I can check if a newer rotki release has added native support, so i can redecode a transaction and get rid of the manual changes.

## Specification

Add a filter option "Only show modified EVM transactions". This filter shall be combinable with existing filters (like asset, address, location, ...)

At a later stage a "diff-view" could be added where you can compare the history event in it's current state with how rotki would decode the transaction.

TripleSpeeder:
Closes #6439

I think the backend changes are ready for review, but I'm not happy with the UI. Current implementation of history filters does not work well with this type of boolean params.

Also not yet looked into testing and usage guide updates.

## Checklist


lukicenturi:
Moved to https://github.com/rotki/rotki/pull/6878

PR #6469, opened 2023-08-08
Bump VCR version

yabirgb:
In the old version there seems to be a bug that was leaking requests to the urllib module instead of using the mocked version


LefterisJP (approved):
LGTM

PR #6470, opened 2023-08-08
Update convex pools

yabirgb:
Updated them now for the release but we need to automate this, we already have the test that does this. I believe the blocker is or was the cache PR


LefterisJP:
The cache PR has still not been done, yes

LefterisJP (approved):
LGTM

PR #6471, opened 2023-08-08
Asset collections graph data should now be sorted and combined

LefterisJP:
Also added a unit test for this

### DIFF
--- a/rotkehlchen/db/dbhandler.py
+++ b/rotkehlchen/db/dbhandler.py
@@ -2220,6 +2220,7 @@ class DBHandler:
                     to_ts=to_ts,
                 ))
 
+        asset_balances.sort(key=lambda x: x.time)
         return combine_asset_balances(asset_balances)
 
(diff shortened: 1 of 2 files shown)


rotkehlchen/db/dbhandler.py
```diff
@@ -2220,6 +2220,7 @@ def query_collection_timed_balances(
                     to_ts=to_ts,
                 ))
 
+        asset_balances.sort(key=lambda x: x.time)
```
LefterisJP, line 2223: this is the one line fix. Rest is to test

yabirgb (approved):
lgtm

LefterisJP:
Seems that `rotkehlchen/tests/api/test_bitcoin.py::test_add_xpub_with_conversion_works` fails quite often lately. We should find out why.

LefterisJP:
Anyway will merge, and may look at it later

PR #6474, opened 2023-08-08
add diva airdrop and fix shapeshift key

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/ethereum/airdrops.py
+++ b/rotkehlchen/chain/ethereum/airdrops.py
@@ -15,6 +15,7 @@ from rotkehlchen.constants.assets import (
     A_CORN,
     A_CRV,
     A_CVX,
+    A_DIVA,
     A_ENS,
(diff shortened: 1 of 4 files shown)


rotkehlchen/chain/ethereum/airdrops.py
```diff
@@ -89,7 +90,7 @@
         A_CVX,
         'https://www.convexfinance.com/',
     ),
-    'fox': (
+    'shapeshift': (
```
LefterisJP, line 93: so this change is somehow messing with the icon on the airdrops page. The DIVA one works fine.

LefterisJP (approved):
LGTM. But also needs a changelog. If all passes we can merge, but if you see this not merged, just add a changelog entry too. Perhaps just combine with the other DIVA one.

If you see it after merging, then add it separately.

PR #6477, opened 2023-08-09
get tx ensures requirements

dimyG:
Closes #(issue_number)

## Checklist


rotkehlchen/db/evmtx.py
```diff
@@ -154,6 +155,67 @@ def get_evm_internal_transactions(
...
+        This function can raise:
+        - pysqlcipher3.dbapi2.OperationalError if the SQL query fails due to invalid
+        filtering arguments.
+        - RemoteError if there is a problem querying the data source.
+        - DeserializationError if the transaction cannot be deserialized from the DB.
+        """
+        self.assert_tx_data_is_pulled(cursor, evm_inquirer, tx_hash, relevant_address)
+        query, bindings = EvmTransactionsFilterQuery.make(tx_hash=tx_hash, chain_id=evm_inquirer.chain_id).prepare()  # noqa: E501
+        query, bindings = self._form_evm_transaction_dbquery(query, bindings, has_premium=True)
+        result = cursor.execute(query, bindings).fetchone()
```
LefterisJP, line 215: What is result is None? I guess it should never be `None` due to `self.assert_tx_data_is_pulled()` above? Perhaps write it in a comment?

rotkehlchen/db/evmtx.py
```diff
@@ -154,6 +155,67 @@ def get_evm_internal_transactions(
...
+            cursor: 'DBCursor',
+            evm_inquirer: 'EvmNodeInquirer',
+            tx_hash: 'EVMTxHash',
+            relevant_address: Optional['ChecksumEvmAddress'],
+    ) -> None:
+        """Makes sure that the receipt of the given transaction is in the database.
+        If not, pulls it and stores it in the database.
+
+        May raise:
+        - RemoteError if there is a problem querying the data source.
```
LefterisJP, line 169:
```suggestion
        - RemoteError if there is a problem querying the data sources or transaction hash does not exist.
```

rotkehlchen/db/evmtx.py
```diff
@@ -154,6 +155,67 @@ def get_evm_internal_transactions(
...
+        )
+        self.add_receipt_data(
+            write_cursor=cursor,
+            chain_id=evm_inquirer.chain_id,
+            data=raw_receipt_data,
+            )
+
+    def get_or_create_transaction(
+            self,
+            cursor: 'DBCursor',
```
LefterisJP, line 196: If this is a write cursor then specify it in the name. We rely on the name being like this to know what type of cursor it is
```suggestion
            write_cursor: 'DBCursor',
```

rotkehlchen/db/evmtx.py
```diff
@@ -154,6 +155,67 @@ def get_evm_internal_transactions(
...
+            evm_transactions=[transaction],
+            relevant_address=relevant_address,
+        )
+        self.add_receipt_data(
+            write_cursor=cursor,
+            chain_id=evm_inquirer.chain_id,
+            data=raw_receipt_data,
+            )
+
+    def get_or_create_transaction(
```
LefterisJP, line 194: Okay so I left you a comment about the cursor name being changed to write_cursor since you pass a writing one but that is a problem.

You should not pass a write_cursor here as passing a write_cursor forces you to have one open for long. If you recall from the chat we had that is a bad thing. Haning an open Database transaction for long can lead to errors and deadlocks.

What you should do instead is:
- Pass a read cursor, and use it to read from the DB
- If needed pull the transaction/receipt data from the remote (this is what may take long - as it's a remote call). You absolutely **do not** want an open write cursor, thus an open DB transaction while a remote query is happening
- Once done and you have the data, then and only then open a write cursor, which opens a DB transaction to write it.

If you need to do multiple calls, then gather the data in memory after each call and then write them all together under a single write_cursor once you are done.

rotkehlchen/db/utils.py
```diff
@@ -430,6 +431,9 @@ def db_tuple_to_str(
             f'for {data[5]} closed at timestamp {data[3]}'
         )
 
+    if tuple_type == 'optimism_transaction':
```
LefterisJP, line 434: Do we need this extra information? If it's just for debugging and printing then just don't do it. KEep it all as evm transaction.

rotkehlchen/db/optimismtx.py
```diff
@@ -70,6 +107,6 @@ def _build_evm_transaction(self, result: tuple[Any, ...]) -> OptimismTransaction
             gas_used=int(result[9]),
             input_data=result[10],
             nonce=result[11],
-            l1_fee=0 if result[12] is None else int(result[12]),
+            l1_fee=0 if result[12] is None else int(result[12]),  # this check is only needed when _build_evm_transaction is called from the get_evm_transactions method.  # noqa: E501
```
LefterisJP, line 110: Since this may change in the future have a more specific comment

```suggestion
            l1_fee=0 if result[12] is None else int(result[12]),  # this check is only needed when _build_evm_transaction is called from a code path that does not call assert_tx_data_is_pulled().  # noqa: E501
```

rotkehlchen/tests/db/test_evmtx.py
```diff
@@ -128,6 +131,65 @@ def test_add_get_evm_transactions(data_dir, username, sql_vm_instructions_cb):
         assert result == [tx1, tx3]
 
 
+@pytest.mark.parametrize('ethereum_manager_connect_at_start', [(INFURA_ETH_NODE,)])
+def test_get_or_create_transaction(ethereum_inquirer):
+    """Tests that get_or_create_transaction works fine. By testing get_or_create_transaction 
+    it checks that the requirements for a transaction are met before returning it."""
+    msg_aggregator = MessagesAggregator()
+    tx_hash = deserialize_evm_tx_hash(b'.h\xdd\x82\x85\x94\xeaq\xfe\n\xfc\xcf\xadwH\xc9\x0f\xfc\xd0\xf1\xad\xd4M\r$\x9b\xf7\x98\x87\xda\x93\x18')  # noqa: E501
+    with ethereum_inquirer.database.conn.write_ctx() as cursor:
```
LefterisJP, line 140: no need to do this. Use the `ethereum_accounts` fixture to populate those

rotkehlchen/tests/db/test_evmtx.py
```diff
@@ -128,6 +131,65 @@ def test_add_get_evm_transactions(data_dir, username, sql_vm_instructions_cb):
...
+    )
+
+    with ethereum_inquirer.database.conn.write_ctx() as cursor:
+        dbevmtx = DBEvmTx(ethereum_inquirer.database)
+        # check that the transaction is properly added to the DB
+        returned_tx = dbevmtx.get_or_create_transaction(cursor, ethereum_inquirer, tx_hash, relevant_address=ETH_ADDRESS3)  # noqa: E501
+        errors = msg_aggregator.consume_errors()
+        warnings = msg_aggregator.consume_warnings()
+        assert len(errors) == 0
+        assert len(warnings) == 0
```
LefterisJP, line 171: shorten here and below

```suggestion
        assert(len(msg_aggregator.consume_errors()) == 0
        assert(len(msg_aggregator.consume_warnings()) == 0
```

rotkehlchen/tests/db/test_evmtx.py
```diff
@@ -128,6 +131,65 @@ def test_add_get_evm_transactions(data_dir, username, sql_vm_instructions_cb):
...
+    with ethereum_inquirer.database.conn.write_ctx() as cursor:
+        ethereum_inquirer.database.add_blockchain_accounts(
+            write_cursor=cursor,
+            account_data=[
+                BlockchainAccountData(chain=SupportedBlockchain.ETHEREUM, address=ETH_ADDRESS1),
+                BlockchainAccountData(chain=SupportedBlockchain.ETHEREUM, address=ETH_ADDRESS2),
+                BlockchainAccountData(chain=SupportedBlockchain.ETHEREUM, address=ETH_ADDRESS3),
+            ],
+        )
+    tx = EvmTransaction(
```
LefterisJP, line 149:
```suggestion
    expected_transaction = EvmTransaction(
```

rotkehlchen/tests/db/test_evmtx.py
```diff
@@ -128,6 +131,65 @@ def test_add_get_evm_transactions(data_dir, username, sql_vm_instructions_cb):
         assert result == [tx1, tx3]
 
 
+@pytest.mark.parametrize('ethereum_manager_connect_at_start', [(INFURA_ETH_NODE,)])
+def test_get_or_create_transaction(ethereum_inquirer):
```
LefterisJP, line 135: thsi test needs to eventually be VCRed

rotkehlchen/tests/unit/decoders/test_optimism_bridge.py
```diff
@@ -100,7 +100,7 @@ def test_deposit_eth(database, ethereum_inquirer, ethereum_accounts):
     ]
 
 
-@pytest.mark.vcr()
+# @pytest.mark.vcr()
```
LefterisJP, line 103: Why all these removing of VCR here? Changes order of queries eh?

dimyG: hmm, after the review changes there is no need for new cassetes.

rotkehlchen/chain/evm/transactions.py
```diff
@@ -264,31 +264,9 @@ def _query_and_save_internal_transactions_for_range_or_parent_hash(
...
-                        self.dbevmtx.add_receipt_data(
-                            write_cursor=write_cursor,
-                            chain_id=self.evm_inquirer.chain_id,
-                            data=raw_receipt_data,
-                        )
-                    timestamp = transaction.timestamp
-                else:
-                    timestamp = result[0].timestamp
+                with self.database.conn.read_ctx() as read_cursor:
+                    result = self.dbevmtx.get_or_create_transaction(read_cursor, self.evm_inquirer, internal_tx.parent_tx_hash, address)  # noqa: E501
```
LefterisJP, line 268: didn't I tell you to use kwargs?

rotkehlchen/chain/evm/transactions.py
```diff
@@ -393,29 +371,9 @@ def _get_erc20_transfers_for_ranges(
...
-                                    write_cursor=write_cursor,
-                                    chain_id=self.evm_inquirer.chain_id,
-                                    data=raw_receipt_data,
-                                )
-                            timestamp = transaction.timestamp
-                        else:
-                            timestamp = result[0].timestamp
-
+                        with self.database.conn.read_ctx() as read_cursor:
+                            result = self.dbevmtx.get_or_create_transaction(read_cursor, self.evm_inquirer, tx_hash, address)  # noqa: E501
```
LefterisJP, line 375: kwargs

rotkehlchen/chain/evm/transactions.py
```diff
@@ -264,31 +264,9 @@ def _query_and_save_internal_transactions_for_range_or_parent_hash(
...
-                        )
-                        self.dbevmtx.add_receipt_data(
-                            write_cursor=write_cursor,
-                            chain_id=self.evm_inquirer.chain_id,
-                            data=raw_receipt_data,
-                        )
-                    timestamp = transaction.timestamp
-                else:
-                    timestamp = result[0].timestamp
+                with self.database.conn.read_ctx() as read_cursor:
```
LefterisJP, line 267: no need
```suggestion
                with self.database.conn.read_ctx() as cursor:
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -555,20 +555,25 @@ def decode_transaction_hashes(
...
-            except RemoteError as e:
-                raise InputError(f'{self.evm_inquirer.chain_name} hash {tx_hash.hex()} does not correspond to a transaction') from e  # noqa: E501
-
             # TODO: Change this if transaction filter query can accept multiple hashes
-            with self.database.conn.read_ctx() as cursor:
-                txs = self.dbevmtx.get_evm_transactions(
-                    cursor=cursor,
-                    filter_=EvmTransactionsFilterQuery.make(tx_hash=tx_hash, chain_id=self.evm_inquirer.chain_id),  # noqa: E501
-                    has_premium=True,  # ignore limiting here
+            with self.database.conn.read_ctx() as read_cursor:
```
LefterisJP, line 559: no need

```suggestion
            with self.database.conn.read_ctx() as cursor:
```

rotkehlchen/chain/evm/transactions.py
```diff
@@ -534,12 +492,14 @@ def get_or_query_transaction_receipt(self, tx_hash: EVMTxHash) -> 'EvmTxReceipt'
...
         - RemoteError if the transaction hash can't be found in any of the connected nodes
         """
-        with self.database.conn.read_ctx() as cursor:
+        with self.database.conn.read_ctx() as read_cursor:
             # If the transaction is not in the DB then query it and add it
-            result = self.dbevmtx.get_evm_transactions(
-                cursor=cursor,
-                filter_=EvmTransactionsFilterQuery.make(tx_hash=tx_hash, chain_id=self.evm_inquirer.chain_id),  # noqa: E501
-                has_premium=True,  # we don't need any limiting here
+            transaction = self.dbevmtx.get_or_create_transaction(read_cursor, self.evm_inquirer, tx_hash, relevant_address=None)  # noqa: E501
```
LefterisJP, line 497: kwargs

```suggestion
            transaction = self.dbevmtx.get_or_create_transaction(read_cursor, self.evm_inquirer, tx_hash, relevant_address=None)  # noqa: E501
```

rotkehlchen/tests/data/data_migration_v11.db
LefterisJP, line 1: What happend here? Why did you change the DB?

LefterisJP (commented):
Left a review.A lot of tests failed. I suspect many just need re-recording. Please check all failures

rotkehlchen/tests/unit/decoders/test_optimism_bridge.py
LefterisJP: I think there is

rotkehlchen/tests/data/data_migration_v11.db
dimyG: This was used by the test_migration_11 for checking the optimism transactions related stuff. It isn't needed anymore.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -555,20 +555,25 @@ def decode_transaction_hashes(
...
+                )
+                receipt = self.dbevmtx.get_receipt(
+                    cursor=read_cursor,
+                    tx_hash=tx_hash,
+                    chain_id=self.evm_inquirer.chain_id,
+                )
+            if receipt is None:
+                raise DeserializationError(
+                    f'{self.evm_inquirer.chain_name} hash {tx_hash.hex()} does not have a corresponding receipt in the database',  # noqa: E501
                 )
```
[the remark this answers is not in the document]
dimyG: The receipt is needed by the _get_or_decode_transaction_events which is called next. The check is indeed not needed becasue the recipt will not be None but mypy compains. I can just type ignore the warning instead.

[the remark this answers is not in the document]
LefterisJP: Ofc you should type ignore the comment. We don't add extra runtime code just to make mypy happy. Explain why it should exist there.

What's more since get_or_create_transaction also returns the receipt internally just make it return tuple of transaction/receipt.

Also make `assert_tx_data_is_pulled` return them too. I just noticed that in `get_or_create_transaction` you are actually requerying the DB to get the transaction which is silly

rotkehlchen/db/utils.py
```diff
@@ -432,6 +432,8 @@ def db_tuple_to_str(
 
     # else can only be evm transaction
     assert tuple_type == 'evm_transaction', 'only DBTupleType possible here is evm_transaction'
+    if isinstance(data[0], int):
+        return f'Optimism transaction with hash 0x{data[1].hex()} and l1_fee {data[0]}'
```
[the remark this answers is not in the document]
dimyG: Ok, but if the flow ends up here without it, it would raise an exception "int has no method hex()".

rotkehlchen/chain/optimism/transactions.py
```diff
@@ -7,6 +7,7 @@
 
 if TYPE_CHECKING:
     from rotkehlchen.db.dbhandler import DBHandler
+
```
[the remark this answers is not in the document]
dimyG: This is from isort (probably because it is a relative import)

rotkehlchen/db/utils.py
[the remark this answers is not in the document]
dimyG: The query used for optimism is different and needs the l1_fee as the first argument of the tuple. I'm not sure if the query can be changed so that tx_hash is first (or even if it's worth it). Maybe a new tuple_type makes sense since here we add the optimism_transaction in it's own table instead of a common evm_transaction?

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -50,6 +51,7 @@
 )
 from rotkehlchen.utils.mixins.customizable_date import CustomizableDateMixin
 
+from ..constants import GENESIS_HASH
```
LefterisJP, line 54: never relative import if not from same module

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -555,20 +557,58 @@ def decode_transaction_hashes(
...
-            except RemoteError as e:
-                raise InputError(f'{self.evm_inquirer.chain_name} hash {tx_hash.hex()} does not correspond to a transaction') from e  # noqa: E501
-
-            # TODO: Change this if transaction filter query can accept multiple hashes
-            with self.database.conn.read_ctx() as cursor:
-                txs = self.dbevmtx.get_evm_transactions(
-                    cursor=cursor,
-                    filter_=EvmTransactionsFilterQuery.make(tx_hash=tx_hash, chain_id=self.evm_inquirer.chain_id),  # noqa: E501
-                    has_premium=True,  # ignore limiting here
+            if tx_hash == GENESIS_HASH:
```
LefterisJP, line 560: wtf? From line 560 to line 596 you just copied paste the genesis tx hash code from `get_or_query_transaction_receipt()`. In what universe do you think this is acceptable?

 This needs to be abstracted into its own function and called in both places properly.

Code that does such sloppy copy pasta would never be accepted into the repo. Don't ask for review with code like this.

rotkehlchen/chain/optimism/transactions.py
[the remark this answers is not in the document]
LefterisJP: Even if it is from isort this file **does not need to be touched**. Undo it. No isort configuration changed.

If this file needed to change due to our isort config all linting runs of the CI in develop would fail.

 If it's your own local editor that does this then configure your editor.

rotkehlchen/db/evmtx.py
```diff
@@ -154,6 +155,72 @@ def get_evm_internal_transactions(
...
+        If not, pulls them and stores them. For most chains this is the transaction and the
+        receipt. Can be extended by subclasses for chain-specific information.
+
+        May raise:
+        - RemoteError if there is a problem querying the data sources or transaction hash does
+        not exist.
+        """
+        query, bindings = EvmTransactionsFilterQuery.make(tx_hash=tx_hash, chain_id=evm_inquirer.chain_id).prepare()  # noqa: E501
+        query, bindings = self._form_evm_transaction_dbquery(query=query, bindings=bindings, has_premium=True)  # noqa: E501
+        tx_data = cursor.execute(query, bindings).fetchone()
```
LefterisJP, line 175: kwargs in all of these functions. Please do not make me repeat things across multiple reviews

rotkehlchen/tests/unit/test_evm_tx_decoding.py
```diff
@@ -214,12 +214,13 @@ def test_query_and_decode_transactions_works_with_different_chains(
     assert len(hashes) == 1
 
 
-@pytest.mark.vcr()
+# @pytest.mark.vcr()
```
LefterisJP, line 217: Ehm ... you have commented out code here, specifically a VCR marker

LefterisJP (commented):
Still needs some work. Also not sure how the tests will do.

rotkehlchen/db/utils.py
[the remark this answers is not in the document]
LefterisJP: Didn't you change this? I saw that you changed the order as I asked, didn't you?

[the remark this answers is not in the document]
LefterisJP: Another reason this **has** to be done as I said is that you would break this logic by what you did. https://github.com/rotki/rotki/blob/a51be249f1f38cf2b28e0bf28ebdfbcb33844dcc/rotkehlchen/db/dbhandler.py#L1554

The code for relevant address detection expects either evm transaction or internal tx. So accidentally without this change all relevant address for optimism transactions would take wrong transaction hash

rotkehlchen/db/utils.py
```diff
@@ -432,6 +432,7 @@ def db_tuple_to_str(
 
     # else can only be evm transaction
     assert tuple_type == 'evm_transaction', 'only DBTupleType possible here is evm_transaction'
+
```
LefterisJP, line 435: Stupid lefty, remove unneeded newline change if anything needs to change.

rotkehlchen/tests/db/test_evmtx.py
```diff
@@ -1,3 +1,4 @@
+
```
LefterisJP, line 1: same here lefty. Probably `ruff --fix` did this

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -739,6 +739,7 @@ def test_reshuffle(ordered_events, events_list, result_list, msg='events_should_
     '0xA1E4380A3B1f749673E270229993eE55F35663b4',
     '0x756F45E3FA69347A9A973A725E3C98bC4db0b5a0',
 ]])
+@pytest.mark.parametrize('ethereum_manager_connect_at_start', ['DEFAULT'])
```
LefterisJP, line 742: I think we wll have to change this. I already had to change one test and re-record due to this. By giving `DEFAULT` as the nodes to use, the test's remote calls are no longer deterministic. If VCR is used, we can only use one node to connect per chain.

rotkehlchen/tests/unit/decoders/test_main.py
```diff
@@ -796,6 +797,7 @@ def test_genesis_transaction(database, ethereum_inquirer, ethereum_accounts):
 
 
 @pytest.mark.vcr()
+@pytest.mark.parametrize('ethereum_manager_connect_at_start', ['DEFAULT'])
```
LefterisJP, line 800: same here

rotkehlchen/tests/unit/test_optimism_transactions.py
```diff
@@ -49,3 +51,9 @@ def assert_tx_okay(transactions, should_have_l1):
             has_premium=True,
         )
         assert_tx_okay(transactions, should_have_l1=False)
+
+    # check that the get_or_create_transaction method makes sure that the requirements for an
+    # optimism transaction (l1_fee existing in the database) are met before returning it.
+    with optimism_transactions.database.conn.read_ctx() as cursor:
+        tx, _ = dbevmtx.get_or_create_transaction(cursor, optimism_inquirer, tx_hash, string_to_evm_address(to_address))  # noqa: E501
```
LefterisJP, line 58: hoooo I think this will go boom boom. Moved `get_or_create_transaction` to `EVMTransactions`. Linting passed which means linting skips this test file. Will need to fix.

LefterisJP (commented):
Leaving myself a review, since I already see things I need to change by reading the diff after my changes.

LefterisJP (approved):
All green. LGTM. Merging.

PR #6484, opened 2023-08-10
Fix compound decoder

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/compound/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/compound/decoder.py
@@ -1,9 +1,10 @@
 import logging
-from typing import TYPE_CHECKING, Any, Optional, cast
+from typing import TYPE_CHECKING, Any, Optional
 
 from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.assets.asset import EvmToken
-from rotkehlchen.assets.utils import get_crypto_asset_by_symbol, get_or_create_evm_token
+from rotkehlchen.assets.utils import get_or_create_evm_token
+from rotkehlchen.chain.ethereum.modules.compound.utils import get_compound_underlying_token
 from rotkehlchen.chain.ethereum.utils import asset_normalized_value, token_normalized_value
 from rotkehlchen.chain.evm.decoding.constants import ERC20_OR_ERC721_TRANSFER
 from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface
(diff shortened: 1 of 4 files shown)


rotkehlchen/chain/ethereum/modules/compound/decoder.py
```diff
@@ -132,24 +133,23 @@ def _decode_redeem(
...
         redeem_tokens_raw = hex_or_bytes_to_int(tx_log.data[64:96])
-        underlying_token = get_crypto_asset_by_symbol(
-            symbol=compound_token.symbol[1:],
-            chain_id=compound_token.chain_id,
-        )
-        if underlying_token is None:
+        try:
+            underlying_asset = get_compound_underlying_token(compound_token)
+        except UnknownAsset:
+            return DEFAULT_DECODING_OUTPUT
```
LefterisJP, line 139: same

rotkehlchen/chain/ethereum/modules/compound/decoder.py
```diff
@@ -317,7 +315,7 @@ def decode_compound_token_movement(
             compound_token: EvmToken,
     ) -> DecodingOutput:
         if context.tx_log.topics[0] == MINT_COMPOUND_TOKEN:
-            log.debug(f'Hash: {context.transaction.tx_hash.hex()}')
+            log.debug(f'Decoding mint: {context.transaction.tx_hash.hex()}')
```
LefterisJP, line 318: This feels like a forgotten debug log. Perhaps just remove?

rotkehlchen/chain/ethereum/modules/compound/utils.py
```diff
@@ -0,0 +1,17 @@
+from typing import Optional
+from rotkehlchen.assets.asset import CryptoAsset, EvmToken
+from rotkehlchen.chain.ethereum.modules.compound.constants import CTOKEN_MAPPING
+
+
+def get_compound_underlying_token(token: EvmToken) -> Optional[CryptoAsset]:
+    """
+    Return the underlying token for a compound token. If the provide token is
```
LefterisJP, line 8:
```suggestion
    Returns the underlying token for a compound token. If the provided token is
```

rotkehlchen/chain/ethereum/modules/compound/utils.py
```diff
@@ -0,0 +1,17 @@
...
+    unknown it returns None.
+
+    May raise:
+    - UnknownAsset: If the underlying asset can't be resolved
```
LefterisJP, line 12: Hm .... I don't think this is a good interface.

You either:
- Always return and in case of error, raise
- Do not raise, but in case of any error return None.

Choose one. Why? In the 3 places you use it you also end up having 3 extra if checks. So you check both for exception and if it's None. Makes sense?

LefterisJP (approved):
LGTM. Thanks

PR #6489, opened 2023-08-11
If node_inquier._query raises TxNotFound and must exist keep looking

LefterisJP:
Before we were stopping and raising RemoteError, which would make sense if it was inside a single query, so that a next query can happen, but this was inside the for loop, so no other nodes were tried and the entire thing failed.

### DIFF
--- a/rotkehlchen/chain/evm/node_inquirer.py
+++ b/rotkehlchen/chain/evm/node_inquirer.py
@@ -529,10 +529,9 @@ class EvmNodeInquirer(metaclass=ABCMeta):
                 log.warning(f'Failed to query {node_info} for {method!s} due to {e!s}')
                 # Catch all possible errors here and just try next node call
                 continue
-            except TransactionNotFound as e:
+            except TransactionNotFound:
                 if kwargs.get('must_exist', False) is True:
-                    tx_hash = kwargs.get('tx_hash', GENESIS_HASH)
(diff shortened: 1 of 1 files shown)


rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -529,10 +529,9 @@ def _query(self, method: Callable, call_order: Sequence[WeightedNode], **kwargs:
                 log.warning(f'Failed to query {node_info} for {method!s} due to {e!s}')
                 # Catch all possible errors here and just try next node call
                 continue
-            except TransactionNotFound as e:
+            except TransactionNotFound:
                 if kwargs.get('must_exist', False) is True:
-                    tx_hash = kwargs.get('tx_hash', GENESIS_HASH)
-                    raise RemoteError(f'{self.chain_name} transaction call should exist for {tx_hash.hex()}') from e  # noqa: E501
+                    continue  # try other nodes, as transaction has to exist
```
yabirgb, line 534: what happens if for some reason the user has only one node and it fails to find the transaction? 

1. Reading the code seems that the returned variable `result` could not be defined if query fails and nodes are exhausted
2. Shouldn't you check at the end of the for loop if result got a value and otherwise raise Remoterror?
3. If remoteerror is not raised the docstring needs to be updated

yabirgb (approved):
LGTM

LefterisJP:
the failure is not related but it's related to this VCR playback when non-deterministic

PR #6490, opened 2023-08-11
Etherscan queries will now backoff for 429 responses too

LefterisJP:
Before they were only backing off in the special case, that I am not even sure is required anymore, where the url contained the error "Max retries exceeded with url".

Now we make it backoff either when that happens or when a 429 is returned.


LefterisJP:
failed test is timeout and passed locally

yabirgb (approved):
lgtm

PR #6491, opened 2023-08-11
Fix arbitrum one bridge token withdrawals

LefterisJP:
The assumption that all token gateways have the same ABI was wrong. Function parameters are nice and fine. What is wrong is that the name differs and is comprised of the symbol of the respective L1 token.

Which is absolutely ridiculous as a design.

Also may be a wrong assumption itself and we need a different way. May need more research.

**Edit:** It was a wrong assumption and can break for some tokens. Thanks to Daniel here, we just use the L2 Gateway router itself to get the info we need irrespective of token or its gateway. https://twitter.com/DZack23/status/1690075661954678786

Kudos!


rotkehlchen/tests/unit/decoders/test_arbitrum_one_bridge.py
```diff
@@ -302,6 +303,50 @@ def test_withdraw_erc20_from_arbitrum_one_to_ethereum(database, arbitrum_one_inq
     ]
 
 
+@pytest.mark.vcr()
+@pytest.mark.parametrize('arbitrum_one_accounts', [['0x9531C059098e3d194fF87FebB587aB07B30B1306']])
+def test_withdraw_dai_from_arbitrum_one_to_ethereum(database, arbitrum_one_inquirer, arbitrum_one_accounts):  # noqa: E501
+    """
+    Test that DAO withdrawals from arbitrum to L1 work fine. This is just to test that
```
LefterisJP, line 310: still thinking of DAOs huh?

```suggestion
    Test that DAI withdrawals from arbitrum to L1 work fine. This is just to test that
```

LefterisJP:
All green. Will merge.

PR #6492, opened 2023-08-12
Staking events from cointracking CSV can now be properly imported 

LefterisJP:
- [Staking events from cointracking CSV can now be properly imported](https://github.com/rotki/rotki/commit/1deafd677c3bcb9ebc8deaae732eb17d8ff5bd32)

- [Rename cursor to write_cursor in data import to underline difference](https://github.com/rotki/rotki/commit/1493b3eef4d489bc3b18419596894c25922d9438) 

It's really important to understand we have a write_cursor and
transaction open in the DB for this entire time, which is also
probably not the best idea


LefterisJP:
All green. That's nice. Merging

PR #6493, opened 2023-08-13
Bump linting

LefterisJP:


rotkehlchen/accounting/structures/base.py
```diff
@@ -119,22 +119,22 @@ def __init__(
         self.notes = notes
         self.identifier = identifier
 
-    def __eq__(self, other: Any) -> bool:
+    def __eq__(self, other: object) -> bool:
```
yabirgb, line 122: shouldn't the type here be

```suggestion
    def __eq__(self, other: HistoryBaseEntry) -> bool:
```

also similar thing in the places where the same pattern has changed

rotkehlchen/api/v1/fields.py
```diff
@@ -482,13 +482,13 @@ class AssetField(fields.Field):
...
-                type[CryptoAsset],
-                type[EvmToken],
-            ],
+            expected_type: type[Union[
+                Asset,
+                AssetWithNameAndType,
+                AssetWithOracles,
+                CryptoAsset,
+                EvmToken,
+            ]],
```
yabirgb, line 491: in python 3.10 we can use `|` for the types. Don't know if it solves the issue of having to add `type`

rotkehlchen/accounting/structures/base.py
yabirgb: I believe the right type is object since you have to compare against anything but inside what is used are properties of HistoryBaseEntry that will fail if `other` is not of a subtype of `HistoryBaseEntry`

rotkehlchen/accounting/structures/base.py
```diff
@@ -119,22 +119,22 @@ def __init__(
...
         if type(self) is not type(other):  # pylint: disable=unidiomatic-typecheck
             return False
```
yabirgb, line 124: why not to use the isinstance as suggested here? https://pylint.pycqa.org/en/latest/user_guide/messages/convention/unidiomatic-typecheck.html

rotkehlchen/api/v1/fields.py
LefterisJP: no it doesn't. Would just switch out the union and replace with `X | Y | Z`

rotkehlchen/accounting/structures/base.py
LefterisJP: Because it's not the same thing. Perhaps we should add a comment there.

let's say you have a `Parent` class and a `Child` class inheriting from `Parent`. Try this:


```python
class Parent:

    def __init__(self, x: int) -> None:
        self.x = 5

    def __eq__(self, other: object):
        if not isinstance(other, Parent):
        # if type(other) != type(self):
            return False

        return self.x == other.x
            

class Child(Parent):
    pass


class OtherChild(Parent):
    pass


child = Child(x=42)
otherchild = OtherChild(x=42)

print(child == otherchild)  # prints True for isinstance and False (what you want) for type check
```

PR #6495, opened 2023-08-13
Do not count pnl in any deposit/withdrawal

LefterisJP:
Also fix accounting event settings for bridge withdrawal


LefterisJP:
nope this is not what we need

PR #6496, opened 2023-08-14
Bridge changes

LefterisJP:
- Harmonize ...a tiny bit ... code for bridging so at least the notes appear in the same way
- Remove bridge/deposit/withdrawal accounting since it messed with cost basis. They should count as transfers, which means not taken into account at all atm. At some point we do need to connect deposits/with wihtdrawals for bridges and CEX<->chain
- Add arbitrum one modules for the 1.30 binaries.


yabirgb (approved):
`rotkehlchen/tests/api/test_pickle.py::test_pickle_dill` passes locally and the other 2 tests failing seem to be an issue with the balancer subgraph

PR #6498, opened 2023-08-14
Fix newer deposits of optimism bridge

LefterisJP:
For those ones l1Fee was missing from the receipt, since it was 0. But the code always assumed l1Fee would be there. This commit fixes this.


LefterisJP:
The failure is unrelated

PR #6500, opened 2023-08-14
feat: use ui library

lukicenturi:
Closes #(issue_number)

## Checklist


frontend/app/tests/e2e/pages/rotki-app.ts
```diff
@@ -135,7 +133,11 @@ export class RotkiApp {
     };
 
     const menuClass = `.navigation__${menu}`;
-    click(menuClass);
+    if (
+      !Cypress.$(menuClass).parent().parent().hasClass('v-list-group--active')
```
kelsos, line 137: isn't this the same as cy.get()?

kelsos: > isn't this the same as cy.get()?

It probably is similar without the chainable?

frontend/app/src/components/helper/Card.vue
```diff
@@ -46,39 +44,26 @@ const otherHeights = computed(() => {
...
     :class="{
-      [css['no-radius-bottom']]: noRadiusBottom,
-      [css['full-height']]: fullHeight
+      ['h-full']: fullHeight
     }"
+    :outlined="outlined"
     v-on="rootListeners"
   >
-    <VCardTitle v-if="slots.title" :class="{ 'pt-6': slots.icon }">
-      <slot v-if="slots.icon" name="icon" />
```
kelsos, line 55: We don't use the slot anywhere? app or components?

lukicenturi: I replace it to other component. So it's only used in LP (I put it in LpPoolIcon), because I feel the style is hacky, and all usage can be moved there.

frontend/app/src/composables/dark-mode.ts
```diff
@@ -1,13 +1,56 @@
...
   const { darkModeEnabled } = storeToRefs(store);
 
   const updateDarkMode = (enabled: boolean) => {
     set(darkModeEnabled, enabled);
   };
 
+  watchDeep(
+    theme,
+    theme => {
+      // @ts-ignore
```
kelsos, line 16: what are we ignoring here? :)

frontend/app/src/locales/en.json
```diff
@@ -3132,7 +3132,7 @@
     "api_keys_sub": {
       "exchanges": "Exchanges",
       "external_services": "External Services",
-      "premium": "rotki Premium"
+      "premium": "Rotki Premium"
```
kelsos, line 3135: don't we keep rotki lower case everywhere?

frontend/app/src/components/helper/Card.vue
```diff
@@ -46,39 +44,26 @@ const otherHeights = computed(() => {
...
     </VCardSubtitle>
     <VCardText
       ref="bodyRef"
       :class="{
         [css.contained]: contained,
-        [css['no-padding']]: noPadding
+        ['no-padding']: noPadding
       }"
     >
       <slot name="search" />
```
kelsos, line 69: Can't select the actual place but can we replace the VSheet with a normal div for actions? from what I can see in the new designs there is no outline and we need to move as much as possible away from Vuetify either way

frontend/app/src/composables/dark-mode.ts
lukicenturi: I need access to `parsedTheme`, but it is non enumerable, and not defined in the type

PR #6509, opened 2023-08-15
Add hook for content_hash

yabirgb:
ENS decoder was failing to load because `content_hash` was not properly picked up by pyinstaller. There are several submodules that need to be imported and they were not automatically imported.

The reason this got unnoticed is because we suppress the module not found error when loading the decoders and this is the same error that was appearing due to the missing lib


LefterisJP (approved):
LGTM as yabir says the generated binary makes it all work fine

Issue #5038, opened 2022-10-27
Upload data to server died with exception: database plaintext is locked


LefterisJP:
## Problem Definition

This message popped up 2 times in more than 2 weeks of testing. Since it's just a non-critical error of a task that gets retried later we did not push it but it should be figured out and fixed.

```
Upload data to server died with exception: database plaintext is locked. Check the logs for more details
```

## Logs

This is what is seen in the logs

```
[27/10/2022 23:35:11 CEST] ERROR rotkehlchen.greenlets Greenlet with id 140709858290144: Upload data to server died with exception: database plaintext is locked.
Exception Name: <class 'pysqlcipher3.dbapi2.OperationalError'>
Exception Info: database plaintext is locked
Traceback:
   File "src/gevent/greenlet.py", line 908, in gevent._gevent_cgreenlet.Greenlet.run
  File "rotkehlchen/premium/sync.py", line 142, in maybe_upload_data_to_server
  File "rotkehlchen/data_handler.py", line 198, in compress_and_encrypt_db
  File "rotkehlchen/db/dbhandler.py", line 483, in export_unencrypted
  File "rotkehlchen/db/drivers/gevent.py", line 264, in executescript

[27/10/2022 23:35:11 CEST] ERROR rotkehlchen.user_messages Greenlet with id 140709858290144: Upload data to server died with exception: database plaintext is locked. Check the logs for more details
```

## Task

Fix it

kelsos:
Didn't this happen in the past too? Did we fix it or we could not reproduce?

LefterisJP:
I saw it at least once. Thought I had fixed it. My self at 01 at night did not remember. We would have to investigate

LefterisJP:
The reason for this issue was misunderstood and it keeps happening. I think I reproduced and as such will re-open and re-fix.

LefterisJP:
Fixed by https://github.com/rotki/rotki/pull/5254 again

LefterisJP:
I saw this **again** with the same large DB in the bugfixes branch (1.27.1.dev)

LefterisJP:
This is still happening

LefterisJP:
Hello darkness my old friend. This happened again. And is also happening in production (1.39.1) and in current develop (will be 1.40)

```
[29/07/2025 16:25:40 CEST] ERROR rotkehlchen.greenlets.manager Greenlet with id 4478465408: Upload data to server died with exception: database table is locked.
[...]

PR #6512, opened 2023-08-15
Third time is the charm

LefterisJP:
### [Fix DB plaintext locked -- third try](https://github.com/rotki/rotki/commit/5dece5ffa9bb148506154869dad5f85727e5675b) 

Fix https://github.com/rotki/rotki/issues/5038 ... for the third time.

The idea is to put the entire export unencrypted DB under a critical
DB section.

It's needed because a context switch
from inside this execute script can result in:
1. coming into this code again from another greenlet which can result
to DB plaintext already in use
2. Having a DB transaction open between the attach and detach and not
closed when we detach which will result in DB plaintext locked.

### [Do not constantly retry uploads. Only try an upload per hour](https://github.com/rotki/rotki/commit/9819996bd76600f4259b616c82d6e3894aa52070) 

Forcing an upload via the UI will still work.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -24,6 +24,8 @@ Changelog
 * :feature:`-` The balances snapshot csv file exported from rotki now contains an asset symbol column.
 * :bug:`-` Remote errors should no longer affect the ethereum staking deposits decoded event view.
 * :bug:`-` Newer deposits to zksync lite should be decoded properly in the history events view.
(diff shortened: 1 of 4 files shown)


LefterisJP:
The single test `rotkehlchen/tests/unit/test_tokens.py::test_detect_tokens_for_addresse` failed due to network timeout

PR #6517, opened 2023-08-16
Prevent watchers endpoint from failing when there is an error in rotki.com

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/premium/premium.py
+++ b/rotkehlchen/premium/premium.py
@@ -5,6 +5,7 @@ import logging
 import time
 from base64 import b64decode, b64encode
 from binascii import Error as BinasciiError
+from collections.abc import Sequence
 from enum import Enum
 from http import HTTPStatus
 from typing import Any, Literal, NamedTuple, Optional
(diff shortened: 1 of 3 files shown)


rotkehlchen/premium/premium.py
```diff
@@ -47,15 +47,23 @@ class RemoteMetadata(NamedTuple):
...
+    """
+    Check the rotki.com response and if status code is not in the expected list
+    log the error and raise RemoteError
+    """
     if response.status_code not in HANDLABLE_STATUS_CODES:
-        raise RemoteError(
-            f'Unexpected status response({response.status_code}) from '
-            f'rotki server. {response.text=}',
+        log.critical(
+            'rotki server answerer with an error response with status'
```
LefterisJP, line 57:
```suggestion
            'rotki server responded with an error response with status'
```

LefterisJP (dismissed):
Shouldn't this check also happen in `pull_data`?

Also noticed that the ctor has a double line (not introduced in this PR but perhaps fix?)

`self.session.mount('https://', adapter)`

rotkehlchen/premium/premium.py
```diff
@@ -406,6 +414,7 @@ def watcher_query(
             log.error(msg)
             raise RemoteError(msg) from e
 
+        check_response_status_code(response)
```
LefterisJP, line 417: Can you tell me here what was the status code of the response from the server? Was it non-200? Are you sure? If yes and your fix is abstracting this it should be used in all other places not only here.

I will amend the PR.

LefterisJP (commented):
I will just take it over to finish what I needed here. This should have been abstracted more properly

LefterisJP:
Amended, squashed and lets' see how it goes. I didn't do much.
Just fixed the smaller things I asked but the biggest was to better abstract the new function you introduced and use it everywhere.

The tests will probably still fail in the etherscan problems but this is unrelated.

LefterisJP (approved):
LGTM.

Have a look too yabir to see I did not do any booboo.

All failed tests are the etherscan thing, which is now fixed. I contacted them.

PR #6521, opened 2023-08-17
Fix issues related to icons

yabirgb:
- svg icons will work
- when uploading an icon with a different extension existing copies will be deleted

### DIFF
--- a/rotkehlchen/icons.py
+++ b/rotkehlchen/icons.py
@@ -35,6 +35,23 @@ log = RotkehlchenLogsAdapter(logger)
 ALLOWED_ICON_EXTENSIONS = ('.png', '.svg', '.jpeg', '.jpg', '.webp')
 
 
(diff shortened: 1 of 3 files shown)


lukicenturi:
Works perfectly

rotkehlchen/icons.py
```diff
@@ -35,6 +35,23 @@
 ALLOWED_ICON_EXTENSIONS = ('.png', '.svg', '.jpeg', '.jpg', '.webp')
 
 
+def _extract_http_type_from_extension(image_path: Path) -> str:
```
LefterisJP, line 38: Why abstract this in a function when it's only used in _build_http_header_for_images? Why not just put it inside there? Both are really small functions and if used in only 1 location there is no reason to do this

LefterisJP (approved):

PR #6525, opened 2023-08-18
Fix missing gitcoin donations in optimism

LefterisJP:
Taking this huge transaction as an example it was noticed that multiple donations were missing:
https://optimistic.etherscan.io/tx/0x5d85b436f5f177de6019baa9ecebae285e0def4924546307fac40556bece4cd7

This was due to not having many voting_impl addresses in the decoder. This is unfortunately a manual process which will need fixing often.

Also fixed a problem in the decoder where the original big transfer of the token to the contract for splitting was not removed.

Before:

https://github.com/rotki/rotki/assets/1658405/60821f04-c211-4977-bc23-87f7b6684987

Now:

https://github.com/rotki/rotki/assets/1658405/f6adf8f5-580c-4d3c-b1ab-30137e5f90e1

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* :bug:`-` Fix an issue where certain gitcoin donations were not detected in optimism and where the big transfer to the contract which later splits into the donations was mistakenly kept.
+
 * :release:`1.30.0 <2023-08-17>`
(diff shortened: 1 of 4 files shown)


yabirgb (approved):
lgtm

LefterisJP:
Moved to bugfixes

PR #6530, opened 2023-08-19
Improve handling of spam assets between user + global DB

### LINKED ISSUE #6528: Improve handling of spam assets for multiple accounts in the same globaldb
## Abstract

the problem is:

Account `A` detects the spam token `EVIL TOKEN`. It marks it as spam when executing `get_or_create_evm_token` and adds to ignored assets of user `A`.

Then you create user `B` and detects token. Since we don't exclude in https://github.com/rotki/rotki/blob/69bd559198fc33df1d7aa34e2e0991ccae9fce89/rotkehlchen/globaldb/handler.py#L826  the protocol `SPAM`, `EVIL TOKEN` appears in the set of tokens to be checked and B adds it. Also B doesn't have automatically added `EVIL TOKEN` to the ignore list since there is no logic that does this.

## Specification

1. In the function linked use a filter and exclude spam protocol from the search (this allows us to query less tokens)
2. On logging or in a background task (tbd, not sure if the operation is too heavy) check the identifiers for spam tokens in the globaldb and if any is missing for the user's ignore list just add it

LefterisJP:
Closes #6528

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -3047,7 +3047,7 @@ class RestAPI:
 
         if result is None:
             with self.rotkehlchen.data.db.user_write() as cursor:
-                self.rotkehlchen.data.db.add_globaldb_assetids(cursor)
+                self.rotkehlchen.data.db.sync_globaldb_assets(cursor)
             return OK_RESULT
 
         return {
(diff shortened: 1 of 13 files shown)


LefterisJP:
The only failed test passes locally and is a timeout so merging.

PR #6531, opened 2023-08-20
Process min and max version limits for all data updates

### LINKED ISSUE #6527: Cleanup v2 spam asset update and specify min version for consumption of data updates
## Problem Definition

This issue is to describe two problems that are tied together.

@yabirgb ended up duplicating all the spam asset data that had the wrong chain name "matic" in https://raw.githubusercontent.com/rotki/data/develop/updates/spam_assets/v2.json because if users downloaded the spam update before 1.29 they would need the name matic. After they would need the name "polygon_pos".

This is showing a problem of these willy nilly data updates that can happen whenever data needs to have different forms across updates due to the app having different logic.

## Task

- We need to remove spam asset chain "matic" from v2 spam asset update. Just remove all duplicates and keep only the polygon pos name.
- Create logic in the app that checks for each data update if there is a minimum rotki version that needs to consume it. And don't consume it if version is lower. We already do this but with DB version, for assets using the info.json: https://github.com/rotki/assets/blob/develop/updates/info.json. Updates also have an info.json. So we should use that: https://github.com/rotki/data/blob/develop/updates/info.json. Such logic will solve many problems we have had and may have again in the future about fearing how an update will be seen by users who have not updated. Simpy **do not let them consume it**. They will when they update.

LefterisJP:
Also refactor code around data updates to make it better and more reusable

Fix #6527

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,6 +36,9 @@ filetype==1.2.0
 # for fuzzy search of assets
 polyleven==0.8
 
+# We used to only use this for packaging, but now use it for the version comparison functionality
+packaging==23.1
+
 bip-utils==2.7.1  # require coincurve < 18.0.0 so can't upgrade yet
 
 typing-extensions==4.5.0 # used for typing.ParamSpec and typing.Concatenate (in stdlib since python 3.10). Also limiting temporarily to 4.5.0 due to a bug in multiformats which is used by rotki-content-hash: https://github.com/hashberg-io/multiformats/issues/10
--- a/rotkehlchen/assets/spam_assets.py
+++ b/rotkehlchen/assets/spam_assets.py
@@ -79,12 +79,18 @@ def update_spam_assets(db: 'DBHandler', assets_info: list[dict[str, Any]]) -> in
     # and in that case query_token_spam_list add it to the global DB
     with db.conn.read_ctx() as cursor:
         ignored_asset_ids = db.get_ignored_asset_ids(cursor)
-    assets_added = 0
+
+    write_data = []
     for token in spam_tokens:
         if token.identifier in ignored_asset_ids:
             continue
 
-        with db.user_write() as write_cursor:
-            db.add_to_ignored_assets(write_cursor=write_cursor, asset=token)
-        assets_added += 1
+        write_data.append(('ignored_asset', token.identifier))
+
+    with db.user_write() as write_cursor:
+        write_cursor.executemany(
+            'INSERT OR IGNORE INTO multisettings(name, value) VALUES(?, ?)', write_data,
+        )
+        assets_added = write_cursor.rowcount
+
     return assets_added
--- a/rotkehlchen/db/updates.py
+++ b/rotkehlchen/db/updates.py
@@ -1,10 +1,11 @@
 import json
 import logging
 from collections.abc import Sequence
-from typing import TYPE_CHECKING, Any, Literal, Union, overload
+from typing import TYPE_CHECKING, Any
 
 import requests
 
+from packaging import version as pversion
 from rotkehlchen.assets.spam_assets import update_spam_assets
 from rotkehlchen.db.addressbook import DBAddressbook
 from rotkehlchen.db.settings import CachedSettings
@@ -22,6 +23,7 @@ from rotkehlchen.types import (
 )
 from rotkehlchen.utils.misc import is_production, ts_now
 from rotkehlchen.utils.network import query_file
+from rotkehlchen.utils.version_check import get_current_version
 
 if TYPE_CHECKING:
     from rotkehlchen.db.dbhandler import DBCursor, DBHandler
@@ -31,12 +33,6 @@ from .constants import LAST_DATA_UPDATES_KEY, UpdateType
 
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
-
-
-SPAM_ASSETS_URL = 'https://raw.githubusercontent.com/rotki/data/{branch}/updates/spam_assets/v{version}.json'  # noqa: E501
-RPC_NODES_URL = 'https://raw.githubusercontent.com/rotki/data/{branch}/updates/rpc_nodes/v{version}.json'  # noqa: E501
-CONTRACTS_URL = 'https://raw.githubusercontent.com/rotki/data/{branch}/updates/contracts/v{version}.json'  # noqa: E501
-GLOBAL_ADDRESSBOOK_URL = 'https://raw.githubusercontent.com/rotki/data/{branch}/updates/global_addressbook/v{version}.json'  # noqa: E501
 
 
 class RotkiDataUpdater:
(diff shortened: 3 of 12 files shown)


rotkehlchen/assets/spam_assets.py
```diff
@@ -79,12 +79,18 @@ def update_spam_assets(db: 'DBHandler', assets_info: list[dict[str, Any]]) -> in
...
     for token in spam_tokens:
         if token.identifier in ignored_asset_ids:
             continue
 
-        with db.user_write() as write_cursor:
-            db.add_to_ignored_assets(write_cursor=write_cursor, asset=token)
-        assets_added += 1
+        write_data.append(('ignored_asset', token.identifier))
+
+    with db.user_write() as write_cursor:
```
LefterisJP, line 90: This is a an important adjustment of this PR. Using db.add_to_ignored_assets in a loop should be avoided as it:

1. Makes as many writes as there is assets
2. Takes an asset and calls many asset specific functions while you just need the identifier

requirements.txt
```diff
@@ -36,6 +36,9 @@ filetype==1.2.0
 # for fuzzy search of assets
 polyleven==0.8
 
+# We used to only use this for packaging, but now use it for the version comparison functionality
+packaging==23.1
```
LefterisJP, line 40: Moved packaging as a requirement to the app itself so we can use version comparison.

rotkehlchen/db/updates.py
```diff
@@ -59,7 +55,56 @@ def __init__(self, msg_aggregator: 'MessagesAggregator', user_db: 'DBHandler') -
             UpdateType.RPC_NODES: self.update_rpc_nodes,
             UpdateType.CONTRACTS: self.update_contracts,
             UpdateType.GLOBAL_ADDRESSBOOK: self.update_global_addressbook,
-        }
+        }  # If we ever change this also change tests/unit/test_data_updates::reset_update_type_mappings  # noqa: E501
+        self.version = pversion.parse(get_current_version().our_version)
+
+    def update_single(
```
LefterisJP, line 61: Made this function contain all the common functionality between all the data update functions

rotkehlchen/db/updates.py
```diff
@@ -59,7 +55,56 @@ def __init__(self, msg_aggregator: 'MessagesAggregator', user_db: 'DBHandler') -
...
+                log.warning(f'Failed to update {update_type.value} due to {e!s}')
+                continue  # perhaps broken file? Skipping
+
+            updated_data = updates.get(update_type.value)
+            if updated_data is None:
+                log.error(f'Remote update {file_url} does not contain {update_type.value} key')
+                continue  # perhaps broken file? Skipping
+
+            # At this point we can apply the update. Type ignore is due to different sigs of data
+            self.update_type_mappings[update_type](updated_data, update_version)  # type: ignore
```
LefterisJP, line 102: here we call the update type specific functionality

rotkehlchen/tests/data_migrations/test_migrations.py
```diff
@@ -375,25 +380,24 @@ def test_migration_11(
...
+            ],
+        )
     migration_patch = patch(
         'rotkehlchen.data_migrations.manager.MIGRATION_LIST',
         new=[MIGRATION_LIST[5]],
     )
 
     arbitrum_one_addresses = [ethereum_accounts[1]]
-    with ExitStack() as stack:
-        setup_evm_addresses_activity_mock(
```
LefterisJP, line 386: no need to use both VCR and mocking evm queries. Only kind of hard thing was to find an address that at the time of the test was active **only** in mainnet and arbitrum. Did it manually by searching

rotkehlchen/assets/spam_assets.py
```diff
@@ -79,12 +79,18 @@ def update_spam_assets(db: 'DBHandler', assets_info: list[dict[str, Any]]) -> in
...
+
+    write_data = []
     for token in spam_tokens:
         if token.identifier in ignored_asset_ids:
             continue
 
-        with db.user_write() as write_cursor:
-            db.add_to_ignored_assets(write_cursor=write_cursor, asset=token)
-        assets_added += 1
+        write_data.append(('ignored_asset', token.identifier))
```
yabirgb, line 88: Maybe is faster here to use list comprhension?

rotkehlchen/tests/fixtures/rotkehlchen.py
```diff
@@ -124,6 +126,8 @@ def patch_and_enter_before_unlock(
         ksm_rpc_endpoint,
         ethereum_manager_connect_at_start,
         optimism_manager_connect_at_start,
+        polygon_pos_manager_connect_at_start,
```
LefterisJP, line 129: not for here but man we do need to abstract those connection mocks better ... or remove/adjust them

rotkehlchen/tests/fixtures/rotkehlchen.py
```diff
@@ -133,10 +137,23 @@ def patch_and_enter_before_unlock(
 ) -> None:
     # Do not connect to the usual nodes at start by default. Do not want to spam
     # them during our tests. It's configurable per test, with the default being nothing
-    rpc_nodes_result = ethereum_manager_connect_at_start + optimism_manager_connect_at_start if network_mocking is False else []  # noqa: E501
+    def mock_get_rpc_nodes(blockchain, only_active):  # pylint: disable=unused-argument
```
LefterisJP, line 140: So .. before what was happening was that `db.get_rpc_nodes()` was returning "optimism + ethereum" nodes to connect to for **any** chain. yes ... lol ...

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -208,19 +282,64 @@ def test_no_update_performed(data_updater: RotkiDataUpdater) -> None:
...
+    assert all(patch.call_count == 0 for patch in patches)
+    with data_updater.user_db.conn.read_ctx() as cursor:
+        cursor.execute(  # also make sure latest DB value is not changed
+            'SELECT value from settings WHERE name IN(?, ?, ?, ?)',
+            [x.serialize() for x in UpdateType],
+        )
+        assert {x[0] for x in cursor} == {'999'}
+
+
+@pytest.mark.parametrize('our_version', ['1.35.0'])  # set a normal version for CI
```
LefterisJP, line 303: putting this here only for CI, since in local we do `pip install -e .` and have a proper version but in CI we don't

rotkehlchen/tests/utils/blockchain.py
```diff
@@ -679,3 +682,36 @@ def mock_arbitrum_one_has_activity(account):
         side_effect=mock_arbitrum_one_has_activity,
     ))
     return stack
+
+
+def maybe_modify_rpc_nodes(
```
LefterisJP, line 687: abstracted this out from `tests/fixtures/blockchain.py` since this was missing from rotkehlchen_api_server fixture and needed to also be called there.

rotkehlchen/db/updates.py
```diff
@@ -59,7 +55,56 @@ def __init__(self, msg_aggregator: 'MessagesAggregator', user_db: 'DBHandler') -
...
+            try:
+                updates = query_file(file_url, True)
+            except RemoteError as e:
+                log.warning(f'Failed to update {update_type.value} due to {e!s}')
+                continue  # perhaps broken file? Skipping
+
+            updated_data = updates.get(update_type.value)
+            if updated_data is None:
+                log.error(f'Remote update {file_url} does not contain {update_type.value} key')
+                continue  # perhaps broken file? Skipping
```
yabirgb, line 99: this is the same comment as in the remote error. I guess remote error should be broken path? or is it intended to be the same?

yabirgb (commented):
looks okey, just have two minor comments

rotkehlchen/db/updates.py
LefterisJP: broken file is intended to include both as a meaning but can specify if you want. I just wanted to explain why "continue"

rotkehlchen/assets/spam_assets.py
LefterisJP: Maybe. I did not do it since I kept `assets_added` at first, but then thought of `cursor.rowcount` for it. Will see if it makes sense and add in next iteration/PR

LefterisJP:
All green so will merge.

Issue #6533, opened 2023-08-21
Add a notification to the user when there are missed premium sync


kelsos:
## Abstract

There is no indicator that a premium sync has not happened. The user has to go to a menu and manually check for the date.

We should add some type of notification to the user to make them aware of missed syncs. Along with an action to perform a sync on demand.

## Motivation

Ensure that users don't miss syncs without being aware, and can react to missed syncs.

## Specification

- [ ] TBD

PR #6535, opened 2023-08-21
Add a websocket message that provides DB upload result info

LefterisJP:
We now have a way to provide information to the user for the periodic DB upload task and let the know if something went wrong or if they can do something to force push (if they are absolutely sure).

Backend part of #6533


yabirgb (approved):
lgtm

Issue #2000, opened 2020-12-29
Differentiate between claimed and unclaimed airdrops


LefterisJP:
## Abstract

As a follow up of #972 we should now also detect which aidrops a user has already claimed and show them in a separate area below the unclaimed ones so that it's obvious which airdrops consist of action items for the user and which is more historical information.

## Task

- [ ] Implement backend code to detect when any of the given airdrops is claimed
- [ ] Add the frontend changes to distinguish between claimed and unclaimed airdrops

LefterisJP:
This has been implemented by https://github.com/rotki/rotki/pull/6544 and by whichever the frontend PR was.

PR #6544, opened 2023-08-23
marks airdrops claimed or not

dimyG:
Backend part for #2000

## Checklist

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -56,7 +56,7 @@ from rotkehlchen.assets.asset import (
     ResolvedAsset,
 )
 from rotkehlchen.assets.resolver import AssetResolver
-from rotkehlchen.assets.types import AssetType
+from rotkehlchen.assets.types import ASSET_TYPES_EXCLUDED_FOR_USERS, AssetType
 from rotkehlchen.balances.manual import (
     ManuallyTrackedBalance,
     add_manually_tracked_balances,
@@ -78,6 +78,7 @@ from rotkehlchen.chain.ethereum.modules.yearn.utils import query_yearn_vaults
 from rotkehlchen.chain.ethereum.utils import try_download_ens_avatar
 from rotkehlchen.chain.evm.names import find_ens_mappings, search_for_addresses_names
 from rotkehlchen.chain.evm.types import WeightedNode
+from rotkehlchen.constants import ONE
 from rotkehlchen.constants.assets import A_ETH
 from rotkehlchen.constants.limits import (
     FREE_ASSET_MOVEMENTS_LIMIT,
@@ -88,14 +89,12 @@ from rotkehlchen.constants.limits import (
     FREE_USER_NOTES_LIMIT,
 )
 from rotkehlchen.constants.misc import (
-    ASSET_TYPES_EXCLUDED_FOR_USERS,
     DEFAULT_MAX_LOG_BACKUP_FILES,
     DEFAULT_MAX_LOG_SIZE_IN_MB,
     DEFAULT_SQL_VM_INSTRUCTIONS_CB,
     HTTP_STATUS_INTERNAL_DB_ERROR,
-    ONE,
-    ZERO_PRICE,
 )
+from rotkehlchen.constants.prices import ZERO_PRICE
 from rotkehlchen.constants.resolver import ChainID
 from rotkehlchen.constants.timing import ENS_AVATARS_REFRESH
 from rotkehlchen.data_import.manager import DataImportSource
@@ -2223,7 +2222,8 @@ class RestAPI:
         try:
             data = check_airdrops(
                 addresses=self.rotkehlchen.chains_aggregator.accounts.eth,
-                data_dir=self.rotkehlchen.data_dir,
+                database=self.rotkehlchen.data.db,
+                tolerance_for_amount_check=FVal('0.00000000000001000'),
             )
         except (RemoteError, UnableToDecryptRemoteData) as e:
             return wrap_in_fail_result(str(e), status_code=HTTPStatus.BAD_GATEWAY)
--- a/rotkehlchen/chain/ethereum/airdrops.py
+++ b/rotkehlchen/chain/ethereum/airdrops.py
@@ -8,7 +8,11 @@ from typing import Any
 
 import requests
 
+from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.assets.asset import Asset
 from rotkehlchen.chain.ethereum.utils import token_normalized_value_decimals
+from rotkehlchen.chain.evm.types import string_to_evm_address
+from rotkehlchen.constants import ZERO
 from rotkehlchen.constants.assets import (
     A_1INCH,
     A_COMBO,
@@ -26,10 +30,12 @@ from rotkehlchen.constants.assets import (
     A_UNI,
     A_VCOW,
 )
+from rotkehlchen.db.dbhandler import DBHandler
 from rotkehlchen.db.settings import CachedSettings
 from rotkehlchen.errors.misc import RemoteError, UnableToDecryptRemoteData
+from rotkehlchen.fval import FVal
 from rotkehlchen.logging import RotkehlchenLogsAdapter
-from rotkehlchen.types import ChecksumEvmAddress
+from rotkehlchen.types import ChecksumEvmAddress, FValWithTolerance
 from rotkehlchen.utils.serialization import jsonloads_dict, rlk_jsondumps
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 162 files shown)


rotkehlchen/chain/ethereum/airdrops.py
```diff
@@ -282,9 +289,36 @@ def get_poap_airdrop_data(name: str, data_dir: Path) -> dict[str, Any]:
...
+    )
+    with database.conn.read_ctx() as cursor:
+        events = DBHistoryEvents(database).get_history_events(
+            cursor=cursor,
+            filter_query=db_filter,
+            has_premium=True,  # no need to limit filtered events here
+        )
+        for event in events:
+            if event.balance.amount == amount:
+                return True
```
yabirgb, line 314: to avoid this loop you could add a filter on the amounts to the superclass of `EvmEventFilterQuery` and then use `get_history_events_count` so you could just check that 

```
return DBHistoryEvents(database).get_history_events_count(cursor, db_filter)[0] != 0
```

rotkehlchen/tests/unit/test_ethereum_airdrops.py
```diff
@@ -28,10 +36,37 @@ def _fixture_airdrop_list(airdrop_list):
 
 @pytest.mark.parametrize('number_of_eth_accounts', [2])
 @pytest.mark.parametrize('use_clean_caching_directory', [True])
-def test_check_airdrops(ethereum_accounts, data_dir):
+def test_check_airdrops(ethereum_accounts, data_dir, database):
+    claim_event = EvmEvent(
+        tx_hash=deserialize_evm_tx_hash('0x5edc23d5a05e347afc60e64a4d5831ed2551985c21dceb85d267926ca2e2c13e'),  # A random tx_hash  # noqa: E501
```
yabirgb, line 41: use

```suggestion
        tx_hash=make_evm_tx_hash()
```

rotkehlchen/chain/ethereum/airdrops.py
```diff
@@ -282,9 +289,36 @@ def get_poap_airdrop_data(name: str, data_dir: Path) -> dict[str, Any]:
     return data_dict
 
 
+def is_airdrop_claimed(
+        address: ChecksumEvmAddress,
+        amount: FVal,
+        asset: Asset,
+        database: DBHandler,
```
yabirgb, line 296: since this logic is executed inside a nested loop consider creating `events_db = DBHistoryEvents(database)` outside in `def check_airdrops` and use it here. for the cursor you can do `events_db.db.conn.read_ctx()` since is an attribute. Doing this you avoid creating an object `NUMBER_OF_AIDROPS * NUMBER OF ADDRESSES` times

yabirgb (commented):
This is missing a chagelog entry

rotkehlchen/db/filtering.py
```diff
@@ -989,6 +990,14 @@ def make(
...
                 operator='NOT IN',
             ))
+        if amounts is not None:
+            str_amounts = [str(amount) for amount in amounts]
+            filters.append(DBMultiStringFilter(
+                and_op=True,
+                column='amount',
+                values=str_amounts,
+                operator='IN',
+            ))
```
yabirgb, line 1000: Something minor but you can just do

```suggestion
            filters.append(DBMultiStringFilter(
                and_op=True,
                column='amount',
                values=[str(amount) for amount in amounts],
                operator='IN',
            ))
```

yabirgb (commented):
just something minor

rotkehlchen/api/rest.py
```diff
@@ -2224,6 +2224,7 @@ def get_ethereum_airdrops(self) -> dict[str, Any]:
             data = check_airdrops(
                 addresses=self.rotkehlchen.chains_aggregator.accounts.eth,
                 data_dir=self.rotkehlchen.data_dir,
+                database=self.rotkehlchen.data.db,
```
LefterisJP, line 2227: Can't check as still mobile, but doesn't the database also contain the data directory or not? If it does perhaps you can remove it from above as an argument.

LefterisJP (commented):
Left some comments

rotkehlchen/chain/ethereum/airdrops.py
```diff
@@ -282,16 +290,52 @@ def get_poap_airdrop_data(name: str, data_dir: Path) -> dict[str, Any]:
...
+        addresses=addresses,
+        assets=tuple(assets),
+        amounts=amounts,
+    )
+    events_db = DBHistoryEvents(database)
+    with database.conn.read_ctx() as cursor:
+        events = events_db.get_history_events(
+            cursor=cursor,
+            filter_query=db_filter,
+            has_premium=True,
```
LefterisJP, line 311:
```suggestion
            has_premium=True,  # needed for completeness sake anyway here
```

rotkehlchen/db/filtering.py
```diff
@@ -568,6 +569,30 @@ def prepare(self) -> tuple[list[str], list[T]]:
...
+        parameters = []
+
+        for val in self.values:
+            value = val[0]
+            tolerance = val[1] * FVal(0.5)
+            conditions.append(f'CAST({self.column} AS REAL) BETWEEN ? AND ?')
+            parameters.extend([str(value - tolerance), str(value + tolerance)])
+
+        return (
+            [' OR '.join(conditions)],
```
LefterisJP, line 591: Why OR? Shouldn't it be AND? Or at least make this configurable? It already is an argument and you are giving it. The `and_op=True` one.

rotkehlchen/types.py
```diff
@@ -881,6 +881,11 @@ def deserialize_from_db(cls, entry: tuple[int, str, str, str, int, int]) -> 'Use
         )
 
 
+class FValWithTolerance(NamedTuple):
```
LefterisJP, line 884: A small description of what this does in a docstring. Oneliner

rotkehlchen/types.py
```diff
@@ -881,6 +881,11 @@ def deserialize_from_db(cls, entry: tuple[int, str, str, str, int, int]) -> 'Use
...
+    value: FVal
+    tolerance: FVal = FVal('0')
```
LefterisJP, line 886:
```suggestion
    tolerance: FVal = ZERO
```

rotkehlchen/db/filtering.py
```diff
@@ -568,6 +569,30 @@ def prepare(self) -> tuple[list[str], list[T]]:
...
+    column: str
+    values: list[FValWithTolerance]
+
+    def prepare(self) -> tuple[list[str], list[str]]:
+        conditions = []
+        parameters = []
+
+        for val in self.values:
+            value = val[0]
+            tolerance = val[1] * FVal(0.5)
```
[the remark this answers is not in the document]
dimyG: Because the range is created by adding half the tolerance to the given value and removing half the tolerance from it so that the total tolerance is the given tolerance number. 
`parameters.extend([str(value - tolerance), str(value + tolerance)])`

rotkehlchen/db/filtering.py
dimyG: It needs to be OR in this case because we query for a list of amounts. here is an example: 
image
That being said, I updated the filter to make the operation (AND or OR) configurable using the and_op parameter.

rotkehlchen/types.py
dimyG: Moved some things around, to solve circular dependency issues for importing ZERO here.

rotkehlchen/types.py
```diff
@@ -199,6 +200,8 @@ def deserialize_evm_tx_hash(val: Union[Web3HexBytes, bytearray, bytes, str]) ->
 T_TradeID = str
 TradeID = NewType('TradeID', T_TradeID)
 
+ZERO_PRICE = Price(ZERO)
```
LefterisJP, line 203: Why did you move this here? This is a constant, not a type. It should be inside the constants where it was before.

rotkehlchen/constants/misc.py
```diff
@@ -20,8 +16,6 @@
 KRAKEN_BASE_URL = 'https://api.kraken.com'
 KRAKEN_API_VERSION = '0'
 
-ASSET_TYPES_EXCLUDED_FOR_USERS = {AssetType.NFT}
```
LefterisJP, line 23: This is fine moving. But not the zero price.

rotkehlchen/db/filtering.py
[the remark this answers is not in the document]
LefterisJP: I see. This makes sense. Can you maybe add a tiny comment next to it explaining this. In case others reading the code have the same question?

rotkehlchen/db/filtering.py
LefterisJP: In that case there is a problem. This query should match each individual claim event. it does not.
Could also potentially match mistaken events if another event has same asset or any other combination of things here.


 I don't think our filters can work here. Cause you are trying to apply a multifilter. I really think raw sqlite is better here as I had written on Friday.

rotkehlchen/tests/unit/test_ethereum_airdrops.py
```diff
@@ -28,10 +37,46 @@ def _fixture_airdrop_list(airdrop_list):
 
 @pytest.mark.parametrize('number_of_eth_accounts', [2])
 @pytest.mark.parametrize('use_clean_caching_directory', [True])
-def test_check_airdrops(ethereum_accounts, data_dir):
+def test_check_airdrops(ethereum_accounts, database):
+    # create airdrop claim events to test the claimed attribute
+    tolerance_for_amount_check = FVal('0.1')
```
LefterisJP, line 42: Shouldn't tolerance be smaller? Or wait .. this is just to test the tolerance attribute right?

LefterisJP (commented):
Some minor comments and questions

LefterisJP (approved):
LGTM. Thanks.

PR #6553, opened 2023-08-27
Fix wrongly custom imported rotki events

LefterisJP:


rotkehlchen/data_migrations/migrations/migration_12.py
```diff
@@ -0,0 +1,50 @@
...
+from rotkehlchen.logging import RotkehlchenLogsAdapter
+
+if TYPE_CHECKING:
+    from rotkehlchen.data_migrations.progress import MigrationProgressHandler
+    from rotkehlchen.rotkehlchen import Rotkehlchen
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+PREFIX = f'{ROTKI_EVENT_PREFIX}_%'
```
yabirgb, line 15: Shouldn't `ROTKI_EVENT_PREFIX` be hardcoded here instead of using the variable?

rotkehlchen/data_migrations/migrations/migration_12.py
```diff
@@ -0,0 +1,50 @@
...
+    with rotki.data.db.conn.write_ctx() as write_cursor:
+        for to_type, to_subtype, from_type, from_subtype in CHANGES:
+            write_cursor.execute(
+                'UPDATE history_events SET type=?, subtype=? WHERE '
+                'event_identifier LIKE ? AND type=? AND subtype=?',
+                (
+                    to_type.serialize(), to_subtype.serialize(), PREFIX,
+                    from_type.serialize(), from_subtype.serialize(),
+                ),
+            )
```
yabirgb, line 48: why not make this a executemany?

rotkehlchen/data_migrations/migrations/migration_12.py
LefterisJP: What do you mean? What if it changes?

rotkehlchen/data_migrations/migrations/migration_12.py
LefterisJP: Does it work? Let me check. For some reason I think it did not.

rotkehlchen/data_migrations/migrations/migration_12.py
yabirgb: yes

rotkehlchen/data_migrations/migrations/migration_12.py
yabirgb: you need to create the list of serialized values but otherwise it should be faster than len(changes) executes

rotkehlchen/data_migrations/migrations/migration_12.py
LefterisJP: Ah I see. Okay yeah. makes sense, but will add a comment for it since I was trying to find places we hard code those and replace with variables (except for db upgrades and migrations)

PR #6560, opened 2023-08-29
Speed up decoders tests

yabirgb:
Run tests for decoders in parallel

### DIFF
--- a/rotkehlchen/tests/fixtures/db.py
+++ b/rotkehlchen/tests/fixtures/db.py
@@ -54,18 +54,18 @@ def fixture_session_data_dir(tmpdir_factory) -> Path:
 
 
 @pytest.fixture(name='user_data_dir')
-def fixture_user_data_dir(data_dir, username) -> Path:
+def fixture_user_data_dir(data_dir: Path, username: str, worker_id: str) -> Path:
     """Create and return the user data directory"""
-    user_data_dir = data_dir / username
-    user_data_dir.mkdir(exist_ok=True)
+    user_data_dir = data_dir / worker_id / username
+    user_data_dir.mkdir(exist_ok=True, parents=True)
     return user_data_dir
 
 
 @pytest.fixture(scope='session', name='session_user_data_dir')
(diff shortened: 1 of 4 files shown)


rotkehlchen/tests/fixtures/db.py
```diff
@@ -54,10 +55,11 @@ def fixture_session_data_dir(tmpdir_factory) -> Path:
 
 
 @pytest.fixture(name='user_data_dir')
-def fixture_user_data_dir(data_dir, username) -> Path:
+def fixture_user_data_dir(data_dir: Path, username: str, worker_id: str) -> Path:
     """Create and return the user data directory"""
-    user_data_dir = data_dir / username
-    user_data_dir.mkdir(exist_ok=True)
+    user_data_dir = data_dir / worker_id / username
+    shutil.rmtree(user_data_dir, ignore_errors=True)
```
LefterisJP, line 61: Why are you doing this? Isn't this why we have the `@pytest.mark.parametrize('use_clean_caching_directory', [True])` fixture?

LefterisJP (commented):
The fixture change looks wrong and also there is a relevant test failure

rotkehlchen/tests/fixtures/db.py
yabirgb: the reason to do `data_dir / worker_id / username` is that there are 3 parallel tests running and you can't have the same db, this is why to use the worker_id. As for the rmtree i got that if a test failed it couldn't run because the folder was populated but probably is due to how `use_clean_caching_directory` works. I'm going to check if this can be improved

LefterisJP (commented):
This is still not progressed right? Just checking if I need to review anything more

yabirgb:
no, not yet

yabirgb:
I've added pytest-pretty and the outputs looks like this
image

is known that xdist doesn't show the progress and this can be an issue in the CI if we can't know if the task is going and what is the current status https://github.com/pytest-dev/pytest-xdist/issues/450

yabirgb:
Closing the PR, seems that the server hangs and is innestable when running all the suite :(

PR #6562, opened 2023-08-30
Prevent double login of users

yabirgb:
It is possible for a client to send twice the request for login fast enough to skip the check that we have for logged in users as can bee seen in this trace

```
[30/08/2023 16:22:52 WIB] ERROR rotkehlchen.api.rest Greenlet with id 4728519808: Greenlet for task 3 dies with exception: set_database was called on a cryptocompare instance that already has a DB.
Exception Name: <class 'AssertionError'>
Exception Info: set_database was called on a cryptocompare instance that already has a DB
Traceback:
   File "src/gevent/greenlet.py", line 908, in gevent._gevent_cgreenlet.Greenlet.run
  File "/Users/lukicenturi/Sites/rotki/rotkehlchen/api/rest.py", line 369, in _do_query_async
    result = command(self, **kwargs)
  File "/Users/lukicenturi/Sites/rotki/rotkehlchen/api/rest.py", line 1138, in user_login
    self.rotkehlchen.unlock_user(
  File "/Users/lukicenturi/Sites/rotki/rotkehlchen/rotkehlchen.py", line 298, in unlock_user
    self.cryptocompare.set_database(self.data.db)
  File "/Users/lukicenturi/Sites/rotki/rotkehlchen/externalapis/cryptocompare.py", line 251, in set_database
    assert self.db is None, msg
```

This PR adds a lock during login to prevent from two tasks accessing to the login logic

The diff looks terrible but is just indenting the code

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -303,6 +303,19 @@ def async_api_call() -> Callable:
     return wrapper
 
 
+def login_lock() -> Callable:
+    """
+    This is a decorator that uses the login lock at RestAPI to avoid a race condition between
+    async tasks using the user unlock logic.
+    """
+    def wrapper(func: Callable[..., Response]) -> Callable[..., Response]:
+        def inner(rest_api: 'RestAPI', **kwargs: Any) -> Response:
+            with rest_api.login_lock:
+                return func(rest_api, **kwargs)
+        return inner
+    return wrapper
+
+
 class RestAPI:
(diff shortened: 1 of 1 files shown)


LefterisJP (commented):
It looks okay but since you are just putting those 2 functions just under the lock and it makes little sense to increase the indentation so much why not make it a decorator? It would be much cleaner.

LefterisJP (approved):
LGTM

PR #6563, opened 2023-08-30
premium tier: device limits

yabirgb:
Closes #(issue_number)

## Checklist


rotkehlchen/premium/premium.py
```diff
@@ -83,7 +85,7 @@ def _process_dict_response(
         raise RemoteError('Could not decode rotki server response as json. Check logs') from e
 
     if response.status_code == HTTPStatus.UNAUTHORIZED:
-        raise PremiumAuthenticationError(result_dict.get('message', 'no message given'))
+        raise PremiumAuthenticationError(result_dict.get('error', 'no message given'))
```
LefterisJP, line 88: this is unrelated to this PR. This needs to go in bugfixes. Seems this was wrong for a long time. Fix it in other places if required.

rotkehlchen/premium/premium.py
```diff
@@ -176,10 +178,13 @@ def __init__(self, credentials: PremiumCredentials):
             allowed_methods=False,  # retry on all method verbs
         )
         self.session.mount('https://', adapter)
+        self.session.verify = False
```
LefterisJP, line 181: what is this? Add a comment

rotkehlchen/premium/premium.py
```diff
@@ -176,10 +178,13 @@ def __init__(self, credentials: PremiumCredentials):
...
         self.apiversion = '1'
-        self.rotki_api = f'https://rotki.com/api/{self.apiversion}/'
-        self.rotki_nest = f'https://rotki.com/nest/{self.apiversion}/'
+        self.rotki_api = f'https://localhost/api/{self.apiversion}/'
+        self.rotki_web = f'https://localhost/webapi/{self.apiversion}/'
```
LefterisJP, line 184: Did we agree on this url for that? What was the latest talk with kelsos?

rotkehlchen/premium/premium.py
```diff
@@ -207,6 +212,78 @@ def set_credentials(self, credentials: PremiumCredentials) -> None:
...
+            device_found = True
+            break
+
+        if not device_found:
+            if num_devices < devices_limit:
+                # try to sign up the device
+                self._register_new_device(device_id)
+            else:
+                # user has to edit his devices
+                raise PremiumAuthenticationError('The limit of devices has been reached')
```
LefterisJP, line 262:
```suggestion
                raise PremiumAuthenticationError(f'The limit of {devices_limit} devices has been reached')
```

rotkehlchen/premium/premium.py
```diff
@@ -207,6 +212,78 @@ def set_credentials(self, credentials: PremiumCredentials) -> None:
...
+        for device in device_data['devices']:
+            if device['device_identifier'] != device_id:
+                continue
+            device_found = True
+            break
+
+        if not device_found:
+            if num_devices < devices_limit:
+                # try to sign up the device
+                self._register_new_device(device_id)
```
LefterisJP, line 259:
```suggestion
                # try to register the device
                self._register_new_device(device_id)
```

Also question. How do you handle errors and show them to the user here? It's possible that the user uses a modified rotki trying to cheat, and the check above is skipped or the server logic has an error and then we just get the failure from the server

rotkehlchen/premium/premium.py
```diff
@@ -207,6 +212,78 @@ def set_credentials(self, credentials: PremiumCredentials) -> None:
...
+            else:
+                # user has to edit his devices
+                raise PremiumAuthenticationError('The limit of devices has been reached')
+
+    def _register_new_device(self, device_id: str) -> dict:
```
LefterisJP, line 264: add a docstring

rotkehlchen/premium/premium.py
```diff
@@ -207,6 +212,78 @@ def set_credentials(self, credentials: PremiumCredentials) -> None:
...
+        log.debug(f'Registering new device {device_id}')
+        method = 'devices'
+        device_name = platform.system()
+        signature, data = self.sign(
+            method=method,
+            device_identifier=device_id,
+            device_name=device_name,
+        )
+        self.session.headers.update({'API-SIGN': base64.b64encode(signature.digest())})
```
LefterisJP, line 273: This signature creation is both here and in the previous function. Abstract it and call it in both places

rotkehlchen/premium/premium.py
```diff
@@ -207,6 +212,78 @@ def set_credentials(self, credentials: PremiumCredentials) -> None:
...
+
+        try:
+            resposne = self.session.put(
```
LefterisJP, line 276:
```suggestion
            response = self.session.put(
```

rotkehlchen/tests/fixtures/accounting.py
```diff
@@ -105,13 +105,14 @@ def fixture_accountant(
...
         rotki_premium_credentials,
+        username,
 ) -> Optional[Accountant]:
     if not start_with_logged_in_user:
         return None
 
     premium = None
     if start_with_valid_premium:
-        premium = Premium(rotki_premium_credentials)
+        premium = Premium(rotki_premium_credentials, username)
```
LefterisJP, line 115: use kwargs

rotkehlchen/tests/fixtures/eth2.py
```diff
@@ -20,10 +20,11 @@ def fixture_eth2(
         beaconchain,
         network_mocking,
         eth2_mock_data,
+        username,
 ):
     premium = None
     if start_with_valid_premium:
-        premium = Premium(rotki_premium_credentials)
+        premium = Premium(rotki_premium_credentials, username=username)
```
LefterisJP, line 27: here and in other places too

rotkehlchen/tests/integration/test_premium.py
```diff
@@ -620,3 +621,51 @@ def mock_error_upload_data_to_server(
 
     assert status is False
     assert error == 'Size limit reached'
+
+
+@pytest.mark.parametrize('start_with_valid_premium', [True])
+@pytest.mark.parametrize('device_limit', [1, 2])
+def test_device_limits(rotkehlchen_instance: 'Rotkehlchen', device_limit: int) -> None:
+    """
+    Test that registering new devices works both when we can register the devices and when the
```
LefterisJP, line 630:
```suggestion
    Test that registering new devices works both when we can register a new device and when the
```

rotkehlchen/tests/unit/test_aave.py
```diff
@@ -97,7 +98,7 @@ def test_aave_v1_events_stats(
     aave = Aave(
         ethereum_inquirer=ethereum_inquirer,
         database=database,
-        premium=Premium(rotki_premium_credentials),
+        premium=Premium(rotki_premium_credentials, username),
```
LefterisJP, line 101: in all these places. There is many files. Please make sure they are all checked.

LefterisJP (commented):
Left some comments

requirements.txt
```diff
@@ -22,6 +22,7 @@ miniupnpc==2.0.2; sys_platform != 'win32'
 miniupnpc==2.2.3; sys_platform == 'win32'
 cryptography==41.0.3
 jsonschema==4.17.3  # jsonchema 4.18 introduces a dependency that is missing wheels for macos in our CI. We have reported it but we will pin the version until is fixed https://github.com/crate-py/rpds/issues/11
+py-machineid==0.4.3
```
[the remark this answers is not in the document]
yabirgb: The company behind looks serious and the code is small. I can fork it and install from github if you prefer it. I've read it and is not doing anything else from obtaining the information to compute the id

rotkehlchen/premium/premium.py
yabirgb: This logic is used by `premium_create_and_verify` that already raises the possible errors `PremiumAuthenticationError` and `RemoteError`. For the frontend at the moment on error there is no custom message and the effect is a common error message saying that the premium couldn't be activated

LefterisJP: Still please take it out.

rotkehlchen/premium/premium.py
yabirgb: This is my error since I made this change last week to test and forgot to revert the values used for testing

rotkehlchen/premium/premium.py
yabirgb: testing, forgot to remove. is for requests to avoid the ssl verification since the certificate exists but is not in any registry

rotkehlchen/premium/premium.py
yabirgb: Moved it inside the sign method and removed the header set from where it is called

requirements.txt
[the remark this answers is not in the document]
LefterisJP: that's fine. If they abandon it we can recreate on our own.

rotkehlchen/premium/premium.py
```diff
@@ -178,8 +180,10 @@ def __init__(self, credentials: PremiumCredentials):
         self.session.mount('https://', adapter)
         self.apiversion = '1'
         self.rotki_api = f'https://rotki.com/api/{self.apiversion}/'
+        self.rotki_web = f'https://rotki.com/webapi/{self.apiversion}/'
```
LefterisJP, line 183: Did not even notice it was localhost before. What I said in the comment still stands. Did you actually agree on this api url endpoint for rotki web? Is this what we discussed with kelsos and agreed on? This is the rotkehlchen-web api you said would be needed right?

rotkehlchen/premium/premium.py
```diff
@@ -207,6 +211,75 @@ def set_credentials(self, credentials: PremiumCredentials) -> None:
...
+        data = _process_dict_response(response)
+        return data
+
+    def authenticate_device(self) -> None:
+        """
+        Check if the device is in the list of the devices and if it isn't add it when possible.
+        May raise:
+        - PremiumAuthenticationError: when the device can't be registered
+        """
+        device_data = self.get_remote_devices_information()
```
LefterisJP, line 242: All the dict in devices data is not checked for key errors. You need to catch them and handle all possible missing keys and raise possibly a remote error

rotkehlchen/premium/premium.py
```diff
@@ -207,6 +211,75 @@ def set_credentials(self, credentials: PremiumCredentials) -> None:
...
+            raise RemoteError(msg) from e
+
+        data = _process_dict_response(response)
+        return data
+
+    def authenticate_device(self) -> None:
+        """
+        Check if the device is in the list of the devices and if it isn't add it when possible.
+        May raise:
+        - PremiumAuthenticationError: when the device can't be registered
```
LefterisJP, line 240: Shouldn't you also add a remote error here? There is two different calls to rotki server in this function

rotkehlchen/premium/premium.py
```diff
@@ -207,6 +211,75 @@ def set_credentials(self, credentials: PremiumCredentials) -> None:
...
+                # try to register the device
+                self._register_new_device(device_id)
+            else:
+                # user has to edit his devices
+                raise PremiumAuthenticationError(
+                    f'The limit of {devices_limit} devices has been reached',
+                )
+
+    def _register_new_device(self, device_id: str) -> dict:
+        """Register a new device at the rotki server using the provided id"""
```
LefterisJP, line 261: Docstring has to contain the possible exceptions that can be raised

rotkehlchen/premium/premium.py
yabirgb: we didn't talk about this with kelsos. The webapi is something that maps in traefik to django api so in theory should be fine but we can change it if needed when kelsos get back

rotkehlchen/premium/premium.py
```diff
@@ -207,6 +211,75 @@ def set_credentials(self, credentials: PremiumCredentials) -> None:
...
+        devices_limit = device_data['limit']
+        device_id = machineid.hashed_id(self.username)
+
+        for device in device_data['devices']:
+            if device['device_identifier'] == device_id:
+                break
+        else:  # device not found
+            if num_devices < devices_limit:
+                # try to register the device
+                self._register_new_device(device_id)
```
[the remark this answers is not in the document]
yabirgb: What I mean is what the user sees. That message irc is always the same since is one from the frontend. The error message that the backend generates is exposed from the backend to the frontend in the message field for the api response

LefterisJP (approved):
Thank you Yabir. LGTM.

PR #6566, opened 2023-08-31
Add base bridging

yabirgb:
image


rotkehlchen/chain/ethereum/modules/base_bridge/decoder.py
```diff
@@ -0,0 +1,165 @@
...
+            return DEFAULT_DECODING_OUTPUT
+
+        # Read information from event's topics & data
+        if context.tx_log.topics[0] == TRANSACTION_DEPOSITED:
+            from_address = hex_or_bytes_to_address(context.tx_log.topics[1])
+            to_address = hex_or_bytes_to_address(context.tx_log.topics[2])
+        else:
+            from_address = context.transaction.from_address
+            to_address = context.transaction.from_address
+
```
dimyG, line 105: Maybe use a variable for context.transaction.from_address since it is used twice here? Not a big deal of course.

dimyG (commented):
LGTM. Just a small comment.

rotkehlchen/chain/ethereum/modules/base_bridge/decoder.py
yabirgb: mmm I believe it is not needed since you will do an extra memory pointer for no reason

LefterisJP (approved):
tiny nitpick. All backend tests passed. And I guess it's fine and tested by you. Tests look okay. Will merge.

PR #6567, opened 2023-08-31
fixes airdrops claimed

dimyG:
Closes #(issue_number)

## Checklist


LefterisJP (commented):
code change makes sense. But perhaps would make sense to be checked by the people who tested with the frontend too

yabirgb:
@LefterisJP I have checked it and works. What is left is the cow airdrop in gnosis that is not shown as claimed because we don't support gnosis yet but is a known limitation

LefterisJP:
> @LefterisJP I have checked it and works. What is left is the cow airdrop in gnosis that is not shown as claimed because we don't support gnosis yet but is a known limitation

That's unrelated to the PR I guess since it's not possible to show as claimed without gnosis support.

LefterisJP (approved):
LGTM then. LEt's see how tests do

PR #6573, opened 2023-09-05
Point to new yearn API domain

LefterisJP:


TomAFrench:
@LefterisJP By the way, updating to `actions/cache@v4` should fix the CI issues on this PR.

LefterisJP:
> @LefterisJP By the way, updating to `actions/checkout@v4` should fix the CI issues on this PR. We were running into the same issue yesterday.

Thank you. Merged your PR for this.

LefterisJP:
So seems that just as all is green here api.yearn.finance is back online. But still they said they would keep both runnign in parallel so we can still do this change.

PR #6576, opened 2023-09-05
Adds velodrome data to globaldb cache

dimyG:
Closes #(issue_number)

## Checklist


rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -0,0 +1,83 @@
...
+            base_tools: 'BaseDecoderTools',
+            msg_aggregator: 'MessagesAggregator',
+    ) -> None:
+        super().__init__(
+            evm_inquirer=optimism_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+        )
+        self.curve_pools, self.curve_gauges = read_velodrome_pools_and_gauges_from_cache()
+        self.inquirer = optimism_inquirer
```
yabirgb, line 47: this is already set in `super` as `evm_inquirer`, no need to re-assign

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -0,0 +1,83 @@
...
+            optimism_inquirer: 'OptimismInquirer',
+            base_tools: 'BaseDecoderTools',
+            msg_aggregator: 'MessagesAggregator',
+    ) -> None:
+        super().__init__(
+            evm_inquirer=optimism_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+        )
+        self.curve_pools, self.curve_gauges = read_velodrome_pools_and_gauges_from_cache()
```
yabirgb, line 46: copy paste, rename

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -0,0 +1,83 @@
...
+        """Make sure velodrome pools are recently queried from the chain, saved in the DB
+        and loaded to the decoder's memory.
+
+        If a query happens and any new mappings are generated they are returned,
+        otherwise `None` is returned.
+        """
+        self.inquirer.ensure_velodrome_cache_data_is_updated()
+        new_pools, new_gauges = read_velodrome_pools_and_gauges_from_cache()
+        pools_diff = new_pools - self.curve_pools
+        gauges_diff = new_gauges - self.curve_gauges
```
yabirgb, line 68: copy pasted name

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def _decode_velodrome_gauge_events(self) -> None:
+        pass
+
+    def counterparties(self) -> list['CounterpartyDetails']:
+        return [CounterpartyDetails(identifier=CPT_VELODROME, label='velodrome_finance', image='velodrome.svg')]  # noqa: E501
+
+    def reload_data(self) -> Optional[Mapping[ChecksumEvmAddress, tuple[Any, ...]]]:
+        """Make sure velodrome pools are recently queried from the chain, saved in the DB
+        and loaded to the decoder's memory.
```
yabirgb, line 60: if the logic is the same as in curve (i believe is almost identical) maybe you can abstract it

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
```diff
@@ -0,0 +1,218 @@
...
+            ))
+        try:
+            db_addressbook.add_addressbook_entries(
+                write_cursor=write_cursor,
+                entries=addresbook_entries,
+            )
+        except InputError as e:
+            log.debug(
+                f'Velodrome address book names for pool {pool.pool_address} were not added. '
+                f'Probably names were added by the user earlier. {e!s}')
```
yabirgb, line 81:
```suggestion
            log.debug(
                f'Velodrome address book names for pool {pool.pool_address} were not added. '
                f'Probably names were added by the user earlier. {e}',
           )
```

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
```diff
@@ -0,0 +1,218 @@
...
+            )
+
+
+def read_velodrome_pools_and_gauges_from_cache() -> tuple[set[ChecksumEvmAddress], set[ChecksumEvmAddress]]:  # noqa: E501
+    """
+    Reads globaldb cache and returns:
+    - A set of all known addresses of velodrome pools.
+    - A set of all known addresses of velodrome gauges.
+
+    Doesn't raise anything unless cache entries were inserted incorrectly.
```
yabirgb, line 102: then it can raise something no? write what it is as I don't have clear what do you mean

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+    gauge_addresses = set()
+    pool_addresses = set()
+    with GlobalDBHandler().conn.read_ctx() as cursor:
+        str_pool_addresses = globaldb_get_general_cache_values(cursor=cursor, key_parts=(CacheType.VELODROME_POOL_ADDRESS,))  # noqa: E501
+        for pool_address in str_pool_addresses:
+            gauge_address = globaldb_get_unique_cache_value(cursor=cursor, key_parts=(CacheType.VELODROME_GAUGE_ADDRESS, pool_address))  # noqa: E501
+            if gauge_address is None:
+                continue
```
yabirgb, line 111: shouldn't you always save the pool address?

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
```diff
@@ -0,0 +1,218 @@
...
+
+    May raise:
+    - RemoteError if there is an error connecting with the remote source of data
+    """
+    data_contract = inquirer.contracts.contract(VELODROME_SUGAR_V2_CONTRACT)
+    pool_data = data_contract.call(
+        node_inquirer=inquirer,
+        method_name='all',
+        arguments=[1000, 0, '0x48a264d1BC1765598D327bB60789bEd541b12504'],  # limit, offset, address  # no need to use a specific address here we just want the pool data  # noqa: E501
+    )
```
yabirgb, line 134: Maybe you should have a lower limit and loop trough the responses. As you have it now there are potential pools that you are missing. Also I'm not sure but I believe it can happen that the query fails depending on the configuration for the max gas used in the queried node.

I would change this to having a lower limit and playing with the offset. Also can you explain that the `address` is something from the design of this interface and to cover one of the use cases that the developers had in mind and that we don't need it since all the pools are returned

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
```diff
@@ -0,0 +1,218 @@
...
+
+    pools = []
+    for pool in pool_data:
+        try:
+            pool_address = deserialize_evm_address(pool[0])
```
yabirgb, line 139: the addresses already come checksummed... isn't that the case? If not you would need to catch the possible error here but I believe you only need `string_to_evm_address`. The issue with `deserialize_evm_address` is taht is expensive to call

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
```diff
@@ -0,0 +1,218 @@
...
+            if pool_address in existing_pools:
+                continue
+            gauge_address = deserialize_evm_address(pool[11])
+            pools.append(
+                VelodromePoolData(
+                    pool_address=pool_address,
+                    pool_name=pool[1],
+                    gauge_address=gauge_address if gauge_address != ZERO_ADDRESS else None,
+                    coins=[deserialize_evm_address(coin) for coin in [pool[5], pool[8]]],  # token0 and token1 of the pool  # noqa: E501
```
yabirgb, line 148: same comment as before for `deserialize_evm_address`

rotkehlchen/data/global.db
yabirgb, line 1: can you post the diff in the globaldb?

yabirgb (commented):
Some comments about the approach, if any is not clear please let's discuss it in discord

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
dimyG: No they are not checksummed so we need these deserialize_evm_address calls. What error you refer to? There is already an except statement for catching DeserializationError.

rotkehlchen/data/global.db
dimyG: https://gist.github.com/dimyG/1e96254d482b174cf4bac14462987b09  currently it's only the new VELODROME_SUGAR_V2_CONTRACT contract.  Let's finalize the form of the velodrome cache data and I will update it with that data too.

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -0,0 +1,83 @@
...
+from rotkehlchen.types import ChecksumEvmAddress
+
+if TYPE_CHECKING:
+    from rotkehlchen.chain.evm.decoding.base import BaseDecoderTools
+    from rotkehlchen.chain.optimism.manager import OptimismInquirer
+    from rotkehlchen.user_messages import MessagesAggregator
+
+
+ROUTER_V2 = string_to_evm_address('0xa062ae8a9c5e11aaa026fc2670b0d65ccc8b2858')
+ADD_LIQUIDITY_EVENT = b'L \x9b_\xc8\xadPu\x8f\x13\xe2\xe1\x08\x8b\xa5jV\r\xffi\n\x1co\xef&9OL\x03\x82\x1cO'  # Mint event (mints LP tokens) # 0x4c209b5fc8ad50758f13e2e1088ba56a560dff690a1c6fef26394f4c03821c4f  # noqa: E501
```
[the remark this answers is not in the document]
dimyG: ok I removed them. These where there just temporarily because the work on this PR needs to continue by adding the decoding functionality.

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
dimyG: It depends on what new mappings have to be added. The decoding logic is not written yet.

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
```diff
@@ -0,0 +1,201 @@
...
+            for address in globaldb_get_general_cache_like(cursor=cursor, key_parts=(CacheType.VELODROME_POOL_ADDRESS,))  # noqa: E501
+        ]
+    pools_data = None
+    try:
+        pools_data = query_velodrome_data_from_chain(
+            inquirer=inquirer,
+            existing_pools=existing_pools,
+        )
+    except RemoteError as err:
+        log.error(f'Could not query chain for velodrome pools due to: {err}')
```
yabirgb, line 199:
```suggestion
    try:
        pools_data = query_velodrome_data_from_chain(
            inquirer=inquirer,
            existing_pools=existing_pools,
        )
    except RemoteError as err:
        log.error(f'Could not query chain for velodrome pools due to: {err}')
        pools_data = None
```

yabirgb (commented):
This is missing for what I see:

1. In the task manager the task that updates the velodrome cache
2. Unless I missed it something like in `ethereum/manager.py` for `self.node_inquirer.ensure_cache_data_is_updated(`
3. In `api/rest.py` we have `refresh_general_cache` that also should refresh velodrome pools

dimyG:
> 1. In the task manager the task that updates the velodrome cache
> 2. Unless I missed it something like in ethereum/manager.py for self.node_inquirer.ensure_cache_data_is_updated(

As we discussed in discord there is no need for such tasks. The curve related task could be removed too. A new issue will be opened for this.

dimyG:
This is the latest sqldiff for the packaged db  https://gist.github.com/dimyG/cfe04d6672f520bdfe82a0422d18259f

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -1386,3 +1389,70 @@ def __init__(
...
+            contract_multicall=contract_multicall,
+            rpc_timeout=rpc_timeout,
+            native_token=native_token,
+            dsproxy_registry=dsproxy_registry,
+        )
+
+    @protect_with_lock()
+    def ensure_cache_data_is_updated(
+            self,
+            cache_type: 'CacheType',
```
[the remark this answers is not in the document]
yabirgb: this hasn't been done but can be done after lefteris review

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
```diff
@@ -0,0 +1,217 @@
...
+                )
+            except NotERC20Conformant as e:
+                skip_pool = True
+                log.error(
+                    f'Skipping token {token} because it is not a valid ERC20 token. {e}',
+                )
+
+        if skip_pool is True:
+            log.error(f'Skipping velodrome pool {pool_address} because one of its tokens is not a valid ERC20')  # noqa: E501
+            continue
```
yabirgb, line 184: why this and not just add the continue in the except part?

yabirgb (commented):
2 comments

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
dimyG: It will continue the token for loop, not the pool loop.

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
```diff
@@ -0,0 +1,212 @@
...
+                get_or_create_evm_token(
+                    userdb=inquirer.database,
+                    evm_address=token,
+                    chain_id=ChainID.OPTIMISM,
+                    evm_inquirer=inquirer,
+                    seen=TokenSeenAt(description='Querying velodrome pools'),
+                )
+            except NotERC20Conformant as e:
+                log.error(
+                    f'Skipping token {token} because it is not a valid ERC20 token. {e}',
```
LefterisJP, line 177:
```suggestion
                    f'Skipping velodrome token {token} because it is not a valid ERC20 token. {e}',
```

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
```diff
@@ -0,0 +1,212 @@
...
+                    pool_name=pool[1],
+                    gauge_address=gauge_address if gauge_address != ZERO_ADDRESS else None,
+                    coins=[token0, token1],
+                ),
+            )
+
+    return pools
+
+
+def query_velodrome_data(inquirer: 'OptimismInquirer') -> Optional[list[VelodromePoolData]]:
```
LefterisJP, line 193: This function is identical to `query_curve_data`. Again can you please abstract this common functionality?

rotkehlchen/types.py
```diff
@@ -905,10 +905,15 @@ class CacheType(Enum):
...
     CURVE_POOL_UNDERLYING_TOKENS = auto()  # get underlying tokens by pool address
+    VELODROME_POOL_ADDRESS = auto()  # get pool address information
+    VELODROME_GAUGE_ADDRESS = auto()  # get gauge address by pool address
 
     def serialize(self) -> str:
         # Using custom serialize method instead of SerializableEnumMixin since mixin replaces
         # `_` with ` ` and we don't need spaces here
+        if 'VELODROME' in self.name:
+            parts = self.name.split('_')
+            return parts[0][:4] + parts[1][0]  # Shorten the name that is stored in the db to save space. For example: VELODROME_POOL_ADDRESS -> VELOP  # noqa: E501
```
LefterisJP, line 916: for the future: We probably need to do similar for all. And maybe even shorter

LefterisJP (commented):
Thanks. Looking all right for a first review. Left a few comments and some asks, mostly related to abstraction of common functionality across all caches.

LefterisJP (approved):
LGTM.

As discussed morework needs to happen on abstraction but this can also go in a subsequent PR.

PR #6578, opened 2023-09-07
feat:add the collateral ratio of the Liquity protocol in troves.py

### LINKED ISSUE #3971: Add status of the Liquity protocol to the frontend
The collateral ratio of the Liquity protocol is important to detect when liquidations in recovery mode can happen. More at https://docs.liquity.org/faq/recovery-mode. It would be nice to have it in rotki since now you have to visit other frontends to know if your position is safe or not

For this to be added the backend needs to call the contracts and inspect the status of the protocol and the frontend needs to properly display this information.

We need to call https://etherscan.io/address/0xA39739EF8b0231DbFA0DcdA07d7e29faAbCf4bb2#readContract for 
image

and do
```
(systemColl * eth price) / (systemDebt * lusd price)
```

3scava1i3r:
Closes #3971 

## Checklist

- [x] The PR modified the backend for getting collateral ratio of liquity protocol and changes the way get_position() data is retrieved

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/liquity/trove.py
+++ b/rotkehlchen/chain/ethereum/modules/liquity/trove.py
@@ -84,8 +84,32 @@ class Liquity(HasDSProxy):
     def get_positions(
             self,
             given_addresses: Sequence[ChecksumEvmAddress],
-    ) -> dict[ChecksumEvmAddress, Trove]:
-        """Query liquity contract to detect open troves"""
+    ) -> dict[str, Any]:
+        """Query liquity contract to get the collateral ratio and detect open troves"""
+
+        eth_price = Inquirer().find_usd_price(A_ETH)
+        lusd_price = Inquirer().find_usd_price(A_LUSD)
+
+        try:
+            system_collateral = self.trove_manager_contract.call(
+                node_inquirer=self.ethereum,
+                method_name='getEntireSystemColl',
+                arguments=[],
+            )
+            system_debt = self.trove_manager_contract.call(
+                node_inquirer=self.ethereum,
+                method_name='getEntireSystemDebt',
+                arguments=[],
+            )
+
+            if system_debt != 0:
+                final_collateral_ratio = eth_price * system_collateral / system_debt * lusd_price
+            else:
+                final_collateral_ratio = None
+
+        except RemoteError as e:
+            log.debug(f'Failed {e!s}')
+
         addresses = list(given_addresses)  # turn to a mutable list copy to add proxies
         proxied_addresses = self.ethereum.proxies_inquirer.get_accounts_having_proxy()
         proxies_to_address = {v: k for k, v in proxied_addresses.items()}
@@ -101,8 +125,6 @@ class Liquity(HasDSProxy):
         )
 
         data: dict[ChecksumEvmAddress, Trove] = {}
-        eth_price = Inquirer().find_usd_price(A_ETH)
-        lusd_price = Inquirer().find_usd_price(A_LUSD)
         for idx, output in enumerate(outputs):
             status, result = output
             if status is True:
@@ -159,7 +181,8 @@ class Liquity(HasDSProxy):
                         f'Ignoring Liquity trove information. '
                         f'Failed to decode contract information. {e!s}.',
                     )
-        return data
+        final_result = {'balances': data, 'final_collateral_ratio': final_collateral_ratio}
+        return final_result
 
     def _query_deposits_and_rewards(
             self,
--- a/rotkehlchen/chain/aggregator.py
+++ b/rotkehlchen/chain/aggregator.py
@@ -1072,7 +1072,7 @@ class ChainsAggregator(CacheableMixIn, LockableQueryMixIn):
             liquity_addresses = self.queried_addresses_for_module('liquity')
             # Get trove information
             liquity_balances = liquity_module.get_positions(given_addresses=liquity_addresses)
-            for address, deposits in liquity_balances.items():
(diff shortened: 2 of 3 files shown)


rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -84,8 +84,32 @@ def __init__(
...
+            )
+
+            if system_debt != 0:
+                final_collateral_ratio = eth_price * system_collateral / system_debt * lusd_price
+            else:
+                final_collateral_ratio = None
+
+        except RemoteError as e:
+            log.debug(f'Failed {e!s}')
+
```
yabirgb, line 112: We try to keep try/except statements as short as possible and only around the code that can raise it. Can you add it only around the contract calls. Also is easier to search later if it has a better message

```suggestion

        except RemoteError as e:
            log.error(f'Failed to query liquity contracts for protocol debt: {e}')

```

rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -84,8 +84,32 @@ def __init__(
...
+                arguments=[],
+            )
+
+            if system_debt != 0:
+                final_collateral_ratio = eth_price * system_collateral / system_debt * lusd_price
+            else:
+                final_collateral_ratio = None
+
+        except RemoteError as e:
+            log.debug(f'Failed {e!s}')
```
yabirgb, line 111: could you move this to a "private" method to keep this logic cleaner?

rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -84,8 +84,32 @@ def __init__(
...
+                arguments=[],
+            )
+            system_debt = self.trove_manager_contract.call(
+                node_inquirer=self.ethereum,
+                method_name='getEntireSystemDebt',
+                arguments=[],
+            )
+
+            if system_debt != 0:
+                final_collateral_ratio = eth_price * system_collateral / system_debt * lusd_price
```
yabirgb, line 106: can we rename this to `total_collateral_ratio`? I believe is what they use

rotkehlchen/chain/ethereum/modules/liquity/trove.py
yabirgb: Also when moving it make it take the price as argument and restore the call to where it was. I believe the reason is that if any remote query fails you don't make extra price queries that won't be used

rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -84,8 +84,32 @@ def __init__(
     def get_positions(
             self,
             given_addresses: Sequence[ChecksumEvmAddress],
-    ) -> dict[ChecksumEvmAddress, Trove]:
-        """Query liquity contract to detect open troves"""
+    ) -> dict[str, Any]:
+        """Query liquity contract to get the collateral ratio and detect open troves"""
+
```
yabirgb, line 89:
```suggestion
        """Detect open troves for users and query total collateral ratio of the protocol"""

```

rotkehlchen/chain/ethereum/modules/liquity/trove.py
```diff
@@ -84,8 +84,32 @@ def __init__(
     def get_positions(
             self,
             given_addresses: Sequence[ChecksumEvmAddress],
-    ) -> dict[ChecksumEvmAddress, Trove]:
-        """Query liquity contract to detect open troves"""
+    ) -> dict[str, Any]:
```
yabirgb, line 87: Can you maybe use a namedtuple here or a typed dict (https://peps.python.org/pep-0589/) This will help mypy and prevent errors for example in the aggregator

yabirgb (commented):
You forgot to update the `docs/api.rst` file that contains information about the response of the endpoint that you modified, sadly we don't have autogenerated docs

3scava1i3r:
> You forgot to update the `docs/api.rst` file that contains information about the response of the endpoint that you modified, sadly we don't have autogenerated docs

added the same

rotkehlchen/chain/ethereum/modules/liquity/trove.py
3scava1i3r: for the response object right?
have added one for the same

yabirgb (commented):
Thank you very much and sorry for taking so long to look at this. Just a few minor comments. Was also just checking yesterday on the frontend team in case there was an issue for the api design and if it would make sense to show the collateral with the troves, they said that is a good place for it and then the api design is okey.

If you notice there is a test that failed to execute. It has some special mocking that we were doing before introducing VCR. The reason why it fails is because there is a remote call to the contract (for the two values that you request) that are not found there. Maybe it would make sense to remove this and migrate to VCR as we do for other tests, the setup is a bit more complicated since you need to create a PR in other repo with the cassette. THis would help you https://rotki.readthedocs.io/en/stable/contribute.html#mocking-networking-in-the-tests

Please if you find any issue with this ping me in discord and I'll check with to answer asap. Also sorry for the delay in the review again.

3scava1i3r:
Just confirming that I am still working on the same

LefterisJP:
> Just confirming that I am still working on the same

hey @3scava1i3r the PR has been open for 9 months with no activity. Please respond within this week if you can finish it or we can either close it or take it over and finish it ourselves.

LefterisJP:
Closing due to inactivity

PR #6580, opened 2023-09-08
Prevent users from merging evm tokens

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/api/v1/schemas.py
+++ b/rotkehlchen/api/v1/schemas.py
@@ -40,6 +40,7 @@ from rotkehlchen.chain.substrate.utils import (
 )
 from rotkehlchen.constants.assets import A_ETH, A_ETH2
 from rotkehlchen.constants.misc import ONE, ZERO
+from rotkehlchen.constants.resolver import EVM_CHAIN_DIRECTIVE
 from rotkehlchen.data_import.manager import DataImportSource
 from rotkehlchen.db.filtering import (
     AssetMovementsFilterQuery,
@@ -2114,6 +2115,18 @@ class AssetsMappingSchema(Schema):
 class AssetsReplaceSchema(Schema):
     source_identifier = fields.String(required=True)
     target_asset = AssetField(required=True, expected_type=Asset, form_with_incomplete_data=True)
+
+    @validates_schema
+    def validate_schema(
+            self,
+            data: dict[str, Any],
+            **_kwargs: Any,
+    ) -> None:
+        if data['source_identifier'].startswith(EVM_CHAIN_DIRECTIVE):
+            raise ValidationError(
+                message="Can't merge two evm tokens",
+                field_name='source_identifier',
+            )
 
 
 class QueriedAddressesSchema(Schema):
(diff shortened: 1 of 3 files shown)


rotkehlchen/api/v1/schemas.py
```diff
@@ -2115,6 +2116,25 @@ class AssetsReplaceSchema(Schema):
...
+            self,
+            data: dict[str, Any],
+            **_kwargs: Any,
+    ) -> None:
+        try:
+            source_is_evm_token = Asset(data['source_identifier']).is_evm_token()
+        except UnknownAsset:
+            can_merge = True  # we allow to merge assets that are not in the globaldb
+        else:
+            can_merge = not (source_is_evm_token and data['target_asset'].is_evm_token())
```
LefterisJP, line 2130: This can fail. If we get to the except above, `source_is_evm_token` will not exist. Weird that linters did not see this

rotkehlchen/api/v1/schemas.py
```diff
@@ -2115,6 +2116,25 @@ class AssetsReplaceSchema(Schema):
...
+
+        if can_merge is False:
```
LefterisJP, line 2132: This logic also can improve.

We need the source to not be an evm token.
The target can be an evm token I think. Say an asset got merged into an evm token so for ease of use the user wants to replace all old appearances of the non evm token with the evm token.

And then since the source token can also not be in the global DB we need to check for the evm chain directive instead to detect the problem also if it's not in the DB.

```suggestion
        if data['source_identifier'].startswith(EVM_CHAIN_DIRECTIVE)
```

rotkehlchen/tests/api/test_user_assets.py
```diff
@@ -872,6 +872,20 @@ def assert_db() -> None:
     )
     assert_proper_response(response)
 
+    # combine two tokens
+    response = requests.put(
```
LefterisJP, line 876: Do in a for loop all possible combos of evm tokens/nonevm assets

rotkehlchen/api/v1/schemas.py
yabirgb: The else executes when there is no exception raised. Finally is what always executes

LefterisJP: ah yes indeed. But read the other comment :P

rotkehlchen/tests/api/test_user_assets.py
```diff
@@ -863,14 +863,29 @@ def assert_db() -> None:
...
 
-    response = requests.put(
-        api_url_for(
-            rotkehlchen_api_server,
-            'assetsreplaceresource',
-        ),
-        json={'source_identifier': glm_id, 'target_asset': 'ICP'},
-    )
-    assert_proper_response(response)
+    # try to merge combinations of EVM tokens/other assets. When the source is a evm token it
```
LefterisJP, line 866:
```suggestion
    # try to merge combinations of EVM tokens/other assets. When the source is an evm token it
```

PR #6581, opened 2023-09-08
Removes the update curve cache task

dimyG:
* The cache is updated on demand by the reload_data method of the ReloadableDecoderMixin decoders (decoders that need data from the cache) whenever an undecoded event needs to be decoded. Curve's decoder is one of them so there is no need for an additional task that updates the curve related cache tables.

Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/tests/unit/test_tasks_manager.py
+++ b/rotkehlchen/tests/unit/test_tasks_manager.py
@@ -75,7 +75,6 @@ def fixture_task_manager(
         exchange_manager=exchange_manager,
         deactivate_premium=lambda: None,
         query_balances=lambda: None,
-        update_curve_pools_cache=lambda: None,
         activate_premium=lambda _: None,
         msg_aggregator=messages_aggregator,
         data_updater=RotkiDataUpdater(msg_aggregator=messages_aggregator, user_db=database),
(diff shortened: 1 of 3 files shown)


rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -320,31 +319,6 @@ def test_update_snapshot_balances(task_manager):
         raise AssertionError(f'Update snapshot balances was not completed within {timeout} seconds') from e  # noqa: E501
 
 
-@pytest.mark.parametrize('globaldb_upgrades', [[]])
-@pytest.mark.parametrize('run_globaldb_migrations', [False])
-@pytest.mark.parametrize('custom_globaldb', ['v4_global_before_migration1.db'])
```
LefterisJP, line 325: Is `v4_global_before_migration1.db` used anywhere else? If not you should also remove that file.

LefterisJP (commented):
I would like @yabirgb to also have a look. Left a comment.

Also can you write in the PR why you remove the task. Explain why it is not needed.

rotkehlchen/tests/unit/test_tasks_manager.py
dimyG: Yes, it is used in a few other tests too so we still need it

PR #6585, opened 2023-09-09
Decoder for optimism bridge 2-step withdrawal process

LefterisJP:
More info: https://blog.oplabs.co/two-step-withdrawals/


LefterisJP (commented):
Will merge. All is green.

PR #6591, opened 2023-09-12
All kraken special assets should now be handled properly.

LefterisJP:
Fix #6857 . All Kraken special & staking assets, like staking assets bonded for a specific time will now be handled properly.


LefterisJP:
Failed tests are timeouts

PR #6594, opened 2023-09-14
Remove constraint in requirements

yabirgb:
multiformats had a dependency on jsonchema but one dependency for jsonschema didn't provice wheels and it used rust to compile. Now we have wheels for our platforms and the app seems to work now
`


LefterisJP (approved):
LGTM

PR #6598, opened 2023-09-14
Add Padding to ENS label_hash decoding to correctly resolve values with leading 0s

### LINKED ISSUE #6597: ENS label_hash values starting with 0 get truncated
## Problem Definition

If the labelhash value of an ENS record starts with 0, the leading 0, possibly 0s are truncated and missing, leading to an incorrect length value, halting transaction decoding.

In the example shared below in Logs, querying Graph adding a 0 in front of 126... resolves their ens correctly afterwards

I have a suggested PR to come that fixes this on line 201 of ens/decoder.py 

## Logs

Borrowing from another user's report in Discord:

`decode 246 ethereum trasactions died with exception: The Graph query to query{domains(first:1, where:{labelhash:"0x126792ee6b36f6e438e270b25a6f7e91acddfcb032104dbcb681371a902802b"}){labelName}} failed due to {'message': 'Failed to decode `Bytes` value: `Odd number of digits`'}. No retries left.. Check the logs for more details
`

-->

### System Description

Operating system:  LFS Linux and Mac
Rotki version: v1.30.1

solaasan:
Closes #6597

Padding the value to 64 resolves ENS leading with 0 fine now, no longer producing a Graph lookup from the invalid length request.

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/ens/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/ens/decoder.py
@@ -198,7 +198,7 @@ class EnsDecoder(DecoderInterface, CustomizableDateMixin):
             log.error(f'Could not decode an ERC721 transfer for an ENS name transfer: {context.transaction.tx_hash.hex()}')  # noqa: E501
             return DEFAULT_DECODING_OUTPUT
 
-        label_hash = hex(transfer_event.extra_data['token_id'])  # type: ignore[index]  # noqa: E501  # ERC721 transfer always has extra data
+        label_hash = "0x{:064x}".format(transfer_event.extra_data['token_id'])  # type: ignore[index]  # noqa: E501  # ERC721 transfer always has extra data
         try:
             result = self.graph.query(
                 querystr=f'query{{domains(first:1, where:{{labelhash:"{label_hash}"}}){{labelName}}}}')  # noqa: E501


rotkehlchen/chain/ethereum/modules/ens/decoder.py
```diff
@@ -198,7 +198,7 @@ def _decode_name_transfer(self, context: DecoderContext) -> DecodingOutput:
             log.error(f'Could not decode an ERC721 transfer for an ENS name transfer: {context.transaction.tx_hash.hex()}')  # noqa: E501
             return DEFAULT_DECODING_OUTPUT
 
-        label_hash = hex(transfer_event.extra_data['token_id'])  # type: ignore[index]  # noqa: E501  # ERC721 transfer always has extra data
+        label_hash = "0x{:064x}".format(transfer_event.extra_data['token_id'])  # type: ignore[index]  # noqa: E501  # ERC721 transfer always has extra data
```
LefterisJP, line 201: We use single quote strings in the codebase

```suggestion
        label_hash = '0x{:064x}'.format(transfer_event.extra_data['token_id'])  # type: ignore[index]  # noqa: E501  # ERC721 transfer always has extra data
```

LefterisJP (commented):
Hey thanks for the PR!

So two things.

1. We will need a test reproducing the error. It would need to be a decoding test like this: https://github.com/rotki/rotki/blob/238c6ab05d0acc3cc90417742bbc27e3a3794687/rotkehlchen/tests/unit/decoders/test_ens.py#L540 which just tries to decode a transaction hash which shows this problems and would fail without the fix

2. If your hypothesis is correct the fix may be correct but you should add a comment explaining why you need to add the padding. Explain that the graph query below expects a fully 64 bits padded hexstring

3. The problem I see is that this failure should not force a full error of all decoding. Unless I am missing something this error that is raised is a `RemoteError`. It's raised from inside `self.graph.query`. Then it's caught in the `except` clause below, an error is logged and `name_to_show` is made to be empty. But there should not be a total failure of decoding. Can you provide more information of how the error happened and maybe a trace?

LefterisJP (commented):
Also since this is a bugfix could you add a changelog entry?

LefterisJP:
Closing this PR since there has been no response in favour of https://github.com/rotki/rotki/pull/6610

Issue #6582, opened 2023-09-08
Introduce unsupported asset events cache for exchanges (and more?)


LefterisJP:
## Abstract

At the moment in all exchanges we skip an event completely if the asset is unknown. This can lead to missing events if we query an exchange's trades while a certain asset mapping is not yet made but is needed. For exampe I downloaded my kraken history before ETHPOW was added, so all related events are missing:

https://github.com/rotki/rotki/blob/47c84776384610ca3da216233d6bafec85f4446a/rotkehlchen/exchanges/kraken.py#L243

Think if this would also make sense in other places in the app. For other events where they are skipped in case of unknown asset.

## Specification

- Add a `skipped_external_events` table that will hold all events that would be skipped for unknown asset. It should have two columns. One should be the location and the other the raw json of the event before the decoding.
- In the linked code place and all other places across all exchanges where an event is skipped due to an unknown asset, instead of just skipping add the event to the aforementioned table.
- In either a periodic task, or a user generated process or both allow for reprocesssing of all skipped events, depending on their location. If something is processed properly, create a normal event for it and remove it from the skipped_external_events table. If not leave it there.
- Think if this should also apply somewhere else and not only exchanges.

yabirgb:
I'm not sure if a periodic task would make sense here since you need user actions / remote actions to trigger it. One scenario can be to run it after an asset update but having a periodic task running all the time is wasting cpu time

PR #6600, opened 2023-09-17
Introduce logic to save failed kraken processed events to the DB

LefterisJP:
An attempt to handle https://github.com/rotki/rotki/issues/6582 for Kraken. Also introduce an api call to request reprocessing of skipped events.


LefterisJP:
all green. Merging.

PR #6603, opened 2023-09-18
Add dependency needed for builds

yabirgb:
When building `rpds-py` in mac arm we need to build it from source and it uses rust so we need the mac arm target for rust in order to being able to build it


kelsos (approved):
lgtm, binary seems to work fine

PR #6606, opened 2023-09-18
bitstamp import

### LINKED ISSUE #6275: Support for Bitstamp csv import
## Abstract

Add support to import Bitstamp transactions csv

## Specification

Write a new parser for Bitstamp csv format.

Format example:

```
Type,Datetime,Account,Amount,Value,Rate,Fee,Sub Type
Deposit,"Jan. 28, 2022, 12:13 AM",Main Account,2.00000000 ETH,,,,
Market,"Jan. 28, 2022, 12:31 AM",Main Account,1.00000000 ETH,2214.01 EUR,2214.01 EUR,10.87005 EUR,Sell
Withdrawal,"Jan. 30, 2022, 11:40 AM",Main Account,2211.01 EUR,,,3.00 EUR,
```

jbrit:
Closes #6275

## Checklist

- [ ] The PR added a new importer for bitstamp transactions.


rotkehlchen/data_import/importers/bitstamp.py
```diff
@@ -0,0 +1,112 @@
...
+        - UnknownAsset
+        - DeserializationError
+        """
+        timestamp = deserialize_timestamp_from_date(
+            date=csv_row['Datetime'],
+            formatstr=timestamp_format,
+            location='Bitstamp',
+        )
+
+        amount, amount_symbol = csv_row['Amount'].split(' ')
```
yabirgb, line 39: this can raise `KeyError`. You already handle it but state it in the comment

yabirgb: Also this can raise `ValueError` if there is no enough values to unpack. That should be also be handled/stated in the docstring

rotkehlchen/data_import/importers/bitstamp.py
```diff
@@ -0,0 +1,112 @@
...
+                link='',
+                notes='',
+            )
+            self.add_trade(write_cursor, trade)
+        elif csv_row['Type'] == 'Deposit' or csv_row['Type'] == 'Withdrawal':
+            asset_movement = AssetMovement(
+                location=Location.BITSTAMP,
+                category=AssetMovementCategory.DEPOSIT
+                if csv_row['Type'] == 'Deposit'
+                else AssetMovementCategory.WITHDRAWAL,
```
yabirgb, line 76: for this we if/else conditions in assignments we always use one line and the comment # noqa: E501. It keeps the code cleaner

rotkehlchen/data_import/importers/bitstamp.py
```diff
@@ -0,0 +1,112 @@
...
+            bought_symbol = asset_from_bitstamp(amount_symbol)
+            sold_symbol = asset_from_bitstamp(value_symbol)
+
+            if trade_type == TradeType.SELL:
+                bought_amount, sold_amount = sold_amount, bought_amount
+                sold_symbol, bought_symbol = bought_symbol, sold_symbol
+
+            trade = Trade(
+                timestamp=timestamp,
+                location=Location.BITSTAMP,
```
yabirgb, line 59: Checl the review comment

rotkehlchen/data_import/importers/bitstamp.py
```diff
@@ -0,0 +1,112 @@
...
+                category=AssetMovementCategory.DEPOSIT
+                if csv_row['Type'] == 'Deposit'
+                else AssetMovementCategory.WITHDRAWAL,
+                address=None,
+                transaction_id=None,
+                timestamp=timestamp,
+                asset=asset_from_bitstamp(amount_symbol),
+                amount=amount,
+                fee=deserialize_fee(None),
+                fee_asset=A_USD,
```
yabirgb, line 83: is the fee always in USD?

rotkehlchen/tests/api/test_data_import.py
```diff
@@ -611,3 +612,21 @@ def test_bitcoin_tax_import(rotkehlchen_api_server):
...
+
+
+def test_bitstamp_import(rotkehlchen_api_server):
+    """Test that data import works for Bitstamp transaction csv files"""
+    rotki = rotkehlchen_api_server.rest_api.rotkehlchen
+    dir_path = Path(__file__).resolve().parent.parent
+
+    # First test a trades type csv import
+    filepath = dir_path / 'data' / 'bitstamp.csv'
+    json_data = {'source': 'bitstamp', 'file': str(filepath)}
```
yabirgb, line 624: please also update the api documentation under `docs/api.rst` and the user guide in the same folder

rotkehlchen/data_import/importers/bitstamp.py
```diff
@@ -0,0 +1,112 @@
...
+            trade = Trade(
+                timestamp=timestamp,
+                location=Location.BITSTAMP,
+                base_asset=bought_symbol,
+                quote_asset=sold_symbol,
+                trade_type=trade_type,
+                amount=bought_amount,
+                rate=rate,
+                fee=deserialize_fee(fee_amount),
+                fee_currency=asset_from_bitstamp(fee_symbol),
```
yabirgb, line 66: could you deserialize the assets earlier? if the asset is unknown for some reason you execute less logic

yabirgb (commented):
I believe we didn't mention this to you. When the app started we had the `Trade` and `AssetMovement` classes to represent events but we went to a more generic format based on what we call `HistoryEvents` that are like "building blocks" for events. We are starting the process to migrate the old logic to the new and standarize it. For reference you can check how we do it in other importers like https://github.com/rotki/rotki/blob/0c4b1999eb5146b3f5d632af560931dce249df89/rotkehlchen/data_import/importers/bitcoin_tax.py#L96

I'm not sure if the PR needs frontend work too. I believe is just adding one line somwhere with the icon but can be done in other PR. Just let you know that we add an entry to docs/changelog.rst when a issue is closed with a PR but if there is still frontend work we do this in the frontend PR.

Last please squash your commits to keep the history clean. Don't leave commits wiht fixing things inside the PR or cleaning

rotkehlchen/data_import/importers/bitstamp.py
```diff
@@ -0,0 +1,160 @@
...
+            bought_currency = asset_from_bitstamp(amount_symbol)
+            sold_currency = asset_from_bitstamp(value_symbol)
+
+            if trade_type == TradeType.SELL:
+                bought_amount, sold_amount = sold_amount, bought_amount
+                sold_currency, bought_currency = bought_currency, sold_currency
+
+            fee_amount = deserialize_fee(fee_amount)
+            fee_currency = asset_from_bitstamp(fee_symbol)
+
```
yabirgb, line 66: no need for this space
```suggestion
```

rotkehlchen/data_import/importers/bitstamp.py
```diff
@@ -0,0 +1,160 @@
...
+                event_identifier=event_identifier,
+                sequence_index=0,
+                timestamp=timestamp,
+                location=Location.BITSTAMP,
+                asset=sold_currency,
+                balance=Balance(
+                    amount=sold_amount,
+                    usd_value=ZERO,
+                ),
+                notes='',
```
yabirgb, line 77: Could you add a note to this events? if you open the app and navigate to history events you will see that there is redeable message with each event. Notes are important for the user to know what is happening.

Here you have an example of how we format them for evm events https://github.com/rotki/rotki/blob/7f4165af0c7ad43431e522345a36c244343bacd6/rotkehlchen/tests/unit/decoders/test_curve.py#L1216

rotkehlchen/tests/data/bitstamp.csv
yabirgb, line 1: Did you manage to get any other example from github/google?

yabirgb (commented):
Thanks, few more comments

rotkehlchen/tests/data/bitstamp.csv
jbrit: yes, I only saw fields that matched what we had

rotkehlchen/data_import/importers/bitstamp.py
```diff
@@ -0,0 +1,159 @@
...
+                sequence_index=2,
+                timestamp=timestamp,
+                location=Location.BITSTAMP,
+                asset=fee_currency,
+                balance=Balance(
+                    amount=fee_amount,
+                    usd_value=ZERO,
+                ),
+                notes=f'Fee of {fee_amount} {fee_currency} as the result of a trade on Bitstamp',
+                event_type=HistoryEventType.TRADE,
```
LefterisJP, line 105: If you check https://github.com/rotki/rotki/blob/develop/rotkehlchen/accounting/constants.py you will see that the combination of `type:TRADE, subtype:FEE` is not a valid/known combination.

This will result in the event as being invalid. The fee event is shown only when the type is SPEND and subtype FEE. So the event in this case won't appear in the UI, which is the exact reason we had to do https://github.com/rotki/rotki/pull/6553 for.

I think this may be a miscalculation on our side. And that perhaps we should just:

1. Add TRADE/FEE as a possible combination with `EventCategory.FEE` as result.
2. Replace SPEND/FEE with TRADE/FEE where needed, where it's just the 3rd part of a trade.

What do you think @yabir? Also do you think there is a nice way to make sure no part of our code produces invalid combinations? Otherwise we will end up finding this out when a user reports it

rotkehlchen/data_import/importers/bitstamp.py
```diff
@@ -0,0 +1,159 @@
...
+                except DeserializationError as e:
+                    self.db.msg_aggregator.add_warning(
+                        f'Deserialization error during Bitstamp CSV import. '
+                        f'{e!s}. Ignoring entry',
+                    )
+                    continue
+                except KeyError as e:
+                    raise InputError(f'Could not find key {e!s} in csv row {row!s}') from e
+                except ValueError as e:
+                    raise InputError(f'Could not parse some values {e!s} in csv row {row!s}') from e  # noqa: E501
```
LefterisJP, line 159: 1. You don't need to continue. There is nothing to continue for.
2. Why is a ValueError a fatal error skipping all other lines? Same question for KeyError.

```suggestion
            for row in data:
                try:
                    self._consume_bitstamp_transaction(write_cursor, row, **kwargs)
                except UnknownAsset as e:
                    self.db.msg_aggregator.add_warning(
                        f'During Bitstamp CSV import found action with unknown '
                        f'asset {e.identifier}. Ignoring entry',
                    )
                except DeserializationError as e:
                    self.db.msg_aggregator.add_warning(
                        f'Deserialization error during Bitstamp CSV import. '
                        f'{e!s}. Ignoring entry',
                    )
                except KeyError as e:
                    raise InputError(f'Could not find key {e!s} in csv row {row!s}') from e
                except ValueError as e:
                    raise InputError(f'Could not parse some values {e!s} in csv row {row!s}') from e  # noqa: E501
```

LefterisJP (commented):
left a review

rotkehlchen/data_import/importers/bitstamp.py
yabirgb: I would add Trade/Fee there. For the decoded events we have this list of expected events and (I have to double check if it is there) but there should be a piece of code that during tests check that the produced value for each event is in the known combinations for that decoder. We should do something similar for the imports

yabirgb: https://github.com/rotki/rotki/blob/fc3d4973db9c1274de4490f44ef4926ab7c4acb0/rotkehlchen/chain/evm/decoding/decoder.py#L961

jbrit: should I change it to SPEND/FEE pending when TRADE/FEE is added to valid combinations @LefterisJP

LefterisJP: No. Add Trade/FEE to valid combinations. Don't push changes when there is a pending review. It runs thetests for nothing

frontend/app/src/components/import/Bitstamp.vue
tewshi, line 1: rename to BitstampImport.vue

frontend/app/src/components/import/GroupedImport.vue
```diff
@@ -81,6 +81,12 @@ const sources = [
       () => import('@/components/import/BitcoinImport.vue')
     )
   },
+  {
+    identifier: 'bitstamp',
+    name: t('import_data.bitstamp.name'),
+    logo: './assets/images/protocols/bitstamp.svg',
+    form: defineAsyncComponent(() => import('@/components/import/Bitstamp.vue'))
```
tewshi, line 88: after renaming the file, also rename the import here.

LefterisJP (approved):
LGTM

LefterisJP:
Failing test is unrelated

PR #6610, opened 2023-09-19
Various ENS related bugfixes

LefterisJP:
### [Properly decode ENS labelhashes with odd number of digits](https://github.com/rotki/rotki/commit/be093b1cd0b0ec468efae7c3864e7a3c48d81445) 

Fix https://github.com/rotki/rotki/issues/6597

Also the wrong import of RemoteError from multiprocessing.managers in
two parts of the code is fixed here. That lead to the RemoteError not
being caught and as such the entire decoding process halting with the
graph error.

### [Fix event amount for ENS registrations with a refund](https://github.com/rotki/rotki/commit/d12caf11f7429ddd874cc7359f262cb93984d341)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` ENS registrations that involve a refund to the user due to paying too much should reflect the proper amount in the decoded event.
 * :bug:`6597` Transactions involving ENS labelhashes with odd number of digits will now be decoded correctly and the entire decoding process should not halt if they are encountered.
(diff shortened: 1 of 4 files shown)


LefterisJP:
Only failed test is unrelated.

```
rotkehlchen/tests/unit/test_nfts.py::test_addresses_queried_for_nfts[ethereum_modules0-True-ethereum_accounts0] - rotkehlchen.errors.misc.RemoteError: Reached opensea backoff limit after we incrementally backed off for https://api.opensea.io/api/v1/collection/mimatic-vault-v2/stats?name=mimatic-vault-v2
```

Perhaps it would make sense to VCR it (making sure to not record api key)? Opensea queries can do this quite often

PR #6612, opened 2023-09-19
adds decoder for velodrome

dimyG:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/optimism/modules/velodrome/decoder.py
+++ b/rotkehlchen/chain/optimism/modules/velodrome/decoder.py
@@ -1,27 +1,54 @@
+import logging
 from collections.abc import Mapping
 from typing import TYPE_CHECKING, Any, Optional
 
+from rotkehlchen.accounting.structures.evm_event import EvmEvent, EvmProduct
+from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.chain.ethereum.utils import asset_normalized_value
+from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
 from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface, ReloadableDecoderMixin
-from rotkehlchen.chain.evm.decoding.types import CounterpartyDetails
+from rotkehlchen.chain.evm.decoding.structures import (
+    DEFAULT_DECODING_OUTPUT,
+    DecoderContext,
+    DecodingOutput,
+)
+from rotkehlchen.chain.evm.decoding.types import CounterpartyDetails, EventCategory
+from rotkehlchen.chain.evm.decoding.utils import maybe_reshuffle_events
+from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.chain.optimism.modules.velodrome.constants import CPT_VELODROME
 from rotkehlchen.chain.optimism.modules.velodrome.velodrome_cache import (
     query_velodrome_data,
     read_velodrome_pools_and_gauges_from_cache,
     save_velodrome_data_to_cache,
 )
-from rotkehlchen.types import CacheType, ChecksumEvmAddress
+from rotkehlchen.logging import RotkehlchenLogsAdapter
+from rotkehlchen.types import CacheType, ChecksumEvmAddress, DecoderEventMappingType
+from rotkehlchen.utils.misc import hex_or_bytes_to_address, hex_or_bytes_to_int
 
 if TYPE_CHECKING:
     from rotkehlchen.chain.evm.decoding.base import BaseDecoderTools
+    from rotkehlchen.chain.evm.structures import EvmTxReceiptLog
     from rotkehlchen.chain.optimism.manager import OptimismInquirer
     from rotkehlchen.user_messages import MessagesAggregator
 
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+ROUTER_V2 = string_to_evm_address('0xa062aE8A9c5e11aaA026fc2670B0D65cCc8B2858')
+ROUTER_V1 = string_to_evm_address('0x9c12939390052919aF3155f41Bf4160Fd3666A6f')
+ADD_LIQUIDITY_EVENT = b'L \x9b_\xc8\xadPu\x8f\x13\xe2\xe1\x08\x8b\xa5jV\r\xffi\n\x1co\xef&9OL\x03\x82\x1cO'  # Mint event (mints LP tokens) same for v1 and v2  # noqa: E501
+REMOVE_LIQUIDITY_EVENT_V2 = b']bJ\xa9\xc1H\x15:\xb3Dl\x1b\x15Of\x0e\xe7p\x1eT\x9f\xe9\xb6-\xabqq\xb1\xc8\x0eo\xa2'  # Burn event (burns LP tokens)  # noqa: E501
+REMOVE_LIQUIDITY_EVENT_V1 = b'\xdc\xcdA/\x0b\x12R\x81\x9c\xb1\xfd3\x0b\x93"L\xa4&\x12\x89+\xb3\xf4\xf7\x89\x97nm\x81\x93d\x96'  # noqa: E501
+SWAP_V2 = b'\xb3\xe2w6\x06\xab\xfd6\xb5\xbd\x919K:T\xd19\x836\xc6P\x05\xba\xf7\xbfz\x05\xef\xef\xfa\xf7['  # noqa: E501
+SWAP_V1 = b'\xd7\x8a\xd9_\xa4l\x99KeQ\xd0\xda\x85\xfc\'_\xe6\x13\xce7e\x7f\xb8\xd5\xe3\xd10\x84\x01Y\xd8"'  # noqa: E501
+GAUGE_DEPOSIT_V2 = b'UH\xc87\xab\x06\x8c\xf5j,$y\xdf\x08\x82\xa4\x92/\xd2\x03\xed\xb7Qs!\x83\x1d\x95\x07\x8c_b'  # noqa: E501
+GAUGE_WITHDRAW_V2 = b'\x88N\xda\xd9\xceo\xa2D\r\x8aT\xcc\x124\x90\xeb\x96\xd2v\x84y\xd4\x9f\xf9\xc76a%\xa9BCd'  # noqa: E501
+CLAIM_REWARDS_V2 = b"\x1f\x89\xf9c3\xd3\x130\x00\xeeDts\x15\x1f\xa9`eC6\x8f\x02'\x1c\x9d\x95\xae\x14\xf1;\xccg"  # noqa: E501
+
 
 class VelodromeDecoder(DecoderInterface, ReloadableDecoderMixin):
-    """
-    A decoder class for velodrome related events.
-    TODO: This is mainly a skeleton for now. It should be implemented properly.
-    """
+    """A decoder class for velodrome related events."""
 
     def __init__(
             self,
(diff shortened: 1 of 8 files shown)


rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -36,11 +61,227 @@ def __init__(
...
+            ):
+                event.event_type = HistoryEventType.DEPOSIT
+                event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
+                event.counterparty = CPT_VELODROME
+                event.notes = f'Deposit {event.balance.amount} {crypto_asset.symbol} in velodrome pool {event.address}'  # noqa: E501
+                event.product = EvmProduct.POOL
+            elif (
+                    event.event_type == HistoryEventType.RECEIVE and
+                    event.event_subtype == HistoryEventSubType.NONE and
+                    event.address == ZERO_ADDRESS
```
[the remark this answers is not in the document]
dimyG: To add an additional check like event.asset = asset_from_topics right? No there is no asset in the topics. It is a trasnfer event from the zero address to an address and the amount in the data.

rotkehlchen/tests/unit/decoders/test_velodrome.py
```diff
@@ -0,0 +1,510 @@
...
+            location=Location.OPTIMISM,
+            event_type=HistoryEventType.WITHDRAWAL,
+            event_subtype=HistoryEventSubType.REMOVE_ASSET,
+            asset=A_WETH_OPT,
+            balance=Balance(FVal('0.002487103206849621')),
+            location_label=user_address,
+            counterparty=CPT_VELODROME,
+            address=WETH_OP_POOL_ADDRESS,
+            product=EvmProduct.POOL,
+            notes=f'Remove 0.002487103206849621 WETH from velodrome pool {WETH_OP_POOL_ADDRESS}',
```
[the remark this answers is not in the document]
dimyG: ok, indeed it would be better that this is consistent. Maybe its better to do this in a new PR because it will affect a lot of unrelated code for uniswap, curve etc.

dimyG:
This is the sqldiff for the packaged globaldb: https://gist.github.com/dimyG/535a7f076774b0799a162854fe63cd22

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -36,11 +64,224 @@ def __init__(
...
+            tx_log: 'EvmTxReceiptLog',
+            decoded_events: list['EvmEvent'],
+    ) -> DecodingOutput:
+        """Decodes events that add liquidity to a velodrome pool"""
+        for event in decoded_events:
+            try:
+                crypto_asset = event.asset.resolve_to_crypto_asset()
+            except (UnknownAsset, WrongAssetType):
+                self.notify_user(event=event, counterparty=CPT_VELODROME)
+                raise
```
yabirgb, line 78: As you can check in https://github.com/rotki/rotki/blob/1169de3db17ef203a34d87a5fa49b8e54264ffe9/rotkehlchen/chain/evm/decoding/decoder.py#L362 we don't handle the wrongAssetType error. Should be there too

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -36,11 +64,224 @@ def __init__(
...
         self.pools, self.gauges = read_velodrome_pools_and_gauges_from_cache()
 
-    def _decode_velodrome_events(self) -> None:
-        pass
+    def _decode_add_liquidity_events(
+            self,
+            tx_log: 'EvmTxReceiptLog',
+            decoded_events: list['EvmEvent'],
+    ) -> DecodingOutput:
+        """Decodes events that add liquidity to a velodrome pool"""
```
yabirgb, line 72: state the errors that it can raise and also the version of velodrome that this is compatible with

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -36,11 +64,224 @@ def __init__(
...
+        return DEFAULT_DECODING_OUTPUT
 
-    def _decode_velodrome_gauge_events(self) -> None:
-        pass
+    def _decode_remove_liquidity_events(
+            self,
+            tx_log: 'EvmTxReceiptLog',
+            decoded_events: list['EvmEvent'],
+    ) -> DecodingOutput:
+        """Decodes events that remove liquidity from a velodrome pool"""
```
yabirgb, line 112: state that this is compatible with v1 and v2

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -36,11 +64,224 @@ def __init__(
...
+            ):
+                event.event_type = HistoryEventType.WITHDRAWAL
+                event.event_subtype = HistoryEventSubType.REMOVE_ASSET
+                event.counterparty = CPT_VELODROME
+                event.notes = f'Remove {event.balance.amount} {crypto_asset.symbol} from velodrome pool {tx_log.address}'  # noqa: E501
+                event.product = EvmProduct.POOL
+            elif (
+                event.event_type == HistoryEventType.INFORMATIONAL and
+                event.event_subtype == HistoryEventSubType.APPROVE
+            ):
```
yabirgb, line 142: the information events are also emitted by the pool address? If that is this case you can do the check `event.address in self.pools` (it might be in the topic) earlier and exit faster

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -36,11 +64,224 @@ def __init__(
...
+        )
+        return DEFAULT_DECODING_OUTPUT
+
+    def _decode_velodrome_pool_events(self, context: DecoderContext) -> DecodingOutput:
+        if context.tx_log.topics[0] in (REMOVE_LIQUIDITY_EVENT_V2, REMOVE_LIQUIDITY_EVENT_V1):
+            return self._decode_remove_liquidity_events(
+                tx_log=context.tx_log,
+                decoded_events=context.decoded_events,
+            )
+        if context.tx_log.topics[0] == ADD_LIQUIDITY_EVENT:
```
yabirgb, line 203: the add liquidity topic is the same for both?

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -36,11 +64,224 @@ def __init__(
...
+                tx_log=context.tx_log,
+                decoded_events=context.decoded_events,
+            )
+        if context.tx_log.topics[0] in (SWAP_V2, SWAP_V1):
+            return self._decode_swap(context=context)
+
+        return DEFAULT_DECODING_OUTPUT
+
+    def _decode_velodrome_gauge_events(self, context: DecoderContext) -> DecodingOutput:
+        if context.tx_log.topics[0] not in (GAUGE_DEPOSIT_V2, GAUGE_WITHDRAW_V2, CLAIM_REWARDS_V2):
```
yabirgb, line 214: missing docstring

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -36,11 +64,224 @@ def __init__(
...
+            return DEFAULT_DECODING_OUTPUT
+
+        user_or_contract_address = hex_or_bytes_to_address(context.tx_log.topics[1])
+        gauge_address = context.tx_log.address
+        raw_amount = hex_or_bytes_to_int(context.tx_log.data)
+        found_event_modifying_balances = False
+        for event in context.decoded_events:
+            crypto_asset = event.asset.resolve_to_crypto_asset()
```
yabirgb, line 222: here for example you don't do the try/catch. I would just remove it from the other ones and let the error raise naturally

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -36,11 +64,224 @@ def __init__(
...
+            return DEFAULT_DECODING_OUTPUT
+
+        maybe_reshuffle_events(
+            ordered_events=[spend_event, receive_event],
+            events_list=context.decoded_events,
+        )
+        return DEFAULT_DECODING_OUTPUT
+
+    def _decode_velodrome_pool_events(self, context: DecoderContext) -> DecodingOutput:
+        if context.tx_log.topics[0] in (REMOVE_LIQUIDITY_EVENT_V2, REMOVE_LIQUIDITY_EVENT_V1):
```
yabirgb, line 198: add small docstring

rotkehlchen/tests/unit/decoders/test_velodrome.py
```diff
@@ -0,0 +1,845 @@
...
+    token_type=EvmTokenKind.ERC20,
+)
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('optimism_accounts', [['0x78C13393Aee675DD7ED07ce992210750D1F5dB88']])
+def test_add_liquidity(optimism_transaction_decoder, optimism_accounts):
+    """Check that adding liquidity to a velodrome pool is properly decoded."""
+    tx_hex = deserialize_evm_tx_hash('0xa1cf038ca08b51971314b4f8e006b455f2a82f113899e5eb7818ac1528605bfc')  # noqa: E501
+    evmhash = deserialize_evm_tx_hash(tx_hex)
```
yabirgb, line 42: why are you doing this twice?

rotkehlchen/tests/unit/decoders/test_velodrome.py
```diff
@@ -0,0 +1,845 @@
...
+    ]
+    assert events == expected_events
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('optimism_accounts', [['0xE1343a4b5e64d47B0c0f208d05Fb4B5973443818']])
+def test_add_liquidity_v1(optimism_transaction_decoder, optimism_accounts):
+    """Check that adding liquidity to a velodrome pool is properly decoded."""
+    tx_hex = deserialize_evm_tx_hash('0x779c2c762ed7ca5d2236ac548f9e9bdc723a8bc9d6d8feba9e3baea9456bbac9')  # noqa: E501
+    evmhash = deserialize_evm_tx_hash(tx_hex)
```
yabirgb, line 128: I believe this is not needed. Same in other tests where you do this

yabirgb (commented):
A few comments

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
dimyG: yes same for v1 and v2

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
dimyG: No they are emitted from the router.

rotkehlchen/chain/optimism/modules/velodrome/decoder.py
```diff
@@ -36,11 +63,219 @@ def __init__(
...
+
+    def _decode_swap(self, context: DecoderContext) -> DecodingOutput:
+        """Decodes events that swap eth or tokens in a velodrome v1 or v2 pool"""
+        spend_event, receive_event = None, None
+        for event in context.decoded_events:
+            crypto_asset = event.asset.resolve_to_crypto_asset()
+            if (
+                event.event_type == HistoryEventType.SPEND and
+                event.event_subtype == HistoryEventSubType.NONE and
+                (event.address in (ROUTER_V2, ROUTER_V1) or event.address in self.pools)
```
LefterisJP, line 149: wouldn't it make sense to add something like `self.protocol_addresses = {ROUTER_V2, ROUTER_v1}.union(self.pools)` in the constructor? And then use this everywhere with a single `in` check here?

rotkehlchen/data/global.db
LefterisJP, line 1: In discord you had mentioned that the diff also shows the last migration changing? Have you found out why? And can you maybe undo this change if not needed (which seems to not be)

rotkehlchen/tests/unit/decoders/test_velodrome.py
```diff
@@ -0,0 +1,832 @@
...
+from rotkehlchen.types import ChainID, EvmTokenKind, Location, TimestampMS, deserialize_evm_tx_hash
+
+WETH_OP_POOL_ADDRESS = string_to_evm_address('0xd25711EdfBf747efCE181442Cc1D8F5F8fc8a0D3')
+WETH_OP_GAUGE_ADDRESS = string_to_evm_address('0xCC53CD0a8EC812D46F0E2c7CC5AADd869b6F0292')
+WETH_OP_LP_TOKEN = evm_address_to_identifier(
+    address=WETH_OP_POOL_ADDRESS,
+    chain_id=ChainID.OPTIMISM,
+    token_type=EvmTokenKind.ERC20,
+)
+VELO_V2_TOKEN = evm_address_to_identifier(
```
LefterisJP, line 25: The Velo token is not in the global DB? It should be. @yabirgb please teach how to add it and how to add it in the next global DB upgrade.

You can add it yourself but it's good to also teach Dimitries so he knows.

And I think we need a cleanup. This is what is currently in my global DB:



There is an asset called `VELO` which is https://www.coingecko.com/en/coins/velo, and it's not our velodrome VELO.  There is `eip155:56/erc20:0xf486ad071f3bEE968384D2E39e2D8aF0fCf6fd46|VELO|velo|||1637827741|` which is exactly the same as that velo I just mentioned. It's in binance chain.

And there is `eip155:10/erc20:0x3c8B650257cFb5f272f799F5e2b4e65093a11a05|VELO|velodrome-finance|||1654113683|` which is the v1 asset.

So we should at least:

- Remove `VELO` in favor of `eip155:56/erc20:0xf486ad071f3bEE968384D2E39e2D8aF0fCf6fd46|VELO|velo|||1637827741|`. This will required DB upgrade.
- Add The v2 Velo token and have them handled as the same thing by making an asset mapping.

Do the above make sense? Not in this PR but we need either an issue for this or some kind of task. As discussed in our workshop we can easily add tasks in a release: https://github.com/orgs/rotki/projects/11/views/1?layout=board

rotkehlchen/tests/unit/decoders/test_velodrome.py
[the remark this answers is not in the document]
LefterisJP: yes don't do it in this PR. We can adjust notes style in subsequent PR. Not sure exactly what style yabir means here but sure can be donelater. Keep this PR simple.

rotkehlchen/tests/unit/decoders/test_velodrome.py
```diff
@@ -0,0 +1,832 @@
...
+            timestamp=timestamp,
+            location=Location.OPTIMISM,
+            event_type=HistoryEventType.TRADE,
+            event_subtype=HistoryEventSubType.RECEIVE,
+            asset=Asset('eip155:10/erc20:0x9485aca5bbBE1667AD97c7fE7C4531a624C8b1ED'),
+            balance=Balance(FVal('0.130694520044655821')),
+            location_label=user_address,
+            counterparty=CPT_VELODROME,
+            address=string_to_evm_address('0x7866C6072B09539fC0FDE82963846b80203d7beb'),
+            notes=F'Receive 0.130694520044655821 agEUR as the result of a swap in {CPT_VELODROME}',
```
LefterisJP, line 446: didn't know capital `F` also works. But please keep it consisten to lowercase across the file. Search if you got any more and adjust them. Won't comment in the PR if I see more.

```suggestion
            notes=f'Receive 0.130694520044655821 agEUR as the result of a swap in {CPT_VELODROME}',
```

rotkehlchen/tests/unit/decoders/test_velodrome.py
```diff
@@ -0,0 +1,832 @@
...
+        ),
+    ]
+    assert events == expected_events
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('optimism_accounts', [['0x2359497cc3F8F11A80d775715367d5CB3D0fD274']])
+def test_swap_token_to_eth(optimism_accounts, optimism_transaction_decoder):
```
LefterisJP, line 454: Add in docstring which version this is for. Don't change name or else you will have to modify the filename of the cassettes and don't want to complicate this for you

rotkehlchen/tests/unit/decoders/test_velodrome.py
```diff
@@ -0,0 +1,832 @@
...
+            notes=F'Receive 0.181508840163183003 ETH as the result of a swap in {CPT_VELODROME}',
+        ),
+    ]
+    assert events == expected_events
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('optimism_accounts', [['0x78C13393Aee675DD7ED07ce992210750D1F5dB88']])
+def test_swap_tokens(optimism_accounts, optimism_transaction_decoder):
+    """Check that swapping tokens in velodrome is properly decoded."""
```
LefterisJP, line 580: same as above here and in other places. Specify version. Let's be future proof (maybe a v3 comes along etc.)

Check rest of the test functions and do that there too. I won't leave more comments in the PR for this.

rotkehlchen/tests/unit/decoders/test_velodrome.py
```diff
@@ -0,0 +1,832 @@
...
+            notes=f'Deposit 0.005 WETH in velodrome pool {WETH_OP_POOL_ADDRESS}',
+        ), EvmEvent(
+            tx_hash=evmhash,
+            sequence_index=69,
+            timestamp=timestamp,
+            location=Location.OPTIMISM,
+            event_type=HistoryEventType.INFORMATIONAL,
+            event_subtype=HistoryEventSubType.APPROVE,
+            asset=A_OP,
+            balance=Balance(FVal('0')),
```
LefterisJP, line 84: Here and in the other 6 places you use `FVal('0')` replace with `ZERO`

rotkehlchen/tests/unit/decoders/test_velodrome.py
```diff
@@ -0,0 +1,832 @@
...
+            sequence_index=109,
+            timestamp=timestamp,
+            location=Location.OPTIMISM,
+            event_type=HistoryEventType.TRADE,
+            event_subtype=HistoryEventSubType.RECEIVE,
+            asset=Asset(VELO_V2_TOKEN),
+            balance=Balance(FVal('50.633275729041219862')),
+            location_label=user_address,
+            counterparty=CPT_VELODROME,
+            address=string_to_evm_address('0x0af32614EEa68B8D2232B9592FbdB6512ab6DA73'),  # fbomb_velo_pool_address  # noqa: E501
```
LefterisJP, line 639: What's fbomb? Just the token name?

LefterisJP (commented):
So left some comments and questions, some of which require short input by @yabirgb.

Not much to do at least in this PR.

rotkehlchen/tests/unit/decoders/test_velodrome.py
dimyG: yes just a token name. I removed that comment since it is not something we should care I guess.

rotkehlchen/data/global.db
dimyG: ok, i redone the changes in the packaged db without running migrations. this is the new sql diff: https://gist.github.com/dimyG/fe9f41db042b4afde886cc776157c3f4

LefterisJP (approved):
LGTM. We have some followups:

-  the abstractions as another task
- https://github.com/orgs/rotki/projects/11?pane=issue&itemId=39416686

PR #6626, opened 2023-09-21
Better handle errors from yearn api

yabirgb:
Closes #(issue_number)

## Checklist


LefterisJP (commented):
Looks okay but ... yearn just rugged us. The old API just moved.

https://api.yexporter.io/v1/chains/1/vaults/all

Can you check if it's the same API? And run tests on it? If schema is same etc.

LefterisJP (approved):
Then LGTM

PR #6628, opened 2023-09-21
fix: display only error text when merging assets

### LINKED ISSUE #6625: Improve how error message appears on merge assets form
## Problem Definition

image
Use vuelidate to parse the external messages and show the validation error to the proper field

## Tasks

- [ ] display only the message at the proper field

kelsos:
Closes #6625 
## Checklist

### DIFF
--- a/frontend/app/src/components/asset-manager/MergeDialog.vue
+++ b/frontend/app/src/components/asset-manager/MergeDialog.vue
@@ -3,13 +3,17 @@ import useVuelidate from '@vuelidate/core';
 import { helpers, requiredUnless } from '@vuelidate/validators';
 import { toMessages } from '@/utils/validation';
 
+type Errors = Partial<
+  Record<'targetIdentifier' | 'sourceIdentifier', string[]>
+>;
+
 defineProps<{ value: boolean }>();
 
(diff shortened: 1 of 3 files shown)


frontend/app/src/components/asset-manager/MergeDialog.vue
```diff
@@ -5,46 +5,47 @@ import { toMessages } from '@/utils/validation';
...
     set(done, true);
   } else {
-    set(errorMessages, [
-      ...get(errorMessages),
-      result.message ?? t('merge_dialog.error').toString()
-    ]);
+    set(
+      errorMessages,
+      typeof result.message === 'string'
+        ? { sourceIdentifier: result.message ?? t('merge_dialog.error') }
```
tewshi, line 46:
```typescript
result.message ?? t('merge_dialog.error')
```
since at this point, `result.message` is a string, then `?? t('merge_dialog.error')` will never run

because it will only run when `result.message` is `null` or `undefined`

perhaps `|| t('merge_dialog.error')`

kelsos: true good catch, let me change that

Issue #630, opened 2020-01-17
Add ability to export history events (unprocessed)


coinyon:
Edit by Lef: Newest edit and what we consider the spec as of 21/09/2023

## Abstract

The user should be able to export their history of events out of rotki. They should be able to go to the history events view, press export and get an export of all the events in their history in the rotki events format.

They should also be able to filter down the selection by the already existing filters of that view. So if a filter is applied the export will only export all events matching that filter.

At the current state of rotki, this will miss some events. Namely trades, ledger actions and asset movements. But once:

- https://github.com/rotki/rotki/issues/6097
- https://github.com/rotki/rotki/issues/6096
- https://github.com/rotki/rotki/issues/6099
- https://github.com/rotki/rotki/issues/6623

are done, and everything has been moved to history events, this will actually export everything.

## Motivation

Some effort wents into rotki to setup all the info. Rotki should not "lock in" the user but provide an export functionality in an unencrypted way. Even though the PnL export exists, this is for processed events. Users should be able to also export their raw history events or any subset of them.

## Specification

- [ ] Create a new API endpoint for exporting history events. That endpoint should accept the exact same filters (without pagination part) that https://rotki.readthedocs.io/en/latest/api.html#dealing-with-history-events accepts. It should also accept either a filepath for electron app or nothing for docker. It should then export the events in our raw format in CSV file to the filepath for electron or prompt for a download of a file in docker.
- [ ] Premium limits should not apply here. We do not hold users data hostage for money. They can export all data no matter the premium level.
- [ ] Write tests for it
- [ ] Create the frontend part, including integrating the export in a nice and visible part in the UI of the history events view.

LefterisJP:
What format would be the best for such an export @coinyon?

coinyon:
You can choose between CSV, XML or JSON. It would probably be JSON nowadays. You could export `Trades.json`, `OTCTrades.json`, `Balances.json`, and so on, and define a JSON-Schema for each, but thats optional.

LefterisJP:
Trades + OTCTrades would be together. Since the format is the same. You can actually look at all these if you go into your DB manually and see it with sqlite queries.

JSON huh? I was more thinking of CSV for such a thing but let's see. Will do some research as to what would fit best.

For balances what do you mean? The current balances? There is no need for that since each time it changes depending on your configured assets.

Addresses per blockchain and owned tokens definitely.

coinyon:
Yep, you are right about the balances.

[...]

PR #6633, opened 2023-09-25
exporting base history events csv

jbrit:
Closes the backend part of #630 

## Checklist

- [x] Create a new API endpoint for exporting history events without premium restrictions
- [ ] Write tests for it

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -4677,23 +4677,8 @@ Dealing with History Events
 
    :reqjson int limit: This signifies the limit of records to return as per the `sql spec <https://www.sqlite.org/lang_select.html#limitoffset>`__.
    :reqjson int offset: This signifies the offset from which to start the return of records per the `sql spec <https://www.sqlite.org/lang_select.html#limitoffset>`__.
-   :reqjson list[string] order_by_attributes: This is the list of attributes of the transaction by which to order the results.
-   :reqjson list[bool] ascending: Should the order be ascending? This is the default. If set to false, it will be on descending order.
-   :reqjson bool group_by_event_ids: A boolean determining if results should be grouped by common event identifiers. If true, the result will return only first event of each group but also the number of events the group has. Default is false.
-   :reqjson int from_timestamp: The timestamp from which to start querying. Default is 0.
-   :reqjson int to_timestamp: The timestamp until which to query. Default is now.
-   :reqjson list[string] event_identifiers: An optional list of event identifiers to filter for.
-   :reqjson list[string] event_types: An optional list of event types by which to filter the decoded events.
-   :reqjson list[string] event_subtypes: An optional list of event subtypes by which to filter the decoded events.
-   :reqjson list location: An optional location name to filter events only for that location.
-   :reqjson list[string] location_labels: A list of location labels to optionally filter by. Location label is a string field that allows to provide more information about the location. When we use this structure in blockchains, it is used to specify the user address. For exchange events it's the exchange name assigned by the user.
-   :reqjson object entry_types: An object with two keys named 'values' and 'behaviour'. 'values' is a list of entry types to optionally filter by. 'behaviour' is optional and is a string with the value 'include' or 'exclude' which defines the filtering behaviour. It defaults to 'include'. Entry type is the event category and defines the schema. Possible values are: ``"history event"``, ``"evm event"``, ``"eth withdrawal event"``, ``"eth block event"``, ``"eth deposit event"``.
-   :reqjson string asset: The asset to optionally filter by.
-   :reqjson list[string] tx_hashes: An optional list of transaction hashes to filter for. This will make it an EVM event query.
-   :reqjson list[string] counterparties: An optional list of counterparties to filter by. List of strings. This will make it an EVM event query. We currently have a special exception for ``"eth2"`` as a counterparty. It filters for all eth staking events if given. It can't be given along with other counterparties in a filter. Or with an entry types filter.
-   :reqjson list[string] products: An optional list of product type to filter by. List of strings. This will make it an EVM event query.
-   :reqjson list[string] addresses: An optional list of EVM addresses to filter by in the set of counterparty addresses. This will make it an EVM event query.
-   :reqjson list[int] validator_indices: An optional list of validator indices to filter by. This makes it an EthStakingevent query
+   
+   :ref:`filter-request-args-label`
 
    **Example Response**:
 
@@ -4991,6 +4976,82 @@ Dealing with History Events
    :statuscode 409: No user is logged in or one of the identifiers to delete did not correspond to an event in the DB or one of the identifiers was for the last event in the corresponding transaction hash and force_delete was false..
    :statuscode 500: Internal rotki error
 
+Exporting History Events
+============================================
+
+.. http:post:: /api/(version)/history/events/export
+
+   Doing a POST on this endpoint with the given filter parameters will export a csv with all history events matching the filter to a file in the provided directory. Only the 'directory_path' argument is required. If no filter is used all the events will be exported.
+
+   **Example Request**:
+
+   .. http:example:: curl wget httpie python-requests
+
+      POST /api/1/history/events/export HTTP/1.1
+      Host: localhost:5042
+      Content-Type: application/json;charset=UTF-8
+
+      {
+          "directory_path": "/home",
+          "from_timestamp": 1500,
+          "to_timestamp": 999999
+      }
+
+   .. _history_export_schema_section:
+   
+   :reqjson string directory_path: The directory in which to write the exported CSV file
+   
+   :ref:`filter-request-args-label`
+
+   
+   **Example Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+          "result": true,
+          "message" "",
+      }
+
+   :statuscode 200: Events succesfully exported
+   :statuscode 400: Provided JSON is in some way malformed
+   :statuscode 409: No user is logged in or failure at event export.
+   :statuscode 500: Internal rotki error
+
+.. http:put:: /api/(version)/history/events/export
+
+   Doing a PUT on this endpoint with the given filter parameters will download a csv with all history events matching the filter. All arguments are optional. If no filter is used all the events will be downloaded.
+
+   **Example Request**:
+
+   .. http:example:: curl wget httpie python-requests
+
+      PUT /api/1/history/events/export HTTP/1.1
+      Host: localhost:5042
+      Content-Type: application/json;charset=UTF-8
+
+      {
+          "from_timestamp": 1500,
+          "to_timestamp": 999999
+      }
+
+   
+   :ref:`filter-request-args-label`
+
+   **Example Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: text/csv
+
+   :statuscode 200: Events succesfully downloaded
+   :statuscode 400: Provided JSON is in some way malformed
+   :statuscode 409: No user is logged in or failure at event download.
+   :statuscode 500: Internal rotki error
 
 Querying online events
 ============================================
@@ -12455,3 +12516,31 @@ Dealing with skipped external events
   :statuscode 200: Reprocessing went fine.
   :statuscode 409: An issue ocurred during reprocessing
   :statuscode 500: Internal rotki error
+
+Schema
+***********
+
+.. _filter-request-args-label:
+
+HistoryEvents Filter Request Arguments
+========================================
+
+These filter request arguments are common to multiple api endpoints that deal with history events filtering.
+
+- :reqjson list[string] order_by_attributes: This is the list of attributes of the transaction by which to order the results.
+- :reqjson list[bool] ascending: Should the order be ascending? This is the default. If set to false, it will be on descending order.
+- :reqjson bool group_by_event_ids: A boolean determining if results should be grouped by common event identifiers. If true, the result will return only the first event of each group but also the number of events the group has. Default is false.
+- :reqjson int from_timestamp: The timestamp from which to start querying. Default is 0.
+- :reqjson int to_timestamp: The timestamp until which to query. Default is now.
+- :reqjson list[string] event_identifiers: An optional list of event identifiers to filter for.
+- :reqjson list[string] event_types: An optional list of event types by which to filter the decoded events.
+- :reqjson list[string] event_subtypes: An optional list of event subtypes by which to filter the decoded events.
+- :reqjson list location: An optional location name to filter events only for that location.
+- :reqjson list[string] location_labels: A list of location labels to optionally filter by. Location label is a string field that allows you to provide more information about the location. When used in blockchains, it is used to specify the user's address. For exchange events, it's the exchange name assigned by the user.
+- :reqjson object entry_types: An object with two keys named 'values' and 'behavior'. 'values' is a list of entry types to optionally filter by. 'behavior' is optional and is a string with the value 'include' or 'exclude' which defines the filtering behavior. It defaults to 'include'. Entry type is the event category and defines the schema. Possible values are: "history event," "evm event," "eth withdrawal event," "eth block event," "eth deposit event."
+- :reqjson string asset: The asset to optionally filter by.
+- :reqjson list[string] tx_hashes: An optional list of transaction hashes to filter for. This will make it an EVM event query.
+- :reqjson list[string] counterparties: An optional list of counterparties to filter by. List of strings. This will make it an EVM event query. We currently have a special exception for "eth2" as a counterparty. It filters for all eth staking events if given. It can't be given along with other counterparties in a filter. Or with an entry types filter.
+- :reqjson list[string] products: An optional list of product type to filter by. List of strings. This will make it an EVM event query.
+- :reqjson list[string] addresses: An optional list of EVM addresses to filter by in the set of counterparty addresses. This will make it an EVM event query.
+- :reqjson list[int] validator_indices: An optional list of validator indices to filter by. This makes it an EthStakingevent query.
--- a/rotkehlchen/accounting/export/csv.py
+++ b/rotkehlchen/accounting/export/csv.py
@@ -1,9 +1,10 @@
 import json
 import logging
+from collections.abc import Collection
 from csv import DictWriter
 from pathlib import Path
 from tempfile import mkdtemp
-from typing import TYPE_CHECKING, Any, Literal
+from typing import TYPE_CHECKING, Any, Literal, Optional
 from zipfile import ZIP_DEFLATED, ZipFile
 
 from rotkehlchen.accounting.pnl import PnlTotals
@@ -30,6 +31,7 @@ logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
 FILENAME_ALL_CSV = 'all_events.csv'
+FILENAME_HISTORY_EVENTS_CSV = 'history_events.csv'
 ETHERSCAN_EXPLORER_TX_URL = 'https://{base_url}/tx/'
 
 ACCOUNTING_SETTINGS = (
@@ -48,7 +50,11 @@ class CSVWriteError(Exception):
     pass
 
 
-def _dict_to_csv_file(path: Path, dictionary_list: list) -> None:
+def dict_to_csv_file(
+        path: Path,
+        dictionary_list: list,
+        headers: Optional[Collection] = None,
+) -> None:
     """Takes a filepath and a list of dictionaries representing the rows and writes them
     into the file as a CSV
 
@@ -61,7 +67,7 @@ def _dict_to_csv_file(path: Path, dictionary_list: list) -> None:
         return
 
     with open(path, 'w', newline='', encoding='utf-8') as f:
-        w = DictWriter(f, fieldnames=dictionary_list[0].keys())
+        w = DictWriter(f, fieldnames=dictionary_list[0].keys() if headers is None else headers)
         w.writeheader()
         try:
             for dic in dictionary_list:
@@ -323,7 +329,7 @@ class CSVExporter(CustomizableDateMixin):
         self._maybe_add_summary(events=serialized_events, pnls=pnls)
         try:
             directory.mkdir(parents=True, exist_ok=True)
-            _dict_to_csv_file(
+            dict_to_csv_file(
                 directory / FILENAME_ALL_CSV,
                 serialized_events,
             )
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -30,7 +30,11 @@ from rotkehlchen.accounting.constants import (
     FREE_REPORTS_LOOKUP_LIMIT,
 )
 from rotkehlchen.accounting.debugimporter.json import DebugHistoryImporter
-from rotkehlchen.accounting.export.csv import CSVWriteError
+from rotkehlchen.accounting.export.csv import (
+    FILENAME_HISTORY_EVENTS_CSV,
+    CSVWriteError,
+    dict_to_csv_file,
+)
 from rotkehlchen.accounting.ledger_actions import LedgerAction
 from rotkehlchen.accounting.structures.balance import Balance, BalanceType
 from rotkehlchen.accounting.structures.base import HistoryBaseEntryType, StakingEvent
@@ -4413,3 +4417,72 @@ class RestAPI:
             result=_wrap_in_ok_result({'total': total, 'successfull': successfull}),
             status_code=HTTPStatus.OK,
         )
+
+    def export_history_events(
+            self,
+            filter_query: HistoryBaseEntryFilterQuery,
+            directory_path: Optional[Path],
+    ) -> Response:
+        """ Export or Download history events data to a CSV file."""
+        dbevents = DBHistoryEvents(self.rotkehlchen.data.db)
+        with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
+            history_events, _, _ = dbevents.get_history_events_and_limit_info(
+                cursor=cursor,
+                filter_query=filter_query,
+                has_premium=True,
+                entries_limit=None,
+            )
+
+        if len(history_events) == 0:
+            return api_response(
+                wrap_in_fail_result('No history processed in order to perform an export'),
+                status_code=HTTPStatus.CONFLICT,
+            )
+
+        serialized_history_events = []
+        headers: dict[str, None] = {}
+        for event in history_events:
+            serialized_event = event.serialize_for_csv()
+            serialized_history_events.append(serialized_event)
+            # maintain insertion order without storing extra info
+            headers.update({key: None for key in serialized_event})
+
+        if directory_path is None:  # on download
+            with tempfile.TemporaryDirectory() as temp_dir:
+                file_path = Path(temp_dir) / FILENAME_HISTORY_EVENTS_CSV
+
+                try:
+                    dict_to_csv_file(
+                        file_path,
+                        serialized_history_events,
+                        headers.keys(),
+                    )
+                    return send_file(
+                        path_or_file=file_path,
+                        mimetype='text/csv',
+                        as_attachment=True,
+                        download_name=FILENAME_HISTORY_EVENTS_CSV,
+                    )
+                except (CSVWriteError, PermissionError) as e:
+                    return api_response(
+                        wrap_in_fail_result(str(e)),
+                        status_code=HTTPStatus.CONFLICT,
+                    )
+                except FileNotFoundError:
+                    return api_response(
+                        wrap_in_fail_result('No file was found'),
+                        status_code=HTTPStatus.NOT_FOUND,
+                    )
+
+        try:  # from here and below we do a direct export to filesystem
+            directory_path.mkdir(parents=True, exist_ok=True)
+            file_path = directory_path / FILENAME_HISTORY_EVENTS_CSV
+            dict_to_csv_file(
+                file_path,
+                serialized_history_events,
+                headers.keys(),
+            )
+        except (CSVWriteError, PermissionError) as e:
+            return api_response(wrap_in_fail_result(str(e)), status_code=HTTPStatus.CONFLICT)
+
+        return api_response(OK_RESULT, status_code=HTTPStatus.OK)
--- a/rotkehlchen/api/v1/resources.py
+++ b/rotkehlchen/api/v1/resources.py
@@ -82,6 +82,7 @@ from rotkehlchen.api.v1.schemas import (
     ExchangesResourceAddSchema,
     ExchangesResourceEditSchema,
     ExchangesResourceRemoveSchema,
+    ExportHistoryEventSchema,
     ExternalServicesResourceAddSchema,
     ExternalServicesResourceDeleteSchema,
     FileListSchema,
@@ -2953,3 +2954,19 @@ class DefiMetadataResource(BaseMethodView):
 
     def get(self) -> Response:
         return self.rest_api.get_defi_metadata()
+
+
+class ExportHistoryEventResource(BaseMethodView):
+
+    post_schema = ExportHistoryEventSchema()
+    put_schema = ExportHistoryEventSchema(exclude=('directory_path',))
+
+    @require_loggedin_user()
+    @use_kwargs(post_schema, location='json_and_query')
+    def post(self, filter_query: 'HistoryBaseEntryFilterQuery', directory_path: Path) -> Response:
+        return self.rest_api.export_history_events(filter_query=filter_query, directory_path=directory_path)  # noqa: E501
+
+    @require_loggedin_user()
+    @use_kwargs(put_schema, location='json_and_query')
+    def patch(self, filter_query: 'HistoryBaseEntryFilterQuery') -> Response:
+        return self.rest_api.export_history_events(filter_query=filter_query, directory_path=None)
--- a/rotkehlchen/api/v1/schemas.py
+++ b/rotkehlchen/api/v1/schemas.py
@@ -669,14 +669,12 @@ class HistoryEventSchema(TimestampRangeSchema, DBPaginationSchema, DBOrderBySche
             should_query_eth_staking_event = True
             should_query_evm_event = False
 
-        common_arguments = {
+        common_arguments = self.make_extra_filtering_arguments(data) | {
             'order_by_rules': create_order_by_rules_list(
                 data=data,  # descending timestamp and ascending sequence index
                 default_order_by_fields=['timestamp', 'sequence_index'],
                 default_ascending=[False, True],
             ),
-            'limit': data['limit'],
-            'offset': data['offset'],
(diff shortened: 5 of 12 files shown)


rotkehlchen/accounting/export/csv.py
```diff
@@ -331,3 +340,20 @@ def export(
             return False, str(e)
 
         return True, ''
+
+    def export_history(
+            self,
+            events: list['HistoryBaseEntry'],
+            directory: Path,
+    ) -> tuple[bool, str]:
+        serialized_events = [self.to_csv_entry(x) for idx, x in enumerate(events)]
```
yabirgb, line 349: As mentioned in discord this should export everything not only the current PnL csv format for the history events https://discord.com/channels/657906918408585217/1080783409640648765/1155766596539777065

rotkehlchen/api/v1/schemas.py
```diff
@@ -3027,3 +3027,136 @@ def make_history_event_filter(
 
 class EthStakingHistoryStatsDetails(EthStakingHistoryStats, AsyncIgnoreCacheQueryArgumentSchema):
     """Schema for querying ethereum staking history details"""
+
+
+class ExportHistoryEventSchema(TimestampRangeSchema, DBOrderBySchema):
+    """Schema for quering history events"""
+    exclude_ignored_assets = fields.Boolean(load_default=True)
+    event_identifiers = DelimitedOrNormalList(fields.String(), load_default=None)
```
yabirgb, line 3035: Don't duplicate code. This filters already exists in other schema so just inherit ir or use it completly. Not sure if there is need to create a new one

yabirgb (commented):
I saw the PR and just to guide you in case you were in the wrong direction

rotkehlchen/api/v1/schemas.py
jbrit: The schema that has all these is the `HistoryEventSchema`, but I'm removing some fields and adding one, or maybe I can create one that both inherit from?

rotkehlchen/accounting/export/csv.py
jbrit: Yes, I export everything for history events

rotkehlchen/api/v1/schemas.py
```diff
@@ -3031,3 +3031,134 @@ class EthStakingHistoryStatsDetails(EthStakingHistoryStats, AsyncIgnoreCacheQuer
...
+    def make_history_event_filter(
+            self,
+            data: dict[str, Any],
+            **_kwargs: Any,
+    ) -> dict[str, Any]:
+        should_query_eth_staking_event = data['validator_indices'] is not None
+        should_query_evm_event = any(data[x] is not None for x in ('products', 'counterparties', 'tx_hashes', 'addresses'))  # noqa: E501
+        counterparties = data['counterparties']
+        entry_types = data['entry_types']
+        if counterparties is not None and CPT_ETH2 in counterparties:
```
yabirgb, line 3096: this function is also a copy paste with the exception of the limit/offset. If you wanted to remove this fields only you could have made an abstraction that took the common arguments as argument. 

Have you coordinated with the frontend how are they going to call this endpoint. If they don't pass the limit/offset is enough with using the same schema. What I would do is use the same methods in both but have 


```python
class HistoryEventSchema(TimestampRangeSchema, DBOrderBySchema):
    # the one you are adding in this PR

class PaginatedHistoryEventSchema(DBPaginationSchema,  HistoryEventSchema):
    # the existing one
```

and in this one instead of doing

```python
            'limit': data['limit'],
            'offset': data['offset'],
```

do 

```python
            'limit': data.get('limit'),
            'offset': data.get('offset'),
```

`.get` in python returns `None` when the value doesn't exist. If you can find a better way of doing this go with it but don't duplicate ever code. If you need to duplicate code something is bad designed

rotkehlchen/api/v1/resources.py
```diff
@@ -2953,3 +2954,13 @@ class DefiMetadataResource(BaseMethodView):
...
+
+
+class ExportHistoryEventResource(BaseMethodView):
+
+    post_schema = ExportHistoryEventSchema()
+
+    @require_loggedin_user()
+    @use_kwargs(post_schema, location='json_and_query')
+    def post(self, filter_query: 'HistoryBaseEntryFilterQuery', directory_path: Path) -> Response:
+        return self.rest_api.export_history_events(filter_query=filter_query, directory_path=directory_path)  # noqa: E501
```
yabirgb, line 2966: about the path for exporting the file I'm not sure what was the discussion... can the frontend provide it for docker and electron? in some endpoints we need to methods, one to download from the browser and one to save in the fs. Have you coordinated on this?

rotkehlchen/accounting/export/csv.py
```diff
@@ -289,10 +292,28 @@ def create_zip(
...
-    def to_csv_entry(self, event: 'ProcessedAccountingEvent') -> dict[str, Any]:
+    def to_csv_entry(self, event: Union['ProcessedAccountingEvent', 'HistoryBaseEntry']) -> dict[str, Any]:  # noqa: E501
         """Prepare the provided event to have a common format for the accounting
         CSV exported file.
         """
+        if isinstance(event, HistoryBaseEntry):
+            dict_event = {
+                'identifier': event.identifier,
+                'entry_type': event.entry_type.serialize(),
+                'event_identifier': event.event_identifier,
```
yabirgb, line 303: This looks really wrong. You have one method that is for exporting the accounting and your are adding the check for the type here and then breaking the logic. You can move this to the `HistoryBaseEntry` class and add this as method. Also you are exporting the base fields but evm events and staking events have other fields that you also have to consider

rotkehlchen/accounting/export/csv.py
```diff
@@ -331,3 +352,23 @@ def export(
             return False, str(e)
 
         return True, ''
+
+    def export_history(
+            self,
+            events: list['HistoryBaseEntry'],
+            directory: Path,
+    ) -> tuple[bool, str]:
+        if len(events) == 0:
```
yabirgb, line 361: I would move this logic to the `rest.py` file. You don't need the class information for anything and the `to_csv`logic should by an attribute of the event class

rotkehlchen/accounting/export/csv.py
```diff
@@ -331,3 +352,23 @@ def export(
...
+            return False, 'No history processed in order to perform an export'
+
+        serialized_events = [self.to_csv_entry(x) for idx, x in enumerate(events)]
```
yabirgb, line 364: why do you need the `enumerate` here?

rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -0,0 +1,147 @@
...
+    else:
+        serialized.pop('identifier')  # there is `identifier`: `None` which we have to remove
+    serialized.pop('event_identifier')
+    serialized.pop('entry_type')
+    return serialized
+
+
+def add_entries(server: 'APIServer', events_db: DBHistoryEvents, add_directly: bool = False) -> list['HistoryBaseEntry']:  # noqa: E501
+    entries = [EvmEvent(
+        tx_hash=deserialize_evm_tx_hash('0x64f1982504ab714037467fdd45d3ecf5a6356361403fc97dd325101d8c038c4e'),  # noqa: E501
```
yabirgb, line 40: add a docstring explaining what this function is doing

rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -0,0 +1,147 @@
...
+
+if TYPE_CHECKING:
+    from rotkehlchen.api.server import APIServer
+
+
+def entry_to_input_dict(
+        entry: 'EvmEvent',
+        include_identifier: bool,
+) -> dict[str, Any]:
+    serialized = entry.serialize()
```
yabirgb, line 27: add a doctring explaining what it does

rotkehlchen/tests/api/test_history_export.py
```diff
@@ -0,0 +1,129 @@
...
+        rotkehlchen_api_server_with_exchanges,
+        tmpdir_factory,
+):
+    """Test that the csv export/download REST API endpoint works correctly."""
+    rotki = rotkehlchen_api_server_with_exchanges.rest_api.rotkehlchen
+    db = DBHistoryEvents(rotki.data.db)
+    add_entries(server=rotkehlchen_api_server_with_exchanges, events_db=db, add_directly=True)
+
+    csv_dir = str(tmpdir_factory.mktemp('test_csv_dir'))
+    csv_dir2 = str(tmpdir_factory.mktemp('test_csv_dir2'))
```
yabirgb, line 57: if you use the same folder won't the file by overwritten?

rotkehlchen/tests/api/test_history_export.py
```diff
@@ -0,0 +1,129 @@
...
+from rotkehlchen.accounting.export.csv import FILENAME_HISTORY_EVENTS_CSV
+from rotkehlchen.db.history_events import DBHistoryEvents
+from rotkehlchen.tests.utils.api import api_url_for, assert_error_response, assert_proper_response
+from rotkehlchen.tests.utils.history import prepare_rotki_for_history_processing_test
+from rotkehlchen.tests.utils.history_base_entry import add_entries
+
+
+def assert_csv_export_response(response, csv_dir, expected_count=6):
+    assert_proper_response(response)
+    data = response.json()
```
yabirgb, line 18: add a docstring explaining what it does

rotkehlchen/accounting/export/csv.py
yabirgb, line 303: We already have the serialize method and this seems to be doing the same. Can't you just call it and then for the balances flattern it manually? meaning pop the value in serialize and add to the mapping the two values required

yabirgb (commented):
Many comments about the approach. Most important don't copy paste code. If you do it probably something is wrong about the design

docs/api.rst
```diff
@@ -4991,6 +4991,64 @@ Dealing with History Events
    :statuscode 409: No user is logged in or one of the identifiers to delete did not correspond to an event in the DB or one of the identifiers was for the last event in the corresponding transaction hash and force_delete was false..
    :statuscode 500: Internal rotki error
 
+Exporting History Events
+============================================
+
+.. http:post:: /api/(version)/history/events/export
+
+   Doing a POST on this endpoint with the given filter parameters will return a csv with all history events matching the filter. All arguments are optional. If nothing is given all events will be returned.
+
```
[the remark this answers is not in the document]
jbrit: done

rotkehlchen/api/v1/schemas.py
```diff
@@ -3027,3 +3027,134 @@ def make_history_event_filter(
...
+    @validates_schema
+    def validate_history_event_schema(
+            self,
+            data: dict[str, Any],
+            **_kwargs: Any,
+    ) -> None:
+        valid_ordering_attr = {None, 'timestamp'}
+        if (
+            data['order_by_attributes'] is not None and
+            not set(data['order_by_attributes']).issubset(valid_ordering_attr)
```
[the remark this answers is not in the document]
jbrit: done

rotkehlchen/api/v1/schemas.py
jbrit: I did a work around to extend the previous class without breaking any logic

rotkehlchen/accounting/export/csv.py
jbrit: done

rotkehlchen/accounting/export/csv.py
jbrit: done

rotkehlchen/accounting/export/csv.py
jbrit: didn't need it

rotkehlchen/tests/utils/history_base_entry.py
jbrit: done

rotkehlchen/tests/utils/history_base_entry.py
jbrit: done

rotkehlchen/tests/api/test_history_export.py
jbrit: they're different folders

rotkehlchen/tests/api/test_history_export.py
jbrit: done

rotkehlchen/api/v1/resources.py
jbrit: seen, I can export, I want to add a query param that will be used for this, something like ?download=true, which would trigger a download instead of an export alone

rotkehlchen/accounting/export/csv.py
```diff
@@ -48,7 +50,9 @@ class CSVWriteError(Exception):
     pass
 
 
-def _dict_to_csv_file(path: Path, dictionary_list: list) -> None:
+def dict_to_csv_file(path: Path,
+                     dictionary_list: list,
+                     headers: Optional[Collection] = None) -> None:
```
yabirgb, line 55: This is the style that we follow in the code. New line for arguments and double indentation

```suggestion
def dict_to_csv_file(
         path: Path,
         dictionary_list: list,
         headers: Optional[Collection] = None,
) -> None:
```

rotkehlchen/api/v1/resources.py
yabirgb: but have you asked the frontend about this? If they can handle it or if they have any special requirement. Can you contact them in discord to coordinate in this topic?

rotkehlchen/api/v1/schemas.py
```diff
@@ -705,10 +704,26 @@ def make_history_event_filter(
...
+        """
+        return {
+            'limit': data['limit'],
+            'offset': data['offset'],
+        }
+
+    def get_extra_fields(self, data: dict[str, Any]) -> dict[str, Any]:
+        """
+        generates the extra fields to be included in the base_filter dictionary
+        """
```
yabirgb, line 725: this naming is confusing. I was thinking that those were things that you add to the filter but instead are extra arguments that get returned after the processing.

```suggestion
    def validation_proxied_information(self, data: dict[str, Any]) -> dict[str, Any]:
        """
        extra fields that will be returned after validation
        """
```

yabirgb: if you find a better name is good. What you had really confused me

rotkehlchen/api/v1/schemas.py
```diff
@@ -705,10 +704,26 @@ def make_history_event_filter(
         else:
             filter_query = HistoryEventFilterQuery.make(**common_arguments)
 
-        return {
+        base_filter = {
             'filter_query': filter_query,
-            'group_by_event_ids': data['group_by_event_ids'],
         }
+        base_filter.update(self.get_extra_fields(data))
+        return base_filter
```
yabirgb, line 711: maybe is cleaner to do

```python
     return self.get_extra_fields(data) | {
            'filter_query': filter_query,
            'group_by_event_ids': data['group_by_event_ids'],
        }
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -705,10 +704,26 @@ def make_history_event_filter(
...
+
+    def make_extra_common_arguments(self, data: dict[str, Any]) -> dict[str, Any]:
```
yabirgb, line 713:
```suggestion
    def make_extra_filtering_arguments(self, data: dict[str, Any]) -> dict[str, Any]:
```

a name that suggests where this will be used is better. `common` doesn't really means anything here or add value to get a quick idea for where this will be used

rotkehlchen/tests/api/test_history_export.py
```diff
@@ -0,0 +1,139 @@
...
+from rotkehlchen.tests.utils.history_base_entry import add_entries
+
+
+def assert_csv_export_response(response, csv_dir, expected_count=6):
+    """
+    Asserts that a CSV export response meets certain criteria.
+    Args:
+        response (Response): The response object returned from the CSV export request.
+        csv_dir (str): The directory where the CSV files are expected to be located.
+        expected_count (int): The expected number of rows in the CSV files.
```
yabirgb, line 22: you can avoid the types if you add them to the function definition instead. 

```python
def assert_csv_export_response(response: Response, csv_dir: Path, expected_count:int = 6):
```

rotkehlchen/tests/api/test_history_export.py
```diff
@@ -0,0 +1,139 @@
...
+    with open(os.path.join(csv_dir, FILENAME_HISTORY_EVENTS_CSV), newline='', encoding='utf-8') as csvfile:  # noqa: E501
+        reader = csv.DictReader(csvfile)
+        count = 0
+        for row in reader:
+            # base history entry field count
+            assert len(row) >= 13
+            assert row['location'] in (
+                'kraken',
+                'ethereum',
+            )
```
yabirgb, line 41:
```suggestion
            assert row['location'] in ('kraken', 'ethereum')
```

rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -0,0 +1,155 @@
...
+    serialized.pop('event_identifier')
+    serialized.pop('entry_type')
+    return serialized
+
+
+def add_entries(server: 'APIServer', events_db: DBHistoryEvents, add_directly: bool = False) -> list['HistoryBaseEntry']:  # noqa: E501
+    """
+    Adds a list of history entries to the database, either directly or through the
+    API, depending on the `add_directly` flag.
+
```
yabirgb, line 45:
```suggestion
```

docs/api.rst
```diff
@@ -4991,6 +4991,64 @@ Dealing with History Events
    :statuscode 409: No user is logged in or one of the identifiers to delete did not correspond to an event in the DB or one of the identifiers was for the last event in the corresponding transaction hash and force_delete was false..
    :statuscode 500: Internal rotki error
 
+Exporting History Events
+============================================
+
+.. http:post:: /api/(version)/history/events/export
+
+   Doing a POST on this endpoint with the given filter parameters will return a csv with all history events matching the filter. All arguments are optional. If no filter is used all the events will be exported.
+
```
yabirgb, line 5000: now all the arguments are not optional since you have the directory path

yabirgb (commented):
Quite a few things to improve

rotkehlchen/api/rest.py
```diff
@@ -4413,3 +4417,44 @@ def reprocess_skipped_external_events(self) -> Response:
...
+                filter_query=filter_query,
+                has_premium=True,
+                entries_limit=None,
+            )
+
+        if len(history_events) == 0:
+            return api_response(
+                wrap_in_fail_result('No history processed in order to perform an export'),
+                status_code=HTTPStatus.CONFLICT,
+            )
```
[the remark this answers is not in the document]
jbrit: returning an empty list seems weird, the error occurs from a user exporting if he has no events (or no events matching the criteria) since the endpoint exports a {result: boolean} normally, it just  returns {result: None}

rotkehlchen/api/rest.py
```diff
@@ -4413,3 +4417,44 @@ def reprocess_skipped_external_events(self) -> Response:
...
+
+        serialized_history_events = []
+        headers = {}
+        for event in history_events:
+            serialized_event = event.serialize()
+            balance = serialized_event.pop('balance')
+            serialized_event['balance_amount'] = balance['amount']
+            serialized_event['balance_usd_value'] = balance['usd_value']
```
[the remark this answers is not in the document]
jbrit: I would have to implement this on all the classes then @yabirgb

rotkehlchen/api/rest.py
```diff
@@ -4413,3 +4417,44 @@ def reprocess_skipped_external_events(self) -> Response:
...
+            serialized_history_events.append(serialized_event)
+            headers.update(serialized_event)
```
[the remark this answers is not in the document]
jbrit: I tried with the set first, but it was unordered, I wanted it to maintain a fairly constant arrangement in the csv

rotkehlchen/api/rest.py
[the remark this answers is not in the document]
yabirgb: they inherit from the basehistoryevent class. you should only implement it in one

rotkehlchen/api/rest.py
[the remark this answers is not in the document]
yabirgb: ah with an empty list I meant creating a file with no events but maybe is better to leave it as an error

rotkehlchen/api/rest.py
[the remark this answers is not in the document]
yabirgb: For that you can use the dict and `headers.update({key: None for x in serialized_event.keys()})` since in python 3.7 and forward (not sure if 3.6) dicts keep the inserion order. But add a comment that is for this reason. Otherwise you are storing extra information that you don't need and is not a reference but the actual values I fear

rotkehlchen/api/v1/schemas.py
jbrit: yes, it's cleaner

re: group_by_event_ids
I don't want it in the second schema, so it's still in the extra_fields

rotkehlchen/api/rest.py
```diff
@@ -4413,3 +4417,63 @@ def reprocess_skipped_external_events(self) -> Response:
...
+            serialized_event = event.serialize_for_csv()
+            serialized_history_events.append(serialized_event)
+            # maintain insertion order without storing extra info
+            headers.update({key: None for key in serialized_event})
+
+        try:
+            with tempfile.TemporaryDirectory() as temp_dir:
+                # on download
+                if directory_path is None:
+                    file_path = Path(temp_dir) / FILENAME_HISTORY_EVENTS_CSV
```
yabirgb, line 4453: mm this works I believe because Path calls str on the object or something like this but can you use

```suggestion
                    file_path = Path(temp_dir.name) / FILENAME_HISTORY_EVENTS_CSV
```

for me at least is clearer

rotkehlchen/api/rest.py
```diff
@@ -4413,3 +4417,63 @@ def reprocess_skipped_external_events(self) -> Response:
...
+        headers: dict[str, None] = {}
+        for event in history_events:
+            serialized_event = event.serialize_for_csv()
+            serialized_history_events.append(serialized_event)
+            # maintain insertion order without storing extra info
+            headers.update({key: None for key in serialized_event})
+
+        try:
+            with tempfile.TemporaryDirectory() as temp_dir:
+                # on download
```
yabirgb, line 4451: this is a bit ugly because you are creating a temp file even if you don't need it. Why not move 

```
                dict_to_csv_file(
                    file_path,
                    serialized_history_events,
                    headers.keys(),
                )

                if directory_path is None:
                    return send_file(
                        path_or_file=file_path,
                        mimetype='text/csv',
                        as_attachment=True,
                        download_name=FILENAME_HISTORY_EVENTS_CSV,
                    )
```

to an auxiliar function. and then call it in the both possible places but avoiding the creation of the temp folder always. I would define a function inside this method but I believe lefteris doesn't like them much. Using another method could be okey

rotkehlchen/api/v1/schemas.py
```diff
@@ -3031,3 +3044,18 @@ class EthStakingHistoryStatsDetails(EthStakingHistoryStats, AsyncIgnoreCacheQuer
...
+    directory_path = DirectoryField(required=True)
+
+    def make_extra_filtering_arguments(self, data: dict[str, Any]) -> dict[str, Any]:
+        return {}
+
+    def validation_proxied_information(self, data: dict[str, Any]) -> dict[str, Any]:
+        extra_fields = {}
+        directory_path = data.get('directory_path', None)
+        if directory_path is not None:
+            extra_fields['directory_path'] = directory_path
```
yabirgb, line 3060: `.get` returns `None` by default when not found

```suggestion
        if (directory_path := data.get('directory_path')) is not None:
            extra_fields['directory_path'] = directory_path
```

rotkehlchen/tests/api/test_history_export.py
```diff
@@ -0,0 +1,151 @@
...
+from rotkehlchen.tests.utils.api import api_url_for, assert_error_response, assert_proper_response
+from rotkehlchen.tests.utils.history import prepare_rotki_for_history_processing_test
+from rotkehlchen.tests.utils.history_base_entry import add_entries
+
+
+def assert_csv_export_response(
+        response: requests.Response,
+        csv_dir: Path,
+        expected_count: int = 6,
+        is_download: bool = False):
```
yabirgb, line 19:
```suggestion
        expected_count: int = 6,
        is_download: bool = False,
):
```

docs/api.rst
```diff
@@ -4991,6 +4991,100 @@ Dealing with History Events
    :statuscode 409: No user is logged in or one of the identifiers to delete did not correspond to an event in the DB or one of the identifiers was for the last event in the corresponding transaction hash and force_delete was false..
    :statuscode 500: Internal rotki error
 
+Exporting History Events
+============================================
+
+.. http:post:: /api/(version)/history/events/export
```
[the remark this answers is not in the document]
jbrit: combining both, from what Lefteris did here, they have the same request format, it's different for me

rotkehlchen/api/v1/schemas.py
jbrit: thanks just noticed the operator too ':='

rotkehlchen/api/rest.py
jbrit: so I only create the temp_dir when necessary, but abstracting with the dict_to_csv_file was looking too tacky because of the error handling it'll still require

rotkehlchen/api/rest.py
```diff
@@ -4413,3 +4417,63 @@ def reprocess_skipped_external_events(self) -> Response:
...
+                        as_attachment=True,
+                        download_name=FILENAME_HISTORY_EVENTS_CSV,
+                    )
+        except (CSVWriteError, PermissionError) as e:
+            return api_response(wrap_in_fail_result(str(e)), status_code=HTTPStatus.CONFLICT)
+        except FileNotFoundError:
+            return api_response(
+                wrap_in_fail_result('No file was found'),
+                status_code=HTTPStatus.NOT_FOUND,
+            )
```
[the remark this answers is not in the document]
jbrit: dict_to_csv_file -> can raise CSVWriteError, PermissionError
send_file -> can raise FileNotFoundError

I didn't want to create two different try/except statements

rotkehlchen/api/rest.py
```diff
@@ -4413,3 +4417,72 @@ def reprocess_skipped_external_events(self) -> Response:
...
+                        wrap_in_fail_result(str(e)),
+                        status_code=HTTPStatus.CONFLICT,
+                    )
+                except FileNotFoundError:
+                    return api_response(
+                        wrap_in_fail_result('No file was found'),
+                        status_code=HTTPStatus.NOT_FOUND,
+                    )
+
+        try:
```
LefterisJP, line 4477: Specify this is the different case

```suggestion
        try:  # from here and below we do a direct export to filesytem
```

rotkehlchen/api/v1/resources.py
```diff
@@ -2953,3 +2954,19 @@ class DefiMetadataResource(BaseMethodView):
 
     def get(self) -> Response:
         return self.rest_api.get_defi_metadata()
+
+
+class ExportHistoryEventResource(BaseMethodView):
+
+    post_schema = ExportHistoryEventSchema()
+    patch_schema = ExportHistoryEventSchema(exclude=('directory_path',))
```
LefterisJP, line 2962: Just use post/post. The reason I used patch in my case was I had no other verb left. Patch should mostly be used for editing.

```suggestion
    put_schema = ExportHistoryEventSchema(exclude=('directory_path',))
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -705,11 +704,25 @@ def make_history_event_filter(
...
-        return {
+        return self.validation_proxied_information(data) | {
             'filter_query': filter_query,
-            'group_by_event_ids': data['group_by_event_ids'],
         }
 
+    def make_extra_filtering_arguments(self, data: dict[str, Any]) -> dict[str, Any]:
+        """
+        generates the extra fields to be included in the filter_query dictionary
+        """
```
LefterisJP, line 714: fits in a line

```suggestion
        """Generates the extra fields to be included in the filter_query dictionary"""
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -705,11 +704,25 @@ def make_history_event_filter(
...
+        return {
+            'limit': data['limit'],
+            'offset': data['offset'],
+        }
+
+    def validation_proxied_information(self, data: dict[str, Any]) -> dict[str, Any]:
+        """
+        extra fields that will be returned after validation
+        """
```
LefterisJP, line 723: Same here
```suggestion
        """Generates extra fields that will be returned after validation"""
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -705,11 +704,25 @@ def make_history_event_filter(
...
+    def make_extra_filtering_arguments(self, data: dict[str, Any]) -> dict[str, Any]:
+        """
+        generates the extra fields to be included in the filter_query dictionary
+        """
+        return {
+            'limit': data['limit'],
+            'offset': data['offset'],
+        }
+
+    def validation_proxied_information(self, data: dict[str, Any]) -> dict[str, Any]:
```
LefterisJP, line 720: This name makes more sense. Could not understand the previous one

```suggestion
    def genrate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any]:
```

rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -0,0 +1,154 @@
...
+        serialized.pop('identifier')  # there is `identifier`: `None` which we have to remove
+    serialized.pop('event_identifier')
+    serialized.pop('entry_type')
+    return serialized
+
+
+def add_entries(server: 'APIServer', events_db: DBHistoryEvents, add_directly: bool = False) -> list['HistoryBaseEntry']:  # noqa: E501
+    """
+    Adds a list of history entries to the database, either directly or through the
+    API, depending on the `add_directly` flag.
```
LefterisJP, line 44:
```suggestion
    Adds a pre-set list of history entries to the database, either directly or
    through the API, depending on the `add_directly` flag.
```

rotkehlchen/tests/api/test_history_export.py
```diff
@@ -0,0 +1,152 @@
+import csv
```
LefterisJP, line 1: please rename the file to: `test_history_events_export.py`


Test history export can also be confused to be the PnL history as that name is used a lot there

rotkehlchen/tests/api/test_history_export.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # check the csv files were generated succesfully
+    with open(os.path.join(csv_dir, FILENAME_HISTORY_EVENTS_CSV), newline='', encoding='utf-8') as csvfile:  # noqa: E501
+        reader = csv.DictReader(csvfile)
+        count = 0
+        for row in reader:
+            # base history entry field count
+            assert len(row) >= 13
+            assert row['location'] in ('kraken', 'ethereum')
+            assert row['identifier'] is not None
```
LefterisJP, line 46: Can't you also just for a for loop in the attributes of the `HistoryEvents` class here?

rotkehlchen/tests/api/test_history_export.py
```diff
@@ -0,0 +1,152 @@
...
+    temporary_file.touch()
+    response = requests.post(
+        api_url_for(rotkehlchen_api_server_with_exchanges, 'exporthistoryeventresource'),
+        json={'directory_path': str(temporary_file)},
+    )
+    assert_error_response(
+        response=response,
+        contained_in_msg='is not a directory',
+        status_code=HTTPStatus.BAD_REQUEST,
+    )
```
LefterisJP, line 152: I need you to add a new test in this file, or augment the first one.

That test should export all possible types of history events to see that exports are done fine for all types of history events. That means, the basic one, evm events, eth deposits, eth withdrawals, eth block production etc.

It's mainly to test that the headers are done correctly and that all the attributes, of all events are exported succesfully. Also keep in mind we will add a lot more history event types so this should work flawlessly

LefterisJP (commented):
Thanks Ajibola. Left lots of comments/questions. Please address them and ping me again.

docs/api.rst
```diff
@@ -4991,6 +4975,106 @@ Dealing with History Events
    :statuscode 409: No user is logged in or one of the identifiers to delete did not correspond to an event in the DB or one of the identifiers was for the last event in the corresponding transaction hash and force_delete was false..
    :statuscode 500: Internal rotki error
 
+.. _filter-request-args-label:
```
LefterisJP, line 4978: This shouldn't be in the middle of all endpoints. It should be at the very end. Also did you try to do `make html` and look at the generated docs to make sure the link works fine and renders fine?

docs/api.rst
```diff
@@ -4991,6 +4975,106 @@ Dealing with History Events
...
+
+Filter Request Arguments
+------------------------
+
+These request arguments are common to multiple parts of the documentation:
```
LefterisJP, line 4983:
```suggestion
These filter request arguments are common to multiple api endpoints that deal with history events filtering.
```

docs/api.rst
```diff
@@ -4991,6 +4975,106 @@ Dealing with History Events
    :statuscode 409: No user is logged in or one of the identifiers to delete did not correspond to an event in the DB or one of the identifiers was for the last event in the corresponding transaction hash and force_delete was false..
    :statuscode 500: Internal rotki error
 
+.. _filter-request-args-label:
+
+Filter Request Arguments
```
LefterisJP, line 4980:
```suggestion
HistoryEvents Filter Request Arguments
```

docs/api.rst
```diff
@@ -4991,6 +4975,106 @@ Dealing with History Events
...
+      Host: localhost:5042
+      Content-Type: application/json;charset=UTF-8
+
+      {
+          "directory_path": "/home",
+          "from_timestamp": 1500,
+          "to_timestamp": 999999
+      }
+
+   .. _history_export_schema_section:
```
LefterisJP, line 5024: you should also use the link in the other endpoints where you saw them. Not just in the new endpoints.

LefterisJP (commented):
Left some more minor comments.

docs/api.rst
jbrit: so the set of filters we use here were only used in these two

docs/api.rst
```diff
@@ -12455,3 +12516,31 @@ Dealing with skipped external events
   :statuscode 200: Reprocessing went fine.
   :statuscode 409: An issue ocurred during reprocessing
   :statuscode 500: Internal rotki error
+
+Schema
```
LefterisJP, line 12520: nitpick title change for later

```suggestion
Schema Appendix
```

LefterisJP (approved):
LGTM. Failing tests should be handled separately and are the cryptocompare tests failing for a day now

PR #6648, opened 2023-09-27
refactor: updates latest prices layout

kelsos:
Closes #(issue_number)

## Checklist


frontend/app/src/components/inputs/AssetSelect.vue
```diff
@@ -267,6 +267,9 @@ const listeners = useListeners();
         />
       </div>
     </template>
+    <template #prepend>
+      <slot name="prepend" />
+    </template>
```
lukicenturi, line 272: image

`#prepend-inner` looks better I think (?)

kelsos: It does but it also pushes the label in the asset select itself when there is a selection so using it is a bit more complicated

lukicenturi (approved):
LGTM!

frontend/app/src/components/inputs/AssetSelect.vue
kelsos: We can do it in a new PR at some point, but I didn't want to mess with the AssetSelect css 😢

PR #6651, opened 2023-09-28
Easier opensea api key handling

LefterisJP:
Now that opensea api key is possible to be made by end users we can stop querying our server for it, and let users use their own, with a local backup of ours that is a last resort choice.


kelsos (approved):
lgtm

PR #6653, opened 2023-09-28
Adds velodrome lp prices and abstractions

dimyG:
* adds special price calculation for velodrome lp tokens
* abstracts common price calcualtion code

Closes #(issue_number)
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=39473786

## Checklist


dimyG:
This is the sqldiff for the packaged db. It adds the abi for velo v2 pools and adds the lp tokens as assets with a protocol attribute. 
https://gist.github.com/dimyG/856451934658411733c913af106bf83a

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,142 @@
...
+token_protocol_to_abi_name: dict[str, 'LP_TOKEN_CONTRACT_ABI_NAMES'] = {
+    UNISWAP_PROTOCOL: 'UNISWAP_V2_LP',
+    VELODROME_POOL_PROTOCOL: 'VELO_V2_LP',
+}
+
+
+def find_lp_price(
+        evm_inquirer: 'EvmNodeInquirer',
+        token: EvmToken,
+        token_price_func: Callable,
```
yabirgb, line 34: it should be stated in the docstring and possibly in the name that this is for uniswap like pools. This should work too for sushiswap and any other fork but not curve for example or balancer

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,142 @@
...
+if TYPE_CHECKING:
+    from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
+    from rotkehlchen.types import LP_TOKEN_CONTRACT_ABI_NAMES
+
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+token_protocol_to_abi_name: dict[str, 'LP_TOKEN_CONTRACT_ABI_NAMES'] = {
```
yabirgb, line 25: this should be uppercase I believe. Also is it possible to make it `Final`? I'm not sure about that

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    address = token.evm_address
+    if abi_name is None:
+        abi_name = token_protocol_to_abi_name[token.protocol]
+    contract = EvmContract(
+        address=address,
+        abi=evm_inquirer.contracts.abi(abi_name),
+        deployed_block=0,
+    )
+    methods = ['token0', 'token1', 'totalSupply', 'getReserves', 'decimals']
```
yabirgb, line 57: can you maybe make this a tuple?

```suggestion
    methods = ('token0', 'token1', 'totalSupply', 'getReserves', 'decimals')
```

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,142 @@
...
+                f'query failed. Should implement direct queries',
+            )
+            return None
+
+        if block_identifier <= 12336033:
+            multicall_method = evm_inquirer.multicall
+
+    try:
+        output = multicall_method(
+            require_success=True,
```
yabirgb, line 72: If you require sucess is the same to use multicall than multicall2. You can remove the logic around  chossing which one to use. What you should check is to not call multicall before it was deployed

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,142 @@
...
+            calls=[(address, contract.encode(method_name=method)) for method in methods],
+            block_identifier=block_identifier,
+        )
+    except RemoteError as e:
```
yabirgb, line 76: we usually don't catch it but this call can also raise `BlockchainQueryError` if I'm not wrong

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,142 @@
...
+            else:
+                decoded.append(decoded_method)
+        else:
+            log.debug(
+                f'Multicall to {abi_name} failed to fetch field {method_name} '
+                f'for token {token.evm_address}',
+            )
+            return None
+
+    chain_id = token.chain_id
```
yabirgb, line 106: truth is that there is no reason to assign this to a variable. You can access directly `token.chain_id` in the two places where it is used since there is no extra cost on it

rotkehlchen/chain/optimism/types.py
```diff
@@ -1,7 +1,14 @@
...
+from typing import TYPE_CHECKING, Any, Literal, Optional
 
 from rotkehlchen.types import ChainID, ChecksumEvmAddress, EvmTransaction, EVMTxHash, Timestamp
 
+if TYPE_CHECKING:
+    # This is a list of all optimism abi that are guaranteed to be returned
+    # by EthereumContracts as they are stored in the file/DB
+    OPTIMISM_KNOWN_ABI = Literal[
+        'VELO_V2_LP',
+    ]
```
yabirgb, line 10:
```suggestion
    OPTIMISM_KNOWN_ABI = Literal['VELO_V2_LP']
```

rotkehlchen/inquirer.py
```diff
@@ -161,14 +164,17 @@ def get_underlying_asset_price(token: EvmToken) -> tuple[Optional[Price], Curren
     """
     price, oracle = None, CurrentPriceOracle.BLOCKCHAIN
     if token.protocol == UNISWAP_PROTOCOL:
-        price = Inquirer().find_uniswap_v2_lp_price(token)
+        price = Inquirer().find_lp_price(token)
```
yabirgb, line 167: Don't add more elif statmentes. Instead

```python
if token.protocol in (VELODROME_POOL_PROTOCOL, UNISWAP_PROTOCOL):
    price = Inquirer().find_lp_price(token)
    oracle = CurrentPriceOracle.BLOCKCHAIN
```

I believe is fine to go with blockchain here. `UniswapV2` exists because we also have

```
class UniswapV2Oracle(UniswapOracle):
```

that queries prices on chain using the pools but I would standarize this and use blockchain for the location

rotkehlchen/inquirer.py
```diff
@@ -342,11 +348,13 @@ def inject_evm_managers(evm_managers: Sequence[tuple[ChainID, 'EvmManager']]) ->
...
-        ethereum_manager = Inquirer()._evm_managers.get(ChainID.ETHEREUM)
-        assert ethereum_manager is not None, 'ethereum manager should have been injected'
-        return ethereum_manager  # type: ignore  # should be ethereum manager
+    def get_evm_manager(self, chain_id: ChainID) -> 'EvmManager':
+        evm_manager = self._evm_managers.get(chain_id)
+        assert evm_manager is not None, f'evm manager for chain id {chain_id} should have been injected'  # noqa: E501
+        return evm_manager
+
+    def get_ethereum_manager(self) -> 'EthereumManager':
+        return self.get_evm_manager(ChainID.ETHEREUM)  # type: ignore  # should be ethereum manager
```
yabirgb, line 357: why don't you use `get_evm_manager ` directly in the places where this method is required? Is just a one line wrapper that adds overhead

rotkehlchen/inquirer.py
```diff
@@ -722,18 +730,18 @@ def _find_usd_price(
         )
         return price, oracle, used_main_currency
 
-    def find_uniswap_v2_lp_price(
+    def find_lp_price(
             self,
             token: EvmToken,
+            abi_name: Optional['LP_TOKEN_CONTRACT_ABI_NAMES'] = None,
```
yabirgb, line 736: do you ever provide the abi? why have this argument?

rotkehlchen/tests/unit/test_inquirer.py
```diff
@@ -334,7 +334,23 @@ def test_find_uniswap_v2_lp_token_price(inquirer, ethereum_manager):
...
-    price = inquirer.find_uniswap_v2_lp_price(EvmToken(identifier))
+    price = inquirer.find_lp_price(
+        token=EvmToken(identifier),
+        abi_name='UNISWAP_V2_LP',
+    )
+    assert price is not None
+
+
+@pytest.mark.parametrize('use_clean_caching_directory', [True])
+@pytest.mark.parametrize('should_mock_current_price_queries', [True])
```
yabirgb, line 345: If you mock the price I believe you never contact the inquirer, that the fixture returns a price before asking it

rotkehlchen/tests/unit/test_inquirer.py
```diff
@@ -334,7 +334,23 @@ def test_find_uniswap_v2_lp_token_price(inquirer, ethereum_manager):
     address = '0xa2107FA5B38d9bbd2C461D6EDf11B11A50F6b974'
     identifier = ethaddress_to_identifier(address)
     inquirer.inject_evm_managers([(ChainID.ETHEREUM, ethereum_manager)])
-    price = inquirer.find_uniswap_v2_lp_price(EvmToken(identifier))
+    price = inquirer.find_lp_price(
+        token=EvmToken(identifier),
+        abi_name='UNISWAP_V2_LP',
```
yabirgb, line 339: why do you use the abi here? the code doesn't use the abi argument when querying the price

yabirgb (commented):
Few comments about the inquirer logic

rotkehlchen/tests/unit/test_inquirer.py
dimyG: This was done because there is no protocol in the uniswap v2 lp tokens. As discussed in discord, price calculation for uniswap v2 lp tokens is done based on the underlying tokens of the lp token, not based on the protocol attribute. There is no need for the abi_name parameter and it is removed.

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # decode output
+    decoded = []
+    for (method_output, method_name) in zip(output, methods):
+        call_success = True
+        if multicall_method == evm_inquirer.multicall_2:
+            call_success = method_output[0]
+            call_result = method_output[1]
+        else:
+            call_result = method_output
```
yabirgb, line 85: this logic can be removed too

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    address = token.evm_address
+    abi_name = TOKEN_PROTOCOL_TO_ABI_NAME[token.protocol]  # the lp token's contract abi that will be used to query the info needed for lp token price calculation  # noqa: E501
+    contract = EvmContract(
+        address=address,
+        abi=evm_inquirer.contracts.abi(abi_name),
+        deployed_block=0,
+    )
+    methods = ('token0', 'token1', 'totalSupply', 'getReserves', 'decimals')
+    multicall_method = evm_inquirer.multicall
```
yabirgb, line 56: don't assign variables that are used only once

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,135 @@
...
+    if isinstance(block_identifier, int) and block_identifier <= 7929876:
+        log.error(
+            f'No multicall contract at the block {block_identifier}. {abi_name} '
+            f'query failed. Should implement direct queries',
+        )
+        return None
```
yabirgb, line 62: don't leave magic numbers in the code. Define as constant

```python
MULTICALL_DEPLOYMENT_BLOCK = 7929876
````

and then 

```suggestion
    if isinstance(block_identifier, int) and block_identifier <= MULTICALL_DEPLOYMENT_BLOCK:
        log.error(
            f'No multicall contract at the block {block_identifier}. {abi_name} '
            f'query failed. Should implement direct queries',
        )
        return None
```

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,135 @@
...
+            call_result = method_output[1]
+        else:
+            call_result = method_output
+        if call_success and len(call_result) != 0:
+            decoded_method = contract.decode(call_result, method_name)
+            if len(decoded_method) == 1:
+                # https://github.com/PyCQA/pylint/issues/4739
+                decoded.append(decoded_method[0])
+            else:
+                decoded.append(decoded_method)
```
yabirgb, line 92: can you check if this is needed when using multicall? I believe is due to the difference in multicall_2/multicall but I'm not sure

rotkehlchen/inquirer.py
```diff
@@ -160,15 +160,12 @@ def get_underlying_asset_price(token: EvmToken) -> tuple[Optional[Price], Curren
     due to recursive import problems
     """
     price, oracle = None, CurrentPriceOracle.BLOCKCHAIN
-    if token.protocol == UNISWAP_PROTOCOL:
-        price = Inquirer().find_uniswap_v2_lp_price(token)
-        oracle = CurrentPriceOracle.UNISWAPV2
```
yabirgb, line 165: don't remove this. As we said we are going to use the protocol

rotkehlchen/tests/unit/test_inquirer.py
```diff
@@ -328,13 +329,40 @@ def test_price_underlying_tokens(inquirer, globaldb):
...
-    address = '0xa2107FA5B38d9bbd2C461D6EDf11B11A50F6b974'
+@pytest.mark.parametrize('should_mock_current_price_queries', [False])
+def test_find_uniswap_v2_lp_token_price(inquirer, ethereum_manager, globaldb):
+    """Tests that the Uniswap v2 lp token price is correctly found. The price calculation
+    requires the underlying tokens of the pool. Pool and lp token are the same contract"""
+    address = '0xa2107FA5B38d9bbd2C461D6EDf11B11A50F6b974'  # LINK ETH POOL
+    tuples = [
+        ('eip155:1/erc20:0x514910771AF9Ca656af840dff83E8264EcF986CA', 0.5, f'eip155:1/erc20:{address}'),  # noqa: E501
+        ('eip155:1/erc20:0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2', 0.5, f'eip155:1/erc20:{address}'),  # noqa: E501
+    ]
```
yabirgb, line 342: probably this will need to change

yabirgb: as I mentioned 0.5 is only if the pool is balanced but doesn't need to be the case in general

rotkehlchen/types.py
```diff
@@ -82,6 +84,7 @@
 YEARN_VAULTS_V1_PROTOCOL = 'yearn_vaults_v1'
 YEARN_VAULTS_V2_PROTOCOL = 'yearn_vaults_v2'
 CURVE_POOL_PROTOCOL = 'curve_pool'
+VELODROME_POOL_PROTOCOL = 'velodrome_pool'
```
yabirgb, line 87: this I have added in my PR so one of the two will have to rebase

yabirgb (commented):
some comments

rotkehlchen/chain/evm/utils.py
dimyG: It is needed as I saw.

rotkehlchen/data/global.db
yabirgb, line 1: can you maybe also update the protocol of the uniswap token being testes so you don't have to edit the db?

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,131 @@
...
+        evm_inquirer: 'EvmNodeInquirer',
+        token: EvmToken,
+        token_price_func: Callable,
+        token_price_func_args: list[Any],
+        block_identifier: BlockIdentifier,
+) -> Optional[Price]:
+    """
+    Calculate the price for an LP token the contract of which is also the contract of the
+    pool it represents. For example velodrome lp tokens. The price is calculated like this:
+    price = (Total value of liquidity pool) / (Current supply of LP tokens)
```
yabirgb, line 42: can you state at the begining of the docstring that this works for any uniswap like pool

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,131 @@
...
+    We need:
+    - Price of token 0
+    - Price of token 1
+    - Pooled amount of token 0
+    - Pooled amount of token 1
+    - Total supply of pool token
+    """
+    address = token.evm_address
+    abi_name = TOKEN_PROTOCOL_TO_ABI_NAME[token.protocol]  # the lp token's contract abi that will be used to query the info needed for lp token price calculation  # noqa: E501
```
yabirgb, line 51: this can raise keyerror

rotkehlchen/chain/evm/utils.py
yabirgb: Why is it needed? When can it have more than one element here?

yabirgb (commented):
Still a few more things

rotkehlchen/tests/unit/test_inquirer.py
```diff
@@ -328,13 +329,35 @@ def test_price_underlying_tokens(inquirer, globaldb):
...
+    address = '0xa2107FA5B38d9bbd2C461D6EDf11B11A50F6b974'  # LINK ETH POOL
     identifier = ethaddress_to_identifier(address)
+    with globaldb.conn.write_ctx() as write_cursor:
+        write_cursor.execute(
+            'UPDATE evm_tokens SET protocol=? WHERE identifier=?',
+            ('UNI-V2', identifier),
+        )
     inquirer.inject_evm_managers([(ChainID.ETHEREUM, ethereum_manager)])
-    price = inquirer.find_uniswap_v2_lp_price(EvmToken(identifier))
+    price, _ = get_underlying_asset_price(token=EvmToken(identifier))
```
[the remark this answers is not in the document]
dimyG: The get_underlying_asset_price calls the find_pool_price that finds the uniswap price so it is ok, but I can changed it to call the function directly.

rotkehlchen/data/global.db
dimyG: Hmm, do you think this is the best thing to do? I mean updating only one of the uniswap v2 tokens and have all the rest without protocol. It wouldd not be consistent and maybe lead to difficualties underatnding why in the future? Maybe we can update all of them in the packaged db. I think we can identify them from the symbol right?

LefterisJP: no no Yabir explained why you can't do this in another post. I also wrote about this yesterday. We can't find them all and we can't identify by symbol. This is error prone and can be tricked by fake tokens.

There is no problem by having some without the protocol. What matters is to have it during decoding, and when we are decoding something uniswap related.

. You just update them as you find them

rotkehlchen/chain/evm/utils.py
dimyG: The token supplies are returned as a tuple of ints. So in that case the decoded_method is a tuple and we need to add the tuple in the decoded list. later we access elements of that tuple (decoded[3][0] and decoded[3][1]:

```
        token0_supply = FVal(decoded[3][0] * 10**-token0.decimals)
        token1_supply = FVal(decoded[3][1] * 10**-token1.decimals)
```

dimyG:
the new sqldiff applied on latest develop (that has Base chain related changes). 
https://gist.github.com/dimyG/5a06ae65cf8757481a0180f9f31ab5c8

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,140 @@
...
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+LP_TOKEN_AS_POOL_PROTOCOL_TO_ABI_NAME: Final[dict[str, 'LP_TOKEN_AS_POOL_CONTRACT_ABIS']] = {
+    VELODROME_POOL_PROTOCOL: 'VELO_V2_LP',
+    UNISWAP_PROTOCOL: 'UNISWAP_V2_LP',
+}
+MULTICALL_DEPLOYMENT_BLOCK = 7929876
```
LefterisJP, line 29: This is wrong here. You are using a constant for the multicall we use in mainnet. That is a mainnet block for a particular multicall version.

The code here is supposed to be **EVM chain agnostic**.

We are lucky though, since we have this data per chain. Look further down in the review.

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def lp_price_from_pool_contract(
```
LefterisJP, line 32: Isn't this a better name if it's uniswaplike? 

```suggestion
def lp_price_from_uniswaplike_pool_contract(
```

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,140 @@
...
+    tokens. The price is calculated like this:
+    price = (Total value of liquidity pool) / (Current supply of LP tokens)
+    We need:
+    - Price of token 0
+    - Price of token 1
+    - Pooled amount of token 0
+    - Pooled amount of token 1
+    - Total supply of pool token
+    """
+    address = token.evm_address
```
LefterisJP, line 51: I find a general name like `address` in a variable here really confusing and hard to read. Just use `token.evm_address` directly wherever you need it. It's much more explicit

```suggestion
```

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,140 @@
...
+        )
+        return None
+
+    contract = EvmContract(
+        address=address,
+        abi=evm_inquirer.contracts.abi(abi_name),
+        deployed_block=0,
+    )
+    methods = ('token0', 'token1', 'totalSupply', 'getReserves', 'decimals')
+    if isinstance(block_identifier, int) and block_identifier <= MULTICALL_DEPLOYMENT_BLOCK:
```
LefterisJP, line 67: We thankfully keep deployed block attributes in `EvmContract` class! So easy fix here. Pheww!

```suggestion
    if isinstance(block_identifier, int) and block_identifier <= evm_inquirer.contract_multicall.deployed_block:
```

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,140 @@
...
+        log.error(
+            f'No multicall contract at the block {block_identifier}. {abi_name} '
```
LefterisJP, line 69:
```suggestion
            f'No multicall contract at {evm_inquirer.chain_name} block {block_identifier}. {abi_name} '
```

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,140 @@
...
+        return None
+
+    # decode output
+    decoded = []
+    for (method_output, method_name) in zip(output, methods):
+        call_success = True
+        if call_success and len(method_output) != 0:
+            decoded_method = contract.decode(method_output, method_name)
+            if len(decoded_method) == 1:
+                # https://github.com/PyCQA/pylint/issues/4739
```
LefterisJP, line 94: This was supposed to explain a pylint ignore, if you check the link. I don't see the pylint ignore statement there anymore.

```suggestion
```

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,140 @@
...
+            chain_id=token.chain_id,
+            token_type=EvmTokenKind.ERC20,
+        ))
+        token1 = EvmToken(evm_address_to_identifier(
+            address=decoded[1],
+            chain_id=token.chain_id,
+            token_type=EvmTokenKind.ERC20,
+        ))
+    except (UnknownAsset, WrongAssetType):
+        log.debug(f'Unknown assets while querying {abi_name} price {decoded[0]} {decoded[1]}')
```
LefterisJP, line 117:
```suggestion
        log.debug(f'Unknown assets while querying {evm_inquirer.chain_name} {abi_name} price {decoded[0]} {decoded[1]}')
```

rotkehlchen/inquirer.py
```diff
@@ -722,14 +718,14 @@ def _find_usd_price(
         )
         return price, oracle, used_main_currency
 
-    def find_uniswap_v2_lp_price(
+    def find_lp_price_from_pool_contract(
```
LefterisJP, line 721: As Yabir also mentioned before. Use different name.
```suggestion
    def find_lp_price_from_uniswaplike_pool(
```

rotkehlchen/inquirer.py
```diff
@@ -722,14 +718,14 @@ def _find_usd_price(
...
             self,
             token: EvmToken,
     ) -> Optional[Price]:
-        # BAD BAD BAD. TODO: Need to rethinking placement of modules here
-        from rotkehlchen.chain.ethereum.modules.uniswap.utils import find_uniswap_v2_lp_price  # isort:skip  # noqa: E501  # pylint: disable=import-outside-toplevel
-        return find_uniswap_v2_lp_price(
-            ethereum=self.get_ethereum_manager().node_inquirer,
+        """Calculates the price for an LP token the contract of which is also the contract of
```
LefterisJP, line 725: mention uniswaplike here, but rest of description sounds okay
```suggestion
        """Calculates the price for a uniswaplike LP token the contract of which is also the contract of
```

rotkehlchen/data/global.db
LefterisJP, line 1: Why is this PR adding all these diff? https://gist.github.com/dimyG/5a06ae65cf8757481a0180f9f31ab5c8

This is crazy. We don't need to add all the possible pools for velodrome. This is adding 1500 assets to the packaged global.db

Undo this change. Just like we don't have every possible uniswap pool (they are more than 300k), we can't have every possible velodrom pool.

WE only add them in the globalDB as they are discovered through decoding, because the user interacted with them.

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,140 @@
...
+            else:
+                decoded.append(decoded_method)
+        else:
+            log.debug(
+                f'Multicall to {abi_name} failed to fetch field {method_name} '
+                f'for token {token.evm_address}',
+            )
+            return None
+
+    try:
```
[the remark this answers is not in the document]
dimyG: OK. I don't think it can raise any other exception

rotkehlchen/data/global.db
dimyG: ok, this is the new sqldiff:
https://gist.github.com/dimyG/0a57034b670f3aeef2ff4376b4ba6644 it has only one asset used in tests

rotkehlchen/chain/evm/utils.py
dimyG: yes, that was my question too. Maybe it has been fixed now although the linked issue is still open.

rotkehlchen/chain/evm/utils.py
```diff
@@ -0,0 +1,170 @@
...
+        ))
+    except (UnknownAsset, WrongAssetType):
+        log.debug(
+            f'Unknown assets {decoded[0]} {decoded[1]} while querying price from '
+            f'{evm_inquirer.chain_name} {abi_name} for token {token.evm_address}',
+        )
+        return None
+
+    name = 'token supply'
+    location = 'uniswap-like pool price query'
```
LefterisJP, line 137: Remove them as function local variables since in esense they are constants.

Declare them in the top ofthe file with the `Final` modifier to mark them as constants and use them in the function

```suggestion
```

rotkehlchen/data/global.db
LefterisJP, line 1: The DB looks better now but still you mentioned [here](https://github.com/rotki/rotki/pull/6653#discussion_r1341457087) that you addded to the production global DB an asset for tests. Please undo this and provide the new diff: https://gist.github.com/dimyG/0a57034b670f3aeef2ff4376b4ba6644

If an asset is needed in tests you can easily add it in the particular test it's needed.

rotkehlchen/chain/evm/utils.py
LefterisJP: no no the issue is not fixed. But at some point code changed, type:ignore was removed since it was not needed but the comment explaining why it was there was not removed.

rotkehlchen/data/global.db
dimyG: ok, this is the new diff:
https://gist.github.com/dimyG/beb59c97502782a5024ae61b0338016f

LefterisJP (commented):
Looks okay now but tests are failing :(

dimyG:
this is the new sql diff. i removed the addition of the protocol attribute. it's only the new abi now. 
https://gist.github.com/dimyG/71f6c71f15d6b426f8c0c6a373ecc276

LefterisJP (approved):
LGTM. Merging.

rotkehlchen/tests/unit/test_inquirer.py
```diff
@@ -328,13 +336,37 @@ def test_price_underlying_tokens(inquirer, globaldb):
...
-def test_find_uniswap_v2_lp_token_price(inquirer, ethereum_manager):
-    address = '0xa2107FA5B38d9bbd2C461D6EDf11B11A50F6b974'
-    identifier = ethaddress_to_identifier(address)
+@pytest.mark.parametrize('should_mock_current_price_queries', [False])
+def test_find_uniswap_v2_lp_token_price(inquirer, ethereum_manager, globaldb):
+    """Tests that the Uniswap v2 lp token's price is correctly found. The special price
+    calculation that is needed, is applied based on the protocol attribute of the lp token"""
+    identifier = ethaddress_to_identifier(string_to_evm_address('0xa2107FA5B38d9bbd2C461D6EDf11B11A50F6b974'))  # LINK ETH POOL  # noqa: E501
+    with globaldb.conn.write_ctx() as write_cursor:
+        write_cursor.execute(  # the protocol attribute is missing from the packaged db for this token and is necessary for price calculation  # noqa: E501
```
LefterisJP, line 347: Two things here.

1. This comment should mention the time
2. Making this small fix for the packaged global DB would not be bad if the token is already there and you just fix the protocol. (but will merge now as it's already green -- can adjust later)
```suggestion
        write_cursor.execute(  # the protocol attribute is missing from the packaged db as of this commit for this token and is necessary for price calculation  # noqa: E501
```

LefterisJP (approved):

PR #6655, opened 2023-09-28
Fix balances not working if address has only locked cvx

yabirgb:
If the address doesn't have convex deposits and only locked cvx the address is not present in the balances mapping and therefore it fails when adding the CVX balances

### DIFF
--- a/rotkehlchen/tests/unit/test_protocol_balances.py
+++ b/rotkehlchen/tests/unit/test_protocol_balances.py
@@ -117,6 +117,39 @@ def test_convex_staking_balances(
 
 
 @pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('ethereum_accounts', [['0x0c3ce74FCB2B93F9244544919572818Dc2AC0641']])
+def test_convex_staking_balances_without_gauges(
+        ethereum_inquirer: 'EthereumInquirer',
+        ethereum_transaction_decoder: 'EthereumTransactionDecoder',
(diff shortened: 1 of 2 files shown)


LefterisJP (commented):
I can see the failure here. But I would like a small unit test where this function fails. We need to make the code well tested, so that if someone removes code arbitrarily tests tell us that a fuckup was just made

rotkehlchen/tests/unit/test_protocol_balances.py
```diff
@@ -116,6 +116,39 @@ def test_convex_staking_balances(
...
+def test_convex_staking_balances_without_gauges(
+        ethereum_inquirer: 'EthereumInquirer',
+        ethereum_transaction_decoder: 'EthereumTransactionDecoder',
+        ethereum_accounts: list[ChecksumEvmAddress],
+        inquirer: 'Inquirer',  # pylint: disable=unused-argument
+) -> None:
+    """
+    Check that convex balances are correctly propagated if one account doesn't have gauges
+    deposits but has staked CVX since staked/locked CVX is added to the balances returned from
+    gauges.
```
LefterisJP, line 130:
```suggestion
    deposits but has staked CVX. The reason for this test is that staked/locked CVX is added to the balances 
    returned from the gauges and those balances before this commit were not a defaultdict and could lead to a failure.
```

LefterisJP (approved):
LGTM. Auto-applied change and will squash merge

PR #6670, opened 2023-10-01
Fix wrong name in some ENS events. Use name/labelhash mapping cache

### LINKED ISSUE #6667: ENS decoders can report wrong ENS name for some activities
## Problem Definition

At the moment the ENS decoders try to show the ENS name involved in any ENS action whenevr possible.

That sometimes does not work. I noticed it for my own name. [This](https://etherscan.io/tx/0x07aa7d1ac61fc03f6416a25c0d6cf96f286e2ce84e9b350dd2a9a1bd6426aef2#eventlog) transaction changed the avatar of hania.eth which is controlled by me (lefteris.eth).

rotki shows it as if I changed the avatar of lefteris.eth. Which is wrong.

The reason for this is this part of the logic: 
https://github.com/rotki/rotki/blob/06aabefe900253ff67369e42d6a7a63feb82193c/rotkehlchen/chain/ethereum/modules/ens/decoder.py#L275-L312

- We find the address that owns/controls the node
- Then we find the primary ens name of that address

But that logic is wrong! As can be seen by the transaction above `0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12` which has `lefteris.eth` as its primary ENS names, sets the avatar of node `0xa1e797aff66753be23f117cef7bb915d73410a71bf7a3f34ffa215a2145440a7`.
 
currently rotki sees that `0xa1e797aff66753be23f117cef7bb915d73410a71bf7a3f34ffa215a2145440a7` is controlled by `0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12` and then sees that its primary name is `lefteris.eth`.

But what we need rotki to do is find the name corresponding to the namehas/hone `0xa1e797aff66753be23f117cef7bb915d73410a71bf7a3f34ffa215a2145440a7` which is hania.eth

## Task

Apparently ... this is hard!
https://twitter.com/gregskril/status/1707828814699495617

The ways to go seem to either use a subgraph or a mapping if you know the names already.

So for us the way to go here is probably multipronged.

1. Add a node_to_ensname mapping to the DB cache. Have it populated via the decoding of events. If any event registers a name or has anything else that shows you plain text name + node, add them to the mapping
2. When we get to this `_get_name_to_show` function look up this mapping we may have populated.
    - a. If it's there use it
    - b. If not there either, give up and use no name in the event description ... or use subgraph and if that fails fall back to no name.

LefterisJP:
Fix #6667

- Save namehash and labelhash to ENS name mapping in the DB whenever possible.
- When finding an ENS node/namehash ask DB if we know it. If not ask the subgraph, to get the final ens name
- Adjust the ENS decoder to fix inconsistencies with the ENS name and utilize this new namehash/labelhash querying logic

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -10,6 +10,7 @@ Changelog
 * :feature:`-` Transactions bridging from/to Base using the official bridge will be now decoded.
 * :bug:`-` History event header will now be visible only on mobile.
 * :bug:`6647` When PnL report generation is finished, users will not be redirected to the report page, but will get notified instead.
+* :bug:`6667` Wrong ENS name should no longer be reported for some edge case of ENS actions and the name should also appear in more events than before.
 
 * :release:`1.30.2 <2023-09-21>`
 * :feature:`-` Improved support for importing Binance CSV files.
--- a/rotkehlchen/chain/ethereum/modules/ens/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/ens/decoder.py
@@ -2,14 +2,13 @@ import logging
 from typing import TYPE_CHECKING, Any, Optional
 
 import content_hash
-from eth_utils import to_checksum_address
+from ens import ENS
 
 from rotkehlchen.accounting.structures.balance import Balance
 from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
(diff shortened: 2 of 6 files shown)


LefterisJP:
All green at first try. I am that good.

PR #6673, opened 2023-10-02
paginated addressbook

jbrit:
## Checklist

- [x] The PR modified the backend to modify the address book endpoint

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -5020,9 +5020,9 @@ Exporting History Events
    :statuscode 409: No user is logged in or failure at event export.
    :statuscode 500: Internal rotki error
 
-.. http:patch:: /api/(version)/history/events/export
+.. http:put:: /api/(version)/history/events/export
 
-   Doing a PATCH on this endpoint with the given filter parameters will download a csv with all history events matching the filter. All arguments are optional. If no filter is used all the events will be downloaded.
+   Doing a PUT on this endpoint with the given filter parameters will download a csv with all history events matching the filter. All arguments are optional. If no filter is used all the events will be downloaded.
 
    .. _filter-request-args-label:
 
@@ -5030,7 +5030,7 @@ Exporting History Events
 
    .. http:example:: curl wget httpie python-requests
 
-      PATCH /api/1/history/events/export HTTP/1.1
+      PUT /api/1/history/events/export HTTP/1.1
       Host: localhost:5042
       Content-Type: application/json;charset=UTF-8
 
@@ -11156,7 +11156,7 @@ Get mappings from addressbook
 
 .. http:post:: /api/(version)/names/addressbook
 
-    Doing a POST on the addressbook endpoint with either /global or /private postfix with a list of addresses will return found address mappings for specified addresses. If addresses parameter isn't specified, all known mappings are returned.
+    Doing a POST on the addressbook endpoint with either /global or /private postfix with a list of addresses will return found address mappings with the specified filter arguments. If no filter argument is specified, all known mappings are returned.
 
     **Example Request**
 
@@ -11167,13 +11167,21 @@ Get mappings from addressbook
         Content-Type: application/json;charset=UTF-8
 
         {
+          "offset": 0,
+          "limit": 1,
+          "blockchain": "eth",
+          "name_substring": "neighbour",
           "addresses": [
             {"address": "0x9531c059098e3d194ff87febb587ab07b30b1306", "blockchain": "eth"},
             {"address": "0x8A4973ABBCEd48596D6D79ac6B53Ceda65e342CD"}
            ]
         }
 
-    :reqjson object addresses: List of addresses that the backend should find names for. When the blockchain field is missing it will delete the other entries for the same address and the name will be used for all blockchains.
+    :reqjson int[optional] limit: This signifies the limit of records to return as per the `sql spec <https://www.sqlite.org/lang_select.html#limitoffset>`__.
+    :reqjson int[optional] offset: This signifies the offset from which to start the return of records per the `sql spec <https://www.sqlite.org/lang_select.html#limitoffset>`__.
+    :reqjson str[optional] name_substring: The substring to use as filter for the name to be found in the addressbook.
+    :reqjson str[optional] blockchain: The chain in which to use the provided name.
+    :reqjson object[optional] addresses: List of addresses that the backend should find names for.
 
     **Example Response**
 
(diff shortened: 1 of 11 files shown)


yabirgb (commented):
the new address in the evm detection is expected to fail. For the other error I'm not sure what happened... are you rebased?

rotkehlchen/db/addressbook.py
```diff
@@ -48,45 +49,28 @@ def write_ctx(self, book_type: AddressbookType) -> Generator['DBCursor', None, N
     def get_addressbook_entries(
             self,
             cursor: 'DBCursor',
-            optional_chain_addresses: Optional[list[OptionalChainAddress]] = None,
-    ) -> list[AddressbookEntry]:
+            filter_query: Optional['AddressbookFilterQuery'] = None,
```
yabirgb, line 52: why make it optional? It complicates the logic and irc in other similar functions we always provide it, even if is empty

docs/api.rst
```diff
@@ -5020,17 +5020,17 @@ Exporting History Events
    :statuscode 409: No user is logged in or failure at event export.
    :statuscode 500: Internal rotki error
 
-.. http:patch:: /api/(version)/history/events/export
+.. http:put:: /api/(version)/history/events/export
 
-   Doing a PATCH on this endpoint with the given filter parameters will download a csv with all history events matching the filter. All arguments are optional. If no filter is used all the events will be downloaded.
+   Doing a PUT on this endpoint with the given filter parameters will download a csv with all history events matching the filter. All arguments are optional. If no filter is used all the events will be downloaded.
```
yabirgb, line 5025: is this change intentional? was it wrong?

yabirgb (commented):
One issue and a few questions

rotkehlchen/db/addressbook.py
jbrit: You mean, I shouldn't provide a default value of None?, or I should not make it optional at all? i.e. users should create an empty filter?

docs/api.rst
jbrit: It's intentional, I'm pretty sure this was on develop before

docs/api.rst
```diff
@@ -5020,17 +5020,17 @@ Exporting History Events
    :statuscode 409: No user is logged in or failure at event export.
    :statuscode 500: Internal rotki error
 
-.. http:patch:: /api/(version)/history/events/export
+.. http:put:: /api/(version)/history/events/export
```
LefterisJP, line 5023: This is a correct change but there is one more thing off. The name of the function here should be `put`:

https://github.com/rotki/rotki/blob/2b358c20a052729238f79c24733d88bc8448e6d9/rotkehlchen/api/v1/resources.py#L2971

Please do it since you are changing this here in the docs

rotkehlchen/db/filtering.py
```diff
@@ -580,6 +582,25 @@ class DBMultiIntegerFilter(DBMultiValueFilter[int]):
...
+class DBOptionalChainAddressesFilter(DBFilter):
+    """Filter the address column by a selection of optional chain addresses"""
+    optional_chain_addresses: Optional[list[OptionalChainAddress]]
+
+    def prepare(self) -> tuple[list[str], list[str]]:
+        query_filters = []
+        bindings: list[Union[str, ChecksumEvmAddress]] = []
+        if self.optional_chain_addresses is not None:
+            for optional_chain_address in self.optional_chain_addresses:
+                query_part = 'address = ?'
```
LefterisJP, line 595: This is a weird way of doing it. Instead I want you to use already existing functionality for filtering for a chain of addresses.

Check `DBTransactionsPendingDecodingFilter` in this same file: https://github.com/rotki/rotki/blob/2b358c20a052729238f79c24733d88bc8448e6d9/rotkehlchen/db/filtering.py#L174

And instead abstract that logic and move it here. And then use this `DBOptionalChainAddressesFilter` in `DBTransactionsPendingDecodingFilter` through inheritance as is done in other places in this file.

rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -56,7 +57,7 @@ def test_get_addressbook(
         ),
     )
     result = assert_proper_response_with_result(response)
-    deserialized_entries = [AddressbookEntry.deserialize(data=raw_entry) for raw_entry in result]
+    deserialized_entries = [AddressbookEntry.deserialize(data=raw_entry) for raw_entry in result['entries']]  # noqa: E501
```
LefterisJP, line 60: here assert for entries found and total too

rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -73,9 +74,8 @@ def test_get_addressbook(
         },
     )
     result = assert_proper_response_with_result(response=response)
-    expected_entries = generated_entries[0:3]
-    deserialized_entries = [AddressbookEntry.deserialize(data=raw_entry) for raw_entry in result]
-    assert set(deserialized_entries) == set(expected_entries)
+    deserialized_entries = [AddressbookEntry.deserialize(data=raw_entry) for raw_entry in result['entries']]  # noqa: E501
```
LefterisJP, line 77: same here

rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -86,14 +86,72 @@ def test_get_addressbook(
...
+                    'blockchain': SupportedBlockchain.OPTIMISM.serialize(),
+                },
             ],
         },
     )
     result = assert_proper_response_with_result(response=response)
-    expected_entries = [generated_entries[1]]
-    deserialized_entries = [AddressbookEntry.deserialize(data=raw_entry) for raw_entry in result]
-    assert set(deserialized_entries) == set(expected_entries)
+    deserialized_entries = [AddressbookEntry.deserialize(data=raw_entry) for raw_entry in result['entries']]  # noqa: E501
```
LefterisJP, line 97: same here

rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -86,14 +86,72 @@ def test_get_addressbook(
...
+            'offset': 1,
+            'limit': 1,
+            'blockchain': SupportedBlockchain.ETHEREUM.serialize(),
+        },
+    )
+    result = assert_proper_response_with_result(response=response)
+    deserialized_entries = [AddressbookEntry.deserialize(data=raw_entry) for raw_entry in result['entries']]  # noqa: E501
+    assert set(deserialized_entries) == {generated_entries[2]}
+    assert result['entries_found'] == 2
+    assert result['entries_total'] == 5
```
LefterisJP, line 154: After this you should also add a filter by a list of addresses. Make sure to have at least two tests. One for 1 address and 1 for 2 addresses.

LefterisJP (commented):
Left some comments

rotkehlchen/db/filtering.py
jbrit: They two filters are very different, and check different fields

LefterisJP: Discussed this in a call and this is indeed two different things as this is using:

```python
class OptionalChainAddress(NamedTuple):
    address: ChecksumAddress
    blockchain: Optional[SupportedBlockchain]
    ```

My comment is invalid.

LefterisJP (approved):
LGTM. Thank you Ajibolah

PR #6678, opened 2023-10-03
Fix issue with trade locations for Ledger actions

tewshi:
this TradeLocation enum is used in defining trade locations, and breaks when parsing non-existing values, switch to string so the location accepts what backend sends


frontend/app/src/types/history/trade/location.ts
```diff
@@ -7,7 +7,9 @@ export const TradeLocation = z.enum([
   ...SUPPORTED_EXCHANGES,
   ...SUPPORTED_TRADE_LOCATIONS,
   'gitcoin',
-  'loopring'
+  'loopring',
+  'polygon pos',
+  'arbitrum one'
```
lukicenturi, line 12: these two are inside `SUPPORTED_TRADE_LOCATIONS`. So just change the hardcoded enum. I think it's the same with what I do https://github.com/rotki/rotki/pull/6676

tewshi: i've removed them, what is inside `SUPPORTED_TRADE_LOCATIONS` has underscores
`polygon_pos` `arbitrum_one`

lukicenturi: > i've removed them, what is inside `SUPPORTED_TRADE_LOCATIONS` has underscores `polygon_pos` `arbitrum_one`

yeah just remove the underscore (?). Because the correct location is without underscore. I think the `<LocationDisplay/>` still breaks

tewshi: Okay, thanks

frontend/app/src/data/defaults.ts
```diff
@@ -71,9 +71,9 @@ export const TRADE_LOCATION_ETHEREUM = 'ethereum';
 
 export const TRADE_LOCATION_OPTIMISM = 'optimism';
 
-export const TRADE_LOCATION_POLYGON_POS = 'polygon_pos';
+export const TRADE_LOCATION_POLYGON_POS = 'polygon pos';
 
-export const TRADE_LOCATION_ARBITRUM_ONE = 'arbitrum_one';
+export const TRADE_LOCATION_ARBITRUM_ONE = 'arbitrum one';
```
lukicenturi, line 76: You still need to put it inside `SUPPORTED_TRADE_LOCATIONS`

tewshi: missed it... thanks

kelsos (approved):
lgtm

PR #6679, opened 2023-10-03
Abstract code for reloadable decoders

dimyG:
Closes #(issue_number)
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=39482896

## Checklist

### DIFF
--- a/rotkehlchen/chain/evm/decoding/interfaces.py
+++ b/rotkehlchen/chain/evm/decoding/interfaces.py
@@ -1,6 +1,7 @@
+import logging
 from abc import ABCMeta, abstractmethod
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, Any, Callable, Optional
+from typing import TYPE_CHECKING, Any, Callable, Optional, Union
 
 from rotkehlchen.accounting.structures.balance import Balance
 from rotkehlchen.accounting.structures.evm_event import EvmProduct
@@ -11,16 +12,30 @@ from rotkehlchen.chain.evm.decoding.structures import (
     DecodingOutput,
 )
 from rotkehlchen.constants.assets import A_ETH
-from rotkehlchen.types import ChecksumEvmAddress, DecoderEventMappingType
+from rotkehlchen.logging import RotkehlchenLogsAdapter
+from rotkehlchen.types import CacheType, ChecksumEvmAddress, DecoderEventMappingType
 from rotkehlchen.utils.misc import hex_or_bytes_to_address, hex_or_bytes_to_int
 
 if TYPE_CHECKING:
     from rotkehlchen.accounting.structures.evm_event import EvmEvent
+    from rotkehlchen.chain.ethereum.modules.curve.curve_cache import (
+        READ_CURVE_DATA_TYPE,
+        CurvePoolData,
+    )
+    from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
     from rotkehlchen.chain.evm.decoding.types import CounterpartyDetails
     from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
+    from rotkehlchen.chain.optimism.modules.velodrome.velodrome_cache import VelodromePoolData
+    from rotkehlchen.chain.optimism.node_inquirer import OptimismInquirer
+    from rotkehlchen.db.dbhandler import DBHandler
+    from rotkehlchen.db.drivers.gevent import DBCursor
     from rotkehlchen.user_messages import MessagesAggregator
 
     from .base import BaseDecoderTools
+
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
 
 
 class DecoderInterface(metaclass=ABCMeta):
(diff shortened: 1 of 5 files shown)


rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -171,8 +171,64 @@ def _decode_vote_cast(self, context: DecoderContext) -> DecodingOutput:
 
 class ReloadableDecoderMixin(metaclass=ABCMeta):
 
+    def __init__(
+            self,
+            cache_type_to_check_for_freshness: CacheType,
+            query_data_method: Callable,
+            save_data_to_cache_method: Callable,
+            read_pools_and_gauges_from_cache_method: Callable,
+    ) -> None:
```
yabirgb, line 180: I make a similar comment to what happened in the big velodrome PR. This interface is tough against porotocls with pools and gauges but `ReloadableDecoderMixin` as it was before can be used for anything for example yearn vaults or anything that might require to reload datal. I would create a subinterface that implements what you did

rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -171,8 +171,64 @@ def _decode_vote_cast(self, context: DecoderContext) -> DecodingOutput:
 
 class ReloadableDecoderMixin(metaclass=ABCMeta):
```
yabirgb, line 173: I would add a docstring for the usage that this interface is intended for since the arguments names and defenitions are too generic

rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -171,8 +171,64 @@ def _decode_vote_cast(self, context: DecoderContext) -> DecodingOutput:
...
 
+    def __init__(
+            self,
+            cache_type_to_check_for_freshness: CacheType,
+            query_data_method: Callable,
+            save_data_to_cache_method: Callable,
+            read_pools_and_gauges_from_cache_method: Callable,
```
yabirgb, line 179: not sure if you can add more information to the types in the return part of the callable.

```suggestion
            read_pools_and_gauges_from_cache_method: Callable[[], [list[str], list[str]]],
```

rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -171,8 +171,64 @@ def _decode_vote_cast(self, context: DecoderContext) -> DecodingOutput:
...
+        self.evm_inquirer.ensure_cache_data_is_updated(  # type: ignore  # mypy doesn't understand that the optimism inquirer is a DSProxyInquirerWithCacheData with an ensure_cache_data_is_updated method  # noqa: E501
+            cache_type=self.cache_type_to_check_for_freshness,
+            query_method=self.query_data_method,
+            save_method=self.save_data_to_cache_method,
+        )
+        new_pools, new_gauges = self.read_pools_and_gauges_from_cache_method()
+        if isinstance(new_pools, dict):  # a curve pool is a dict of lp token to pool addresses
+            pools_diff = set(new_pools.keys()) - set(self.pools.keys())
+        else:
+            pools_diff = new_pools - self.pools
```
yabirgb, line 218:
```suggestion
        if isinstance(new_pools, dict):  # a pool is a dict of lp token to pool addresses
            pools_diff = set(new_pools.keys()) - set(self.pools.keys())
        else:
            pools_diff = new_pools - self.pools
```

rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -171,8 +171,64 @@ def _decode_vote_cast(self, context: DecoderContext) -> DecodingOutput:
...
+            return None
+
+        self.pools = new_pools
+        self.gauges = new_gauges
+        new_mapping: dict[ChecksumEvmAddress, tuple[Any, ...]] = {
+            pool_address: (self._decode_pool_events,)
+            for pool_address in pools_diff
+        }
+        new_mapping.update({
+            # addresses of pools and gauges don't intersect, so combining like this is fine
```
yabirgb, line 230: you are not combining anything so I'm not sure about this comment

rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -171,8 +171,64 @@ def _decode_vote_cast(self, context: DecoderContext) -> DecodingOutput:
...
+        """
+        self.cache_type_to_check_for_freshness = cache_type_to_check_for_freshness
+        self.query_data_method = query_data_method
+        self.save_data_to_cache_method = save_data_to_cache_method
+        self.read_pools_and_gauges_from_cache_method = read_pools_and_gauges_from_cache_method
+        self.pools, self.gauges = self.read_pools_and_gauges_from_cache_method()
+
     @abstractmethod
+    def _decode_pool_events(self, context: DecoderContext) -> DecodingOutput:
+        """Subclasses should implement this method. It decodes pool related events."""
```
yabirgb, line 196: This comment is a bit useless since the abstractmethod decorator already tells you that needs to be implemented.

```suggestion
        """Decodes events related to protocol pools""
```

Same for the gauges

rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -171,8 +171,64 @@ def _decode_vote_cast(self, context: DecoderContext) -> DecodingOutput:
...
-        """Subclasses may implement this to be able to reload some of the decoder's properties
-        Returns only new mappings of addresses to decode functions
+        """Make sure pools and gauges are recently queried from the remote source,
+        saved in the DB and loaded to the decoder's memory.
+
+        If a query happens and any new mappings are generated they are returned,
+        otherwise `None` is returned.
         """
+        self.evm_inquirer.ensure_cache_data_is_updated(  # type: ignore  # mypy doesn't understand that the optimism inquirer is a DSProxyInquirerWithCacheData with an ensure_cache_data_is_updated method  # noqa: E501
+            cache_type=self.cache_type_to_check_for_freshness,
```
yabirgb, line 210: this is not totally correct. This is missing the evm inquirer in the definition of the interface and this is why it fails first. Later it fails because of the dsproxy issue but this is missing the evm_inquierer definition in the interface

rotkehlchen/chain/evm/decoding/interfaces.py
dimyG: I'm not sure about the value of data types for the other callables here. mypy seems to not handle them very well yet. it requires type:ignone wehn passing the methods as arguments. 
image

rotkehlchen/chain/evm/decoding/interfaces.py
dimyG: I'm not sure about the value of data types for the other callables here. mypy seems to not handle them very well yet. it requires type:ignone wehn passing the methods as arguments. 
image

rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -170,9 +185,91 @@ def _decode_vote_cast(self, context: DecoderContext) -> DecodingOutput:
...
+    """Used by decoders of protocols that have pools and gauges stored in the globaldb cache
+    tables. It can reload them to the decoder's memory.
+    """
+
+    def __init__(
+            self,
+            evm_inquirer: 'EvmNodeInquirer',
+            cache_type_to_check_for_freshness: CacheType,
+            query_data_method:
+            Callable[['OptimismInquirer'], Optional[list['VelodromePoolData']]] |
```
LefterisJP, line 207: As mentioned in discord using `|` is incosistent with the rest of the codebase. Though we will switch to that notation soon.

rotkehlchen/chain/evm/decoding/interfaces.py
```diff
@@ -170,9 +185,91 @@ def _decode_vote_cast(self, context: DecoderContext) -> DecodingOutput:
...
+    remote data source, to the decoder's memory."""
 
     @abstractmethod
     def reload_data(self) -> Optional[Mapping[ChecksumEvmAddress, tuple[Any, ...]]]:
         """Subclasses may implement this to be able to reload some of the decoder's properties
-        Returns only new mappings of addresses to decode functions
+        Returns only new mappings of addresses to decode functions"""
+
+
+class ReloadablePoolsAndGaugesDecoderMixin(ReloadableDecoderMixin):
```
LefterisJP, line 197: Shouldn't this also be an abstract class? It should never be instantiated on its own.

LefterisJP (commented):
Looks good. Just 2 small comments and a changelog for velodrome addition request. There is no frontend specific work for this right? So changelog should be added and the issue closed!

rotkehlchen/chain/evm/decoding/interfaces.py
LefterisJP: This seems wrong. mypy handles callable arguments quite well. But I guess you guys figured it out later? Please check mypy errors very well. type: ignore is only for mypy errors and bugs or things we can't fix yet.

LefterisJP (approved):
LGTM. Merging

PR #6690, opened 2023-10-04
Remove LedgerActions from the codebase

### LINKED ISSUE #6096: Replace ledger actions with history events

LefterisJP:
Fix #6096


rotkehlchen/accounting/structures/types.py
```diff
@@ -37,8 +37,8 @@
 class ActionType(DBCharEnumMixIn):
     TRADE = 1
     ASSET_MOVEMENT = 2
-    EVM_TRANSACTION = 3
-    LEDGER_ACTION = 4
```
yabirgb, line 41: Should we delete ledger action from the db too?

rotkehlchen/chain/evm/accounting/airdrop.py
```diff
@@ -26,10 +25,9 @@ def __init__(
 
     def event_settings(self, pot: 'AccountingPot') -> dict[str, 'TxEventSettings']:
         """Being defined at function call time is fine since this function is called only once"""
-        airdrops_taxable = LedgerActionType.AIRDROP in pot.settings.taxable_ledger_actions
         return {
             get_tx_event_type_identifier(HistoryEventType.RECEIVE, HistoryEventSubType.AIRDROP, protocol): TxEventSettings(  # noqa: E501
-                taxable=airdrops_taxable,
+                taxable=False,  # this used to depend on ledger actions. Needs https://github.com/rotki/rotki/issues/4341  # noqa: E501
```
yabirgb, line 30: 👍

rotkehlchen/db/upgrades/v39_v40.py
```diff
@@ -83,6 +128,114 @@ def _add_base_chain_data(write_cursor: 'DBCursor') -> None:
...
+                event_type = 'spend'
+                event_subtype = 'none'
+            elif ledger_action_type == 'E':
+                event_type = 'receive'
+                event_subtype = 'donate'
+            elif ledger_action_type == 'F':
+                event_type = 'receive'
+                event_subtype = 'airdrop'
+            else:
+                raise AssertionError('Should never happen. Guaranteed by DB consistency')
```
yabirgb, line 186: should we just continue and avoid fatal errors?

rotkehlchen/db/upgrades/v39_v40.py
```diff
@@ -83,6 +128,114 @@ def _add_base_chain_data(write_cursor: 'DBCursor') -> None:
...
+
+            history_events.append((
+                1,  # entry_type -> HISTORY_EVENT
+                f'{MIGRATION_PREFIX}{uuid4().hex}',  # event_identifier
+                0,  # sequence_index
+                entry[0] * 1000,  # timestamp
```
yabirgb, line 192: Not using `ts_sec_to_ms` here is to avoid one extra context switch?

rotkehlchen/db/upgrades/v39_v40.py
```diff
@@ -83,6 +128,114 @@ def _add_base_chain_data(write_cursor: 'DBCursor') -> None:
...
+
+    with conn.read_ctx() as cursor:  # migrate coinbase ledger action query ranges
+        cursor.execute("SELECT name FROM used_query_ranges WHERE name LIKE '%_ledger_actions_%'")
+        for entry in cursor:
+            name = entry[0]
+            try:
+                write_cursor.execute(
+                    'UPDATE used_query_ranges SET name = ? WHERE name = ?',
+                    (name.replace('_ledger_actions_', '_history_events_'), name),
+                )
```
yabirgb, line 228: is it worth the extra variable storage just for the case of the query raising the exception?

rotkehlchen/db/upgrades/v39_v40.py
```diff
@@ -83,6 +128,114 @@ def _add_base_chain_data(write_cursor: 'DBCursor') -> None:
...
+            except sqlcipher.IntegrityError as e:  # pylint: disable=no-member
+                log.error(f'Failed to update used query range {name} due to {e}')
+                continue
+
+    # Clean up any ignored ledger actions # THINK: Migrate the ignoring if action exists and is migrated to history event or let user do it again? Probably let's keep it simple  # noqa: E501
+    write_cursor.execute('DELETE FROM ignored_actions WHERE type=?', ('D',))
+
```
yabirgb, line 235: extra white space?

rotkehlchen/db/upgrades/v39_v40.py
```diff
@@ -83,6 +128,114 @@ def _add_base_chain_data(write_cursor: 'DBCursor') -> None:
...
+            usd_value = ZERO  # try to use a rate asset if given only if it's against USD. Unfortunately this means we can lose data here but there is no way to translate this to a single history event  # noqa: E501
+            if entry[5] is not None and entry[6] is not None and entry[6] == 'USD':
+                try:
+                    amount = deserialize_fval(entry[3], 'amount', 'ledger actions migration')
+                    rate = deserialize_fval(entry[5], 'rate', 'ledger actions migration')
+                except DeserializationError as e:
+                    log.error(f'Skipped a ledger action from migration to history event due {e}')  # noqa: E501
+                    continue
+
+                usd_value = amount * rate
```
yabirgb, line 163: should we fail here or set price to 0?

rotkehlchen/exchanges/exchange.py
```diff
@@ -417,27 +417,26 @@ def query_income_loss_expense(
...
             # has_premium True is fine here since the result of this is not user facing atm
-            ledger_actions = db.get_ledger_actions(cursor, filter_query=filter_query, has_premium=True)  # noqa: E501
+            events = db.get_history_events(cursor, filter_query=filter_query, has_premium=True)
             if only_cache:
-                return ledger_actions
+                return events
 
             ranges = DBQueryRanges(self.db)
-            location_string = f'{self.location!s}_ledger_actions_{self.name}'
+            location_string = f'{self.location!s}_history_events_{self.name}'
```
yabirgb, line 439: is the `!s` needed here?

yabirgb (commented):
image

rotkehlchen/db/upgrades/v39_v40.py
LefterisJP: this will never happen. Perhaps we should log something too but this can't happen. It's like an

```if 1 ==1:
    pass
else:
    raise AssertionError
```

If that happens means DB is corrupt.

rotkehlchen/accounting/structures/types.py
LefterisJP: The PR is doing that

rotkehlchen/db/upgrades/v39_v40.py
LefterisJP: hmmm not really, just to not depend on one extra thing and since no typing is needed here

rotkehlchen/db/upgrades/v39_v40.py
LefterisJP: I don't understand the question

rotkehlchen/db/upgrades/v39_v40.py
LefterisJP: no this separates from the logging

rotkehlchen/db/upgrades/v39_v40.py
LefterisJP: what do you mean? There is no failure in the 3 lines you quoted

rotkehlchen/exchanges/exchange.py
LefterisJP: Was before. Explicitness instead of relying on language's quirks I guess.

rotkehlchen/accounting/structures/types.py
yabirgb: yes, from the schema is what I mean

rotkehlchen/db/upgrades/v39_v40.py
yabirgb: I see the point of having it and agree that should not happen but not 100% what to do here. I see why to stop the upgrade

rotkehlchen/db/upgrades/v39_v40.py
yabirgb: so you use `name = entry[0]` to reuse it in the error message but then you have one extra variable creation just for reusing it in the except log mesasge. Why not use entry[0] in both places?

rotkehlchen/db/upgrades/v39_v40.py
yabirgb: is in the `log.error(`. What I mean is that if you fail to deserialize the values for amount and rate instead of doing a continue just the the usd_value = ZERO

rotkehlchen/db/upgrades/v39_v40.py
LefterisJP: readability and size of the code I guess. This code changed a lot as I was working on the PR

rotkehlchen/db/upgrades/v39_v40.py
LefterisJP: no if the values of the entry are wrong the entry makes no sense.  How can you have an entry with a non-valid amount?

rotkehlchen/accounting/structures/types.py
LefterisJP: done

rotkehlchen/db/upgrades/v39_v40.py
LefterisJP: If it makes you happy I can remove it also

LefterisJP (commented):
Went through it quickly and left 2 comments

yabirgb (approved):
May god bless the consecuences of this PR

PR #6692, opened 2023-10-05
improving binance import

jbrit:
## Checklist

- [ ] The PR modified the backend, to improve the support for importing csvs in finance.


rotkehlchen/tests/utils/dataimport.py
```diff
@@ -1556,7 +1558,45 @@ def assert_binance_import_results(rotki: Rotkehlchen):
...
+        ),
+        Trade(
+            timestamp=Timestamp(1690570138),
+            location=Location.BINANCE,
+            base_asset=A_USDT,
+            quote_asset=A_SOL,
+            trade_type=TradeType.BUY,
+            amount=FVal('26.82490000'),
+            rate=Price(FVal('25.07')),
+            fee=FVal(-0.00008316),
```
yabirgb, line 1580: fee should be a positve amount. We always work with positve amounts

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -1711,6 +1871,145 @@ def assert_binance_import_results(rotki: Rotkehlchen):
...
+        ), HistoryEvent(
+            identifier=31,
+            event_identifier='BNC_e5621aea7c02961b0e01fadb78c9faf2bfae1df38ebe1a31e7083d8e4f5fae09',
+            sequence_index=0, timestamp=1694623421000, location=Location.BINANCE,
+            event_type=HistoryEventType.SPEND,
+            event_subtype=HistoryEventSubType.PAYBACK_DEBT,
+            asset=A_USDT,
+            balance=Balance(amount=FVal(0.03480000)),
+            location_label='CSV import',
+            notes='0.03480000 eip155:1/erc20:0xdAC17F958D2ee523a2206206994597C13D831ec7(Tether) realized loss on binance',
```
yabirgb, line 2012: This should show the symbol and not the identifier

yabirgb (commented):
Wanted to try it but unfortunatly the UI is not working. If the tests are okey I guess is fine

rotkehlchen/data_import/importers/binance.py
```diff
@@ -287,7 +340,12 @@ class BinanceDepositWithdrawEntry(BinanceSingleEntry):
     """This class processes Deposit and Withdraw actions
         which are AssetMovements in the internal representation"""
 
-    AVAILABLE_OPERATIONS: Final[tuple[str, ...]] = ('Deposit', 'Withdraw')  # type: ignore[misc]  # noqa: E501  # figure out how to mark final only in this class
+    AVAILABLE_OPERATIONS: Final[tuple[str, ...]] = (  # type: ignore[misc]  # figure out how to mark final only in this class  # noqa: E501
+        'Deposit',
+        'Buy Crypto',
```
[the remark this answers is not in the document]
jbrit: It's a buy, with regular payment methods, so no record for the payment method, since it's not crypto

rotkehlchen/data_import/importers/binance.py
```diff
@@ -434,6 +539,61 @@ def process_entry(
...
+            event_subtype=HistoryEventSubType.NONE,
+            asset=data['Coin'],
+            balance=Balance(amount=amount),
+            location_label='CSV import',
+            notes=data['Operation'],
+        )
+        if data['Operation'] in ('Fee', 'Funding Fee'):
+            history_event.event_type = HistoryEventType.SPEND
+            history_event.event_subtype = HistoryEventSubType.FEE
+            history_event.notes = f'{amount} {data["Coin"]} fee paid on binance'
```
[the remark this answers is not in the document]
jbrit: I don't understand

rotkehlchen/tests/utils/dataimport.py
jbrit: I think it's different for trades:
https://github.com/rotki/rotki/blob/03f738d44412f7429105f1edf24c5a4cc315e70b/rotkehlchen/tests/utils/dataimport.py#L1515

yabirgb: Why you mention it. Checking the logic when processed is assumed to be positive as seen here https://github.com/rotki/rotki/blob/e064d8db4898c7e7b681ea360c8d73607e528c24/rotkehlchen/exchanges/data_structures.py#L441

jbrit: If I search the current codebase for this: `Fee(FVal('-` there're valid results [here](https://github.com/search?q=repo%3Arotki%2Frotki+Fee%28FVal%28%27-&type=code).
So it's probably a bug that has been merged into develop and I should fix everything with mine?

LefterisJP: Seems to be binance import tests again?

There is no such thing as a negative fee. If there is a negative fee. It does not make sense. Something is wrong if there is one.

rotkehlchen/data_import/importers/binance.py
```diff
@@ -63,7 +63,7 @@ class BinanceSingleEntry(BinanceEntry, metaclass=abc.ABCMeta):
     """
     AVAILABLE_OPERATIONS: tuple[str, ...]
 
-    def is_entry(self, requested_operation: str) -> bool:
+    def is_entry(self, requested_operation: str, account: str, change: int) -> bool:  # pylint: disable=unused-argument
```
LefterisJP, line 66: you need to adjust the docstring to explain what `account` and `change` are used for.

rotkehlchen/data_import/importers/binance.py
```diff
@@ -98,6 +98,63 @@ def process_entries(
...
+        transfer_events: list[HistoryBaseEntry] = [
+            HistoryEvent(
+                event_identifier=event_identifier,
+                sequence_index=index,
+                timestamp=timestamp_ms,
+                location=Location.BINANCE,
+                event_type=HistoryEventType.TRANSFER,
+                event_subtype=HistoryEventSubType.NONE,
+                asset=row['Coin'],
+                balance=Balance(amount=abs(row['Change'])),
```
LefterisJP, line 133: Hmm you are performing `abs()` on a raw string/int value from the csv. This should be deserialized as an FVal properly and checked for errors. Is it? I guess it's done in `_group_binance_rows`. Same as for the `asset` in `csv_row['Coin']`.

But that is very hard to understand and error prone. Is it possible to change the csv row to a typed dict? We need to be sure of the types flying around in these functions. Otherwise I see a `row['key']` and I assume, and so will every new reader, that this is the raw string/int from the CSV.

rotkehlchen/data_import/importers/binance.py
```diff
@@ -111,20 +168,27 @@ def are_entries(self, requested_operations: list) -> bool:
             - DOESN'T support mixed Fee and Non-fee Trades
             - (Transaction Related + Transaction Related) * N
             - (Small assets exchange BNB + Small assets exchange BNB) * N
+            - (Asset Conversion Transfer + Asset Conversion Transfer) * N
+            - (Binance Convert + Binance Convert) * N
             - (Buy + Transaction Related) * N
         """
         counted = Counter(requested_operations)
         counted.pop('Fee', None)
+        counted.pop('Transaction Fee', None)
```
LefterisJP, line 177: Can you add a comment here, explaining why are both of those keys popped out?

rotkehlchen/data_import/importers/binance.py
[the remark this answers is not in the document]
LefterisJP: Can you add a comment explaining that and why it's here?

rotkehlchen/data_import/importers/binance.py
```diff
@@ -298,7 +367,7 @@ def process_entry(
     ) -> None:
         amount = data['Change']
         asset = data['Coin']
-        category = AssetMovementCategory.DEPOSIT if data['Operation'] == 'Deposit' else AssetMovementCategory.WITHDRAWAL  # noqa: E501
+        category = AssetMovementCategory.WITHDRAWAL if data['Operation'] == 'Withdraw' else AssetMovementCategory.DEPOSIT  # noqa: E501
```
LefterisJP, line 370: Why this change? Isn't it the same, but you just changed the order of the if? If the else covers more stuff then this is tricky and needs to be explained. If not undo this diff.

rotkehlchen/data_import/importers/binance.py
```diff
@@ -317,6 +386,45 @@ def process_entry(
         importer.add_asset_movement(write_cursor=write_cursor, asset_movement=asset_movement)
 
 
+class BinanceDistributionEntry(BinanceSingleEntry):
```
LefterisJP, line 389: add short 1-line docstring explaining what this class is used for

rotkehlchen/data_import/importers/binance.py
```diff
@@ -317,6 +386,45 @@ def process_entry(
...
+            timestamp: Timestamp,
+            data: BinanceCsvRow,
+    ) -> None:
+        """
+        Process distributions from binance. It includes
+        Cash Voucher Distribution and Mission Reward Distribution. May raise:
+        - KeyError
+        - DeserializationError: if the event is malformed when being stored in the db
+        """
+        asset = data['Coin']
```
LefterisJP, line 408: no need to declare a variable here. Just use it directly in the constructor below
```suggestion
```

rotkehlchen/data_import/importers/binance.py
```diff
@@ -317,6 +386,45 @@ def process_entry(
...
+        amount = abs(data['Change'])
+        event_identifier = f'{EVENT_IDENTIFIER_PREFIX}{hash_csv_row(data)}'
```
LefterisJP, line 410: Same for this and all the rest. Just add them to the constructor directly. Why create temporary variables?

rotkehlchen/data_import/importers/binance.py
```diff
@@ -434,6 +550,72 @@ def process_entry(
...
+            timestamp=timestamp_ms,
+            location=Location.BINANCE,
+            event_type=event_type,
+            event_subtype=event_subtype,
+            asset=data['Coin'],
+            balance=Balance(amount=amount),
+            location_label='CSV import',
+            notes=notes,
+        )
+        return history_event
```
LefterisJP, line 600: don't create temporary variables that are used only in one place

```suggestion
        return HistoryEvent(
            event_identifier=event_identifier,
            sequence_index=0,
            timestamp=timestamp_ms,
            location=Location.BINANCE,
            event_type=event_type,
            event_subtype=event_subtype,
            asset=data['Coin'],
            balance=Balance(amount=amount),
            location_label='CSV import',
            notes=notes,
        )
```

rotkehlchen/data_import/importers/binance.py
```diff
@@ -434,6 +550,72 @@ def process_entry(
...
+    def _get_event(self, timestamp: Timestamp, data: BinanceCsvRow) -> HistoryEvent:
+        """
+        Process USDM Program Events. It includes
+        Fee, Funding Fee, and Realized Profit and Loss. May raise:
+        - KeyError
+        """
+        event_identifier = f'{EVENT_IDENTIFIER_PREFIX}{hash_csv_row(data)}'
+        timestamp_ms = ts_sec_to_ms(timestamp)
+        amount = abs(data['Change'])
+        asset: AssetWithOracles = data['Coin']
```
LefterisJP, line 576: All these can just go directly in the constructor. No need for temporary variables
```suggestion
```

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -1641,6 +1705,102 @@ def assert_binance_import_results(rotki: Rotkehlchen):
...
+            sequence_index=0,
+            timestamp=ts_sec_to_ms(Timestamp(1640910825)),
+            location=Location.BINANCE,
+            asset=A_USDT,
+            balance=Balance(amount=FVal(500.00000000)),
+            event_type=HistoryEventType.TRANSFER,
+            event_subtype=HistoryEventSubType.NONE,
+            location_label='CSV import',
+            notes='Transfer Between Main and Funding Wallet',
+        ), HistoryEvent(
```
LefterisJP, line 1718: What is the purpose of these events?

`BNC_ac8a1a8ccd32547766f312428a59ac8ea667f72c33c0c4695b3fc82a7896645f`?

It's a "transfer" and appears twice. So this is wrong.

If it's internal transfers we:

1. Probably don't need to show them. We don't care about them.
2. We definitely should not show them duplicated.

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -1641,6 +1705,102 @@ def assert_binance_import_results(rotki: Rotkehlchen):
...
+            identifier=8,
+            event_identifier='BNC_29a2d52a91b9bbb7213d4710ad190d587c4febc197373291289d85ae275ec7a5',
```
LefterisJP, line 1732: Same for these 2 events `BNC_29a2d52a91b9bbb7213d4710ad190d587c4febc197373291289d85ae275ec7a5`

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -1641,6 +1705,102 @@ def assert_binance_import_results(rotki: Rotkehlchen):
...
+            location=Location.BINANCE,
+            asset=A_USDT,
+            balance=Balance(amount=FVal(100.00000000)),
+            event_type=HistoryEventType.TRANSFER,
+            event_subtype=HistoryEventSubType.NONE,
+            location_label='CSV import',
+            notes='Transfer Between Spot Account and UM Futures Account',
+        ), HistoryEvent(
+            identifier=10,
+            event_identifier='BNC_33a197c8698d3de4db57bb5d8a43e48addb28538587f73cd624792036b69a870',
```
LefterisJP, line 1756: Same for `BNC_33a197c8698d3de4db57bb5d8a43e48addb28538587f73cd624792036b69a870`

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -1641,6 +1705,102 @@ def assert_binance_import_results(rotki: Rotkehlchen):
...
+            location=Location.BINANCE,
+            asset=A_USDT,
+            balance=Balance(amount=FVal(60.00000000)),
+            event_type=HistoryEventType.TRANSFER,
+            event_subtype=HistoryEventSubType.NONE,
+            location_label='CSV import',
+            notes='Transfer Between Spot Account and CM Futures Account',
+        ), HistoryEvent(
+            identifier=12,
+            event_identifier='BNC_a5e6e8af4008679025ce68ac818ee4ae3f8b9f461d70eeb3d30f20f3c9a8d92f',
```
LefterisJP, line 1780: Same for `BNC_a5e6e8af4008679025ce68ac818ee4ae3f8b9f461d70eeb3d30f20f3c9a8d92f`

Also why have so many similar events in a test? If we ever need to modify the tests (which we do often due to logic changes), doing so will become a nightmare!

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -1711,6 +1871,147 @@ def assert_binance_import_results(rotki: Rotkehlchen):
...
+            sequence_index=0,
+            timestamp=ts_sec_to_ms(Timestamp(1694623421)),
+            location=Location.BINANCE,
+            event_type=HistoryEventType.SPEND,
+            event_subtype=HistoryEventSubType.PAYBACK_DEBT,
+            asset=A_USDT,
+            balance=Balance(amount=FVal('1.29642000')),
+            location_label='CSV import',
+            notes='1.29642000 USDT realized loss on binance USD-MFutures',
+        ), HistoryEvent(
```
LefterisJP, line 1943: Why did you add this even in the CSV for the test? It's same type as above. So already tested.

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -1711,6 +1871,147 @@ def assert_binance_import_results(rotki: Rotkehlchen):
...
+            sequence_index=0,
+            timestamp=ts_sec_to_ms(Timestamp(1694623421)),
+            location=Location.BINANCE,
+            event_type=HistoryEventType.SPEND,
+            event_subtype=HistoryEventSubType.FEE,
+            asset=A_USDT,
+            balance=Balance(amount=FVal(0.11955856)),
+            location_label='CSV import',
+            notes='0.11955856 USDT fee paid on binance USD-MFutures',
+        ), HistoryEvent(
```
LefterisJP, line 1955: Same for this and all other events underneath  here!!!

LefterisJP (commented):
Okay so left many comments.

Please as a general guideline keep the tests simple and only one test event per type you need. All these will need modifications next time logic is touched. Simplify it.

rotkehlchen/data_import/importers/binance.py
```diff
@@ -63,9 +72,11 @@ class BinanceSingleEntry(BinanceEntry, metaclass=abc.ABCMeta):
     """
     AVAILABLE_OPERATIONS: tuple[str, ...]
 
-    def is_entry(self, requested_operation: str) -> bool:
+    def is_entry(self, requested_operation: str, account: str, change: AssetAmount) -> bool:  # pylint: disable=unused-argument
         """This method checks whether row with "requested_operation" could be processed
-        by a class on which this method has been called
+        by a class on which this method has been called.
+        Some subclasses require combined checks with the "account" and "change" to
```
LefterisJP, line 78: nitpick: Here you should explain that account and change are keys of the CSV. Need to be explaining what arguments are.

rotkehlchen/data_import/importers/binance.py
```diff
@@ -287,7 +360,12 @@ class BinanceDepositWithdrawEntry(BinanceSingleEntry):
     """This class processes Deposit and Withdraw actions
         which are AssetMovements in the internal representation"""
 
-    AVAILABLE_OPERATIONS: Final[tuple[str, ...]] = ('Deposit', 'Withdraw')  # type: ignore[misc]  # figure out how to mark final only in this class
+    AVAILABLE_OPERATIONS: Final[tuple[str, ...]] = (  # type: ignore[misc]  # figure out how to mark final only in this class
+        'Deposit',
+        'Buy Crypto',  # it's a buy, with regular the payment methods, that deposits crypto
```
LefterisJP, line 365: Hmm this doesn't make sense. Is it a typo? What do you mean by regular payment methods? Something like, you deposit FIAT and it immediately appears as crypto? So something like this?

```suggestion
        'Buy Crypto',  # it's a direct buy where you deposit fiat and get credited with crypto
```

rotkehlchen/data_import/importers/binance.py
```diff
@@ -296,27 +374,59 @@ def process_entry(
             timestamp: Timestamp,
             data: BinanceCsvRow,
     ) -> None:
-        amount = data['Change']
         asset = data['Coin']
-        category = AssetMovementCategory.DEPOSIT if data['Operation'] == 'Deposit' else AssetMovementCategory.WITHDRAWAL  # noqa: E501
-        if category == AssetMovementCategory.WITHDRAWAL:
-            amount = -amount
-
+        category = AssetMovementCategory.WITHDRAWAL if data['Operation'] == 'Withdraw' else AssetMovementCategory.DEPOSIT  # covers 'Buy Crypto' & 'Fiat Deposit'  # noqa: E501
```
LefterisJP, line 378:
```suggestion
        category = AssetMovementCategory.WITHDRAWAL if data['Operation'] == 'Withdraw' else AssetMovementCategory.DEPOSIT  # else clause also covers 'Buy Crypto' & 'Fiat Deposit'  # noqa: E501
```

rotkehlchen/data_import/importers/binance.py
```diff
@@ -499,9 +677,15 @@ def _group_binance_rows(
                 formatstr=timestamp_format,
                 location='binance',
             )
-            csv_row['Coin'] = asset_from_binance(csv_row['Coin'])
-            csv_row['Change'] = deserialize_asset_amount(csv_row['Change'])
-            multirows[timestamp].append(csv_row)
+            multirows[timestamp].append(BinanceCsvRow(  # maintain insertion order for consistent hash  # noqa: E501
```
LefterisJP, line 680: Hmmm that's not what I had in mind when I mentioned TypedDict.

It's true you would need to do something like that to turn an existing dict into typeddict, and probably can't go around it here.

In these cases probably a simple dataclass. And I would do all typing possible here. In this function. Not let it for later.

So I would also save serialized all possible serialized attributes. Like `UTC_Time', 'Account' etc.

But .... do you need those rows? It seems you don't. For example check above. The only part where `UTC_Time` is used is in like 676 right above here. Nowhere else. Then no need to add it here. Same with `User_ID`, 'Remark'. And maybe more?

If nobody reads them do not have them in the dataclass.

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -1711,6 +1823,99 @@ def assert_binance_import_results(rotki: Rotkehlchen):
...
+            timestamp=ts_sec_to_ms(Timestamp(1694505602)),
+            location=Location.BINANCE,
+            event_type=HistoryEventType.SPEND,
+            event_subtype=HistoryEventSubType.FEE,
+            asset=A_USDT,
+            balance=Balance(amount=FVal(0.07779065)),
+            location_label='CSV import',
+            notes='0.07779065 USDT fee paid on binance USD-MFutures',
+        ), HistoryEvent(
+            identifier=21,
```
LefterisJP, line 1884: Delete this event. Event with id 20 and 21 test same thing.

rotkehlchen/data_import/importers/binance.py
jbrit: I can create the data class, I should probably hash the row and store immediately to the dataclass field `hashed`, that’s the only reason I converted the entire row

LefterisJP: No, why wuld you do that? You can just add a dataclass method to generate the event identifer which does that for you.

jbrit: I made the other fixes, when trying to move this to a dataclass, I realized it would be more messy, unless I did not understand you correctly.

- currently all the fields are accessed when we call `hash_csv_row`
- if I move it to a dataclass, i still have to define all the fields since I need them for the hash, and when serializing the dataclass, I'll be creating a dictionary matching the default out put from dataclass, just to hash it too

LefterisJP (approved):
LGTM. Thank you Ajibolah. Appreciate it. Totally forgot but this is missing changelog entry. Feel free to add it in another PR.

PR #6695, opened 2023-10-05
Custom accounting rules

yabirgb:
This PR allows to create custom rules for accounting. What it includes is:

- Rest endpoints for creating, editing, querying and deleting rules
- Change in the logic to query first db for rules and fallback in module rules and default rules
- db migration to create the new table

What needs to be done in a different PR:

- [ ] Port harcoded rules to db
- [ ] Move rules from ledger actions to this new system in the db upgrade
- [ ] frontend changes for this PR


rotkehlchen/chain/evm/accounting/aggregator.py
```diff
@@ -123,3 +128,59 @@ def reset(self) -> None:
...
+
+    def _query_db_rules(self) -> None:
+        """"
+        """
+        query = (
+            'SELECT type, subtype, counterparty, taxable, count_entire_amount_spend, '
+            'count_cost_basis_pnl, method, accounting_treatment FROM accounting_rules'
+        )
+        with self.database.conn.read_ctx() as cursor:
+            cursor.execute(query)
```
LefterisJP, line 154: why put it in a variable? Takes more space and is unused.

```suggestion
        with self.database.conn.read_ctx() as cursor:
            cursor.execute(
                'SELECT type, subtype, counterparty, taxable, count_entire_amount_spend, '
                'count_cost_basis_pnl, method, accounting_treatment FROM accounting_rules'
            )
```

rotkehlchen/chain/evm/accounting/aggregator.py
```diff
@@ -123,3 +128,59 @@ def reset(self) -> None:
...
+            for entry in cursor:
+                rule = BaseEventSettings.deserialize_from_db(entry[3:])
```
LefterisJP, line 156: perhaps we can combine this with the `get_tx_event_type_identifier` method? 

Also now that I noticed it that's quite a silly method to combine them. Why did we add two `__` between them and not one? To be able to separate them again by checking where there is a double underscore? Why just not append those strings?

rotkehlchen/chain/evm/accounting/aggregator.py
```diff
@@ -123,3 +128,59 @@ def reset(self) -> None:
...
+                key = f'{entry[0]}__{entry[1]}'
+                if entry[2] != NULL_ACCOUNTING_COUNTERPARTY:
+                    key += f'__{entry[2]}'
+
+                self.event_settings[key] = rule
+
+    def get_event_settings(self, event: HistoryBaseEntry) -> Optional[BaseEventSettings]:
+        """
+        Return a matching rule for the event if it exists
+        """
```
LefterisJP, line 166:
```suggestion
        """Return a matching rule for the event if it exists"""
```

rotkehlchen/chain/evm/accounting/aggregator.py
```diff
@@ -123,3 +128,59 @@ def reset(self) -> None:
...
+        self.event_settings = (  # Using | operator is fine since keys are unique
+            self.aggregators.get_accounting_settings(self.pot) |
+            make_default_accounting_settings(self.pot)
+        )
+        self._query_db_rules()
+
+    def clean_rules(self) -> None:
+        """
+        Remove the rules from memory. Should be done after finishing with the accounting process
+        to avoid having many dynamic objects in memory since the rules come also from the database
```
LefterisJP, line 184:
```suggestion
        to avoid having many dynamic objects in memory since the rules come from the database
```

rotkehlchen/accounting/accountant.py
```diff
@@ -212,6 +212,10 @@ def process_history(
             total_actions=actions_length,
             pnls=self.pots[0].pnls,
         )
+
+        for pot in self.pots:
```
LefterisJP, line 216: Add a small docstring explaining what this is here for

rotkehlchen/accounting/default_rules.py
```diff
@@ -0,0 +1,114 @@
...
+
+from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.chain.evm.accounting.structures import BaseEventSettings, TxAccountingTreatment
+from rotkehlchen.chain.evm.decoding.constants import CPT_GAS
+
+if TYPE_CHECKING:
+    from rotkehlchen.accounting.pot import AccountingPot
+
+
+def make_default_accounting_settings(pot: 'AccountingPot') -> dict[str, BaseEventSettings]:
```
LefterisJP, line 11: What's this file and docstring? This comes from `accounting/history_base_entries.py` and even has the docstring from before which references the issue you work on.

This should go away. These will be written in the DB we said.

I don't even see this used anywhere.

rotkehlchen/db/accounting_rules.py
```diff
@@ -0,0 +1,140 @@
...
+    def __init__(self, db_handler: 'DBHandler') -> None:
+        self.db = db_handler
+
+    @classmethod
+    def _formulate_primary_key(
+            cls,
+            event_type: HistoryEventType,
+            event_subtype: HistoryEventSubType,
+            counterparty: Optional[str],
+    ) -> tuple[str, str, str]:
```
LefterisJP, line 26: 1. Add docstring.
2. Is this really needed to be its own method? It's really realy simple logic which can just be added in the 2 places you use it. Not a dealbreaker for me ofc, but just looks weird.

rotkehlchen/db/constants.py
```diff
@@ -18,6 +18,8 @@
 
 LAST_DATA_UPDATES_KEY: Final = 'last_data_updates_ts'
 
+NULL_ACCOUNTING_COUNTERPARTY = 'NONE'
```
LefterisJP, line 21:
```suggestion
NO_ACCOUNTING_COUNTERPARTY = 'NONE'
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -0,0 +1,140 @@
...
+            counterparty if counterparty is not None else NULL_ACCOUNTING_COUNTERPARTY,
+        )
+
+    def add_accounting_rule(
+            self,
+            event_type: HistoryEventType,
+            event_subtype: HistoryEventSubType,
+            counterparty: Optional[str],
+            rule: 'BaseEventSettings',
+    ) -> None:
```
LefterisJP, line 39: docstring

rotkehlchen/db/accounting_rules.py
```diff
@@ -0,0 +1,140 @@
...
+        query = (
+            'INSERT INTO accounting_rules(type, subtype, counterparty, taxable, '
+            'count_entire_amount_spend, count_cost_basis_pnl, method, accounting_treatment) '
+            'VALUES (?, ?, ?, ?, ?, ?, ?, ?)'
+        )
+        with self.db.conn.write_ctx() as write_cursor:
+            try:
+                write_cursor.execute(
+                    query,
```
LefterisJP, line 48: same as before. Don't use a `query` variable if it's only used in the query itself once
```suggestion
        with self.db.conn.write_ctx() as write_cursor:
            try:
                write_cursor.execute(
                    'INSERT INTO accounting_rules(type, subtype, counterparty, taxable, '
                    'count_entire_amount_spend, count_cost_basis_pnl, method, accounting_treatment) '
                    'VALUES (?, ?, ?, ?, ?, ?, ?, ?)'
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -0,0 +1,140 @@
...
+                            event_type=event_type,
+                            event_subtype=event_subtype,
+                            counterparty=counterparty,
+                        ),
+                        *rule.serialize_for_db(),
+                    ),
+                )
+            except sqlcipher.IntegrityError as e:  # pylint: disable=no-member
+                raise InputError(
+                    f'Rule for ({event_type.serialize()}, {event_subtype.serialize()}, {counterparty}) already exists',  # noqa: E501
```
LefterisJP, line 60: This Rule for .... bla bla is repeated a lot. Can't you add this as a method of the rule class?

rotkehlchen/db/accounting_rules.py
```diff
@@ -0,0 +1,140 @@
...
+                    f'{counterparty}) does not exist',
+                )
+
+    def update_accounting_rule(
+            self,
+            event_type: HistoryEventType,
+            event_subtype: HistoryEventSubType,
+            counterparty: Optional[str],
+            rule: 'BaseEventSettings',
+            row_id: int,
```
LefterisJP, line 92:
```suggestion
            identifier: int,
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -0,0 +1,140 @@
...
+    ) -> None:
+        """
+        Edit accounting rule properties for the (type, subtype, counterparty) provided.
```
LefterisJP, line 95:
```suggestion
        Edit accounting rule properties (type, subtype, counterparty) for the rule with the provided identifier
```

rotkehlchen/db/schema.py
```diff
@@ -659,6 +659,20 @@
...
+CREATE TABLE IF NOT EXISTS accounting_rules(
+    type TEXT NOT NULL,
+    subtype TEXT NOT NULL,
+    counterparty TEXT NOT NULL,
+    taxable INTEGER NOT NULL CHECK (taxable IN (0, 1)),
+    count_entire_amount_spend INTEGER NOT NULL CHECK (count_entire_amount_spend IN (0, 1)),
+    count_cost_basis_pnl INTEGER NOT NULL CHECK (count_cost_basis_pnl IN (0, 1)),
+    method TEXT,
+    accounting_treatment TEXT,
+    PRIMARY KEY(type, subtype, counterparty)
```
LefterisJP, line 672: This is a bad idea. And you can see it that you are actually trying to work with rowids, so exposing implementation details into our logic. Very bad.

If you need to work with a single identifier, and you seem to want to then make the schema explicit. This is what happens underneath so long as we don't use this `NOROWID` thing.

```suggestion
    identifier INTEGER NOT NULL PRIMARY KEY,
    type TEXT NOT NULL,
    subtype TEXT NOT NULL,
    counterparty TEXT NOT NULL,
    taxable INTEGER NOT NULL CHECK (taxable IN (0, 1)),
    count_entire_amount_spend INTEGER NOT NULL CHECK (count_entire_amount_spend IN (0, 1)),
    count_cost_basis_pnl INTEGER NOT NULL CHECK (count_cost_basis_pnl IN (0, 1)),
    method TEXT,
    accounting_treatment TEXT,
    UNIQUE(type, subtype, counterparty)
```

This is also suggested by their docs: https://www.sqlite.org/rowidtable.html

rotkehlchen/db/accounting_rules.py
```diff
@@ -0,0 +1,140 @@
...
+        May raise:
+        - InputError: if not event gets updated
+        """
+        query = (
```
LefterisJP, line 99: same thing for this query. Just move it in the execute please. Use identifier

rotkehlchen/db/accounting_rules.py
```diff
@@ -0,0 +1,140 @@
...
+            if write_cursor.rowcount != 1:
+                raise InputError(
+                    f'Tried to update accounting rule for ({event_type.serialize()}, '
+                    f'{event_subtype.serialize()}, {counterparty}) but it was not found',
+                )
+
+    def query_rules(
+            self,
+            filter_query: AccountingRulesFilterQuery,
+    ) -> tuple[
```
LefterisJP, line 124: docstring

rotkehlchen/api/rest.py
```diff
@@ -4435,3 +4442,86 @@ def export_history_events(
...
+        return api_response(_wrap_in_ok_result(True), status_code=HTTPStatus.OK)
+
+    def query_accounting_rules(self, filter_query: AccountingRulesFilterQuery) -> Response:
+        db = DBAccountingRules(self.rotkehlchen.data.db)
+        entries, total_filter_count = db.query_rules(filter_query=filter_query)
+        entries_result = []
+        for entry in entries:
+            # serialize the rule and add information about the key to what it applies
+            data = entry[1].serialize()
+            data['rowid'] = entry[0][0]
```
LefterisJP, line 4513:
```suggestion
            data['identifier'] = entry[0][0]
```

rotkehlchen/api/rest.py
```diff
@@ -4435,3 +4442,86 @@ def export_history_events(
...
+            data['event_type'] = entry[0][1]
+            data['event_subtype'] = entry[0][2]
+            data['counterparty'] = entry[0][3] if entry[0][3] != NULL_ACCOUNTING_COUNTERPARTY else None  # noqa: E501
```
LefterisJP, line 4516: hmmm why does this for loop serialization happen here? Can't you have it as part of this query_rules thing or somehow a different method?

rotkehlchen/api/rest.py
```diff
@@ -4435,3 +4442,86 @@ def export_history_events(
...
+            entries_result.append(data)
+
+        with db.db.conn.read_ctx() as cursor:
+            result = {
+                'entries': entries_result,
+                'entries_found': total_filter_count,
+                'entries_total': self.rotkehlchen.data.db.get_entries_count(cursor=cursor, entries_table='accounting_rules'),  # noqa: E501
+                'entries_limit': -1,  # no limit in this endpoint for free users
```
LefterisJP, line 4524:
```suggestion
                'entries_limit': -1,
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -2936,3 +2950,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+    def prepare_rule_info(
+            self,
+            data: dict[str, Any],
+            **kwargs: Any,
+    ) -> dict[str, Any]:
+        return self._create_settings(data, **kwargs)
+
+
+class EditAccountingRuleSchema(CreateAccountingRuleSchema):
+    rowid = fields.Integer(required=True)
```
LefterisJP, line 3000:
```suggestion
    identifier = fields.Integer(required=True)
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -2936,3 +2950,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+
+    @post_load
+    def prepare_rule_info(
+            self,
+            data: dict[str, Any],
+            **kwargs: Any,
+    ) -> dict[str, Any]:
+        return {'rowid': data['rowid'], **self._create_settings(data, **kwargs)}
```
LefterisJP, line 3008:
```suggestion
        return {'identifier': data['identifier'], **self._create_settings(data, **kwargs)}
```

rotkehlchen/api/v1/resources.py
```diff
@@ -2922,3 +2932,64 @@ def post(self, filter_query: 'HistoryBaseEntryFilterQuery', directory_path: Path
...
+            counterparty: Optional[str],
+            rule: 'BaseEventSettings',
+            rowid: int,
+    ) -> Response:
+        return self.rest_api.update_accounting_rule(
+            event_type=event_type,
+            event_subtype=event_subtype,
+            counterparty=counterparty,
+            rule=rule,
+            row_id=rowid,
```
LefterisJP, line 2975:
```suggestion
            identifier: int,
    ) -> Response:
        return self.rest_api.update_accounting_rule(
            event_type=event_type,
            event_subtype=event_subtype,
            counterparty=counterparty,
            rule=rule,
            identifier=identifier,
```

rotkehlchen/db/filtering.py
```diff
@@ -1710,3 +1710,53 @@ def make(
 
         filter_query.filters = filters
         return filter_query
+
+
+class AccountingRulesFilterQuery(DBFilterQuery):
```
LefterisJP, line 1715: Needs a docstring

rotkehlchen/db/filtering.py
```diff
@@ -1710,3 +1710,53 @@ def make(
...
+            and_op: bool = True,
+            order_by_rules: Optional[list[tuple[str, bool]]] = None,
+            limit: Optional[int] = None,
+            offset: Optional[int] = None,
+            event_types: Optional[list[HistoryEventType]] = None,
+            event_subtypes: Optional[list[HistoryEventSubType]] = None,
+            counterparties: Optional[list[str]] = None,
+    ) -> 'AccountingRulesFilterQuery':
+        if order_by_rules is None:
+            order_by_rules = [('rowid', False)]
```
LefterisJP, line 1729:
```suggestion
            order_by_rules = [('identifier', False)]
```

rotkehlchen/tests/api/test_accounting_rules.py
```diff
@@ -0,0 +1,132 @@
...
+    api_url_for,
+    assert_proper_response_with_result,
+    assert_simple_ok_response,
+)
+
+
+def test_manage_rules(rotkehlchen_api_server):
+    """"
+    Test basic operations in the endpoint for managing accounting rules
+    """
```
LefterisJP, line 14: If it fits keep it in one line

```suggestion
    """"Test basic operations in the endpoint for managing accounting rules"""
```

rotkehlchen/tests/db/test_db_accounting_rules.py
```diff
@@ -0,0 +1,128 @@
...
+    # first try to store the rule in the db
+    db.add_accounting_rule(
+        event_type=HistoryEventType.TRADE,
+        event_subtype=HistoryEventSubType.RECEIVE,
+        counterparty='uniswap',
+        rule=rule,
+    )
+    # add another rule without counterparty to test that we respect the primary key
+    # and the edit/delete
+    # first try to store the rule in the db
```
LefterisJP, line 31: Seems like copy pasta

```suggestion
```

rotkehlchen/tests/db/test_db_accounting_rules.py
```diff
@@ -0,0 +1,128 @@
...
+    )
+    with database.conn.read_ctx() as cursor:
+        entries = cursor.execute(query_all_rules).fetchall()
+        assert len(entries) == 0
+
+
+def test_errors_with_rules(database: DBHandler) -> None:
+    """
+    Test different situations that can lead to erros when managing accounting rules
+    """
```
LefterisJP, line 89: Typo + one line

```suggestion
    """Test different situations that can lead to errors when managing accounting rules"""
```

rotkehlchen/tests/unit/accounting/evm/test_balancer.py
```diff
@@ -138,12 +138,12 @@ def test_balancer_multiple_deposit(accountant: 'Accountant', ethereum_inquirer:
...
     events_accountant = pot.events_accountant
-    curve_accountant = Balancerv1Accountant(
+    balancer_accountant = Balancerv1Accountant(
         node_inquirer=ethereum_inquirer,
         msg_aggregator=ethereum_inquirer.database.msg_aggregator,
     )
-    curve_settings = curve_accountant.event_settings(pot)
-    events_accountant.event_settings.update(curve_settings)
+    curve_settings = balancer_accountant.event_settings(pot)
+    events_accountant.accounting_ruler.event_settings.update(curve_settings)
```
LefterisJP, line 146: you bothered to touch it so fix all wrong names

```suggestion
    balancer_settings = balancer_accountant.event_settings(pot)
    events_accountant.accounting_ruler.event_settings.update(balancer_settings)
```

LefterisJP (commented):
A lot of comments on my first pass.

**Important:** Linting failed. So no tests have ran

One thing that I did not see is moving all the default/built-in rules in the DB. Also ... how should we handle the per module rules? They should also be written to the DB no so they can be:

1. Seen by the users so they know what's up
2. Be editable by the users

We probably need a very different way than those "accountants" per decoder. They are not well thought anyway and are full copy pasta.

```python
class SushiswapAccountant(ModuleAccountantInterface):

    def event_settings(self, pot: 'AccountingPot') -> dict[str, 'TxEventSettings']:
        """Being defined at function call time is fine since this function is called only once"""
        ````

File: rotkehlchen/api/rest.py (hunk 18 above)
Comment (yabirgb) [reply rotkehlchen/api/rest.py:4516]: Made it here since the method was more a db operation only but I've added a new method that includes serialization

File: rotkehlchen/chain/evm/accounting/aggregator.py (hunk 2 above)
Comment (yabirgb) [reply rotkehlchen/chain/evm/accounting/aggregator.py:156]: Yes, I believe is helpful for splitting them if needed. I've changed it to a single underscore and unified its usage

File: docs/api.rst [hunk 29]
```diff
@@ -12378,3 +12378,202 @@ Dealing with skipped external events
...
+         "counterparty": "compound",
+         "accounting_treatment":"swap"
+      }
+
+  .. _accounting_rules_fields:
+
+  :reqjsonarr boolean taxable: ``true`` if the event should be considered as taxable
+  :reqjsonarr boolean count_entire_amount_spend: if ``true`` then the entire amount is counted as a spend. Which means an expense (negative pnl).
+  :reqjsonarr boolean count_cost_basis_pnl: if ``true`` then we also count any profit/loss the asset may have had compared to when it was acquired.
+  :reqjsonarr string method: Can be either ``spend`` or ``acquistion``.
```
LefterisJP, line 12473: Important to not have typo here else the frontend will use wrong string :rofl: 

```suggestion
  :reqjsonarr string method: Can be either ``spend`` or ``acquisition``.
```

rotkehlchen/accounting/rules.py
```diff
@@ -0,0 +1,170 @@
...
+    from rotkehlchen.chain.evm.accounting.aggregator import EVMAccountingAggregators
+    from rotkehlchen.db.dbhandler import DBHandler
+
+
+def get_accounting_rule_identifier(
+        event_type: HistoryEventType,
+        event_subtype: HistoryEventSubType,
+        counterparty: Optional[str] = None,
+) -> str:
+    key = f'{event_type.serialize()}_{event_subtype.serialize()}'
```
LefterisJP, line 21: Just as we discussed in our call just concatenate them fren. There is no reason to keep the `_`. The reason for the `__` was to maybe at some point get the type/subtype/counterparty back.

But since that's not needed let's make it an one way function to save space also. So not only contatenate but also **hash**. This will not make it human readable anymore but does it matter?

If not, and I think it does not, then we will end up having much smaller keys.

rotkehlchen/accounting/rules.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def make_default_accounting_settings(pot: 'AccountingPot') -> dict[str, BaseEventSettings]:
+    """
+    Returns accounting settings for events that can come from various decoders and thus don't have
+    any particular protocol. These settings also allow users to customize events in the UI.
+    TODO: Remove this hardcoded rules and add them to the database along the evm modules rules.
+    """
+    result = {}
+    result[get_accounting_rule_identifier(HistoryEventType.SPEND, HistoryEventSubType.FEE, CPT_GAS)] = BaseEventSettings(  # noqa: E501
```
LefterisJP, line 35: not checking validity of those. I suppose you copied them from before, right?

rotkehlchen/accounting/rules.py
```diff
@@ -0,0 +1,170 @@
...
+    result[get_accounting_rule_identifier(HistoryEventType.STAKING, HistoryEventSubType.REMOVE_ASSET)] = BaseEventSettings(  # noqa: E501
+        taxable=False,
+        count_entire_amount_spend=False,
+        count_cost_basis_pnl=False,
+        method='acquisition',
+    )
+    return result
+
+
+class AccountingRulesManager:
```
LefterisJP, line 120: small docstring explaining what the class does.

rotkehlchen/accounting/rules.py
```diff
@@ -0,0 +1,170 @@
...
+            cursor.execute(
+                'SELECT type, subtype, counterparty, taxable, count_entire_amount_spend, '
+                'count_cost_basis_pnl, method, accounting_treatment FROM accounting_rules',
+            )
+            for entry in cursor:
+                rule = BaseEventSettings.deserialize_from_db(entry[3:])
+                key = get_accounting_rule_identifier(
+                    event_type=HistoryEventType.deserialize(entry[0]),
+                    event_subtype=HistoryEventSubType.deserialize(entry[1]),
+                    counterparty=entry[2] if entry[2] != NO_ACCOUNTING_COUNTERPARTY else None,
```
LefterisJP, line 145: Question.

Is this the only place where the logic `entry[2] if entry[2] != NO_ACCOUNTING_COUNTERPARTY else None` is to be used?

If so then it's good to leave it as is. And by checking quickly with Ctrl-f it seems to be.

IF not, then we would need to move it in `get_accounting_rule_identifier` itself.

rotkehlchen/accounting/structures/base.py
```diff
@@ -197,7 +197,7 @@ def entry_type(self) -> HistoryBaseEntryType:
 
     def get_type_identifier(self, **kwargs: Any) -> str:  # pylint: disable=unused-argument
         """Get the type identifier for this event. Subclasses may accept additional arguments"""
-        return str(self.event_type) + '__' + str(self.event_subtype)
+        return str(self.event_type) + '_' + str(self.event_subtype)
```
LefterisJP, line 200: here you should use the function `get_accounting_rule_identifier` you just made above.

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -56,8 +56,16 @@ class EvmProduct(SerializableEnumNameMixin):
     GAUGE = auto()
 
 
-def get_tx_event_type_identifier(event_type: HistoryEventType, event_subtype: HistoryEventSubType, counterparty: str) -> str:  # noqa: E501
-    return str(event_type) + '__' + str(event_subtype) + '__' + counterparty
+def get_accounting_rule_identifier(
```
LefterisJP, line 59: Hmm why is this here? :sweat: 

You have already defined this in `accounting/rules.py`

rotkehlchen/accounting/rules.py
```diff
@@ -0,0 +1,170 @@
...
+from rotkehlchen.chain.evm.decoding.constants import CPT_GAS
+from rotkehlchen.db.constants import NO_ACCOUNTING_COUNTERPARTY
+
+if TYPE_CHECKING:
+    from rotkehlchen.accounting.pot import AccountingPot
+    from rotkehlchen.chain.evm.accounting.aggregator import EVMAccountingAggregators
+    from rotkehlchen.db.dbhandler import DBHandler
+
+
+def get_accounting_rule_identifier(
```
LefterisJP, line 16: Also I know this will make me hate you for changing all lines again but the name is not what we need. Before it was better. This doesn't need to be an accounting specific, or any rule related identifier. This identifies the event type based on type/subtype/cpt. What we do with this identifier depends on the different modules.

You just happen to care about accounting right now.

```suggestion
def get_event_type_identifier(
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -2936,3 +2950,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+            **_kwargs: Any,
+    ) -> dict[str, Any]:
+        rule = BaseEventSettings(
+            taxable=data['taxable'],
+            count_entire_amount_spend=data['count_entire_amount_spend'],
+            count_cost_basis_pnl=data['count_cost_basis_pnl'],
+            method=data['method'],
+            accounting_treatment=data['accounting_treatment'],
+        )
+
```
LefterisJP, line 2982:
```suggestion
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -2936,3 +2950,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+                default_order_by_fields=['identifier', 'type', 'subtype', 'counterparty'],
+                default_ascending=[False, True, True, True],
+            ),
+            limit=data['limit'],
+            offset=data['offset'],
+            event_types=data['event_types'],
+            event_subtypes=data['event_subtypes'],
+            counterparties=data['counterparties'],
+        )
+
```
LefterisJP, line 3035:
```suggestion
```

LefterisJP (commented):
LEft another review

rotkehlchen/accounting/rules.py
yabirgb: yes copied from before

rotkehlchen/accounting/structures/base.py
```diff
@@ -470,3 +473,15 @@ def serialize(self) -> dict[str, Any]:
 
         balance = abs(self.balance).serialize()
         return {**data, **balance}
+
+
+def get_event_type_identifier(
```
LefterisJP, line 478: you did not do what I asked here. I said hash them too: https://github.com/rotki/rotki/pull/6695#discussion_r1350443642

Since we are no longer interested in this being two-way, by hashing we save a lot of space. This turns into an int.

LefterisJP (approved):
just 1 comment left. Is not minor so will leave it to you to decide if you do it in this or other PR.

Issue #6427, opened 2023-07-27
Refactoring accounting and PnL to happen in History view


LefterisJP:
## Abstract

We have discussed multiple times that the current PnL report and the accounting code is hard to maintain and a lot of times collides with the history events.

We end up having all history events as one thing, that truly represents what the user/s did but then we create the AccountingEvents out of all the different types of events and process them. And this has ended up in work duplication and also on complicating the codebase.

We need to fix this

The end result we should aim for is to do all PnL + Accounting from the history events themselves and perhaps as an overlay of the history view.

And it should work with the filters of the history view. This way we can get both full-scale PnL report for a year, but also filtered down specific queries, which right now require the user to run an entire PnL report.

Example queries.

- For myaddress.eth, and January 2022, return how much I paid in gas
- For all polygon addresses that used uniswap, how much profit(or loss) did I make
- How much have I spent in ENS names

## Specification

- First we need everything to become history events. Which means:
    * https://github.com/rotki/rotki/issues/6099
    * https://github.com/rotki/rotki/issues/6097
    * https://github.com/rotki/rotki/issues/6096
    * https://github.com/rotki/rotki/issues/7007
- Related issue which should also be addressed by this: https://github.com/rotki/rotki/issues/5818
- Then we need to ... spec it out (TODO)

LefterisJP:
So after some time I think it's finally time to address this and I will try to take point on this. Let's use this as the main issue. We had a discussion with @yabirgb and @kelsos on this. The outcome is the following.

## Goal

- Be able to see accounting information overlayed on each history event. Cost basis, taxable/not, how much of the asset you are supposed to have at the time via events processing
- Have a way to calculate those (via a button in the history view). The frontend should also tell you there is pending changes and that accounting needs to be recalculated.
- The PnL view should become "Accounting view". There you will be able to run filtered accounting queries, ranging from a PnL for a year to more specific ones like tell me how much gas I spent in arbitrum on 0xfoo address. It should be a way to operate on and query history events with the accounting information on top.

## Spec

- Get rid of the accounting event abstractions (ProcessedAccountingEvent) as it's not needed anymore now that we have everything into history events
- Add a new table for accounting information for history events. It should be an 1-1 relation to history events. It should contain:
    * history event identifier (FK link)
    * amount of asset so far (before the event)
    * cost basis of asset at the point (before the event)
    * is the event taxable?  -- or maybe this is from the accounting rules and we should not duplicate the information here?
[...]

PR #6697, opened 2023-10-05
Fix Aave-v2 accounting

LefterisJP:
## [Aave v2 accounting should now work correctly again](https://github.com/rotki/rotki/commit/41d6fe61250286056d0f0f87b7254c825c50a51d)

Fixed some bugs and wrote some much needed tests for the accounting
part of aave v2

## [Enforce proper event order (out first) for aave v2 withdrawal](https://github.com/rotki/rotki/commit/54bb70d154085ef5159adfe85d3c22814fd3311a)

Noticed the test for aave v2 withdrawal was spitting out the withdrawal (in event) first which was wrong and fixed it

----

We need to be writing tests for these type of Decoding Accountants. Otherwise it's so easy to make such silly errors as these and not have them noticed for a long time.

On the other hand the tests as you can see from this PR need a bit more tooling around them and to abstract some setup in fixtures ... and due to lack of some fields being in the data class init .. a bit of ugly code in testing.

The worse thing is that with all our plans to revamp accounting (https://github.com/rotki/rotki/issues/6427) this may also change a lot so not much sense to invest a lot of time into it.

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/aave/v2/accountant.py
+++ b/rotkehlchen/chain/ethereum/modules/aave/v2/accountant.py
@@ -82,9 +82,9 @@ class Aavev2Accountant(ModuleAccountantInterface):
         key = (string_to_evm_address(event.location_label), event.asset)  # type: ignore[arg-type]  # location_label can't be None here  # noqa: E501
         self.assets_supplied[key] -= event.balance.amount
         if self.assets_supplied[key] < ZERO:
-            gain = -1 * self.assets_borrowed[key]
+            gain = -1 * self.assets_supplied[key]
             resolved_asset = event.asset.resolve_to_asset_with_symbol()
-            pot.add_spend(
+            pot.add_acquisition(
                 event_type=AccountingEventType.TRANSACTION_EVENT,
                 notes=f'Gained {gain} {resolved_asset.symbol} on Aave v2 as interest rate for {event.location_label}',  # noqa: E501
                 location=event.location,
@@ -94,7 +94,7 @@ class Aavev2Accountant(ModuleAccountantInterface):
                 taxable=True,
                 extra_data={'tx_hash': event.tx_hash.hex()},
             )
-            self.assets_borrowed[key] = ZERO
+            self.assets_supplied[key] = ZERO
 
     def event_settings(self, pot: 'AccountingPot') -> dict[str, TxEventSettings]:  # pylint: disable=unused-argument  # noqa: E501
         """Being defined at function call time is fine since this function is called only once"""
--- a/rotkehlchen/tests/unit/accounting/evm/test_aave.py
+++ b/rotkehlchen/tests/unit/accounting/evm/test_aave.py
@@ -1,0 +1,262 @@
+from typing import TYPE_CHECKING
+
+import pytest
+
+from rotkehlchen.accounting.cost_basis.base import (
+    AssetAcquisitionEvent,
+    CostBasisInfo,
+    MatchedAcquisition,
+)
+from rotkehlchen.accounting.mixins.event import AccountingEventType
+from rotkehlchen.accounting.pnl import PNL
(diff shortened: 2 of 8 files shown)


rotkehlchen/chain/ethereum/modules/aave/v2/accountant.py
```diff
@@ -82,9 +82,9 @@ def _process_withdraw(
         key = (string_to_evm_address(event.location_label), event.asset)  # type: ignore[arg-type]  # location_label can't be None here  # noqa: E501
         self.assets_supplied[key] -= event.balance.amount
         if self.assets_supplied[key] < ZERO:
-            gain = -1 * self.assets_borrowed[key]
+            gain = -1 * self.assets_supplied[key]
```
LefterisJP, line 85: this is bug 1

rotkehlchen/chain/ethereum/modules/aave/v2/accountant.py
```diff
@@ -82,9 +82,9 @@ def _process_withdraw(
...
             resolved_asset = event.asset.resolve_to_asset_with_symbol()
-            pot.add_spend(
+            pot.add_acquisition(
```
LefterisJP, line 87: this is bug 2

rotkehlchen/chain/ethereum/modules/aave/v2/accountant.py
```diff
@@ -94,7 +94,7 @@ def _process_withdraw(
                 taxable=True,
                 extra_data={'tx_hash': event.tx_hash.hex()},
             )
-            self.assets_borrowed[key] = ZERO
+            self.assets_supplied[key] = ZERO
```
LefterisJP, line 97: and this is bug 3.

All 3 known as copy pasta without bothering to test the code.

rotkehlchen/tests/unit/accounting/evm/test_aave.py
```diff
@@ -0,0 +1,296 @@
...
+            TS1: Price(ONE),
+            TS2: Price(ONE),
+        },
+    }, A_REN.identifier: {
+        'EUR': {
+            TS1: Price(ONE),
+            TS2: Price(ONE),
+        },
+    },
+}
```
yabirgb, line 56: if the price is always one why not use 

```
@pytest.mark.parametrize('default_mock_price_value', [ONE])
```

rotkehlchen/tests/unit/accounting/evm/test_aave.py
```diff
@@ -0,0 +1,296 @@
...
+
+WITHDRAW_TEST_ENTRIES = [
+    EvmEvent(
```
yabirgb, line 59: This and the payback list is created only once. Also this variable lives as long as the test suite is running... why not define them inside the tests?

rotkehlchen/tests/unit/accounting/evm/test_aave.py
LefterisJP: Hmm ehm ... I started trying to give out a bit more complicated prices but ended up keeping it simple. Your suggestion is probably simpler

rotkehlchen/tests/unit/accounting/evm/test_aave.py
LefterisJP: Thought I would reuse them across two tests but ended up making 2 diferent ones. can definitely do that.

rotkehlchen/tests/unit/accounting/evm/test_aave.py
LefterisJP: done

rotkehlchen/tests/unit/accounting/evm/test_aave.py
LefterisJP: also done

LefterisJP:
Thank you for your review @yabirgb. Addressed it and now all is green. Will merge.

PR #6700, opened 2023-10-06
create/modify all history events

jbrit:
## Checklist

- [ ] The PR modified the backend to support creation and modification of any history_event

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -4851,7 +4851,7 @@ Dealing with History Events
          :reqjson bool is_exit: true if the "eth withdrawal event" is an exit event.
 
    :reqjson string entry_type: The type of the event that will be processed. Different validation is used based on the value for this field. Possible values are: ``"history event"``, ``"evm event"``, ``"eth withdrawal event"``, ``"eth block event"``, ``"eth deposit event"``.
-   :reqjson int timestamp: The timestamp of the entry
+   :reqjson int timestamp: The timestamp of the entry **in milliseconds**.
    :reqjson object balance: The amount/usd value of the event. If not known usd_value can also be "0".
 
    **Example Response**:
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -4689,41 +4689,170 @@ Dealing with History Events
 
 .. http:put:: /api/(version)/history/events
 
-   Doing a PUT on this endpoint can add a new evm event to rotki. For each entry evm chain also has to be specified. The unique identifier for the entry is returned as success.
+   .. _add_event_args_label:
 
-   **Example Request**:
+   Doing a PUT on this endpoint can add a new event to rotki. For each entry type, the specified arguments are different. The unique identifier for the entry is returned as success.
+   
+   .. tab:: History Event
 
-   .. http:example:: curl wget httpie python-requests
+      **Example Request**:
 
-      PUT /api/1/history/events HTTP/1.1
-      Host: localhost:5042
-      Content-Type: application/json;charset=UTF-8
+      .. http:put:: /api/(version)/history/events
 
-      {
-          "tx_hash": "0x64f1982504ab714037467fdd45d3ecf5a6356361403fc97dd325101d8c038c4e",
-          "sequence_index": 162,
-          "timestamp": 1569924574,
-          "location": "ethereum",
-          "event_type": "informational",
-          "asset": "eip155:1/erc20:0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359",
-          "balance": {"amount": "1.542", "usd_value": "1.675"},
-          "location_label": "0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12",
-          "notes": "Approve 1 SAI of 0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12 for spending by 0xdf869FAD6dB91f437B59F1EdEFab319493D4C4cE",
-          "event_subtype": "approve",
-          "counterparty": "0xdf869FAD6dB91f437B59F1EdEFab319493D4C4cE"
-      }
+         .. http:example:: curl wget httpie python-requests
 
-   :reqjson string tx_hash: This is the transaction hash of the evm event
-   :reqjson int sequence_index: This is an index that tries to provide the order of history entries for a single event_identifier.
+            PUT /api/1/history/events HTTP/1.1
+            Host: localhost:5042
+            Content-Type: application/json;charset=UTF-8
+
+            {
+               "entry_type": "history event",
+               "event_identifier": "RE_xxxxxxxxxx",
+               "location": "ethereum",
+               "timestamp": 1569924574,
+               "balance": {"amount": "1.542", "usd_value": "1.675"},
+               "sequence_index": 162,
+               "event_type": "informational",
+               "event_subtype": "approve",
+               "asset": "eip155:1/erc20:0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359",
+               "location_label": "0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12",
+               "notes": "Approve 1 SAI of 0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12 for spending by 0xdf869FAD6dB91f437B59F1EdEFab319493D4C4cE"
+            }
+
+         :reqjson int sequence_index: This is an index that tries to provide the order of history entries for a single event_identifier.
+         :reqjson string location: The location of the entry. Such as "ethereum", "optimism", etc.
+         :reqjson string asset: The asset identifier for this entry
+         :reqjson string event_identifier: The event identifier to be used for the event.
+         :reqjson string event_type: The main event type of the entry. Possible event types can be seen in the `HistoryEventType enum <https://github.com/rotki/rotki/blob/59aa288dacd1776e62682e711a916f32a14c04c2/rotkehlchen/accounting/structures/types.py#L54>`_.
+         :reqjson string event_subtype: The subtype for the entry. Possible event types can be seen in the `HistoryEventSubType enum <https://github.com/rotki/rotki/blob/59aa288dacd1776e62682e711a916f32a14c04c2/rotkehlchen/accounting/structures/types.py#L72>`_.
+         :reqjson string[optional] location_label: location_label is a string field that allows to provide more information about the location. For example when we use this structure in blockchains can be used to specify the source address.
+         :reqjson string[optional] notes: This is a description of the event entry in plain text explaining what is being done. This is supposed to be shown to the user.
+
+   .. tab:: Evm Event
+
+      **Example Request**:
+
+      .. http:put:: /api/(version)/history/events
+
+         .. http:example:: curl wget httpie python-requests
+
+            PUT /api/1/history/events HTTP/1.1
+            Host: localhost:5042
+            Content-Type: application/json;charset=UTF-8
+
+            {
+               "entry_type": "evm event",
+               "tx_hash": "0x64f1982504ab714037467fdd45d3ecf5a6356361403fc97dd325101d8c038c4e",
+               "sequence_index": 162,
+               "timestamp": 1569924574,
+               "location": "ethereum",
+               "event_type": "informational",
+               "asset": "eip155:1/erc20:0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359",
+               "balance": {"amount": "1.542", "usd_value": "1.675"},
+               "location_label": "0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12",
+               "notes": "Approve 1 SAI of 0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12 for spending by 0xdf869FAD6dB91f437B59F1EdEFab319493D4C4cE",
+               "event_subtype": "approve",
+               "counterparty": "0xdf869FAD6dB91f437B59F1EdEFab319493D4C4cE",
+               "extra_data": {}
+            }
+
+         :reqjson string tx_hash: This is the transaction hash of the evm event
+         :reqjson int sequence_index: This is an index that tries to provide the order of history entries for a single event_identifier.
+         :reqjson string location: The location of the entry. Such as "ethereum", "optimism", etc.
+         :reqjson string asset: The asset identifier for this entry
+         :reqjson string event_type: The main event type of the entry. Possible event types can be seen in the `HistoryEventType enum <https://github.com/rotki/rotki/blob/59aa288dacd1776e62682e711a916f32a14c04c2/rotkehlchen/accounting/structures/types.py#L54>`_.
+         :reqjson string event_subtype: The subtype for the entry. Possible event types can be seen in the `HistoryEventSubType enum <https://github.com/rotki/rotki/blob/59aa288dacd1776e62682e711a916f32a14c04c2/rotkehlchen/accounting/structures/types.py#L72>`_.
+         :reqjson string[optional] location_label: location_label is a string field that allows to provide more information about the location. For example when we use this structure in blockchains can be used to specify the source address.
+         :reqjson string[optional] notes: This is a description of the event entry in plain text explaining what is being done. This is supposed to be shown to the user.
+         :reqjson string[optional] counterparty: An identifier for a potential counterparty of the event entry. For a send it's the target. For a receive it's the sender. For bridged transfer it's the bridge's network identifier. For a protocol interaction it's the protocol.
+         :reqjson object[optional] extra_data: An object containing any other data to be stored.
+
+   .. tab:: Eth Block Event
+
+      **Example Request**:
+
+      .. http:put:: /api/(version)/history/events
+
+         .. http:example:: curl wget httpie python-requests
+
+            PUT /api/1/history/events HTTP/1.1
+            Host: localhost:5042
+            Content-Type: application/json;charset=UTF-8
+
+            {
+               "entry_type": "eth block event",
+               "timestamp": 1569924574,
+               "balance": {"amount": "1.542", "usd_value": "1.675"},
+               "block_number": 11,
+               "validator_index": 1,
+               "fee_recipient": "0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12",
+               "is_mev_reward": true
+            }
+
+         :reqjson int block_number: This is the number of the block where the event took place.
+         :reqjson int validator_index: This is the index of the validator.
+         :reqjson string fee_recipient: an evm address field to specify the fee recipient in an "eth block event".
+         :reqjson bool is_mev_reward: true if the "eth block event" is an mev reward event.
+
+   .. tab:: Eth Deposit Event
+
+      **Example Request**:
+
+      .. http:put:: /api/(version)/history/events
+
+         .. http:example:: curl wget httpie python-requests
+
+            PUT /api/1/history/events HTTP/1.1
+            Host: localhost:5042
+            Content-Type: application/json;charset=UTF-8
+
+            {
+               "entry_type": "eth deposit event",
+               "timestamp": 1569924574,
+               "balance": {"amount": "1.542", "usd_value": "1.675"},
+               "tx_hash": "0x64f1982504ab714037467fdd45d3ecf5a6356361403fc97dd325101d8c038c4e",
+               "event_identifier": "RE_xxxxxxxxxx",
+               "validator_index": 1,
+               "sequence_index": 162,
+               "depositor": "0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12",
+               "extra_data": {}
+            }
+
+         :reqjson string tx_hash: This is the transaction hash of the evm event
+         :reqjson int sequence_index: This is an index that tries to provide the order of history entries for a single event_identifier.
+         :reqjson int validator_index: This is the index of the validator.
+         :reqjson string event_identifier: The event identifier to be used for the event.
+         :reqjson string depositor: an evm address field to specify the depositor in an "eth deposit event".
+         :reqjson object[optional] extra_data: An object containing any other data to be stored.
+
+   .. tab:: Eth Withdrawal Event
+
+      **Example Request**:
+
+      .. http:put:: /api/(version)/history/events
+
+         .. http:example:: curl wget httpie python-requests
+
+            PUT /api/1/history/events HTTP/1.1
+            Host: localhost:5042
+            Content-Type: application/json;charset=UTF-8
+
+            {
+               "entry_type": "eth withdrawal event",
+               "timestamp": 1569924574,
+               "balance": {"amount": "1.542", "usd_value": "1.675"},
+               "is_exit": true,
+               "validator_index": 1,
+               "withdrawal_address": "0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12"
+            }
+
+         :reqjson int validator_index: This is the index of the validator.
+         :reqjson string withdrawal_address: an evm address field to specify the withdrawer in an "eth withdrawal event".
+         :reqjson bool is_exit: true if the "eth withdrawal event" is an exit event.
+
+   :reqjson string entry_type: The type of the event that will be processed. Different validation is used based on the value for this field. Possible values are: ``"history event"``, ``"evm event"``, ``"eth withdrawal event"``, ``"eth block event"``, ``"eth deposit event"``.
    :reqjson int timestamp: The timestamp of the entry
-   :reqjson string location: The location of the entry. Such as "ethereum", "optimism", etc.
-   :reqjson string event_type: The main event type of the entry. Possible event types can be seen in HistoryEventType enum.
-   :reqjson string asset: The asset identifier for this entry
    :reqjson object balance: The amount/usd value of the event. If not known usd_value can also be "0".
-   :reqjson string location_label: location_label is a string field that allows to provide more information about the location. For example when we use this structure in blockchains can be used to specify the source address.
-   :reqjson string notes: This is a description of the event entry in plain text explaining what is being done. This is supposed to be shown to the user.
-   :reqjson string event_subtype: Optional. An optional subtype for the entry. Possible event types can be seen in HistoryEventSubType enum.
-   :reqjson string counterparty: Optional. An identifier for a potential counterparty of the event entry. For a send it's the target. For a receive it's the sender. For bridged transfer it's the bridge's network identifier. For a protocol interaction it's the protocol.
 
    **Example Response**:
 
@@ -4756,6 +4885,7 @@ Dealing with History Events
       Content-Type: application/json;charset=UTF-8
 
       {
+          "entry_type": "evm event",
           "identifier": 243,
           "event_identifier": "0x64f1982504ab714037467fdd45d3ecf5a6356361403fc97dd325101d8c038c4e",
           "sequence_index": 162,
@@ -4770,7 +4900,7 @@ Dealing with History Events
           "counterparty": "0xdf869FAD6dB91f437B59F1EdEFab319493D4C4cE"
       }
 
-   The request object is the same as above, a base history entry, with the addition of the identifier which signifies which entry will be edited.
+   The request object uses all the same arguments for each entry type as the `add event endpoint <add_event_args_label_>`_, with the addition of the identifier which signifies which entry will be edited. 
 
    **Example Response**:
 
--- a/rotkehlchen/accounting/structures/base.py
+++ b/rotkehlchen/accounting/structures/base.py
@@ -22,7 +22,7 @@ from rotkehlchen.serialization.deserialize import (
     deserialize_timestamp,
 )
 from rotkehlchen.types import Location, Timestamp, TimestampMS
-from rotkehlchen.utils.misc import timestamp_to_date, ts_ms_to_sec, ts_sec_to_ms
+from rotkehlchen.utils.misc import timestamp_to_date, ts_ms_to_sec
 from rotkehlchen.utils.mixins.enums import DBIntEnumMixIn
 
 from .balance import Balance
@@ -259,7 +259,7 @@ class HistoryBaseEntry(AccountingEventMixin, metaclass=ABCMeta):
             return HistoryBaseEntryData(
                 event_identifier=data['event_identifier'],
                 sequence_index=data['sequence_index'],
-                timestamp=ts_sec_to_ms(deserialize_timestamp(data['timestamp'])),
+                timestamp=TimestampMS(deserialize_timestamp(data['timestamp'])),
                 location=Location.deserialize(data['location']),
                 event_type=HistoryEventType.deserialize(data['event_type']),
                 event_subtype=HistoryEventSubType.deserialize(data['event_subtype']) if data['event_subtype'] is not None else HistoryEventSubType.NONE,  # noqa: E501
--- a/rotkehlchen/accounting/structures/base.py
+++ b/rotkehlchen/accounting/structures/base.py
@@ -209,7 +209,7 @@ class HistoryBaseEntry(AccountingEventMixin, metaclass=ABCMeta):
             'entry_type': self.entry_type.serialize(),
             'event_identifier': self.event_identifier,
             'sequence_index': self.sequence_index,
-            'timestamp': ts_ms_to_sec(self.timestamp),  # serialize to api in seconds MS
+            'timestamp': self.timestamp,
             'location': str(self.location),
             'asset': self.asset.identifier,
             'balance': self.balance.serialize(),
--- a/rotkehlchen/accounting/structures/evm_event.py
+++ b/rotkehlchen/accounting/structures/evm_event.py
@@ -153,6 +153,7 @@ class EvmEvent(HistoryBaseEntry):  # noqa: PLW1641  # hash in superclass
             'counterparty': self.counterparty,
             'product': self.product.serialize() if self.product is not None else None,
             'address': self.address,
+            'extra_data': self.extra_data,
         }
 
     def serialize_for_api(
--- a/rotkehlchen/api/v1/schemas.py
+++ b/rotkehlchen/api/v1/schemas.py
@@ -1,14 +1,19 @@
 import logging
-from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Union, get_args
+from typing import TYPE_CHECKING, Any, Callable, Final, Literal, Optional, Union, get_args
 
 import marshmallow
 import webargs
 from eth_utils import to_checksum_address
-from marshmallow import Schema, fields, post_load, validate, validates_schema
+from marshmallow import INCLUDE, Schema, fields, post_load, validate, validates_schema
 from marshmallow.exceptions import ValidationError
 
 from rotkehlchen.accounting.structures.balance import Balance, BalanceType
-from rotkehlchen.accounting.structures.base import HistoryBaseEntryType
+from rotkehlchen.accounting.structures.base import HistoryBaseEntryType, HistoryEvent
+from rotkehlchen.accounting.structures.eth2 import (
+    EthBlockEvent,
+    EthDepositEvent,
+    EthWithdrawalEvent,
+)
 from rotkehlchen.accounting.structures.evm_event import EvmEvent, EvmProduct
 from rotkehlchen.accounting.structures.types import (
     ActionType,
@@ -131,6 +136,7 @@ from .fields import (
     SerializableEnumField,
     TaxFreeAfterPeriodField,
     TimestampField,
+    TimestampMSField,
     TimestampUntilNowField,
     XpubField,
 )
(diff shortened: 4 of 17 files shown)


rotkehlchen/accounting/structures/base.py
```diff
@@ -315,6 +315,24 @@ def get_identifier(self) -> str:
     def get_assets(self) -> list[Asset]:
         return [self.asset]
 
+    @classmethod
+    def resolve_subclass(cls, entry_type: HistoryBaseEntryType) -> type['HistoryBaseEntry']:
+        from rotkehlchen.accounting.structures.eth2 import (  # pylint: disable=import-outside-toplevel  # noqa: E501
```
LefterisJP, line 320: no no no. Do not ever do that. Don't import outside top level.

rotkehlchen/accounting/structures/evm_event.py
```diff
@@ -156,6 +156,7 @@ def serialize(self) -> dict[str, Any]:
             'counterparty': self.counterparty,
             'product': self.product.serialize() if self.product is not None else None,
             'address': self.address,
+            'extra_data': self.extra_data,
```
LefterisJP, line 159: why did you add this here? The extra data are not supposed to be serialized for the api ..

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
-    address = EvmAddressField(load_default=None)
+    counterparty = fields.String(required=False, allow_none=True)
+    product = SerializableEnumField(enum_class=EvmProduct, required=False, allow_none=True)
+    address = EvmAddressField(required=False, allow_none=True)
+    event_identifier = fields.String(required=False)
+    fee_recipient = EvmAddressField(required=False)
+    depositor = EvmAddressField(required=False)
+    withdrawal_address = EvmAddressField(required=False)
+    tx_hash = EVMTransactionHashField(required=False)
+    block_number = fields.Integer(
```
LefterisJP, line 747: it's `is_exit_or_blocknumber` as you can see from the code.

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+        required=False,
+        validate=webargs.validate.Range(
+            min=0,
+            error='Block number must be an integer >= 0',
```
LefterisJP, line 751:
```suggestion
            error='is_exit_or_blocknumber must be an integer >= 0',
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+        validate=webargs.validate.Range(
+            min=0,
+            error='Validator index must be an integer >= 0',
+        ),
+    )
+    is_mev_reward = fields.Boolean(required=False)
+    is_exit = fields.Boolean(required=False)
+    extra_data = fields.Dict(required=False, allow_none=True)
+
+    def validate_required(self, required_fields: tuple, data: dict[str, Any]) -> None:
```
LefterisJP, line 765: move this function in `validator_required_fields` ... what's the point of having it standalone here?

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
-    tx_hash = EVMTransactionHashField(required=True)
-    sequence_index = fields.Integer(required=True)
     # Timestamp coming in from the API is in seconds, in contrast to what we save in the struct
     timestamp = TimestampField(ts_multiplier=1000, required=True)
-    location = LocationField(required=True)
-    event_type = SerializableEnumField(enum_class=HistoryEventType, required=True)
-    asset = AssetField(required=True, expected_type=Asset, form_with_incomplete_data=True)
     balance = fields.Nested(BalanceSchema, required=True)
+    entry_type = SerializableEnumField(enum_class=HistoryBaseEntryType, required=True)
+    sequence_index = fields.Integer(required=False)
```
LefterisJP, line 729: This is required. But should default to 0 if not given.

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+    location = LocationField(required=False)
```
LefterisJP, line 730: this is required

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+    event_type = SerializableEnumField(enum_class=HistoryEventType, required=False)
```
LefterisJP, line 731: please put subtype and type next to each other. And this is required

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+    asset = AssetField(required=False, expected_type=Asset, form_with_incomplete_data=True)
```
LefterisJP, line 732: this is required

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
-        load_default=HistoryEventSubType.NONE,
         allow_none=True,
     )
-    counterparty = fields.String(load_default=None)
-    product = SerializableEnumField(enum_class=EvmProduct, load_default=None)
-    address = EvmAddressField(load_default=None)
+    counterparty = fields.String(required=False, allow_none=True)
+    product = SerializableEnumField(enum_class=EvmProduct, required=False, allow_none=True)
+    address = EvmAddressField(required=False, allow_none=True)
+    event_identifier = fields.String(required=False)
```
LefterisJP, line 742: This is required.

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+    fee_recipient = EvmAddressField(required=False)
```
LefterisJP, line 743: This is not required. This is the location label

```suggestion
    fee_recipient = EvmAddressField(required=False)
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+    depositor = EvmAddressField(required=False)
+    withdrawal_address = EvmAddressField(required=False)
```
LefterisJP, line 745: not required, this is again location label depending on context
```suggestion
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
     )
-    counterparty = fields.String(load_default=None)
-    product = SerializableEnumField(enum_class=EvmProduct, load_default=None)
-    address = EvmAddressField(load_default=None)
+    counterparty = fields.String(required=False, allow_none=True)
+    product = SerializableEnumField(enum_class=EvmProduct, required=False, allow_none=True)
+    address = EvmAddressField(required=False, allow_none=True)
+    event_identifier = fields.String(required=False)
+    fee_recipient = EvmAddressField(required=False)
+    depositor = EvmAddressField(required=False)
```
LefterisJP, line 744: not required, this is again location label depending on context
```suggestion
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+        ),
+    )
+    validator_index = fields.Integer(
+        required=False,
+        validate=webargs.validate.Range(
+            min=0,
+            error='Validator index must be an integer >= 0',
+        ),
+    )
+    is_mev_reward = fields.Boolean(required=False)
```
LefterisJP, line 761: This is not required. This will be determined from type/subtype. You can abstract the logic

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+    is_exit = fields.Boolean(required=False)
```
LefterisJP, line 762: not required. Taken from `is_exit_or_blocknumber`
```suggestion
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,47 +720,89 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
 
     @post_load
     def make_history_base_entry(
             self,
             data: dict[str, Any],
             **_kwargs: Any,
     ) -> dict[str, Any]:
-        if data.get('event_subtype', None) is None:
+        entry_type = data.pop('entry_type')
+        if entry_type in (HistoryBaseEntryType.HISTORY_EVENT, HistoryBaseEntryType.EVM_EVENT) and data.get('event_subtype') is None:  # noqa: E501
```
LefterisJP, line 799:
```suggestion
        if data.get('event_subtype') is None:  # noqa: E501
```

rotkehlchen/accounting/structures/base.py
```diff
@@ -315,6 +315,24 @@ def get_identifier(self) -> str:
...
+            EthBlockEvent,
+            EthDepositEvent,
+            EthWithdrawalEvent,
+        )
+        from rotkehlchen.accounting.structures.evm_event import \
```
LefterisJP, line 325: what's this? Where in the code have you seen us use `\` to break a line. Never do this.

LefterisJP (commented):
Okay the approach is wrong and needs quite a bit of change.

1. Please be linking to what the PR is for in the body of the PR: https://github.com/orgs/rotki/projects/11?pane=issue&itemId=40691321
2. Never add new code that uses imports inside a function. All these imports run every time the function is called. It's a huge no no .
3. I see you tried to match the constructors of the various events to the arguments in the API. That won't work since you also have to edit them. You have to match them to the representation they have in the structures/DB and then edit accordingly with extra logic for each event. That logic should also verify that the type/subtype and other values for the event ar e legal values
4. Never ever do things that go against the style of the code ilke breaking lines with `\`

rotkehlchen/accounting/structures/evm_event.py
LefterisJP: ah wait no, you need it for editing probably

rotkehlchen/api/v1/schemas.py
jbrit: it is required for history events, I did the validation based on input type in `validate_required_fields`

rotkehlchen/api/v1/schemas.py
jbrit: I did the validation for the events  that required it in validate_required_fields

rotkehlchen/api/v1/schemas.py
jbrit: okay. I did the validation for the events  that required it in `validate_required_fields`

rotkehlchen/api/v1/schemas.py
jbrit: okay, yabir mentioned you didn't like defining functions in functions

rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -16,39 +20,104 @@
...
+    entry.asset = A_USDT
+    entry.balance = Balance(amount=FVal('1500.1'), usd_value=FVal('1499.45'))
+    entry.location_label = '0x9531C059098e3d194fF87FebB587aB07B30B1306'
+    entry.notes = 'Deposit stuff for staking somewhere'
+    entry.event_subtype = HistoryEventSubType.NONE
+    json_data = entry_to_input_dict(entry, include_identifier=True)
+    response = requests.patch(
+        api_url_for(rotkehlchen_api_server, 'historyeventresource'),
+        json=json_data,
+    )
```
yabirgb, line 52: No need to create the variable if it won't be later reused
```suggestion
    response = requests.patch(
        api_url_for(rotkehlchen_api_server, 'historyeventresource'),
        json=entry_to_input_dict(entry, include_identifier=True),
    )
```

rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -16,39 +20,104 @@
...
+    assert_simple_ok_response(response)
+    assert entry.identifier is not None
+
+
+def pop_useless_fields(serialized_event: dict[str, Any], fields: tuple[str, ...]):
+    for field in fields:
+        serialized_event.pop(field)
```
yabirgb, line 59: add a docstring. Notice that this can raise error if the key is not present in the mapping

```suggestion
def pop_multiple_keys(serialized_event: dict[str, Any], fields: tuple[str, ...]):
    for field in fields:
        serialized_event.pop(field)
```

rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -16,39 +20,104 @@
...
+    elif entry.entry_type == HistoryBaseEntryType.ETH_WITHDRAWAL_EVENT:
+        pop_useless_fields(serialized, (
+            'event_identifier',
+            'sequence_index',
+            'location',
+            'asset',
+            'event_type',
+            'event_subtype',
+            'notes',
+        ))
```
yabirgb, line 87: since you have to name the fields that need to be removed... why not create a mapping with the fields that you want to mantain when removing multiples? If you end up adding more fields we will need to update this

rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -148,35 +217,13 @@ def add_entries(server: 'APIServer', events_db: DBHistoryEvents, add_directly: b
...
-    if add_directly is False:
-        assert entries_added_using_api == 4, 'api was used directly and event is missing'
-
+        json_data = entry_to_input_dict(entry, include_identifier=False)
+        response = requests.put(
+            api_url_for(server, 'historyeventresource'),
+            json=json_data,
+        )
+        result = assert_proper_response_with_result(response)
+        assert 'identifier' in result
```
yabirgb, line 227: this is doing an http request, exceuting the validation logic... it is doing more things than what you need. If you want only to add the events call the method directly from the db object instead

rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -16,39 +20,104 @@
...
+            'notes',
+            'product',
+            'counterparty',
+        ))
+        serialized['depositor'] = serialized.pop('location_label')
     return serialized
 
 
-def add_entries(server: 'APIServer', events_db: DBHistoryEvents, add_directly: bool = False) -> list['HistoryBaseEntry']:  # noqa: E501
+def add_entries(server: 'APIServer') -> list['HistoryBaseEntry']:
```
yabirgb, line 115: don't forget to update the docstring

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,49 +725,168 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+            HistoryBaseEntryType.ETH_DEPOSIT_EVENT: self.CreateEthDepositEventEventSchema,
+            HistoryBaseEntryType.ETH_WITHDRAWAL_EVENT: self.CreateEthWithdrawalEventEventSchema,
+            HistoryBaseEntryType.EVM_EVENT: self.CreateEvmEventSchema,
+        }
+        exclude = () if self.include_identifier else ('identifier',)
+        return entry_to_schema[entry_type](exclude=exclude).load(data)
 
-        return {'event': EvmEvent(**data)}
+    class Meta:
+        unknown = INCLUDE
```
yabirgb, line 884: can you add a comment about why this is needed?

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,49 +725,168 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
     ) -> dict[str, Any]:
-        if data.get('event_subtype', None) is None:
-            data['event_subtype'] = HistoryEventSubType.NONE
-
-        if data['location'] not in EVM_LOCATIONS:
-            raise ValidationError(
-                message=f'EVM event location needs to be one of {EVM_LOCATIONS}',
-                field_name='location',
-            )
+        entry_type = data.pop('entry_type')
```
yabirgb, line 872: if you pop it it will be removed from the mapping. I would do a get instead. You won't use later but is something unexpected that if in the future we need to use we will find that is missing from the data

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,49 +725,168 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+        entry_to_schema: dict[HistoryBaseEntryType, type[Schema]] = {
+            HistoryBaseEntryType.HISTORY_EVENT: self.CreateBaseHistoryEventSchema,
+            HistoryBaseEntryType.ETH_BLOCK_EVENT: self.CreateEthBlockEventEventSchema,
+            HistoryBaseEntryType.ETH_DEPOSIT_EVENT: self.CreateEthDepositEventEventSchema,
+            HistoryBaseEntryType.ETH_WITHDRAWAL_EVENT: self.CreateEthWithdrawalEventEventSchema,
+            HistoryBaseEntryType.EVM_EVENT: self.CreateEvmEventSchema,
+        }
+        exclude = () if self.include_identifier else ('identifier',)
+        return entry_to_schema[entry_type](exclude=exclude).load(data)
```
yabirgb, line 881:
```suggestion
        exclude = () if self.include_identifier else ('identifier',)
        return entry_to_schema[data.get('entry_type')](exclude=exclude).load(data)
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,49 +725,168 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
-                field_name='location',
-            )
+        entry_type = data.pop('entry_type')
+        entry_to_schema: dict[HistoryBaseEntryType, type[Schema]] = {
+            HistoryBaseEntryType.HISTORY_EVENT: self.CreateBaseHistoryEventSchema,
+            HistoryBaseEntryType.ETH_BLOCK_EVENT: self.CreateEthBlockEventEventSchema,
+            HistoryBaseEntryType.ETH_DEPOSIT_EVENT: self.CreateEthDepositEventEventSchema,
+            HistoryBaseEntryType.ETH_WITHDRAWAL_EVENT: self.CreateEthWithdrawalEventEventSchema,
+            HistoryBaseEntryType.EVM_EVENT: self.CreateEvmEventSchema,
+        }
```
yabirgb, line 879: maybe define this mapping as a class attribute? Otherwise you are creating it every time the endpoint is executed

yabirgb (commented):
This PR is missing the changes in the docs file

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,49 +725,168 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+        notes = fields.String(load_default=None)
+
+        @post_load
+        def make_history_base_entry(
+                self,
+                data: dict[str, Any],
+                **_kwargs: Any,
+        ) -> dict[str, Any]:
+            if data.get('event_subtype', None) is None:
+                data['event_subtype'] = HistoryEventSubType.NONE
```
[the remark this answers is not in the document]
jbrit: Yes, `allow_none=True` is on the event_subtype field

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,49 +725,168 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+            if data['location'] not in EVM_LOCATIONS:
```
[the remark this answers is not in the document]
jbrit: it's harder to do the validation at this level if I abstract the location field

rotkehlchen/api/v1/schemas.py
jbrit: i need to remove it from the mapping, it throws an error during validation if I don't

rotkehlchen/api/v1/schemas.py
```diff
@@ -720,49 +725,155 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+        balance = fields.Nested(BalanceSchema, required=True)
+        identifier = fields.Integer(required=True)
+
+    class BaseEventSchema(BaseSchema):
+        event_type = SerializableEnumField(enum_class=HistoryEventType, required=True)
+        event_subtype = SerializableEnumField(
+            enum_class=HistoryEventSubType,
+            required=False,
+            load_default=HistoryEventSubType.NONE,
+            allow_none=True,
```
yabirgb, line 745: when you use the load_default + required = False this `allow_none` what does it mean?

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -40,14 +40,40 @@
...
 
 
+def assert_editing_works(
+        entry: 'HistoryBaseEntry',
+        rotkehlchen_api_server: 'APIServer',
+        sequence_index: int,
+):
+    """
+    a test to check editing any subclass of history base entry is successful
+    """
```
yabirgb, line 54:
```suggestion
    """Assert that editing any subclass of history base entry is successful done"""
```

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -99,7 +125,8 @@ def test_add_edit_delete_entries(rotkehlchen_api_server: 'APIServer'):
         contained_in_msg='Failed to add event to the DB. It already exists',
         status_code=HTTPStatus.CONFLICT,
     )
-    # test editing works
+
+    # test editing evm event works
```
yabirgb, line 129: Don't you have here below a copy-paste of what is being done in `assert_editing_works`?

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -40,14 +40,40 @@
...
+    entry.location_label = '0x9531C059098e3d194fF87FebB587aB07B30B1306'
+    entry.notes = 'Deposit stuff for staking somewhere'
+    entry.event_subtype = HistoryEventSubType.NONE
+    json_data = entry_to_input_dict(entry, include_identifier=True)
+    response = requests.patch(
+        api_url_for(rotkehlchen_api_server, 'historyeventresource'),
+        json=json_data,
+    )
+    assert_simple_ok_response(response)
+    assert entry.identifier is not None
```
yabirgb, line 69: shouldn't you also check that the information in the db matches what you expect?

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -119,6 +146,15 @@ def test_add_edit_delete_entries(rotkehlchen_api_server: 'APIServer'):
...
     assert entry.identifier is not None
 
+    # test editing history event
+    assert_editing_works(entries[5], rotkehlchen_api_server, 5)
+    # test editing eth withdrawal event
+    assert_editing_works(entries[6], rotkehlchen_api_server, 6)
+    # test editing eth_block event
+    assert_editing_works(entries[7], rotkehlchen_api_server, 7)
+    # test editing eth_deposit event
+    assert_editing_works(entries[8], rotkehlchen_api_server, 8)
```
yabirgb, line 156: can you assert before sending the patch request that the event has the type that you expect? Then you can remove the comment

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -40,14 +40,40 @@
...
+    a test to check editing any subclass of history base entry is successful
+    """
+    entry.sequence_index = sequence_index
+    entry.timestamp = TimestampMS(1639924575000)
+    entry.event_type = HistoryEventType.DEPOSIT
+    entry.asset = A_USDT
+    entry.balance = Balance(amount=FVal('1500.1'), usd_value=FVal('1499.45'))
+    entry.location_label = '0x9531C059098e3d194fF87FebB587aB07B30B1306'
+    entry.notes = 'Deposit stuff for staking somewhere'
+    entry.event_subtype = HistoryEventSubType.NONE
```
yabirgb, line 62: I would not try to edit everything because is a bit confusing. I would have a if check for the event type and modify only the fields that matter. Maybe also another one not related to check that the logic in marshmallow works as you expect it but not everything for all the event types

docs/api.rst
```diff
@@ -4702,9 +4703,81 @@ Dealing with History Events
...
+          "location": "ethereum",
+          "timestamp": 1569924574,
+          "balance": {"amount": "1.542", "usd_value": "1.675"},
+          "sequence_index": 162,
+          "event_type": "informational",
+          "event_subtype": "approve",
+          "asset": "eip155:1/erc20:0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359",
+          "location_label": "0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12",
+          "notes": "Approve 1 SAI of 0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12 for spending by 0xdf869FAD6dB91f437B59F1EdEFab319493D4C4cE"
+      }
```
yabirgb, line 4728: I'm not sure but is it possible to create sub sections inside this section for the different event types and then have each one with their special fields only and then later the fields that are common for all with a reference? I believe it would improve this section a lot

docs/api.rst
```diff
@@ -4716,6 +4789,14 @@ Dealing with History Events
    :reqjson string notes: This is a description of the event entry in plain text explaining what is being done. This is supposed to be shown to the user.
    :reqjson string event_subtype: Optional. An optional subtype for the entry. Possible event types can be seen in HistoryEventSubType enum.
    :reqjson string counterparty: Optional. An identifier for a potential counterparty of the event entry. For a send it's the target. For a receive it's the sender. For bridged transfer it's the bridge's network identifier. For a protocol interaction it's the protocol.
+   :reqjson int block_number: This is the index of the block the event is contained in.
```
yabirgb, line 4792:
```suggestion
   :reqjson int block_number: This is the number of the block where the event took place.
```

docs/api.rst
```diff
@@ -4716,6 +4789,14 @@ Dealing with History Events
...
+   :reqjson int validator_index: This is the index of the validator.
```
yabirgb, line 4793: can you check the errors? I bleieve the text need updates. For example if there is a constraint that fails like the validator index doesn't exists, I believe there is one like that

docs/api.rst
```diff
@@ -4716,6 +4789,14 @@ Dealing with History Events
...
+   :reqjson string event_identifier: The event identifier to be used for the event.
```
yabirgb, line 4794: This one has to be unique too

docs/api.rst
```diff
@@ -4716,6 +4789,14 @@ Dealing with History Events
...
+   :reqjson string fee_recipient: an evm address field to specify the recipient in an "eth block event".
+   :reqjson string depositor: an evm address field to specify the depositor in an "eth deposit event".
+   :reqjson string withdrawal_address: an evm address field to specify the withdrawer in an "eth withdrawal event".
+   :reqjson bool is_exit: boolean field to specify if the "eth withdrawal event" is an exit.
```
yabirgb, line 4798:
```suggestion
   :reqjson bool is_exit: true if the "eth withdrawal event" is an exit event.
```

docs/api.rst
```diff
@@ -4716,6 +4789,14 @@ Dealing with History Events
...
+   :reqjson bool is_mev_reward: boolean field to specify if the "eth block event" is an mev reward event.
```
yabirgb, line 4799:
```suggestion
   :reqjson bool is_mev_reward: true if the "eth block event" is an mev reward event.
```

docs/api.rst
yabirgb, line 4798:
```suggestion
   :reqjson bool is_exit: true if the "eth withdrawal event" is an exit event.
```

docs/api.rst
```diff
@@ -4702,9 +4703,81 @@ Dealing with History Events
...
+      {
+          "entry_type": "eth withdrawal event",
+          "timestamp": 1569924574,
+          "balance": {"amount": "1.542", "usd_value": "1.675"},
+          "is_exit": true,
+          "validator_index": 1,
+          "withdrawal_address": "0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12"
       }
 
+   :reqjson string entry_type: This is the type of the event.
```
yabirgb, line 4780:
```suggestion
   :reqjson string entry_type: The type of the event that will be processed. Different validation is used based on the value for this field
```

also list the possible values

yabirgb (commented):
A few things to improve

rotkehlchen/api/v1/schemas.py
jbrit: `allow_none` means a null value can be sent from the frontend

docs/api.rst
jbrit: The errors seem correct

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -43,11 +44,47 @@
...
+            rotkehlchen_api_server,
+            'historyeventresource',
+        ),
+        json={'event_identifiers': [entry.event_identifier]},
+    )
+    result = assert_proper_response_with_result(response)
+    for result_entry in result['entries']:
+        if result_entry['entry']['identifier'] == entry.identifier:
+            for valid_key in entry_keys[entry.entry_type]:
+                assert result_entry['entry'][valid_key] == entry.serialize()[valid_key]
```
yabirgb, line 80: I believe to have mentioned it in other review but if you are not testing the api avoid making calls. It executes extra logic that you don't need. Here you can query the db directly using a filter and `get_history_events`. It will return the same information + you don't need to serialize the object if I'm correct

rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -16,39 +18,59 @@
...
+    from rotkehlchen.db.history_events import DBHistoryEvents
+
+
+entry_keys: dict[HistoryBaseEntryType, set[str]] = {
+    HistoryBaseEntryType.HISTORY_EVENT: {'sequence_index', 'location', 'event_type', 'event_subtype', 'asset', 'notes', 'event_identifier'},  # noqa: E501
+    HistoryBaseEntryType.ETH_BLOCK_EVENT: {'validator_index', 'block_number', 'event_subtype'},
+    HistoryBaseEntryType.ETH_DEPOSIT_EVENT: {'tx_hash', 'validator_index', 'sequence_index', 'event_identifier', 'extra_data'},  # noqa: E501
+    HistoryBaseEntryType.ETH_WITHDRAWAL_EVENT: {'validator_index', 'is_exit'},
+    HistoryBaseEntryType.EVM_EVENT: {'tx_hash', 'sequence_index', 'location', 'event_type', 'event_subtype', 'asset', 'notes', 'counterparty', 'product', 'address', 'extra_data', 'event_identifier'},  # noqa: E501
+}
```
yabirgb, line 34: saw the code where it is used and the name is confusing. Also for constant we use capital letters

```suggestion
KEYS_IN_ENTRY_TYPE: dict[HistoryBaseEntryType, set[str]] = {
    HistoryBaseEntryType.HISTORY_EVENT: {'sequence_index', 'location', 'event_type', 'event_subtype', 'asset', 'notes', 'event_identifier'},  # noqa: E501
    HistoryBaseEntryType.ETH_BLOCK_EVENT: {'validator_index', 'block_number', 'event_subtype'},
    HistoryBaseEntryType.ETH_DEPOSIT_EVENT: {'tx_hash', 'validator_index', 'sequence_index', 'event_identifier', 'extra_data'},  # noqa: E501
    HistoryBaseEntryType.ETH_WITHDRAWAL_EVENT: {'validator_index', 'is_exit'},
    HistoryBaseEntryType.EVM_EVENT: {'tx_hash', 'sequence_index', 'location', 'event_type', 'event_subtype', 'asset', 'notes', 'counterparty', 'product', 'address', 'extra_data', 'event_identifier'},  # noqa: E501
}
```

docs/api.rst
```diff
@@ -4702,9 +4703,81 @@ Dealing with History Events
...
+          "balance": {"amount": "1.542", "usd_value": "1.675"},
+          "is_exit": true,
+          "validator_index": 1,
+          "withdrawal_address": "0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12"
       }
 
+   :resjson string entry_type: The type of the event that will be processed. Different validation is used based on the value for this field. Possible values are: ``"history event"``, ``"evm event"``, ``"eth withdrawal event"``, ``"eth block event"``, ``"eth deposit event"``.
    :reqjson string tx_hash: This is the transaction hash of the evm event
    :reqjson int sequence_index: This is an index that tries to provide the order of history entries for a single event_identifier.
    :reqjson int timestamp: The timestamp of the entry
```
yabirgb, line 4783: Don't forget to also split the arguments depending on the type. As it is now is unredeable

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -43,11 +44,47 @@
...
+        rotkehlchen_api_server: 'APIServer',
+        sequence_index: int,
+):
+    """Assert that editing any subclass of history base entry is successful done"""
+    entry.sequence_index = sequence_index
+    entry.timestamp = TimestampMS(1639924575000)
+    entry.balance = Balance(amount=FVal('1500.1'), usd_value=FVal('1499.45'))
+    entry.location_label = '0x9531C059098e3d194fF87FebB587aB07B30B1306'
+    entry.asset = A_USDT
+    entry.notes = 'Deposit stuff for staking somewhere'
```
yabirgb, line 58: can you maybe change this note to make more sense

```suggestion
    entry.notes = 'Updated entry for test using type {type(entry)}'
```

or something that makes them unique

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -43,11 +44,47 @@
...
+        entry: 'HistoryBaseEntry',
+        rotkehlchen_api_server: 'APIServer',
+        sequence_index: int,
+):
+    """Assert that editing any subclass of history base entry is successful done"""
+    entry.sequence_index = sequence_index
+    entry.timestamp = TimestampMS(1639924575000)
+    entry.balance = Balance(amount=FVal('1500.1'), usd_value=FVal('1499.45'))
+    entry.location_label = '0x9531C059098e3d194fF87FebB587aB07B30B1306'
+    entry.asset = A_USDT
```
yabirgb, line 57: Don't change the asset for the eth staking events. Shouldn't it fail the test? Since staking events should have always ETH

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -43,11 +44,47 @@
...
+    entry.notes = 'Deposit stuff for staking somewhere'
+    if entry.entry_type in (HistoryBaseEntryType.EVM_EVENT, HistoryBaseEntryType.HISTORY_EVENT):
+        entry.event_type = HistoryEventType.DEPOSIT
+        entry.event_subtype = HistoryEventSubType.NONE
+    json_data = entry_to_input_dict(entry, include_identifier=True)
+    response = requests.patch(
+        api_url_for(rotkehlchen_api_server, 'historyeventresource'),
+        json=json_data,
+    )
```
yabirgb, line 66: this you didn't change

```suggestion
    response = requests.patch(
        api_url_for(rotkehlchen_api_server, 'historyeventresource'),
        json=entry_to_input_dict(entry, include_identifier=True),
    )
```

rotkehlchen/tests/api/test_history_base_entry.py
jbrit: The asset doesn’t get updated there, it gets pooped in  `pop_multiple_keys`

rotkehlchen/api/v1/schemas.py
```diff
@@ -734,49 +739,155 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
-    )
-    counterparty = fields.String(load_default=None)
-    product = SerializableEnumField(enum_class=EvmProduct, load_default=None)
-    address = EvmAddressField(load_default=None)
+    include_identifier: bool = False
+    entry_type = SerializableEnumField(enum_class=HistoryBaseEntryType, required=True)
+
+    class BaseSchema(Schema):
+        # Timestamp coming in from the API is in seconds, in contrast to what we save in the struct
+        timestamp = TimestampField(ts_multiplier=1000, required=True)
```
LefterisJP, line 749: Change this. Allow the API and instruct the frontend to send timestamp in milliseconds. Otherwise during editing we will lose precision.

Perhaps make an MS timestamp field or somehow make TimestampField be parametrizable, so that typing works fine and is most of all obvious to the reader of the schema what is the precision here.

```suggestion
        timestamp = TimestampMSField(required=True)
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -734,49 +739,155 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+        extra_data = fields.Dict(load_default=None)
+        location = LocationField(required=True, limit_to=EVM_LOCATIONS)
+
+        @post_load
+        def make_history_base_entry(
+                self,
+                data: dict[str, Any],
+                **_kwargs: Any,
+        ) -> dict[str, Any]:
+            if data.get('event_subtype') is None:
```
LefterisJP, line 795: Perhaps time to change this. This used to be from before and is no longer needed since we are making changes here. Let's be explicit. Tell the frontend to send the special value `NONE` and that `null` is no longer a valid value. This way we can get rid of this special rule.

rotkehlchen/api/v1/schemas.py
```diff
@@ -734,49 +739,155 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
+        balance = fields.Nested(BalanceSchema, required=True)
+        identifier = fields.Integer(required=True)
+
+    class BaseEventSchema(BaseSchema):
+        event_type = SerializableEnumField(enum_class=HistoryEventType, required=True)
+        event_subtype = SerializableEnumField(
+            enum_class=HistoryEventSubType,
+            required=False,
+            load_default=HistoryEventSubType.NONE,
+            allow_none=True,
```
LefterisJP, line 759: As mentioned below:

1.  Remove this
2. Communicate the change in the api to the fronted
3. Edit the docs


```suggestion
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -734,49 +739,155 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
...
     ) -> dict[str, Any]:
-        if data.get('event_subtype', None) is None:
-            data['event_subtype'] = HistoryEventSubType.NONE
-
-        if data['location'] not in EVM_LOCATIONS:
-            raise ValidationError(
-                message=f'EVM event location needs to be one of {EVM_LOCATIONS}',
-                field_name='location',
-            )
+        entry_type = data.pop('entry_type')
```
LefterisJP, line 880: Just explain why you pop
```suggestion
        entry_type = data.pop('entry_type')  # already used to decide schema
```

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -40,14 +44,57 @@
...
+        events_db: 'DBHistoryEvents',
+        sequence_index: int,
+):
+    valid_keys = KEYS_IN_ENTRY_TYPE[entry.entry_type]
+
+    def edit_entry(attr: str, value: Any):
+        if attr in valid_keys:
+            setattr(entry, attr, value)
+
+    """Assert that editing any subclass of history base entry is successful done"""
```
LefterisJP, line 63:
```suggestion
    """Assert that editing any subclass of history base entry is successfuly done"""
```

LefterisJP (commented):
Some things left to address, but otherwise pretty nice for my first review

docs/api.rst
```diff
@@ -4681,41 +4681,170 @@ Dealing with History Events
...
+                  "event_subtype": "approve",
+                  "counterparty": "0xdf869FAD6dB91f437B59F1EdEFab319493D4C4cE",
+                  "extra_data": {}
+               }
+
+            :reqjson string tx_hash: This is the transaction hash of the evm event
+            :reqjson int sequence_index: This is an index that tries to provide the order of history entries for a single event_identifier.
+            :reqjson string location: The location of the entry. Such as "ethereum", "optimism", etc.
+            :reqjson string asset: The asset identifier for this entry
+            :reqjson string event_type: The main event type of the entry. Possible event types can be seen in HistoryEventType enum.
```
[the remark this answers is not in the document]
jbrit: that would probably mean I create a separate section just for these two arguments? I'm not sure what you have in mind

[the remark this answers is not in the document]
LefterisJP: No I mean add a link to the definition ofthe enum.

rotkehlchen/api/v1/schemas.py
jbrit: @rotki/frontend event_subtype does not accept null anymore, it accepts the string "none"

docs/api.rst
```diff
@@ -4689,41 +4689,170 @@ Dealing with History Events
...
+               "location_label": "0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12",
+               "notes": "Approve 1 SAI of 0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12 for spending by 0xdf869FAD6dB91f437B59F1EdEFab319493D4C4cE"
+            }
+
+         :reqjson int sequence_index: This is an index that tries to provide the order of history entries for a single event_identifier.
+         :reqjson string location: The location of the entry. Such as "ethereum", "optimism", etc.
+         :reqjson string asset: The asset identifier for this entry
+         :reqjson string event_identifier: The event identifier to be used for the event.
+         :reqjson string event_type: The main event type of the entry. Possible event types can be seen in the `HistoryEventType enum <https://github.com/rotki/rotki/blob/59aa288dacd1776e62682e711a916f32a14c04c2/rotkehlchen/accounting/structures/types.py#L54>`_.
+         :reqjson string[optional] event_subtype: An optional subtype for the entry. Possible event types can be seen in the `HistoryEventSubType enum <https://github.com/rotki/rotki/blob/59aa288dacd1776e62682e711a916f32a14c04c2/rotkehlchen/accounting/structures/types.py#L72>`_.
```
LefterisJP, line 4766: Same here. Why are you duplicating those btw? Shouldn't they be in the base description of the hisory event only?

```suggestion
         :reqjson event_subtype: The subtype for the entry. Possible event types can be seen in the `HistoryEventSubType enum <https://github.com/rotki/rotki/blob/59aa288dacd1776e62682e711a916f32a14c04c2/rotkehlchen/accounting/structures/types.py#L72>`_.
```

LefterisJP (commented):
Very small things left.

docs/api.rst
jbrit: they only appear in two of the subclasses. `entry_type`, `timestamp` & `balance` are the only fields that apply to all subclasses

docs/api.rst
```diff
@@ -4689,41 +4689,170 @@ Dealing with History Events
...
+            }
+
+         :reqjson int validator_index: This is the index of the validator.
+         :reqjson string withdrawal_address: an evm address field to specify the withdrawer in an "eth withdrawal event".
+         :reqjson bool is_exit: true if the "eth withdrawal event" is an exit event.
 
-   :reqjson string tx_hash: This is the transaction hash of the evm event
-   :reqjson int sequence_index: This is an index that tries to provide the order of history entries for a single event_identifier.
+   :reqjson string entry_type: The type of the event that will be processed. Different validation is used based on the value for this field. Possible values are: ``"history event"``, ``"evm event"``, ``"eth withdrawal event"``, ``"eth block event"``, ``"eth deposit event"``.
    :reqjson int timestamp: The timestamp of the entry
```
LefterisJP, line 4854: Please specify here that it's in milliseconds now

LefterisJP (commented):
Left a tiny comment in the docs but since I found a circular import which caused a failure am fixing it with my commit.

LefterisJP (approved):
LGTM. Thank you ajibola

PR #6710, opened 2023-10-06
fix: addressbook entry name search

lukicenturi:
So for the addressbook, the blockchain can be a `blockchain string`, or null (it means for all evm chains).

When we try to fetch the address names, if there's an entry in db with `null` blockchain, it should also be taken into consideration.

### DIFF
--- a/rotkehlchen/db/addressbook.py
+++ b/rotkehlchen/db/addressbook.py
@@ -168,12 +168,18 @@ class DBAddressbook:
             chain_address: OptionalChainAddress,
     ) -> Optional[str]:
         """
-        Returns the name for the specified address and blockchain or None if either there is
-        no name set or the pair (address, blockchain) doesn't exist in the database.
+        Returns the name for the specified address and blockchain.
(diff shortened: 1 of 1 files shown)


rotkehlchen/db/addressbook.py
```diff
@@ -168,14 +168,14 @@ def get_addressbook_entry_name(
             chain_address: OptionalChainAddress,
     ) -> Optional[str]:
         """
-        Returns the name for the specified address and blockchain or None if either there is
-        no name set or the pair (address, blockchain) doesn't exist in the database.
+        Returns the name for the specified address and blockchain.
+        It will search for the pair of address and the exact blockchain.
+        If it's not found, it will search for the pair of address with blockchain (meaning, for all EVM chains).
+        Otherwise, it will return None.
```
LefterisJP, line 174: also in the docstring explain why you need the order by

LefterisJP (approved):
LGTM Luki. Thank you. I will do the test myself in another PR.

PR #6722, opened 2023-10-09
Handle some kraken event edge cases

### LINKED ISSUE #6169: Handle known kraken edge cases with negative amounts properly

LefterisJP:
Fixes #6169 

### [Properly handle negative amounts for kraken staking](https://github.com/rotki/rotki/commit/8bc1afdbc29529e8addf53e47a23c49ee4e965ce) 

This handles part of https://github.com/rotki/rotki/issues/6169 where
negative amounts appeared in Kraken for:

- Moving from/to staking or earn program.
- Internal transfer events such as crediting forked coins, or trading
  away delisted assets
- OTC or private deals

### [Handle kraken failed withdrawals and other events properly](https://github.com/rotki/rotki/commit/01c0bbe23ba4abc480a4bfe95f0af3cae5096782)

Handle failed transactions such as withdrawals deposit that cancel each other out.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -18,6 +18,7 @@ Changelog
 * :bug:`6647` When PnL report generation is finished, users will not be redirected to the report page, but will get notified instead.
 * :bug:`6667` Wrong ENS name should no longer be reported for some edge case of ENS actions and the name should also appear in more events than before.
 * :bug:`-` Aave v2 accounting for deposit interest profit and borrow payback loss should now work correctly again.
+* :bug:`6169` Handle some kraken event categories better and show them in the UI. Categories like failed withdrawals, moving funds from/to staking and credits of forked assets or delisted asset trades.
 
 * :release:`1.30.2 <2023-09-21>`
 * :feature:`-` Improved support for importing Binance CSV files.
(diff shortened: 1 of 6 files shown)


LefterisJP:
all green and the filtering change removed. Since Yabir also checked and gave the okay I am merging.

PR #6726, opened 2023-10-10
Check for correctness of schema during db upgrade

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/tests/db/test_db_upgrades.py
+++ b/rotkehlchen/tests/db/test_db_upgrades.py
@@ -9,6 +9,7 @@ from pysqlcipher3 import dbapi2 as sqlcipher
 
 from rotkehlchen.constants.misc import DEFAULT_SQL_VM_INSTRUCTIONS_CB
 from rotkehlchen.data_handler import DataHandler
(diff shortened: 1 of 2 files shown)


rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1865,6 +1866,11 @@ def test_upgrade_db_39_to_40(user_data_dir):  # pylint: disable=unused-argument
         resume_from_backup=False,
     )
     cursor = db.conn.cursor()
+    sanity_check_impl(  # do sanity check on the db schema
```
LefterisJP, line 1869: Hmm this won't work here.The moment we change the schema this test will fail. The only way to check schema of only the latest upgrade is to put it in `test_latest_upgrade_adds_remove_tables` and then rename the test to a more generic name like `test_latest_upgrade_correctness`.

But this leaves every older upgrade unchecked. Which kind of sucks .... but anyway we cant' fix older upgrades can we? So perhaps doing this starting from now on it will be fine?

LefterisJP (approved):
LGTM. Thank you yabir

PR #6730, opened 2023-10-11
feat(frontend): custom accounting rules

lukicenturi:
Frontend part of https://github.com/rotki/rotki/pull/6695


frontend/app/src/components/display/BooleanDisplay.vue
```diff
@@ -0,0 +1,18 @@
+<script setup lang="ts">
+withDefaults(
+  defineProps<{
+    display: boolean;
```
kelsos, line 4: maybe we could rename this to `success`? 

I would expect display false to mean hidden.

Maybe also rename it to `SuccessDisplay` or something?

frontend/app/src/components/history/events/HistoryEventTypeCounterparty.vue
```diff
@@ -1,52 +1,70 @@
...
-                  v-else-if="counterparty.image"
-                  :src="`${imagePath}${counterparty.image}`"
-                  contain
-                />
+    <DefineImage>
+      <div>
+        <VAvatar v-if="counterparty" size="24">
+          <VIcon v-if="counterparty.icon" :color="counterparty.color">
+            {{ counterparty.icon }}
+          </VIcon>
```
kelsos, line 30: Hmmm damn this would require changes in the backend to migrate from mdi right?

frontend/app/src/composables/history/events/mapping.ts
```diff
@@ -198,7 +194,7 @@ export const useHistoryEventMappings = createSharedComposable(() => {
   const { scrambleData, scrambleHex } = useScramble();
 
   const getEventCounterpartyData = (
-    event: MaybeRef<EvmHistoryEvent | EthDepositEvent>
+    event: MaybeRef<{ counterparty?: string | null; address?: string | null }>
```
kelsos, line 197: When is the type null and when is it undefined?

frontend/app/src/types/settings/accounting.ts
```diff
@@ -0,0 +1,59 @@
...
+  SPEND = 'spend',
+  ACQUISITION = 'acquisition'
+}
+
+const AccountingMethodEnum = z.nativeEnum(AccountingMethod);
+
+export enum AccountingTreatment {
+  SWAP = 'swap',
+  SWAP_WITH_FEE = 'swap with fee'
+}
```
kelsos, line 15: Don't we get these dynamically somehow by the backend?

frontend/app/src/components/settings/accounting/rule/AccountingRuleForm.vue
```diff
@@ -0,0 +1,230 @@
...
+      : t('accounting_settings.rule.add_error');
+
+    if (e instanceof ApiValidationError) {
+      const messages = e.errors;
+
+      set(errorMessages, messages);
+
+      const keys = Object.keys(messages);
+      const knownKeys = Object.keys(payload);
+      const unknownKeys = keys.filter(key => !knownKeys.includes(key));
```
kelsos, line 102: I guess this differs a bit from the `getValidationErrors` method but maybe we could also integrate that in the ApiValidationError class too as an extra method.

kelsos:
image

We should also move the button outside of the table to follow what we do in other places

frontend/app/src/components/history/events/HistoryEventTypeCounterparty.vue
lukicenturi: Yes it is, but we can do it later hahaha, it will not be hard to do

frontend/app/src/composables/history/events/mapping.ts
lukicenturi: Updated, so the counterparty shouldn't be missing (but `string | null`). For the address it's string | null if it's from event. But for the usage in the accounting rule, I don't send it.

frontend/app/src/types/settings/accounting.ts
lukicenturi: no we don't get it from backend. It just stated on the docs

frontend/app/src/components/settings/accounting/rule/AccountingRuleForm.vue
lukicenturi: Sorry that was old implementation, updated to new one that use `getValidationErrors`

frontend/app/src/composables/history/events/mapping.ts
lukicenturi: Updated, so the counterparty shouldn't be missing (but `string | null`). For the address it's string | null if it's from event. But for the usage in the accounting rule, I don't send it.

kelsos (approved):
lgtm

kelsos:
@lukicenturi I guess we also need a changelog entry can you do that too?

lukicenturi:
> @lukicenturi I guess we also need a changelog entry can you do that too?

Will do that

PR #6738, opened 2023-10-12
Fetch remotly accounting rules

yabirgb:
This PR:

- Adds linked settings so properties of rules can be linked to certain preferences in settings
- Remote updates to add accounting rules and linked properties

Remains:
- Consensus with the frontend about the endpoints required to manage the linked properties
- Move the existing rules to the data repo


rotkehlchen/db/accounting_rules.py
```diff
@@ -32,7 +44,7 @@ def add_accounting_rule(
             event_subtype: HistoryEventSubType,
             counterparty: Optional[str],
             rule: 'BaseEventSettings',
-    ) -> None:
+    ) -> int:
```
LefterisJP, line 47: add to the docstring that the identifier of the rule is returned by the function

rotkehlchen/db/accounting_rules.py
```diff
@@ -7,11 +8,22 @@
...
 
 if TYPE_CHECKING:
     from rotkehlchen.db.dbhandler import DBHandler
 
 
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+class RuleInformation(NamedTuple):
```
LefterisJP, line 21: Docstring explaining what this is and especially why the identifier is a tuple (this is such a confusing name) don't use identifier as name for a tuple. And what is overwrites etc.

rotkehlchen/db/accounting_rules.py
```diff
@@ -117,30 +131,85 @@ def update_accounting_rule(
...
+        """
+        with self.db.conn.write_ctx() as write_cursor:
+            try:
+                write_cursor.execute(
+                    'INSERT INTO linked_rules_properties(accounting_rule, property_name, '
+                    'setting_name, active) VALUES (?, ?, ?, ?)',
+                    (rule_identifier, rule_property, setting_name, is_active),
+                )
+            except sqlcipher.IntegrityError as e:  # pylint: disable=no-member
+                raise InputError('Overwritting rule already exists') from e
```
LefterisJP, line 154:
```suggestion
                raise InputError(f'Link of rule for {rule_properly} to setting {setting_name} already exists') from e
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -117,30 +131,85 @@ def update_accounting_rule(
...
 
+    def add_overwritten_setting(
+            self,
+            rule_identifier: int,
+            rule_property: Literal['count_entire_amount_spend', 'count_cost_basis_pnl'],
+            setting_name: Literal['include_gas_costs', 'include_crypto2crypto'],
+            is_active: bool = True,
+    ) -> None:
+        """
+        Store in the database a rule overwritten property that links it to a setting in the
```
LefterisJP, line 142: You use the term "overwrite" here alot. This is probably not the best term. Perhaps just use "link_rule_to_setting" everywhere.

rotkehlchen/db/accounting_rules.py
```diff
@@ -7,11 +8,22 @@
...
 
 
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+class RuleInformation(NamedTuple):
+    identifier: tuple[int, HistoryEventType, HistoryEventSubType, str]
+    rule: BaseEventSettings
+    overwrites: dict[str, Any]
```
LefterisJP, line 24: the overwrites term is confusing as I wrote in the function below. Please rename.

rotkehlchen/db/accounting_rules.py
```diff
@@ -117,30 +131,85 @@ def update_accounting_rule(
...
+                raise InputError('Overwritting rule already exists') from e
+
     def query_rules(
             self,
             filter_query: AccountingRulesFilterQuery,
-    ) -> tuple[
-        list[tuple[tuple[int, HistoryEventType, HistoryEventSubType, Optional[str]], BaseEventSettings]],  # noqa: E501
-        int,
-    ]:
+    ) -> tuple[list[RuleInformation], int]:
```
LefterisJP, line 159: adjust the docstring here

rotkehlchen/db/accounting_rules.py
```diff
@@ -117,30 +131,85 @@ def update_accounting_rule(
...
+                is_active,
+            ) in cursor:
+                if bool(is_active) is False:
+                    continue
+
+                setting_value = getattr(settings, setting_name, None)
+                if setting_value is None:
+                    log.error(
+                        f'Failed to read setting value for {setting_name} in '
+                        'overwrites for accounting rules',
```
LefterisJP, line 197: remember to rename

rotkehlchen/db/accounting_rules.py
```diff
@@ -117,30 +131,85 @@ def update_accounting_rule(
...
+                    continue
+
+                if property_name == 'count_entire_amount_spend':
+                    old_value = rules[accountint_rule_id].rule.count_entire_amount_spend
+                    rules[accountint_rule_id].rule.count_entire_amount_spend = setting_value
+                elif property_name == 'count_cost_basis_pnl':
+                    old_value = rules[accountint_rule_id].rule.count_cost_basis_pnl
+                    rules[accountint_rule_id].rule.count_cost_basis_pnl = setting_value
+                else:
+                    log.error(f'Unknown accounting rule property {property_name}')
```
LefterisJP, line 208: Shouldn't you continue here?
```suggestion
                    log.error(f'Unknown accounting rule property {property_name}')
                    continue
```

rotkehlchen/db/schema.py
```diff
@@ -676,6 +676,16 @@
 );
 """
 
+DB_CREATE_MAPPED_ACCOUNTING_RULES = """
+CREATE TABLE IF NOT EXISTS linked_rules_properties(
+    identifier INTEGER PRIMARY KEY NOT NULL,
+    accounting_rule INTEGER REFERENCES accounting_rules(identifier),
+    property_name TEXT NOT NULL,
+    setting_name TEXT NOT NULL references settings(name),
+    active BOOLEAN
```
LefterisJP, line 685: What's the difference between this and `INTEGER NOT NULL CHECK (status IN (0, 1))`? I think there must have been a reason we are goign with the explicit `INTEGER NOT NULL CHECK (status IN (0, 1))` since it actively fails if you try to pass something like `2` there. Will the `BOOLEAN` also do that? I see nothing in the docs.

rotkehlchen/db/updates.py
```diff
@@ -166,6 +172,63 @@ def update_rpc_nodes(self, data: list[dict[str, Any]], version: int) -> None:
             new_default_nodes=new_default_nodes,
         )
 
+    def update_accounting_rules(self, data: list[dict[str, Any]], version: int) -> None:
```
LefterisJP, line 175: docstring please

rotkehlchen/db/updates.py
```diff
@@ -166,6 +172,63 @@ def update_rpc_nodes(self, data: list[dict[str, Any]], version: int) -> None:
...
+        log.info(f'Applying update for accounting rules to v{version}')
+        rules_db = DBAccountingRules(self.user_db)
+        for rule_data in data:
+            try:
+                event_type = HistoryEventType.deserialize(rule_data['event_type'])
```
LefterisJP, line 180: all these deserializations can also raise `DeserializatonError`. Do you want to handle this? Or let it propagate. The lack of a docstring does not let me understand the intentions.

rotkehlchen/db/updates.py
```diff
@@ -166,6 +172,63 @@ def update_rpc_nodes(self, data: list[dict[str, Any]], version: int) -> None:
...
+                counterparty = rule_data['counterparty'] if rule_data['counterparty'] is not None else NO_ACCOUNTING_COUNTERPARTY  # noqa: E501
+                rule = BaseEventSettings(
+                    taxable=rule_data['taxable'],
+                    count_entire_amount_spend=rule_data['count_entire_amount_spend'],
+                    count_cost_basis_pnl=rule_data['count_cost_basis_pnl'],
+                    method=rule_data['method'],
+                    accounting_treatment=TxAccountingTreatment.deserialize(rule_data['accounting_treatment']) if rule_data['accounting_treatment'] else None,  # noqa: E501
+                )
+            except KeyError as e:
+                log.error(f'Failed to read key {e} while iterating new accounting rules')
```
LefterisJP, line 191: Shouldn't you continue here?
```suggestion
                log.error(f'Failed to read key {e} while iterating new accounting rules')
                continue
```

rotkehlchen/db/updates.py
```diff
@@ -166,6 +172,63 @@ def update_rpc_nodes(self, data: list[dict[str, Any]], version: int) -> None:
...
+            except InputError as e:
+                # there is a conflict in the rule. Notify the frontend about it
+                rules, _ = rules_db.query_rules_and_serialize(
+                    filter_query=AccountingRulesFilterQuery.make(
+                        event_types=[event_type],
+                        event_subtypes=[event_subtype],
+                        counterparties=[counterparty],
+                    ),
+                )
+                if len(rules) == 1:
```
LefterisJP, line 209: If it's > 1 what happens? When can that happen?

rotkehlchen/db/updates.py
```diff
@@ -166,6 +172,63 @@ def update_rpc_nodes(self, data: list[dict[str, Any]], version: int) -> None:
...
+                        data={
+                            'local_rule': rules[0],
+                            'remote_rule': rule_data,
+                        },
+                    )
+
+                log.debug(f'Failed to add accounting rule {rule_data} due to {e}')
+                continue
+
+            if 'settings' not in rule_data:
```
LefterisJP, line 221: Isn't this an error? Should it not be handled somehow?

rotkehlchen/tests/db/test_db_accounting_rules.py
```diff
@@ -123,3 +125,52 @@ def test_errors_with_rules(database: DBHandler) -> None:
             event_subtype=HistoryEventSubType.RECEIVE,
             counterparty='uniswap',
         )
+
+
+@pytest.mark.parametrize('db_settings', [{'include_crypto2crypto': True}])
+@pytest.mark.parametrize('counterparty', ['yabir'])
+def test_accountint_rules_overwrite(database: 'DBHandler', counterparty: str) -> None:
```
LefterisJP, line 132: Typo. Also don't forget to change the name here about the overwrite thing

```suggestion
def test_accounting_rules_overwrite(database: 'DBHandler', counterparty: str) -> None:
```

LefterisJP (commented):
Some first comments

rotkehlchen/db/schema.py
yabirgb: yes we need to use the check. I recall this discussion before and I used boolean because I was thinking in sql and not sqlite. 

```
SQLite does not have a separate Boolean storage class. Instead, Boolean values are stored as integers 0 (false) and 1 (true).

SQLite recognizes the keywords "TRUE" and "FALSE", as of version 3.23.0 (2018-04-02) but those keywords are really just alternative spellings for the integer literals 1 and 0 respectively.
```

https://www.sqlite.org/datatype3.html

rotkehlchen/db/accounting_rules.py
yabirgb: I believe it can stay since esentially is the same

rotkehlchen/db/updates.py
yabirgb: it can be either 0 or 1 since the combination is unique

rotkehlchen/api/v1/schemas.py
```diff
@@ -3065,13 +3065,18 @@ class AccountingRuleIdSchema(Schema):
     counterparty = fields.String(required=False, load_default=None)
 
 
+class LinkedAccountingSetting(Schema):
```
LefterisJP, line 3068: Small docstring explaining the schema. What is the value and how is it linked to the linked setting

rotkehlchen/api/v1/schemas.py
```diff
@@ -3081,16 +3086,22 @@ def _create_settings(
     ) -> dict[str, Any]:
         rule = BaseEventSettings(
             taxable=data['taxable'],
-            count_entire_amount_spend=data['count_entire_amount_spend'],
-            count_cost_basis_pnl=data['count_cost_basis_pnl'],
-            method=data['method'],
+            count_entire_amount_spend=data['count_entire_amount_spend']['value'],
+            count_cost_basis_pnl=data['count_cost_basis_pnl']['value'],
+            method='spend',  # TODO 1.31: REMOVE THE METHOD field
```
LefterisJP, line 3091: Weren't you supposed to remove this already? Or just moving it to a later PR?

rotkehlchen/db/accounting_rules.py
```diff
@@ -57,39 +83,42 @@ def add_accounting_rule(
...
+                self.add_linked_setting(
+                    write_cursor=write_cursor,
+                    rule_identifier=inserted_rule_id,
+                    rule_property=property_name,
+                    setting_name=setting_name,
+                )
+
+            return inserted_rule_id
+
+    def remove_accounting_rule(self, rule_id: int) -> None:
```
LefterisJP, line 97: update docstring

rotkehlchen/db/accounting_rules.py
LefterisJP: you are right. I saw it all get back to one line and thought the second part of the tuple, the int was removed.

rotkehlchen/db/accounting_rules.py
```diff
@@ -117,30 +146,96 @@ def update_accounting_rule(
...
+                )
                 for entry in cursor
-            ]
+            }
             query, bindings = filter_query.prepare(with_pagination=False)
             query = 'SELECT COUNT(*) from accounting_rules ' + query
             total_found_result = cursor.execute(query, bindings).fetchone()[0]
-        return rules, total_found_result
+
+            # check the settings linked for the rule
```
LefterisJP, line 211:
```suggestion
            # check the settings linked to the rule
```

rotkehlchen/db/updates.py
```diff
@@ -166,6 +172,58 @@ def update_rpc_nodes(self, data: list[dict[str, Any]], version: int) -> None:
...
+        for rule_data in data:
+            try:
+                event_type = HistoryEventType.deserialize(rule_data['event_type'])
+                event_subtype = HistoryEventSubType.deserialize(rule_data['event_subtype'])
+                counterparty = rule_data['counterparty'] if rule_data['counterparty'] is not None else NO_ACCOUNTING_COUNTERPARTY  # noqa: E501
+                rule = BaseEventSettings(
+                    taxable=rule_data['taxable'],
+                    count_entire_amount_spend=rule_data['count_entire_amount_spend'],
+                    count_cost_basis_pnl=rule_data['count_cost_basis_pnl'],
+                    method=rule_data['method'],
```
LefterisJP, line 191: guess this also needs to go away, no?

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -121,6 +123,42 @@
...
+            'event_subtype': HistoryEventSubType.FEE.serialize(),
+            'counterparty': 'test_counterparty',
+            'taxable': True,
+            'count_entire_amount_spend': True,
+            'count_cost_basis_pnl': True,
+            'method': 'spend',
+            'accounting_treatment': None,
+            'links': {'count_cost_basis_pnl': 'include_crypto2crypto'},
+        },
+        {
```
LefterisJP, line 139: please do this over this file and other places you do this. Saves space

```suggestion
        }, {
```

LefterisJP (commented):
Small stuff

LefterisJP (approved):
LGTM.

Checked all 2 test failures and seems both were due to connection timeout / remote hiccups:

```
WARNING  rotkehlchen.chain.evm.node_inquirer:node_inquirer.py:532 Greenlet-184: Failed to query NodeName(name='etherscan', endpoint='', owned=False, blockchain=<SupportedBlockchain.ETHEREUM: 'ETH'>) for <bound method EvmNodeInquirer._call_contract of <rotkehlchen.chain.ethereum.node_inquirer.EthereumInquirer object at 0x7f0737dd7370>> due to ethereum Etherscan API request failed due to HTTPSConnectionPool(host='api.etherscan.io', port=443): Read timed out. (read timeout=30)
ERROR    rotkehlchen.chain.evm.node_inquirer:node_inquirer.py:543 Greenlet-184: Failed to query <bound method EvmNodeInquirer._call_contract of <rotkehlchen.chain.ethereum.node_inquirer.EthereumInquirer object at 0x7f0737dd7370>> after trying the following nodes: ['etherscan']
```

PR #6740, opened 2023-10-12
Fix substrate tests and subscan bug

LefterisJP:
### [Fix bug with substrate manager subscan query](https://github.com/rotki/rotki/commit/cb55fdad64ce05681d3f391fdd8e83f95e0cb640) 

Also catch AttributeError as potential error from the underlying
library. They really do no error handling ...

### [Fix substrate tests and fixtures. VCR them too.](https://github.com/rotki/rotki/commit/f83ffce9f7fa6062277e2a8224b1e1d45c7ba1a4) 

- Due to the way the tests were written the
wait_until_all_substrate_nodes_connected always timed out. Since at
the point of the fixture there was no kusama accounts for some tests.
- It also tried to connect to all kusama nodes, of which onfinality is
being bad right now and 409s almost everything. On top of that they
ran with the flaky modifier which means they did not fail but retry up
to 3 times
- VCRed them and adjusted fixtures so all these problems go away

### DIFF
--- a/rotkehlchen/chain/substrate/manager.py
+++ b/rotkehlchen/chain/substrate/manager.py
@@ -406,6 +406,7 @@ class SubstrateManager:
                 SubstrateRequestException,
                 WebSocketException,
                 ValueError,
+                AttributeError,
         ) as e:
             message = (
                 f'{self.chain} failed to request last block '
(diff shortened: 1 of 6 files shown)


yabirgb (approved):
lgtm

PR #6743, opened 2023-10-13
feat(frontend): modify all history events

lukicenturi:
Closes #(issue_number)

## Checklist

- [x] Functionality (add / edit / delete)
- [x] Add more test for the form


frontend/app/src/components/display/DateDisplay.vue
```diff
@@ -37,30 +58,44 @@ const displayTimestamp = computed<number>(() =>
...
 });
+
+const splittedByMillisecondsPart = computed(() =>
+  get(formattedDate).split('.')
+);
 </script>
 
 <template>
   <span>
     <VTooltip top open-delay="400" :disabled="!showTooltip">
```
kelsos, line 84: any chance we can RuiTooltip this? or is there a problem with that?

frontend/app/package.json
```diff
@@ -58,6 +58,7 @@
     "eslint-plugin-vue": "9.17.0",
     "eslint-plugin-vuetify": "1.1.0",
     "imask": "6.6.2",
+    "json-editor-vue": "0.10.19",
```
kelsos, line 61: do you think it would be too much effort to port this to our codebase and use `vanilla-jsoneditor` directly? It seems that the whole library is a single class wrapper around `vanilla-jsoneditor` either way.

frontend/app/src/components/history/events/HistoryEventType.vue
```diff
@@ -7,6 +7,7 @@ import {
   isEvmEventRef,
   isOnlineHistoryEventRef
 } from '@/utils/history/events';
+import HistoryEventTypeCombination from '@/components/history/events/HistoryEventTypeCombination.vue';
```
kelsos, line 10: nitpick: we should not have to explicitly import this right?

frontend/app/src/components/history/events/HistoryEventsAction.vue
```diff
@@ -66,26 +65,28 @@ const resetEvent = (event: EvmHistoryEvent) => emit('reset', event);
...
-          </VListItemContent>
-        </VListItem>
+        <template v-if="evmEvent">
+          <VListItem
+            link
+            :disabled="loading"
+            @click="redecode(toEvmChainAndTxHash(evmEvent))"
+          >
+            <VListItemIcon class="mr-4">
+              <VIcon>mdi-database-refresh</VIcon>
```
kelsos, line 75: nitpick: any chance we can also migrate the icons here?

frontend/app/src/components/history/events/HistoryEventsAction.vue
```diff
@@ -66,26 +65,28 @@ const resetEvent = (event: EvmHistoryEvent) => emit('reset', event);
...
+            </VListItemIcon>
+            <VListItemContent>
+              {{ t('transactions.actions.redecode_events') }}
+            </VListItemContent>
+          </VListItem>
+          <VListItem link :disabled="loading" @click="resetEvent(evmEvent)">
+            <VListItemIcon class="mr-4">
+              <VIcon>mdi-file-restore</VIcon>
```
kelsos, line 83: nitpick-2: and here too? or does it create a problem with `VListIcon`

frontend/app/src/components/history/events/HistoryEventsView.vue
```diff
@@ -447,7 +447,7 @@ watch(
 
 const premium = usePremium();
 const { isLoading: isSectionLoading } = useStatusStore();
-const sectionLoading = isSectionLoading(Section.TX);
+const sectionLoading = isSectionLoading(Section.HISTORY_EVENT);
 const eventTaskLoading = isTaskRunning(TaskType.TX_EVENTS);
```
kelsos, line 451: Since you are renaming the section property maybe it would also make sense to also update this here?

frontend/app/src/components/history/events/forms/EthBlockEventForm.vue
```diff
@@ -0,0 +1,292 @@
...
+import { HistoryEventEntryType } from '@rotki/common/lib/history/events';
+import { helpers, required } from '@vuelidate/validators';
+import { Blockchain } from '@rotki/common/lib/blockchain';
+import {
+  type EthBlockEvent,
+  type NewEthBlockEventPayload
+} from '@/types/history/events';
+import { type Writeable } from '@/types';
+import { toMessages } from '@/utils/validation';
+import HistoryEventAssetPriceForm from '@/components/history/events/forms/HistoryEventAssetPriceForm.vue';
```
kelsos, line 12: nitpick: auto-imported component

frontend/app/src/components/history/events/forms/EthDepositEventForm.vue
```diff
@@ -0,0 +1,361 @@
...
+};
+
+const { setMessage } = useMessageStore();
+
+const { editHistoryEvent, addHistoryEvent } = useHistoryTransactions();
+
+const save = async (): Promise<boolean> => {
+  const timestamp = convertToTimestamp(get(datetime));
+
+  const payload: Writeable<NewEthDepositEventPayload> = {
```
kelsos, line 175: same here if we don't modify the payload after the assignment no need for it to be writable afterwards

frontend/app/src/components/history/events/forms/EthWithdrawalEventForm.vue
```diff
@@ -0,0 +1,270 @@
...
+  set(withdrawalAddress, entry.locationLabel ?? '');
+  set(validatorIndex, entry.validatorIndex.toString());
+  set(datetime, convertFromTimestamp(entry.timestamp));
+};
+
+const { setMessage } = useMessageStore();
+
+const { editHistoryEvent, addHistoryEvent } = useHistoryTransactions();
+
+const save = async (): Promise<boolean> => {
```
kelsos, line 131: So question, how much of the save method, and check props repeats in different components? 

in save the only thing that seems different is the payload.

Do you think we can move the common part to a composable and only pass the pass that differs?

frontend/app/src/components/history/events/forms/EthBlockEventForm.vue
lukicenturi: it is being used as type

frontend/app/src/components/inputs/DateTimePicker.vue
```diff
@@ -43,55 +43,46 @@ const props = withDefaults(
...
-    isValidDate(date, dateFormat) ||
-    isValidDate(date, dateTimeFormatVal) ||
-    (!get(seconds) && isValidDate(date, completeDateTimeFormatVal))
-  );
-};
+const isValidFormat = (date: string): boolean =>
+  isValidDate(date, get(dateOnlyFormat)) ||
+  isValidDate(date, get(dateTimeFormat)) ||
+  isValidDate(date, get(dateTimeFormatWithSecond)) ||
+  (get(milliseconds) && isValidDate(date, get(dateTimeFormatWithMilliseconds)));
```
kelsos, line 77: i wonder if there is a way to simplify this code

frontend/app/src/components/inputs/DateTimePicker.vue
```diff
@@ -343,62 +351,68 @@ const focus = () => {
...
+        transition="scale-transition"
+        :nudge-bottom="56"
+        left
+        max-width="580px"
+        class="date-time-picker"
+      >
+        <template #activator="{ on }">
+          <VBtn icon class="mt-n2" v-on="on">
+            <VIcon>mdi-earth</VIcon>
+          </VBtn>
```
kelsos, line 390: nitpick: migrate icons/buttons?

frontend/app/src/types/history/events/index.ts
```diff
@@ -14,7 +14,7 @@ const LiquityStakingEventExtraData = z.object({
 // ETH Transactions
 export const EvmTransaction = z.object({
   txHash: z.string(),
-  timestamp: z.number(),
+  timestamp: z.number().transform(timestamp => timestamp / 1000),
```
kelsos, line 17: Do we use these timestamps for editing later?
This transformation is converting the timestamp to seconds right?

frontend/app/src/utils/date.ts
```diff
@@ -24,39 +24,44 @@ export function changeDateFormat(
...
+
+      const thirdSplit = secondSplit[2].split('.');
+      if (thirdSplit.length === 2) {
+        format += '.SSS';
+      }
     }
   }
 
-  return dayjs(date, format).unix();
+  return dayjs(date, format).valueOf() / 1000;
```
kelsos, line 52: `unix()` returns the epoch in seconds 
`valueOf()` returns the epoch in ms

1 second is 1000 ms so if you do valueOf()/1000 you get the epoch in seconds, which is what unix does so why change it?

frontend/app/src/types/history/events/index.ts
lukicenturi: yes true, this way I can make the input consistent.
Ideally I think the backend should send everything in milliseconds.

frontend/app/tests/unit/specs/components/history/events/eth-block-event-form.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  const createWrapper = (options: ThisTypedMountOptions<any> = {}) => {
+    const vuetify = new Vuetify();
+    const pinia = createPinia();
+    setActivePinia(pinia);
+    return mount(EthBlockEventForm, {
+      pinia,
+      vuetify,
+      stubs: {
+        VAutocomplete: {
```
kelsos, line 45: Is this stub the same across different specs? if so do you think we can extract it somewhere?

kelsos (approved):
lgtm, if anything else remains we will fix it while testing

PR #6753, opened 2023-10-14
Gitcoin allo protocol payout events will be properly decoded

LefterisJP:
This unfortunately requires, just like with the rounds, maintaining a list of payout_strategy implementations (which can be derived from the round implementation contracts themselves)

Looks like this


LefterisJP (commented):
single failing test is flaky, probably polygon pos node in the test can use different modes. Unrelated to the diff and passes locally so I am merging.

PR #6754, opened 2023-10-15
Bump some packages

LefterisJP:


yabirgb:
Seems that linting has an issue with the new packages

LefterisJP:
> Seems that linting has an issue with the new packages

It's probably webargs. Checked the changelog. Will need more work on my side.

LefterisJP:
Finally all green. Will merge.

PR #6761, opened 2023-10-16
gnosis  xdai bridge

jbrit:
## Checklist

- [ ] The PR modified the backend to include proper decoding for the gnosis xdci bridge.


rotkehlchen/tests/unit/decoders/test_gnosis_bridge.py
```diff
@@ -0,0 +1,153 @@
...
+from rotkehlchen.types import Location, TimestampMS, deserialize_evm_tx_hash
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('ethereum_accounts', [['0x4c1a316De360E08817eB88dD31A0E7305005fB65']])
+def test_bridge_dai_from_ethereum(
+        database,
+        ethereum_inquirer,
+        ethereum_accounts,
+):
```
yabirgb, line 24: this fits in one line. same for the other functions

```suggestion
def test_bridge_dai_from_ethereum(database, ethereum_inquirer, ethereum_accounts):
```

rotkehlchen/chain/gnosis/modules/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,104 @@
...
+
+BRIDGE_ADDRESS = string_to_evm_address('0x7301CFA0e1756B71869E93d4e4Dca5c7d0eb0AA6')
+BRIDGE_TOKEN = b'\x12vP\xbc\xfb\x0b\xa0\x17@\x1a\xbeI1E:@Q@\xa8\xfd6\xfe\xceg\xba\xe2\xdb\x17M?\xddc'  # noqa: E501
+
+
+class XdaiBridgeDecoder(DecoderInterface):
+
+    def __init__(
+            self,
+            ethereum_inquirer: 'EthereumInquirer',
```
yabirgb, line 34: this should be `GnosisInquirer`

```suggestion
            gnosis_inquirer: 'GnosisInquirer',
```

rotkehlchen/chain/gnosis/modules/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,104 @@
...
+from rotkehlchen.utils.misc import hex_or_bytes_to_address, hex_or_bytes_to_int
+
+if TYPE_CHECKING:
+    from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
+    from rotkehlchen.chain.evm.decoding.base import BaseDecoderTools
+    from rotkehlchen.user_messages import MessagesAggregator
+
+
+BRIDGE_ADDRESS = string_to_evm_address('0x7301CFA0e1756B71869E93d4e4Dca5c7d0eb0AA6')
+BRIDGE_TOKEN = b'\x12vP\xbc\xfb\x0b\xa0\x17@\x1a\xbeI1E:@Q@\xa8\xfd6\xfe\xceg\xba\xe2\xdb\x17M?\xddc'  # noqa: E501
```
yabirgb, line 27: in the bridge for what I saw you can only bridge xdai so maybe to avoid confusions is better just

```suggestion
BRIDGE_XDAI= b'\x12vP\xbc\xfb\x0b\xa0\x17@\x1a\xbeI1E:@Q@\xa8\xfd6\xfe\xceg\xba\xe2\xdb\x17M?\xddc'  # noqa: E501
```

rotkehlchen/chain/gnosis/modules/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,104 @@
...
+    ) -> None:
+        super().__init__(
+            evm_inquirer=ethereum_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+        )
+        self.xdai = A_XDAI.resolve_to_crypto_asset()
+
+    def _decode_bridge_tokens(self, context: DecoderContext) -> DecodingOutput:
+        """Decodes a bridging event for xdai. Either a deposit or a withdrawal."""
```
yabirgb, line 46: currently you decode only one right? state it in the docstring that the other is pending

rotkehlchen/chain/gnosis/modules/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,104 @@
...
+        to_address = BRIDGE_ADDRESS
+        amount = asset_normalized_value(
+            amount=hex_or_bytes_to_int(context.tx_log.data[32:64]),
+            asset=self.xdai,
+        )
+        expected_event_type, new_event_type, from_chain, to_chain, expected_location_label = bridge_prepare_data(  # noqa: E501
+            tx_log=context.tx_log,
+            deposit_events=(BRIDGE_TOKEN,),
+            main_chain=ChainID.GNOSIS,
+            l2_chain=ChainID.ETHEREUM,
```
yabirgb, line 60: @LefterisJP I found this name confusing. Maybe could we rename this to `target_chain` or anything better than `l2`??

rotkehlchen/chain/ethereum/modules/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,108 @@
...
+            evm_inquirer=evm_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+        )
+        self.dai = A_DAI.resolve_to_crypto_asset()
+
+    def _decode_bridge_tokens(self, context: DecoderContext) -> DecodingOutput:
+        """Decodes a bridging event for dai. Either a deposit or a withdrawal."""
+        if context.tx_log.topics[0] not in {BRIDGE_TOKEN, WITHDRAWAL_TOKEN}:
+            return DEFAULT_DECODING_OUTPUT
```
yabirgb, line 48: it can only bridge dai right? don't mention tokens since it can be confusing

rotkehlchen/chain/ethereum/modules/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,108 @@
...
+        super().__init__(
+            evm_inquirer=evm_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+        )
+        self.dai = A_DAI.resolve_to_crypto_asset()
+
+    def _decode_bridge_tokens(self, context: DecoderContext) -> DecodingOutput:
+        """Decodes a bridging event for dai. Either a deposit or a withdrawal."""
+        if context.tx_log.topics[0] not in {BRIDGE_TOKEN, WITHDRAWAL_TOKEN}:
```
yabirgb, line 47: The logic seems really similar to the one in the gnosis chain all the `bridge_prepare_data` and the for loop logic is identical I would say with the exception on the direction. I believe you can abstract this class (maybe move it to the gnosis decoder side in a new file and then create subclasses for ethereum and gnosis.

Things that change are:

- the topics that will be allowed (this can be an attribute of the subclass
- you can create a method to extract the `from_address/to_address`
- the `main_chain`/`l2_chain` can be attributes of the subclass (I believe you only need to provice as argument the l2_chain since the main_chain is in the inquirer.chain_id or something like this)

yabirgb (commented):
Main concern is that there is code that can be abstracted

rotkehlchen/chain/gnosis/modules/xdai_bridge/decoder.py
LefterisJP: In that case I would use source/target. And not main/target. Does that make sense?

rotkehlchen/chain/evm/decoding/xdai_bridge/constants.py
```diff
@@ -0,0 +1 @@
+CPT_COWSWAP = 'cowswap'
```
yabirgb, line 1: This is incorrect

rotkehlchen/chain/evm/decoding/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,117 @@
...
+            to_address = hex_or_bytes_to_address(context.tx_log.data[0:32])
+        amount = asset_normalized_value(
+            amount=hex_or_bytes_to_int(context.tx_log.data[32:64]),
+            asset=self.bridged_asset.resolve_to_crypto_asset(),
+        )
+        expected_event_type, new_event_type, from_chain, to_chain, expected_location_label = bridge_prepare_data(  # noqa: E501
+            tx_log=context.tx_log,
+            deposit_events=(self.deposit_event,),
+            main_chain=self.source_chain,
+            l2_chain=self.target_chain,
```
yabirgb, line 73: Can you maybe adjust to what lefteris suggested here?

https://github.com/rotki/rotki/pull/6761#discussion_r1363738777

yabirgb (commented):
Two smaller things

rotkehlchen/chain/evm/decoding/constants.py
```diff
@@ -39,3 +40,8 @@
     label='Gitcoin',
     image='gitcoin.svg',
 )
+XDAI_BRIDGE_CPT_DETAILS = CounterpartyDetails(
+    identifier=CPT_XDAI_BRIDGE,
+    label='Xdai',
```
LefterisJP, line 45:
```suggestion
    label='xDai Bridge',
```

rotkehlchen/chain/evm/decoding/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,117 @@
...
+
+class XdaiBridgeCommonDecoder(DecoderInterface, metaclass=abc.ABCMeta):
+
+    def __init__(
+            self,
+            evm_inquirer: 'EvmNodeInquirer',
+            base_tools: 'BaseDecoderTools',
+            msg_aggregator: 'MessagesAggregator',
+            deposit_event: bytes,
+            withdrawal_event: Optional[bytes],  # withdrawal is currently unsupported on gnosis
```
LefterisJP, line 33:
```suggestion
            deposit_topic: bytes,
            withdrawal_topic: Optional[bytes],  # withdrawal is currently unsupported on gnosis
```

rotkehlchen/chain/evm/decoding/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,117 @@
...
+        self.bridge_address = bridge_address
+        self.bridged_asset = bridged_asset
+        self.deposit_event = deposit_event
+        self.withdrawal_event = withdrawal_event
+        self.source_chain = source_chain
+        self.target_chain = target_chain
+
+    def _decode_bridged_asset(self, context: DecoderContext) -> DecodingOutput:
+        """
+        Decodes a bridging event for `bridged_asset` specified.
```
LefterisJP, line 53:
```suggestion
        Decodes a bridging event for the `bridged_asset`.
```

rotkehlchen/chain/evm/decoding/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,117 @@
...
+        Either a deposit or a withdrawal.
+        """
+        if context.tx_log.topics[0] not in {self.deposit_event, self.withdrawal_event}:
```
LefterisJP, line 56: Creating a set is more expensive than creating a tuple. Keep it simple unless this set later needs to grow, which it does not

```suggestion
        if context.tx_log.topics[0] not in (self.deposit_event, self.withdrawal_event):
```

rotkehlchen/chain/evm/decoding/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,117 @@
...
+            return DEFAULT_DECODING_OUTPUT
+
+        if context.tx_log.topics[0] == self.deposit_event:
```
LefterisJP, line 59: Don't do it like that then. You have one unnecessary conditional check.

```python
if context.tx_log.topics[0] == self.deposit_event
    pass
elif context.tx_log.topics[0] == self.withdrawal_event:
   pass
else:

File: rotkehlchen/chain/evm/decoding/xdai_bridge/decoder.py [hunk 15]
```diff
@@ -0,0 +1,117 @@
...
+    ) -> None:
+        super().__init__(
+            evm_inquirer=evm_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+        )
+        self.bridge_address = bridge_address
+        self.bridged_asset = bridged_asset
+        self.deposit_event = deposit_event
+        self.withdrawal_event = withdrawal_event
```
LefterisJP, line 47:
```suggestion
        self.deposit_topict = deposit_event
        self.withdrawal_topic = withdrawal_event
```

rotkehlchen/chain/evm/decoding/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,117 @@
...
+
+        if context.tx_log.topics[0] == self.deposit_event:
+            from_address = context.transaction.from_address
+            to_address = self.bridge_address
+        else:
+            from_address = self.bridge_address
+            to_address = hex_or_bytes_to_address(context.tx_log.data[0:32])
+        amount = asset_normalized_value(
+            amount=hex_or_bytes_to_int(context.tx_log.data[32:64]),
+            asset=self.bridged_asset.resolve_to_crypto_asset(),
```
LefterisJP, line 67: move this resolution to the constructor please. It's silly to resolve it every time this function runs.

rotkehlchen/chain/evm/decoding/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,117 @@
...
+                    from_address=context.transaction.from_address,
+                    to_address=hex_or_bytes_to_address(context.tx_log.data[0:32]),
+                    from_chain=from_chain,
+                    to_chain=to_chain,
+                    amount=event.balance.amount,
+                    asset=self.bridged_asset.resolve_to_asset_with_symbol(),
+                    expected_event_type=expected_event_type,
+                    new_event_type=new_event_type,
+                    counterparty=XDAI_BRIDGE_CPT_DETAILS,
+                )
```
LefterisJP, line 96: here you should break if it's found. Also add an else clause that logs an error if the decoder could not find the matching transfer. There is many other decoders we do this in.
```suggestion
                )
                break
        else:
            log.error('')
```

rotkehlchen/chain/gnosis/modules/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,36 @@
...
+            gnosis_inquirer: 'GnosisInquirer',
+            base_tools: 'BaseDecoderTools',
+            msg_aggregator: 'MessagesAggregator',
+    ) -> None:
+        super().__init__(
+            evm_inquirer=gnosis_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+            deposit_event=BRIDGE_XDAI,
+            withdrawal_event=None,  # TODO: decode bridge withdrawal, currently unsupported by gnosisscan  # noqa: E501
```
LefterisJP, line 31: Add link to the note/issue you made for this

rotkehlchen/chain/evm/decoding/utils.py
```diff
@@ -102,7 +102,8 @@ def bridge_match_transfer(
     event.event_type = new_event_type
     event.event_subtype = HistoryEventSubType.BRIDGE
     event.counterparty = counterparty.identifier
+    bridge_suffix = '' if counterparty.label.endswith('bridge') else ' bridge'
```
LefterisJP, line 105: I would actually change this and make sure all counterparties passed here are consistent. Either they are all bridges or they are all the protocol without a bridge.

To do that remove the existence of xDAI Bridge as protocol. It's the old name anyway. Use `CPT_GNOSIS_CHAIN` as the protocol to pass for the bridge and let this be as it was.

rotkehlchen/chain/evm/decoding/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,125 @@
...
+    def _decode_bridged_asset(self, context: DecoderContext) -> DecodingOutput:
+        """Decodes a bridging event for the `bridged_asset`, either a deposit or a withdrawal."""
+        if context.tx_log.topics[0] == self.deposit_topic:
+            from_address = context.transaction.from_address
+            to_address = self.bridge_address
+        elif context.tx_log.topics[0] == self.withdrawal_topic:
+            from_address = self.bridge_address
+            to_address = hex_or_bytes_to_address(context.tx_log.data[0:32])
+        else:
+            return DEFAULT_DECODING_OUTPUT
```
LefterisJP, line 67: You need empty line here to separate code areas

```suggestion
            return DEFAULT_DECODING_OUTPUT

```

rotkehlchen/chain/evm/decoding/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,125 @@
...
+        amount = asset_normalized_value(
+            amount=hex_or_bytes_to_int(context.tx_log.data[32:64]),
+            asset=self.bridged_crypto_asset,
+        )
+        expected_event_type, new_event_type, from_chain, to_chain, expected_location_label = bridge_prepare_data(  # noqa: E501
+            tx_log=context.tx_log,
+            deposit_events=(self.deposit_topic,),
```
LefterisJP, line 74: can you also change deposit_events to deposit_topics in this function?

rotkehlchen/chain/evm/decoding/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,125 @@
...
+                    from_chain=from_chain,
+                    to_chain=to_chain,
+                    amount=event.balance.amount,
+                    asset=self.bridged_asset_with_symbol,
+                    expected_event_type=expected_event_type,
+                    new_event_type=new_event_type,
+                    counterparty=XDAI_BRIDGE_CPT_DETAILS,
+                )
+                return DEFAULT_DECODING_OUTPUT
+        log.error(
```
LefterisJP, line 101: Why did you not do it as I said. This way you now have two returns. Please do what I said with the break and the else in the for.

rotkehlchen/chain/evm/decoding/xdai_bridge/decoder.py
```diff
@@ -0,0 +1,125 @@
...
+    ) -> None:
+        super().__init__(
+            evm_inquirer=evm_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+        )
+        self.bridge_address = bridge_address
+        self.bridged_asset = bridged_asset
+        self.bridged_crypto_asset = bridged_asset.resolve_to_crypto_asset()
+        self.bridged_asset_with_symbol = bridged_asset.resolve_to_asset_with_symbol()
```
LefterisJP, line 52: What the fuck?

These are all the same. Why would you keep it 3 times in memory. Just save it resolved, to the type of thing you need. A `CryptoAsset` is an `AssetWithSymbol`

```suggestion
        self.bridged_asset = bridged_asset.resolve_to_crypto_asset()
```

rotkehlchen/chain/evm/decoding/xdai_bridge/decoder.py
jbrit: the break and else in the for doesn't work, that's why I did this.

jbrit: if I do an else, it returns the log for every event that's not the transfer event

LefterisJP (approved):
LGTM. Small nitpick but I can handle it in a followup. 

The failing test is due to basescan being unreachable. Fails in develop also when ran locally.

```
ERROR    rotkehlchen.chain.aggregator:aggregator.py:1419 Greenlet-420: base Etherscan API request failed due to HTTPSConnectionPool(host='api.basescan.org', port=443): Read timed out. (read timeout=30) when checking if 0xc37b40ABdB939635068d3c5f13E7faF686F03B65 is active at base
```

PR #6765, opened 2023-10-17
Fix editing of all history events

LefterisJP:
- Make history events serialize_for_db() method more modular and let it know what queries to use.
- Adjust tests so that all history event editing is tested
- Make sure that event identifier can be edited for all event types
----

Some general notes here:

1. It seems that when editing events at the moment we still adhere to the constructor values. Not sure this is correct as an approach as from the DB editor it's possible to edit everything and perhaps we may end up needing to. This needs more thinking.
2. The tests are convoluted and written terribly. Exactly because of this it took me a lot of time to adjust and fix. They should be rewritten properly and have different test per event type, not this salad of code that's impossible to untangle right now.
3. The `deserialize_from_db()` code is weird and seems to try to edit specific tuples out of the full result which does not make much sense. Perhaps writing a query per event type is simpler here too and more easy to maintain. Let's see.

### DIFF
--- a/rotkehlchen/accounting/structures/base.py
+++ b/rotkehlchen/accounting/structures/base.py
@@ -158,26 +158,38 @@ class HistoryBaseEntry(AccountingEventMixin, metaclass=ABCMeta):
             f'{self.identifier=}',
         ]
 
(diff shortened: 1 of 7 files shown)


LefterisJP (commented):
nobody complained, all tests green so merging

PR #6760, opened 2023-10-16
Add direction to history events

d347h-eth:
Closes [Add in/out information for pairs of event type and subtype](https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=41588023)

PR #6777, opened 2023-10-19
Remove method field

yabirgb:
This PR requires #6760 merged

Removes the method field from the accounting rules and adjusts the accounting logic to use the new directions in the possible events


rotkehlchen/accounting/history_base_entries.py
```diff
@@ -125,7 +150,7 @@ def process(
             )
 
         self.pot.add_asset_change_event(
-            method=event_settings.method,
+            method=method,
```
LefterisJP, line 153: don't do it like that. Don't re-create the old code. Use the `EventDirection` instead here

rotkehlchen/accounting/constants.py
```diff
@@ -10,6 +11,7 @@
 FREE_PNL_EVENTS_LIMIT = 1000
 FREE_REPORTS_LOOKUP_LIMIT = 20
 
+ACCOUNTING_METHOD_TYPE = Literal['spend', 'acquisition']
```
LefterisJP, line 14: This is no longer needed right?

LefterisJP (approved):
LGTM. All green. I left a comment for an unused variable which can probably be removed by you or me later. Wll merge due to green tests

PR #6779, opened 2023-10-19
Fix asset resolution problems

LefterisJP:
### [Fix resolve_asset_to_class. If type in DB is not what we ask bail.](https://github.com/rotki/rotki/commit/aa6673236fe1db34c06586749a013875a9d88d5c) 

And by bailing earlier we don't end up overwriting the user's global
DB with the packaged global DB.

What was happening was that at some point all assets were attempted to
be resolved to FIAT assets. Even if they were not FIAT assets. This
basically got us in this code for each asset, and if the asset was in
the constant assets it automatically replaced itself with what was in
packaged DB.

Now this will no longer happen. Only if asset is not in the global DB
or is not what was expected and in the packaged DB it is, will the
replacement happen.

### [When querying current prices use asset resolution properly](https://github.com/rotki/rotki/commit/f5c39e3edcd8ffe01f1cec97add7cfa4b7f196f7) 

Before the code was doing asset resolution as an attempt to try and
see if it's a certain type of asset. For example for FiatAsset, or
EvmToken and used some kind of duck typing with exception suppression.

That's super ugly and hid a bug that made us overwrite user's
globalDB. (see previous commit)

Now instead of doing that we just resolve the asset once and then act
according to type.

### DIFF
--- a/rotkehlchen/inquirer.py
+++ b/rotkehlchen/inquirer.py
@@ -632,53 +632,57 @@ class Inquirer:
             if cache is not None:
                 return cache.price, cache.oracle, cache.used_main_currency
 
-        with suppress(UnknownAsset, RemoteError, WrongAssetType):
-            asset = asset.resolve_to_fiat_asset()
-            price, oracle = instance._query_fiat_pair(base=asset, quote=instance.usd)
-            return price, oracle, False
+        try:
+            asset = asset.resolve()
+        except UnknownAsset:
+            log.error(f'Tried to ask for {asset.identifier} price but asset is missing from the DB')  # noqa: E501
+            return ZERO_PRICE, CurrentPriceOracle.FIAT, False
+
+        # if asset.asset_type == AssetType.FIAT:  # type: ignore  # class is fully resolved here
+        if isinstance(asset, FiatAsset):
+            with suppress(RemoteError):
+                price, oracle = instance._query_fiat_pair(base=asset, quote=instance.usd)
+                return price, oracle, False
 
         # continue, asset isn't fiat or a price can be found by one of the oracles (CC for example)
 
         # Try and check if it is an ethereum token with specified protocol or underlying tokens
         is_known_protocol = False
         underlying_tokens = None
-        with suppress(UnknownAsset, WrongAssetType):
-            token = asset.resolve_to_evm_token()
-            if token.protocol is not None:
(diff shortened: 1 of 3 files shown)


LefterisJP:
The failed test is again test_detect_evm_accounts due to basescan timeout.

Same happened last night. We may need to look it up and see what's happening there:

```
ERROR    rotkehlchen.chain.aggregator:aggregator.py:1419 Greenlet-420: base Etherscan API request failed due to HTTPSConnectionPool(host='api.basescan.org', port=443): Read timed out. (read timeout=30) when checking if 0xc37b40ABdB939635068d3c5f13E7faF686F03B65 is active at base
```

rotkehlchen/inquirer.py
```diff
@@ -632,53 +632,57 @@ def _find_usd_price(
...
-            asset = asset.resolve_to_fiat_asset()
-            price, oracle = instance._query_fiat_pair(base=asset, quote=instance.usd)
-            return price, oracle, False
+        try:
+            asset = asset.resolve()
+        except UnknownAsset:
+            log.error(f'Tried to ask for {asset.identifier} price but asset is missing from the DB')  # noqa: E501
+            return ZERO_PRICE, CurrentPriceOracle.FIAT, False
+
+        # if asset.asset_type == AssetType.FIAT:  # type: ignore  # class is fully resolved here
```
yabirgb, line 641: left this commented code

rotkehlchen/inquirer.py
```diff
@@ -632,53 +632,57 @@ def _find_usd_price(
...
+                    ethereum=ethereum.node_inquirer,  # type:ignore  # ethereum is an EthereumManager so the inquirer is of the expected type
+                    token=asset,
+                    underlying_asset_price=underlying_asset_price,
                 )
-                return usd_price, oracle, False
-            # else known protocol on-chain query failed. Continue to external oracles
+                if usd_price is None:
+                    price = ZERO_PRICE
+                else:
+                    price = Price(usd_price)
```
yabirgb, line 669:
```suggestion
                price = ZERO_PRICE if usd_price is None else Price(usd_price)
```

LefterisJP: yep good suggestion. I did not change this, it was like that before, just moved it around.

PR #6781, opened 2023-10-19
Remove ResolvedAsset. Seems like a useless abstraction.

LefterisJP:


LefterisJP:
Both errors are due to external apis and also happen in develop

1. `rotkehlchen/tests/api/blockchain/test_evm.py::test_detect_evm_accounts` due to basescan
2. `rotkehlchen/tests/exchanges/test_iconomi.py::test_iconomi_assets_are_known` due to iconomi api being down

PR #6785, opened 2023-10-20
Fix data migration version

yabirgb:
In master we have that the last data migration is 12 https://github.com/rotki/rotki/blob/31d62792a165ce70c4b71d27c092ac9728a3ad56/rotkehlchen/data_migrations/migrations/migration_12.py#L27 for 1.30.X

but for 1.31 we have the same version so it breaks and it doesn't apply on 1.31


LefterisJP (approved):
LGTM. Merging it @yabirgb though I think we should have a way to test we don't re-introduce a deleted migration

PR #6788, opened 2023-10-20
Optimize caches, specifically velodrome

LefterisJP:
### [Optimize velodrome cache query by grouping token info queries](https://github.com/rotki/rotki/commit/00b63f02ffa2377300de447b309865bb48da7644) 

The velodrome cache query was querying token info in a for loop. This
commit optimized it by making sure the token info queries are bundled
before the loop and thus reduces amount of external queries

### [Fix get_general_cache_like](https://github.com/rotki/rotki/commit/6264ccf3d213afa5bf147e54d72de5897c09a246) 

The function was returning a list of single entry tuples. So in the places where it
was used (curve and velodrome cache queries) instead of doing an `in
['0xfoo1', '0xfoo2']` check it was doing `in
[('0xfoo1',), ('0xfoo2',)] check. That was always false for addresses.

### [Turn evm_inquirer erc20/721 caches to LRU caches](https://github.com/rotki/rotki/pull/6788/commits/d8b66ad913ea637694f04cd167397fcc92562d59) 
Also some other fixes in the code and optimization in speed of some tests.

### [Move some commonly used ABIs in memory](https://github.com/rotki/rotki/pull/6788/commits/ecd65bfa9be28e05b43c228a9045a2267021cd97) 
Move erc20, erc721 and univ1 LP token ABIs as class attributes to the
contracts class so that they are only loaded once in memory and not
multiple times for each EVM Chain.

Since they are used quite very often, loading them from the DB and
doing a json importing each time a token query happens ends up getting expensive.

### DIFF
--- a/rotkehlchen/chain/evm/node_inquirer.py
+++ b/rotkehlchen/chain/evm/node_inquirer.py
@@ -31,7 +31,12 @@ from rotkehlchen.assets.asset import CryptoAsset
 from rotkehlchen.chain.constants import DEFAULT_EVM_RPC_TIMEOUT
 from rotkehlchen.chain.ethereum.constants import DEFAULT_TOKEN_DECIMALS
 from rotkehlchen.chain.ethereum.utils import MULTICALL_CHUNKS, should_update_protocol_cache
-from rotkehlchen.chain.evm.constants import FAKE_GENESIS_TX_RECEIPT, GENESIS_HASH
+from rotkehlchen.chain.evm.constants import (
+    ERC20_PROPERTIES,
+    ERC721_PROPERTIES,
+    FAKE_GENESIS_TX_RECEIPT,
+    GENESIS_HASH,
+)
 from rotkehlchen.chain.evm.contracts import EvmContract, EvmContracts
 from rotkehlchen.chain.evm.proxies_inquirer import EvmProxiesInquirer
 from rotkehlchen.chain.evm.types import NodeName, Web3Node, WeightedNode
@@ -1119,40 +1124,49 @@ class EvmNodeInquirer(metaclass=ABCMeta):
             return output
         return [contract.decode(x, method_name, arguments[0]) for x in output]
 
-    def get_erc20_contract_info(self, address: ChecksumEvmAddress) -> dict[str, Any]:
-        """
-        Query an erc20 contract address and return basic information as:
-        - Decimals
-        - name
-        - symbol
-        At all times, the dictionary returned contains the keys; decimals, name & symbol.
-        Although the values might be None.
-        if it is provided in the contract. This method may raise:
-        - BadFunctionCallOutput: If there is an error calling a bad address
-        """
-        cache = self.contract_info_erc20_cache.get(address)
-        if cache is not None:
-            return cache
+    def get_multiple_erc20_contract_info(self, addresses: list[ChecksumEvmAddress]) -> dict[str, Any]:  # noqa: E501
+        all_info: dict[str, Any] = {}
 
-        properties = ('decimals', 'symbol', 'name')
+        if len(addresses) == 0:
+            return {}
+
+        contract = EvmContract(address=addresses[0], abi=self.contracts.abi('ERC20_TOKEN'), deployed_block=0)  # noqa: E501
+
+        for addresses_chunk in get_chunks(addresses, 8):  # chunk number seems to be highest that can work with etherscan url limit  # noqa: E501
+            calls = []
+            for address in addresses_chunk:
+                calls.extend([(address, contract.encode(method_name=prop)) for prop in ERC20_PROPERTIES])  # noqa: E501
+
+            try:
+                # Output contains call status and result
+                output = self.multicall_2(require_success=False, calls=calls)
+            except RemoteError:
+                # If something happens in the connection the output should have
+                # the same length as the tuple of properties * addresses
+                output = [(False, b'')] * len(ERC20_PROPERTIES) * len(addresses_chunk)
+
+            for idx, single_output in enumerate(get_chunks(output, 3)):
+                address = addresses_chunk[idx]
+                info = self._process_and_create_erc20_info(
+                    output=single_output,
+                    address=address,
+                )
+                self.contract_info_erc20_cache[address] = info
+                all_info[address] = info
+
+        return all_info
+
+    def _process_and_create_erc20_info(
+            self,
+            output: list[tuple[bool, bytes]],
+            address: ChecksumEvmAddress,
+    ) -> dict[str, Any]:
         info: dict[str, Any] = {}
-
-        abi = self.contracts.abi('ERC20_TOKEN')
-        contract = EvmContract(address=address, abi=abi, deployed_block=0)
-        try:
-            # Output contains call status and result
-            output = self.multicall_2(
-                require_success=False,
-                calls=[(address, contract.encode(method_name=prop)) for prop in properties],
-            )
-        except RemoteError:
-            # If something happens in the connection the output should have
-            # the same length as the tuple of properties
-            output = [(False, b'')] * len(properties)
+        contract = EvmContract(address=address, abi=self.contracts.abi('ERC20_TOKEN'), deployed_block=0)  # noqa: E501
         try:
             decoded = self._process_contract_info(
                 output=output,
-                properties=properties,
+                properties=ERC20_PROPERTIES,
                 contract=contract,
                 token_kind=EvmTokenKind.ERC20,
             )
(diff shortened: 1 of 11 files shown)


rotkehlchen/globaldb/cache.py
```diff
@@ -133,10 +133,10 @@ def globaldb_get_general_cache_like(
     key_parts should contain neither the "%" nor the "." symbol.
     """
     cache_key = compute_cache_key(key_parts)
-    return cursor.execute(
+    return [x[0] for x in cursor.execute(
```
LefterisJP, line 136: This is the most lol part

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -1169,20 +1183,62 @@ def get_erc20_contract_info(self, address: ChecksumEvmAddress) -> dict[str, Any]
...
+        Query an erc20 contract address and return basic information as:
+        - Decimals
+        - name
+        - symbol
+        At all times, the dictionary returned contains the keys; decimals, name & symbol.
+        Although the values might be None.
+        if it is provided in the contract. This method may raise:
+        - BadFunctionCallOutput: If there is an error calling a bad address
+        """
+        cache = self.contract_info_erc20_cache.get(address)
```
yabirgb, line 1231: shouldn't this cache be a LRU to avoid increasing its size too much? at the moment it only grows

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -1169,20 +1183,62 @@ def get_erc20_contract_info(self, address: ChecksumEvmAddress) -> dict[str, Any]
...
+        if cache is not None:
+            return cache
```
yabirgb, line 1233: nitpick: in those cases the walrus opeartor is nice

```suggestion
        if (cache :=  self.contract_info_erc20_cache.get(address)) is not None:
            return cache
```

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -1199,32 +1255,20 @@ def get_erc721_contract_info(self, address: ChecksumEvmAddress) -> dict[str, Any
...
-            # If something happens in the connection the output should have
-            # the same length as the tuple of properties
-            output = [(False, b'')] * len(properties)
         try:
             decoded = self._process_contract_info(
                 output=output,
-                properties=properties,
-                contract=contract,
+                properties=ERC721_PROPERTIES,
+                contract=EvmContract(address=address, abi=self.contracts.abi('ERC721_TOKEN'), deployed_block=0),  # noqa: E501
```
yabirgb, line 1264: wouldn't it make sense to store 

- self.contracts.abi('ERC20_TOKEN')
- self.contracts.abi('ERC721_TOKEN')

in memory to avoid multiple queries to the db + json load?

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
```diff
@@ -38,6 +38,8 @@
 class VelodromePoolData(NamedTuple):
     pool_address: ChecksumEvmAddress
     pool_name: str
+    token0: ChecksumEvmAddress
+    token1: ChecksumEvmAddress
```
yabirgb, line 42: toke0 and token1 made me think that this was a EvmToken. I would add `_address` to the names

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -1169,20 +1183,62 @@ def get_erc20_contract_info(self, address: ChecksumEvmAddress) -> dict[str, Any]
...
-        self.contract_info_erc20_cache[address] = info
         return info
 
+    def _query_token_contract(
+            self,
+            token_abi_str: Literal['ERC20_TOKEN', 'ERC721_TOKEN'],
+            properties: tuple[str, ...],
+            address: ChecksumEvmAddress,
+    ) -> list[tuple[bool, bytes]]:
+        abi = self.contracts.abi(token_abi_str)
```
yabirgb, line 1205: same comment about storing them in memory to avoid querying multiple times + json load

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -1119,40 +1124,49 @@ def multicall_specific(
...
-        At all times, the dictionary returned contains the keys; decimals, name & symbol.
-        Although the values might be None.
-        if it is provided in the contract. This method may raise:
-        - BadFunctionCallOutput: If there is an error calling a bad address
-        """
-        cache = self.contract_info_erc20_cache.get(address)
-        if cache is not None:
-            return cache
+    def get_multiple_erc20_contract_info(self, addresses: list[ChecksumEvmAddress]) -> dict[str, Any]:  # noqa: E501
+        all_info: dict[str, Any] = {}
```
yabirgb, line 1128: missing docstring

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -1119,40 +1124,49 @@ def multicall_specific(
...
 
-        properties = ('decimals', 'symbol', 'name')
-        info: dict[str, Any] = {}
+        if len(addresses) == 0:
+            return {}
```
yabirgb, line 1131: this can be executed before defining all_info

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -1119,40 +1124,49 @@ def multicall_specific(
...
-        except RemoteError:
-            # If something happens in the connection the output should have
-            # the same length as the tuple of properties
-            output = [(False, b'')] * len(properties)
+        contract = EvmContract(address=addresses[0], abi=self.contracts.abi('ERC20_TOKEN'), deployed_block=0)  # noqa: E501
+
+        for addresses_chunk in get_chunks(addresses, 8):  # chunk number seems to be highest that can work with etherscan url limit  # noqa: E501
+            calls = []
+            for address in addresses_chunk:
+                calls.extend([(address, contract.encode(method_name=prop)) for prop in ERC20_PROPERTIES])  # noqa: E501
```
yabirgb, line 1138: instead of extending multiple times isn't it better to do?

```suggestion
           calls = [
                [(address, contract.encode(method_name=prop)) for prop in ERC20_PROPERTIES])]
                for address in addresses_chunk
          ]
```

rotkehlchen/globaldb/cache.py
yabirgb: yeah saw it

yabirgb (commented):
Few questions/comments

rotkehlchen/chain/evm/node_inquirer.py
LefterisJP: done

rotkehlchen/chain/evm/node_inquirer.py
LefterisJP: done

rotkehlchen/chain/evm/node_inquirer.py
LefterisJP: This is the only thing I have not yet touched. I will push to other commit or PR. I would not store it in `self`. As there is as many `selves` as there is EVM chains. And this is universal across all of them. So need to rethink some stuff here.

rotkehlchen/chain/evm/node_inquirer.py
LefterisJP: added

rotkehlchen/chain/evm/node_inquirer.py
LefterisJP: Removed `all_info` since the function does not need to return anything

rotkehlchen/chain/evm/node_inquirer.py
LefterisJP: yeah sure done. But as you also can see by the mistake you made (the code you wrote is invalid syntax) making multilist python list comprehensions is hard and error prone :wink: 

It's unnatural and hard to read. Heard they optimize it, so not bad to do it but still ... weird.

rotkehlchen/chain/optimism/modules/velodrome/velodrome_cache.py
LefterisJP: done

yabirgb (approved):
Is in the name of the powers that have been gifted to me and for which I thank every day that I approve this PR.

The failing test seems to be that the branch is missing a rebase. I rebased locally and the test works fine

LefterisJP:
> Is in the name of the powers that have been gifted to me and for which I thank every day that I approve this PR.
> 
> The failing test seems to be that the branch is missing a rebase. I rebased locally and the test works fine

Aww thank you. I am working on doing the last thing on a new commit here and will push though. It needs some re-architecturing.

LefterisJP:
All green now!

PR #6798, opened 2023-10-22
Turn evm_contracts fixture into session scope and always use it

LefterisJP:
Depends on globaldb session scope since it needs to load data from there


LefterisJP:
Let's hope I won't need to fix them again :vulcan_salute:

PR #6801, opened 2023-10-23
Refactor the one inch logic to make it work everywhere for v4 and v5

yabirgb:
Abstract v4 code to work with v5 as v5 seems to work like v4 as for what decoding refers.

This PR moves the logic to all evm chains since v4 and v5 are deployed with the same address in all the evm chains


rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -175,6 +177,8 @@ def _add_builtin_decoders(self, rules: DecodingRules) -> None:
         normal loading?
         """
         self._add_single_decoder(class_name='Safemultisig', decoder_class=SafemultisigDecoder, rules=rules)  # noqa: E501
+        self._add_single_decoder(class_name='Oneinchv4', decoder_class=Oneinchv4Decoder, rules=rules)  # noqa: E501
```
LefterisJP, line 180: Hmmm we should not have it for all. Only places where 1inch has deployed. Have they deployed in all ofthe chains we support? I don't think so

rotkehlchen/tests/fixtures/blockchain.py
```diff
@@ -102,6 +102,7 @@ def _initialize_and_yield_evm_inquirer_fixture(
     )
     if mocked_proxies is not None:
         mock_proxies(parent_stack, mocked_proxies)
+    inquirer.contracts.initialize_common_abis()
```
LefterisJP, line 105: again with this one. I told you emove this.

rotkehlchen/chain/evm/decoding/decoder.py
yabirgb: for v5 is deployed in all the chains we support with the same address. Checked it

for v4 is in all the chains we support but for optimism they have a different address. I've move them to every chain and adjusted on optimism

rotkehlchen/tests/fixtures/blockchain.py
yabirgb: I've been all the weekend fighting with this and forgot

LefterisJP (approved):
LGTM. Rebased, fixed conflict and fixed remnant of other PR.

PR #6818, opened 2023-10-25
Added code of conduct file.

Kris248:
Just a code of conduct file. You can merge it if there's a need of this file in this project.


LefterisJP:
Thank you very much for the PR. But I will have to reject it as code of conduct is not something we would get written by an external. If at all we would need to write it ourselves.

PR #6819, opened 2023-10-25
feat: managing custom accounting rule linked setting

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/settings/accounting/rule/AccountingRuleSetting.vue
+++ b/frontend/app/src/components/settings/accounting/rule/AccountingRuleSetting.vue
@@ -1,6 +1,5 @@
 <script setup lang="ts">
 import { type DataTableHeader } from 'vuetify';
-import { pick } from 'lodash-es';
 import { type Collection } from '@/types/collection';
 import {
   type Filters,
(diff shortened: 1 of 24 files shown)


tewshi (approved):
lgtm!

frontend/app/src/types/settings/accounting.ts
```diff
@@ -57,3 +51,9 @@ export interface AccountingRuleRequestPayload
   readonly eventSubtypes?: string[];
   readonly counterparties?: string[];
 }
+
+export interface AccountingRuleLinkedSettingMap {
+  identifier: string;
+  label: string;
+  state: Ref<any>;
```
kelsos, line 58: Can we do something about this any here? the type should not be any right? can we infer the proper type?

frontend/app/src/composables/settings/accounting/rule-mapping.ts
```diff
@@ -0,0 +1,50 @@
...
+
+        if (state) {
+          const label = te(translationKey)
+            ? t(translationKey)
+            : toHumanReadable(item);
+
+          result.push({
+            identifier: item,
+            label,
+            state
```
kelsos, line 39: Do you use this to change the setting then? this is why you pass the ref here?

kelsos: I am not sure I like that we return a ref from a computed.

frontend/app/src/components/settings/accounting/rule/AccountingRuleWithLinkedSettingDisplay.vue
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  return null;
+});
+
+const value: ComputedRef<boolean> = computed(() => {
+  const selectedLinkableSettingVal = get(selectedLinkableSetting);
+  if (selectedLinkableSettingVal) {
+    return get(get(selectedLinkableSettingVal).state);
```
kelsos, line 37: We should not need to do a get-get, the computed property should update any time any of the state updates if we unref inside right? does it not?

frontend/app/src/components/settings/accounting/rule/AccountingRuleSetting.vue
```diff
@@ -149,11 +152,17 @@ const showDeleteConfirmation = (item: AccountingRuleEntry) => {
...
+  getEventTypeData({
+    eventType,
+    eventSubtype
+  });
 </script>
 
 <template>
   <div>
-    <TablePageLayout :title="[t('accounting_settings.rule.title')]">
+    <TablePageLayout child :title="[t('accounting_settings.rule.title')]">
```
kelsos, line 165: Can we make the `TablePageLayout` here the root element and then move the form in it's default slot to remove the extra div?

kelsos (approved):
lgtm

PR #6830, opened 2023-10-26
refactor: reduce the number of rendered elements on history view

kelsos:
Closes #(issue_number)

## Checklist


frontend/app/src/components/history/LocationIcon.vue
```diff
@@ -1,67 +1,80 @@
 <script setup lang="ts">
-import { type TradeLocationData } from '@/types/history/trade/location';
+const props = withDefaults(
+  defineProps<{
+    item: string;
```
tewshi, line 4: thought item should be one of `TradeLocationData | string | null`

kelsos: Yes that union at least for vue 2 doesn't work, if you do this the console will be full of warnings...

Also I think it makes more sense to just go with the identifier as it is guaranteed to be everywhere either way and then go and cleanup stuff.

tewshi (approved):
lgtm!

PR #6831, opened 2023-10-26
Make taxable also linkable for accounting rules

yabirgb:
This PR:

- Fixes an unhandled 500 error
- Allows to link the taxable field (done in the gas fee with `include_gas_costs`)

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -12597,7 +12597,7 @@ Managing custom accounting rules
       Content-Type: application/json;charset=UTF-8
 
       {
-         "taxable":true,
+         "taxable": {"value": true},
          "count_entire_amount_spend":{"value": false, "linked_setting": "include_crypto2crypto"},
          "count_cost_basis_pnl":{"value": true},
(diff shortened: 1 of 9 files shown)


docs/api.rst
```diff
@@ -12608,7 +12608,7 @@ Managing custom accounting rules
 
   .. _accounting_rules_fields:
 
-  :reqjsonarr boolean taxable: ``true`` if the event should be considered as taxable
+  :reqjsonarr object taxable: ``true`` if the event should be considered as taxable. Allows to link the property to an accounting setting.
```
LefterisJP, line 12611: How is an object ``true``? So can be boolean or object? How does it look when an object? Can't understand from the doc entry

yabirgb: I forgot to update the response object. It is always an object and the `value` attribute is what is a boolean. Fixed it along a better comment

LefterisJP (approved):
LGTM. Failure was 5xx error from beaconchain

PR #6832, opened 2023-10-26
Counterparty changes

LefterisJP:
- [Turn counterparties to a static class method](https://github.com/rotki/rotki/commit/20892ff76174bfdc803701e94d24669950a0a1f7)
- [Remove wrong airdrops CPT from accountants. Use direct from decoders](https://github.com/rotki/rotki/commit/8d3df3d3e419d9ca55c1f1bae7ca7fcac76cb6bf)


rotkehlchen/chain/evm/decoding/oneinch/decoder.py
```diff
@@ -99,10 +107,3 @@ def addresses_to_decoders(self) -> dict[ChecksumEvmAddress, tuple[Any, ...]]:
         return {
             self.router_address: (self.decode_action,),
         }
-
-    def counterparties(self) -> list[CounterpartyDetails]:
-        return [CounterpartyDetails(
-            identifier=self.counterparty,
```
LefterisJP, line 105: This is the only thing that made it difficult for me as this was unexpected. You recently made this using `self` for somethign I expected to be static. Though I can understand why. Made some changes to make it work.

PR #6837, opened 2023-10-27
[DO NOT MERGE] Adjust withdrawals beaconchain scraper

LefterisJP:
They changed the DOM but unfortunately also removed precision which is not something we can accept. So this is just a draft commit.

Correct approach is to switch to using multiple APIs for redundancy: https://github.com/orgs/rotki/projects/11/views/2?filterQuery=&pane=issue&itemId=42832092


LefterisJP:
Closing in favor of https://github.com/rotki/rotki/pull/6891

PR #6838, opened 2023-10-27
Remove accounting rules

yabirgb:
Closes #(issue_number)

## Checklist


rotkehlchen/accounting/rules.py
```diff
@@ -105,42 +25,49 @@ def __init__(
...
 
-    def get_event_settings(self, event: HistoryBaseEntry) -> Optional[BaseEventSettings]:
-        """Return a matching rule for the event if it exists"""
-        rule = self.event_settings.get(event.get_type_identifier(), None)
+    def get_event_settings(
+            self,
+            event: HistoryBaseEntry,
+    ) -> tuple[Optional[BaseEventSettings], Optional[EventsAccountantCallback]]:
+        """
+        Return a matching rule for the event if it exists and a optional callback defined for
```
LefterisJP, line 48:
```suggestion
        Return a matching rule for the event if it exists and an optional callback defined for
```

rotkehlchen/accounting/rules.py
```diff
@@ -105,42 +25,49 @@ def __init__(
...
+        """
+        event_identifier = event.get_type_identifier()
+        rule = self.event_settings.get(event_identifier, None)
         if isinstance(event, EvmEvent) is False or rule is not None:
-            return rule
+            return rule, self.event_callbacks.get(event_identifier)
 
-        return self.event_settings.get(event.get_type_identifier(include_counterparty=False), None)
+        event_identifier = event.get_type_identifier(include_counterparty=False)
+        return (
```
LefterisJP, line 57: What happens if there is no rule for the event?

rotkehlchen/chain/evm/accounting/aggregator.py
```diff
@@ -87,14 +86,14 @@ def initialize_all_accountants(self) -> None:
         """Recursively check all submodules to get all accountants and initialize them"""
         self._recursively_initialize_accountants(self.modules_path)
 
-    def get_accounting_settings(self, pot: 'AccountingPot') -> dict[int, TxEventSettings]:
+    def get_accounting_callbacks(self) -> dict[int, EventsAccountantCallback]:
```
LefterisJP, line 89: you will need to also adjust the docstring here

rotkehlchen/chain/evm/accounting/aggregator.py
```diff
@@ -112,15 +111,15 @@ class EVMAccountingAggregators:
     def __init__(self, aggregators: list[EVMAccountingAggregator]) -> None:
         self.aggregators = aggregators
 
-    def get_accounting_settings(self, pot: 'AccountingPot') -> dict[int, TxEventSettings]:
+    def get_accounting_callbacks(self) -> dict[int, EventsAccountantCallback]:
```
LefterisJP, line 114: same here. Adjust docstring

rotkehlchen/chain/evm/accounting/interfaces.py
```diff
@@ -38,7 +37,7 @@ def __init__(
         self.reset()
 
     @abstractmethod
-    def event_settings(self, pot: 'AccountingPot') -> dict[int, 'TxEventSettings']:
+    def event_callbacks(self) -> dict[int, 'EventsAccountantCallback']:
```
LefterisJP, line 40: adjust docstring

rotkehlchen/tests/fixtures/accounting.py
```diff
@@ -95,6 +98,33 @@ def fixture_accounting_initialize_parameters():
...
+    return True
+
+
+@pytest.fixture(name='last_accounting_rules_version', scope='session')
+def fixture_last_accounting_rules_version() -> int:
+    return 1
+
+
+@pytest.fixture(name='latest_accounting_rules', autouse=True, scope='session')
+def fixture_download_rules(last_accounting_rules_version) -> Path:
```
LefterisJP, line 112: Add a docstring on this fixture

rotkehlchen/tests/fixtures/accounting.py
```diff
@@ -95,6 +98,33 @@ def fixture_accounting_initialize_parameters():
...
+    root_dir = default_data_directory().parent / 'test-caching'
+    base_dir = root_dir / 'rules'
```
LefterisJP, line 114:
```suggestion
    base_dir = root_dir / 'accounting_rules'
```

rotkehlchen/tests/fixtures/accounting.py
```diff
@@ -95,6 +98,33 @@ def fixture_accounting_initialize_parameters():
...
+    rules_file = Path(base_dir / f'v{last_accounting_rules_version}.json')
+
+    if rules_file.exists():
+        return rules_file
+
+    response = requests.get(f'https://raw.githubusercontent.com/rotki/data/develop/updates/accounting_rules/v{last_accounting_rules_version}.json')
+    rules_file.parent.mkdir(exist_ok=True, parents=True)
+    with open(rules_file, 'w+') as f:
```
LefterisJP, line 122: why 'w+'`?

rotkehlchen/tests/fixtures/accounting.py
```diff
@@ -114,6 +145,17 @@ def fixture_accountant(
...
 
+    # add accounting rules to the database
+    data_updater = RotkiDataUpdater(
+        msg_aggregator=function_scope_messages_aggregator,
+        user_db=database,
+    )
+    with open(latest_accounting_rules) as f:
+        data_updater.update_accounting_rules(
+            data=json.loads(f.read())['accounting_rules'],
+            version=999999,  # only for logs
```
LefterisJP, line 156: What do you mean only for logs here? So that if you see 999999 to know it's the test?

rotkehlchen/tests/fixtures/accounting.py
```diff
@@ -95,6 +98,33 @@ def fixture_accounting_initialize_parameters():
     return False
 
 
+@pytest.fixture(name='initialize_accounting_rules')
+def fixture_initialize_accounting_rules() -> bool:
```
LefterisJP, line 102: Add a docstring here explainin what True means for this.

rotkehlchen/tests/fixtures/rotkehlchen.py
```diff
@@ -267,10 +269,12 @@ def initialize_mock_rotkehlchen_instance(
         network_mocking,
         have_decoders,
         add_accounts_to_db,
+        latest_accounting_rules,
+        initialize_accounting_rules,
 ) -> None:
     if not start_with_logged_in_user:
         return
-
+    rotki = cast(Rotkehlchen, rotki)
```
LefterisJP, line 277: ?? Why do you need this cast here now?

rotkehlchen/tests/fixtures/rotkehlchen.py
```diff
@@ -622,7 +658,7 @@ def rotkehlchen_api_server_with_exchanges(
 ):
     """Adds mock exchange objects to the rotkehlchen_server fixture"""
     exchanges = rotkehlchen_api_server.rest_api.rotkehlchen.exchange_manager.connected_exchanges
-    rotki = rotkehlchen_api_server.rest_api.rotkehlchen
+    rotki = cast(Rotkehlchen, rotkehlchen_api_server.rest_api.rotkehlchen)
```
LefterisJP, line 661: same here. What changed and why do you need this cast?

rotkehlchen/tests/unit/accounting/evm/test_balancer.py
```diff
@@ -31,20 +29,6 @@
 
 DEPOSIT_ENTRIES = [
     EvmEvent(
-        tx_hash=EVM_HASH,
```
LefterisJP, line 34: why remove 1 entire event here?

rotkehlchen/tests/unit/accounting/evm/test_curve.py
```diff
@@ -64,18 +61,6 @@
 
 DEPOSIT_ENTRIES = [
     EvmEvent(
-        tx_hash=EVM_HASH,
```
LefterisJP, line 67: same question here. Why remove this event?

rotkehlchen/tests/fixtures/accounting.py
yabirgb: The `w+` is to create the file if it doesn't exists. Pathlib creates the path tree and then this the file if needed

rotkehlchen/tests/fixtures/accounting.py
yabirgb: The version number is used inside of `update_accounting_rules` only to display it in the logs and doesn't do any check. If we see this high number we can detect easily were it comes from

rotkehlchen/accounting/rules.py
yabirgb: If there is no rule then this function returns None and the event will be skipped

rotkehlchen/tests/unit/accounting/evm/test_balancer.py
yabirgb: As commented on discord the test didn't have the rules for accounting the gas event and was skipping it. I tried first to add the processed event but it complicated the logic and since this test just tests the logic for the curve decoded events I saw it easier to delete it.

rotkehlchen/accounting/rules.py
LefterisJP: Can we have a log entry for these cases? To let client know something was skipped?

rotkehlchen/tests/fixtures/accounting.py
LefterisJP: That's not true. 'w' will still create the file if it does not exist. In both cases old file will be erased. Difference is you can also read with this. In short you don't need this. You need `'w'`

LefterisJP (approved):
LGTM. Poloniex failed test is due to maintenance.

PR #6840, opened 2023-10-27
Adjust event category details mapping and move direction

LefterisJP:
EventCategory is now a more complicated enum, which also has the direction in it.

The event_category_details which is returned to the frontend for displaying icons and more per event type is now configurable also by counterparty on top of type/subtype.

This is now only used for spend/fee and counterparty gas as a specialization of spend/fee.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -12130,14 +12130,37 @@ Event Mappings
             },
           },
           "event_category_details":{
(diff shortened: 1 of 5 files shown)


LefterisJP:
It's all green. Will just merge so frontend can utilize it.

PR #6846, opened 2023-10-27
Spawn a task for querying ens names during schema validation

yabirgb:
Addressing https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=42744837

Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/api/v1/parser.py
+++ b/rotkehlchen/api/v1/parser.py
@@ -24,10 +24,15 @@ class ResourceReadingParser(FlaskParser):
             validate: Optional[ValidateArg] = None,
             error_status_code: Optional[int] = None,
             error_headers: Optional[Mapping[str, str]] = None,
+            allow_async_validation: Optional[bool] = False,
     ) -> Callable:
         """Decorator that injects parsed arguments into a view function or method.
 
-        Edited from core parser to include the resource object
+        Edited from core parser to include the resource object.
+
+        When allow_async_validation is set to true we avoid doing validation for async requests
+        before spawning the the task and instead any validation error will appear in the async
+        task output.
         """
         location = location or self.location
         request_obj = req
@@ -57,6 +62,7 @@ class ResourceReadingParser(FlaskParser):
                     validate=validate,
                     error_status_code=error_status_code,
                     error_headers=error_headers,
+                    allow_async_validation=allow_async_validation,
                 )
                 args, kwargs = self._update_args_kwargs(
                     args, kwargs, parsed_args, as_kwargs, arg_name,
@@ -79,11 +85,15 @@ class ResourceReadingParser(FlaskParser):
             validate: Optional[ValidateArg] = None,
             error_status_code: Optional[int] = None,
             error_headers: Optional[Mapping[str, str]] = None,
+            allow_async_validation: Optional[bool] = False,
     ) -> dict[str, Any]:
         """Main request parsing method.
 
         Different from core parser is that we also get the resource object and
-        pass it to the schema
+        pass it to the schema.
+
+        If allow_async_validation is set to True then we return the schema that should be
+        validated and the raw data of the request.
         """
         req = req if req is not None else self.get_default_request()
         location = location or self.location
@@ -96,6 +106,17 @@ class ResourceReadingParser(FlaskParser):
             location_data = self._load_location_data(
                 schema=schema, req=req, location=location,
             )
+            if allow_async_validation and location_data.get('async_query', False) is True:
+                # If the validation is async instead of returning the loaded data from the schema
+                # we return the schema and the raw data so the validation can be executed in
+                # the spawned task
+                return {
+                    'do_async_validation': {
+                        'schema': schema,
+                        'data': location_data,
+                    },
+                }
+
             data = schema.load(location_data)
             self._validate_arguments(data, validators)
         except ma_exceptions.ValidationError as error:
--- a/rotkehlchen/api/v1/resources.py
+++ b/rotkehlchen/api/v1/resources.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING, Any, Callable, Literal, Optional
 
 from flask import Blueprint, Request, Response, request as flask_request
 from flask.views import MethodView
-from marshmallow import Schema
+from marshmallow import Schema, ValidationError
 from marshmallow.utils import missing
 from webargs.flaskparser import parser, use_kwargs
 from webargs.multidictproxy import MultiDictProxy
@@ -19,7 +19,12 @@ from rotkehlchen.accounting.structures.types import (
     HistoryEventSubType,
     HistoryEventType,
 )
-from rotkehlchen.api.rest import RestAPI, api_response, wrap_in_fail_result
+from rotkehlchen.api.rest import (
+    RestAPI,
+    api_response,
+    make_response_from_dict,
+    wrap_in_fail_result,
+)
 from rotkehlchen.api.v1.parser import ignore_kwarg_parser, resource_parser
 from rotkehlchen.api.v1.schemas import (
     AccountingReportDataSchema,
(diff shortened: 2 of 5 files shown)


rotkehlchen/api/v1/parser.py
```diff
@@ -79,11 +81,14 @@ def parse(  # type: ignore  # we have added the resource_object on top of parse
...
             error_headers: Optional[Mapping[str, str]] = None,
+            allow_async_validation: Optional[bool] = False,
     ) -> dict[str, Any]:
         """Main request parsing method.
 
         Different from core parser is that we also get the resource object and
-        pass it to the schema
+        pass it to the schema.
+
+        If allow_async_validation is set to True then
```
LefterisJP, line 91: then .... ? 

The suspense is killing me! :rofl:

rotkehlchen/api/v1/parser.py
```diff
@@ -96,6 +101,17 @@ def parse(  # type: ignore  # we have added the resource_object on top of parse
             location_data = self._load_location_data(
                 schema=schema, req=req, location=location,
             )
+            if allow_async_validation and location_data.get('async_query', False) is True:
```
LefterisJP, line 104: What exactly is `location_data`? I thought `data`is where the request args live. Why do you need to check this and not pure `data` from `schema.load()` down there?

rotkehlchen/api/v1/parser.py
```diff
@@ -124,7 +140,6 @@ def _get_schema(
 
         return schema
 
-
```
LefterisJP, line 127: why remove this line? Isn't this a linter error?

rotkehlchen/api/v1/resources.py
```diff
@@ -311,6 +311,62 @@ def load_view_args_file_data(request: Request, schema: Schema) -> MultiDictProxy
     return data
 
 
+def allow_async_validation() -> Callable:
+    """
+    Decorator to be used when validation should happen as an async task.
+
+    This decorator should be used after resource_parser.use_args using the argument
+    allow_async_validation=True. The validation and any possible error will be handled inside
+    the task spawned and if validation succeeds then the endpoint logic will be executed in the
```
LefterisJP, line 320:
```suggestion
    the spawned task and if validation succeeds then the endpoint logic will be executed in the
```

rotkehlchen/api/v1/resources.py
```diff
@@ -311,6 +311,62 @@ def load_view_args_file_data(request: Request, schema: Schema) -> MultiDictProxy
...
+        try:
+            data = schema.load(raw_data)
+            kwargs.update(data)
+        except ValidationError as e:
+            return {
+                'result': None,
+                'message': str(e),
+                'status_code': HTTPStatus.BAD_REQUEST,
+            }
+        kwargs.pop('async_query')  # async query has been already handled so it can be removed
```
LefterisJP, line 342:
```suggestion
        kwargs.pop('async_query')  # async query has already been handled so it can be removed
```

rotkehlchen/api/v1/resources.py
```diff
@@ -311,6 +311,62 @@ def load_view_args_file_data(request: Request, schema: Schema) -> MultiDictProxy
...
+        return method(view, *args, **kwargs)
+
+    def _allow_async_validation(f: Callable) -> Callable:
+        """Determine if the logic needs to be executed in a async or sync way"""
```
LefterisJP, line 346:
```suggestion
        """Determine if the validation logic needs to be executed in a sync or async way"""
```

rotkehlchen/api/v1/resources.py
```diff
@@ -311,6 +311,62 @@ def load_view_args_file_data(request: Request, schema: Schema) -> MultiDictProxy
...
+        @wraps(f)
+        def wrapper(view: BaseMethodView, *args: Any, **kwargs: Any) -> Any:
+            if (async_validation := kwargs.pop('do_async_validation', None)) is None:
+                kwargs.pop('async_query', None)  # we already know that it won't be async
+                print(args, kwargs)
```
LefterisJP, line 351: print statement 
```suggestion
```

rotkehlchen/api/v1/resources.py
```diff
@@ -1455,23 +1511,22 @@ def make_put_schema(self) -> EvmAccountsPutSchema:
...
-    def put(
-            self,
-            accounts: list[dict[str, Any]],
-            async_query: bool,
-    ) -> Response:
+    @resource_parser.use_args(
+        make_put_schema,
+        location='json_and_query',
+        allow_async_validation=True,
+        as_kwargs=True,
```
LefterisJP, line 1518: why is as kwargss needed? Add a comment. Also short comment next to `allow_async_validation=True` on what it does.

rotkehlchen/api/v1/resources.py
```diff
@@ -1455,23 +1511,22 @@ def make_put_schema(self) -> EvmAccountsPutSchema:
...
+    )
+    @allow_async_validation()
```
LefterisJP, line 1520: Can't this somehow be combined with the above? You need both to make it work. Right now it's readable, all good so no need to press it. Just wonder if it can be combined.

rotkehlchen/api/v1/schemas.py
```diff
@@ -3158,4 +3158,4 @@ def make_rules_query(
         )
         return {
             'filter_query': filter_query,
-        }
+        }
```
LefterisJP, line 3161: undo this diff

rotkehlchen/tests/api/blockchain/test_evm.py
```diff
@@ -431,3 +432,63 @@ def patch_single_query(**kwargs):  # pylint: disable=unused-argument
     with rotki.data.db.conn.read_ctx() as cursor:
         accounts = rotki.data.db.get_blockchain_accounts(cursor)
         assert set(accounts.eth) == {api_addies[1]}
+
+
+@pytest.mark.parametrize('number_of_eth_accounts', [0])
+def test_async_task(rotkehlchen_api_server: 'APIServer') -> None:
```
LefterisJP, line 438:
```suggestion
def test_evm_address_async(rotkehlchen_api_server: 'APIServer') -> None:
```

rotkehlchen/tests/api/blockchain/test_evm.py
```diff
@@ -431,3 +432,63 @@ def patch_single_query(**kwargs):  # pylint: disable=unused-argument
...
+    """Test that doing validation async for the evm addresses endpoints works"""
```
LefterisJP, line 439:
```suggestion
    """Test that doing async validation for the evm addresses endpoints works"""
```

rotkehlchen/api/v1/parser.py
yabirgb: location_data is the raw json that the frontend sends. The reason to not check `data` later is because at that moment we have called `.load` adn the `post_load` action has been triggered and that one is making the ens queries

rotkehlchen/api/v1/resources.py
yabirgb: I was thinking on doing it but is hard for two reasons:

1. You need the ApiServer that lives in the view method since is the one that has access to the "task manager"
2. The resource parser is already doing some "magic" and adding more things there complicated the logic

rotkehlchen/api/v1/parser.py
```diff
@@ -24,10 +24,15 @@ def use_args(
...
             error_headers: Optional[Mapping[str, str]] = None,
+            allow_async_validation: Optional[bool] = False,
     ) -> Callable:
         """Decorator that injects parsed arguments into a view function or method.
 
-        Edited from core parser to include the resource object
+        Edited from core parser to include the resource object.
+
+        When allow_async_validation is set to true we avoid doing validation for async requests
+        before spawning the the task and instead any validation error will appear in the async
```
LefterisJP, line 34:
```suggestion
        before spawning the task and instead any validation error will appear in the async
```

rotkehlchen/api/v1/parser.py
LefterisJP: can you also add this to the comment above. Explain what the location data is and why they are used?

rotkehlchen/api/v1/resources.py
LefterisJP: you need to add a comment here explaining why as I said in original request.

LefterisJP (approved):
I will merge since it's all doc/comment related requests and t's all green. But please handle in another PR.

PR #6849, opened 2023-10-29
Upgrade ruff and other linters fixing the code

LefterisJP:
### [Use set for constant literal collection checks](https://github.com/rotki/rotki/commit/b262f3111e51379ac2318502ce807acdafcb2d84) 

According to https://docs.python.org/3/whatsnew/3.2.html#optimizations
since Python 3.2 constant literal collections are automatically loaded
as a frozen set at compile time which make their loading time
insignificant.

So at that point it's suggested to replace constant collections of
list/tuples with sets as set IN comparison is faster.

Used ruff's https://docs.astral.sh/ruff/rules/literal-membership/ for
this but had to be very careful and only accept constant literal
suggestions. Since it suggests all IN collection checks to have this
rule, I also disabled it from ruff's pyproject toml for now and opened
an issue: https://github.com/astral-sh/ruff/issues/8322

### [Use dict.get() instead of if/else](https://github.com/rotki/rotki/commit/1896b5debfc2bac451553e881aea9d69e11f202f) 

This is ruff's rule SIM401

### [Remove unneeded noqas after ruff upgrade and fix bugs](https://github.com/rotki/rotki/commit/a77748a4bdd0ab2ad23f13a6acc44fc6674d4686)

This fixes a bug where `cond1 and cond2 or cond3` should have been `cond1 and (cond2 or cond3)`

### [Always specify encoding for write/read file](https://github.com/rotki/rotki/commit/2c2bfe489c4d22326dc461330015414ce70bcd4d) 

This is fixing ruff rule PLW1514

### DIFF
--- a/rotkehlchen/tests/fixtures/accounting.py
+++ b/rotkehlchen/tests/fixtures/accounting.py
@@ -54,7 +54,7 @@ def fixture_data_dir(use_clean_caching_directory, tmpdir_factory) -> Path:
     for x in data_directory.iterdir():
         directory_with_db = (
             x.is_dir() and
-            (x / 'rotkehlchen.db').exists() or (x / 'rotkehlchen_transient.db').exists()
+            ((x / 'rotkehlchen.db').exists() or (x / 'rotkehlchen_transient.db').exists())
         )
         if directory_with_db:
             shutil.rmtree(x, ignore_errors=True)
--- a/package.py
+++ b/package.py
@@ -474,14 +474,14 @@ class MacPackaging:
 
         :param wheel_metadata: Path to the wheel metadata file
         """
-        with open(wheel_metadata) as file:
+        with open(wheel_metadata, encoding='utf8') as file:
             data = file.readlines()
             for (index, line) in enumerate(data):
                 if not line.startswith('Tag'):
                     continue
                 data[index] = line.replace('x86_64', 'universal2')
 
-        with open(wheel_metadata, 'w') as file:
+        with open(wheel_metadata, 'w', encoding='utf8') as file:
             file.writelines(data)
 
     def __download_patched_pip(self) -> Path:
@@ -509,7 +509,7 @@ class MacPackaging:
         """
         requirements = self.__storage.working_directory / 'requirements.txt'
         package_versions: dict[str, str] = {}
-        with open(requirements) as fp:
+        with open(requirements, encoding='utf8') as fp:
             while True:
                 line = fp.readline()
                 if not line:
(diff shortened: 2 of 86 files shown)


rotkehlchen/tests/api/test_uniswap.py
```diff
@@ -116,7 +116,7 @@ def test_get_balances(
         for lp_asset in lp['assets']:
             lp_asset_type = type(lp_asset['asset'])
 
-            assert lp_asset_type in (str, dict)
+            assert lp_asset_type in {str, dict}
```
LefterisJP, line 119: Actually this is also wrong. I tested and str/dict are not considered constants for the interpreter. will undo

rotkehlchen/tests/api/test_sushiswap.py
```diff
@@ -92,7 +92,7 @@ def test_get_balances(
         for lp_asset in lp['assets']:
             lp_asset_type = type(lp_asset['asset'])
 
-            assert lp_asset_type in (str, dict)
+            assert lp_asset_type in {str, dict}
```
LefterisJP, line 95: same here. Not constant

rotkehlchen/tests/fixtures/accounting.py
```diff
@@ -54,7 +54,7 @@ def fixture_data_dir(use_clean_caching_directory, tmpdir_factory) -> Path:
     for x in data_directory.iterdir():
         directory_with_db = (
             x.is_dir() and
-            (x / 'rotkehlchen.db').exists() or (x / 'rotkehlchen_transient.db').exists()
+            ((x / 'rotkehlchen.db').exists() or (x / 'rotkehlchen_transient.db').exists())
```
LefterisJP, line 57: This was a bug. We needed `is_dir and (filename1 or filename2)` with a parentheses.

rotkehlchen/tests/unit/test_polygon_pos_inquirer.py
```diff
@@ -22,7 +22,7 @@ def test_polygon_pos_nodes_prune_and_archive_status(
         if node_name.endpoint == 'https://polygon-bor.publicnode.com':
             assert web3_node.is_pruned
             assert not web3_node.is_archive
-        elif node_name.endpoint in ('https://rpc.ankr.com/polygon', ALCHEMY_RPC_ENDPOINT):
+        elif node_name.endpoint in {'https://rpc.ankr.com/polygon', ALCHEMY_RPC_ENDPOINT}:
```
LefterisJP, line 25: this has a variable. Should also undo.

PR #6855, opened 2023-10-30
Feat(LoginForm): update login user input

tewshi:
## Checklist

### DIFF
--- a/frontend/app/src/components/account-management/login/LoginForm.vue
+++ b/frontend/app/src/components/account-management/login/LoginForm.vue
@@ -1,11 +1,13 @@
 <script setup lang="ts">
 import useVuelidate from '@vuelidate/core';
 import { helpers, required, requiredIf } from '@vuelidate/validators';
+import { toMessages } from '@/utils/validation';
 import { type LoginCredentials, type SyncApproval } from '@/types/login';
 
 const props = withDefaults(
   defineProps<{
     loading: boolean;
+    isDocker?: boolean;
     errors?: string[];
   }>(),
   {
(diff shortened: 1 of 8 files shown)


frontend/app/src/store/session/auth.ts
```diff
@@ -44,6 +49,38 @@ export const useSessionAuthStore = defineStore('session/auth', () => {
...
+      }
+      loggedUsers.push(user);
+    }
+    return loggedUsers;
+  };
+
+  const users: ComputedRef<string[]> = computed(() =>
+    Object.keys(get(userAccounts) ?? {})
+  );
+
```
kelsos, line 83: Is it really necessary to store the accounts in the pinia store?
Other than avoiding duplicate requests maybe during login?

The information is relevant only while you are at the login screen/login in.

tewshi: I was trying to avoid multiple calls, but, yes, I can revert to use multiple calls over storing in pinia.

kelsos:
So a couple of questions.
- Do you think it is possible to keep the normal text input for when it's the docker build. 
- Also for the non-docker application version do you think it would make sense to make the remember username functionality the default so that every time you open the application the last used account it pre-selected and hide the checkbox?

tewshi:
1. sure
2. sure thing too

kelsos (approved):
lgtm

kelsos:
@tewshi so I tried locally, and debug mode all good so far, headless mode I manage to get at least one failure

kelsos:
I am gonna merge this as it is and we need to figure out separately why the specific test started to fail.

PR #6862, opened 2023-10-31
specify events without accounting rules

jbrit:
This modifies the backend for this task: 
https://github.com/orgs/rotki/projects/11?pane=issue&itemId=42767182

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -4554,6 +4554,7 @@ Dealing with History Events
                       "notes": "Burned 0.00863351371344 ETH for gas",
                       "sequence_index": 0,
                       "timestamp": 1642802807,
+                      "missing_accounting_rule": true,
 		      "tx_hash": "0x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f",
 		      "address": null,
 		      "product": null
@@ -4578,6 +4579,7 @@ Dealing with History Events
                       "notes": "Burned 0.00863351371344 ETH for gas",
                       "sequence_index": 0,
                       "timestamp": 1642802807,
+                      "missing_accounting_rule": true,
 		      "tx_hash": "0x1c822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f",
 		      "address": null,
 		      "product": null
@@ -4600,6 +4602,7 @@ Dealing with History Events
                       "notes": "Withdrew 0.00163351371344 ETH from validator 1454",
                       "sequence_index": 0,
                       "timestamp": 1652802807,
+                      "missing_accounting_rule": true,
 		      "validator_index": 1454,
 		      "is_exit": false
                   },
@@ -4622,6 +4625,7 @@ Dealing with History Events
                       "notes": "Validator 1454 produced block 15534342 with 0.00163351371344 going to 0xce15887E2CEC81434C16D587709f64603b39b545 as the block reward",
                       "sequence_index": 0,
                       "timestamp": 1652802807,
+                      "missing_accounting_rule": true,
 		      "validator_index": 1454,
 		      "block_number": 15534342
                   },
@@ -4644,6 +4648,7 @@ Dealing with History Events
                       "notes": "Deposit 32 ETH to validator 4242",
                       "sequence_index": 15,
                       "timestamp": 1642802807,
+                      "missing_accounting_rule": true,
 		      "tx_hash": "0x2c822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f",
 		      "address": "0x00000000219ab540356cBB839Cbe05303d7705Fa",
 		      "product": "staking",
@@ -4667,6 +4672,7 @@ Dealing with History Events
    :resjson string event_identifier: Common key. An event identifier grouping multiple events under a common group. This is how we group transaction events under a transaction, staking related events under block production etc.
    :resjson int sequence_index: Common key. This is an index that tries to provide the order of history entries for a single event_identifier.
    :resjson int timestamp: Common key. The timestamp of the entry
+   :resjson bool missing_accounting_rule: Common key. Present and `True` if the event is not processed by any accounting rule.
    :resjson string location: Common key. The location of the entry. Such as "ethereum", "optimism", etc.
    :resjson string asset: Common key. The asset involved in the event.
    :resjson object balance: Common key. The balance of the asset involved in the event.
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -3474,22 +3474,29 @@ class RestAPI:
                 action_type=ActionType.HISTORY_EVENT,
             )
 
+        db_accounting = DBAccountingRules(self.rotkehlchen.data.db)
         if group_by_event_ids is True:
-            entries = [  # type: ignore  # mypy doesnt understand significance of boolean check
-                x.serialize_for_api(
+            missing_accounting_rules = db_accounting.missing_accounting_rules(
+                events=[x for _, x in events_result],  # type: ignore
+            )
+            entries = [  # type: ignore  # mypy doesn't understand significance of boolean check
+                x.serialize_for_api(  # type: ignore
                     customized_event_ids=customized_event_ids,
                     ignored_ids_mapping=ignored_ids_mapping,
                     hidden_event_ids=hidden_event_ids,
-                    grouped_events_num=grouped_events_num,
-                ) for grouped_events_num, x in events_result
+                    missing_accounting_rule=missing_accounting_rule,
+                    grouped_events_num=grouped_events_num,  # type: ignore
+                ) for (grouped_events_num, x), missing_accounting_rule in zip(events_result, missing_accounting_rules)  # noqa: E501
             ]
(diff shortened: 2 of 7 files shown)


rotkehlchen/api/rest.py
```diff
@@ -3474,9 +3474,16 @@ def get_history_events(
                 action_type=ActionType.HISTORY_EVENT,
             )
 
+        valid_entry_combinations = DBAccountingRules(self.rotkehlchen.data.db).get_valid_entry_combinations(events_result)  # noqa: E501
+
```
yabirgb, line 3478: for this I would process the events here since you know wether they are a list or a list of tuples and do something like

```python
if group_by_events_ids is False:
    get_valid_entry_combinations(events_result)
else:
    get_valid_entry_combinations([event for _, event in events_result])
```

you can call the method in the if statement below to not check again. The reason is that this way you simplify the logic in `get_valid_entry_combinations` and you adjust the data to the function instead of the function to the data format

rotkehlchen/db/accounting_rules.py
```diff
@@ -204,6 +205,28 @@ def add_linked_setting(
...
+            entries: list[HistoryBaseEntry] | list[tuple[int, HistoryBaseEntry]],
+    ) -> set[tuple[HistoryEventType, HistoryEventSubType, Optional[str]]]:
+        if len(entries) == 0:
+            return set()
+        combinations = {
+            (entry[1].event_type, entry[1].event_subtype, getattr(entry[1], 'counterparty', None))
+            if isinstance(entry, tuple)
+            else (entry.event_type, entry.event_subtype, getattr(entry, 'counterparty', None))
+            for entry in entries
+        }
```
yabirgb, line 219: as I mentioned standarizing the format you can remove complexity from here

rotkehlchen/db/accounting_rules.py
```diff
@@ -204,6 +205,28 @@ def add_linked_setting(
...
+        query = 'SELECT type, subtype, counterparty FROM accounting_rules WHERE'
+        query += ' OR '.join(f'(type="{x[0]}" AND subtype="{x[1]}" AND counterparty="{NO_ACCOUNTING_COUNTERPARTY if x[2] is None else x[2]}")' for x in combinations)  # noqa: E501
```
yabirgb, line 221: I believe is better to use `?` and then populate a list of bindings to use cursor.execute(query, bindings)

rotkehlchen/db/accounting_rules.py
yabirgb: Also you can make the transformation to NO_ACCOUNTING_COUNTERPARTY here

rotkehlchen/db/accounting_rules.py
```diff
@@ -204,6 +205,28 @@ def add_linked_setting(
...
+        with self.db.conn.read_ctx() as cursor:
+            cursor.execute(query)
+            return {(
+                HistoryEventType.deserialize(entry[0]),
+                HistoryEventSubType.deserialize(entry[1]),
+                None if entry[2] == NO_ACCOUNTING_COUNTERPARTY else str(entry[2]),
```
yabirgb, line 227: the call to `str` I believe is not needed

rotkehlchen/db/accounting_rules.py
```diff
@@ -204,6 +205,28 @@ def add_linked_setting(
                 f'Link of rule for {rule_property} to setting {setting_name} already exists',
             ) from e
 
+    def get_valid_entry_combinations(
+            self,
+            entries: list[HistoryBaseEntry] | list[tuple[int, HistoryBaseEntry]],
+    ) -> set[tuple[HistoryEventType, HistoryEventSubType, Optional[str]]]:
+        if len(entries) == 0:
```
yabirgb, line 212: don't forget to add a docstring

rotkehlchen/tests/api/test_ethereum_transactions.py
```diff
@@ -273,6 +273,12 @@ def _write_transactions_to_db(
                 )
 
 
+def assert_equal_events(returned_events, expected_events):
+    for returned_event, expected_event in zip(returned_events, expected_events):
```
yabirgb, line 277: add a small docstring

yabirgb (commented):
Some comments

rotkehlchen/db/accounting_rules.py
jbrit: i would have to do
```python
query += ' OR '.join('(type=? AND subtype=? AND counterparty=?' for _ in bindings)
```
where I would have had to process bindings to a flat map of `(type, subtype, counterparty) for x in combinations`, then run an execute

rotkehlchen/db/accounting_rules.py
jbrit: so for this, I didn't transform to `NO_ACCOUNTING_COUNTERPARTY` here, it wasn't different from leaving as is, and would have been an additional import to this file, unless I didn't understand correctly

rotkehlchen/db/accounting_rules.py
yabirgb: You can do `.join(['(type=? AND subtype=? AND counterparty=?'] * len(bindings))`

rotkehlchen/db/accounting_rules.py
```diff
@@ -204,6 +205,25 @@ def add_linked_setting(
                 f'Link of rule for {rule_property} to setting {setting_name} already exists',
             ) from e
 
+    def get_valid_entry_combinations(
+            self,
+            entries: list[HistoryBaseEntry],
+    ) -> set[tuple[HistoryEventType, HistoryEventSubType, Optional[str]]]:
+        """It returns the valid set of rules for the combinations of
+        event type, subtype and counterparty from a given list of entries"""
```
yabirgb, line 213: you are not returning a set of rules but a set of rules identifiers

yabirgb (commented):
two smaller comments

docs/api.rst
```diff
@@ -4644,6 +4648,7 @@ Dealing with History Events
                       "notes": "Deposit 32 ETH to validator 4242",
                       "sequence_index": 15,
                       "timestamp": 1642802807,
+                      "has_accounting_rule": false,
```
LefterisJP, line 4651: Wouldn't it be simpler and easier for both frontend and api if this is only existing if rule is missing and be `"missing_accounting_rule": true,`

rotkehlchen/db/accounting_rules.py
```diff
@@ -204,6 +205,33 @@ def add_linked_setting(
                 f'Link of rule for {rule_property} to setting {setting_name} already exists',
             ) from e
 
+    def get_valid_entry_combinations(
```
LefterisJP, line 208: don't do it like this. Instead what you want to learn is if a set of entries have valid rules or not. PAss the entries here and do a direct DB check.

rotkehlchen/api/rest.py
```diff
@@ -3474,9 +3474,20 @@ def get_history_events(
                 action_type=ActionType.HISTORY_EVENT,
             )
 
+        db_accounting = DBAccountingRules(self.rotkehlchen.data.db)
+        if group_by_event_ids is True:
+            valid_entry_combinations = db_accounting.get_valid_entry_combinations([event for _, event in events_result])  # type: ignore # noqa: E501
+        else:
+            valid_entry_combinations = db_accounting.get_valid_entry_combinations(events_result)  # type: ignore
+
+        def has_accounting_rule(entry: 'HistoryBaseEntry') -> dict[str, bool]:
```
LefterisJP, line 3483: why do it like this? Why not edit the `serialize_for_api()` function which is the function made for this?

Along with the return of the modified function I asked for in the comment on `get_valid_entry_combination` it should be straightforward and easier.

rotkehlchen/db/accounting_rules.py
```diff
@@ -204,6 +205,19 @@ def add_linked_setting(
                 f'Link of rule for {rule_property} to setting {setting_name} already exists',
             ) from e
 
+    def missing_accounting_rule(
```
LefterisJP, line 208: Hmm that's what I meant but not for a single event. As you end up doing multiple reads in a for loop.

Instead pass a list of events in this funciton and return them a list of true/false for missing rules.

rotkehlchen/db/accounting_rules.py
```diff
@@ -204,6 +205,26 @@ def add_linked_setting(
...
+        bindings = [
+            (
+                event.event_type.serialize(), event.event_subtype.serialize(),
+                NO_ACCOUNTING_COUNTERPARTY if (counterparty := getattr(event, 'counterparty', None)) is None else counterparty,  # noqa: E501
+            ) for event in events
+        ]
+        missing_accounting_rule: list[bool] = []
+        with self.db.conn.read_ctx() as cursor:
+            for idx in range(len(events)):
+                row = cursor.execute(query, bindings[idx]).fetchone()
```
LefterisJP, line 223: wonder if there is any way to do this with a single execute. Can't use the `IN` check here.

LefterisJP (approved):
LGTM

PR #6886, opened 2023-11-03
Format as valid json validation error for async validation

yabirgb:
as in valid json the frontend can parse it properly


LefterisJP (approved):
lgtm

LefterisJP:
LGTM but probably a failed validation test for these type of endpoints is needed. Frontend should not have had to find it out for us

Issue #6871, opened 2023-11-01
Failed to parse withdrawals response from beaconchain


benjaminion:
## Problem Definition

Feel free to close this if it is being dealt with via #6837!

Getting frequent error messages as follows:

```
Periodically query ethereum withdrawals died with exception: Failed to parse withdrawals response from beaconchain. Check logs. Check the logs for more details
```

This is frequent and consistent in an account that has made deposits to validators. I first noticed it a couple of days ago.

## Logs

Validator details redacted:

```
[01/11/2023 07:33:34 GMT Standard Time] DEBUG rotkehlchen.tasks.manager Greenlet-5: Scheduling task to Periodically query ethereum withdrawals
[01/11/2023 07:33:34 GMT Standard Time] DEBUG rotkehlchen.chain.ethereum.modules.eth2.eth2 Greenlet-7: Querying for validator withdrawals up to to_ts=1698824014
[01/11/2023 07:33:34 GMT Standard Time] DEBUG rotkehlchen.chain.ethereum.modules.eth2.eth2 Greenlet-7: Got the following validator data to query for withdrawals: [(XXX, 1695524567000), (YYY, 1695524567000), (ZZZ, 1695742295000), ...]
[01/11/2023 07:33:34 GMT Standard Time] DEBUG rotkehlchen.externalapis.beaconchain Greenlet-7: Querying beaconcha.in API for https://beaconcha.in/api/v1/validator/XXX,YYY,ZZZ,...
[...]

PR #6891, opened 2023-11-04
Adjust withdrawals queries

LefterisJP:
Closes: https://github.com/rotki/rotki/issues/6871 and https://github.com/orgs/rotki/projects/11?pane=issue&itemId=42832092
Supersedes: https://github.com/rotki/rotki/pull/6837

Still WIP.

1. Adjust typing in some DBFiltering and get history events
2. Add etherscan as a source for querying withdrawals
3. [TODO] Add blockscout as a source for quering withdrawals
    * Would need to also expose to frontend for api key addition


LefterisJP:
The only failure, `test_add_btc_blockchain_account_ens_domain` is a networking failure.

Will merge now as is. Note that blockscout implementation has been added but it not used due to it having been discovered to be missing data.

Will open a PR with the usage and the test, which we will set to not merge until we can be sure it's fixed on their side

PR #6900, opened 2023-11-06
clearing cache on blockchain balance refresh

jbrit:
Closes this [double counting issue](https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=42992233)

### DIFF
--- a/rotkehlchen/chain/aggregator.py
+++ b/rotkehlchen/chain/aggregator.py
@@ -921,16 +921,11 @@ class ChainsAggregator(CacheableMixIn, LockableQueryMixIn):
         manager = cast('EvmManager', self.get_chain_manager(chain))
         native_token_usd_price = Inquirer().find_usd_price(manager.node_inquirer.native_token)
         chain_balances = self.balances.get(chain)
-        queried_balances = manager.node_inquirer.get_multi_balance(accounts)
-        for account, balance in queried_balances.items():
-            if balance == ZERO:
-                continue
-
-            usd_value = balance * native_token_usd_price
+        for account, balance in manager.node_inquirer.get_multi_balance(accounts).items():
             chain_balances[account] = BalanceSheet(
                 assets=defaultdict(Balance, {
-                    manager.node_inquirer.native_token: Balance(balance, usd_value),
-                }),
+                    manager.node_inquirer.native_token: Balance(balance, balance * native_token_usd_price),  # noqa: E501
+                } if balance != ZERO else {}),  # accounts (e.g. multisigs) can have zero balances
             )
         self.query_evm_tokens(manager=manager, balances=chain_balances)
 
@@ -1045,8 +1040,7 @@ class ChainsAggregator(CacheableMixIn, LockableQueryMixIn):
     def _add_eth_protocol_balances(self, eth_balances: defaultdict[ChecksumEvmAddress, BalanceSheet]) -> None:  # noqa: E501
         """Also count token balances that may come from various eth protocols"""
         # If we have anything in DSR also count it towards total blockchain balances
-        dsr_module = self.get_module('makerdao_dsr')
-        if dsr_module is not None:
+        if (dsr_module := self.get_module('makerdao_dsr')) is not None:
             current_dsr_report = dsr_module.get_current_dsr()
             for dsr_account, balance_entry in current_dsr_report.balances.items():
 
@@ -1056,8 +1050,7 @@ class ChainsAggregator(CacheableMixIn, LockableQueryMixIn):
                 eth_balances[dsr_account].assets[A_DAI] += balance_entry
 
         # Also count the vault balances
-        vaults_module = self.get_module('makerdao_vaults')
-        if vaults_module is not None:
+        if (vaults_module := self.get_module('makerdao_vaults')) is not None:
             vault_balances = vaults_module.get_balances()
             for address, entry in vault_balances.items():
                 if address not in eth_balances:
@@ -1108,8 +1101,7 @@ class ChainsAggregator(CacheableMixIn, LockableQueryMixIn):
                     balances=defi_balances,
                 )
 
-        pickle_module = self.get_module('pickle_finance')
-        if pickle_module is not None:
+        if (pickle_module := self.get_module('pickle_finance')) is not None:
             pickle_balances_per_address = pickle_module.balances_in_protocol(
                 addresses=self.queried_addresses_for_module('pickle_finance'),
             )
@@ -1117,8 +1109,7 @@ class ChainsAggregator(CacheableMixIn, LockableQueryMixIn):
                 for asset_balance in pickle_balances:
                     eth_balances[address].assets[asset_balance.asset] += asset_balance.balance
 
-        liquity_module = self.get_module('liquity')
-        if liquity_module is not None:
+        if (liquity_module := self.get_module('liquity')) is not None:
             liquity_addresses = self.queried_addresses_for_module('liquity')
             # Get trove information
             liquity_balances = liquity_module.get_positions(given_addresses=liquity_addresses)
--- a/rotkehlchen/tests/api/test_balances.py
+++ b/rotkehlchen/tests/api/test_balances.py
@@ -1,4 +1,5 @@
 import random
+from collections import defaultdict
 from contextlib import ExitStack
 from http import HTTPStatus
 from typing import TYPE_CHECKING, Any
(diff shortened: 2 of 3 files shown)


rotkehlchen/chain/aggregator.py
```diff
@@ -840,6 +840,10 @@ def _update_balances_after_token_query(
     ) -> None:
         # Update the per account token balance and usd value
         for account, token_balances in balance_result.items():
+            if dsr_proxy_append is False:  # reset all account balances
+                for asset in list(balances[account].assets.keys()):
+                    del balances[account].assets[asset]
```
yabirgb, line 845: I believe this needs rethinking. You are deleting balacnes way too late in the process but I don't see any explanation for why we have duplicated balances. There has to be two logics that introduce the same balances twice for some reason.

The balances get empty here https://github.com/rotki/rotki/blob/8b1bfb8636391e6f52fbfc2e929aa3226bc8c216/rotkehlchen/chain/aggregator.py#L930 because this logic is executed first. Also I believe there is a bug because if there was balances before and after we have `balance == ZERO` then we don't clean it. But that is a different thing.


So the question that has not been addressed is where do we have double balances. If I'm not mistaken we clean the balances at the beginning in the pointed code and something later duplicates it. Have you checked if https://github.com/rotki/rotki/blob/8b1bfb8636391e6f52fbfc2e929aa3226bc8c216/rotkehlchen/chain/ethereum/defi/zerionsdk.py#L292 is adding the duplicated balances? What two sources are adding the same balance.

Please check if indeed the balances are not empty at the beginning of the query and at what point you insert the balances and then the duplicates.

This doesn't look like solving the issue because you are at one point of the execution removing balances when it should happen at the beginning of the balance query in any case.

yabirgb (commented):
There are some tests failing. I believe because the change is not correct

rotkehlchen/chain/aggregator.py
jbrit: You're right, but the bug is still the balances not getting cleared, my reason was just wrong

> The balances get empty here
> 
> https://github.com/rotki/rotki/blob/8b1bfb8636391e6f52fbfc2e929aa3226bc8c216/rotkehlchen/chain/aggregator.py#L930


The balances don't get cleared because the user's balance is 0, i pushed an updtae for this
> because this logic is executed first. Also I believe there is a bug because if there was balances before and after we have `balance == ZERO` then we don't clean it. But that is a different thing.
> So the question that has not been addressed is where do we have double balances. If I'm not mistaken we clean the balances at the beginning in the pointed code and something later duplicates it. Have you checked if


There're some tests failing, because they probably didn't count on `balance == ZERO`, I'm taking a look at all that fail after this run

rotkehlchen/tests/api/test_balances.py
```diff
@@ -911,3 +912,35 @@ def test_balances_behaviour_with_manual_current_prices(rotkehlchen_api_server, e
...
+    with account_balance_patch, query_tokens_patch:
+        response = requests.get(api_url_for(
+            rotkehlchen_api_server,
+            'blockchainbalancesresource',
+        ), json={'async_query': False, 'blockchain': 'ETH'})
+        result = assert_proper_response_with_result(response)
+        assert result['per_account']['eth'][ethereum_accounts[0]]['assets'] == {
+            'ETH': {'amount': '0', 'usd_value': '0.0'},
+            A_USDC.resolve_to_evm_token().identifier: {'amount': '23', 'usd_value': '230'},
+            A_DAI.resolve_to_evm_token().identifier: {'amount': '3', 'usd_value': '33'},
```
yabirgb, line 945: you don't need to resolve to get the identifier. The identifier is in teh `Asset` class, is the only thing that they have

```suggestion
            A_USDC.identifier: {'amount': '23', 'usd_value': '230'},
            A_DAI.identifier: {'amount': '3', 'usd_value': '33'},
```

rotkehlchen/tests/api/test_balances.py
```diff
@@ -911,3 +912,35 @@ def test_balances_behaviour_with_manual_current_prices(rotkehlchen_api_server, e
         rdn_result = result['assets']['eip155:1/erc20:0x255Aa6DF07540Cb5d3d297f0D0D4D84cb52bc8e6']
         assert rdn_result['amount'] == '5'
         assert rdn_result['usd_value'] == '150.0'
+
+
+@pytest.mark.parametrize('number_of_eth_accounts', [1])
+def test_blockchain_balances_refresh(
```
yabirgb, line 918: this test is doing remote calls right? Wouldn't it make sense to vcr and use a fixed address?

rotkehlchen/chain/aggregator.py
```diff
@@ -923,9 +923,6 @@ def query_evm_chain_balances(self, chain: SUPPORTED_EVM_CHAINS) -> None:
         chain_balances = self.balances.get(chain)
         queried_balances = manager.node_inquirer.get_multi_balance(accounts)
         for account, balance in queried_balances.items():
-            if balance == ZERO:
-                continue
-
```
yabirgb, line 928: this makes more sense but in the case of the account having 0 ETH I believe you are adding a 0 to the balances. Would be better to create an empty `BalanceSheet` since we don't return 0 balances. Also add a comment explaining why you do it

yabirgb (commented):
Some comments. Don't forget to add a entry in the changelog saying that you fixed this bug with duplicated balances in some cases

rotkehlchen/chain/aggregator.py
jbrit: the way i did it, I'm not sure if it needed a comment

rotkehlchen/chain/aggregator.py
```diff
@@ -919,18 +919,12 @@ def query_evm_chain_balances(self, chain: SUPPORTED_EVM_CHAINS) -> None:
...
-            usd_value = balance * native_token_usd_price
+        for account, balance in manager.node_inquirer.get_multi_balance(accounts).items():
             chain_balances[account] = BalanceSheet(
                 assets=defaultdict(Balance, {
-                    manager.node_inquirer.native_token: Balance(balance, usd_value),
-                }),
+                    manager.node_inquirer.native_token: Balance(balance, balance * Inquirer().find_usd_price(manager.node_inquirer.native_token)),  # noqa: E501
```
yabirgb, line 926: I would keep the price query outside the for loop for different reasons:

1. The asset is constant
2. Is one extra function call that has to context switch
3. You are calling the `Inquirer()` class that is creating the object and this is extra logic that you don't need to execute
4. It can happen that the price goes out of the cache and you make additional queries

I would keep as before and just do usd_value = ZERO if balance == ZERO

rotkehlchen/chain/aggregator.py
```diff
@@ -921,16 +921,11 @@ def query_evm_chain_balances(self, chain: SUPPORTED_EVM_CHAINS) -> None:
...
-            if balance == ZERO:
-                continue
-
-            usd_value = balance * native_token_usd_price
+        for account, balance in manager.node_inquirer.get_multi_balance(accounts).items():
             chain_balances[account] = BalanceSheet(
                 assets=defaultdict(Balance, {
-                    manager.node_inquirer.native_token: Balance(balance, usd_value),
-                }),
+                    manager.node_inquirer.native_token: Balance(balance, balance * native_token_usd_price),  # noqa: E501
```
LefterisJP, line 927: I don't understand the fix. This needs a comment explaining why you added the `if balance != ZERO else {}`. How does that affect the specific case of the user with stETH makerDAO vault?

rotkehlchen/tests/api/test_balances.py
```diff
@@ -911,3 +912,32 @@ def test_balances_behaviour_with_manual_current_prices(rotkehlchen_api_server, e
...
         assert rdn_result['usd_value'] == '150.0'
+
+
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('ethereum_accounts', [['0x7e574e063903b1D6DFf54A9C8B1260e6E068d35e']])
+def test_blockchain_balances_refresh(rotkehlchen_api_server: 'APIServer', ethereum_accounts):
+    """Checks that blockchain balances are refreshed properly when the endpoint is called"""
+    rotki = rotkehlchen_api_server.rest_api.rotkehlchen
+
+    account_balance = {ethereum_accounts[0]: BalanceSheet(
```
LefterisJP, line 923: What is this test doing? You are both mocking and VCRing. Don't. Do one or the other. Otherwise this is going to be impossible to maintain in the future.

Just use VCR and a real balance example where this problem you found is happening. 

**Make sure the test fails** without your fix.

LefterisJP (approved):
LGTM

PR #6910, opened 2023-11-08
Handle missing accounting rules for accoutning treatment

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/db/accounting_rules.py
+++ b/rotkehlchen/db/accounting_rules.py
@@ -5,7 +5,7 @@ from pysqlcipher3 import dbapi2 as sqlcipher
 
 from rotkehlchen.accounting.structures.base import HistoryBaseEntry
 from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
-from rotkehlchen.chain.evm.accounting.structures import BaseEventSettings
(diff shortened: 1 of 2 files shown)


rotkehlchen/db/accounting_rules.py
```diff
@@ -225,9 +225,31 @@ def missing_accounting_rules(self, events: list[HistoryBaseEntry]) -> list[bool]
...
-                missing_accounting_rule.append(row[0] == 0)
+                if (is_missing_rule := row[0] == 0) is False:
+                    missing_accounting_rule.append(is_missing_rule)
+                    continue
+
+                if (  # check for cases that could use accounting_treatment
+                    event.event_type == HistoryEventType.TRADE and
+                    event.event_subtype in (HistoryEventSubType.RECEIVE, HistoryEventSubType.FEE)
+                ):
+                    cursor.execute(
```
LefterisJP, line 238: This is deep inside a for loop of all events. So this happens for many events. Can't you gather them all in one query and run it once only?

Also though I think I kind of understand what you are doing I would need a few more comments explaining what you mean by "could use accounting_treatment".

LefterisJP (approved):
My comment is optimization and comment explanation nitpick, so will merge since all green.

PR #6912, opened 2023-11-08
Fixes on custom assets

yabirgb:
- allow to search for custom assets in the history view (before was not possible because we had as a constraint to use only crypto assets there)
- Fix assets import from backup for custom assets


LefterisJP (approved):
lgtm

PR #6914, opened 2023-11-09
Handle beaconchain performance endpoint shenanigans

LefterisJP:
Beaconchain may return only some of the attributes, and if it's less than a day may also return "performancetoday" as a key ...


yabirgb (approved):
lgtm

PR #6918, opened 2023-11-10
Small fixes backend

yabirgb:
- fix handling of fiat assets when missing price
- fix ws message sent when is not needed


LefterisJP (approved):
for this you could just have vcred no?

PR #6926, opened 2023-11-15
fix: refresh prices after changing the oracle order

lukicenturi:
https://docs.google.com/spreadsheets/d/1MNF6gsUmlZEhVbFg5f3ZzGSSJMu8Ti4aXx8KXSUL4j4/edit#gid=0&range=M350:M351

### DIFF
--- a/frontend/app/src/components/helper/PrioritizedList.vue
+++ b/frontend/app/src/components/helper/PrioritizedList.vue
@@ -103,59 +103,58 @@ const autoCompleteHint: ComputedRef<string> = computed(() => {
 
 <template>
   <div>
-    <VSheet outlined rounded>
-      <div class="pb-0" :class="slots.title ? 'pa-4' : 'pa-0'">
-        <div v-if="slots.title" class="text-h6 pb-4">
+    <RuiCard rounded="md" class="[&>div]:!p-0">
+      <template v-if="slots.title" #header>
+        <div class="p-4">
           <slot name="title" />
         </div>
+      </template>
 
-        <VRow v-if="!disableAdd" no-gutters>
-          <VCol class="pr-4">
-            <VAutocomplete
-              v-model="selection"
-              prepend-inner-icon="mdi-magnify"
-              outlined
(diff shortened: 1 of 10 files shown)


frontend/app/src/components/helper/PrioritizedList.vue
```diff
@@ -103,59 +103,58 @@ const autoCompleteHint: ComputedRef<string> = computed(() => {
 
 <template>
   <div>
-    <VSheet outlined rounded>
-      <div class="pb-0" :class="slots.title ? 'pa-4' : 'pa-0'">
-        <div v-if="slots.title" class="text-h6 pb-4">
+    <RuiCard rounded="md" class="[&>div]:!p-0">
```
tewshi, line 106: we can probably add a prop to card to not add padding (some libs use `flush`).
doesn't affect your PR though.

frontend/app/src/components/helper/PrioritizedList.vue
```diff
@@ -165,78 +164,63 @@ const autoCompleteHint: ComputedRef<string> = computed(() => {
...
-                    icon
-                    small
-                    :disabled="isLast(identifier)"
-                    @click="move(identifier, true)"
-                  >
-                    <VIcon>mdi-chevron-down</VIcon>
-                  </VBtn>
-                </div>
+                <RuiButtonGroup variant="outlined" size="sm" icon vertical>
+                  <template #default>
```
tewshi, line 176: just to be sure, do you need to put it in default slot this way?

frontend/app/src/components/helper/PrioritizedList.vue
lukicenturi: Yeah that sounds good. We can do that for next release

frontend/app/src/components/helper/PrioritizedList.vue
lukicenturi: There is a bug with <ButtonGroup> and <RadioGroup>, the children aren't reactive.
We need to do this until we move to vue 3

tewshi (approved):

PR #6928, opened 2023-11-15
fix: service key always show value when filled

lukicenturi:
- [x] Service key always show value when filled
- [x] Amount Input skeleton issue


tewshi (approved):

kelsos (approved):
lgtm

PR #6940, opened 2023-11-17
Bump linting

LefterisJP:


LefterisJP:
Merging only coingecko api limtis hit

PR #6945, opened 2023-11-19
Periodic task to find which withdrawals have exited

LefterisJP:
Solves https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=43645180


LefterisJP:
This is the only failing test and passes locally: `rotkehlchen/tests/integration/test_price_history.py::test_price_queries`. Was a timeout.

PR #6953, opened 2023-11-20
Accounting rule detection for spend events

yabirgb:
- Check related events to identify if an event will be processed by a accounting treatment rule
- For staking event add an exception since they don't use accounting rules

closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=45155775

----

logic:

for accounting rules what we do is check wether the combination of type/subtype/counterparty is present in the accounting rules to decide if the event will be processed or not.

For all the other things that use accounting treatment the combination of type/subtype for the first and second event is not fixed as you saw with the different cases that you found. 

so let say that for [event_1, event_2] we have that event_1 has accounting rule with accounting_treatment = SWAP and event_2 doesn't have a rule because it will be processed together with event_1.

When checking if event_2 has a rule or not the logic will be

```
1. Check if (event_2.type, event_2.subtype, event_2.counterparty) a matching accounting rule
2. Since it doesn't have fetch the related events, that is [event_1, event_2]
3. Get the index for that event that we are checking in the previous list, in this case 1.
4. Check if the event just before event_2 has the accounting treatment that would affect event_2. In this case since event_1 does a SWAP treatment event_2 will be processed.
```

### DIFF
--- a/rotkehlchen/chain/evm/accounting/interfaces.py
+++ b/rotkehlchen/chain/evm/accounting/interfaces.py
@@ -37,10 +37,15 @@ class ModuleAccountantInterface(metaclass=ABCMeta):
         self.reset()
 
     @abstractmethod
-    def event_callbacks(self) -> dict[int, 'EventsAccountantCallback']:
+    def event_callbacks(self) -> dict[int, tuple[int, 'EventsAccountantCallback']]:
         """
         Subclasses implement this to specify callbacks that should be executed for combinations of
-        type, subtype and counterparty
+        type, subtype and counterparty.
+
+        It returns a mapping of (hashed type, subtype, counterparty) to a tuple of the number of
+        events processed and the logic that will process them. If the number of events
+        is variable it returns -1 and the user needs to actually call the logic to get the number
+        of processed events.
         """
 
     def reset(self) -> None:
@@ -61,7 +66,7 @@ class DepositableAccountantInterface(ModuleAccountantInterface):
             pot: 'AccountingPot',
             event: 'EvmEvent',
             other_events: Iterator['EvmEvent'],
-    ) -> None:
+    ) -> int:
         """
         Process deposits and withdrawals from protocols that allow to deposit multiple assets
         in return for a wrapped token. There are multiple events that we have to consume from
@@ -71,6 +76,8 @@ class DepositableAccountantInterface(ModuleAccountantInterface):
         marking the number of events to consume.
         The return wrapped event needs to have the key `withdrawal_events_num` with the number
         of events in the return event.
+
+        It returns the number of events processed
         """
         if event.event_type == HistoryEventType.RECEIVE:
             # Pool token is received, which means it is a deposit
@@ -85,14 +92,17 @@ class DepositableAccountantInterface(ModuleAccountantInterface):
                 f'Could not find the number of events to consume for a {event.counterparty} '
                 f'deposit/withdrawal transaction {event.tx_hash.hex()}',
             )
-            return
+            return 1
 
         # Consume the events
+        events_consumed = 1  # counting the event that started it
         for idx in range(events_to_consume):
             next_event = next(other_events, None)
             if next_event is None:
                 log.debug(f'Could not consume event nr. {idx} for {event.counterparty} deposit/withdrawal')  # noqa: E501
-                return
+                return events_consumed
+
+            events_consumed += 1
 
             if next_event.balance.amount == ZERO:
                 continue
@@ -110,3 +120,5 @@ class DepositableAccountantInterface(ModuleAccountantInterface):
                 count_cost_basis_pnl=False,
                 extra_data={'tx_hash': next_event.tx_hash.hex()},
             )
+
+        return events_consumed
--- a/rotkehlchen/chain/evm/accounting/structures.py
+++ b/rotkehlchen/chain/evm/accounting/structures.py
@@ -16,12 +16,13 @@ class EventsAccountantCallback(Protocol):
             pot: 'AccountingPot',
             event: 'EvmEvent',
             other_events: Iterator['EvmEvent'],
-    ) -> None:
+    ) -> int:
         """
         Callback to be called by the accounting module.
         If the callback expects more than 1 events, it is supposed to iterate over the
         `other_events` iterator to get them.
         Note that events consumed by the callback from the iterator will not be re-processed later.
+        It returns the number of events processed by the callback.
         """
 
 
--- a/rotkehlchen/db/accounting_rules.py
+++ b/rotkehlchen/db/accounting_rules.py
@@ -1,24 +1,36 @@
 import logging
+from collections.abc import Iterator, Sequence
 from typing import TYPE_CHECKING, Any, NamedTuple, Optional, Union
 
 from pysqlcipher3 import dbapi2 as sqlcipher
 
-from rotkehlchen.accounting.structures.base import HistoryBaseEntry
+from rotkehlchen.accounting.structures.base import HistoryBaseEntry, HistoryEvent
+from rotkehlchen.accounting.structures.eth2 import EthStakingEvent
+from rotkehlchen.accounting.structures.evm_event import EvmEvent
 from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
-from rotkehlchen.chain.evm.accounting.structures import BaseEventSettings, TxAccountingTreatment
+from rotkehlchen.chain.evm.accounting.structures import (
+    BaseEventSettings,
+    EventsAccountantCallback,
+    TxAccountingTreatment,
+)
+from rotkehlchen.chain.evm.decoding.constants import CPT_GAS
 from rotkehlchen.db.constants import (
     LINKABLE_ACCOUNTING_PROPERTIES,
     LINKABLE_ACCOUNTING_SETTINGS_NAME,
     NO_ACCOUNTING_COUNTERPARTY,
 )
-from rotkehlchen.db.drivers.gevent import DBCursor
-from rotkehlchen.db.filtering import AccountingRulesFilterQuery
+from rotkehlchen.db.filtering import AccountingRulesFilterQuery, HistoryEventFilterQuery
+from rotkehlchen.db.history_events import DBHistoryEvents
 from rotkehlchen.db.settings import DEFAULT_INCLUDE_CRYPTO2CRYPTO, DEFAULT_INCLUDE_GAS_COSTS
 from rotkehlchen.errors.misc import InputError
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 
 if TYPE_CHECKING:
+    from rotkehlchen.accounting.accountant import Accountant
+    from rotkehlchen.accounting.pot import AccountingPot
+    from rotkehlchen.chain.evm.accounting.aggregator import EVMAccountingAggregators
     from rotkehlchen.db.dbhandler import DBHandler
+    from rotkehlchen.db.drivers.gevent import DBCursor
 
 
 logger = logging.getLogger(__name__)
@@ -174,7 +186,7 @@ class DBAccountingRules:
 
     def add_linked_setting(
             self,
-            write_cursor: DBCursor,
+            write_cursor: 'DBCursor',
             rule_identifier: int,
             rule_property: LINKABLE_ACCOUNTING_PROPERTIES,
             setting_name: LINKABLE_ACCOUNTING_SETTINGS_NAME,
@@ -208,50 +220,6 @@ class DBAccountingRules:
             raise InputError(
                 f'Link of rule for {rule_property} to setting {setting_name} already exists',
             ) from e
-
-    def missing_accounting_rules(self, events: list[HistoryBaseEntry]) -> list[bool]:
-        """
-        For a list of events returns a list of the same length with boolean values where True
-        means that the event won't be affected by any accounting rule
-        """
-        query = 'SELECT COUNT(*) FROM accounting_rules WHERE (type=? AND subtype=? AND (counterparty=? OR counterparty=?))'  # noqa: E501
-        bindings = [
-            (
-                event.event_type.serialize(),
-                event.event_subtype.serialize(),
-                NO_ACCOUNTING_COUNTERPARTY if (counterparty := getattr(event, 'counterparty', None)) is None else counterparty,  # noqa: E501
-                NO_ACCOUNTING_COUNTERPARTY,  # account for rules based only on type/subtype
-            ) for event in events
-        ]
-        missing_accounting_rule: list[bool] = []
-        with self.db.conn.read_ctx() as cursor:
-            for idx, event in enumerate(events):
-                row = cursor.execute(query, bindings[idx]).fetchone()
-                if (is_missing_rule := row[0] == 0) is False:
-                    missing_accounting_rule.append(is_missing_rule)
-                    continue
-
-                if (  # check for cases that could use accounting_treatment
-                    event.event_type == HistoryEventType.TRADE and
-                    event.event_subtype in (HistoryEventSubType.RECEIVE, HistoryEventSubType.FEE)
-                ):
-                    cursor.execute(
-                        query + ' AND (accounting_treatment=? OR accounting_treatment=?)',
-                        (
-                            HistoryEventType.TRADE.serialize(),
-                            HistoryEventSubType.SPEND.serialize(),
-                            NO_ACCOUNTING_COUNTERPARTY if (counterparty := getattr(event, 'counterparty', None)) is None else counterparty,  # noqa: E501
-                            NO_ACCOUNTING_COUNTERPARTY,
-                            TxAccountingTreatment.SWAP.serialize_for_db(),
-                            TxAccountingTreatment.SWAP_WITH_FEE.serialize_for_db(),
-
-                        ),
-                    )
-                    missing_accounting_rule.append(cursor.fetchone()[0] == 0)
-                else:
-                    missing_accounting_rule.append(is_missing_rule)
-
-        return missing_accounting_rule
 
     def query_rules(
             self,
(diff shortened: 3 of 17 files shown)


rotkehlchen/db/accounting_rules.py
```diff
@@ -224,8 +227,13 @@ def missing_accounting_rules(self, events: list[HistoryBaseEntry]) -> list[bool]
             ) for event in events
         ]
         missing_accounting_rule: list[bool] = []
+        history_db = DBHistoryEvents(self.db)
         with self.db.conn.read_ctx() as cursor:
             for idx, event in enumerate(events):
+                if isinstance(event, EthStakingEvent) is True:
+                    missing_accounting_rule.append(False)  # staking events are processed without need of accounting rules  # noqa: E501
```
LefterisJP, line 234:
```suggestion
                    missing_accounting_rule.append(False)  # staking events all have a process() function for accounting
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -248,8 +256,58 @@ def missing_accounting_rules(self, events: list[HistoryBaseEntry]) -> list[bool]
...
+
+                # check the surrounding events to verify if this particular event will
+                # be processed by a previous one. First query the related events
+                related_events = history_db.get_history_events(
+                    cursor=cursor,
+                    filter_query=HistoryEventFilterQuery.make(event_identifiers=[event.event_identifier]),
+                    has_premium=True,
+                )
+                for related_event_idx, related_event in enumerate(related_events):
+                    counterparty = getattr(event, 'counterparty', None)
```
LefterisJP, line 269: This should be the for loop event right?

```suggestion
                    counterparty = getattr(related_event, 'counterparty', None)
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -248,8 +256,58 @@ def missing_accounting_rules(self, events: list[HistoryBaseEntry]) -> list[bool]
...
+                    )
+                    if (raw_treatment := cursor.fetchone()) is None:
+                        continue
+
+                    accounting_treatment = TxAccountingTreatment.deserialize_from_db(raw_treatment[0])  # noqa: E501
+                    match accounting_treatment:
+                        case TxAccountingTreatment.SWAP:
+                            # ensure that the event that we are going to check can exist
+                            # in the related events. Avoid index error
+                            if len(related_events) <= related_event_idx + 1:
```
LefterisJP, line 291: 1. Don't use yoda conditionals. They are too confusing
2. If something is used multiple times in a for loop define **once** outside of it.
```suggestion
                            if related_event_idx + 1 > related_events_length:
```

But why do you need this condition? The comment seems misleading. if you hit this continue it's over, right?

rotkehlchen/db/accounting_rules.py
```diff
@@ -248,8 +256,58 @@ def missing_accounting_rules(self, events: list[HistoryBaseEntry]) -> list[bool]
...
+                                continue
+
+                            if (
+                                related_events[related_event_idx + 1].identifier == event.identifier  # noqa: E501
+                            ):
+                                missing_accounting_rule.append(False)
+                                break
+                        case TxAccountingTreatment.SWAP_WITH_FEE:
+                            if len(related_events) <= related_event_idx + 2:
```
LefterisJP, line 300: same as above

rotkehlchen/tests/db/test_db_accounting_rules.py
```diff
@@ -209,3 +211,57 @@ def test_missing_accounting_rules_accounting_treatment(database: 'DBHandler') ->
     assert not all(
         db.missing_accounting_rules([swap_event_spend, swap_event_receive, swap_event_fee]),
     )
+
+    # test missing accounting rules for non trade events
```
LefterisJP, line 215: 1. Wouldn't it be best toadd a different test? This is a separate logical test.
2. Don't you need also one for something like swap with fee?

rotkehlchen/chain/evm/accounting/interfaces.py
```diff
@@ -85,14 +92,17 @@ def _process_deposit_or_withdrawal(
                 f'Could not find the number of events to consume for a {event.counterparty} '
                 f'deposit/withdrawal transaction {event.tx_hash.hex()}',
             )
-            return
+            return 0
 
         # Consume the events
+        events_consumed = 0
```
LefterisJP, line 98:
```suggestion
        events_processed = 1  # counting the event that started it
```

rotkehlchen/chain/evm/accounting/structures.py
```diff
@@ -16,7 +16,7 @@ def __call__(
             pot: 'AccountingPot',
             event: 'EvmEvent',
             other_events: Iterator['EvmEvent'],
-    ) -> None:
+    ) -> int:
```
LefterisJP, line 19: Adjust the docstring to explain what is returned

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +303,151 @@ def query_rules_and_serialize(
                 data[linked_property]['linked_setting'] = setting_name
             entries.append(data)
         return entries, total_found_result
+
+
+def _check_if_event_affected_by_treatment(
```
LefterisJP, line 308:
```suggestion
def _is_event_affected_by_treatment(
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +303,151 @@ def query_rules_and_serialize(
...
+                related_event.event_type.serialize(),
+                related_event.event_subtype.serialize(),
+                NO_ACCOUNTING_COUNTERPARTY if counterparty is None else counterparty,
+                NO_ACCOUNTING_COUNTERPARTY,
+            ),
+        )
+
+        if (raw_treatment := cursor.fetchone()) is not None and raw_treatment[0] is not None:
+            accounting_treatment = TxAccountingTreatment.deserialize_from_db(raw_treatment[0])
+            match accounting_treatment:
```
LefterisJP, line 334: don't use pattern matching for this. Use if/else. It's; just 2 cases as I said in the last review

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +303,151 @@ def query_rules_and_serialize(
...
+                case TxAccountingTreatment.SWAP:
+                    # ensure that the event that we are going to check can exist
+                    # in the related events. Avoid index error
+                    if len(related_events) <= related_event_idx + 1:
```
LefterisJP, line 338: read what I wrote in the last review for this: https://github.com/rotki/rotki/pull/6953#discussion_r1399607695

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +303,151 @@ def query_rules_and_serialize(
...
+        if isinstance(related_event, EvmEvent) is False:
+            continue  # only evm events have callbacks
+
+        # if there is no accounting treatment check for callbacks
+        if (callback_data := callbacks.get(related_event.get_type_identifier())) is None:
+            continue
+
+        affected_events_num, callback = callback_data
+        if affected_events_num == 0:
+            continue
```
LefterisJP, line 364:
```suggestion
        processed_events_num, callback = callback_data
        if processed_events_num == 1:
            continue
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +303,151 @@ def query_rules_and_serialize(
...
+
+        # count the number of events that will be consumed
+        consumed_events = callback(
+            pot=pot,
+            event=related_event,  # type: ignore[arg-type] # mypy doesn't recognize that this is a evm event
```
LefterisJP, line 369:
```suggestion
            event=related_event,  # type: ignore[arg-type] # mypy doesn't recognize that this is an evm event
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +303,151 @@ def query_rules_and_serialize(
...
+            other_events=iter(related_events[related_event_idx + 1:]),  # type: ignore[arg-type]  # mypy doesn't recognize that this is a evm event
```
LefterisJP, line 370:
```suggestion
            other_events=iter(related_events[related_event_idx + 1:]),  # type: ignore[arg-type]  # mypy doesn't recognize that this is an evm event
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +303,151 @@ def query_rules_and_serialize(
...
+        # if there is no accounting treatment check for callbacks
+        if (callback_data := callbacks.get(related_event.get_type_identifier())) is None:
+            continue
+
+        affected_events_num, callback = callback_data
+        if affected_events_num == 0:
+            continue
+
+        # count the number of events that will be consumed
+        consumed_events = callback(
```
LefterisJP, line 367:
```suggestion
        # count the number of events that will be processed if accounting ran
        processed_events_num = callback(
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +303,151 @@ def query_rules_and_serialize(
...
+            pot=pot,
+            event=related_event,  # type: ignore[arg-type] # mypy doesn't recognize that this is a evm event
+            other_events=iter(related_events[related_event_idx + 1:]),  # type: ignore[arg-type]  # mypy doesn't recognize that this is a evm event
+        )
+
+        # check if among the consumed events is the one that we are checking
+        return any(
+            event.identifier == consumed_event.identifier
+            for consumed_event in related_events[related_event_idx: related_event_idx + consumed_events + 1]  # noqa: E501
```
LefterisJP, line 376:
```suggestion
            for consumed_event in related_events[related_event_idx: related_event_idx + processed_events_num]  # noqa: E501
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +303,151 @@ def query_rules_and_serialize(
...
+            continue
+
+        # count the number of events that will be consumed
+        consumed_events = callback(
+            pot=pot,
+            event=related_event,  # type: ignore[arg-type] # mypy doesn't recognize that this is a evm event
+            other_events=iter(related_events[related_event_idx + 1:]),  # type: ignore[arg-type]  # mypy doesn't recognize that this is a evm event
+        )
+
+        # check if among the consumed events is the one that we are checking
```
LefterisJP, line 373:
```suggestion
        # check if the given event is among the processed events
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +303,151 @@ def query_rules_and_serialize(
...
+        return any(
```
LefterisJP, line 374: you are iterating two times here by using any. https://docs.python.org/3/library/functions.html#any

Undo this.

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +303,151 @@ def query_rules_and_serialize(
...
+        if affected_events_num == 0:
+            continue
+
+        # count the number of events that will be consumed
+        consumed_events = callback(
+            pot=pot,
+            event=related_event,  # type: ignore[arg-type] # mypy doesn't recognize that this is a evm event
+            other_events=iter(related_events[related_event_idx + 1:]),  # type: ignore[arg-type]  # mypy doesn't recognize that this is a evm event
+        )
+
```
LefterisJP, line 372: **important**: After going through the logic of this function what are you doing here? Why are you re-iterating the events once you see that `N` events need to be processed. At this point you should just return the number of processed events from this function.

This function should return amount of events to skip. not a boolean.

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +303,151 @@ def query_rules_and_serialize(
...
+            # check the surrounding events to verify if this particular event will
+            # be processed by a previous one. First query the related events
+            related_events = history_db.get_history_events(
+                cursor=cursor,
+                filter_query=HistoryEventFilterQuery.make(event_identifiers=[event.event_identifier]),
+                has_premium=True,
+            )
+
+            event_has_treatment = _check_if_event_affected_by_treatment(
```
LefterisJP, line 444: As I pointed out above, this should also return a number of event to skip. If it needs to skip events then in this for loop you should just skip ahead. Probably best way to do that is turn the for loop into an iterator and conume it.

rotkehlchen/tests/db/test_db_accounting_rules.py
```diff
@@ -207,5 +238,204 @@ def test_missing_accounting_rules_accounting_treatment(database: 'DBHandler') ->
...
+    ) == [False, False]
+
+
+@pytest.mark.parametrize('accountant_without_rules', [True])
+def test_events_affeced_by_others_accounting_treatment_with_fee(
+        database: 'DBHandler',
+        accountant: Accountant,
+) -> None:
+    """
+    Test that if a rule has a accounting treatment with fee then the events
```
LefterisJP, line 313:
```suggestion
    Test that if a rule has an accounting treatment with fee then the events
```

rotkehlchen/tests/db/test_db_accounting_rules.py
```diff
@@ -207,5 +238,204 @@ def test_missing_accounting_rules_accounting_treatment(database: 'DBHandler') ->
...
+        accounting_pot=accountant.pots[0],
+        evm_accounting_aggregator=accountant.pots[0].events_accountant.evm_accounting_aggregators,
+        events=events,
+    ) == [False, False, False]
+
+
+@pytest.mark.parametrize('ethereum_accounts', [['0x7716a99194d758c8537F056825b75Dd0C8FDD89f']])
+@pytest.mark.parametrize('accountant_without_rules', [True])
+@pytest.mark.parametrize('default_mock_price_value', [ONE])
+def test_events_affeced_by_others_callbacks(
```
LefterisJP, line 379:
```suggestion
def test_events_affected_by_others_callbacks(
```

rotkehlchen/db/accounting_rules.py
yabirgb: I like the pattern matching more because you need to be exhaustive and avoids error but okey

rotkehlchen/db/accounting_rules.py
LefterisJP: 1. you won't need this with the changes I recommended.
2.  I removed the +1 since you will have +1 on the processed_events_num

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +304,160 @@ def query_rules_and_serialize(
...
+            event.event_type.serialize(),
+            event.event_subtype.serialize(),
+            NO_ACCOUNTING_COUNTERPARTY if (counterparty := getattr(event, 'counterparty', None)) is None else counterparty,  # noqa: E501
+            NO_ACCOUNTING_COUNTERPARTY,  # account for rules based only on type/subtype
+        ) for event in related_events
+    ]
+
+    callbacks = evm_accounting_aggregator.get_accounting_callbacks()
+    bindings_and_events_iterator = zip(bindings, related_events)
+    processed_events: dict[int, bool] = {}  # map of all the events related to the one requested to wether they will be processed in accounting or not  # noqa: E501
```
LefterisJP, line 413: Why is this mapping needed? You already have the cache right?

```suggestion
    processed_events: dict[int, bool] = {}  # map of events to whether they are processed in accounting or not  # noqa: E501
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +304,160 @@ def query_rules_and_serialize(
...
+        cursor: 'DBCursor',
+        callbacks: dict[int, tuple[int, EventsAccountantCallback]],
+        events_iterator: Iterator[tuple[tuple[Any, ...], 'HistoryEvent']],
+        next_events: Sequence[HistoryBaseEntry],
+        event: HistoryBaseEntry,
+        pot: 'AccountingPot',
+) -> list[int]:
+    """
+    Returns a list of ids processed after checking possible accounting
+    treatments and callbacks.
```
LefterisJP, line 319:
```suggestion
    Returns a list of event identifiers processed after checking possible accounting
    treatments and callbacks.
```

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +304,160 @@ def query_rules_and_serialize(
...
+            if isinstance(event, EthStakingEvent) is True:
+                # staking events all have a process() function for accounting
+                processed_events[event.identifier] = False
+                current_event_index += 1
+                continue
+
+            # check the rule in the database
+            row = cursor.execute(query, event_binding).fetchone()
+            is_missing_rule = row[0] == 0
+            processed_events[event.identifier] = is_missing_rule
```
LefterisJP, line 435: The reason you don't stop here if this is False is to also populate the callback if existing? Why exactly?

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +304,160 @@ def query_rules_and_serialize(
...
+    # count the number of events that will be processed if accounting ran
+    processed_events_num = callback(
+        pot=pot,
+        event=event,  # type: ignore[arg-type] # mypy doesn't recognize that this is an evm event
+        other_events=iter(next_events),  # type: ignore[arg-type]  # mypy doesn't recognize that this is an evm event
+    )
+
+    for _ in range(processed_events_num - 1):  # -1 because we exclude the current event here
+        _, next_event = next(events_iterator)
+        assert next_event.identifier is not None, 'We should only process events with identifiers'
```
LefterisJP, line 371: for all these assertions perhaps please just make the to type: ignores

rotkehlchen/db/accounting_rules.py
```diff
@@ -336,3 +304,160 @@ def query_rules_and_serialize(
...
+
+                # update information about the new events
+                for processed_event_id in new_missing_accounting_rule:
+                    accountant.processable_events_cache.add(processed_event_id, False)
+                    processed_events[processed_event_id] = False
+
+    # get the list of wether events are processable for the events requested and not the requested
+    # plus the related to them
+    is_missing_rule_list = [processed_events.get(event.identifier, False) for event in events]  # type: ignore[arg-type]
+    return is_missing_rule_list
```
LefterisJP, line 463: Why not use the cache directly? Why do you need `processed_events` to exist at all?

```suggestion
    return [processed_events.get(event.identifier, False) for event in events]  # type: ignore[arg-type]
```

rotkehlchen/db/accounting_rules.py
yabirgb: Because the event in addition to having an accounting rule could have a callback that affects events that come after and is not enough to check the accounting rule

LefterisJP (approved):
LGTM. Only failed test is an etherscan timeout.

```
WARNING  rotkehlchen.chain.evm.node_inquirer:node_inquirer.py:539 Greenlet-278: Failed to query NodeName(name='etherscan', endpoint='', owned=False, blockchain=<SupportedBlockchain.ETHEREUM: 'ETH'>) for <bound method EvmNodeInquirer._call_contract of <rotkehlchen.chain.ethereum.node_inquirer.EthereumInquirer object at 0x7f0c50e5e3e0>> due to Unexpected format of ethereum Etherscan response for request https://api.etherscan.io/api?module=proxy&action=eth_call&to=0x06FE76B2f432fdfEcAEf1a7d4f6C3d41B5861672&data=0x85c6a7930000000000000000000000005c4d8cee7de74e31ce69e76276d862180545c3070000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000002800000000000000000000000000000000000000000000000000000000000005000000000000000000000000000000000000000000000000000000000000000540000000000000000000000000000000000000000000000000000000000000058000000000000000000000000000000000000000000000000000000000000005c000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000640000000000000000000000000000000000000000000000000000000000000068000000000000000000000000000000000000000000000000000000000000006c000000000000000000000000000000000000000000000000000000000000007000000000000000000000000000000000000000000000000000000000000000740000000000000000000000000000000000000000000000000000000000000078000000000000000000000000000000000000000000000000000000000000007c000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000840000000000000000000000000000000000000000000000000000000000000088000000000000000000000000000000000000000000000000000000000000008c000000000000000000000000000000000000000000000000000000000000009000000000000000000000000000000000000000000000000000000000000000940000000000000000000000000000000000000000000000000000000000000098000000000000000000000000000000000000000000000000000000000000009c00000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000000a400000000000000000000000000000000000000000000000000000000000000a800000000000000000000000000000000000000000000000000000000000000ac00000000000000000000000000000000000000000000000000000000000000b000000000000000000000000000000000000000000000000000000000000000b400000000000000000000000000000000000000000000000000000000000000b800000000000000000000000000000000000000000000000000000000000000bc00000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000000c400000000000000000000000000000000000000000000000000000000000000c800000000000000000000000000000000000000000000000000000000000000cc00000000000000000000000000000000000000000000000000000000000000d000000000000000000000000000000000000000000000000000000000000000d400000000000000000000000000000000000000000000000000000000000000d800000000000000000000000000000000000000000000000000000000000000dc00000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000e400000000000000000000000000000000000000000000000000000000000000e800000000000000000000000000000000000000000000000000000000000000ec0000000000000000000000000000000000000000000000000000000000000001a437572766520e280a2204c697175696469747920476175676573000000000000000000000000000000000000000000000000000000000000000000000000001179676f762e66696e616e63652028763129000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001179676f762e66696e616e6365202876322900000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000136d537461626c6520e280a2205374616b696e6700000000000000000000000000000000000000000000000000000000000000000000000000000000000000001b53776572766520e280a2204c697175696469747920476175676573000000000000000000000000000000000000000000000000000000000000000000000000185069636b6c652046696e616e636520e280a2204661726d730000000000000000000000000000000000000000000000000000000000000000000000000000001a5069636b6c652046696e616e636520e280a2205374616b696e6700000000000000000000000000000000000000000000000000000000000000000000000000104161766520e280a2205374616b696e67000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000016432e522e452e412e4d2e20e280a2205374616b696e67000000000000000000000000000000000000000000000000000000000000000000000000000000000013436f6d706f756e6420476f7665726e616e636500000000000000000000000000000000000000000000000000000000000000000000000000000000000000000c7a6c6f742e66696e616e63650000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000846696e4e65787573000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e5069636b6c652046696e616e63650000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004444f444f000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000007426572657a6b61000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003625a780000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000a432e522e452e412e4d2e0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000065377657276650000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000b53617368696d69537761700000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000007486172766573740000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000064b494d434849000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000095375736869537761700000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000c4e65787573204d757475616c000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000094d6f6f6e6973776170000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000054d617469630000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000006417261676f6e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000054d656c6f6e0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000018796561726e2e66696e616e636520e280a2205661756c7473000000000000000000000000000000000000000000000000000000000000000000000000000000094b656570657244414f000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000076d537461626c650000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000084b7962657244414f000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000d4444455820e280a22053706f7400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000f4444455820e280a2204d617267696e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000104444455820e280a2204c656e64696e6700000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000a416d706c65666f7274680000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000104d616b657220476f7665726e616e636500000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000f476e6f7369732050726f746f636f6c0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001543686920476173746f6b656e2062792031696e63680000000000000000000000000000000000000000000000000000000000000000000000000000000000001649646c6520e280a2205269736b2d41646a75737465640000000000000000000000000000000000000000000000000000000000000000000000000000000000174161766520e280a220556e6973776170204d61726b6574000000000000000000&apikey=R7QNMZJF1Z5EZM96GMSZSQKHQK3V2TBKW5. Missing a result in response. Response was: {"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"execution aborted (timeout = 5s)"}}

ERROR    rotkehlchen.chain.evm.node_inquirer:node_inquirer.py:550 Greenlet-278: Failed to query <bound method EvmNodeInquirer._call_contract of <rotkehlchen.chain.ethereum.node_inquirer.EthereumInquirer object at 0x7f0c50e5e3e0>> after trying the following nodes: ['etherscan']
```

PR #6965, opened 2023-11-22
HotFix assets 1.31

yabirgb:
This PR is a hot fix adding locations needed during the assets update. It:

- Adds new locations for assets
- Handles a deserialization error when the ChainID is unknown during assets updates. We were using the enum directly insted of using the deserialize method.
- Adds mappings for assets in exchanges


rotkehlchen/globaldb/updates.py
```diff
@@ -267,7 +271,7 @@ def _parse_evm_token_data(
             insert_text=insert_text,
         )
         if chain_value is not None:
-            chain_id = ChainID(chain_value)
+            chain_id = ChainID.deserialize(chain_value)
```
LefterisJP, line 274: the potential `DeserializationError` here is already handled from what I see, right?

LefterisJP (approved):
LGTM

PR #6968, opened 2023-11-22
feat: add changelog linking message after update

kelsos:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/composables/update-message.ts
+++ b/frontend/app/src/composables/update-message.ts
@@ -1,0 +1,34 @@
+export const useUpdateMessage = createSharedComposable(() => {
+  const lastUsedVersion = useLocalStorage<string>('rotki.last_version', null);
+  const showReleaseNotes = ref(false);
+
+  const { appVersion } = storeToRefs(useMainStore());
+
+  const link = computed(
+    () => `https://github.com/rotki/rotki/releases/tag/${get(appVersion)}`
+  );
+
+  const setShowNotes = (appVersion: string, lastUsed: string): void => {
+    if (!appVersion || appVersion.includes('dev')) {
+      return;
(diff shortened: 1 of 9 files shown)


frontend/app/src/pages/user/login/index.vue
```diff
@@ -92,6 +93,7 @@ onMounted(async () => fetchMessages());
         <p v-if="false" class="text-body-2 text-rui-primary">
           {{ t('login.welcome_update_message') }}
         </p>
+        <NewReleaseChangelog v-if="showReleaseNotes" class="mt-4" />
         <WelcomeMessageDisplay
```
kelsos, line 97: maybe the second one should be v-else-if so that they don't both show at the same time?

frontend/app/src/composables/update-message.ts
```diff
@@ -0,0 +1,34 @@
...
+  const showReleaseNotes = ref(false);
+
+  const { appVersion } = storeToRefs(useMainStore());
+
+  const link = computed(
+    () => `https://github.com/rotki/rotki/releases/tag/${get(appVersion)}`
+  );
+
+  const setShowNotes = (appVersion: string, lastUsed: string): void => {
+    if (!appVersion || appVersion.includes('dev')) {
```
tewshi, line 12: will this not prevent it from showing on dev...

kelsos: The idea here is that the message should only show before the first login on a new production version.

kelsos: It doesn't make sense to show it on dev since the link will not work either way.

tewshi (approved):
lgtm!

PR #6971, opened 2023-11-22
Perform some kraken tweaks after reading a user's logs

LefterisJP:
- Add handling for kraken ledger type: invite bonus
- Note that we don't process the margin trade kraken ledger entries and name which they are according to docs
- Adjust the used_query ranges saved for kraken ledgers query. Do not save anything if there was an error


yabirgb (approved):
lgtm

PR #6975, opened 2023-11-23
Remove address from the filter for undecoded transactions

yabirgb:
We noticed that the query to decode undecoded transactions was failing because the filter when used with addresses was only checking for transactions where the address in to/from fields and this excluded other transactions where there was a transfer executed by another contract like in bridges.


LefterisJP (approved):
LGTM

PR #6978, opened 2023-11-23
Feat: generic error for paginated data calls

tewshi:
## Checklist

### DIFF
--- a/frontend/app/src/composables/filter-paginate.ts
+++ b/frontend/app/src/composables/filter-paginate.ts
@@ -218,7 +218,7 @@ export const usePaginationFilters = <
         }
 
         logger.error(error);
-        if (code && ['400', '500'].includes(code)) {
+        if (Number(code) >= 400) {
           notify({
             title: t('error.generic.title'),
             message: t('error.generic.message', { code, message, path }),
--- a/frontend/app/src/composables/filter-paginate.ts
+++ b/frontend/app/src/composables/filter-paginate.ts
@@ -58,6 +58,8 @@ export const usePaginationFilters = <
     };
   } = {}
 ) => {
+  const { t } = useI18n();
+  const { notify } = useNotificationsStore();
   const router = useRouter();
   const route = useRoute();
   const paginationOptions: Ref<TablePagination<V>> = ref(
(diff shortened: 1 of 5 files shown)


frontend/app/src/composables/filter-paginate.ts
```diff
@@ -202,7 +204,29 @@ export const usePaginationFilters = <
     {
       immediate: false,
       resetOnExecute: false,
-      delay: 0
+      delay: 0,
+      onError(e) {
+        const { notify } = useNotificationsStore();
+        const { t } = useI18n();
```
kelsos, line 210: Shouldn't we pull this on the upper composable scope? 
I am thinking if there is a chance that these two run outside of the right context and failing.

frontend/app/src/composables/filter-paginate.ts
```diff
@@ -202,7 +204,29 @@ export const usePaginationFilters = <
...
+        const path = error.config?.url;
+        let { message, code } = error;
+
+        if (error.response) {
+          message = error.response.data.message;
+          code = error.response.status.toString();
+        }
+
+        logger.error(error);
+        if (code === '500') {
```
kelsos, line 221: Maybe show the error on any 400/500 error?

In general the api should not fail but if it returns any 400/500 failure we should notify the user.

frontend/app/src/composables/filter-paginate.ts
tewshi: ok, will move them up

frontend/app/src/composables/filter-paginate.ts
tewshi: okay, will do

frontend/app/src/composables/filter-paginate.ts
```diff
@@ -218,7 +218,7 @@ export const usePaginationFilters = <
         }
 
         logger.error(error);
-        if (code === '500') {
+        if (code && ['400', '500'].includes(code)) {
```
kelsos, line 221: This is not exactly what I meant though, this will only show the error for 400 and 500 explicitly. I had in mind more of https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses so any 4xx or 5xx error code

kelsos (approved):
lgtm

PR #6982, opened 2023-11-24
ws message for transactions by hash

yabirgb:
1. Set the ws message when decoding transactions by hash so it can be used by the frontend
2. Remove code only used in tests for the accountant


LefterisJP (approved):
LGTM

PR #6983, opened 2023-11-24
Peek at next event in accounting to figure out if fee exists

LefterisJP:
This needs: https://github.com/rotki/data/pull/38

- Remove the SWAP_WITH_FEE treatment as it is rendered useless
- Use peekable from more_itertools (it's already a dependency via other packages -- just made it explicit)
- Nasty typing hacks due to the in-between implementation of accounting at the moment.


rotkehlchen/accounting/history_base_entries.py
```diff
@@ -21,26 +23,6 @@
 T = TypeVar('T', bound=HistoryBaseEntry)
 
 
-def history_base_entries_iterator(
```
LefterisJP, line 24: This I see and am wondering what the fuck were we thinking. This seems to be just for typing and error logging. And to re-ierate the iterator every time it is called?

rotkehlchen/chain/evm/accounting/structures.py
```diff
@@ -15,7 +16,7 @@ def __call__(
             self,
             pot: 'AccountingPot',
             event: 'EvmEvent',
-            other_events: Iterator['EvmEvent'],
+            other_events: "peekable['EvmEvent']",
```
LefterisJP, line 19: Adding this in `""` is only due to this limitation: https://github.com/python/mypy/issues/5264#issuecomment-399407428

Which is probably still an issue?

As mypy works fine, but when running the code it goes BOOM. With

```
rotkehlchen/chain/evm/accounting/structures.py:19: in EventsAccountantCallback                                                                                                                                                                 
    other_events: peekable['EvmEvent'],                                                                                                                                                                                                        
E   TypeError: 'type' object is not subscriptable 
```

rotkehlchen/accounting/history_base_entries.py
yabirgb: is not re-itearting it, just yielding. So consume the next event, check and yield the consumed event

rotkehlchen/db/accounting_rules.py
```diff
@@ -339,8 +340,9 @@ def _events_to_consume(
         if accounting_treatment == TxAccountingTreatment.SWAP:
             _, next_event = next(events_iterator)
             ids_processed.append(next_event.identifier)  # type: ignore[arg-type]
-        else:
-            for _ in range(2):  # swap_with_fee consumes two events
+            _, peeked_event = events_iterator.peek()
+            if peeked_event.event_identifier == event.event_identifier and peeked_event.event_subtype == HistoryEventSubType.FEE:  # noqa: E501
+                # consume the related fee if it exists
```
yabirgb, line 345: this affects

```
FAILED rotkehlchen/tests/db/test_db_accounting_rules.py::test_events_affeced_by_others_accounting_treatment[True-True] - StopIteration
FAILED rotkehlchen/tests/db/test_db_accounting_rules.py::test_events_affeced_by_others_accounting_treatment_with_fee[True-True] - assert [False, False, True] == [False, False, False]
```
I believe

PR #6987, opened 2023-11-24
Pick correct accounting treatment

yabirgb:
When a event is affected by both a generic rule and a rule with counterparty we were using the treatment from the generic rule instead of the one for the more specific rule due to the order in which they appeared in the query.

I went with the approach of making two queries because the we might in the future introduce more types of treatments and then it might be possible to have to rules for an event (generic and specific) one with one treatment `A` and the other with treatment `B`. To prevent such issues I decided to be explicit in what is choosen


LefterisJP (approved):
LGTM

PR #6997, opened 2023-11-24
Fix welcome message during version upgrade broken tag link

LefterisJP:
The welcome message at first login after a version upgrade will now have the correct link to the release notes.

It was a broken link since we prepend our tags with a `v` and it was missing that.


yabirgb (approved):
:(

tewshi (approved):
lgtm

PR #7000, opened 2023-11-25
Python 3.10 style type annotation

LefterisJP:
## [Switch to python 3.10 type annotation style](https://github.com/rotki/rotki/commit/9426bf5b00adb8e5eca70829279365fc8c53f639) 

Use ruff's auto-fix for this. Along with some manual tweaking.

## [Solve B905 error. zip() without specifying strict argument](https://github.com/rotki/rotki/commit/dbefae325bfdc16b71b9a9a29084d3f4403d0a4e) 

The purpose of this argument is to let the reader know if it's okay to
have zip silently create a zip of the length of the shortest
list (strict=False) or if the zipped lists are not equal length to
raise a `ValueError` (strict=True)

-----

Adding this to bugfixes due to it being a huge diff and wanting to avoid conflicts.

### DIFF
--- a/package.py
+++ b/package.py
@@ -9,10 +9,10 @@ import stat
 import subprocess
 import sys
 import urllib.request
-from collections.abc import Generator
+from collections.abc import Callable, Generator
 from pathlib import Path
 from tempfile import NamedTemporaryFile
-from typing import Any, Callable, Literal, Optional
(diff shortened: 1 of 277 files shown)


LefterisJP:
Finally all green. Merging.

PR #7001, opened 2023-11-26
Invalidate cache for accounting rules

yabirgb:
Editing accounting rules was not expiring the cache for processed events making that if you modify them the frontend won't correctly update this infomation

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -38,7 +38,11 @@ from rotkehlchen.accounting.export.csv import (
 )
 from rotkehlchen.accounting.pot import AccountingPot
 from rotkehlchen.accounting.structures.balance import Balance, BalanceType
-from rotkehlchen.accounting.structures.base import HistoryBaseEntryType, StakingEvent
+from rotkehlchen.accounting.structures.base import (
+    HistoryBaseEntryType,
+    StakingEvent,
+    get_event_type_identifier,
+)
 from rotkehlchen.accounting.structures.evm_event import EvmProduct
 from rotkehlchen.accounting.structures.processed_event import AccountingEventExportType
 from rotkehlchen.accounting.structures.types import (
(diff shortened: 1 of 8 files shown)


rotkehlchen/api/rest.py
```diff
@@ -4474,6 +4479,22 @@ def export_history_events(
 
         return api_response(OK_RESULT, status_code=HTTPStatus.OK)
 
+    def _invalidate_accounting_rule_cache(
```
LefterisJP, line 4482: you don't invalidate the entire cache. I got confused by the name
```suggestion
    def _invalidate_cache_for_accounting_rule(
```

rotkehlchen/api/rest.py
```diff
@@ -4518,15 +4544,40 @@ def update_accounting_rule(
...
+        self._invalidate_accounting_rule_cache(
+            event_type=event_type,
+            event_subtype=event_subtype,
+            counterparty=counterparty,
+        )
         return api_response(_wrap_in_ok_result(True), status_code=HTTPStatus.OK)
 
     def delete_accounting_rule(self, rule_id: int) -> Response:
         db = DBAccountingRules(self.rotkehlchen.data.db)
+        with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
```
LefterisJP, line 4556: What are you doing here? Isn't this (and shouldn't this) logic be part of `db.remove_accounting_rule()`? It raises an `InputError` if this happens.

rotkehlchen/tests/api/test_accounting_rules.py
```diff
@@ -365,3 +374,127 @@ def test_listing_conflicts(
         },
     ]
     assert result['entries_total'] == 2
+
+
+@pytest.mark.parametrize('initialize_accounting_rules', [False])
+def test_cache_invalidation(rotkehlchen_api_server: APIServer):
+    """
+    Test that cache for events affected by accounting rules gets correctly invalidated
```
LefterisJP, line 382:
```suggestion
    Test that the cache for events affected by an accounting rule gets correctly invalidated
```

rotkehlchen/tests/api/test_accounting_rules.py
```diff
@@ -365,3 +374,127 @@ def test_listing_conflicts(
...
+    when operations happen modifying accounting rules that affect them.
```
LefterisJP, line 383:
```suggestion
    when operations happen modifying the rule that affects them.
```

rotkehlchen/api/rest.py
yabirgb: I didn't do it because it was a extra query that might not be needed in other places but this function is used only here so I did as you suggested

LefterisJP (approved):
LGTM.

Single failed test is etherscan problem. `test_query_all_protocol_balances_for_account`

```
[27/11/2023 09:03:11 UTC] ERROR rotkehlchen.chain.evm.node_inquirer Main Greenlet: Failed to query <bound method EvmNodeInquirer._call_contract of <rotkehlchen.chain.ethereum.node_inquirer.EthereumInquirer object at 0x7f371a100430>> after trying the following nodes: ['etherscan']
```

Issue #6885, opened 2023-11-03
Add option to opt out from auto-detecting/adding layer2s like Polygon to avoid spam tokens appearing in portfolio


kilrau:
## Abstract

I'd like to opt out from this
![image](https://github.com/rotki/rotki/assets/19181985/0197f2ca-9240-49e0-bd1c-25b3e6276fdd)
to prevent this
![image](https://github.com/rotki/rotki/assets/19181985/80c46e2f-3c17-48ef-96b6-b053e6bf64df)

## Motivation

Quite tedious to go to `Accounts & Balances` -> `Polygon POS BALANCES` -> Select all -> Delete every time I'm opening rotki. Also I would claim in general the behavior to auto-add Polygon is rather questionable since I've never used Polygon with my addresses in rotki before.

## Specification

`v1.30.2`

LefterisJP:
Hey @kilrau this is a valid request. Polygon is unfortunately a source of spam and many users use it so the problem is figuring ourt spam versus real events.

It's an ever continuing battle. Please hop on to our discord if you wanna chat more about it: https://discord.rotki.com/

But yeah opting out of certain EVM chains completely is a good ask.

kilrau:
Joined your discord.

> opting out of certain EVM chains completely is a good ask.

That's basically it. I know which chains I'm interested in in which not :)

kilrau:
Can you consider changing default to opt-in instead of opt-out? I think you should be able to use the merged backend PR as-is just let the frontend opt-out from all by default. Today Gnosis was suddenly added and I don't want to spend time opting out from newly added chains forever.

In future considering scanning chains in the background and only auto-add if >0 value was detected.

yabirgb:
Closing the issue as completed in https://github.com/rotki/rotki/pull/7016

PR #7003, opened 2023-11-26
Add setting to determine EVM chains for which to skip autodetection

LefterisJP:
Backend part of https://github.com/rotki/rotki/issues/6885
Also does the renaming mentioned in https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=45696490


rotkehlchen/db/settings.py
```diff
@@ -351,6 +359,11 @@ def serialize_db_setting(
         value = value.serialize()  # pylint: disable=no-member
     elif setting == 'address_name_priority' and is_modifiable is True:
         value = json.dumps(value)
+    elif setting == 'evmchains_to_skip_detection':
+        if is_modifiable is True:
+            value = json.dumps([x.to_name() for x in value])
+        else:
+            value = [x.to_name() for x in value]
```
yabirgb, line 366: maybe is better to just do 

```suggestion
        value = [x.to_name() for x in value]
        if is_modifiable is True:
            value = json.dumps(value)
```

yabirgb (approved):
lgtm

rotkehlchen/db/settings.py
LefterisJP: maybe later since all green now

PR #7008, opened 2023-11-27
Be more aggresive in spam token detection

yabirgb:
Changed the checks made in the spam detection based on symbols to cover a wider range of tokens.

There are some cases where they use a url without the protocol param and is harder to filter since for example bridged USDC uses `USDC.e` and similar tokens could fall in this


rotkehlchen/assets/utils.py
```diff
@@ -30,6 +31,17 @@
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
+SPAM_ASSET_REGEX = r"""
+    https:// |                          # Matches 'https://'
```
LefterisJP, line 35: This would match https only in the beginning right? Or not?  Same with Claim or visit?

Need more tests as I wrote in discord

LefterisJP (commented):
So left a review but still I am missing a way for users to solve false positives.

We have false positives already in your test. We can't release this with no ability to mark something as false positive as I already told you in discord.

rotkehlchen/tasks/assets.py
```diff
@@ -0,0 +1,42 @@
...
+    'A.identifier=B.identifier WHERE B.PROTOCOL IS NOT ? ORDER BY C.symbol'
+)
+
+
+def autodetect_spam_assets_in_db(user_db: DBHandler) -> None:
+    """Autodetect spam tokens already in the globaldb and ignore them"""
+    globaldb = GlobalDBHandler()
+    detected_spam_assets = []
+
+    # get the spam tokens already in the globaldb
```
LefterisJP, line 19: This comment really confused me

```suggestion
    # Check if any tokens in the global DB not marked as spam are actually matching the spam patterns
```

rotkehlchen/tasks/assets.py
```diff
@@ -0,0 +1,42 @@
...
+        for symbol, name, identifier in cursor:
+            if check_if_spam_token(symbol=symbol, name=name) is False:
+                continue
+
+            detected_spam_assets.append(identifier)
+
+    if len(detected_spam_assets) == 0:
+        return
+
+    # set the spam protocols for the new detected tokens
```
LefterisJP, line 31:
```suggestion
    # set the spam protocol for the newly detected tokens
```

rotkehlchen/tests/unit/globaldb/test_globaldb.py
```diff
@@ -1223,10 +1233,79 @@ def test_for_spam_tokens(database: 'DBHandler', ethereum_inquirer) -> None:
...
+    # check that for the tokens in the database we wouldn't raise a false positive
+    # this checks also for:
+    # - yearn tokens
+    # - convex tokens
+    # - curve pools
+    # - velodrome pools
+    # in addition to normal tokens
+    with globaldb.conn.read_ctx() as cursor:
+        for asset_row in cursor.execute(SYMBOL_AND_NAME_ASSETS_QUERY, (SPAM_PROTOCOL,)):
+            if asset_row[2] in (  # These are known false positives already in GlobalDB -- which are a grey area  # noqa: E501
```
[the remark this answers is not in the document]
yabirgb: I will make it a subset and just show a warning for the others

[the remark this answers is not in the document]
yabirgb: I was thinking in just having warning but is true that it doesn't make much sense. Will remove this section

rotkehlchen/tasks/assets.py
```diff
@@ -0,0 +1,51 @@
...
+            query,
+            [(SPAM_PROTOCOL, identifier) for identifier in detected_spam_assets],
+        )
+
+    # update the user list of ignored assets
+    with user_db.conn.write_ctx() as write_cursor:
+        for asset_id in detected_spam_assets:
+            user_db.add_to_ignored_assets(write_cursor=write_cursor, asset=Asset(asset_id))
+
+        write_cursor.execute(  # remember last time evm addresses were detected
```
LefterisJP, line 48: Dude come on ...

```suggestion
        write_cursor.execute(  # remember last time spam detection ran
```

LefterisJP (approved):
LGTM. Left a nitipck but will merge.

PR #7020, opened 2023-11-28
Restructure the rotki data directory

LefterisJP:
Properly structure the rotki data directory so that data is organized properly in various directories and not just all thrown under a root directory with potential clashes between usernames and other stuff.

Fix https://github.com/rotki/rotki/issues/4841

### DIFF
--- a/rotkehlchen/tests/test_users.py
+++ b/rotkehlchen/tests/test_users.py
@@ -5,6 +5,7 @@ from unittest import mock
 
 import pytest
 
+from rotkehlchen.constants.misc import USERDB_NAME, USERSDIR_NAME
 from rotkehlchen.data_handler import DataHandler
 from rotkehlchen.errors.misc import SystemPermissionError
 from rotkehlchen.user_messages import MessagesAggregator
@@ -99,12 +100,13 @@ def test_users_query_permission_error(
         function_scope_messages_aggregator: MessagesAggregator,
         sql_vm_instructions_cb: int,
 ):
-    not_allowed_dir = os.path.join(data_dir, 'notallowed')
-    allowed_user_dir = os.path.join(data_dir, 'allowed_user')
-    os.mkdir(not_allowed_dir)
+    users_dir = data_dir / USERSDIR_NAME
+    not_allowed_dir = users_dir / 'notallowed'
+    allowed_user_dir = users_dir / 'allowed_user'
+    not_allowed_dir.mkdir(parents=True)
     os.chmod(not_allowed_dir, 0o200)
-    os.mkdir(allowed_user_dir)
-    Path(Path(allowed_user_dir) / 'rotkehlchen.db').touch()
+    allowed_user_dir.mkdir(parents=True)
+    (allowed_user_dir / USERDB_NAME).touch()
     handler = DataHandler(
         data_directory=data_dir,
         msg_aggregator=function_scope_messages_aggregator,
--- a/rotkehlchen/utils/datadir.py
+++ b/rotkehlchen/utils/datadir.py
@@ -1,0 +1,140 @@
+import logging
+from pathlib import Path
+from shutil import Error, copytree, move, rmtree
+
+from rotkehlchen.constants.misc import (
+    ALLASSETIMAGESDIR_NAME,
+    APPDIR_NAME,
+    ASSETIMAGESDIR_NAME,
+    AVATARIMAGESDIR_NAME,
+    CUSTOMASSETIMAGESDIR_NAME,
+    GLOBALDIR_NAME,
+    IMAGESDIR_NAME,
+    MISCDIR_NAME,
+    USERDB_NAME,
+    USERSDIR_NAME,
+)
+from rotkehlchen.logging import RotkehlchenLogsAdapter
+from rotkehlchen.utils.misc import ts_now
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+def _create_directory_with_potential_backup(data_dir: Path, name: str) -> Path:
+    """Create a subdirectory under the main data directory. If the directory
+    already exists and has a user database in it, that means that someone made
+    a user with that name. In which edge case we will try to rename and backup"""
+    subdir = data_dir / name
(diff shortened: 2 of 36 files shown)


rotkehlchen/utils/datadir.py
```diff
@@ -0,0 +1,136 @@
...
+            elif x.is_dir() and x.name == 'airdrops':
+                airdrops_path = x
+            elif x.is_dir() and x.name == 'airdrops_poap':
+                airdrops_poap_path = x
+            elif x.is_dir() and x.name == MISCDIR_NAME:
+                misc_path = x
+            elif x.is_dir() and x.name == 'icons':
+                icons_path = x
+            else:
+                other_paths.append(x)
```
yabirgb, line 111: isn't it better to do

```suggestion
            if not x.is_dir():
                other_paths.append(x)
                continue
  
            if  (x / USERDB_NAME).exists():
                user_paths.append(x)
            elif x.name == 'global_data':
                global_path = x
            elif x.name == 'airdrops':
                airdrops_path = x
            elif x.name == 'airdrops_poap':
                airdrops_poap_path = x
            elif x.name == MISCDIR_NAME:
                misc_path = x
            elif x.name == 'icons':
                icons_path = x
            else:
                other_paths.append(x)
```

rotkehlchen/utils/datadir.py
```diff
@@ -0,0 +1,136 @@
...
+        target_path_dir = target_dir / path.name
+        try:
+            move(path, target_path_dir)
+        except (PermissionError, Error) as e:
+            log.error(f'During restructuring data directory failed to copy {path} to {target_path_dir} due to {e}')  # noqa: E501
+
+
+def _handle_images(data_dir: Path, icons_path: Path | None) -> None:
+    if icons_path is not None:
+        # create the images data directory and move paths there
```
yabirgb, line 54: this one is missing a docstring

rotkehlchen/utils/datadir.py
```diff
@@ -0,0 +1,136 @@
...
+    if (data_dir / USERSDIR_NAME).is_dir() and (data_dir / GLOBALDIR_NAME).is_dir():
+        return  # already restructured
+
+    user_paths = []
+    other_paths = []
+    global_path = None
+    airdrops_path = None
+    airdrops_poap_path = None
+    misc_path = None
+    icons_path = None
```
yabirgb, line 95: just in case you didn't know it is possible to do

```suggestion
    global_path = airdrops_path = airdrops_poap_path = misc_path = icons_path = None
```

rotkehlchen/utils/datadir.py
```diff
@@ -0,0 +1,136 @@
...
+        all_assets_dir = targetpath
+        for name, target_dir in ((AVATARIMAGESDIR_NAME, img_data_dir), (CUSTOMASSETIMAGESDIR_NAME, assets_dir)):  # noqa: E501
+            subdir = all_assets_dir / name
+            if subdir.exists():
+                targetpath = target_dir / name
+                try:
+                    move(subdir, targetpath)
+                except (PermissionError, Error) as e:
+                    log.error(f'During restructuring data directory failed to move {subdir} to {targetpath} due to {e}')  # noqa: E501
+                    continue
```
yabirgb, line 75: don't know if you prefer to reduce the indentation modifying the flow but

```suggestion
            if not subdir.exists():
                continue
            targetpath = target_dir / name
            try:
                move(subdir, targetpath)
            except (PermissionError, Error) as e:
               log.error(f'During restructuring data directory failed to move {subdir} to {targetpath} due to {e}')  # noqa: E501
```

also the last continue isn't needed since there isn't anything after ir

rotkehlchen/utils/datadir.py
```diff
@@ -0,0 +1,136 @@
...
+
+def _create_directory_with_potential_backup(data_dir: Path, name: str) -> Path:
+    """Create a subdirectory under the main data directory. If the directory
+    already exists and has a user database in it, that means that someone made
+    a user with that name. In which edge case we will try to rename and backup"""
+    subdir = data_dir / name
+    if (subdir / USERDB_NAME).exists():
+        # someone had a user, called name ... and did not get moved with initial for loop
+        try:  # this should not happen actually ...
+            copytree(subdir, f'{ts_now()}_backup_user')
```
yabirgb, line 32: maybe for debugging it could be interesting to have a log message here?

rotkehlchen/tests/test_users.py
```diff
@@ -104,7 +105,7 @@ def test_users_query_permission_error(
     os.mkdir(not_allowed_dir)
     os.chmod(not_allowed_dir, 0o200)
     os.mkdir(allowed_user_dir)
-    Path(Path(allowed_user_dir) / 'rotkehlchen.db').touch()
+    Path(Path(allowed_user_dir) / USERDB_NAME).touch()
```
yabirgb, line 108: this second path is not need... is it?

```suggestion
    (Path(allowed_user_dir) / USERDB_NAME).touch()
```

rotkehlchen/utils/datadir.py
LefterisJP: no it's not. The point is to move anything not fiting into other paths. Including directories.

rotkehlchen/utils/datadir.py
LefterisJP: no the old code is better here. Readability above all.

LefterisJP: hm no no no no. I thought it's like the other suggestion in my previous comment. I am applying this change too.

rotkehlchen/tests/test_users.py
LefterisJP: actuallly none of them are!

rotkehlchen/utils/datadir.py
LefterisJP: Okay I will add it. It's trivial difference. A bit harder to read, but also perhaps a tiny bit faster due to the x.is_dir() check. Had not seen the else when I made the comment above.

LefterisJP:
All green. Merging.

PR #7031, opened 2023-11-30
Improve UX for uniswap and sushiswap tokens

### LINKED ISSUE #7024: Uniswap LP tokens are missing the protocol information
## Abstract

A user reported to me issues querying prices for Uniswap LP tokens. It turns out that the issue was that the asset was missing the `UNI-V2` protocol that is used by the inquierer to execute special logic and query the prices for the LP. In addition is hard to identify the asset since all use the same `UNI-V2` symbol.

## Specification

1. In https://github.com/rotki/rotki/blob/119397e52cc4ea5210fd50093c7e5098fdff16f9/rotkehlchen/chain/ethereum/modules/uniswap/v2/common.py#L297 we are not adding the information about the protocol. In the case of uniswap we have special logic to query prices https://github.com/rotki/rotki/blob/119397e52cc4ea5210fd50093c7e5098fdff16f9/rotkehlchen/inquirer.py#L163C30-L163C30
2. When adding or editing the token since we know the underlying tokens we can add their symbol to the name and symbol of the lp token

yabirgb:
1. Fixes a case where if the user extended the symbol of a uniswap token making the logic fail to populate correctly the protocol of the token
2. Adds information about underlying tokens to symbols of uniswap and sushiswap LP tokens

imagen

Closes #7024

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/uniswap/v2/common.py
+++ b/rotkehlchen/chain/ethereum/modules/uniswap/v2/common.py
@@ -7,22 +7,21 @@ from web3 import Web3
 
 from rotkehlchen.accounting.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.assets.asset import CryptoAsset, EvmToken, UnderlyingToken
+from rotkehlchen.assets.resolver import AssetResolver
 from rotkehlchen.assets.utils import (
     TokenEncounterInfo,
     get_or_create_evm_token,
     set_token_protocol_if_missing,
 )
 from rotkehlchen.chain.ethereum.modules.constants import AMM_ASSETS_SYMBOLS
+from rotkehlchen.chain.ethereum.modules.uniswap.constants import CPT_UNISWAP_V2
 from rotkehlchen.chain.ethereum.utils import asset_normalized_value, generate_address_via_create2
 from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
 from rotkehlchen.chain.evm.decoding.constants import ERC20_OR_ERC721_TRANSFER
 from rotkehlchen.chain.evm.decoding.structures import (
     DEFAULT_DECODING_OUTPUT,
-    FAILED_ENRICHMENT_OUTPUT,
     ActionItem,
     DecodingOutput,
-    EnricherContext,
-    TransferEnrichmentOutput,
 )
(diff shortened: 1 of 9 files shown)


rotkehlchen/chain/ethereum/modules/uniswap/v2/common.py
```diff
@@ -352,7 +367,7 @@ def enrich_uniswap_v2_like_lp_tokens_transfers(
     """This function enriches LP tokens transfers of Uniswap V2 like AMMs."""
     resolved_asset = context.event.asset.resolve_to_crypto_asset()
     if (
-        resolved_asset.symbol == lp_token_symbol and
+        resolved_asset.symbol.startswith(lp_token_symbol) and
```
LefterisJP, line 370: Why startswith here? Due to the `object.__setattr__(pool_token, 'symbol', f'{pool_token.symbol} {token0.symbol}-{token1.symbol}')` that the way you have it now happens only at 2nd run?

yabirgb: yes

rotkehlchen/chain/ethereum/modules/uniswap/v2/common.py
```diff
@@ -295,88 +276,106 @@ def decode_uniswap_like_deposit_and_withdrawals(
...
         if len(pool_token.underlying_tokens) == 0:
             pool_token = dataclasses.replace(pool_token, underlying_tokens=underlying_tokens)
             GlobalDBHandler().edit_evm_token(pool_token)
+            AssetResolver().clean_memory_cache(pool_token.identifier)
+
+        if pool_token.symbol in {UNISWAP_PROTOCOL, SUSHISWAP_PROTOCOL}:
+            # uniswap and sushiswap provide the same symbol for all the LP tokens. In order to
+            # provide a a better UX if the default symbol is used then change the symbol to
+            # include the symbols of the underlying tokens.
+            object.__setattr__(pool_token, 'symbol', f'{pool_token.symbol} {token0.symbol}-{token1.symbol}')  # noqa: E501
```
LefterisJP, line 291: perhaps it would make sense, not for this PR, but to abstract editing a token somehow out of here as it seems to involve multiple things such as editing an immutable object (LOL), editing the DB and then cleaning the cache.

Don't we do it anywhere else too?

LefterisJP (approved):
LGTM

PR #7034, opened 2023-12-01
Fix loading for invalid event type/subtype combinations

LefterisJP:
Before this patch any invalid type/subtype combination would raise an error in the history view.

And we can have invalid rules that were just created from versions before 1.31.0.

### DIFF
--- a/rotkehlchen/tests/unit/test_history_events.py
+++ b/rotkehlchen/tests/unit/test_history_events.py
@@ -1,0 +1,51 @@
+from rotkehlchen.accounting.constants import EVENT_CATEGORY_MAPPINGS
+from rotkehlchen.accounting.structures.balance import Balance
+from rotkehlchen.accounting.structures.base import (
+    HistoryEvent,
+    HistoryEventSubType,
+    HistoryEventType,
+)
+from rotkehlchen.constants.assets import A_ETH
+from rotkehlchen.fval import FVal
+from rotkehlchen.types import Location, TimestampMS
+
+
+def test_serialize_with_invalid_type_subtype():
+    """Test that serialize an event with invalid type/subtype does not raise exception"""
+    event_type = HistoryEventType.TRANSFER
+    event_subtype = HistoryEventSubType.SPEND
+    assert event_subtype not in EVENT_CATEGORY_MAPPINGS[event_type]
+    event = HistoryEvent(
+        event_identifier='1',
+        sequence_index=1,
+        timestamp=TimestampMS(1),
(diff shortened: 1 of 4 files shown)


rotkehlchen/tests/unit/test_history_events.py
```diff
@@ -0,0 +1,51 @@
...
+        event_identifier='1',
+        sequence_index=1,
+        timestamp=TimestampMS(1),
+        location=Location.KRAKEN,
+        event_type=event_type,
+        event_subtype=HistoryEventSubType.NONE,
+        asset=A_ETH,
+        balance=Balance(amount=FVal(1)),
+    )
+    event.event_subtype = event_subtype  # do here cause ctor will raise for invalid subtype
```
yabirgb, line 28: I believe this is a typo? what is ctor? I believe you deletec constructor

rotkehlchen/tests/unit/test_history_events.py
```diff
@@ -0,0 +1,51 @@
...
+    assert event.serialize_for_api(
+        customized_event_ids=[],
+        ignored_ids_mapping={},
+        hidden_event_ids=[],
+        missing_accounting_rule=True,  # needed to recreate the error thistestsfor
```
yabirgb, line 33: `thistestsfor` doesn't belong here either

rotkehlchen/tests/unit/test_history_events.py
```diff
@@ -0,0 +1,51 @@
...
+    assert event_subtype not in EVENT_CATEGORY_MAPPINGS[event_type]
+    event = HistoryEvent(
+        event_identifier='1',
+        sequence_index=1,
+        timestamp=TimestampMS(1),
+        location=Location.KRAKEN,
+        event_type=event_type,
+        event_subtype=HistoryEventSubType.NONE,
+        asset=A_ETH,
+        balance=Balance(amount=FVal(1)),
```
yabirgb, line 26:
```suggestion
        balance=Balance(amount=ONE),
```

yabirgb (commented):
Just a few small things. Looks okey

rotkehlchen/tests/unit/test_history_events.py
LefterisJP: ctor is a common abbreviation for constructor

rotkehlchen/tests/unit/test_history_events.py
LefterisJP: it just needs spaces

PR #7078, opened 2023-12-10
Rewrite coinbase module to only use transactions API

### LINKED ISSUE #7075: Coinbase may not work with newer api keys
## Problem Definition

Due to incompatibility with newer api keys some users may see an error like the following:

> The setup of coinbase failed due to: Coinbase query https://api.coinbase.com/v2/accounts/xxxxxx/buys responded with error status code: 410 and text: This API has been deprecated. Please use the Transactions API (https://docs.cloud.coinbase.com/sign-in-with-coinbase/docs/api-transactions#list-transactions) to view your past transactions.

~~That is because apparently newer api version deprecated some endpoints we use. Coinbase endpoints depend on when the api key was created and what api version it uses~~ :rofl:

Nope. They really just went and removed an entire endpoint without versioning. 

## Task

~~Thankfully this is easy to handle. They provide a header to specify the coinbase api version and we can provide a version we know the current rotki implementation can work fine with.~~

https://docs.cloud.coinbase.com/sign-in-with-coinbase/docs/versioning

That's not it. We were already providing the version. They just decided to deprecate buy/sells endpoint wihout respecting the versioning schema. Setting up a new account even with my 2016 api key does not work.

We have to make a patch replacing the deprecated buys/sells endpoints. Also I did not see it in my normal app as I have coinbase history sync off.

LefterisJP:
Fix #7075 and stop using the removed buy/sell apis.

Main changes here are:

- Only query coinbase transactions endpoint. Unfortunately needs to be queried once per account. All other endpoints seem to have same/duplicate data
- Remember last time each coinbase account was queried and also last transaction id. This helps not querying same events again and improves speed
- Do not query for accounts with created_at == updated_at. We assume that this means they have had no activity.
- Added support for interest events, and for understanding withdrawals from coinbase pro

-----

This is WIP. All tests would need to be rewritten too. Tested it with my account/api key and works fine and shows all events.


rotkehlchen/exchanges/coinbase.py
```diff
@@ -278,68 +222,72 @@ def validate_api_key(self) -> tuple[bool, str]:
...
-            # and now try to get all sells of an account to see if that's possible
-            method = f'accounts/{account_ids[0]}/sells'
+            method = f'accounts/{account_info[0][0]}/transactions'
             result, msg = self._validate_single_api_key_action(method)
             if result is None:
                 return False, msg
 
             # and now try to get all deposits of an account to see if that's possible
-            method = f'accounts/{account_ids[0]}/deposits'
+            method = f'accounts/{account_info[0][0]}/deposits'
```
LefterisJP, line 241: can remove those too

rotkehlchen/exchanges/coinbase.py
```diff
@@ -278,68 +222,72 @@ def validate_api_key(self) -> tuple[bool, str]:
...
+            updated_at = account_data.get('updated_at')
+            try:
+                timestamp = deserialize_timestamp_from_date(updated_at, 'iso8601', 'coinbase')
+            except DeserializationError:
+                log.error(f'Skipping coinbase account {account_data} due to inability to deserialize timestamp')  # noqa: E501
+                continue
+
+            log.debug(f'Found coinbase account: {account_data}')
+
+            if account_data.get('created_at') == updated_at:
```
LefterisJP, line 285: This `if` saves **a lot** of time. Can't find the assumption documented in their API docs, but their API docs are terribly outdated and don't even have half the transaction types I used in coinbase and I am not a heavy user at all.

```suggestion
            if account_data.get('created_at') == updated_at:
```

rotkehlchen/exchanges/coinbase.py
```diff
@@ -384,6 +330,11 @@ def _api_query(
             except requests.exceptions.RequestException as e:
                 raise RemoteError(f'Coinbase API request failed due to {e!s}') from e
 
+            if response.status_code in (401, 429) and had_4xx is False:
+                had_4xx = True
+                gevent.sleep(.5)
+                continue  # do a single retry since they don't have info on retries
```
LefterisJP, line 336: This I put to try and combat these random 401 authentication error warnings. They also mention them in the forums but no idea why it happens. Seems to be on the coinbase side.

LefterisJP:
The 3 failed tests are all external reasons unrelated to the PR.



1. Bitmex sandbox is down.
```
FAILED rotkehlchen/tests/exchanges/test_bitmex.py::test_bitmex_api_withdrawals_deposit_and_query_after_subquery - rotkehlchen.errors.misc.RemoteError: Bitmex api request for https://testnet.bitmex.com/api/v1/user/walletHistory failed with HTTP status code 400
FAILED rotkehlchen/tests/exchanges/test_bitmex.py::test_bitmex_margin_history - rotkehlchen.errors.misc.RemoteError: Bitmex api request for https://testnet.bitmex.com/api/v1/user/walletHistory failed with HTTP status code 400
```

2. One test failed due to ethersan remote error and passes locally
```
rotkehlchen/tests/unit/test_evm_tx_decoding.py::test_query_and_decode_transactions_works_with_different_chains
```

PR #7087, opened 2023-12-12
Fix: debounce settings save cb and fix dashboard assets search 

tewshi:
part of https://github.com/orgs/rotki/projects/8/views/1?pane=issue&itemId=47017656

## Checklist

### DIFF
--- a/frontend/app/src/components/settings/controls/SettingsOption.vue
+++ b/frontend/app/src/components/settings/controls/SettingsOption.vue
@@ -54,7 +54,7 @@ const getMessage = (
   return message(value);
 };
 
-const update = async (newValue: any) => {
+const update = useDebounceFn(async (newValue: any) => {
   stop();
   clear();
   set(loading, true);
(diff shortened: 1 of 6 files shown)


frontend/app/src/components/settings/controls/SettingsOption.vue
```diff
@@ -83,7 +83,7 @@ const update = async (newValue: any) => {
     setError(result.error);
   }
   emit('finished');
-};
+}, 1500);
```
lukicenturi, line 86: I'm not sure if we should place the debounce here. 
Additionally, what if the input is blurred? Wouldn't it be better to save without any delay?

tewshi: for this, we might reduce the timeout, but it won't be too noticeable on blur. also we don't use the blur event to save for all types

kelsos: Yes update is always attached to the `input` event, so debouncing the update makes more sense because you don't want a network request on every keystroke.

kelsos (approved):
lgtm @tewshi thanks.

This didn't have any type errors after rebasing so we can merge it as soon as the `CI` passes

PR #7090, opened 2023-12-12
refactor: unify external link components

### LINKED ISSUE #7088: Unify `<BaseExternalLink>`, `<ExternalLink>` and `<ExternalLinkButton>` into single component.
Unify `<BaseExternalLink>`, `<ExternalLink>` and `<ExternalLinkButton>` into single component, since all of them serve same purpose

lukicenturi:
Closes #7088

## Checklist


frontend/app/src/components/helper/ExternalLink.vue
```diff
@@ -1,12 +1,69 @@
-<script lang="ts" setup>
-const props = defineProps<{
-  url: string;
-}>();
+<script setup lang="ts">
+const props = withDefaults(
+  defineProps<{
+    url?: string | null;
```
kelsos, line 4:
```suggestion
    url?: string;
```

and make the default `undefined`. In the past we needed null because undefined was not triggering reactivity after being set, but this has not be a problem with composition api.

frontend/app/src/composables/links.ts
```diff
@@ -1,4 +1,4 @@
-export const useLinks = (url?: Ref<string>) => {
+export const useLinks = (url?: Ref<string | null>) => {
```
kelsos, line 1: this might need to become <string | undefined> after the change I suggested above.

kelsos:
Code looks ok, I will do a quick check in the morning but this also need a rebase on top of develop.

kelsos (approved):
lgtm

PR #7091, opened 2023-12-12
More resilient, less annoying coinbase balance queries

LefterisJP:
It seems that for at least one user the coinbase API will randomly return for some accounts the balance object having the 'value' key instead of the 'amount' key. Even though we specify the api version in every call.

Also this commit changes the annoying messages to simple log error/warnings.


LefterisJP:
bitmex test failure is unrelated. But we need to probably adjust those tests.

PR #7096, opened 2023-12-12
Bitstamp deposit/withdrawals show address + tx_id

### LINKED ISSUE #1863: Add support for Bitstamp crypto transactions
As per [Bitstamp API v2 documentation](https://www.bitstamp.net/api/), the JSON returned by the user transactions endpoint does not include any chain-related data (e.g. txid, address). This data is obtained by requesting the crypto transactions endpoint.

Be aware the crypto transactions endpoint is not paginated like the user transactions one (e.g. it has no `since_id`).
 
Related #1856 and more specifically: https://github.com/rotki/rotki/pull/1856#discussion_r536031605

LefterisJP:
Fix #1863

Achieve that by querying an extra endpoint for bitstamp deposit/withdrawals. The crypto_transactions.

Unfortunately we need to query both user_transactions and crypto_transactions for the full picture.

The first tells us the fees and the second tells us the address and tx_id.

After querying them, correlating them by matching them via timestamp, type, asset is also done.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`1863` Bitstamp deposit/withdrawals should now also show the address and transaction id.
 * :bug:`7075` Coinbase api should now be usable again. Additionally history retrieval speed has improved by orders of magnitude.
 * :bug:`-` Fix the issue where user profiles on the login screen are empty after logging out.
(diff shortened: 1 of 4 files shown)


LefterisJP:
Note to self: The bitstramp integration is old, quite badly written (a bit complicated) and could do with a rewrite but ... not a priority. Just had to whine a bit ...

PR #7100, opened 2023-12-13
Fix bitmex API queries

LefterisJP:
Changes enforced by them is:

- Made the currency parameter a requirement
- Removed the fee key in the deposit events for user wallet transactions

This PR reacts to them and adjusts tests.


LefterisJP:
all green. Such a beautiful sight.

PR #7108, opened 2023-12-14
Fix/#7099: add `exclude_ignored_assets` in trades and asset_movements query

### LINKED ISSUE #7099: Trades and Deposits/withdrawals should not return events with ignored assets
## Problem Definition

At the moment the history events API towards which we want to go is by default not returning any events that contain an ignored_asset. This is tweakable behaviour via the `exclude_ignored_assets` API argument but it's always set to true.

We also have events in the old format in the trades API endpoing and the asset movements API endpoint. Those endpoints do not follow the same logic. They will return all events despite containing an ignored asset.

The plan is to move trades and deposit/withdrawals into history events and essentially remove those api endpoints, but until that happens we need to have common functionality between them and not show to the user events with ignored assets.

## Task

Backend: Add same functionality we have in the history events endpoint to trades and deposit/withdrawals. Essentially a default true `exclude_ignored_assets` argument. Add it to the filter and make sure ignored asset events are skipped. Add tests.
Frontend: Make it tweakable.

OjusWiZard:
Closes #7099

## Checklist
- [x] add `exclude_ignored_assets` query parameter in trades and asset_movements query
- [x] add their tests
- [x] update the API docs

### DIFF
--- a/rotkehlchen/tests/api/test_exchanges.py
+++ b/rotkehlchen/tests/api/test_exchanges.py
@@ -33,6 +33,7 @@ from rotkehlchen.exchanges.kucoin import API_KEY_ERROR_CODE_ACTION as KUCOIN_API
 from rotkehlchen.fval import FVal
 from rotkehlchen.globaldb.binance import GlobalDBBinance
 from rotkehlchen.globaldb.handler import GlobalDBHandler
+from rotkehlchen.tests.utils.accounting import toggle_ignore_an_asset
 from rotkehlchen.tests.utils.api import (
     api_url_for,
     assert_error_response,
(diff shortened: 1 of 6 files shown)


rotkehlchen/tests/api/test_exchanges.py
```diff
@@ -801,6 +803,46 @@ def assert_okay(response):
...
+        without_ignore_response = assert_proper_response_with_result(without_ignore_response)
+
+        toggle_ignore_an_asset(rotkehlchen_api_server_with_exchanges, A_BTC)
+        data |= {'exclude_ignored_assets': True}
+        ignored_response = requests.get(api_url_for(server, 'assetmovementsresource'), json=data)
+        ignored_response = assert_proper_response_with_result(ignored_response)
+        default_response = requests.get(api_url_for(server, 'assetmovementsresource'))
+        default_response = assert_proper_response_with_result(default_response)
+        msg = '`exclude_ignored_assets` should be True by default'
+        assert ignored_response == default_response, msg
```
LefterisJP, line 822: nitipck. Keep things compact here and in other places in the code

```suggestion
        assert ignored_response == default_response, '`exclude_ignored_assets` should be True by default'  # noqa: E501
```

rotkehlchen/tests/utils/accounting.py
```diff
@@ -345,3 +350,47 @@ def assert_csv_export(
                 expected_csv_data=expected_csv_data,
                 expected_pnls=expected_pnls,
             )
+
+
+def toggle_ignore_an_asset(
+        rotkehlchen_api_server: APIServer,
+        asset_to_ignore: Asset,
+) -> None:
+    """Utility functions to add/remove an ignored asset"""
```
LefterisJP, line 359:
```suggestion
    """Utility function to add/remove an ignored asset"""
```

rotkehlchen/tests/utils/accounting.py
```diff
@@ -345,3 +350,47 @@ def assert_csv_export(
...
+                'ignoredassetsresource',
+            ), json={'assets': [asset_to_ignore.identifier]},
+        )
+        assert response.status_code == 200
+        response_result = response.json().get('result', [])
+        assert asset_to_ignore.identifier in response_result
+
+        with rotki.data.db.conn.read_ctx() as cursor:
+            result = rotki.data.db.get_ignored_asset_ids(cursor)
+        assert asset_to_ignore.identifier in result
```
LefterisJP, line 396: Can also do this for readability and compactness

```suggestion
    return
    
    # else add the asset to the ignored list
    response = requests.put(
        api_url_for(
            rotkehlchen_api_server,
            'ignoredassetsresource',
        ), json={'assets': [asset_to_ignore.identifier]},
    )
    assert response.status_code == 200
    response_result = response.json().get('result', [])
    assert asset_to_ignore.identifier in response_result

    with rotki.data.db.conn.read_ctx() as cursor:
        result = rotki.data.db.get_ignored_asset_ids(cursor)
    assert asset_to_ignore.identifier in result
```

rotkehlchen/tests/utils/history.py
```diff
@@ -736,6 +736,81 @@ def assert_kraken_asset_movements(
     assert_asset_movements(expected, to_check_list, deserialized, movements_to_check)
 
 
+def assert_ignored_btc_asset_movements(
```
LefterisJP, line 739: do not do it like this. If it's not used in at least 2 places do not create a new function.

rotkehlchen/tests/api/test_exchanges.py
OjusWiZard: Oh I was trying to avoid `# noqa: E501`, but okay.

LefterisJP: yeah I get it. Some times for compactness it's okay to have it.

LefterisJP (approved):
LGTM.

Only thing missing is a changelog entry as this has a subtle but meaningful change for users. Can add later if at all

PR #7112, opened 2023-12-15
Feat: history events section filter to show/hide entries with ignored assets

### LINKED ISSUE #7099: Trades and Deposits/withdrawals should not return events with ignored assets

lukicenturi:
Closes #7099

## Checklist


kelsos (approved):
lgtm

PR #7113, opened 2023-12-15
fix (#6979): remove case-sensitivity of generic import trade type

### LINKED ISSUE #6979: Feedback on generic impact functionality
Type	Location	Base Currency	Buy Amount	Sell Amount	Quote Currency	Timestamp	Fee	Fee Currency
Sell	binance	ETH	1	2	EUR	1		
Sell	binance	ETH	1	2	EUR	1		
Buy	binance	XMR	1	2.     ETH	1		
Buy	binance	XMR	1	2.     ETH	1		

A few things I just noticed

- if I define "Type" "sell" then rotki will import it as "Buy". If I define "Type" as "Sell" it properly imports it as "Sell"
- the docs say that Fee, Fee Currency and Description are optional, but the importer at least requires the headers. IMO it'd be nice if it could tolerate not getting those headers

OjusWiZard:
Closes #6979

## Checklist
- [x] checking the trade type after striping and lower-casing them
- [x] raising an error if an incorrect type is inputted
- [x] updated the test


rotkehlchen/data_import/importers/rotki_trades.py
```diff
@@ -39,6 +39,12 @@ def _consume_rotki_trades(
         amount_sold = deserialize_asset_amount(csv_row['Sell Amount'])
         amount_bought = deserialize_asset_amount(csv_row['Buy Amount'])
         asset, fee, fee_currency, location, timestamp = process_rotki_generic_import_csv_fields(csv_row, 'Base Currency')  # noqa: E501
+        if csv_row['Type'].strip().lower() == 'buy':
+            trade_type = TradeType.BUY
+        elif csv_row['Type'].strip().lower() == 'sell':
+            trade_type = TradeType.SELL
+        else:
+            raise InputError(f'Unknown trade type {csv_row["Type"]}')
```
yabirgb, line 47: The docstring is not updated to add this error but I believe it should be DeserializationError instead

rotkehlchen/data_import/importers/rotki_trades.py
```diff
@@ -39,6 +39,12 @@ def _consume_rotki_trades(
         amount_sold = deserialize_asset_amount(csv_row['Sell Amount'])
         amount_bought = deserialize_asset_amount(csv_row['Buy Amount'])
         asset, fee, fee_currency, location, timestamp = process_rotki_generic_import_csv_fields(csv_row, 'Base Currency')  # noqa: E501
+        if csv_row['Type'].strip().lower() == 'buy':
+            trade_type = TradeType.BUY
+        elif csv_row['Type'].strip().lower() == 'sell':
+            trade_type = TradeType.SELL
```
yabirgb, line 45: here if you need to got to the elif you are doing twice the `csv_row['Type'].strip().lower()`. I believe is better to create a variable (`sanitized_type`?) and store there `csv_row['Type'].strip().lower()` and then do the `if/else`

yabirgb: I just noticed that we have a `deserialize`method in `TradeType`that we weren't using. Instead of comparing against `buy/sell` I would use that

yabirgb (commented):
Just a few comments

LefterisJP (approved):
LGTM

Issue #7119, opened 2023-12-15
Remove bittrex and allow CSV import


LefterisJP:
## Problem Definition

Bittrex is finally completely dead. Today the API stopped responding. https://twitter.com/LefterisJP/status/1735800944225300522

![2023-12-16_00-09](https://github.com/rotki/rotki/assets/1658405/aae3ac66-17cf-4701-a517-34e1181152f7)

## Task

So let's remove the exchange completely.

### Task 1 - Remove

Remove bittrex and anything related to its API from the app. Keep the bittrex location.
### Task 2 - Allow importing bittrex data via CSV

![2023-12-16_00-18](https://github.com/rotki/rotki/assets/1658405/a075f3cf-fd32-4e2d-bd52-e21f3e5c361e)

There is 2 types of files that people can get from the website right now.

**transactions history.csv**

```csv
Date,Currency,Type,Address,Memo/Tag,TxId,Amount
2023-03-09 01:46:42.043,BTC,WITHDRAWAL,"ADDRESS","","TXID",-0.0001234
```

**order history per year.csv**

```csv
TXID,Time (UTC),Transaction,Order Type,Market,Base,Quote,Price,Quantity (Base),Fees (Quote),Total (Quote),Approx Value (USD),Time In Force,Notes
xx-yy-zz-dd,2017-09-01T12:03:05,Bought,Limit,SWT/BTC,SWT,BTC,0.00034100,731.28354007,0.00062343,0.24999841,,Good til cancelled,"Bought 431.18354007 SWT for 0.16999841 BTC"
```

-> Also in next develop upgrade go through the credentials of all removed exchanges and remove them since they are no longer possible to be used anyway and shouldn't stay in the DB. That is FTX, Bittrex and what's more? Note here: https://github.com/orgs/rotki/projects/11?pane=issue&itemId=47743438

LefterisJP:
Fixed by https://github.com/rotki/rotki/pull/7121

DeLo1585:
.reopen

> Invalid CSV format: The given bittrex CSV file format can not be recognized

today i recieved my history through the bittrex.Zendesk Support. history contains data from about 2018 to 2020

actual format that was given from the csv are (just each 2 lines from whole csv)

<closed_Deposit>.csv
 
```
Deposit Uuid,Currency,Amount,State,Confirmations,Last Updated (UTC),TxId,Crypto Address
[...]

PR #7121, opened 2023-12-16
Remove bittrex exchange

LefterisJP:
Remove bittrex exchange as an API-connectable exchange since it's now dead and API stopped working.

Part of https://github.com/rotki/rotki/issues/7119


LefterisJP:
It's all green (apart from a rate limit test), so merging.

PR #7122, opened 2023-12-18
Add augmented detection of spam tokens

### LINKED ISSUE #7066: Augment spam token detection algorithm
## Abstract

Currently we have a regex/pattern matching detection for spam tokens. This is helping get rid of many spam tokens but it's not perfect and many fall through the cracks.

The idea is to create a more aggressive methodology which can find and mark as spam almost all of the tokens by figuring out the methodology

## Specification

1. Get all owned tokens. Not all tokens in the global DB as this check is a bit heavier.
2. For each token check the events. If the only event touching this token is a receive event and nothing else go to (3)
3. Get the transaction receipt of the event and check the log topics. If the transaction receipt consists of multiple (more than `X` -- let's think of the X value ... perhaps 5?) erc20 token transfer topics for that asset it's a multi-send, which is what spammers do.
4. Query only defillama for evm chain + address price (not the entire oracle order as it takes a shit-ton of time). And if no price can be found, or if there is an error then consider this spam.

The above should run as a periodict task. It should probably be combined with the regex detection but be made sure to have the heavy part run only periodically and not on each token addition.

yabirgb:
closes #7066

Closes #(issue_number)

## Checklist


rotkehlchen/assets/asset.py
```diff
@@ -470,7 +470,7 @@ class EvmToken(CryptoAsset):
     chain_id: ChainID = field(init=False)
     token_kind: EvmTokenKind = field(init=False)
     decimals: int = field(init=False)
-    protocol: str = field(init=False)
+    protocol: str | None = field(init=False)
```
yabirgb, line 473: In the globaldb we use none and in the initialize field we had None as default value but the type didn't allow none. While writting the test this check failed and I have updated it

rotkehlchen/chain/evm/utils.py
```diff
@@ -51,6 +51,9 @@ def lp_price_from_uniswaplike_pool_contract(
     - Pooled amount of token 1
     - Total supply of pool token
     """
+    if token.protocol is None:
+        return None
+
```
yabirgb, line 56: this is the only significant change as consequence of protocol being Nullable

rotkehlchen/tasks/assets.py
```diff
@@ -17,6 +36,25 @@
...
+    """Update the protocol to SPAM_PROTOCOL for the list of assets provided and also ignores it"""
+    with globaldb.conn.write_ctx() as write_cursor:
+        query = 'UPDATE evm_tokens SET protocol=? WHERE identifier=?'
+        write_cursor.executemany(
+            query,
+            [(SPAM_PROTOCOL, identifier) for identifier in detected_spam_assets],
+        )
+
+    for asset_id in detected_spam_assets:
+        user_db.add_to_ignored_assets(write_cursor=user_db_write_ctx, asset=Asset(asset_id))
```
LefterisJP, line 57: we have had this discussion before. This is really bad and heavy code. Instead use `add_to_ignored_action_ids` which was made to avoid the resolution and lookup in the DB for every entry of a for loop

```suggestion
    user_db.add_to_ignored_action_ids(write_cursor=user_db_write_cursor, detected_spam_assets)
```

rotkehlchen/tasks/assets.py
```diff
@@ -17,6 +36,25 @@
...
     'A.identifier=B.identifier WHERE B.PROTOCOL IS NOT ? ORDER BY C.symbol'
 )
+MULTISEND_SPAM_THRESHOLD = 5
+
+
+def _add_spam_asset(
+        detected_spam_assets: Sequence[str],
+        globaldb: GlobalDBHandler,
+        user_db: DBHandler,
+        user_db_write_ctx: DBCursor,
```
LefterisJP, line 46:
```suggestion
        user_db_write_cursor: DBCursor,
```

rotkehlchen/tasks/assets.py
```diff
@@ -45,18 +83,14 @@ def autodetect_spam_assets_in_db(user_db: DBHandler) -> None:
...
-        for asset_id in detected_spam_assets:
-            user_db.add_to_ignored_assets(write_cursor=write_cursor, asset=Asset(asset_id))
+        _add_spam_asset(
+            detected_spam_assets=detected_spam_assets,
+            globaldb=globaldb,
+            user_db=user_db,
+            user_db_write_ctx=write_cursor,
+        )
 
         write_cursor.execute(  # remember last time spam detection ran
```
LefterisJP, line 95:
```suggestion
        )
        write_cursor.execute(  # remember last time spam detection ran
```

rotkehlchen/tasks/assets.py
```diff
@@ -74,3 +108,116 @@ def autodetect_spam_assets_in_db(user_db: DBHandler) -> None:
...
+
+            # check if all the events are receive and nothing else
+            if (
+                len(events) == 0 or
+                not all(event.event_type == HistoryEventType.RECEIVE and event.event_subtype == HistoryEventSubType.NONE for event in events)  # noqa: E501
+            ):
+                log.debug(
+                    f'Skipping {asset=} during spam detection because not '
+                    'all the events are receive',
+                )
```
LefterisJP, line 156: 1. Dude no no no. Don't log this. There is no value in this and it will spam the logs.
2. please keep things compact

```suggestion
            
            if (
                len(events) == 0 or
                not all(event.event_type == HistoryEventType.RECEIVE and event.event_subtype == HistoryEventSubType.NONE for event in events)  # noqa: E501
            ):  # check if all the events are receive and nothing else
```

rotkehlchen/tasks/assets.py
```diff
@@ -74,3 +108,116 @@ def autodetect_spam_assets_in_db(user_db: DBHandler) -> None:
...
+            asset = Asset(asset_id)
+            if (
+                not asset.is_evm_token() or
+                asset.identifier in spam_assets
+            ):
+                continue
+
+            events: list['EvmEvent'] = history_db.get_history_events(
+                cursor=cursor,
+                filter_query=EvmEventFilterQuery.make(assets=(asset,)),
```
LefterisJP, line 143: Be explicit here. Even if it is now the default, you have to be explicit ... and I think it's not the default and this will return the latest event first.

From what I understand you want to return the events, earliest first. Also please add it in a compact comment.

```suggestion
                filter_query=EvmEventFilterQuery.make(assets=(asset,)),
```

rotkehlchen/tasks/assets.py
```diff
@@ -74,3 +108,116 @@ def autodetect_spam_assets_in_db(user_db: DBHandler) -> None:
...
+
+            # check if the transaction could be a multisend
+            receipt = evm_tx_db.get_receipt(
+                cursor=cursor,
+                tx_hash=events[0].tx_hash,
+                chain_id=ChainID(events[0].location.to_chain_id()),
+            )
+            if receipt is None:
+                log.info(f'Could not find receipt for event {events[0].identifier=}')
+                continue
```
LefterisJP, line 167: 1. Write compact code
2. Not finding a receipt is an error.

```suggestion
            if (receipt := evm_tx_db.get_receipt(
                cursor=cursor,
                tx_hash=events[0].tx_hash,
                chain_id=ChainID(events[0].location.to_chain_id()),
            )) is None:  # check if the transaction could be a multisend
                log.error(f'Could not find evm receipt for event {events[0].identifier=}')
                continue
```

rotkehlchen/tasks/assets.py
```diff
@@ -74,3 +108,116 @@ def autodetect_spam_assets_in_db(user_db: DBHandler) -> None:
...
+
+            transfer_counter = 0
+            for log_event in receipt.logs:
+                if log_event.topics[0] == ERC20_OR_ERC721_TRANSFER:
+                    transfer_counter += 1
+
+                if transfer_counter == MULTISEND_SPAM_THRESHOLD:
```
LefterisJP, line 174:
```suggestion
                if transfer_counter >= MULTISEND_SPAM_THRESHOLD:
```

rotkehlchen/tasks/assets.py
```diff
@@ -74,3 +108,116 @@ def autodetect_spam_assets_in_db(user_db: DBHandler) -> None:
...
+                cursor=cursor,
+                tx_hash=events[0].tx_hash,
+                chain_id=ChainID(events[0].location.to_chain_id()),
+            )
+            if receipt is None:
+                log.info(f'Could not find receipt for event {events[0].identifier=}')
+                continue
+
+            transfer_counter = 0
+            for log_event in receipt.logs:
```
LefterisJP, line 170:
```suggestion
            for log_event in receipt.logs:  # Check if there is enough transfers to consider it spam
```

rotkehlchen/tasks/assets.py
```diff
@@ -74,3 +108,116 @@ def autodetect_spam_assets_in_db(user_db: DBHandler) -> None:
...
+                if log_event.topics[0] == ERC20_OR_ERC721_TRANSFER:
+                    transfer_counter += 1
+
+                if transfer_counter == MULTISEND_SPAM_THRESHOLD:
+                    break
+            else:
+                log.debug(f'Transaction {events[0].tx_hash=} was not detected as a spam transaction')  # noqa: E501
+                continue
```
LefterisJP, line 178: Dude this is too much logs spam. Please don't do this!

```suggestion
                continue  # not spam
```

rotkehlchen/tasks/assets.py
```diff
@@ -74,3 +108,116 @@ def autodetect_spam_assets_in_db(user_db: DBHandler) -> None:
...
+
+            # check if defillama has a price for the token as a last check
+            token = asset.resolve_to_evm_token()
+            try:
```
LefterisJP, line 182:
```suggestion
            token = asset.resolve_to_evm_token()
            try:  # check if defillama has a price for the token as a last check
```

rotkehlchen/tasks/assets.py
```diff
@@ -74,3 +108,116 @@ def autodetect_spam_assets_in_db(user_db: DBHandler) -> None:
...
+                price, _ = inquirer._defillama.query_current_price(
+                    from_asset=token,
+                    to_asset=usd_asset,
+                    match_main_currency=True,
+                )
+            except RemoteError:
+                log.debug(
```
LefterisJP, line 189:
```suggestion
                log.error(
```

rotkehlchen/tasks/assets.py
```diff
@@ -74,3 +108,116 @@ def autodetect_spam_assets_in_db(user_db: DBHandler) -> None:
...
+                    f'Failed to query defillama when doing spam detection on {asset=}. '
+                    'Skipping it for now',
+                )
+                continue
+
+            log.debug(f'Detected defillama price for {asset=} to be {price=} during spam check')
+            if price == ZERO_PRICE:
+                log.info(f'Detecting {asset} as spam asset')
+                spam_assets.add(asset.identifier)
+                chains_with_spam.add(token.chain_id)
```
LefterisJP, line 199:
```suggestion
            if price == ZERO_PRICE:
                log.info(f'Determined {asset} is a spam asset. Marking it as such.')
                spam_assets.add(asset.identifier)
                chains_with_spam.add(token.chain_id)
```

rotkehlchen/tasks/assets.py
```diff
@@ -74,3 +108,116 @@ def autodetect_spam_assets_in_db(user_db: DBHandler) -> None:
...
+            globaldb=globaldb,
+            user_db=user_db,
+            user_db_write_ctx=write_cursor,
+        )
+        write_cursor.execute(  # remember last time augmented spam detection ran
+            'INSERT OR REPLACE INTO settings (name, value) VALUES (?, ?)',
+            (LAST_AUGMENTED_SPAM_ASSETS_DETECT_KEY, str(ts_now())),
+        )
+
+    for chain in chains_with_spam:
```
LefterisJP, line 213:
```suggestion
    for chain in chains_with_spam:  # let frontend know so they refresh balances
```

rotkehlchen/tasks/manager.py
```diff
@@ -711,6 +716,22 @@ def _maybe_detect_new_spam_tokens(self) -> Optional[list[gevent.Greenlet]]:
             user_db=self.database,
         )]
 
+    def _maybe_detect_new_spam_tokens_augmented(self) -> Optional[list[gevent.Greenlet]]:
```
LefterisJP, line 719:
```suggestion
    def _maybe_augmented_detect_new_spam_tokens(self) -> Optional[list[gevent.Greenlet]]:
```

rotkehlchen/tasks/manager.py
```diff
@@ -711,6 +716,22 @@ def _maybe_detect_new_spam_tokens(self) -> Optional[list[gevent.Greenlet]]:
...
+        """
+        This function queries the globaldb looking for assets that look like spam tokens
+        and ignores them in addition to marking them as spam tokens
```
LefterisJP, line 722: This is copy pasta from the other task ...

```suggestion
        This function runs the augmented token detection algorithm which is a heavier and more
        time consuming analysis on user's asset that involves external calls in order to find and detect
        potential spam tokens.
```

rotkehlchen/tasks/assets.py
yabirgb: It doesn't happen for every event log. Is just after executing them. Anyway removed it

rotkehlchen/tasks/assets.py
yabirgb: I like the walrus operator but in this case for me is harder to read due the multiple lines and parenthesis. Applied the change

rotkehlchen/tasks/assets.py
yabirgb: As mentioned in discord Asset doesn't query the globaldb. I have created a new method that is a `executemany`

rotkehlchen/tasks/assets.py
LefterisJP: It's an unecessary spamming log entry

LefterisJP (approved):
LGTM. But tests still need to run and be checked.

PR #7128, opened 2023-12-19
fix: improve graph UX

lukicenturi:
https://gitlab.com/rotkehlchenio/premium-components/-/merge_requests/173

Fix what yabir said about invisible decimal detail in the chart


frontend/app/src/components/graphs/AssetAmountAndValueDialog.vue
```diff
@@ -0,0 +1,86 @@
+<script setup lang="ts">
```
kelsos, line 1: Do we use this component anywhere else in the app? If not this should not be the right place for it.

PR #7131, opened 2023-12-19
Bitstamp match withdrawals due to different activity/transaction timestamp

proteansec:


rotkehlchen/exchanges/bitstamp.py
```diff
@@ -299,7 +305,7 @@ def query_online_deposits_withdrawals(
         for idx in sorted(indices_to_delete, reverse=True):
             del crypto_asset_movements[idx]  # remove the crypto asset movements whose data we matched in the DB  # noqa: E501
 
-        return asset_movements + crypto_asset_movements
+        return asset_movements
```
LefterisJP, line 308: I think this log from when debugging would be useful to have here for all users. It would show us if there is another failure of matching and for which crypto transactions that was

```suggestion
        log.debug(f'Remaining Bitstramp unmatched {crypto_asset_movements=}')
        return asset_movements
```

rotkehlchen/exchanges/bitstamp.py
```diff
@@ -269,7 +275,7 @@ def query_online_deposits_withdrawals(
         indices_to_delete = []
         for asset_movement in asset_movements:
             for idx, crypto_movement in enumerate(crypto_asset_movements):
-                if crypto_movement.timestamp == asset_movement.timestamp and crypto_movement.category == asset_movement.category and crypto_movement.asset == asset_movement.asset:  # noqa: E501
+                if crypto_movement.category == asset_movement.category and crypto_movement.asset == asset_movement.asset and crypto_movement.amount == asset_movement.amount + asset_movement.fee and abs(crypto_movement.timestamp - asset_movement.timestamp) < BITSTAMP_MATCHING_TOLERANCE:
```
LefterisJP, line 278: you will need the noqa: E501 due to the line legth. Also Do `<=`

```suggestion
                if crypto_movement.category == asset_movement.category and crypto_movement.asset == asset_movement.asset and crypto_movement.amount == asset_movement.amount + asset_movement.fee and abs(crypto_movement.timestamp - asset_movement.timestamp) <= BITSTAMP_MATCHING_TOLERANCE:  # noqa: E501
```

rotkehlchen/exchanges/bitstamp.py
proteansec: Fixed

rotkehlchen/exchanges/bitstamp.py
proteansec: Fixed

rotkehlchen/exchanges/bitstamp.py
```diff
@@ -299,7 +305,8 @@ def query_online_deposits_withdrawals(
         for idx in sorted(indices_to_delete, reverse=True):
             del crypto_asset_movements[idx]  # remove the crypto asset movements whose data we matched in the DB  # noqa: E501
 
-        return asset_movements + crypto_asset_movements
+        log.debug(f'Remaining Bitstramp unmatched {crypto_asset_movements=}')
```
LefterisJP, line 308:
```suggestion
        log.debug(f'Remaining Bitstamp unmatched {crypto_asset_movements=}')
```

LefterisJP (commented):
Ah sorry I made a typo in my suggestion. can you fix? Also if you can please squash all commits in one, don't make multiple commits for this. If not, don't worry we can just squash at merging.

LefterisJP (approved):
LGTM. Both failing tests are already addressed in latest bugfixes.

PR #7125, opened 2023-12-18
fix (#7082): add `location_name` in trade query

OjusWiZard:
Closes #7082 

## Checklist

- [x] Update `query_location_latest_trades` to use `excluded_instances` to filter along with names
- [x] Update `_query_services_for_trades` to pass `excluded_instances`
- [x] Add unit test to check this functionality
- [x] Also, add missing documentation for `non_syncing_exchanges` in PUT /settings

PR #7136, opened 2023-12-19
Fix/#7123: Check for balance of EOA address on ethereum for checking its activity

### LINKED ISSUE #7123: Adding an EOA address that has no activity apart from withdrawals/blocks fails.
## Problem Definition

If you try to add an EOA address that has no other activity apart from eth2 withdrawals or non-mev block productions in rotki it fails. The reason is that no transactions/erc20 transfers etc. happened and we don't query for withdrawals or blocks and as such for rotki it seems as if the address has had no activity.

An example address is: https://etherscan.io/address/0x3c69bc9b9681683890ad82953fe67d13cd91d5ee

## Task

When adding/detecting activity for an address, if it's mainnet also check if it has any ETH balance.

OjusWiZard:
Closes #7123

## Checklist

- [x] Add check for balance in `has_activity` of Etherscan
- [x] Add test for `has_activity`
- [x] Cleanup from #7125

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`7123` Adding an EVM EOA address that has only withdrawals/blocks activity will no longer fail.
 * :bug:`7082` Now disabling sync for an exchange instance won't prevent other instances in the same exchange from querying new trades.
 * :bug:`7071` Fix the issue where users on mobile devices need to scroll to login.
 * :bug:`7120` Fix the issue where after removing an exchange key, an error notification is shown.
--- a/rotkehlchen/externalapis/etherscan.py
+++ b/rotkehlchen/externalapis/etherscan.py
@@ -55,10 +55,12 @@ class EtherscanHasChainActivity(Enum):
     """
     Classify the type of transaction first found in etherscan. TRANSACTIONS means that the endpoint
     for transactions/internal transactions had entries, TOKENS means that the tokens endpoint had
-    entries and NONE means that no entry has been found in the different endpoints.
+    entries, BALANCE means that the address has a non-zero native asset balance and
+    NONE means that no activity was found.
     """
     TRANSACTIONS = auto()
     TOKENS = auto()
+    BALANCE = auto()
     NONE = auto()
 
 
@@ -532,8 +534,8 @@ class Etherscan(ExternalServiceWithApiKey, metaclass=ABCMeta):
         yield _hashes_tuple_to_list(hashes)
 
     def has_activity(self, account: ChecksumEvmAddress) -> EtherscanHasChainActivity:
-        """Queries transactions, internal_txs and tokentx for an address with limit=1
-        just to quickly determine if the account has had any activity in the chain.
+        """Queries native asset balance, transactions, internal_txs and tokentx for an address
+        with limit=1 just to quickly determine if the account has had any activity in the chain.
         We make a distinction between transactions and ERC20 transfers since ERC20
         are often spammed. If there was no activity at all we return the enum value
         NONE.
@@ -548,6 +550,10 @@ class Etherscan(ExternalServiceWithApiKey, metaclass=ABCMeta):
         result = self._query(module='account', action='tokentx', options=options)
         if len(result) != 0:
             return EtherscanHasChainActivity.TOKENS
+        if self.chain in {SupportedBlockchain.ETHEREUM, SupportedBlockchain.GNOSIS}:
+            balance = self._query(module='account', action='balance', options={'address': account})
+            if int(balance) != 0:
+                return EtherscanHasChainActivity.BALANCE
         return EtherscanHasChainActivity.NONE
 
     def get_latest_block_number(self) -> int:
--- a/rotkehlchen/tests/external_apis/test_etherscan.py
+++ b/rotkehlchen/tests/external_apis/test_etherscan.py
@@ -12,6 +12,7 @@ from rotkehlchen.chain.evm.constants import GENESIS_HASH, ZERO_ADDRESS
 from rotkehlchen.db.dbhandler import DBHandler
 from rotkehlchen.db.evmtx import DBEvmTx
 from rotkehlchen.db.filtering import EvmTransactionsFilterQuery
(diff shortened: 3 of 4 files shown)


rotkehlchen/tests/external_apis/test_etherscan.py
```diff
@@ -192,3 +193,14 @@ def test_etherscan_get_contract_abi(temp_etherscan):
     abi = temp_etherscan.get_contract_abi('0x5a464C28D19848f44199D003BeF5ecc87d090F87')
     assert abi == json.loads(ILK_REGISTRY_ABI)
     assert temp_etherscan.get_contract_abi('0x9531C059098e3d194fF87FebB587aB07B30B1306') is None
+
+
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+def test_has_activity(temp_etherscan):
+    """Test to check if an address has any activity on ethereum mainnet"""
+
```
yabirgb, line 201: same comment as in the other PR don't leave empty lines

```suggestion
```

rotkehlchen/externalapis/etherscan.py
```diff
@@ -548,6 +548,9 @@ def has_activity(self, account: ChecksumEvmAddress) -> EtherscanHasChainActivity
         result = self._query(module='account', action='tokentx', options=options)
         if len(result) != 0:
             return EtherscanHasChainActivity.TOKENS
+        balance = self._query(module='account', action='balance', options={'address': str(account)})  # noqa: E501
+        if int(balance) != 0:
+            return EtherscanHasChainActivity.TRANSACTIONS
```
yabirgb, line 553: seems that we don't make a distinction and we only check for tokens but I would add a `BALANCE` entry to the enum and use it here

yabirgb: Also update the docstring of this function regarding what it does

yabirgb (commented):
This one also closes a bug right? It should have a changelog entry I believe

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`7123` Now adding an EOA address that has only withdrawals/blocks activity will not fail.
```
```suggestion
* :bug:`7123` Adding an EVM EOA address that has only withdrawals/blocks activity will no longer fail.
```

rotkehlchen/externalapis/etherscan.py
```diff
@@ -55,10 +55,12 @@ class EtherscanHasChainActivity(Enum):
     """
     Classify the type of transaction first found in etherscan. TRANSACTIONS means that the endpoint
     for transactions/internal transactions had entries, TOKENS means that the tokens endpoint had
-    entries and NONE means that no entry has been found in the different endpoints.
+    entries, BALANCE means that the address has a non-zero balance and
```
LefterisJP, line 58:
```suggestion
    entries, BALANCE means that the address has a non-zero native asset balance and
```

rotkehlchen/externalapis/etherscan.py
```diff
@@ -532,9 +534,9 @@ def get_token_transaction_hashes(
         yield _hashes_tuple_to_list(hashes)
 
     def has_activity(self, account: ChecksumEvmAddress) -> EtherscanHasChainActivity:
-        """Queries transactions, internal_txs and tokentx for an address with limit=1
+        """Queries balance, transactions, internal_txs and tokentx for an address with limit=1
```
LefterisJP, line 537:
```suggestion
        """Queries native asset balance, transactions, internal_txs and tokentx for an address with limit=1
```

rotkehlchen/externalapis/etherscan.py
```diff
@@ -532,9 +534,9 @@ def get_token_transaction_hashes(
...
         just to quickly determine if the account has had any activity in the chain.
-        We make a distinction between transactions and ERC20 transfers since ERC20
+        We make a distinction between balance, transactions and ERC20 transfers since ERC20
```
LefterisJP, line 539: why mention balance here?

```suggestion
        We make a distinction between transactions and ERC20 transfers since ERC20
```

rotkehlchen/externalapis/etherscan.py
```diff
@@ -548,6 +550,9 @@ def has_activity(self, account: ChecksumEvmAddress) -> EtherscanHasChainActivity
         result = self._query(module='account', action='tokentx', options=options)
         if len(result) != 0:
             return EtherscanHasChainActivity.TOKENS
+        balance = self._query(module='account', action='balance', options={'address': str(account)})  # noqa: E501
```
LefterisJP, line 553: Hmm @yabirgb this checks for all chains and not only ethereum mainnet. Does this make sense? Should we do it? Say in Gnosis ... does it take native asset (ETH?) from withdrawals/blocks? In polygon what is it? In the rollups there is no such concept. In other chains?

yabirgb: ah true I forgot the other chains. If this is possible in ethereum it should be possible in gnosis as it the same process. In polygon I'm not aware how the staking works and in the rollups the concept doesn't exists so I would limit this query only to ethereum and gnosis

rotkehlchen/externalapis/etherscan.py
```diff
@@ -532,8 +534,8 @@ def get_token_transaction_hashes(
         yield _hashes_tuple_to_list(hashes)
 
     def has_activity(self, account: ChecksumEvmAddress) -> EtherscanHasChainActivity:
-        """Queries transactions, internal_txs and tokentx for an address with limit=1
-        just to quickly determine if the account has had any activity in the chain.
+        """Queries native asset balance, transactions, internal_txs and tokentx for an address
+        with limit=1 just to quickly determine if the account has had any activity in the chain.
```
LefterisJP, line 538: Not sure if this has a trailing whitespace as I am checking from mobile. If it does fix it in another PR

LefterisJP (approved):
LGTM with a small question/ nitpick

rotkehlchen/externalapis/etherscan.py
OjusWiZard: I don't see any, ` make format` did the job I guess..

PR #7149, opened 2023-12-21
Process scientific notation timestamps

LefterisJP:
If a user imports a timestamp in scientific notation from a CSV or anywhere else it should now work properly.

Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=47247446


yabirgb (approved):
As a mathematician I can approve now

PR #7151, opened 2023-12-21
Remove current price comparison for defillama.

yabirgb:
This test is only comparing if the prices are the same but we have tested already defillama and this test is expensive against the rate limit. Removing it


LefterisJP (approved):
LGTM

PR #7152, opened 2023-12-21
chore: rename managed assets to assets across

tewshi:
## Checklist

### DIFF
--- a/frontend/app/src/locales/en.json
+++ b/frontend/app/src/locales/en.json
@@ -3202,7 +3202,7 @@
     "manage_assets": "Manage Assets",
     "manage_assets_sub": {
       "custom_assets": "Custom Assets",
-      "managed_assets": "Assets",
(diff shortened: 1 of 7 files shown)


frontend/app/src/locales/en.json
```diff
@@ -3202,7 +3202,7 @@
     "manage_assets": "Manage Assets",
     "manage_assets_sub": {
       "custom_assets": "Custom Assets",
-      "managed_assets": "Managed Assets",
+      "managed_assets": "Assets",
```
LefterisJP, line 3205: Shouldn't you also change the json key? Otherwise developers will wonder why the name difference in the key and the value in the future.

tewshi: on it now.

LefterisJP (approved):
LGTM. Thank you

PR #7153, opened 2023-12-21
Periodically update owned assets

yabirgb:
This PR ensures that history events are queried when checking for owned assets Notify
and also adds a periodic task to update the list of owned assets.


rotkehlchen/constants/timing.py
```diff
@@ -21,3 +21,4 @@
 ENS_AVATARS_REFRESH = DAY_IN_SECONDS
 SPAM_ASSETS_DETECTION_REFRESH = HOUR_IN_SECONDS * 6
 AUGMENTED_SPAM_ASSETS_DETECTION_REFRESH = DAY_IN_SECONDS
+OWNED_ASSETS_UPDATE = HOUR_IN_SECONDS * 2
```
LefterisJP, line 24: Can also be every hour. Is fine. It takes very little time, right? Just one sqlite query

rotkehlchen/db/dbhandler.py
```diff
@@ -145,7 +145,7 @@
 
 
 # Tuples that contain first the name of a table and then the columns that
-# reference assets ids. This is used to query all assets that a user owns.
+# reference assets ids. This is used to query all assets that a user has ever own.
```
LefterisJP, line 148:
```suggestion
# reference assets ids. This is used to query all assets that a user has ever owned.
```

rotkehlchen/tests/api/test_assets.py
```diff
@@ -76,13 +81,29 @@ def test_query_owned_assets(
             id=-1,
             asset=A_EUR,
             label='My EUR bank',
-            amount=FVal('1550'),
+            amount=FVal(1550),
```
LefterisJP, line 84: it's fine to do this change, but why did you do it here and now?

LefterisJP (commented):
Also as I wrote in discord add a changelog entry. Something like: "When filtering for owned assets in the assets view you should be able to see the most recent assets even if you do not restart the app"

rotkehlchen/constants/timing.py
yabirgb: yes shouldn't take much. Is more: one query per table + the insert but is fast

rotkehlchen/tests/api/test_assets.py
yabirgb: just because I was editing the test and saw it

LefterisJP (approved):
lgtm, merging

PR #7163, opened 2023-12-22
Prevent errors when removing exchanges

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/db/dbhandler.py
+++ b/rotkehlchen/db/dbhandler.py
@@ -314,6 +314,7 @@ class DBHandler:
         is older than the one supported.
         - DBSchemaError if database schema is malformed.
         """
+        # This logic executes only for the user db
         # Run upgrades if needed
         fresh_db = DBUpgradeManager(self).run_upgrades()
         if fresh_db:  # create tables during the first run and add the DB version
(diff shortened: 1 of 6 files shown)


rotkehlchen/exchanges/manager.py
```diff
@@ -283,6 +283,9 @@ def initialize_exchanges(
         self.database = database
         # initialize exchanges for which we have keys and are not already initialized
         for location, credentials_list in exchange_credentials.items():
+            if location not in SUPPORTED_EXCHANGES:
```
LefterisJP, line 286: add a comment here explaining what this does
```suggestion
            if location not in SUPPORTED_EXCHANGES:  # in case a no longer supported exchange key is in the DB
```

LefterisJP (commented):
I think we also need 1 more thing. And that is a way to remove non-supported exchanges from the non-syncing exchanges dict when returnign it in case there is somehow a leftover in there.

rotkehlchen/db/settings.py
```diff
@@ -330,7 +331,13 @@ def db_settings_from_dict(
             specified_args[key] = [HistoricalPriceOracle.deserialize(oracle) for oracle in oracles]
         elif key == 'non_syncing_exchanges':
             values = json.loads(value)
-            specified_args[key] = [ExchangeLocationID.deserialize(x) for x in values]
+            non_syncing_exchanges = []
+            for exchange_info in values:  # exclude possible leftovers of removed exchanges
+                location_id = ExchangeLocationID.deserialize(exchange_info)
+                if location_id.location in SUPPORTED_EXCHANGES:
```
LefterisJP, line 337: Sorry to be annoying here, but I also meant to delete it from the DB. Since this will stay here forever and always hit this special case

rotkehlchen/db/dbhandler.py
```diff
@@ -291,6 +319,7 @@ def _run_actions_after_first_connection(self) -> None:
         is older than the one supported.
         - DBSchemaError if database schema is malformed.
         """
+        # This logic executes only for the user db
         # Run upgrades if needed
```
LefterisJP, line 323:
```suggestion
        # Run upgrades if needed -- only for user DB
```

rotkehlchen/db/dbhandler.py
```diff
@@ -300,30 +329,33 @@ def _run_actions_after_first_connection(self) -> None:
...
-                ('version', str(ROTKEHLCHEN_TRANSIENT_DB_VERSION)),
-            )
-            self.conn_transient.commit()
 
+        # run checks on the database
         self.conn.schema_sanity_check()
+        self._check_settings()
+
+        # This logic executes only for the transient db
+        # set up transient connection.
```
LefterisJP, line 338:
```suggestion
        # This logic executes only for the transient db
```

LefterisJP (commented):
some nitpics

LefterisJP (approved):
LGTM

PR #7168, opened 2023-12-23
Fix negative amount appearing in some coinbase trades

LefterisJP:


LefterisJP:
user tested and confirmed this works

PR #7170, opened 2023-12-25
Optimizations on inquirer and assets

yabirgb:
- We were using instances of the Inquirer class to call staticmethods. This uses one extra method call and a check to see if there is already an instance for the singleton that are not needed.
- If we had evmtokens already resolved and called again resolve on them we would hit again the database. This happens some times in the decoders and is not really needed

### DIFF
--- a/rotkehlchen/inquirer.py
+++ b/rotkehlchen/inquirer.py
@@ -200,7 +200,7 @@ def get_underlying_asset_price(token: EvmToken) -> tuple[Price | None, CurrentPr
         usd_price = ZERO
         for underlying_token in custom_token.underlying_tokens:
             token = EvmToken(underlying_token.get_identifier(parent_chain=custom_token.chain_id))
-            underlying_asset_price, oracle, _ = Inquirer().find_usd_price_and_oracle(token)
+            underlying_asset_price, oracle, _ = Inquirer.find_usd_price_and_oracle(token)
             usd_price += underlying_asset_price * underlying_token.weight
 
         if usd_price != ZERO_PRICE:
@@ -357,7 +357,7 @@ class Inquirer:
             cache_key: tuple[Asset, Asset],
             match_main_currency: bool,
     ) -> CachedPriceEntry | None:
-        cache = Inquirer()._cached_current_price.get(cache_key, None)
+        cache = Inquirer._cached_current_price.get(cache_key, None)
         if cache is None or ts_now() - cache.time > CURRENT_PRICE_CACHE_SECS or cache.used_main_currency != match_main_currency:  # noqa: E501
             return None
 
@@ -371,9 +371,9 @@ class Inquirer:
             assets_to_invalidate.add(asset_a)
             assets_to_invalidate.add(asset_b)
 
-        for asset_pair in list(Inquirer()._cached_current_price):
+        for asset_pair in list(Inquirer._cached_current_price):
             if asset_pair[0] in assets_to_invalidate or asset_pair[1] in assets_to_invalidate:
-                Inquirer()._cached_current_price.pop(asset_pair, None)
+                Inquirer._cached_current_price.pop(asset_pair, None)
 
     @staticmethod
     def remove_cached_current_price_entry(cache_key: tuple[Asset, Asset]) -> None:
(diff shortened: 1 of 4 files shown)


rotkehlchen/tests/unit/globaldb/test_globaldb.py
```diff
@@ -1285,3 +1286,28 @@ def test_get_evm_tokens(globaldb):
     tokens = globaldb.get_evm_tokens(chain_id=ChainID.ETHEREUM, protocol=CPT_COMPOUND, exceptions=(exception_address,))  # noqa: E501
     assert len(tokens) == tokens_without_exception - 1
     assert not any(token.evm_address == exception_address for token in tokens)
+
+
+def test_asset_resolution(globaldb: GlobalDBHandler):  # pylint: disable=unused-argument
+    """Test that resolution of assets doesn't hit the database if the asset
+    has been already resolved
```
LefterisJP, line 1293:
```suggestion
    has already been resolved
```

LefterisJP (commented):
Looks okay. Just a nitpick typo and a question. Shouldn't this go to bugfixes? As it's really small, is an improvement and would not create any problems?

LefterisJP (commented):
There is some test failure heres

rotkehlchen/tests/unit/globaldb/test_globaldb.py
```diff
@@ -3,6 +3,7 @@
 from pathlib import Path
 from shutil import copyfile
 from typing import TYPE_CHECKING
+from unittest.mock import patch
```
LefterisJP, line 6: why is this needed here now?

LefterisJP (commented):
lint failure

rotkehlchen/tests/unit/globaldb/test_globaldb.py
yabirgb: Is me working in 3 things and having the wrong import commited. Has had to be after running the linter.sry

rotkehlchen/tests/unit/test_defi_oracles.py
```diff
@@ -71,42 +71,43 @@ def test_uniswap_oracles_special_cases(inquirer_defi, socket_enabled):  # pylint
...
+        return original_resolver_to_class(identifier, **kwargs)
+
+    with (
+        patch.object(asset_resolver, 'resolve_asset', wraps=mocked_asset_getter),
+        patch.object(asset_resolver, 'resolve_asset_to_class', wraps=mocked_asset_getter),
+    ):
         weth = EvmToken(A_WETH.identifier)
         assert weth.decimals is None
-        with pytest.raises(DefiPoolError):
+        with pytest.raises(DefiPoolError):  # type: ignore[unreachable]
```
LefterisJP, line 110: why is this type ignore needed?

rotkehlchen/inquirer.py
```diff
@@ -654,8 +652,8 @@ def _find_usd_price(
             underlying_tokens = asset.underlying_tokens
 
             # Check if it is a special token
-            if asset.identifier in instance.special_tokens:
-                ethereum = instance.get_evm_manager(chain_id=ChainID.ETHEREUM)
+            if asset.identifier in Inquirer.special_tokens:
+                ethereum = Inquirer().get_evm_manager(chain_id=ChainID.ETHEREUM)
```
LefterisJP, line 656: why do you need the parentheses in this one?

LefterisJP (commented):
2 questions but all green

rotkehlchen/tests/unit/test_defi_oracles.py
yabirgb: explained in the comment

rotkehlchen/inquirer.py
yabirgb: Because is a method and not an staticmethod. I've changed it

rotkehlchen/tests/unit/test_defi_oracles.py
```diff
@@ -71,42 +71,43 @@ def test_uniswap_oracles_special_cases(inquirer_defi, socket_enabled):  # pylint
...
+        return original_resolver_to_class(identifier, **kwargs)
+
+    with (
+        patch.object(asset_resolver, 'resolve_asset', wraps=mocked_asset_getter),
+        patch.object(asset_resolver, 'resolve_asset_to_class', wraps=mocked_asset_getter),
+    ):
         weth = EvmToken(A_WETH.identifier)
         assert weth.decimals is None
-        with pytest.raises(DefiPoolError):
+        with pytest.raises(DefiPoolError):  # type: ignore[unreachable]  # this is needed because decimals is not nullable in the dataclass but it is in the database and the previous line is always evaluated as False by mypy.
```
LefterisJP, line 110: should add a link to this in the note I asked you to create. Also a link to the note from here

LefterisJP (approved):
LGTM

PR #7171, opened 2023-12-25
Change API error 409 to 401 when user is not logged in

LefterisJP:
Backend part of https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=42744707

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -165,7 +165,7 @@ Handling user creation, sign-in, log-out and querying
    :resjson object result: For successful requests, result contains the currently connected exchanges, and the user's settings. For details on the user settings refer to the `Getting or modifying settings`_ section.
    :statuscode 200: Adding the new user was successful
    :statuscode 400: Provided JSON is in some way malformed
-   :statuscode 409: User already exists. Another user is already logged in. Given Premium API credentials are invalid. Permission error while trying to access the directory where rotki saves data.
+   :statuscode 409: Another user is already logged in. User already exists. Given Premium API credentials are invalid. Permission error while trying to access the directory where rotki saves data.
    :statuscode 500: Internal rotki error
 
 .. http:post:: /api/(version)/users/(username)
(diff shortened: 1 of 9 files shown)


rotkehlchen/api/rest.py
```diff
@@ -1042,7 +1042,7 @@ def create_new_user(
             return {
                 'result': None,
                 'message': message,
-                'status_code': HTTPStatus.CONFLICT,
+                'status_code': HTTPStatus.UNAUTHORIZED,
```
kelsos, line 1045: Shouldn't this stay a conflict contextually?

LefterisJP: I don't know. This is what I understood from what you asked. Just let me know

kelsos: 401 makes sense for other endpoints but I still feel that for account creation or login 409 remains the valid response.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409

> The HTTP 409 Conflict response status code indicates a request conflict with the current state of the target resource.

At least for me trying to login with a user while another user is still logged in feels like a conflict rather than being unauthorized.

LefterisJP: I have no opinions at all here. Before was fine for me, whatever you want changed is fine for me. So I will chanfge it back for accoutn creation and login

kelsos: So the way I have it in mind is that any "a user is not logged in" should be a 401 so that the frontend can have an interceptor that redirects you to the login screen. Does this clarify things?

LefterisJP: There is more 401s around the code for other things that match the unaothorized one. Not sure if it will always work. You can see if this is what you want, otherwise tell me which endpoint needs changing. Premium user only for example also get 401s.

The amount of manual changes in the docs make this a really annoying change to make. So please whatever you want to have specify very well beforehand, otherwise we have to pass over the docs again to make a lot of changes.

LefterisJP:
So after our call with Kelsos I tried to do the following:

- Return 401 Unauthorized whenever there is no user logged in and we need this. They want to use a reflector for this status code to redirect to login screen
- To not use 401 in missing premium, I used 403 Forbidden for api endpoints that need premium

Still need to check if we use 401 Unauthorized as a return value in any other case

yabirgb (commented):
looks okey but the tests also need to be updated

LefterisJP:
tests updated. Should be fine to build upon.

LefterisJP:
as discussed in discord merging, and frontend should do their part in a different PR

PR #7172, opened 2023-12-26
fix: add exchange detail in location mapping

lukicenturi:
Part of #7174

### DIFF
--- a/rotkehlchen/constants/location_details.py
+++ b/rotkehlchen/constants/location_details.py
@@ -65,10 +65,11 @@ LOCATION_DETAILS: dict = {
 }
 for key, value in LOCATION_DETAILS.items():
     if key in ALL_SUPPORTED_EXCHANGES:
-        value['is_exchange'] = True
         if key in SUPPORTED_EXCHANGES:
-            value['exchange_data'] = {'is_exchange_with_key': True}
+            value['exchange_details'] = {'is_exchange_with_key': True}
             if key in EXCHANGES_WITH_PASSPHRASE:
-                value['exchange_data']['is_exchange_with_passphrase'] = True
+                value['exchange_details']['is_exchange_with_passphrase'] = True
             if key in EXCHANGES_WITHOUT_API_SECRET:
-                value['exchange_data']['is_exchange_without_api_secret'] = True
+                value['exchange_details']['is_exchange_without_api_secret'] = True
+            continue
+        value['is_exchange'] = True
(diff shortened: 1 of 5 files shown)


rotkehlchen/constants/location_details.py
```diff
@@ -59,5 +64,12 @@
     Location.WOO: {'image': 'woo.svg'},
 }
 for key, value in LOCATION_DETAILS.items():
-    if key in SUPPORTED_EXCHANGES:
-        value['is_exchange_with_key'] = True
+    if key in ALL_SUPPORTED_EXCHANGES:
+        value['is_exchange'] = True
+        if key in SUPPORTED_EXCHANGES:
+            value['exchange_data'] = {}
+            value['exchange_data']['is_exchange_with_key'] = True
```
yabirgb, line 71:
```suggestion
            value['exchange_data'] = {'is_exchange_with_key': True}
```

rotkehlchen/constants/location_details.py
```diff
@@ -59,5 +64,12 @@
     Location.WOO: {'image': 'woo.svg'},
 }
 for key, value in LOCATION_DETAILS.items():
-    if key in SUPPORTED_EXCHANGES:
-        value['is_exchange_with_key'] = True
+    if key in ALL_SUPPORTED_EXCHANGES:
```
yabirgb, line 67: this will include the dead exchanges. Maybe you want to mark somehow if the exchange is dead?

yabirgb (commented):
two comments

rotkehlchen/constants/location_details.py
lukicenturi: I need both dead exchanges, and exchanges that can be added with key

the dead exchanges, I need it in purge data section

yabirgb: what I mean is that before you didn't have them and now you don't know that they are dead because there is no key telling you it unless I'm missing something

lukicenturi: I mark all exchanges as `is_exchange = True`,
and the ones that are still alive I mark is as `is_exchange_with_key  = True`.

It should be enough to separate them, no?

yabirgb: Ah okey I understood now. Yes should be fien

lukicenturi:
Okay thanks, I will add the test

LefterisJP (commented):
There is some redundancy in this api design. You have both `is_exchange` and `exchange_detail` in the same response. IT shuold have either or. If there is details to share it should have `exchange_details` (notice the plural), and if it's not then just the `is_exchange` being True

LefterisJP (approved):
LGTM

PR #7177, opened 2023-12-26
Avoid extra queries of prices for assets in collections

### LINKED ISSUE #7147: Inquirer doesn't use cached price for groupped assets
We have collections of assets that use the same oracle id. The problem is that the inquirer doesn't use this information and if it has already queried the USDC price for the ethereum token it queries it again for the USDC in optimism.

This has a negative impact due to rate limits and slower queries.

## Task

Make use of the cache when querying prices of assets that belong to a collection and any of the members has already been queried

yabirgb:
Closes #7147

### DIFF
--- a/rotkehlchen/globaldb/handler.py
+++ b/rotkehlchen/globaldb/handler.py
@@ -1936,3 +1936,23 @@ class GlobalDBHandler:
                     (abi_name, serialized_abi),
                 )
                 return cursor.lastrowid
+
+    @staticmethod
+    def get_assets_in_same_collection(identifier: str) -> tuple[Asset, ...]:
+        """
+        Query the assets that belong to the collection of the queried asset. If the
+        asset isn't in any collection we return a list with the asset queried.
+        """
+        with GlobalDBHandler().conn.read_ctx() as cursor:
+            cursor.execute(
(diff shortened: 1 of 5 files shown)


rotkehlchen/globaldb/handler.py
```diff
@@ -1936,3 +1936,24 @@ def get_or_write_abi(serialized_abi: str, abi_name: str | None = None) -> int:
...
+        asset isn't in any collection we return a list with the asset queried.
+        """
+        with GlobalDBHandler().conn.read_ctx() as cursor:
+            cursor.execute(
+                'SELECT MM.asset FROM multiasset_mappings AS MM JOIN asset_collections AS AC '
+                'ON MM.collection_id = AC.id WHERE MM.collection_id = (SELECT collection_id '
+                'FROM multiasset_mappings WHERE asset=?)',
+                (identifier,),
+            )
+
```
LefterisJP, line 1953:
```suggestion
```

LefterisJP (commented):
Small nitpick and please add changelog.

"rotki should no longer query price multple times for the same asset across different chains"

LefterisJP (approved):
The 2 test failures are unrelated

PR #7178, opened 2023-12-26
Feat/cache: Move unrelated values from `settings` to `key_value_cache`

OjusWiZard:
## Checklist

- [x] Update `v40_v41` upgrade to move unrelated settings to the `key_value_cache` table
- [x] Update `/settings` API to keep returning old settings values from both tables now
- [x] Add `DBCache` class for serializing and deserializing
- [x] Clean some temporary settings' code for cache values
- [x] Update `settings` with `key_value_cache` in related parts of the codebase and tests

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -111,6 +111,7 @@ from rotkehlchen.constants.timing import ENS_AVATARS_REFRESH
 from rotkehlchen.data_import.manager import DataImportSource
 from rotkehlchen.db.accounting_rules import DBAccountingRules, query_missing_accounting_rules
 from rotkehlchen.db.addressbook import DBAddressbook
+from rotkehlchen.db.cache import serialize_cache_for_api
 from rotkehlchen.db.constants import (
     HISTORY_MAPPING_KEY_STATE,
     HISTORY_MAPPING_STATE_CUSTOMIZED,
@@ -449,12 +450,15 @@ class RestAPI:
 
         with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
             new_settings = process_result(self.rotkehlchen.get_settings(cursor))
-        result_dict = {'result': new_settings, 'message': ''}
+            cache = process_result(self.rotkehlchen.get_cache(cursor))
+        result_dict = {'result': new_settings | serialize_cache_for_api(cache), 'message': ''}
         return api_response(result=result_dict, status_code=HTTPStatus.OK)
 
     def get_settings(self) -> Response:
         with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
-            result_dict = _wrap_in_ok_result(process_result(self.rotkehlchen.get_settings(cursor)))
+            settings = process_result(self.rotkehlchen.get_settings(cursor))
+            cache = process_result(self.rotkehlchen.get_cache(cursor))
+        result_dict = _wrap_in_ok_result(settings | serialize_cache_for_api(cache))
         return api_response(result=result_dict, status_code=HTTPStatus.OK)
 
     def query_tasks_outcome(self, task_id: int | None) -> Response:
--- a/rotkehlchen/data_migrations/manager.py
+++ b/rotkehlchen/data_migrations/manager.py
@@ -45,15 +45,14 @@ class DataMigrationManager:
 
     def maybe_migrate_data(self) -> None:
         with self.rotki.data.db.conn.read_ctx() as cursor:
-            settings = self.rotki.data.db.get_settings(cursor)
-        last_migration_version = settings.last_data_migration
+            last_migration_version = self.rotki.data.db.get_setting(cursor, 'last_data_migration')
 
         self.progress_handler = MigrationProgressHandler(
             messages_aggregator=self.rotki.msg_aggregator,
             target_version=LAST_DATA_MIGRATION,
         )
         for migration in MIGRATION_LIST:
-            if last_migration_version < migration.version:
+            if last_migration_version is not None and last_migration_version < migration.version:
                 if self._perform_migration(migration) is False:
                     break  # a migration failed -- no point continuing
 
--- a/rotkehlchen/db/cache.py
+++ b/rotkehlchen/db/cache.py
@@ -1,0 +1,30 @@
+from rotkehlchen.errors.serialization import DeserializationError
+from rotkehlchen.types import Timestamp
+from rotkehlchen.utils.mixins.enums import Enum
+
+
+class DBCache(Enum):
+    """It contains all the values that can be stored in the `key_value_cache` table of the DB"""
+    LAST_BALANCE_SAVE = 'last_balance_save'
+    LAST_DATA_UPLOAD_TS = 'last_data_upload_ts'
+    LAST_DATA_UPDATES_TS = 'last_data_updates_ts'
+    LAST_OWNED_ASSETS_UPDATE = 'last_owned_assets_update'
+    LAST_EVM_ACCOUNTS_DETECT_TS = 'last_evm_accounts_detect_ts'
+    LAST_SPAM_ASSETS_DETECT_KEY = 'last_spam_assets_detect_key'
+    LAST_AUGMENTED_SPAM_ASSETS_DETECT_KEY = 'last_augmented_spam_assets_detect_key'
+
+    @classmethod
+    def deserialize(cls: type['DBCache'], value: str) -> 'DBCache':
+        """The method to deserialize a string to an enum object"""
+        try:
+            return getattr(cls, value.upper())
+        except AttributeError as e:
+            raise DeserializationError(f'Failed to deserialize {cls.__name__} value {value}') from e  # noqa: E501
+
+
+def serialize_cache_for_api(cache: dict[DBCache, int]) -> dict[str, Timestamp]:
+    """Serialize the cache for /settings API consumption."""
+    return {
+        DBCache.LAST_DATA_UPLOAD_TS.value: Timestamp(cache.get(DBCache.LAST_DATA_UPLOAD_TS, 0)),
+        DBCache.LAST_BALANCE_SAVE.value: Timestamp(cache.get(DBCache.LAST_BALANCE_SAVE, 0)),
+    }
--- a/rotkehlchen/db/dbhandler.py
+++ b/rotkehlchen/db/dbhandler.py
@@ -44,6 +44,7 @@ from rotkehlchen.constants.limits import (
 )
 from rotkehlchen.constants.misc import NFT_DIRECTIVE, USERDB_NAME
 from rotkehlchen.constants.timing import HOUR_IN_SECONDS
+from rotkehlchen.db.cache import DBCache
 from rotkehlchen.db.constants import (
     BINANCE_MARKETS_KEY,
     EVM_ACCOUNTS_DETAILS_LAST_QUERIED_TS,
@@ -63,6 +64,7 @@ from rotkehlchen.db.misc import detect_sqlcipher_version
 from rotkehlchen.db.schema import DB_SCRIPT_CREATE_TABLES
 from rotkehlchen.db.schema_transient import DB_SCRIPT_CREATE_TRANSIENT_TABLES
 from rotkehlchen.db.settings import (
+    DEFAULT_LAST_DATA_MIGRATION,
     DEFAULT_PREMIUM_SHOULD_SYNC,
     ROTKEHLCHEN_DB_VERSION,
     ROTKEHLCHEN_TRANSIENT_DB_VERSION,
@@ -190,10 +192,10 @@ class DBHandler:
         self.setting_to_default_type = {
             'version': (int, ROTKEHLCHEN_DB_VERSION),
             'last_write_ts': (int, Timestamp(0)),
-            'last_data_upload_ts': (int, Timestamp(0)),
             'premium_should_sync': (str_to_bool, DEFAULT_PREMIUM_SHOULD_SYNC),
             'main_currency': (lambda x: Asset(x).resolve(), A_USD.resolve_to_fiat_asset()),
             'ongoing_upgrade_from_version': (int, None),
+            'last_data_migration': (int, DEFAULT_LAST_DATA_MIGRATION),
         }
         self.conn: DBConnection = None  # type: ignore
         self.conn_transient: DBConnection = None  # type: ignore
@@ -340,7 +342,7 @@ class DBHandler:
         ...
 
     @overload
-    def get_setting(self, cursor: 'DBCursor', name: Literal['last_write_ts', 'last_data_upload_ts']) -> Timestamp:  # noqa: E501
+    def get_setting(self, cursor: 'DBCursor', name: Literal['last_write_ts']) -> Timestamp:
         ...
 
     @overload
@@ -355,16 +357,20 @@ class DBHandler:
     def get_setting(self, cursor: 'DBCursor', name: Literal['ongoing_upgrade_from_version']) -> int | None:  # noqa: E501
         ...
 
+    @overload
+    def get_setting(self, cursor: 'DBCursor', name: Literal['last_data_migration']) -> int | None:
+        ...
+
     def get_setting(
             self,
             cursor: 'DBCursor',
             name: Literal[
                 'version',
                 'last_write_ts',
-                'last_data_upload_ts',
                 'premium_should_sync',
                 'main_currency',
                 'ongoing_upgrade_from_version',
+                'last_data_migration',
             ],
     ) -> int | None | (Timestamp | (bool | AssetWithOracles)):
         deserializer, default_value = self.setting_to_default_type[name]
@@ -383,7 +389,6 @@ class DBHandler:
             name: Literal[
                 'version',
                 'last_write_ts',
-                'last_data_upload_ts',
                 'premium_should_sync',
                 'ongoing_upgrade_from_version',
                 'main_currency',
@@ -582,6 +587,38 @@ class DBHandler:
             list(settings_dict.items()),
         )
         CachedSettings().update_entries(settings)
+
+    @overload
+    def get_cache(self, cursor: 'DBCursor', name: DBCache) -> Timestamp:
+        ...
+
+    @overload
+    def get_cache(self, cursor: 'DBCursor') -> dict[DBCache, Timestamp]:
+        ...
+
+    def get_cache(
+            self,
+            cursor: 'DBCursor',
+            name: DBCache | None = None,
+    ) -> dict[DBCache, Timestamp] | Timestamp:
+        """Returns the cache values from the `key_value_cache` table of the DB
+        - If `name` is not passed, it returns all the values in the table
+        - If `name` is passed, it returns the value of the given name, defaults to Timestamp(0)"""
+        if name is None:
+            cursor.execute('SELECT name, value FROM key_value_cache;')
+            return {
+                DBCache.deserialize(q[0]): Timestamp(int(q[1]))
+                for q in cursor
+            }
+        value = cursor.execute('SELECT value FROM key_value_cache WHERE name=?;', (name.value,)).fetchone()  # noqa: E501
+        return Timestamp(0) if value is None else Timestamp(int(value[0]))
+
+    def set_cache(self, write_cursor: 'DBCursor', cache: dict[DBCache, Timestamp]) -> None:
+        """Save the values of the cache to the `key_value_cache` table of the DB"""
+        write_cursor.executemany(
+            'INSERT OR REPLACE INTO key_value_cache(name, value) VALUES(?, ?)',
+            [(name.value, value) for name, value in cache.items()],
+        )
 
     @need_writable_cursor('user_write')
     def add_external_service_credentials(
--- a/rotkehlchen/db/settings.py
+++ b/rotkehlchen/db/settings.py
@@ -4,16 +4,9 @@ from dataclasses import dataclass, field, fields
 from typing import Any, Literal, NamedTuple, Optional
 
 from rotkehlchen.assets.asset import Asset, AssetWithOracles
-from rotkehlchen.chain.constants import LAST_EVM_ACCOUNTS_DETECT_KEY
 from rotkehlchen.constants.assets import A_USD
-from rotkehlchen.constants.misc import (
-    LAST_AUGMENTED_SPAM_ASSETS_DETECT_KEY,
-    LAST_OWNED_ASSETS_UPDATE,
-    LAST_SPAM_ASSETS_DETECT_KEY,
-)
 from rotkehlchen.constants.timing import YEAR_IN_SECONDS
 from rotkehlchen.data_migrations.constants import LAST_DATA_MIGRATION
-from rotkehlchen.db.constants import LAST_DATA_UPDATES_KEY, UpdateType
 from rotkehlchen.db.utils import str_to_bool
 from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.history.types import DEFAULT_HISTORICAL_PRICE_ORACLES_ORDER, HistoricalPriceOracle
@@ -101,22 +94,12 @@ STRING_KEYS = (
     'date_display_format',
     'frontend_settings',
 )
-TIMESTAMP_KEYS = ('last_write_ts', 'last_data_upload_ts', 'last_balance_save')
-IGNORED_KEYS = (
-    LAST_EVM_ACCOUNTS_DETECT_KEY,
-    LAST_DATA_UPDATES_KEY,
-    LAST_SPAM_ASSETS_DETECT_KEY,
-    LAST_AUGMENTED_SPAM_ASSETS_DETECT_KEY,
-    LAST_OWNED_ASSETS_UPDATE,
-) + tuple(x.serialize() for x in UpdateType)
-
 
 CachedDBSettingsFieldNames = Literal[
     'have_premium',
     'version',
     'premium_should_sync',
     'include_crypto2crypto',
-    'last_data_upload_ts',
     'ui_floating_precision',
     'taxfree_after_period',
     'balance_save_frequency',
@@ -125,7 +108,6 @@ CachedDBSettingsFieldNames = Literal[
     'dot_rpc_endpoint',
     'main_currency',
     'date_display_format',
-    'last_balance_save',
     'submit_usage_analytics',
     'active_modules',
     'frontend_settings',
@@ -155,7 +137,6 @@ CachedDBSettingsFieldNames = Literal[
 DBSettingsFieldTypes = (
     bool |
     int |
-    Timestamp |
     str |
     Asset |
     Sequence[ModuleName] |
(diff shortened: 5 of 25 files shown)


rotkehlchen/api/rest.py
```diff
@@ -449,12 +449,15 @@ def set_settings(self, settings: ModifiableDBSettings) -> Response:
...
-        result_dict = {'result': new_settings, 'message': ''}
+            cache = process_result(self.rotkehlchen.get_cache(cursor))
+        result_dict = {'result': new_settings | cache, 'message': ''}
         return api_response(result=result_dict, status_code=HTTPStatus.OK)
 
     def get_settings(self) -> Response:
         with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
-            result_dict = _wrap_in_ok_result(process_result(self.rotkehlchen.get_settings(cursor)))
+            settings = self.rotkehlchen.get_settings(cursor)
+            cache = self.rotkehlchen.get_cache(cursor)
```
yabirgb, line 459: why do you need it? for the `last_balance_save`? if is only one key maybe you could create a method to query only certains keys exposed to the frontend and return them instead of everything

rotkehlchen/db/cache.py
```diff
@@ -0,0 +1,31 @@
+
+from dataclasses import dataclass, field, fields
+from typing import Any
+
+from rotkehlchen.types import Timestamp
+
+
+DBCacheFieldTypes = int | Timestamp
```
yabirgb, line 8: why this type? aren't all the values timestamps?

rotkehlchen/db/dbhandler.py
```diff
@@ -583,6 +574,20 @@ def set_settings(self, write_cursor: 'DBCursor', settings: ModifiableDBSettings)
         )
         CachedSettings().update_entries(settings)
 
+    def get_cache(self, cursor: 'DBCursor') -> DBCache:
+        cursor.execute('SELECT name, value FROM key_value_cache;')
+        cache_dict = {}
+        for q in cursor:
+            cache_dict[q[0]] = q[1]
+        return DBCache.deserialize(cache_dict)
```
yabirgb, line 582: instead of this I believe you can just do

```suggestion
        return DBCache.deserialize(dict(cursor))
```

but check if it works

rotkehlchen/db/dbhandler.py
```diff
@@ -583,6 +574,20 @@ def set_settings(self, write_cursor: 'DBCursor', settings: ModifiableDBSettings)
         )
         CachedSettings().update_entries(settings)
 
+    def get_cache(self, cursor: 'DBCursor') -> DBCache:
+        cursor.execute('SELECT name, value FROM key_value_cache;')
```
yabirgb, line 578: add a small docstring explaining what it does

rotkehlchen/db/dbhandler.py
```diff
@@ -583,6 +574,20 @@ def set_settings(self, write_cursor: 'DBCursor', settings: ModifiableDBSettings)
...
+        for q in cursor:
+            cache_dict[q[0]] = q[1]
+        return DBCache.deserialize(cache_dict)
+
+    def set_cache(self, write_cursor: 'DBCursor', cache: DBCache) -> None:
+        cache_dict = cache.serialize()
+        write_cursor.executemany(
+            'INSERT OR REPLACE INTO key_value_cache(name, value) VALUES(?, ?)',
+            list(cache_dict.items()),
+        )
```
yabirgb, line 589: don't create variables that are used only once. Also add a small docstring

```suggestion
        write_cursor.executemany(
            'INSERT OR REPLACE INTO key_value_cache(name, value) VALUES(?, ?)',
            list(cache.serialize().items()),
        )
```

rotkehlchen/db/drivers/gevent.py
```diff
@@ -366,12 +366,19 @@ def write_ctx(self, commit_ts: bool = False) -> Generator['DBCursor', None, None
...
-                    # last_write_ts in not cached to cached settings. This is a critical section
-                    # and adding even one more function call can have very ugly and
+                    try:
+                        cursor.execute(
+                            'INSERT OR REPLACE INTO key_value_cache(name, value) VALUES(?, ?)',
+                            ('last_write_ts', str(ts_now())),
+                        )
+                    except sqlcipher.OperationalError:  # pylint: disable=no-member
+                        cursor.execute(  # when key_value_cache is missing go back to settings
+                            # it happens while upgrading
```
yabirgb, line 376: leave the comment in one line or add it over the cursor.execute statement. Also Maybe add that this is needed as long as we have v40->v41 upgrade

rotkehlchen/db/cache.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=False)
+class DBCache:
```
yabirgb, line 12: aren't you missing the UpdateType version keys?

rotkehlchen/db/settings.py
```diff
@@ -339,8 +313,8 @@ def db_settings_from_dict(
         elif key == 'address_name_priority':
             specified_args[key] = json.loads(value)
         else:
-            if key == 'eth_rpc_endpoint':
-                continue  # temporary since setting is removed in migration and may still get here
+            if key in {'eth_rpc_endpoint'} | SETTINGS_MOVED:  # temporary since these settings
+                continue  # are moved in migrations/updates and may still get here
```
yabirgb, line 317: do we load the settings during the upgrade?

rotkehlchen/db/dbhandler.py
yabirgb: Also this will contain the keys from the updates right? Shouldn't you handle it somehow?

rotkehlchen/db/upgrades/v40_v41.py
```diff
@@ -22,15 +36,41 @@ def _add_cache_table(write_cursor: 'DBCursor') -> None:
     log.debug('Exit _add_cache_table')
 
 
+def _get_unrelated_mappings(read_cursor: 'DBCursor') -> dict:
+    """Get the unrelated key-value pairs from `settings`"""
+    log.debug('Enter _get_unrelated_mappings')
+    settings = read_cursor.execute('SELECT name, value FROM settings').fetchall()
+    log.debug('Exit _get_unrelated_mappings')
+    return {name: value for name, value in settings if name in SETTINGS_MOVED}
```
yabirgb, line 44: don't create a function for this is just a query that you can do in the other function

rotkehlchen/db/upgrades/v40_v41.py
```diff
@@ -22,15 +36,41 @@ def _add_cache_table(write_cursor: 'DBCursor') -> None:
     log.debug('Exit _add_cache_table')
 
 
+def _get_unrelated_mappings(read_cursor: 'DBCursor') -> dict:
+    """Get the unrelated key-value pairs from `settings`"""
+    log.debug('Enter _get_unrelated_mappings')
+    settings = read_cursor.execute('SELECT name, value FROM settings').fetchall()
```
yabirgb, line 42: don't do a fetchall if you are going to iterate after. Fetchall is already a loop in the cursor and then you loop it again

rotkehlchen/db/upgrades/v40_v41.py
```diff
@@ -22,15 +36,41 @@ def _add_cache_table(write_cursor: 'DBCursor') -> None:
     log.debug('Exit _add_cache_table')
 
 
+def _get_unrelated_mappings(read_cursor: 'DBCursor') -> dict:
```
yabirgb, line 39: why call it unrelated?

rotkehlchen/premium/sync.py
```diff
@@ -42,7 +43,7 @@ def __init__(self, migration_manager: DataMigrationManager, data: DataHandler) -
         # Initialize this with the value saved in the DB
         with data.db.conn.read_ctx() as cursor:
             # These 2 vars contain the timestamp of our side. When did this DB try to upload
-            self.last_data_upload_ts = data.db.get_setting(cursor, name='last_data_upload_ts')
+            self.last_data_upload_ts = data.db.get_cache(cursor).last_data_upload_ts
```
yabirgb, line 46: could we keep the other interface for selecting the name of the key that we want? the problem with the approach of querying the object is that you do a serialization and we build all the object with all the fields when we need only one

rotkehlchen/premium/sync.py
```diff
@@ -247,7 +248,7 @@ def maybe_upload_data_to_server(
         self.last_upload_attempt_ts = self.last_data_upload_ts
         self.last_remote_data_upload_ts = self.last_data_upload_ts
         with self.data.db.user_write() as cursor:
-            self.data.db.set_setting(cursor, name='last_data_upload_ts', value=self.last_data_upload_ts)  # noqa: E501
+            self.data.db.set_cache(cursor, DBCache(last_data_upload_ts=self.last_data_upload_ts))
```
yabirgb, line 251: if you do this aren't you overwritting all the other values in the database with the default? `0`

yabirgb: This is why I like the other interface more. Needing to build all the object to modify a single entry is cumbersome

rotkehlchen/tests/db/test_db.py
```diff
@@ -487,33 +489,34 @@ def test_settings_entry_types(database):
...
-    assert isinstance(res.date_display_format, str)
-    assert res.date_display_format == '%d/%m/%Y %H:%M:%S %z'
-    assert isinstance(res.submit_usage_analytics, bool)
-    assert res.submit_usage_analytics is False
-    assert isinstance(res.active_modules, tuple)
-    assert res.active_modules == DEFAULT_ACTIVE_MODULES
-    assert isinstance(res.frontend_settings, str)
-    assert res.frontend_settings == ''
+        settings = database.get_settings(cursor)
+        cache = database.get_cache(cursor)
```
yabirgb, line 493: did you forget to check for the types in the cache?

rotkehlchen/tests/db/test_db.py
```diff
@@ -487,33 +489,34 @@ def test_settings_entry_types(database):
...
+    assert isinstance(settings.include_crypto2crypto, bool)
+    assert settings.include_crypto2crypto is True
+    assert isinstance(settings.ui_floating_precision, int)
+    assert settings.ui_floating_precision == 1
+    assert isinstance(settings.taxfree_after_period, int)
+    assert settings.taxfree_after_period == 1
+    assert isinstance(settings.balance_save_frequency, int)
+    assert settings.balance_save_frequency == 24
+    assert isinstance(cache.last_balance_save, int)
+    assert cache.last_balance_save == 0
```
yabirgb, line 509: there is only this one If I'm not wrong. Maybe create a different function for this one and test the other fields?

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1989,8 +1989,7 @@ def test_upgrade_db_40_to_41(user_data_dir):
     )""") is True
         cursor.execute('INSERT INTO key_value_cache VALUES (?, ?)', ('test', 'test'))
         cursor.execute('SELECT * FROM key_value_cache')
-        assert cursor.fetchone() == ('test', 'test')
-        assert cursor.fetchall() == []
+        assert ('test', 'test') in cursor.fetchall()
```
yabirgb, line 1992:
```suggestion
        assert [('test', 'test')] == cursor.fetchall()
```

yabirgb: You should add to the test db values for all the settings and check that they got correctly moved

rotkehlchen/tests/unit/test_tokens.py
```diff
@@ -211,9 +212,9 @@ def _do_spawn(database):
     while True:
         gevent.spawn(_do_read, database)
         with database.user_write() as write_cursor:
-            database.set_setting(write_cursor, 'last_write_ts', 15)
+            database.set_cache(write_cursor, DBCache(last_write_ts=Timestamp(15)))
```
yabirgb, line 215: same comment. With this approach I believe you are overwritting all the settings. Check it

yabirgb (commented):
Some comments. If I'm not wrong I believe you are overwritting the settings when saving them to the cache. Check it to verify.

rotkehlchen/db/settings.py
OjusWiZard: It happens during the migration and fails in the this test: `rotkehlchen/tests/data_migrations/test_migrations.py::test_migration_1[None-False-0-data_migration_v0.db] `

OjusWiZard:
The above suggestions are resolved. But I missed updating some places where `get_settings` should be replaced by `get_cache`, checking those.

rotkehlchen/db/dbhandler.py
```diff
@@ -382,10 +388,8 @@ def set_setting(
             write_cursor: 'DBCursor',
             name: Literal[
                 'version',
-                'last_write_ts',
-                'last_data_upload_ts',
                 'premium_should_sync',
-                'ongoing_upgrade_from_version',
```
yabirgb, line 388: Maybe instead of using literals it would make sense to have an enum like we do for `class CacheType(Enum):`. You can make it serializableEnumMixing

rotkehlchen/api/rest.py
```diff
@@ -449,12 +449,18 @@ def set_settings(self, settings: ModifiableDBSettings) -> Response:
...
 
     def get_settings(self) -> Response:
         with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
-            result_dict = _wrap_in_ok_result(process_result(self.rotkehlchen.get_settings(cursor)))
+            settings = process_result(self.rotkehlchen.get_settings(cursor))
+            cache = process_result(self.rotkehlchen.get_cache(cursor))
+        result_dict = _wrap_in_ok_result(settings | {
+            name: value for name, value in cache.items()
+            if name in {'last_write_ts', 'last_data_upload_ts', 'last_balance_save'}
+        })
```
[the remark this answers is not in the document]
OjusWiZard: They do parse it in `frontend/app/src/types/user.ts` and it's also present in the docs, so by including it here we can avoid changing the docs and this frontend file.

rotkehlchen/db/dbhandler.py
```diff
@@ -583,6 +581,52 @@ def set_settings(self, write_cursor: 'DBCursor', settings: ModifiableDBSettings)
...
+        """Returns the cache values from the `key_value_cache` table of the DB
+        If the given name doesn't exist in the table, returns 0 for timestamps, otherwise None"""
+        if name is None:
+            cursor.execute('SELECT name, value FROM key_value_cache;')
+            return {
+                DBCache.deserialize(q[0]): get_db_cache_value(DBCache.deserialize(q[0]), q[1])
+                for q in cursor
+            }
+        value = cursor.execute('SELECT value FROM key_value_cache WHERE name=?;', (name.serialize(),)).fetchone()  # noqa: E501
+        return 0 if value is None else get_db_cache_value(name, value[0])
```
yabirgb, line 621: shouldn't this be?

```suggestion
        return get_db_cache_value(name, 0) if value is None else get_db_cache_value(name, value[0])
```

rotkehlchen/db/upgrades/v40_v41.py
```diff
@@ -22,15 +26,35 @@ def _add_cache_table(write_cursor: 'DBCursor') -> None:
     log.debug('Exit _add_cache_table')
 
 
+def _move_non_settings_mappings_to_cache(write_cursor: 'DBCursor') -> None:
+    """Move the non-settings value from `settings` to a seperate `key_value_cache` table"""
+    log.debug('Enter _move_non_settings_mappings_to_cache')
+    movable_settings = write_cursor.execute(
+        f'SELECT name, value FROM settings WHERE name IN ({",".join(["?"] * len(SETTINGS_MOVED))});',  # noqa: E501
+        tuple(SETTINGS_MOVED),
+    ).fetchall()
```
yabirgb, line 35: if it is used in one here don't create a global variable

```suggestion
    settings_moved = {x.serialize() for x in DBCache}
    movable_settings = write_cursor.execute(
        f'SELECT name, value FROM settings WHERE name IN ({",".join(["?"] * len(settings_moved))});',  # noqa: E501
        tuple(settings_moved),
    ).fetchall()
```

all of them have the same name right?

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1971,7 +1974,13 @@ def test_upgrade_db_40_to_41(user_data_dir):
         msg_aggregator=msg_aggregator,
         resume_from_backup=False,
     )
+    # add some settings that should move in the upgrade
+    settings = {
+        name: str(randint(0, 1000)) for name in SETTINGS_MOVED
+        if name is not DBCache.LAST_WRITE_TS.serialize()
+    }
```
yabirgb, line 1981: instead of doing this using structures from the version hardcode them. the reason is that if for some reason we change the logic the test won't be valid anymore

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1971,7 +1974,13 @@ def test_upgrade_db_40_to_41(user_data_dir):
...
     with db_v40.conn.write_ctx() as cursor:
+        cursor.executemany('INSERT INTO settings VALUES (?, ?)', tuple(settings.items()))
```
yabirgb, line 1983: do you need the tuple thing?

yabirgb (commented):
Unfortunatly there are still some issues

rotkehlchen/db/upgrades/v40_v41.py
```diff
@@ -22,15 +23,36 @@ def _add_cache_table(write_cursor: 'DBCursor') -> None:
     log.debug('Exit _add_cache_table')
 
 
+def _move_non_settings_mappings_to_cache(write_cursor: 'DBCursor') -> None:
+    """Move the non-settings value from `settings` to a seperate `key_value_cache` table"""
+    log.debug('Enter _move_non_settings_mappings_to_cache')
+    settings_moved = [x.serialize() for x in DBCache]
```
yabirgb, line 29: nitpick: about what I said in the other comment I believe is better to have them hardcoded since if we delete something or rename something this can fail

docs/api.rst
```diff
@@ -139,7 +139,6 @@ Handling user creation, sign-in, log-out and querying
               "settings": {
                   "have_premium": true,
                   "version": "6",
-                  "last_write_ts": 1571552172,
```
LefterisJP, line 142: Where is this going to if it's removed from here?

rotkehlchen/db/db_cache.py
```diff
@@ -0,0 +1,74 @@
+from enum import auto
+from rotkehlchen.errors.serialization import DeserializationError
+from rotkehlchen.types import Timestamp
+from rotkehlchen.utils.mixins.enums import Enum
```
LefterisJP, line 4: change the name of this file. It's already under the DB directory. Name it `cache.py`

rotkehlchen/data_migrations/manager.py
```diff
@@ -45,15 +45,16 @@ def __init__(self, rotki: 'Rotkehlchen'):
 
     def maybe_migrate_data(self) -> None:
         with self.rotki.data.db.conn.read_ctx() as cursor:
-            settings = self.rotki.data.db.get_settings(cursor)
-        last_migration_version = settings.last_data_migration
+            # it's important to use get_setting instead of get_settings here, because during
+            # old migrations we should not pass last_data_migration through db_settings_from_dict
+            last_migration_version = self.rotki.data.db.get_setting(cursor, 'last_data_migration')
```
LefterisJP, line 50: I told you to leave the last data migration in the settings. Same as version. Do not do this. There is no need to add extra logic and backwards incompatible changes here

rotkehlchen/db/db_cache.py
```diff
@@ -0,0 +1,74 @@
...
+    """It contains all the values that can be stored in the `key_value_cache` table of the DB"""
+    LAST_WRITE_TS = auto()
+    LAST_BALANCE_SAVE = auto()
+    LAST_DATA_UPLOAD_TS = auto()
+    LAST_DATA_UPDATES_TS = auto()
+    LAST_OWNED_ASSETS_UPDATE = auto()
+    LAST_EVM_ACCOUNTS_DETECT_TS = auto()
+    LAST_SPAM_ASSETS_DETECT_KEY = auto()
+    LAST_AUGMENTED_SPAM_ASSETS_DETECT_KEY = auto()
+    SPAM_ASSETS_VERSION = auto()
```
LefterisJP, line 20: Aren't all these `_VERSION` keys charactherizations of the DB? They are not part of the cache. I don't see why you need them to be here.

rotkehlchen/db/db_cache.py
```diff
@@ -0,0 +1,74 @@
...
+    RPC_NODES_VERSION = auto()
+    CONTRACTS_VERSION = auto()
+    GLOBAL_ADDRESSBOOK_VERSION = auto()
+    ACCOUNTING_RULES_VERSION = auto()
+
+    def __str__(self) -> str:
+        """string constructor"""
```
LefterisJP, line 27: useless comment

```suggestion
```

rotkehlchen/db/db_cache.py
```diff
@@ -0,0 +1,74 @@
...
+    elif name in {
+        DBCache.SPAM_ASSETS_VERSION,
+        DBCache.RPC_NODES_VERSION,
+        DBCache.CONTRACTS_VERSION,
+        DBCache.GLOBAL_ADDRESSBOOK_VERSION,
+        DBCache.ACCOUNTING_RULES_VERSION,
+    }:
+        return 0 if value is None else int(value)
+    else:  # some untracked key-value pair is passed, returning as it is
+        return value
```
LefterisJP, line 74: Souldn't this be an assertion? This is a programming error.

rotkehlchen/db/dbhandler.py
```diff
@@ -583,6 +581,52 @@ def set_settings(self, write_cursor: 'DBCursor', settings: ModifiableDBSettings)
...
+        DBCache.GLOBAL_ADDRESSBOOK_VERSION,
+        DBCache.ACCOUNTING_RULES_VERSION,
+    ]) -> int:
+        ...
+
+    @overload
+    def get_cache(self, cursor: 'DBCursor') -> dict[DBCache, DBCacheFieldTypes]:
+        ...
+
+    def get_cache(self, cursor: 'DBCursor', name: DBCache | None = None) -> dict[DBCache, DBCacheFieldTypes] | DBCacheFieldTypes:  # noqa: E501
```
LefterisJP, line 611: Split each argument in its own line.

rotkehlchen/db/dbhandler.py
```diff
@@ -583,6 +581,52 @@ def set_settings(self, write_cursor: 'DBCursor', settings: ModifiableDBSettings)
...
+        """Returns the cache values from the `key_value_cache` table of the DB
+        If the given name doesn't exist in the table, returns 0 for timestamps, otherwise None"""
```
LefterisJP, line 613: Waht do you mean if the given name does not exist in the table? Don't you mean if there is nothing passed?

rotkehlchen/db/drivers/gevent.py
```diff
@@ -366,12 +367,20 @@ def write_ctx(self, commit_ts: bool = False) -> Generator['DBCursor', None, None
...
             else:
                 if commit_ts is True:
-                    cursor.execute(
-                        'INSERT OR REPLACE INTO settings(name, value) VALUES(?, ?)',
-                        ('last_write_ts', str(ts_now())),
-                    )
-                    # last_write_ts in not cached to cached settings. This is a critical section
-                    # and adding even one more function call can have very ugly and
+                    try:
+                        cursor.execute(
```
LefterisJP, line 371: This is not good. This is a very critical section of the app. I don't want this change here. Even the Enum.serialize() part is bad. This happens for every single DB write. This should not go here. U/ndo this change completely.

rotkehlchen/db/db_cache.py
```diff
@@ -0,0 +1,74 @@
...
+from rotkehlchen.types import Timestamp
+from rotkehlchen.utils.mixins.enums import Enum
+
+
+DBCacheFieldTypes = int | Timestamp
+
+
+class DBCache(Enum):
+    """It contains all the values that can be stored in the `key_value_cache` table of the DB"""
+    LAST_WRITE_TS = auto()
```
LefterisJP, line 12: This is a very silly way to do this enum. You are adding A LOT of extra logic to "serialize/deserialize" and calling the "serialize()` function in every single DB write. This is a very very bad idea. Just use the value of the enum and change it away from auto() to the string value.

LefterisJP (commented):
Not very happy with how this looks like. Need many changes.

rotkehlchen/data_migrations/manager.py
OjusWiZard: Actually, the comment was outdated, it's about `last_write_ts` instead of `last_data_migration`. Even if we keep `last_data_migration` in the settings table, but then call `.get_setttings`. The `last_write_ts` will be fetched from the settings table in old DB versions and passed through `db_settings_from_dict`. Which raises a warning: `Unknown DB setting last_write_ts given. Ignoring it. Should not happen so please open an issue in Github.`

rotkehlchen/db/dbhandler.py
OjusWiZard: Updated the comment to make it more clear:
```
Returns the cache values from the `key_value_cache` table of the DB
- If `name` is not passed, it returns all the values in the table
- If `name` is passed, it returns the value of the given name, defaults to Timestamp(0)
````

File: docs/api.rst (hunk 25 above)
Comment (OjusWiZard) [reply docs/api.rst:142]: We decided to remove `last_write_ts` from the API because it is not being used on the frontend.

File: rotkehlchen/db/drivers/gevent.py (hunk 33 above)
Comment (OjusWiZard) [reply rotkehlchen/db/drivers/gevent.py:371]: Updated it like this:
```py
try:
    cursor.execute(
        'INSERT OR REPLACE INTO key_value_cache(name, value) VALUES(?, ?)',
        ('last_write_ts', str(ts_now())),
    )
except sqlcipher.OperationalError:  # pylint: disable=no-member
    # when key_value_cache is missing, go back to settings
    # it happens while upgrading until DB version v41
    cursor.execute(
        ('last_write_ts', str(ts_now())),
    )
# last_write_ts is updated to settings/key_value_cache. This is a critical
# section and adding even one more function call can have very ugly and
# detrimental effects in the entire codebase as everything calls this.
```

The `.serialize()` function is removed, but still using try-except. Because during db upgrades and migrations in old versions, `key_value_cache` table is not present.

docs/api.rst
yabirgb: Is not that we decided unilaterally. We asked and confirmed that this field is not used

rotkehlchen/db/db_cache.py
yabirgb: you didn't answer this one

rotkehlchen/db/db_cache.py
yabirgb: you didn't do this one

yabirgb: I believe you pushed this file instead of renamign

rotkehlchen/db/drivers/gevent.py
yabirgb: this try/except happens only for the case of the upgrade right? so the usual path is just doing the insert and exiting

rotkehlchen/db/db_cache.py
OjusWiZard: Yes, I've removed them from the cache now.

rotkehlchen/db/drivers/gevent.py
LefterisJP: Still not good. I am sorry but I don't think you realize what the critical path means.  try/excpet is not free. And having an unecessary try/except just for that 1 edge case is not acceptable.

LefterisJP (approved):
LGTM. Made a note for last_write_ts: https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=48663574

PR #7183, opened 2023-12-27
feat(frontend): allow running task cancellation

### LINKED ISSUE #6254: Introduce the ability to cancel async queries
## Abstract

We need the endpoint `DELETE /api/1/tasks/{task_id}` to provide the ability to terminate an in-progress `async_query`.

Integrate this functionality with the pending tasks in the notification area so you can stop the execution of a pending task.

## Motivation

Sometimes you might need to cancel long-running tasks for queries that have been rate-limited for hours.

## Specification

- `DELETE /api/1/tasks/{task_id}`
- Frontend integration

tewshi:
Closes #6254 

## Checklist


frontend/app/src/locales/en.json
```diff
@@ -1518,7 +1518,8 @@
     "title": "Exchange Trades"
   },
   "collapsed_pending_tasks": {
-    "title": "{count} pending task | {count} pending tasks"
+    "title": "{count} pending task | {count} pending tasks",
+    "cancel_task": "cancel task"
```
lukicenturi, line 1522: "Cancel task"

lukicenturi:
I think a confirmation dialog is needed when you click the cancel task button.

lukicenturi:
I wonder if we somehow can unify the implementation for the `UserCancelledTaskError` in catch.
Since the only thing it does is `logger.debug`.

frontend/app/src/store/tasks/index.ts
```diff
@@ -169,6 +190,18 @@ export const useTaskStore = defineStore('tasks', () => {
           } else if (result === null) {
             let errorMessage: string;
             if (message) {
+              if (message === USER_CANCELLED_TASK) {
+                if (checkIfDevelopment() && !import.meta.env.VITE_TEST) {
```
lukicenturi, line 194: maybe put the logger here.
and in other task caller, in the `catch part`, just do:

`if (!(e instanceof UserCancelledTaskError)) {...`

tewshi:
> I think a confirmation dialog is needed when you click the cancel task button.

okay, should do.

frontend/app/src/store/tasks/index.ts
tewshi: okay, I see, makes sense.

frontend/app/src/composables/assets/retrieval.ts
```diff
@@ -154,13 +154,15 @@ export const useAssetInfoRetrieval = () => {
...
       return result;
     } catch (e: any) {
-      notify({
-        title: t('actions.assets.erc20.error.title', payload),
-        message: t('actions.assets.erc20.error.description', {
-          message: e.message
-        }),
-        display: true
-      });
+      if (!(e instanceof UserCancelledTaskError)) {
```
lukicenturi, line 157: After I think again, is it makes sense to move this to a function instead?
Something like `processError` that returns boolean

lukicenturi: and also if possible, use return early.
`if(!processError(e)) return`

tewshi: done

lukicenturi (approved):
LGTM!

PR #7185, opened 2023-12-27
Switch the DB to WAL mode

LefterisJP:
https://www.sqlite.org/wal.html

yolo for testing


kelsos:
isn't WAL the default nowadays?

LefterisJP:
> isn't WAL the default nowadays?

no no no. Sqlite defaults are etched in stone and really old. Got lots of issues to "adjust stuff" because of it.

And the reason to switch to WAL is multi/read-write works correctly (the multiprocesses problem we want to solve) with it and also unexpected shutdowns recover properly.

Finally wanna see how it works with gevent and if we have less problems ... though not holding my breath there.

PR #7194, opened 2023-12-29
No etherscan backoff for max retries exceeded with url

LefterisJP:
This was an old way we used to check for 429 backoff from etherscan. This error though will also pop up if the server has problems, if we can't resolve the domain name and more.

In all those cases rotki would backoff slowly and retry, taking a lot more time instead of failing immediately.

I spoke with etherscan people and they should now always use 429 when they need you to backoff. So this code is no longer necessary


yabirgb (approved):
lgtm

PR #7196, opened 2023-12-29
Cache timestamp to block queries to etherscan

yabirgb:
When a user tracks many addresses we end up repeating the same queries for each addresses. In the case of the PnL report we were querying for example for each address the block for the to/from. This was causing rate limits and slowing the app due to back offs


LefterisJP (approved):
LGTM

PR #7201, opened 2024-01-01
Fix (#7178): Follow-up fixes

OjusWiZard:
## Checklist

- [x] Add missing cache fields for settings in `/users` and `/history/debug`
- [x] Add support for `UpdateType`_versions in `DBSettings`
- [x] Fix bugs for not moving `UpdateType`_versions in DB upgrade

### DIFF
--- a/rotkehlchen/db/settings.py
+++ b/rotkehlchen/db/settings.py
@@ -7,6 +7,7 @@ from rotkehlchen.assets.asset import Asset, AssetWithOracles
 from rotkehlchen.constants.assets import A_USD
 from rotkehlchen.constants.timing import YEAR_IN_SECONDS
 from rotkehlchen.data_migrations.constants import LAST_DATA_MIGRATION
+from rotkehlchen.db.constants import UpdateType
 from rotkehlchen.db.utils import str_to_bool
 from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.history.types import DEFAULT_HISTORICAL_PRICE_ORACLES_ORDER, HistoricalPriceOracle
@@ -94,6 +95,8 @@ STRING_KEYS = (
     'date_display_format',
     'frontend_settings',
 )
+
+UPDATE_TYPES_VERSIONS = {x.serialize() for x in UpdateType}
 
 CachedDBSettingsFieldNames = Literal[
     'have_premium',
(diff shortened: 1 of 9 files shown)


rotkehlchen/db/settings.py
```diff
@@ -278,6 +289,8 @@ def db_settings_from_dict(
             specified_args[key] = int(value)
         elif key in STRING_KEYS:
             specified_args[key] = str(value)
+        elif key in UPDATE_TYPES:
+            specified_args[key] = value
```
yabirgb, line 293: this will never return None right? so maybe return int(value)?

rotkehlchen/db/updates.py
```diff
@@ -372,7 +372,7 @@ def check_for_updates(self, updates: Sequence[UpdateType] = tuple(UpdateType)) -
     def _check_for_last_version(cursor: 'DBCursor', update_type: UpdateType) -> int:
         """This method checks the database for the last local version of `update_type`"""
         found_version = cursor.execute(
-            'SELECT value FROM key_value_cache WHERE name=?',
+            'SELECT value FROM settings WHERE name=?',
```
yabirgb, line 375: why this change? Weren't they suppose to be in the value cache table?

OjusWiZard: We kept them in the settings table in the last PR https://github.com/rotki/rotki/pull/7178/#discussion_r1438117221

LefterisJP, line 375: This is really bad that the change was not detected. Why did the tests here not fail before? 

Can you add a test that uses this so that this thing actually works and is covered?

rotkehlchen/tests/api/test_settings.py
```diff
@@ -100,7 +101,8 @@ def test_querying_settings(rotkehlchen_api_server, username):
     assert json_data['message'] == ''
     assert result['version'] == ROTKEHLCHEN_DB_VERSION
     for setting in dataclasses.fields(DBSettings):
-        assert setting.name in result
+        if setting.name not in UPDATE_TYPES:
```
LefterisJP, line 104: why do this? Makes no sense to inroduce this change

LefterisJP (approved):
LGTM

PR #7208, opened 2024-01-02
Adjust tests to work in windows

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -3638,7 +3638,7 @@ class RestAPI:
                 if 'assets.json' not in zip_file.namelist():
                     raise ValidationError('assets.json could not be found in the provided zip file.')  # noqa: E501
 
-                with tempfile.TemporaryDirectory() as tempdir:
+                with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as tempdir:  # needed on windows, see https://tinyurl.com/tmp-win-err  # noqa: E501
                     zip_file.extract('assets.json', tempdir)
                     import_assets_from_file(
                         path=Path(tempdir) / 'assets.json',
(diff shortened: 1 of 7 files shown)


rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2237,5 +2237,8 @@ def test_unfinished_upgrades(user_data_dir):
 
             for f in os.listdir(user_data_dir):
                 if f.endswith('backup'):
-                    (Path(user_data_dir) / f).unlink()
+                    try:
```
LefterisJP, line 2240: use `with suppress`

LefterisJP (commented):
Reviewed. Some tests failed, but did not check out why

rotkehlchen/utils/misc.py
```diff
@@ -115,12 +115,17 @@ def timestamp_to_date(
         formatstr: str = '%d/%m/%Y %H:%M:%S',
         treat_as_local: bool = False,
 ) -> str:
-    """Transforms a timestamp to a datestring depending on given formatstr and UTC/local choice"""
+    """
+    Transforms a timestamp to a datestring depending on given formatstr and UTC/local choice
```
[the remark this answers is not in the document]
yabirgb: Is this issue again https://github.com/python/cpython/issues/107078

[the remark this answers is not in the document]
LefterisJP: Then add a link to it

rotkehlchen/api/rest.py
```diff
@@ -4469,9 +4469,16 @@ def export_history_events(
             headers.update({key: None for key in serialized_event})
 
         if directory_path is None:  # on download
-            with tempfile.TemporaryDirectory() as temp_dir:
+            with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as temp_dir:
```
LefterisJP, line 4472: Delete the huge wall of text. This is a known issue

- https://www.scivision.dev/python-tempfile-permission-error-windows/
- https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

Just add a link on the right as a comment along with probably noqa:501 to the first link

LefterisJP: We already do this in a few places. But not in all the places where the `tempfile.TemporaryDirectory` is used. You can grep about it.

Do you think this would mean that we may end up getting this bug in production?

LefterisJP (approved):
LGTM. Let's see the tests

PR #7216, opened 2024-01-03
Bump dependencies

LefterisJP:


LefterisJP:
Merging. Only failed test is the one that started popping up today:

`rotkehlchen/tests/db/test_async.py::test_callback_segfault_simple`

and a coingecko rate limit

PR #7222, opened 2024-01-04
Make oracle penalty settings  user-modifiable at the backend

### LINKED ISSUE #5143: Make oracle penalty info user configurable
## Problem
https://github.com/rotki/rotki/pull/5133 introduced a mechanism that penalizes misbehaving oracle(s) in runtime.

However, the approach hardcodes the constants that determine which oracle(s) should be penalized.
This should not be the case, it should be user configurable.

## Specification
Instead of hardcoding it, move it to `DBSettings` so it can be updated/changed from time to time by the user.

programmer-ke:
Approach:
- Added settings `oracle_penalty_threshold_count` and `oracle_penalty_duration` to replace the constants that were being used by the `PenalizablePriceOracleMixin`.
- Made the settings updateable via the REST API.
- Made the settings accessible from `CachedSettings`.

I noticed that the previously `CachedSettings` was getting initialized by default settings, and updated with any settings updated in the logged in session. However this means that the next time the user logs in, cached settings would have default values but not any saved custom settings.

Changed this so that `CachedSettings` is updated with saved settings when the user is unlocked, so that previously customized settings are also cached.

Closes #5143 

## Checklist

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -611,6 +611,8 @@ Getting or modifying settings
               "ssf_graph_multiplier": 2,
               "non_sync_exchanges": [{"location": "binance", "name": "binance1"}],
               "cost_basis_method": "fifo",
+              "oracle_penalty_threshold_count": 5,
+              "oracle_penalty_duration": 1800,
               "address_name_priority": ["private_addressbook", "blockchain_account",
                                         "global_addressbook", "ethereum_tokens",
                                         "hardcoded_mappings", "ens_names"],
@@ -645,6 +647,8 @@ Getting or modifying settings
    :resjson int query_retry_limit: The number of times to retry a query to external services before giving up. Default is 5.
    :resjson int connect_timeout: The number of seconds to wait before giving up on establishing a connection to an external service. Default is 30.
    :resjson int read_timeout: The number of seconds to wait for the first byte after a connection to an external service has been established. Default is 30.
+   :resjson int oracle_penalty_threshold_count: The number of failures after which an oracle is penalized. Default is 5.
+   :resjson int oracle_penalty_duration: The duration in seconds for which an oracle is penalized. Default is 1800.
 
    :statuscode 200: Querying of settings was successful
    :statuscode 409: There is no logged in user
@@ -689,6 +693,8 @@ Getting or modifying settings
    :resjson int query_retry_limit: The number of times to retry a query to external services before giving up. Default is 5.
    :resjson int connect_timeout: The number of seconds to wait before giving up on establishing a connection to an external service. Default is 30.
    :resjson int read_timeout: The number of seconds to wait for the first byte after a connection to an external service has been established. Default is 30.
+   :resjson int oracle_penalty_threshold_count: The number of failures after which an oracle is penalized. Default is 5.
+   :resjson int oracle_penalty_duration: The duration in seconds for which an oracle is penalized. Default is 1800.
 
    **Example Response**:
 
--- a/rotkehlchen/api/v1/schemas.py
+++ b/rotkehlchen/api/v1/schemas.py
@@ -1261,6 +1261,20 @@ class ModifiableSettingsSchema(Schema):
         ),
         load_default=None,
     )
+    oracle_penalty_threshold_count = fields.Integer(
+        load_default=None,
+        validate=webargs.validate.Range(
+            min=1,
+            error='The count should be >= 1',
+        ),
+    )
+    oracle_penalty_duration = fields.Integer(
+        load_default=None,
+        validate=webargs.validate.Range(
+            min=1,
+            error='The penalty should be >= 1 seconds',
+        ),
+    )
 
     @validates_schema
     def validate_settings_schema(
@@ -1316,6 +1330,8 @@ class ModifiableSettingsSchema(Schema):
             query_retry_limit=data['query_retry_limit'],
             connect_timeout=data['connect_timeout'],
             read_timeout=data['read_timeout'],
+            oracle_penalty_threshold_count=data['oracle_penalty_threshold_count'],
+            oracle_penalty_duration=data['oracle_penalty_duration'],
         )
 
 
--- a/rotkehlchen/db/settings.py
+++ b/rotkehlchen/db/settings.py
@@ -55,6 +55,8 @@ DEFAULT_INFER_ZERO_TIMED_BALANCES = False  # If True the asset amount and value 
 DEFAULT_QUERY_RETRY_LIMIT = 5
 DEFAULT_CONNECT_TIMEOUT = 30
 DEFAULT_READ_TIMEOUT = 30
+DEFAULT_ORACLE_PENALTY_THRESHOLD_COUNT = 5
+DEFAULT_ORACLE_PENALTY_DURATION = 1800
 
 JSON_KEYS = (
     'current_price_oracles',
@@ -88,6 +90,8 @@ INTEGER_KEYS = (
     'query_retry_limit',
     'connect_timeout',
     'read_timeout',
+    'oracle_penalty_threshold_count',
+    'oracle_penalty_duration',
 )
 STRING_KEYS = (
     'ksm_rpc_endpoint',
@@ -135,6 +139,8 @@ CachedDBSettingsFieldNames = Literal[
     'query_retry_limit',
     'connect_timeout',
     'read_timeout',
+    'oracle_penalty_threshold_count',
+    'oracle_penalty_duration',
 ]
 
 DBSettingsFieldTypes = (
@@ -190,6 +196,8 @@ class DBSettings:
     query_retry_limit: int = DEFAULT_QUERY_RETRY_LIMIT
     connect_timeout: int = DEFAULT_CONNECT_TIMEOUT
     read_timeout: int = DEFAULT_READ_TIMEOUT
+    oracle_penalty_threshold_count: int = DEFAULT_ORACLE_PENALTY_THRESHOLD_COUNT
+    oracle_penalty_duration: int = DEFAULT_ORACLE_PENALTY_DURATION
 
     def serialize(self) -> dict[str, Any]:
         settings_dict = {}
@@ -243,6 +251,8 @@ class ModifiableDBSettings(NamedTuple):
     query_retry_limit: int | None = None
     connect_timeout: int | None = None
     read_timeout: int | None = None
+    oracle_penalty_threshold_count: int | None = None
+    oracle_penalty_duration: int | None = None
 
     def serialize(self) -> dict[str, Any]:
         settings_dict = {}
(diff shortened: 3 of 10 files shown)


rotkehlchen/api/v1/schemas.py
```diff
@@ -1261,6 +1261,20 @@ class ModifiableSettingsSchema(Schema):
...
+        validate=webargs.validate.Range(
+            min=1,
+            error='The count should be >= 1',
+        ),
+    )
+    oracle_penalty_duration = fields.Integer(
+        load_default=None,
+        validate=webargs.validate.Range(
+            min=1,
+            error='The penalty should be > 0 seconds',
```
yabirgb, line 1275: for consistency the message should use the same format in the string

```suggestion
            error='The count should be >= 1',
        ),
    )
    oracle_penalty_duration = fields.Integer(
        load_default=None,
        validate=webargs.validate.Range(
            min=1,
            error='The penalty should be >= 1 seconds',
```

rotkehlchen/rotkehlchen.py
```diff
@@ -310,7 +310,9 @@ def unlock_user(
         Inquirer()._manualcurrent.set_database(database=self.data.db)
 
         # Initialize the cached settings singleton
-        CachedSettings()
+        with self.data.db.conn.read_ctx() as cursor:
+            settings = self.get_settings(cursor)
+        CachedSettings().update_entries(settings)
```
yabirgb, line 315: instead of opening two cursors move the update of the cached settings to where the settings were queried before

rotkehlchen/tests/api/test_settings.py
```diff
@@ -35,7 +35,8 @@
 )
 
 
-def test_cached_settings(rotkehlchen_api_server, username):
+@pytest.mark.mock_rotki_settings(False)
```
yabirgb, line 38: This is something that is really only used here. Instead you can use the fixture that we have to set settings. An example of how to use it is

```
@pytest.mark.parametrize('db_settings', [
    {'pnl_csv_with_formulas': False},
])
```

and you can modify some of them and test them

rotkehlchen/tests/fixtures/rotkehlchen.py
```diff
@@ -484,14 +486,19 @@ def fixture_rotkehlchen_api_server(
...
     api_server = create_api_server(
         rotki=uninitialized_rotkehlchen,
         rest_port_number=rest_api_port,
     )
-
+    marker = request.node.get_closest_marker('mock_rotki_settings')
+    if marker is None:
+        mock_rotki_settings = True
+    else:
+        mock_rotki_settings = marker.args[0]
```
yabirgb, line 501: remove this logic

rotkehlchen/utils/mixins/penalizable_oracle.py
```diff
@@ -19,7 +17,8 @@ def note_failure_or_penalize(self) -> None:
         It is called when an oracle fails, if the failure count reaches a threshold,
         it is punished for a duration i.e. skipped. Otherwise, the failure count is incremented.
         """
-        if self.query_failures_count >= ORACLE_PENALTY_THRESHOLD_COUNT:
+        settings = CachedSettings().get_settings()
+        if self.query_failures_count >= settings.oracle_penalty_threshold_count:
```
yabirgb, line 21: instead of all the settings can't you just query the key?

```suggestion
        if self.query_failures_count >= CachedSettings().get_entry('oracle_penalty_threshold_count'):
```

rotkehlchen/tests/unit/test_inquirer.py
```diff
@@ -558,15 +558,16 @@ def test_punishing_of_oracles_works(inquirer):
                 assert defillama_mock.called is True
 
         # move the current time forward and check that coingecko is still penalized
+        penalty_duration = CachedSettings().get_entry('oracle_penalty_duration')
```
yabirgb, line 561: you did it here

rotkehlchen/rotkehlchen.py
```diff
@@ -310,7 +310,9 @@ def unlock_user(
         Inquirer()._manualcurrent.set_database(database=self.data.db)
 
         # Initialize the cached settings singleton
-        CachedSettings()
+        with self.data.db.conn.read_ctx() as cursor:
+            settings = self.get_settings(cursor)
```
yabirgb, line 314: You need to update the cachedsettings docstring since the behaviour has changed

rotkehlchen/db/settings.py
```diff
@@ -394,8 +404,15 @@ def reset(self) -> None:
...
-        for attr in settings._fields:
+    def update_entries(self, settings: ModifiableDBSettings | DBSettings) -> None:
+
+        setting_fields: tuple[str, ...] | list[str]
+        if isinstance(settings, ModifiableDBSettings):
+            setting_fields = settings._fields
+        else:
+            setting_fields = [field.name for field in fields(settings)]
+
+        for attr in setting_fields:
```
yabirgb, line 415: don't leave the empty line. Also why not make both a tuple?
```suggestion
    def update_entries(self, settings: ModifiableDBSettings | DBSettings) -> None:
        if isinstance(settings, ModifiableDBSettings):
            setting_fields = settings._fields
        else:
            setting_fields = tuple(field.name for field in fields(settings))

        for attr in setting_fields:
```

docs/api.rst
```diff
@@ -686,6 +690,8 @@ Getting or modifying settings
    :resjson int query_retry_limit: The number of times to retry a query to external services before giving up. Default is 5.
    :resjson int connect_timeout: The number of seconds to wait before giving up on establishing a connection to an external service. Default is 30.
    :resjson int read_timeout: The number of seconds to wait for the first byte after a connection to an external service has been established. Default is 30.
+   :resjson int oracle_penalty_threshold_count: The count of failures after which an oracle is penalized. Default is 5.
+   :resjson int oracle_penalty_duration: The duration in seconds for which an oracle is penalized. Default is 1800.
```
yabirgb, line 694:
```suggestion
   :resjson int oracle_penalty_threshold_count: The number of failures after which an oracle is penalized. Default is 5.
   :resjson int oracle_penalty_duration: The duration in seconds for which an oracle is penalized. Default is 1800.
```

docs/api.rst
```diff
@@ -642,6 +644,8 @@ Getting or modifying settings
...
```
yabirgb, line 648:
```suggestion
   :resjson int oracle_penalty_threshold_count: The number of failures after which an oracle is penalized. Default is 5.
   :resjson int oracle_penalty_duration: The duration in seconds for which an oracle is penalized. Default is 1800.
```

yabirgb (commented):
Thank you! SOme tests failed. Please check them

rotkehlchen/utils/mixins/penalizable_oracle.py
programmer-ke: Okay, the reason for this was that the type checker wasn't pleased with the use of `get_entry`, because it returns
a union of types yet the one I need is an int.

It would therefore flag the comparison operations as invalid as they don't work with the non-integer types. But I could be missing a way of making it work in case you have a suggestion. Another approach would be to add a property method for each setting I need.

rotkehlchen/tests/unit/test_inquirer.py
programmer-ke: I think the type checker doesn't check tests, so this wasn't flagged

rotkehlchen/utils/mixins/penalizable_oracle.py
yabirgb: As options you can:

- Create an overrade of the method so when the literal `oracle_penalty_threshold_count` is provided the return type is int.
- Use a `cast()`
- Use a `# type: ignore` explaining why

I would do the first or the last (probably the last and 1 in a smaller refactor) because the cast has a small overhead in production

rotkehlchen/tests/unit/test_inquirer.py
yabirgb: Yes, for non typed environments mypy doens't check on the tests

programmer-ke:
> Also @programmer-ke try to squash your commits. We keep 1 commit per change. So no smaller commits updating things that were added in the same PR

Cool, I can squash the commits before the finally merging to the base branch

rotkehlchen/tests/api/test_settings.py
programmer-ke: I took this approach, but ended up having to create a new fixture to control the relevant flag

programmer-ke:
ping @yabirgb, just a reminder

programmer-ke:
I see an api test failed. From the message, it looks like it could be a throttling issue on Coingecko

yabirgb:
> I see an api test failed. From the message, it looks like it could be a throttling issue on Coingecko

Yes don't worry! I've pinged @LefterisJP to do a final review since I don't have any more comments

rotkehlchen/db/settings.py
```diff
@@ -394,8 +408,14 @@ def reset(self) -> None:
     def update_entry(self, attr: str, value: DBSettingsFieldTypes) -> None:
         setattr(self._settings, attr, value)
 
-    def update_entries(self, settings: ModifiableDBSettings) -> None:
-        for attr in settings._fields:
+    def update_entries(self, settings: ModifiableDBSettings | DBSettings) -> None:
```
LefterisJP, line 411: Undo this change. This should only get ModifiableDBSettings.

rotkehlchen/rotkehlchen.py
```diff
@@ -336,6 +333,7 @@ def unlock_user(
 
         with self.data.db.conn.read_ctx() as cursor:
             settings = self.get_settings(cursor)
+            CachedSettings().update_entries(settings)  # update cached settings from DB
```
LefterisJP, line 336: oooof so it was never initialized? Damn ...

So the main thing here is the logic needs changing.

`update_entries` should only get a `ModifiableDBSettings` argument. So here you need to turn the returned settings to a ModifiableDBSettings or just more simply create an `initialize()` function which just does what `reset()` does but with an argument of `DBSettings`.

LefterisJP (commented):
Some comments

programmer-ke:
Hey @LefterisJP , I've pushed an update addressing your comments.

docs/api.rst
```diff
@@ -611,6 +611,8 @@ Getting or modifying settings
               "ssf_graph_multiplier": 2,
               "non_sync_exchanges": [{"location": "binance", "name": "binance1"}],
               "cost_basis_method": "fifo",
+          "oracle_penalty_threshold_count": 5,
```
LefterisJP, line 614: still not aligned

LefterisJP (commented):
All looks good. The small tiny alignment thing I will push myself since it seems CI needs approval again.

LefterisJP (approved):
LGTM. Failed test is due to bitmex sandbox changing their data and is being fixed in another PR.

docs/api.rst
programmer-ke: lol, I applied you patch. I thought it'd be fixed

programmer-ke:
Thanks for the POAP, first time I'm getting this. Interesting concept

PR #7225, opened 2024-01-04
Allow to set an asset as spam when ignoring it

### LINKED ISSUE #7124: Allow to mark assets as spam instead of just ignoring them
## Abstract

Marking an asset as spam works when the global db is shared between multiple users. At the moment we allow to ignore an asset easily but not to mark it as spam

## Specification

1. In the list of newly detected tokens we should allow to mark as spam instead of just ignoring it
2. In the asset view we should allow both: ignore and mark as spam

The endpoint for ignoring assets should take an additional argument to mark an asset as spam. The endpoint currently handling this is `/api/1/assets/ignored`

yabirgb:
This commit allows to send a `is_spam` field that will set the protocol to spam for evm tokens when ignoring them

Backend part of #7124

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -10011,9 +10011,14 @@ Dealing with ignored assets
       Host: localhost:5042
       Content-Type: application/json;charset=UTF-8
 
-      {"assets": ["eip155:1/erc20:0x6810e776880C02933D47DB1b9fc05908e5386b96"]}
+      {"assets": [{
+         "asset": "eip155:1/erc20:0x6810e776880C02933D47DB1b9fc05908e5386b96",
+         "is_spam": true
+      }]}
 
-   :reqjson list assets: A list of asset symbols to add to the ignored assets.
+   :reqjson list assets: A list of assets to be ignored
+   :reqjson string asset: The id of the asset to be ignored
+   :reqjson bool is_spam: True if the asset should also be marked as spam in addition to putting to the ignore list.
 
    **Example Response**:
 
@@ -10035,7 +10040,7 @@ Dealing with ignored assets
 
 .. http:delete:: /api/(version)/assets/ignored/
 
-   Doing a DELETE on the ignored assets endpoint will remove the given assets from the ignored assets list. Returns the new list without the removed assets in the response.
+   Doing a DELETE on the ignored assets endpoint will remove the given assets from the ignored assets list. It will also remove the assets from the spam list. Returns the new list without the removed assets in the response.
 
 
    **Example Request**:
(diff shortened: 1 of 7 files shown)


docs/api.rst
```diff
@@ -10011,9 +10011,14 @@ Dealing with ignored assets
...
-      {"assets": ["eip155:1/erc20:0x6810e776880C02933D47DB1b9fc05908e5386b96"]}
+      {"assets": [{
+         "asset": "eip155:1/erc20:0x6810e776880C02933D47DB1b9fc05908e5386b96",
+         "is_spam": true
+      }]}
 
-   :reqjson list assets: A list of asset symbols to add to the ignored assets.
+   :reqjson list assets: A list of assets to be ignored
+   :reqjson string asset: The id of the asset to be ignored
+   :reqjson bool is_spam: True if the asset should be marked as spam
```
LefterisJP, line 10021:
```suggestion
   :reqjson bool is_spam: True if the asset should also be marked as spam in addition to putting to the ignore list.
```

rotkehlchen/api/rest.py
```diff
@@ -1959,17 +1959,46 @@ def get_ignored_assets(self) -> Response:
             result = self.rotkehlchen.data.db.get_ignored_asset_ids(cursor)
         return api_response(_wrap_in_ok_result(list(result)), status_code=HTTPStatus.OK)
 
-    def add_ignored_assets(self, assets: list[Asset]) -> Response:
-        result, msg = self.rotkehlchen.data.add_ignored_assets(assets=assets)
+    def add_ignored_assets(self, assets_to_ignore: list[tuple[Asset, bool]]) -> Response:
```
LefterisJP, line 1962: add a docstring here. It's no longer self-evident since there is a boolean argument.

rotkehlchen/tests/api/test_assets.py
```diff
@@ -209,6 +220,12 @@ def test_ignored_assets_endpoint_errors(rotkehlchen_api_server, method):
     )
     assert_proper_response(response)
 
+    def prepare_request_body(assets: list[str]) -> list[dict[str, str]] | list[str]:
+        if method == 'put':
+            return [{'asset': asset} for asset in assets]
+        else:
+            return assets
```
LefterisJP, line 227:
```suggestion
        return [{'asset': asset} for asset in assets]  if method == 'put' else assets
```

LefterisJP (commented):
Some nitpicks.

Also do not write close issue. This is the backend part of the issue and needs frontend work and changelo.

LefterisJP (approved):
LGTM

PR #7233, opened 2024-01-05
fix(frontend): give detailed information after adding EVM account

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/composables/blockchain/index.ts
+++ b/frontend/app/src/composables/blockchain/index.ts
@@ -1,7 +1,7 @@
+/* eslint-disable max-lines */
 import { Blockchain } from '@rotki/common/lib/blockchain';
 import { Severity } from '@rotki/common/lib/messages';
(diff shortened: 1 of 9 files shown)


frontend/app/src/composables/blockchain/index.ts
```diff
@@ -110,32 +111,112 @@ export const useBlockchains = () => {
...
-            logger.error(`${chain} was not a valid blockchain`);
-            continue;
-          }
-          startPromise(finishAddition(chain, addresses[chain]));
+
+        const listOfFailureText = [];
+        if (noActivity) {
+          listOfFailureText.push(
+            getChainsText(
+              Object.values(noActivity)[0],
```
kelsos, line 174: why do we only take the first value here?

lukicenturi: because we separate the tasks, so 1 task for 1 address. The response will show only 1 key there

PR #7234, opened 2024-01-05
Upgrade to python 3.11

LefterisJP:
python311logo


LefterisJP:
Only `rotkehlchen/tests/unit/test_gnosis_inquirer.py::test_gnosis_nodes_prune_and_archive_status` failed and passes locally. Most probably a proxy behind the node changing between pruned and non pruned.

Pushing to build

LefterisJP:
As discussed in discord. Will merge this as is for now.

But what remains to be done is packaging for win32 and macos to make miniupnpc to work correctly. Kelsos is on it.

PR #7235, opened 2024-01-05
fix (#7176): add label for addresses in CSV export

### LINKED ISSUE #7176: CSV export should also contain human readable addresses from the users + global addressbook
## Abstract

The notes for transaction events of the PnL export are exported raw. In the UI all hexadecimal addresses are checked against addressbook and replaced if existing. But the CSV export does not do this.

It has been requested we do this as it is a value add for the user. cc @Buttaa 

## Task

- If the event in the CSV export has any addresses in the notes add the label next to it in brackets. For example `<address> [label]` This will require some text processing with regexes, similar to what the frontend does on top of the notes.

OjusWiZard:
Closes #7176

## Checklist

- [x] Match EVM address regex in notes of exported CSV and add a label after them if present.
- [x] Add test for this functionality

### DIFF
--- a/rotkehlchen/accounting/structures/processed_event.py
+++ b/rotkehlchen/accounting/structures/processed_event.py
@@ -1,5 +1,6 @@
 import builtins
 import json
+import re
 from collections.abc import Callable
 from dataclasses import dataclass, field
 from enum import Enum, auto
@@ -9,13 +10,16 @@ from rotkehlchen.accounting.cost_basis import CostBasisInfo
 from rotkehlchen.accounting.mixins.event import AccountingEventType
 from rotkehlchen.accounting.pnl import PNL
 from rotkehlchen.assets.asset import Asset
+from rotkehlchen.chain.evm.constants import EVM_ADDRESS_REGEX
+from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.constants import ZERO
+from rotkehlchen.db.dbhandler import DBHandler
 from rotkehlchen.errors.asset import UnknownAsset
 from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.fval import FVal
 from rotkehlchen.history.deserialization import deserialize_price
 from rotkehlchen.serialization.deserialize import deserialize_fval
-from rotkehlchen.types import Location, Price, Timestamp
+from rotkehlchen.types import EVM_LOCATIONS, ChainAddress, ChainID, Location, Price, Timestamp
 from rotkehlchen.utils.serialization import rlk_jsondumps
 
 T = TypeVar('T', bound='ProcessedAccountingEvent')
@@ -61,11 +65,25 @@ class ProcessedAccountingEvent:
 
         return desc
 
+    def _maybe_add_label_with_address(
+            self,
+            database: DBHandler,
+            matched_address: re.Match[str],
+    ) -> str:
+        """Aux method to enrich addresses in the event notes using the addressbook"""
+        chain_address = ChainAddress(
+            address=string_to_evm_address(matched_address.group()),
+            blockchain=ChainID(self.location.to_chain_id()).to_blockchain(),
+        )
+        name = database.get_blockchain_account_label(chain_address)
+        return f'{chain_address.address} [{name}]' if name else chain_address.address
+
     @overload
     def to_exported_dict(
             self,
             ts_converter: Callable[[Timestamp], str],
             export_type: Literal[AccountingEventExportType.CSV],
+            database: DBHandler,
             evm_explorer: str | None,
     ) -> dict[str, Any]:
         ...
@@ -82,6 +100,7 @@ class ProcessedAccountingEvent:
             self,
             ts_converter: Callable[[Timestamp], str],
             export_type: AccountingEventExportType,
+            database: DBHandler | None = None,
             evm_explorer: str | None = None,
     ) -> dict[str, Any]:
         """These are the fields that will appear in CSV, report API and are also exported to the
@@ -117,6 +136,15 @@ class ProcessedAccountingEvent:
                 exported_dict['asset'] = ''
             if tx_hash is not None:
                 exported_dict['notes'] = f'{evm_explorer}{tx_hash}  ->  {self.notes}'
+            if self.location in EVM_LOCATIONS and database is not None:
+                # call _maybe_add_label_with_address on each address in the note
+                exported_dict['notes'] = EVM_ADDRESS_REGEX.sub(
+                    repl=lambda matched_address: self._maybe_add_label_with_address(
+                        database=database,
+                        matched_address=matched_address,
+                    ),
+                    string=exported_dict['notes'],  # type: ignore [call-overload]  # exported_dict['notes'] is always a string
+                )
         else:  # for the other types of export we include the cost basis information
             cost_basis = None
             if self.cost_basis is not None:
--- a/rotkehlchen/chain/evm/constants.py
+++ b/rotkehlchen/chain/evm/constants.py
@@ -1,3 +1,4 @@
+import re
 from typing import Final
 
 from rotkehlchen.types import deserialize_evm_tx_hash
@@ -9,6 +10,7 @@ ZERO_ADDRESS = string_to_evm_address('0x0000000000000000000000000000000000000000
 ETH_SPECIAL_ADDRESS = string_to_evm_address('0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE')
 ZERO_32_BYTES_HEX = '0x' + '0' * 64
 GENESIS_HASH = deserialize_evm_tx_hash(ZERO_32_BYTES_HEX)  # hash for transactions in genesis block
+EVM_ADDRESS_REGEX = re.compile(r'\b0x[a-fA-F0-9]{40}\b')
 
 # Fake receipt with values taken from ethereum mainnet, to emulate a receipt for the
 # genesis transactions
--- a/rotkehlchen/tests/api/test_pnl_csv.py
+++ b/rotkehlchen/tests/api/test_pnl_csv.py
@@ -16,7 +16,9 @@ from rotkehlchen.accounting.export.csv import FILENAME_ALL_CSV
 from rotkehlchen.accounting.mixins.event import AccountingEventType
 from rotkehlchen.accounting.structures.balance import Balance
 from rotkehlchen.assets.asset import CustomAsset
+from rotkehlchen.chain.accounts import BlockchainAccountData
 from rotkehlchen.chain.evm.constants import GENESIS_HASH
+from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.constants.assets import A_ETH
 from rotkehlchen.db.custom_assets import DBCustomAssets
 from rotkehlchen.fval import FVal
@@ -29,7 +31,7 @@ from rotkehlchen.tests.utils.constants import ETH_ADDRESS1, ETH_ADDRESS2, ETH_AD
 from rotkehlchen.tests.utils.factories import make_evm_tx_hash
 from rotkehlchen.tests.utils.history import prepare_rotki_for_history_processing_test, prices
 from rotkehlchen.tests.utils.pnl_report import query_api_create_and_get_report
-from rotkehlchen.types import Location, Timestamp, TimestampMS
+from rotkehlchen.types import Location, SupportedBlockchain, Timestamp, TimestampMS
 from rotkehlchen.utils.misc import create_timestamp
 
 
@@ -82,6 +84,7 @@ def assert_csv_export_response(response, csv_dir, is_download=False, expected_nu
             rows.append(row)
 
     assert count == expected_num_of_events
+    return rows
 
 
 @pytest.mark.parametrize('have_decoders', [True])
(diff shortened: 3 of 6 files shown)


rotkehlchen/accounting/structures/processed_event.py
```diff
@@ -75,13 +79,15 @@ def to_exported_dict(
             self,
             ts_converter: Callable[[Timestamp], str],
             export_type: Literal[AccountingEventExportType.API, AccountingEventExportType.DB],
+            database: None,
```
yabirgb, line 82: can't you remove it from here and give a default value in the main definition?

```python
database: DBHandler | None = None

File: rotkehlchen/accounting/structures/processed_event.py [hunk 2]
```diff
@@ -117,6 +123,21 @@ def to_exported_dict(
...
+            if self.location in EVM_LOCATIONS and database is not None:
+                def _maybe_add_label_with_address(matched_address: re.Match[str]) -> str:
+                    """Adds a label of the address if it exists in a given string"""
+                    chain_address = ChainAddress(
+                        address=string_to_evm_address(matched_address.group(0)),
+                        blockchain=ChainID(self.location.to_chain_id()).to_blockchain(),
+                    )
+                    name = database.get_blockchain_account_label(chain_address)
+                    return f'{chain_address.address} [{name}]' if name else chain_address.address
+                exported_dict['notes'] = re.sub(
```
yabirgb, line 135:
```suggestion
                    return f'{chain_address.address} [{name}]' if name else chain_address.address

                exported_dict['notes'] = re.sub(
```

rotkehlchen/accounting/structures/processed_event.py
```diff
@@ -117,6 +123,21 @@ def to_exported_dict(
                 exported_dict['asset'] = ''
             if tx_hash is not None:
                 exported_dict['notes'] = f'{evm_explorer}{tx_hash}  ->  {self.notes}'
+            if self.location in EVM_LOCATIONS and database is not None:
+                def _maybe_add_label_with_address(matched_address: re.Match[str]) -> str:
+                    """Adds a label of the address if it exists in a given string"""
+                    chain_address = ChainAddress(
+                        address=string_to_evm_address(matched_address.group(0)),
```
yabirgb, line 130: group(0) can raise an error if for some reason the address doesn't appear right? Can you check that? also I believe you need to check if it is none or not before creating the chainAddress?

rotkehlchen/accounting/structures/processed_event.py
```diff
@@ -117,6 +123,21 @@ def to_exported_dict(
...
+                        blockchain=ChainID(self.location.to_chain_id()).to_blockchain(),
+                    )
+                    name = database.get_blockchain_account_label(chain_address)
+                    return f'{chain_address.address} [{name}]' if name else chain_address.address
+                exported_dict['notes'] = re.sub(
+                    # call _maybe_add_label_with_address on each address in the note
+                    pattern=r'\b0x[a-fA-F0-9]{40}\b',
+                    repl=_maybe_add_label_with_address,
+                    string=str(exported_dict['notes']),
```
yabirgb, line 139: isn't this already a string? why the cast to string?

rotkehlchen/accounting/structures/processed_event.py
```diff
@@ -135,7 +156,7 @@ def to_exported_dict(
 
     def serialize_to_dict(self, ts_converter: Callable[[Timestamp], str]) -> dict[str, Any]:
         """This is used to serialize to dict for saving to the DB"""
-        data = self.to_exported_dict(ts_converter=ts_converter, export_type=AccountingEventExportType.DB)  # noqa: E501
+        data = self.to_exported_dict(ts_converter=ts_converter, export_type=AccountingEventExportType.DB, database=None)  # noqa: E501
```
yabirgb, line 159: I believe you can remove it with the thing I mentioned before

rotkehlchen/tests/api/test_pnl_csv.py
```diff
@@ -184,13 +187,30 @@ def test_history_export_download_csv(
...
+    # query it again and check if known addresses are replace by their labels in the exported notes
+    with rotkehlchen_api_server_with_exchanges.rest_api.rotkehlchen.data.db.user_write() as cursor:
+        rotkehlchen_api_server_with_exchanges.rest_api.rotkehlchen.data.db.edit_blockchain_accounts(
+            write_cursor=cursor,
+            account_data=[BlockchainAccountData(
+                chain=SupportedBlockchain.ETHEREUM,
+                address=string_to_evm_address(ETH_ADDRESS1),
+                label='ETH_ADDRESS1',
+            )],
+        )
```
yabirgb, line 207: can't you do it in the first query? to avoid having to read it again

rotkehlchen/accounting/structures/processed_event.py
```diff
@@ -117,6 +123,21 @@ def to_exported_dict(
...
+                    """Adds a label of the address if it exists in a given string"""
+                    chain_address = ChainAddress(
+                        address=string_to_evm_address(matched_address.group(0)),
+                        blockchain=ChainID(self.location.to_chain_id()).to_blockchain(),
+                    )
+                    name = database.get_blockchain_account_label(chain_address)
+                    return f'{chain_address.address} [{name}]' if name else chain_address.address
+                exported_dict['notes'] = re.sub(
+                    # call _maybe_add_label_with_address on each address in the note
+                    pattern=r'\b0x[a-fA-F0-9]{40}\b',
```
yabirgb, line 137: is the `\b` needed in the regex? Also maybe makes sense to move it as a constant at have it like `EVM_ADDRESS_REGEX`. You don't need to create the string each time

rotkehlchen/accounting/structures/processed_event.py
```diff
@@ -117,6 +123,21 @@ def to_exported_dict(
                 exported_dict['asset'] = ''
             if tx_hash is not None:
                 exported_dict['notes'] = f'{evm_explorer}{tx_hash}  ->  {self.notes}'
+            if self.location in EVM_LOCATIONS and database is not None:
+                def _maybe_add_label_with_address(matched_address: re.Match[str]) -> str:
+                    """Adds a label of the address if it exists in a given string"""
```
yabirgb, line 128: insted of defining it here each time maybe define it in the file as a private method. Otherwise you are creating the function each time

yabirgb (commented):
If it closses the issue I believe it needs a changelog entry

rotkehlchen/accounting/structures/processed_event.py
OjusWiZard: I tested with the cases where nothing matches with the regex or multiple matches are found. It works in both cases because the function `_maybe_add_label_with_address` is called only when something is matched, so inside the function, it always returns the matched string. Also, the default argument of `.group()` is 0, so removing that.

rotkehlchen/accounting/structures/processed_event.py
OjusWiZard: Initially, I did it without `/b` but then it also captured partial transaction hashes and Etherscan URLs. So, I had to specify that it should be a separate word.

rotkehlchen/accounting/structures/processed_event.py
OjusWiZard: mypy doesn't like it because, `exported_dict` was of type `dict[str, Any]`

LefterisJP: we never make more work on the production code to satisfy typing. You should fix typing instead str() on a string is wasting resources.

rotkehlchen/accounting/structures/processed_event.py
```diff
@@ -61,11 +65,25 @@ def to_string(self, ts_converter: Callable[[Timestamp], str]) -> str:
 
         return desc
 
+    def _maybe_add_label_with_address(
+            self,
+            database: DBHandler,
+            matched_address: re.Match[str],
+    ) -> str:
+        """Adds a label of the address if it exists in a given string"""
+        chain_address = ChainAddress(
```
yabirgb, line 74: THe docsstirng is not correct

```suggestion
        """Aux method to enrich addresses in the event notes using the addressbook""
        chain_address = ChainAddress(
```

yabirgb (commented):
Just one small comment

rotkehlchen/accounting/structures/processed_event.py
```diff
@@ -117,6 +136,16 @@ def to_exported_dict(
                 exported_dict['asset'] = ''
             if tx_hash is not None:
                 exported_dict['notes'] = f'{evm_explorer}{tx_hash}  ->  {self.notes}'
+            if self.location in EVM_LOCATIONS and database is not None:
+                # call _maybe_add_label_with_address on each address in the note
+                exported_dict['notes'] = re.sub(
+                    pattern=EVM_ADDRESS_REGEX,
```
LefterisJP, line 142: Instead of using this kind of regex matching, since the regex is constant compile the expression and use it: https://docs.python.org/3/library/re.html#re.compile. Since you use it only here you can probably compile it only in this module or well .. wherever you want.

rotkehlchen/chain/evm/constants.py
```diff
@@ -9,6 +9,7 @@
 ETH_SPECIAL_ADDRESS = string_to_evm_address('0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE')
 ZERO_32_BYTES_HEX = '0x' + '0' * 64
 GENESIS_HASH = deserialize_evm_tx_hash(ZERO_32_BYTES_HEX)  # hash for transactions in genesis block
+EVM_ADDRESS_REGEX = r'\b0x[a-fA-F0-9]{40}\b'
```
LefterisJP, line 12: Can you also add some unit tests for matching (and not matching) of this pattern?

rotkehlchen/accounting/structures/processed_event.py
```diff
@@ -117,6 +136,16 @@ def to_exported_dict(
                 exported_dict['asset'] = ''
             if tx_hash is not None:
                 exported_dict['notes'] = f'{evm_explorer}{tx_hash}  ->  {self.notes}'
+            if self.location in EVM_LOCATIONS and database is not None:
+                # call _maybe_add_label_with_address on each address in the note
+                exported_dict['notes'] = re.sub(
+                    pattern=re.compile(EVM_ADDRESS_REGEX),
```
LefterisJP, line 142: not what I asked. Check discord.

rotkehlchen/tests/unit/test_evm_misc.py
```diff
@@ -22,3 +24,20 @@ def test_asset_id_is_evm_token():
...
+        'https://etherscan.io/tx/0x9b675024d8648c3b590eff411fcf75a1199d10d1a3fe2ddbe50e166ce8b87cc9': False,  # etherscan link of a transaction  # noqa: E501
+        '0x0F5D2fB29fb7d3CFeE444a200298f468908cC942a': False,  # invalid address length
+        '0x0F5D2fB29fb7d3CFeE444a200298f468908zC942': False,  # invalid address characters
+        '0x0F5D2fB29fb7d3CFeE444a200298f468908cC942 Note with address at the start': True,
+        'Note with address in the end 0x0F5D2fB29fb7d3CFeE444a200298f468908cC942': True,
+        'A full stop after the address 0x0F5D2fB29fb7d3CFeE444a200298f468908cC942.': True,
+    }
+
+    for case, expected_result in cases.items():
+        assert (re.search(re.compile(EVM_ADDRESS_REGEX), case) is not None) == expected_result
```
LefterisJP, line 43: same here ... no no no no

rotkehlchen/accounting/structures/processed_event.py
LefterisJP, line 142: You are not using the compiled regex. Please when I say read the docs and look at the rest of the codebase do so

```suggestion
                exported_dict['notes'] = EVM_ADDRESS_REGEX.sub(
```

rotkehlchen/tests/unit/test_evm_misc.py
```diff
@@ -22,3 +24,20 @@ def test_asset_id_is_evm_token():
...
+        'https://etherscan.io/tx/0x9b675024d8648c3b590eff411fcf75a1199d10d1a3fe2ddbe50e166ce8b87cc9': False,  # etherscan link of a transaction  # noqa: E501
+        '0x0F5D2fB29fb7d3CFeE444a200298f468908cC942a': False,  # invalid address length
+        '0x0F5D2fB29fb7d3CFeE444a200298f468908zC942': False,  # invalid address characters
+        '0x0F5D2fB29fb7d3CFeE444a200298f468908cC942 Note with address at the start': True,
+        'Note with address in the end 0x0F5D2fB29fb7d3CFeE444a200298f468908cC942': True,
+        'A full stop after the address 0x0F5D2fB29fb7d3CFeE444a200298f468908cC942.': True,
+    }
+
+    for case, expected_result in cases.items():
+        assert (re.search(EVM_ADDRESS_REGEX, case) is not None) == expected_result
```
LefterisJP, line 43: dame here

```suggestion
        assert (EVM_ADDRESS_REGEX.search(case) is not None) == expected_result
```

LefterisJP (commented):
still not okay

LefterisJP (approved):
LGTM. All failed tests are due to etherscan unreachability

PR #7238, opened 2024-01-06
Upgrade gql and other dependencies

LefterisJP:


LefterisJP:
Both test failures are unrelated.

One for coingecko rate limit and other etherscan hiccup.

PR #7251, opened 2024-01-09
Add entry type mappings

yabirgb:
I didn't use directly `EthWithdrawalEvent.entry_type` because `entry_type` is an instance method and making `entry_type` a classmethod disallows from using it as a property. Didn't want to set it as a class attrribute because is something that we need to change and don't want someone from forgetting of changing it.

Created a new file for this mappings since otherwise we get a import cycle due the usage of `HistoryBaseEntryType` (is the same error if using `EthWithdrawalEvent`)

### DIFF
--- a/rotkehlchen/history/events/structures/types.py
+++ b/rotkehlchen/history/events/structures/types.py
@@ -152,8 +152,9 @@ class EventCategory(Enum):
     UPDATE_PROJECT = 33, EventDirection.NEUTRAL
     APPLY = 34, EventDirection.NEUTRAL
     STAKE_DEPOSIT = 35, EventDirection.OUT
-    STAKE_WITHDRAWAL = 36, EventDirection.IN
+    UNSTAKE = 36, EventDirection.IN
     ATTEST = 37, EventDirection.NEUTRAL
+    STAKE_EXIT = 38, EventDirection.IN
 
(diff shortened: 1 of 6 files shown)


rotkehlchen/history/events/structures/types.py
```diff
@@ -154,6 +154,7 @@ class EventCategory(Enum):
     STAKE_DEPOSIT = 35, EventDirection.OUT
     STAKE_WITHDRAWAL = 36, EventDirection.IN
     ATTEST = 37, EventDirection.NEUTRAL
+    STAKING_EXIT = 38, EventDirection.IN
```
LefterisJP, line 157:
```suggestion
    STAKE_EXIT = 38, EventDirection.IN
```

rotkehlchen/tests/api/test_misc.py
```diff
@@ -442,6 +442,16 @@ def test_events_mappings(rotkehlchen_api_server_with_exchanges):
     )
     result = assert_proper_response_with_result(response)
     assert 'global_mappings' in result
+    assert result['type_mappings'] == {
```
LefterisJP, line 445: this fails in tests

LefterisJP (approved):
LGTM

PR #7259, opened 2024-01-10
feat(frontend): mark asset as spam asset

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/asset-manager/managed/ManagedAssetIgnoringMore.vue
+++ b/frontend/app/src/components/asset-manager/managed/ManagedAssetIgnoringMore.vue
@@ -1,10 +1,17 @@
 <script setup lang="ts">
-const props = defineProps<{
-  identifier: string;
-}>();
+const props = withDefaults(
+  defineProps<{
+    identifier: string;
+    isSpam?: boolean;
+  }>(),
+  {
+    isSpam: false
+  }
+);
 
 const emit = defineEmits<{
   (e: 'toggle-whitelist'): void;
+  (e: 'toggle-spam'): void;
 }>();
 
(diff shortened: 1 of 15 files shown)


frontend/app/src/composables/assets/spam.ts
```diff
@@ -0,0 +1,57 @@
...
+      await markAssetAsSpamCaller(token);
+      await fetchWhitelistedAssets();
+      await fetchIgnoredAssets();
+      return { success: true };
+    } catch (e: any) {
+      notify({
+        title: t('ignore.spam.failed.mark_title').toString(),
+        message: t('ignore.spam.failed.mark_message', {
+          message: e.message
+        }).toString(),
```
kelsos, line 26: The `.toString()` call here is redundant we don't need it anymore since a while now.

So we can skip it.

frontend/app/src/components/asset-manager/managed/ManagedAssetIgnoringMore.vue
```diff
@@ -29,26 +40,32 @@ const toggleWhitelist = () => {
...
-          v-if="!isWhitelisted"
-          variant="list"
-          @click="toggleWhitelist()"
-        >
+        <RuiButton variant="list" @click="toggleWhitelist()">
           <template #prepend>
-            <RuiIcon name="checkbox-circle-line" />
+            <RuiCheckbox
+              color="primary"
+              class="[&_span]:-mr-1 [&_span]:!py-0"
```
kelsos, line 52: Isn't this something that we should do as part of the component itself instead of overriding the css outside?

PR #7262, opened 2024-01-10
Properly detect LUSD borrowing fee and stability pool frontend fee

LefterisJP:
Fix #7526


LefterisJP:
all green. Will merge.

PR #7268, opened 2024-01-11
fix (Fval): large precision

OjusWiZard:
## Checklist

- [x] Support precise `FVal` arithmetic operations for very large numbers
- [ ] Fix the affected tests


rotkehlchen/fval.py
```diff
@@ -22,6 +22,9 @@ class FVal:
 
     def __init__(self, data: AcceptableFValInitInput = 0):
 
+        decimal_context = getcontext()
+        decimal_context.prec = 78
+        setcontext(decimal_context)
```
yabirgb, line 27: Initializing a `FVal` is something that happens a lot and I don't know how expensive this operation is but it adds an overhead. I'm reading in the docs that this takes effect per thread so theorically you only need to execute it once when starting the app

rotkehlchen/fval.py
```diff
@@ -22,6 +22,9 @@ class FVal:
 
     def __init__(self, data: AcceptableFValInitInput = 0):
 
+        decimal_context = getcontext()
+        decimal_context.prec = 78
```
yabirgb, line 26: Why take 78? where does this value come from?

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` Tokens with large amount or decimal value are calculated with full precision.
```
yabirgb, line 5: probably is better to not add this in the docstring. The working has to change because as a user I don't know what it means

LefterisJP: yes this should not have a changelog

rotkehlchen/fval.py
```diff
@@ -7,6 +8,9 @@
 AcceptableFValInitInput = Union[float, bytes, Decimal, int, str, 'FVal']
 AcceptableFValOtherInput = Union[int, 'FVal']
 
+DefaultContext.prec = ceil(log10(2 ** 256))  # support upto uint256 max value
```
LefterisJP, line 11:
```suggestion
DefaultContext.prec = ceil(log10(2 ** 256))  # support up to uint256 max value
```

LefterisJP (approved):
just a typo. Otherwise lgtm.

PR #7282, opened 2024-01-13
Fix gnosis chain bridging DAI from ethereum edge case

LefterisJP:
There is some events (newer ones) that seem to no longer emit the USER_REQUESTED_FOR_AFFIRMATION_EVENT so there we need to just check if it's a DAI transfer to the bridge and treat it as bridging.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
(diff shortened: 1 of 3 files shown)


LefterisJP:
Only 2 failed test are due to remote timeouts and unrelated.

PR #7283, opened 2024-01-14
Periodic task to query monerium API and enrich transactions

### LINKED ISSUE #7027: Integrate monerium
## Abstract

Monerium is a way with which to associate an IBAN with an ethereum address (privately) and be able to make payments in the SEPA system. Their stablecoin EURe, is 1-1 exchangeable with eur. Receiving EUR to your IBAN mints EURe to your address, and sending EUR to the SEPA system burns them.

The minting/burning system can now also be used to transfer between the chains that EURe exists.

At the moment rotki sees all such transactions as transfers from/to 0x0.00 address. Since it has no other information.

By utilizing monerium's API we can augment the info for the user and actually show the SEPA transfers and all relevant info.

## Specification

- The monerium API at the moment is not meant to be consumed by local apps, but we are in talks with them about this, so they can make a good old plain API key.

 - Also they don't have permissions yet, so the access key can also transfer funds or associate IBAN with address which is not something rotki wants to touch (yet). We will want to when active management is implemented, but for now not. So having permissions that can be set to read only would be good.
 
 Until api keys are implemented we can use basic authentication: https://monerium.dev/docs/api#auth-client-credentials

At least for development purposes and see how to proceed from there.

LefterisJP:
Fix #7027

### DIFF
--- a/rotkehlchen/db/dbhandler.py
+++ b/rotkehlchen/db/dbhandler.py
@@ -1667,6 +1667,10 @@ class DBHandler:
                     f'the app with an older version. {e!s}',
                 )
                 continue
+
+            if location not in SUPPORTED_EXCHANGES:
+                continue
+
             credentials[location].append(ExchangeApiCredentials(
                 name=entry[0],
                 location=location,
--- a/rotkehlchen/tasks/assets.py
+++ b/rotkehlchen/tasks/assets.py
@@ -234,3 +234,9 @@ def update_owned_assets(user_db: DBHandler) -> None:
     """Wrapper to be used in async task to update owned assets"""
     with user_db.conn.read_ctx() as cursor:
         user_db.update_owned_assets_in_globaldb(cursor=cursor)
+
+    with user_db.conn.write_ctx() as write_cursor:
+        write_cursor.execute(  # remember last task ran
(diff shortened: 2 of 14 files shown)


rotkehlchen/db/dbhandler.py
```diff
@@ -1667,6 +1667,10 @@ def get_exchange_credentials(
                     f'the app with an older version. {e!s}',
                 )
                 continue
+
+            if location not in SUPPORTED_EXCHANGES:
```
LefterisJP, line 1671: This is because we may add user credentials (like with monerium) that are not exchanges. It's already happening above with the rotki premium credentials (but they got no location, so it defaults to external).

rotkehlchen/tasks/assets.py
```diff
@@ -234,3 +234,9 @@ def update_owned_assets(user_db: DBHandler) -> None:
     """Wrapper to be used in async task to update owned assets"""
     with user_db.conn.read_ctx() as cursor:
         user_db.update_owned_assets_in_globaldb(cursor=cursor)
+
+    with user_db.conn.write_ctx() as write_cursor:
```
LefterisJP, line 238: Added this since I did not see it being written anywhere

LefterisJP:
Will merge since now it's all green and this fixes develop tests too.

Remaining is an API to set the monerium credentials.

PR #7297, opened 2024-01-17
fix: improve notification for missing Etherscan API key

lukicenturi:
Closes #(issue_number)

## Checklist


kelsos (changes requested):
So code and changes look fine, what we need is to make sure that the notification does not get dismissed when you press the external link, as per discord and we should be good.

kelsos (approved):
lgtm thank you

PR #7310, opened 2024-01-18
Feat (paraswap): curve, balancer, and uniswapV2_buy support

OjusWiZard:
## Checklist

- [x] Add support for Curve v1 and v2 swaps, Balancer v2 swap, and Uniswap v2 buy in Paraswap decoder
- [x] Add the corresponding tests

### DIFF
--- a/rotkehlchen/chain/evm/decoding/paraswap/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/paraswap/decoder.py
@@ -4,9 +4,7 @@ from typing import TYPE_CHECKING, Any
 
 from rotkehlchen.accounting.structures.balance import Balance
 from rotkehlchen.chain.ethereum.modules.paraswap.constants import PARASWAP_AUGUSTUS_ROUTER
-from rotkehlchen.chain.ethereum.modules.uniswap.v2.constants import (
-    SWAP_SIGNATURE as UNISWAP_V2_SWAP_SIGNATURE,
-)
+from rotkehlchen.chain.ethereum.modules.uniswap.v2.constants import SWAP_SIGNATURE
 from rotkehlchen.chain.ethereum.modules.uniswap.v3.constants import DIRECT_SWAP_SIGNATURE
 from rotkehlchen.chain.ethereum.utils import asset_normalized_value
 from rotkehlchen.chain.evm.decoding.constants import ERC20_OR_ERC721_TRANSFER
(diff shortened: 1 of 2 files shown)


rotkehlchen/tests/unit/decoders/test_paraswap.py
```diff
@@ -22,6 +22,11 @@
 A_TKST = Asset('eip155:1/erc20:0x7CdBfC86A0BFa20F133748B0CF5cEa5b787b182c')
 A_PRISMA = Asset('eip155:1/erc20:0xdA47862a83dac0c112BA89c6abC2159b95afd71C')
 A_UDOODLE = Asset('eip155:1/erc20:0xa07dCC1aBFe20D29D87a32E2bA89876145DAfB0a')
+A_KP3R = Asset('eip155:1/erc20:0x1cEB5cB57C4D4E2b2433641b95Dd330A33185A44')
+A_AUCTION = Asset('eip155:1/erc20:0xA9B1Eb5908CfC3cdf91F9B8B3a74108598009096')
+A_PSP = Asset('eip155:1/erc20:0xcAfE001067cDEF266AfB7Eb5A286dCFD277f3dE5')
```
yabirgb, line 27: This one you can import it from the constants

rotkehlchen/tests/unit/decoders/test_paraswap.py
```diff
@@ -22,6 +22,11 @@
 A_TKST = Asset('eip155:1/erc20:0x7CdBfC86A0BFa20F133748B0CF5cEa5b787b182c')
 A_PRISMA = Asset('eip155:1/erc20:0xdA47862a83dac0c112BA89c6abC2159b95afd71C')
 A_UDOODLE = Asset('eip155:1/erc20:0xa07dCC1aBFe20D29D87a32E2bA89876145DAfB0a')
+A_KP3R = Asset('eip155:1/erc20:0x1cEB5cB57C4D4E2b2433641b95Dd330A33185A44')
```
yabirgb, line 25: this one too

rotkehlchen/chain/evm/decoding/paraswap/decoder.py
```diff
@@ -101,6 +103,8 @@ def _decode_swap(
             return DEFAULT_DECODING_OUTPUT
 
         if partial_refund_event is not None:  # if some in_asset is returned back
+            if in_event.asset == out_event.asset:  # in_event is the partial refund instead of partial_refund_event  # noqa: E501
+                partial_refund_event, in_event = in_event, partial_refund_event
```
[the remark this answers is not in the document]
OjusWiZard: to determine the `partial_refund_event` we need to check if its asset is the same as `out_event`'s asset. but in the above for loop the `out_event` can be None. We can suppress the linter if we're sure that the `in_event` is always captured after the `out_event`, but I'm not sure if that's always the order.

the cases I've seen is `out > in > refund` and `out > refund > in` 
but if, we get `out_event` in between or in the last, then in the for loop we will reach that condition without `out_event`, so we can't check the asset similarity

rotkehlchen/chain/evm/decoding/paraswap/decoder.py
```diff
@@ -101,6 +99,9 @@ def _decode_swap(
             return DEFAULT_DECODING_OUTPUT
 
         if partial_refund_event is not None:  # if some in_asset is returned back
+            if in_event.asset == out_event.asset:  # check the similarity of asset with out_event
+                # due to order of events it can happen that this two get mislabeled and need to be swapped  # noqa: E501
```
LefterisJP, line 103: Can you also maybe describe in a tiny bit more detail like you did to @yabirgb 's question in the comment here? A potential reader would need to understand when something may happen and why a check is in place.

```suggestion
                # due to order of events it can happen that these two get mislabeled and need to be swapped  # noqa: E501
```

LefterisJP (approved):
LGTM. For the comment I left then let's do it in your next PRs.

PR #7316, opened 2024-01-19
Remove covalent external service

yabirgb:
- Make AVAX balance queries using the avax RPC node
- Remove covalent code
- Upgrade step to remove covalent from the external services

db diff:

```
(rotki-main) ➜  rotki git:(develop) ✗ python diff_encrypted_db.py rotkehlchen/tests/data/v40_rotkehlchen.db v40_rotkehlchen.db 
/var/folders/zn/n9fpvbnj3fn971b5s466fzd40000gn/T/tmp8b1ofg0q /var/folders/zn/n9fpvbnj3fn971b5s466fzd40000gn/T/tmpohsj562m
INSERT INTO external_service_credentials(rowid,name,api_key) VALUES(3,'covalent','lollol');
```


rotkehlchen/chain/avalanche/manager.py
```diff
@@ -50,35 +45,20 @@ def get_latest_block_number(self) -> int:
...
-            balance = ZERO
-            for entry in result:
-                if entry.get('contract_ticker_symbol') == 'AVAX':
-                    balance = from_wei(FVal(entry.get('balance', 0)))
-                    break
-            if balance == ZERO:
-                balance = from_wei(FVal(self.w3.eth.get_balance(account)))
-        return FVal(balance)
+        except Web3Exception as e:
+            log.error(f'Failed to connect to AVAX rpc node. {e}')
```
LefterisJP, line 53: no need to catch this error here, right? Perhaps just say that this can raise Web3Exception and then handle it in the only place where this function is called.

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2035,7 +2035,7 @@ def test_upgrade_db_40_to_41(user_data_dir):
             name VARCHAR[30] NOT NULL PRIMARY KEY,
             api_key TEXT
             )""") is True
-        assert cursor.execute('SELECT * FROM external_service_credentials').fetchall() == [('etherscan', 'LOL'), ('blockscout', 'LOL2')]  # noqa: E501
+        assert cursor.execute('SELECT * FROM external_service_credentials').fetchall() == [('etherscan', 'LOL'), ('blockscout', 'LOL2'), ('covalent', 'lollol')]  # noqa: E501
```
LefterisJP, line 2038: you also need to assert it's no longer there after the upgrade

LefterisJP (commented):
So where `get_avax_balance` is called we also need changes.

```python
                    try:
                        # just check balance and nonce in avalanche
                        has_activity = (
                            avax_manager.w3.eth.get_transaction_count(address) != 0 or
                            avax_manager.get_avax_balance(address) != ZERO
                        )
                    except (requests.exceptions.RequestException, RemoteError) as e:
                        log.error(f'Failed to check {address} activity in avalanche due to {e!s}')
                        failed_to_query_chains.append(chain)
                        has_activity = False
```

Change to

```python
                    try:
                        # just check balance and nonce in avalanche
                        has_activity = (
                            avax_manager.w3.eth.get_transaction_count(address) != 0 or
                            avax_manager.get_avax_balance(address) != ZERO
                        )
                    except (requests.exceptions.RequestException, Web3Exception) as e:
                        log.error(f'Failed to check {address} activity in avalanche due to {e!s}')
                        failed_to_query_chains.append(chain)
                        has_activity = False
```

or something like that right?

rotkehlchen/tests/db/test_db_upgrades.py
yabirgb: It happens here https://github.com/rotki/rotki/pull/7316/files#diff-dc14676416c357a3a363ef102d35370ee0bd2ebc98976e6509f3e4805a12e87fR2088

imagen

LefterisJP: ah well lol ... yeah you are right. Thanks for the screenshot cause link does not work in PR reviews

LefterisJP (approved):
LGTM

PR #7333, opened 2024-01-22
Fix decoding of WETH in uniswap v3

yabirgb:
As it was before if a user deposited WETH instead of ETH the decoder was not detecting correctly the deposit event

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
@@ -37,7 +37,6 @@ from ..constants import CPT_UNISWAP_V2, CPT_UNISWAP_V3, UNISWAP_ICON, UNISWAP_LA
 
 if TYPE_CHECKING:
     from rotkehlchen.accounting.structures.evm_event import EvmEvent
-    from rotkehlchen.assets.asset import CryptoAsset
     from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
     from rotkehlchen.chain.evm.decoding.base import BaseDecoderTools
     from rotkehlchen.user_messages import MessagesAggregator
@@ -60,7 +59,7 @@ UNISWAP_ROUTERS = {UNISWAP_AUTO_ROUTER_V1, UNISWAP_AUTO_ROUTER_V2, UNISWAP_UNIVE
 
 class CryptoAssetAmount(NamedTuple):
     """This is used to represent a pair of resolved crypto asset to an amount."""
-    asset: 'CryptoAsset'
+    asset: 'EvmToken'
     amount: FVal
 
 
@@ -87,6 +86,20 @@ def _find_from_asset_and_amount(events: list['EvmEvent']) -> tuple[Asset, FVal] 
     if from_asset is None:
         return None
     return from_asset, from_amount
+
+
+def _compare_with_maybe_eth_token(
+        event_asset: 'Asset',
+        pool_token: 'EvmToken',
+) -> tuple[bool, str]:
+    """
+    Compares an asset with a pool token, considering ETH equal to WETH. It returns the symbol
+    of the asset to be used to avoid having extra resolutions
+    """
+    if pool_token == A_WETH and event_asset == A_ETH:
+        return True, 'ETH'
+
+    return event_asset == pool_token, pool_token.symbol
 
 
 def _find_to_asset_and_amount(events: list['EvmEvent']) -> tuple[Asset, FVal] | None:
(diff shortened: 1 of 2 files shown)


rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -415,7 +423,6 @@ def _maybe_decode_v3_deposit_or_withdrawal(
                 evm_inquirer=self.evm_inquirer,
                 encounter=TokenEncounterInfo(tx_hash=context.transaction.tx_hash),
             )
-            token_with_data = self.eth if token_with_data == A_WETH else token_with_data
```
yabirgb, line 418: this line was making that the check was always against ETH instead of ETH/WETH. The uniswap pools always have WETH afaik and allow to withdraw/deposit ETH

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -89,6 +89,14 @@ def _find_from_asset_and_amount(events: list['EvmEvent']) -> tuple[Asset, FVal]
     return from_asset, from_amount
 
 
+def _asset_is_eth(event_asset: 'Asset', pool_token: 'Asset') -> bool:
+    """Compare two assets evaluating as equal WETH and ETH"""
```
LefterisJP, line 93: From the current name and description I don't understand what this did. Also isn't pool token always an EVM token?

```suggestion
def _compare_with_maybe_eth_token(asset: 'Asset', pool_token: 'EVMToken') -> bool:
    """Compares an asset with a pool token, considering ETH equal to WETH"""
```

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -425,7 +432,7 @@ def _maybe_decode_v3_deposit_or_withdrawal(
         for event in context.decoded_events:
             # search for the event of the first token
             if (
-                event.asset == resolved_assets_and_amounts[0].asset and
+                _asset_is_eth(event.asset, resolved_assets_and_amounts[0].asset) and
```
LefterisJP, line 435: here use kwargs too

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -434,16 +441,17 @@ def _maybe_decode_v3_deposit_or_withdrawal(
                 event.event_type = to_event_type[0]
                 event.event_subtype = to_event_type[1]
                 event.counterparty = CPT_UNISWAP_V3
+                event_asset = event.asset.resolve_to_crypto_asset()
                 event.notes = notes.format(
                     amount=event.balance.amount,
-                    asset=resolved_assets_and_amounts[0].asset.symbol,
+                    asset=event_asset.symbol,
```
LefterisJP, line 447: Why this change? Isn't `event_asset` and `resolved_assets_and_amounts[0].asset` the same thanks to the if check? This way you don't have to do the resolution you added up there.

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
```diff
@@ -452,9 +460,10 @@ def _maybe_decode_v3_deposit_or_withdrawal(
                 event.event_type = to_event_type[0]
                 event.event_subtype = to_event_type[1]
                 event.counterparty = CPT_UNISWAP_V3
+                event_asset = event.asset.resolve_to_crypto_asset()
                 event.notes = notes.format(
                     amount=event.balance.amount,
-                    asset=resolved_assets_and_amounts[1].asset.symbol,
+                    asset=event_asset.symbol,
```
LefterisJP, line 466: same question here

rotkehlchen/chain/ethereum/modules/uniswap/v3/decoder.py
yabirgb: No because the asset could be ETH and then it would differ from the WETH that is resolved before. In `resolved_assets_and_amounts` we only have EVm tokens and in the events we can have evm tokens or ETH  I can add a if check and do the resolution only if it is eth the asset

LefterisJP: Since this is already seen in the utility function you made can you modify it to return this information too? We know ETH's symbol so no need to resolve in that case.

LefterisJP (approved):
LGTM. Thank you Yabir

PR #7334, opened 2024-01-23
style: change layout for the ETH2 form

lukicenturi:
To improve the clarity, so the user doesn't get confused.
Make it clear that only one of `validatorIndex` or `publicKey` needed.

image

### DIFF
--- a/frontend/app/src/components/accounts/blockchain/Eth2Input.vue
+++ b/frontend/app/src/components/accounts/blockchain/Eth2Input.vue
@@ -9,6 +9,7 @@ import {
 } from '@vuelidate/validators';
 import { isEmpty } from 'lodash-es';
 import { toMessages } from '@/utils/validation';
+import { consistOfNumbers } from '@/utils/text';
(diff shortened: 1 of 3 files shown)


frontend/app/src/components/accounts/blockchain/Eth2Input.vue
```diff
@@ -92,10 +92,23 @@ watch(
 </script>
 
 <template>
-  <div class="grid gap-4 grid-cols-2">
-    <div class="col-span-2">
+  <div class="grid gap-4 grid-cols-3 mt-3">
+    <div class="col-span-3 md:col-span-1">
+      <AmountInput
```
kelsos, line 97: Do you think we can also replace the `AmountInput` with a normal `RuiTextInput`. 

Validator index is an identifier that happens to be a number, we don't show it as an amount which is correct, but we add it as an amount which we should not.

lukicenturi: updated

kelsos (approved):
lgtm

PR #7338, opened 2024-01-23
Fix (Paraswap): python/cpython/issues/107078

OjusWiZard:
## Checklist

- [x] Avoiding https://github.com/python/cpython/issues/107078, in `accounting/evm/test_paraswap.py`

### DIFF
--- a/rotkehlchen/tests/unit/accounting/evm/test_paraswap.py
+++ b/rotkehlchen/tests/unit/accounting/evm/test_paraswap.py
@@ -11,17 +11,10 @@ from rotkehlchen.constants.assets import A_ETH, A_WBTC
 from rotkehlchen.fval import FVal
 from rotkehlchen.history.events.structures.evm_event import EvmEvent
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
-from rotkehlchen.tests.utils.accounting import (
-    TIMESTAMP_0_MS,
-    TIMESTAMP_0_SEC,
-    TIMESTAMP_1_MS,
-    TIMESTAMP_1_SEC,
-    accounting_history_process,
-    assert_pnl_totals_close,
-)
+from rotkehlchen.tests.utils.accounting import accounting_history_process, assert_pnl_totals_close
 from rotkehlchen.tests.utils.factories import make_evm_address, make_evm_tx_hash
 from rotkehlchen.tests.utils.messages import no_message_errors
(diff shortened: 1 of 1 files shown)


yabirgb (approved):

rotkehlchen/tests/unit/accounting/evm/test_paraswap.py
```diff
@@ -11,30 +11,28 @@
...
+from rotkehlchen.tests.utils.accounting import accounting_history_process, assert_pnl_totals_close
 from rotkehlchen.tests.utils.factories import make_evm_address, make_evm_tx_hash
 from rotkehlchen.tests.utils.messages import no_message_errors
-from rotkehlchen.types import Location, Price
+from rotkehlchen.types import Location, Price, Timestamp, TimestampMS
 
 if TYPE_CHECKING:
     from rotkehlchen.accounting.accountant import Accountant
 
+TIMESTAMP_1700000000_SEC = Timestamp(1700000000)
```
LefterisJP, line 22: There is no need to define constants for these, or name the constant with the value of the timestamp. Come on wtf?\

If you really want to define them call them `TIMESTAMP_A`and `TIMESTAMP_B`

rotkehlchen/tests/unit/accounting/evm/test_paraswap.py
```diff
@@ -101,13 +99,13 @@ def test_paraswap_swap_with_fee(accountant: 'Accountant', db_settings: dict):
...
-        start_ts=TIMESTAMP_0_SEC,
-        end_ts=TIMESTAMP_1_SEC,
+        start_ts=TIMESTAMP_1700000000_SEC,
+        end_ts=TIMESTAMP_1700000001_SEC,
         history_list=events,
     )
     no_message_errors(accountant.msg_aggregator)
 
-    fee_spent_event = accountant.pots[0].processed_events[4]
+    fee_spent_event = accountant.pots[0].processed_events[3]
```
LefterisJP, line 108: Why did thus change here? The timing of events should not have changed at all.

OjusWiZard: this is because earlier we were getting a "Prefork acquisition for ETC" processed event because the timestamp was < ETH_DAO_FORK_TS. Now it's not

LefterisJP (approved):
LGTM

PR #7340, opened 2024-01-23
Notify user on fatal error during report

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -5295,7 +5295,8 @@ Querying complete action history
 
    :statuscode 200: History processed and returned successfully
    :statuscode 400: Provided JSON is in some way malformed.
-   :statuscode 409: No user is currently logged in.
+   :statuscode 401: No user is currently logged in.
+   :statuscode 409: Fatal accounting error while processing the report.
    :statuscode 500: Internal rotki error.
 
 
--- a/rotkehlchen/accounting/accountant.py
+++ b/rotkehlchen/accounting/accountant.py
@@ -15,7 +15,7 @@ from rotkehlchen.chain.evm.accounting.aggregator import EVMAccountingAggregators
 from rotkehlchen.db.reports import DBAccountingReports
 from rotkehlchen.db.settings import DBSettings
 from rotkehlchen.errors.asset import UnknownAsset, UnprocessableTradePair, UnsupportedAsset
-from rotkehlchen.errors.misc import RemoteError
+from rotkehlchen.errors.misc import AccountingError, RemoteError
 from rotkehlchen.errors.price import NoPriceForGivenTimestamp, PriceQueryUnsupportedAsset
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.premium.premium import Premium
@@ -202,6 +202,10 @@ class Accountant:
                     reason='inability to reach an external service at that point in time',
                 )
                 continue
+            except AccountingError as e:
+                log.error(f'Found critical error {e} when processing history. Stopping.')
+                e.report_id = report_id
+                raise
 
             if processed_events_num == 0:
                 break  # we reached the period end
--- a/rotkehlchen/accounting/cost_basis/base.py
+++ b/rotkehlchen/accounting/cost_basis/base.py
@@ -11,6 +11,7 @@ from rotkehlchen.assets.asset import Asset
 from rotkehlchen.constants import ZERO
 from rotkehlchen.constants.assets import A_ETH, A_WETH
 from rotkehlchen.db.settings import DBSettings
+from rotkehlchen.errors.misc import AccountingError
 from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.fval import FVal
 from rotkehlchen.logging import RotkehlchenLogsAdapter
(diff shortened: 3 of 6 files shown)


rotkehlchen/errors/misc.py
```diff
@@ -80,3 +80,11 @@ class DBSchemaError(Exception):
 
 class GreenletKilledError(Exception):
     """Raised when a greenlet is killed"""
+
+
+class AccountingError(Exception):
```
LefterisJP, line 85: Name is super generic ... but I don't think we have any other right now. So perhaps fine to leave as is. It's a fatal error in accounting processing. In the future we may have other errors in other parts of accounting, or non fatal ones. 

But guess can easily modify stuff if needed.

rotkehlchen/errors/misc.py
```diff
@@ -80,3 +80,11 @@ class DBSchemaError(Exception):
...
+    """Error processing the """
```
LefterisJP, line 86:
```suggestion
    """Fatal error while processing accounting events during a PnL report"""
```

rotkehlchen/accounting/cost_basis/base.py
```diff
@@ -375,6 +379,16 @@ def consume_result(self, used_amount: FVal) -> None:
         `current_amount` is guaranteed to be greater than zero since `consume_result` is
         supposed to be called under `processing_iterator`.
         """
+        if self.current_amount == ZERO:
+            # this shouldn't happen but a user reported it in
+            # https://github.com/rotki/rotki/issues/7273. We couldn't find the reason for it so we
+            # decided to protect against it raising a error shown in the frontend
```
LefterisJP, line 385:
```suggestion
            # decided to protect against it by raising an error shown in the frontend
```

rotkehlchen/errors/misc.py
```diff
@@ -80,3 +80,11 @@ class DBSchemaError(Exception):
...
+
+    def __init__(self, message: str = '', report_id: int | None = None):
```
LefterisJP, line 88: Why message is an optional argument? It should not be.

As for report_id, please explain in the docstring it's not visible in the place where error is raises and is later set when caught.

rotkehlchen/tests/api/test_history.py
```diff
@@ -200,6 +204,37 @@ def test_query_history_remote_errors(rotkehlchen_api_server_with_exchanges):
...
+        response = requests.get(
+            api_url_for(rotkehlchen_api_server, 'historyprocessingresource'),
+            json={'from_timestamp': 0, 'to_timestamp': 1, 'async_query': True},
+        )
+        task_id = assert_ok_async_response(response)
+        outcome = wait_for_async_task(
+            rotkehlchen_api_server,
+            task_id,
+        )
+        assert outcome == {'result': 1, 'message': 'mocked error', 'status_code': 409}
```
LefterisJP, line 235:
```suggestion
        assert outcome == {'result': 1, 'message': 'mocked error', 'status_code': HttpStatus.CONFLICT}
```

rotkehlchen/accounting/cost_basis/base.py
```diff
@@ -375,6 +379,16 @@ def consume_result(self, used_amount: FVal) -> None:
...
+            log.error(f'Division by zero error when processing report using ACB. {self._acquisitions_heap}')  # noqa: E501
+            raise AccountingError(
+                "Can't calculate the average cost basis because the remaining amount of the "
+                "asset is 0. Contact support and provide the log file for more information",
```
LefterisJP, line 389: Is this a user-facing message? Seems so. Do you really think this is informative? It gives half-details and those are even not clear. Why does he care about the remaining amount? What is that even? User does not know. Which asset?

 The user can't fix this on their own so they will anyway need to speak to us.
```suggestion
                "Remaining amount error during ACB calculation for {asset}. Contact support and provide the log file for more information",
```

To provide the {asset} it's possible you will need to add it to `BaseCostBasisMethod` though it may already be retrievable somehow. Not sure.

rotkehlchen/accounting/accountant.py
```diff
@@ -200,6 +200,10 @@ def process_history(
                     reason='inability to reach an external service at that point in time',
                 )
                 continue
+            except AccountingError as e:
+                log.error(f'Found critical error {e} when processing history. Stopping')
```
LefterisJP, line 204:
```suggestion
                log.error(f'Found critical error {e} when processing history. Stopping.')
```

docs/api.rst
```diff
@@ -5295,7 +5295,7 @@ Querying complete action history
 
    :statuscode 200: History processed and returned successfully
    :statuscode 400: Provided JSON is in some way malformed.
-   :statuscode 409: No user is currently logged in.
+   :statuscode 409: No user is currently logged in. Critical error in the data was found while processing the report.
```
LefterisJP, line 5298: This should be like that no?

```suggestion
   :statuscode 401: No user is currently logged in. 
   :statuscode 409: Fatal accounting error while processing the report.
```

LefterisJP (approved):
LGTM

PR #7341, opened 2024-01-23
Avoid querying ignored assets more than once

yabirgb:
Ignored events were being queried for every event instead of only once for the processing

### DIFF
--- a/rotkehlchen/accounting/accountant.py
+++ b/rotkehlchen/accounting/accountant.py
@@ -71,6 +71,7 @@ class Accountant:
         # map event rules signatures to a list of event identifiers affected by them
         # used to know which events need to be invalidated when updating a rule
         self.processable_events_cache_signatures: DefaultLRUCache[int, list[int]] = DefaultLRUCache(default_factory=list, maxsize=PROCESSABLE_EVENTS_CACHE_SIZE)  # noqa: E501
+        self.ignored_asset_ids: set[str] = set()  # populated in process_history
 
     def activate_premium_status(self, premium: Premium) -> None:
         self.premium = premium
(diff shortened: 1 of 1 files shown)


rotkehlchen/accounting/accountant.py
```diff
@@ -71,6 +71,7 @@ def __init__(
         # map event rules signatures to a list of event identifiers affected by them
         # used to know which events need to be invalidated when updating a rule
         self.processable_events_cache_signatures: DefaultLRUCache[int, list[int]] = DefaultLRUCache(default_factory=list, maxsize=PROCESSABLE_EVENTS_CACHE_SIZE)  # noqa: E501
+        self.ignored_asset_ids: set[str] = set()  # populated in process_history
```
LefterisJP, line 74:
```suggestion
        self.ignored_asset_ids: set[str] = set()  # populated in process_history so that we load them in memory once during accounting and not reload them from the DB for every single event processing  # noqa: E501
```

rotkehlchen/accounting/accountant.py
```diff
@@ -230,6 +232,7 @@ def process_history(
         for pot in self.pots:  # delete rules stored in memory since they won't be needed and can be queried again from the db  # noqa: E501
             pot.events_accountant.rules_manager.clean_rules()
 
+        self.ignored_asset_ids.clear()
```
LefterisJP, line 235:
```suggestion
        self.ignored_asset_ids.clear()  # clean ignored assets from memory once PnL report run concludes
```

LefterisJP (approved):
LGTM. Just left some comments you should adjust in another PR. Please remember that.

PR #7354, opened 2024-01-26
Fix (pagination): free limit with ignored assets

OjusWiZard:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=51005645

## Checklist
- [x] Apply limit after the spam assets filter in `get_history_events_count` (just like `rotkehlchen/db/history_events.py::get_history_events`)
- [ ] check if any tests fails


rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -507,11 +507,24 @@ def test_get_events(rotkehlchen_api_server: 'APIServer'):
...
+    with db_history_events.db.user_write() as cursor:
+        for limit, exclude_ignored, total, found in (
+            (9999999999, False, 6, 6), (9999999999, True, 2, 2), (1, False, 6, 1), (1, True, 2, 1),
+        ):
+            assert db_history_events.get_history_events_count(
+                cursor=cursor,
+                group_by_event_ids=True,
+                query_filter=HistoryEventFilterQuery.make(exclude_ignored_assets=exclude_ignored),
+                entries_limit=limit,
+            ) == (total, found)
```
yabirgb, line 526: maybe you could create a more realistic test with some groups having a couple of events and filtering with something that is not 1? I don't understand very well this test

yabirgb (commented):
Please add a changelog entry for this fix and for https://github.com/rotki/rotki/issues/7359

```
NFTs should be queried and displayed properly again
```

rotkehlchen/db/history_events.py
```diff
@@ -355,36 +368,35 @@ def get_history_events(
...
-                special_free_query = True
-
-            type_idx = 1
+                base_query = f'{base_prefix} * FROM (SELECT COUNT(*), {base_suffix} '
+                base_query += f'{free_query_group_by} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?) '  # noqa: E501
+            else:
+                base_query = f'{base_prefix} * FROM (SELECT {base_suffix} '
+                # if we don't group them but still apply free limit, then apply this limit on DISTINCT event_identifiers in a subquery  # noqa: E501
+                # so that we have enough sub_events in the outer filters in prepared_query
+                base_query += f'WHERE event_identifier IN (SELECT DISTINCT event_identifier FROM history_events {free_query_group_by} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?)) '  # noqa: E501
```
yabirgb, line 392: no need to create two operations if you can do it in one

```suggestion
                # if we don't group them but still apply free limit, then apply this limit on DISTINCT event_identifiers in a subquery  # noqa: E501
                # so that we don't remove in the limit events that should appear because they are part of a queried group
                base_query += f'{base_prefix} * FROM (SELECT {base_suffix} WHERE event_identifier IN (SELECT DISTINCT event_identifier FROM history_events {free_query_group_by} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?)) '  # noqa: E501
```

rotkehlchen/db/history_events.py
```diff
@@ -608,10 +620,17 @@ def get_history_events_count(
...
+                with_order=False,
+                with_pagination=False,
+            )
+
             query = 'SELECT * ' + query_filter.get_join_query()
-            if group_by_event_ids:  # we take the groups before the limit has been applied
-                query += ' GROUP BY event_identifier '
-            query += ' ORDER BY timestamp DESC LIMIT ?'
+            # we take the groups before the limit has been applied
+            query += free_query_group_by + 'GROUP BY event_identifier ORDER BY timestamp DESC, sequence_index ASC LIMIT ?'  # noqa: E501
```
yabirgb, line 633: same, don't do it in two operations if it fits in one

```suggestion
            # we take the groups before the limit has been applied
            query = (
                f'SELECT * {query_filter.get_join_query()} {free_query_group_by}'
                'GROUP BY event_identifier ORDER BY timestamp DESC, sequence_index ASC LIMIT ?'
            )
```

rotkehlchen/db/history_events.py
```diff
@@ -355,36 +368,35 @@ def get_history_events(
         TODO: To not query all columns with all joins for all cases, we perhaps can
         peek on the entry type of the filter and adjust the SELECT fields accordingly?
         """
-        free_query_group_by = ''
-        free_query_count = ''
         base_prefix = 'SELECT '
-        type_idx = 0
+        base_suffix = f'{HISTORY_BASE_ENTRY_FIELDS}, {EVM_EVENT_FIELDS}, {ETH_STAKING_EVENT_FIELDS} {ALL_EVENTS_DATA_JOIN}'  # noqa: E501
+        bindings = []
+        type_idx = 1 if group_by_event_ids else 0
```
yabirgb, line 374:
```suggestion
        type_idx = 1 if group_by_event_ids is True else 0
```

rotkehlchen/db/history_events.py
```diff
@@ -355,36 +368,35 @@ def get_history_events(
...
+            base_query = f'{base_prefix} {base_suffix}'
+        else:
+            if group_by_event_ids is True:
+                special_free_query = True  # a bit ugly conditions to keep limit at groups for free users  # noqa: E501
                 free_query_group_by += 'GROUP BY event_identifier'
-                special_free_query = True
-
-            type_idx = 1
+                base_query = f'{base_prefix} * FROM (SELECT COUNT(*), {base_suffix} '
+                base_query += f'{free_query_group_by} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?) '  # noqa: E501
```
yabirgb, line 387: same if you can avoid having the `+=`operator is better. Reason is that you have to mutate and store twice the string and that is harder to read since you are just adding to the end but mentally you have to read what is in the variable. Is easier to read it in just in once

```suggestion
                base_query = (
                    f'{base_prefix} * FROM (SELECT COUNT(*), {base_suffix} '
                    f'{free_query_group_by} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?) '
                )
```

rotkehlchen/db/history_events.py
```diff
@@ -355,36 +368,39 @@ def get_history_events(
...
-                special_free_query = True
-
-            type_idx = 1
+                base_query = (
+                    f'{base_prefix} * FROM (SELECT COUNT(*), {base_suffix} '
+                    f'{free_query_group_by} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?) '
+                )
+            else:
+                # if we don't group them but still apply free limit, then apply this limit on DISTINCT event_identifiers in a subquery  # noqa: E501
+                # so that we have enough sub_events in the outer filters in prepared_query
```
LefterisJP, line 392: You are adding noqa: E501 and still split the text into the next line. Also text does not really make sense. You are not describing the check, you are describing what is in your mind in the current code, when compared to how it was before.

```suggestion
                # if we don't group them but still apply free limit, then apply this limit only on any event_identifiers that fall inside the limit
```

LefterisJP (approved):
LGTM. Small nitpick will fix myself in bump PR. But this will need quite a lot of testing in the history events section for bugfixes.

PR #7360, opened 2024-01-28
feat(opensea): Update OpenSea API integration to v2

### LINKED ISSUE #7359: NFT queries no longer work - switch to opensea v2
## Problem Definition

It seems that Opensea which was the sole provider of NFT data for us has dev-rugged us: https://twitter.com/LefterisJP/status/1751561509929369940

It was probably not noticed fast enough due to the fact we query cached DB nft data: https://github.com/rotki/rotki/blob/3d267d0591a342f53a5a1110f0e25f1cd10bb6ab/rotkehlchen/rotkehlchen.py#L947

So it only got noticed after a user reporting seeing an NFT they sold, still show in rotki even after a restart.

We generally need to not rely only on opensea for NFT data: https://github.com/rotki/rotki/issues/4911

## Task

For the purpose of this issue we should keep it simple and just switch to using opensea v2: https://docs.opensea.io/reference/list_nfts_by_account so that NFT detection works properly again

moxw:
Closes #7359

## Checklist


rotkehlchen/externalapis/opensea.py
```diff
@@ -140,12 +136,14 @@ def _query(
...
+            query_str = f'https://api.opensea.io/api/v2/collections/{options["name"]}/stats'  # type: ignore
+        elif endpoint == 'collection':
+            query_str = f'https://api.opensea.io/api/v2/collections/{options["collection_slug"]}'  # type: ignore
         elif endpoint == 'asset':
-            query_str = f'https://api.opensea.io/api/v1/asset/{options["address"]}/{options["item_id"]}'  # type: ignore
+            query_str = f'https://api.opensea.io/api/v2/chain/ethereum/contract/{options["address"]}/nfts/{options["item_id"]}'  # type: ignore
             options = None
         else:
-            query_str = f'https://api.opensea.io/api/v1/{endpoint}'
+            query_str = f'https://api.opensea.io/api/v2/chain/ethereum/account/{options["address"]}/nfts'  # type: ignore
```
LefterisJP, line 146: Something seems off here.

The possible endpoints are given in the literal above `'assets', 'collections', 'collectionstats', 'asset', 'collection'`.

Then at the if elif/else we are missing `collections` and `assets`. So the `else` case can't be both.

rotkehlchen/externalapis/opensea.py
```diff
@@ -246,12 +244,12 @@ def _deserialize_nft(
             collection = None
             # NFT might not be part of a collection
             if 'collection' in entry:
-                saved_entry = self.collections.get(entry['collection']['name'])
+                saved_entry = self.collections.get(entry['collection'])
```
LefterisJP, line 247: At line 259, `entry['collection']` is reused.

And we have some keys that won't work the way they are there:

```
                        banner_image=collection_data['banner_image_url'],
                        description=collection_data['description'],	                        description=collection_data['description'],
                        large_image=collection_data['large_image_url'],	                        
                        large_image=collection_data['large_image_url'],
```

LefterisJP (commented):
These related nft tests failed:

```
rotkehlchen/tests/api/test_nfts.py::test_nft_balances_and_prices
rotkehlchen/tests/unit/test_evm_names.py::test_download_ens_avatar
rotkehlchen/tests/unit/test_nfts.py::test_addresses_queried_for_nfts
```

They are all VCRed so you would need to also update the VCR cassettes of these tests as shown here: https://rotki.readthedocs.io/en/latest/contribute.html#using-vcr

We can help with anything in discord where we are already talking. In the worst case we can just take over and put finishing touches at any point.

rotkehlchen/externalapis/opensea.py
moxw: Removed 'collections' as it's not relevant anymore.

rotkehlchen/externalapis/opensea.py
moxw: raising UnknownAsset error instead

rotkehlchen/externalapis/opensea.py
```diff
@@ -246,38 +226,35 @@ def _deserialize_nft(
...
                     if saved_entry.floor_price is not None:
                         floor_price = saved_entry.floor_price
                 else:  # should not happen. That means collections endpoint doesnt return anything
-                    collection_data = entry['collection']
-                    collection = Collection(
-                        name=collection_data['name'],
-                        banner_image=collection_data['banner_image_url'],
-                        description=collection_data['description'],
-                        large_image=collection_data['large_image_url'],
+                    raise UnknownAsset(
```
yabirgb, line 241: Change this to `DeserializationError` instead since UnknownAsset is for when we don't have an asset locally

yabirgb (commented):
Hello! Left a tiny change. The cassettes failed to reproduce the tests. You need to record them and also create a PR to the cassettes repo with the cassettes that you recorded. The name of the branch needs to be the same as the name of this one. https://rotki.readthedocs.io/en/stable/usage_guide.html#token-detection It might not let you create a branch so create a PR instead. I believe it should work but lets see. Ping us in discord if you need any help with it

rotkehlchen/externalapis/opensea.py
LefterisJP: And also add that to the docstring as by reading the docstring it seems as if the only thing this can raise is a RemoteError. But it does also potentially raise `DeserializationError` elsewhere too

LefterisJP (approved):
LGTM. We need a changelog as I said in discord

PR #7369, opened 2024-01-30
Remove contract check in opensea api calls

yabirgb:
This check was avoiding some NFTs from being rendered because the collections weren't queried. I've checked with the v2 endpoint and I didn't find issues with this logic.

Closes #(issue_number)

## Checklist


LefterisJP (approved):
LGTM

PR #7374, opened 2024-01-31
Fix customized events filter in free version

LefterisJP:
Follow up from https://github.com/rotki/rotki/pull/7366.

There was one more query that needed adjustment. Ended up defining what "*" means per filter since the ambiguity can lead to errors like this.


yabirgb (approved):
tested the changes with my account and seems to work

LefterisJP:
Rerunning failed tests to see if test-caching will be picked up from master

LefterisJP:
All green. Creating a master branch in test-caching made it work thanks to the conftest.py VCR setup

PR #7381, opened 2024-02-01
Remove Kraken public methods that were not used by us

LefterisJP:
So removing the extra logic and in the one test that it's being queried we just use requests directly.

This was triggered by them removing the ability to query the public methods with POST: https://twitter.com/LefterisJP/status/1752847650720768145


yabirgb (approved):
lgtm

PR #7384, opened 2024-02-01
Add decoder for Octant rewards

LefterisJP:


LefterisJP:
Will merge. Failing test is due to etherscan error, though it shoudl be VCRed.

PR #7391, opened 2024-02-02
Fix (Paraswap): Direct UniV3 swaps by address

OjusWiZard:
## Checklist

- [x] Update Paraswap decoder to catch direct Uniswap V3 swap by address


yabirgb (approved):
lgtm. It just replaces the function signature by a check in the topic and test passed

LefterisJP (approved):
LGTM

PR #7394, opened 2024-02-05
Add gmx v1 staking decoders

yabirgb:
Closes #(issue_number)

## Checklist


rotkehlchen/chain/arbitrum_one/modules/gmx/decoder.py
```diff
@@ -197,12 +199,49 @@ def decode_position_change(self, context: DecoderContext) -> DecodingOutput:
...
+        for event in context.decoded_events:
+            if (
+                event.location_label == account and
+                event.event_type == HistoryEventType.SPEND and
+                event.balance.amount == amount
+            ):
+                event.counterparty = CPT_GMX
+                event.event_type = HistoryEventType.STAKING
+                event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
+                # the contract when staking GMX also creates transfers for (sGMX, sbGMX, sbfGMX)
```
LefterisJP, line 223: why mention this here? What do we need to do for those as developers of the decoder? Is it the elif?

yabirgb: added the comment because instead of using GMX as symbol I resolve the asset to use the symbol. The reason is because it catches not only GMX but all the variants that they have until the more staked version of the token

LefterisJP (approved):
lgtm

PR #7395, opened 2024-02-05
Feat (BaseDecoderTools): Support internal txns

OjusWiZard:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=49582891

## Checklist

- [x] Add `maybe_get_internal_txs_to_address` to `BaseDecoderTools`
- [x] Check if it's working with `test_simple_swap_eth_fee`
- [ ] Don't query some transactions multiple times


rotkehlchen/chain/evm/transactions.py
```diff
@@ -541,6 +550,53 @@ def ensure_tx_data_exists(
         tx_receipt = self.dbevmtx.get_receipt(cursor, tx_hash, self.evm_inquirer.chain_id)
         return tx_data, tx_receipt  # type: ignore  # tx_data can't be None here
 
+    def get_and_ensure_internal_txns_of_parent_in_db(
+            self,
+            tx_hash: 'EVMTxHash',
+            to_address: 'ChecksumEvmAddress',
+            for_address: 'ChecksumEvmAddress',
```
yabirgb, line 557: is this always a tx `for_address` sending to `to_address`. The naming is confusing. Maybe `from_address` and `to_address`?

rotkehlchen/chain/polygon_pos/modules/paraswap/decoder.py
```diff
@@ -24,4 +25,5 @@ def __init__(
             msg_aggregator=msg_aggregator,
             router_address=PARASWAP_AUGUSTUS_ROUTER,
             fee_receiver_address=PARASWAP_FEE_CLAIMER,
+            native_currency=A_POLYGON_POS_MATIC.resolve_to_crypto_asset(),
```
yabirgb, line 28: same, you don't need to resolve

rotkehlchen/chain/evm/transactions.py
```diff
@@ -541,6 +550,53 @@ def ensure_tx_data_exists(
...
+
+        May raise:
+        - RemoteError if there is a problem querying the data sources or transaction hash does
+        not exist."""
+        db_internal_txs = self.dbevmtx.get_evm_internal_transactions(
+            parent_tx_hash=tx_hash,
+            blockchain=CHAINID_TO_SUPPORTED_BLOCKCHAIN[self.evm_inquirer.chain_id],
+        )
+        if len(db_internal_txs) > 0:
+            return db_internal_txs
```
yabirgb, line 570: shouldn't you check the from and to here? so it is `len(checked_transactions) > 0` Otherwise I see that you could have internal transaction that could beling to an address that you are not looking for

rotkehlchen/chain/evm/transactions.py
```diff
@@ -541,6 +550,53 @@ def ensure_tx_data_exists(
...
+        # check cache if this parent hash was queried before for this receiver and affected address
+        with self.database.conn.read_ctx() as cursor:
+            affected_address = self.database.get_dynamic_cache(
+                cursor=cursor,
+                name=DBCacheDynamic.EXTRA_INTERNAL_TX,
+                tx_hash=tx_hash,
+                receiver=to_address,
+            )
+        if affected_address == for_address:  # if we have queried them before
+            return db_internal_txs
```
yabirgb, line 581: for clairity this means that there was none in the db and there aren't any new so

```suggestion
            return []
```

rotkehlchen/chain/evm/decoding/base.py
```diff
@@ -333,6 +335,24 @@ def get_or_create_evm_asset(self, address: ChecksumEvmAddress) -> CryptoAsset:
...
+        """Adds internal txs to the given address into the DB if they did not exists,
+        while decoding an event for the given address and then returns them."""
+        return [
+            internal_tx for internal_tx
+            in self.evm_txns.get_and_ensure_internal_txns_of_parent_in_db(
+                tx_hash=tx_hash,
+                to_address=to_address,
+                for_address=for_address,
+            )
+            if internal_tx.to_address == to_address
```
yabirgb, line 353: don't you need to also check with the for_address?

rotkehlchen/db/cache.py
```diff
@@ -57,7 +64,11 @@ def get_db_key(self, **kwargs: Unpack[LabeledLocationIdArgsType]) -> str:
     def get_db_key(self, **kwargs: Unpack[AddressArgType]) -> str:
         ...
 
-    def get_db_key(self, **kwargs: str) -> str:
+    @overload
+    def get_db_key(self, **kwargs: Unpack[ExtraTxArgType]) -> str:
+        ...
+
+    def get_db_key(self, **kwargs: Any) -> str:
```
yabirgb, line 71: why did you need to use the Any here?

rotkehlchen/db/dbhandler.py
```diff
@@ -706,11 +707,20 @@ def get_dynamic_cache(
...
+            **kwargs: Unpack[ExtraTxArgType],
+    ) -> Timestamp | None:
+        ...
+
     def get_dynamic_cache(
             self,
             cursor: 'DBCursor',
             name: DBCacheDynamic,
-            **kwargs: str,
+            **kwargs: Any,
```
yabirgb, line 723: why was this change to any forced?

rotkehlchen/db/evmtx.py
```diff
@@ -155,7 +155,7 @@ def get_evm_internal_transactions(
                 trace_id=result[0],
                 from_address=result[1],
                 to_address=result[2],
-                value=result[3],
+                value=int(result[3]),
```
yabirgb, line 158: why this change here?

rotkehlchen/chain/evm/transactions.py
OjusWiZard: This function returns all of them. Filtering them with from/to in BaseDecoderTools

rotkehlchen/db/evmtx.py
OjusWiZard: So, this was because the value is typed as an int in the `EvmInternalTransaction`. But, here we get it from DB, so it's returned as str. This raises an error [here in the decoder](https://github.com/rotki/rotki/pull/7395/files#diff-dfc6191ba411ce3c637694024e930e2a9bb812dbbda6b3b7306c2adfe88f976aR144) FVal doesn't accept str.

yabirgb: It was only used in one place and the test was expecting str instead of int https://github.com/rotki/rotki/blob/64beb7ec4c6e38f67d2d5f4c6444df88d525c1d2/rotkehlchen/tests/external_apis/test_etherscan.py#L185

I believe it was always executing the else branch of this condition https://github.com/rotki/rotki/blob/64beb7ec4c6e38f67d2d5f4c6444df88d525c1d2/rotkehlchen/chain/evm/decoding/decoder.py#L664

rotkehlchen/chain/evm/transactions.py
```diff
@@ -248,8 +257,8 @@ def _get_transactions_for_range(
 
     def _query_and_save_internal_transactions_for_range_or_parent_hash(
             self,
-            address: ChecksumEvmAddress | None,
             period_or_hash: TimestampOrBlockRange | EVMTxHash,
+            address: ChecksumEvmAddress | None = None,
```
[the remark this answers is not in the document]
OjusWiZard: This was to make `None` its default argument value so that we don't have to explicitly assign None, in the place where I've called it.

rotkehlchen/db/evmtx.py
```diff
@@ -508,6 +508,11 @@ def delete_transactions(
             'DELETE FROM evm_tx_mappings WHERE tx_id=? AND value=?',
             [(x, HISTORY_MAPPING_STATE_DECODED) for x in tx_ids],
         )
+        # Delete any key_value_cache entries
+        write_cursor.executemany(
+            'DELETE FROM key_value_cache WHERE name LIKE ?',
+            [(f'extrainternaltx_{tx_hash.hex()}%',) for tx_hash in tx_hashes],
+        )
```
yabirgb, line 515: Add a small test for this

rotkehlchen/chain/evm/transactions.py
```diff
@@ -541,6 +550,54 @@ def ensure_tx_data_exists(
...
+
+        May raise:
+        - RemoteError if there is a problem querying the data sources or transaction hash does
+        not exist."""
+        db_internal_txs = self.dbevmtx.get_evm_internal_transactions(
+            parent_tx_hash=tx_hash,
+            blockchain=CHAINID_TO_SUPPORTED_BLOCKCHAIN[self.evm_inquirer.chain_id],
+        )
+        if len(db_internal_txs) > 0:
+            return db_internal_txs
```
yabirgb, line 571: this was not done

rotkehlchen/chain/evm/decoding/paraswap/decoder.py
```diff
@@ -130,6 +127,17 @@ def _decode_swap(
...
                 fee_raw = hex_or_bytes_to_int(log_event.data)
                 break
+        else:
+            # if no fee is found yet, check if eth fee is paid through internal transactions
+            internal_txs_to_fee_claimer = self.base.maybe_get_internal_txs(
+                tx_hash=context.transaction.tx_hash,
+                from_address=self.router_address,
+                to_address=self.fee_receiver_address,
+                user_address=sender,
+            )
```
yabirgb, line 137: you are not handling the remote error right? It should catch and log

rotkehlchen/chain/evm/decoding/base.py
```diff
@@ -333,6 +335,22 @@ def get_or_create_evm_asset(self, address: ChecksumEvmAddress) -> CryptoAsset:
...
+            user_address: ChecksumEvmAddress,
+    ) -> list[EvmInternalTransaction]:
+        """Adds internal txs from/to the given address into the DB if they did not exists,
+        while decoding an event for the given user_address and then returns them."""
+        return self.evm_txns.get_and_ensure_internal_txns_of_parent_in_db(
+            tx_hash=tx_hash,
+            from_address=from_address,
+            to_address=to_address,
+            user_address=user_address,
+        )
```
yabirgb, line 352: THis became a wrapper. Just call it directly from the decoder

rotkehlchen/tests/api/test_ethereum_transactions.py
```diff
@@ -1265,6 +1266,16 @@ def query_transactions(rotki):
...
     tx4_events[0]['entry']['identifier'] = result['identifier']
 
+    # Also add a cache entry for a transaction
+    with rotki.data.db.user_write() as write_cursor:
+        rotki.data.db.set_dynamic_cache(
+            write_cursor=write_cursor,
+            name=DBCacheDynamic.EXTRA_INTERNAL_TX,
+            value='',
+            tx_hash='0x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f',
+            receiver='',
```
yabirgb, line 1276: fill the receiver and value fields

rotkehlchen/tests/api/test_ethereum_transactions.py
```diff
@@ -1291,6 +1302,14 @@ def query_transactions(rotki):
             assert cursor.execute(f'SELECT COUNT(*) from {name}').fetchone()[0] == count
         customized_events = dbevents.get_history_events(cursor, EvmEventFilterQuery.make(), True)
 
+        # Check if related cache is removed
+        assert rotki.data.db.get_dynamic_cache(
+            cursor=cursor,
+            name=DBCacheDynamic.EXTRA_INTERNAL_TX,
+            tx_hash='0x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f',
+            receiver='',
```
yabirgb, line 1310: fill the receiver with a `make_evm_address`

rotkehlchen/tests/db/test_db.py
```diff
@@ -354,6 +355,13 @@ def test_writing_fetching_data(data_dir, username, sql_vm_instructions_cb):
             blockchain=SupportedBlockchain.POLYGON_POS,
             tokens=[A_POLYGON_POS_MATIC],
         )
+        data.db.set_dynamic_cache(
+            write_cursor=write_cursor,
+            name=DBCacheDynamic.EXTRA_INTERNAL_TX,
+            value=string_to_evm_address('0xd36029d76af6fE4A356528e4Dc66B2C18123597D'),
+            tx_hash='',
```
yabirgb, line 362: use `make_evm_tx_hash`

rotkehlchen/db/evmtx.py
```diff
@@ -508,6 +519,11 @@ def delete_transactions(
             'DELETE FROM evm_tx_mappings WHERE tx_id=? AND value=?',
             [(x, HISTORY_MAPPING_STATE_DECODED) for x in tx_ids],
         )
+        # Delete any key_value_cache entries
+        write_cursor.executemany(
+            'DELETE FROM key_value_cache WHERE name LIKE ?',
+            [(f'extrainternaltx_{tx_hash.hex()}%',) for tx_hash in tx_hashes],
```
LefterisJP, line 525: make this into a constant final variable

```suggestion
            [(f'{EXTRAINTERNALTXPREFIX}_{tx_hash.hex()}%',) for tx_hash in tx_hashes],
```

rotkehlchen/db/dbhandler.py
```diff
@@ -2032,6 +2052,10 @@ def delete_data_for_evm_address(
             'DELETE FROM evm_accounts_details WHERE account=? AND chain_id=?',
             (address, blockchain.to_chain_id().serialize_for_db()),
         )
+        write_cursor.execute(
+            'DELETE FROM key_value_cache WHERE name LIKE "extrainternaltx_%" AND value = ?',
```
LefterisJP, line 2056: make a constant final variable

```suggestion
            'DELETE FROM key_value_cache WHERE name LIKE "{PREFIX}_%" AND value = ?',
```

rotkehlchen/db/cache.py
```diff
@@ -36,6 +36,12 @@ class AddressArgType(TypedDict):
     address: ChecksumEvmAddress
 
 
+class ExtraTxArgType(TypedDict):
+    """Type of kwargs, used to get the value of `DBCacheDynamic.EXTRA_INTERNAL_TX`"""
+    tx_hash: str
```
LefterisJP, line 41: Isnt the type `EVMTxHash` here?
```suggestion
    tx_hash: EVMTxHash
```

rotkehlchen/db/dbhandler.py
```diff
@@ -706,18 +707,27 @@ def get_dynamic_cache(
...
             **kwargs: str,
-    ) -> Timestamp | None:
+    ) -> Timestamp | ChecksumEvmAddress | None:
         """Returns the cache value from the `key_value_cache` table of the DB
         according to the given `name` and `kwargs`. Defaults to `None` if not found."""
         value = cursor.execute(
             'SELECT value FROM key_value_cache WHERE name=?;', (name.get_db_key(**kwargs),),
         ).fetchone()
-        return None if value is None else Timestamp(int(value[0]))
+        return None if value is None else value[0]
```
LefterisJP, line 730: this changes logic here, not only typing. If returning a timestamp from string this used to work and now it will no longer work. That is dangerous. What are the possibilities here?

rotkehlchen/tests/db/test_db.py
```diff
@@ -354,6 +355,14 @@ def test_writing_fetching_data(data_dir, username, sql_vm_instructions_cb):
...
             tokens=[A_POLYGON_POS_MATIC],
         )
+        random_tx_hash_in_cache = make_evm_tx_hash()
+        data.db.set_dynamic_cache(
+            write_cursor=write_cursor,
+            name=DBCacheDynamic.EXTRA_INTERNAL_TX,
+            value=string_to_evm_address('0xd36029d76af6fE4A356528e4Dc66B2C18123597D'),
+            tx_hash=random_tx_hash_in_cache.hex(),  # pylint: disable=no-member
+            receiver=string_to_evm_address('0xd36029d76af6fE4A356528e4Dc66B2C18123597D'),
+        )
```
LefterisJP, line 365: can you add a get of the cache here also just to make sure it's set properly?

rotkehlchen/tests/api/test_ethereum_transactions.py
```diff
@@ -1265,6 +1266,17 @@ def query_transactions(rotki):
...
+    # Also add a cache entry for a transaction
+    with rotki.data.db.user_write() as write_cursor:
+        random_receiver_in_cache = make_evm_address()
+        rotki.data.db.set_dynamic_cache(
+            write_cursor=write_cursor,
+            name=DBCacheDynamic.EXTRA_INTERNAL_TX,
+            value=make_evm_address(),
+            tx_hash='0x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f',
+            receiver=random_receiver_in_cache,
+        )
```
LefterisJP, line 1278: can you add a get of the cache here just to check it's set properly?

rotkehlchen/db/dbhandler.py
OjusWiZard: It's hard to track the possibilities. Maybe we can handle the typing similar to DBSettings like this?
```py
    def get_typed_value(self, value: str) -> Timestamp | ChecksumEvmAddress:
        """Returns the given value with the correct type for the current key"""
        if self == DBCacheDynamic.EXTRA_INTERNAL_TX:
            return string_to_evm_address(value)
        else:
            return Timestamp(int(value))
...
return None if value is None else name.get_typed_value(value[0])
```

rotkehlchen/db/dbhandler.py
```diff
@@ -706,18 +708,27 @@ def get_dynamic_cache(
...
             **kwargs: str,
-    ) -> Timestamp | None:
+    ) -> Timestamp | ChecksumEvmAddress | None:
         """Returns the cache value from the `key_value_cache` table of the DB
         according to the given `name` and `kwargs`. Defaults to `None` if not found."""
         value = cursor.execute(
             'SELECT value FROM key_value_cache WHERE name=?;', (name.get_db_key(**kwargs),),
         ).fetchone()
-        return None if value is None else Timestamp(int(value[0]))
+        return None if value is None else name.get_typed_value(value[0])
```
LefterisJP, line 731: no this is not a good solution. You are adding an extra if, and basically move the deserialization inside the getter function. 

While the caller of `get_dynamic_cache` knows what to expect and should just use it properly. Can't you just return value[0] and then the caller deserializes properly?

Also again I ask is the timestamp saved as an int or as a string in the DB cache?

rotkehlchen/db/cache.py
```diff
@@ -44,6 +51,7 @@ class DBCacheDynamic(Enum):
     LAST_QUERY_ID = '{location}_{location_name}_{account_id}_last_query_id'
     WITHDRAWALS_TS = 'ethwithdrawalsts_{address}'
     WITHDRAWALS_IDX = 'ethwithdrawalsidx_{address}'
+    EXTRA_INTERNAL_TX = f'{EXTRAINTERNALTXPREFIX}_{{tx_hash}}_{{receiver}}'
```
LefterisJP, line 54: Got a better idea after all.

Make this enum have 2 values. Example for one line.
```suggestion
    EXTRA_INTERNAL_TX = f'{EXTRAINTERNALTXPREFIX}_{{tx_hash}}_{{receiver}}', _simple_callback
```

And then modify the `get_db_key` to use `self.value[0]`
And also add a `@property` like so:

```python
@property 
def deserialize_callback -> Callback[[Any], Any]:
    return self.value[1]
```

And then define the simple callback above and give it to all lines.

```python
T = TypeVar('T')

def _simple_callback(value: T) -> T | None :
    return None if value is None else value
```

And for timestamps define and add this one

```python
def _deserialize_timestamp_from_str(value: str) -> Timestamp:
    return Timestamp(int(value))
```

rotkehlchen/exchanges/bitstamp.py
```diff
@@ -270,7 +270,7 @@ def query_online_deposits_withdrawals(
                 location=self.location.serialize(),
                 location_name=self.name,
             )) is not None:
-                offset = result
+                offset = int(result)
```
LefterisJP, line 273: this will also need to change. Make sure typing in LAST_CRYPTOTX_OFFSET is int.

rotkehlchen/externalapis/blockscout.py
```diff
@@ -131,7 +131,7 @@ def query_withdrawals(self, address: ChecksumEvmAddress) -> set[int]:
                 name=DBCacheDynamic.WITHDRAWALS_IDX,
                 address=address,
             )) is not None:
-                last_known_withdrawal_idx = idx_result
+                last_known_withdrawal_idx = int(idx_result)
```
LefterisJP, line 134: same here

rotkehlchen/externalapis/blockscout.py
```diff
@@ -195,7 +195,7 @@ def query_withdrawals(self, address: ChecksumEvmAddress) -> set[int]:
                 self.db.set_dynamic_cache(
                     write_cursor=write_cursor,
                     name=DBCacheDynamic.WITHDRAWALS_IDX,
-                    value=index_to_write,
+                    value=int(index_to_write),
```
LefterisJP, line 198: this is not needed. Index is an int already.
```suggestion
                    value=index_to_write,
```

rotkehlchen/exchanges/bitstamp.py
```diff
@@ -372,7 +372,7 @@ def _query_crypto_transactions(self, offset: int) -> list[AssetMovement]:
                 self.db.set_dynamic_cache(
                     write_cursor=write_cursor,
                     name=DBCacheDynamic.LAST_CRYPTOTX_OFFSET,
-                    value=Timestamp(options['offset']),
+                    value=int(options['offset']),
```
LefterisJP, line 375: not needed. This is an int already no?

LefterisJP (approved):
lgtm.

1 test failed due to missing cassette which is in develop

2 more tests failed due to a remote error.
- `rotkehlchen/tests/unit/test_arbitrum_one_inquirer.py::test_arbitrum_one_nodes_prune_and_archive_status` which can fail if pruned status node is queried ... we need to change those tests 
- `rotkehlchen/tests/api/test_user_evm_tokens.py::test_adding_evm_token_with_underlying_token` failed due to coingecko rate limit

PR #7400, opened 2024-02-06
Filter staking performance by status + more correct APR calculation

LefterisJP:


LefterisJP:
All tests green, will merge

PR #7412, opened 2024-02-07
Fix and add tests for own beacon node

LefterisJP:


LefterisJP:
The failed test is a remote query in a non-vcred test: `test_icons_and_avatars_cache_deletion`.

unrelated

PR #7415, opened 2024-02-07
fix: some styling from the text-field

lukicenturi:
Fix some styling from the textfield
Remove some vuetify classes


kelsos (approved):
lgtm, I rebased on top of develop so we can merge as soon as the CI is done

PR #7417, opened 2024-02-07
Fixes wal issues

yabirgb:
We were getting autogenerated whl files in the packaged globaldb folder when running the dev app and errors when querying all the assets if several queries were attempted.

What I did was:

- The query won't fail if the database is lock. We were commiting the wal file since when we attach a database the attached database won't see everything unless it is commited. It happens that if we spawn several times the query very fast two queries lock the database
- To avoid creating the whl in the packaged db I changed the mode to remove the WAL mode in the packaged db ~one and I always attach it in read only mode what avoids changing the mode~ Edit: I was wrongly thinking that the packaged db was the one getting attached but it was the user db to the user globaldb. 

The diff of the database is empty since what I changed is the wal mode

```
PRAGMA journal_mode=DELETE
```

```
(rotki-main) ➜  rotki git:(fixes-wal) ✗ sqldiff ~/Downloads/global-wal.db rotkehlchen/data/global.db    
(rotki-main) ➜  rotki git:(fixes-wal) ✗ 
```

### DIFF
--- a/rotkehlchen/globaldb/handler.py
+++ b/rotkehlchen/globaldb/handler.py
@@ -237,6 +237,23 @@ class GlobalDBHandler:
         if self._packaged_db_conn is not None:
             self._packaged_db_conn.close()
 
+    def _wal_checkpoint(self) -> None:
+        """
+        Commit wal file. If database is locked we ignore it. This is needed when attaching
+        databases and we want to have up to date information. It can happen that if two tasks
+        are spawned in a short time and both of them try to commit the WAL one of them finds
+        the database locked. This doesn't prevent any query later from executing.
+
+        https://www.sqlite.org/pragma.html#pragma_wal_checkpoint
+        """
+        try:
+            self.conn.execute('PRAGMA wal_checkpoint(PASSIVE);')
+        except sqlite3.OperationalError as e:
+            if 'database table is locked' in (err := str(e)):
+                log.warning(f'Could not commit globaldb wal file. Ignoring. {err}')
+            else:
+                raise
+
     @staticmethod
     def packaged_db_conn() -> DBConnection:
         """Return a DBConnection instance for the packaged global db."""
@@ -368,9 +385,9 @@ class GlobalDBHandler:
 
         with userdb.conn.read_ctx() as cursor:
             globaldb = GlobalDBHandler()
-            globaldb.conn.execute('PRAGMA wal_checkpoint;')  # needed to have up to date information when attaching  # noqa: E501
+            globaldb._wal_checkpoint()
             cursor.execute(
-                f'ATTACH DATABASE "file:{globaldb.filepath()!s}?more=ro" AS globaldb KEY "";',
+                f'ATTACH DATABASE "{globaldb.filepath()!s}" AS globaldb KEY "";',
             )
             try:
                 # get all underlying tokens
@@ -477,11 +494,11 @@ class GlobalDBHandler:
         query, bindings = filter_query.prepare()
         query = ALL_ASSETS_TABLES_QUERY.format(dbprefix='globaldb.') + query
         resolved_eth = A_ETH.resolve_to_crypto_asset()
-        globaldb.conn.execute('PRAGMA wal_checkpoint;')  # needed to have up to date information when attaching  # noqa: E501
+        globaldb._wal_checkpoint()
         with db.conn.read_ctx() as cursor:
             treat_eth2_as_eth = db.get_settings(cursor).treat_eth2_as_eth
             cursor.execute(
-                f'ATTACH DATABASE "file:{globaldb.filepath()!s}?more=ro" AS globaldb KEY "";',
+                f'ATTACH DATABASE "{globaldb.filepath()!s}" AS globaldb KEY "";',
             )
             try:
                 cursor.execute(query, bindings)
--- a/rotkehlchen/globaldb/handler.py
+++ b/rotkehlchen/globaldb/handler.py
@@ -370,7 +370,7 @@ class GlobalDBHandler:
             globaldb = GlobalDBHandler()
             globaldb.conn.execute('PRAGMA wal_checkpoint;')  # needed to have up to date information when attaching  # noqa: E501
             cursor.execute(
-                f'ATTACH DATABASE "{globaldb.filepath()!s}" AS globaldb KEY "";',
+                f'ATTACH DATABASE "file:{globaldb.filepath()!s}?more=ro" AS globaldb KEY "";',
             )
             try:
                 # get all underlying tokens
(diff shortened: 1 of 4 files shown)


rotkehlchen/globaldb/handler.py
```diff
@@ -237,6 +237,20 @@ def cleanup(self) -> None:
         if self._packaged_db_conn is not None:
             self._packaged_db_conn.close()
 
+    @staticmethod
+    def _wal_checkpoint(conn: DBConnection) -> None:
+        """
+        Commit wal file. If database is locked we ignore it. This is needed when attacking
```
LefterisJP, line 243:
```suggestion
        Commit wal file. If database is locked we ignore it. This is needed when attaching
```

rotkehlchen/globaldb/handler.py
```diff
@@ -237,6 +237,20 @@ def cleanup(self) -> None:
...
+        databases and we want to have up to date information
+        """
```
LefterisJP, line 245: Also can you add a link to the docs here:  https://www.sqlite.org/pragma.html#pragma_wal_checkpoint ?

rotkehlchen/globaldb/handler.py
```diff
@@ -237,6 +237,20 @@ def cleanup(self) -> None:
         if self._packaged_db_conn is not None:
             self._packaged_db_conn.close()
 
+    @staticmethod
+    def _wal_checkpoint(conn: DBConnection) -> None:
+        """
+        Commit wal file. If database is locked we ignore it. This is needed when attacking
+        databases and we want to have up to date information
```
LefterisJP, line 244:
```suggestion
        databases and we want to have up to date information.
```

rotkehlchen/globaldb/handler.py
```diff
@@ -237,6 +237,20 @@ def cleanup(self) -> None:
         if self._packaged_db_conn is not None:
             self._packaged_db_conn.close()
 
+    @staticmethod
```
LefterisJP, line 240: I don't understand why you are making it a static method and then you are passing the connection? And the connection you pass in all cases is the globaldb own connection. 

This means this should be a normal method with no arguments and use `self.conn` inside

rotkehlchen/globaldb/handler.py
```diff
@@ -237,6 +237,20 @@ def cleanup(self) -> None:
...
+    def _wal_checkpoint(conn: DBConnection) -> None:
+        """
+        Commit wal file. If database is locked we ignore it. This is needed when attacking
+        databases and we want to have up to date information
+        """
+        try:
+            conn.execute('PRAGMA wal_checkpoint(PASSIVE);')
+        except sqlite3.OperationalError as e:
+            if 'database table is locked' in (err := str(e)):
+                log.warning(f'Could not commit globaldb wal file. Ignoring. {err}')
```
LefterisJP, line 250: are you sure this is safe to continue? Won't subsequent DB conn.execute() calls also fail with the same error?

rotkehlchen/globaldb/handler.py
```diff
@@ -368,9 +382,9 @@ def retrieve_assets(userdb: 'DBHandler', filter_query: 'AssetsFilterQuery') -> t
 
         with userdb.conn.read_ctx() as cursor:
             globaldb = GlobalDBHandler()
-            globaldb.conn.execute('PRAGMA wal_checkpoint;')  # needed to have up to date information when attaching  # noqa: E501
+            globaldb._wal_checkpoint(conn=globaldb.conn)
             cursor.execute(
-                f'ATTACH DATABASE "{globaldb.filepath()!s}" AS globaldb KEY "";',
+                f'ATTACH DATABASE "file:{globaldb.filepath()!s}?more=ro" AS globaldb KEY "";',
```
LefterisJP, line 387: why this change? What does it do? This needs a comment in the code explaining it.

rotkehlchen/globaldb/handler.py
```diff
@@ -477,11 +491,11 @@ def search_assets(
         query, bindings = filter_query.prepare()
         query = ALL_ASSETS_TABLES_QUERY.format(dbprefix='globaldb.') + query
         resolved_eth = A_ETH.resolve_to_crypto_asset()
-        globaldb.conn.execute('PRAGMA wal_checkpoint;')  # needed to have up to date information when attaching  # noqa: E501
+        globaldb._wal_checkpoint(conn=globaldb.conn)
         with db.conn.read_ctx() as cursor:
             treat_eth2_as_eth = db.get_settings(cursor).treat_eth2_as_eth
             cursor.execute(
-                f'ATTACH DATABASE "{globaldb.filepath()!s}" AS globaldb KEY "";',
+                f'ATTACH DATABASE "file:{globaldb.filepath()!s}?more=ro" AS globaldb KEY "";',
```
LefterisJP, line 498: same here

rotkehlchen/globaldb/handler.py
yabirgb: in both cases where I use it I already have the globaldb object initialized by `GlobalDbHandler()` and It takes the conn for that initailized instance

rotkehlchen/globaldb/handler.py
yabirgb: No, they don't fail. The reason is that the other operation that is operating in the database finishes and unlocks the database

rotkehlchen/globaldb/handler.py
yabirgb: imagen

yabirgb: imagen

rotkehlchen/globaldb/handler.py
yabirgb: One thing that I couldn't get working is that the initialization of the connection accepts a timeout argument that tells how much python should wait before raising OperationalError if the database is locked. With pragma seems that it doesn't work (?) https://docs.python.org/3/library/sqlite3.html#sqlite3.connect

rotkehlchen/globaldb/handler.py
yabirgb: Is not needed. It was my fault. As stated in the edit I was trying different things and when setting the mode to read only in the connection string of the attachment I didn't see the -whl file generated but since I was working with 3 databases I mixed the one being attached here. This is not needed

rotkehlchen/globaldb/handler.py
yabirgb: About this I'm not coming up with it. I reproduced as joseph found it and if it fails you can query again the endpoint or do any globaldb query and it works fine

rotkehlchen/tests/unit/globaldb/test_globaldb.py
```diff
@@ -1302,3 +1302,16 @@ def test_assets_in_same_collection(globaldb: GlobalDBHandler):
 
     # check an asset with no related assets
     assert globaldb.get_assets_in_same_collection(identifier=A_ETH.identifier) == (A_ETH,)
+
+
+def test_check_wal_mode_of_package_db(globaldb: GlobalDBHandler) -> None:
+    """
+    If the packaged db is modified and the wal mode is activated users can have issues
+    when the database is in a read only device. To avoid such issues we need to ensure
+    that no new files are created and this is why we prevent from shipping the database
```
LefterisJP, line 1311:
```suggestion
    that no new files are created and this is why we prevent from shipping the packaged database
```

LefterisJP (approved):
LGTM. The docstring change can go in another PR.

PR #7419, opened 2024-02-07
Include end of timerange for staking performance.

LefterisJP:
Also handle edge case where no events for a given validator could lead to KeyError for 'sum'


yabirgb (approved):
lgtm

PR #7422, opened 2024-02-08
Remove spammy warnings when some etherscan is down

LefterisJP:
When some source of transactions (etherscan atm) is not responding we don't fetch latest transactions for an address. This can happen quite often and it's quite spammy for the user.

No need to show it like this.

e.g. it's happening right now with polygon being down


yabirgb (approved):
really needed

PR #7429, opened 2024-02-08
More staking

LefterisJP:
Now address filtering should take into account associated addresses
properly. Matches a validator index if the address given is:

- A depositor for that index
- A withdrawal address for that index
- A fee recipient for that index

Also the address filter and validator indices filters can be combined

Also add a log and more comments to restructuring data directory


LefterisJP:
all green

PR #7433, opened 2024-02-09
Allow filtering eth2 daily stats by associated address

LefterisJP:
Same as staking performance endpoint


LefterisJP:
all green. Love to see it.

PR #7439, opened 2024-02-10
Forbid merging same asset to itself

LefterisJP:
Before this commit you could replace/merge an asset by itself, which killed the DB logic and effectively deleted the asset from your global DB and some parts of user DB.

Result was a botched DB which needed manual intervention. This should no longer be allowed by the API.


LefterisJP:
Only failure is `test_gnosis_nodes_prune_and_archive_status` due to the unfortunately usual by now hitting pruned node. Those tests need to change.

PR #7440, opened 2024-02-11
Common filtering across the eth staking endpoints

LefterisJP:
Now all eth staking endpoints accept a list of associated addresses, specific validator indices and a validator status filter.

Also fix a bug where if a validator had exited daily stats would not be pulled for it at all.


LefterisJP:
all green. Merging.

PR #7454, opened 2024-02-13
fix: add validator refresh on staking page

kelsos:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/staking/eth/EthStakingPage.vue
+++ b/frontend/app/src/components/staking/eth/EthStakingPage.vue
@@ -1,10 +1,14 @@
 <script setup lang="ts">
 import { Blockchain } from '@rotki/common/lib/blockchain';
+import { objectOmit } from '@vueuse/core';
+import { isEmpty } from 'lodash-es';
 import { EthStaking } from '@/premium/premium';
 import { Module } from '@/types/modules';
 import { Section } from '@/types/status';
-import { useBlockchainBalances } from '@/composables/blockchain/balances';
-import type { EthStakingCombinedFilter, EthStakingFilter } from '@rotki/common/lib/staking/eth2';
+import { TaskType } from '@/types/task-type';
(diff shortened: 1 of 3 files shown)


frontend/app/src/components/staking/eth/EthStakingPage.vue
```diff
@@ -48,20 +59,30 @@ const premium = usePremium();
...
   };
 
-  const validatorRefresh: Promise<void>[] = (userInitiated ? [refreshValidators()] : []);
+  const updatePerformance = async (userInitiated = false): Promise<void> => {
+    await refreshPerformance(userInitiated);
+    // if the number of validators is bigger than the total entries in performance
+    // force a refresh of performance to pick the missing performance entries.
+    const totalValidators = get(eth2Validators).entriesFound;
+    const totalPerformanceEntries = get(performance).entriesTotal;
+    if (totalValidators > totalPerformanceEntries) {
```
tewshi, line 76: if there won't be a case where `totalValidators < totalPerformanceEntries` is truthy, then it is all good to me.

kelsos: Unless you manually go an remove the validators I don't think there would be a scenario where this could happen automatically. This is supposed to handle an edge case where the missing validators don't make it to performance due to timing.

frontend/app/src/components/staking/eth/EthStakingPage.vue
```diff
@@ -4,6 +4,8 @@ import { EthStaking } from '@/premium/premium';
 import { Module } from '@/types/modules';
 import { Section } from '@/types/status';
 import { useBlockchainBalances } from '@/composables/blockchain/balances';
```
lukicenturi, line 6: this can be removed right?

LefterisJP (approved):
LGTM

PR #7459, opened 2024-02-14
fix: asset update stuck after completed

lukicenturi:
Issue: asset update after login, get stuck in the white screen after completed


kelsos (approved):
lgtm

PR #7466, opened 2024-02-14
Fix (Airdrops): Optional icon_url

OjusWiZard:
## Checklist

- [x] Add optional field in response `icon_url`

**What problem does the PR solve?**
The problem is for the assets whose icons are not already present in the repo. If we add a new airdrop remotely for these assets, the frontend will have no way to show their icons.

**What was the approach taken?**
Now we can pass an optional `icon_path` (to their location in the data repo) in the metadata JSON for these new assets. When the backend receives them, it will convert them to full URLs and return them via the API. So, frontend will receive these new assets' icons' URLs which can be directly rendered from GitHub.

**What the frontend needs to do?**
When parsing the response from [GET /api/(version)/blockchains/eth/airdrops](https://rotki.readthedocs.io/en/latest/api.html#get--api-(version)-blockchains-eth-airdrops) or [GET /api/(version)/airdrops/metadata](https://rotki.readthedocs.io/en/latest/api.html#get--api-(version)-airdrops-metadata), also check if `icon_url` field is present in the response. If it is present use that URL instead of trying to use a packaged icon

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -4392,6 +4392,8 @@ class RestAPI:
                     'name': airdrop.name,
                     'icon': airdrop.icon,
                 })
+                if airdrop.icon_url is not None:
+                    result[-1]['icon_url'] = airdrop.icon_url
         except RemoteError as e:
             return api_response(wrap_in_fail_result(str(e)), status_code=HTTPStatus.BAD_GATEWAY)
         return api_response(result=_wrap_in_ok_result(result=result))
--- a/rotkehlchen/chain/ethereum/airdrops.py
+++ b/rotkehlchen/chain/ethereum/airdrops.py
@@ -54,6 +54,7 @@ class Airdrop(NamedTuple):
     url: str
     name: str
     icon: str
+    icon_url: str | None = None
     cutoff_time: Timestamp | None = None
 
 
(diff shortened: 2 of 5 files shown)


rotkehlchen/chain/ethereum/airdrops.py
```diff
@@ -118,6 +119,7 @@ def _parse_airdrops(database: 'DBHandler', airdrops_data: dict[str, Any]) -> dic
                 url=airdrop_data['url'],
                 name=airdrop_data['name'],
                 icon=airdrop_data['icon'],
+                icon_url=f"{AIRDROPS_REPO_BASE}/{airdrop_data['icon_path']}" if 'icon_path' in airdrop_data else None,  # noqa: E501
```
yabirgb, line 122: you name the attribute icon_url but in the json is icon_path. I believe it should be icon_url in all the places

yabirgb (commented):
one small comment

rotkehlchen/chain/ethereum/airdrops.py
OjusWiZard: this was because in the metadata, we are returning the path, and in the api response we are returning the url

rotkehlchen/api/rest.py
```diff
@@ -4391,6 +4391,7 @@ def get_airdrops_metadata(self) -> Response:
                     'identifier': identifier,
                     'name': airdrop.name,
                     'icon': airdrop.icon,
+                    'icon_url': airdrop.icon_url,
```
LefterisJP, line 4394: Why not put it only if it's there? Should be optional right?

rotkehlchen/chain/ethereum/airdrops.py
LefterisJP: I don't understand here why the difference between icon path and icon url

rotkehlchen/tests/unit/test_ethereum_airdrops.py
```diff
@@ -55,6 +55,7 @@
         'url': 'https://claim.harvest.finance/',
         'name': 'Grain',
         'icon': 'grain.png',
+        'icon_path': 'airdrops/icons/grain.svg',
```
LefterisJP, line 58: why this? What is icon_path?

LefterisJP (approved):
LGTM

PR #7467, opened 2024-02-14
Monerium API only for premium

LefterisJP:
Backend work.

- Monerium credentials can only be given to rotki if user has an active premium
- The task to query the monerium API and decorate all related transactions with the banking information will now only run if the user has an active premium.


LefterisJP:
all green. Merging.

PR #7468, opened 2024-02-15
Use the correct RemoteError exception

LefterisJP:
Code that imported wrong RemoteError should handle remote errors properly again


yabirgb (approved):
I'm going to change the configuration in the editor to avoid the imports

PR #7475, opened 2024-02-15
Fix deletion of staking daily stats at validator refresh

LefterisJP:
During the DB upgrade we were adding new fields to the eth2_validators table and making the validator index being a UNIQUE key instead of a primary key.

When detecting new data for validators the code was doing INSERT OR REPLACE which essentially replaced the entire row with a new one.

The problem here is that daily stats is tied to eth2 validators via foreign key connection to the validator index.

This replacing dropped all eth daily stats. We now do a more complicated query to detect if it's a new validator addition in which case we have to insert it or if it's an old one in which case we have to update some keys. Then we need to find which keys to update.

### DIFF
--- a/rotkehlchen/db/eth2.py
+++ b/rotkehlchen/db/eth2.py
@@ -231,7 +231,7 @@ class DBEth2:
             self,
             write_cursor: 'DBCursor',
             index: int,
-            withdrawable_ts: Timestamp,
+            withdrawable_timestamp: Timestamp,
     ) -> None:
         """If the validator has withdrawal events, find last one and mark as exit if after withdrawable ts"""  # noqa: E501
         write_cursor.execute(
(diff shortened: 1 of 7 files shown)


rotkehlchen/db/eth2.py
```diff
@@ -259,22 +259,45 @@ def add_or_update_validators_except_ownership(
...
+                    if getattr(db_validator, attr) != (new_value := getattr(validator, attr)):
+                        write_cursor.execute(
+                            f'UPDATE eth2_validators SET {attr}=? WHERE public_key=?',
+                            (new_value, validator.public_key),
+                        )
+            else:  # insertion case
+                write_cursor.execute(
+                    'INSERT INTO '
+                    'eth2_validators(validator_index, public_key, ownership_proportion, withdrawal_address, activation_timestamp, withdrawable_timestamp) VALUES(?, ?, ?, ?, ?, ?)',  # noqa: E501
+                    validator.serialize_for_db(),
```
yabirgb, line 299: sqlite has upsert that I believe is what you need here? https://www.sqlite.org/lang_upsert.html Mentioning it here just in case it could be useful

LefterisJP:
only failed test is remote error `test_detect_tokens_for_addresses`

rotkehlchen/db/eth2.py
LefterisJP: upsert has some edge cases if you read and it's fairly new. After I started working on rotki ;).

But may work. Can try it later.

PR #7484, opened 2024-02-16
Increase spam threshold

yabirgb:
Increase the threshold we use to mark a token as spam in a transaction since when claiming multiple tokens we can get more than 5


LefterisJP (approved):
LGTM

PR #7495, opened 2024-02-16
Update some dependencies

LefterisJP:


LefterisJP:
ah good to see it green with those updates!

PR #7499, opened 2024-02-17
Gotta go fast (use uv in CI)

yabirgb:
It speed ups docker building times and ci:

- Uses minimal installation for rust
- Uses uv in docker container
- Uses uv in the github ci (can be improved with https://github.com/astral-sh/uv/issues/1526)


yabirgb:
The CI built at https://github.com/rotki/rotki/actions/runs/8172650278

yabirgb:
The linting job compared in this PR vs a PR in develop with pip
imagen
imagen

requirements_dev.txt
```diff
@@ -7,7 +7,7 @@ pytest-cov==4.1.0
 coverage==7.4.3
 psutil==5.9.8
 pytest-freezer==0.4.8
-git+https://github.com/rotki/pytest-deadfixtures@87d2be8#egg=pytest-deadfixtures # temporarily due to false positive
+pytest-deadfixtures @ git+https://github.com/rotki/pytest-deadfixtures@87d2be8#egg=pytest-deadfixtures # temporarily due to false positive
```
LefterisJP, line 10: Do we still need this?

LefterisJP (commented):
Some questions:

1. Should we also switch our dev environment to using this uv thing?
2. Shouldn't we also change dev docs for this?

yabirgb:
@LefterisJP 

> Should we also switch our dev environment to using this uv

It is not a requirement since the requirements files are not modified. It is indeed faster installing dependencies but pip works the same. `uv` implements a subset of pip that is compatible with most of the features

> Shouldn't we also change dev docs for this?

I don't believe so since uv is a choice and pip is familiar for a lot of users already

LefterisJP (approved):
LGTM

Let's hope building stuff will work fine now.

PR #7501, opened 2024-02-18
Remove deprecated pkg_resources and simplify packaging a bit

LefterisJP:
- Remove pkg_resources usage in favor of importlib metadata and others as per deprecation warnings and official docs:
https://setuptools.pypa.io/en/latest/pkg_resources.html
- Also simplified some things for packaging. And probably can do more if enough time is spent to remember how it worked


LefterisJP:
Hmmm so building locally worked but in the CI not. Will have to play with it

PR #7507, opened 2024-02-19
Fix (Binance): Lending APIs

OjusWiZard:
## Checklist

- [x] [Switch](https://www.binance.com/en/support/announcement/updates-on-binance-simple-earn-sapi-endpoints-2023-09-21-8060cafe466b4d2c87c79a37a0757aeb) to new Simple Earn APIs and stop using `/sapi/v1/lending/union/*` endpoints
- [x] Update tests

### DIFF
--- a/rotkehlchen/exchanges/binance.py
+++ b/rotkehlchen/exchanges/binance.py
@@ -5,9 +5,9 @@ import logging
 from collections import defaultdict
 from contextlib import suppress
 from json.decoder import JSONDecodeError
-from typing import TYPE_CHECKING, Any, Literal
+from sqlite3 import IntegrityError
+from typing import TYPE_CHECKING, Any, Final, Literal
 from urllib.parse import urlencode
-from uuid import uuid4
 
 import gevent
 import requests
@@ -80,34 +80,33 @@ log = RotkehlchenLogsAdapter(logger)
 
 # Binance launched at 2017-07-14T04:00:00Z (12:00 GMT+8, Beijing Time)
 # https://www.binance.com/en/support/articles/115000599831-Binance-Exchange-Launched-Date-Set
-BINANCE_LAUNCH_TS = Timestamp(1500001200)
-API_TIME_INTERVAL_CONSTRAINT_TS = 7776000  # 90 days
-BINANCE_LENDING_INTEREST_HISTORY_INTERVAL_TS = 2592000  # 30 days
+BINANCE_LAUNCH_TS: Final = Timestamp(1500001200)
+API_TIME_INTERVAL_CONSTRAINT_TS: Final = 7776000  # 90 days
 
 # this determines the length of the data returned, 100 is the maximum value possible.
-BINANCE_LENDING_INTEREST_HISTORY_PAGE_SIZE = 100
+BINANCE_SIMPLE_EARN_HISTORY_PAGE_SIZE: Final = 100
 
-V3_METHODS = (
+V3_METHODS: Final = (
     'account',
     'myTrades',
     'openOrders',
     'exchangeInfo',
     'time',
 )
-PUBLIC_METHODS = ('exchangeInfo', 'time')
+PUBLIC_METHODS: Final = ('exchangeInfo', 'time')
 
-RETRY_AFTER_LIMIT = 60
+RETRY_AFTER_LIMIT: Final = 60
 # Binance api error codes we check for (all below apis seem to have the same)
 # https://binance-docs.github.io/apidocs/spot/en/#error-codes-2
 # https://binance-docs.github.io/apidocs/futures/en/#error-codes-2
 # https://binance-docs.github.io/apidocs/delivery/en/#error-codes-2
-REJECTED_MBX_KEY = -2015
+REJECTED_MBX_KEY: Final = -2015
 
 
 BINANCE_API_TYPE = Literal['api', 'sapi', 'dapi', 'fapi']
 
-BINANCE_BASE_URL = 'binance.com/'
-BINANCEUS_BASE_URL = 'binance.us/'
+BINANCE_BASE_URL: Final = 'binance.com/'
+BINANCEUS_BASE_URL: Final = 'binance.us/'
 
 
 class BinancePermissionError(RemoteError):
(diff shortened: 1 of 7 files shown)


rotkehlchen/exchanges/binance.py
```diff
@@ -593,24 +626,28 @@ def query_lending_interests_history(
...
         May raise:
         - RemoteError
         - BinancePermissionError
-        """  # noqa: E501
+        """
         ranges = DBQueryRanges(self.db)
-        range_query_name = f'{self.location}_lending_history_{self.name}'
+
+        # Query and save flexible simple earn history
+        range_query_name = f'{self.location}_flexible_lending_history_{self.name}'
```
LefterisJP, line 650: Don't change the name here. We already will have `'{self.location}_lending_history` in the DB. So we need to keep same name.

Unless it's a completely different one.

Or if you want to keep the same, or if it's a different one then you need to do a DB upgrade renaming or removing the previous key values.

rotkehlchen/exchanges/binance.py
```diff
@@ -719,6 +756,112 @@ def query_lending_interests_history(
                     queried_ranges=[(start_ts, end_ts)],
                 )
 
+        # Query and save locked simple earn history
+        range_query_name = f'{self.location}_locked_lending_history_{self.name}'
```
LefterisJP, line 760: this is new right?

rotkehlchen/exchanges/binance.py
```diff
@@ -719,6 +756,112 @@ def query_lending_interests_history(
...
+                    usd_value = usd_price * interest_received
+                except NoPriceForGivenTimestamp as e:
+                    log.warning(
+                        f'Could not find USD price of {asset} at {timestamp}. {e!s} '
+                        f'Using zero usd_value for lending history entry.',
+                    )
+                    usd_value = ZERO
+
+                event = HistoryEvent(
+                    event_identifier=uuid4().hex,
```
LefterisJP, line 831: Can't we do anything better than completely random event id?

rotkehlchen/tests/api/test_balances.py
```diff
@@ -303,7 +303,7 @@ def test_query_all_balances_ignore_cache(
         )
         for fn in function_call_counters:
             if fn._mock_wraps == original_binance_query_dict:
-                assert fn.call_count == 3
+                assert fn.call_count == 2
```
LefterisJP, line 306: Why is everything one less query? I saw 2 endpoints added 1 removed no?

rotkehlchen/exchanges/binance.py
OjusWiZard: [Found it](https://www.binance.com/en/support/announcement/introducing-simple-earn-the-evolution-of-binance-savings-binance-staking-b067f764f4f14e5aa46cca1bf2b55ab8)
TL;DR: Simple Earn is a rebranding of the same thing.
So, we can use the same range_query_name

rotkehlchen/tests/api/test_balances.py
OjusWiZard: so this is because these mock function call count are for `api_query_dict`, we were using it for the previous API. But now we're using `api_query_list` for the new API. That is tested [here](https://github.com/rotki/rotki/pull/7507/files#diff-2147b8264bfde8a089c45c3387721b3f48a50b4a8fb94d0721c8d666457b5001R548).

rotkehlchen/exchanges/binance.py
```diff
@@ -534,56 +534,88 @@ def _query_lending_balances(
...
+                        f'Found unsupported {self.name} asset {e.identifier}. '
+                        f'Ignoring its lending balance query.',
+                    )
+                    continue
+                except UnknownAsset as e:
+                    self.msg_aggregator.add_warning(
+                        f'Found unknown {self.name} asset {e.identifier}. '
+                        f'Ignoring its lending balance query.',
+                    )
+                    continue
```
LefterisJP, line 595: Like you did elsewhere, I think those can be combined with an extra if check in the case of the exception

```suggestion
combine them
```

rotkehlchen/exchanges/binance.py
```diff
@@ -712,6 +748,98 @@ def query_lending_interests_history(
...
+            if len(events) != 0:
+                history_events_db = DBHistoryEvents(self.db)
+                with self.db.user_write() as write_cursor:
+                    try:
+                        history_events_db.add_history_events(write_cursor, events)
+                    except InputError as e:
+                        self.msg_aggregator.add_error(
+                            f'Failed to save Binance {self.name} lending interest history from '
+                            f'{query_start_ts} to {query_end_ts} in the database. {e!s}',
+                        )
```
LefterisJP, line 841: Use the single one. The one you use is just a for loop. If for some reason an event is requeried with the change you made it won't be duplicated in the DB due to the event identifier but would fail with an sqlcipher Integrity error. Also why do you catch InputError. Where did you see this?

The probable exceptions according to the code are:
```
        - DeserializationError if the event could not be serialized for the DB
        - sqlcipher.IntegrityError: If the asset of the added history event does not exist in
        the DB. Can only happen if an event with an unresolved asset is passed.
```

```suggestion
                    for event in events:
                    try:
                        history_events_db.add_history_event(write_cursor, event)
                    except (IntegrityError, DeserializationError) as e:
                        log.error(f'Did not add lending history event {event} to the DB due to {e})
```

rotkehlchen/exchanges/binance.py
```diff
@@ -719,6 +832,11 @@ def query_lending_interests_history(
                     queried_ranges=[(start_ts, end_ts)],
                 )
 
+        if events_failed_to_save > 0:
+            self.msg_aggregator.add_error(
+                f'Failed to save {events_failed_to_save} binance lending history events in the DB. Check logs.',  # noqa: E501
+            )
```
LefterisJP, line 838: Remove this logic. We don't need to spam user with this. If they are missing events they can go and check the logs.

```suggestion
```

LefterisJP (approved):
LGTM

PR #7516, opened 2024-02-21
refactor: VSwitch to RuiSwitch

lukicenturi:
Closes #(issue_number)

## Checklist


kelsos (approved):
lgtm as soon as we fix the failing tests

btw @lukicenturi do you think we can do something about the highlight circle? 

Rather than it lasting while focusing on the switch would it be possible have it disappear a bit after clicking?

lukicenturi:
> lgtm as soon as we fix the failing tests
> 
> btw @lukicenturi do you think we can do something about the highlight circle?
> 
> Rather than it lasting while focusing on the switch would it be possible have it disappear a bit after clicking?

Yeah I can do that. I just need to change the condition from 'focus' to 'active' instead.

PR #7526, opened 2024-02-22
Process kraken rewards program events

yabirgb:
This PR addresses two issues:

- Handle .F assets comming from kraken
- Process deposits for the rewards program.

It is missing:

- the unstaking event since we don't have data. I requested @tewshi to unstake a little amount to generate an event and check but it takes 7 days

edit: I've created a ticket to kraken since it is a bit confusing and I'm waiting for a response


yabirgb:
Closing since the approach seems to be wrong. More info at https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=62706896

PR #7534, opened 2024-02-24
Fix the curve pool token price calculation

### LINKED ISSUE #7532: Price calculation for curve pools is wrong in some cases
## Problem Definition

It seems that the calculation for the price of a curve pool is wrong in some cases. Presumably the result of `find_curve_pool_price` is what is wrong.

An example is this: `Curve.fi Factory Crypto Pool: YFI/ETH`. Address of the pool is `0x29059568bb40344487d62f7450e78b8e6c74e0e5`. The price seen for the amount of LP tokens seems to be roughly half.

Seen in 1.32.1.

## Task

Figure out and fix it.

LefterisJP:
Fix #7532

Also @yabirgb I see many curve pools are missing from our global DB. Such as the `TryLSD` one: `0x2570f1bD5D2735314FC102eb12Fc1aFe9e6E7193`. And a looooot more.

Shouldn't these be autodetected and added to our shipped global DB periodically?

I see them in neither the packaged DB nor in my own one.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* :bug:`7532` Curve LP token price calculation should now be correct.
+
 * :release:`1.32.1 <2024-02-23>`
 * :bug:`-` rotki will now automatically update the local airdrops' CSVs when needed, without manually deleting them.
 * :bug:`-` Fix the issue where an incorrect amount of ETH is displayed in the Loopring account table.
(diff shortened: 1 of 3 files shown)


yabirgb:
We have both the api and an onchain query for the curve pools. The api seems to be missing the optimized curve pools like the one you linked

https://github.com/curvefi/curve-api/issues/305

image


I searched manually and it isn't in any of the endpoints that we query.

Second the on chain query is not correct and we do https://github.com/rotki/rotki/blob/f1a7410f577e707304ba9e46717051fbf7179aef/rotkehlchen/chain/ethereum/modules/curve/curve_cache.py#L361

where we query for the 7th registry instead of using the value from `max_id` https://etherscan.io/address/0x0000000022D53366457F9d5E68Ec105046FC4383#readContract

yabirgb (approved):
lgtm

PR #7541, opened 2024-02-26
Pixel binance

LefterisJP:
Corresponding assets PR https://github.com/rotki/assets/pull/136

Also shows the global DB diff

Fixes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=54372339.

I can't find LDAI 

Also turn some msg aggregator spam to simple log errors/


yabirgb (commented):
Isn't this missing a change in the assets url queried? https://github.com/rotki/rotki/blob/1ad44f2a33057400ada3931da7415a032f46e9eb/rotkehlchen/globaldb/updates.py#L129

LefterisJP:
> Isn't this missing a change in the assets url queried?
> 
> https://github.com/rotki/rotki/blob/1ad44f2a33057400ada3931da7415a032f46e9eb/rotkehlchen/globaldb/updates.py#L129

hmmm I am not sure. Do we do that anywhere else? Will we be changing the branch when merging? I don't think we should. Not sure if it can be done properly. The dynamic check we have is production or not.

PR #7549, opened 2024-02-27
Update asset updates regexes.

LefterisJP:
To be able to work with the changes here:
https://github.com/rotki/assets/pull/133

Essentially to:

1. Handle "(" inside a string
2. Handle no space between ) and VALUES


yabirgb (approved):
It is missing tests and I don't trust regex but I entrust myself to Godt  and we can fix/add in a different PR

PR #7560, opened 2024-02-29
Remove initialization of singleton when calling static methods

yabirgb:
When calling static methods we are calling in some cases the constructor of the singleton and those are additional operations that we can avoid.

Closes #(issue_number)

## Checklist


LefterisJP (approved):
LGTM

PR #7578, opened 2024-03-04
refactor(frontend): history events redecoding status

tewshi:
final part of https://github.com/orgs/rotki/projects/10/views/1?pane=issue&itemId=48079086

## Checklist


tewshi:
> So lgtm things seems to work as expected.
> 
> I noticed something and I am not sure why it is, I don't remember it seeing before.
> 
> It seems that when you hover it goes to normal grey and when you don't it returns to the secondary color.
> 
> Can you check if you can replicate after installing the latest dependencies on develop? If you can, make sure that the color switching does not happen. At least from what I see the color difference seems quite big and it does not look good.

I did a rebase and had conflict on that bar, will check if something changed from what I had before the rebase.

PR #7581, opened 2024-03-05
Fix (7572): LAST_QUERY_ID query in Coinbase

OjusWiZard:
to test with binaries


LefterisJP:
as discussed this is not the way

PR #7600, opened 2024-03-12
feat(frontend): cex asset mapping

### LINKED ISSUE #2698: Move asset CEX mapping to the asset entries in the global DB
## Problem Definition

All the location (CEX) mapppings of assets to identifier per location are in huge hard coded mappings which end up in memory. We need a better solution that is only used when needed and is editable by the user.

### New table

Create a new table in the global DB for location asset mappings. Something like:

```sqlite
CREATE TABLE IF NOT EXISTS location_asset_mappings (
    location TEXT NOT NULL,
    value TEXT NOT NULL
    target_asset TEXT NOT NULL COLLATE NOCASE,
    FOREIGN KEY(target_asset) REFERENCES assets(identifier) ON UPDATE CASCADE ON DELETE CASCADE,
    PRIMARY KEY (location, value)
);
```

### Move mappings

Move all mappings from memory into that table

### Create API endpoints

The user should be able to view, add and edit those mappings. So that we don't need to "fix" mappings by making new releases, but just asking the user to add it to the mapping.

### Allow pulling of remote data

These mappings should also be added to our remote data pulling mechanism. So if we push an update of the mappings it's updated for all clients. Our choices should override users as there should be only one source of truth.

lukicenturi:
Closes #2698
## Checklist


frontend/app/src/components/inputs/ExchangeInput.vue
```diff
@@ -0,0 +1,46 @@
...
+  value: string | null;
+}>();
+
+const emit = defineEmits<{
+  (e: 'input', value: string): void;
+}>();
+
+const vModel = useSimpleVModel(props, emit);
+const { exchangesWithKey }
+  = storeToRefs(useLocationStore());
```
kelsos, line 12: was this supposed to go to the next line?

frontend/app/src/components/inputs/ExchangeInput.vue
```diff
@@ -0,0 +1,46 @@
...
+
+const rootAttrs = useAttrs();
+</script>
+
+<template>
+  <VAutocomplete
```
kelsos, line 18: Didn't we already had a component to select exchanges? what are the differences? is that a normal select?

kelsos: Scratch it I see that this is the component it was not just extracted before

frontend/app/src/locales/en.json
```diff
@@ -1132,6 +1132,30 @@
   "asset_management": {
     "add_error": "Could not add asset: {message}",
     "add_title": "Add a new asset",
+    "cex_mapping": {
+      "add_error": "Could not add the mapping: {message}",
+      "add_mapping": "Add Mapping",
+      "add_title": "Add a new mapping",
+      "all_exchanges": "All exchanges",
+      "confirm_delete": {
+        "message": "Are you sure you want to delete asset mapping for symbol {asset} in {location}?",
```
kelsos, line 1141:
```suggestion
        "message": "Are you sure you want to delete the asset mapping for the symbol {asset} in {location}?",
```

kelsos (approved):
lgtm

PR #7547, opened 2024-02-26
Refactor (2698): Migrate exchange mappings

OjusWiZard:
## Checklist
- [x] Move the hardcoded exchange asset mappings to globalDB
- [x] https://github.com/rotki/assets/pull/137

- SQL diff of `rotkehlchen/data/global.db` is
```sql
CREATE TABLE IF NOT EXISTS location_asset_mappings (
    location TEXT,
    exchange_symbol TEXT NOT NULL,
    local_id TEXT NOT NULL COLLATE NOCASE,
    UNIQUE (location, exchange_symbol)
);
```
\+ `rotkehlchen/data/populate_location_asset_mappings.sql` +
```sql
DELETE FROM settings WHERE rowid=39;
INSERT INTO settings(rowid,name,value) VALUES(51,'version','7');
```

- `rotkehlchen/data/populate_location_asset_mappings.sql` generated using:
```py
cursor.executemany(  # populate the mappings for all the exchanges
    'INSERT OR REPLACE INTO location_asset_mappings(location, exchange_symbol, local_id) VALUES (?, ?, ?)', [
        (location, value, identifier)
        for location, mapping in [
            (None, COMMON_ASSETS_MAPPINGS),
            (Location.BINANCE.serialize_for_db(), {k: v for k, v in WORLD_TO_BINANCE.items() if (k, v) not in COMMON_ASSETS_MAPPINGS.items()}),
            (Location.BITFINEX.serialize_for_db(), {k: v for k, v in WORLD_TO_BITFINEX.items() if (k, v) not in COMMON_ASSETS_MAPPINGS.items()}),
            (Location.BITPANDA.serialize_for_db(), {k: v for k, v in WORLD_TO_BITPANDA.items() if (k, v) not in COMMON_ASSETS_MAPPINGS.items()}),
            (Location.BITSTAMP.serialize_for_db(), {k: v for k, v in WORLD_TO_BITSTAMP.items() if (k, v) not in COMMON_ASSETS_MAPPINGS.items()}),
            (Location.BITTREX.serialize_for_db(), {k: v for k, v in WORLD_TO_BITTREX.items() if (k, v) not in COMMON_ASSETS_MAPPINGS.items()}),
            (Location.BLOCKFI.serialize_for_db(), {k: v for k, v in WORLD_TO_BLOCKFI.items() if (k, v) not in COMMON_ASSETS_MAPPINGS.items()}),
[...]

PR #7608, opened 2024-03-13
Feat (2698): Remote updates for the mappings

### LINKED ISSUE #2698: Move asset CEX mapping to the asset entries in the global DB

OjusWiZard:
Closes #2698

## Checklist

- [x] Allow remote updates for location_asset_mappings from data repo
- [x] Update `info.json` there: https://github.com/rotki/data/pull/69

It's a follow-up PR from https://github.com/rotki/rotki/pull/7547, where the hardcoded mappings are moved to the packaged GlobalDB. And, also an upgrade is made to move these mapping for already existing users.

### DIFF
--- a/rotkehlchen/db/updates.py
+++ b/rotkehlchen/db/updates.py
@@ -7,6 +7,7 @@ import requests
 
 from packaging import version as pversion
 from rotkehlchen.api.websockets.typedefs import WSMessageType
+from rotkehlchen.assets.asset import Asset
 from rotkehlchen.assets.spam_assets import update_spam_assets
 from rotkehlchen.chain.evm.accounting.structures import BaseEventSettings, TxAccountingTreatment
 from rotkehlchen.db.accounting_rules import DBAccountingRules
@@ -25,6 +26,9 @@ from rotkehlchen.serialization.deserialize import deserialize_evm_address
 from rotkehlchen.types import (
     AddressbookEntry,
     AddressbookType,
+    Location,
+    LocationAssetMappingDeleteEntry,
+    LocationAssetMappingUpdateEntry,
     OptionalChainAddress,
     SupportedBlockchain,
 )
@@ -63,6 +67,7 @@ class RotkiDataUpdater:
             UpdateType.CONTRACTS: self.update_contracts,
             UpdateType.GLOBAL_ADDRESSBOOK: self.update_global_addressbook,
             UpdateType.ACCOUNTING_RULES: self.update_accounting_rules,
+            UpdateType.LOCATION_ASSET_MAPPINGS: self.update_location_asset_mappings,
         }  # If we ever change this also change tests/unit/test_data_updates::reset_update_type_mappings  # noqa: E501
         self.version = get_current_version().our_version
 
@@ -326,6 +331,32 @@ class RotkiDataUpdater:
                 write_cursor=write_cursor,
                 entries=entries_to_add,
             )
+
+    def update_location_asset_mappings(self, data: dict[str, list[dict[str, Any]]], version: int) -> None:  # noqa: E501
+        """Applies location asset mappings updates in the global DB"""
+        log.info(f'Applying update for location asset mappings to v{version}')
+        for update_function, entry_type, raw_data_key in (
+            (GlobalDBHandler.add_location_asset_mappings, LocationAssetMappingUpdateEntry, 'additions'),  # noqa: E501
+            (GlobalDBHandler.update_location_asset_mappings, LocationAssetMappingUpdateEntry, 'updates'),  # noqa: E501
+            (GlobalDBHandler.delete_location_asset_mappings, LocationAssetMappingDeleteEntry, 'deletions'),  # noqa: E501
+        ):
+            entries, raw_data = [], data.get(raw_data_key)
+            if raw_data is None:
+                continue  # no data to update
+            for raw_entry in raw_data:
+                try:
+                    if (asset_id := raw_entry.get('asset')) is not None:
+                        raw_entry['asset'] = Asset(asset_id)
+                    if (raw_location := raw_entry.get('location')) is not None:
+                        raw_entry['location'] = Location.deserialize(raw_location)
+                    entries.append(entry_type.deserialize(raw_entry))  # type: ignore[attr-defined]  # deserialize is defined in both
+                except DeserializationError as e:
+                    log.error(f'Could not deserialize {entry_type.__name__} {raw_entry!s}: {e!s}')
+
+            try:
+                update_function(entries=entries)  # type: ignore[operator]  # update_function is known
+            except InputError as e:
+                log.error(f'Could not update {entry_type.__name__} due to {e!s}')
 
     def check_for_updates(self, updates: Sequence[UpdateType] = tuple(UpdateType)) -> None:
         """Retrieve the information about the latest available update"""
--- a/rotkehlchen/tests/api/test_location_asset_mappings.py
+++ b/rotkehlchen/tests/api/test_location_asset_mappings.py
@@ -1,5 +1,5 @@
 from http import HTTPStatus
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Final
 
 import requests
 
@@ -11,6 +11,8 @@ from rotkehlchen.tests.utils.api import (
 
 if TYPE_CHECKING:
     from rotkehlchen.api.server import APIServer
+
+NUM_ASSETS_MAPPINGS_V1_32: Final = 1500
 
 
 def _get_all_location_mappings(rotkehlchen_api_server: 'APIServer') -> Any:
@@ -28,7 +30,7 @@ def test_location_asset_mappings_query(rotkehlchen_api_server: 'APIServer') -> N
     """Test the location asset mappings API for querying the mappings."""
     # query all the mappings
     result = _get_all_location_mappings(rotkehlchen_api_server)
-    assert len(result['entries']) == result['entries_found'] == result['entries_total'] == 1500
+    assert len(result['entries']) == result['entries_found'] == result['entries_total'] == NUM_ASSETS_MAPPINGS_V1_32  # noqa: E501
 
     # query all common mappings
     response = requests.post(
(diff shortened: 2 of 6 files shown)


rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -553,3 +599,64 @@ def test_accounting_rules_updates(data_updater: RotkiDataUpdater) -> None:
...
+
+
+def test_location_asset_mappings_updates(
+        data_updater: RotkiDataUpdater,
+        globaldb: 'GlobalDBHandler',
+) -> None:
+    """Test that remote updates for location asset mappings work"""
+    # check state of the location asset mappings before updating
+    with globaldb.conn.read_ctx() as cursor:
+        assert cursor.execute('SELECT COUNT(*) FROM location_asset_mappings').fetchone()[0] == 1500
```
yabirgb, line 611: This constant will change when we add more right? Define a variable named

```python
NUM_ASSETS_MAPPINS_V1_32: Final = 1500
```

or similar and then 

```python
number_of_asset_mappings = assert cursor.execute('SELECT COUNT(*) FROM location_asset_mappings').fetchone()[0] 
number_of_asset_mappings >= 1500
```

rotkehlchen/types.py
```diff
@@ -849,9 +849,15 @@ class LocationAssetMappingDeleteEntry:
...
+        -DeserializationError if required keys are missing
         """
-        return cls(**data)
+        try:
+            return cls(
+                location=data['location'],
+                location_symbol=data['location_symbol'],
+            )
+        except KeyError as e:
+            raise DeserializationError(f'Missing key {e!s}') from e
```
yabirgb, line 860: Where is. this used? Don't you need to adjust the error handled in other pkace?

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -553,3 +599,64 @@ def test_accounting_rules_updates(data_updater: RotkiDataUpdater) -> None:
...
+                    Location.deserialize(addition['location']).serialize_for_db(),
+                    addition['location_symbol'],
+                ),
+            ).fetchone()[0] == addition['asset']
+
+        for update in LOCATION_ASSET_MAPPINGS_DATA['location_asset_mappings']['updates']:
+            assert cursor.execute(  # updates have been made
+                'SELECT local_id FROM location_asset_mappings WHERE location IS ? AND exchange_symbol IS ?',  # noqa: E501
+                (Location.deserialize(update['location']).serialize_for_db(), update['location_symbol']),  # noqa: E501
+            ).fetchone()[0] == update['asset']
```
yabirgb, line 656: can't you create an auxiliar function to avoid repeating the code in those checks? It could take an argument `check_before_update: bool`

and then you can do something like

```suggestion
        for update in LOCATION_ASSET_MAPPINGS_DATA['location_asset_mappings']['updates']:
            local_id = cursor.execute(  # updates have been made
                'SELECT local_id FROM location_asset_mappings WHERE location IS ? AND exchange_symbol IS ?',  # noqa: E501
                (Location.deserialize(update['location']).serialize_for_db(), update['location_symbol']),  # noqa: E501
            ).fetchone()[0]
            
            if check_before_update is True:
                 assert  == update['asset']
            else:
                assert != update['asset'] 
```

same for the delete check, you can do both in the same function and avoid repeating code.

yabirgb: I believe you can use XOR to do it in one assert line

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`2698` rotki will now remotely update asset symbols' mappings for different exchanges.
```
yabirgb, line 5: luki is already adding it here
image

Also the issue is not about remote updates but managing them right? Remove the changelog entry from here

rotkehlchen/db/updates.py
```diff
@@ -327,6 +332,32 @@ def update_global_addressbook(self, data: list[dict[str, Any]], version: int) ->
                 entries=entries_to_add,
             )
 
+    def update_location_asset_mappings(self, data: dict[str, list[dict[str, Any]]], version: int) -> None:  # noqa: E501
+        """Applies location asset mappings updates in the global DB"""
+        log.info(f'Applying update for location asset mappings to v{version}')
+        for update_function, entry_type, raw_data_key in (
```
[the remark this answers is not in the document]
OjusWiZard: for this, the min and max versions are checked in `.update_single` function.
image

rotkehlchen/types.py
OjusWiZard: handled them in `rotkehlchen/api/v1/schemas.py` and `rotkehlchen/db/updates.py`
it's only used in these two placed

rotkehlchen/tests/api/test_location_asset_mappings.py
```diff
@@ -12,6 +12,8 @@
 if TYPE_CHECKING:
     from rotkehlchen.api.server import APIServer
 
+NUM_ASSETS_MAPPINS_V1_32: Final = 1500
```
yabirgb, line 15: typo: 
```suggestion
NUM_ASSETS_MAPPINGS_V1_32: Final = 1500
```

rotkehlchen/tests/api/test_location_asset_mappings.py
```diff
@@ -28,7 +30,7 @@ def test_location_asset_mappings_query(rotkehlchen_api_server: 'APIServer') -> N
     """Test the location asset mappings API for querying the mappings."""
     # query all the mappings
     result = _get_all_location_mappings(rotkehlchen_api_server)
-    assert len(result['entries']) == result['entries_found'] == result['entries_total'] == 1500
+    assert len(result['entries']) == result['entries_found'] == result['entries_total'] == NUM_ASSETS_MAPPINS_V1_32  # noqa: E501
```
yabirgb, line 33:
```suggestion
    assert len(result['entries']) == result['entries_found'] == result['entries_total'] == NUM_ASSETS_MAPPINGS_V1_32  # noqa: E501
```

rotkehlchen/tests/api/test_location_asset_mappings.py
```diff
@@ -196,7 +198,7 @@ def test_location_asset_mappings_pagination(rotkehlchen_api_server: 'APIServer')
     )
     result = assert_proper_response_with_result(response=response)
     assert result['entries'] == [all_mappings[1]]
-    assert result['entries_found'] == 1500
+    assert result['entries_found'] == NUM_ASSETS_MAPPINS_V1_32
```
yabirgb, line 201:
```suggestion
    assert result['entries_found'] == NUM_ASSETS_MAPPINGS_V1_32
```

same in all the places where you use it

yabirgb (commented):
Just a typo

rotkehlchen/db/updates.py
```diff
@@ -327,6 +332,32 @@ def update_global_addressbook(self, data: list[dict[str, Any]], version: int) ->
...
+            (GlobalDBHandler.update_location_asset_mappings, LocationAssetMappingUpdateEntry, 'updates'),  # noqa: E501
+            (GlobalDBHandler.delete_location_asset_mappings, LocationAssetMappingDeleteEntry, 'deletions'),  # noqa: E501
+        ):
+            entries, raw_data = [], data.get(raw_data_key)
+            if raw_data is None:
+                continue  # no data to update
+            for raw_entry in raw_data:
+                try:
+                    if 'asset' in raw_entry:
+                        raw_entry['asset'] = Asset(raw_entry['asset'])
```
LefterisJP, line 349:
```suggestion
                    if (asset_id := raw_entry.get('asset')) is not None:
                        raw_entry['asset'] = Asset(asset_id)
```

rotkehlchen/db/updates.py
```diff
@@ -327,6 +332,32 @@ def update_global_addressbook(self, data: list[dict[str, Any]], version: int) ->
...
+                    if raw_entry['location'] is not None:
+                        raw_entry['location'] = Location.deserialize(raw_entry['location'])
```
LefterisJP, line 351: Do this only if location can be missing. If not then ignore

```suggestion
                    if (raw_location := raw_entry.get('location')) is not None:
                        raw_entry['location'] = Location.deserialize(raw_location)
```

rotkehlchen/db/updates.py
```diff
@@ -327,6 +332,32 @@ def update_global_addressbook(self, data: list[dict[str, Any]], version: int) ->
...
+                    entries.append(entry_type.deserialize(raw_entry))  # type: ignore[attr-defined]  # deserialize is defined in both
+                except (KeyError, DeserializationError) as e:
```
LefterisJP, line 353: Here could skip KeyError if location can be missing above as nothign else should raise key error with the suggestions I left

rotkehlchen/tests/api/test_location_asset_mappings.py
LefterisJP: +1

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -553,3 +603,52 @@ def test_accounting_rules_updates(data_updater: RotkiDataUpdater) -> None:
             'counterparty': 'test_counterparty',
         },
     ]
+
+
+def _check_location_asset_mappings(cursor: 'DBCursor', after_upgrade: bool) -> None:
+    """Auxiliar function to check the db values before and after the upgrade"""
```
LefterisJP, line 609: This is not Spanish :sweat_smile: 

```suggestion
    """Auxiliary function to check the db values before and after the upgrade"""
```

LefterisJP (approved):
LGTM. Failed test is flaky `rotkehlchen/tests/unit/test_polygon_pos_inquirer.py::test_polygon_pos_nodes_prune_and_archive_status`

PR #7613, opened 2024-03-14
Ensure that latest price is correctly queried for nfts

yabirgb:
the floor price can be given using an asset different than ETH and this was not correctly handled. Now we query the token and save the price using the correct asset.

**This PR requires a small frontend change and @lukicenturi  is already aware of that. Not including the changelog entry due to that**

### DIFF
--- a/rotkehlchen/externalapis/opensea.py
+++ b/rotkehlchen/externalapis/opensea.py
@@ -6,7 +6,9 @@ from typing import TYPE_CHECKING, Any, Final, Literal, NamedTuple
 
 import gevent
 import requests
+from eth_utils import to_checksum_address
 
+from rotkehlchen.assets.asset import Asset
 from rotkehlchen.assets.utils import get_or_create_evm_token
 from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
 from rotkehlchen.chain.ethereum.utils import asset_normalized_value
@@ -19,6 +21,7 @@ from rotkehlchen.errors.misc import RemoteError
 from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.externalapis.interface import ExternalServiceWithApiKey
 from rotkehlchen.fval import FVal
+from rotkehlchen.globaldb.handler import GlobalDBHandler
 from rotkehlchen.inquirer import Inquirer
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.serialization.deserialize import (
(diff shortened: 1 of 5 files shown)


rotkehlchen/externalapis/opensea.py
```diff
@@ -42,12 +45,45 @@
...
+
+    TODO: The api has a bug reported in the ticket #1455100 and the chain is reported
+    always as ethereum even if the asset is not an ethereum token.
+    When fixed adjust this function to use the chain.
+
+    May raise:
+    - UnknownAsset if the token wasn't found in the collection payment tokens.
+    """
+    for payment_token in collection_payment_tokens:
+        if payment_token['symbol'] == floor_price_symbol:
```
LefterisJP, line 64: This and the other key lookups can raise KeyError. You need to handle it or put it in the docstring this propagates upwards and handle it at the caller

rotkehlchen/externalapis/opensea.py
```diff
@@ -42,12 +45,45 @@
...
+            address = to_checksum_address(payment_token['address'])
+            with GlobalDBHandler().conn.read_ctx() as cursor:
+                asset_identifier = cursor.execute(
+                    'SELECT identifier FROM evm_tokens WHERE address=?',
```
LefterisJP, line 68: what if the address is more than once in the DB? Supposedly same address for different chains. It's not impossible. Current code will take the first one only.

So you consider it acceptable error until they fix on their side? If so write it here.

rotkehlchen/externalapis/opensea.py
```diff
@@ -57,6 +93,7 @@ def serialize(self) -> dict[str, Any]:
             'description': self.description,
             'large_image': self.large_image,
             'floor_price': str(self.floor_price) if self.floor_price else None,
+            'floor_price_asset': self.floor_price_asset.identifier if self.floor_price_asset else None,  # noqa: E501
```
LefterisJP, line 96: When can this be None? You initialize it as A_ETH and the type is `Asset`, not `Asset | None`

rotkehlchen/externalapis/opensea.py
```diff
@@ -240,14 +279,25 @@ def _deserialize_nft(
...
                         f'Could not find collection {entry["collection"]} in opensea collections '
                         f'endpoint',
                     )
 
-            price_in_eth = max(last_price_in_eth, floor_price)
-            price_in_usd = price_in_eth * eth_usd_price
+            last_price_in_usd = last_price_in_eth * eth_usd_price
+            floor_price_in_usd = floor_price * Inquirer.find_usd_price(asset=floor_price_asset) if floor_price != ZERO else ZERO  # noqa: E501
+
+            if floor_price_in_usd > last_price_in_usd:
```
LefterisJP, line 292: So this logic keeps the highest? It's what we have in the docs right? So we don't need to change anything in the description?

> An estimation of the value of the NFTs you own is counted into your total net worth. The estimation strategy is currently the maximum of either the floor price of the collection or the last sale of the NFT. If a manual price has been given this is always preferred.

rotkehlchen/externalapis/opensea.py
```diff
@@ -315,12 +366,30 @@ def gather_account_collections(self, account: ChecksumEvmAddress) -> None:
                 name='floor price',
                 location='opensea',
             )
+
+            if (
+                len(payment_tokens := collection.get('payment_tokens', [])) == 0 or
+                stats_result['total']['floor_price_symbol'] in ('', 'ETH')
```
LefterisJP, line 372: Both in your new code and some lines above this function does not seem to handle `KeyError` properly for `stats_result`. Please do so and if you decide to handle it at the caller side, note in the docstring that the function raises a KeyError

LefterisJP (commented):
Some comments. Also doesn't this require us to do any DB changes? Since you added the last price asset in the NFT class? But weirdly in develop that seems to exist in the DB? What gives?

rotkehlchen/externalapis/opensea.py
yabirgb: Yes, I considered it acceptable until we can correctly get it from opensea

rotkehlchen/externalapis/opensea.py
yabirgb: i forgot to change it when doing changes. I've removed it

rotkehlchen/externalapis/opensea.py
LefterisJP: > Yes, I considered it acceptable until we can correctly get it from opensea

So did you write it here in a comment? I told you to do so. Please read what I write. When we look at this code in the future the question will come up again.

LefterisJP (approved):
lgtm

PR #7615, opened 2024-03-14
Fix (7531): Prioritise manual prices above oracles

### LINKED ISSUE #7531: Manual price is ignored if asset has special price calculation
## Problem Definition

If the asset has a special price calculation such as for curve pools, then we don't get to go to the manual price oracle, so any manually input price is ignored. So if our special price calculation method is broken, then there is no way for a user to fix it.

Seen in 1.32.1.

## Task

Fix this and make sure manual price oracle is preferred above all other methods.

OjusWiZard:
Closes #7531

## Checklist

- [x] Query manual price entries in DB before using the oracles

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`7531` rotki will now prioritize manual prices above other price oracles and calculations.
 * :bug:`-` rotki will display NFT prices correctly across all asset types.
 * :bug:`7572` Fix issue where coinbase data query could result in an error popup.
 * :bug:`7611` USDT balances and trades in bitcoin.de should now work properly.
--- a/rotkehlchen/inquirer.py
+++ b/rotkehlchen/inquirer.py
@@ -5,7 +5,7 @@ from contextlib import suppress
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, NamedTuple, Optional, Union
 
-from rotkehlchen.assets.asset import Asset, EvmToken, FiatAsset, UnderlyingToken
+from rotkehlchen.assets.asset import Asset, AssetWithOracles, EvmToken, FiatAsset, UnderlyingToken
 from rotkehlchen.assets.utils import TokenEncounterInfo, get_or_create_evm_token
 from rotkehlchen.chain.ethereum.defi.price import handle_defi_price_query
 from rotkehlchen.chain.ethereum.utils import token_normalized_value_decimals
@@ -382,8 +382,8 @@ class Inquirer:
             "Oracles can't be empty or have repeated items"
         )
         instance = Inquirer()
-        instance._oracles = oracles
-        instance._oracle_instances = [getattr(instance, f'_{oracle!s}') for oracle in oracles]
+        instance._oracles = [oracle for oracle in oracles if oracle is not CurrentPriceOracle.MANUALCURRENT]  # noqa: E501  # TODO: remove MANUALCURRENT from the settings
+        instance._oracle_instances = [getattr(instance, f'_{oracle!s}') for oracle in instance._oracles]  # noqa: E501
         instance._oracles_not_onchain = []
         instance._oracle_instances_not_onchain = []
         for oracle, oracle_instance in zip(instance._oracles, instance._oracle_instances, strict=True):  # noqa: E501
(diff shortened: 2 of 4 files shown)


rotkehlchen/inquirer.py
```diff
@@ -648,7 +668,24 @@ def _find_usd_price(
...
+                to_asset=A_USD,
+                coming_from_latest_price=coming_from_latest_price,
+                match_main_currency=match_main_currency,
+            )
+            Inquirer.set_cached_price(cache_key=cache_key, cached_price=CachedPriceEntry(
+                price=price,
+                time=ts_now(),
+                oracle=CurrentPriceOracle.MANUALCURRENT,
+                used_main_currency=used_main_currency,
+            ))
```
yabirgb, line 685: can we add this to `_try_oracle_price_query`? If you check the other function we check if it is zero and if it is not then we log and add to the cache

rotkehlchen/inquirer.py
```diff
@@ -648,7 +668,24 @@ def _find_usd_price(
...
+            return price, CurrentPriceOracle.MANUALCURRENT, used_main_currency
+
+        # continue, price can be found by one of the oracles (CC for example)
```
yabirgb, line 689: No need for the extra space + this comment belongs right after querying the oracles, not here

rotkehlchen/inquirer.py
```diff
@@ -648,7 +668,24 @@ def _find_usd_price(
                 price, oracle = Inquirer._query_fiat_pair(base=asset, quote=Inquirer.usd)
                 return price, oracle, False
 
-        # continue, asset isn't fiat or a price can be found by one of the oracles (CC for example)
+        # continue, asset isn't fiat, check manual prices
+        if GlobalDBHandler.get_manual_current_price(asset=asset) is not None:
+            price, used_main_currency, _ = Inquirer._try_oracle_price_query(
```
yabirgb, line 673: Internally `Inquirer._manualcurrent` calls `GlobalDBHandler.get_manual_current_price(asset=asset)`. Should you just query `Inquirer._manualcurrent` and if the prize returned is not zero then return? Otherwise you are repeating a query

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`7531` rotki will now prioritize manual prices above other price oracles.
```
LefterisJP, line 5: Also mention other price calculations here.

rotkehlchen/tests/api/test_current_assets_price.py
```diff
@@ -318,8 +318,8 @@ def test_manual_current_prices_loop(inquirer):
 
 @pytest.mark.vcr()
 @pytest.mark.parametrize('ignore_mocked_prices_for', ['ETH'])
-def test_inquirer_oracles_affect_manual_price(inquirer):
-    """Checks that change of oracles order affects manual current price usage.
+def test_inquirer_oracles_does_not_affect_manual_price(inquirer):
+    """Checks that change of oracles order does not affects manual current price usage.
```
LefterisJP, line 322: Does not affect

rotkehlchen/inquirer.py
```diff
@@ -411,7 +461,6 @@ def _query_oracle_instances(
         `coming_from_latest_price` is used by manual latest price oracle to handle price loops.
         """
         instance = Inquirer()
```
LefterisJP, line 463: Why is that removed? Where did it go?

rotkehlchen/inquirer.py
```diff
@@ -648,7 +672,17 @@ def _find_usd_price(
                 price, oracle = Inquirer._query_fiat_pair(base=asset, quote=Inquirer.usd)
                 return price, oracle, False
 
-        # continue, asset isn't fiat or a price can be found by one of the oracles (CC for example)
+        # continue, asset isn't fiat, check manual prices
+        if (price_result := Inquirer._try_oracle_price_query(
```
LefterisJP, line 676: Hmm okay so you now run manual oracle before of o red it correctly ... But the manual oracle is reran in the for loop no? Didn't we say it should be removed during stand up?

rotkehlchen/inquirer.py
OjusWiZard: not sure, what you pointed but we moved some common code in `_try_oracle_price_query` function

LefterisJP: I was pointing the line below. The cache key ... damn mobile app

rotkehlchen/tests/api/test_current_assets_price.py
```diff
@@ -318,8 +318,8 @@ def test_manual_current_prices_loop(inquirer):
 
 @pytest.mark.vcr()
 @pytest.mark.parametrize('ignore_mocked_prices_for', ['ETH'])
-def test_inquirer_oracles_affect_manual_price(inquirer):
-    """Checks that change of oracles order affects manual current price usage.
+def test_inquirer_oracles_does_not_affect_manual_price(inquirer):
```
LefterisJP, line 321: Avove this it is failing in `rotkehlchen/tests/api/test_current_assets_price.py::test_manual_current_prices_loop` with 

```
AssertionError: assert FVal(0) == FVal(1570.92) + where FVal(1570.92) = FVal('1570.92')
```

LefterisJP:
Pushed a fix. 

```diff
modified   rotkehlchen/inquirer.py
@@ -435,6 +435,7 @@ class Inquirer:
                 f'Was not able to find price from {from_asset!s} to {to_asset!s} since your '
                 f'manual latest prices form a loop. For now, other oracles will be used.',
             )
+            is_error = False
 
         if (price, used_main_currency) != (ZERO_PRICE, False):
             Inquirer.set_cached_price(
```

I think that by moving the oracle price query out of the inquirer you ended up considering the manual price loop case as an error while it should not be and should be allowedto try other oracles.

Amended your commit with the diff above.

LefterisJP (approved):
LGTM. All seems green now

PR #7619, opened 2024-03-16
Support Coinbase staking_reward tx type

coinyon:
When digging into coinbase API responses, I found that rotki ignored the `staking_reward` tx type. This PR adds support in the same way as `inflation_reward` and `interest`.

As a related note, I don't know why all three (`interest`, `inflation_reward` and `staking_reward`) are reported with Event Subtype NONE where they could be reported as REWARD. If you want me to change this as well with this PR, let me know.

## Checklist

### DIFF
--- a/rotkehlchen/exchanges/coinbase.py
+++ b/rotkehlchen/exchanges/coinbase.py
@@ -641,7 +641,7 @@ class Coinbase(ExchangeInterface):
                 if (trade := self._process_coinbase_trade(event=transaction)):
                     trades.append(trade)
             elif (
-                    tx_type in ('interest', 'inflation_reward') or
+                    tx_type in ('interest', 'inflation_reward', 'staking_reward') or
                     (
                         tx_type == 'send' and 'from' in transaction and
                         'resource' in transaction['from'] and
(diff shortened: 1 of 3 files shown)


rotkehlchen/exchanges/coinbase.py
```diff
@@ -538,7 +538,7 @@ def _query_single_account_transactions(
             elif tx_type in ('buy', 'sell'):
                 self._process_normal_trade(event=transaction, trades=trades)
             elif (
-                    tx_type in ('interest', 'inflation_reward') or
```
LefterisJP, line 541: If this exists as a transaction type then it's a good addition.

Can you also edit the test at `test_coinbase_query_income_loss_expense` and add a mock event with this type and test it in the test?

LefterisJP (commented):
Left a comment about the change.

For the subtype, would need to check.

I think the interest/inflation_reward/staking_rewards can be like this.

But if you see the place you edited has another elif which is afaics (but I am not sure since I don't use coinbase and did not write the code) concerns internal movements/receivals. So that part should stay as is.

rotkehlchen/exchanges/coinbase.py
lordi: Will do!

coinyon: Done. This is how I get the data from the API, unfortunately without any notes

coinyon:
@LefterisJP The test failures seem unrelated

LefterisJP (approved):
LGTM

PR #7628, opened 2024-03-18
Refactor (2698): Move Bitfinex currency_map to DB

OjusWiZard:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=54770713

## Checklist

- [x] Use `location_asset_mappings` table of GlobalDB instead of in-memory `currency_map` for Bitfinex

### DIFF
--- a/rotkehlchen.spec
+++ b/rotkehlchen.spec
@@ -85,6 +85,7 @@ a = Entrypoint(
         ('rotkehlchen/data/populate_asset_collections.sql', 'rotkehlchen/data'),
         ('rotkehlchen/data/populate_multiasset_mappings.sql', 'rotkehlchen/data'),
         ('rotkehlchen/data/populate_location_asset_mappings.sql', 'rotkehlchen/data'),
+        ('rotkehlchen/data/populate_location_unsupported_assets.sql', 'rotkehlchen/data'),
         # TODO
         # We probably should have a better way to specify some data should be loaded
         # by a module in pyinstaller. Should be loaded dynamically by rotki and not
--- a/rotkehlchen/exchanges/bitfinex.py
+++ b/rotkehlchen/exchanges/bitfinex.py
@@ -16,6 +16,7 @@ from requests.adapters import Response
 
 from rotkehlchen.accounting.structures.balance import Balance
 from rotkehlchen.assets.converters import BITFINEX_EXCHANGE_TEST_ASSETS, asset_from_bitfinex
+from rotkehlchen.assets.utils import symbol_to_asset_or_token
 from rotkehlchen.constants import ZERO
 from rotkehlchen.errors.asset import UnknownAsset, UnsupportedAsset
 from rotkehlchen.errors.misc import RemoteError
@@ -84,12 +85,6 @@ class CurrenciesResponse(NamedTuple):
     success: bool
     response: Response
     currencies: list[str]
-
-
-class CurrencyMapResponse(NamedTuple):
-    success: bool
-    response: Response
-    currency_map: dict[str, str]
 
 
 class ExchangePairsResponse(NamedTuple):
@@ -478,10 +473,7 @@ class Bitfinex(ExchangeInterface):
                 f'Unexpected bitfinex movement with status: {raw_result[5]}. '
                 f'Only completed movements are processed. Raw movement: {raw_result}',
             )
-        fee_asset = asset_from_bitfinex(
-            bitfinex_name=raw_result[1],
-            currency_map=self.currency_map,
-        )
+        fee_asset = asset_from_bitfinex(bitfinex_name=raw_result[1])
 
         amount = deserialize_asset_amount(raw_result[12])
         category = (
@@ -541,18 +533,9 @@ class Bitfinex(ExchangeInterface):
                 f'Raw trade: {raw_result}',
             )
 
-        base_asset = asset_from_bitfinex(
-            bitfinex_name=bfx_base_asset_symbol,
-            currency_map=self.currency_map,
-        )
-        quote_asset = asset_from_bitfinex(
-            bitfinex_name=bfx_quote_asset_symbol,
-            currency_map=self.currency_map,
-        )
-        fee_asset = asset_from_bitfinex(
-            bitfinex_name=raw_result[10],
-            currency_map=self.currency_map,
-        )
+        base_asset = asset_from_bitfinex(bitfinex_name=bfx_base_asset_symbol)
+        quote_asset = asset_from_bitfinex(bitfinex_name=bfx_quote_asset_symbol)
+        fee_asset = asset_from_bitfinex(bitfinex_name=raw_result[10])
 
         trade = Trade(
             timestamp=Timestamp(int(raw_result[2] / 1000)),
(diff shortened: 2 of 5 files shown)


rotkehlchen/exchanges/bitfinex.py
```diff
@@ -654,23 +641,29 @@ def _query_currency_map(self) -> CurrencyMapResponse:
...
+                            )
+                    except UnknownAsset:
+                        log.info(f'Found new asset symbol {bfx_symbol} for {symbol} in Bitfinex. Support for it has to be added.')  # noqa: E501
 
-        return CurrencyMapResponse(
-            success=was_successful,
-            response=response,
-            currency_map=currency_map,
-        )
+        return CurrencyMapResponse(success=was_successful, response=response)
```
yabirgb, line 666: Can't you raise remoterror here and remove the usage where you call this method? This allows you to get rid of `CurrencyMapResponse`. Just be sure to use the same text that we have where you use this method

rotkehlchen/exchanges/bitfinex.py
```diff
@@ -654,23 +641,29 @@ def _query_currency_map(self) -> CurrencyMapResponse:
...
-                    cursor.execute(
-                        'SELECT exchange_symbol, local_id FROM location_asset_mappings WHERE location IS ? OR location IS NULL;',  # noqa: E501
-                        (Location.BITFINEX.serialize_for_db(),),
-                    )
-                    for bfx_symbol, asset_id in cursor:
-                        currency_map[bfx_symbol] = asset_id
+                test_assets = set(BITFINEX_EXCHANGE_TEST_ASSETS)
+                bitfinex_db_serialized = Location.BITFINEX.serialize_for_db()
+                for bfx_symbol, symbol in response_list[0]:
+                    with GlobalDBHandler().conn.read_ctx() as cursor:
```
yabirgb, line 647: instead of creating a cursor for each element of the loop create the cursor once

rotkehlchen/exchanges/bitfinex.py
```diff
@@ -654,23 +641,29 @@ def _query_currency_map(self) -> CurrencyMapResponse:
...
+                        if cursor.execute(
+                            'SELECT COUNT(*) FROM location_unsupported_assets WHERE location=? AND exchange_symbol=?',  # noqa: E501
+                            (bitfinex_db_serialized, bfx_symbol),
+                        ).fetchone()[0] > 0 or bfx_symbol in test_assets:
+                            continue
```
yabirgb, line 652: do the check `bfx_symbol in test_assets` first to avoid the extra trip to db

rotkehlchen/exchanges/bitfinex.py
```diff
@@ -654,23 +641,29 @@ def _query_currency_map(self) -> CurrencyMapResponse:
...
+
+                    try:
+                        with GlobalDBHandler().conn.write_ctx() as write_cursor:
+                            write_cursor.execute(
+                                'INSERT OR IGNORE INTO location_asset_mappings (location, exchange_symbol, local_id) VALUES (?, ?, ?);', (  # noqa: E501
+                                    bitfinex_db_serialized,
+                                    bfx_symbol,
+                                    symbol_to_asset_or_token(symbol).serialize(),
+                                ),
+                            )
```
yabirgb, line 662: I'm not sure to follow. Why do you need to do this? To check if any of the symbols can be directly mapped? Why not to do a executemany instead? ~Also shouldn't you try only symbols that are not in the db?~

rotkehlchen/exchanges/bitfinex.py
```diff
@@ -654,23 +641,29 @@ def _query_currency_map(self) -> CurrencyMapResponse:
...
+                        ).fetchone()[0] > 0 or bfx_symbol in test_assets:
+                            continue
+
+                    try:
+                        with GlobalDBHandler().conn.write_ctx() as write_cursor:
+                            write_cursor.execute(
+                                'INSERT OR IGNORE INTO location_asset_mappings (location, exchange_symbol, local_id) VALUES (?, ?, ?);', (  # noqa: E501
+                                    bitfinex_db_serialized,
+                                    bfx_symbol,
+                                    symbol_to_asset_or_token(symbol).serialize(),
```
yabirgb, line 660: It feels like doing `symbol_to_asset_or_token` it is skipping the known mappings for bitfinex and you are doing extra job that you shouldn't be doing? Am I missing something?

yabirgb (commented):
A test is failing. Also I'm not sure to be following the logic correctly. Left some questions

rotkehlchen/exchanges/bitfinex.py
```diff
@@ -89,7 +90,6 @@ class CurrenciesResponse(NamedTuple):
 class CurrencyMapResponse(NamedTuple):
     success: bool
     response: Response
-    currency_map: dict[str, str]
```
yabirgb, line 92: you can delete this class

rotkehlchen/exchanges/bitfinex.py
```diff
@@ -654,23 +641,42 @@ def _query_currency_map(self) -> CurrencyMapResponse:
...
             else:
-                currency_map = {
-                    bfx_symbol: symbol for bfx_symbol, symbol in response_list[0]
-                    if bfx_symbol not in set(BITFINEX_EXCHANGE_TEST_ASSETS)
-                }
-                with GlobalDBHandler().conn.read_ctx() as cursor:
-                    cursor.execute(
-                        'SELECT exchange_symbol, local_id FROM location_asset_mappings WHERE location IS ? OR location IS NULL;',  # noqa: E501
-                        (Location.BITFINEX.serialize_for_db(),),
+                # add the mappings in the globalDB
```
yabirgb, line 644: so this is done to add possible assets that have no conflict on the symbol and aren't already in the db right? cAn you state that in the comment?

rotkehlchen/exchanges/bitfinex.py
```diff
@@ -816,13 +822,7 @@ def first_connection(self) -> None:
...
             )
 
-        currency_map_response = self._query_currency_map()
-        if currency_map_response.success is False:
-            raise RemoteError(
-                f'bitfinex failed to request exchange currency map. '
-                f'Response status code: {currency_map_response.response.status_code}. '
-                f'Response text: {currency_map_response.response.text}.',
-            )
+        self._query_currency_map()
```
yabirgb, line 825: Add to the docstring that this function raises remoteError

rotkehlchen.spec
```diff
@@ -85,6 +85,7 @@ a = Entrypoint(
         ('rotkehlchen/data/populate_asset_collections.sql', 'rotkehlchen/data'),
         ('rotkehlchen/data/populate_multiasset_mappings.sql', 'rotkehlchen/data'),
         ('rotkehlchen/data/populate_location_asset_mappings.sql', 'rotkehlchen/data'),
+        ('rotkehlchen/data/populate_location_unsupported_assets.sql', 'rotkehlchen/data'),
```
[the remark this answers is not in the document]
OjusWiZard: added it in a separate commit

LefterisJP (approved):
LGTM

PR #7639, opened 2024-03-20
Add zksync lite transactions query, balance query and address addition API

LefterisJP:
- [X] Transactions query API endpoint: `/blockchains/evmlike/transactions`
- [X] Balances query API: The balances one with chain=zksync_lite
- [x] Address addition/edit/delete API: The standard one with chain = zksync_lite


LefterisJP:
All green finally. And transactions, balance and accounts query/addition/editing should be done (hopefully) from the backend.

PR #7652, opened 2024-03-21
Handle kraken new yield bearing product balances

LefterisJP:
Kraken balances in new yield-bearing products and automatic staking for new kraken users should now appear properly.


LefterisJP:
This was actually already done in https://github.com/rotki/rotki/pull/7526 but not merged for over a month. Merging to handle the case in next patch release.

PR #7654, opened 2024-03-22
Feat (Aave): Add v3 decoder

OjusWiZard:
## Checklist

- [x] Add Aave v3 decoder for mainnet


LefterisJP (commented):
Left some comments.

Also why the heck did this test fail:

```
rotkehlchen/tests/unit/test_ethereum_inquirer.py::test_ethereum_nodes_prune_and_archive_status[ethereum_manager_connect_at_start0] - AssertionError: assert 0 == (4 - 1)
```

I thought they were VCRed?

rotkehlchen/chain/ethereum/modules/aave/constants.py
```diff
@@ -1,3 +1,40 @@
...
+
+
+class AaveLabel(Enum):
+    """Protocol label constants for the AAVE protocol."""
+    AAVE_LABEL: Final = 'Aave'
+    AAVE_V2_LABEL: Final = 'AAVE v2'
+    AAVE_V3_LABEL: Final = 'AAVE v3'
+
+
+class AaveCpt(Enum):
```
LefterisJP, line 18: If you wanted to use an enum, why not use a `StrEnum` here and below? https://docs.python.org/3/library/enum.html#enum.StrEnum

LefterisJP (approved):
LGTM.

Failed test is `test_get_contract_deployed_block` due to remote error

PR #7659, opened 2024-03-24
Add a test that dead tasks log traceback

LefterisJP:
Just a test that the functionality works. The test works. Problem is (probably) some greenlets for some errors can have no traceback.


LefterisJP:
Same as the previous PR. Only error is gnosis scan api down:

```
rotkehlchen/tests/unit/decoders/test_1inch.py::test_1inch_gnosis_v5_swap[
```

PR #7660, opened 2024-03-24
Clarify logs for no price and etherscan

LefterisJP:
- Failure to _get_historical_assets_price via the API should just log a warning. Not error.
- etherscan logs add the chain to missing logs

Makes it easier to read logs


LefterisJP:
Only remote errors here

Issue #4663, opened 2022-08-06
Support Coinbase Advanced Trade


coinyon:
## Abstract

Current Coinbase implementation in Rotki does not fetch trades done in the new "Advanced Trade" tool.

## Motivation

Coinbase phases out "Coinbase Pro" and will replace it wilth "Advanced Trade". While there is no timeline for the migration, Advanced Trade is already live and possibly in use by Rotki users.

## Specification

For this to be implemented, we probably need to wait to an API spec from Coinbase. See [this thread](https://twitter.com/LefterisJP/status/1555983232239390728).

LefterisJP:
Thanks for bringing this to our attention @coinyon. We will wait until coinbase pro gives more information on advance trading and creates an api for it. We also need info on how to query CBPro history once it's sunset and becomes part of "advance trading".

The blog post by coinbase is here: https://blog.coinbase.com/hello-advanced-trade-goodbye-coinbase-pro-5b0715b03ef0
Ccxt, the most well know exchange integration library, also has the same open questions so it's something I guess we all are waiting on. https://github.com/ccxt/ccxt/issues/14571

LefterisJP:
Pushing this to 1.27.0

Got this email. I don't like this email.
![2022-10-28_18-02](https://user-images.githubusercontent.com/1658405/198682178-409c66c2-c98a-40be-96b0-0afb81407597.png)

pottertech:
[This](https://www.nerdwallet.com/article/investing/coinbase-vs-coinbase-pro) article indicates that Coinbase Pro is shutting down on November 9, 2022. If you have funds on Coinbase Pro will be moved into the new "Advanced Trade" automatically.

The issue still remains that Coinbase has not provided any indication of when an API would be made available for us to integrate into ccxt.

LefterisJP:
related post showing that nothing is yet clear: https://twitter.com/LefterisJP/status/1588237926634332161

LefterisJP:
New update: https://forums.coinbasecloud.dev/t/about-the-advanced-trade-api-category/1743

LefterisJP:
Api docs: https://docs.cloud.coinbase.com/advanced-trade-api/docs/welcome

coinyon:
I think you do not need to use the "Advanced Trade API". Advanced trade fills do appear in the normal "transactions" endpoint

LefterisJP:
All right then we close this :)

PR #7668, opened 2024-03-25
Support transaction type `advanced_trade_fill` on Coinbase

FrederikBolding:
Relevant to https://github.com/rotki/rotki/issues/4663

Adds support for the type `advanced_trade_fill` returned by the Coinbase API.

This turned out to be more complicated than anticipated and thus warranted an additional function `_process_advanced_trade`. There are multiple differences in how these are processed compared to the previous implementation:
- Each side of an `advanced_trade_fill` seems to be returned from different accounts with different IDs, so correlating them seems difficult. Instead I chose to skip one of them.
- Both sides of the trade are not obvious by looking at just the data returned, instead we attempt to parse this from the `product_id` (e.g. `ETH-USDC`)
- Since both sides of the trade are not represented this also has reduced precision in pricing. The Coinbase API only seems to return the rate with up to 3 digits of decimal precision.

Also note that a few assumptions were made in how `BUY` orders work as I couldn't find a good example of the API return types for that and I didn't have any of my own data to test with.

## Checklist

### DIFF
--- a/rotkehlchen/exchanges/coinbase.py
+++ b/rotkehlchen/exchanges/coinbase.py
@@ -535,8 +535,9 @@ class Coinbase(ExchangeInterface):
                         f'Transaction of type trade doesnt have the '
                         f'expected structure {transaction}',
                     )
-            elif tx_type in ('buy', 'sell'):
-                self._process_normal_trade(event=transaction, trades=trades)
+            elif tx_type in ('buy', 'sell', 'advanced_trade_fill'):
+                if (trade := self._process_coinbase_trade(event=transaction)):
+                    trades.append(trade)
             elif (
                     tx_type in ('interest', 'inflation_reward') or
                     (
@@ -610,13 +611,10 @@ class Coinbase(ExchangeInterface):
                 )
                 continue
 
-    def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> None:
-        """Turns a coinbase transaction into a rotki trade and adds it to the list.
+    def _process_coinbase_trade(self, event: dict[str, Any]) -> Trade | None:
+        """Turns a coinbase transaction into a rotki trade and returns it.
 
-        Sometimes the amounts can be negative which breaks rotki's logic which is why we use abs().
-
-        https://docs.cloud.coinbase.com/sign-in-with-coinbase/docs/api-transactions#transaction-types
-        If the coinbase transaction is not a trade related transaction nothing happens.
+        Uses underlying trade processing functions based on the type of trade.
         """
         try:
             if event['status'] != 'completed':
@@ -632,32 +630,12 @@ class Coinbase(ExchangeInterface):
                 raw_time = event['payout_at']
 
             timestamp = deserialize_timestamp_from_date(raw_time, 'iso8601', 'coinbase')
-            trade_type = TradeType.deserialize(event['type'])
-            amount = AssetAmount(abs(deserialize_asset_amount(event['amount']['amount'])))
-            tx_asset = asset_from_coinbase(event['amount']['currency'], time=timestamp)
-            native_amount = abs(deserialize_asset_amount(event['native_amount']['amount']))
-            native_asset = asset_from_coinbase(event['native_amount']['currency'], time=timestamp)
-            # rate is how much you get/give in quotecurrency if you buy/sell 1 unit of basecurrency
-            rate = Price(native_amount / amount)
 
-            fee_amount = fee_asset = None
-            if 'fee' in event:
-                fee_amount = abs(deserialize_fee(event['fee']['amount']))
-                fee_asset = asset_from_coinbase(event['fee']['currency'], time=timestamp)
+            if event['type'] in ('buy', 'sell'):
+                return self._process_normal_trade(event, timestamp)
+            elif event['type'] == 'advanced_trade_fill':
+                return self._process_advanced_trade(event, timestamp)
 
-            trades.append(Trade(
-                timestamp=timestamp,
-                location=Location.COINBASE,
-                # in coinbase you are buying/selling tx_asset for native_asset
-                base_asset=tx_asset,
-                quote_asset=native_asset,
-                trade_type=trade_type,
-                amount=amount,
-                rate=rate,
-                fee=fee_amount,  # type: ignore  # abs() doesn't propagate Fee type
-                fee_currency=fee_asset,
-                link=str(event['id']),
-            ))
         except UnknownAsset as e:
             self.msg_aggregator.add_warning(
                 f'Found coinbase transaction with unknown asset '
@@ -668,7 +646,7 @@ class Coinbase(ExchangeInterface):
                 f'Found coinbase trade with unsupported asset '
                 f'{e.identifier}. Ignoring it.',
             )
-        except (DeserializationError, KeyError) as e:
+        except (DeserializationError, KeyError, IndexError, ZeroDivisionError) as e:
             msg = str(e)
             if isinstance(e, KeyError):
                 msg = f'Missing key entry for {msg}.'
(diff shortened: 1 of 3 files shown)


rotkehlchen/exchanges/coinbase.py
```diff
@@ -684,6 +686,98 @@ def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> N
...
+            if event.get('created_at') is not None:
+                raw_time = event['created_at']
+            else:
+                raw_time = event['payout_at']
+
+            timestamp = deserialize_timestamp_from_date(raw_time, 'iso8601', 'coinbase')
+            trade_type = TradeType.deserialize(event['advanced_trade_fill']['order_side'])
+
+            if trade_type not in (TradeType.BUY, TradeType.SELL):
+                return None
```
yabirgb, line 712: This part is common to both functions and can be moved to a third function that returns the type, the timestamp and the amount.

I would create a function `_process_coinbase_trade` and based on the type it executes the common logic and passes it to `_process_advanced_trade` or `_process_normal_trade` and it returns the Trade object generated by the respective of them

rotkehlchen/exchanges/coinbase.py
```diff
@@ -684,6 +686,98 @@ def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> N
...
+
+            # If the trade is a buy instead, flip the assets.
+            if trade_type == TradeType.BUY:
+                quote_asset = tx_asset
+                base_asset = other_side_asset
+
+            rate = Price(event['advanced_trade_fill']['fill_price'])
+
+            fee_amount = abs(deserialize_fee(event['advanced_trade_fill']['commission']))
+            fee_asset = base_asset
```
yabirgb, line 741: Is the comission key always present? In the normal trade It seems to assume that it is not always there

rotkehlchen/exchanges/coinbase.py
```diff
@@ -684,6 +686,98 @@ def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> N
...
+            if ((trade_type == TradeType.SELL and amount > 0) or
+                    (trade_type == TradeType.BUY and amount < 0)):
+                return None
+
+            tx_asset_identifier = event['amount']['currency']
+            tx_asset = asset_from_coinbase(tx_asset_identifier, time=timestamp)
+            asset_identifiers = event['advanced_trade_fill']['product_id'].split('-')
+            other_side_identifier = (asset_identifiers[0]
+                                     if asset_identifiers[0] != tx_asset_identifier
+                                     else asset_identifiers[1])
```
yabirgb, line 727: Can you catch a possible indexError? If the product for some reason doesn't follow the `ASSET1-ASSET2` format the split will return a list with a single element and won't be correctly indexed. Also not sure if there are symbols with a `-`, something like `BTC-L`. Can you add a check to ensure that you have the expected number of symbols and if not log an error and exit?

rotkehlchen/tests/exchanges/test_coinbase.py
```diff
@@ -860,6 +860,218 @@ def test_asset_conversion_choosing_fee_asset():
...
+    assert len(trades) == 4
+    expected_trades = [Trade(
+        timestamp=1709799171,
+        location=Location.COINBASE,
+        base_asset=A_USDC,
+        quote_asset=A_EUR,
+        trade_type=TradeType.SELL,
+        amount=FVal('192.790000'),
+        rate=FVal('0.9174'),
+        fee=FVal('0'),
```
yabirgb, line 1035: we have a constant `ZERO` that you can use instaed of `FVal(0)`. Use it here and in the other places

rotkehlchen/tests/exchanges/test_coinbase.py
```diff
@@ -860,6 +860,218 @@ def test_asset_conversion_choosing_fee_asset():
...
+                "order_side": "sell"
+            },
+            "details": {
+                "title": "Filled USDC on sell",
+                "subtitle": null,
+                "header": null,
+                "health": "positive"
+            },
+            "hide_native_amount": false
+        },
```
yabirgb, line 900: can you add the other side of a buy/sell to check that it correctly gets skipped? Also to have it documented with an example

rotkehlchen/exchanges/coinbase.py
```diff
@@ -684,6 +686,98 @@ def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> N
...
+            timestamp = deserialize_timestamp_from_date(raw_time, 'iso8601', 'coinbase')
+            trade_type = TradeType.deserialize(event['advanced_trade_fill']['order_side'])
+
+            if trade_type not in (TradeType.BUY, TradeType.SELL):
+                return None
+
+            # Notice that we do not use abs() yet
+            amount = deserialize_asset_amount(event['amount']['amount'])
+
+            # Each trade has two sides, we ignore one of them
```
yabirgb, line 717:
```suggestion
            # Each trade has two sides, we ignore one of them
```

remove the new line

rotkehlchen/tests/exchanges/test_coinbase.py
```diff
@@ -860,6 +860,218 @@ def test_asset_conversion_choosing_fee_asset():
...
+        if 'transaction' in url:
+            if 'next-page' in url:
+                return MockResponse(200, TRANSACTIONS_RESPONSE)
+            # else
+            return MockResponse(200, mock_transactions_response)
+        if 'accounts' in url:
+            # keep it simple just return a single ID and ignore the rest of the fields
+            return MockResponse(200, '{"data": [{"id": "5fs23", "updated_at": "2020-06-08T02:32:16Z"}]}')  # noqa: E501
+        # else
+        raise AssertionError(f'Unexpected url {url} for test')
```
yabirgb, line 1013: can you define this as a `_function` in the file and use it in both places? Probably with a better name and a docstring explaining what it does

rotkehlchen/tests/exchanges/test_coinbase.py
```diff
@@ -860,6 +860,218 @@ def test_asset_conversion_choosing_fee_asset():
...
+            start_ts=0,
+            end_ts=ts_now(),
+            only_cache=False,
+        )
+
+    warnings = coinbase.msg_aggregator.consume_warnings()
+    errors = coinbase.msg_aggregator.consume_errors()
+    assert len(warnings) == 0
+    assert len(errors) == 0
+    assert len(trades) == 4
```
yabirgb, line 1026: this check is not needed since you already check for all the entries in the list

yabirgb (commented):
Thanks! The approach ignoring one of the sides seems correct

rotkehlchen/exchanges/coinbase.py
FrederikBolding: I haven't seen any instances of it not being present. It seems to be set to 0 when not applicable.

rotkehlchen/exchanges/coinbase.py
```diff
@@ -535,8 +535,10 @@ def _query_single_account_transactions(
                         f'Transaction of type trade doesnt have the '
                         f'expected structure {transaction}',
                     )
-            elif tx_type in ('buy', 'sell'):
-                self._process_normal_trade(event=transaction, trades=trades)
+            elif tx_type in ('buy', 'sell', 'advanced_trade_fill'):
+                trade = self._process_coinbase_trade(event=transaction)
+                if trade:
+                    trades.append(trade)
```
yabirgb, line 541: a bit of a nitpick but this can be done now in 3.11
```suggestion
                if (trade := self._process_coinbase_trade(event=transaction)):
                    trades.append(trade)
```

rotkehlchen/exchanges/coinbase.py
```diff
@@ -684,6 +663,97 @@ def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> N
...
         return None
 
+    def _process_normal_trade(self, event: dict[str, Any], timestamp: Timestamp) -> Trade | None:
+        """Turns a normal coinbase transaction into a rotki trade and returns it.
+
+        Sometimes the amounts can be negative which breaks rotki's logic which is why we use abs().
+
+        https://docs.cloud.coinbase.com/sign-in-with-coinbase/docs/api-transactions#transaction-types
+        If the coinbase transaction is not a trade related transaction nothing happens.
+        """
```
yabirgb, line 673: can you add to the docstring the errors that those functions can raise? It is deserializationError, KeyError, UnknownAsset and UnsupportedAsset

rotkehlchen/exchanges/coinbase.py
```diff
@@ -684,6 +663,97 @@ def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> N
...
+        trade_type = TradeType.deserialize(event['type'])
+        amount = AssetAmount(abs(deserialize_asset_amount(event['amount']['amount'])))
+        tx_asset = asset_from_coinbase(event['amount']['currency'], time=timestamp)
+        native_amount = abs(deserialize_asset_amount(event['native_amount']['amount']))
+        native_asset = asset_from_coinbase(event['native_amount']['currency'], time=timestamp)
+        # rate is how much you get/give in quotecurrency if you buy/sell 1 unit of basecurrency
+        rate = Price(native_amount / amount)
```
yabirgb, line 680: It shouldn't happen but I just noticed that this can raise DivisionByZeroError. Can you also catch it? I recall one exchange raising it once 🙈

rotkehlchen/exchanges/coinbase.py
```diff
@@ -684,6 +663,97 @@ def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> N
...
+        if ((trade_type == TradeType.SELL and amount > 0) or
+                (trade_type == TradeType.BUY and amount < 0)):
+            return None
+
+        tx_asset_identifier = event['amount']['currency']
+        tx_asset = asset_from_coinbase(tx_asset_identifier, time=timestamp)
+        asset_identifiers = event['advanced_trade_fill']['product_id'].split('-')
+
+        if len(asset_identifiers) != 2:
+            return None
```
yabirgb, line 724: can you add a log error here? Otherwise it is harder to debug later

rotkehlchen/exchanges/coinbase.py
```diff
@@ -684,6 +663,97 @@ def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> N
...
+
+        other_side_identifier = (asset_identifiers[0]
+                                 if asset_identifiers[0] != tx_asset_identifier
+                                 else asset_identifiers[1])
+        other_side_asset = asset_from_coinbase(other_side_identifier, time=timestamp)
+
+        base_asset = tx_asset
+        quote_asset = other_side_asset
+
```
yabirgb, line 733:
```suggestion
        base_asset, quote_asset = tx_asset, other_side_asset
```

rotkehlchen/exchanges/coinbase.py
```diff
@@ -684,6 +663,97 @@ def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> N
...
+            fee_currency=fee_asset,
+            link=str(event['id']),
+        )
+
+    def _process_advanced_trade(self, event: dict[str, Any], timestamp: Timestamp) -> Trade | None:
+        """Turns an advanced_trade_fill transaction into a rotki trade returns it.
+
+        https://docs.cloud.coinbase.com/sign-in-with-coinbase/docs/api-transactions#transaction-types
+        If the coinbase transaction is not a trade related transaction nothing happens.
+        """
```
yabirgb, line 706: this one in addition to the previous errors can also raise indexError. Please add it to the docstring since it helps later to identify unhandled errors

yabirgb (commented):
Thank you very much! Can you add also an entry to the changelog mentioning that this kind of trades are now handled in coinbase?

https://github.com/rotki/rotki/blob/develop/docs/changelog.rst

rotkehlchen/exchanges/coinbase.py
```diff
@@ -684,6 +662,107 @@ def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> N
...
+                                 if asset_identifiers[0] != tx_asset_identifier
+                                 else asset_identifiers[1])
+        other_side_asset = asset_from_coinbase(other_side_identifier, time=timestamp)
+
+        base_asset, quote_asset = tx_asset, other_side_asset
+
+        rate = Price(event['advanced_trade_fill']['fill_price'])
+
+        fee_amount = abs(deserialize_fee(event['advanced_trade_fill']['commission']))
+        fee_asset = quote_asset
```
yabirgb, line 751: nitpick: if variables are used only once there is no need to create a variable

yabirgb:
Will ping left to take a final look

rotkehlchen/exchanges/coinbase.py
```diff
@@ -684,6 +662,107 @@ def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> N
...
+        trade_type = TradeType.deserialize(event['advanced_trade_fill']['order_side'])
+
+        if trade_type not in (TradeType.BUY, TradeType.SELL):
+            return None
+
+        # Notice that we do not use abs() yet
+        amount = deserialize_asset_amount(event['amount']['amount'])
+        # Each trade has two sides, we ignore one of them
+        if ((trade_type == TradeType.SELL and amount > 0) or
+                (trade_type == TradeType.BUY and amount < 0)):
```
LefterisJP, line 727: style nit

```suggestion
        if (
            (trade_type == TradeType.SELL and amount > 0) or
            (trade_type == TradeType.BUY and amount < 0)
        ):
```

rotkehlchen/exchanges/coinbase.py
```diff
@@ -684,6 +662,107 @@ def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> N
...
+            return None
+
+        other_side_identifier = (asset_identifiers[0]
+                                 if asset_identifiers[0] != tx_asset_identifier
+                                 else asset_identifiers[1])
+        other_side_asset = asset_from_coinbase(other_side_identifier, time=timestamp)
+
+        base_asset, quote_asset = tx_asset, other_side_asset
+
+        rate = Price(event['advanced_trade_fill']['fill_price'])
```
LefterisJP, line 748:
```suggestion
        base_asset, quote_asset = tx_asset, other_side_asset
        rate = Price(event['advanced_trade_fill']['fill_price'])
```

rotkehlchen/exchanges/coinbase.py
```diff
@@ -684,6 +662,107 @@ def _process_normal_trade(self, event: dict[str, Any], trades: list[Trade]) -> N
...
+
```
LefterisJP, line 749:
```suggestion
```

LefterisJP (approved):
Looks okay. Got some styling comments but since this is a PR from an external won't press.

But this should go to bugfixes branch as it's a small adjustment to existing logic and could also get released on a patch release.

LefterisJP (approved):
LGTM. Failign test is a remote error hiccup in the tests.

LefterisJP:
Thank you @FrederikBolding

PR #7669, opened 2024-03-25
Fix (count_with_limit): Make grouping configurable

OjusWiZard:
## Checklist

- [x] For counting with limit, only do the group by sub-query if `group_by_event_ids` is True.

### DIFF
--- a/rotkehlchen/db/history_events.py
+++ b/rotkehlchen/db/history_events.py
@@ -628,12 +628,15 @@ class DBHistoryEvents:
             )
 
             free_query_group_by = maybe_filter_ignore_asset(query_filter)
+            group_by_query = ''
+            if group_by_event_ids:
+                group_by_query = 'GROUP BY event_identifier ORDER BY timestamp DESC, sequence_index ASC LIMIT ?'  # noqa: E501
+                bindings.insert(0, entries_limit)  # add limit's binding before prepared_query's bindings  # noqa: E501
             count_with_limit = cursor.execute(
                 f'SELECT COUNT(*) FROM ('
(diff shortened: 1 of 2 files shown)


rotkehlchen/db/history_events.py
```diff
@@ -628,12 +628,15 @@ def get_history_events_count(
             )
 
             free_query_group_by = maybe_filter_ignore_asset(query_filter)
+            group_by_query = ''
+            if group_by_event_ids:
+                group_by_query = 'GROUP BY event_identifier ORDER BY timestamp DESC, sequence_index ASC LIMIT ?'  # noqa: E501
```
yabirgb, line 633: how does this change affect the frontend? Does it break something in their assumptions? It should be documented in the PR description if it is changing the default behaviour

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -391,7 +391,7 @@ def test_get_events(rotkehlchen_api_server: 'APIServer'):
         ),
     )
     result = assert_proper_response_with_result(response)
-    assert result['entries_found'] == 6
+    assert result['entries_found'] == 9
```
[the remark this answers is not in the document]
OjusWiZard: it's there just below this one, on the line 418.

rotkehlchen/db/history_events.py
OjusWiZard: I tested it by running the app, and all the paginated events were showing as expected. Also, I've asked the frontend devs, if they ever use `group_by_event_ids: false` on this API.

OjusWiZard: @tewshi confirmed that it's not used.

LefterisJP (approved):
LGTM

PR #7677, opened 2024-03-27
Add metadata for when an airdrop is missing a decoder

yabirgb:
a new key in the index file has been introduced that lists the airdrops for which we don't have a decoder. In the api we will return a new attribute `missing_decoder` of type bool always true when the airdrop doesn't have a decoder.

Part of https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=53148627

### DIFF
--- a/rotkehlchen/chain/ethereum/airdrops.py
+++ b/rotkehlchen/chain/ethereum/airdrops.py
@@ -135,9 +135,10 @@ def _parse_airdrops(database: 'DBHandler', airdrops_data: dict[str, Any]) -> dic
     return airdrops
 
 
-def fetch_airdrops_metadata(database: 'DBHandler') -> tuple[dict[str, Airdrop], dict[str, list[str]]]:  # noqa: E501
+def fetch_airdrops_metadata(database: 'DBHandler') -> tuple[dict[str, Airdrop], dict[str, list[str]], set[str]]:  # noqa: E501
     """Fetches airdrop metadata from the rotki/data repository. If it's not cached, parses them,
     and returns them in two parts: a dict of Airdrop instances and a dict of POAP airdrops data.
+    Also returns the list of airdrops that don't have decoders yet.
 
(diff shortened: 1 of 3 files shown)


rotkehlchen/chain/ethereum/airdrops.py
```diff
@@ -135,9 +135,10 @@ def _parse_airdrops(database: 'DBHandler', airdrops_data: dict[str, Any]) -> dic
     return airdrops
 
 
-def fetch_airdrops_metadata(database: 'DBHandler') -> tuple[dict[str, Airdrop], dict[str, list[str]]]:  # noqa: E501
+def fetch_airdrops_metadata(database: 'DBHandler') -> tuple[dict[str, Airdrop], dict[str, list[str]], set[str]]:  # noqa: E501
     """Fetches airdrop metadata from the rotki/data repository. If it's not cached, parses them,
     and returns them in two parts: a dict of Airdrop instances and a dict of POAP airdrops data.
+    Also returns the list of airdrops that don't have decoders yet provided by the remote index.
```
LefterisJP, line 141: Provided by the remote index? What do you mean?

```suggestion
    Also returns the list of airdrops that don't have decoders yet provided by the remote index.
```

rotkehlchen/chain/ethereum/airdrops.py
```diff
@@ -386,6 +388,8 @@ def check_airdrops(
                     'link': airdrop_data.url,
                     'claimed': False,
                 }
+                if protocol_name in airdrops_without_decoder:
+                    temp_found_data[addr][protocol_name]['missing_decoder'] = True  # type: ignore
```
LefterisJP, line 392: not for this PR but we probably should change the way data is gathered here. We are repeating `asset`, `link` and now `missing_decoder` for each address of a given airdrop

LefterisJP (approved):
lgtm

PR #7683, opened 2024-03-28
Add API to manage user calendar

yabirgb:
Closes #(issue_number)

## Checklist


rotkehlchen/api/rest.py
```diff
@@ -4871,3 +4873,65 @@ def remove_token_from_spam(self, token: EvmToken) -> Response:
...
         return api_response(OK_RESULT, status_code=HTTPStatus.OK)
+
+    def create_calendar_entry(
+            self,
+            name: str,
+            timestamp: int,
+            description: str,
+            counterparty: str | None,
+            address: BlockchainAddress | None,
+    ) -> Response:
```
LefterisJP, line 4884: docstring

rotkehlchen/api/rest.py
```diff
@@ -4871,3 +4873,65 @@ def remove_token_from_spam(self, token: EvmToken) -> Response:
         self.rotkehlchen.data.remove_ignored_assets(assets=[token])
         AssetResolver().clean_memory_cache(token.identifier)
         return api_response(OK_RESULT, status_code=HTTPStatus.OK)
+
+    def create_calendar_entry(
+            self,
+            name: str,
+            timestamp: int,
```
LefterisJP, line 4880: This should be `Timestamp` type

rotkehlchen/api/rest.py
```diff
@@ -4871,3 +4873,65 @@ def remove_token_from_spam(self, token: EvmToken) -> Response:
...
+            description=description,
+            counterparty=counterparty,
+            address=address,
+        )
+        return api_response(_wrap_in_ok_result(
+            {'entry_id': calendar_event_id}),
+            status_code=HTTPStatus.OK,
+        )
+
+    def delete_calendar_entry(self, identifier: int) -> Response:
```
LefterisJP, line 4897: small docstring for all of those

rotkehlchen/api/rest.py
```diff
@@ -4871,3 +4873,65 @@ def remove_token_from_spam(self, token: EvmToken) -> Response:
...
+        )
+        return api_response(_wrap_in_ok_result(
+            result=process_result(result),
+            status_code=HTTPStatus.OK,
+        ))
+
+    def update_calendar_entry(
+            self,
+            name: str,
+            timestamp: int,
```
LefterisJP, line 4916:
```suggestion
            timestamp: Timestamp,
```

rotkehlchen/api/rest.py
```diff
@@ -4871,3 +4873,65 @@ def remove_token_from_spam(self, token: EvmToken) -> Response:
...
+            description: str,
```
LefterisJP, line 4917: Isn't description also optional?

rotkehlchen/api/v1/resources.py
```diff
@@ -3243,3 +3249,59 @@ def post(self, token: EvmToken) -> Response:
...
+    delete_schema = DeleteCalendarEntrySchema()
+    update_schema = UpdateCalendarSchema()
+    query_schema = QueryCalendarSchema()
+
+    @require_loggedin_user()
+    @use_kwargs(create_schema, location='json_and_query')
+    def put(
+            self,
+            name: str,
+            timestamp: int,
```
LefterisJP, line 3266: same questions for all of these. Type of `Timestamp` and optionality of description

rotkehlchen/api/v1/schemas.py
```diff
@@ -3531,3 +3532,55 @@ class FalsePositveSpamTokenSchema(Schema):
 
 class SpamTokenSchema(Schema):
     token = AssetField(expected_type=EvmToken, required=True)
+
+
+class CalendarCommonEntrySchema(Schema):
+    name = fields.String(required=True)
+    description = fields.String(load_default='')
```
LefterisJP, line 3539: Aha so if missing you have it empty? Isn't it better to just keep `None` for consistency with the DB schema also? It's optional and can be `NULL` there

rotkehlchen/api/v1/schemas.py
```diff
@@ -3531,3 +3532,55 @@ class FalsePositveSpamTokenSchema(Schema):
...
+    counterparty = fields.String(load_default=None)
+    address = fields.String(load_default=None)
+
+
+class NewCalendarEntrySchema(CalendarCommonEntrySchema):
+    timestamp = TimestampField(required=True)
+
+
+class DeleteCalendarEntrySchema(IntegerIdentifierSchema):
```
LefterisJP, line 3548: You don't need to create a class if nothing new is added. Just use `IntegerIdentifierSchema`

rotkehlchen/api/v1/schemas.py
```diff
@@ -3531,3 +3532,55 @@ class FalsePositveSpamTokenSchema(Schema):
...
+class UpdateCalendarSchema(NewCalendarEntrySchema, IntegerIdentifierSchema):
+    ...
+
+
+class QueryCalendarSchema(DBPaginationSchema, TimestampRangeSchema, DBOrderBySchema):
+    name = fields.String(load_default=None)
+    description = fields.String(load_default=None)
+    counterparty = fields.String(load_default=None)
+    addresses = fields.List(fields.String, load_default=None)
+    to_timestamp = TimestampField(load_default=None)
```
LefterisJP, line 3561: This is already part of `TimestampRangeSchema`

rotkehlchen/api/v1/schemas.py
```diff
@@ -3531,3 +3532,55 @@ class FalsePositveSpamTokenSchema(Schema):
...
+
+class UpdateCalendarSchema(NewCalendarEntrySchema, IntegerIdentifierSchema):
+    ...
+
+
+class QueryCalendarSchema(DBPaginationSchema, TimestampRangeSchema, DBOrderBySchema):
+    name = fields.String(load_default=None)
+    description = fields.String(load_default=None)
+    counterparty = fields.String(load_default=None)
+    addresses = fields.List(fields.String, load_default=None)
```
LefterisJP, line 3560: This should be typed properly no? Don't we have a BlockchainAddressed field which checks stuff?

Otherwise someone could write invalid addresses as api input here

rotkehlchen/api/v1/schemas.py
```diff
@@ -3531,3 +3532,55 @@ class FalsePositveSpamTokenSchema(Schema):
 
 class SpamTokenSchema(Schema):
     token = AssetField(expected_type=EvmToken, required=True)
+
+
+class CalendarCommonEntrySchema(Schema):
+    name = fields.String(required=True)
+    description = fields.String(load_default='')
+    counterparty = fields.String(load_default=None)
```
LefterisJP, line 3540: Who/where checks given counterpart is a valid counterparty?

rotkehlchen/db/calendar.py
```diff
@@ -0,0 +1,219 @@
...
+
+class DBCalendar:
+
+    def __init__(self, database: 'DBHandler') -> None:
+        self.db = database
+
+    def create_calendar_entry(
+            self,
+            name: str,
+            timestamp: int,
```
LefterisJP, line 118: again same concerns for type of timestamp and description as before. Entire file

rotkehlchen/db/calendar.py
```diff
@@ -0,0 +1,219 @@
...
+                    'VALUES (?, ?, ?, ?, ?) RETURNING identifier',
+                    (
+                        name,
+                        timestamp,
+                        description,
+                        counterparty,
+                        address,
+                    ),
+                )
+            except sqlcipher.IntegrityError as e:  # pylint: disable=no-member
```
LefterisJP, line 140: Why integrity error? There is no foreign keys. So the only error I can see is already existing. Why not do `INSERT OR IGNORE` and then check if cursor write count is 1?

rotkehlchen/db/calendar.py
```diff
@@ -0,0 +1,219 @@
...
+                raise InputError(f'Failed to insert calendar entry {name} due to {e}') from e
+
+            identifier = write_cursor.fetchone()[0]
+        return identifier
+
+    def query_calendar_entry(self, filter_query: CalendarFilterQuery) -> dict[str, Any]:
+        """Query events using the provided filter and return the information required to
+        use pagination filters.
```
LefterisJP, line 148:
```suggestion
        """Query events using the provided filter and return the result along with the attributes used in pagination
```

rotkehlchen/db/calendar.py
```diff
@@ -0,0 +1,219 @@
...
+            result = [CalendarRow(
+                identifier=entry[0],
+                name=entry[1],
+                timestamp=entry[2],
+                description=entry[3],
+                counterparty=entry[4],
+                address=entry[5],
+            ) for entry in cursor]
+
+            entries_total = cursor.execute('SELECT COUNT(*) FROM calendar').fetchone()[0]
```
LefterisJP, line 166: can also use `get_entries_count` right?

rotkehlchen/db/calendar.py
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+    def delete_calendar_entry(self, identifier: int) -> None:
+        """May raise:
+        - InputError: if no event got deleted
+        """
+        with self.db.user_write() as write_cursor:
+            write_cursor.execute('DELETE FROM calendar WHERE identifier=?', (identifier,))
+            if write_cursor.rowcount != 1:
+                raise InputError('Tried to delete an unexistent calendar entry')
```
LefterisJP, line 187:
```suggestion
                raise InputError('Tried to delete a non existent calendar entry')
```

rotkehlchen/db/calendar.py
yabirgb: I can use it but it is basically the same

rotkehlchen/api/v1/schemas.py
yabirgb: For the history events we don't check it either. I'll add the same logic we use to provide them to validate it

rotkehlchen/api/v1/schemas.py
yabirgb: I've been checking and we don't have it. We have in `BlockchainAccountsDeleteSchema` something like it but is an address + blockchain thing

rotkehlchen/db/calendar.py
yabirgb: It can't happen either since the primary key is an auto increasing integer

LefterisJP: > It can't happen either since the primary key is an auto increasing integer

But then we would have duplicate calendar events. Perhaps make title unique?

rotkehlchen/api/v1/schemas.py
LefterisJP: So if you do it as a field you can type it properly here too. Please do.

rotkehlchen/api/v1/schemas.py
LefterisJP: Then do not inherit from it and redefine it. 

But this is silly. Just inherit and and not redefine. Keep consistent api defaults across the app. It's up to the frontend what it will call it with.

rotkehlchen/db/upgrades/v41_v42.py
```diff
@@ -84,14 +84,33 @@ def _upgrade_evmchains_to_skip_detection(write_cursor: 'DBCursor') -> None:
...
 
 
+@enter_exit_debug_log()
+def _add_calendar_table(write_cursor: 'DBCursor') -> None:
+    write_cursor.execute("""CREATE TABLE IF NOT EXISTS calendar (
+    identifier INTEGER PRIMARY KEY NOT NULL,
+    name TEXT NOT NULL,
+    timestamp INTEGER NOT NULL,
+    description TEXT,
+    counterparty TEXT,
```
LefterisJP, line 94: you did not change this. you absolutely need to change this!!!!!!

rotkehlchen/api/v1/fields.py
```diff
@@ -1029,3 +1029,32 @@ def _deserialize(
...
+        self.counterparties = counterparties
+
+    def _deserialize(
+            self,
+            value: str,
+            attr: str | None,  # pylint: disable=unused-argument
+            data: Mapping[str, Any] | None,
+            **_kwargs: Any,
+    ) -> str:
+        if self.counterparties is None:
```
LefterisJP, line 1054: Why is this needed? Can't you just make it required at input? This can only be a developer error. At the most it should be an assertion if anything

rotkehlchen/api/v1/schemas.py
```diff
@@ -3531,3 +3533,99 @@ class FalsePositveSpamTokenSchema(Schema):
...
+    address = fields.String(load_default=None)
+    blockchain = BlockchainField(load_default=None)
+
+    @validates_schema
+    def validate_schema(
+            self,
+            data: dict[str, Any],
+            **_kwargs: Any,
+    ) -> None:
+        """Check that the combination of address and blockchaina re correct"""
```
LefterisJP, line 3553:
```suggestion
        """Check that the combination of address and blockchain are correct"""
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -3531,3 +3533,99 @@ class FalsePositveSpamTokenSchema(Schema):
...
+                'If any of address or blockchain is provided both need to be provided',
+            )
+
+        if data['blockchain'] is None:
+            return
+
+        blockchain = cast(SupportedBlockchain, data['blockchain'])
+        if ((
+            blockchain == SupportedBlockchain.BITCOIN and
+            is_valid_btc_address(data['address']) is False
```
LefterisJP, line 3565:
```suggestion
            not is_valid_btc_address(data['address'])
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -3531,3 +3533,99 @@ class FalsePositveSpamTokenSchema(Schema):
...
+        ) or (
+            blockchain == SupportedBlockchain.BITCOIN_CASH and
+            is_valid_bitcoin_cash_address(data['address']) is False
```
LefterisJP, line 3568:
```suggestion
            not is_valid_bitcoin_cash_address(data['address'])
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -3531,3 +3533,99 @@ class FalsePositveSpamTokenSchema(Schema):
...
+        ) or (
+            blockchain.get_chain_type() == 'substrate' and
+            is_valid_substrate_address(chain=blockchain, value=data['address']) is False  # type: ignore  # expects polkadot or kusama
```
LefterisJP, line 3571:
```suggestion
            not is_valid_substrate_address(chain=blockchain, value=data['address'])  # type: ignore  # expects polkadot or kusama
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -3531,3 +3533,99 @@ class FalsePositveSpamTokenSchema(Schema):
...
+            raise ValidationError(
+                f'Given value {data["address"]} is not a {blockchain} address',
+                field_name='address',
+            )
+
+
+class CalendarCommonEntrySchema(AnyBlockchainAddress):
+    name = fields.String(required=True)
+    description = fields.String(load_default=None)
+    counterparty = EvmCounterpartyField()
```
LefterisJP, line 3585: So if counterparty is never given at constructor remove it from the constructor.

rotkehlchen/api/v1/schemas.py
```diff
@@ -3531,3 +3533,99 @@ class FalsePositveSpamTokenSchema(Schema):
...
+
+
+class UpdateCalendarSchema(NewCalendarEntrySchema, IntegerIdentifierSchema):
+    ...
+
+
+class QueryCalendarSchema(DBPaginationSchema, TimestampRangeSchema, DBOrderBySchema):
+    name = fields.String(load_default=None)
+    description = fields.String(load_default=None)
+    counterparty = fields.String(load_default=None)
```
LefterisJP, line 3605: here counterparty and addresses are still not checked afaics

rotkehlchen/db/calendar.py
```diff
@@ -0,0 +1,223 @@
...
+    def serialize(self) -> dict[str, Any]:
+        return {
+            'identifier': self.identifier,
+            'name': self.name,
+            'description': self.description,
+            'counterparty': self.counterparty,
+            'timestamp': self.timestamp,
+            'address': self.address,
+            'blockchain': self.blockchain.serialize() if self.blockchain is not None else None,
+        }
```
LefterisJP, line 40: I asked you to add a serialize_for_db() and deserialize_from_db method for this and use it below in the various parts in this file and elsewhere (if needed)

rotkehlchen/api/v1/schemas.py
yabirgb: Had to add a bit of logic since to check the address you need the blockchain.

rotkehlchen/api/v1/schemas.py
```diff
@@ -3531,3 +3535,160 @@ class FalsePositveSpamTokenSchema(Schema):
...
+        self.allow_nullable_blockchain = allow_nullable_blockchain
+
+    @validates_schema
+    def validate_schema(
+            self,
+            data: dict[str, Any],
+            **_kwargs: Any,
+    ) -> None:
+        """Check that the combination of address and blockchain are correct"""
+
```
LefterisJP, line 3591: unnecessary newline

```suggestion
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -3531,3 +3535,160 @@ class FalsePositveSpamTokenSchema(Schema):
...
+
+class NewCalendarEntrySchema(CalendarCommonEntrySchema):
+    timestamp = TimestampField(required=True)
+
+    @post_load
+    def make_calendar_entry(self, data: dict[str, Any], **_kwargs: dict[str, Any]) -> dict[str, Any]:  # noqa: E501
+        return {
+            'calendar': CalendarEntry(
+                identifier=data.get('identifier', 0),  # not present here but when editing the calendar  # noqa: E501
```
LefterisJP, line 3628:
```suggestion
                identifier=data.get('identifier', 0),  # not present here but used in UpdateCalendarSchema  # noqa: E501
```

rotkehlchen/db/calendar.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def serialize_for_db(self) -> tuple[
+        str,  # name
```
LefterisJP, line 48: why not put identifier here first too and then have it declared as a type and have it the same in serialize and deserialize?

rotkehlchen/types.py
```diff
@@ -951,6 +951,11 @@ class OptionalChainAddress(NamedTuple):
     blockchain: SupportedBlockchain | None
 
 
+class OptionalBlockchainAddress(NamedTuple):
```
LefterisJP, line 954: Can't this just be the same as above with a Typevar for the type of address? No need to create another tuple

LefterisJP (approved):
lgtm

PR #7688, opened 2024-03-29
VCR api test_ethereum_tokens_detection

LefterisJP:


LefterisJP:
Only `rotkehlchen/tests/unit/test_tokens.py::test_detect_tokens_for_addresses` failed. And we have an issue to VCR it but seems to not be as simple.

PR #7692, opened 2024-03-31
Fix Arbitrum One bridge directions for L2 events

zinkkrysty:
When you deposit funds from L1 and receiving them on L2:  L1 transaction is a deposit, L2 transaction is a withdrawal

When you withdraw from L2 and receive them on L1: L2 transaction is a deposit, L1 is a withdrawal

It's relative to the address: The address deposits to a bridge or the address withdraws from a bridge. All we care about is direction of funds


rotkehlchen/chain/arbitrum_one/modules/arbitrum_one_bridge/decoder.py
```diff
@@ -158,7 +158,7 @@ def _decode_erc20_withdraw_event(
             from_event_subtype=HistoryEventSubType.NONE,
             asset=from_token,
             amount=amount,
-            to_event_type=HistoryEventType.WITHDRAWAL,
+            to_event_type=HistoryEventType.DEPOSIT,
```
LefterisJP, line 161: Can you also adjust the docstrings for all these functions?

yabirgb (approved):
lgtm

PR #7696, opened 2024-04-01
Feat (API): Allow sorting in addressbook

### LINKED ISSUE #7423: Add sorting support to `/names/addressbook/`
Enable the support in the backend and adjust the frontend to sort by `name` or `address`

OjusWiZard:
Closes #7423

## Checklist

- [x] Allow sorting by 'name' and 'address' in `/names/addressbook`

### DIFF
--- a/rotkehlchen/tests/api/test_addressbook.py
+++ b/rotkehlchen/tests/api/test_addressbook.py
@@ -66,6 +66,7 @@ def test_get_addressbook(
     result = assert_proper_response_with_result(response)
     deserialized_entries = [AddressbookEntry.deserialize(data=raw_entry) for raw_entry in result['entries']]  # noqa: E501
     assert deserialized_entries == entries_set
+    assert deserialized_entries != sorted(deserialized_entries, key=lambda x: x.name)  # not sorted
     assert result['entries_found'] == len(entries_set)
     assert result['entries_total'] == len(entries_set)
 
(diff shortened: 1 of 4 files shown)


rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -200,6 +200,40 @@ def test_get_addressbook(
...
+            ),
+            json={
+                'order_by_attributes': [attribute],
+                'ascending': [False],
+            },
+        )
+        result = assert_proper_response_with_result(response=response)
+        descending_events = [AddressbookEntry.deserialize(data=raw_entry) for raw_entry in result['entries']]  # noqa: E501
+        descending_events.reverse()  # reverse the descending ordered events
+        assert descending_events == ascending_events  # order is ascending now
```
yabirgb, line 235: you don't need to test that the logic for swapping the order works. This should be already tested since it is common logic what you are using

rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -200,6 +200,40 @@ def test_get_addressbook(
...
+                book_type=book_type,
+            ),
+            json={
+                'order_by_attributes': [attribute],
+                'ascending': [True],
+            },
+        )
+        result = assert_proper_response_with_result(response=response)
+        ascending_events = [AddressbookEntry.deserialize(data=raw_entry) for raw_entry in result['entries']]  # noqa: E501
+        assert len(ascending_events) > 0
```
yabirgb, line 218: you can check

```python
entries = [result['entry'][attribute]] for entry in result['entries']
assert sorted(entries) == entries
```

LefterisJP (approved):
LGTM.

PR #7699, opened 2024-04-02
Fix some errors in EVMLike api

LefterisJP:


rotkehlchen/api/rest.py
```diff
@@ -2838,6 +2838,7 @@ def decode_evmlike_transactions(
                     transaction=transaction,
                     tracked_addresses=tracked_addresses,
                 )
+            result = True
```
LefterisJP, line 2841: wtf? NO This is not what I typed.

LefterisJP:
okay all green and no funny business in the diff. Merging.

PR #7706, opened 2024-04-04
Use deserialization method for supportedblockchains

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/api/v1/fields.py
+++ b/rotkehlchen/api/v1/fields.py
@@ -560,7 +560,7 @@ class EvmChainLikeNameField(fields.Field):
             **_kwargs: Any,
     ) -> SupportedBlockchain:
         try:
-            chain = SupportedBlockchain.deserialize_from_name(value)
+            chain = SupportedBlockchain.deserialize(value)
         except DeserializationError as e:
             raise ValidationError(str(e)) from e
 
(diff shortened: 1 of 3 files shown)


rotkehlchen/api/v1/fields.py
```diff
@@ -560,7 +560,7 @@ def _deserialize(
             **_kwargs: Any,
     ) -> SupportedBlockchain:
         try:
-            chain = SupportedBlockchain.deserialize_from_name(value)
+            chain = SupportedBlockchain.deserialize(value)
```
LefterisJP, line 563: the one place where my change must have fucked things up is this one. And probably this is not needed. But the function and especially settings.py 333 is absolutely needed.

This duality is annoying but it's due to upgrading from evm chains to something more generic.

rotkehlchen/db/settings.py
```diff
@@ -330,7 +330,7 @@ def db_settings_from_dict(
             specified_args[key] = [ExchangeLocationID.deserialize(x) for x in values]
         elif key == 'evmchains_to_skip_detection':
             values = json.loads(value)
-            specified_args[key] = [SupportedBlockchain.deserialize_from_name(x) for x in values]
+            specified_args[key] = [SupportedBlockchain.deserialize(x) for x in values]
```
LefterisJP, line 333: if this went in, it would go boom

LefterisJP (changes requested):

LefterisJP (approved):
LGTM. As discussed the code in the end stores blockchain as value. So simple deserialize works.

Though we should work towards consolidation on naming. We should not have entries for the same thing in the DB as 'ETH` and 'ethereum'.

PR #7709, opened 2024-04-05
Implement zksync lite forced exit event

LefterisJP:


LefterisJP:
Okay, all green here. Merging.

PR #7710, opened 2024-04-05
Feat (Aave): Support v3 in Defi sections

### LINKED ISSUE #6636: Support Aave v3 positions in all chains
## Abstract

Auto detect aave v3 positions in mainnet (and other chains) and show them to the users. That means at least detect borrowed loans and show them as liabilities. The normal v3 tokens that are being lent should also be auto-detected and if they are not we should make sure they are.

Regarding a dedicated view for Aave v3 I am not sure if it is warranted as dedicated views don't scale well, are hard to maintain and for projects that have not paid for them it makes no financial sense for us.

## Motivation

We support v2 positions but v3 not. So we should support them at least in mainnet

## Specification

Figure out what's needed to do and do it.

OjusWiZard:
Closes #6636

## Checklist

- [x] Support Aave v3 lending and borrowing balances in Defi sections

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`6636` Aave v3 positions and liabilities will now be properly shown in the dashboard.
 * :feature:`7423` Users will be able to sort address book entries by displayed name and address.
 
 * :feature:`7281` rotki will now properly decode the transactions done on Aave v2 and v3 on all the supported chains.
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -2428,23 +2428,46 @@ class RestAPI:
             method='get_balances',
             # We need to query defi balances before since defi_balances must be populated
             query_specific_balances_before=['defi'],
-            # Giving the defi balances as a lambda function here so that they
-            # are retrieved only after we are sure the defi balances have been
+            # Giving the eth/defi balances as a lambda functions here so that they
+            # are retrieved only after we are sure the eth/defi balances have been
             # queried.
             given_defi_balances=lambda: self.rotkehlchen.chains_aggregator.defi_balances,
+            given_eth_balances=lambda: self.rotkehlchen.chains_aggregator.balances.eth,
         )
 
     @async_api_call()
-    def get_module_stats_using_balances(
+    def get_aave_stats_using_balances(
             self,
-            module: Literal['aave', 'compound'],
+            module: Literal['aave'],
             from_timestamp: Timestamp,
             to_timestamp: Timestamp,
     ) -> dict[str, Any]:
-        """
-        Query the provided module for statistics using the tracked addresses for such module.
-        This function uses the defi balances to enrich statistics.
-        """
+        """Query Aave module for statistics using the tracked addresses for aave.
+        This function uses the eth/defi balances to enrich statistics."""
+        return self._eth_module_query(
+            module_name=module,
+            method='get_stats_for_addresses',
+            # We need to query defi balances before since defi_balances must be populated
+            query_specific_balances_before=['defi'],
+            addresses=self.rotkehlchen.chains_aggregator.queried_addresses_for_module(module),
+            from_timestamp=from_timestamp,
+            to_timestamp=to_timestamp,
+            # Giving the eth/defi balances as lambda functions here so that they
+            # are retrieved only after we are sure the eth/defi balances have been
+            # queried.
+            given_defi_balances=lambda: self.rotkehlchen.chains_aggregator.defi_balances,
+            given_eth_balances=lambda: self.rotkehlchen.chains_aggregator.balances.eth,
+        )
+
+    @async_api_call()
+    def get_compound_stats_using_balances(
+            self,
+            module: Literal['compound'],
+            from_timestamp: Timestamp,
+            to_timestamp: Timestamp,
+    ) -> dict[str, Any]:
+        """Query Compound module for statistics using the tracked addresses for compound.
+        This function uses the defi balances to enrich statistics."""
         return self._eth_module_query(
             module_name=module,
             method='get_stats_for_addresses',
--- a/rotkehlchen/chain/ethereum/modules/aave/aave.py
+++ b/rotkehlchen/chain/ethereum/modules/aave/aave.py
@@ -7,12 +7,16 @@ from gevent.lock import Semaphore
 from rotkehlchen.accounting.structures.balance import Balance
 from rotkehlchen.assets.asset import Asset, CryptoAsset, EvmToken
 from rotkehlchen.chain.ethereum.constants import RAY
-from rotkehlchen.chain.evm.decoding.aave.constants import CPT_AAVE_V1, CPT_AAVE_V2
+from rotkehlchen.chain.ethereum.modules.aave.v3.constants import AAVE_V3_DATA_PROVIDER
+from rotkehlchen.chain.evm.decoding.aave.constants import CPT_AAVE_V1, CPT_AAVE_V2, CPT_AAVE_V3
 from rotkehlchen.chain.evm.types import string_to_evm_address
-from rotkehlchen.constants.resolver import ethaddress_to_identifier
+from rotkehlchen.constants.misc import ONE
+from rotkehlchen.constants.resolver import ethaddress_to_identifier, strethaddress_to_identifier
+from rotkehlchen.constants.timing import YEAR_IN_SECONDS
 from rotkehlchen.db.filtering import EvmEventFilterQuery
 from rotkehlchen.db.history_events import DBHistoryEvents
 from rotkehlchen.errors.asset import UnknownAsset, WrongAssetType
+from rotkehlchen.errors.misc import RemoteError
 from rotkehlchen.fval import FVal
 from rotkehlchen.history.events.structures.evm_event import EvmEvent
 from rotkehlchen.history.events.structures.types import HistoryEventSubType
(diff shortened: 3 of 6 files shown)


rotkehlchen/chain/ethereum/modules/aave/aave.py
```diff
@@ -174,6 +204,39 @@ def _get_balances(
...
+        ).items():
+            for asset, balance in balance_sheet.assets.items():
+                if (
+                    asset.is_evm_token() and
+                    (asset := asset.resolve_to_evm_token()).protocol == CPT_AAVE_V3 and
+                    (reserve_data := self._get_v3_reserve_data(asset)) is not None
+                ):
+                    aave_balances[account].lending[CryptoAsset(
+                        strethaddress_to_identifier(asset.underlying_tokens[0].address),  # type: ignore[attr-defined]  # asset is resolved to evm token
+                    )] = AaveLendingBalance(
```
yabirgb, line 218: can you check if the underlying tokens is not empty before? I'm not sure if we set it to None or empty list but I see a user modifying it and creating an error here

rotkehlchen/chain/ethereum/modules/aave/aave.py
```diff
@@ -74,33 +77,60 @@ def __init__(
...
+        if (
+            (reserve_data := self.v3_reserve_cache.get(reserve_token)) is None and
+            reserve_token.underlying_tokens is not None and
+            len(reserve_token.underlying_tokens) == 1
+        ):
+            reserve_result = self.v3_data_provider.call(
+                node_inquirer=self.ethereum,
+                method_name='getReserveData',
+                arguments=[reserve_token.underlying_tokens[0].address],
+            )
```
yabirgb, line 94: this can raise remote error

rotkehlchen/chain/ethereum/modules/aave/aave.py
yabirgb: I just noticed that you make this check in `_get_v3_reserve_data`. Add a comment mentioning that is checked there

rotkehlchen/chain/ethereum/modules/aave/aave.py
```diff
@@ -74,33 +77,60 @@ def __init__(
...
+            given_eth_balances: 'GIVEN_ETH_BALANCES',
     ) -> dict[ChecksumEvmAddress, AaveBalances]:
         """Retrieves the aave balances
 
         Receives the defi balances from zerion as an argument. They can either be directly given
         as the defi balances mapping or as a callable that will retrieve the
         balances mapping when executed.
         """
-        aave_balances = {}
+        aave_balances: defaultdict[ChecksumEvmAddress, AaveBalances] = defaultdict(lambda: AaveBalances(lending={}, borrowing={}))  # noqa: E501
```
yabirgb, line 125: update the docstring of this function since now it include things from the balances

rotkehlchen/chain/ethereum/modules/aave/aave.py
```diff
@@ -74,33 +77,60 @@ def __init__(
...
+        ):
+            reserve_result = self.v3_data_provider.call(
+                node_inquirer=self.ethereum,
+                method_name='getReserveData',
+                arguments=[reserve_token.underlying_tokens[0].address],
+            )
+            reserve_data = ReserveData(
+                liquidity_rate=FVal(reserve_result[5] / RAY),
+                variable_borrow_rate=FVal(reserve_result[6] / RAY),
+                stable_borrow_rate=FVal(reserve_result[7] / RAY),
```
yabirgb, line 98: here you are dividing before converting to FVal

```suggestion
                liquidity_rate=FVal(reserve_result[5]) / RAY,
                variable_borrow_rate=FVal(reserve_result[6]) / RAY,
                stable_borrow_rate=FVal(reserve_result[7]) / RAY,
```

rotkehlchen/tests/api/test_aave.py
```diff
@@ -25,78 +30,138 @@
...
-    TODO: Here we should use a test account for which we will know what balances
-    it has and we never modify
-    """
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('ethereum_accounts', [[
+    AAVE_BALANCESV1_TEST_ACC, AAVE_BALANCESV2_TEST_ACC, AAVE_BALANCESV3_TEST_ACC,
+]])
+@pytest.mark.parametrize('ethereum_modules', [['aave']])
+def test_query_aave_balances(rotkehlchen_api_server: APIServer) -> None:
+    """Check querying the aave balances endpoint works. Uses real data."""
```
yabirgb, line 55: It is real data for v1 and v2 and mocked for v3 right?

OjusWiZard: yeah, updating the docstring

rotkehlchen/chain/ethereum/modules/aave/aave.py
```diff
@@ -74,33 +78,62 @@ def __init__(
...
+            try:
+                reserve_result = self.v3_data_provider.call(
+                    node_inquirer=self.ethereum,
+                    method_name='getReserveData',
+                    arguments=[reserve_token.underlying_tokens[0].address],
+                )
+            except RemoteError as e:
+                self.msg_aggregator.add_error(
+                    f'Failed to get reserve data from Aave v3 data provider for token {reserve_token!s}, due to {e}',  # noqa: E501
+                )
```
yabirgb, line 100: you need to return None here otherwise reserve_result is not defined

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`6636` Aave v3 positions and liabilities will now be properly shown in the dashboard and the defi sections.
```
yabirgb, line 5: I would omit the defi section since it is only for ethereum but let @LefterisJP say here

rotkehlchen/chain/ethereum/modules/aave/aave.py
```diff
@@ -174,6 +207,41 @@ def _get_balances(
...
+                    aave_balances[account].borrowing[CryptoAsset(
+                        # existence of underlying_tokens is checked in _get_v3_reserve_data
+                        strethaddress_to_identifier(liability.underlying_tokens[0].address),  # type: ignore[attr-defined]  # asset is resolved to evm token
+                    )] = AaveBorrowingBalance(
+                        balance=balance,
+                        variable_apr=reserve_data.variable_borrow_rate,
+                        stable_apr=reserve_data.stable_borrow_rate,
+                        version=3,
+                    )
+
```
yabirgb, line 244: shouldn't you clean the cache? The apy will be different between calls I believe and as it is now you will need to close the app to get updated values

docs/changelog.rst
LefterisJP: yes + 1. I agree. Remove defi section.

rotkehlchen/api/rest.py
```diff
@@ -2428,23 +2428,46 @@ def get_aave_balances(self) -> dict[str, Any]:
...
+        This function uses the eth/defi balances to enrich statistics."""
+        return self._eth_module_query(
+            module_name=module,
+            method='get_stats_for_addresses',
+            # We need to query defi balances before since defi_balances must be populated
+            query_specific_balances_before=['defi'],
+            addresses=self.rotkehlchen.chains_aggregator.queried_addresses_for_module(module),
+            from_timestamp=from_timestamp,
+            to_timestamp=to_timestamp,
+            # Giving the eth/defi balances as a lambda functions here so that they
```
LefterisJP, line 2455:
```suggestion
            # Giving the eth/defi balances as lambda functions here so that they
```

rotkehlchen/chain/ethereum/modules/aave/aave.py
```diff
@@ -74,33 +78,63 @@ def __init__(
         self.history_lock = Semaphore()
         self.balances_lock = Semaphore()
         self.counterparties = [CPT_AAVE_V1, CPT_AAVE_V2]
+        self.v3_data_provider = self.ethereum.contracts.contract(AAVE_V3_DATA_PROVIDER)
+        self.v3_reserve_cache: dict[EvmToken, ReserveData] = {}
+
+    def _get_v3_reserve_data(self, reserve_token: 'EvmToken') -> 'ReserveData | None':
```
LefterisJP, line 84: Aren't these constant once queried? Why save them in memory and not global DB cache?

rotkehlchen/tests/api/test_aave.py
```diff
@@ -25,78 +30,138 @@
...
-    TODO: Here we should use a test account for which we will know what balances
-    it has and we never modify
-    """
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('ethereum_accounts', [[
+    AAVE_BALANCESV1_TEST_ACC, AAVE_BALANCESV2_TEST_ACC, AAVE_BALANCESV3_TEST_ACC,
+]])
+@pytest.mark.parametrize('ethereum_modules', [['aave']])
+def test_query_aave_balances(rotkehlchen_api_server: APIServer) -> None:
+    """Check querying the aave balances endpoint works. Uses real data for v1/v2."""
```
LefterisJP, line 55: What about v3? It's mocked? Why? Why not find an online position since it's a deployed and recent version and mock it in another test? At the least you should write it's mocked here in the docstring.

rotkehlchen/chain/ethereum/modules/aave/aave.py
OjusWiZard: They can be stored in globalDb cache to make it faster. But I could not find anywhere, how frequently they are updated. If they are updated very frequently then better to query every time.

rotkehlchen/tests/api/test_aave.py
OjusWiZard: I avoided that for v3 because for it to work it needs the token balances and the underlying token mappings in the db.

LefterisJP (approved):
LGTM

PR #7712, opened 2024-04-07
Bump ruff and fix tests

LefterisJP:
Also clean up code a bit thanks to ruff warnings


rotkehlchen/tests/unit/decoders/test_aave_v3.py
```diff
@@ -194,7 +193,7 @@ def test_aave_v3_deposit(database, ethereum_inquirer, ethereum_accounts) -> None
             location=Location.ETHEREUM,
             event_type=HistoryEventType.DEPOSIT,
             event_subtype=HistoryEventSubType.DEPOSIT_ASSET,
-            asset=A_LDO,
+            asset=EvmToken('eip155:1/erc20:0x5A98FcBEA516Cf06857215779Fd812CA3beF1B32'),  # LDO
```
LefterisJP, line 196: since my removal of LDO in bugfixes, this fails. Fixing it here.

We should not keep constant assets only for tests.

PR #7716, opened 2024-04-09
Proper error on insert conflict

yabirgb:
This PR adds proper a good error message if we failed to insert due to a conflict. I opted to keep the `IGNORE` part and check the identifier instead of catching the error because it is easier to understand for the user than the UNIQUE constraint failed error that sql returns


rotkehlchen/api/v1/schemas.py
```diff
@@ -3609,6 +3609,7 @@ class CalendarCommonEntrySchema(AnyBlockchainAddress):
     name = fields.String(required=True)
     description = fields.String(load_default=None)
     counterparty = EvmCounterpartyField()
+    color = fields.String(load_default=None)
```
LefterisJP, line 3612: We already have a color field. Please use that. Search for `ColorField` in this file.

rotkehlchen/db/calendar.py
```diff
@@ -31,6 +31,7 @@ class CalendarEntry(NamedTuple):
     counterparty: str | None
     address: BlockchainAddress | None
     blockchain: SupportedBlockchain | None
+    color: str | None
```
LefterisJP, line 34: The type we use for this should be same as the tag one.

Use that please. It's in types.py
```
T_HexColorCode = str
HexColorCode = NewType('HexColorCode', T_HexColorCode)
```

LefterisJP (approved):
LGTM.

PR #7718, opened 2024-04-09
Fix (test_query_aave_balances): Remove flakiness

OjusWiZard:
## Checklist

- [x] Only test with a synchronized API call to avoid flakiness in the VCRed test.


yabirgb (approved):
lgtm

PR #7727, opened 2024-04-10
fix(frontend): save correct values for evms to ignore token detection in settings

tewshi:
frontend part of #7723

## Checklist


frontend/app/src/components/settings/general/EvmChainsToIgnoreSettings.vue
```diff
@@ -79,6 +75,7 @@ function removeChain(chain: Blockchain) {
         :success-messages="success"
         :error-messages="error"
         :class="css['chain-select']"
+        class="general-settings__fields__account-chains-to-skip-detection"
```
kelsos, line 78: I guess here we pass the data-cy to another element than the one the class goes right?

tewshi: yes, so the data-cy goes to the autocomplete input itself...

kelsos (approved):
lgtm

PR #7730, opened 2024-04-11
Introduce zksync lite swaps as decoded events

LefterisJP:


LefterisJP:
all green. MERGEMENTATION IS OF BEGINNING

PR #7737, opened 2024-04-11
Test fixes and a bug fix

LefterisJP:


rotkehlchen/chain/ethereum/utils.py
```diff
@@ -223,13 +222,3 @@ def try_download_ens_avatar(
...
-
-
-def has_tracked_addresses(
-        blockchain: SupportedBlockchain,
-        database: 'DBHandler',
-) -> bool:
-    with database.conn.read_ctx() as cursor:
-        return database.get_blockchain_accounts(cursor).get(
-            blockchain=blockchain,
-        ) != 0
```
LefterisJP, line 235: This function was broken and never worked. Always returned true.

Blockchain accounts is an object of `BlockchainAccounts`. Will never be 0.

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -548,19 +548,20 @@ def test_maybe_augmented_detect_new_spam_tokens(
...
 def test_tasks_dont_schedule_if_no_eth_address(task_manager: TaskManager) -> None:
     """Test that we don't execute extra logic in tasks if no ethereum address is tracked"""
-    task_manager.should_schedule = True
-    for func in (
-        task_manager._maybe_update_yearn_vaults,
-        task_manager._maybe_update_ilk_cache,
-    ):
-        with (
-            patch('rotkehlchen.tasks.utils.should_run_periodic_task') as mocked_func,
+    with gevent.Timeout(5):  # this should not take long. Otherwise a long running task ran
```
LefterisJP, line 551: added a timeout to make test fail even if we change the way task manager's task `_maybe_update_yearn_vaults` and `_maybe_update_ilk_cache` works.

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -548,19 +548,20 @@ def test_maybe_augmented_detect_new_spam_tokens(
...
+            task_manager._maybe_update_yearn_vaults,
+            task_manager._maybe_update_ilk_cache,
         ):
-            task_manager.potential_tasks = [func]
-            task_manager.schedule()
-            if len(task_manager.running_greenlets) != 0:
-                gevent.joinall(task_manager.running_greenlets[func])
-            assert mocked_func.call_count == 0
+            with (  # if we get to should_update_protocol_cache it means check did not work
+                patch('rotkehlchen.chain.ethereum.utils.should_update_protocol_cache') as mocked_func,  # noqa: E501
```
LefterisJP, line 558: this was also a bug in the test introduced after OJ's changes. Before it was checking it never got to `rotkehlchen.tasks.utils.should_run_periodic_task`. But with OJ's changes that should now be `rotkehlchen.tasks.utils.should_update_protocol_cache`

rotkehlchen/tasks/manager.py
```diff
@@ -653,11 +653,9 @@ def _maybe_run_events_processing(self) -> Optional[list[gevent.Greenlet]]:
         )]
 
     def _maybe_update_yearn_vaults(self) -> Optional[list[gevent.Greenlet]]:
-        if has_tracked_addresses(
```
LefterisJP, line 656: so both for yearn vaults and ilk cache we always got in this code, even if user was tracking no ethereum addresses.

LefterisJP:
Only failed test is `rotkehlchen/tests/api/test_ens.py::test_reverse_ens` due to a remote error.

VCRing it is a tiny bit more complicated, seems to fail at replay for some reason. So will address it elsewhere.

PR #7744, opened 2024-04-13
refactor: Remove unused airdrops_list from accountants

zinkkrysty:
I noticed this code is not used, and the tests seems to pass without it. Is there a reason why this was made? I guess if there is need for this code, there should be at least tests for it.


LefterisJP (approved):
To tell you the truth @zinkkrysty I am not sure.

I seem to recall using them inside some of the module specific accountants. But as you say this is no longer the case.

As you see it's given to all the EVM Base decoder classes which later provide it as kwargs to all per module accountants if there is an Airdrops specific module.

I guess this is no longer needed as we handle airdrop accounting in a more generic way now.

Thank you for the PR!

PR #7745, opened 2024-04-14
Sunday is ruff day

LefterisJP:
Bump ruff to 0.3.7

There is a lot of noqa for augmented assignment due to this bug: https://github.com/astral-sh/ruff/issues/10925

We can remove once next version of ruff comes out

### DIFF
--- a/rotkehlchen/chain/ethereum/graph.py
+++ b/rotkehlchen/chain/ethereum/graph.py
@@ -68,7 +68,7 @@ class Graph:
             prefix += json.dumps(param_types).replace('"', '').replace('{', '(').replace('}', ')')
             prefix += '{'
 
-        querystr = prefix + querystr
+        querystr = prefix + querystr  # https://github.com/astral-sh/ruff/issues/10925 # noqa: E501 PLR6104
         log.debug(f'Querying The Graph for {querystr}')
 
         retries_left = CachedSettings().get_query_retry_limit()
--- a/rotkehlchen/db/drivers/gevent.py
+++ b/rotkehlchen/db/drivers/gevent.py
@@ -261,7 +261,7 @@ class DBConnection:
             )
         else:
             self._conn = sqlcipher.connect(  # pylint: disable=no-member
-                database=path,
+                database=str(path),
                 check_same_thread=False,
                 isolation_level=None,
             )
--- a/rotkehlchen/icons.py
+++ b/rotkehlchen/icons.py
@@ -187,9 +187,7 @@ class IconManager:
         if response.status_code != HTTPStatus.OK:
             return False
 
-        with open(self.iconfile_path(asset), 'wb') as f:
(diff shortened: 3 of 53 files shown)


rotkehlchen/icons.py
```diff
@@ -187,9 +187,7 @@ def query_coingecko_for_icon(self, asset: AssetWithNameAndType, coingecko_id: st
         if response.status_code != HTTPStatus.OK:
             return False
 
-        with open(self.iconfile_path(asset), 'wb') as f:
-            f.write(response.content)
-
+        self.iconfile_path(asset).write_bytes(response.content)
```
yabirgb, line 190: gigabrain. I didn't know this existed

yabirgb (approved):

yabirgb:
The failing tests from the subgraph it seems that the graph changed?

```
[14/04/2024 14:27:20 UTC] ERROR rotkehlchen.user_messages Greenlet-95: Failed to request the Balancer subgraph due to Failed to query the graph for poolShares ( first: $limit, skip: $offset, where: { userAddress_in: $addresses, balance_gt: $balance, } ) { userAddress { id } balance poolId { id tokens { address balance decimals denormWeight name symbol } totalShares totalWeight } }}query ($limit: Int!, $offset: Int!, $addresses: [ID!], $balance: BigDecimal!){ due to Syntax Error: Unexpected Name 'poolShares'.
```

needs to be removed asap. The subghaph doesn't seem to have changed https://thegraph.com/hosted-service/subgraph/balancer-labs/balancer

yabirgb (changes requested):
image
Seems that also some tests need fixing

LefterisJP:
There is no need to review an ongoing PR lol. I am looking into the failed tests. They are seem related to my changes as they pass in develop

rotkehlchen/chain/ethereum/graph.py
```diff
@@ -68,7 +68,7 @@ def query(
             prefix += json.dumps(param_types).replace('"', '').replace('{', '(').replace('}', ')')
             prefix += '{'
 
-        querystr = prefix + querystr
+        querystr += prefix
```
LefterisJP, line 71: this is why the subgraph tests failed. It's the ruff bug, but weird. I did not use autofix so may have done this on my own without paying attention.

rotkehlchen/db/drivers/gevent.py
```diff
@@ -261,7 +261,7 @@ def __init__(
             )
         else:
             self._conn = sqlcipher.connect(  # pylint: disable=no-member
-                database=path,
+                database=str(path),
```
LefterisJP, line 264: The other part of the failure is this one, which is ratther weird. Only for sqlcipher if a path gets here it fails with type error. sqlite.connect works fine with paths.

PR #7748, opened 2024-04-15
Introduce api for reminders

yabirgb:
This PR introduces the API with verbs to create, edit, delete and query reminders for calendar events.

After this changes we need to trigger them from the backend


rotkehlchen/api/rest.py
```diff
@@ -4938,3 +4941,47 @@ def update_calendar_entry(self, calendar: CalendarEntry) -> Response:
             {'entry_id': calendar_event_id}),
             status_code=HTTPStatus.OK,
         )
+
+    def create_calendar_reminder(self, entry: ReminderEntry) -> Response:
```
LefterisJP, line 4945: tiny docstring

rotkehlchen/api/v1/resources.py
```diff
@@ -3295,3 +3298,31 @@ def patch(self, calendar: CalendarEntry) -> Response:
     @resource_parser.use_kwargs(make_query_schema, location='json_and_query')
     def post(self, filter_query: CalendarFilterQuery) -> Response:
         return self.rest_api.query_calendar(filter_query=filter_query)
+
+
+class CalendarReminderResource(BaseMethodView):
```
LefterisJP, line 3303: Same as the endpoint

```suggestion
class CalendarRemindersResource(BaseMethodView):
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -3694,3 +3701,54 @@ def make_calendar_query(
             identifiers=data['identifiers'],
         )
         return {'filter_query': filter_query}
+
+
+class CalendarReminderCommonEntrySchema(Schema):
+    AT_TIME_REGEX = re.compile(r'^(2[0-3]|[0-1][0-9]):[0-5][0-9]$')
```
LefterisJP, line 3707: Mark it as constant. This should be a constant class variable shared by all inherited classes right? Not repeating in memory.
```suggestion
    AT_TIME_REGEX: Final = re.compile(r'^(2[0-3]|[0-1][0-9]):[0-5][0-9]$')
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -3694,3 +3701,54 @@ def make_calendar_query(
...
+    @validates_schema
+    def validate_schema(
+            self,
+            data: dict[str, Any],
+            **_kwargs: Any,
+    ) -> None:
+        if data['at_time'] is None:
+            return
+
+        if not self.AT_TIME_REGEX.match(data['at_time']):
```
LefterisJP, line 3728: Why not use a timestamp? Let the frontend decide how to translate it into time based on the rules and settings we already have. Having a regex and a string is kind of silly no?

rotkehlchen/api/v1/schemas.py
```diff
@@ -3694,3 +3701,54 @@ def make_calendar_query(
...
+                at_time=data['at_time'],
+            ),
+        }
+
+
+class UpdateCalendarReminderSchema(NewCalendarReminderSchema, IntegerIdentifierSchema):
+    ...
+
+
+class QueryCalendarReminderSchema(Schema):
```
LefterisJP, line 3753: Don't use this. Use `IntegerIdentifierSchema`. Don't remake multiple schema classes for the same thing but change the name.

rotkehlchen/db/calendar.py
```diff
@@ -91,6 +94,53 @@ def deserialize_from_db(
         )
 
 
+class ReminderEntry(NamedTuple):
```
LefterisJP, line 97: you need a docstring about this and what all its fields are here

rotkehlchen/db/calendar.py
```diff
@@ -91,6 +94,53 @@ def deserialize_from_db(
...
+    event_id: int
+    amount: int
```
LefterisJP, line 99: What are all these? You just need an event id and a timestamp or list of timestamps.

LefterisJP (commented):
I think this needs rewriting as far as the reminder schema is concerned.

rotkehlchen/db/schema.py
```diff
@@ -786,6 +786,18 @@
...
+DB_CREATE_CALENDAR_REMINDER = """
+CREATE TABLE IF NOT EXISTS reminder (
+    identifier INTEGER PRIMARY KEY NOT NULL,
+    event_id INTEGER NOT NULL,
+    amount INTEGER NOT NULL,
+    unit TEXT NOT NULL,
+    at_time TEXT,
+    FOREIGN KEY(event_id) REFERENCES calendar(identifier) ON DELETE CASCADE
+);
+"""
```
LefterisJP, line 798: plural ... just like we do with history events, trades etc.


```suggestion
DB_CREATE_CALENDAR_REMINDERS = """
CREATE TABLE IF NOT EXISTS calendar_reminders (
    identifier INTEGER PRIMARY KEY NOT NULL,
    event_id INTEGER NOT NULL,
    secs_before INTEGER NOT NULL,
    executed INTEGER NOT NULL CHECK (status IN (0, 1)),
    FOREIGN KEY(event_id) REFERENCES calendar(identifier) ON DELETE CASCADE
);
"""
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -3694,3 +3694,30 @@ def make_calendar_query(
...
+
+
+class CalendarReminderCommonEntrySchema(Schema):
+    event_id = fields.Integer(
+        required=True,
+        validate=webargs.validate.Range(min=1, error='event_id has to be bigger than 0'),
+    )
+    secs_before = fields.Integer(
+        required=True,
+        validate=webargs.validate.Range(min=1, error='secs_berfore has to be bigger than 0'),
```
LefterisJP, line 3706:
```suggestion
        validate=webargs.validate.Range(min=1, error='secs_before has to be bigger than 0'),
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -3694,3 +3694,30 @@ def make_calendar_query(
             identifiers=data['identifiers'],
         )
         return {'filter_query': filter_query}
+
+
+class CalendarReminderCommonEntrySchema(Schema):
+    event_id = fields.Integer(
+        required=True,
+        validate=webargs.validate.Range(min=1, error='event_id has to be bigger than 0'),
```
LefterisJP, line 3702: Why do you need this check? I know sqlite rowid has to start from 1 but no need to check this here. This is just an id the frontend will give us. We give it to them anyway. No need for a check here. If they give us a non-existing id it will be seen by the backend anyway later and handled properly.

rotkehlchen/db/schema.py
```diff
@@ -786,6 +786,16 @@
 """  # noqa: E501
 
 
+DB_CREATE_CALENDAR_REMINDER = """
```
LefterisJP, line 789:
```suggestion
DB_CREATE_CALENDAR_REMINDERS = """
```

rotkehlchen/db/calendar.py
```diff
@@ -24,6 +24,9 @@
     from rotkehlchen.db.dbhandler import DBHandler
 
 
+VALID_REMINDER_UNITS = Literal['days', 'weeks']
+
+
```
LefterisJP, line 29: This needs to go
```suggestion
```

rotkehlchen/db/calendar.py
```diff
@@ -234,3 +279,51 @@ def update_calendar_entry(self, calendar: CalendarEntry) -> int:
...
+                raise InputError(f'Could not update reminder due to {e}') from e
+
+        return reminder.identifier
+
+    def query_reminder_entry(self, event_id: int) -> dict[str, list[ReminderEntry]]:
+        """Query reminder using the id of the event linked"""
+        with self.db.conn.read_ctx() as cursor:
+            cursor.execute(
+                'SELECT identifier, event_id, secs_before FROM calendar_reminders '
+                'WHERE event_id=?',
```
LefterisJP, line 326: Do an order BY here. either ASC or DESC secs_before, depending on how this is going to be consumed. And add it to the docstring. So we always have a guaranteed order of returned reminders based on distance from event

rotkehlchen/tests/api/test_calendar.py
```diff
@@ -14,6 +15,9 @@
 from rotkehlchen.types import ChecksumEvmAddress, SupportedBlockchain
 
 
+THREE_DAYS = DAY_IN_SECONDS * 3
+
+
```
LefterisJP, line 20: overkill. Used only in this test and only in 2 places. Just use the value directly

```suggestion
```

LefterisJP (commented):
Rebased PR on develop so tests run.

1. So this PR just adds DB schema and API operations for calendar reminder right? Everything else in other PRs?
2. Shouldn't we also have the API docs for these in this PR?

yabirgb:
1. yes
2. adding them

docs/api.rst
```diff
@@ -13685,6 +13685,175 @@ Managing calendar entries
...
+  :statuscode 409: Failed to validate the data. Entry doesn't exist.
+  :statuscode 500: Internal rotki error.
+
+
+Managing calendar reminders
+============================
+
+.. http:post:: /api/(version)/calendar/reminders
+
+   Doing a POST on this endpoint will allow querying the calendar reminders using the identifier of the calendar entry associated.
```
LefterisJP, line 13710:
```suggestion
   Doing a POST on this endpoint will allow querying the calendar reminders using the identifier of the associated calendar entry.
```

docs/api.rst
```diff
@@ -13685,6 +13685,175 @@ Managing calendar entries
...
+
+      {
+        "result":{
+          "entries":[
+            {
+              "identifier":1,
+              "event_id": 1,
+              "secs_before": 213234124
+            },
+            {
```
LefterisJP, line 13739:
```suggestion
            }, {
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -3694,3 +3694,27 @@ def make_calendar_query(
...
+        validate=webargs.validate.Range(min=1, error='secs_before has to be bigger than 0'),
+    )
+
+
+class NewCalendarReminderSchema(CalendarReminderCommonEntrySchema):
+    @post_load
+    def make_calendar_entry(self, data: dict[str, Any], **_kwargs: dict[str, Any]) -> dict[str, Any]:  # noqa: E501
+        return {
+            'reminder': ReminderEntry(
+                identifier=data.get('identifier', 0),
```
LefterisJP, line 3712: why give a defautl value at all here? When can it not be given? And why should the default value be 0?

rotkehlchen/db/calendar.py
```diff
@@ -210,14 +246,20 @@ def query_calendar_entry(self, filter_query: CalendarFilterQuery) -> dict[str, A
...
 
-    def delete_calendar_entry(self, identifier: int) -> None:
-        """May raise:
+    def delete_entry(
+            self,
+            identifier: int,
+            entry_type: Literal['calendar', 'calendar_reminders'],
+    ) -> None:
+        """
+        Delete by identifier either calendar entries or reminders based on the entry_type
```
LefterisJP, line 255:
```suggestion
        Delete calendar entries or reminders by identifier based on the entry_type
```

rotkehlchen/db/calendar.py
```diff
@@ -234,3 +276,51 @@ def update_calendar_entry(self, calendar: CalendarEntry) -> int:
...
+                    'UPDATE calendar_reminders SET event_id=?, secs_before=? WHERE identifier=?',
+                    reminder.serialize_for_db(),
+                )
+            except sqlcipher.IntegrityError as e:  # pylint: disable=no-member
+                raise InputError(f'Could not update reminder due to {e}') from e
+
+        return reminder.identifier
+
+    def query_reminder_entry(self, event_id: int) -> dict[str, list[ReminderEntry]]:
+        """Query reminder using the id of the event linked"""
```
LefterisJP, line 319: Also explain in the docstring what the order they are returned in is.

```suggestion
        """Query reminder using the id of the linked event"""
```

LefterisJP (approved):
Multiple smaller comments. But will merge since green and you are working on subsequent PRs on this

PR #7749, opened 2024-04-15
Remove 'manualcurrent' from price oracles

moxw:
SQL diff of `rotkehlchen/tests/data/v41_rotkehlchen.db`:
```sql
INSERT INTO settings (name, value)
VALUES ('current_price_oracles', '["manualcurrent", "coingecko", "defillama", "cryptocompare", "uniswapv2", "uniswapv3"]');
```


rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2259,6 +2259,11 @@ def test_upgrade_db_41_to_42(user_data_dir, messages_aggregator):
             ChainID.deserialize_from_name(x) for x in json.loads(raw_list)
         ] == [ChainID.POLYGON_POS, ChainID.GNOSIS]
 
+        # get settings and confirm manualcurrent is present before the upgrade
+        cursor.execute('SELECT value FROM settings WHERE name="current_price_oracles"')
+        oracles = json.loads(cursor.fetchone()[0])
+        assert 'manualcurrent' in oracles
```
OjusWiZard, line 2265: can avoid creating extra variables
```suggestion
        assert 'manualcurrent' in json.loads(cursor.execute(
            'SELECT value FROM settings WHERE name="current_price_oracles"'
        ).fetchone()[0])
```

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2259,6 +2259,11 @@ def test_upgrade_db_41_to_42(user_data_dir, messages_aggregator):
             ChainID.deserialize_from_name(x) for x in json.loads(raw_list)
         ] == [ChainID.POLYGON_POS, ChainID.GNOSIS]
 
+        # get settings and confirm manualcurrent is present before the upgrade
+        cursor.execute('SELECT value FROM settings WHERE name="current_price_oracles"')
+        oracles = json.loads(cursor.fetchone()[0])
```
OjusWiZard, line 2264: this test is failing here, did you try running it? It fails probably because this setting is not already present in the table of the DB that is being tested (v41_rotkehlchen.db).

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2288,6 +2293,11 @@ def test_upgrade_db_41_to_42(user_data_dir, messages_aggregator):
             SupportedBlockchain.deserialize(x) for x in json.loads(raw_list)
         ] == [SupportedBlockchain.POLYGON_POS, SupportedBlockchain.GNOSIS]
 
+        # get current oracles and check that manualcurrent was removed and all others remain.
+        settings = db.get_settings(cursor=cursor)
+        expected_oracles = [CurrentPriceOracle.deserialize(oracle) for oracle in oracles if oracle != 'manualcurrent']  # noqa: E501
```
OjusWiZard, line 2298: generally avoid creating the variables that are used only once, like settings and expected_oracles here

rotkehlchen/db/upgrades/v41_v42.py
```diff
@@ -118,6 +118,24 @@ def _add_calendar_table(write_cursor: 'DBCursor') -> None:
     );""")  # noqa: E501
 
 
+@enter_exit_debug_log()
+def _remove_manualcurrent_oracle(write_cursor: 'DBCursor') -> None:
```
OjusWiZard, line 122: add a docstring to this function to explain what it does

rotkehlchen/db/upgrades/v41_v42.py
```diff
@@ -118,6 +118,24 @@ def _add_calendar_table(write_cursor: 'DBCursor') -> None:
...
+    write_cursor.execute('SELECT value FROM settings WHERE name="current_price_oracles"')
+    if (data := write_cursor.fetchone()) is None:
+        return  # oracles not configured
+    try:
```
OjusWiZard, line 126: some space for readability
```suggestion

    try:
```

rotkehlchen/db/upgrades/v41_v42.py
```diff
@@ -118,6 +118,24 @@ def _add_calendar_table(write_cursor: 'DBCursor') -> None:
...
+        oracles: list[str] = json.loads(data[0])
+    except json.JSONDecodeError as e:
+        log.debug(f'Failed to read oracles from user db. {e!s}')
```
OjusWiZard, line 129: should be logged as an error
```suggestion
        log.error(f'Failed to read oracles from user db. {e!s}')
```

rotkehlchen/db/upgrades/v41_v42.py
```diff
@@ -118,6 +118,24 @@ def _add_calendar_table(write_cursor: 'DBCursor') -> None:
...
+        return
+
+    new_oracles = [oracle for oracle in oracles if oracle != 'manualcurrent']
+    write_cursor.execute(
+        'INSERT OR REPLACE INTO settings(name, value) VALUES(?, ?)',
+        ('current_price_oracles', json.dumps(new_oracles)),
+    )
```
OjusWiZard, line 136:
```suggestion
    write_cursor.execute(
        'INSERT OR REPLACE INTO settings(name, value) VALUES(?, ?)', (
            'current_price_oracles',
            json.dumps([oracle for oracle in oracles if oracle != 'manualcurrent'])
        ),
    )
```

rotkehlchen/db/upgrades/v41_v42.py
```diff
@@ -136,3 +154,5 @@ def upgrade_v41_to_v42(db: 'DBHandler', progress_handler: 'DBUpgradeProgressHand
         progress_handler.new_step()
         _add_calendar_table(write_cursor)
         progress_handler.new_step()
+        _remove_manualcurrent_oracle(write_cursor)
```
OjusWiZard, line 157: update the docstring of `upgrade_v41_to_v42` function

rotkehlchen/db/upgrades/v41_v42.py
```diff
@@ -136,3 +154,5 @@ def upgrade_v41_to_v42(db: 'DBHandler', progress_handler: 'DBUpgradeProgressHand
...
+        progress_handler.new_step()
```
OjusWiZard, line 158: this will increment the step by one, but total steps are set as 4 above. It should be 5.

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2288,6 +2288,10 @@ def test_upgrade_db_41_to_42(user_data_dir, messages_aggregator):
             SupportedBlockchain.deserialize(x) for x in json.loads(raw_list)
         ] == [SupportedBlockchain.POLYGON_POS, SupportedBlockchain.GNOSIS]
 
+        # get current oracles and check that manualcurrent was removed and all others remain.
+        settings = db.get_settings(cursor=cursor)
+        assert CurrentPriceOracle.MANUALCURRENT not in settings.current_price_oracles
```
OjusWiZard, line 2293: this test is not valid yet. That's because `db.get_settings` returns settings that are stored in Db + default values for settings that are not stored. So, if you check `rotkehlchen/db/settings.py` line 186, the default is used without check the DB table.

OjusWiZard: Test should be like this:
1. Add the setting first in the table before upgrade. Can also add it directly in `rotkehlchen/tests/data/v41_rotkehlchen.db` (which is even better, because it saves us from doing an extra DB query). Also, if you edit any DB in the codebase, comment a SQL diff of it in the PR.
2. Ensure it's added by doing a SELECT query.
3. Upgrade
4. Check that the manualcurrent is removed.

rotkehlchen/db/upgrades/v41_v42.py
```diff
@@ -118,15 +118,38 @@ def _add_calendar_table(write_cursor: 'DBCursor') -> None:
...
+
+    try:
+        oracles: list[str] = json.loads(data[0])
+    except json.JSONDecodeError as e:
+        log.error(f'Failed to read oracles from user db. {e!s}')
+        return
+
+    write_cursor.execute(
+        'INSERT OR REPLACE INTO settings(name, value) VALUES(?, ?)',
+        (
```
OjusWiZard, line 136: nitpick, but making it compact
```suggestion
        'INSERT OR REPLACE INTO settings(name, value) VALUES(?, ?)', (
```

OjusWiZard (changes requested):

OjusWiZard (changes requested):
also please remove the TODO and filter from `set_oracles_order()` in `Inquirer` class, because it's not needed now.

rotkehlchen/db/upgrades/v41_v42.py
moxw: I used Black for formatting to ensure consistency. Should I adjust this to fit a different style preference?

OjusWiZard: no issues then, just make sure `make lint` passes

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2259,6 +2259,11 @@ def test_upgrade_db_41_to_42(user_data_dir, messages_aggregator):
             ChainID.deserialize_from_name(x) for x in json.loads(raw_list)
         ] == [ChainID.POLYGON_POS, ChainID.GNOSIS]
 
+        # get settings and confirm saddle is present before the upgrade
```
LefterisJP, line 2262: saddle?

OjusWiZard (approved):

LefterisJP (approved):
LGTM

PR #7751, opened 2024-04-15
Bump web3.py to 6.17.1

LefterisJP:
Should fix also the damn eth-typing problems that suddenly hit us in the tests due to the improper pinning the old web3.py had.


LefterisJP:
all green. Merging.

PR #7754, opened 2024-04-16
Fix (Uniswap): NFT Contract data for other chains

OjusWiZard:
## Checklist

- [x] Add contract data for Uniswap NFT manager in global.db
- [x] Enable enricher in other chains' decoders
- [x] Test for optimism

SQL diff for `global.db`:
```sql
INSERT INTO contract_data(rowid,address,chain_id,abi,deployed_block) VALUES(162,'0x03a520b32C04BF3bEEf7BEb72E919cf822Ed34f1',8453,15,1371714);
INSERT INTO contract_data(rowid,address,chain_id,abi,deployed_block) VALUES(163,'0xC36442b4a4522E871399CD717aBDD847Ab11FE88',42161,15,173);
INSERT INTO contract_data(rowid,address,chain_id,abi,deployed_block) VALUES(164,'0xC36442b4a4522E871399CD717aBDD847Ab11FE88',10,15,0);
INSERT INTO contract_data(rowid,address,chain_id,abi,deployed_block) VALUES(165,'0xC36442b4a4522E871399CD717aBDD847Ab11FE88',137,15,22760586);
```

### DIFF
--- a/rotkehlchen/chain/optimism/decoding/decoder.py
+++ b/rotkehlchen/chain/optimism/decoding/decoder.py
@@ -2,10 +2,6 @@ import logging
 from typing import TYPE_CHECKING
 
 from rotkehlchen.chain.evm.decoding.base import BaseDecoderToolsWithDSProxy
-from rotkehlchen.chain.evm.decoding.structures import (
-    FAILED_ENRICHMENT_OUTPUT,
-    TransferEnrichmentOutput,
-)
 from rotkehlchen.chain.evm.l2_with_l1_fees.decoding.decoder import L2WithL1FeesTransactionDecoder
 from rotkehlchen.constants.assets import A_ETH
 from rotkehlchen.db.l2withl1feestx import DBL2WithL1FeesTx
@@ -13,7 +9,6 @@ from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.types import ChecksumEvmAddress
 
 if TYPE_CHECKING:
-    from rotkehlchen.chain.evm.decoding.structures import EnricherContext
     from rotkehlchen.chain.optimism.node_inquirer import OptimismInquirer
     from rotkehlchen.chain.optimism.transactions import OptimismTransactions
     from rotkehlchen.db.dbhandler import DBHandler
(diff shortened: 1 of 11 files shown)


rotkehlchen/constants/assets.py
```diff
@@ -32,6 +32,7 @@
 A_PETH = Asset('eip155:1/erc20:0xf53AD2c6851052A81B42133467480961B2321C09')
 A_USDT = Asset('eip155:1/erc20:0xdAC17F958D2ee523a2206206994597C13D831ec7')
 A_USDC = Asset('eip155:1/erc20:0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48')
+A_OPTIMISM_USDC = Asset('eip155:10/erc20:0x0b2C639c533813f4Aa9D7837CAf62653d097Ff85')
```
yabirgb, line 35: I don't see it used but only in tests. Moving it here it is special and you should do it only for production code, not for tests

rotkehlchen/chain/optimism/decoding/decoder.py
```diff
@@ -48,7 +49,21 @@ def __init__(
 
     # -- methods that need to be implemented by child classes --
 
-    def _enrich_protocol_tranfers(self, context: 'EnricherContext') -> TransferEnrichmentOutput:  # pylint: disable=unused-argument
+    def _enrich_protocol_tranfers(self, context: 'EnricherContext') -> TransferEnrichmentOutput:
+        for enrich_call in self.rules.token_enricher_rules:
```
yabirgb, line 53: you are repeating code... Does it need to exists for all the chains? What happens for example with scroll? If so instead of modifying it overwrite the method in the evm class. If not you can avoid calling the logic by modifying `_enrich_protocol_tranfers` in the base class and having something like

```python
if not self.should_enrich_transfers:
     return Failed_enrichment output

_this_logic_that_you_repeat_
```

rotkehlchen/tests/unit/decoders/test_uniswapv3.py
```diff
@@ -1338,3 +1340,83 @@ def test_swap_on_polygon_pos(database, polygon_pos_inquirer, polygon_pos_account
         ),
     ]
     assert events == expected_events
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('optimism_accounts', [['0x9A539f692cDE873D6B882fc326c8d62D4cEA8048']])
+def test_add_liquidity_on_optimism(database, optimism_inquirer, optimism_accounts):
+    evmhash = deserialize_evm_tx_hash('0x96bd0e37e1734b5e73f9abdf30b39c4e4a6879667c2d01a7be2d95a85cc0b0cc')  # noqa: E501
+    events, _ = get_decoded_events_of_transaction(
```
yabirgb, line 1349: is this fixing something that should be in the changelog or it was already there?

OjusWiZard: there is already a changelog for this, I've updated it now

LefterisJP (approved):
LGTM. Merging

PR #7755, opened 2024-04-16
Add task for reminders

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -13707,7 +13707,7 @@ Managing calendar reminders
 
 .. http:post:: /api/(version)/calendar/reminders
 
-   Doing a POST on this endpoint will allow querying the calendar reminders using the identifier of the calendar entry associated.
+   Doing a POST on this endpoint will allow querying the calendar reminders using the identifier of the associated calendar entry.
 
 
   **Example Request**
@@ -13735,8 +13735,7 @@ Managing calendar reminders
               "identifier":1,
               "event_id": 1,
               "secs_before": 213234124
-            },
-            {
+            }, {
               "identifier":2,
               "event_id": 1,
               "secs_before": 2132341253
@@ -13754,7 +13753,7 @@ Managing calendar reminders
 
 .. http:put:: /api/(version)/calendar/reminders
 
-  Doing a PUT request on this endpoint will allow to create a new calendar reminder
+  Doing a PUT request on this endpoint will allow to create new calendar reminder entries. If any of the entries fails to get added the key ``failed`` will be populated in the response.
 
   **Example Request**
 
@@ -13765,8 +13764,10 @@ Managing calendar reminders
       Content-Type: application/json;charset=UTF-8
 
       {
-        "secs_before": 1869737344,
-        "event_id": 1
+        "reminders": [
+         {"secs_before": 1869737344, "event_id": 1},
+         {"secs_before": 1869737344, "event_id": 100}
+        ]
       }
 
   .. _calendar_reminder_fields:
@@ -13782,11 +13783,11 @@ Managing calendar reminders
       Content-Type: application/json
 
       {
-         "result": {"entry_id": 1},
+         "result": {"success": [1], "failed": [100]},
          "message": ""
       }
 
-  :resjson object result: object with the identifier of the calendar reminder created. 
+  :resjson object result: object with the ids of the events for which the reminders were created succesfully and for which it failed. If none failed the failed key is not returned.
   :statuscode 200: Entry correctly stored.
   :statuscode 401: No user is currently logged in.
   :statuscode 409: Failed to validate the data.
--- a/rotkehlchen/db/calendar.py
+++ b/rotkehlchen/db/calendar.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Literal, NamedTuple
 
 from pysqlcipher3 import dbapi2 as sqlcipher
@@ -91,18 +92,26 @@ class CalendarEntry(NamedTuple):
         )
 
 
-class ReminderEntry(NamedTuple):
-    """Store basic information to warn the user based on the event timestamp"""
-    event_id: int
+@dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=True)
+class BaseReminderData:
+    """Base fields queried for reminders"""
     secs_before: int
-    identifier: int = 0
+    identifier: int
 
     def serialize(self) -> dict[str, Any]:
         return {
             'identifier': self.identifier,
-            'event_id': self.event_id,
             'secs_before': self.secs_before,
         }
+
+
+@dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=True)
+class ReminderEntry(BaseReminderData):
+    """Store basic information to warn the user based on the event timestamp"""
+    event_id: int
+
+    def serialize(self) -> dict[str, Any]:
+        return super().serialize() | {'event_id': self.event_id}
 
     def serialize_for_db(self) -> tuple[
         int,  # event_id
@@ -222,7 +231,7 @@ class DBCalendar:
                     'address and blockchain already exist',
                 )
 
-            return identifier_row[0]
+        return identifier_row[0]
 
     def query_calendar_entry(self, filter_query: CalendarFilterQuery) -> dict[str, Any]:
         """Query events using the provided filter and return the result along with the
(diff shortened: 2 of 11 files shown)


yabirgb:
I FORGOT THAT I HAVE COMMENTS TO FIX FROM HERE. WILL PUSH THEM AFTER REVIEW https://github.com/rotki/rotki/pull/7748#discussion_r1567042825

rotkehlchen/db/calendar.py
```diff
@@ -320,7 +326,7 @@ def query_reminder_entry(self, event_id: int) -> dict[str, list[ReminderEntry]]:
         with self.db.conn.read_ctx() as cursor:
             cursor.execute(
                 'SELECT identifier, event_id, secs_before FROM calendar_reminders '
-                'WHERE event_id=? ORDER BY secs_before ASC',
+                'WHERE event_id=? and secs_before != 0 ORDER BY secs_before ASC',
```
LefterisJP, line 329: why this limit? Perhaps do not allow a `secs_before` equal to 0 in the first place if you don't want this?

rotkehlchen/tasks/calendar.py
```diff
@@ -0,0 +1,43 @@
...
+
+from rotkehlchen.api.websockets.typedefs import WSMessageType
+from rotkehlchen.db.calendar import CalendarEntry
+from rotkehlchen.db.dbhandler import DBHandler
+from rotkehlchen.user_messages import MessagesAggregator
+
+
+class CalendarNotification(NamedTuple):
+    """Basic reminder information along the calendar event linked to it"""
+    event: CalendarEntry
```
LefterisJP, line 11: Isn't this copying the structure from yesterday's PR? Just re-use it. Why duplicate it here?

rotkehlchen/tasks/calendar.py
```diff
@@ -0,0 +1,43 @@
...
+    reminder_id: int | None
+    secs_before: int | None
+
+    def serialize(self) -> dict[str, str | int]:
+        data = self.event.serialize()
+        if self.secs_before != 0:  # it was a user created reminder instead of the basic one
```
LefterisJP, line 17: Why this distinction? It makes no sense. Even if it's created by the user if it's automatic it should always have a "secs_before" entry.

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -667,3 +672,120 @@ def test_maybe_update_aave_v3_underlying_assets(
...
+@pytest.mark.freeze_time('2023-04-16 22:31:11 GMT')
+@pytest.mark.parametrize('legacy_messages_via_websockets', [True])
+def test_send_ws_calendar_reminder(
+        rotkehlchen_api_server: 'APIServer',
+        websocket_connection: WebsocketReader,
+        legacy_messages_via_websockets: bool,  # pylint: disable=unused-argument
+        freezer,
+) -> None:
+    """
+    Check that reminders work correctly by:
```
LefterisJP, line 687:
```suggestion
   Test that reminders work correctly by:
```

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -667,3 +672,120 @@ def test_maybe_update_aave_v3_underlying_assets(
...
+    - Check we get the notifications
+    - Check we remove the reminders once fired
+    - Check starting the app after an event with a reminder passes triggers the reminder
```
LefterisJP, line 690:
```suggestion
    - Checking we get the notifications
    - Checking we remove the reminders once fired
    - Checking that starting the app after an event with a reminder passes triggers the reminder
```

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -667,3 +672,120 @@ def test_maybe_update_aave_v3_underlying_assets(
...
+    }
+
+    # check that the reminder got deleted
+    with database.conn.read_ctx() as cursor:
+        assert cursor.execute(
+            'SELECT COUNT(*) FROM calendar_reminders WHERE event_id=1',
+        ).fetchone()[0] == 0
+
+    # move to the timestamp to trigger the second reminder but we are after
+    # the event. We should get the notification
```
LefterisJP, line 763:
```suggestion
    # move to the timestamp to trigger the second reminder. Even though we are after
    # the event, we should still get the notification
```

LefterisJP (commented):
Some comments/questions. Also you have forgotten all the remaining comments from: https://github.com/rotki/rotki/pull/7748#pullrequestreview-2003104850

rotkehlchen/tasks/calendar.py
yabirgb: is not the same. the other used only the event id and this one has all the information about the event since the frontend doesn't keep it in memory

LefterisJP: So can't you inherit yesterday's one? Since it just has one extra attribute? Why duplicate?

yabirgb: It isn't an extra attribute. The other one uses event_id and this one uses the whole calendar event. I can abstract the two fields but I'm not sure if it hurts that much

docs/api.rst
```diff
@@ -13754,7 +13753,7 @@ Managing calendar reminders
 
 .. http:put:: /api/(version)/calendar/reminders
 
-  Doing a PUT request on this endpoint will allow to create a new calendar reminder
+  Doing a PUT request on this endpoint will allow to create new calendar entries. If any of the entries fails to get added the key ``failed`` will be populated in the response.
```
LefterisJP, line 13756:
```suggestion
  Doing a PUT request on this endpoint will allow to create new calendar reminder entries. If any of the entries fails to get added the key ``failed`` will be populated in the response.
```

docs/api.rst
```diff
@@ -13765,8 +13764,10 @@ Managing calendar reminders
       Content-Type: application/json;charset=UTF-8
 
       {
-        "secs_before": 1869737344,
-        "event_id": 1
+        "reminders": [
+         {"secs_before": 1869737344, "event_id": 1},
+         {"secs_before": 1869737344, "event_id": 100},
```
LefterisJP, line 13769: probably this comma is why docs building failed

rotkehlchen/tasks/manager.py
```diff
@@ -807,6 +810,37 @@ def _maybe_query_monerium(self) -> Optional[list[gevent.Greenlet]]:
             method=monerium.get_and_process_orders,
         )]
 
+    def _maybe_trigger_calendar_reminder(self) -> Optional[list[gevent.Greenlet]]:
+        """Get upcoming reminders and maybe process them"""
+        with self.database.conn.read_ctx() as cursor:
```
LefterisJP, line 815: Two things about this function.

1. Shouldn't it run every X minutes? This will now get always scheduled and at least the sqlite query will run continuously. I know this should indeed run frequently but wondering if perhaps a variable in the task manager (in-memory variable) to remember last query would be fine, to say do this check every 5 mins, and not every time this is scheduled. Not sure here ...
2. How/when do events that have passed get deleted?

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -667,3 +672,99 @@ def test_maybe_update_aave_v3_underlying_assets(
...
+        rotkehlchen_api_server: 'APIServer',
+        websocket_connection: WebsocketReader,
+        legacy_messages_via_websockets: bool,  # pylint: disable=unused-argument
+        freezer,
+) -> None:
+    """
+    Test that reminders work correctly by:
+    - Checking we get the notifications
+    - Checking we remove the reminders once fired
+    - Checking that starting the app after an event with a reminder passes triggers the reminder
```
LefterisJP, line 690:
```suggestion
    - Checking that starting the app after an event with a reminder passes, triggers the reminder
```

rotkehlchen/tasks/manager.py
yabirgb: The events get deleted after the ws is sent in the function that gets called?

LefterisJP: I don't see that.

rotkehlchen/tasks/manager.py
```diff
@@ -807,6 +812,41 @@ def _maybe_query_monerium(self) -> Optional[list[gevent.Greenlet]]:
...
+
+        with self.database.conn.read_ctx() as cursor:
+            cursor.execute(
+                'SELECT event.identifier, event.name, event.description, event.counterparty, '
+                'event.timestamp, event.address, event.blockchain, event.color, '
+                'reminder.identifier, reminder.secs_before FROM calendar_reminders AS reminder '
+                'LEFT JOIN calendar AS event '
+                'ON reminder.event_id = event.identifier WHERE '
+                '? > event.timestamp - reminder.secs_before',
+                (ts_now(),),
```
LefterisJP, line 828:
```suggestion
                (now,),
```

LefterisJP (approved):
LGTM.

PR #7757, opened 2024-04-17
feat: revamp notification and handle reminder

lukicenturi:
Closes #(issue_number)

## Checklist


frontend/app/src/composables/message-handling.ts
```diff
@@ -209,6 +211,47 @@ export function useMessageHandling() {
...
+    let message = '';
+
+    if (data.address && data.blockchain)
+      message += `Account: ${data.address} (${data.blockchain}) \n`;
+
+    if (data.counterparty)
+      message += `Counterparty: ${data.counterparty} \n`;
+
+    if (data.description)
+      message += `Description: ${data.description}`;
```
kelsos, line 235: don't we have any resources for these? can't we use them instead of hardcoding?

kelsos (commented):
only a small comment about the hardcoded text otherwise it looks good

PR #7783, opened 2024-04-19
Feat (Calendar): Reminders for ENS expiry

### LINKED ISSUE #7074: Calendar of on chain events
## Abstract

It is common for users to have special dates in the calendar where they have to take actions. Examples of those are for example dates to renew ENS domains, CRV votes, yst-eth votes... If the user wants to keep track of all of them it has to keep a manual entry for information that can be easily scheduled with data from the chain.

## Specification

- Create a list of events where we link a string, and address and a timestamp and optionally counterparty
- provide an api to query the events (probably allowing to filter by timestamp period and address)
- for certain decoded events we can detect actions that need to be taken in the future and add them automatically to the calendar
- allow the user to manually add entries to the calendar

OjusWiZard:
closes #7074

## Checklist

- [x] Add `auto_create_calendar_reminders` setting. Defaults to `True`.
- [x] Add task to auto create calendar reminder, first is ENS.

### DIFF
--- a/rotkehlchen/tasks/calendar.py
+++ b/rotkehlchen/tasks/calendar.py
@@ -1,10 +1,33 @@
+import logging
 from dataclasses import dataclass
+from typing import TYPE_CHECKING
 
 from rotkehlchen.api.websockets.typedefs import WSMessageType
-from rotkehlchen.db.calendar import BaseReminderData, CalendarEntry
+from rotkehlchen.chain.ethereum.modules.ens.constants import CPT_ENS
+from rotkehlchen.chain.evm.types import string_to_evm_address
+from rotkehlchen.constants.timing import DAY_IN_SECONDS, WEEK_IN_SECONDS
+from rotkehlchen.db.calendar import (
+    BaseReminderData,
+    CalendarEntry,
+    CalendarFilterQuery,
+    DBCalendar,
+    ReminderEntry,
+)
 from rotkehlchen.db.dbhandler import DBHandler
+from rotkehlchen.db.filtering import EvmEventFilterQuery
+from rotkehlchen.db.history_events import DBHistoryEvents
+from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.logging import RotkehlchenLogsAdapter
+from rotkehlchen.types import ChainID, OptionalBlockchainAddress, Timestamp
 from rotkehlchen.user_messages import MessagesAggregator
 from rotkehlchen.utils.misc import ts_now
+from rotkehlchen.utils.mixins.customizable_date import CustomizableDateMixin
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+if TYPE_CHECKING:
+    from rotkehlchen.history.events.structures.evm_event import EvmEvent
 
 
 @dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=True)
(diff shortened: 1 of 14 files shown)


yabirgb:
Comment on discord

```
You need to add a:

user setting that would decide if reminders are created automatically for the calendar event entry or not. Default should be YES.
create a task that will go through specific history events and create calendar event entries (if they are gonna be in the future and if they don't already exist)
start with ENS name registration and ENS name updating/re-registration. Give it either no automatic reminders (if setting is set not to) or if yes dunno ... let's say 1 week before and 1 day before?
```

https://discord.com/channels/657906918408585217/1080783409640648765/1230139087994749021

rotkehlchen/tasks/calendar.py
```diff
@@ -47,3 +70,98 @@ def delete_past_calendar_entries(database: DBHandler) -> None:
...
+
+    _, failed_to_add = db_calendar.create_reminder_enties(reminders=[
+        ReminderEntry(
+            identifier=idx,  # this is only used for logging below, it's auto generated in db
+            event_id=calendar_identifier,
+            secs_before=secs_before,
+        )
+        for idx, (calendar_identifier, _) in enumerate(ens_to_event.values())
+        for secs_before in (WEEK_IN_SECONDS, DAY_IN_SECONDS)
+    ])
```
yabirgb, line 161: mmm checking it the `calendar_reminders` doesn't have any kind of check so you need to avoid creating reminders for calendar entries that already have them

rotkehlchen/tasks/manager.py
```diff
@@ -817,6 +818,27 @@ def _maybe_query_monerium(self) -> Optional[list[gevent.Greenlet]]:
...
+            ):
+                return None
+
+        return [self.greenlet_manager.spawn_and_track(
+            after_seconds=None,
+            task_name='Maybe create ENS reminders',
+            exception_is_error=True,
+            method=maybe_create_ens_reminders,
+            database=self.database,
+        )]
```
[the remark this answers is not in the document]
yabirgb: As I understood it the task should be somethign like

```
for event in all_events:
     check_calendar_automations(event)


...


def check_calendar_automations(event):
    for calendar_check in (ens_calendar_check,):
          calendar_check(event)

```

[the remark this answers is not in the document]
yabirgb: but let's sync in discortd about it because maybe I misunderstood something

rotkehlchen/tasks/calendar.py
```diff
@@ -47,3 +70,98 @@ def delete_past_calendar_entries(database: DBHandler) -> None:
...
         )
+
+
+def maybe_create_ens_reminders(database: DBHandler) -> None:
+    """Check ENS registration and renewal history events and create reminders if needed."""
+    db_calendar = DBCalendar(database=database)
+    db_history_events = DBHistoryEvents(database=database)
+    customizable_date = CustomizableDateMixin(database=database)
+    with database.conn.read_ctx() as cursor:
+        registrations: list['EvmEvent'] = db_history_events.get_history_events(
```
yabirgb, line 81: is the type hint needed? You provided `EvmEventFilterQuery` that should be enough

rotkehlchen/tasks/calendar.py
```diff
@@ -47,3 +70,98 @@ def delete_past_calendar_entries(database: DBHandler) -> None:
...
+            cursor=cursor,
+            has_premium=True,  # to not hit the free limit while internal processing
+            group_by_event_ids=False,
+            filter_query=EvmEventFilterQuery.make(
+                and_op=True,
+                counterparties=[CPT_ENS],
+                event_types=[HistoryEventType.RENEW],
+                event_subtypes=[HistoryEventSubType.NONE],
+            ),
+        )
```
yabirgb, line 102: you can put this in a for loop whith 3 params (counterparty, type, subtype

rotkehlchen/tasks/calendar.py
```diff
@@ -47,3 +70,98 @@ def delete_past_calendar_entries(database: DBHandler) -> None:
...
+            ens_event.extra_data['expires'] > ts_now()  # those with expiry in the future
+        ):
+            ens_name: str = ens_event.extra_data['fullname']
+            ens_expires: Timestamp = ens_event.extra_data['expires']
+            blockchain = ChainID.deserialize(ens_event.location.to_chain_id()).to_blockchain()
+            if (calendar_entries := db_calendar.query_calendar_entry(
+                filter_query=CalendarFilterQuery.make(
+                    and_op=True,
+                    name=ens_name,
+                    addresses=[OptionalBlockchainAddress(
```
yabirgb, line 123: an ethereum address can have multiple ens names. You need to check the timestamp too

OjusWiZard: using the name filter as `f'{ens_name} expiry'` so that should be enough to differentiate them

rotkehlchen/tasks/calendar.py
```diff
@@ -47,3 +70,96 @@ def delete_past_calendar_entries(database: DBHandler) -> None:
...
+                filter_query=EvmEventFilterQuery.make(
+                    and_op=True,
+                    counterparties=[CPT_ENS],
+                    event_types=[event_type],
+                    event_subtypes=[event_subtype],
+                ),
+            ))
+
+    with database.conn.read_ctx() as cursor:
+        blockchain_accounts = database.get_blockchain_accounts(cursor=cursor)
```
yabirgb, line 99: exit earlier `if len(ens_events) == 0`

rotkehlchen/tasks/calendar.py
```diff
@@ -47,3 +70,96 @@ def delete_past_calendar_entries(database: DBHandler) -> None:
             'DELETE FROM calendar WHERE timestamp < ? AND auto_delete=1',
             (ts_now(),),
         )
+
+
+def maybe_create_ens_reminders(database: DBHandler) -> None:
+    """Check ENS registration and renewal history events and create reminders if needed."""
+    db_calendar = DBCalendar(database=database)
+    db_history_events = DBHistoryEvents(database=database)
+    customizable_date = CustomizableDateMixin(database=database)
```
yabirgb, line 79: can you maybe initialize the things that you need only if there are events later? if there are no events you will create things that won't be used

rotkehlchen/tasks/calendar.py
```diff
@@ -47,3 +70,96 @@ def delete_past_calendar_entries(database: DBHandler) -> None:
...
+            ens_to_event[ens_name] = (calendar_entry_id, ens_event)
+
+    _, failed_to_add = db_calendar.create_reminder_enties(reminders=[
+        ReminderEntry(
+            identifier=idx,  # this is only used for logging below, it's auto generated in db
+            event_id=calendar_identifier,
+            secs_before=secs_before,
+        )
+        for idx, (calendar_identifier, _) in enumerate(ens_to_event.values())
+        if len(db_calendar.query_reminder_entry(event_id=calendar_identifier)['entries']) == 0  # if not already present  # noqa: E501
```
yabirgb, line 155: I would do just a count since here you are doing deserializations that you don't need if you are just counting

yabirgb: Also shoudn't you check if the `event_id=calendar_identifier` but also if secs_before is the amount that you are adding the reminder for? With the logic as it is you could be adding one entry and then the second won't be added because you added already one and the count == 1

rotkehlchen/tasks/calendar.py
```diff
@@ -47,3 +70,96 @@ def delete_past_calendar_entries(database: DBHandler) -> None:
...
+        for secs_before in (WEEK_IN_SECONDS, DAY_IN_SECONDS)
+    ])
+
+    if len(failed_to_add) == 0:
+        return
+
+    enses = list(ens_to_event.keys())
+    log.error(
+        f'Failed to add the ENS expiry reminder(s) for {[enses[i] for i in failed_to_add]}.',
+    )
```
yabirgb, line 165: not sure how this login works. The failed to add gives you the calendar's event_id so not sure how you use the list here. I believe as you have now only works because you have in the test a small amount of events and this will raise indexerror in the code

rotkehlchen/tasks/calendar.py
```diff
@@ -47,3 +70,96 @@ def delete_past_calendar_entries(database: DBHandler) -> None:
...
+    ens_to_event: dict[str, tuple[int, EvmEvent]] = {}
+    for ens_event in ens_events:
+        if not (
+            ens_event.location_label is not None and
+            ens_event.extra_data is not None and
+            'fullname' in ens_event.extra_data and
+            'expires' in ens_event.extra_data and (
+                user_address := string_to_evm_address(ens_event.location_label)
+            ) in blockchain_accounts.get(
+                blockchain := ChainID.deserialize(ens_event.location.to_chain_id()).to_blockchain(),  # noqa: E501
```
yabirgb, line 111: does .get return empty list or None here?

rotkehlchen/tasks/calendar.py
```diff
@@ -47,3 +70,96 @@ def delete_past_calendar_entries(database: DBHandler) -> None:
...
+        blockchain_accounts = database.get_blockchain_accounts(cursor=cursor)
+
+    current_ts = ts_now()
+    ens_to_event: dict[str, tuple[int, EvmEvent]] = {}
+    for ens_event in ens_events:
+        if not (
+            ens_event.location_label is not None and
+            ens_event.extra_data is not None and
+            'fullname' in ens_event.extra_data and
+            'expires' in ens_event.extra_data and (
```
yabirgb, line 108:
```suggestion
            (expires_at := ens_event.extra_data.get('expires')) is not None (
```

and you can use `expires_at` later

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -843,3 +846,80 @@ def test_calendar_entries_get_deleted(
...
+    ]
+
+    task_manager.potential_tasks = [task_manager._maybe_create_calendar_reminder]
+    task_manager.schedule()
+    if len(task_manager.running_greenlets):
+        gevent.joinall(task_manager.running_greenlets[task_manager._maybe_create_calendar_reminder])
+
+    if db_settings['auto_create_calendar_reminders'] is False:
+        assert calendar_db.query_calendar_entry(CalendarFilterQuery.make())['entries_total'] == 0
+        return
```
yabirgb, line 897: add a comment saying that you finish testing because of the setting configuration

yabirgb (commented):
There are some tests that failed before and you haven't adjusted. Please check them before pushing

rotkehlchen/tasks/calendar.py
OjusWiZard: I did not check the `secs_before` because it's possible that the user could have already set their own custom reminder, in that case, we do not add it. Also, the count should not increase because it's a list comprehension before the full query, and then all of them are added at once.

rotkehlchen/tasks/calendar.py
```diff
@@ -47,3 +70,101 @@ def delete_past_calendar_entries(database: DBHandler) -> None:
...
+    db_history_events = DBHistoryEvents(database=database)
+    ens_events: list['EvmEvent'] = []
+    with database.conn.read_ctx() as cursor:
+        for event_type, event_subtype in (
+            (HistoryEventType.TRADE, HistoryEventSubType.SPEND),
+            (HistoryEventType.RENEW, HistoryEventSubType.NONE),
+        ):
+            ens_events.extend(db_history_events.get_history_events(
+                cursor=cursor,
+                has_premium=True,  # to not hit the free limit while internal processing
```
LefterisJP, line 86:
```suggestion
                has_premium=True,  # not limiting here
```

rotkehlchen/tasks/calendar.py
```diff
@@ -47,3 +70,101 @@ def delete_past_calendar_entries(database: DBHandler) -> None:
...
+        else:  # else
+            calendar_entry_id = calendar_entries['entries'][0].identifier
+
+        if (
+            ens_name not in ens_to_event or
+            ens_expires > ens_to_event[ens_name][1].extra_data['expires']  # type: ignore[index]  # extra_data is not None, checked above
+        ):  # insert mapping for the latest expiry timestamp
+            ens_to_event[ens_name] = (calendar_entry_id, ens_event)
+
+    _, failed_to_add = db_calendar.create_reminder_enties(reminders=[
```
LefterisJP, line 150: 1. What happens if reminder already exists?
2. The function you call here has a typo

`create_reminder_enties` -> `create_reminder_entries`

OjusWiZard: if the reminder entry already exists then we don't add it for that calendar entry, because it's possible that the user could have already set their own custom reminder
it's checked by this line
```py
if db_calendar.count_reminder_entries(event_id=calendar_identifier) == 0  # if not already present  # noqa: E501
```

LefterisJP (approved):
LGTM

PR #7792, opened 2024-04-20
Fix customized events fetching + zksync lite filtering

LefterisJP:
- Fix the issue where zksync lite filtering in history events did not work
- Customized events should not be filtered by chain ID but by location as events exist in many places and not just evm chains.


LefterisJP:
all green. merging

PR #7795, opened 2024-04-21
Decode harvest finance grain claim events

LefterisJP:
Also astract common code for MerkleDrop claim events by putting it in a common Decoder mixin


LefterisJP:
all green. Le merge.

PR #7808, opened 2024-04-22
Fix zksync lite decoded transactions returned number

LefterisJP:


LefterisJP:
All green. Will merge.

PR #7809, opened 2024-04-23
feat: change asset filter suggestion text

lukicenturi:
Instead of showing "no suggestions for asset", give hint that user need to type the symbol to see some suggestions.


kelsos (approved):
lgtm

PR #7814, opened 2024-04-23
Fix balancer trades of token to token

yabirgb:
Closes #(issue_number)

## Checklist


yabirgb:
image
With this changes all the cases that I was aware off get correctly processed

PR #7827, opened 2024-04-24
Update mappings with globaldb

yabirgb:
When filling the mappings for the globaldb those need to be checked locally. Also when checking the assets in tests we need to have an updated version of the mappings. This PR:

- Introduces a new script to populate the generated globaldb with the remote mappings
- Ensure that the insertion of an already existing mapping doesn't error (if we pull remote updates in a packaged globaldb this can happen)


yabirgb:
closing as it is not needed

PR #7828, opened 2024-04-25
fix(frontend): loading UI when asset reset is running

lukicenturi:
Closes #(issue_number)

## Checklist


kelsos (approved):
@lukicenturi looks ok but can you give a quick explanation on the matcher order change and why is it needed?

lukicenturi:
> @lukicenturi looks ok but can you give a quick explanation on the matcher order change and why is it needed?

So the example is we have filter `protocol` and `location`. Then I search by `l`

Previously it will show `protocol` before `location`. 
We didn't do any sorting there, just check if `text.includes(search)`.
This changes just add sorting using the levehnstein so the `location` will appear before `protocol`

PR #7833, opened 2024-04-26
fix(frontend): remove unneeded error on version update check setting

lukicenturi:
Remove unneded error on version update check setting. If the switch disabled, the error on the textfield should be removed.


kelsos (approved):
lgtm

PR #7842, opened 2024-04-26
Fix (Calendar): Update last entry for same ENS

OjusWiZard:
## Checklist

- [x] If a calendar entry is already present with outdated expiry, update it instead of creating a new one

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -22,6 +22,7 @@ Changelog
 * :feature:`-` The claim event of the Degen airdrop 2 will be correctly decoded.
 * :bug:`-` Binance balances will now include funding wallet's balances.
 * :feature:`-` rotki now supports Coinbase transactions using Advanced Trading.
+* :bug:`-` Velodrome and Aerodrome LP pools will now refresh properly.
 * :bug:`-` Fix a bug where the custom price is not applied in the balance view under the account table.
 * :bug:`-` Deleting an EVM RPC node will no longer fail sometimes with a "Can't delete etherscan node" error. Additionally non-mainnet etherscan nodes are no longer deletable.
 * :bug:`-` Nigerian Naira should be correctly detected in binance.
--- a/rotkehlchen/tasks/calendar.py
+++ b/rotkehlchen/tasks/calendar.py
@@ -6,6 +6,7 @@ from rotkehlchen.api.websockets.typedefs import WSMessageType
 from rotkehlchen.chain.ethereum.modules.ens.constants import CPT_ENS
 from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.constants.timing import DAY_IN_SECONDS, WEEK_IN_SECONDS
+from rotkehlchen.db.cache import DBCacheStatic
 from rotkehlchen.db.calendar import (
     BaseReminderData,
     CalendarEntry,
@@ -116,6 +117,7 @@ def maybe_create_ens_reminders(database: DBHandler) -> None:
         ):
             continue
 
+        calendar_entry_description = f'{ens_name} expires on {customizable_date.timestamp_to_date(ens_expires)}'  # noqa: E501
         if (calendar_entries := db_calendar.query_calendar_entry(
             filter_query=CalendarFilterQuery.make(
                 and_op=True,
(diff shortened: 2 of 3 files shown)


rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -896,14 +897,19 @@ def test_maybe_create_calendar_reminder(
         assert calendar_db.query_calendar_entry(CalendarFilterQuery.make())['entries_total'] == 0
         return  # finish this test case because this setting will prevent creation of reminders
 
+    latest_expiry_of_ens: dict[str, Timestamp] = defaultdict(lambda: Timestamp(0))
+    for event in ens_events:
+        if event.extra_data is not None and latest_expiry_of_ens[event.extra_data['name']] < event.extra_data['expires']:  # noqa: E501
+            latest_expiry_of_ens[event.extra_data['name']] = event.extra_data['expires']
```
yabirgb, line 903: this is something that you know beforehand when creating the test right? Instead of creating dinamically the test create it manually. Sometimes there are bugs that are not spotted because you expect data that you don't have

rotkehlchen/tasks/calendar.py
```diff
@@ -138,10 +138,25 @@ def maybe_create_ens_reminders(database: DBHandler) -> None:
...
-        ) > 0:  # and already has a reminder entry
-            continue  # we don't add any automatic reminders
+        else:  # else calendar entry already exists
+            calendar_entry = calendar_entries['entries'][0]
+            if ens_expires > calendar_entry.timestamp:  # if a later expiry is found
+                db_calendar.update_calendar_entry(CalendarEntry(
+                    identifier=calendar_entry.identifier,
+                    name=f'{ens_name} expiry',
+                    timestamp=ens_expires,  # update the calendar entry
+                    description=f'{ens_name} expires on {customizable_date.timestamp_to_date(ens_expires)}',  # noqa: E501
```
yabirgb, line 148: can you assign the description to a variable? The reason is that when you modify it you don't need to change multiple places

rotkehlchen/tasks/manager.py
```diff
@@ -831,6 +831,13 @@ def _maybe_create_calendar_reminder(self) -> Optional[list[gevent.Greenlet]]:
         ):
             return None
 
+        with self.database.conn.write_ctx() as write_cursor:
+            self.database.set_static_cache(  # remember last task ran
+                write_cursor=write_cursor,
+                name=DBCacheStatic.LAST_CREATE_REMINDER_CHECK_TS,
+                value=ts_now(),
+            )
```
yabirgb, line 839: This should be done in the task itself. The reason is that if for some reason the task dies before executing (say that you close the app) then it won't have executed but the timestamp would have been recorded

```suggestion
            self.database.set_static_cache(
                write_cursor=write_cursor,
                name=DBCacheStatic.LAST_CREATE_REMINDER_CHECK_TS,
                value=ts_now(),
            )
```

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` Velodrome and Aerodrome LP pools will now refresh properly.
```
yabirgb, line 5: can you put it with the other bugfixes? To not have them mixed features and bugs

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -923,3 +929,8 @@ def test_maybe_create_calendar_reminder(
         assert reminders[0].event_id == reminders[1].event_id == calendar_entry.identifier
         assert reminders[0].secs_before == DAY_IN_SECONDS
         assert reminders[1].secs_before == WEEK_IN_SECONDS
+
+    with task_manager.database.conn.read_ctx() as cursor:
+        assert task_manager.database.get_static_cache(
+            cursor=cursor, name=DBCacheStatic.LAST_CREATE_REMINDER_CHECK_TS,
+        ) is not None
```
yabirgb, line 936:
```suggestion
        assert task_manager.database.get_static_cache(
            cursor=cursor, name=DBCacheStatic.LAST_CREATE_REMINDER_CHECK_TS,
        ) - ts_now() < 5  # executed recently
```

LefterisJP (approved):
Can see nothing wrong. If this fixes the problem you saw merging it.

PR #7844, opened 2024-04-26
Tx deletion and more

LefterisJP:


LefterisJP:
not finished. Need to also make it delete history events of related transactions. But opened it to check tests and show API

LefterisJP:
all green. Merging.

PR #7862, opened 2024-04-30
fix(backend): update counterparty image and icon

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/evm/decoding/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/decoder.py
@@ -150,7 +150,7 @@ class EVMTransactionDecoder(ABC):
         decoder that should be included for this decoder modules.
         """
         self.database = database
-        self.misc_counterparties = [CounterpartyDetails(identifier=CPT_GAS, label='gas', image='gas.svg')] + misc_counterparties  # noqa: E501
+        self.misc_counterparties = [CounterpartyDetails(identifier=CPT_GAS, label='gas', icon='fire-line')] + misc_counterparties  # noqa: E501
         self.evm_inquirer = evm_inquirer
         self.transactions = transactions
         self.msg_aggregator = database.msg_aggregator
--- a/rotkehlchen/chain/evm/decoding/types.py
+++ b/rotkehlchen/chain/evm/decoding/types.py
@@ -2,6 +2,22 @@ from typing import NamedTuple
 
 
 class CounterpartyDetails(NamedTuple):
+    """
+    Details for a counterparty returned to the api consumer.
+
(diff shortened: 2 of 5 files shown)


rotkehlchen/api/rest.py
```diff
@@ -4446,11 +4446,9 @@ def get_evm_counterparties_details(self) -> Response:
         removing duplicates.
         """
         return api_response(
-            result=process_result(
-                _wrap_in_ok_result(
-                    result=list(self.rotkehlchen.chains_aggregator.get_all_counterparties()),
-                ),
-            ),
+            result=process_result(_wrap_in_ok_result(
```
LefterisJP, line 4449: what did you do here? What changed?

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -150,7 +150,7 @@ def __init__(
         decoder that should be included for this decoder modules.
         """
         self.database = database
-        self.misc_counterparties = [CounterpartyDetails(identifier=CPT_GAS, label='gas', image='gas.svg')] + misc_counterparties  # noqa: E501
+        self.misc_counterparties = [CounterpartyDetails(identifier=CPT_GAS, label='gas', icon='fire-line')] + misc_counterparties  # noqa: E501
```
LefterisJP, line 153: is there any difference to before in production?

rotkehlchen/chain/evm/decoding/types.py
```diff
@@ -2,6 +2,21 @@
 
 
 class CounterpartyDetails(NamedTuple):
+    """
+    details for a counterparty returned to the api consumer.
+    icon is used when the counterparty uses an icon from the icon package instead of
+    a preselected image.
```
LefterisJP, line 8:
```suggestion
    Details for a counterparty returned to the api consumer.
    
    The icon is used when the counterparty uses an icon from the icon package instead of
    a preselected image.
```

LefterisJP (commented):
Some questions. Also shouldn't there be any test?

rotkehlchen/chain/evm/decoding/decoder.py
lukicenturi: yes the icons wasn't showing in production because the image doesn't exist

rotkehlchen/api/rest.py
lukicenturi: before the image should exist.
now it can return either image or icon

LefterisJP (approved):
LGTM

PR #7863, opened 2024-04-30
Fix debug import

yabirgb:
- Ensure that all the types are imported
- If an event is already ignored don't fail and skip it instead

Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/accounting/debugimporter/json.py
+++ b/rotkehlchen/accounting/debugimporter/json.py
@@ -15,6 +15,7 @@ from rotkehlchen.errors.misc import InputError
 from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.exchanges.data_structures import AssetMovement, Loan, MarginPosition, Trade
 from rotkehlchen.history.events.structures.base import HistoryEvent
+from rotkehlchen.history.events.structures.evm_event import EvmEvent
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


lukicenturi:
works fine for me

rotkehlchen/accounting/debugimporter/json.py
```diff
@@ -51,18 +52,23 @@ def import_history_debug(
...
+                        events.append(MarginPosition.deserialize(event))
+                    case AccountingEventType.LOAN:
+                        events.append(Loan.deserialize(event))
+                    case AccountingEventType.STAKING:
+                        events.append(ValidatorDailyStats.deserialize(event))
+                    case AccountingEventType.HISTORY_EVENT:
+                        events.append(HistoryEvent.deserialize(event))
+                    case AccountingEventType.TRANSACTION_EVENT:
+                        events.append(EvmEvent.deserialize(event))
+                    case AccountingEventType.FEE | AccountingEventType.PREFORK_ACQUISITION:
```
LefterisJP, line 70: if a case is not covered does it warn you? And who warns? mypy? or the interpreter?

LefterisJP (approved):
lgtm

rotkehlchen/accounting/debugimporter/json.py
yabirgb: I changed to match case because I recalled the interpreter warning me but seems that if the optinion is missing it becomes a no-op as per the docs and mypy doesn't detect it properly yet https://github.com/python/mypy/issues/13597

PR #7865, opened 2024-04-30
fix: exchange input for cex mapping

lukicenturi:
Closes #(issue_number)

## Checklist


kelsos:
@lukicenturi what are we fixing here exactly?

lukicenturi:
> @lukicenturi what are we fixing here exactly?

so the `exchange` data from `locations` endpoint.
In the exchange form I want to only show the ones with api key
while in the manage cex mapping, I want to show all exchange data.

kelsos:
OK so on my side it looks fine if it's ok with @yabirgb too we can proceed and merge.

yabirgb (approved):
L
G
T
M

PR #7869, opened 2024-04-30
Fix (_maybe_create_calendar_reminder): Include it (PR of the day)

OjusWiZard:


rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -58,6 +58,16 @@
...
 
 
+def test_potential_maybe_schedule_task(task_manager: TaskManager):
+    """Check that all the _maybe_... tasks are included in the potential tasks."""
+    tasks, count = {function.__name__ for function in task_manager.potential_tasks}, 0
+    for function in dir(task_manager):
+        if function.startswith('_maybe_'):
+            count += 1
+            assert function in tasks
+    assert count == 26
```
yabirgb, line 68: the count I believe is not that useful and will be a pain because we will forget each time we need to add a new task.
I would do

```suggestion
    assert all(function in tasks for function in dir(task_manager) if function.startswith('_maybe_'))
```

LefterisJP (approved):
lol, lgtm

PR #7848, opened 2024-04-28
Fix compound-v3 handling

gianluca-pub-dev:
Current compound-v3 handling was using the deposit/borrow functions in case of supply/withdraw collateral and the opposite for supply/withdraw. This resulted in compound events not detected.

Noticed the issue while testing for Compound-V3 WETH contract 0xA17581A9E3356d9A858b789D68B4d866e593aE94, and the interactions for the 4 supported method IDs:
COMPOUND_V3_SUPPLY
COMPOUND_V3_SUPPLY_COLLATERAL
COMPOUND_V3_WITHDRAW
COMPOUND_V3_WITHDRAW_COLLATERAL

Closes #(issue to be created)

## Checklist

- [NO] The PR modified the frontend, and updated the [user guide](https://github.com/rotki/rotki/blob/develop/docs/usage_guide.rst) to reflect the changes.

PR #7872, opened 2024-05-01
Fix (Compound v3): Remap the event decoders

OjusWiZard:
Signed-off-by: OjusWiZard <ojuswimail@gmail.com>

- [x] Decode seperate events for adding and removal of collateral assets.
- [x] Change the common decoding logic from `_decode_supply_or_withdraw_event` and `_decode_repay_or_borrow_event`  to `_decode_supply_or_repay_event` and `_decode_withdraw_or_borrow_event`
- [x] Fix ~an~ two underlying_asset mapping for `cUSDCv3` on arbitrum. ~Checking if any other is wrong...~
- [x] Adjust the test for wrongly decoded events

SQL diff for global.db
```sql
UPDATE underlying_tokens_list SET identifier='eip155:42161/erc20:0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8' WHERE rowid=193;
UPDATE underlying_tokens_list SET identifier='eip155:8453/erc20:0xd9aAEc86B65D86f6A7B5B1b0c42FFA531710b6CA' WHERE rowid=196;
```

### DIFF
--- a/rotkehlchen/chain/evm/decoding/compound/v3/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/compound/v3/decoder.py
@@ -2,9 +2,12 @@ import logging
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
 
+from rotkehlchen.accounting.structures.balance import Balance
 from rotkehlchen.assets.asset import EvmToken
 from rotkehlchen.assets.utils import get_or_create_evm_token
 from rotkehlchen.chain.ethereum.utils import asset_normalized_value
+from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
+from rotkehlchen.chain.evm.decoding.constants import ERC20_OR_ERC721_TRANSFER
 from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface
 from rotkehlchen.chain.evm.decoding.structures import (
     DEFAULT_DECODING_OUTPUT,
@@ -13,6 +16,7 @@ from rotkehlchen.chain.evm.decoding.structures import (
     DecodingOutput,
 )
 from rotkehlchen.chain.evm.decoding.types import CounterpartyDetails
+from rotkehlchen.chain.evm.decoding.utils import maybe_reshuffle_events
 from rotkehlchen.constants.resolver import evm_address_to_identifier
 from rotkehlchen.errors.asset import UnknownAsset, WrongAssetType
 from rotkehlchen.globaldb.handler import GlobalDBHandler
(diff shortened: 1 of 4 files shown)


rotkehlchen/chain/evm/decoding/compound/v3/decoder.py
```diff
@@ -209,10 +248,69 @@ def _decode_repay_or_borrow_event(self, context: DecoderContext) -> DecodingOutp
...
+                    address=event.address,
+                )
+                break
+        else:
+            log.error(
+                f'Could not find any compound v3 supply/withdraw collateral event in tx '
+                f'{context.transaction.tx_hash.hex()}.',
+            )
+
+        context.decoded_events.append(collateral_event)
```
yabirgb, line 309: I might be missing something but I believe there is a scenario in the code if something goes bad where collateral_event is udnefined here

yabirgb (commented):
If the user checked and is fine I guess we can go with it

rotkehlchen/chain/evm/decoding/compound/v3/decoder.py
OjusWiZard: right, adding `return DEFAULT_DECODING_OUTPUT` now

gianluca-pub-dev (commented):
Just as contribute from a manual test done using this WIP PR.

I tested with some transaction and it didn't worked as expected: 
1. supply collateral to compound-v3 not detected but reported as "generic send" to compound-v3 token contract
2. Borrow detected instead
3. As in the previous aave case, there is a flash loan from another procotol detected erroneoulsy as "withdraw" from compound
4. Repay event is detected
5. Withdraw collateral is not detected, it is shown as generic "receive" from the compound token contract

If it could be useful as reference, the same transactions are working with the draft PR: https://github.com/rotki/rotki/pull/7848

OjusWiZard:
@gianluca-pub-dev I've pushed the fix based on these two transactions that you shared on discord
https://etherscan.io/tx/0xbf59eb763331573f9be61a75d6fad7d280aefc5202bc5df47090fdef61b9acbb
https://etherscan.io/tx/0xc7bec769b7ecc66a7450048e06a6e5fe2917fba668990a8573c393692da7cc14
Now it decodes the supply/withdrawal collateral events and flash loans should not show as compound events.
Thanks for helping with these cases!

gianluca-pub-dev:
> @gianluca-pub-dev I've pushed the fix based on these two transactions that you shared on discord https://etherscan.io/tx/0xbf59eb763331573f9be61a75d6fad7d280aefc5202bc5df47090fdef61b9acbb https://etherscan.io/tx/0xc7bec769b7ecc66a7450048e06a6e5fe2917fba668990a8573c393692da7cc14 Now it decodes the supply/withdrawal collateral events and flash loans should not show as compound events. Thanks for helping with these cases!

Thanks @OjusWiZard , I tried again pulling this branch and this time the compound-v3 transactions in my wallet history are recognized! So I confirm this current commit works for thise type of compound-v3 transactions.

rotkehlchen/chain/evm/decoding/compound/v3/decoder.py
```diff
@@ -106,47 +110,75 @@ def decode_reward_claim(self, context: DecoderContext) -> DecodingOutput:
...
 
+        receiving_ctoken = False
+        for tx_log in context.all_logs:
+            if (
+                tx_log.address == compound_token.evm_address and
+                tx_log.topics[0] == ERC20_OR_ERC721_TRANSFER and
+                hex_or_bytes_to_address(tx_log.topics[1]) == ZERO_ADDRESS and  # from
+                hex_or_bytes_to_address(tx_log.topics[2]) == hex_or_bytes_to_address(context.tx_log.topics[2])  # to  # noqa: E501
+            ):
+                receiving_ctoken = True
```
gianluca-pub-dev, line 154: Hello @OjusWiZard , what about the compound-v3 transactions that are not generating any ctoken? I have seen in some compound transaction of [WETH contract](https://etherscan.io/address/0xA17581A9E3356d9A858b789D68B4d866e593aE94) that there is no ctoken, generated or transferred, for the supply/borrow/repay/withdraw operations. I am not sure if I fully got this code section, but it seems to me that this line can be reached even if there is any other protocol token generation in the same transaction (and maybe this is the reason it is working with the transactions I tried, even if there is no ctoken generation there)

OjusWiZard: yes, It will only be true when new compound tokens are minted and transferred to the tracked user. That only happens in the case of supplying. We compare the tx_log.address with compound_token address because the transfer log will emit from the compound token, and check tx_log.topics[2] with context.tx_log.topics[2] which means the token is being transfer to the same address for whome the supplying is done.

rotkehlchen/chain/evm/decoding/compound/v3/decoder.py
```diff
@@ -106,47 +110,74 @@ def decode_reward_claim(self, context: DecoderContext) -> DecodingOutput:
...
```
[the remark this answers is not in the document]
yabirgb: this doesn't seem to have been done

rotkehlchen/chain/evm/decoding/compound/v3/decoder.py
```diff
@@ -209,10 +251,86 @@ def _decode_repay_or_borrow_event(self, context: DecoderContext) -> DecodingOutp
...
+                event.asset == collateral_asset and
+                event.address == compound_token.evm_address
+            ):
+                transfer_event = event
+                event.counterparty = CPT_COMPOUND_V3
+                if event.event_type == HistoryEventType.SPEND:
+                    notes_action = 'Enable'
+                    event.event_type = HistoryEventType.DEPOSIT
+                    event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
+                    event.notes = f'Deposit {collateral_amount} {collateral_asset.symbol} into compound v3'  # noqa: E501
```
yabirgb, line 303: Can you change here and in the other places `compound v3` to `Compound v3`. TO match what we do in aave

rotkehlchen/chain/evm/decoding/compound/v3/decoder.py
yabirgb: Here this also seems to not have changed. TGhe collateral_event could be missing here where it is used

LefterisJP (commented):
Question smaller things

rotkehlchen/chain/evm/decoding/compound/v3/decoder.py
```diff
@@ -106,47 +110,76 @@ def decode_reward_claim(self, context: DecoderContext) -> DecodingOutput:
...
+                if receiving_ctoken:
                     event.event_type = HistoryEventType.DEPOSIT
                     event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
-                    event.notes = f'Deposit {amount} {underlying_token.symbol} to compound v3'
+                    event.notes = f'Deposit {amount} {underlying_token.symbol} into Compound v3'
+                    paired_event = event
                     action_from_event_type = HistoryEventType.RECEIVE
                     action_to_event_subtype = HistoryEventSubType.RECEIVE_WRAPPED
-                    action_to_notes = f'Receive {{amount}} {compound_token.symbol} from compound v3'  # {amount} to be replaced in post decoding  # noqa: E501
+                    action_to_notes = f'Receive {{amount}} {compound_token.symbol} from Compound v3'  # {amount} to be replaced in post decoding  # noqa: E501
```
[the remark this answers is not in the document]
OjusWiZard: We kept it consistent with aave. I see balancer also uppercased but some other lowercase.

LefterisJP (approved):
Lgtm

PR #7873, opened 2024-05-01
Fix for a problem with makerdao ETH vaults recognition of withdraw events

gianluca-pub-dev:
Event history was not reporting withdraw collateral events because ETH vaults were mapped to native ETH tokens instead than WETH


OjusWiZard:
Hey @gianluca-pub-dev thanks for raising this PR for this fix! Could you rebase it to `bugfixes` branch because we'll put this in the next bugfix release, also please fix the tests that broke in the CI.

gianluca-pub-dev:
> Hey @gianluca-pub-dev thanks for raising this PR for this fix! Could you rebase it to `bugfixes` branch because we'll put this in the next bugfix release, also please fix the tests that broke in the CI.

Hello @OjusWiZard , first of all, thanks for the review. For rebase I can do, and I will shortly, instead for updating the test cases I need to find the time, maybe during the weekend but I am not sure, because it needs time to me because I should analyze those test cases to see how they were written and the transactions. Let's see, but if urgent for your project, again I would ask to take over.

CC: @LefterisJP

Thanks

rotkehlchen/tests/unit/decoders/test_makerdao_sai.py
```diff
@@ -2299,31 +2300,23 @@ def test_makerdao_sai_proxy_interaction(ethereum_transaction_decoder):
...
-            timestamp=1579044372000,
-            location=Location.ETHEREUM,
-            event_type=HistoryEventType.WITHDRAWAL,
-            event_subtype=HistoryEventSubType.REMOVE_ASSET,
-            asset=A_ETH,
-            balance=Balance(FVal('0.022255814')),
-            location_label='0xca482bCd75A6E0697aD6A1732aa187310b8372Df',
-            notes='Withdraw 0.022255814 ETH from ETH-A MakerDAO vault',
-            counterparty=CPT_VAULT,
-            address=string_to_evm_address('0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2'),
```
gianluca-pub-dev, line 2324: @LefterisJP , uou can see from this, how the processing of events was bugged. This "0xC02aaA...." is the WETH contract address, and here the event was detected as one "withdraw from ETH-A MakerDAO vault" operation. Also notice the sequence_index number, 0. Basically the decoding here was working "by chance". I think the general decoding could be re-worked for being more deterministic (and reproducible). I have some idea in mind if you are interested on discussing possible sw requirements and architecture.

LefterisJP: The address field is manually added per decoder and means different things depending on the decoder. Coce can have bugs.

Always open to suggestions, but only things that make sense but for things that make sense you would need a perfect understanding of how the current thing we have works. And that should be in discord, not here.

rotkehlchen/chain/ethereum/modules/makerdao/decoder.py
```diff
@@ -557,31 +571,31 @@ def decode_saidai_migration(self, context: DecoderContext) -> DecodingOutput:
...
-            string_to_evm_address('0x3BC3A58b4FC1CbE7e98bB4aB7c99535e8bA9b8F1'): (self.decode_makerdao_vault_action, A_UNI.resolve_to_crypto_asset(), 'UNI-A'),  # noqa: E501
-            string_to_evm_address('0xFD5608515A47C37afbA68960c1916b79af9491D0'): (self.decode_makerdao_vault_action, A_RENBTC.resolve_to_crypto_asset(), 'RENBTC-A'),  # noqa: E501
-            string_to_evm_address('0x24e459F61cEAa7b1cE70Dbaea938940A7c5aD46e'): (self.decode_makerdao_vault_action, A_AAVE.resolve_to_crypto_asset(), 'AAVE-A'),  # noqa: E501
-            string_to_evm_address('0x885f16e177d45fC9e7C87e1DA9fd47A9cfcE8E13'): (self.decode_makerdao_vault_action, A_ETH_MATIC.resolve_to_crypto_asset(), 'MATIC-A'),  # noqa: E501
+            string_to_evm_address('0x3D0B1912B66114d4096F48A8CEe3A56C231772cA'): (self.decode_makerdao_vault_action, string_to_evm_address('0x3D0B1912B66114d4096F48A8CEe3A56C231772cA'), A_BAT.resolve_to_crypto_asset(), 'BAT-A'),  # noqa: E501
+            string_to_evm_address(MAKERDAO_GEM_JOIN_ETHA_ADDRESS): (self.decode_makerdao_vault_action, string_to_evm_address(MAKERDAO_GEM_JOIN_ETHA_ADDRESS), A_WETH.resolve_to_crypto_asset(), 'ETH-A'),  # noqa: E501
```
LefterisJP, line 575: I am not sure if simply replacing ETH with WETH works here. It may work for some cases and fail for others.

I think the solution is to work on the `decode_makerdao_vault_action` to treat ETH and WETH the same

rotkehlchen/tests/unit/decoders/test_makerdao.py
```diff
@@ -13,15 +18,19 @@
 
 
 @pytest.mark.vcr()
-@pytest.mark.parametrize('ethereum_accounts', [['0x648aA14e4424e0825A5cE739C8C68610e143FB79']])
+# Analyze EOA and user's DSProxy address
```
LefterisJP, line 21: no this is not the way. 

1. We don't put comments here, but only in docstrings of test
2. We never manually track DSProxies. This is not the solution. The solution is to make it get detected by rotki. Make sure the code that detects the proxies runs.

rotkehlchen/tests/unit/decoders/test_makerdao_sai.py
```diff
@@ -2299,31 +2300,23 @@ def test_makerdao_sai_proxy_interaction(ethereum_transaction_decoder):
 @pytest.mark.vcr(filter_query_parameters=['apikey'])
 @pytest.mark.parametrize('ethereum_accounts', [['0xca482bCd75A6E0697aD6A1732aa187310b8372Df']])
 def test_makerdao_sai_cdp_migration(ethereum_transaction_decoder, ethereum_accounts):
+    """Data taken from
+    https://etherscan.io/tx/0x03620c6bf5edb7a7935953337ffcfac70d631cf2012d6c80d36828d636063318
+    """
+
```
LefterisJP, line 2306: useless comment. 
```suggestion
    """

```

rotkehlchen/tests/unit/decoders/test_makerdao_sai.py
```diff
@@ -2299,31 +2300,23 @@ def test_makerdao_sai_proxy_interaction(ethereum_transaction_decoder):
...
     """Check that a Sai CDP migration is decoded properly"""
     tx_hex = deserialize_evm_tx_hash('0x03620c6bf5edb7a7935953337ffcfac70d631cf2012d6c80d36828d636063318 ')  # noqa: E501
     evmhash = deserialize_evm_tx_hash(tx_hex)
-    user_address = ethereum_accounts[0]
+    # TODO: ROTKI TEAM: The address you used here is the DSProxy address, not the user address.
```
LefterisJP, line 2310: there may be errors in this test, but manually tracking the DSProxy is not a solution

LefterisJP (commented):
So @gianluca-pub-dev we have more pressing issues we work on so we have no time to devote to this at the moment.

For when someone has time to look at it, please give us transaction hashes and what you expect vs what you get.

The issue you get with ds proxies should not be solved by tracking the DSproxy manually. That's not a solution. Also replacing ETH with WETH from a first look does not look correct to me. Botjh can be used. Your change would break for transactions that use pure ETH and get it auto wrapped to WETH.

rotkehlchen/tests/unit/decoders/test_makerdao_sai.py
gianluca-pub-dev: ok, that type of comment was already in other test cases in the same files, so I thought to add it also here for consistency. I can remove, no problem

rotkehlchen/tests/unit/decoders/test_makerdao_sai.py
gianluca-pub-dev: @LefterisJP , yes, but this is what you had already in this test case, so this is the proof that automatico recognition of DSProxy addresses is not working as you intended

rotkehlchen/tests/unit/decoders/test_makerdao.py
gianluca-pub-dev: 2. That currently doesn't work

rotkehlchen/chain/ethereum/modules/makerdao/decoder.py
gianluca-pub-dev: Without it, events are not correctly decoded when transactions are complex. In this way instead the events are "deterministic" and they can be compared with the logs on blockchain for reviewing. I will push another PR on top of this branch for a fix on the WETH module, just for discussing. Because I prefer to show what I mean using code.

rotkehlchen/tests/unit/decoders/test_makerdao_sai.py
```diff
@@ -2299,31 +2300,21 @@ def test_makerdao_sai_proxy_interaction(ethereum_transaction_decoder):
 @pytest.mark.vcr(filter_query_parameters=['apikey'])
 @pytest.mark.parametrize('ethereum_accounts', [['0xca482bCd75A6E0697aD6A1732aa187310b8372Df']])
```
gianluca-pub-dev, line 2301: Hello @LefterisJP , the account checked in this test case was already a DSProxy address, I didn't change it. https://etherscan.io/address/0xca482bCd75A6E0697aD6A1732aa187310b8372Df 

So it seems that rotki already needed to check the DSProxy address for decoding some event.

LefterisJP:
closing due to inactivity

PR #7874, opened 2024-05-02
fix: update message for add single account that already exist

lukicenturi:
How to replicate:

1. add an ETH account.
2. after it has been added, add again the same account, but turn off the "Enable for all EVM account"
3. the error message should be correct, previously it's not


tewshi (approved):
lgtm!

PR #7877, opened 2024-05-02
Ensure that a nonerc20token error

yabirgb:
We saw with kelsos that if the queried node was behind and couldn't find the pool token details. To prevent this we ensure that no error makes the logic fail and we return the pools that have been veried locally

Closes #(issue_number)

## Checklist


LefterisJP (approved):
LGTM

Issue #7570, opened 2024-03-02
Premium DB syncing allow choice of automatic handling sync errors


LefterisJP:
## Abstract

Quite often rotki may inform users it has not uploaded the local DB to the rotki server due to a size discrepancy. And ask user to confirm they want to push it.

This is how rotki syncing works with the server. We just give a message as an fyi. Since the database is nothing but a blob of encrypted data.

We know nothing else about it apart from its size and last time of upload. We can't see what's in there. So out of abundance of caution we just do not upload automatically to the server if there is a size difference and ask for action from you.

The size discrepancies can happen for various reasons, as periodic tasks run underneath either adding more data or cleaning up data

## Motivation

Some users have asked for the ability to automatically choose a strategy of uploading there via a setting and not to have to manually press force push whenever such a size discrepancy happens.

It's a bit less safe option as automatically overwriting data of bigger size can lead to loss of data if the local is an old DB from an older machine.

But if the users are aware of the dangers (say they only use one machine) then it's a positive quality of life change.

## Specification

- Add a setting to ignore size discrepancies in premium DB upload
- Use the value of this setting when syncing to decide what to do. Whether to force push or ask user for input
- Frontend; When asking user to force push due to size discrepancy, also have a checkbox dialog to "Do this action every time size discrepancy occurs" or something like that.

Mithil467:
Hello 👋,
I'd like to take this up, if no one else is already working on it.

LefterisJP:
Fixed by https://github.com/rotki/rotki/pull/7879

PR #6265, opened 2023-06-15
Improve online backups

yabirgb:
Closes #(issue_number)

## Checklist

- [ ] The PR modified the frontend, and updated the [user guide](https://github.com/rotki/rotki/blob/develop/docs/usage_guide.rst) to reflect the changes.

PR #7879, opened 2024-05-03
feat(backend): Premium DB syncing allow choice of automatic handling sync errors

Mithil467:
This PR aims to implement the feature #7570.

1. Added new setting `ask_user_upon_size_discrepancy` with as bool based on which either continue asking the user every time for input, or set to force push always.
2. [This PR](https://github.com/rotki/rotki/pull/6265) previously changed the upload request call from `put` to `post`. Certain tests were rendered ineffective as they were still mocking the `session.put` method. Updated the tests to correctly mock the `post` method.
3. Even if a test would fail by incorrectly calling the mocked request, the response being `None` would cause the test to fail but not due to the intended `assert not post_mock.called` check, but because of an exception as code refers to `response.status` and `response.text`. Updated the return value to a mocked response.

### DIFF
--- a/rotkehlchen/api/v1/schemas.py
+++ b/rotkehlchen/api/v1/schemas.py
@@ -118,6 +118,7 @@ from rotkehlchen.types import (
     OptionalBlockchainAddress,
     OptionalChainAddress,
     SupportedBlockchain,
+    SyncMethodUponSizeDiscrepancy,
     Timestamp,
     TradeType,
     UserNote,
@@ -1332,6 +1333,10 @@ class ModifiableSettingsSchema(Schema):
     )
     auto_delete_calendar_entries = fields.Boolean(load_default=None)
     auto_create_calendar_reminders = fields.Boolean(load_default=None)
+    sync_method_upon_size_discrepancy = SerializableEnumField(
+        enum_class=SyncMethodUponSizeDiscrepancy,
+        load_default=None,
+    )
 
     @validates_schema
     def validate_settings_schema(
@@ -1392,6 +1397,7 @@ class ModifiableSettingsSchema(Schema):
             oracle_penalty_duration=data['oracle_penalty_duration'],
             auto_delete_calendar_entries=data['auto_delete_calendar_entries'],
             auto_create_calendar_reminders=data['auto_create_calendar_reminders'],
+            sync_method_upon_size_discrepancy=data['sync_method_upon_size_discrepancy'],
         )
 
 
--- a/rotkehlchen/api/v1/schemas.py
+++ b/rotkehlchen/api/v1/schemas.py
@@ -118,7 +118,6 @@ from rotkehlchen.types import (
     OptionalBlockchainAddress,
     OptionalChainAddress,
     SupportedBlockchain,
-    SyncMethodUponSizeDiscrepancy,
     Timestamp,
     TradeType,
     UserNote,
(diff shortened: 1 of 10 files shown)


rotkehlchen/api/v1/schemas.py
```diff
@@ -1332,6 +1333,10 @@ class ModifiableSettingsSchema(Schema):
     )
     auto_delete_calendar_entries = fields.Boolean(load_default=None)
     auto_create_calendar_reminders = fields.Boolean(load_default=None)
+    sync_method_upon_size_discrepancy = SerializableEnumField(
```
LefterisJP, line 1336: Using an enum is not a bad idea, but since at the moment it's just 2 choices you can just use a boolean.

rotkehlchen/tests/integration/test_premium.py
```diff
@@ -155,10 +156,10 @@ def test_upload_data_to_server_same_hash(rotkehlchen_instance):
     _, our_hash = rotkehlchen_instance.data.compress_and_encrypt_db()
     remote_hash = our_hash
 
-    patched_put = patch.object(
+    patched_post = patch.object(
         rotkehlchen_instance.premium.session,
-        'put',
-        return_value=None,
+        'post',
```
LefterisJP, line 161: nice find

LefterisJP (commented):
Thank you for the PR.

Some comments:

1. `make lint` failed
2. You would need to update the docs to mention the new setting. docs/api.rst
3. Probably need to change to boolean
4. Would need to talk to @rotki/frontend to see what they can do for the last bullet point of the issue

rotkehlchen/tests/integration/test_premium.py
```diff
@@ -169,43 +169,51 @@ def test_upload_data_to_server_same_hash(rotkehlchen_instance):
...
+def test_upload_data_to_server_smaller_db(rotkehlchen_instance, db_settings: dict[str, bool]):
     """Test that if the server has bigger DB size no upload happens"""
     with rotkehlchen_instance.data.db.user_write() as cursor:
         last_ts = rotkehlchen_instance.data.db.get_static_cache(
             cursor=cursor, name=DBCacheStatic.LAST_DATA_UPLOAD_TS,
         )
         assert last_ts is None
         # Write anything in the DB to set a non-zero last_write_ts
-        rotkehlchen_instance.data.db.set_settings(cursor, ModifiableDBSettings(main_currency=A_EUR))  # noqa: E501
+        rotkehlchen_instance.data.db.set_settings(cursor, ModifiableDBSettings(main_currency=A_EUR.resolve_to_asset_with_oracles()))  # noqa: E501
```
Mithil467, line 191: Made changes at a few places as MyPy started complaining -
```
Expected main_currency: `AssetWithOracles | None` got `Asset`
Expected saved_data: `bytes` got `str`
```

Mithil467:
Thanks for the review! I have addressed the comments. Also fixed the documented HTTP method for the `/premium/sync` API.

LefterisJP (commented):
hey @Mithil467 multiple related tests have failed. Can you make sure all tests pass? You can also pop in our discord to ask for help from the dev team: https://discord.rotki.com/

LefterisJP (approved):
Thanks @Mithil467! LGTM. Merging.

PR #7884, opened 2024-05-03
Repull internal txs when redecoding transactions

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -2748,6 +2748,21 @@ class RestAPI:
                 'DELETE FROM evm_transactions WHERE tx_hash=? AND chain_id=?',
                 (tx_hash, evm_chain.serialize_for_db()))
         try:
+            chain_manager.transactions.get_or_query_transaction_receipt(tx_hash=tx_hash)
+        except RemoteError as e:
+            return {
+                'result': False,
+                'message': f'Failed to request evm transaction decoding due to hash {tx_hash.hex()} does not correspond to a transaction at {evm_chain.name}. {e!s}',  # noqa: E501
+                'status_code': HTTPStatus.CONFLICT,
+            }
+        except DeserializationError as e:
+            return {
+                'result': False,
+                'message': f'Failed to request evm transaction decoding due to {e!s}',
+                'status_code': HTTPStatus.CONFLICT,
+            }
+
+        try:
             chain_manager.transactions_decoder.decode_transaction_hashes(
                 tx_hashes=[tx_hash],
                 send_ws_notifications=True,
--- a/rotkehlchen/tests/api/test_ethereum_transactions.py
+++ b/rotkehlchen/tests/api/test_ethereum_transactions.py
@@ -27,6 +27,7 @@ from rotkehlchen.externalapis.etherscan import Etherscan
 from rotkehlchen.fval import FVal
 from rotkehlchen.history.events.structures.evm_event import EvmProduct
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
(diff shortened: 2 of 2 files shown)


rotkehlchen/api/rest.py
```diff
@@ -2747,6 +2747,21 @@ def decode_evm_transaction(
             write_cursor.execute(
                 'DELETE FROM evm_transactions WHERE tx_hash=? AND chain_id=?',
                 (tx_hash, evm_chain.serialize_for_db()))
+        try:
```
LefterisJP, line 2750: this looks corecct and was missing. Even docstring says it should do it.

One important point here is that perhaps we need to have these abstracted in proper functions and reused where needed. Probably this is already used inside the code elsewhere either completely or in parts, while this code is only on this api call.

rotkehlchen/tests/api/test_ethereum_transactions.py
```diff
@@ -1474,3 +1475,61 @@ def test_count_transactions_missing_decoding(rotkehlchen_api_server: 'APIServer'
...
+    dbevents = DBHistoryEvents(database)
+    ethereum_inquirer = rotki.chains_aggregator.ethereum
+    get_decoded_events_of_transaction(
+        evm_inquirer=ethereum_inquirer.node_inquirer,
+        database=database,
+        tx_hash=tx_hash,
+    )
+
+    # query the prices so when we later compare the events the price doesn't affect.
+    # Prices are mocked
```
LefterisJP, line 1500:
```suggestion
    # query the prices so when we later compare the events the price doesn't affect.
    # Prices are mocked
```

The first line here makes no sense. What do you mean?

rotkehlchen/tests/api/test_ethereum_transactions.py
```diff
@@ -1474,3 +1475,61 @@ def test_count_transactions_missing_decoding(rotkehlchen_api_server: 'APIServer'
...
+    filter_query = EvmEventFilterQuery.make(tx_hashes=[tx_hash])
+    entries = dbevents.get_base_entries_missing_prices(filter_query)
+    query_missing_prices_of_base_entries(
+        database=database,
+        entries_missing_prices=entries,
+    )
+
+    with database.conn.read_ctx() as cursor:
+        events_first_decode = dbevents.get_history_events(
```
LefterisJP, line 1509:
```suggestion
        events_before_redecoding = dbevents.get_history_events(
```

rotkehlchen/tests/api/test_ethereum_transactions.py
```diff
@@ -1474,3 +1475,61 @@ def test_count_transactions_missing_decoding(rotkehlchen_api_server: 'APIServer'
...
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('have_decoders', [True])
+@pytest.mark.parametrize('ethereum_accounts', [['0xc37b40ABdB939635068d3c5f13E7faF686F03B65']])
+@pytest.mark.parametrize('should_mock_price_queries', [True])
+@pytest.mark.parametrize('default_mock_price_value', [ONE])
+def test_repulling_transaction_with_internal_txs(rotkehlchen_api_server: 'APIServer'):
+    """Check that re-decoding a transaction that has internal ETH transfers correctly
+    repulls those"""
```
LefterisJP, line 1487:
```suggestion
    repulls them"""
```

rotkehlchen/api/rest.py
yabirgb: I looked around on friday and seems that we only call those two functions in a test. When querying them normally we do step by step fetching from etherscan. I'll add a note and we can refactor after the release

LefterisJP (approved):
LGTM

PR #7887, opened 2024-05-05
Track in events the blockchain log index when created from a specific log withing a transaction 

gianluca-pub-dev:
Design idea: track the blockhain log index when possible, so to map the event 1 to 1 with the blockchain, and so to improve review and test cases

Closes #(issue_number)

## Checklist


gianluca-pub-dev:
@LefterisJP , this low priority PR is for discussing a possible idea for having events more deterministic

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -443,6 +443,7 @@ def _decode_transaction(
                     result = DEFAULT_DECODING_OUTPUT
 
                 if result.event:
+                    result.event.tx_log_index = tx_log.log_index
```
LefterisJP, line 446: > @LefterisJP , this low priority draft PR is for discussing a possible idea for having events more deterministic

Hey thanks for the PR but we do not need this. Determinism in the sequence index is not needed as we don't want to map our events to the log event index order. A lot of contracts show events in order that makes no sense. So our decoders tend to change them to match an order that makes sense.

What's more rotki's events are not only 1-1 to ethereum events. All events, bitcoin transactions, CEX transactions etc, are mapped to rotki events. They are supposed to be the lowest common denominator.

Sequence index is the attribute that started being the log index of the evm event when an evm event was the source of the event. But we wanted to change the sequence index some times to make sense as mentioned above.

gianluca-pub-dev: Thanks for your review @LefterisJP .

Please notice that I am not proposing to drop the sequence_index attribute or changing its logic. I am just proposing to store additionally the original log event index as attribute, so the detection would be less error-prone (for complex transactions), and also that information could be used for reviewing the event on the blockchain. I have seen that rotki is creating also new events as with make_next_event, and this the reason I left "None" as possible value for tx_log_index. In that case the user/review/developer can know that is a "virtual" event.

PR #7891, opened 2024-05-06
Fix (location_asset_mappings): Already existing

OjusWiZard:
## Checklist

- [x] Fix bug for remote updates of `location_asset_mappings` that already exists, where it was failing with first mapping and not continuing with others.

### DIFF
--- a/rotkehlchen/db/updates.py
+++ b/rotkehlchen/db/updates.py
@@ -355,10 +355,7 @@ class RotkiDataUpdater:
                 except DeserializationError as e:
                     log.error(f'Could not deserialize {entry_type.__name__} {raw_entry!s}: {e!s}')
 
-            try:
-                update_function(entries=entries)  # type: ignore[operator]  # update_function is known
-            except InputError as e:
-                log.error(f'Could not update {entry_type.__name__} due to {e!s}')
+            update_function(entries=entries, skip_errors=True)  # type: ignore[operator]  # update_function is known
 
     def update_location_unsupported_assets(self, data: dict[str, dict[str, list[str]]], version: int) -> None:  # noqa: E501
         """Applies location unsupported assets updates in the global DB"""
--- a/rotkehlchen/globaldb/handler.py
+++ b/rotkehlchen/globaldb/handler.py
@@ -2045,11 +2045,14 @@ class GlobalDBHandler:
             } for identifier, location, symbol in cursor], mappings_count, mappings_total
 
     @staticmethod
-    def add_location_asset_mappings(entries: list[LocationAssetMappingUpdateEntry]) -> None:
+    def add_location_asset_mappings(
+            entries: list[LocationAssetMappingUpdateEntry],
+            skip_errors: bool = False,
+    ) -> None:
         """Adds the given mapping entries of asset identifiers and their symbols in the given
         location to the location_asset_mappings table.
 
-        May Raise:
+        May Raise (if skip_errors is False):
         - InputError if any of the pairs of location and exchange_symbol already exist"""
         with GlobalDBHandler().conn.write_ctx() as cursor:
             for entry in entries:
@@ -2062,17 +2065,24 @@ class GlobalDBHandler:
                         ),
                     )
                 except sqlite3.IntegrityError as e:
-                    raise InputError(
-                        f'Failed to add the location asset mapping of {entry.location_symbol} in '
-                        f'{entry.location} because it already exists in the DB.',
-                    ) from e
+                    error_msg = (
+                        f'Failed to add the location asset mapping of {entry.location_symbol} '
+                        f'in {entry.location} because it already exists in the DB.'
+                    )
+                    if skip_errors:
+                        log.error(error_msg)
+                    else:
+                        raise InputError(error_msg) from e
 
     @staticmethod
-    def update_location_asset_mappings(entries: list[LocationAssetMappingUpdateEntry]) -> None:
+    def update_location_asset_mappings(
+            entries: list[LocationAssetMappingUpdateEntry],
+            skip_errors: bool = False,
+    ) -> None:
         """Updates the mapped asset identifiers in the location_asset_mappings table based on their
         location symbol.
 
-        May Raise:
+        May Raise (if skip_errors is False):
         - InputError if any of the pairs of location and exchange_symbol does not exist"""
         with GlobalDBHandler().conn.write_ctx() as cursor:
             for entry in entries:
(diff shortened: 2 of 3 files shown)


rotkehlchen/db/updates.py
```diff
@@ -337,28 +335,40 @@ def update_global_addressbook(self, data: list[dict[str, Any]], version: int) ->
...
-                    log.error(f'Could not deserialize {entry_type.__name__} {raw_entry!s}: {e!s}')
+                    log.error(f'Could not deserialize {raw_entry!s}: {e!s}')
 
-            try:
-                update_function(entries=entries)  # type: ignore[operator]  # update_function is known
-            except InputError as e:
-                log.error(f'Could not update {entry_type.__name__} due to {e!s}')
+                try:
+                    with GlobalDBHandler().conn.write_ctx() as write_cursor:
+                        write_cursor.execute(update_query, [raw_entry[key] for key in entry_keys])
```
yabirgb, line 369: this can raise keyError

rotkehlchen/db/updates.py
```diff
@@ -337,28 +335,40 @@ def update_global_addressbook(self, data: list[dict[str, Any]], version: int) ->
...
         ):
-            entries, raw_data = [], data.get(raw_data_key)
+            raw_data = data.get(raw_data_key)
             if raw_data is None:
                 continue  # no data to update
             for raw_entry in raw_data:
                 try:
                     if (asset_id := raw_entry.get('asset')) is not None:
-                        raw_entry['asset'] = Asset(asset_id)
+                        raw_entry['asset'] = Asset(asset_id).serialize()
```
yabirgb, line 361: can't this raise UnknownAsset?

rotkehlchen/db/updates.py
```diff
@@ -337,28 +335,40 @@ def update_global_addressbook(self, data: list[dict[str, Any]], version: int) ->
...
         log.info(f'Applying update for location asset mappings to v{version}')
-        for update_function, entry_type, raw_data_key in (
-            (GlobalDBHandler.add_location_asset_mappings, LocationAssetMappingUpdateEntry, 'additions'),  # noqa: E501
-            (GlobalDBHandler.update_location_asset_mappings, LocationAssetMappingUpdateEntry, 'updates'),  # noqa: E501
-            (GlobalDBHandler.delete_location_asset_mappings, LocationAssetMappingDeleteEntry, 'deletions'),  # noqa: E501
+        for update_query, entry_keys, raw_data_key in (
+            (
+                'INSERT OR IGNORE INTO location_asset_mappings(local_id, location, exchange_symbol) VALUES(?, ?, ?);',  # noqa: E501
+                ('asset', 'location', 'location_symbol'),
+                'additions',
```
yabirgb, line 342: so the change is ignoring/replacing right? Does it make sense to change the logic in the `GlobalDBHandler`? Otherwise we will change the logic and forget about changing this. We could have a ignore_conflicts argument but don't know if it is tricking too much the logic. TBH I'm not sure what is better here

rotkehlchen/db/updates.py
OjusWiZard: It doesn't raise UnknownAsset error

rotkehlchen/globaldb/handler.py
```diff
@@ -2047,67 +2047,70 @@ def query_location_asset_mappings(
...
+                        ).fetchone()[0] > 0:  # it already exists
+                            continue  # we don't do OR IGNORE with NULL location, otherwise it creates new row in DB considering NULL as distinct  # noqa: E501
+                    else:
+                        location = entry.location.serialize_for_db()
+
+                    cursor.execute(query, (entry.asset.serialize(), location, entry.location_symbol))  # noqa: E501
                 except sqlite3.IntegrityError as e:
-                    raise InputError(
+                    log.error(
                         f'Failed to add the location asset mapping of {entry.location_symbol} in '
```
yabirgb, line 2068: If you do INSERT OR IGNORE there can't be integrity error

rotkehlchen/globaldb/handler.py
```diff
@@ -2047,67 +2047,70 @@ def query_location_asset_mappings(
...
-                        ),
-                    )
+                    query = 'INSERT OR IGNORE INTO location_asset_mappings(local_id, location, exchange_symbol) VALUES(?, ?, ?)'  # noqa: E501
+                    if entry.location is None:
+                        location = None
+                        if cursor.execute(
+                            'SELECT COUNT(*) FROM location_asset_mappings WHERE location IS NULL AND exchange_symbol=?',  # noqa: E501
+                            (entry.location_symbol,),
+                        ).fetchone()[0] > 0:  # it already exists
+                            continue  # we don't do OR IGNORE with NULL location, otherwise it creates new row in DB considering NULL as distinct  # noqa: E501
```
yabirgb, line 2061: log here what you are skipping so we later know

rotkehlchen/globaldb/handler.py
```diff
@@ -2047,67 +2047,70 @@ def query_location_asset_mappings(
...
-                    'UPDATE location_asset_mappings SET local_id=? WHERE location IS ? AND exchange_symbol=?', (  # noqa: E501
-                        entry.asset.serialize(),
-                        None if entry.location is None else entry.location.serialize_for_db(),
-                        entry.location_symbol,
-                    ),
-                )
-                if cursor.rowcount != 1:
-                    raise InputError(
+                try:
+                    query = 'INSERT OR REPLACE INTO location_asset_mappings(local_id, location, exchange_symbol) VALUES(?, ?, ?)'  # noqa: E501
```
yabirgb, line 2079: Why not do a update instead of a insert or replace?

yabirgb: Also sqlite has `https://sqlite.org/lang_upsert.html`

rotkehlchen/globaldb/handler.py
```diff
@@ -2047,67 +2047,70 @@ def query_location_asset_mappings(
...
-                    raise InputError(
+                try:
+                    cursor.execute(
+                        'DELETE FROM location_asset_mappings WHERE location IS ? AND exchange_symbol=?', (  # noqa: E501
+                            None if entry.location is None else entry.location.serialize_for_db(),
+                            entry.location_symbol,
+                        ),
+                    )
+                except sqlite3.IntegrityError as e:
+                    log.error(
```
yabirgb, line 2111: when can the delete raise an integrityError?

LefterisJP (approved):
LGTM

PR #7901, opened 2024-05-07
Add ckzg to the build dependencies

yabirgb:
This is a dependency that doesn't have wheels and we need to build and link manually. We need to pin the version so it can be later retrieved

Closes #(issue_number)

## Checklist

### DIFF
--- a/package.py
+++ b/package.py
@@ -493,31 +493,40 @@ class MacPackaging:
     def __get_versions(self, packages: list[str]) -> dict[str, str]:
         """
         Gets the versions of specified packages from requirements.txt
+        and requirements_packaging.txt.
 
         :param packages: A list of package names for which we need versions from
         the requirements.txt
         :returns: A Dict where the key is the package and the value is the package version
         """
-        requirements = self.__storage.working_directory / 'requirements.txt'
         package_versions: dict[str, str] = {}
-        with open(requirements, encoding='utf8') as fp:
-            while True:
-                line = fp.readline()
-                if not line:
-                    break
-                if len(line.strip()) == 0 or line.startswith('#'):
-                    continue
-                requirement = line.split('#')[0]
-                req = requirement.split(';')
-                requirement = req[0]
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -46,3 +46,6 @@ bip-utils==2.9.3
 
 #constraints
 pycparser<=2.17 # ugly -- to avoid https://github.com/eliben/pycparser/pull/198
+
+# used for the packaging script that requires to have a version pinned for it
+ckzg==1.0.1
```
LefterisJP, line 51: Is this used in production or packaging? If only packaging then this should go to requirements_packaging.txt

package.py
```diff
@@ -498,26 +498,28 @@ def __get_versions(self, packages: list[str]) -> dict[str, str]:
...
-                if not line:
-                    break
-                if len(line.strip()) == 0 or line.startswith('#'):
-                    continue
-                requirement = line.split('#')[0]
-                req = requirement.split(';')
-                requirement = req[0]
-                if len(req) > 1 and req[1].strip() == "sys_platform == 'win32'":
-                    continue
+        for requirement_file_name in ('requirements.txt', 'requirements_packaging.txt'):
```
LefterisJP, line 502: please also adjust the docstring to mention it uses both files

package.py
```diff
@@ -498,26 +498,28 @@ def __get_versions(self, packages: list[str]) -> dict[str, str]:
...
+                    req = requirement.split(';')
+                    requirement = req[0]
+                    if len(req) > 1 and req[1].strip() == "sys_platform == 'win32'":
+                        continue
+
+                    split_requirement = requirement.split('==')
+                    package_name = split_requirement[0]
+                    if package_name in packages:
+                        package_version = split_requirement[1]
+                        package_versions[package_name.strip()] = package_version.strip()
```
LefterisJP, line 521: Also here please raise an error if `package_versions` already contains a version for this. This would mean that a file redefines a package.

LefterisJP (approved):
LGTM

PR #7923, opened 2024-05-11
Feat (History Events): Consistent free and premium

### LINKED ISSUE #6115: Rethink filtering history events for non-premium
## Abstract

At the moment we have disabled most filtering for non-premium users in the history events page. The reason for that is not business driven, but technical.

It was realized during this PR: https://github.com/rotki/rotki/pull/6105

Essentially when performing paginated filter queries with the new method we use where we define an event as the group of common event_identifiers we can't apply a limit **and** a filter at the same time.

The limit can either be applied after the filtering, in which case it makes no sense anymore since it's no longer an event limit but a filter limit and then by simple pagination or date range users can get all events.

Or it can be applied before filtering as is done here: https://github.com/rotki/rotki/blob/6b1a538ca6a1ec235933b0d3936698be9ebc72ee/rotkehlchen/db/history_events.py#L403

The problem with this approach is that the filter is then applied to the GROUP BY limited query. And that query only includes the first event of each group. So if say you have an event group of a transaction of sending 10 DAI to someone. The first event is the gas burning. If you filter by DAI asset it won't work as the filter would only be applied in the first event, the gas burning.

This problem occurs only on the not premium part of the above code as that's the only part where a number limitation on the overall events (meaning grouped by events) needs to be applied.

Before https://github.com/rotki/rotki/pull/6105, when the limit was not on the grouped events, but on the individual events that problem did not exist. But then we had a more serious problem of pagination not working as, say a limit of 100 events would not be 10x10 pages, but depending on events per group would be maybe ~70 and pagination was off.

This is what that PR fixed. 

## Task

Now the problem here is that by fixing pagination and limiting  for free users the filtering no longer works. Which was an acceptable choice for 1.28.

The problem is we want to go toward tiered premium, where there will be different limits dependin on the tier. We can't have filtering off there on any but the lowest tier. We generally need a way to have both a limit **and** filtering and have it all applied to the grouped by events.

So think about this and come up with a proper solution.

Change this test to use both flags when fixed: https://github.com/rotki/rotki/pull/6105/files#diff-1bf17d119fd2bf5ac8b87baa10349638966aecfc5309660d1675564616845c9aR1312

OjusWiZard:
Closes #6115

## Checklist

- [x] Modify and test `get_history_events()` to return consistent results with premium and free modes.

### DIFF
--- a/rotkehlchen/db/history_events.py
+++ b/rotkehlchen/db/history_events.py
@@ -66,15 +66,20 @@ log = RotkehlchenLogsAdapter(logger)
 
 def maybe_filter_ignore_asset(
         filter_query: HistoryBaseEntryFilterQuery,
+        include_ignored_assets: bool = False,
 ) -> str:
     """An auxiliar function to find if query_filter contains `DBIgnoredAssetsFilter`. If it does
     then return that filter clause. This is done where we want to filter ignored assets
-    before applying the free limit."""
+    before applying the free limit. If `include_ignored_assets` is true then the filter is returned
+    to include them."""
     for fil in filter_query.filters:
         if isinstance(fil, DBIgnoredAssetsFilter):
             # Also don't count spam asset transactions in the limit
-            return 'WHERE (asset IS NULL OR asset NOT IN (SELECT value FROM multisettings WHERE name="ignored_asset")) '  # noqa: E501
-    return ' '
+            ignored_asset_subquery = 'SELECT value FROM multisettings WHERE name="ignored_asset")'
+            if include_ignored_assets:
+                return f'WHERE (asset IN ({ignored_asset_subquery}) '
+            return f'WHERE (asset IS NULL OR asset NOT IN ({ignored_asset_subquery}) '
+    return ''
 
 
 class DBHistoryEvents:
@@ -310,6 +315,46 @@ class DBHistoryEvents:
 
         return deserialized
 
+    def _create_history_events_query(
+            self,
+            filter_query: HistoryBaseEntryFilterQuery,
+            entries_limit: int,
+            has_premium: bool,
+            group_by_event_ids: bool = False,
+    ) -> tuple[str, list]:
+        """Returns the sql queries and bindings for the history events without pagination."""
+        base_suffix = f'{HISTORY_BASE_ENTRY_FIELDS}, {EVM_EVENT_FIELDS}, {ETH_STAKING_EVENT_FIELDS} {ALL_EVENTS_DATA_JOIN}'  # noqa: E501
+        if (ignore_asset_filter := maybe_filter_ignore_asset(filter_query, include_ignored_assets=True)) != '':  # noqa: E501
+            ignore_asset_filter = (
+                f' WHERE event_identifier NOT IN '
+                f'(SELECT DISTINCT event_identifier FROM history_events {ignore_asset_filter})'
+            )
+
+        premium_base_suffix = f'{base_suffix} {ignore_asset_filter}'
+        free_base_suffix = (
+            f'* FROM (SELECT {base_suffix}) WHERE event_identifier IN ('
+            f'SELECT DISTINCT event_identifier FROM history_events {ignore_asset_filter} '
+            'ORDER BY timestamp DESC,sequence_index ASC LIMIT ?)'  # free query only select the last LIMIT groups  # noqa: E501
+        )
+
+        if has_premium:
+            suffix, limit = premium_base_suffix, []
+        else:
+            suffix, limit = free_base_suffix, [entries_limit]
+
+        if group_by_event_ids:
+            filters, query_bindings = filter_query.prepare(
+                with_group_by=True,
+                with_pagination=False,
+                without_ignored_asset_filter=True,
+            )
+            prefix = 'SELECT COUNT(*), *'
+        else:
+            filters, query_bindings = filter_query.prepare(with_pagination=False)
+            prefix = 'SELECT *'
+
+        return f'{prefix} FROM (SELECT {suffix}) {filters}', limit + query_bindings
+
     @overload
     def get_history_events(
             self,
@@ -397,40 +442,19 @@ class DBHistoryEvents:
         TODO: To not query all columns with all joins for all cases, we perhaps can
         peek on the entry type of the filter and adjust the SELECT fields accordingly?
         """
-        base_prefix = 'SELECT '
-        base_suffix = f'{HISTORY_BASE_ENTRY_FIELDS}, {EVM_EVENT_FIELDS}, {ETH_STAKING_EVENT_FIELDS} {ALL_EVENTS_DATA_JOIN}'  # noqa: E501
-        bindings = []
-        type_idx = 1 if group_by_event_ids is True else 0
-        special_free_query = False
-        free_query_group_by = maybe_filter_ignore_asset(filter_query)
-
-        if has_premium is True:
-            if group_by_event_ids is True:
-                base_prefix += 'COUNT(*), '
-            base_query = f'{base_prefix} {base_suffix}'
-        else:
-            if group_by_event_ids is True:
-                special_free_query = True  # a bit ugly conditions to keep limit at groups for free users  # noqa: E501
-                free_query_group_by += 'GROUP BY event_identifier'
-                base_query = (
-                    f'{base_prefix} * FROM (SELECT COUNT(*), {base_suffix} '
-                    f'{free_query_group_by} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?) '
-                )
-            else:  # if we don't group them but still apply free limit, then apply this limit only on any event_identifiers that fall inside the limit  # noqa: E501
-                base_query = (
-                    f'{base_prefix} * FROM (SELECT {base_suffix} WHERE event_identifier IN '
-                    f'(SELECT DISTINCT event_identifier FROM history_events {free_query_group_by} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?)) '  # noqa: E501
-                )
-            bindings = [FREE_HISTORY_EVENTS_LIMIT]
-
-        prepared_query, prepared_bindings = filter_query.prepare(
-            with_group_by=group_by_event_ids,
-            special_free_query=special_free_query,
+        base_query, filters_bindings = self._create_history_events_query(
+            has_premium=has_premium,
+            filter_query=filter_query,
+            group_by_event_ids=group_by_event_ids,
+            entries_limit=FREE_HISTORY_EVENTS_LIMIT,
         )
-        bindings.extend(prepared_bindings)
 
-        cursor.execute(base_query + prepared_query, bindings)
+        if filter_query.pagination is not None:
+            base_query = f'SELECT * FROM ({base_query}) {filter_query.pagination.prepare()}'
+
+        cursor.execute(base_query, filters_bindings)
         output: list[HistoryBaseEntry] | list[tuple[int, HistoryBaseEntry]] = []
+        type_idx = 1 if group_by_event_ids else 0
         data_start_idx = type_idx + 1
         for entry in cursor:
             entry_type = HistoryBaseEntryType(entry[type_idx])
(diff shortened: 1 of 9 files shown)


rotkehlchen/db/history_events.py
```diff
@@ -401,35 +402,40 @@ def get_history_events(
...
-                free_query_group_by += 'GROUP BY event_identifier'
                 base_query = (
-                    f'{base_prefix} * FROM (SELECT COUNT(*), {base_suffix} '
-                    f'{free_query_group_by} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?) '
+                    f'{base_prefix} COUNT(*), * FROM (SELECT {base_suffix} WHERE event_identifier '
+                    f'IN (SELECT DISTINCT event_identifier FROM history_events {filters} LIMIT ?))'
                 )
+                bindings.extend(filters_bindings + [self.free_history_events_limit])
+                if filter_query.group_by is not None:
+                    final_clause += f' {filter_query.group_by.prepare()}'
```
yabirgb, line 428: in both branches (premium and notpremium) this seems to be the same logic. we already have in line 404 for `if group_by_event_ids` so you can do there

```python
if group_by_event_ids:
    type_idx = 
    if filter_query.group_by is not None:
        final_clause += f' {filter_query.group_by.prepare()}'
else:
    type_idx = 0

File: rotkehlchen/db/history_events.py [hunk 2]
```diff
@@ -401,35 +402,40 @@ def get_history_events(
         base_suffix = f'{HISTORY_BASE_ENTRY_FIELDS}, {EVM_EVENT_FIELDS}, {ETH_STAKING_EVENT_FIELDS} {ALL_EVENTS_DATA_JOIN}'  # noqa: E501
         bindings = []
         type_idx = 1 if group_by_event_ids is True else 0
-        special_free_query = False
-        free_query_group_by = maybe_filter_ignore_asset(filter_query)
+        final_clause = ''
```
yabirgb, line 405: this is only for grouping and ordering right? rather than final_clause can you come up with a name that is more descreptive? Or at least having a comment explaining what you use this for

rotkehlchen/db/history_events.py
```diff
@@ -401,35 +402,40 @@ def get_history_events(
...
-                )
-            bindings = [FREE_HISTORY_EVENTS_LIMIT]
+                bindings.extend(filters_bindings + [self.free_history_events_limit])
+                base_query = f'{base_prefix} * FROM (SELECT {base_suffix} {filters} LIMIT ?)'
 
-        prepared_query, prepared_bindings = filter_query.prepare(
-            with_group_by=group_by_event_ids,
-            special_free_query=special_free_query,
-        )
-        bindings.extend(prepared_bindings)
```
yabirgb, line 430: We have this `special_free_query` for ignoring the spam assets in the free serach. This one is missing in your current logic. Can you also adjust the test to have on of those? or maybe create a new test for it

rotkehlchen/db/history_events.py
```diff
@@ -401,35 +402,40 @@ def get_history_events(
         base_suffix = f'{HISTORY_BASE_ENTRY_FIELDS}, {EVM_EVENT_FIELDS}, {ETH_STAKING_EVENT_FIELDS} {ALL_EVENTS_DATA_JOIN}'  # noqa: E501
         bindings = []
         type_idx = 1 if group_by_event_ids is True else 0
-        special_free_query = False
-        free_query_group_by = maybe_filter_ignore_asset(filter_query)
+        final_clause = ''
 
         if has_premium is True:
+            filters, filters_bindings = filter_query.prepare(with_order=False, with_pagination=False)  # noqa: E501
+            bindings.extend(filters_bindings)
```
yabirgb, line 409:
```suggestion
        if has_premium:
            filters, filters_bindings = filter_query.prepare(with_order=False, with_pagination=False)  # noqa: E501
            bindings.extend(filters_bindings)
```

rotkehlchen/tests/db/test_history_events.py
```diff
@@ -311,3 +312,172 @@ def test_delete_last_event(database):
...
+            assert history_events.free_history_events_limit == 2
+            for filters in (  # trying different filters with some combinations
+                HistoryEventFilterQuery.make(event_types=[HistoryEventType.RECEIVE]),
+                HistoryEventFilterQuery.make(assets=(A_ETH, A_USDT)),
+                HistoryEventFilterQuery.make(location=Location.COINBASE),
+                HistoryEventFilterQuery.make(from_ts=Timestamp(2)),
+                HistoryEventFilterQuery.make(assets=(A_ETH, A_USDC), from_ts=Timestamp(2)),
+                HistoryEventFilterQuery.make(exclude_ignored_assets=True),
+            ):
+                premium_grouped_result = history_events.get_history_events(  # when grouping
```
yabirgb, line 454: what are you checking here? changing the free history event limit?

rotkehlchen/db/history_events.py
```diff
@@ -81,6 +81,7 @@ class DBHistoryEvents:
 
     def __init__(self, database: 'DBHandler') -> None:
         self.db = database
+        self.free_history_events_limit = FREE_HISTORY_EVENTS_LIMIT
```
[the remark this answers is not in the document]
OjusWiZard: This allowed me to mock it in the test, and also we may need it as a variable when allowing different tiers of premium. But, lmk if we should still keep it as constant, it's tricky because the test is interpreted after the constant is imported so the mocked value is not used in the test case.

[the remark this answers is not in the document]
yabirgb: Yes, you shouldn't have code just for tests. It is possible to mock a global variable too, we do it in other places and for the tier premium is an adjustment that we can do later. Is not a significant change

rotkehlchen/tests/db/test_history_events.py
OjusWiZard: This checks that when results of a query hits the free limit, then we should not return more the FREE_LIMIT results. `len(free_grouped_result) == min(2, len(premium_grouped_result))`
And the N limited results that are returned, are the first N results returned in premium query with the same filters. `free_grouped_result == premium_grouped_result[:2]`

rotkehlchen/db/history_events.py
```diff
@@ -397,39 +397,65 @@ def get_history_events(
...
-                free_query_group_by += 'GROUP BY event_identifier'
-                base_query = (
-                    f'{base_prefix} * FROM (SELECT COUNT(*), {base_suffix} '
-                    f'{free_query_group_by} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?) '
+        if has_premium:
+            if group_by_event_ids:
+                filters, filters_bindings = filter_query.prepare(
+                    with_order=False,
+                    with_pagination=False,
+                    special_free_query=True,
```
yabirgb, line 416: In the premium I believe we don´t use it since there are no limits. For free users we wanted to prevent them from getting no events due to spam

OjusWiZard: yeah that was the reason we had this flag, but there is another reason to seperate ignored asset filter from other filters.

When excluding the ignored assets and enabling grouping, if we apply the ignored assets exclusion filter in the outer query along with other filters then the group having one event with ignored asset is also returned because other events of that group have normal assets.
So, now I'm eliminating them in the sub query for both cases, so that in the outer query none of the events are present which are part of a group in which at least one event have ignored event.

This is how the query is forming with current logic:
```sql
SELECT COUNT(*), * FROM (
    SELECT entry_type, history_events.identifier AS history_events_identifier, event_identifier, sequence_index, timestamp, location, location_label, asset, amount, usd_value, notes, type, subtype , tx_hash, counterparty, product, address, extra_data, validator_index, is_exit_or_blocknumber FROM history_events
    LEFT JOIN evm_events_info ON history_events.identifier=evm_events_info.identifier
    LEFT JOIN eth_staking_events_info ON history_events.identifier=eth_staking_events_info.identifier
    WHERE event_identifier NOT IN (
        SELECT event_identifier FROM history_events
        WHERE (asset IN (SELECT value FROM multisettings WHERE name="ignored_asset"))
    )
)
GROUP BY event_identifier
ORDER BY timestamp ASC,sequence_index ASC
```

This is formed when doing it along with the filters:
```sql
SELECT COUNT(*), * FROM (
    SELECT entry_type, history_events.identifier AS history_events_identifier, event_identifier, sequence_index, timestamp, location, location_label, asset, amount, usd_value, notes, type, subtype , tx_hash, counterparty, product, address, extra_data, validator_index, is_exit_or_blocknumber FROM history_events
    LEFT JOIN evm_events_info ON history_events.identifier=evm_events_info.identifier
    LEFT JOIN eth_staking_events_info ON history_events.identifier=eth_staking_events_info.identifier
)
WHERE (asset IS NULL OR asset NOT IN (SELECT value FROM multisettings WHERE name="ignored_asset"))
GROUP BY event_identifier
ORDER BY timestamp ASC,sequence_index ASC
```

rotkehlchen/db/history_events.py
```diff
@@ -66,15 +66,19 @@
 
 def maybe_filter_ignore_asset(
         filter_query: HistoryBaseEntryFilterQuery,
+        negation: bool = False,
```
yabirgb, line 69: negation is a really bad argument name here. The reason is that is very generic and is not documented in the docstring. If you are the person that reads this code negation doesn't mean anything to you and is confusing making you go to the function and read how it works.

What do you thing of using `include_ignored_assets`? Also document it in the docstring

rotkehlchen/db/history_events.py
```diff
@@ -397,39 +401,55 @@ def get_history_events(
...
-                free_query_group_by += 'GROUP BY event_identifier'
-                base_query = (
-                    f'{base_prefix} * FROM (SELECT COUNT(*), {base_suffix} '
-                    f'{free_query_group_by} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?) '
+            filters, filters_bindings = filter_query.prepare(with_pagination=False)
+
+        if has_premium:
+            if group_by_event_ids:
+                base_query = group_by_event_id_query.format(
+                    exclude_ignore_assets=ignore_asset_filter,
```
yabirgb, line 429: `exclude_ignore_assets` makes me thing it is a boolean but instead is a string filter. Can you rename it to explicit that is a filter?

rotkehlchen/db/history_events.py
```diff
@@ -397,39 +401,55 @@ def get_history_events(
...
-            bindings = [FREE_HISTORY_EVENTS_LIMIT]
+            else:  # if we don't group them but still apply free limit, then apply this limit only on any event_identifiers that fall inside the limit  # noqa: E501
+                base_query = f'{base_prefix} * FROM (SELECT {base_suffix} {filters} LIMIT ?)'
 
-        prepared_query, prepared_bindings = filter_query.prepare(
-            with_group_by=group_by_event_ids,
-            special_free_query=special_free_query,
-        )
-        bindings.extend(prepared_bindings)
+            bindings.extend(filters_bindings + [FREE_HISTORY_EVENTS_LIMIT])
```
yabirgb, line 447: you can do this extend at line 425 and here just do it 

```suggestion
            bindings.extend([FREE_HISTORY_EVENTS_LIMIT])
```

rotkehlchen/db/history_events.py
```diff
@@ -397,39 +401,55 @@ def get_history_events(
         TODO: To not query all columns with all joins for all cases, we perhaps can
         peek on the entry type of the filter and adjust the SELECT fields accordingly?
         """
+        bindings, type_idx = [], 0
         base_prefix = 'SELECT '
```
yabirgb, line 405: base prefix is always select right? Why not use it always? THat avoids replacing that is more expensive

rotkehlchen/db/history_events.py
```diff
@@ -397,40 +441,18 @@ def get_history_events(
...
-                base_query = (
-                    f'{base_prefix} * FROM (SELECT {base_suffix} WHERE event_identifier IN '
-                    f'(SELECT DISTINCT event_identifier FROM history_events {free_query_group_by} ORDER BY timestamp DESC, sequence_index ASC LIMIT ?)) '  # noqa: E501
-                )
-            bindings = [FREE_HISTORY_EVENTS_LIMIT]
+        base_query, filters_bindings = self._create_history_events_query(
+            filter_query=filter_query,
+            group_by_event_ids=group_by_event_ids,
+            entries_limit=FREE_HISTORY_EVENTS_LIMIT,
+        )[0 if has_premium else 1]
```
yabirgb, line 448: I see how this improved the function but the list indexing here is a bad pattern. `has_premium` should be an argument and in `_create_history_events_query` you should create only the strings that you need. You can have the if has_premium in there and only create and return what you need

yabirgb: This is error prone and harder to follow

rotkehlchen/db/history_events.py
```diff
@@ -642,36 +664,28 @@ def get_history_events_count(
...
+            entries_limit=FREE_HISTORY_EVENTS_LIMIT if entries_limit is None else entries_limit,
+        )
+
+        count_without_limit = cursor.execute(
+            f'SELECT COUNT(*) FROM ({premium_query})',
+            premium_bindings,
+        ).fetchone()[0]
+
+        if entries_limit is None:
+            return count_without_limit, count_without_limit
```
yabirgb, line 682: after this check is when you can call `_create_history_events_query` with `has_premium=False`

LefterisJP (approved):
Code changes look good to me. If after this filtering works both for premium and for free and has been tested it's fine. I am not testing it myself though so am taking @OjusWiZard and @yabirgb word for it.

lukicenturi:
Frontend code looks fine, also tried the functionality in frontend and it works

lukicenturi (approved):

LefterisJP:
Failed test is a remote error.

PR #7926, opened 2024-05-13
Fix (Asset updates): Allow absent files

OjusWiZard:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=62277464

## Checklist

- [x] Allow asset updates if one or more update files are missing, and test.


rotkehlchen/globaldb/updates.py
```diff
@@ -725,15 +725,25 @@ def _retrieve_update_files(
             asset_collections_url = ASSET_COLLECTIONS_UPDATES_URL.format(branch=self.branch, version=version)  # noqa: E501
             asset_collections_mappings_url = ASSET_COLLECTIONS_MAPPINGS_UPDATES_URL.format(branch=self.branch, version=version)  # noqa: E501
 
-            assets_file = query_file(url=assets_url, is_json=False)
+            assets_file, asset_collections_file, asset_collections_mappings_file = '', '', ''
+            try:
+                assets_file = query_file(url=assets_url, is_json=False)
+            except RemoteError as e:
+                log.warning(f'Could not query assets update {version}: {e!s}')
```
yabirgb, line 732: won't this skip the asset update and then skip it forever? since in the globaldb it will write as if this update was performed

LefterisJP (approved):
LGTM. Thanks!

PR #7931, opened 2024-05-14
feat(frontend): setting to force push when time discrepancy occurs

### LINKED ISSUE #7570: Premium DB syncing allow choice of automatic handling sync errors

lukicenturi:
Closes #7570

## Checklist

### DIFF
--- a/frontend/app/src/components/settings/general/AskUserUponSizeDiscrepancySetting.vue
+++ b/frontend/app/src/components/settings/general/AskUserUponSizeDiscrepancySetting.vue
@@ -1,0 +1,46 @@
+<script setup lang="ts">
+withDefaults(
+  defineProps<{
+    dialog: boolean;
+    confirm: boolean;
+  }>(),
+  { dialog: false, confirm: false },
+);
+
+const value = ref(false);
+const { askUserUponSizeDiscrepancy } = storeToRefs(useGeneralSettingsStore());
+
+watchImmediate(askUserUponSizeDiscrepancy, (askUser) => {
+  set(value, !get(askUser));
+});
+
+const { t } = useI18n();
+</script>
+
+<template>
+  <SettingsOption
+    #default="{ error, success, updateImmediate }"
+    setting="askUserUponSizeDiscrepancy"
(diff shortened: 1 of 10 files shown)


frontend/app/src/components/settings/general/AskUserUponSizeDiscrepancySetting.vue
```diff
@@ -0,0 +1,46 @@
+<script setup lang="ts">
+withDefaults(
+  defineProps<{
+    dialog: boolean;
+    confirm: boolean;
```
kelsos, line 5:
```suggestion
    dialog?: boolean;
    confirm?: boolean;
```
Otherwise the props are required and it gives us a warning

lukicenturi: updated

kelsos (approved):
lgtm

Issue #3433, opened 2021-09-09
Support ETH Staking via Lido (multi-protocol liquid staking solution)


isidorosp:
## Motivation

It would be great for users to be able to easily see the work that their liquid staking tokens are doing, whether those are simply held in wallets or have been put into DeFi protocols.

## Abstract

Lido (http://lido.fi/) is a multi-protocol liquid staking solution (i.e. when you stake you receive a token that represents your staked assets, which either accrues staking interest or is automatically rebased to include staking interest on a periodic basis, and this token can be used in DeFi in lieu of the staked asset). Currently Lido staking offerings exist on [Ethereum](http://lido.fi/), [Terra](https://lido.fi/terra), and [Solana](http://solana.lido.fi/). The staking product for each protocol works a little differently, so  I would propose that we just start with Ethereum. 

To that end, as a representative of the LEGO (Lido Ecosystem Grants Organisation) I'm happy to be able to offer a grant for the completion of an integration of Lido into rotki. Since there are pieces in this issue starting from basic (i.e. just accounting for ROI based on stETH held) to complex that can only be completed once further underlying protocols are supported by rotki, we can also break the tasks and grants into stages/milestones. I propose a total grant size of ~1K LDO (1 LDO ~= 5 USD atm), but to be aligned on based on the amount of work that's expected after some initial research. If a much larger grant is deemed necessary, it can be discussed and a more formal LEGO proposal can be made.

Given my role at Lido and my active contribution to rotki in the past, I would want to make clear that I am not be eligible to pick up / work on this issue, nor will I be receiving any compensation from rotki, directly or indirectly with regards to this issue.

## Specification

**Ethereum**
Lido's staked Ether comes in two forms: [stETH](https://www.coingecko.com/en/coins/lido-staked-ether), which rebases daily at midnight GMT, and [wstETH](https://docs.lido.fi/contracts/wsteth), a wrapped version which is non-rebasing. This is because some DeFi protocols play nice with rebasing tokens (e.g. Curve) while others don't (e.g. Uniswap).

[...]

PR #7949, opened 2024-05-18
Added support for decoding Lido contracts events on ethereum network - Part 1 - Submit topic

gianluca-pub-dev:
Added support for Lido contract on eth blockchain for the following events:
- staking submitting ETH to stETH contract, and receiving stETH in return

temporarly removed from the PR:
```
- staking transferring ETH to wstETH contract and receiving wstETH in return, tracking the calculated wstETH ratio at the time of the staking
- wrapping stETH receiving wstETH, tracking the calculated wstETH ratio at the time of the wrapping
- unwrapping wstETH receiving stETH, , tracking the calculated wstETH ratio at the time of the unwrapping
```

https://docs.lido.fi/contracts/lido/#submit-1
https://docs.lido.fi/contracts/wsteth
https://docs.lido.fi/contracts/lido/#transfershares

This maybe covers the backend side of B.2 point from issue: https://github.com/rotki/rotki/issues/3433

## Checklist


rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+                event.notes = f'Sent {wsteth_unwrapped_amount} {A_WSTETH.symbol_or_name()} for unwrapping at wstETH ratio {paired_wsteth_ratio}'  # noqa: E501
+                out_event = event
+            elif (
+                event.event_type == HistoryEventType.RECEIVE and
+                event.event_subtype == HistoryEventSubType.NONE and
+                event.asset == A_STETH and
+                event.balance.amount == sth_received_amount and
+                event.address == wsteth_evm_address
+            ):
+                event.event_type = HistoryEventType.WITHDRAWAL  # TODO: decide the right mappings
```
gianluca-pub-dev, line 296: Rotki team: I would like to we discuss with you about how to map the lido events, because it is a staking, but users receive tokens in return that have a deep market, and often they are exchanged without returning the tokens to the Lido contract, so I am not sure how to map (I have never used the STAKING event type for this reason)

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+        collateral_amount = asset_normalized_value(
+            amount=amount_raw,
+            asset=A_ETH.resolve_to_crypto_asset(),
+        )
+
+        # Searching for the exact paired stETH token reception for validating the decoding
+        steth_minted_tokens = None
+        # Here searching for the paired stETH mint operation, with an hacky way for the issue:
+        # https://github.com/lidofinance/lido-dao/issues/442
+        for tx_log in context.all_logs:
```
LefterisJP, line 58: If the to is the sender address it will be seen as an ERC20 transfer by rotki. Why do you need to do this here. This does not make sense to me.

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+                amount=steth_minted_tokens,
+                to_event_subtype=action_to_event_subtype,
+                to_notes=action_to_notes,
+                to_counterparty=CPT_LIDO_ETH,
+                paired_event_data=(paired_event, True),
+                extra_data={'staked_eth': str(collateral_amount)},
+            ))
+        else:  # did not break/find anything
+            log.error(
+                'At lido steth submit decoding of tx '
```
LefterisJP, line 118: add the tx hash always in such errors as I showed you above.

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+                'not found the decoded event of the ETH transfer',
+            )
+            return DEFAULT_DECODING_OUTPUT
+
+        return DecodingOutput(action_items=action_items, matched_counterparty=CPT_LIDO_ETH)
```
LefterisJP, line 123: So for this rule I don't understand what you are trying to do. Why bother with this issue at all? 

The actual transfer from 0x0000 to the sender maps **exactly** to the amount of the internal ETH transfer.

So you really need to do nothing else other than intercept that transfer of stETH and check if the sender is the zero address. If it is then get the corresponding already decoded event of ETH transfer and you are done. All in a simple decoder.

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+    def _map_erc20_lido_token_transfer(
+            self,
+            decoded_events: list['EvmEvent'],
+            transaction: 'EvmTransaction',  # pylint: disable=unused-argument
+            all_logs: list['EvmTxReceiptLog'],  # pylint: disable=unused-argument
+    ) -> list['EvmEvent']:
+        """Post decoding function to map the ERC20 Lido tokens reception"""
+        """TODO: To be checked with Rotki team, I am adding post decoding because I see that ERC20
+        transfers are not available in decoded_events at the time of decode by address rules
+        from what I have seen in the code. I got this solution looking at compound rules"""
```
LefterisJP, line 134: Never add two docstrings.

```suggestion
        """TODO: To be checked with Rotki team, I am adding post decoding because I see that ERC20
        transfers are not available in decoded_events at the time of decode by address rules
        from what I have seen in the code. I got this solution looking at compound rules"""
```

I don't understand what this is supposed to do. As I said post decoding rules should be avoided like the plague as they are really slow and are adding extra steps in our event processing flow.

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+            ):
+                event.notes = event.notes.format(amount=event.balance.amount)  # set the amount  # noqa: E501
+                break
+        return decoded_events
+
+    def _decode_lido_transfer_shares_for_steth_wrap(
+            self, context: DecoderContext,
+    ) -> DecodingOutput:
+        """Decode steth wrapping in wsteth from TransferShares log event"""
+
```
LefterisJP, line 151:
```suggestion
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+            if (
+                tx_log.log_index == expected_steth_transfer_index and
+                tx_log.topics[0] == ERC20_OR_ERC721_TRANSFER and
+                tx_log.address == steth_evm_address and
+                hex_or_bytes_to_address(tx_log.topics[1]) == user_address and  # from
+                hex_or_bytes_to_address(tx_log.topics[2]) == wsteth_evm_address  # to # wsteth
+            ):
+                sth_wrapped_amount = asset_normalized_value(
+                    amount=hex_or_bytes_to_int(tx_log.data[:32]),
+                    asset=A_STETH.resolve_to_crypto_asset(),
```
LefterisJP, line 173: don't do that all the time. If you want to use them, have them as class attributes but best to not use asset_normalize_value at all. We know both of those have 18 decimals. Just use the functon that determines amount from decimals. No need to resolve anything. Resolution is non-trivial.

Same for all the file and any future changes

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+        #  expected log index for the transferred wsteth is -3
+        expected_steth_transfer_index = context.tx_log.log_index - 1
+        expected_wsteth_transfer_index = context.tx_log.log_index - 3
+        paired_wsteth_ratio = None
+        sth_wrapped_amount = None
+        wsteth_received_amount = None
+
+        for tx_log in context.all_logs:
+            if (
+                tx_log.log_index == expected_steth_transfer_index and
```
LefterisJP, line 165: do not count on the log index. Please as I told you in another PR this does not help and is not deterministic. Just cause you saw it in some events does not make it a rule. Log index is non-deterministic and depends on other events in the same transaction. Remove this as a check.

rotkehlchen/chain/ethereum/modules/lido/decoder.py
LefterisJP: I would map it with same events as we do in compound, aave and other tokens. Not as staking. So same events type/subtypes as we use in compound, aave etc for wrapping and unwrapping a token

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+                sth_wrapped_amount = asset_normalized_value(
+                    amount=hex_or_bytes_to_int(tx_log.data[:32]),
+                    asset=A_STETH.resolve_to_crypto_asset(),
+                )
+            elif (
+                tx_log.log_index == expected_wsteth_transfer_index and
+                tx_log.topics[0] == ERC20_OR_ERC721_TRANSFER and
+                tx_log.address == wsteth_evm_address and
+                hex_or_bytes_to_address(tx_log.topics[1]) == ZERO_ADDRESS and  # from
+                hex_or_bytes_to_address(tx_log.topics[2]) == user_address  # to # wsteth
```
LefterisJP, line 180: like I told you before. There is no need to search the logs to see this. If it's not available when you are decoding this particular event it will be a bit later. This is what we have action items for. Study how they are used.

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+            if sth_wrapped_amount and wsteth_received_amount:
+                paired_wsteth_ratio = FVal(sth_wrapped_amount) / FVal(wsteth_received_amount)
+                break
+
+        if paired_wsteth_ratio is None:
+            return DEFAULT_DECODING_OUTPUT
+
+        in_event = None
+        out_event = None
+        for event in context.decoded_events:
```
LefterisJP, line 196: okay so this function has gotten super complicated. Iterating all log entries. And then all decoded events. I do not understand what you are trying to achieve. It seems too complicated and perhaps you are not understanding how the decoders work.

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+                event.event_type == HistoryEventType.SPEND and
+                event.event_subtype == HistoryEventSubType.NONE and
+                event.asset == A_STETH and
+                event.balance.amount == sth_wrapped_amount and
+                event.address == wsteth_evm_address
+            ):
+                event.event_type = HistoryEventType.DEPOSIT
+                event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
+                event.counterparty = CPT_LIDO_ETH
+                event.notes = f'Sent {sth_wrapped_amount} {A_STETH.symbol_or_name()} for wrapping at wstETH ratio {paired_wsteth_ratio}'  # noqa: E501
```
LefterisJP, line 207: Something to keep in mind for all events. I won't comment on all others.

1. Always present tense
2. no need to run function when needed. We know symbol of the token.
3. Why do you need the ratio? Just informational purposes?
```suggestion
                event.notes = f'Send {sth_wrapped_amount} stETH for wrapping at wstETH ratio {paired_wsteth_ratio}'  # noqa: E501
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+
+        return DecodingOutput(matched_counterparty=CPT_LIDO_ETH, refresh_balances=True)
+
+    def _decode_lido_transfer_shares_for_wsteth_unwrap(
+            self, context: DecoderContext,
+    ) -> DecodingOutput:
+        """Decode wsteth unwrapping in steth from TransferShares log event"""
+
+        user_address = hex_or_bytes_to_address(context.tx_log.topics[2])
+
```
LefterisJP, line 235: please don't add so many random empty lines

```suggestion
        user_address = hex_or_bytes_to_address(context.tx_log.topics[2])
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+        paired_wsteth_ratio = None
+        # we need to retrieve the transferred amount deterministically.
+        # log index baseline: transferShares log index
+        # expect log index for the transferred steth is -1
+        # expect log index for the transferred steth is -2
+        expected_steth_transfer_index = context.tx_log.log_index - 1
+        expected_wsteth_transfer_index = context.tx_log.log_index - 2
+        sth_received_amount = None
+        wsteth_unwrapped_amount = None
+        for tx_log in context.all_logs:
```
LefterisJP, line 245: again iterating all log events of the sender. This is not how decoders work. This is extremely slow

LefterisJP (commented):
Hey @gianluca-pub-dev thanks for the time you put at this PR!

I left a lot of comments but now after 40 mins of trying to understand what you are trying to do I have to take a step back as I can't dedicate more time at this.

I tried to leave some comments to explain how decoders are supposed to work.

My advice is the following.

1. Stash the current PR
2. Read and understand all the comments/feedback
3. Make a much smaller PR for just a single function. Say only to deposit ETH for stETH, along with a test. Nothing else. Only that. A really small and clean PR. And let's discusss for this single method first and try to see how to do it properly.

The point to understand is how decoders work. Please follow the code of the main logic of how [`decode_transaction`](https://github.com/rotki/rotki/blob/664484982a3b2eb23f9afbe2e12688d67e153840/rotkehlchen/chain/evm/decoding/decoder.py#L409) works and what is called and in what order. Also what kind of rules we have. The most normal ones, based on addresses of logs, then on input data, then post decoding etc.

But most importantly this should help you understand what gets called and how often. To see why what you wrote here suboptimal and why it would slow down the decoder system a lot.

LefterisJP (commented):
Also a test that failed: `rotkehlchen/tests/unit/decoders/test_main.py::test_decoders_initialization` needs super small tweak. Whenever you add a new decoder file you need to tweak this in two places to take the new decoder into account. Should be super easy to fix.

gianluca-pub-dev:
> Hey @gianluca-pub-dev thanks for the time you put at this PR!
> 
> I left a lot of comments but now after 40 mins of trying to understand what you are trying to do I have to take a step back as I can't dedicate more time at this.
> 
> I tried to leave some comments to explain how decoders are supposed to work.
> 
> My advice is the following.
> 
>     1. Stash the current PR
> 
>     2. Read and understand all the comments/feedback
> 
>     3. Make a much smaller PR for just a single function. Say only to deposit ETH for stETH, along with a test. Nothing else. Only that. A really small and clean PR. And let's discusss for this single method first and try to see how to do it properly.
> 
> 
> The point to understand is how decoders work. Please follow the code of the main logic of how [`decode_transaction`](https://github.com/rotki/rotki/blob/664484982a3b2eb23f9afbe2e12688d67e153840/rotkehlchen/chain/evm/decoding/decoder.py#L409) works and what is called and in what order. Also what kind of rules we have. The most normal ones, based on addresses of logs, then on input data, then post decoding etc.
> 
> But most importantly this should help you understand what gets called and how often. To see why what you wrote here suboptimal and why it would slow down the decoder system a lot.

Thank you @LefterisJP  for the time you spent for the review. I will take care of all the comments.

rotkehlchen/chain/ethereum/modules/lido/decoder.py
gianluca-pub-dev: As I reported in the comment, I was forced to use this solution because "submitted" log event happens before the erc20 transfer, and the decoded event erc20 transfer is not yet processed by rotki when the decoder for "submitted" is run. Also, basically I used exactly the same mechanism used in compound_v3 rules that was used for solving the same problem I think: https://github.com/rotki/rotki/blob/develop/rotkehlchen/chain/evm/decoding/compound/v3/decoder.py#L388

rotkehlchen/chain/ethereum/modules/lido/decoder.py
gianluca-pub-dev: 1 and 2, I will fix.
For 3, imho it is important to track the valute of the ratio at the moment the user obtained it, because it is not a fixed ratio, but it grows with time and stETH tokens obtained by Lido staking. With that, user can know its current gaining, looking at the original and current ratio.

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,526 @@
...
+                context=context,
+            )
+        else:
+            return DEFAULT_DECODING_OUTPUT
+
+    # -- DecoderInterface methods
+
+    def addresses_to_decoders(self) -> dict[ChecksumEvmAddress, tuple[Any, ...]]:
+        return {
+            steth_evm_address: (self._decode_lido_eth_staking_contract,),
```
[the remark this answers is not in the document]
gianluca-pub-dev: Ok, so please propose and educate me, which other address we can use for decoding the following eth submit transaction? https://etherscan.io/tx/0x23a3ee601475424e91bdc0999a780afe57bf37cbcce6d1c09a4dfaaae1765451#eventlog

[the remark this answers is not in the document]
LefterisJP: Can't use decoder mapping for this. It's a bad case where our decoders will have to be slow I am afraid. I would like to hear also @yabirgb's view.

But perhaps putting nothing except for a post decoding rule where you just use the existing decoded transfer from 0x.00 to the tracked address and transform this is the most performant thing to do.

rotkehlchen/chain/ethereum/modules/lido/decoder.py
gianluca-pub-dev: I added the searching of tx_log for 2 reasons:
1. stETH ERC20 transfer decoded event is not yet available at this point, so we need to add an ActionItem as done below.
2. The expected amount of stETH tokens is not exactly known a this point, due to the lido rounding issue: https://github.com/lidofinance/lido-dao/issues/442. Thus, I am iterating the tx log for retreiving the exact stETH amount to be passed to the ActionItem "_amount=steth_minted_tokens_", using a custom STETH_MAX_ROUND_ERROR_WEI for finding the exact amount in tx logs (set as 8 for now, but we need to decide the final value, maybe checking with Lido team).

rotkehlchen/chain/ethereum/modules/lido/decoder.py
gianluca-pub-dev: Unfortunately the amount is not the same, there is 1 wei difference in your example, as in the one I added in the test cases. Please see my reply to the previous comment, we can continue there.

LefterisJP:
hey @gianluca-pub-dev can't have multple threads in the review. It's impossible to follow. I asked a simple thing in the last review. Reduce the PR to just this and nothing else.

> Make a much smaller PR for just a single function. Only to deposit ETH for stETH, along with a test. Nothing else. Only that. A really small and clean PR. And let's discusss for this single method first and try to see how to do it properly.

Now for how to do it properly just use the single post decoding rule I mentioned. You start from the zero -> tracked address stETH transfer.

Then you can use whatever logic to find the ETH itnernal transfer of amount + difference. And that's it. You got both events.

Just do a PR with this single simple thing please. Force push on this one.

rotkehlchen/chain/ethereum/modules/lido/decoder.py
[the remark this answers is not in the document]
gianluca-pub-dev: Ok, using only post decoding it is something I can try, it seems to me it should work, and I could move the logic there. Only difference is that the code would not check anymore that "submit" method was called, but it should work because it seems it is only case where stETH are generated from zero address and transferred to the user address.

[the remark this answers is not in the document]
gianluca-pub-dev: but at the same time, you said in other comment that post decoding rules should be avoided, right? https://github.com/rotki/rotki/pull/7949#discussion_r1605903531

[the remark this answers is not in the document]
LefterisJP: yes it should be avoided so long as other options can work fine. Here they won't.

And address decoder based on a token's address is also pretty bad since it will run for every transfer/approval/other event related to the token.

From the information you shared, so far as I can see with the time I have now it's indeed a case of an event that can only be done via suboptimal decoders :(

gianluca-pub-dev:
> hey @gianluca-pub-dev can't have multple threads in the review. It's impossible to follow. I asked a simple thing in the last review. Reduce the PR to just this and nothing else.
> 
> > Make a much smaller PR for just a single function. Only to deposit ETH for stETH, along with a test. Nothing else. Only that. A really small and clean PR. And let's discusss for this single method first and try to see how to do it properly.
> 
> Now for how to do it properly just use the single post decoding rule I mentioned. You start from the zero -> tracked address stETH transfer.
> 
> Then you can use whatever logic to find the ETH itnernal transfer of amount + difference. And that's it. You got both events.
> 
> Just do a PR with this single simple thing please. Force push on this one.

Ok, I will reduce the size of the PR with only the eth submit for helping the review. But I am not sure that the solution of searching the stETH transfer in post_decoding is better as performance, since the function would be run for each transaction https://github.com/rotki/rotki/blob/664484982a3b2eb23f9afbe2e12688d67e153840/rotkehlchen/chain/evm/decoding/decoder.py#L480 .

I am going to wait some time before making this modification, so to give more time for taking a final decision.

LefterisJP:
So @gianluca-pub-dev the post decoding events have limitations. They run only if certain counterparties or certain input data have been found. So you would need to also set those conditions. Specifically here it being stETH transfer. Runnign always would indeed be bad.

But I remembered another decoding rule that is probably a perfect fit here that we made for such situations.

As I said above you need to find the stETH transfer from 0x0..00 to the tracked address right? So this is essentially augmenting a normal transfer. This is something we already do for some transfers via `_enrich_protocol_tranfers` logic.

All decoders also have the possiility of adding a token enricher rule. This is exactly what we need here. So add some extra logic to run once the token transfer is made and only then.

An example for this is already done in xdai bridge logic: https://github.com/rotki/rotki/blob/664484982a3b2eb23f9afbe2e12688d67e153840/rotkehlchen/chain/ethereum/modules/xdai_bridge/decoder.py#L46-L71

gianluca-pub-dev:
> .

Ok, but `_enrich_protocol_tranfers ` would be run for each token transfer regardless of the token type, right? https://github.com/rotki/rotki/blob/664484982a3b2eb23f9afbe2e12688d67e153840/rotkehlchen/chain/evm/decoding/decoder.py#L947

And with your proposal, inside its logic I should check if the token transfer is the one we are insterested, and then proceed.

But this would not work for mapping the stETH wrap or unwrap in wstETH, because 2 token transfers are involved on that operation, and both token transfers would not be already decoded at that time since `_enrich_protocol_tranfers ` is run in the context of processing one token transfer. And an `ActionItem` could not be used in this case without iterating the tx_log as I proposed, because the wstETH ratio is not known.

Also I need to say that I arrived to propose this PR after looking at all the options I found in rotki code, including enricher rules. It still seems to me that the best option is the iteration of tx_log for matching the operations, after checking that the topic is the one we are interested on. But I am open to listen any feedback.

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,156 @@
...
+                    token_amount=hex_or_bytes_to_int(tx_log.data[:32]),
+                    token_decimals=18,
+                )
+                break
+        else:  # did not break/find anything
+            log.error(
+                f'At lido steth submit decoding of tx {context.transaction.tx_hash.hex()}'
+                f'did not find the related stETH token generation',
+            )
+            return DEFAULT_DECODING_OUTPUT
```
LefterisJP, line 78: So I had an idea here to improve this approach.

I think you don't need to do this search of all logs. Instead since the ActionItem will already do this matching for you for everything **except** this `STETH_MAX_ROUND_ERROR_WEI` why not move this to the `ActionItem` class itself? So have an extra optional attribute. Something like `amount_error_tolerance` in the `ActionItem` class? 

And here:

https://github.com/rotki/rotki/blob/664484982a3b2eb23f9afbe2e12688d67e153840/rotkehlchen/chain/evm/decoding/decoder.py#L906

modify the check if the `action_item.amount_error_tolerance is not None` to do the absolute check you have there.

So you keep what you have here except this check of all logs. Does this make sense?

```suggestion
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,156 @@
...
+                event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
+                event.notes = f'Submit {collateral_amount} ETH to Lido for receiving stETH in exchange'  # noqa: E501
+                event.counterparty = CPT_LIDO_ETH
+                #  preparing next action to be processed when erc20 transfer will be decoded by rotki  # noqa: E501
+                #  needed because submit levent is emitted prior of erc20 transfer, so it is not decoded yet  # noqa: E501
+                # TODO: to be confirmed with ROTKI team if it is not possible to have the erc20 event available before this decoder is called  # noqa: E501
+                paired_event = event
+                action_from_event_type = HistoryEventType.RECEIVE
+                action_to_event_subtype = HistoryEventSubType.RECEIVE_WRAPPED
+                action_to_notes = f'Receive {steth_minted_tokens} stETH in exchange of the deposited ETH'  # noqa: E501
```
LefterisJP, line 101: So you will probably need an extra action in the `ActionItem` processing in evm/decoding/decoder.py that if there is an amount tolerance check to change the amount in the notes when matching happens.

So here: https://github.com/rotki/rotki/blob/664484982a3b2eb23f9afbe2e12688d67e153840/rotkehlchen/chain/evm/decoding/decoder.py#L923

(pseudocode)

```python
                if action_item.to_notes is not None:
                    transfer.notes = action_item.to_notes if action_item.amount_error_tolerance is None else action_item.to_notes.format(amount=transfer.balance.amount)
```

effectively making it take the correct amount, which is the amount of the matched ERC20 token transfer.

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,156 @@
...
+                event.asset == A_ETH and
+                event.balance.amount == collateral_amount and
+                event.event_type == HistoryEventType.SPEND and
+                event.location_label == sender_address
+            ):
+                event.event_type = HistoryEventType.DEPOSIT
+                event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
+                event.notes = f'Submit {collateral_amount} ETH to Lido for receiving stETH in exchange'  # noqa: E501
+                event.counterparty = CPT_LIDO_ETH
+                #  preparing next action to be processed when erc20 transfer will be decoded by rotki  # noqa: E501
```
LefterisJP, line 95:
```suggestion
                #  preparing next action to be processed when erc20 transfer will be decoded by rotki
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,156 @@
...
+                #  needed because submit levent is emitted prior of erc20 transfer, so it is not decoded yet  # noqa: E501
```
LefterisJP, line 96:
```suggestion
                #  needed because submit event is emitted prior to the erc20 transfer
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,156 @@
...
+                # TODO: to be confirmed with ROTKI team if it is not possible to have the erc20 event available before this decoder is called  # noqa: E501
```
LefterisJP, line 97:
```suggestion
```

rotkehlchen/chain/ethereum/modules/lido/constants.py
```diff
@@ -0,0 +1,13 @@
+from typing import Final
+
+CPT_LIDO_ETH: Final = 'lido'
```
LefterisJP, line 3:
```suggestion
CPT_LIDO: Final = 'lido'
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,156 @@
...
+                paired_event = event
+                action_from_event_type = HistoryEventType.RECEIVE
+                action_to_event_subtype = HistoryEventSubType.RECEIVE_WRAPPED
+                action_to_notes = f'Receive {steth_minted_tokens} stETH in exchange of the deposited ETH'  # noqa: E501
+                break
```
LefterisJP, line 102:
```suggestion
                break
            else:  # did not find anything
                log.error(
                    f'At lido steth submit decoding of tx {context.transaction.tx_hash.hex()}'
                    f' did not find the related ETH transfer',
            )
            return DEFAULT_DECODING_OUTPUT
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,156 @@
...
+
+        if (
+            context.tx_log.topics[0] == LIDO_STETH_SUBMITTED and
+            self.base.any_tracked([
+                hex_or_bytes_to_address(context.tx_log.topics[1]),  # sender address
+            ])
+        ):
+            return self._decode_lido_staking_in_steth(
+                context=context,
+            )
```
LefterisJP, line 138: request: Dont' break code in multiple lines if it fits in one

```suggestion
            context.tx_log.topics[0] == LIDO_STETH_SUBMITTED and
            (sender := self.base.any_tracked([
                hex_or_bytes_to_address(context.tx_log.topics[1]),
            ]))
        ):
            return self._decode_lido_staking_in_steth(context=context, sender=sender)
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,156 @@
...
+            return DEFAULT_DECODING_OUTPUT
+
+    # -- DecoderInterface methods
+
+    def addresses_to_decoders(self) -> dict[ChecksumEvmAddress, tuple[Any, ...]]:
+        return {
+            self.steth_evm_address: (self._decode_lido_eth_staking_contract,),
+        }
+
+    def addresses_to_counterparties(self) -> dict['ChecksumEvmAddress', str]:
```
LefterisJP, line 149: Where are you using this? Why are you mapping address to counterparty here?

rotkehlchen/tests/unit/decoders/test_lido_eth.py
```diff
@@ -0,0 +1,71 @@
...
+from rotkehlchen.chain.evm.decoding.constants import CPT_GAS
+from rotkehlchen.constants.assets import A_ETH, A_STETH
+from rotkehlchen.fval import FVal
+from rotkehlchen.history.events.structures.evm_event import EvmEvent
+from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.tests.utils.ethereum import get_decoded_events_of_transaction
+from rotkehlchen.types import Location, TimestampMS, deserialize_evm_tx_hash
+
+
+# @pytest.mark.vcr()
```
LefterisJP, line 15: Don't leave commented out code in a PR. This you have to add to our cassettes. I think we discussed how you can do in dev-support right? If not let's chat there.

rotkehlchen/tests/unit/decoders/test_lido_eth.py
```diff
@@ -0,0 +1,71 @@
...
+    evmhash = deserialize_evm_tx_hash(tx_hex)
+    events, _ = get_decoded_events_of_transaction(
+        evm_inquirer=ethereum_inquirer,
+        database=database,
+        tx_hash=tx_hex,
+    )
+    timestamp = TimestampMS(1710486191000)
+    gas_str = '0.002846110430778206'
+    amount_deposited_eth_str = '1.12137397'
+    amount_minted_steth_str = '1.121373969999999999'
```
LefterisJP, line 28: Please put all those extra test vars in one line. And maybe simplify naming.
```suggestion
    timestamp, gas_str, amount_deposited, amount_minted = TimestampMS(1710486191000), '0.002846110430778206', '1.12137397', '1.121373969999999999'  # noqa: E501
```

LefterisJP (commented):
hey @gianluca-pub-dev I left another review of this single action PR.

> And with your proposal, inside its logic I should check if the token transfer is the one we are insterested, and then proceed.
>But this would not work for mapping the stETH wrap or unwrap in wstETH, because 2 token transfers are involved on that operation, and both token transfers would not be already decoded at that time since _enrich_protocol_tranfers  is run in the context of processing one token transfer. And an ActionItem could not be used in this case without iterating the tx_log as I proposed, because the wstETH ratio is not known.
> Also I need to say that I arrived to propose this PR after looking at all the options I found in rotki code, including enricher rules. It still seems to me that the best option is the iteration of tx_log for matching the operations, after checking that the topic is the one we are interested on. But I am open to listen any feedback.

So the reason I suggested enrich transfer is because your current approach:

1. It will run for all stETH related events. Approvals, transfers, deposits etc. Anything that will touch this contract's address.
2. You are iterating all logs once to find the stETH ERC20 transfer due to the amount difference
3. You are then iterating all decoded events to find the internal ETH transfer
4. Then your are using the action item functionality which will match the decoded ERC20 transfer in the future

This is all a lot of things to do and quite often as stETH is one of the most used tokens in ethereum mainnet.

Alternatively my suggestion was to use enrich transfer due to the fact that with just a simple if condition addition to all ERC20 transfers you can go to the logic you want. So it would just add a single if condition.

-----

But after looking at the smaller PR (there is a very good reason we ask for small PRs so they are easy to understand and review) I see some more improvements which can make toyour suggestion even more preferrable to anything else.

I left them in the review itsef. Please talk to me and/or other devs in dev-support channel in discord if something is unclear

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,126 @@
...
+                event.balance.amount == collateral_amount and
+                event.event_type == HistoryEventType.SPEND and
+                event.location_label == sender
+            ):
+                event.event_type = HistoryEventType.DEPOSIT
+                event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
+                event.notes = f'Submit {collateral_amount} ETH to Lido for receiving stETH in exchange'  # noqa: E501
+                event.counterparty = CPT_LIDO
+                #  preparing next action to be processed when erc20 transfer will be decoded by rotki  # noqa: E501
+                #  needed because submit event is emitted prior to the erc20 transfer
```
LefterisJP, line 75: nitpick. Can just move comment up/down to not need the noqa

```suggestion
                #  preparing next action to be processed when erc20 transfer will be decoded
                #  by rotki needed because submit event is emitted prior to the erc20 transfer
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,126 @@
...
+        super().__init__(
+            evm_inquirer=evm_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+        )
+        self.steth_evm_address = A_STETH.resolve_to_evm_token().evm_address
+
+    def _decode_lido_staking_in_steth(self, context: DecoderContext, sender: ChecksumEvmAddress) -> DecodingOutput:  # noqa: E501
+        """Decode the submit of eth to lido contract for obtaining steth in return"""
+        # decode the ETH collateral amount submitted for staking in Lido
```
LefterisJP, line 51:
```suggestion
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,126 @@
...
+                event.event_type = HistoryEventType.DEPOSIT
+                event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
+                event.notes = f'Submit {collateral_amount} ETH to Lido for receiving stETH in exchange'  # noqa: E501
+                event.counterparty = CPT_LIDO
+                #  preparing next action to be processed when erc20 transfer will be decoded by rotki  # noqa: E501
+                #  needed because submit event is emitted prior to the erc20 transfer
+                paired_event = event
+                action_from_event_type = HistoryEventType.RECEIVE
+                action_to_event_subtype = HistoryEventSubType.RECEIVE_WRAPPED
+                action_to_notes = f'Receive {{amount}} stETH in exchange of the deposited ETH'  # {amount} to be set by transform ActionItem  # noqa: E501,F541
```
LefterisJP, line 79: wording.

Also you don't need `f` sincethis is not an fstring. You are not using any variable in there. The lint warning `F541` seems correct.

```suggestion
                action_to_notes = 'Receive {amount} stETH in exchange for the deposited ETH'  # {amount} to be set by transform ActionItem  # noqa: E501
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,126 @@
...
+            if (
+                event.address == self.steth_evm_address and
+                event.asset == A_ETH and
+                event.balance.amount == collateral_amount and
+                event.event_type == HistoryEventType.SPEND and
+                event.location_label == sender
+            ):
+                event.event_type = HistoryEventType.DEPOSIT
+                event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
+                event.notes = f'Submit {collateral_amount} ETH to Lido for receiving stETH in exchange'  # noqa: E501
```
LefterisJP, line 72: Can keep it simple since the corresponding receive event already will explain it nicely

```suggestion
                event.notes = f'Submit {collateral_amount} ETH to Lido'  
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,126 @@
...
+
+        action_items = []  # also create an action item for the reception of the stETH tokens
+        if paired_event is not None and action_from_event_type is not None:
+            action_items.append(ActionItem(
+                action='transform',
+                from_event_type=action_from_event_type,
+                from_event_subtype=HistoryEventSubType.NONE,
+                asset=A_STETH,
+                amount=collateral_amount,
+                amount_error_tolerance=FVal(token_raw_value_decimals(STETH_MAX_ROUND_ERROR_WEI, 18)),  # passing the maximum rounding error for finding the related stETH transfer  # noqa: E501
```
LefterisJP, line 96: Can use the `from_wei` function. Simpler. Also make sure to be rebased on develop so it works with int and returns FVal.

```suggestion
                amount_error_tolerance=from_wei(STETH_MAX_ROUND_ERROR_WEI),  # passing the maximum rounding error for finding the related stETH transfer  # noqa: E501
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,126 @@
...
+                to_counterparty=CPT_LIDO,
+                paired_event_data=(paired_event, True),
+                extra_data={'staked_eth': str(collateral_amount)},
+            ))
+
+        return DecodingOutput(action_items=action_items, matched_counterparty=CPT_LIDO)
+
+    def _decode_lido_eth_staking_contract(self, context: DecoderContext) -> DecodingOutput:
+        """Decode interactions with stETH ans wstETH contracts"""
+
```
LefterisJP, line 108:
```suggestion
```

rotkehlchen/chain/ethereum/modules/lido/decoder.py
```diff
@@ -0,0 +1,126 @@
...
+        if (
+            context.tx_log.topics[0] == LIDO_STETH_SUBMITTED and
+            self.base.any_tracked([hex_or_bytes_to_address(context.tx_log.topics[1])])
+        ):
+            return self._decode_lido_staking_in_steth(context=context, sender=hex_or_bytes_to_address(context.tx_log.topics[1]))  # noqa: E501
```
LefterisJP, line 113: As I pointed out here: https://github.com/rotki/rotki/pull/7949#discussion_r1606113674

1. Use the walrus operator. This way you don't run `hex_or_bytes_to_address` twice.
2. Use is_tracked. It's just 1 address.

```suggestion
        if (
            context.tx_log.topics[0] == LIDO_STETH_SUBMITTED and
            self.base.is_tracked(sender := hex_or_bytes_to_address(context.tx_log.topics[1]))
        ):
            return self._decode_lido_staking_in_steth(context=context, sender=sender)
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -903,7 +903,14 @@ def _maybe_decode_erc20_721_transfer(
                     action_item.asset == found_token and
                     action_item.from_event_type == transfer.event_type and
                     action_item.from_event_subtype == transfer.event_subtype and
-                    (action_item.amount is None or action_item.amount == transfer.balance.amount) and  # noqa: E501
+                    (
+                        (action_item.amount is None or action_item.amount == transfer.balance.amount) or  # noqa: E501
+                        (
+                            action_item.amount_error_tolerance is not None and
+                            action_item.amount is not None and
+                            abs(action_item.amount - transfer.balance.amount) < action_item.amount_error_tolerance  # abs used for supporting also different cases than stETH, if they exist # noqa: E501
```
LefterisJP, line 911: no need for the comment. Makes sense to always use abs for tolerance.

```suggestion
                            abs(action_item.amount - transfer.balance.amount) < action_item.amount_error_tolerance
```

rotkehlchen/chain/evm/decoding/structures.py
```diff
@@ -29,6 +29,8 @@ class ActionItem(NamedTuple):
     # Optional event data that pairs it with the event of the action item
     # Contains a tuple with the paired event and whether it's an out event (True) or in event
     paired_event_data: tuple['EvmEvent', bool] | None = None
+    # Error tolerance that can be used for protocols having rounding errors as stETH
```
LefterisJP, line 32: Just a few more details.

```suggestion
    # Error tolerance that can be used for protocols having rounding errors. Such as with stETH (https://github.com/lidofinance/lido-dao/issues/442)
    # In those cases the notes should also be formatted to have an amount as format string since at
    # action item matching this format will populate the note with the actual amount
```

LefterisJP (commented):
Left some more comments. Also pushed your VCR PR to upstream so tests run properly.

rotkehlchen/chain/evm/decoding/decoder.py
gianluca-pub-dev: I left the comment only temporarly for explaining during the review.

rotkehlchen/tests/unit/decoders/test_lido_eth.py
```diff
@@ -0,0 +1,68 @@
...
+            tx_hash=evmhash,
+            sequence_index=1,
+            timestamp=timestamp,
+            location=Location.ETHEREUM,
+            event_type=HistoryEventType.DEPOSIT,
+            event_subtype=HistoryEventSubType.DEPOSIT_ASSET,
+            asset=A_ETH,
+            balance=Balance(FVal(amount_deposited)),
+            location_label=ethereum_accounts[0],
+            notes=f'Submit {amount_deposited} {A_ETH.symbol_or_name()} to Lido',
```
LefterisJP, line 49: This is constant. No need to resolve (as resolution is DB roundtrip)

```suggestion
            notes=f'Submit {amount_deposited} ETH to Lido',
```

rotkehlchen/tests/unit/decoders/test_lido_eth.py
```diff
@@ -0,0 +1,68 @@
...
+            tx_hash=evmhash,
+            sequence_index=2,
+            timestamp=timestamp,
+            location=Location.ETHEREUM,
+            event_type=HistoryEventType.RECEIVE,
+            event_subtype=HistoryEventSubType.RECEIVE_WRAPPED,
+            asset=A_STETH,
+            balance=Balance(FVal(amount_minted)),
+            location_label=ethereum_accounts[0],
+            notes=f'Receive {amount_minted} {A_STETH.symbol_or_name()} in exchange for the deposited {A_ETH.symbol_or_name()}',  # noqa: E501
```
LefterisJP, line 62: sam

```suggestion
            notes=f'Receive {amount_minted} stETH in exchange for the deposited ETH',  # noqa: E501
```

LefterisJP (approved):
LGTM. Left 2 small nitpicks that can just be done in next PR/s

PR #7955, opened 2024-05-20
Feat (ZkSync-lite): Notify decoding progress

### LINKED ISSUE #7905: zksync lite websockets messages
When we decode zksync lite transactions we don't send a websocket message to update the frontend about the progress and they don't know when the job finished.

Task:

- add websockets messages like we do for evm transactions to update about the progress
- coordinate with the frontend to ensure that those are correctly processed

OjusWiZard:
## Checklist

- [x] Notify progress of decoding zksync-lite transactions through a new WS message type

### DIFF
--- a/rotkehlchen/chain/zksync_lite/manager.py
+++ b/rotkehlchen/chain/zksync_lite/manager.py
@@ -11,6 +11,7 @@ import requests
 from pysqlcipher3.dbapi2 import IntegrityError
 
 from rotkehlchen.accounting.structures.balance import Balance
+from rotkehlchen.api.websockets.typedefs import WSMessageType
 from rotkehlchen.assets.asset import Asset, CryptoAsset
 from rotkehlchen.assets.utils import TokenEncounterInfo, get_or_create_evm_token
 from rotkehlchen.chain.ethereum.utils import asset_normalized_value
@@ -32,6 +33,7 @@ from rotkehlchen.serialization.deserialize import deserialize_evm_address, deser
 from rotkehlchen.types import (
     ChainID,
     ChecksumEvmAddress,
+    EvmlikeChain,
     EVMTxHash,
     Fee,
     Location,
@@ -838,9 +840,14 @@ class ZksyncLiteManager:
                 (1, transaction.tx_hash),
             )
 
-    def decode_undecoded_transactions(self, force_redecode: bool) -> int:
+    def decode_undecoded_transactions(
+            self,
+            force_redecode: bool,
+            send_ws_notifications: bool = False,
+    ) -> int:
         """Decodes undecoded zksync lite transactions. If force redecode is True
-        then all transactions are redecoded.
+        then all transactions are redecoded. If send_ws_notifications is True then the decoding
+        progress is broadcasted via websocket messages.
         Returns the number of decoded transactions (not events in transactions)
         """
         queryfilter, bindings = '', ()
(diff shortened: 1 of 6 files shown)


rotkehlchen/chain/zksync_lite/manager.py
```diff
@@ -860,4 +866,24 @@ def decode_undecoded_transactions(self, force_redecode: bool) -> int:
 
             self.decode_transaction(transaction, tracked_addresses)
 
+            if send_ws_notifications and tx_index % 10 == 0:
+                self.msg_aggregator.add_message(
+                    message_type=WSMessageType.EVMLIKE_UNDECODED_TRANSACTIONS,
+                    data={
+                        'evm_chain': self.ethereum_inquirer.chain_name,
+                        'total': len(transactions),
```
yabirgb, line 874: define as a variable. I know that is O(1) but left prefers having the call outiside of the loop. Having them in python can have a big impact on performance

rotkehlchen/chain/zksync_lite/manager.py
```diff
@@ -860,4 +866,24 @@ def decode_undecoded_transactions(self, force_redecode: bool) -> int:
...
+                    },
+                )
+
+        if send_ws_notifications:
+            self.msg_aggregator.add_message(
+                message_type=WSMessageType.EVMLIKE_UNDECODED_TRANSACTIONS,
+                data={
+                    'evm_chain': self.ethereum_inquirer.chain_name,
+                    'total': len(transactions),
+                    'processed': len(transactions),
```
yabirgb, line 885: use the variable instead of calling twice

rotkehlchen/chain/zksync_lite/manager.py
```diff
@@ -860,4 +866,24 @@ def decode_undecoded_transactions(self, force_redecode: bool) -> int:
...
+                    message_type=WSMessageType.EVMLIKE_UNDECODED_TRANSACTIONS,
+                    data={
+                        'evm_chain': self.ethereum_inquirer.chain_name,
+                        'total': len(transactions),
+                        'processed': tx_index,
+                    },
+                )
+
+        if send_ws_notifications:
+            self.msg_aggregator.add_message(
```
yabirgb, line 880: should it have a docstring entry? is fixing the issue where the total number of decoded transactions is not tracked properly

rotkehlchen/chain/zksync_lite/manager.py
```diff
@@ -61,6 +62,7 @@ def __init__(
             database: 'DBHandler',
     ) -> None:
         self.database = database
+        self.msg_aggregator = database.msg_aggregator
```
yabirgb, line 65: no need to create a new variable for this. Access self.database.msg_aggregator directly

rotkehlchen/chain/zksync_lite/manager.py
```diff
@@ -860,4 +868,24 @@ def decode_undecoded_transactions(self, force_redecode: bool) -> int:
...
+                        'total': total_transactions,
+                        'processed': tx_index,
+                    },
+                )
+
+        if send_ws_notifications:
+            self.msg_aggregator.add_message(
+                message_type=WSMessageType.EVM_UNDECODED_TRANSACTIONS,
+                data={
+                    'evm_chain': self.ethereum_inquirer.chain_name,
```
yabirgb, line 885: this is wrong, will return ethereum instead of zksync_lit

rotkehlchen/chain/zksync_lite/manager.py
```diff
@@ -860,4 +868,24 @@ def decode_undecoded_transactions(self, force_redecode: bool) -> int:
 
             self.decode_transaction(transaction, tracked_addresses)
 
+            if send_ws_notifications and tx_index % 10 == 0:
+                self.msg_aggregator.add_message(
+                    message_type=WSMessageType.EVM_UNDECODED_TRANSACTIONS,
+                    data={
+                        'evm_chain': self.ethereum_inquirer.chain_name,
```
yabirgb, line 875: can you change here `evm_chain` to `chain` as we do in places where we support evm and evm like? Change it in all the places. Also in the `websockets.rst` file and update the docs there

LefterisJP (approved):
LGTM

PR #7963, opened 2024-05-21
Chore (CI): Fail if VCR is not full rebased

OjusWiZard:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=63040434

## Checklist

- [x] Fail CI if the VCR branch is not fully rebased on target branch

### DIFF
--- a/.github/workflows/task_backend_tests.yml
+++ b/.github/workflows/task_backend_tests.yml
@@ -15,6 +15,8 @@ jobs:
     env:
       CI: true
       TEST_ENVIRONMENT: ${{ inputs.test_environment }}
+      TARGET_BRANCH: ${{ github.base_ref }}
+      SOURCE_BRANCH: ${{ github.head_ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/task_backend_tests.yml
```diff
@@ -32,6 +34,16 @@ jobs:
         with:
           repository: rotki/test-caching
           path: test-caching
+          fetch-depth: 0
+      - name: Ensure VCR branch is fully rebased
+        run: |
+          cd $HOME/work/rotki/rotki/test-caching
```
kelsos, line 40: here we could potentially replace that with working directory https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsworking-directory but if it works as it is it's also ok

kelsos (approved):
looks good

PR #7973, opened 2024-05-23
ci: remove no-optional flag and disable cypress binary

kelsos:
Closes #(issue_number)

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ ENV CYPRESS_INSTALL_BINARY=0
 WORKDIR /app
 COPY frontend/ .
 RUN if [ "$BUILDARCH" != "amd64" ]; then apt-get update && apt-get install -y build-essential python3 --no-install-recommends; fi
-RUN npm install -g pnpm@9 && pnpm install --no-optional --frozen-lockfile
+RUN npm install -g pnpm@9 && pnpm install --frozen-lockfile
 RUN pnpm run docker:build
 
 FROM python:3.11-buster as backend-build-stage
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -7,7 +7,7 @@ ENV CYPRESS_INSTALL_BINARY=0
 WORKDIR /app
 COPY frontend/ .
 RUN if [ "$BUILDARCH" != "amd64" ]; then apt-get update && apt-get install -y build-essential python3 --no-install-recommends; fi
-RUN npm install -g pnpm@9 && pnpm install --no-optional --frozen-lockfile
+RUN npm install -g pnpm@9 && pnpm install --frozen-lockfile
```
yabirgb, line 10: in docker don't you need to keep it? or set the environment variable too?

package.py
```diff
@@ -96,6 +96,7 @@ def __init__(self) -> None:
         os.environ.pop(CERTIFICATE_KEY, None)
         os.environ.pop(APPLE_ID, None)
         os.environ.pop(APPLE_ID_PASS, None)
+        os.environ.setdefault('CYPRESS_INSTALL_BINARY', '0')
```
yabirgb, line 99: is this related to the removal of the no-optional?

kelsos: yes this stops the installation from packaging from downloading the Cypress binary when installing because we don't really need it, the whole idea behind --no-optional was to avoid downloading cypress on packaging and save some time.

Dockerfile
kelsos: It's already there in line 6 of the dockerfile so no need to set it

yabirgb (approved):
thanks kelsos. LGTM

PR #7975, opened 2024-05-24
Provide default key for subgraphs

yabirgb:
Also avoid notifications for balancer subgraph

### DIFF
--- a/rotkehlchen/chain/ethereum/graph.py
+++ b/rotkehlchen/chain/ethereum/graph.py
@@ -12,12 +12,13 @@ from gql.transport.requests import RequestsHTTPTransport
 from graphql.error import GraphQLError
 
 from rotkehlchen.api.websockets.typedefs import WSMessageType
+from rotkehlchen.chain.ethereum.modules.ens.constants import CPT_ENS
 from rotkehlchen.db.settings import CachedSettings
 from rotkehlchen.errors.api import APIKeyNotConfigured
 from rotkehlchen.errors.misc import RemoteError
 from rotkehlchen.externalapis.interface import ExternalServiceWithApiKey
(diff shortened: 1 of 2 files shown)


rotkehlchen/chain/ethereum/modules/ens/decoder.py
```diff
@@ -119,7 +119,7 @@ def __init__(  # pylint: disable=super-init-not-called
         self.graph = Graph(
             subgraph_id='5XqPmWe6gjyrJtFn9cLy237i4cWw2j9HcUJEXsP5qGtH',
             database=self.database,
-            label='ENS',
+            label=ENS_GRAPH_LABEL,
```
LefterisJP, line 122: why not jsut use same label as `CPT_ENS`? Does it need to be upper case? Why add another constant which is the ens constant but just upper?

rotkehlchen/chain/ethereum/graph.py
```diff
@@ -82,7 +89,7 @@ def _maybe_create_client(self) -> Client:
                     },
                 )
                 self.warning_given = True
-
+        elif api_key is None:  # don't bother the user and fail silently. We avoid exhausting the default key  # noqa: E501
```
LefterisJP, line 92: is this ApiKeyNotConfigured exception handled in a way that it just skips the query? I guess yes but need to confirm

yabirgb: Yes, in the ens decoder and in balancer, for example here https://github.com/rotki/rotki/blob/3856b9b9f5a7ce740511618cc2e5b31f93ce8333/rotkehlchen/chain/ethereum/modules/balancer/balancer.py#L119 but wherever it is called

PR #7980, opened 2024-05-24
Clean code

yabirgb:
Closes #(issue_number)

## Checklist


rotkehlchen/chain/gnosis/modules/airdrops/decoder.py
```diff
@@ -1,36 +1,17 @@
...
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
 
 class AirdropsDecoder(DecoderInterface):
 
-    def __init__(
-            self,
-            ethereum_inquirer: 'EthereumInquirer',
-            base_tools: 'BaseDecoderTools',
```
yabirgb, line 23: should we just remove this and move it to cowswap?

yabirgb: This doesn't seem to be useful. Removing it

rotkehlchen/chain/gnosis/modules/airdrops/decoder.py
```diff
@@ -1,36 +0,0 @@
-import logging
```
LefterisJP, line 1: Hmm so it seems this was an empty decoder class? Right?

LefterisJP (approved):
LGTM

rotkehlchen/chain/gnosis/modules/airdrops/decoder.py
yabirgb: yes, the only thing it was doing was to add the cowswap counterparty but we already have it with the decoder for the swaps. This one was a no-op it seems

PR #7982, opened 2024-05-25
Feat (CI): Post SQL diff on PRs

OjusWiZard:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=56761592

## Checklist

- [x] Add GitHub action to post SQL diff on PRs that modify any database.


kelsos:
this we can close, we don't need it anymore @OjusWiZard. I ported whatever we needed to the action so it should be ok now.

PR #7990, opened 2024-05-28
feat(frontend): graph service key

lukicenturi:
Closes #(issue_number)

## Checklist


frontend/app/src/composables/message-handling.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
```
kelsos, line 1: We should probably figure out how to split the file further here. not now but we should start thinking about it

kelsos (approved):
lgtm

PR #8004, opened 2024-05-29
Fix (Airdrops): Always include `has_decoder` field

OjusWiZard:
## Checklist

- [x] Always include `has_decoder` field in the metadata index, no matter if it's true or false


LefterisJP (approved):
yolo lgtm merge.

PR #8010, opened 2024-05-30
Adjust tests regarding thegraph

yabirgb:
We were creating a mock client and then call it but since we adjusted the key the app correctly detected that the password changed and the client was being recreated. Since we weren't using the mocked client the expected error wasn't being raised.


LefterisJP (approved):
Lgtm if tests pass

PR #8015, opened 2024-05-31
feat(frontend): export and import accounting rules

### LINKED ISSUE #7675: Make Accounting Rules Exportable/Importable
## Abstract

The Idea here would be to have an export button for the accounting rules so you could save them if you change them. 
With the import button you could actually restore the accounting rules again.

## Motivation

This would allow to firstly store & restore the accounting rules as well as you would be able to create different accounting rules that could also be shareable. 

Lastly an reset button could make sense to also restore the defaults 

## Specification

I guess from  a technical side it should not be a too complex thing to do :smiley:

lukicenturi:
Closes #7675

## Checklist


kelsos (approved):
The PR looks good, I only have a few really minor comments.

PR #8017, opened 2024-05-31
fix: change taskkill promise to ensure task execution order

kelsos:
Closes #(issue_number)

## Checklist


kelsos:
The promise before was completely broken and the restart would start the moment taskill exited regardless if the sub processes where still pending.

With this change it fixes that, so that the quit and install won't run until we ensure no background processes.

lukicenturi (approved):
LGTM!

PR #8025, opened 2024-06-03
Fix wrong order for nfts

### LINKED ISSUE #8022: NFT balances order seems to break if asset price is not in ETH
## Problem Definition

It seems that NFT balances order is not from highest to lowest if the asset that the pricing happens in is not ETH. And this seems to come from the backend response.

## Task

Figure out why and fix it

yabirgb:
Sort correctly by fiat price when querying the db with filters. It was using the last price that uses the last price assets but that order is not correct when multiple assets are used in the nft table

Closes #8022

Tried to do it in memory but since we have pagination is hard and it scaled in complexity


rotkehlchen/chain/ethereum/modules/nft/nfts.py
```diff
@@ -71,12 +71,15 @@ def _deserialize_nft_price(
...
-    """From a db tuple extract the information required by the API for a NFT"""
-    price_in_asset, price_asset, usd_price = _deserialize_nft_price(
-        last_price=entry[2],
-        last_price_asset=entry[3],
-    )
+def _deserialize_nft_from_db(
+        entry: NFT_DB_READ_TUPLE,
+        price_tuple: tuple[FVal, Asset, FVal],
+) -> dict[str, Any]:
+    """From a db tuple extract the information required by the API for a NFT.
```
LefterisJP, line 78:
```suggestion
    """From a db tuple extract the information required by the API for an NFT.
```

rotkehlchen/chain/ethereum/modules/nft/nfts.py
```diff
@@ -172,29 +175,42 @@ def get_all_info(
...
+
+        with self.db.user_write() as write_ctx:
+            write_ctx.executemany(
+                'UPDATE nfts SET usd_price=? WHERE identifier=?',
+                [(float(price_tuple[2]), ident) for ident, price_tuple in usd_price_tuples.items()],  # noqa: E501
+            )  # update the usd value since the consumer of the api always sorts by usd value
+
+        query, bindings = filter_query.prepare()
+        with self.db.conn.read_ctx() as cursor:
+            cursor.execute(  # this sometimes causes https://github.com/rotki/rotki/issues/5432
```
LefterisJP, line 205: This is from before or did you see it now?

rotkehlchen/db/schema.py
```diff
@@ -625,6 +625,9 @@
 """  # noqa: E501
 
 
+# usd value is a column of the table because we sort by value in the fiat currency and that price
```
LefterisJP, line 628: is it value or price? Please adjust either the docstring or the name of the column accordingly

rotkehlchen/db/upgrades/v42_v43.py
```diff
@@ -0,0 +1,31 @@
...
+    write_cursor.execute(
+        'ALTER TABLE nfts ADD COLUMN usd_price REAL NOT NULL DEFAULT 0;',
+    )
+
+
+@enter_exit_debug_log(name='UserDB v42->v43 upgrade')
+def upgrade_v42_to_v43(db: 'DBHandler', progress_handler: 'DBUpgradeProgressHandler') -> None:
+    """Upgrades the DB from v42 to v43. This was in v1.34 release.
+
+    - add usd_value to the nfts table
```
LefterisJP, line 26: again here you are mixing the term value and price. Across the whole PR and codebase please use only one. And probably that one should be price.

rotkehlchen/tests/unit/test_nfts.py
```diff
@@ -45,3 +46,35 @@ def test_nfts_floor_prices(
...
     assert nfts[0].collection.floor_price_asset == Asset('eip155:137/erc20:0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174')  # USDC.e on polygon # noqa: E501
+
+
+@pytest.mark.parametrize('ethereum_accounts', [[TEST_ACC1]])
+@pytest.mark.parametrize('start_with_valid_premium', [True])
+@pytest.mark.parametrize('ethereum_modules', [['nfts']])
+@pytest.mark.parametrize('mocked_current_prices', [{A_USDC.identifier: 1, A_ETH: 3600}])
+def test_sorting_nfts(blockchain: ChainsAggregator):
+    """Tests that nfts are only queried for addresses stored in the database preventing the
+    IntegrityError described in https://github.com/rotki/rotki/issues/4456
```
LefterisJP, line 57: The docstring seems to point to another issue. And the description is also off.

rotkehlchen/db/upgrades/v42_v43.py
yabirgb: I had it initially as value and then changed it to price since is the price what we are returning. Will adjust the comments

rotkehlchen/chain/ethereum/modules/nft/nfts.py
yabirgb: This is from before, I just moved the line at the start of the logic since is needed later

lukicenturi:
Frontend code is fine

rotkehlchen/chain/ethereum/modules/nft/nfts.py
```diff
@@ -172,29 +175,42 @@ def get_all_info(
...
+                usd_price_tuples[db_entry[2]] = (price_in_asset, price_asset, usd_price)  # save all the prices to reuse them when creating the response  # noqa: E501
                 total_usd_value += usd_price
                 entries_found += 1
-            entries_total = cursor.execute('SELECT COUNT(*) FROM nfts').fetchone()[0]
+
+        with self.db.user_write() as write_ctx:
+            write_ctx.executemany(
+                'UPDATE nfts SET usd_price=? WHERE identifier=?',
+                [(float(price_tuple[2]), ident) for ident, price_tuple in usd_price_tuples.items()],  # noqa: E501
+            )  # update the usd value since the consumer of the api always sorts by usd value
```
LefterisJP, line 201:
```suggestion
            )  # update the usd price since the consumer of the api always sorts by usd price
```

LefterisJP (approved):
Failing frontend test is something I asked frontend devs.

Failing backend is flaky and I made a VCRed version of the test here: https://github.com/rotki/rotki/pull/8033

PR #8035, opened 2024-06-06
Feat (Balances): Auto-detect tokens

### LINKED ISSUE #5103: Improve token scanning and balance detection combination
## Problem Definition

your net worth is shown in the main dashboard along with a graph of its value over time.   this major headline number is no longer accurate with the breaking change thats recently been made.

a breaking UI change has been made where Ethereum addresses are now no longer scanned by default for token values.  Whilst this optimisation makes address scanning faster, it does so at a large cost of leaving out major assets (tokens!) from your net worth so the headline net worth figure in the dashboard will no longer represent an accurate picture of your net worth, = clearly, this is undesirable from a UI point of view and defeats an important use case of Rotki, to know a true figure of your net worth and all of your assets that is automatically maintained.

if you really must have this optimisation in there, then I would recommend two UI fixes to make it less onerous.

1.  there should be a 're-detect all tokens' button sitting right next to 'blockchain balances' refresh button on the main dashboard.

2.  make sure that when your 'net worth' is re-scanned every 12 hours (or whatever its set to)... that it not only scans blockchain balances, but also tokens.   otherwise your net worth as shown on the headline is meaningless.

## Logs

no logs required.

-->

### System Description

all rotki versions.

Operating system: MacOS
Rotki version: 1.26.1

OjusWiZard:
Closes #5103

## Checklist

- [x] Check for active addresses which scheduling query_balance and detect their tokens

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`5103` rotki will now automatically detect tokens along with their balances if the setting has been turned on. Set to true by default.
 * :feature:`7675` Users will now be able to export & import accounting rules.
 * :feature:`8003` 1inch v6 events should now be properly decoded for all supported EVM chains.
 * :feature:`7981` Circle's CCTP bridge transfers are decoded properly for all supported EVM chains.
--- a/rotkehlchen/db/settings.py
+++ b/rotkehlchen/db/settings.py
@@ -61,6 +61,7 @@ DEFAULT_ORACLE_PENALTY_DURATION = 1800
 DEFAULT_AUTO_DELETE_CALENDAR_ENTRIES = True
 DEFAULT_AUTO_CREATE_CALENDAR_REMINDERS = True
 DEFAULT_ASK_USER_UPON_SIZE_DISCREPANCY = True
+DEFAULT_AUTO_DETECT_TOKENS = True
 
 JSON_KEYS = (
     'current_price_oracles',
@@ -86,6 +87,7 @@ BOOLEAN_KEYS = (
     'auto_delete_calendar_entries',
     'auto_create_calendar_reminders',
     'ask_user_upon_size_discrepancy',
+    'auto_detect_tokens',
 )
 INTEGER_KEYS = (
     'version',
@@ -214,6 +216,7 @@ class DBSettings:
     auto_delete_calendar_entries: bool = DEFAULT_AUTO_DELETE_CALENDAR_ENTRIES
     auto_create_calendar_reminders: bool = DEFAULT_AUTO_CREATE_CALENDAR_REMINDERS
     ask_user_upon_size_discrepancy: bool = DEFAULT_ASK_USER_UPON_SIZE_DISCREPANCY
+    auto_detect_tokens: bool = DEFAULT_AUTO_DETECT_TOKENS
 
     def serialize(self) -> dict[str, Any]:
         settings_dict = {}
@@ -273,6 +276,7 @@ class ModifiableDBSettings(NamedTuple):
     auto_delete_calendar_entries: bool | None = None
     auto_create_calendar_reminders: bool | None = None
     ask_user_upon_size_discrepancy: bool | None = None
+    auto_detect_tokens: bool | None = None
 
     def serialize(self) -> dict[str, Any]:
         settings_dict = {}
--- a/rotkehlchen/tasks/manager.py
+++ b/rotkehlchen/tasks/manager.py
@@ -14,6 +14,8 @@ from rotkehlchen.chain.ethereum.modules.makerdao.cache import (
 )
 from rotkehlchen.chain.ethereum.modules.yearn.utils import query_yearn_vaults
 from rotkehlchen.chain.ethereum.utils import should_update_protocol_cache
+from rotkehlchen.chain.evm.constants import EVM_ADDRESS_REGEX
+from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.constants.timing import (
     AAVE_V3_ASSETS_UPDATE,
     AUGMENTED_SPAM_ASSETS_DETECTION_REFRESH,
@@ -26,15 +28,22 @@ from rotkehlchen.constants.timing import (
 )
 from rotkehlchen.db.cache import DBCacheDynamic, DBCacheStatic
 from rotkehlchen.db.calendar import CalendarEntry
+from rotkehlchen.db.constants import EVM_EVENT_FIELDS, HISTORY_BASE_ENTRY_FIELDS
 from rotkehlchen.db.evmtx import DBEvmTx
-from rotkehlchen.db.filtering import EvmTransactionsFilterQuery, HistoryEventFilterQuery
-from rotkehlchen.db.history_events import DBHistoryEvents
+from rotkehlchen.db.filtering import (
+    EVM_EVENT_JOIN,
+    EvmTransactionsFilterQuery,
+    HistoryEventFilterQuery,
+)
+from rotkehlchen.db.history_events import DBHistoryEvents, filter_ignore_asset_query
 from rotkehlchen.db.settings import CachedSettings
 from rotkehlchen.errors.api import PremiumAuthenticationError
 from rotkehlchen.errors.asset import UnknownAsset, WrongAssetType
 from rotkehlchen.errors.misc import RemoteError
 from rotkehlchen.externalapis.monerium import Monerium
 from rotkehlchen.globaldb.handler import GlobalDBHandler
+from rotkehlchen.history.events.structures.evm_event import EvmEvent
+from rotkehlchen.history.events.structures.types import EventDirection
 from rotkehlchen.history.types import HistoricalPriceOracle
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.premium.premium import Premium, premium_create_and_verify
@@ -53,6 +62,7 @@ from rotkehlchen.tasks.calendar import (
 from rotkehlchen.tasks.utils import query_missing_prices_of_base_entries, should_run_periodic_task
 from rotkehlchen.types import (
     EVM_CHAINS_WITH_TRANSACTIONS,
+    EVM_LOCATIONS,
     SUPPORTED_BITCOIN_CHAINS,
     CacheType,
     ChecksumEvmAddress,
@@ -63,7 +73,7 @@ from rotkehlchen.types import (
     Timestamp,
     get_args,
 )
-from rotkehlchen.utils.misc import ts_now
+from rotkehlchen.utils.misc import ts_now, ts_sec_to_ms
 
 from .events import process_events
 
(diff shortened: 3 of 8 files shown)


rotkehlchen/tasks/manager.py
```diff
@@ -538,6 +541,36 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
...
+                if CachedSettings().get_settings().auto_detect_tokens is True:
+                    type_filter = [event_type.serialize() for event_type in [
+                        HistoryEventType.RECEIVE,
+                        HistoryEventType.WITHDRAWAL,
+                    ]]
+                    subtype_filter = [event_subtype.serialize() for event_subtype in [
+                        HistoryEventSubType.RECEIVE,
+                        HistoryEventSubType.REWARD,
+                        HistoryEventSubType.REMOVE_ASSET,
+                        HistoryEventSubType.MEV_REWARD,
```
yabirgb, line 554: mev reward can only be in eth right? I believe we don't need to count it

rotkehlchen/tasks/manager.py
```diff
@@ -538,6 +541,36 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
...
+                    ]]
+                    last_save_time = self.database.get_last_balance_save_time(cursor)
+                    for chain in EVM_CHAIN_IDS_WITH_TRANSACTIONS:
+                        if len(location_labels := cursor.execute(  # number of earliest receive events of distinct assets after last_save_time  # noqa: E501
+                            'SELECT DISTINCT location_label FROM '
+                            '(SELECT MIN(timestamp), location_label FROM history_events '
+                            f'WHERE (type IN ({",".join(["?"] * len(type_filter))}) OR '
+                            f'subtype IN ({",".join(["?"] * len(subtype_filter))})) '
```
yabirgb, line 562: One thing, the location label can't only be addresses. For kraken I believe is the exchange name. YOu need to add additional filters to limit the location to `EVM_LOCATIONS_TYPE`

rotkehlchen/rotkehlchen.py
```diff
@@ -901,6 +906,7 @@ def query_balances(
         If a timestamp is given then that is the time that the balances are going
         to be saved in the DB
         If ignore_cache is True then all underlying calls that have a cache ignore it
+        New tokens are detected for the addresses per blockchain in detect_tokens_for_addresses.
```
yabirgb, line 909: what is `detect_tokens_for_addresses`?

rotkehlchen/tasks/manager.py
```diff
@@ -538,6 +541,36 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
...
+                            'AND location = ? AND timestamp >= ? GROUP BY asset);',
+                            type_filter + subtype_filter + [
+                                Location.from_chain_id(chain).serialize_for_db(),
+                                last_save_time * 1000,  # to ms
+                            ],
+                        ).fetchall()) > 0:
+                            detect_tokens[chain.to_blockchain()] = [
+                                string_to_evm_address(location_label[0])
+                                for location_label in location_labels
+                            ]
```
yabirgb, line 572: am I missing something or you didn't use the variable in the task below?

should be something like

```
detect_tokens=detect_tokens if len(detec_tokens) else None

File: rotkehlchen/rotkehlchen.py [hunk 5]
```diff
@@ -910,6 +916,17 @@ def query_balances(
...
         )
 
+        if detect_tokens is not None:
+            for blockchain, addresses in detect_tokens.items():
+                manager: EvmManager = self.chains_aggregator.get_chain_manager(blockchain)
+                try:
+                    manager.tokens.detect_tokens(
+                        only_cache=False,
+                        addresses=addresses,
+                    )
```
yabirgb, line 926: fits in one line

rotkehlchen/tasks/manager.py
```diff
@@ -538,6 +541,36 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
...
+                        HistoryEventSubType.RECEIVE,
+                        HistoryEventSubType.REWARD,
+                        HistoryEventSubType.REMOVE_ASSET,
+                        HistoryEventSubType.MEV_REWARD,
+                    ]]
+                    last_save_time = self.database.get_last_balance_save_time(cursor)
+                    for chain in EVM_CHAIN_IDS_WITH_TRANSACTIONS:
+                        if len(location_labels := cursor.execute(  # number of earliest receive events of distinct assets after last_save_time  # noqa: E501
+                            'SELECT DISTINCT location_label FROM '
+                            '(SELECT MIN(timestamp), location_label FROM history_events '
```
yabirgb, line 560: I was thinking that if we have many addresses this task can be really slow. Not sure if it is worth to split the job per chain but it complicates the logic

yabirgb (commented):
relevant test failed

rotkehlchen/tasks/manager.py
OjusWiZard: also we query different (ether)scan APIs for each chain

rotkehlchen/tasks/manager.py
OjusWiZard: we filter the location using `Location.from_chain_id(chain).serialize_for_db()` where chain is from `EVM_CHAIN_IDS_WITH_TRANSACTIONS`, so we should only get the evm events

rotkehlchen/rotkehlchen.py
```diff
@@ -910,6 +916,14 @@ def query_balances(
             save_despite_errors=save_despite_errors,
         )
 
+        if detect_tokens is not None:
+            for blockchain, addresses in detect_tokens.items():
+                manager: EvmManager = self.chains_aggregator.get_chain_manager(blockchain)
+                try:
+                    manager.tokens.detect_tokens(only_cache=False, addresses=addresses)
+                except (RemoteError, BadFunctionCallOutput) as e:
+                    log.error(f'Failed to detect tokens: {e}')
```
yabirgb, line 925: can you provide here the chain and the addresses quried? ALso mention that the detection for that chain is skipped

rotkehlchen/db/settings.py
```diff
@@ -61,6 +61,7 @@
 DEFAULT_AUTO_DELETE_CALENDAR_ENTRIES = True
 DEFAULT_AUTO_CREATE_CALENDAR_REMINDERS = True
 DEFAULT_ASK_USER_UPON_SIZE_DISCREPANCY = True
+DEFAULT_AUTO_DETECT_TOKENS = True
```
LefterisJP, line 64: This should be false.

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`5103` rotki will now automatically detect tokens along with their balances.
```
```suggestion
* :feature:`5103` rotki will now automatically detect tokens along with their balances if the setting has been turned on
```

rotkehlchen/tasks/manager.py
```diff
@@ -538,6 +541,35 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
         """
         with self.database.conn.read_ctx() as cursor:
             if self.database.should_save_balances(cursor):
+                detect_tokens: dict[SupportedBlockchain, list[ChecksumEvmAddress]] = defaultdict(list)  # noqa: E501
+                if CachedSettings().get_settings().auto_detect_tokens is True:
+                    type_filter = [event_type.serialize() for event_type in [
```
LefterisJP, line 546: shouldn't you make this smarter? You can never guess all possible combinations and you are 100% definitely missing events here. You should use the direction attribute of the event. IN/OUT/NEUTRAL.

rotkehlchen/tasks/manager.py
```diff
@@ -538,6 +541,35 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
...
+                            f'subtype IN ({",".join(["?"] * len(subtype_filter))})) '
+                            'AND location = ? AND timestamp >= ? GROUP BY asset);',
+                            type_filter + subtype_filter + [
+                                Location.from_chain_id(chain).serialize_for_db(),
+                                last_save_time * 1000,  # to ms
+                            ],
+                        ).fetchall()) > 0:
+                            detect_tokens[chain.to_blockchain()] = [
+                                string_to_evm_address(location_label[0])
+                                for location_label in location_labels
```
LefterisJP, line 570: also what if the location label is not there?

Btw question. This "detect tokens" is different to the force detect tokens we run when we do balance checks?

rotkehlchen/tasks/manager.py
```diff
@@ -549,6 +581,7 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
                     save_despite_errors=False,
                     timestamp=None,
                     ignore_cache=True,
+                    detect_tokens=detect_tokens if len(detect_tokens) else None,
```
LefterisJP, line 584: Hmm I think this is silly. You are using the events and not really utilizing the info you get and then you do the extremely heavy token detection task. This is wrong.

LefterisJP (commented):
Okay so this approach is not good.

The detect tokens task is **very very** heavy. Do you understand what it does? It goes through **all the tokens** of a chain for a single address and checks their balances with many many multicalls.

The more tokens we add for a chain the slower it gets.

To give you an idea of how heavy and how slow it now is for mainnet. For my 20 addresses it takes about 15 mins. It's insanely slow. You don't want this to run by default for all addresses.

 What I see you are doing:

1. Set the new detect tokens setting to true by default. For reference of what damage this does look at what I said for me. 20 accounts, 15 mins.
2. Check some random receive events and find only the addresses they may touch
3. Then for these addresses do a token detection, but ... FOR ALL TOKENS IN EXISTENCE.

See the problem?

You actually bother to compare and then do the silly thing.


You need to understand that the app is slow as fuck. We want to make it smarter. I need you guys to think a bit more and work on improving the performance. If this went in, we would be killing performance.

1. What you should actually do is check for all token receive events with the direction attribute (IN/OUT/NEUTRAL).
2. For all the IN you should just add it to the detected tokens for the address if they are not there yet
3. (optional -- perhaps other PR or future issue when balances are saved in DB) For the OUT ones if the amount going out is same as the amount of the balance of the address in the token, remove the token from the detected tokens.

Under no circumstances should you do the entire heavy token detection. This is what we do to detect all potential balances but it's getting slower and slower and it should **never ever** happen automatically.

rotkehlchen/db/settings.py
OjusWiZard: it should be efficient enough now so keeping the default to true (https://discord.com/channels/657906918408585217/1080783409640648765/1248392708624613429)

rotkehlchen/tasks/manager.py
```diff
@@ -536,8 +546,50 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
...
+                        'GROUP BY asset) WHERE timestamp >= ?;',
+                        (last_save_time * 1000,),  # to ms
+                    ):
+                        event = EvmEvent.deserialize_from_db(event_data[2:])
+                        if (
+                            event.location not in EVM_EVMLIKE_LOCATIONS or
+                            event.location_label is None or
+                            EVM_ADDRESS_REGEX.fullmatch(event.location_label) is None
+                        ):
+                            log.warning(
```
LefterisJP, line 567: I am not sure I understand the warning here. It's a big if condition and can match for many events right? Doesn't the query above protect you from it? If yes so is this more like an assertion of something that should not happen? I am a but confused here.

Otherwise if this warning is going to be happening often then it's gonna spam the logs.

But from what I see it's more like an assertion?

rotkehlchen/tasks/manager.py
```diff
@@ -536,8 +546,50 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
         Update the balances of a user if the difference between last time they were updated
         and the current time exceeds the `balance_save_frequency`.
         """
-        with self.database.conn.read_ctx() as cursor:
-            if self.database.should_save_balances(cursor):
+        with self.database.conn.write_ctx() as write_cursor:
+            if self.database.should_save_balances(write_cursor):
+                if CachedSettings().get_settings().auto_detect_tokens is True:
```
LefterisJP, line 551: Can you maybe add a small comment here describing the currrent algorithm for token detection here?

rotkehlchen/tasks/manager.py
OjusWiZard: yeah, it's more like an assertion that if we look for evm events and their location label is empty or not a valid evm address, then we log this warning.

LefterisJP (approved):
LGTM

PR #8041, opened 2024-06-10
fix: remove notification tooltip

lukicenturi:
Not sure if this entirely fix the notification issue, but at least I got some terrible experience with tooltip there (clicking the tooltip will close the navigation drawer), and we don't need anyway, since the action button is pretty straightforward


kelsos (approved):
lgtm, thanks @lukicenturi 

Do we need to add a changelog entry for this?

lukicenturi:
> lgtm, thanks @lukicenturi
> 
> Do we need to add a changelog entry for this?

I've never seen any user reported this, so should be fine

PR #8046, opened 2024-06-10
feat(api): history event export with selected currency

### LINKED ISSUE #7722: History Events CSV Export does not respect profit currency in event value
## Problem Definition

When exporting a CSV file of filtered historical events, the balance value is always in `USD` even if your Profit Currency has been set as something else ex: `SGD`.

## Logs

N.A

### System Description

Operating system: Linux
Rotki version: v1.32.2

moxw:
Closes #7722


rotkehlchen/api/rest.py
```diff
@@ -4675,15 +4678,35 @@ def export_history_events(
...
+                    )
+                except (PriceQueryUnsupportedAsset, RemoteError):
+                    price = ZERO_PRICE
+                except NoPriceForGivenTimestamp as e:
+                    # In the case of NoPriceForGivenTimestamp when we got rate limited we let
+                    # it propagate so the user can take action after the report is made
+                    if e.rate_limited is True:
+                        raise
+                    price = ZERO_PRICE
+                event.balance.usd_value = event.balance.amount * price
```
yabirgb, line 4708: this looks wrong. here is not the usd price but the price in the main currency. THis is unexpected and can lead to errors

rotkehlchen/api/rest.py
```diff
@@ -4675,15 +4678,35 @@ def export_history_events(
...
 
+        with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
+            settings = self.rotkehlchen.get_settings(cursor)
+            currency = settings.main_currency.resolve_to_asset_with_oracles()
+
         serialized_history_events = []
         headers: dict[str, None] = {}
         for event in history_events:
-            serialized_event = event.serialize_for_csv()
+            if event.balance.usd_value != ZERO and currency != A_USD:
```
yabirgb, line 4693: The approach here looks wrong because you are trying to edit a field that means usd value to mean something else

yabirgb: THis is somehow related to a different issue that we have to get rid of usd values and handle them differently but is a different topic

rotkehlchen/api/rest.py
```diff
@@ -4675,15 +4678,35 @@ def export_history_events(
...
+                try:
+                    price = self.get_rate_in_profit_currency(
+                        asset=event.asset,
+                        timestamp=ts_ms_to_sec(event.timestamp),
+                        currency=currency,
+                    )
```
yabirgb, line 4699: I'm thinking that this approach is good because you ask the different oracles for the prices in currency instead of querying the fiat rates. If this is the reason I would add a comment for the reader

rotkehlchen/api/rest.py
```diff
@@ -4675,15 +4678,35 @@ def export_history_events(
...
+                except (PriceQueryUnsupportedAsset, RemoteError):
+                    price = ZERO_PRICE
+                except NoPriceForGivenTimestamp as e:
+                    # In the case of NoPriceForGivenTimestamp when we got rate limited we let
+                    # it propagate so the user can take action after the report is made
+                    if e.rate_limited is True:
+                        raise
+                    price = ZERO_PRICE
+                event.balance.usd_value = event.balance.amount * price
+            serialized_event = event.serialize_for_csv(currency)
```
yabirgb, line 4709: instead of the approach of modifying the event.balance data what I would do is to provice an argument for this function `price_in_native_currency` and what you do is querying the pricce in the native currency if it is != USD and provide it as argument.

Since you don't have the `price_in_usd` you can set `price_in_native_currency` with default to None and check 

```
if price_in_native_currency is not None:
     fiat_value = price_in_native_currency * event.balance.amount
else:
     fiat_value = event.balance.value
```

does it make sense? Also properly document it

rotkehlchen/history/events/structures/base.py
```diff
@@ -253,7 +253,7 @@ def serialize_for_csv(self) -> dict[str, Any]:
             new_dict[key] = value
             if key == 'asset':  # input the amount/usd value right after the asset
                 new_dict['amount'] = balance['amount']
-                new_dict['usd_value'] = balance['usd_value']
+                new_dict[f'{currency.symbol_or_name().lower()}_value'] = balance['usd_value']
```
yabirgb, line 256: don't do it since the frontend will need to handle specific cases and doesn't help. Call it `fiat_value`. Having variable keys is worst. Also this change needs an update in the api docs

rotkehlchen/api/v1/schemas.py
```diff
@@ -3374,9 +3374,10 @@ def make_staking_performance_query(
 
 class SkippedExternalEventsExportSchema(Schema):
     directory_path = DirectoryField(load_default=None)
+    async_query = fields.Boolean(load_default=False)
```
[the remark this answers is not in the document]
moxw: this is actually not required at all, probably mistakenly pasted, removed.

rotkehlchen/api/rest.py
```diff
@@ -4723,9 +4764,27 @@ def export_history_events(
...
+        timestamp from the price oracle
+        - RemoteError if there is a problem reaching the price oracle server
+        or with reading the response returned by the server
+        """
+        rate = PriceHistorian.query_historical_price(
+            from_asset=asset,
+            to_asset=currency,
+            timestamp=timestamp,
+        )
+        return rate
```
yabirgb, line 4787: why did you keep it? We said about removing it. Now is just a wrapper around `query_historical_price`

rotkehlchen/history/events/structures/base.py
```diff
@@ -244,16 +245,16 @@ def serialize(self) -> dict[str, Any]:
...
-    def serialize_for_csv(self) -> dict[str, Any]:
+    def serialize_for_csv(self, fiat_value: FVal) -> dict[str, Any]:
         """Serialize event data for CSV export"""
         entry = self.serialize()
         balance = entry.pop('balance')
         new_dict = {}
         for key, value in entry.items():
             new_dict[key] = value
-            if key == 'asset':  # input the amount/usd value right after the asset
+            if key == 'asset':  # input the amount/fiat value right after the asset
```
yabirgb, line 255: mention here that we don't use the usd_value here because we sport the value in the user selected currency

rotkehlchen/tests/api/test_history_events_export.py
```diff
@@ -113,6 +113,9 @@ def test_history_export_download_csv(
     # now query the download CSV endpoint
     response = requests.put(
         api_url_for(rotkehlchen_api_server_with_exchanges, 'exporthistoryeventresource'),
+        json={
+            'async_query': False,
+        },
```
yabirgb, line 120:
```suggestion
        json={'async_query': False},
```

rotkehlchen/api/rest.py
```diff
@@ -4675,15 +4680,50 @@ def export_history_events(
...
 
+        with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
+            settings = self.rotkehlchen.get_settings(cursor)
+            currency = settings.main_currency.resolve_to_asset_with_oracles()
+
         serialized_history_events = []
         headers: dict[str, None] = {}
         for event in history_events:
-            serialized_event = event.serialize_for_csv()
+            if event.balance.usd_value != ZERO and currency != A_USD:
```
yabirgb, line 4695: I'm not sure about the condition here. What happens if the event has `usd_value == ZERO and currency == A_USD`. In that case you won't enter this section and the price will be exported as 0 right?

rotkehlchen/api/rest.py
```diff
@@ -4675,15 +4680,50 @@ def export_history_events(
...
+                    if price != ZERO_PRICE:
+                        fiat_value = event.balance.amount * price
+                    else:
+                        # if the price is zero, we query the fiat rate for currency against USD
+                        price = self.get_rate_in_profit_currency(
+                            asset=A_USD,
+                            timestamp=ts_ms_to_sec(event.timestamp),
+                            currency=currency,
+                        )
+                        fiat_value = event.balance.usd_value * price
```
yabirgb, line 4713: this doesn't make sense. If you are exporting in the native currency you don't care about USD. Having usd mixed with the data is worst than having a 0

rotkehlchen/api/rest.py
```diff
@@ -4675,15 +4680,50 @@ def export_history_events(
...
+                    price = ZERO_PRICE
+                except NoPriceForGivenTimestamp as e:
+                    # In the case of NoPriceForGivenTimestamp when we got rate limited we let
+                    # it propagate so the user can take action after the report is made
+                    if e.rate_limited is True:
+                        raise
+                    price = ZERO_PRICE
+            else:
+                # if the asset is USD or the value is zero, we don't need to ask for the rate
+                fiat_value = event.balance.usd_value
```
yabirgb, line 4724: exactly, if the value is zero you need to query it. By default is zero. What you don't need to query is if the amount is zero, but value zero means that the price was not queried yet

rotkehlchen/history/price.py
```diff
@@ -167,7 +167,7 @@ def query_historical_price(
             from_asset: The ticker symbol of the asset for which we want to know
                         the price.
             to_asset: The ticker symbol of the asset against which we want to
-                      know the price.
+                        know the price.
             timestamp: The timestamp at which to query the price
```
yabirgb, line 171: why this change?

rotkehlchen/api/rest.py
```diff
@@ -4675,15 +4680,50 @@ def export_history_events(
...
+                    else:
+                        # if the price is zero, we query the fiat rate for currency against USD
+                        price = self.get_rate_in_profit_currency(
+                            asset=A_USD,
+                            timestamp=ts_ms_to_sec(event.timestamp),
+                            currency=currency,
+                        )
+                        fiat_value = event.balance.usd_value * price
+                except (PriceQueryUnsupportedAsset, RemoteError):
+                    price = ZERO_PRICE
```
yabirgb, line 4715: I believe is not price what you want here but `fiat_value`.

what you can do is define price here and do the finally in the try_except and there do 

```
                        fiat_value = event.balance.usd_value * price
```

yabirgb (commented):
Some comments. I don't see it correct

rotkehlchen/history/price.py
moxw: aligning with spaces of 4, it had a block of two spaces, was showing red on vscode.

rotkehlchen/api/rest.py
moxw: right, i had the wrong assumption that `usd_value` is always fetched with the event.

yabirgb:
For the frontend changes luki said to merge them and he will adjust what remains

rotkehlchen/api/rest.py
```diff
@@ -4675,15 +4680,48 @@ def export_history_events(
...
+        if currency == A_USD:  # if the currency is not USD don't loose time querying prices that won't be used  # noqa: E501
+            entries = dbevents.get_base_entries_missing_prices(EvmEventFilterQuery.make())
+            query_missing_prices_of_base_entries(
+                database=self.rotkehlchen.data.db,
+                entries_missing_prices=entries,
+            )
+
         for event in history_events:
-            serialized_event = event.serialize_for_csv()
+            if currency != A_USD:
```
LefterisJP, line 4702: keep it simple. Remove the above query_missing_prices_of_base_entries and if it's A_USD and `usd_value` is zero also continue with the query here

docs/api.rst
```diff
@@ -5298,6 +5298,7 @@ Exporting History Events
    :statuscode 400: Provided JSON is in some way malformed
    :statuscode 409: No user is logged in or failure at event export.
    :statuscode 500: Internal rotki error
+   :statuscode 502: Couldn't fetch prices for all the events due to being rate limited.
```
LefterisJP, line 5301: Shouldnt't the async change also be mentioned in the docs?

moxw: added

LefterisJP (approved):
LGTM. Frontend will adjust some things so ignoring failed frontend test.

PR #8050, opened 2024-06-11
Fix duplicate entries in globaldb conflicts

yabirgb:
When updating assets it can happen that two remote updates by mistake add the same token and maybe with extra data. Before this change the multiple conflicts were being returned causing errors in the logic for the api consumer since it was assumed uniqness.

This PR ensures that only the latest appearence is took as conflict

Closes #(issue_number)

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,6 +3,7 @@ Changelog
 =========
 
 * :bug:`-` Fix the issue where the popup notification breaks the experience when the notification sidebar is already open.
+* :bug:`-` Fixed an issue with the assets updates where conflicts couldn't be resolved for multiple changes in a single asset.
 
 * :release:`1.33.1 <2024-05-29>`
 * :bug:`-` Fix the issue where airdrops aren't properly filtered by status.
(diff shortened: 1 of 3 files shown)


docs/changelog.rst
```diff
@@ -3,6 +3,7 @@ Changelog
 =========
 
 * :bug:`-` Fix the issue where the popup notification breaks the experience when the notification sidebar is already open.
+* :bug:`-` Fixed an issue assets updates where conflicts couldn't be resolved.
```
LefterisJP, line 6:
```suggestion
* :bug:`-` Fixed an issue with the assets updates where conflicts couldn't be resolved for multiple changes in a single asset.
```

rotkehlchen/globaldb/updates.py
```diff
@@ -485,7 +485,7 @@ def _handle_asset_update(
                 serialized=False,
                 specific_ids=[local_asset.identifier],
             )[0]
-            self.conflicts.append((local_data, remote_asset_data))
+            self.conflicts[local_asset.identifier] = (local_data, remote_asset_data)
```
LefterisJP, line 488:
```suggestion
            # always take the last one, if there is multiple conflicts for a single asset
            self.conflicts[local_asset.identifier] = (local_data, remote_asset_data)
```

LefterisJP (approved):
LGTM

PR #8054, opened 2024-06-11
Decode older stakedao bribes

LefterisJP:
Stakedao bribes older than January 2023 were not decoded properly since it was a different contract and event format.

This PR fixes that and adds a test


LefterisJP:
all green. Merging.

PR #8055, opened 2024-06-11
Feat (Curve): Add curve for other chains

### LINKED ISSUE #7431: Support curve in other EVM chains
## Abstract

Curve is deployed in almost (or all) the evm chains that we support. We should be able to decode events in those chains too

## Specification

We need to adjust the logic to query pools and gauges in the other chains too. Also update the cache keys to take into consideration the evm chain where they are being queried. If we don't track addresses in one chain we shouldn't query pools/gauges for those chains.

OjusWiZard:
Closes #7431

## Checklist

- [x] Add curve for other chains
- [x] Update meta registry, zap, and exchange addresses and logic 

Note: `last_queried_ts` is 0 in SQL diff, so the values will be refreshed when the decoder is loaded in prod

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/curve/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/curve/decoder.py
@@ -1,16 +1,12 @@
-import logging
 from typing import TYPE_CHECKING, Any
 
-from rotkehlchen.chain.ethereum.modules.curve.constants import (
-    AAVE_POOLS,
-    CRV_ADDRESS,
-    CURVE_DEPOSIT_CONTRACTS,
-    CURVE_SWAP_ROUTER,
-    GAUGE_BRIBE_V2,
-    GAUGE_CONTROLLER,
-)
+from rotkehlchen.accounting.structures.balance import Balance
 from rotkehlchen.chain.evm.decoding.constants import ERC20_OR_ERC721_TRANSFER
-from rotkehlchen.chain.evm.decoding.curve.constants import CPT_CURVE
+from rotkehlchen.chain.evm.decoding.curve.constants import (
+    CPT_CURVE,
+    CURVE_SWAP_ROUTER_NG,
+    GAUGE_VOTE,
+)
 from rotkehlchen.chain.evm.decoding.curve.decoder import CurveCommonDecoder
 from rotkehlchen.chain.evm.decoding.structures import (
     DEFAULT_DECODING_OUTPUT,
@@ -21,17 +17,23 @@ from rotkehlchen.chain.evm.decoding.structures import (
 from rotkehlchen.constants.assets import A_CRV, A_ETH
 from rotkehlchen.history.events.structures.evm_event import EvmProduct
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
-from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.types import ChecksumEvmAddress
-from rotkehlchen.utils.misc import hex_or_bytes_to_address
+from rotkehlchen.utils.misc import hex_or_bytes_to_address, hex_or_bytes_to_int
+
+from .constants import (
+    AAVE_POOLS,
+    CRV_ADDRESS,
+    CURVE_DEPOSIT_CONTRACTS,
+    CURVE_SWAP_ROUTER,
+    DEPOSIT_AND_STAKE_ZAP,
+    GAUGE_BRIBE_V2,
+    GAUGE_CONTROLLER,
+)
 
 if TYPE_CHECKING:
     from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
     from rotkehlchen.chain.evm.decoding.base import BaseDecoderTools
     from rotkehlchen.user_messages import MessagesAggregator
-
-logger = logging.getLogger(__name__)
-log = RotkehlchenLogsAdapter(logger)
 
 
 class CurveDecoder(CurveCommonDecoder):
@@ -48,9 +50,8 @@ class CurveDecoder(CurveCommonDecoder):
             msg_aggregator=msg_aggregator,
             native_currency=A_ETH,
             aave_pools=AAVE_POOLS,
-            curve_deposit_contracts=CURVE_DEPOSIT_CONTRACTS,
-            curve_swap_router=CURVE_SWAP_ROUTER,
-            gauge_controller=GAUGE_CONTROLLER,
+            curve_deposit_contracts=CURVE_DEPOSIT_CONTRACTS | {DEPOSIT_AND_STAKE_ZAP},
+            curve_swap_routers={CURVE_SWAP_ROUTER, CURVE_SWAP_ROUTER_NG},
         )
 
     def _decode_gauge_bribe(self, context: DecoderContext) -> DecodingOutput:
@@ -74,6 +75,37 @@ class CurveDecoder(CurveCommonDecoder):
 
         return DEFAULT_DECODING_OUTPUT
 
+    def _decode_curve_gauge_votes(self, context: DecoderContext) -> DecodingOutput:
+        if context.tx_log.topics[0] != GAUGE_VOTE:
+            return DEFAULT_DECODING_OUTPUT
+
+        if not self.base.is_tracked(user_address := hex_or_bytes_to_address(context.tx_log.data[32:64])):  # noqa: E501
+            return DEFAULT_DECODING_OUTPUT
+
+        user_note = '' if user_address == context.transaction.from_address else f' from {user_address}'  # noqa: E501
+        gauge_address = hex_or_bytes_to_address(context.tx_log.data[64:96])
+        vote_weight = hex_or_bytes_to_int(context.tx_log.data[96:128])
+        if vote_weight == 0:
+            verb = 'Reset vote'
+            weight_note = ''
+        else:
+            verb = 'Vote'
+            weight_note = f' with {vote_weight / 100}% voting power'
+        event = self.base.make_event_from_transaction(
+            transaction=context.transaction,
+            tx_log=context.tx_log,
+            event_type=HistoryEventType.INFORMATIONAL,
+            event_subtype=HistoryEventSubType.GOVERNANCE,
+            asset=self.native_currency,
+            balance=Balance(),
+            location_label=context.transaction.from_address,
+            notes=f'{verb}{user_note} for {gauge_address} curve gauge{weight_note}',
+            address=gauge_address,
+            counterparty=CPT_CURVE,
+            product=EvmProduct.GAUGE,
+        )
+        return DecodingOutput(event=event, refresh_balances=False)
+
     # -- DecoderInterface methods
     @staticmethod
     def possible_products() -> dict[str, list[EvmProduct]]:
@@ -83,5 +115,6 @@ class CurveDecoder(CurveCommonDecoder):
 
     def addresses_to_decoders(self) -> dict[ChecksumEvmAddress, tuple[Any, ...]]:
         return super().addresses_to_decoders() | {
+            GAUGE_CONTROLLER: (self._decode_curve_gauge_votes,),
             CRV_ADDRESS: (self._decode_gauge_bribe,),
         }
--- a/rotkehlchen/chain/evm/decoding/curve/constants.py
+++ b/rotkehlchen/chain/evm/decoding/curve/constants.py
@@ -9,6 +9,7 @@ ADD_LIQUIDITY_EVENTS: Final = {
     b'B?d\x95\xa0\x8f\xc6RB\\\xf4\xed\r\x1f\x9e7\xe5q\xd9\xb9R\x9b\x1c\x1c#\xcc\xe7\x80\xb2\xe7\xdf\r',  # ADD_LIQUIDITY  # noqa: E501
     b'&\xf5Z\x85\x08\x1d$\x97N\x85\xc6\xc0\x00E\xd0\xf0E9\x91\xe9Xs\xf5+\xff\r!\xaf@y\xa7h',  # ADD_LIQUIDITY_2_ASSETS  # noqa: E501
     b'?\x19\x15w^\x0c\x9a8\xa5z{\xb7\xf1\xf9\x00_Ho\xb9\x04\xe1\xf8J\xa2\x156MVs\x19\xa5\x8d',  # ADD_LIQUIDITY_4_ASSETS  # noqa: E501
+    b'\x18\x9cb;fk\x1bE\xb8=qx\xf3\x9b\x8c\x08|\xb0\x97t1|\xa2\xf5<-<7&\xf2"\xa2',  # AddLiquidity in DepositAndStake Zap  # noqa: E501
 }
 REMOVE_LIQUIDITY_IMBALANCE: Final = {
     b'\xb9d\xb7/s\xf5\xef[\xf0\xfd\xc5Y\xb2\xfa\xb9\xa7\xb1*9\xe4x\x17\xa5G\xf1\xf0\xae\xe4\x7f\xeb\xd6\x02',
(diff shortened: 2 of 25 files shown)


rotkehlchen/chain/evm/decoding/curve/decoder.py
```diff
@@ -725,6 +720,6 @@ def post_decoding_rules(self) -> dict[str, list[tuple[int, Callable]]]:
     def enricher_rules(self) -> list[Callable]:
         return [self._maybe_enrich_curve_transfers]
 
-    @staticmethod
+    @ staticmethod
```
yabirgb, line 723:
```suggestion
    @staticmethod
```

rotkehlchen/chain/evm/decoding/curve/decoder.py
```diff
@@ -559,6 +540,21 @@ def _handle_post_decoding(
                 )
                 break  # We only need to handle zap contracts once
 
+        for event in decoded_events:
+            if (
+                event.address == ZERO_ADDRESS and
+                event.event_subtype == HistoryEventSubType.NONE and
+                event.asset.is_evm_token() and
+                (event_asset := event.asset.resolve_to_evm_token()).evm_address in self.gauges
```
yabirgb, line 548: have you checked the case for eth on this?

rotkehlchen/chain/evm/decoding/curve/decoder.py
```diff
@@ -352,39 +350,49 @@ def _decode_curve_trades(self, context: DecoderContext) -> DecodingOutput:
...
-                bought_token_address = self.pools[pool_address][bought_token_id]
-        else:  # EXCHANGE_MULTIPLE
-            swapping_contract = self.curve_swap_router
+        swapping_contracts: set[ChecksumEvmAddress] = set()
+        for tx_log in context.all_logs:
+            if tx_log.topics[0] in (TOKEN_EXCHANGE, TOKEN_EXCHANGE_UNDERLYING, TOKEN_EXCHANGE_NG):
+                pool_address = tx_log.address
+                swapping_contracts = {pool_address}
+                # When a single pool is used, spender and receiver is always the same
+                spender_address = receiver_address = hex_or_bytes_to_address(tx_log.topics[1])
```
yabirgb, line 359: so this is fixing swapping when there are more than 1 pool in the route?

rotkehlchen/chain/evm/decoding/curve/decoder.py
```diff
@@ -352,39 +350,49 @@ def _decode_curve_trades(self, context: DecoderContext) -> DecodingOutput:
...
+            swapping_contracts = self.curve_swap_routers
             spender_address = hex_or_bytes_to_address(context.tx_log.topics[1])
             receiver_address = hex_or_bytes_to_address(context.tx_log.topics[2])
-            raw_sold_amount = hex_or_bytes_to_int(context.tx_log.data[-64:-32])
+            if raw_sold_amount is None:  # if it's not already set in token exchange event
+                raw_sold_amount = hex_or_bytes_to_int(context.tx_log.data[-64:-32])
+
             raw_bought_amount = hex_or_bytes_to_int(context.tx_log.data[-32:])
+            route_length = 11 if context.tx_log.topics[0] == EXCHANGE_NG else 9
+
```
yabirgb, line 384: is this documented in the curve docs? If so link them

rotkehlchen/chain/evm/decoding/curve/decoder.py
```diff
@@ -352,39 +350,49 @@ def _decode_curve_trades(self, context: DecoderContext) -> DecodingOutput:
...
-                pool_address in self.pools and
-                len(self.pools[pool_address]) > max(sold_token_id, bought_token_id)  # Make sure that tokens of the pool are cached  # noqa: E501
-            ):
-                sold_token_address = self.pools[pool_address][sold_token_id]
-                bought_token_address = self.pools[pool_address][bought_token_id]
-        else:  # EXCHANGE_MULTIPLE
-            swapping_contract = self.curve_swap_router
+        swapping_contracts: set[ChecksumEvmAddress] = set()
+        for tx_log in context.all_logs:
+            if tx_log.topics[0] in (TOKEN_EXCHANGE, TOKEN_EXCHANGE_UNDERLYING, TOKEN_EXCHANGE_NG):
```
yabirgb, line 355: what are you doing here and what is the break condition?

OjusWiZard: So, here we check if before using the router any pool is used separately, like in this tx https://basescan.org/tx/0x6fee2438337aefe297a69d4361ff4d743865ef7fce6637e9e3e544af0c19184f. If that's the case then we extract the sold token address and amount here. As soon as we find the info, we break out of the loop.

rotkehlchen/chain/evm/decoding/curve/decoder.py
OjusWiZard: When there is a some pool used outside of the route.

rotkehlchen/chain/evm/decoding/curve/decoder.py
OjusWiZard: We look for the gauge token here so it cannot be a native currency

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -34,7 +46,47 @@ def __init__(
...
+            curve_deposit_contracts=CURVE_DEPOSIT_CONTRACTS | {DEPOSIT_AND_STAKE_ZAP},
+            curve_swap_routers={CURVE_SWAP_ROUTER, CURVE_SWAP_ROUTER_NG},
         )
+
+    def _decode_curve_gauge_votes(self, context: DecoderContext) -> DecodingOutput:
+        if context.tx_log.topics[0] != GAUGE_VOTE:
+            return DEFAULT_DECODING_OUTPUT
+
+        user_address = hex_or_bytes_to_address(context.tx_log.data[32:64])
+        if not self.base.is_tracked(user_address):
```
LefterisJP, line 58: I know you just moved it but let's improve

```suggestion        
        if not self.base.is_tracked(user_address := hex_or_bytes_to_address(context.tx_log.data[32:64])):
```

rotkehlchen/chain/ethereum/modules/curve/decoder.py
```diff
@@ -34,7 +46,47 @@ def __init__(
...
+            return DEFAULT_DECODING_OUTPUT
+
+        user_note = ''
+        if user_address != context.transaction.from_address:
+            user_note = f' from {user_address}'
```
LefterisJP, line 63:
```suggestion
        user_note = '' if user_address == context.transaction.from_address else f' from {user_address}'
```

rotkehlchen/chain/evm/decoding/curve/constants.py
```diff
@@ -27,26 +28,29 @@
 GAUGE_VOTE: Final = b'E\xca\x9aL\x8d\x01\x19\xeb2\x9eX\r(\xfeh\x9eHN\x1b\xe20\xda\x807\xad\xe9T}-%\xcc\x91'  # noqa: E501
 TOKEN_EXCHANGE: Final = b'\x8b>\x96\xf2\xb8\x89\xfaw\x1cS\xc9\x81\xb4\r\xaf\x00_c\xf67\xf1\x86\x9fppR\xd1Z=\xd9q@'  # noqa: E501
 TOKEN_EXCHANGE_UNDERLYING: Final = b'\xd0\x13\xca#\xe7ze\x00<,e\x9cTB\xc0\x0c\x80Sq\xb7\xfc\x1e\xbdL lA\xd1Sk\xd9\x0b'  # noqa: E501
+TOKEN_EXCHANGE_NG: Final = b'\x14?\x1f\x8e\x86\x1f\xbd\xed\xdd[F\xe8D\xb7\xd3\xac{\x86\xa1"\xf3n\x8cF8Y\xeeh\x11\xb1\xf2\x9c'  # noqa: E501
```
LefterisJP, line 31: What are these 2? What is `NG`? CAn you add a comment explaining?

rotkehlchen/chain/evm/decoding/curve/decoder.py
```diff
@@ -352,39 +350,52 @@ def _decode_curve_trades(self, context: DecoderContext) -> DecodingOutput:
...
+                    pool_address in self.pools and
+                    len(self.pools[pool_address]) > max(sold_token_id, bought_token_id)  # Make sure that tokens of the pool are cached  # noqa: E501
+                ):
+                    sold_token_address = self.pools[pool_address][sold_token_id]
+                    bought_token_address = self.pools[pool_address][bought_token_id]
+                break  # As soon as we find the info, we break out of the loop.
+        else:  # only the router is used
+            raw_sold_amount = None
+
+        if context.tx_log.topics[0] in (EXCHANGE_MULTIPLE, EXCHANGE_NG):
```
LefterisJP, line 377: Add a comment also on what case this is

rotkehlchen/chain/evm/decoding/curve/decoder.py
```diff
@@ -559,6 +543,21 @@ def _handle_post_decoding(
                 )
                 break  # We only need to handle zap contracts once
 
+        for event in decoded_events:
```
LefterisJP, line 546: Couldn't this post decoding rule be switched to action items and thus be more performant?

rotkehlchen/tests/api/test_caching.py
```diff
@@ -152,7 +152,7 @@ def test_general_cache_refresh(rotkehlchen_api_server: 'APIServer'):
         ))
         assert_proper_response(response)
         assert patched_convex_query.call_count == 1, 'Convex pools should have been queried despite should_update_protocol_cache being False'  # noqa: E501
-        assert patched_curve_query.call_count == 1, 'Curve pools should have been queried despite should_update_protocol_cache being False'  # noqa: E501
+        assert patched_curve_query.call_count == 5, 'Curve pools should have been queried despite should_update_protocol_cache being False'  # noqa: E501
```
LefterisJP, line 155: hmm this test queries everything? Isn't it slow?

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -1725,3 +1735,307 @@ def test_no_zap_event(ethereum_transaction_decoder, ethereum_accounts):
...
+@pytest.mark.parametrize('polygon_pos_accounts', [['0x1c362DFE864a4c4b3311eC97bf0b8320CB0a4952']])
+def test_curve_deposit_polygon(database, polygon_pos_inquirer, polygon_pos_accounts):
+    tx_hash = deserialize_evm_tx_hash('0x6cb9d7ceb55a1063c17b58cb643e699525ca6037e711c34283cf0f3d6e81716e')  # noqa: E501
+    events, _ = get_decoded_events_of_transaction(
+        evm_inquirer=polygon_pos_inquirer,
+        database=database,
+        tx_hash=tx_hash,
+    )
+    timestamp, pool_address = TimestampMS(1718015936000), string_to_evm_address('0x445FE580eF8d70FF569aB36e80c647af338db351')  # noqa: E501
+    deposit_amount, approve_amount, received_amount, gas_fees = '2.12', '0.88', '1.87538744644899407', '0.0127938000149261'  # noqa: E501
```
LefterisJP, line 1750: Combine all in same line. Already need noqa for each line so just combine them. Same in all other tests

LefterisJP (commented):
LEft some comments. Also this needs a changelog entry

rotkehlchen/tests/api/test_caching.py
OjusWiZard: it's fast because all the queries are mocked and rest is VCRed

rotkehlchen/chain/evm/decoding/curve/constants.py
OjusWiZard: NG means new generation, we include these new contracts' topics to decode the events done through new contracts

rotkehlchen/chain/evm/decoding/curve/decoder.py
```diff
@@ -300,31 +332,64 @@ def _decode_curve_deposit_events(
...
+
+        if received_asset is None:  # find it through the logs
+            deposit_addresses = {event.address for event in deposit_events} | {ZERO_ADDRESS}
+            for _log in all_logs[all_logs.index(tx_log) + 1:]:
+                if (
+                    (_log.topics[0] == ERC20_OR_ERC721_TRANSFER) and
+                    (hex_or_bytes_to_address(_log.topics[1]) in deposit_addresses) and
+                    self.base.is_tracked(hex_or_bytes_to_address(_log.topics[2]))
+                ):
+                    received_asset = Asset(evm_address_to_identifier(
```
LefterisJP, line 345: you should use `self.base.get_or_create_evm_token()`. Never initialize a token like this from external data

rotkehlchen/chain/evm/decoding/curve/decoder.py
```diff
@@ -300,31 +332,64 @@ def _decode_curve_deposit_events(
                 event.counterparty == CPT_CURVE
             ):
                 receive_event = event
+                received_asset = event.asset
+
+        if received_asset is None:  # find it through the logs
+            deposit_addresses = {event.address for event in deposit_events} | {ZERO_ADDRESS}
+            for _log in all_logs[all_logs.index(tx_log) + 1:]:
+                if (
+                    (_log.topics[0] == ERC20_OR_ERC721_TRANSFER) and
```
LefterisJP, line 341: shouldn't there be any asset address + amount check here?

rotkehlchen/chain/evm/decoding/curve/decoder.py
```diff
@@ -352,39 +417,53 @@ def _decode_curve_trades(self, context: DecoderContext) -> DecodingOutput:
...
-                context.tx_log.topics[0] == TOKEN_EXCHANGE and
-                pool_address in self.pools and
-                len(self.pools[pool_address]) > max(sold_token_id, bought_token_id)  # Make sure that tokens of the pool are cached  # noqa: E501
-            ):
-                sold_token_address = self.pools[pool_address][sold_token_id]
-                bought_token_address = self.pools[pool_address][bought_token_id]
-        else:  # EXCHANGE_MULTIPLE
-            swapping_contract = self.curve_swap_router
+        swapping_contracts: set[ChecksumEvmAddress] = set()
+        # we check if before using the router any pool is used separately.
```
LefterisJP, line 421: IS this comment still valid? Why are you iterating all logs here? This makes no sense.

This function is called for 

```python
        if context.tx_log.topics[0] in (
            TOKEN_EXCHANGE,
            TOKEN_EXCHANGE_UNDERLYING,
            TOKEN_EXCHANGE_NG,
            EXCHANGE_MULTIPLE,
            EXCHANGE_NG,
        ):
```

And then in here you reiterate all logs and then have a similar (subset) if condition?



Also does the function's docstring still make sense? It seems to describe something different? Can you make sure this code make sense?

rotkehlchen/chain/evm/decoding/curve/decoder.py
```diff
@@ -505,161 +555,61 @@ def _decode_gauge_events(self, context: DecoderContext) -> DecodingOutput:
...
                 event.product = EvmProduct.GAUGE
                 found_event_modifying_balances = True
                 if context.tx_log.topics[0] == GAUGE_DEPOSIT:
                     event.event_type = HistoryEventType.DEPOSIT
                     event.event_subtype = HistoryEventSubType.DEPOSIT_ASSET
-                    event.notes = f'Deposit {event.balance.amount} {crypto_asset.symbol} into {gauge_address} curve gauge'  # noqa: E501
+                    event.notes = f'Deposit {event.balance.amount} {evm_asset.symbol} into {gauge_address} curve gauge'  # noqa: E501
+                    from_event_type = HistoryEventType.RECEIVE
+                    pair_subtype = HistoryEventSubType.RECEIVE_WRAPPED
+                    pair_note = 'Receive {{amount}} {symbol} after depositing in {address} curve gauge'  # noqa: E501
```
LefterisJP, line 587: why double {} here for amount?.

edit: Okay I see. Hmm this is weird. Think if you can make it a bit better / easier to read.

At least do add a comment on why the double braces here and below (I know why but need to explain)

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -941,15 +944,15 @@ def _maybe_decode_erc20_721_transfer(
...
 
-                if action_item.paired_event_data is not None:
+                if action_item.paired_events_data is not None:
                     # If there is a paired event to this, take care of the order
-                    out_event = transfer
-                    in_event = action_item.paired_event_data[0]
-                    if action_item.paired_event_data[1] is True:
-                        out_event = action_item.paired_event_data[0]
-                        in_event = transfer
+                    out_events: Sequence[EvmEvent] = [transfer]
```
LefterisJP, line 949: can't there also be multiple out events for some other type of action item?

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -1725,3 +1738,319 @@ def test_no_zap_event(ethereum_transaction_decoder, ethereum_accounts):
...
+            tx_hash=tx_hash,
+            sequence_index=2484,
+            timestamp=timestamp,
+            location=Location.POLYGON_POS,
+            event_type=HistoryEventType.DEPOSIT,
+            event_subtype=HistoryEventSubType.DEPOSIT_ASSET,
+            asset=A_POLYGON_POS_USDT,
+            balance=Balance(amount=FVal(deposit_amount)),
+            location_label=polygon_pos_accounts[0],
+            notes=f'Deposit {deposit_amount} USDT in curve pool {pool_address}',
```
LefterisJP, line 1802: So this test shows that the order is wrong. Should be OUT first, IN afterwards.

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -1725,3 +1738,319 @@ def test_no_zap_event(ethereum_transaction_decoder, ethereum_accounts):
...
+            tx_hash=tx_hash,
+            sequence_index=1106,
+            timestamp=timestamp,
+            location=Location.GNOSIS,
+            event_type=HistoryEventType.SPEND,
+            event_subtype=HistoryEventSubType.RETURN_WRAPPED,
+            asset=Asset(f'eip155:100/erc20:{gauge_address}'),
+            balance=Balance(amount=FVal(withdraw_amount)),
+            location_label=gnosis_accounts[0],
+            notes=f'Return {withdraw_amount} crvUSDsDAI-gauge after withdrawing from {gauge_address} curve gauge',  # noqa: E501
```
LefterisJP, line 1928: Same with this test. Return is the out event. Should come before.

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -1725,3 +1738,319 @@ def test_no_zap_event(ethereum_transaction_decoder, ethereum_accounts):
...
+            tx_hash=tx_hash,
+            sequence_index=10,
+            timestamp=timestamp,
+            location=Location.ARBITRUM_ONE,
+            event_type=HistoryEventType.DEPOSIT,
+            event_subtype=HistoryEventSubType.DEPOSIT_ASSET,
+            asset=arbitrum_usdt,
+            balance=Balance(amount=FVal(deposit_amount)),
+            location_label=arbitrum_one_accounts[0],
+            notes=f'Deposit {deposit_amount} USDT in curve pool {pool_address}',
```
LefterisJP, line 2051: same here. Deposit is the out event. Should be first.

rotkehlchen/chain/evm/decoding/curve/decoder.py
```diff
@@ -505,161 +571,63 @@ def _decode_gauge_events(self, context: DecoderContext) -> DecodingOutput:
...
-            f'Expected to see a return pool token event and withdrawal events '
-            f'for a curve pool, but have not found them. '
-            f'Tx_hash: {transaction.tx_hash.hex()}',
+                    event.notes = f'Withdraw {event.balance.amount} {evm_asset.symbol} from {gauge_address} curve gauge'  # noqa: E501
+                    from_event_type = HistoryEventType.SPEND
+                    pair_subtype = HistoryEventSubType.RETURN_WRAPPED
+                    pair_note = 'Return {{amount}} {symbol} after withdrawing from {address} curve gauge'  # noqa: E501
+
+        # `pair_note` is formatted twice, first below with `symbol` and `address`. `amount` is left
+        # due to double {}. Then later in the action item `amount` is replaces by its value.
```
LefterisJP, line 612:
```suggestion
        # due to double {}. Then later in the action item `amount` is replaced by its value.
```

LefterisJP (approved):
LGTM

PR #8058, opened 2024-06-11
feat(evm): support Gearbox protocol in price queries

### LINKED ISSUE #7544: Support Gearbox protocol
## Abstract

We should support the Gearbox protocol.

We can start with passive earn pools and detect them.

https://app.gearbox.fi/pools

## Specification

- Properly decode all gearbox related transactions in the history events
- Detect gearbox pools and automatically find balances for them.

moxw:
Closes #7544

### DIFF
--- a/rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
+++ b/rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
@@ -12,7 +12,7 @@ from rotkehlchen.chain.evm.decoding.gearbox.constants import (
 )
 from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.constants.misc import ONE, ZERO
-from rotkehlchen.errors.misc import NotERC20Conformant, RemoteError
+from rotkehlchen.errors.misc import BlockchainQueryError, NotERC20Conformant, RemoteError
 from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.globaldb.cache import (
     globaldb_get_general_cache_keys_and_values_like,
@@ -197,6 +197,56 @@ def ensure_gearbox_tokens_existence(
         pool.lp_tokens = lp_tokens_from_db
         verified_pools.append(pool)
     return verified_pools
+
+
+def ensure_gearbox_lp_underlying_tokens(
+        node_inquirer: 'EvmNodeInquirer',
+        token_identifier: str,
+        lp_contract: 'EvmContract',
+) -> EvmToken | None:
+    """Ensures that the underlying token of a Gearbox liquidity pool (LP) is stored in the global
+    database. If the underlying token is not already saved, it queries the chain to retrieve
+    and save it."""
+    try:
+        remote_underlying_token = lp_contract.call(node_inquirer, 'underlyingToken')
+    except (RemoteError, BlockchainQueryError) as e:
+        log.error(f'Failed to query underlyingToken method in {node_inquirer.chain_name} Gearbox pool {lp_contract.address}. {e!s}')  # noqa: E501
+        return None
+
+    try:
+        underlying_token_address = deserialize_evm_address(remote_underlying_token)
+    except DeserializationError:
+        log.error(f'underlyingToken call of {node_inquirer.chain_name} {lp_contract.address} returned invalid address {remote_underlying_token}')  # noqa: E501
+        return None
+
+    try:  # make sure it's in the global DB
+        underlying_token = get_or_create_evm_token(
+            userdb=node_inquirer.database,
+            evm_address=underlying_token_address,
+            chain_id=node_inquirer.chain_id,
+            encounter=TokenEncounterInfo(
+                description='Detecting Gearbox pool underlying tokens',
+            ),
+        )
+    except NotERC20Conformant as e:
+        log.error(f'Error fetching {node_inquirer.chain_name} token {underlying_token_address} while detecting underlying tokens of {lp_contract.address!s}: {e!s}')  # noqa: E501
+        return None
+
+    # store it in the DB, so next time no need to query chain
+    with GlobalDBHandler().conn.write_ctx() as write_cursor:
+        GlobalDBHandler._add_underlying_tokens(
+            write_cursor=write_cursor,
+            parent_token_identifier=token_identifier,
+            underlying_tokens=[
+                UnderlyingToken(
+                    address=underlying_token_address,
+                    token_kind=EvmTokenKind.ERC20,
+                    weight=ONE,  # all gearbox vaults have single underlying
+                ),
+            ],
+            chain_id=node_inquirer.chain_id,
+        )
+    return underlying_token
 
 
 def get_gearbox_pool_tokens(inquirer: 'EvmNodeInquirer', pool_data: list[str], underlying_token_address: ChecksumEvmAddress) -> set[ChecksumEvmAddress] | None:  # noqa: E501
--- a/rotkehlchen/inquirer.py
+++ b/rotkehlchen/inquirer.py
@@ -11,6 +11,10 @@ from rotkehlchen.chain.ethereum.defi.price import handle_defi_price_query
 from rotkehlchen.chain.ethereum.utils import token_normalized_value_decimals
 from rotkehlchen.chain.evm.constants import ETH_SPECIAL_ADDRESS
 from rotkehlchen.chain.evm.contracts import EvmContract
+from rotkehlchen.chain.evm.decoding.gearbox.gearbox_cache import (
+    ensure_gearbox_lp_underlying_tokens,
+    read_gearbox_data_from_cache,
+)
 from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.chain.evm.utils import lp_price_from_uniswaplike_pool_contract
 from rotkehlchen.constants import ONE, ZERO
@@ -60,7 +64,7 @@ from rotkehlchen.constants.assets import (
 )
 from rotkehlchen.constants.misc import CURRENCYCONVERTER_API_KEY
 from rotkehlchen.constants.prices import ZERO_PRICE
-from rotkehlchen.constants.resolver import ethaddress_to_identifier
+from rotkehlchen.constants.resolver import ethaddress_to_identifier, evm_address_to_identifier
 from rotkehlchen.constants.timing import DAY_IN_SECONDS, MONTH_IN_SECONDS
 from rotkehlchen.errors.asset import UnknownAsset, WrongAssetType
 from rotkehlchen.errors.defi import DefiPoolError
@@ -87,6 +91,7 @@ from rotkehlchen.oracles.structures import CurrentPriceOracle
 from rotkehlchen.serialization.deserialize import deserialize_evm_address
 from rotkehlchen.types import (
     CURVE_POOL_PROTOCOL,
+    GEARBOX_PROTOCOL,
     LP_TOKEN_AS_POOL_PROTOCOLS,
     YEARN_VAULTS_V2_PROTOCOL,
     CacheType,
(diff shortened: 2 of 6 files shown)


rotkehlchen/inquirer.py
```diff
@@ -920,6 +924,112 @@ def find_curve_pool_price(
...
+            log.error(f'Error fetching {node_inquirer.chain_name} token while detecting underlying tokens of {token.evm_address!s}')  # noqa: E501
+            return None
+
+        underlying_token_price = self.find_usd_price(underlying_token)
+        try:
+            price_per_share = lp_contract.call(
+                node_inquirer=node_inquirer,
+                method_name='convertToAssets',
+                arguments=[1 * 10 ** (decimals := token.get_decimals())],
+            )
```
yabirgb, line 1025:
```suggestion
                arguments=[10 ** (decimals := token.get_decimals())],
            )
```

rotkehlchen/inquirer.py
```diff
@@ -920,6 +924,112 @@ def find_curve_pool_price(
...
+        except (RemoteError, BlockchainQueryError) as e:
+            log.error(f'Failed to query convertToAssets method in {node_inquirer.chain_name} Gearbox Pool. {e!s}')  # noqa: E501
```
yabirgb, line 1027: add the address of the asset that was being queried

rotkehlchen/types.py
```diff
@@ -85,7 +85,7 @@
 AERODROME_POOL_PROTOCOL = 'aerodrome_pool'
 PICKLE_JAR_PROTOCOL = 'pickle_jar'
 SPAM_PROTOCOL = 'spam'
-GEARBOX_POOL_PROTOCOL = 'gearbox_pool'
+GEARBOX_POOL_PROTOCOL = 'gearbox'
```
yabirgb, line 88: why this change? Doesn't it affect the assets in the asset repo?

rotkehlchen/inquirer.py
```diff
@@ -920,6 +924,112 @@ def find_curve_pool_price(
...
+
+        if token in farming_tokens:
+            farming_contract = EvmContract(
+                address=token.evm_address,
+                abi=node_inquirer.contracts.abi('GEARBOX_FARMING_POOL'),
+                deployed_block=0,  # not used here
+            )
+            try:
+                lp_token = farming_contract.call(node_inquirer=node_inquirer, method_name='stakingToken')  # noqa: E501
+                lp_token = deserialize_evm_address(lp_token)
```
yabirgb, line 955: the deserialize can raise `DeserializationError`. Handle it

rotkehlchen/inquirer.py
```diff
@@ -920,6 +924,112 @@ def find_curve_pool_price(
...
+            except (RemoteError, BlockchainQueryError) as e:
+                log.error(f'Failed to query stakingToken method in Gearbox Pool. {e!s}')
```
yabirgb, line 957: inclide the asset in the error messages otherwise we won't know what happeened

rotkehlchen/inquirer.py
```diff
@@ -920,6 +924,112 @@ def find_curve_pool_price(
...
+            address=lp_token,
+            abi=node_inquirer.contracts.abi('GEARBOX_LP'),
+            deployed_block=0,  # not used here
+        )
+        if maybe_underlying_tokens is None or len(maybe_underlying_tokens) != 1:
+            # underlying token not recorded in the DB. Ask the chain
+            try:
+                remote_underlying_token = lp_contract.call(node_inquirer, 'underlyingToken')
+            except (RemoteError, BlockchainQueryError) as e:
+                log.error(f'Failed to query underlying token method in {node_inquirer.chain_name} Gearbox pool. {e!s}')  # noqa: E501
```
yabirgb, line 972: this logic seems to be for the case where the asset doesn't have the underlying token right? Can you move it to its own function in the gearbox evm module? something like `ensure_gearbox_lp_underlying_tokens`. Maybe a better name. This will help clearing the logic for the reader

rotkehlchen/inquirer.py
```diff
@@ -920,6 +924,112 @@ def find_curve_pool_price(
...
+                underlying_token = get_or_create_evm_token(
+                    userdb=node_inquirer.database,
+                    evm_address=underlying_token_address,
+                    chain_id=token.chain_id,
+                    encounter=TokenEncounterInfo(
+                        description='Detecting Gearbox pool underlying tokens',
+                    ),
+                )
+            except NotERC20Conformant as e:
+                log.error(f'Error fetching {node_inquirer.chain_name} token {underlying_token_address} while detecting underlying tokens of {token.evm_address!s}: {e!s}')  # noqa: E501
```
yabirgb, line 991: you are not returning here so the logic will continue and error later because you don't have the underlying token

rotkehlchen/inquirer.py
```diff
@@ -920,6 +924,112 @@ def find_curve_pool_price(
...
+                evm_address_to_identifier(
+                    address=maybe_underlying_tokens[0].address,
+                    chain_id=token.chain_id,
+                    token_type=EvmTokenKind.ERC20,
+                ),
+            )
+
+        if underlying_token is None:
+            log.error(f'Error fetching {node_inquirer.chain_name} token while detecting underlying tokens of {token.evm_address!s}')  # noqa: E501
+            return None
```
yabirgb, line 1017: how can it be none? The logic should have failed before if it wasn't a valid gearbox token before right?

rotkehlchen/tests/unit/test_inquirer.py
```diff
@@ -622,6 +624,27 @@ def test_find_yearn_vaults_v2_price(inquirer_defi, globaldb):
             assert result and result[0].address == underlying_token.resolve_to_evm_token().evm_address  # noqa: E501
 
 
+@pytest.mark.parametrize('use_clean_caching_directory', [True])
+@pytest.mark.parametrize('should_mock_current_price_queries', [False])
+def test_find_gearbox_lp_price(inquirer: 'Inquirer', arbitrum_one_manager: 'ArbitrumOneManager'):
+    dwethv3 = EvmToken('eip155:42161/erc20:0x04419d3509f13054f60d253E0c79491d9E683399')
+    sdwethv3 = EvmToken('eip155:42161/erc20:0x6773fF780Dd38175247795545Ee37adD6ab6139a')
```
yabirgb, line 631: shouldn't this test be vcred since it does external calls? Also since you will have it vcred assert for the right price

rotkehlchen/tests/unit/test_inquirer.py
```diff
@@ -622,6 +624,27 @@ def test_find_yearn_vaults_v2_price(inquirer_defi, globaldb):
...
+            'DELETE from underlying_tokens_list WHERE parent_token_entry=?',
+            (sdwethv3.identifier,),
+        )
+    inquirer.inject_evm_managers([(ChainID.ARBITRUM_ONE, arbitrum_one_manager)])
+    for token, underlying_token in ((sdwethv3, A_WETH_ARB), (dwethv3, A_WETH_ARB)):
+        price = inquirer.find_gearbox_price(token)
+        assert price is not None
+
+        with GlobalDBHandler().conn.read_ctx() as cursor:
+            result = GlobalDBHandler().fetch_underlying_tokens(cursor, token.identifier)
```
yabirgb, line 644:
```suggestion
            result = GlobalDBHandler.fetch_underlying_tokens(cursor, token.identifier)
```

rotkehlchen/types.py
moxw: So this wasn't used anywhere in the code before, but the assets in the repo have been saved as protocol being `gearbox`. So, the change.

rotkehlchen/inquirer.py
moxw: Was getting linting error because I missed the `return None` earlier, this is not needed anymore, removed.

rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
```diff
@@ -199,6 +199,56 @@ def ensure_gearbox_tokens_existence(
...
+    try:
+        remote_underlying_token = lp_contract.call(node_inquirer, 'underlyingToken')
+    except (RemoteError, BlockchainQueryError) as e:
+        log.error(f'Failed to query underlyingToken method in {node_inquirer.chain_name} Gearbox pool {lp_contract.address}. {e!s}')  # noqa: E501
+        return None
+
+    try:
+        underlying_token_address = deserialize_evm_address(remote_underlying_token)
+    except DeserializationError:
+        log.error(f'underlying token call of {lp_contract.address} returned invalid address {remote_underlying_token}')  # noqa: E501
```
LefterisJP, line 219: add chain name in this error

rotkehlchen/inquirer.py
```diff
@@ -920,6 +927,68 @@ def find_curve_pool_price(
 
         return (total_assets_value / total_supply) * (10 ** lp_token.get_decimals())
 
+    def find_gearbox_price(self, token: EvmToken) -> Price | None:
+        evm_manager = self.get_evm_manager(chain_id=token.chain_id)
+        node_inquirer = evm_manager.node_inquirer
```
LefterisJP, line 932: do not have variables used in only one place
```suggestion
        node_inquirer = self.get_evm_manager(chain_id=token.chain_id).node_inquirer
```

rotkehlchen/inquirer.py
```diff
@@ -920,6 +927,68 @@ def find_curve_pool_price(
...
+            maybe_underlying_tokens = GlobalDBHandler.fetch_underlying_tokens(
+                cursor=cursor,
+                parent_token_identifier=token.identifier,
+            )
+        lp_contract = EvmContract(
+            address=lp_token,
+            abi=node_inquirer.contracts.abi('GEARBOX_LP'),
+            deployed_block=0,  # not used here
+        )
+        if maybe_underlying_tokens is None or len(maybe_underlying_tokens) != 1:
```
LefterisJP, line 960: add some comments here on what this nested if is checking. It's not clear to me.

rotkehlchen/inquirer.py
```diff
@@ -920,6 +927,68 @@ def find_curve_pool_price(
...
+            if (
+                underlying_token := ensure_gearbox_lp_underlying_tokens(
+                    token_identifier=token.identifier,
+                    node_inquirer=node_inquirer,
+                    lp_contract=lp_contract,
+                )
+            ) is None:
+                return None
```
LefterisJP, line 968: Isn't this an error? Shouldn't it log one?

rotkehlchen/inquirer.py
```diff
@@ -920,6 +927,68 @@ def find_curve_pool_price(
...
+        else:
```
LefterisJP, line 969:
```suggestion
        else:  # explain what this is in a short phrase
```

rotkehlchen/inquirer.py
```diff
@@ -920,6 +927,68 @@ def find_curve_pool_price(
...
+            underlying_token = EvmToken(
+                evm_address_to_identifier(
+                    address=maybe_underlying_tokens[0].address,
+                    chain_id=token.chain_id,
+                    token_type=EvmTokenKind.ERC20,
+                ),
+            )
+
+        underlying_token_price = self.find_usd_price(underlying_token)
```
LefterisJP, line 978: This is only used in the else part ofthe try/except/else. Put it there and don't create a new variable for it.

Also ... can it raise anything? Is it handled?

rotkehlchen/types.py
LefterisJP, line 88:
```suggestion
GEARBOX_PROTOCOL = 'gearbox'
```

LefterisJP (commented):
Some comments.

Also sqldiff is just the one ABI you use right?

moxw:
> Some comments.
> 
> Also sqldiff is just the one ABI you use right?

yes

rotkehlchen/inquirer.py
moxw: added

rotkehlchen/inquirer.py
moxw: From the docs, other usages of this function and looking at the code, it looks like the errors are handled in the function itself.  It instead `ZERO_PRICE` when it has exhausted all the options, so adding a error log if price == ZERO_PRICE.

LefterisJP (approved):
LGTM

PR #8059, opened 2024-06-11
Compound finance COMP claim when transfer after event

LefterisJP:
Make sure decoder works for Compound finance COMP claim when the ERC20 transfer event comes after the claim event.

Part of https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=66951521

There is one more case with no event, not handled: https://etherscan.io/tx/0xc50cd2f79e33e0bfbb429d73a00c3745aba5116f05d86abd48e8532aa4c60d0d

But that is a bit trickier and seems to be a recent thing.


LefterisJP:
all is green, all is fine

PR #8060, opened 2024-06-11
Add paladin bribes decoding

LefterisJP:


LefterisJP:
All green but frontend should not have anything to do with the PR. For some reason it fails and the frontend devs should figure it out

PR #8068, opened 2024-06-12
HOP governance and some other fixes

LefterisJP:
- HOP governance events should now be decoded properly
- Remove unnecessary constants file
- Rename hop counterparty and add DB upgrade for it


LefterisJP:
All green. License CLA has bugged. I am the founder so lol.

PR #8077, opened 2024-06-13
fix: various issues because of vuetify removal

lukicenturi:
Closes #(issue_number)

- [x] Container not centered
- [x] Navigation menu may be invisible when toggled
- [x] Fonts looks weird (bolder than usual)


kelsos (approved):
lgtm thanks :)

PR #8078, opened 2024-06-13
Decode more ENS events

LefterisJP:
What prompted this is seeing some transactions for my ENS with only gas burned. So added two new events:

- Transfer subnode ownership for a node to a new owner.
- Address changed for a name

Unfortunately this required changes in many of the tests :sweat_smile: 

Also made the graph query fail earlier if the API key is malformed. cc @yabirgb

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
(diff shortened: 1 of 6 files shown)


LefterisJP:
Finally. All green. This took so much fixing the tests ...

PR #8081, opened 2024-06-13
feat(hop): add Hop ARB reward contract

moxw:


LefterisJP (approved):
There is an issue with the HOP implementation but the PR looks fine. So I will fix it in another PR.

PR #8083, opened 2024-06-13
Fix token detection on scroll

yabirgb:
The balance scanner contract in scroll is different from the one in other chains and it doesn't work with some tokens as already reported. In scroll we saw it with ascrusdc


LefterisJP (approved):
LGTM

PR #8087, opened 2024-06-14
Decode 1inch v4 transactions using OrderFilledRFQ

LefterisJP:


rotkehlchen/tests/unit/decoders/test_1inch.py
```diff
@@ -297,6 +297,60 @@ def test_1inchv4_swap_on_uniswapv3(database, ethereum_inquirer):
     assert expected_events == events
 
 
+@pytest.mark.vcr()
```
yabirgb, line 300: you didn't add the api key filter here

```suggestion
@pytest.mark.vcr(filter_query_parameters=['apikey'])
```

I don't recall where is required but I belive the test will run fine

LefterisJP: it will run fine

LefterisJP: or maybe not :rofl:

LefterisJP: hmmmm it's unrelated. I think I did not push the cassettes PR ffs

LefterisJP:
all green now

PR #8100, opened 2024-06-16
Curve vote escrow events

LefterisJP:
Curve vote escrow locking and withdrawal events will now be properly decoded on ethereum mainnet.


LefterisJP:
all green

PR #8114, opened 2024-06-17
Feat (Curve): Detect prices of LP and Gauge tokens

OjusWiZard:
## Checklist

- [x] Detect price of LP and Gauge tokens of Curve on all supported chains

### DIFF
--- a/rotkehlchen/chain/evm/node_inquirer.py
+++ b/rotkehlchen/chain/evm/node_inquirer.py
@@ -173,7 +173,7 @@ def _query_web3_get_logs(
     return events
 
 
-class EvmNodeInquirer(ABC):
+class EvmNodeInquirer(ABC, LockableQueryMixIn):
     """Class containing generic functionality for querying evm nodes
 
     The child class must implement the following methods:
@@ -226,6 +226,7 @@ class EvmNodeInquirer(ABC):
         self.contract_info_erc20_cache: LRUCacheWithRemove[ChecksumEvmAddress, dict[str, Any]] = LRUCacheWithRemove(maxsize=1024)  # noqa: E501
         self.contract_info_erc721_cache: LRUCacheWithRemove[ChecksumEvmAddress, dict[str, Any]] = LRUCacheWithRemove(maxsize=512)  # noqa: E501
         self.maybe_connect_to_nodes(when_tracked_accounts=True)
+        LockableQueryMixIn.__init__(self)
 
     def maybe_connect_to_nodes(self, when_tracked_accounts: bool) -> None:
         """Start async connect to the saved nodes for the given evm chain if needed.
@@ -1396,52 +1397,8 @@ class EvmNodeInquirer(ABC):
     def _get_pruned_check_tx_hash(self) -> EVMTxHash:
         """Returns a transaction hash that can used for checking whether a node is pruned."""
 
-    def _additional_receipt_processing(self, tx_receipt: dict[str, Any]) -> None:  # noqa: B027
+    def _additional_receipt_processing(self, tx_receipt: dict[str, Any]) -> None:
         """Performs additional tx_receipt processing where necessary"""
-
-
-class EvmNodeInquirerWithDSProxy(EvmNodeInquirer):
-    def __init__(
-            self,
-            greenlet_manager: GreenletManager,
-            database: 'DBHandler',
-            etherscan: Etherscan,
-            blockchain: SUPPORTED_EVM_CHAINS_TYPE,
-            etherscan_node: WeightedNode,
-            etherscan_node_name: str,
-            contracts: EvmContracts,
-            contract_scan: 'EvmContract',
-            contract_multicall: 'EvmContract',
-            dsproxy_registry: 'EvmContract',
-            native_token: CryptoAsset,
-            rpc_timeout: int = DEFAULT_EVM_RPC_TIMEOUT,
-    ) -> None:
-        super().__init__(
-            greenlet_manager=greenlet_manager,
-            database=database,
-            etherscan=etherscan,
-            blockchain=blockchain,
-            etherscan_node=etherscan_node,
-            etherscan_node_name=etherscan_node_name,
-            contracts=contracts,
-            contract_scan=contract_scan,
-            contract_multicall=contract_multicall,
-            rpc_timeout=rpc_timeout,
-            native_token=native_token,
-        )
-        self.proxies_inquirer = EvmProxiesInquirer(
-            node_inquirer=self,
-            dsproxy_registry=dsproxy_registry,
-        )
-
-
-class UpdatableCacheDataMixin(LockableQueryMixIn):
-    def __init__(
-            self,
-            database: 'DBHandler',
-    ) -> None:
-        super().__init__()
-        self.database = database
 
     @protect_with_lock()
     def ensure_cache_data_is_updated(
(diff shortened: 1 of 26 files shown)


rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -173,7 +173,59 @@ def _query_web3_get_logs(
...
+                'database': self.database,
+                'new_data': new_data,
+            }
+            if chain_id is not None:
+                save_method_kwargs['chain_id'] = chain_id
+            save_method(**save_method_kwargs)
+        return True
+
+
+class EvmNodeInquirer(ABC, UpdatableCacheDataMixin):
```
LefterisJP, line 228: if you do this then:

1. Is it used **anywhere** else? If not then why not just make it part of evm node inquirer itself?
2. Everywhere else that UpdatableCacheDataMixin is used should be removed. Atm every chain Inquirer inherits from it. This should be undone. Grep for uses.

rotkehlchen/rotkehlchen.py
```diff
@@ -426,6 +426,10 @@ def unlock_user(
         Inquirer().inject_evm_managers([
             (ChainID.ETHEREUM, ethereum_manager),
             (ChainID.OPTIMISM, optimism_manager),
+            (ChainID.POLYGON_POS, polygon_pos_manager),
+            (ChainID.ARBITRUM_ONE, arbitrum_one_manager),
+            (ChainID.BASE, base_manager),
+            (ChainID.GNOSIS, gnosis_manager),
```
LefterisJP, line 432: What about scroll? Should it be all here? If yes then we need this to be an iteration of the `EVM_CHAINS_WITH_TRANSACTIONS`

rotkehlchen/tests/unit/test_inquirer.py
```diff
@@ -422,35 +429,79 @@ def test_find_velodrome_v2_lp_token_price(inquirer, optimism_manager):
...
+    with GlobalDBHandler().conn.write_ctx() as write_cursor:  # querying curve lp token price normally triggers curve cache query. Set all query ts to now, so it does not happen.  # noqa: E501
+        write_cursor.execute('UPDATE general_cache SET last_queried_ts=? WHERE key LIKE ?', (ts_now(), 'CURVE_LP_TOKENS%'))  # noqa: E501
+
+    inquirer.inject_evm_managers([
+        (ethereum_manager.node_inquirer.chain_id, ethereum_manager),
+        (optimism_manager.node_inquirer.chain_id, optimism_manager),
+        (polygon_pos_manager.node_inquirer.chain_id, polygon_pos_manager),
+        (arbitrum_one_manager.node_inquirer.chain_id, arbitrum_one_manager),
+        (base_manager.node_inquirer.chain_id, base_manager),
+        (gnosis_manager.node_inquirer.chain_id, gnosis_manager),
```
LefterisJP, line 497: same logic missing here

LefterisJP (commented):
There is some test failures. Also left some comments.

rotkehlchen/inquirer.py
```diff
@@ -344,10 +351,25 @@ def inject_evm_managers(evm_managers: Sequence[tuple[ChainID, 'EvmManager']]) ->
...
+    @overload
+    @staticmethod
+    def get_evm_manager(
+            chain_id: ChainID,
+            for_curve: bool,
+    ) -> 'ArbitrumOneManager | BaseManager | EthereumManager | GnosisManager | OptimismManager | PolygonPOSManager':  # noqa: E501
+        ...
+
+    @staticmethod
+    def get_evm_manager(chain_id: ChainID, for_curve: bool = False) -> 'EvmManager':
```
LefterisJP, line 368: What are you doing here lol. Come on. Adding an argument for type cheking? This is not how typing works. You would need to add different argument values for chain id and proper overloading. Please see how it's done in the codebase

rotkehlchen/inquirer.py
```diff
@@ -344,10 +351,25 @@ def inject_evm_managers(evm_managers: Sequence[tuple[ChainID, 'EvmManager']]) ->
...
         evm_manager = Inquirer._evm_managers.get(chain_id)
         assert evm_manager is not None, f'evm manager for chain id {chain_id} should have been injected'  # noqa: E501
+        if for_curve:
```
LefterisJP, line 371: Absolutely not. This is not how typing works. Remove

rotkehlchen/tests/unit/test_evm_misc.py
```diff
@@ -55,19 +54,3 @@ def test_weth_is_supported():
         set(CHAIN_ID_TO_WETH_MAPPING.keys()) ==
         set(typing.get_args(SUPPORTED_CHAIN_IDS)) - CHAINS_WITHOUT_NATIVE_ETH
     )
-
-
-def test_evminquirer_has_cachable_mixin(blockchain):
```
LefterisJP, line 60: Why is this removed? Has the problem been solved? You moved the logic to the manager and not the inquirer? So this is not needed?

OjusWiZard: ah no, I thought it's testing something else

OjusWiZard: so this was testing for `ensure_cache_data_is_updated`. But now thats the part of base EvmInquirer class. So, I'm changing it to check the `CurveManagerMixin` now

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -173,7 +173,7 @@ def _query_web3_get_logs(
     return events
 
 
-class EvmNodeInquirer(ABC):
+class EvmNodeInquirer(ABC, LockableQueryMixIn):
```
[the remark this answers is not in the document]
OjusWiZard: `UpdatableCacheDataMixin` is removed now right? and its function `ensure_cache_data_is_updated` is now moved to this class.
as per https://github.com/rotki/rotki/pull/8114#discussion_r1643514586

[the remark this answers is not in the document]
OjusWiZard: Using `LockableQuery` because `ensure_cache_data_is_updated` use `@protect_with_lock` that needs the attributes defined in this mixin

rotkehlchen/inquirer.py
```diff
@@ -819,8 +834,13 @@ def find_curve_pool_price(
         May raise:
         - RemoteError
         """
-        ethereum = self.get_evm_manager(chain_id=lp_token.chain_id)
-        ethereum.assure_curve_cache_is_queried_and_decoder_updated()  # type:ignore  # ethereum is an EthereumManager here
+        assert lp_token.chain_id in CURVE_CHAIN_IDS, f'{lp_token} is not on a curve support chain'
```
LefterisJP, line 837:
```suggestion
        assert lp_token.chain_id in CURVE_CHAIN_IDS, f'{lp_token} is not on a curve supported chain'
```

rotkehlchen/inquirer.py
```diff
@@ -819,8 +834,13 @@ def find_curve_pool_price(
...
+        evm_manager = self.get_evm_manager(chain_id=lp_token.chain_id)
+        assert isinstance(evm_manager, CurveManagerMixin), f'{lp_token} is not on a curve support chain'  # noqa: E501
```
LefterisJP, line 839: Why do you need this? Doesn't it understand chain id is of type Curve chain id? Force it somehow above and please avoid this assertion

rotkehlchen/tests/unit/test_evm_misc.py
```diff
@@ -57,17 +61,17 @@ def test_weth_is_supported():
     )
 
 
-def test_evminquirer_has_cachable_mixin(blockchain):
+def test_evmmanager_has_cachable_mixin(blockchain: 'ChainsAggregator') -> None:
```
LefterisJP, line 64: if typing works fine then this test is not needed

rotkehlchen/inquirer.py
```diff
@@ -818,14 +832,20 @@ def find_curve_pool_price(
 
         May raise:
         - RemoteError
+        - AssertionError if used for a token whose chain_id is not in CURVE_CHAIN_IDS
```
LefterisJP, line 835: AssertionError is never raised in production. `assert` statements are going off
```suggestion
```

rotkehlchen/inquirer.py
```diff
@@ -818,14 +832,20 @@ def find_curve_pool_price(
...
         """
-        ethereum = self.get_evm_manager(chain_id=lp_token.chain_id)
-        ethereum.assure_curve_cache_is_queried_and_decoder_updated()  # type:ignore  # ethereum is an EthereumManager here
+        assert lp_token.chain_id in CURVE_CHAIN_IDS, f'{lp_token} is not on a curve support chain'
```
LefterisJP, line 837: you did not fix the typo here
```suggestion
        assert lp_token.chain_id in CURVE_CHAIN_IDS, f'{lp_token} is not on a curve supported chain'
```

OjusWiZard: oh I did this and then had to reset everything because many things were messed while I tried to fix the typing, and then forgot to do this again

LefterisJP (approved):
LGTM

PR #8122, opened 2024-06-18
Fix cowswap bug with other DEX

LefterisJP:
Fix a cowswap bug where detection did not happen if another DEX had decoded the event in a half-finished state first.

https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=67816340


LefterisJP:
all green. Merging

PR #8127, opened 2024-06-19
feat(htx): Add huobi exchange mappings

moxw:


rotkehlchen/exchanges/htx.py
```diff
@@ -93,7 +93,9 @@ def _sign_request(self, url: str, params: dict[str, Any]) -> dict[str, Any]:
 
     def _query(self, absolute_path: str) -> Any:
         url = f'https://api.huobi.pro{absolute_path}'
-        signed_payload = self._sign_request(url=url, params={})
+        signed_payload = {}
```
LefterisJP, line 96: Since this endpoint is not used in production you should not add it here. As you are adding an extra if condition for every  huobi remote call the app will ever do that is for nothing.

Instead in `test_assets_are_known` mock the `_sign_request()` function to be a noop returning `{}`. This would have the same effect and not add production logic for a test

LefterisJP (commented):
I can't really check the global DB diff but left one comment for the rest.

LefterisJP (approved):
LGTM

rkkwns5pxw-gif (commented):
I am removing all verifications and need all funds and coin in my new wallets address please help me

PR #8140, opened 2024-06-21
Fix balances for validators

yabirgb:
When removing a validator the balances in the api weren't being updated because the cache wasn't being cleared

Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/aggregator.py
+++ b/rotkehlchen/chain/aggregator.py
@@ -1393,23 +1393,18 @@ class ChainsAggregator(CacheableMixIn, LockableQueryMixIn):
             ownership_proportion=ownership_proportion,
         )
         self.flush_eth2_cache()
-        self.flush_cache('query_balances')
-        self.flush_cache('query_balances', blockchain=SupportedBlockchain.ETHEREUM_BEACONCHAIN)
-        self.flush_cache('query_balances', blockchain=None, ignore_cache=False)
-        self.flush_cache('query_balances', blockchain=None, ignore_cache=True)
-        self.flush_cache('query_balances', blockchain=SupportedBlockchain.ETHEREUM_BEACONCHAIN, ignore_cache=False)  # noqa: E501
-        self.flush_cache('query_balances', blockchain=SupportedBlockchain.ETHEREUM_BEACONCHAIN, ignore_cache=True)  # noqa: E501
 
     def delete_eth2_validators(self, validator_indices: list[int]) -> None:
         """May raise:
         - ModuleInactive if eth2 module is not activated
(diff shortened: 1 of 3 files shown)


rotkehlchen/chain/aggregator.py
```diff
@@ -1713,6 +1707,11 @@ def flush_eth2_cache(self) -> None:
         self.flush_cache('refresh_eth2_get_daily_stats')
         self.flush_cache('get_eth2_daily_stats')
         self.flush_cache('query_eth2_balances')
+        self.flush_cache('query_balances')
```
LefterisJP, line 1710: Can you also change the docstring and write why you are all flushing also all balances cache?

Since this is anything that is touched by eth2 changes and needs to be invalidated.

You did not move this:`self.flush_cache('query_balances', blockchain=SupportedBlockchain.ETHEREUM_BEACONCHAIN)`. Why?

rotkehlchen/chain/aggregator.py
```diff
@@ -1393,23 +1393,17 @@ def add_eth2_validator(
...
         """May raise:
         - ModuleInactive if eth2 module is not activated
         - InputError if the validator is not found in the DB
         """
-        self.flush_eth2_cache()
         eth2 = self.get_module('eth2')
         if eth2 is None:
             raise ModuleInactive('Cant delete eth2 validator since eth2 module is not active')
-        return DBEth2(self.database).delete_validators(validator_indices)
+        self.flush_eth2_cache()
```
LefterisJP, line 1405: This is indeed better but still wrong. The cache should be flushed only if the validator/s got succesfully deleted. So it shuld either:

1. Just go after deletion.
2. PErhaps change delete_validators to also return success or fail? But looking at its implementation it will always succeed if the validator exists? If there if a problem it will raise. You see about it

rotkehlchen/tests/api/test_eth2.py
```diff
@@ -1249,3 +1249,42 @@ def test_get_validators(rotkehlchen_api_server):
...
+
+    response = requests.delete(  # delete validator
+        api_url_for(
+            rotkehlchen_api_server,
+            'eth2validatorsresource',
+        ), json={'validators': [10000]},
+    )
+    assert_simple_ok_response(response)
+
+    response = requests.get(api_url_for(  # query again balances
```
LefterisJP, line 1284:
```suggestion
    response = requests.get(api_url_for(  # query balances again
```

LefterisJP (commented):
Some questions/comments

LefterisJP (approved):
LGTM

PR #8141, opened 2024-06-21
feat(tools): location_mappings support for packageddb

moxw:
The `main.py` script now fetches the latest location asset mappings and applies them using the `RotkiDataUpdater`.


yabirgb (commented):
The approach is not correct, please check the link that I provide in the comment for how to do it

PR #8142, opened 2024-06-21
Add setting to not query daily stats

### LINKED ISSUE #8120: Replace beaconcha.in daily stats scraping with API
## Abstract

We are currently taking the ETH2 validator daily stats by scraping since at the time there was no way to get it via API: https://github.com/rotki/rotki/blob/cb8549028040d7457a7aa5fcbf2e2ec2ddcec3a3/rotkehlchen/chain/ethereum/eth2_utils.py#L50

Now there is an API that we can use: https://beaconcha.in/api/v1/docs/index.html#/Validator/get_api_v1_validator_stats__index_

## Task

- [x] Create a setting if the user wants the daly stats queried. Since this would considerably increase rate limiting depending on amount of validators it needs to be configurable
- [x] Replace the scraping logic with the API

yabirgb:
Closes #8120

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -654,6 +654,7 @@ Getting or modifying settings
    :resjson int oracle_penalty_threshold_count: The number of failures after which an oracle is penalized. Default is 5.
    :resjson int oracle_penalty_duration: The duration in seconds for which an oracle is penalized. Default is 1800.
    :resjson bool auto_create_calendar_reminders: A boolean denoting whether reminders are created automatically for calendar entries based on the decoded history events. Default is ``true``.
+   :resjson bool query_daily_stats_from_beaconchain: A boolean denoting whether we are allowed to query beaconchain for daily stats. Default is ``true``.
 
    :statuscode 200: Querying of settings was successful
    :statuscode 409: There is no logged in user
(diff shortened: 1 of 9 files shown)


LefterisJP (commented):
You did what we discussed! 

But looking at this I got a bit confused and torn if this is the right approach. What do we want to achieve with this PR?

1. To have a way to not query daily stats automatically if we want to. We can decide whether this is gonna be default or not, but that's another question.
2. To be able to query them manually if we want to.

As far as I know daily stats do not get populated by any of the long running tasks right?
So seeing that our current function has the "only_cache" argument already ... isn't this setting redundant for the backend?

Couldn't we just make this a frontend only setting? The frontend would query with `only_cache: True` at the page load. And then depending on if the setting is on it would also do an only_cache: False` to trigger a remote query only if this setting is on.

And if the user ever wanted to manually refresh the daily stats, irrespective of the setting value they should press the "refresh" button on daily stats and have them refresh.

-------------

That's on the setting.

Now I also have a question on the new daily stats API. Does it return data only for completed days? I hope yes. Otherwise we may have problems.

yabirgb:
Closing since we are adding it as a frontend setting

PR #8151, opened 2024-06-24
feat(frontend): lazy loader for account table

lukicenturi:


kelsos (approved):
looks better than having the rows lazy-load, thanks @lukicenturi

PR #8159, opened 2024-06-24
Chore (tests): Unload protocol's cache by default

OjusWiZard:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=68349975

## Checklist

- [x] Add `load_global_caches` fixture parameter
- [x] Update tests to load cache only when needed

### DIFF
--- a/rotkehlchen/tests/fixtures/globaldb.py
+++ b/rotkehlchen/tests/fixtures/globaldb.py
@@ -17,6 +17,7 @@ from rotkehlchen.globaldb.upgrades.manager import UPGRADES_LIST
 from rotkehlchen.globaldb.utils import GLOBAL_DB_VERSION
 from rotkehlchen.history.types import HistoricalPrice, HistoricalPriceOracle
 from rotkehlchen.tests.utils.database import mock_db_schema_sanity_check
+from rotkehlchen.tests.utils.decoders import patch_decoder_reload_data
 from rotkehlchen.tests.utils.globaldb import patch_for_globaldb_upgrade_to
 from rotkehlchen.types import Price, Timestamp
 
@@ -74,6 +75,12 @@ def fixture_remove_global_assets() -> list[str]:
     return []
 
 
+@pytest.fixture(name='load_global_caches')
+def fixture_load_global_caches() -> list[str]:
+    """A list of counterparties of the decoders for which the global cache should be loaded."""
+    return []
+
+
 def create_globaldb(
         data_directory,
         sql_vm_instructions_cb,
@@ -100,6 +107,7 @@ def _initialize_fixture_globaldb(
         run_globaldb_migrations,
         empty_global_addressbook,
         remove_global_assets,
+        load_global_caches,
 ) -> GlobalDBHandler:
     # clean the previous resolver memory cache, as it
     # may have cached results from a discarded database
@@ -132,6 +140,7 @@ def _initialize_fixture_globaldb(
             stack.enter_context(mock_db_schema_sanity_check())
             patch_for_globaldb_upgrade_to(stack, target_globaldb_version)
 
+        stack.enter_context(patch_decoder_reload_data(load_global_caches))
         globaldb = create_globaldb(new_data_dir, sql_vm_instructions_cb)
 
     if empty_global_addressbook is True:
@@ -160,6 +169,7 @@ def fixture_globaldb(
         run_globaldb_migrations,
         empty_global_addressbook,
         remove_global_assets,
+        load_global_caches,
 ):
     globaldb = _initialize_fixture_globaldb(
         custom_globaldb=custom_globaldb,
@@ -172,6 +182,7 @@ def fixture_globaldb(
         run_globaldb_migrations=run_globaldb_migrations,
         empty_global_addressbook=empty_global_addressbook,
         remove_global_assets=remove_global_assets,
+        load_global_caches=load_global_caches,
     )
     yield globaldb
 
--- a/rotkehlchen/tests/unit/decoders/test_aerodrome.py
+++ b/rotkehlchen/tests/unit/decoders/test_aerodrome.py
@@ -50,14 +50,16 @@ WETH_BASE = Asset(evm_address_to_identifier(
 
 
 @pytest.mark.vcr()
+@pytest.mark.parametrize('load_global_caches', [[CPT_AERODROME]])
 @pytest.mark.parametrize('base_accounts', [['0x514c4BA193c698100DdC998F17F24bDF59c7b6fB']])
-def test_add_liquidity(base_transaction_decoder, base_accounts):
+def test_add_liquidity(base_transaction_decoder, base_accounts, load_global_caches):
     evmhash = deserialize_evm_tx_hash('0xb71a1339c700a110d61655387d422bb982252a3b55de7f571ced3b9f00d9beee')  # noqa: E501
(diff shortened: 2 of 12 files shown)


rotkehlchen/tests/unit/decoders/test_aerodrome.py
```diff
@@ -136,14 +138,16 @@ def test_add_liquidity(base_transaction_decoder, base_accounts):
...
 @pytest.mark.parametrize('base_accounts', [['0x514c4BA193c698100DdC998F17F24bDF59c7b6fB']])
-def test_stake_lp_token_to_gauge(base_accounts, base_transaction_decoder):
+def test_stake_lp_token_to_gauge(base_accounts, base_transaction_decoder, load_global_caches):
     evmhash = deserialize_evm_tx_hash('0x9a0cd1ab0b8e5dbf2718b1c87dad239f7f3a9ed8ff2e07643922b190f80ae898 ')  # noqa: E501
     user_address = base_accounts[0]
     events, _ = get_decoded_events_of_transaction(
         evm_inquirer=base_transaction_decoder.evm_inquirer,
         database=base_transaction_decoder.database,
         tx_hash=evmhash,
+        load_global_caches=load_global_caches,
```
yabirgb, line 150: question: we need to pass load_global_cache here because we don't use the `[chain_name]_transaction_decoder` and instead we create a new one, instead of using the one that you have also modified to have the patch... right?

if instead of `evm_inquirer` and `database` we provide as argument  `[chain_name]_transaction_decoder` to  get_decoded_events_of_transaction we can avoid building it there and having to patch it... also we don't need to care about passing `load_global_caches`. That requires modifying almost all the tests and I believe we can do it after... what do you think? I would add a todo

yabirgb: I mention it because having `load_global_caches` is not really doing nothing but defining a constant

rotkehlchen/tests/unit/decoders/test_velodrome.py
```diff
@@ -380,15 +388,17 @@ def test_remove_liquidity_v1(optimism_transaction_decoder, optimism_accounts):
 
 
 @pytest.mark.vcr()
+@pytest.mark.parametrize('load_global_caches', [['velodrome']])
 @pytest.mark.parametrize('optimism_accounts', [['0x78C13393Aee675DD7ED07ce992210750D1F5dB88']])
-def test_swap_eth_to_token_v2(optimism_accounts, optimism_transaction_decoder):
+def test_swap_eth_to_token_v2(optimism_accounts, optimism_transaction_decoder, load_global_caches):
```
yabirgb, line 393: here for example we do as I said. Providing `optimism_transaction_decoder` as argument we wouldn't need to patch later neither build again the object. Here we are building it twice

rotkehlchen/tests/fixtures/globaldb.py
```diff
@@ -74,6 +74,12 @@ def fixture_remove_global_assets() -> list[str]:
     return []
 
 
+@pytest.fixture(name='load_global_caches')
+def fixture_load_global_caches() -> list[str]:
+    """The decoder for which the global cache should be loaded."""
```
yabirgb, line 79: Can be multiple right?

```suggestion
    """The decoders for which the global cache should be loaded."""
```

yabirgb: Can you also document here that what we use is the name of the folder for the decoder? Do you think it would make sense to use the counterparty? I believe is less error prone. Instead of checking for `in` when you patch you will add a new column for the counterparty

rotkehlchen/tests/utils/decoders.py
```diff
@@ -1,9 +1,31 @@
...
+        stack.enter_context(patch(  # patch to not refresh cache by not downloading new data
+            target='rotkehlchen.chain.evm.node_inquirer.should_update_protocol_cache',
+            new=lambda *args, **kwargs: False,
+        ))
+
+        for path, patch_general, patch_unique in (  # patch to not load cache from DB by default
+            ('rotkehlchen.chain.ethereum.modules.convex.convex_cache', True, True),
+            ('rotkehlchen.chain.evm.decoding.curve.curve_cache', True, True),
+            ('rotkehlchen.chain.evm.decoding.gearbox.gearbox_cache', True, True),
+            ('rotkehlchen.chain.evm.decoding.velodrome.velodrome_cache', True, False),
```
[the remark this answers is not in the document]
OjusWiZard: they both work in the same file `rotkehlchen/chain/evm/decoding/velodrome/velodrome_cache.py`

rotkehlchen/tests/unit/decoders/test_aerodrome.py
OjusWiZard: yeah this makes sense

rotkehlchen/tests/utils/decoders.py
```diff
@@ -1,9 +1,36 @@
...
-        target='rotkehlchen.chain.evm.node_inquirer.should_update_protocol_cache',
-        new=lambda *args, **kwargs: False,
-    )
+    with ExitStack() as stack:
+        stack.enter_context(patch(  # patch to not refresh cache by not downloading new data
+            target='rotkehlchen.chain.evm.node_inquirer.should_update_protocol_cache',
+            new=lambda *args, **kwargs: False,
+        ))
+
+        for counterparties, path, patch_general, patch_unique in (  # patch to not load cache from DB by default  # noqa: E501
```
LefterisJP, line 18: What is patch general and patch unique? Can you either add a comment here or add more info to the function's docstring?

rotkehlchen/tests/utils/decoders.py
```diff
@@ -1,9 +1,36 @@
...
+            ({CPT_CONVEX}, 'rotkehlchen.chain.ethereum.modules.convex.convex_cache', True, True),
```
LefterisJP, line 19: Why pass the counterparty and not the decoder path itself?

rotkehlchen/tests/unit/decoders/test_velodrome.py
```diff
@@ -848,15 +872,17 @@ def test_unstake_lp_token_to_gauge_v2(optimism_accounts, optimism_transaction_de
...
-def test_get_reward_from_gauge_v2(optimism_accounts, optimism_transaction_decoder):
+def test_get_reward_from_gauge_v2(optimism_accounts, optimism_transaction_decoder, load_global_caches):  # noqa: E501
     """Check claiming rewards from a velodrome v2 gauge is properly decoded."""
     evmhash = deserialize_evm_tx_hash('0x9d0eae3c2d2cda853e77a2571a7c702507beb015c08b40406e03baa4b1dde505')  # noqa: E501
     user_address = optimism_accounts[0]
     events, _ = get_decoded_events_of_transaction(
         evm_inquirer=optimism_transaction_decoder.evm_inquirer,
         database=optimism_transaction_decoder.database,
         tx_hash=evmhash,
+        load_global_caches=load_global_caches,
```
LefterisJP, line 885: This is a small but quality of life improvement. If you always pass load_global_caches to this function can't you somehow make it load the value from the fixture automatically? We do

LefterisJP (approved):
LGTM. As discussed today during standup the rest happens in other PRs

PR #8164, opened 2024-06-25
feat: update file uploader

lukicenturi:
image

### DIFF
--- a/frontend/app/src/components/import/FileUpload.vue
+++ b/frontend/app/src/components/import/FileUpload.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { size } from '../../utils/data';
 import type { ImportSourceType } from '@/types/upload-types';
 
 const props = withDefaults(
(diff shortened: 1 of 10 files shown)


frontend/app/src/components/import/FileUpload.vue
```diff
@@ -70,11 +76,32 @@ function onSelect(event: Event) {
...
+function clearTimeoutHandler(timeout: Ref<NodeJS.Timeout | null>) {
+  const timeoutVal = get(timeout);
+  if (timeoutVal) {
+    clearTimeout(timeoutVal);
+    set(timeout, null);
+  }
+}
+
+const errorTimeout: Ref<NodeJS.Timeout | null> = ref(null);
+const uploadedTimeout: Ref<NodeJS.Timeout | null> = ref(null);
```
kelsos, line 88:
```suggestion
const errorTimeout = ref<NodeJS.Timeout>();
const uploadedTimeout = ref<NodeJS.Timeout>();
```

Rather than using explicit `null` let's use undefined as default.
For most of the places now that we removed vuetify and we don't have vue 2.16 we can use `undefined` and keep things simpler.

kelsos (approved):

PR #8165, opened 2024-06-25
Correctly calculate PnL for validators when exiting

### LINKED ISSUE #8095: Excess of 32 ETH from full withdrawal counts as loss
When generating PnL reports with validator exits the excess of 32 Eth (the original staked amount) results in a loss. I think the following line is the wrong way around? I'd expect the 32 Eth to act neutral in PnL, while an excess is a profit, while less than 32 Eth must imply a loss (slashing or offline perhaps).

https://github.com/rotki/rotki/blob/44f76031dfa823873ee1ed0073353e8e9e70bf88/rotkehlchen/history/events/structures/eth2.py#L213

yabirgb:
Closes #8095. PnL will be calculated correctly since we were obtaining a negative PnL when balance was over 32 ETH


LefterisJP (approved):
LGTM

PR #8166, opened 2024-06-25
Feat (Cache): Notify refresh progress

OjusWiZard:
## Checklist

- [x] Notify the progress when the protocol's cache is refreshing
- [x] Add a comment for booleans from [this PR](https://github.com/rotki/rotki/pull/8159#discussion_r1652162870)
- [x] Add a `log.error` in the HOP decoder

### DIFF
--- a/rotkehlchen/chain/evm/decoding/curve/curve_cache.py
+++ b/rotkehlchen/chain/evm/decoding/curve/curve_cache.py
@@ -6,6 +6,7 @@ from rotkehlchen.assets.utils import TokenEncounterInfo, get_or_create_evm_token
 from rotkehlchen.chain.evm.constants import ETH_SPECIAL_ADDRESS, ZERO_ADDRESS
 from rotkehlchen.chain.evm.contracts import EvmContract
 from rotkehlchen.chain.evm.decoding.curve.constants import (
+    CPT_CURVE,
     CURVE_ADDRESS_PROVIDER,
     CURVE_API_URL,
     CURVE_CHAIN_ID,
@@ -13,6 +14,10 @@ from rotkehlchen.chain.evm.decoding.curve.constants import (
     IGNORED_CURVE_POOLS,
 )
 from rotkehlchen.chain.evm.types import string_to_evm_address
+from rotkehlchen.chain.evm.utils import (
+    maybe_notify_cache_query_status,
+    maybe_notify_new_pools_status,
+)
 from rotkehlchen.constants import ONE
 from rotkehlchen.db.addressbook import DBAddressbook
 from rotkehlchen.errors.misc import (
@@ -40,6 +45,7 @@ from rotkehlchen.types import (
     ChainID,
     ChecksumEvmAddress,
     EvmTokenKind,
+    Timestamp,
 )
 from rotkehlchen.utils.network import request_get_dict
 
@@ -47,6 +53,7 @@ if TYPE_CHECKING:
     from gevent import DBCursor
     from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
     from rotkehlchen.db.dbhandler import DBHandler
+    from rotkehlchen.user_messages import MessagesAggregator
 
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
@@ -102,6 +109,7 @@ def read_curve_pools_and_gauges(chain_id: ChainID) -> tuple[dict[ChecksumEvmAddr
 def _ensure_curve_tokens_existence(
         evm_inquirer: 'EvmNodeInquirer',
         all_pools: list[CurvePoolData],
+        msg_aggregator: 'MessagesAggregator',
 ) -> list[CurvePoolData]:
     """This function receives data about curve pools and ensures that lp tokens and pool coins
     exist in rotki's database.
@@ -109,8 +117,17 @@ def _ensure_curve_tokens_existence(
     Since is possible that a pool has an invalid token we keep a mapping of the valid ones
     and return it.
     """
-    verified_pools = []
-    for pool in all_pools:
+    verified_pools, last_notified_ts, all_pools_length = [], Timestamp(0), len(all_pools)
+    for idx, pool in enumerate(all_pools):
+        last_notified_ts = maybe_notify_cache_query_status(
+            msg_aggregator=msg_aggregator,
+            last_notified_ts=last_notified_ts,
+            protocol=CPT_CURVE,
+            chain=evm_inquirer.chain_id,
+            processed=idx + 1,
+            total=all_pools_length,
+        )
+
         # Ensure pool coins exist in the globaldb.
         # We have to create underlying tokens only if pool utilizes them.
         if pool.underlying_coins is None or pool.underlying_coins == pool.coins:
@@ -318,8 +335,7 @@ def _query_curve_data_from_api(
     May raise:
     - RemoteError if failed to query curve api
     """
-    all_api_pools = []
-    api_url = CURVE_API_URL.format(curve_blockchain_id=CURVE_CHAIN_ID[chain_id])
+    all_api_pools, api_url = [], CURVE_API_URL.format(curve_blockchain_id=CURVE_CHAIN_ID[chain_id])
     log.debug(f'Querying curve api {api_url}')
     response_json = request_get_dict(api_url)
     if response_json['success'] is False:
@@ -364,6 +380,7 @@ def _query_curve_data_from_api(
 def _query_curve_data_from_chain(
         evm_inquirer: 'EvmNodeInquirer',
         existing_pools: list[ChecksumEvmAddress],
+        msg_aggregator: 'MessagesAggregator',
 ) -> list[CurvePoolData] | None:
     """
     Query all curve information(lp tokens, pools, gagues, pool coins) from metaregistry.
@@ -388,8 +405,17 @@ def _query_curve_data_from_chain(
         deployed_block=0,  # deployment_block is not used and the contract is dynamic
     )
     pool_count = metaregistry.call(node_inquirer=evm_inquirer, method_name='pool_count')
-    new_pools = []
+    new_pools: list[CurvePoolData] = []
+    last_notified_ts = Timestamp(0)
     for pool_index in range(pool_count):
+        last_notified_ts = maybe_notify_new_pools_status(
+            msg_aggregator=msg_aggregator,
+            last_notified_ts=last_notified_ts,
+            protocol=CPT_CURVE,
+            chain=evm_inquirer.chain_id,
+            get_new_pools_count=lambda: len(new_pools),
+        )
+
         raw_address = metaregistry.call(
             node_inquirer=evm_inquirer,
             method_name='pool_list',
@@ -453,6 +479,7 @@ def _query_curve_data_from_chain(
 def query_curve_data(
         inquirer: 'EvmNodeInquirer',
         cache_type: Literal[CacheType.CURVE_LP_TOKENS],
+        msg_aggregator: 'MessagesAggregator',
 ) -> list[CurvePoolData] | None:
     """Query curve lp tokens, curve pools and curve gauges.
     First tries to find data via curve api and if fails to do so, queries the chain (metaregistry).
@@ -481,6 +508,7 @@ def query_curve_data(
             pools_data = _query_curve_data_from_chain(
                 evm_inquirer=inquirer,
                 existing_pools=existing_pools,
+                msg_aggregator=msg_aggregator,
             )
         except RemoteError as err:
             log.error(f'Could not query chain for curve pools due to: {err}')
@@ -492,6 +520,7 @@ def query_curve_data(
     verified_pools = _ensure_curve_tokens_existence(
         evm_inquirer=inquirer,
         all_pools=pools_data,
+        msg_aggregator=msg_aggregator,
     )
     return verified_pools
 
@@ -501,8 +530,7 @@ def get_lp_and_gauge_token_addresses(
         chain_id: ChainID,
 ) -> set['ChecksumEvmAddress']:
     """Reads the db to get the lp and gauge token addresses for the given pool address"""
-    addresses = set()
-    chain_id_str = str(chain_id.serialize_for_db())
+    addresses, chain_id_str = set(), str(chain_id.serialize_for_db())
     with GlobalDBHandler().conn.read_ctx() as cursor:
         if (key := cursor.execute(
             'SELECT key FROM unique_cache WHERE value = ?',
--- a/rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
+++ b/rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
@@ -11,6 +11,10 @@ from rotkehlchen.chain.evm.decoding.gearbox.constants import (
     CPT_GEARBOX,
 )
 from rotkehlchen.chain.evm.types import string_to_evm_address
+from rotkehlchen.chain.evm.utils import (
+    maybe_notify_cache_query_status,
+    maybe_notify_new_pools_status,
+)
 from rotkehlchen.constants.misc import ONE, ZERO
 from rotkehlchen.errors.misc import BlockchainQueryError, NotERC20Conformant, RemoteError
 from rotkehlchen.errors.serialization import DeserializationError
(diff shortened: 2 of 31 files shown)


rotkehlchen/chain/evm/decoding/curve/curve_cache.py
```diff
@@ -388,8 +408,21 @@ def _query_curve_data_from_chain(
...
+            msg_aggregator.add_message(
+                message_type=WSMessageType.PROTOCOL_CACHE_UPDATES,
+                data={
+                    'protocol': CPT_CURVE,
+                    'chain': evm_inquirer.chain_id.to_name(),
+                    'processed': 0,
+                    'total': len(new_pools),
+                },
+            )
+            last_notified_ts = current_time
```
yabirgb, line 424: this logic can be abstracted. Create a function `notify_cache_query_status` that takes as argument the counterparty, the current processed, the total and the chain name and then you do

```
last_notified_ts = notify_cache_query_status(...)
```

rotkehlchen/chain/evm/decoding/curve/curve_cache.py
```diff
@@ -388,8 +408,21 @@ def _query_curve_data_from_chain(
...
+    new_pools: list[CurvePoolData] = []
+    last_notified_ts = 0
     for pool_index in range(pool_count):
+        if (current_time := ts_now()) - last_notified_ts > 5:
+            msg_aggregator.add_message(
+                message_type=WSMessageType.PROTOCOL_CACHE_UPDATES,
+                data={
+                    'protocol': CPT_CURVE,
+                    'chain': evm_inquirer.chain_id.to_name(),
+                    'processed': 0,
```
yabirgb, line 420: notice how you are sending 0 here always

rotkehlchen/chain/evm/decoding/curve/curve_cache.py
```diff
@@ -388,8 +408,21 @@ def _query_curve_data_from_chain(
...
+                    'total': len(new_pools),
```
yabirgb, line 421: left doesn't like calling len in loops. Create a variable outside

rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
```diff
@@ -157,11 +160,26 @@ def register_token(evm_inquirer: 'EvmNodeInquirer', token_address: str, pool: Ge
...
+    all_pools_length = len(all_pools)
+    last_notified_ts = 0
+    for idx, pool in enumerate(all_pools):
+        if (current_time := ts_now()) - last_notified_ts > 5:
+            msg_aggregator.add_message(
+                message_type=WSMessageType.PROTOCOL_CACHE_UPDATES,
+                data={
+                    'protocol': CPT_GEARBOX,
+                    'chain': evm_inquirer.chain_id.to_name(),
+                    'processed': idx + 1,
```
yabirgb, line 177: isn't idx here correct? WHen you start you send the message because is the frist time and you have processed 0 yet

rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
```diff
@@ -277,8 +296,21 @@ def query_gearbox_data_from_chain(
...
+    new_pools: list[GearboxPoolData] = []
+    last_notified_ts = 0
     for pool_data in pools_data:
+        if (current_time := ts_now()) - last_notified_ts > 5:
+            msg_aggregator.add_message(
+                message_type=WSMessageType.PROTOCOL_CACHE_UPDATES,
+                data={
+                    'protocol': CPT_GEARBOX,
+                    'chain': evm_inquirer.chain_id.to_name(),
+                    'processed': 0,
```
yabirgb, line 308: same here, you hardcoded 0.

rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
```diff
@@ -277,8 +296,21 @@ def query_gearbox_data_from_chain(
...
-    new_pools = []
+    new_pools: list[GearboxPoolData] = []
+    last_notified_ts = 0
     for pool_data in pools_data:
+        if (current_time := ts_now()) - last_notified_ts > 5:
+            msg_aggregator.add_message(
+                message_type=WSMessageType.PROTOCOL_CACHE_UPDATES,
+                data={
+                    'protocol': CPT_GEARBOX,
+                    'chain': evm_inquirer.chain_id.to_name(),
```
yabirgb, line 307: don't call .to_name on each loop iteration. Calling functions in python has a cost, define it as a variable before the loop

rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
```diff
@@ -277,8 +296,21 @@ def query_gearbox_data_from_chain(
     pools_data = evm_inquirer.contracts.contract(
         CHAIN_ID_TO_DATA_COMPRESSOR[evm_inquirer.chain_id],
     ).call(node_inquirer=evm_inquirer, method_name='getPoolsV3List')
-    new_pools = []
+    new_pools: list[GearboxPoolData] = []
+    last_notified_ts = 0
     for pool_data in pools_data:
+        if (current_time := ts_now()) - last_notified_ts > 5:
```
yabirgb, line 302: do you really need to send it every 5 seconds? Can't it be every 10-15 processed entries? Asking for the tiemstamp is slower than checking the %

OjusWiZard: I kept it timewise because it's simple this way to avoid flooding the WS messages. and it ensures that user see the progress not too faster/slower than required

rotkehlchen/chain/evm/decoding/curve/curve_cache.py
OjusWiZard: in some places the `processed` is 0 because 0 pools are processed at that point. so, it happens in two iterations.
first we iterate to find how many new pools are there, this is where we populate the `total` value and will increment it at every message sent.
then in the next iteration `total` number is constant, and `processed` is incremented.

rotkehlchen/chain/evm/decoding/curve/curve_cache.py
```diff
@@ -388,8 +406,19 @@ def _query_curve_data_from_chain(
...
+    last_notified_ts = Timestamp(0)
+    chain_name = evm_inquirer.chain_id.to_name()
     for pool_index in range(pool_count):
+        last_notified_ts = maybe_notify_cache_query_status(
+            msg_aggregator=msg_aggregator,
+            last_notified_ts=last_notified_ts,
+            protocol=CPT_CURVE,
+            chain_name=chain_name,
+            processed=0,
+            total=len(new_pools),
```
yabirgb, line 419: can you add a small comment mentioning that you keep processed 0 here and increment `new_pools` and the processed get incremented in the function `add_name_here`. It was confusing even after explaining it in discord but having the other function name helped here.

rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
```diff
@@ -277,8 +293,19 @@ def query_gearbox_data_from_chain(
...
+    last_notified_ts = Timestamp(0)
+    chain_name = evm_inquirer.chain_id.to_name()
     for pool_data in pools_data:
+        last_notified_ts = maybe_notify_cache_query_status(
+            msg_aggregator=msg_aggregator,
+            last_notified_ts=last_notified_ts,
+            protocol=CPT_GEARBOX,
+            chain_name=chain_name,
+            processed=0,
+            total=len(new_pools),
```
yabirgb, line 306: same here

yabirgb:
image

rotkehlchen/chain/evm/decoding/curve/curve_cache.py
```diff
@@ -388,8 +406,19 @@ def _query_curve_data_from_chain(
...
+    new_pools: list[CurvePoolData] = []
+    last_notified_ts = Timestamp(0)
+    chain_name = evm_inquirer.chain_id.to_name()
     for pool_index in range(pool_count):
+        last_notified_ts = maybe_notify_cache_query_status(
+            msg_aggregator=msg_aggregator,
+            last_notified_ts=last_notified_ts,
+            protocol=CPT_CURVE,
+            chain_name=chain_name,
+            processed=0,  # 0 because here we only query pools to find the total number of them, it will be incremented while processing the pools in `_ensure_curve_tokens_existence`  # noqa: E501
```
LefterisJP, line 418:
```suggestion
            processed=0,  # 0 because here we only query pools to find their total number. It will be incremented while processing the pools in `_ensure_curve_tokens_existence`  # noqa: E501
```

rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
```diff
@@ -157,11 +159,24 @@ def register_token(evm_inquirer: 'EvmNodeInquirer', token_address: str, pool: Ge
 def ensure_gearbox_tokens_existence(
         evm_inquirer: 'EvmNodeInquirer',
         all_pools: list[GearboxPoolData],
+        msg_aggregator: 'MessagesAggregator',
 ) -> list[GearboxPoolData]:
     """This function receives data about gearbox pools and ensures that lp tokens and pool coins
     exist in rotki's database."""
     verified_pools = []
-    for pool in all_pools:
+    all_pools_length = len(all_pools)
```
LefterisJP, line 167: nitpick: same comments as in the previous file for compression of variable initialization. Won't repeat anymore. Please take care of it. `verified_pools` is also part of this initialization.

rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
```diff
@@ -277,8 +293,19 @@ def query_gearbox_data_from_chain(
...
+    new_pools: list[GearboxPoolData] = []
+    last_notified_ts = Timestamp(0)
+    chain_name = evm_inquirer.chain_id.to_name()
     for pool_data in pools_data:
+        last_notified_ts = maybe_notify_cache_query_status(
+            msg_aggregator=msg_aggregator,
+            last_notified_ts=last_notified_ts,
+            protocol=CPT_GEARBOX,
+            chain_name=chain_name,
+            processed=0,  # 0 because here we only query pools to find the total number of them, it will be incremented while processing the pools in `ensure_gearbox_tokens_existence`  # noqa: E501
```
LefterisJP, line 305:
```suggestion
            processed=0,  # 0 because here we only query pools to find their total number. It will be incremented while processing the pools in `ensure_gearbox_tokens_existence`  # noqa: E501
```

rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
```diff
@@ -277,8 +293,19 @@ def query_gearbox_data_from_chain(
...
+            total=len(new_pools),  # notifying the increase in len(pools) every 5 seconds as we query them  # noqa: E501
```
LefterisJP, line 306: Hmmm here this repeats again like in curve. And I guess you do it in other files too. Can't you create a version of `maybe_notify_cache_query_status` that's only for this? To find the total number of pools? It would just be calling this function with processed=0, and total=len(new_pools) in a for loop. And have a proper name so you don't need to repeat this comment and explain what it's for.

Something like (pseudocode):

```python
def find_total_number_of_pools(...) -> int:
for pool_data in pools_data:
    maybe_notify_cache_query_status(
            last_notified_ts=last_notified_ts,
            protocol=CPT_GEARBOX,
            chain_name=chain_name,
            processed=0,  # 0 because here we only query pools to find their total number. It will be incremented while processing the pools in `ensure_gearbox_tokens_existence`  # noqa: E501
total=len(new_pools), # notifying the increase in len(pools) every 5 seconds as we query them # noqa: E501
 )

File: rotkehlchen/chain/evm/decoding/hop/decoder.py [hunk 14]
```diff
@@ -380,6 +380,10 @@ def _decode_common_liquidity(self, context: DecoderContext, first_token_raw: byt
             return None
 
         if (swap_asset_id := self.swaps_to_asset.get(context.tx_log.address)) is None:
+            log.error(
+                f'Could not find asset for the saddle swap address {context.tx_log.address!r} '
```
LefterisJP, line 384: This is not what we need to add in such logs. The important things are missing. The things we need to know to try and debug. Those are:

1. Chain name (notice this is deployed in many chains)
2. Transaction hash.


The rest is not necessary as we would see it the moment we debug and they would instead flood the logs with too munch info.

rotkehlchen/chain/evm/decoding/velodrome/velodrome_cache.py
```diff
@@ -164,15 +169,28 @@ def query_velodrome_data_from_chain_and_maybe_create_tokens(
...
         protocol = AERODROME_POOL_PROTOCOL
+        counterparty = CPT_AERODROME
 
-    pool_data = []
+    pool_data: list[dict] = []
     offset = 0
     limit = 200
     pool_data_chunk: list[dict] = []
+    last_notified_ts = Timestamp(0)
+    chain_name = inquirer.chain_id.to_name()
```
LefterisJP, line 183: same here for compression of initialization. Too many stuff are initialized in all those lines before the while

rotkehlchen/chain/evm/decoding/velodrome/velodrome_cache.py
```diff
@@ -164,15 +169,28 @@ def query_velodrome_data_from_chain_and_maybe_create_tokens(
...
     while len(pool_data_chunk) == limit or (len(pool_data_chunk) == 0 and offset == 0):
+        last_notified_ts = maybe_notify_cache_query_status(
```
LefterisJP, line 185: use the function I told you to create here

rotkehlchen/chain/evm/utils.py
```diff
@@ -171,3 +180,28 @@ def lp_price_from_uniswaplike_pool_contract(
     numerator = token0_supply * token0_price + token1_supply * token1_price
     share_value = numerator / total_supply
     return Price(share_value)
+
+
+def maybe_notify_cache_query_status(
+        msg_aggregator: 'MessagesAggregator',
+        last_notified_ts: Timestamp,
+        protocol: str,
+        chain_name: str,
```
LefterisJP, line 189: why pass chain name and not either ChainID or Blockchain enum here? It's best to do conversion/serialization when stuff leave our code. So right here.

This is informally called the sandwich pattern. Decide on a format for your codebase. For us it's these enums. And code that comes from the outside can be anything but needs to be serialized into it at point of entry (API, CSV import etc) and deserialized to it at point of exit (such as these websocket messages).

rotkehlchen/chain/evm/decoding/curve/curve_cache.py
```diff
@@ -388,8 +405,17 @@ def _query_curve_data_from_chain(
         deployed_block=0,  # deployment_block is not used and the contract is dynamic
     )
     pool_count = metaregistry.call(node_inquirer=evm_inquirer, method_name='pool_count')
-    new_pools = []
+    new_pools: list[CurvePoolData] = []
```
LefterisJP, line 408: nitpick: could have also compacted this one

rotkehlchen/chain/evm/decoding/gearbox/gearbox_cache.py
```diff
@@ -277,8 +293,17 @@ def query_gearbox_data_from_chain(
     pools_data = evm_inquirer.contracts.contract(
         CHAIN_ID_TO_DATA_COMPRESSOR[evm_inquirer.chain_id],
     ).call(node_inquirer=evm_inquirer, method_name='getPoolsV3List')
-    new_pools = []
+    new_pools: list[GearboxPoolData] = []
```
LefterisJP, line 296: same here

rotkehlchen/chain/evm/decoding/hop/decoder.py
```diff
@@ -380,6 +380,10 @@ def _decode_common_liquidity(self, context: DecoderContext, first_token_raw: byt
             return None
 
         if (swap_asset_id := self.swaps_to_asset.get(context.tx_log.address)) is None:
+            log.error(
+                f'Could not find asset for the saddle swap address while decoding '
+                f'{context.transaction.tx_hash.hex()} in chain {self.evm_inquirer.chain_id}.',
```
LefterisJP, line 385: Data is all there now but keep consistent with other such error logs. It's not the chain id but the chain name and should read like:

```suggestion
                f'{self.evm_inquirer.chain_name} transaction {context.transaction.tx_hash.hex()}',
```

LefterisJP (approved):
LGTM if it passes. Left some nitpick comments you can handle in a next PR

PR #8168, opened 2024-06-25
Make the abi string unique in the globaldb schema

yabirgb:
Note https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=68498209


LefterisJP (approved):
LGTM

PR #8182, opened 2024-06-27
Fix (Aave): Missing aTokens and order

OjusWiZard:
Closes:
- https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=67740083
- https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=68806760

## Checklist

- [x] Add missing tokens in Aave v2 and v3
- [x] Fix order of events with approval events
- [x] Log change in HOP decoder

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` Deposit events in Aave will now show in the correct order.
 * :feature:`3971` Show the total collateral ratio in the Liquity Trove section.
 * :feature:`2323` Add support for Huobi exchange.
 * :feature:`-` Curve vote escrow locking and withdrawal events will now be properly decoded on ethereum mainnet.
--- a/rotkehlchen/chain/evm/decoding/aave/common.py
+++ b/rotkehlchen/chain/evm/decoding/aave/common.py
@@ -1,3 +1,4 @@
+import logging
 from abc import abstractmethod
 from typing import TYPE_CHECKING, Any, Literal, Optional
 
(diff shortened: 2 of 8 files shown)


docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` Deposit events in Aave will now show in correct order with approval events.
```
```suggestion
* :bug:`-` Deposit events in Aave will now show in the correct order.
```

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -119,7 +124,7 @@ def _decode_deposit(
             token: 'EvmToken',
             tx_log: EvmTxReceiptLog,
             decoded_events: list['EvmEvent'],
-    ) -> None:
+    ) -> tuple[EvmEvent | None, EvmEvent | None]:
```
yabirgb, line 127: update the docstrings to state what is returned

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -303,15 +316,19 @@ def _decode_repay(
                     resolved_asset = event.asset.resolve_to_asset_with_symbol()
                     event.counterparty = self.counterparty
                     event.notes = f'Return {event.balance.amount} {resolved_asset.symbol} to {self.label}'  # noqa: E501
+                    return_event = event
```
yabirgb, line 319: here the two events are of type spend. Is it needed the order for the accounting rules? If not maybe you can just return None, None and avoid tracking them

OjusWiZard: it is to avoid any other event (like approval) coming in between and keeping it consistent with others, so that token movements are shown in the last.

yabirgb:
Failing test is a remote error. Please take note of it and vcr it

LefterisJP (approved):
LGTM

PR #8187, opened 2024-06-27
Abstract gitcoin cgrants to be multievm chain capable

LefterisJP:
Also handle the other kind of funds claimed possible event log which unfortunately has the same topics[0] hash


LefterisJP:
all green. This means ...... MERGEMENTATION

PR #8188, opened 2024-06-27
Chore (History Events): Notify unknown asset case

OjusWiZard:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=68262589

## Checklist

- [x] Notify error with this rare case where we fetch the history events of an asset that is unknown


rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -527,6 +529,27 @@ def test_get_events(rotkehlchen_api_server: 'APIServer'):
...
+    ):
+        response = requests.post(
+            api_url_for(
+                rotkehlchen_api_server,
+                'historyeventresource',
+            ),
+        )
+    assert_proper_sync_response_with_result(response)
+    assert rotkehlchen_api_server.rest_api.rotkehlchen.msg_aggregator.consume_errors() == [
+        "Failed to deserialize history event (1, 6, 'STARK-STARK-STARK', 0, 1673146287380, 'B', 'Kraken', 'ETH2', '0.0000400780', '0.051645312360', 'Staking reward from kraken', 'staking', 'reward', None, None, None, None, None, None, None) due to Unknown asset ETH2 provided. Try redecoding the event or adding the asset manually.",  # noqa: E501
```
yabirgb, line 548: This part of the error `history event (1, 6, 'STARK-STARK-STARK', 0, 1673146287380, 'B', 'Kraken', 'ETH2', '0.0000400780', '0.051645312360', 'Staking reward from kraken', 'staking', 'reward', None, None, None, None, None, None, None)` is not useful for the user and add noise. Consider that we display errors in a small notification and the action that the user needs to perform probably will be dismissed

rotkehlchen/db/history_events.py
```diff
@@ -503,7 +505,12 @@ def get_history_events(
                     data = entry[data_start_idx:HISTORY_BASE_ENTRY_LENGTH + 1]
                     deserialized_event = HistoryEvent.deserialize_from_db(entry[data_start_idx:])
             except (DeserializationError, UnknownAsset) as e:
-                log.debug(f'Failed to deserialize history event {entry} due to {e!s}')
+                error = f'Failed to deserialize history event {entry} due to {e!s}'
+                log.error(error)
+                if msg_aggregator is not None:
```
yabirgb, line 510: As you can see the entry is in the log but not in the message. It is useful there because we would ask the user for the logs but is not useful for the user to have the history event tuple. I would tell the user the identifier of the event and then if the reason for the error is `UnknownAsset` you can show the try redecoding the event. If it is a deserialization error I believe is not useful to re-decode and they should contact us

rotkehlchen/db/history_events.py
```diff
@@ -500,10 +502,18 @@ def get_history_events(
...
 
                 else:
-                    data = entry[data_start_idx:HISTORY_BASE_ENTRY_LENGTH + 1]
-                    deserialized_event = HistoryEvent.deserialize_from_db(entry[data_start_idx:])
+                    data = entry[data_start_idx:]
+                    deserialized_event = HistoryEvent.deserialize_from_db(data)
             except (DeserializationError, UnknownAsset) as e:
-                log.debug(f'Failed to deserialize history event {entry} due to {e!s}')
+                error = f'Failed to deserialize history event {entry} due to {e!s}'
+                log.error(error)
```
yabirgb, line 509: why create a variable if it is only used here? 

```suggestion
                log.error(f'Failed to deserialize history event {entry} due to {e!s}')
```

rotkehlchen/db/history_events.py
```diff
@@ -437,6 +438,7 @@ def get_history_events(
             filter_query: HistoryEventFilterQuery | EvmEventFilterQuery | EthDepositEventFilterQuery | EthWithdrawalFilterQuery,  # noqa: E501
             has_premium: bool,
             group_by_event_ids: bool = False,
+            msg_aggregator: 'MessagesAggregator | None' = None,
```
LefterisJP, line 441: hmm what are you doing here? Why are you passing the message aggregator here and in the rest? Please remove

rotkehlchen/db/history_events.py
```diff
@@ -500,10 +502,17 @@ def get_history_events(
...
 
                 else:
-                    data = entry[data_start_idx:HISTORY_BASE_ENTRY_LENGTH + 1]
-                    deserialized_event = HistoryEvent.deserialize_from_db(entry[data_start_idx:])
+                    data = entry[data_start_idx:]
+                    deserialized_event = HistoryEvent.deserialize_from_db(data)
             except (DeserializationError, UnknownAsset) as e:
-                log.debug(f'Failed to deserialize history event {entry} due to {e!s}')
+                log.error(f'Failed to deserialize history event {entry} due to {e!s}')
+                if msg_aggregator is not None:
```
LefterisJP, line 509: hmmm this is wrong. You don't need to pass it as optional and check it in here. You already have it.

`self.database.msg_aggregator`

rotkehlchen/db/history_events.py
```diff
@@ -500,10 +502,17 @@ def get_history_events(
...
+                    msg_aggregator.add_error(
+                        f'Could not deserialize history event: {data[0]}. ' + (
```
LefterisJP, line 511: Okay this is also too much. Unless I misunderstand what data[0] is.

Keep in mind. `msg_aggregator.add_error()` pops up as an error popup to the user. That means it need to be short and succinct.

What we need to show to them is that their DB is probably messed up through an error of theirs and they need to take action. And we should have more details in a log error.

LefterisJP (approved):
LGTM

PR #8191, opened 2024-06-28
Decode aave v2 incentives claiming

LefterisJP:
Also abstract common code between aave v2 and v3 incentives claiming


LefterisJP:
Al green. Merging as I got one more coming.

PR #8192, opened 2024-06-28
Fix 1inchv4 decoding in polygon

LefterisJP:
Fix https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=69113977


LefterisJP:
forgot this one. Green is green. Merge is now.

PR #8193, opened 2024-06-28
Refactor (Exchanges): Purge coinbasepro

### LINKED ISSUE #8143: Purge Coinbasepro
## Problem Definition

Fun fact, we still had this exchange lying around since the API was working and it was querying both history and balances.

Seems coinbase finally removed it.

## Task

Purge it with fire. Naturally preserve all history.

OjusWiZard:
Signed-off-by: OjusWiZard <ojuswimail@gmail.com>

Closes #8143

## Checklist

- [x] Remove coinbasepro, retaining support for previous data


rotkehlchen/exchanges/okx.py
```diff
@@ -82,7 +82,7 @@ def edit_exchange_credentials(self, credentials: ExchangeAuthCredentials) -> boo
         if credentials.api_key is not None:
             self.session.headers.update({'OK-ACCESS-KEY': self.api_key})
         if credentials.passphrase is not None:
-            self.session.headers.update({'OK-ACCESS-PASSPHRASE': self.passphrase})
+            self.session.headers.update({'OK-ACCESS-PASSPHRASE': credentials.passphrase})
```
yabirgb, line 85: this is a bug right? Please add a changelog entry for this. Also add one for coinbasePro. For ftx we had

```
* :feature:`5933` FTX and FTX US support is removed since the exchanges no longer exist
```

rotkehlchen/exchanges/constants.py
yabirgb, line 1: Check with the frontend if this needs any change or allows to remove any code. I believe passhprase is a property that we add and everything is dinamic

OjusWiZard: > I don't see anything from a quick search
~ kelsos

LefterisJP (commented):
It's missign removal of credentials if existing in the database.

LefterisJP (approved):
lgtm

PR #8197, opened 2024-06-29
Add more gitcoin grants round mappings

LefterisJP:
Also better wording in clrfund events


LefterisJP:
`rotkehlchen/tests/unit/test_tokens.py::test_last_queried_ts` has a network error. VCRing or modifying would probably make sense.

PR #8203, opened 2024-07-01
Chore (Tools): Sync remote updates with packageddb

OjusWiZard:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=68169085

## Checklist

- [x] Add script to add remote updates of `location_asset_mappings` and `location_unsupported_assets`.
- [x] Add test to check the consistency of these updates with packaged db.

### DIFF
--- a/rotkehlchen/tests/unit/globaldb/test_globaldb_consistency.py
+++ b/rotkehlchen/tests/unit/globaldb/test_globaldb_consistency.py
@@ -1,3 +1,4 @@
+import json
 from collections import defaultdict
 from dataclasses import dataclass
 from pathlib import Path
@@ -10,6 +11,8 @@ import pytest
 from rotkehlchen.assets.asset import Asset
 from rotkehlchen.chain.evm.decoding.aave.constants import CPT_AAVE_V3
 from rotkehlchen.constants.misc import GLOBALDIR_NAME, ONE
+from rotkehlchen.db.constants import UpdateType
+from rotkehlchen.db.updates import RotkiDataUpdater
 from rotkehlchen.fval import FVal
 from rotkehlchen.globaldb.updates import AssetsUpdater
 from rotkehlchen.tests.fixtures.globaldb import create_globaldb
@@ -23,10 +26,12 @@ from rotkehlchen.types import (
     YEARN_VAULTS_V1_PROTOCOL,
     YEARN_VAULTS_V2_PROTOCOL,
     ChainID,
+    Location,
     Timestamp,
 )
 
 if TYPE_CHECKING:
+    from rotkehlchen.db.dbhandler import DBHandler
     from rotkehlchen.globaldb.handler import GlobalDBHandler
     from rotkehlchen.types import ChecksumEvmAddress
     from rotkehlchen.user_messages import MessagesAggregator
@@ -79,7 +84,7 @@ class DBToken:
         }
 
 
-def test_update_consistency_with_packaged_db(
+def test_asset_updates_consistency_with_packaged_db(
         tmpdir_factory: 'pytest.TempdirFactory',
         messages_aggregator: 'MessagesAggregator',
 ):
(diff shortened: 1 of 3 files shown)


rotkehlchen/tests/unit/globaldb/test_globaldb_consistency.py
```diff
@@ -371,3 +376,87 @@ def test_oracle_ids_in_asset_collections(globaldb: 'GlobalDBHandler'):
...
+        messages_aggregator: 'MessagesAggregator',
+        database: 'DBHandler',
+):
+    """Test that the remote updates are consistent with the packaged db for:
+    - Location asset mappings
+    - Location unsupported assets"""
+    temp_data_dir = Path(tmpdir_factory.mktemp(GLOBALDIR_NAME))
+    (old_db_dir := temp_data_dir / GLOBALDIR_NAME).mkdir(parents=True, exist_ok=True)
+    request.urlretrieve(
+        url='https://github.com/rotki/rotki/raw/v1.33.0/rotkehlchen/data/global.db',
```
yabirgb, line 392: here you have hardcoded the version. Shouldn't this be dinamic? or it doesn't matter because it will update it

tools/assets_database/populate_with_mappings.py
```diff
@@ -0,0 +1,48 @@
+"""
+Tool to create a globaldb with the remote updates in the data repo for location assets mappings and
+unsupported assets.
+
+To call the script do:
+python -m tools.assets_database.populate_with_mappings --start-db-version 9 --target-directory data
+"""
```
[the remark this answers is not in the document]
OjusWiZard: hmmm we should have a start-db-path argument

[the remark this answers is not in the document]
yabirgb: how I see it is that usually what you would want to do is either do the updates or the updates + remote data so the script to populate with mappings instead would be an option of the first `--add-remote-mappings`

rotkehlchen/tests/unit/globaldb/test_globaldb_consistency.py
OjusWiZard: this version is hardcoded because the `location_asset_mappings` and `location_unsupported_assets` tables were added from this version and after this they will be updated

tools/assets_database/utils.py
```diff
@@ -30,10 +22,19 @@ def parse_args() -> argparse.Namespace:
...
+    )
+    p.add_argument(
+        '--start-db-hash',
+        type=str,
+        help='The starting DB commit hash',
+    )
+    p.add_argument(
+        '--start-db-path',
+        type=str,
+        help='The starting DB path',
```
yabirgb, line 37: I find it confusing. What happens if I provide both db-path and db-hash? This should be documented in the help message and also avoid any such possible error. Also can't you make it so if you add the assets to also add the remote updates?

OjusWiZard:
`rotkehlchen/tests/api/test_yearn_vaults.py::test_query_yearn_vault_v2_balances` failed earlier due to some CI hiccup.

rotkehlchen/tests/unit/globaldb/test_globaldb_consistency.py
```diff
@@ -371,3 +376,87 @@ def test_oracle_ids_in_asset_collections(globaldb: 'GlobalDBHandler'):
...
+    # find all the location_unsupported_assets that are present in packaged db but not in remote updates  # noqa: E501
+    missing_unsupported_assets = defaultdict(list)
+    for location, symbol in packaged_location_unsupported_assets:
+        if (location, symbol) not in updated_location_unsupported_assets:
+            missing_unsupported_assets[Location.deserialize_from_db(location).serialize()].append(symbol)
+
+    if len(missing_unsupported_assets) > 0:
+        missing_in_remote_updates.append(missing_unsupported_assets)
+
+    if len(missing_in_packaged_db) > 0:
```
LefterisJP, line 458: Why not fail here?

tools/assets_database/main.py
```diff
@@ -1,123 +1,92 @@
...
+        up_to_version=parse_args().target_version,
+        conflicts=None,
+    )) is not None:
+        assert assets_updater.perform_update(
+            up_to_version=None,
+            conflicts={
+                Asset(conflict['identifier']): 'remote'
+                for conflict in conflicts
+            },
+        ) is None
```
LefterisJP, line 57:
```suggestion
        ) is None, 'ADD AN ERROR TO INFORM THE CALLER WHAT WENT WRONG'
```

tools/assets_database/main.py
```diff
@@ -1,123 +1,92 @@
...
+
+
+def populate_location_mappings():
+    """Apply remote updates to query mappings for assets"""
+    msg_aggregator = MessagesAggregator()
+    with TemporaryDirectory() as tmp_dir:
+        user_db = DBHandler(  # the RotkiDataUpdater takes a user db argument but is not used for the UpdateTypes that we use.  # noqa: E501
```
LefterisJP, line 64: If it's not used just pass `None`, leave the comment you have and if there is typin checks here do type: ignore next to the comment you left explaining it's; not used

tools/assets_database/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    else:
+        url = f'https://github.com/rotki/rotki/raw/{commit_hash}/rotkehlchen/data/global.db'
+
+    try:
+        response = requests.get(url=url, timeout=CachedSettings().get_timeout_tuple())
+    except requests.exceptions.RequestException:
+        print(f'Couldnt download v{version} global.db from github')
+        sys.exit(1)
+
+    if response.status_code != 200:
```
LefterisJP, line 90:
```suggestion
    if response.status_code != HTTPStatus.OK:
```

tools/assets_database/utils.py
```diff
@@ -0,0 +1,134 @@
...
+        version=args.start_db_version,
+        commit_hash=args.start_db_hash,
+        path=args.start_db_path,
+        branch=args.assets_branch,
+    )
+    return GlobalDBHandler(data_dir=target_directory, sql_vm_instructions_cb=0), target_directory
+
+
+def clean_folder(globaldb: GlobalDBHandler, target_directory: Path):
+    """Due to the way globaldb initializes we have two of them. Clean up the extra one"""
```
LefterisJP, line 129: what?

tools/assets_database/main.py
OjusWiZard: so.. it is used, I just checked. Earlier it was a leftover comment from  previous script. I'm removing the comment now

rotkehlchen/tests/unit/globaldb/test_globaldb_consistency.py
OjusWiZard: This is because, generally we keep adding the remote updates whenever we want to add an asset, without touching the packaged db. Then at the end just before releasing, we add all of them together in the packagedb at once (final packaging). So, we don't fail it here otherwise it will always fail and we will have to do twice PRs whenever we add an update (one extra PR each to add in the packaged db).

LefterisJP (approved):
LGTM.

PR #8216, opened 2024-07-04
Fix (Airdrops): Incorrect claim of mutiple aidrops

OjusWiZard:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=69560658

## Checklist

- [x] Fix `asset_to_protocol` mapping to return along with calculated `claimed_airdrops`

### DIFF
--- a/rotkehlchen/chain/ethereum/airdrops.py
+++ b/rotkehlchen/chain/ethereum/airdrops.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from collections import defaultdict
 from collections.abc import Callable, Sequence
@@ -43,6 +44,7 @@ AIRDROPS_REPO_BASE: Final = f'https://raw.githubusercontent.com/rotki/data/{"mai
 AIRDROPS_INDEX: Final = f'{AIRDROPS_REPO_BASE}/airdrops/index_v3.json'
 ETAG_CACHE_KEY: Final = 'ETag'
 JSON_PATH_SEPARATOR_API_AIRDROPS: Final = '/'
+AIRDROP_IDENTIFIER_KEY: Final = 'airdrop_identifier'
 
 
 @dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=False, kw_only=True)  # noqa: E501
@@ -80,6 +82,7 @@ class AirdropAPIData(Airdrop):
 
 class AirdropClaimEventQueryParams(NamedTuple):
     """Params used in the SQL query to detect as claimed an airdrop"""
+    protocol: str
     event_type: HistoryEventType
     event_subtype: HistoryEventSubType
     location_label: ChecksumEvmAddress
@@ -333,7 +336,7 @@ def get_poap_airdrop_data(airdrop_data: list[str], name: str, data_dir: Path) ->
 def calculate_claimed_airdrops(
         airdrop_data: Sequence[AirdropClaimEventQueryParams],
         database: DBHandler,
-) -> Sequence[tuple[ChecksumEvmAddress, str, FVal]]:
+) -> Sequence[tuple[ChecksumEvmAddress, str]]:
     """Calculates which of the given airdrops have been claimed.
     It does so by checking if there is a claim history event in the database
     that matches the airdrop data (address, asset and amount). It returns a list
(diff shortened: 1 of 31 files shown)


rotkehlchen/chain/ethereum/airdrops.py
```diff
@@ -334,37 +337,49 @@ def _process_json(response: Response, filename: Path) -> None:
...
+                    airdrop_info.event_type.serialize(),
+                    airdrop_info.event_subtype.serialize(),
+                    airdrop_info.location_label,
+                    airdrop_info.asset.serialize(),
+                    str(amount - half_range),
+                    str(amount + half_range),
+                ),
+            ):
+                if db_extra_data is None:
+                    continue
```
yabirgb, line 369: shouldn't we log an error message here? maybe even show a database.message_aggregator.add_warning error?

rotkehlchen/chain/ethereum/airdrops.py
```diff
@@ -334,37 +337,49 @@ def _process_json(response: Response, filename: Path) -> None:
...
+                try:
+                    extra_data = json.loads(db_extra_data)
+                except json.JSONDecodeError as e:
+                    log.error(
+                        f'Failed to read extra_data when reading EvmEvent entry '
+                        f'{db_extra_data} from the DB due to {e!s}. Skipping airdrop claim check.',
+                    )
+
+                if airdrop_info.protocol in extra_data.get(AIRDROP_IDENTIFIER_KEY):
+                    claimed_events.append((address, airdrop_info.protocol))
```
yabirgb, line 380: sholdn't the else here add an error? It is missing the key right?

yabirgb:
Failing test is a remote error `failed with HTTP status code 503 and text The service is unavailable.`

rotkehlchen/chain/evm/decoding/airdrops.py
```diff
@@ -15,6 +16,7 @@ def match_airdrop_claim(
         amount: 'FVal',
         asset: 'Asset',
         counterparty: str,
+        airdrop_identifier: str,
```
LefterisJP, line 19: Couldn't we do a `Literal['shutter', 'cornichon', ..., 'degen']` here? Would it not make sense to have the possible literals so that type checker helps us? Would help whenever it's not dynamically read

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -375,6 +375,7 @@ def _decode_normal_claim(self, context: DecoderContext) -> DecodingOutput:
                 amount=amount,
                 asset=self.vcow_token,
                 counterparty=CPT_COWSWAP,
+                airdrop_identifier='cow_mainnet' if self.evm_inquirer.chain_id == ChainID.ETHEREUM else 'cow_gnosis',  # noqa: E501
```
LefterisJP, line 378: 1. Move this out of the for.
2. Just doing (1) is probably fine. But perhaps can also instead move it to the constructor as a property of the decoder to not do the if check every time.Though it's a memory vs cpu tradeoff and perhps not worth it?

OjusWiZard: 2. it's probably okay to have the if check, because this decoder function will only execute once (only if the user ever claimed it). But, if we pass it through the constructor then it will always occupy memory

rotkehlchen/chain/ethereum/modules/shutter/decoder.py
```diff
@@ -84,6 +85,7 @@ def _decode_shutter_claim(self, context: DecoderContext) -> DecodingOutput:
             notes=f'Claim {amount} SHU from shutter airdrop into the vesting contract: {vesting_contract_address}',  # noqa: E501
             counterparty=CPT_SHUTTER,
             address=context.transaction.to_address,
+            extra_data={AIRDROP_IDENTIFIER_KEY: 'shutter'},
```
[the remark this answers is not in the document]
OjusWiZard: we use the string literal only in one place in the production code, so not adding to avoid occupying memory

LefterisJP (approved):
LGTM. Let's see how tests go

PR #8228, opened 2024-07-08
feat(frontend): use mono font for address display

lukicenturi:
Use mono font for address display, for better readibility.
Also help us to calculate the width (since every character has same width), so we can truncate the address better.

### DIFF
--- a/frontend/app/src/components/accounts/AccountGroupHeader.vue
+++ b/frontend/app/src/components/accounts/AccountGroupHeader.vue
@@ -84,10 +84,14 @@ const usdSum = computed<BigNumber>(() => balanceUsdValueSum(get(items)));
         <div class="font-medium">
           {{ t('account_group_header.xpub') }}
         </div>
-        <div :class="{ blur: !shouldShowAmount }">
+        <div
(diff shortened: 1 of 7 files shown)


frontend/app/src/components/accounts/AccountGroupHeader.vue
```diff
@@ -37,7 +37,7 @@ const label = computed<string | undefined>(() => get(xpub).label);
 const xpubTags = computed<string[] | undefined>(() => get(xpub).tags);
 
 const displayXpub = computed<string>(() =>
-  scrambleAddress(truncateAddress(get(xpub).data.xpub, truncationPoints[get(breakpoint)] ?? 4)),
+  truncateAddress(get(xpub).data.xpub, truncationPoints[get(breakpoint)] ?? 4),
```
kelsos, line 40: Doesn't that mean that the `xpub` won't be scrambled anymore?

lukicenturi: my bad, updated

kelsos (approved):
lgtm

PR #8229, opened 2024-07-08
fix(frontend): optimize balances query

lukicenturi:
Only fetch balances for a chain if there is account from that chain


yabirgb:
lol I noticed this the other day and forgot to comment. GOOOOOOOOOD

kelsos (approved):
lgtm

PR #8230, opened 2024-07-08
Fix (Globaldb): Remove ATTACH from asset search

### LINKED ISSUE #7533: Asset search, database globaldb already in use
## Problem Definition

Some times (very rarely) when trying to add a latest price for an asset and searching for the asset via the asset field search box it can throw a 500 error. In the frontend nothing  show. Just searching for assets stops working.

 If you inspect the return of the search endpoint you see 500 errors. And then in the logs you see that the error is `database globaldb is already in use`.

Seen in 1.32.1.

## Logs

```
[24/02/2024 01:08:50 CET] DEBUG rotkehlchen.api.server Greenlet-2: end rotki api OPTIONS /api/1/assets/search/levenshtein view_args={}, query_string=b'', status_code=200, result=None
[24/02/2024 01:08:50 CET] DEBUG rotkehlchen.api.server Greenlet-2: start rotki api POST /api/1/assets/search/levenshtein view_args={}, query_string=b''
[24/02/2024 01:08:50 CET] CRITICAL rotkehlchen.api.server Greenlet-2: Unhandled exception when processing endpoint request exc_info=True, exception=database globaldb is already in use
[24/02/2024 01:08:50 CET] DEBUG rotkehlchen.api.server Greenlet-2: end rotki api POST /api/1/assets/search/levenshtein view_args={}, query_string=b'', status_code=500, result={'result': None, 'message': 'database globaldb is already in use'}
```

### System Description

Operating system: Linux
Rotki version: 1.32.1

OjusWiZard:
Closes #7533 and https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=69334133

## Checklist

- [x] Remove ATTACH from `retrieve_assets`, `search_assets`, and `_search_only_assets_levenstein`

### DIFF
--- a/rotkehlchen/db/filtering.py
+++ b/rotkehlchen/db/filtering.py
@@ -1495,6 +1495,8 @@ class AddressbookFilterQuery(DBFilterQuery):
 
 
 class AssetsFilterQuery(DBFilterQuery):
+    ignored_assets_handling: IgnoredAssetsHandling = IgnoredAssetsHandling.NONE
+
     @classmethod
     def make(
             cls: type['AssetsFilterQuery'],
@@ -1526,6 +1528,7 @@ class AssetsFilterQuery(DBFilterQuery):
             order_by_rules=order_by_rules,
             order_by_case_sensitive=False,
         )
+        filter_query.ignored_assets_handling = ignored_assets_handling
         filters: list[DBFilter] = []
         if name is not None:
             filters.append(DBSubStringFilter(
@@ -1572,12 +1575,6 @@ class AssetsFilterQuery(DBFilterQuery):
                 select_value='asset_id',
                 select_table='user_owned_assets',
                 select_condition=None,
-            ))
-        if ignored_assets_handling is not IgnoredAssetsHandling.NONE:
-            filters.append(DBIgnoredAssetsFilter(
-                and_op=True,
-                asset_key=identifier_column_name,
-                operator=ignored_assets_handling.operator(),
             ))
         if chain_id is not None:
             filters.append(DBEqualsFilter(
@@ -1815,6 +1812,7 @@ class LevenshteinFilterQuery(MultiTableFilterQuery):
     Is used for querying both assets and nfts.
     """
     substring_search: str
+    ignored_assets_handling: IgnoredAssetsHandling = IgnoredAssetsHandling.NONE
 
     @classmethod
     def make(
@@ -1829,6 +1827,7 @@ class LevenshteinFilterQuery(MultiTableFilterQuery):
             filters=[],
             substring_search=substring_search,
         )
+        filter_query.ignored_assets_handling = ignored_assets_handling
         filters: list[tuple[DBFilter, str]] = []  # filter + table name for which to use it.
 
         name_filter = DBSubStringFilter(
@@ -1868,14 +1867,6 @@ class LevenshteinFilterQuery(MultiTableFilterQuery):
                 value=chain_id.serialize_for_db(),
             )
             filters.append((new_filter, 'assets'))
-
-        if ignored_assets_handling is not IgnoredAssetsHandling.NONE:
-            ignored_assets_filter = DBIgnoredAssetsFilter(
-                and_op=True,
-                asset_key='assets.identifier',
-                operator=ignored_assets_handling.operator(),
-            )
-            filters.append((ignored_assets_filter, 'assets'))
 
         filter_query.filters = filters
         return filter_query
--- a/rotkehlchen/db/search_assets.py
+++ b/rotkehlchen/db/search_assets.py
@@ -45,61 +45,58 @@ def _search_only_nfts_levenstein(
 
 
 def _search_only_assets_levenstein(
-        cursor: 'DBCursor',
-        db: 'DBHandler',
+        userdb_cursor: 'DBCursor',
+        userdb: 'DBHandler',
         filter_query: 'LevenshteinFilterQuery',
 ) -> list[tuple[int, dict[str, Any]]]:
     search_result: list[tuple[int, dict[str, Any]]] = []
-    resolved_eth = A_ETH.resolve_to_crypto_asset()
-    globaldb = GlobalDBHandler()
-    treat_eth2_as_eth = db.get_settings(cursor).treat_eth2_as_eth
-    globaldb.conn.execute('PRAGMA wal_checkpoint;')  # needed to have up to date information when attaching  # noqa: E501
-    with db.conn.critical_section():  # needed due to ATTACH. Must not context switch out of this
-        cursor.execute(
-            f'ATTACH DATABASE "{globaldb.filepath()!s}" AS globaldb KEY "";',
-        )
-        try:
-            query, bindings = filter_query.prepare('assets')
-            query = ALL_ASSETS_TABLES_QUERY.format(dbprefix='globaldb.') + query
-            cursor.execute(query, bindings)
-            found_eth = False
-            for entry in cursor:
-                lev_dist_min = 100
-                if entry[1] is not None:
-                    lev_dist_min = min(
-                        lev_dist_min,
-                        levenshtein(filter_query.substring_search, entry[1].casefold()),
-                    )
-                if entry[2] is not None:
-                    lev_dist_min = min(
-                        lev_dist_min,
-                        levenshtein(filter_query.substring_search, entry[2].casefold()),
-                    )
-                if treat_eth2_as_eth is True and entry[0] in (A_ETH.identifier, A_ETH2.identifier):
-                    if found_eth is False:
-                        search_result.append((lev_dist_min, {
-                            'identifier': resolved_eth.identifier,
-                            'name': resolved_eth.name,
-                            'symbol': resolved_eth.symbol,
-                            'asset_type': AssetType.OWN_CHAIN.serialize(),
-                        }))
-                        found_eth = True
-                    continue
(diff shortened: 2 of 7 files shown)


yabirgb:
I suspect this PR also closes https://github.com/rotki/rotki/issues/7533

rotkehlchen/db/search_assets.py
```diff
@@ -45,61 +45,58 @@ def _search_only_nfts_levenstein(
...
-        finally:
-            cursor.execute('DETACH globaldb;')
+    ignored_asset_handler = filter_query.ignored_assets_handling.get_handler()
+    with db.conn.read_ctx() as cursor:
+        treat_eth2_as_eth = db.get_settings(cursor).treat_eth2_as_eth
+        ignored_assets = db.get_ignored_asset_ids(cursor)
+
+    with GlobalDBHandler().conn.read_ctx() as cursor:
+        query, bindings = filter_query.prepare('assets')
+        query = ALL_ASSETS_TABLES_QUERY.format(dbprefix='') + query
```
yabirgb, line 59: you can remove the dbprefix part of the queries if I'm not mistaken and also the format then

rotkehlchen/db/search_assets.py
```diff
@@ -45,61 +45,58 @@ def _search_only_nfts_levenstein(
...
-                if entry[5] is not None:
-                    entry_info['custom_asset_type'] = entry[5]
-
-                search_result.append((lev_dist_min, entry_info))
-        finally:
-            cursor.execute('DETACH globaldb;')
+    ignored_asset_handler = filter_query.ignored_assets_handling.get_handler()
+    with db.conn.read_ctx() as cursor:
+        treat_eth2_as_eth = db.get_settings(cursor).treat_eth2_as_eth
+        ignored_assets = db.get_ignored_asset_ids(cursor)
```
yabirgb, line 55: if you check where this function is called we already have an open cursor there, the one that you removed from the args, instead of opening one again and creating a new context just reuse it

rotkehlchen/db/search_assets.py
```diff
@@ -114,7 +111,6 @@ def search_assets_levenshtein(
     search_result = []
     with db.conn.read_ctx() as cursor:
         search_result = _search_only_assets_levenstein(
-            cursor=cursor,
```
yabirgb, line 117: this is the only place where it is kept. Don't remove the argumetn

rotkehlchen/globaldb/handler.py
```diff
@@ -487,51 +485,47 @@ def get_assets_mappings(identifiers: list[str]) -> tuple[dict[str, dict], dict[s
...
-                    if entry[5] is not None:
-                        entry_info['custom_asset_type'] = entry[5]
-
-                    search_result.append(entry_info)
-            finally:
-                cursor.execute('DETACH DATABASE globaldb;')
+            ignored_assets = db.get_ignored_asset_ids(cursor)
+
+        query, bindings = filter_query.prepare(without_ignored_asset_filter=True)
+        query = ALL_ASSETS_TABLES_QUERY.format(dbprefix='') + query
```
yabirgb, line 497: this is the other place where I beleive is used. REmove it

rotkehlchen/globaldb/handler.py
```diff
@@ -487,51 +485,47 @@ def get_assets_mappings(identifiers: list[str]) -> tuple[dict[str, dict], dict[s
...
-    ) -> list[dict[str, Any]]:
+    def search_assets(filter_query: 'AssetsFilterQuery', db: 'DBHandler') -> list[dict[str, Any]]:
         """Returns a list of asset details that match the search query provided."""
         search_result = []
-        globaldb = GlobalDBHandler()
-        query, bindings = filter_query.prepare()
-        query = ALL_ASSETS_TABLES_QUERY.format(dbprefix='globaldb.') + query
-        resolved_eth = A_ETH.resolve_to_crypto_asset()
-        globaldb._wal_checkpoint()
+        ignored_asset_handler = filter_query.ignored_assets_handling.get_handler()
```
yabirgb, line 491: this is hard to read because you later do a `not ignored_asset_handler` that makes it complex to understand. Can you maybe find a better name to it and maybe change the logic so you don't have to do a `not logic`?

rotkehlchen/db/filtering.py
```diff
@@ -1815,6 +1812,7 @@ class LevenshteinFilterQuery(MultiTableFilterQuery):
     Is used for querying both assets and nfts.
     """
     substring_search: str
+    ignored_assets_handling: IgnoredAssetsHandling = IgnoredAssetsHandling.NONE
```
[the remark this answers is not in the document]
OjusWiZard: yeah, it's used in `_search_only_assets_levenstein`

rotkehlchen/db/search_assets.py
```diff
@@ -45,61 +45,58 @@ def _search_only_nfts_levenstein(
...
+    treat_eth2_as_eth = db.get_settings(db_cursor).treat_eth2_as_eth
+    ignored_assets = db.get_ignored_asset_ids(db_cursor)
+
+    with GlobalDBHandler().conn.read_ctx() as cursor:
+        query, bindings = filter_query.prepare('assets')
+        query = ALL_ASSETS_TABLES_QUERY + query
+        cursor.execute(query, bindings)
+        found_eth = False
+        for entry in cursor:
+            if should_skip(entry[0], ignored_assets):
```
LefterisJP, line 63: This is a really slow way of doing things. Loading all ignored assets in memory and then doing the check in here for each loop iteration is costly.

It's done to avoid the attaching I suppose right? Can you add a todo comment/note here? Something like

```
# We knowingly load the entired ignored assets in memory here since it's from a different database and
# need to avoid attaching detaching user DB here due to: LINKISSUE1, LINKISSUE2

File: rotkehlchen/db/search_assets.py [hunk 8]
```diff
@@ -45,61 +45,58 @@ def _search_only_nfts_levenstein(
 
 
 def _search_only_assets_levenstein(
-        cursor: 'DBCursor',
+        db_cursor: 'DBCursor',
```
LefterisJP, line 48: still not a good name. Rename to `userdb_cursor`.

db cursor can be any.

rotkehlchen/globaldb/handler.py
```diff
@@ -381,71 +381,69 @@ def retrieve_assets(userdb: 'DBHandler', filter_query: 'AssetsFilterQuery') -> t
...
+            ignored_assets = userdb.get_ignored_asset_ids(cursor)
+
+        with GlobalDBHandler().conn.read_ctx() as cursor:
+            # get all underlying tokens
+            for entry in cursor.execute(underlying_tokens_query, bindings):
+                underlying_tokens[entry[0]].append(UnderlyingToken.deserialize_from_db((entry[1], entry[2], entry[3])).serialize())  # noqa: E501
+
+            cursor.execute(query, bindings)
+            for entry in cursor:
+                if should_skip(entry[0], ignored_assets):
```
LefterisJP, line 399: maybe same comment here? Or hmmm perhaps find a more nice central single point to write this comment? I will 100% definitely come in the future and ask why are we doing it this way if there is no comment. Also a TODO to maybe improve it (can't think of how right now, late at night)

rotkehlchen/globaldb/handler.py
```diff
@@ -487,51 +485,47 @@ def get_assets_mappings(identifiers: list[str]) -> tuple[dict[str, dict], dict[s
...
-                cursor.execute('DETACH DATABASE globaldb;')
+            ignored_assets = db.get_ignored_asset_ids(cursor)
+
+        query, bindings = filter_query.prepare(without_ignored_asset_filter=True)
+        query = ALL_ASSETS_TABLES_QUERY + query
+        with GlobalDBHandler().conn.read_ctx() as cursor:
+            cursor.execute(query, bindings)
+            found_eth = False
+            for entry in cursor:
+                if should_skip(entry[0], ignored_assets):
```
LefterisJP, line 502: yeah same here. So perhaps in the definition of should_ship put the comment and TODO? not sure.

rotkehlchen/globaldb/handler.py
```diff
@@ -487,51 +485,47 @@ def get_assets_mappings(identifiers: list[str]) -> tuple[dict[str, dict], dict[s
...
+                    continue
+
+                if treat_eth2_as_eth is True and entry[0] in (A_ETH.identifier, A_ETH2.identifier):
```
LefterisJP, line 505: Also isn't; there a lot of common functionality in the parts of the code you touched that can be abstracted? Or is it just me?

LefterisJP (commented):
Left some comments.

1. Better explanation on what is being done in the code. I will definitely 100% come back and ask why its done like that and not in some other way that's more efficient.
2. As i wrote in the comments add a TODO and let's think how to improve this.
3. Can't you add a unit test for this? I don't see a test reproducing any of those problems this PR is supposed to fix. Is it super hard to add one? If not then add one. I was given the impression you had one. These are useful for regression purposes.
4. Perhaps it's just my tired eyes late at night but seems like the code you touched can be abstracted a lot?

rotkehlchen/globaldb/handler.py
OjusWiZard: so they seems similar because only these are common in all the functions: ignored_assets and should_skip that I already abstracted. Other things are slightly different in all of them.

LefterisJP (approved):
LGTM. Failing test `test_query_historical_dsr_with_a_zero_withdrawal` needs to be VCRed as it's a remote errors

PR #8232, opened 2024-07-09
Use default nodes when querying logs

yabirgb:
By default we were using only owned nodes and etherscan and this was triggering a rate limit in gnosiscan when querying withdrawal logs

Closes #(issue_number)

## Checklist


LefterisJP (approved):
LGTM. `test_query_token_with_info` needs to be VCRed.

PR #8241, opened 2024-07-09
Fix HTX test_trades

LefterisJP:
Since HTX can't query trades more than 120 days in the past then the test needs to have its time frozen.


LefterisJP:
all green

PR #8243, opened 2024-07-10
Chore (GlobalDB): Package for v1.34

OjusWiZard:
## Checklist

- [x] Add asset and remote updates into packaged globaldb for release v1.34
- [x] Fix a bug in asset updates where the foreign key entries are removed when an asset update conflict is resolved through 'remote' option
 
bug is that we have an asset A and B
1. we add an underlying asset mapping A -> B, in update v24
2. then we update the asset A in v25 that results in a conflict
3. we resolve the conflict with 'remote' option
4. after forcing remote data in the update, the underlying asset mapping A -> B is gone

this is because when we force the remote data in update, we first remove the asset, and then apply the full insert query 
when asset is removed, the underlying asset entry is also removed because of foreign key, and is not added by in the full_insert because it was not that tokens underlying asset 
it was for B

Update log: [File here](https://cdn.discordapp.com/attachments/524705301644050467/1260530380377821215/update.log?ex=668fa7e6&is=668e5666&hm=4c3b5476720e211ce9c43015c48f7d5bd44c08f55e829dd4adc5a9fb2d02f7af&)


rotkehlchen/tests/unit/decoders/test_aave_v3.py
```diff
@@ -1205,7 +1205,7 @@ def test_aave_v3_withdraw_xdai(database, gnosis_inquirer, gnosis_accounts) -> No
             event_type=HistoryEventType.SPEND,
             event_subtype=HistoryEventSubType.RETURN_WRAPPED,
             asset=Asset('eip155:100/erc20:0xd0Dd6cEF72143E22cCED4867eb0d5F2328715533'),
-            balance=Balance(amount=FVal(returned_amount)),  # TODO: This is a bug and should be withdrawn_amount but the decimals need to be edited in the assets repo. OJ is on that  # noqa: E501
+            balance=Balance(amount=FVal(returned_amount)),
```
yabirgb, line 1208:
```suggestion
            balance=Balance(amount=FVal(withdrawn_amount)),
```

rotkehlchen/tests/unit/decoders/test_aave_v3.py
```diff
@@ -1145,7 +1145,7 @@ def test_aave_v3_withdraw_xdai(database, gnosis_inquirer, gnosis_accounts) -> No
         database=database,
         tx_hash=tx_hash,
     )
-    timestamp, gained_amount, withdrawn_amount, gas_fees, gateway_address, approval_amount, returned_amount = TimestampMS(1720459795000), '0.0000000000000000000000000000000000000000000000000000000000000000000000000000000000076355892637370336', '5.076355892637370336', '0.0008300288', string_to_evm_address('0xfE76366A986B72c3f2923e05E6ba07b7de5401e4'), FVal('0.0000000000000000000000115792089237316195423570985008687907853269984665640564039452507652020492269599'), '0.0000000000000000000000000000000000000000000000000000000000000000000000000000000005076355892637370336'  # noqa: E501
+    timestamp, gained_amount, withdrawn_amount, gas_fees, gateway_address, approval_amount, returned_amount = TimestampMS(1720459795000), '0.076355892637370336', '5.076355892637370336', '0.0008300288', string_to_evm_address('0xfE76366A986B72c3f2923e05E6ba07b7de5401e4'), FVal('115792089237316195423570985008687907853269984665640564039452.507652020492269599'), '5.076355892637370336'  # noqa: E501
```
yabirgb, line 1148:
```suggestion
    timestamp, gained_amount, withdrawn_amount, gas_fees, gateway_address, approval_amount = TimestampMS(1720459795000), '0.076355892637370336', '5.076355892637370336', '0.0008300288', string_to_evm_address('0xfE76366A986B72c3f2923e05E6ba07b7de5401e4'), FVal('115792089237316195423570985008687907853269984665640564039452.507652020492269599')  # noqa: E501
```

rotkehlchen/tests/unit/test_inquirer.py
```diff
@@ -435,11 +435,11 @@ def test_find_curve_lp_token_price(inquirer: 'Inquirer', blockchain: 'ChainsAggr
         # Curve.fi amDAI/amUSDC/amUSDT (am3CRV)
         ChainID.POLYGON_POS: ('0xE7a24EF0C5e95Ffb0f6684b813A78F2a3AD7D171', FVal('1.129260')),
         # crvUSDT-gauge
-        ChainID.ARBITRUM_ONE: ('0xB08FEf57bFcc5f7bF0EF69C0c090849d497C8F8A', FVal('2.024704')),
+        ChainID.ARBITRUM_ONE: ('0xB08FEf57bFcc5f7bF0EF69C0c090849d497C8F8A', FVal('2.025168')),
```
yabirgb, line 438: why did this test changed?

rotkehlchen/tests/unit/globaldb/test_asset_updates.py
```diff
@@ -512,6 +513,16 @@ def test_conflict_updates(assets_updater: AssetsUpdater, globaldb: GlobalDBHandl
     """Test that the logic doesn't add duplicates for assets that were inserted twice
     in the globaldb assets updates
     """
+    with globaldb.conn.write_ctx() as write_cursor:
+        assert write_cursor.execute(
+            'SELECT COUNT(*) FROM underlying_tokens_list WHERE parent_token_entry=?;',
+            ('eip155:42161/erc20:0xA5EDBDD9646f8dFF606d7448e414884C7d905dCA',),
+        ).fetchone()[0] > 0
```
yabirgb, line 520: since you know exactly how many set them in a variable and check that the amount didn't change

yabirgb: Also add in the docstring that you test that remote errors don't alter the underlying tokens/multiasset mappinggs and this is a regression test for `_force_remote_asset`

rotkehlchen/globaldb/updates.py
```diff
@@ -93,13 +93,37 @@ def _force_remote_asset(cursor: DBCursor, local_asset: Asset, full_insert: str)
...
+        cursor.executemany(  # add the old multiasset mappings
+            'INSERT INTO multiasset_mappings (collection_id, asset) VALUES (?, ?);',
+            multiasset_mappings,
+        )
+    if len(underlying_assets) > 0:
+        cursor.executemany(  # add the old underlying assets
+            'INSERT INTO underlying_tokens_list (parent_token_entry, weight, identifier) '
+            'VALUES (?, ?, ?);',
+            underlying_assets,
+        )
```
yabirgb, line 125: what happens with errors here? Should you just catch them and log an error? This is somehow critical and you can't have issues with those. What happens if this section fails?

yabirgb: It is handled already, marking as resolved

rotkehlchen/tests/unit/test_inquirer.py
OjusWiZard: This is because earlier we did not have coingecko/cryptocompare id for one of its lp token `eip155:100/erc20:0xaBEf652195F98A91E490f047A5006B71c85f058d`, so they were getting queried from defillama. Now, they get queried from coingecko/cryptocompare, so price is slightly varied.

rotkehlchen/assets/utils.py
```diff
@@ -371,7 +371,15 @@ def get_crypto_asset_by_symbol(
...
     )
-    if len(assets_data) != 1:
+    if len(assets_data) < 1:
+        return None
+
+    if len(assets_data) > 1:
+        # check if all the assets are part of the same asset_collection
+        multiassets = set(GlobalDBHandler.get_assets_in_same_collection(assets_data[0].identifier))
+        if len(set(assets_data).difference(multiassets)) == 0:
+            return assets_data[0]
```
yabirgb, line 381: this is bad because if for some reason the asset order changed in the db it will affect the events and it can lead to errors. I believe you added it due to the tron issue right? YOu can add a mapping for all the locations for this case

yabirgb (approved):
lgtm

PR #8247, opened 2024-07-10
fix: show xpub even if no addresses inside

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,8 @@
 Changelog
 =========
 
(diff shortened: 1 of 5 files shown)


kelsos (approved):
looks good on my side

lukicenturi:
I didn't see that the xpub only data is also inside the `balances`

kelsos:
> I didn't see that the xpub only data is also inside the `balances`

yes after the refactor they where there the only problem is that the header required items to show.

PR #8248, opened 2024-07-10
Fix price conversion when querying TCR for liquity

yabirgb:
When the price had high precission the FVal couldn't be converted to int exactly and this was raising an unhandled error. We don't need such procession here
## Checklist

### DIFF
--- a/rotkehlchen/tests/unit/test_stats.py
+++ b/rotkehlchen/tests/unit/test_stats.py
@@ -108,7 +108,7 @@ def test_compound_events_stats(
 
 @pytest.mark.vcr(filter_query_parameters=['apikey'])
 @pytest.mark.parametrize('ethereum_accounts', [TEST_ACCOUNTS])
-def test_liquity_tcr(ethereum_inquirer: 'EthereumInquirer'):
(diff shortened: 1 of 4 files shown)


rotkehlchen/tests/unit/test_stats.py
```diff
@@ -103,3 +104,18 @@ def test_compound_events_stats(
     # Check rewards mappings
     rewards_0 = stats['rewards']['0x65304d6aff5096472519ca86a6a1fea31cb47Ced']
     assert rewards_0[A_COMP].amount == FVal('0.06366460157923238')
+
+
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('ethereum_accounts', [TEST_ACCOUNTS])
+def test_liquity_tcr(ethereum_inquirer: 'EthereumInquirer'):
```
LefterisJP, line 111: nitpick, it's not the TCR bu the having a non-exact price

```suggestion
def test_liquity_tcr_non_exact_int(ethereum_inquirer: 'EthereumInquirer'):
```

LefterisJP (approved):
LGTM

PR #8268, opened 2024-07-15
Bump ruff to 0.5.2

LefterisJP:
- Removes parentheses from fixtures with no args
- S603 different line
- Enforce using .items() to access value in dict iteration (https://docs.astral.sh/ruff/rules/dict-index-missing-items/)


LefterisJP:
Two tests failed with etherscan 502.

Should be VCRed.

```
rotkehlchen/tests/unit/test_tokens.py::test_last_queried_ts
rotkehlchen/tests/unit/globaldb/test_globaldb.py::test_for_spam_tokens
```

PR #8272, opened 2024-07-15
Fix (Hop): Price of LP tokens

### LINKED ISSUE #8262: Hop LP balances not detected for Base - Price
## Problem Definition

With 1.34.0 we finally added HOP-LP detection of balances. Staked or unstaked. I have an address that does not get its HOP LPing detected in Base. Suspect something is off.

EDIT: The price is not detected. The LP balance is there.
Image

## Task

Fix it and write a test for it.

OjusWiZard:
Closes #8262 

## Checklist

- [x] Calculate the prices for HOP LP tokens

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`8262` Prices of HOP LP tokens will now properly show up for all pools.
 * :bug:`8261` Windows users will no longer be greeted with an unhandled exception dialog when closing the application.
 * :bug:`8263` Fixes manual pagination not working.
 
--- a/rotkehlchen/chain/evm/decoding/hop/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/hop/decoder.py
@@ -2,7 +2,8 @@ import logging
 from typing import TYPE_CHECKING, Any
 
 from rotkehlchen.accounting.structures.balance import Balance
-from rotkehlchen.assets.asset import Asset
+from rotkehlchen.assets.asset import Asset, EvmToken
+from rotkehlchen.assets.resolver import AssetResolver
 from rotkehlchen.chain.ethereum.utils import (
     token_normalized_value,
     token_normalized_value_decimals,
@@ -22,12 +23,18 @@ from rotkehlchen.chain.evm.decoding.utils import maybe_reshuffle_events
 from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.constants.assets import A_ETH
 from rotkehlchen.constants.misc import ZERO
+from rotkehlchen.errors.asset import UnknownAsset, WrongAssetType
 from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.fval import FVal
+from rotkehlchen.globaldb.cache import (
+    globaldb_get_unique_cache_value,
+    globaldb_set_unique_cache_value,
+)
+from rotkehlchen.globaldb.handler import GlobalDBHandler
 from rotkehlchen.history.events.structures.evm_event import EvmEvent, EvmProduct
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.logging import RotkehlchenLogsAdapter
-from rotkehlchen.types import ChainID, ChecksumEvmAddress
+from rotkehlchen.types import HOP_PROTOCOL_LP, CacheType, ChainID, ChecksumEvmAddress
 from rotkehlchen.utils.misc import hex_or_bytes_to_address, hex_or_bytes_to_int
 
 from .constants import (
@@ -106,6 +113,36 @@ class HopCommonDecoder(DecoderInterface):
             target_str = f'at address {recipient} '
 
         return f'Bridge {amount} {asset.symbol_or_name()} {chain_label}{target_str}via Hop protocol'  # noqa: E501
+
+    def _process_hop_lp_token(self, lp_token: EvmToken, pool_address: ChecksumEvmAddress) -> None:
+        """Save the protocol value of the LP token and cache its pool address."""
+        with GlobalDBHandler().conn.write_ctx() as write_cursor:
+            # Save the protocol value of the lp token if needed
+            if lp_token.protocol is None:
+                write_cursor.execute(
+                    'UPDATE evm_tokens SET protocol = ? WHERE identifier = ?;',
+                    (HOP_PROTOCOL_LP, lp_token.identifier),
+                )
+                AssetResolver.clean_memory_cache(lp_token.identifier)
+
+            # Cache the pool address if needed
+            if globaldb_get_unique_cache_value(
+                cursor=write_cursor,
+                key_parts=(
+                    CacheType.HOP_POOL_ADDRESS,
+                    str(lp_token.chain_id.value),
+                    lp_token.evm_address,
+                ),
+            ) is None:
+                globaldb_set_unique_cache_value(
+                    write_cursor=write_cursor,
+                    key_parts=(
+                        CacheType.HOP_POOL_ADDRESS,
+                        str(lp_token.chain_id.value),
+                        lp_token.evm_address,
+                    ),
+                    value=pool_address,
+                )
 
     def _decode_withdrawal_bonded(self, context: DecoderContext) -> DecodingOutput:
         """This function is used to decode the WithdrawalBonded events on Hop protocol."""
@@ -340,6 +377,13 @@ class HopCommonDecoder(DecoderInterface):
                 event.counterparty = CPT_HOP
                 event.notes = f'Receive {event.balance.amount} {event.asset.symbol_or_name()} after providing liquidity in Hop'  # noqa: E501
                 in_event = event
+                try:
+                    self._process_hop_lp_token(
+                        lp_token=event.asset.resolve_to_evm_token(),
+                        pool_address=context.tx_log.address,
+                    )
+                except (WrongAssetType, UnknownAsset) as e:
+                    log.error(f'Could not resolve {event.asset!s} in {self.evm_inquirer.chain_name} while decoding AddLiquidity in Hop: {e!s}')  # noqa: E501
 
         maybe_reshuffle_events(
             ordered_events=[out_event1, out_event2, in_event],
--- a/rotkehlchen/inquirer.py
+++ b/rotkehlchen/inquirer.py
@@ -93,6 +93,7 @@ from rotkehlchen.serialization.deserialize import deserialize_evm_address
 from rotkehlchen.types import (
     CURVE_POOL_PROTOCOL,
     GEARBOX_PROTOCOL,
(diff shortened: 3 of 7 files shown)


rotkehlchen/chain/evm/decoding/hop/decoder.py
```diff
@@ -333,13 +342,77 @@ def _decode_add_liquidity(self, context: DecoderContext) -> DecodingOutput:
...
+                        abi=self.evm_inquirer.contracts.abi('HOP_POOL'),
+                        deployed_block=0,  # not used here
+                    )
+                    index, asset_addresses = 0, []
+                    try:
+                        while (response := contract.call(
+                                node_inquirer=self.evm_inquirer,
+                                method_name='getToken',
+                                arguments=[index],
+                        )) is not None:
```
yabirgb, line 388: this does two calls at least can raise remoteERror. YOu should handle it

rotkehlchen/chain/evm/decoding/hop/decoder.py
```diff
@@ -333,13 +342,77 @@ def _decode_add_liquidity(self, context: DecoderContext) -> DecodingOutput:
...
                 event.event_type = HistoryEventType.RECEIVE
                 event.event_subtype = HistoryEventSubType.RECEIVE_WRAPPED
                 event.counterparty = CPT_HOP
                 event.notes = f'Receive {event.balance.amount} {event.asset.symbol_or_name()} after providing liquidity in Hop'  # noqa: E501
                 in_event = event
+                lp_token = event.asset.resolve_to_evm_token()
+                GlobalDBHandler.edit_evm_token(
+                    entry=EvmToken.initialize(
+                        address=lp_token.evm_address,
+                        chain_id=self.evm_inquirer.chain_id,
```
yabirgb, line 357: All this new logic can you move it to a method for populating the pool in the database? I believe it would benefict from having a proper docstring and a good function name with the errors that it can raise

rotkehlchen/chain/evm/decoding/hop/decoder.py
```diff
@@ -333,13 +342,77 @@ def _decode_add_liquidity(self, context: DecoderContext) -> DecodingOutput:
...
+                                node_inquirer=self.evm_inquirer,
+                                method_name='getToken',
+                                arguments=[index],
+                        )) is not None:
+                            index += 1
+                            asset_addresses.append(response)
+                    except RemoteError:
+                        # exit out of loop as soon as an error is encountered
+                        # this means no more assets are left to query
+                        ...
```
yabirgb, line 394: if you don't want to do anything you can do a supress(RemoteError)

rotkehlchen/chain/evm/decoding/hop/decoder.py
yabirgb, line 388: there is no way to know how many tokens are right? I need at least 2 calls for that and I believe that is what you do on average

rotkehlchen/inquirer.py
```diff
@@ -1094,6 +1101,63 @@ def find_yearn_price(
...
+
+        contract = EvmContract(
+            address=string_to_evm_address(amm_address),
+            abi=evm_manager.node_inquirer.contracts.abi('HOP_POOL'),
+            deployed_block=0,
+        )
+        pool_supply = ZERO
+        for idx, _ in enumerate(pool_assets):
+            pool_supply += Price(contract.call(
+                node_inquirer=evm_manager.node_inquirer,
```
yabirgb, line 1154: it isn't a price right? it is a FVal

rotkehlchen/chain/evm/decoding/hop/decoder.py
```diff
@@ -333,13 +342,77 @@ def _decode_add_liquidity(self, context: DecoderContext) -> DecodingOutput:
...
+                    )
+                    index, asset_addresses = 0, []
+                    try:
+                        while (response := contract.call(
+                                node_inquirer=self.evm_inquirer,
+                                method_name='getToken',
+                                arguments=[index],
+                        )) is not None:
+                            index += 1
+                            asset_addresses.append(response)
```
yabirgb, line 390: I would do a to_checksum_address to prevent errors if for some reason they don't come checksummed

rotkehlchen/chain/evm/decoding/hop/decoder.py
```diff
@@ -333,13 +342,77 @@ def _decode_add_liquidity(self, context: DecoderContext) -> DecodingOutput:
...
+                        deployed_block=0,  # not used here
+                    )
+                    index, asset_addresses = 0, []
+                    try:
+                        while (response := contract.call(
+                                node_inquirer=self.evm_inquirer,
+                                method_name='getToken',
+                                arguments=[index],
+                        )) is not None:
+                            index += 1
```
yabirgb, line 389: do you need to keep track of the index variable? You can do len(asset_addresses)

rotkehlchen/chain/evm/decoding/hop/decoder.py
OjusWiZard: yes right

rotkehlchen/chain/evm/decoding/hop/decoder.py
```diff
@@ -107,6 +112,44 @@ def _generate_bridge_note(
...
+                entry=EvmToken.initialize(
+                    address=lp_token.evm_address,
+                    chain_id=self.evm_inquirer.chain_id,
+                    token_kind=EvmTokenKind.ERC20,
+                    name=lp_token.name,
+                    symbol=lp_token.symbol,
+                    decimals=lp_token.decimals,
+                    protocol=HOP_PROTOCOL_LP,
+                ),
+            )
```
yabirgb, line 128: I just saw that the logic here is very complicated and error prone. You are initializating the lp token and you might or might not be missing the underlying tokem, then calling trhe edit evm token that are 3 calls at least only to change a value and so on. I would just do a single sql query

```
UPDATE evm_tokens SET protocol=? WHERE identifier=?
```

with `(HOP_PROTOCOL_LP, lp_token.identifier)`. Maybe for the sake of not having logic split everywhere you could create a method that does that + the cache cleaning

rotkehlchen/inquirer.py
```diff
@@ -1094,6 +1097,51 @@ def find_yearn_price(
...
+            ))) * FVal(contract.call(
+                node_inquirer=evm_manager.node_inquirer,
+                method_name='getVirtualPrice',
+            )) / 10 ** lp_token.get_decimals()
+        except DeserializationError as e:
+            log.error(f'Failed to deserialize pool token address of {amm_address!s}. {e!s}')
+            return ZERO_PRICE
+        except RemoteError as e:
+            log.error(f'Failed to query virtual price of {lp_token.evm_address!s}. {e!s}')
+            return ZERO_PRICE
```
yabirgb, line 1143: include in the errors that is a hop protocol token

rotkehlchen/chain/evm/decoding/hop/decoder.py
```diff
@@ -333,13 +369,18 @@ def _decode_add_liquidity(self, context: DecoderContext) -> DecodingOutput:
             elif (
                 event.event_type == HistoryEventType.RECEIVE and
                 event.event_subtype == HistoryEventSubType.NONE and
-                event.location_label == user_address
+                event.location_label == user_address and
+                event.asset.is_evm_token()
```
LefterisJP, line 373: why this check?

rotkehlchen/inquirer.py
```diff
@@ -1094,6 +1097,51 @@ def find_yearn_price(
...
+                    arguments=[0],
+                )),
+                chain_id=lp_token.chain_id,
+                token_type=EvmTokenKind.ERC20,
+            ))) * FVal(contract.call(
+                node_inquirer=evm_manager.node_inquirer,
+                method_name='getVirtualPrice',
+            )) / 10 ** lp_token.get_decimals()
+        except DeserializationError as e:
+            log.error(f'Failed to deserialize Hop pool token address of {amm_address!s}. {e!s}')
```
LefterisJP, line 1139: need to add the chain here

rotkehlchen/inquirer.py
```diff
@@ -1094,6 +1097,51 @@ def find_yearn_price(
...
+            return ZERO_PRICE
+        except RemoteError as e:
+            log.error(f'Failed to query virtual price of Hop lp token {lp_token.evm_address!s}. {e!s}')  # noqa: E501
```
LefterisJP, line 1142: same here

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`8262` Prices of HOP LP tokens will now show accurately.
```
```suggestion
* :bug:`8262` Prices of HOP LP tokens will now properly show up for all pools.
```

rotkehlchen/chain/evm/decoding/hop/decoder.py
OjusWiZard: this was because we `resolve_to_evm_token()` below, so to ensure it doesn't raise any error. But I'm removing it now and try/catching the error

rotkehlchen/chain/evm/decoding/hop/decoder.py
```diff
@@ -340,6 +377,13 @@ def _decode_add_liquidity(self, context: DecoderContext) -> DecodingOutput:
                 event.counterparty = CPT_HOP
                 event.notes = f'Receive {event.balance.amount} {event.asset.symbol_or_name()} after providing liquidity in Hop'  # noqa: E501
                 in_event = event
+                try:
+                    self._process_hop_lp_token(
+                        lp_token=event.asset.resolve_to_evm_token(),
+                        pool_address=context.tx_log.address,
+                    )
+                except (WrongAssetType, UnknownAsset) as e:
+                    log.error(f'Could not resolve {event.asset!s} while decoding AddLiquidity in Hop: {e!s}')  # noqa: E501
```
LefterisJP, line 386: need chain here

rotkehlchen/inquirer.py
```diff
@@ -1094,6 +1097,51 @@ def find_yearn_price(
...
+                token_type=EvmTokenKind.ERC20,
+            ))) * FVal(contract.call(
+                node_inquirer=evm_manager.node_inquirer,
+                method_name='getVirtualPrice',
+            )) / 10 ** lp_token.get_decimals()
+        except DeserializationError as e:
+            log.error(f'Failed to deserialize Hop pool token address of {amm_address!s} in {lp_token.chain_id!s}. {e!s}')  # noqa: E501
+            return ZERO_PRICE
+        except RemoteError as e:
+            log.error(f'Failed to query virtual price of Hop lp token {lp_token.evm_address!s} in {lp_token.chain_id!s}. {e!s}')  # noqa: E501
```
LefterisJP, line 1142: also please don't use chain id. Use chain_name. here and above.

LefterisJP (approved):
lgtm. Let's hope tests pass.

PR #8273, opened 2024-07-16
Add default order to logs queries

yabirgb:
Logs queries default to etherscan if no default node is provided. Those calls required a list of nodes to be provided

Closes #(issue_number)

## Checklist


yabirgb:
Remote error is a 502 from etherscan at `test_evm_address_async`

yabirgb:
Those are inteonatly not using the call order since only etherscan has optimizations

PR #8275, opened 2024-07-16
Decode transactions claiming SAFE tokens from vesting.

LefterisJP:
Fix https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=70040930


rotkehlchen/chain/ethereum/modules/safe/constants.py
```diff
@@ -0,0 +1,7 @@
+from typing import Final
+
+from rotkehlchen.chain.evm.types import string_to_evm_address
+
+CPT_SAFE: Final = 'safe'
```
yabirgb, line 5: we already have `CPT_SAFE_MULTISIG`. Does it make sense to keep only `safe` and not have multiple safe things?

rotkehlchen/chain/ethereum/modules/safe/decoder.py
```diff
@@ -0,0 +1,61 @@
...
+                to_address=SAFE_VESTING,
+            )],
+        )
+
+    # -- DecoderInterface methods
+
+    def addresses_to_decoders(self) -> dict[ChecksumEvmAddress, tuple[Any, ...]]:
+        return {
+            SAFE_VESTING: (self._decode_safe_vesting,),
+        }
```
yabirgb, line 53: nitpick: fits in one line

rotkehlchen/chain/ethereum/modules/safe/constants.py
LefterisJP: It's not the same protocol. One is the multisig. The other is the token. Not sure if it makes sense

rotkehlchen/chain/ethereum/modules/safe/decoder.py
LefterisJP: done

yabirgb (approved):
lgtm

PR #8281, opened 2024-07-17
fix: convert backend termination to synchronous

kelsos:
An improperly implemented promise was causing the
termination to hang on windows.

Closes #(issue_number)

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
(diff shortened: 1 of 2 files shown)


kelsos:
@yabirgb we also have to make sure that there is no side-effect on mac/linux

I did a test from a local build both windows/linux but an extra confirmation would be really appreciated.

yabirgb:
I'm going to check macos

yabirgb (approved):
Tested in windows and macos and works fine

PR #8285, opened 2024-07-18
feat: add placeholder for image component

lukicenturi:
In some cases the NFT image url is broken, so we need to give placeholder here.


kelsos (approved):
lgtm

Issue #8101, opened 2024-06-16
Specify when an EVM transaction has failed


LefterisJP:
## Abstract

At the moment there is no way to see that a history even of gas burning is due to a failed transaction.

As an example check this https://etherscan.io/tx/0xfbfd35db096d0acb26a988895841d786baafe08f6cf55265338e0b5db58350ee transaction.

In rotki it appears like this

![2024-06-16_21-56](https://github.com/rotki/rotki/assets/1658405/4640ede3-f620-492e-befb-0438c21ed8c5)

It just shows the burned gas from the failed transaction but nothing else. It looks like a not properly decoded event and not a failed transaction. It would be good to somehow specify it and show in the UI.

## Task

- Figure out the best way to show it it in the backend. Perhaps specific type/subtype + CPT_GAS?
- Figoure out how to best represent in the frontend. Specific event type showing a failed icon would make sense.

PR #8288, opened 2024-07-18
Display failed transactions

LefterisJP:
Fix https://github.com/rotki/rotki/issues/8101

Looks like this

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`8101` Failed transactions will now be displayed as failed and not just show "burned gas" event only.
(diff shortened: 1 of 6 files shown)


yabirgb:
Seems that is missing the accounting rule. We need to add it also

LefterisJP:
> Seems that is missing the accounting rule. We need to add it also

do we have tests for these? Something should fail no?

LefterisJP:
answer: no nothing failed about accounting

LefterisJP:
The failed test is already marked as a TODO in the VCR tests `rotkehlchen/tests/unit/test_tokens.py::test_last_queried_ts `

PR #8289, opened 2024-07-18
chore: downgrade electron-store because it breaks dev

kelsos:
Closes #(issue_number)

## Checklist


kelsos:
Probably because version 9.0.0 went full ESM and we need to adjust the tsconfig for the main process. Need to figure it out.

Issue #7979, opened 2024-05-24
Remove evm event logs "removed" attribute


LefterisJP:
## Abstract

As discussed with @yabirgb in discord, he noticed this is probably not needed.

DB SChema: https://github.com/rotki/rotki/blob/400e4755a25dae7f7bd480d8a64c0958ac0935b1/rotkehlchen/db/schema.py#L430
Event log class: https://github.com/rotki/rotki/blob/400e4755a25dae7f7bd480d8a64c0958ac0935b1/rotkehlchen/chain/evm/structures.py#L15

## Motivation

`removed` is a very special field that will always be false except for when a log is discarded due to a chain reorg

This can't happen with our event processing since for all of them we get them from already existing transactions. Except for:

1. Places where we actually query logs of an address. There we can get logs that are removed. At that place we should ignore them.
2. If we query etherscan transactions, get a transaction that's at the tip, process it, decode it but the block ends up getting reorged and then it's all wrong.

Nr. 1 -> Easy. We just don't save this field, and if any ofthe events has removed == True we skip it.
Nr. 2 -> Super bad edge case that we can't fix.  (but this can happen today too)

## Task

1. Remove it from the python class -> save memory
2. Remove it from the DB -> save space
3. Upgrade the DB to remove it from all existing DBs -> save space

PR #8294, opened 2024-07-18
Remove "removed" receipt log column

LefterisJP:
It's not used and we save space by removing an unused column for each log event.

Fix https://github.com/rotki/rotki/issues/7979


LefterisJP:
sqldiff test failed due to too many changes to the upgrade test DB. I ran history event query on it to get some test data.

LefterisJP:
All green here. Will merge.

PR #8295, opened 2024-07-18
Purge dead code

LefterisJP:


rotkehlchen/globaldb/handler.py
```diff
@@ -240,23 +240,6 @@ def cleanup(self) -> None:
         if self._packaged_db_conn is not None:
             self._packaged_db_conn.close()
 
-    def _wal_checkpoint(self) -> None:
```
LefterisJP, line 243: Notice that this was no longer used @yabirgb

yabirgb: Yes, oj removed all the usage

PR #8299, opened 2024-07-19
Add new api for yearn

yabirgb:
- Adds ydemon as new source for yearn vaults. Since it doesn't support the v1 and
the old api wasn't being updated with the new vaults we need to aggreagate information
from both of them
- Also updates the logics to update only the protocol of tokens since it was a common
path in the app


rotkehlchen/chain/ethereum/modules/yearn/utils.py
```diff
@@ -72,30 +73,62 @@ def _maybe_reset_yearn_cache_timestamp(data: dict[str, Any] | None) -> bool:
...
+    for vault in data:
+        if (vault_address := vault.get('address')) is None:
+            log.error(f'Unexpected vault schema in yearn api for {vault}. Skipping')
+            continue
+
+        if (version := vault.get('version')) is not None and version.startswith('3.'):
+            log.debug(f'Skipping yearn v3 vault {vault.get("address")}')
+            continue  # skip v3 vaults until we add them #7540
+
+        if vault_address in vaults_seen:
```
LefterisJP, line 115: why not make them both sets and avoid this extra if? Since you are adding it if it's not there already and that's a set. So that's not needed right?

rotkehlchen/globaldb/handler.py
```diff
@@ -1036,8 +1037,23 @@ def edit_evm_token(entry: EvmToken) -> str:
...
+        if token.protocol == new_protocol:
+            return
+
+        with GlobalDBHandler().conn.write_ctx() as write_cursor:
+            write_cursor.execute(
+                'UPDATE evm_tokens SET protocol = ? WHERE identifier = ?;',
+                (new_protocol, token.identifier),
+            )
+
+        AssetResolver.clean_memory_cache(token.identifier)
```
LefterisJP, line 1055: setting these here both in this and edit function changes behaviour for many places. That's correct though right?

LefterisJP (approved):
Lgtm

PR #8303, opened 2024-07-21
Bump linting packages and adjust code

LefterisJP:


LefterisJP:
The only failed test is fixed in https://github.com/rotki/rotki/pull/8304

PR #8308, opened 2024-07-22
Fix (globaldb): CC IDs and non EVM assets

OjusWiZard:
## Checklist

- [x] Update all the wrong/missing cryptocompare IDs of assets
- [x] Fix the details of non-EVM assets


LefterisJP (approved):
code is probably fine. Let's see tests.

PR #8313, opened 2024-07-23
Chore (Migration): Set protocol of Hop LP tokens

OjusWiZard:
## Checklist

- [x] Add data migration 15 to set protocol value as 'hop' for the Hop LP tokens

### DIFF
--- a/rotkehlchen/data_migrations/migrations/migrations_15.py
+++ b/rotkehlchen/data_migrations/migrations/migrations_15.py
@@ -1,0 +1,47 @@
+import logging
+from typing import TYPE_CHECKING
+
+from rotkehlchen.assets.asset import EvmToken
+from rotkehlchen.chain.evm.decoding.hop.constants import CPT_HOP
+from rotkehlchen.errors.asset import UnknownAsset, WrongAssetType
+from rotkehlchen.globaldb.handler import GlobalDBHandler
+from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.logging import RotkehlchenLogsAdapter, enter_exit_debug_log
+from rotkehlchen.types import HOP_PROTOCOL_LP
+
+if TYPE_CHECKING:
+    from rotkehlchen.data_migrations.progress import MigrationProgressHandler
+    from rotkehlchen.rotkehlchen import Rotkehlchen
+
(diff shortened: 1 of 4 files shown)


rotkehlchen/data_migrations/migrations/migrations_15.py
```diff
@@ -0,0 +1,53 @@
...
+            'LEFT JOIN history_events as HE on HE.identifier=EE.identifier '
+            'WHERE counterparty=? AND type=? AND subtype=?;', (
+                CPT_HOP,
+                HistoryEventType.RECEIVE.serialize(),
+                HistoryEventSubType.RECEIVE_WRAPPED.serialize(),
+            ),
+        ).fetchall()
+
+    progress_handler.set_total_steps(len(hop_lp_assets))
+
```
yabirgb, line 39: the updates will happen fast so I wouldn't worry about setting the steps. Set just one step (update the protocol field in the assets).

Once that is done don't call in fetchall. Iterate over the cursor. This avoids iterating twice the result of the query

rotkehlchen/data_migrations/migrations/migrations_15.py
```diff
@@ -0,0 +1,53 @@
...
+        except (WrongAssetType, UnknownAsset) as e:
+            log.error(f'{e!s} when deserializing an EvmToken in data migration 15')
+            progress_handler.new_step(identifier)
+            continue
+
+        progress_handler.new_step(asset.name)
+        GlobalDBHandler.set_token_protocol_if_missing(
+            token=asset,
+            new_protocol=CPT_HOP,
+        )
```
yabirgb, line 53: doesn't it fit in one line?

rotkehlchen/tests/data_migrations/test_migrations.py
```diff
@@ -492,6 +505,52 @@ def test_migration_14(
     )
 
 
+@pytest.mark.parametrize('data_migration_version', [14])
+@pytest.mark.parametrize('perform_upgrades_at_unlock', [True])
+def test_migration_15(rotkehlchen_api_server) -> None:
+    """
+    Test migration 15
```
yabirgb, line 512:
```suggestion
    """Test migration 15
```

LefterisJP (approved):
LGTM

PR #8315, opened 2024-07-23
Feat (Odos): Decode v1 swap events

### LINKED ISSUE #7629: Understand and decode swaps via Odos v2 router in all supported chains
## Abstract
The Odos router is another router for amm swaps used across multiple chains: https://docs.odos.xyz/product/sor/v2/
We need to support it.

Example (optimism) transaction: https://optimistic.etherscan.io/tx/0x125741a9097e4204f80682d77ae6ffd0cbcf143b608e4a56d779b49822f87459

Router address in optimism: https://optimistic.etherscan.io/address/0xca423977156bb05b13a2ba3b76bc5419e2fe9680

All relevant deployments are in the above link in a specific subsection: https://docs.odos.xyz/product/sor/v2/#deployments

## Specification

Decode all possible Odos related swaps properly across all chains and write tests for them.

OjusWiZard:
Closes #7629

## Checklist

- [x] Decoder for swaps done via Odos v1

### DIFF
--- a/rotkehlchen/chain/evm/decoding/odos/common.py
+++ b/rotkehlchen/chain/evm/decoding/odos/common.py
@@ -1,0 +1,183 @@
+import logging
+from collections import defaultdict
+from typing import TYPE_CHECKING
+
+from rotkehlchen.accounting.structures.balance import Balance
+from rotkehlchen.assets.asset import Asset, EvmToken
+from rotkehlchen.chain.ethereum.utils import asset_normalized_value
+from rotkehlchen.chain.evm.decoding.constants import ERC20_OR_ERC721_TRANSFER
+from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface
+from rotkehlchen.chain.evm.decoding.structures import (
+    DEFAULT_DECODING_OUTPUT,
+    DecoderContext,
+    DecodingOutput,
+)
+from rotkehlchen.chain.evm.decoding.utils import maybe_reshuffle_events
+from rotkehlchen.chain.evm.transactions import EvmTransactions
+from rotkehlchen.constants.misc import ZERO
+from rotkehlchen.constants.resolver import evm_address_to_identifier
+from rotkehlchen.errors.misc import RemoteError
+from rotkehlchen.fval import FVal
+from rotkehlchen.history.events.structures.evm_event import EvmEvent
+from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.logging import RotkehlchenLogsAdapter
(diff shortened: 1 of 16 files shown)


rotkehlchen/chain/evm/decoding/odos/common.py
```diff
@@ -0,0 +1,202 @@
...
+            evm_inquirer=evm_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+        )
+        self.evm_txns = EvmTransactions(self.evm_inquirer, self.base.database)
+        self.router_address = router_address
+        self.native_currency = self.evm_inquirer.native_token
+        self.label = self.counterparties()[0].label
+
+    def resolve_tokens_data(
```
LefterisJP, line 56: Isn't this not specific to Odos? Wouldn't it be needed in other decoders too?

rotkehlchen/chain/evm/decoding/odos/common.py
```diff
@@ -0,0 +1,202 @@
...
+                event_subtype=HistoryEventSubType.FEE,
+                asset=held_asset,
+                balance=Balance(amount=amount),
+                location_label=sender,
+                notes=f'Spend {amount} {held_asset.symbol_or_name()} as an {self.label} fee',
+                counterparty=self.counterparties()[0].identifier,
+                address=self.router_address,
+            )
+            for held_asset, amount in router_holding_amount.items()
+            if amount != ZERO
```
LefterisJP, line 145: what happens if the router_holding_amount is negative? I don't see any checks against this.

rotkehlchen/chain/evm/decoding/odos/common.py
```diff
@@ -0,0 +1,202 @@
...
+                ) or (
+                    event.event_type == HistoryEventType.TRADE and
+                    event.event_subtype == HistoryEventSubType.RECEIVE
+                )
+            ):
+                event.event_type = HistoryEventType.TRADE
+                event.event_subtype = HistoryEventSubType.RECEIVE
+                event.notes = f'Receive {event.balance.amount} {event.asset.symbol_or_name()} as the result of a swap in {self.label}'  # noqa: E501
+                event.address = self.router_address
+                event.counterparty = self.counterparties()[0].identifier
```
LefterisJP, line 190: Both for this and for the label in other places we put it as part of the constructor. Why not do the same here? Putting another constructor argument and saving it in memory duplicates info?

OjusWiZard: yeah that was the reason

LefterisJP (approved):
LGTM

PR #8316, opened 2024-07-23
Fix (Migration 15): Redecode HOP LP events

OjusWiZard:
## Checklist

- [x] Redecode the add liquidity transactions for each Hop LP token to set its protocol and pool_address

### DIFF
--- a/rotkehlchen/data_migrations/migrations/migrations_15.py
+++ b/rotkehlchen/data_migrations/migrations/migrations_15.py
@@ -1,13 +1,17 @@
 import logging
 from typing import TYPE_CHECKING
 
-from rotkehlchen.assets.asset import EvmToken
 from rotkehlchen.chain.evm.decoding.hop.constants import CPT_HOP
-from rotkehlchen.errors.asset import UnknownAsset, WrongAssetType
-from rotkehlchen.globaldb.handler import GlobalDBHandler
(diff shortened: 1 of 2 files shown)


rotkehlchen/data_migrations/migrations/migrations_15.py
```diff
@@ -21,27 +20,46 @@
...
                 HistoryEventType.RECEIVE.serialize(),
                 HistoryEventSubType.RECEIVE_WRAPPED.serialize(),
             ),
-        ):  # set their protocol to Hop
+        ):
+            tx_hash = deserialize_evm_tx_hash(db_tx_hash)
+            chain_id: SUPPORTED_CHAIN_IDS = ChainID.deserialize(Location.deserialize_from_db(db_location).to_chain_id())
+            chain_manager = rotki.chains_aggregator.get_evm_manager(chain_id=chain_id)
+            with rotki.data.db.user_write() as write_cursor:
+                write_cursor.execute(
```
LefterisJP, line 41: no no no. Don't do that. This is deleting the transaction. And then repulling it. Check `_reset_decoded_events in the last upgrade`. We should not pull anything again. No remote network stuff should happen.

LefterisJP (approved):
LGTM. Let's hope it works fine when in the binary with my data too

PR #8317, opened 2024-07-23
Fix (Migration 15): Different db cursor from loop

OjusWiZard:
## Checklist

- [x] Fix to not use the same cursor in a loop for other queries in data migration 15

### DIFF
--- a/rotkehlchen/data_migrations/migrations/migrations_15.py
+++ b/rotkehlchen/data_migrations/migrations/migrations_15.py
@@ -27,34 +27,38 @@ def data_migration_15(rotki: 'Rotkehlchen', progress_handler: 'MigrationProgress
 
     Redecodes the distinct Add liquidity events for each Hop LP token."""
     progress_handler.set_total_steps(1)
-
     # find one Add liquidity event for each Hop LP token
-    event_data = []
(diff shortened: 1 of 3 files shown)


rotkehlchen/data_migrations/migrations/migrations_15.py
```diff
@@ -27,34 +27,38 @@ def data_migration_15(rotki: 'Rotkehlchen', progress_handler: 'MigrationProgress
 
     Redecodes the distinct Add liquidity events for each Hop LP token."""
     progress_handler.set_total_steps(1)
-
     # find one Add liquidity event for each Hop LP token
-    event_data = []
-    with rotki.data.db.conn.write_ctx() as write_cursor:
-        for identifier, db_tx_hash, db_location in write_cursor.execute(
```
LefterisJP, line 34: Ah i see. You were using the cursor in the loop and looping its result, and then reusing it inside the loop, which would result in the cursor being off and perhaps loop not looping all or nor deleting all.

Weird that no linter failed there. I recall this being linter-detectable.

Nice.

LefterisJP:
`test_pickle_dill` and `test_protocol_names_are_known` both failed due to etherscan errors

LefterisJP (approved):
LGTM

PR #8320, opened 2024-07-24
Fix Eigenlayer pending withdrawal balances

LefterisJP:
If the timing of the eigenlayer pending withdrawal completion came before the saving of the start of pending withdrawal was in the DB, then the pending withdrawal was never marked as complete.

This resulted in those cases of completed withdrawals counting again in the balances.

This commit fixes this by forcing redecoding of completion if a completed withdrawal is found without having a match to a pending withdrawal whenever eigenlayer balances are queried.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,6 +3,7 @@ Changelog
 =========
 
 * :release:`1.34.1 <2024-07-24>`
+* :bug:`-` Eigenlayer LST pending withdrawals that have been completed should no longer count as user balance.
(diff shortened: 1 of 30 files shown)


yabirgb (approved):
Failing tests is a timeout 

```
due to ethereum Etherscan API request failed due to HTTPSConnectionPool(host='api.etherscan.io', port=443): Read timed out. (read timeout=30)
```

from `test_protocol_names_are_known` and passes locally. Merging

PR #8324, opened 2024-07-25
Fix nightlies tests for macos

yabirgb:
We have the fixture `include_cryptocompare_key` that checks if we are on mac and on CI to include a cc key. Moreover the key is random from a set and this was breaking the VCR query. Added a filter to avoid comparing against it


LefterisJP (approved):
LGTM

PR #8327, opened 2024-07-25
Refactor (API): Use threadpool for async calls

OjusWiZard:
## Checklist

- [ ] Use gevent threadpool for async tasks in API to not block the main API greenlet


yabirgb:
Closing since we saw the issues of the threadpool. Using a db writter process with zmq now

PR #8328, opened 2024-07-26
chore: fix type for unit tests

lukicenturi:
I was playing with tsconfig.json,
but got some issue (and got drowning with other tasks, so pushing this first


kelsos (approved):
thanks @lukicenturi lgtm

PR #8336, opened 2024-07-28
Use some perflint checks to improve performance

LefterisJP:
The tool is a nice idea but it's overwhelming with way too many false positives.

Some rules also are probably an overkill. Like the import full method and not just module and then do the .method always. Though they do make sense it's a lot of work to adjust and maybe too anal.

Let's see.


LefterisJP:
all green

PR #8338, opened 2024-07-29
fix: metamask error if other wallet is installed

### LINKED ISSUE #8334: Metamask account import fails with "Method not supported" when alternative wallets are enabled
## Problem Definition

If the user's browser has another wallet such as Rabby or Rainbow installed and active alongside Metamask, then Metamask account import can fail with the error `Method not supported`, even if the other wallet is configured to not be the default wallet:

image 

Ideally it should just work, e.g. via [EIP-6963 (Multi Injected Provider Discovery)](https://eips.ethereum.org/EIPS/eip-6963).  But if that's not technically feasible, there should at least be a helpful error message suggesting the workaround below to the user.

## Logs

When Rainbow is the wallet interfering, the browser console shows:

```
Uncaught (in promise) TypeError: Error in invocation of runtime.sendMessage(optional string extensionId, any message, optional object options, optional function callback): chrome.runtime.sendMessage() called from a webpage must specify an Extension ID (string) for its first argument.
    at et (inpage.js:1:769615)
    at inpage.js:1:770023
```

### Workaround

Disable the other wallet.

### System Description

Operating system: Linux (openSUSE Tumbleweed)
Rotki version: 1.34.1 (AppImage)

lukicenturi:
Closes #8334

Fix issue that metamask showed error `Method not supported`, especially if `Rainbow wallet` is installed alongside with the Metamask

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`8334` Importing addresses from MetaMask should work when multiple browser wallets are installed.
 * :bug:`-` rotki will now properly run background tasks when logging out and logging in again.
 * :bug:`-` rotki will now detect new tokens right after finishing decoding new events.
 * :bug:`8169` Prevent a recursion error when querying the price of a token.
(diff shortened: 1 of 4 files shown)


LefterisJP (approved):
LGTM as it probably seems to work.

PR #8340, opened 2024-07-29
Fix (Token detection): Do after decoding ends

OjusWiZard:
## Checklist

- [x] Detect received tokens right after finishing the decoding task

### DIFF
--- a/rotkehlchen/chain/evm/decoding/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/decoder.py
@@ -48,6 +48,7 @@ from rotkehlchen.globaldb.handler import GlobalDBHandler
 from rotkehlchen.history.events.structures.evm_event import EvmProduct
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.logging import RotkehlchenLogsAdapter
+from rotkehlchen.tasks.assets import maybe_detect_new_tokens
 from rotkehlchen.types import ChecksumEvmAddress, EvmTokenKind, EvmTransaction, EVMTxHash, Location
 from rotkehlchen.utils.misc import from_wei, hex_or_bytes_to_address, hex_or_bytes_to_int
 from rotkehlchen.utils.mixins.customizable_date import CustomizableDateMixin
@@ -611,6 +612,7 @@ class EVMTransactionDecoder(ABC):
             )
 
         self._post_process(refresh_balances=refresh_balances)
+        maybe_detect_new_tokens(self.database)
         return events
 
     def _get_or_decode_transaction_events(
--- a/rotkehlchen/tasks/assets.py
+++ b/rotkehlchen/tasks/assets.py
@@ -2,7 +2,6 @@ import logging
 from collections import defaultdict
 from typing import TYPE_CHECKING, Final
 
-
 from rotkehlchen.api.websockets.typedefs import WSMessageType
 from rotkehlchen.assets.asset import Asset, UnderlyingToken
 from rotkehlchen.assets.types import AssetType
@@ -18,11 +17,13 @@ from rotkehlchen.chain.ethereum.modules.aave.v3.constants import (
     AAVE_V3_DATA_PROVIDER as AAVE_V3_DATA_PROVIDER_ETH,
 )
 from rotkehlchen.chain.ethereum.utils import MULTICALL_CHUNKS
+from rotkehlchen.chain.evm.constants import EVM_ADDRESS_REGEX
 from rotkehlchen.chain.evm.decoding.aave.constants import CPT_AAVE_V3
 from rotkehlchen.chain.evm.decoding.aave.v3.constants import (
     AAVE_V3_DATA_PROVIDER as AAVE_V3_DATA_PROVIDER_EVM,
 )
 from rotkehlchen.chain.evm.decoding.constants import ERC20_OR_ERC721_TRANSFER
+from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.chain.gnosis.modules.aave.v3.constants import (
     AAVE_V3_DATA_PROVIDER as AAVE_V3_DATA_PROVIDER_GNO,
 )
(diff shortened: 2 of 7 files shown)


rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -611,6 +612,7 @@ def decode_transaction_hashes(
             )
 
         self._post_process(refresh_balances=refresh_balances)
+        maybe_detect_new_tokens(self.database)
         return events
```
yabirgb, line 616: this doesn't make any external request right?

rotkehlchen/tasks/assets.py
```diff
@@ -381,3 +390,51 @@ def update_aave_v3_underlying_assets(chains_aggregator: 'ChainsAggregator') -> N
...
+                continue
+
+            chain = SupportedBlockchain.from_location(event.location)  # type: ignore[arg-type]  # event.location is one of EVM_LOCATIONS
+            evm_asset = event.asset.resolve_to_evm_token()
+            if (
+                event.location_label is None or
+                EVM_ADDRESS_REGEX.fullmatch(event.location_label) is None
+            ):
+                log.warning(
+                    f'Found invalid location label {event.location_label} in history '
```
yabirgb, line 423: shouldn't you check that the location label is the set of tracked addresses for this chain?

rotkehlchen/tasks/assets.py
```diff
@@ -381,3 +390,51 @@ def update_aave_v3_underlying_assets(chains_aggregator: 'ChainsAggregator') -> N
...
+        last_save_time = database.get_last_balance_save_time(write_cursor)
+        detected_tokens = defaultdict(list)
+        # we query the decoded history events that are the first event of each distinct
+        # unignored asset for that account that happened on or after last_save_time.
+        for event_data in write_cursor.execute(
+            # events that are the earliest events of distinct assets after last_save_time
+            f'SELECT {HISTORY_BASE_ENTRY_FIELDS}, '
+            f'{EVM_EVENT_FIELDS} {EVM_EVENT_JOIN} {filter_ignore_asset_query()} '
+            'AND timestamp >= ? GROUP BY asset, location_label;',
+            (ts_sec_to_ms(last_save_time),),
```
yabirgb, line 409: you have detected tokens for only one chain at this point right? Wouldn't it be better to filter by chain? With that you could avoid the check `event.location not in EVM_LOCATIONS`

yabirgb: Don't do it because you can loose events. The function taht you call at the end updates `last_save_time` and you can miss tokens... but then question: this logic will fail because 

1. YOu query gnosis and you have events between timstamps [17, 20]. You call this logic and `last_save_time = 22`
2. You query ethereum and you ahve events between timesatmp [18,21]. YOu call this logic and you do nothing since there are no events after `last_save_time = 22`

Is this expected? Not sure what is the objetive that you and left have regarding this

rotkehlchen/tests/unit/test_evm_tx_decoding.py
```diff
@@ -296,3 +298,22 @@ def delete_transactions_for_address(address: ChecksumEvmAddress) -> None:
...
+
+
+@pytest.mark.vcr()
+@pytest.mark.parametrize('ethereum_accounts', [['0xcBe21204C4b9F1810363D69773b74203376681a2']])
+def test_token_detection_after_decoding(
+        database: 'DBHandler',
+        ethereum_inquirer: 'EthereumInquirer',
+        ethereum_accounts: 'list[ChecksumEvmAddress]',
+) -> None:
+    with patch.object(database, 'save_tokens_for_address') as save_tokens_mock:
```
yabirgb, line 310: add a small docstring

rotkehlchen/chain/evm/decoding/decoder.py
OjusWiZard: no it only process the data in the db

rotkehlchen/tasks/assets.py
OjusWiZard: now we call it outside of the loop from a different cursor, so should be fine now

rotkehlchen/tasks/assets.py
```diff
@@ -381,3 +391,56 @@ def update_aave_v3_underlying_assets(chains_aggregator: 'ChainsAggregator') -> N
...
+
+            chain: SUPPORTED_EVM_EVMLIKE_CHAINS_TYPE = SupportedBlockchain.from_location(event.location)  # type: ignore[arg-type,assignment]  # event.location is one of EVM_LOCATIONS  # noqa: E501
+            evm_asset = event.asset.resolve_to_evm_token()
+            if (
+                event.location_label is None or
+                EVM_ADDRESS_REGEX.fullmatch(event.location_label) is None or
+                event.location_label not in database.get_single_blockchain_addresses(
+                    cursor=cursor,
+                    blockchain=chain,
+                )
```
yabirgb, line 429: mmm this is one query per event... can't you just do

```
{address_tuple[0] for address_tuple in cursor.execute(
            'SELECT UNIQUE account FROM blockchain_accounts;',
        )}
```

at the start of the function?

rotkehlchen/tasks/assets.py
```diff
@@ -381,3 +391,57 @@ def update_aave_v3_underlying_assets(chains_aggregator: 'ChainsAggregator') -> N
             name=DBCacheStatic.LAST_AAVE_V3_ASSETS_UPDATE,
             value=ts_now(),
         )
+
+
+def maybe_detect_new_tokens(database: 'DBHandler') -> None:
```
LefterisJP, line 396: please add a docstring explaining the what this function does and the algorithm it follows (short explanation)

rotkehlchen/tasks/assets.py
```diff
@@ -381,3 +391,57 @@ def update_aave_v3_underlying_assets(chains_aggregator: 'ChainsAggregator') -> N
...
+            'AND timestamp >= ? GROUP BY asset, location_label;',
+            (ts_sec_to_ms(last_save_time),),
+        ):
+            event = EvmEvent.deserialize_from_db(event_data[1:])
+            if (
+                event.location not in EVM_LOCATIONS or
+                not event.asset.is_evm_token() or
+                event.maybe_get_direction() != EventDirection.IN
+            ):
+                # if the location or asset is not evm, or if event is not IN, we skip it
```
LefterisJP, line 422: no need for the comment. It's obvious

```suggestion
            ):
```

rotkehlchen/tasks/assets.py
```diff
@@ -381,3 +391,57 @@ def update_aave_v3_underlying_assets(chains_aggregator: 'ChainsAggregator') -> N
...
+
+            chain: SUPPORTED_EVM_EVMLIKE_CHAINS_TYPE = SupportedBlockchain.from_location(event.location)  # type: ignore[arg-type,assignment]  # event.location is one of EVM_LOCATIONS  # noqa: E501
+            evm_asset = event.asset.resolve_to_evm_token()
+            if (
+                event.location_label is None or
+                EVM_ADDRESS_REGEX.fullmatch(event.location_label) is None or
+                event.location_label not in tracked_accounts
+            ):
+                log.warning(
+                    f'Found invalid location label {event.location_label} in history '
```
LefterisJP, line 433: Why invalid? Shouldn't it be invalid or non-tracked?

LefterisJP (approved):
LGTM

Issue #8040, opened 2024-06-07
Better Account Overview/Management UI/UX


zinkkrysty:
## Abstract

After initial onboarding, it's difficult to find the interface for adding/viewing accounts. I cannot add an account from the dashboard anymore as the button to add account has disappeared:

<img width="200" alt="image" src="https://github.com/rotki/rotki/assets/570771/081c50af-d9cb-46aa-b3a8-90dd64e32412">

Going to "Accounts & Balances" does not offer an overview of the accounts that I added, nor does it allow me to manage my accounts (rename, delete). I can only add more accounts by pressing the button at the top.

<img width="500" alt="image" src="https://github.com/rotki/rotki/assets/570771/30c69481-c1f0-4a45-8243-107f92b2ff10">

As a user I have to discover "Manage Address Book" and navigate to "Private" to manage the accounts that were added. Even then, from the interface it's not sure whether changing anything here would affect the Balance overviews.

<img width="200" alt="image" src="https://github.com/rotki/rotki/assets/570771/8ace829b-a7f0-4c8d-980e-d2caa9b00f94">

## Motivation

Better UX and easier management of accounts - would make Rotki more preferred by users.

## Specification

There are 2 solutions that I can think of that address this problem:

### Solution 1 
Add a button for adding more accounts next to the Dashboard (addresses the question "Where do I add another account?")

<img width="263" alt="image" src="https://github.com/rotki/rotki/assets/570771/1c7ea258-a8a5-4d18-b140-2d6336112227">

### Solution 2
Create a new page where the user can manage their accounts in "Accounts & Balances"

<img width="956" alt="image" src="https://github.com/rotki/rotki/assets/570771/2a06fc66-2abb-495e-82c5-e009f2025772">

**EDIT**: It seems there is already a solution for addressing this in https://github.com/rotki/rotki/pull/7527

PR #8343, opened 2024-07-29
feat: some improvements on account balances

lukicenturi:
- [x] Evmlike to EVM
- [x] Clicking the chain icon will do the filter
- [x] Filter by chains now show correct usdValue on the group row
- [x] Show top assets instead of amount
- [x] Options to show aggregated assets under chains group, instead of address per chain info

For these two I guess I can do it in different PR
- [ ] Mass delete account
- [ ] Mass edit account

### DIFF
--- a/frontend/app/src/components/AssetBalances.vue
+++ b/frontend/app/src/components/AssetBalances.vue
@@ -162,7 +162,7 @@ function expand(item: AssetBalanceWithPrice) {
         v-if="!hideBreakdown && isEvmNativeToken(row.asset)"
         blockchain-only
         :identifier="row.asset"
-        class="bg-white dark:bg-[#1E1E1E]"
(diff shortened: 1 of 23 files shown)


LefterisJP (commented):
Hey @lukicenturi looks really cool.

Some more things.

1. Clicking the chain icon will do the filtering. Here what I imagined was the opposite. Clicking the chain icon should basically gray out that chain, still keep it there and keep the rest only. This is so you can filter for multiple chains. Because with the current approach you can only filter one by one.

kelsos (approved):
I am gonna merge and we can take care of whatever need fixing/remaining on develop

kelsos:
For the filter on a specific address we need to think a bit

kelsos:
related to #8040

PR #8346, opened 2024-07-30
Fixes for etherscan

yabirgb:
- [Fail fast with etherscan in ethereum](https://github.com/rotki/rotki/commit/84aa9f614976a713fda2019ee7e925655b3fa37e)
- [Avoid querying a evm chain if there are no accounts for it](https://github.com/rotki/rotki/commit/fccc5e09f0f69420a39980b604a3ef5bb17c779d)
- [ethereum nodes to connect without addresses](https://github.com/rotki/rotki/commit/94cc6b3821ba1fcc76ea7be0816f221a27c49671)

### DIFF
--- a/rotkehlchen/chain/aggregator.py
+++ b/rotkehlchen/chain/aggregator.py
@@ -544,6 +544,9 @@ class ChainsAggregator(CacheableMixIn, LockableQueryMixIn):
                 xpub_manager.check_for_new_xpub_addresses(blockchain=blockchain)  # type: ignore # is checked in the if
         else:  # all chains
             for chain in SupportedBlockchain:
+                if chain.is_evm() and len(self.accounts.get(chain)) == 0:  # don't check eth2 and bitcoin since we might need to query new addresses  # noqa: E501
+                    continue
+
                 query_method = f'query_{chain.get_key()}_balances'
                 getattr(self, query_method)(ignore_cache=ignore_cache)
                 if ignore_cache is True and chain.is_bitcoin():
--- a/rotkehlchen/chain/evm/node_inquirer.py
+++ b/rotkehlchen/chain/evm/node_inquirer.py
@@ -233,6 +233,9 @@ class EvmNodeInquirer(ABC, LockableQueryMixIn):
 
         If `when_tracked_accounts` is True then it will connect when we have some
         tracked accounts in the DB. Otherwise when we have none.
+
+        In ethereum case always connect to nodes. Needed for ENS resolution.
+        For other EVM chains we respect `when_tracked_accounts`.
         """
         if self.connected_to_any_web3() or self.greenlet_manager.has_task(_connect_task_prefix(self.chain_name)):  # noqa: E501
             return
@@ -241,7 +244,10 @@ class EvmNodeInquirer(ABC, LockableQueryMixIn):
             accounts = self.database.get_blockchain_accounts(cursor)
 
         tracked_accounts_num = len(accounts.get(self.blockchain))
-        if (tracked_accounts_num != 0 and when_tracked_accounts) or (when_tracked_accounts is False and tracked_accounts_num == 0):  # noqa: E501
+        if (
+            (tracked_accounts_num != 0 and when_tracked_accounts) or
+            (tracked_accounts_num == 0 and (when_tracked_accounts is False or self.chain_id == ChainID.ETHEREUM))  # noqa: E501
+        ):
             rpc_nodes = self.database.get_rpc_nodes(blockchain=self.blockchain, only_active=True)
             self.connect_to_multiple_nodes(rpc_nodes)
 
--- a/rotkehlchen/externalapis/etherscan.py
+++ b/rotkehlchen/externalapis/etherscan.py
@@ -240,9 +240,9 @@ class Etherscan(ExternalServiceWithApiKey, ABC):
                     },
                 )
                 self.warning_given = True
-        else:
-            query_str += f'&apikey={api_key}'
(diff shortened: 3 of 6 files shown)


rotkehlchen/chain/aggregator.py
```diff
@@ -544,6 +544,9 @@ def query_balances(
                 xpub_manager.check_for_new_xpub_addresses(blockchain=blockchain)  # type: ignore # is checked in the if
         else:  # all chains
             for chain in SupportedBlockchain:
+                if chain.is_evm() and len(self.accounts.get(chain)) == 0:
```
LefterisJP, line 547: why only for evm?

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -225,7 +225,12 @@ def __init__(
         # A cache for erc20 and erc721 contract info to not requery the info
         self.contract_info_erc20_cache: LRUCacheWithRemove[ChecksumEvmAddress, dict[str, Any]] = LRUCacheWithRemove(maxsize=1024)  # noqa: E501
         self.contract_info_erc721_cache: LRUCacheWithRemove[ChecksumEvmAddress, dict[str, Any]] = LRUCacheWithRemove(maxsize=512)  # noqa: E501
-        self.maybe_connect_to_nodes(when_tracked_accounts=True)
+        # Etherscan for ethereum requires an API key. For a new user if we don't
```
LefterisJP, line 228: Assume all etherscans. That's their plan. Don't do only for ethereum.

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -225,7 +225,12 @@ def __init__(
...
+        # connect to any web3 node the experience is bad since we can't add any
+        # new address without. We have error messages that point to add the key
+        # but it can give a bad first impression or be confusing. Ethereum is
+        # needed since is where we resolve ENS names.
```
LefterisJP, line 232:
```suggestion
        # needed since that is where we resolve ENS names.
```

rotkehlchen/externalapis/etherscan.py
```diff
@@ -240,6 +240,8 @@ def _query(
                     },
                 )
                 self.warning_given = True
+            if self.chain == SupportedBlockchain.ETHEREUM:
```
LefterisJP, line 243: as I said assume they all do.

rotkehlchen/chain/aggregator.py
yabirgb: eth2 is an exception since we don't have a balance structure for them. Left only that one unchecked

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -225,7 +225,9 @@ def __init__(
         # A cache for erc20 and erc721 contract info to not requery the info
         self.contract_info_erc20_cache: LRUCacheWithRemove[ChecksumEvmAddress, dict[str, Any]] = LRUCacheWithRemove(maxsize=1024)  # noqa: E501
         self.contract_info_erc721_cache: LRUCacheWithRemove[ChecksumEvmAddress, dict[str, Any]] = LRUCacheWithRemove(maxsize=512)  # noqa: E501
-        self.maybe_connect_to_nodes(when_tracked_accounts=True)
+        # Only in the case of ethereum connect to RPC nodes since if there are no accounts yet
+        # and the api key for etherscan is exhausted we can't resolve ENS names.
```
LefterisJP, line 229:
```suggestion
        # In ethereum case always connect to nodes. Needed for ENS resolution.
        # For other EVM chains connect only if addresses are tracked.
```

rotkehlchen/externalapis/etherscan.py
```diff
@@ -240,9 +252,15 @@ def _query(
                     },
                 )
                 self.warning_given = True
-        else:
-            query_str += f'&apikey={api_key}'
 
+            if (raw_key := ROTKI_INCLUDED_KEYS.get(self.chain)) is not None:
```
LefterisJP, line 256: why the if here? This should **always** be there. We control that mapping.

rotkehlchen/externalapis/etherscan.py
```diff
@@ -240,9 +252,15 @@ def _query(
...
+                api_key = ApiKey(raw_key)  # we warn right before that a key is needed
+                log.debug(f'Using default etherscan key for {self.chain}')
+            else:
+                self.msg_aggregator.add_error(msg := f'Etherscan for {self.chain} was queried but requires an API key. Please add one.')  # noqa: E501
```
LefterisJP, line 260: not sure I get this. What we need to achieve is:

1. If there is no api key use default. But warn user as we do now that they need an api key
2. If there is an api key from the user use that.

rotkehlchen/externalapis/etherscan.py
```diff
@@ -240,9 +252,11 @@ def _query(
                     },
                 )
                 self.warning_given = True
-        else:
-            query_str += f'&apikey={api_key}'
 
+            api_key = ApiKey(ROTKI_INCLUDED_KEYS[self.chain])
+            log.debug(f'Using default etherscan key for {self.chain}')
```
LefterisJP, line 257: Do we need this? Won';t this spam the logs? We can probably see that the default key is being used as api key is also logged in the logs right?

On the other hand seeing this ins the logs would point us to a user who needs to take action. Not sure what's best.

LefterisJP (approved):
LGTM

PR #8347, opened 2024-07-30
Fix (Async tasks): No pending after logout

OjusWiZard:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=72778103

## Checklist

- [x] Clear the tracked greenlets list after they are all killed at logout

I confirmed that they are all dead, but were still tracked after logout.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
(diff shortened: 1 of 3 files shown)


LefterisJP (approved):
LGTM

PR #8348, opened 2024-07-31
fix: change metamask flow for better clarity

lukicenturi:
Closes #(issue_number)

## Checklist


frontend/app/src/electron-main/http.ts
```diff
@@ -119,7 +119,8 @@ function handleRequests(req: IncomingMessage, res: ServerResponse, cb: Callback)
       return;
     }
   }
-  const basePath = checkIfDevelopment() ? path.join(__dirname, '..', 'public') : __dirname;
+  const dirname = path.dirname(new URL(import.meta.url).pathname);
```
kelsos, line 122: we are on node > 20.11 so we should be able to use `import.meta.dirname`

https://stackoverflow.com/a/50052194

frontend/app/src/components/accounts/blockchain/MetamaskAddressesImport.vue
```diff
@@ -0,0 +1,139 @@
...
+const { isPackaged, metamaskImport } = useInterop();
+
+async function importAddresses() {
+  const addresses = await (isPackaged ? metamaskImport() : getMetamaskAddresses());
+  emit('update:addresses', addresses);
+}
+</script>
+
+<template>
+  <div>
```
kelsos, line 33: Is the extra div around absolutely needed? do we pass style here?

kelsos (commented):
Ok so small comments plus we need to make sure that the address field doesn't keep the state when changing chains, especially if e.g. we go from ETH to BTC :)

lukicenturi:
should be updated now

> Ok so small comments plus we need to make sure that the address field doesn't keep the state when changing chains, especially if e.g. we go from ETH to BTC :)

kelsos (approved):
lgtm

PR #8352, opened 2024-08-02
Refactor (Process): DB writer

### LINKED ISSUE #7272: Database writer process
## Abstract

As discussed we should try to move some stuff to different subprocesses in order to modularize the design and allow for some services to be written even in other languages.

For communication between processes we should use some fast IPC that works in all OSes such as https://zeromq.org/ with a schema such as protobuf: https://protobuf.dev/

This way we can slowly have various services that do stuff like, periodic task workers, DB writer, etc.

As a start and test of the concept we want a single process that just writes to the DB in order to:
1. Test the system of ipc communication in all OSes
2. Create a very simple new process
3. Make all db writing sequential to avoid all the problems that having multiple coroutines/greenlets writing to sqlite asynchronously can create

## Task

Do it + spec out more/

OjusWiZard:
Closes #7272

## Checklist

- [x] Seperate DB write process from API process
- [x] Communication between both the processes
- [x] Adjust existing code to support this change

### DIFF
--- a/rotkehlchen/db/dbhandler.py
+++ b/rotkehlchen/db/dbhandler.py
@@ -57,7 +57,7 @@ from rotkehlchen.db.constants import (
     KRAKEN_ACCOUNT_TYPE_KEY,
     USER_CREDENTIAL_MAPPING_KEYS,
 )
-from rotkehlchen.db.drivers.gevent import DBConnection, DBConnectionType, DBCursor
+from rotkehlchen.db.drivers.client import DBConnection, DBConnectionType, DBCursor, DBWriterClient
 from rotkehlchen.db.evmtx import DBEvmTx
 from rotkehlchen.db.filtering import (
     AssetMovementsFilterQuery,
@@ -188,6 +188,7 @@ class DBHandler:
             initial_settings: ModifiableDBSettings | None,
             sql_vm_instructions_cb: int,
             resume_from_backup: bool,
+            db_writer_port: int,
     ):
         """Database constructor
 
@@ -219,14 +220,15 @@ class DBHandler:
         # Lock to make sure that 2 callers of get_or_create_evm_token do not go in at the same time
         self.get_or_create_evm_token_lock = Semaphore()
         self.password = password
+        self._db_writer_port = db_writer_port
         self._connect()
         self._check_unfinished_upgrades(resume_from_backup=resume_from_backup)
         self._run_actions_after_first_connection()
-        with self.user_write() as cursor:
+        with self.user_read_write() as write_cursor:
             if initial_settings is not None:
-                self.set_settings(cursor, initial_settings)
-            self.update_owned_assets_in_globaldb(cursor)
-            self.sync_globaldb_assets(cursor)
+                self.set_settings(write_cursor, initial_settings)  # type: ignore[arg-type]  # here we can use DBCursor because no other process will be writing
+            self.update_owned_assets_in_globaldb(write_cursor)
+            self.sync_globaldb_assets(write_cursor)  # type: ignore[arg-type]  # same reason as above
 
     def _check_unfinished_upgrades(self, resume_from_backup: bool) -> None:
         """
@@ -337,12 +339,12 @@ class DBHandler:
         # Run upgrades if needed -- only for user DB
         fresh_db = DBUpgradeManager(self).run_upgrades()
         if fresh_db:  # create tables during the first run and add the DB version
-            self.conn.executescript(DB_SCRIPT_CREATE_TABLES)
-            cursor = self.conn.cursor()
-            cursor.execute(
-                'INSERT OR REPLACE INTO settings(name, value) VALUES(?, ?)',
-                ('version', str(ROTKEHLCHEN_DB_VERSION)),
-            )
+            with self.user_write() as write_cursor:
+                write_cursor.executescript(DB_SCRIPT_CREATE_TABLES)
+                write_cursor.execute(
+                    'INSERT OR REPLACE INTO settings(name, value) VALUES(?, ?)',
+                    ('version', str(ROTKEHLCHEN_DB_VERSION)),
+                )
 
         # run checks on the database
         self.conn.schema_sanity_check()
@@ -351,24 +353,26 @@ class DBHandler:
         # This logic executes only for the transient db
         self._connect(conn_attribute='conn_transient')
         transient_version = 0
-        cursor = self.conn_transient.cursor()
-        with suppress(sqlcipher.DatabaseError):  # pylint: disable=no-member  # not created yet
-            result = cursor.execute('SELECT value FROM settings WHERE name=?', ('version',)).fetchone()  # noqa: E501
-            if result is not None:
-                transient_version = int(result[0])
+        with (
+            self.conn_transient.read_ctx() as cursor,
+            self.conn_transient.write_ctx() as write_cursor,
+        ):
+            with suppress(sqlcipher.DatabaseError):  # pylint: disable=no-member  # not created yet
+                result = cursor.execute('SELECT value FROM settings WHERE name=?', ('version',)).fetchone()  # noqa: E501
+                if result is not None:
+                    transient_version = int(result[0])
 
-        if transient_version != ROTKEHLCHEN_TRANSIENT_DB_VERSION:
-            # "upgrade" transient DB
-            tables = list(cursor.execute('select name from sqlite_master where type is "table"'))
-            cursor.executescript('PRAGMA foreign_keys = OFF;')
-            cursor.executescript(';'.join([f'DROP TABLE IF EXISTS {name[0]}' for name in tables]))
-            cursor.executescript('PRAGMA foreign_keys = ON;')
-        self.conn_transient.executescript(DB_SCRIPT_CREATE_TRANSIENT_TABLES)
-        cursor.execute(
-            'INSERT OR IGNORE INTO settings(name, value) VALUES(?, ?)',
-            ('version', str(ROTKEHLCHEN_TRANSIENT_DB_VERSION)),
-        )
-        self.conn_transient.commit()
+            if transient_version != ROTKEHLCHEN_TRANSIENT_DB_VERSION:
+                # "upgrade" transient DB
+                tables = list(cursor.execute('select name from sqlite_master where type is "table"'))  # noqa: E501
+                write_cursor.executescript('PRAGMA foreign_keys = OFF;')
+                write_cursor.executescript(';'.join([f'DROP TABLE IF EXISTS {name[0]}' for name in tables]))  # noqa: E501
+                write_cursor.executescript('PRAGMA foreign_keys = ON;')
+            self.conn_transient.executescript(DB_SCRIPT_CREATE_TRANSIENT_TABLES)
+            write_cursor.execute(
+                'INSERT OR IGNORE INTO settings(name, value) VALUES(?, ?)',
+                ('version', str(ROTKEHLCHEN_TRANSIENT_DB_VERSION)),
+            )
 
     def get_md5hash(self, transient: bool = False) -> str:
         """Get the md5hash of the DB
@@ -444,7 +448,7 @@ class DBHandler:
 
     def set_setting(
             self,
-            write_cursor: 'DBCursor',
+            write_cursor: 'DBWriterClient',
             name: Literal[
                 'version',
                 'last_write_ts',
@@ -481,6 +485,7 @@ class DBHandler:
                 path=str(fullpath),
                 connection_type=connection_type,
                 sql_vm_instructions_cb=self.sql_vm_instructions_cb,
+                db_writer_port=self._db_writer_port,
             )
         except sqlcipher.OperationalError as e:  # pylint: disable=no-member
             raise SystemPermissionError(
@@ -492,15 +497,16 @@ class DBHandler:
         if self.sqlcipher_version == 3:
             script += f'PRAGMA kdf_iter={KDF_ITER};'
         try:
-            conn.executescript(script)
-            conn.execute('PRAGMA foreign_keys=ON')
-            # Optimizations for the combined trades view
-            # the following will fail with DatabaseError in case of wrong password.
-            # If this goes away at any point it needs to be replaced by something
-            # that checks the password is correct at this same point in the code
-            conn.execute('PRAGMA cache_size = -32768')
-            # switch to WAL mode: https://www.sqlite.org/wal.html
-            conn.execute('PRAGMA journal_mode=WAL;')
+            for _conn in (conn._conn, conn):  # both reader and writer connections
+                _conn.executescript(script)
+                _conn.execute('PRAGMA foreign_keys=ON')
+                # Optimizations for the combined trades view
+                # the following will fail with DatabaseError in case of wrong password.
+                # If this goes away at any point it needs to be replaced by something
+                # that checks the password is correct at this same point in the code
+                _conn.execute('PRAGMA cache_size = -32768')
+                # switch to WAL mode: https://www.sqlite.org/wal.html
+                _conn.execute('PRAGMA journal_mode=WAL;')
         except sqlcipher.DatabaseError as e:  # pylint: disable=no-member
             conn.close()
             raise AuthenticationError(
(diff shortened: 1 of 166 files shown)


rotkehlchen/start.py
```diff
@@ -0,0 +1,65 @@
...
+        self.command = command
+        self.process = process
+
+
+class ProcessesManager:
+    def __init__(self) -> None:
+        self.processes_running = True
+        self.processes: dict[ProcessType, ProcessData] = {
+            ProcessType.API_SERVER: ProcessData(
+                command=[sys.executable, '-m', 'rotkehlchen'] + sys.argv[1:],
```
yabirgb, line 28: can you try calling pyinstaller manually and build only the backend? You can check package.py to see what logic is followed. I believe `sys.executable` doesn't exists there since what pyinstlaler does is bundle a python and then create an exe. Irc the frontend calls the binaries directly but check if this works

rotkehlchen/db/dbhandler.py
```diff
@@ -2144,46 +2166,47 @@ def delete_data_for_evm_address(
...
-        ]
-        for hashes_chunk in get_chunks(hashes_to_remove, n=1000):  # limit num of hashes in a query
-            write_cursor.execute(  # delete transactions themselves
-                f'DELETE FROM zksynclite_transactions WHERE tx_hash IN '
-                f'({",".join("?" * len(hashes_chunk))})',
-                hashes_chunk,
+        with self.conn.read_ctx() as cursor:
+            other_addresses = self.get_single_blockchain_addresses(
+                cursor=cursor,
+                blockchain=SupportedBlockchain.ZKSYNC_LITE,
```
yabirgb, line 2180: just for the indentation... you don't need the read cursor for more than two queries right? Can't you do them first and then operate only with the write cursor?

rotkehlchen/tests/db/test_evmtx.py
```diff
@@ -93,23 +93,28 @@ def test_add_get_evm_transactions(data_dir, username, sql_vm_instructions_cb):
         assert len(errors) == 0
         assert len(warnings) == 0
         filter_query = EvmTransactionsFilterQuery.make(chain_id=ChainID.ETHEREUM)
+    with data.db.conn.read_ctx() as cursor:
         returned_transactions = dbevmtx.get_evm_transactions(cursor, filter_query, True)
         assert returned_transactions == [tx1, tx2]
 
+    with data.db.user_write() as cursor:
```
yabirgb, line 100: why not opening the cuirsor as

cursor = data.db.conn.cusor() and then close it manually?

rotkehlchen/tests/db/test_ranges.py
```diff
@@ -44,20 +44,24 @@ def test_update_used_query_range(database):
...
         start_ts = 250
         end_ts = 500
         query_range = dbranges.get_location_query_ranges(cursor, location2, start_ts, end_ts)
+    with database.user_write() as cursor:
         dbranges.update_used_query_range(
             cursor,
             location2,
             queried_ranges=[(start_ts, end_ts)] + query_range,
         )
+    with database.conn.read_ctx() as cursor:
```
yabirgb, line 66: same comment about opening the read cursor once

rotkehlchen/tests/unit/test_data_updates.py
```diff
@@ -467,17 +467,18 @@ def test_update_rpc_nodes(data_updater: RotkiDataUpdater) -> None:
...
-        )
-        write_cursor.execute('SELECT COUNT(*) FROM rpc_nodes')
-        assert write_cursor.fetchone()[0] == default_rpc_nodes_count + 1
+        with data_updater.user_db.conn.write_ctx() as write_cursor:
+            write_cursor.execute(
+                'INSERT INTO rpc_nodes(name, endpoint, owned, active, weight, blockchain) '
+                'VALUES(?, ?, ?, ?, ?, ?)',
+                custom_node_tuple,
+            )
+        cursor.execute('SELECT COUNT(*) FROM rpc_nodes')
```
yabirgb, line 480: you need to close the cursor before reading to ensure that the other connection can see it right?

rotkehlchen/tests/unit/test_loopring.py
```diff
@@ -43,17 +43,23 @@ def test_loopring_accountid_mapping(database):
     addr1 = make_evm_address()
     addr2 = make_evm_address()
 
-    with database.user_write() as cursor:
+    with database.conn.read_ctx() as cursor:
```
yabirgb, line 46: isn't it better in cases like this test to open the read/write cursor that you have?

tools/pyinstaller_hooks/hook-zmq.py
yabirgb, line 1: Did you check if this file was needed? I believe it might be but not sure

rotkehlchen/tests/utils/premium.py
```diff
@@ -146,12 +146,14 @@ def setup_starting_environment(
...
-            # Emulate already having the api keys in the DB
-            rotkehlchen_instance.data.db.set_rotkehlchen_premium(premium_credentials)
+    if not first_time:
+        # Emulate already having the api keys in the DB
+        rotkehlchen_instance.data.db.set_rotkehlchen_premium(premium_credentials)
 
+    with rotkehlchen_instance.data.db.user_write() as cursor:
         rotkehlchen_instance.data.db.set_setting(cursor, name='premium_should_sync', value=db_can_sync_setting)  # noqa: E501
+
+    with rotkehlchen_instance.data.db.conn.read_ctx() as cursor:
```
yabirgb, line 156: jejejej same comment about read/write cursor

rotkehlchen/db/dbhandler.py
```diff
@@ -491,15 +494,16 @@ def _connect(self, conn_attribute: Literal['conn', 'conn_transient'] = 'conn') -
...
+            for _conn in (conn._conn, conn):  # both reader and writer connections
+                _conn.executescript(script)
+                _conn.execute('PRAGMA foreign_keys=ON')
+                # Optimizations for the combined trades view
+                # the following will fail with DatabaseError in case of wrong password.
+                # If this goes away at any point it needs to be replaced by something
+                # that checks the password is correct at this same point in the code
+                _conn.execute('PRAGMA cache_size = -32768')
+                # switch to WAL mode: https://www.sqlite.org/wal.html
+                _conn.execute('PRAGMA journal_mode=WAL;')
```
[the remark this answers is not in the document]
OjusWiZard: I checked the docs and also tried from two connections. These queries are only applied to the current database connection.

rotkehlchen/db/drivers/server.py
```diff
@@ -0,0 +1,244 @@
...
+                raise KeyError(f'No cursor for {call_data.db_path} found')
+        else:
+            starting_reference = db_connection
+
+        return getattr(starting_reference, call_data.method)(*call_data.args, **call_data.kwargs)
+
+    def listen(self) -> None:
+        while True:
+            try:
+                call_data: ZMQCallData = self.zmq_connection.recv_pyobj()
```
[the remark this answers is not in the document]
OjusWiZard: right

rotkehlchen/db/drivers/server.py
```diff
@@ -0,0 +1,244 @@
...
+    ) -> None:
+        self.path = path
+        self.connection_type = connection_type
+        self.sqlite_vm_instructions = sqlite_vm_instructions
+
+
+class ZMQCallData:
+    """The data packet sent through ZMQ call"""
+    connection_type: DBConnectionType
+    db_path: str
```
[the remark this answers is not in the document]
OjusWiZard: I found some places where we set temporary databases like `import_unencrypted`, also it supports parallel execution of tests.

rotkehlchen/db/drivers/server.py
```diff
@@ -0,0 +1,244 @@
...
+                    del self.db_cursors[call_data.db_path][call_data.cursor_name or 'default']
+                    self.zmq_connection.send_pyobj(None)
+                    continue
+
+                result = self._handle_query(db_connection, call_data)
+
+            except BaseException as e:
+                if self.shutting_down:
+                    break
+                self.zmq_connection.send_pyobj(ZMQReturnData(error=e).serialize())
```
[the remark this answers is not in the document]
OjusWiZard: I was thinking to switch to protobuf in another PR because this was already getting big

rotkehlchen/tests/db/test_evmtx.py
OjusWiZard: the context also does the same. And, also commits when they are closed.

rotkehlchen/tests/unit/test_data_updates.py
OjusWiZard: Yes.

rotkehlchen/tests/db/test_evmtx.py
yabirgb: I mean for the read cursor. That would avoid having to open an close every time. In the test I marked you had many read cursors getting open and close irc

OjusWiZard: yeah for the reader, that should work

rotkehlchen/tests/unit/test_loopring.py
OjusWiZard: reader can be used like that, but writer has to commit changes

rotkehlchen/db/eth2.py
```diff
@@ -229,19 +229,20 @@ def get_associated_with_addresses_validator_indices(
...
-            return  # no event found so nothing to do
+        with self.db.conn.read_ctx() as cursor:
+            cursor.execute(
+                'SELECT HE.identifier, HE.timestamp, HE.amount FROM history_events HE LEFT JOIN '
+                'eth_staking_events_info SE ON SE.identifier = HE.identifier WHERE '
+                'SE.validator_index=? AND HE.entry_type=? ORDER BY HE.timestamp DESC LIMIT 1',
+                (index, HistoryBaseEntryType.ETH_WITHDRAWAL_EVENT.value),
+            )
+            if (latest_result := cursor.fetchone()) is None:
+                return  # no event found so nothing to do
```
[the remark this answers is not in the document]
OjusWiZard: the write_cursor is used below in the function

rotkehlchen/db/dbhandler.py
```diff
@@ -350,24 +351,26 @@ def _run_actions_after_first_connection(self) -> None:
...
+                result = cursor.execute('SELECT value FROM settings WHERE name=?', ('version',)).fetchone()  # noqa: E501
+                if result is not None:
+                    transient_version = int(result[0])
+
+            if transient_version != ROTKEHLCHEN_TRANSIENT_DB_VERSION:
+                # "upgrade" transient DB
+                tables = list(cursor.execute('select name from sqlite_master where type is "table"'))  # noqa: E501
+                write_cursor.executescript('PRAGMA foreign_keys = OFF;')
+                write_cursor.executescript(';'.join([f'DROP TABLE IF EXISTS {name[0]}' for name in tables]))  # noqa: E501
+                write_cursor.executescript('PRAGMA foreign_keys = ON;')
```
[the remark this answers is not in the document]
OjusWiZard: it's needed in both the connection

rotkehlchen/db/drivers/server.py
```diff
@@ -0,0 +1,246 @@
...
+    lastrowid: int
+
+
+class SerializedZMQReturnData(TypedDict):
+    """Dictionary type for serialized ZMQ return data"""
+    result: SerializedDBWriteResult | None
+    error: Any
+
+
+class DbMethod(StrEnum, SerializableEnumNameMixin):
```
yabirgb, line 54: if you use `SerializableEnumNameMixin` do you need `StrEnum`?

rotkehlchen/db/drivers/server.py
```diff
@@ -0,0 +1,246 @@
...
+    @classmethod
+    def deserialize(cls, data: SerializedZMQCallData) -> 'ZMQCallData':
+        return cls(
+            connection_type=DBConnectionType(data['connection_type']),
+            db_path=data['db_path'],
+            cursor_name=data['cursor_name'],
+            method=DbMethod.deserialize(data['method']),
+            args=data['args'],
+            kwargs=data['kwargs'],
+        )
```
yabirgb, line 95: this method cna raise deserialization error and valueerror afaik. Are those handled? State it in the docstring

rotkehlchen/db/drivers/server.py
```diff
@@ -0,0 +1,246 @@
...
+    about the DB operation to be executed. The server will respond with a
+    ZMQReturnData message containing the result of the operation or an error if it fails.
+    """
+
+    def __init__(self, port: int) -> None:
+        self.db_connections: dict[str, UnderlyingConnection] = {}
+        self.zmq_connection = zmq.Context().socket(zmq.REP)
+        self.zmq_connection.bind(f'{DB_CONNECTION_ADDRESS}:{port}')
+        self.db_cursors: defaultdict[str, defaultdict[str, UnderlyingCursor | None]] = defaultdict(defaultdict)  # noqa: E501
+        self.shutting_down = False
```
yabirgb, line 137: can you explain in a comment why you need this variable?

rotkehlchen/start.py
```diff
@@ -0,0 +1,96 @@
...
+        description='The rotki orchestrator to run other rotki processes.',
+    ).parse_args()
+
+    if args.process == 'all':
+        executable = [sys.executable] + sys.orig_argv[1:]
+        for process_type, process_data in manager.processes.items():
+            process_data.process = Popen(executable + ['--process', process_type.value])  # noqa: S603
+
+        if (backend_process := manager.processes[ProcessType.API_SERVER].process) is not None:
+            backend_process.wait()
```
yabirgb, line 90: One thing that I noticed is that you don't check that all the processes are running right? It can happen that the db writter fails to start and you don't check it

rotkehlchen/db/drivers/server.py
```diff
@@ -0,0 +1,227 @@
...
+            try:
+                call_data: ZMQCallData = self.zmq_connection.recv_pyobj()
+                if call_data.method.value == DbMethod.INITIALIZE.value:
+                    self._handle_connection(call_data)
+                    self.zmq_connection.send_pyobj(None)
+                    continue
+
+                try:
+                    db_connection = self.db_connections[call_data.db_path]
+                except KeyError:
```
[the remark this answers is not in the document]
OjusWiZard: here we store the connections in a mapping with their paths as keys, so we have only one connection per db. so no two connections can write to the same db at the same time.

rotkehlchen/db/drivers/server.py
```diff
@@ -0,0 +1,227 @@
...
+                    continue
+
+                if call_data.method.value == DbMethod.OPEN_CURSOR.value:
+                    self.db_cursors[call_data.db_path][call_data.cursor_name or 'default'] = db_connection.cursor()  # noqa: E501
+                    self.zmq_connection.send_pyobj(None)
+                    continue
+                if call_data.method.value == DbMethod.CLOSE_CURSOR.value:
+                    del self.db_cursors[call_data.db_path][call_data.cursor_name or 'default']
+                    self.zmq_connection.send_pyobj(None)
+                    continue
```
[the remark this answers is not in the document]
OjusWiZard: in that case, it will raise a KeyError because it will try to get the cursor of that db_path and find nothing. This error will be caught below and returned back to the client, and then raised there.

rotkehlchen/db/drivers/client.py
```diff
@@ -253,58 +359,64 @@ def __init__(
...
-            )
-        else:
-            self._conn = sqlcipher.connect(  # pylint: disable=no-member
-                database=str(path),
-                check_same_thread=False,
-                isolation_level=None,
-            )
+        connect_func = sqlite3.connect if connection_type == DBConnectionType.GLOBAL else sqlcipher.connect  # noqa: E501  # pylint: disable=no-member
+        connect_kwargs = {
+            'database': str(path),
```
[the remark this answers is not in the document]
OjusWiZard: so I had to add `self.db_path` because we also need to call `self.write_conn` sometimes, there we need to pass the `self.db_path`.

PR #8353, opened 2024-08-02
Properly handle curve pool with a specific coin == underlying

LefterisJP:
So finally found that bug that made USDT and PAXUSD have themselves as underlying tokens and create an infinite price recursion loop.

Since for curve it can happen that for a given pool some coins in the list have their own underlying coins and some have themselves use a set in a for loop to not repeat ourselves

Example pools. 

https://etherscan.io/address/0x06364f10B501e868329afBc005b3492902d6C763#readContract

https://etherscan.io/address/0xA5407eAE9Ba41422680e2e00537571bcC53efBfD#readContract

### DIFF
--- a/.github/workflows/task_backend_tests.yml
+++ b/.github/workflows/task_backend_tests.yml
@@ -97,7 +97,6 @@ jobs:
           (Get-Command python).Path
       - name: Install dependencies
         run: |
-          uv pip install --system 'setuptools<72'
           uv pip install --system --upgrade wheel pytest-cov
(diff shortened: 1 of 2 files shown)


LefterisJP:
Cleaned the cache of the CI. Will try again

LefterisJP:
I will merge this PR since all bugfixes CI runs seem to fail the tests/others test as seen here: https://github.com/rotki/rotki/pull/8354

We need to figure this out.

PR #8356, opened 2024-08-04
Fix ci tests

LefterisJP:


yabirgb (approved):
lgtm. In develop the decoders also run in parallel with xdist so there will be conflicts when merging

LefterisJP:
> lgtm. In develop the decoders also run in parallel with xdist so there will be conflicts when merging

YAAAAAS 
giphy (17)

PR #8357, opened 2024-08-04
Bump the regex package

LefterisJP:
In reality I just want to see full tests run in develop too as something may have broken during merge.


LefterisJP:
okay perfect. All green.

PR #8358, opened 2024-08-05
fix: show balances loading from beginning

lukicenturi:
show loading for dashboard percentage and chart from beginning, to not show sudden decrease when balances is not loaded

### DIFF
--- a/frontend/app/src/components/dashboard/OverallBalances.vue
+++ b/frontend/app/src/components/dashboard/OverallBalances.vue
@@ -23,9 +23,12 @@ const { totalNetWorth } = storeToRefs(statistics);
 const frontendStore = useFrontendSettingsStore();
 const { visibleTimeframes } = storeToRefs(frontendStore);
 
-const { isLoading: isSectionLoading } = useStatusStore();
+const { shouldShowLoadingScreen, isLoading: isSectionLoading } = useStatusStore();
 
-const isLoading = isSectionLoading(Section.BLOCKCHAIN);
+const isLoading = logicOr(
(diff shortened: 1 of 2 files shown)


frontend/app/src/components/dashboard/OverallBalances.vue
```diff
@@ -23,9 +23,20 @@ const { totalNetWorth } = storeToRefs(statistics);
...
+const isLoading = logicOr(isBlockchainLoading, isDashboardLoading);
+
+watch(isBlockchainLoading, (loading) => {
+  if (!loading) {
+    setStatus({
+      section: Section.DASHBOARD,
+      status: Status.LOADED,
+    });
+  }
+});
```
kelsos, line 39: What happens if you navigate away before loading completes, wait say in balances for the loading to complete and then navigate back to the dashboard?

lukicenturi: yeah you're right, let me update

lukicenturi: updated, should be more straightforward now

frontend/app/src/store/status.ts
```diff
@@ -87,7 +87,7 @@ export const useStatusStore = defineStore('status', () => {
   const shouldShowLoadingScreen = (section: Section, subsection: string = defaultSection) => computed<boolean>(() => {
     const statuses = get(status)[section];
     if (!statuses)
-      return false;
+      return true;
```
lukicenturi, line 90: this one should be true, because if it's not found, it means the status is None, hence it should be true

kelsos (approved):
lgtm

PR #8360, opened 2024-08-05
Add default for ERC20 properties of curve gauges

yabirgb:
Old curve gauges don't have the symbol, name, decimals methods. I asked in telegram and the default is 18. This PR sets default values for this attributes to avoid errors at the moment of populating the curve cache

Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/assets/utils.py
+++ b/rotkehlchen/assets/utils.py
@@ -278,7 +278,10 @@ def get_or_create_evm_token(
             except UnknownAsset:
                 asset_exists = False
 
-            if evm_inquirer is not None:
+            if (
+                None in (name, symbol, decimals) and
+                evm_inquirer is not None
+            ):
                 try:
                     name, symbol, decimals = _query_or_get_given_token_info(
                         evm_inquirer=evm_inquirer,
--- a/rotkehlchen/assets/utils.py
+++ b/rotkehlchen/assets/utils.py
@@ -215,6 +215,9 @@ def get_or_create_evm_token(
         encounter: TokenEncounterInfo | None = None,
         coingecko: str | None = None,
         cryptocompare: str | None = None,
+        fallback_decimals: int | None = None,
+        fallback_name: str | None = None,
+        fallback_symbol: str | None = None,
 ) -> EvmToken:
     """Given a token address return the <EvmToken>
 
@@ -225,6 +228,9 @@ def get_or_create_evm_token(
 
     Optionally the caller can provide a transaction hash of where the token was seen.
     This is used in the websocket message to provide information to the frontend.
+
+    If fallback values are provided and the token isn't ERC20 conformant we use those
+    as values for the decimal, name and symbol attributes.
 
     Note: if the token already exists but the other arguments don't match the
     existing token will still be silently returned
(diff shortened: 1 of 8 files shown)


rotkehlchen/chain/evm/decoding/curve/decoder.py
```diff
@@ -87,7 +87,7 @@ def __init__(
         ReloadablePoolsAndGaugesDecoderMixin.__init__(
             self,
             evm_inquirer=evm_inquirer,
-            cache_type_to_check_for_freshness=CacheType.CURVE_LP_TOKENS,
+            cache_type_to_check_for_freshness=CacheType.CURVE_POOL_ADDRESS,
```
LefterisJP, line 90: so this was a bug?

rotkehlchen/chain/evm/decoding/curve/curve_cache.py
```diff
@@ -265,6 +265,9 @@ def _ensure_curve_tokens_existence(
                         token_kind=EvmTokenKind.ERC20,
                         weight=ONE,
                     )],
+                    decimals=18,  # 18 is the default for old gauges that don't implement the decimals method https://t.me/curvefi/654915  # noqa: E501
```
LefterisJP, line 268:
```suggestion
                    decimals=18,  # all gauges have 18 decimals https://t.me/curvefi/654915. Hard code since some contracts have no name/symbol/decimals.  # noqa: E501
```

rotkehlchen/chain/evm/decoding/curve/curve_cache.py
```diff
@@ -493,13 +496,14 @@ def query_curve_data(
...
     with GlobalDBHandler().conn.read_ctx() as cursor:
-        existing_pools = [
-            string_to_evm_address(address)
-            for address in globaldb_get_general_cache_like(
-                cursor=cursor,
-                key_parts=(cache_type, str(inquirer.chain_id.serialize_for_db())),
+        existing_pools = {  # query the pools that we already have in the db
+            string_to_evm_address(address[0])
+            for address in cursor.execute(
+                'SELECT value FROM unique_cache WHERE key LIKE ?',
```
LefterisJP, line 502: why this change?

LefterisJP (commented):
Ah also @yabirgb we will need a changelog here

rotkehlchen/chain/evm/decoding/curve/curve_cache.py
yabirgb: We need to access the unique_cache instead of the general cache

rotkehlchen/chain/evm/decoding/curve/decoder.py
yabirgb: yes, reading the wrong cache tyype

rotkehlchen/assets/utils.py
```diff
@@ -215,6 +215,9 @@ def get_or_create_evm_token(
         encounter: TokenEncounterInfo | None = None,
         coingecko: str | None = None,
         cryptocompare: str | None = None,
+        fallback_decimals: int | None = None,
```
LefterisJP, line 218: Why do this? In our usecase it's not a fallback.

In our usecase they are all the same, no ?

If provided instead just don't query the chain at all. Why query, only to see it's missing and then use the given one? That makes no sense.

rotkehlchen/tests/unit/test_tokens.py
```diff
@@ -261,3 +265,75 @@ def test_flaky_binding_parameter_zero(
     with database.conn.read_ctx() as cursor:
         for address in ethereum_accounts:
             database.get_tokens_for_address(cursor, address, SupportedBlockchain.ETHEREUM)
+
+
+@pytest.mark.parametrize('number_of_eth_accounts', [1])
+def test_old_curve_gauge(ethereum_inquirer: 'EthereumInquirer'):
+    """Test that querying new and old gauges get the data correctly
```
LefterisJP, line 272:
```suggestion
    """Test that querying new and old gauges get the data correctly.
```

rotkehlchen/tests/unit/test_tokens.py
```diff
@@ -261,3 +265,75 @@ def test_flaky_binding_parameter_zero(
...
+    Old one should peak the default values provided and the new one should
```
LefterisJP, line 273:
```suggestion
    Old one should pick the default values provided and the new one should
```

LefterisJP (approved):
LGTM

PR #8365, opened 2024-08-06
feat(frontend): accounts mass edit and delete

### LINKED ISSUE #8344: Allow editing accounts label and tags groups
# Description

For evm groups enable editing on the group level that allows you to edit both the label and tags for a group.

- hide the labels in the evm account breakdown and for edit make sure that whatever label/tag you are setting it updates the whole group

# Specification
## Backend:
Create endpoint to edit group of accounts with same address
Should work like https://rotki.readthedocs.io/en/latest/api.html#patch--api-(version)-blockchains-(name)-accounts, but no chain need to be passed.

## Frontend
Adjust the table, put edit button on the group of chains too.

lukicenturi:
Closes #8344

## Checklist

### DIFF
--- a/frontend/app/src/components/accounts/AccountBalancesTable.vue
+++ b/frontend/app/src/components/accounts/AccountBalancesTable.vue
@@ -270,8 +270,6 @@ defineExpose({
           v-if="!('virtual' in row) || !row.virtual"
           class="account-balance-table__actions"
           :edit-tooltip="t('account_balances.edit_tooltip')"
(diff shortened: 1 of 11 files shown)


kelsos (approved):
There is a thing I noticed that needs some investigation but we can do it after merging.

When you edit the group and add the tags, under some conditions when going to the account breakdown for the first time it won't show you the new tag.

kelsos:
related to #8040

PR #8366, opened 2024-08-06
Updated yanked dep

yabirgb:
https://pypi.org/project/types-pkg-resources/ states to use types-setuptools


LefterisJP (approved):
LGTM if tests pass

PR #8371, opened 2024-08-07
Add payment/grant event subtypes and event categories

LefterisJP:


rotkehlchen/accounting/constants.py
```diff
@@ -33,6 +33,8 @@
         HistoryEventSubType.DONATE: EventCategory.RECEIVE_DONATION,
         HistoryEventSubType.NONE: EventCategory.RECEIVE,
         HistoryEventSubType.LIQUIDATE: EventCategory.LIQUIDATION_REWARD,
+        HistoryEventSubType.PAYMENT: EventCategory.RECEIVE_PAYMENT,
+        HistoryEventSubType.GRANT: EventCategory.RECEIVE_GRANT,
```
yabirgb, line 37: where are those 2 defined? I don't see them in bugfixes

LefterisJP: In this PR. But one mapping is missing which I will push.

yabirgb (approved):
lgtm

PR #8377, opened 2024-08-08
fix(frontend): only refresh account that being deleted

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/composables/accounts/blockchain/use-account-delete.ts
+++ b/frontend/app/src/composables/accounts/blockchain/use-account-delete.ts
@@ -1,3 +1,4 @@
+import { Blockchain } from '@rotki/common/lib/blockchain';
 import type { Account } from '@rotki/common/lib/account';
 import type {
   BlockchainAccountGroupWithBalance,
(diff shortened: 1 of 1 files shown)


frontend/app/src/composables/accounts/blockchain/use-account-delete.ts
```diff
@@ -82,11 +88,15 @@ export function useAccountDelete() {
...
+      for (const xpub of xpubs) {
+        chainsToRefresh.push(xpub.chain);
         await deleteXpub(xpub);
+      }
     }
 
-    startPromise(refreshAccounts());
+    chainsToRefresh.filter(uniqueStrings).forEach((chain) => {
+      startPromise(refreshAccounts(chain));
+    });
```
kelsos, line 99: wouldn't it make sense here to limit how many tasks we start in parallel?

kelsos: Ok that's pre-unificatioan so it's a single chain per delete most of the time.

kelsos (approved):
lgtm

PR #8385, opened 2024-08-09
Reduce the work when starting the yearn model

yabirgb:
Avoids querying all the token details and the underlying tokens if we only need the address of the token

Closes #(issue_number)

## Checklist


LefterisJP (approved):
LGTM

PR #8393, opened 2024-08-12
Eigenlayer do not double count eigenpod

LefterisJP:
Eigenlayer native restaking exited balances residing in eigenpod will no longer be double counted.


LefterisJP:
I may have to revert most of this PR

PR #8400, opened 2024-08-13
Fix native eth restaking withdrawal double counting

LefterisJP:
Reverses most of https://github.com/rotki/rotki/pull/8393 as Eigenpod balances should be counted still.

But makes sure to not count pending withdrawals if it's native ETH since their balance is already in the Eigenpod.

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/eigenlayer/balances.py
+++ b/rotkehlchen/chain/ethereum/modules/eigenlayer/balances.py
@@ -162,6 +162,9 @@ class EigenlayerBalances(ProtocolWithBalance):
         )
         for address, event_list in addresses_with_withdrawals.items():
             for event in event_list:
+                if event.asset == A_ETH:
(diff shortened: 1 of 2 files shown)


rotkehlchen/chain/ethereum/modules/eigenlayer/balances.py
```diff
@@ -162,6 +162,9 @@ def _query_token_pending_withdrawals(self, balances: 'BalancesSheetType') -> 'Ba
         )
         for address, event_list in addresses_with_withdrawals.items():
             for event in event_list:
+                if event.asset == A_ETH:
```
LefterisJP, line 165: All of the PR is re-adding what I removed before. Except this. This is the actual fix that should have been done.

We count all in-flight withdrawals from here by checking withdrawal events. For native restaking we also check the eigenpod. So for the native ETH asset we should not count the withdrawals.

PR #8402, opened 2024-08-14
Add decoding of payments for cowswap vCOW claims

nicholasyoder:


rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -370,6 +370,10 @@ def _decode_normal_claim(self, context: DecoderContext) -> DecodingOutput:
         amount = asset_normalized_value(amount=raw_amount, asset=self.vcow_token)
         airdrop_identifier: Literal['cow_mainnet', 'cow_gnosis'] = 'cow_mainnet' if self.evm_inquirer.chain_id == ChainID.ETHEREUM else 'cow_gnosis'  # noqa: E501
         for event in context.decoded_events:
+            if event.event_type == HistoryEventType.SPEND and event.counterparty != 'gas':
+                event.event_subtype = HistoryEventSubType.PAYMENT
+                event.notes = f'Pay {event.balance.amount} {event.asset.symbol_or_name()} to claim vCOW'  # noqa: E501
+                continue
```
yabirgb, line 376: image
Checking the transaction there is an attribute claimType that I suspect would tell you if you need to perform this check or not. You can read it by using `hex_or_bytes_to_int(context.tx_log.data[32:64])`. Before entering the loop you can check that value and do something like

```suggestion
            if claim_has_payment and event.event_type == HistoryEventType.SPEND and event.counterparty != 'gas':
                event.event_subtype = HistoryEventSubType.PAYMENT
                event.notes = f'Pay {event.balance.amount} {event.asset.symbol_or_name()} to claim vCOW'  # noqa: E501
                continue
```

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -370,6 +370,10 @@ def _decode_normal_claim(self, context: DecoderContext) -> DecodingOutput:
         amount = asset_normalized_value(amount=raw_amount, asset=self.vcow_token)
         airdrop_identifier: Literal['cow_mainnet', 'cow_gnosis'] = 'cow_mainnet' if self.evm_inquirer.chain_id == ChainID.ETHEREUM else 'cow_gnosis'  # noqa: E501
         for event in context.decoded_events:
+            if event.event_type == HistoryEventType.SPEND and event.counterparty != 'gas':
```
yabirgb, line 373: Is rare that you need to compare against strings in our code. There is almost always a constant 

```suggestion
            if event.event_type == HistoryEventType.SPEND and event.counterparty != CPT_GAS:
```

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
yabirgb, line 376: notice that below the code that you added there is a break statment. I believe that the vcow receive always happens after the payment but you should add a comment stating this. Someone that is not aware could change the code and break the assumption here

yabirgb: Forgot it in the review. There in the data of the log event you also have the address. Check too that the address is the same as the location label of the event

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -369,7 +370,13 @@ def _decode_normal_claim(self, context: DecoderContext) -> DecodingOutput:
         raw_amount = hex_or_bytes_to_int(context.tx_log.data[128:160])
         amount = asset_normalized_value(amount=raw_amount, asset=self.vcow_token)
         airdrop_identifier: Literal['cow_mainnet', 'cow_gnosis'] = 'cow_mainnet' if self.evm_inquirer.chain_id == ChainID.ETHEREUM else 'cow_gnosis'  # noqa: E501
+        claim_has_payment = hex_or_bytes_to_int(context.tx_log.data[32:64]) in (1, 2)
```
yabirgb, line 373: add a comment explaining what 1 and 2 mean. You or anyone in the team will come in the future and will need to research it

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -369,7 +370,13 @@ def _decode_normal_claim(self, context: DecoderContext) -> DecodingOutput:
...
         for event in context.decoded_events:
+            # Claim event always follows payment. Continue on payment, break on claim.
+            if claim_has_payment and event.event_type == HistoryEventType.SPEND and event.counterparty != CPT_GAS:  # noqa: E501
```
yabirgb, line 376: I added it later but also requested to check against the location label since the address making the payment should be in the data too

rotkehlchen/tests/unit/decoders/test_cowswap.py
```diff
@@ -879,6 +879,136 @@ def test_gnosis_vested_claim(database, gnosis_inquirer, gnosis_accounts):
...
+        database=database,
+        tx_hash=tx_hash,
+    )
+    timestamp, gas_amount, claim1_amount, payment_amount, claim2_amount = (
+        TimestampMS(1645776935000),
+        '0.000259548002076384',
+        '3559.387459031416241036',
+        '2.085578589276220452',
+        '3559.387459031416239786',
+    )
```
yabirgb, line 953: I believe you had this in one line but keep it in one line like the other test

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
yabirgb, line 376: this if check you can split it in different lines using parenthesis

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -369,7 +370,20 @@ def _decode_normal_claim(self, context: DecoderContext) -> DecodingOutput:
         raw_amount = hex_or_bytes_to_int(context.tx_log.data[128:160])
         amount = asset_normalized_value(amount=raw_amount, asset=self.vcow_token)
         airdrop_identifier: Literal['cow_mainnet', 'cow_gnosis'] = 'cow_mainnet' if self.evm_inquirer.chain_id == ChainID.ETHEREUM else 'cow_gnosis'  # noqa: E501
+        # claimTypes with payment: 1=GnoOption, 2=UserOption, 3=Investor
+        claim_has_payment = hex_or_bytes_to_int(context.tx_log.data[32:64]) in (1, 2, 3)
+        claimant_address = hex_or_bytes_to_address(context.tx_log.data[64:96])
```
LefterisJP, line 375: Shouldn't you also check if claimant address is tracked?

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -369,7 +370,20 @@ def _decode_normal_claim(self, context: DecoderContext) -> DecodingOutput:
...
         for event in context.decoded_events:
+            # Claim event always follows payment. Continue on payment, break on claim.
+            if (
+                claim_has_payment and
+                event.location_label == claimant_address and
+                event.event_type == HistoryEventType.SPEND and
+                event.counterparty != CPT_GAS
```
LefterisJP, line 382: Shouldn't you check for the asset here? Excluding gas seems arbitrary .

The assets with which you could pay for this are definitely limited

LefterisJP (commented):
Left some comments. But also from an accounting perspective perhaps I am wrong but the payment does not make sense.

Should it not be seen as a swap instead? Otherwise accounting for this won't work.

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
yabirgb: Yes missed this. Can be only ETH, XDAI or GNO unless I'm mistaken

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
yabirgb: For this @nicholasyoder we have in self.base a `is_tracked` method

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -369,7 +372,27 @@ def _decode_normal_claim(self, context: DecoderContext) -> DecodingOutput:
...
+        # claimTypes with payment: 1=GnoOption, 2=UserOption, 3=Investor
+        claim_supports_payment = hex_or_bytes_to_int(context.tx_log.data[32:64]) in (1, 2, 3)
+        claimant_address = hex_or_bytes_to_address(context.tx_log.data[64:96])
+        claim_has_payment = False
+        out_event = in_event = None
         for event in context.decoded_events:
+            # Claim event always follows payment. Continue on payment, break on claim.
+            if (
+                claim_supports_payment and
+                self.base.is_tracked(claimant_address) and
```
yabirgb, line 384: claimant_address doesn't get modified so you don't need to check this on every iteration of the loop. If it is not tracked you need to exit this logic but before entering the for loop

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -378,11 +401,21 @@ def _decode_normal_claim(self, context: DecoderContext) -> DecodingOutput:
...
+                if claim_has_payment:
+                    event.event_type = HistoryEventType.TRADE
+                    event.event_subtype = HistoryEventSubType.RECEIVE
+                    in_event = event
                 break
 
         else:
             log.error(f'Could not find the normal COW token claim for {self.evm_inquirer.chain_name} transaction {context.transaction.tx_hash.hex()}')  # noqa: E501
 
+        if claim_has_payment:
```
yabirgb, line 413: to prevent errors check that `None not in (out_event, in_event)`

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -333,6 +335,7 @@ def __init__(
         )
         self.vcow_token = vcow_token.resolve_to_evm_token()
         self.cow_token = cow_token.resolve_to_evm_token()
+        self.gno_token = gno_token
```
LefterisJP, line 338: you need to do the same as above. `resolve_to_evm_token()`.

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
LefterisJP: This is not needed. If either is None then `maybe_reshuffle_events` will do nothing. But since yet it is an error I would log an error instead here.

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -369,7 +371,28 @@ def _decode_normal_claim(self, context: DecoderContext) -> DecodingOutput:
         raw_amount = hex_or_bytes_to_int(context.tx_log.data[128:160])
         amount = asset_normalized_value(amount=raw_amount, asset=self.vcow_token)
         airdrop_identifier: Literal['cow_mainnet', 'cow_gnosis'] = 'cow_mainnet' if self.evm_inquirer.chain_id == ChainID.ETHEREUM else 'cow_gnosis'  # noqa: E501
+        # claimTypes with payment: 1=GnoOption, 2=UserOption, 3=Investor
+        claim_supports_payment = hex_or_bytes_to_int(context.tx_log.data[32:64]) in (1, 2, 3)
+        claimant_address = hex_or_bytes_to_address(context.tx_log.data[64:96])
+        if not self.base.is_tracked(claimant_address):
+            return DEFAULT_DECODING_OUTPUT
+        claim_has_payment = False
```
LefterisJP, line 379: Leave newlines between codeblocks like this for readability

```suggestion
            return DEFAULT_DECODING_OUTPUT

        claim_has_payment = False
```

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -378,11 +401,24 @@ def _decode_normal_claim(self, context: DecoderContext) -> DecodingOutput:
                 counterparty=CPT_COWSWAP,
                 airdrop_identifier=airdrop_identifier,
             ):
+                if claim_has_payment:
+                    event.event_type = HistoryEventType.TRADE
+                    event.event_subtype = HistoryEventSubType.RECEIVE
+                    in_event = event
                 break
```
LefterisJP, line 408: Shouldn't the break be in there? Or we break on outer if and in inner if just change type and remember in event?

```suggestion
                    in_event = event
                    break
```

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -369,7 +371,28 @@ def _decode_normal_claim(self, context: DecoderContext) -> DecodingOutput:
...
+                event.event_type == HistoryEventType.SPEND and
+                event.event_subtype == HistoryEventSubType.NONE and
+                event.asset in (self.evm_inquirer.native_token, self.gno_token)
+            ):
+                event.event_type = HistoryEventType.TRADE
+                event.event_subtype = HistoryEventSubType.SPEND
+                event.notes = f'Pay {event.balance.amount} {event.asset.symbol_or_name()} to claim vCOW'  # noqa: E501
+                claim_has_payment = True
+                out_event = event
+                continue
```
LefterisJP, line 395: same nitpick here

```suggestion
                continue
                
```

rotkehlchen/tests/unit/decoders/test_cowswap.py
```diff
@@ -879,6 +879,130 @@ def test_gnosis_vested_claim(database, gnosis_inquirer, gnosis_accounts):
...
+        ), EvmEvent(
+            sequence_index=10,
+            timestamp=timestamp,
+            location=Location.GNOSIS,
+            event_type=HistoryEventType.RECEIVE,
+            event_subtype=HistoryEventSubType.AIRDROP,
+            asset=A_GNOSIS_VCOW,
+            balance=Balance(amount=FVal(claim1_amount)),
+            location_label=user_address,
+            notes=f'Claim {claim1_amount} vCOW from cowswap airdrop',
```
LefterisJP, line 970: So in this transaction we have 1 normal airdrop claim and 1 we pay for?

nicholasyoder: That seems to be the case as far as I can tell, but I'm certainly not an expert on understanding exactly what is going on in a transaction like this. You probably know better than I do.

LefterisJP: I have seen many transactions but for each protocol you study, by studying the code of the contract and going in the discord/forums of the protocol it is your job to also understand what happens, and write the decoder correctly.

I also would not know how each protocol works.

LefterisJP (approved):
LGTM.

PR #8404, opened 2024-08-15
Handle tags format in db upgrade

yabirgb:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=73549327

### DIFF
--- a/rotkehlchen/db/upgrades/v43_v44.py
+++ b/rotkehlchen/db/upgrades/v43_v44.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from typing import TYPE_CHECKING
 
 from rotkehlchen.logging import RotkehlchenLogsAdapter, enter_exit_debug_log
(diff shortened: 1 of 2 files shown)


rotkehlchen/db/upgrades/v43_v44.py
```diff
@@ -62,18 +62,64 @@ def _remove_log_removed_column(write_cursor: 'DBCursor') -> None:
...
+    depend on the supported blockchain
+    """
+    object_references: list[tuple[str, str]] = write_cursor.execute(
+        'SELECT object_reference, tag_name FROM tag_mappings',
+    ).fetchall()
+    if len(object_references) == 0:
+        return
+
+    remove_keys, insert_entries = [], set()
+    suppported_blockchains = (  # supported blockchains at the moment of writing the upgrade
```
OjusWiZard, line 76:
```suggestion
    supported_blockchains = (  # supported blockchains at the moment of writing the upgrade
```

rotkehlchen/db/upgrades/v43_v44.py
```diff
@@ -62,18 +62,64 @@ def _remove_log_removed_column(write_cursor: 'DBCursor') -> None:
...
+        'GNOSIS',
+        'SCROLL',
+        'ZKSYNC_LITE',
+    )
+    for (object_reference, tag_name) in object_references:
+        for supported_chain in suppported_blockchains:
+            if object_reference.startswith(supported_chain):
+                remove_keys.append(object_reference)
+                insert_entries.add((object_reference.removeprefix(supported_chain), tag_name))
+                break
```
OjusWiZard, line 97: we can use a regex here like: `^(ETH|ETH2|BTC|BCH|KSM|AVAX|DOT|OPTIMISM|POLYGON_POS|ARBITRUM_ONE|BASE|GNOSIS|SCROLL|ZKSYNC_LITE)`
and match it with the `object_reference`, then use `re.sub` to remove that part. This way we can avoid running a nested loop for each entry.

rotkehlchen/db/upgrades/v43_v44.py
```diff
@@ -62,18 +62,64 @@ def _remove_log_removed_column(write_cursor: 'DBCursor') -> None:
...
+
+    write_cursor.executemany(
+        'DELETE FROM tag_mappings WHERE object_reference=?',
+        [(reference,) for reference in remove_keys],
```
OjusWiZard, line 101: we can directly append `(object_reference,)` above and avoid another loop here

yabirgb:
Wen your college preferes regex 
image

LefterisJP (approved):
LGTM

Issue #8350, opened 2024-08-02
Prevent addressbook duplication addition


lukicenturi:
So for endpoint <PUT> names/addressbook/private or global
I am not prohibited to put multiple entries with same blockchain and address

<img width="1143" alt="image" src="https://github.com/user-attachments/assets/03365bb1-85fc-4f4f-b252-fcb72d63ea5b">

This should not happen,
If I add new entry with a blockchain and address, the previous entry with same blockchain & address should be deleted.

yabirgb:
Is probably my old friend: nullable column in composed primary key https://github.com/rotki/rotki/blob/a8bdd0db91d4f3b6060127dc7a0adba383e4e999/rotkehlchen/globaldb/schema.py#L247

yabirgb:
for this task:

- In bugfixes create a data migration that uses a constant instead of NULL in the db to avoid the unique check with the NULL constant
- In develop create a db upgrade to change the type of the column to `NOT NULL` to avoid having this issue in the future

PR #8405, opened 2024-08-15
Fix addressbook

yabirgb:
Set the blockchain value to a constant string instead of None since the primary key of the table includes the nullable column.

This avoid having duplicated entries. Also ship a data migration for both the user db and the globaldb to update the tuples. In the case of duplicates we delete them except the one with the highest row id.

Related #8350

**NOTE**:

This is for bugfixes. In develop we need to issue a new db upgrade step where we mark the column as non nullable

### DIFF
--- a/rotkehlchen/types.py
+++ b/rotkehlchen/types.py
@@ -910,6 +910,9 @@ class CostBasisMethod(SerializableEnumNameMixin):
     ACB = auto()
 
 
+ANY_BLOCKCHAIN_ADDRESS_BOOK_VALUE: Final = 'NONE'  # blockchain value used to mark in the DB that the address entry is valid for any blockchain  # noqa: E501
+
+
 class AddressbookEntry(NamedTuple):
     address: BlockchainAddress
     name: str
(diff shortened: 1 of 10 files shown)


LefterisJP:
> NOTE:
> This is for bugfixes. In develop we need to issue a new db upgrade step where we mark the column as non nullable

Can you then:

1. Not close this here
2. Describe in the issue what needs to be done in bugfixes and what needs to be done in develop?

rotkehlchen/types.py
```diff
@@ -910,6 +910,9 @@ class CostBasisMethod(SerializableEnumNameMixin):
     ACB = auto()
 
 
+ANY_BLOCKCHAIN_ADDRESS_BOOK_VALUE: Final = 'NONE'  # value used to mark in the DB that the address entry is valid for any blockchain  # noqa: E501
```
LefterisJP, line 913: Just a suggestion to shorten it. Addressbook can be one word. Also comment.

```suggestion
ANY_BLOCKCHAIN_ADDRESSBOOK_VALUE: Final = 'NONE'  # blockchain value used to mark in the DB that the address entry is valid for any blockchain  # noqa: E501
```

yabirgb:
Changed the close and added a comment in the issue

rotkehlchen/types.py
```diff
@@ -910,6 +910,9 @@ class CostBasisMethod(SerializableEnumNameMixin):
     ACB = auto()
 
 
+ANY_BLOCKCHAIN_ADDRESS_BOOK_VALUE: Final = 'NONE'  # blockchain value used to mark in the DB that the address entry is valid for any blockchain  # noqa: E501
```
LefterisJP, line 913: didn't change this as requested

LefterisJP (approved):
LGTM. Just missing a change but not important enough to ask for repush.

PR #8407, opened 2024-08-16
Improvements to addressbook

yabirgb:
- Avoid loading the fetchers on every request to the api. The frontend has a cache of ~10 mins but every time a new address is displayed the app was adding the fetchers
- Rename variable from previous PR


rotkehlchen/rotkehlchen.py
```diff
@@ -509,6 +510,8 @@ def unlock_user(
             method=self.data_updater.check_for_updates,
         )
 
+        self.addressbook_prioritizer = NamePrioritizer(self.data.db)
+
```
LefterisJP, line 514: You need to explain why we keep this in memory here. Also remove empty newline

```suggestion
        self.addressbook_prioritizer = NamePrioritizer(self.data.db)  # initialize here since it's reused by the api for addressbook endpoints
```

LefterisJP (approved):
LGTM

PR #8409, opened 2024-08-16
Feat (Aave): Decode interests separately

OjusWiZard:
## Checklist

- [x] Decode interest earned from Aave separately as receive/interest events


rotkehlchen/chain/evm/decoding/aave/v3/decoder.py
```diff
@@ -116,6 +117,19 @@ def _decode_incentives(self, context: 'DecoderContext') -> DecodingOutput:
...
         )
 
+    def _decode_mint_from_transfer(self, context: 'DecoderContext') -> DecodingOutput:
+        for event in context.decoded_events:
+            if (
+                event.event_type == HistoryEventType.RECEIVE and
+                event.event_subtype == HistoryEventSubType.NONE and
+                event.address == ZERO_ADDRESS
+            ):
+                event.event_subtype = HistoryEventSubType.INTEREST
```
yabirgb, line 127: we have it as receive wrapped. I believe we shouldn't change it... is there a reason to introduce a new type?

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`-` Rotki will now decode interests earned from aave separately
```
```suggestion
* :feature:`-` Rotki will now decode interest earned from aave as independent events.
```

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -123,33 +126,62 @@ def _decode_collateral_events(
...
             asset=token,
         )
+
+        earned_event = None  # find the mint event of aToken
+        for _log in context.all_logs:
+            if (
+                _log.topics[0] == MINT and ((
+                    on_behalf_of is not None and
+                    hex_or_bytes_to_address(_log.topics[2]) == on_behalf_of
+                ) or on_behalf_of is None) and
```
yabirgb, line 152: move the `on_behalf_of is None` first and since I expect this to be the most common case you will avoid the hex_or_bytes_to_address check

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -171,36 +204,75 @@ def _decode_deposit(
...
+        return deposit_event, receive_event, earned_event
 
     def _decode_withdrawal(
             self,
             token: 'EvmToken',
-            tx_log: EvmTxReceiptLog,
-            decoded_events: list['EvmEvent'],
-    ) -> tuple[EvmEvent | None, EvmEvent | None]:
+            context: 'DecoderContext',
+    ) -> tuple[EvmEvent | None, EvmEvent | None, EvmEvent | None]:
```
yabirgb, line 218: have you checked if the logic works the same in v2 and v3? From the chat in tg I kind of recall that it was the same

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -171,36 +204,75 @@ def _decode_deposit(
...
-            return None, None
+            return None, None, None
         amount = asset_normalized_value(
-            amount=hex_or_bytes_to_int(tx_log.data),
+            amount=hex_or_bytes_to_int(context.tx_log.data),
             asset=token,
         )
+
+        earned_event = None  # find the mint or burn event of aToken
+        for _log in context.all_logs:
```
yabirgb, line 235: question... does it find only once? If it is only once add a break statment in the for loop. If it is more then I believe you want this to be a list and not a single event

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -171,36 +204,75 @@ def _decode_deposit(
...
+                context.decoded_events.append(earned_event := self.base.make_event_from_transaction(  # noqa: E501
+                    transaction=context.transaction,
+                    tx_log=_log,
+                    event_type=HistoryEventType.RECEIVE,
+                    event_subtype=HistoryEventSubType.INTEREST,
+                    asset=(earned_token := Asset(evm_address_to_identifier(
+                        address=earned_token_address,
+                        chain_id=self.evm_inquirer.chain_id,
+                        token_type=EvmTokenKind.ERC20,
+                    )).resolve_to_crypto_asset()),
```
yabirgb, line 264: use get_or_create_evm_token

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -123,33 +126,62 @@ def _decode_collateral_events(
...
-            return None, None
+            return None, None, None
         amount = asset_normalized_value(
-            amount=hex_or_bytes_to_int(tx_log.data[32:64]),
+            amount=hex_or_bytes_to_int(context.tx_log.data[32:64]),
             asset=token,
         )
+
+        earned_event = None  # find the mint event of aToken
+        for _log in context.all_logs:
```
yabirgb, line 147: It is supoosed to be only one event? If so break the for loop when you find it. Otherwise it might be a list what you want here

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -171,36 +204,75 @@ def _decode_deposit(
...
         ):
-            return None, None
+            return None, None, None
         amount = asset_normalized_value(
-            amount=hex_or_bytes_to_int(tx_log.data),
+            amount=hex_or_bytes_to_int(context.tx_log.data),
             asset=token,
         )
+
+        earned_event = None  # find the mint or burn event of aToken
```
yabirgb, line 234: can you maybe rename this to `accrue_interest_event`

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -171,36 +204,75 @@ def _decode_deposit(
...
+                    )
+                ) > 0
+            ):  # parse the mint amount and balance_increase
+                earned_token_address = _log.address
+                if _log.topics[0] == BURN:
+                    amount = asset_normalized_value(
+                        amount=hex_or_bytes_to_int(_log.data[:32]),
+                        asset=token,
+                    )
+                    earned_token_address = reserve
```
yabirgb, line 254: `accrued_token_address` here. I believe is clearer here becayuse you can accrue debt that you pay I believe

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -171,36 +204,75 @@ def _decode_deposit(
...
+                        asset=token,
+                    )
+                ) > 0
+            ):  # parse the mint amount and balance_increase
+                earned_token_address = _log.address
+                if _log.topics[0] == BURN:
+                    amount = asset_normalized_value(
+                        amount=hex_or_bytes_to_int(_log.data[:32]),
+                        asset=token,
+                    )
```
yabirgb, line 253: add a comment explaining what this amount represents

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -221,19 +293,26 @@ def _decode_withdrawal(
...
+                    event.balance.amount = amount
                     withdraw_event = event
                     event.counterparty = self.counterparty
                 elif (
-                    event.event_type != HistoryEventType.RECEIVE and
-                    (event.address == ZERO_ADDRESS or event.address in self.native_gateways) and
-                    self._token_is_aave_contract(event.asset)
+                    event.event_type != HistoryEventType.RECEIVE or ((
+                        earned_event is not None and
+                        earned_event.balance.amount > event.balance.amount
```
yabirgb, line 302: can you add a comment stating that this is the case of accrued interest

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -221,19 +293,26 @@ def _decode_withdrawal(
...
+                    ) and (
+                        event.address == ZERO_ADDRESS or
+                        event.address in self.native_gateways
```
yabirgb, line 305: and thhis case I'm nto sure what it is tbh

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -171,36 +204,75 @@ def _decode_deposit(
...
         amount = asset_normalized_value(
-            amount=hex_or_bytes_to_int(tx_log.data),
+            amount=hex_or_bytes_to_int(context.tx_log.data),
             asset=token,
         )
+
+        earned_event = None  # find the mint or burn event of aToken
+        for _log in context.all_logs:
+            if (
+                _log.topics[0] in {MINT, BURN} and
```
yabirgb, line 237: in this case I believe is better to have a tuple than a set because it builds faster

rotkehlchen/chain/evm/decoding/aave/v3/decoder.py
```diff
@@ -116,6 +117,19 @@ def _decode_incentives(self, context: 'DecoderContext') -> DecodingOutput:
             amount_raw=context.tx_log.data[32:64],
         )
 
+    def _decode_mint_from_transfer(self, context: 'DecoderContext') -> DecodingOutput:
+        for event in context.decoded_events:
+            if (
```
yabirgb, line 122: is it only one? If so you can break the loop after finding it

rotkehlchen/tests/unit/decoders/test_aave_v3.py
```diff
@@ -220,6 +232,73 @@ def test_aave_v3_deposit(database, ethereum_inquirer, ethereum_accounts) -> None
...
+            tx_hash=tx_hash,
+            sequence_index=3,
+            timestamp=timestamp,
+            location=Location.ETHEREUM,
+            event_type=HistoryEventType.RECEIVE,
+            event_subtype=HistoryEventSubType.INTEREST,
+            asset=EvmToken('eip155:1/erc20:0x5Ee5bf7ae06D1Be5997A1A72006FE6C607eC6DE8'),
+            balance=Balance(amount=FVal(interest_amount)),
+            location_label=ethereum_accounts[0],
+            notes=f'Receive {interest_amount} aEthWBTC as interest earned from AAVE v3',
```
yabirgb, line 295: I would use accrued rather than earned but you can also ask lefteris that has a better english than me xd

rotkehlchen/tests/unit/decoders/test_aave_v3.py
```diff
@@ -1226,3 +1396,57 @@ def test_aave_v3_withdraw_xdai(database, gnosis_inquirer, gnosis_accounts) -> No
         ),
     ]
     assert events == expected_events
+
+
+@pytest.mark.vcr
+@pytest.mark.parametrize('ethereum_accounts', [['0x79A4D7448B110bf926d96b726D113f7d7De6781f']])
+def test_aave_v3_interest_on_transfer(database, ethereum_inquirer, ethereum_accounts) -> None:
+    tx_hash = deserialize_evm_tx_hash('0x84a70d14ef53987e1ee267435492100f07a4480b7d8022a8d3508e3a4ee70fbf')  # noqa: E501
+    events, _ = get_decoded_events_of_transaction(
```
yabirgb, line 1405: do you have a test for withdrawing the 100% of the position?

rotkehlchen/tests/unit/decoders/test_aave_v3.py
```diff
@@ -1226,3 +1396,57 @@ def test_aave_v3_withdraw_xdai(database, gnosis_inquirer, gnosis_accounts) -> No
         ),
     ]
     assert events == expected_events
+
+
+@pytest.mark.vcr
```
yabirgb, line 1401: can you add to the cassettes the skip api_key param?

rotkehlchen/tests/unit/decoders/test_aave_v3.py
```diff
@@ -265,11 +344,90 @@ def test_aave_v3_withdraw(database, ethereum_inquirer, ethereum_accounts) -> Non
...
+            balance=Balance(amount=FVal(interest_amount)),
+            location_label=ethereum_accounts[0],
+            notes=f'Receive {interest_amount} USDT as interest earned from AAVE v3',
+            counterparty=CPT_AAVE_V3,
+        ),
+    ]
+    assert events == expected_events
+
+
+@pytest.mark.vcr
```
yabirgb, line 369: same for the other tests

rotkehlchen/chain/evm/decoding/aave/v3/decoder.py
OjusWiZard: yeah, this new type is to differentiate it from the regular wrapped event. This event is for the amount that is earned from the interest and help to do its accounting differently.

rotkehlchen/tests/unit/decoders/test_aave_v3.py
OjusWiZard: I guess accrued is when it's not receive in our wallet yet, and earned is when it's received. But yeah, can check with ser @LefterisJP also

rotkehlchen/tests/unit/decoders/test_aave_v3.py
OjusWiZard: yeah `test_aave_v3_disable_collateral` is that case

rotkehlchen/chain/evm/decoding/aave/common.py
OjusWiZard: yeah it's only one

rotkehlchen/chain/evm/decoding/aave/common.py
OjusWiZard: they seemingly have a different mechanism, but this logic doesn't interfere with them because the MINT and BURN topic is different for them

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -123,33 +127,64 @@ def _decode_collateral_events(
...
+            amount=hex_or_bytes_to_int(context.tx_log.data[32:64]),
             asset=token,
         )
+
+        earned_event = None  # find the mint event of aToken
+        for _log in context.all_logs:
+            if (
+                _log.topics[0] == MINT and (on_behalf_of is None or (
+                    on_behalf_of is not None and
+                    hex_or_bytes_to_address(_log.topics[2]) == on_behalf_of
```
LefterisJP, line 152: since this is not used anywhere else and you don't use a variable for it, can you explain what `topics[2]` is here in a comment?

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -123,33 +127,64 @@ def _decode_collateral_events(
...
+                )) and
+                user == hex_or_bytes_to_address(_log.topics[1]) and (
```
LefterisJP, line 154: same here for topics[1]

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -123,33 +127,64 @@ def _decode_collateral_events(
...
+                    event_subtype=HistoryEventSubType.INTEREST,
+                    asset=(a_token := get_or_create_evm_token(
+                        userdb=self.evm_inquirer.database,
+                        evm_address=_log.address,
+                        chain_id=self.evm_inquirer.chain_id,
+                        evm_inquirer=self.evm_inquirer,
+                    )),
+                    balance=Balance(amount=balance_increase),
+                    location_label=user,
+                    notes=f'Receive {balance_increase} {a_token.symbol} as interest earned from {self.label}',  # noqa: E501
```
LefterisJP, line 174: hmm isn't this only for aave v3? I think the v2 logic was different. Just notice it's in the common.py decoder.

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -171,36 +207,78 @@ def _decode_deposit(
                 ):
                     event.event_subtype = HistoryEventSubType.RECEIVE_WRAPPED
                     resolved_asset = event.asset.resolve_to_asset_with_symbol()
-                    event.notes = f'Receive {event.balance.amount} {resolved_asset.symbol} from {self.label}'  # noqa: E501
+                    event.balance.amount = amount
```
LefterisJP, line 210: Didn't you say the amount should be minus the earned interest? So that we now split the events in 2?

EDIT: Looking at the v3 test changes you made it seems to be as you described. Though I somehow don't see it from the code.

Where is the amount coming from? Is it the deposit amount? Perhaps then rename the variable here (and in the other place I ask the same question) to make it obvious.

rotkehlchen/chain/evm/decoding/aave/common.py
LefterisJP: I also asked this in another part of the PR. If this is only for v3 it should not be in the common logic. It goes against the expectations

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -171,36 +207,78 @@ def _decode_deposit(
...
 
     def _decode_withdrawal(
             self,
             token: 'EvmToken',
-            tx_log: EvmTxReceiptLog,
-            decoded_events: list['EvmEvent'],
-    ) -> tuple[EvmEvent | None, EvmEvent | None]:
+            context: 'DecoderContext',
+    ) -> tuple[EvmEvent | None, EvmEvent | None, EvmEvent | None]:
         """Decode aave v2/v3 withdrawal event. Returns the Return and Withdraw events."""
```
LefterisJP, line 222: you have to adjust the docstring here

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -171,36 +207,78 @@ def _decode_deposit(
...
             asset=token,
         )
+
+        earned_event = None  # find the mint or burn event of aToken
+        for _log in context.all_logs:
+            if (
+                _log.topics[0] in (MINT, BURN) and
+                user == hex_or_bytes_to_address(_log.topics[1]) and ((
+                    to is not None and
+                    hex_or_bytes_to_address(_log.topics[2]) == to
```
LefterisJP, line 243: same comment as above with explaining what topics[2] is

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -221,19 +299,29 @@ def _decode_withdrawal(
...
+                        earned_event.balance.amount > event.balance.amount
+                    ) and (  # minted directly to us or through native gateway
+                        # docs: https://docs.aave.com/developers/periphery-contracts/wethgateway
+                        event.address == ZERO_ADDRESS or
+                        event.address in self.native_gateways
+                    ) and self._token_is_aave_contract(event.asset))
                 ):
+                    event.event_type = HistoryEventType.SPEND
                     event.event_subtype = HistoryEventSubType.RETURN_WRAPPED
+                    event.balance.amount = amount
```
LefterisJP, line 319: Same question about the amount difference here. I am probably missing something.

rotkehlchen/chain/evm/decoding/aave/v3/decoder.py
```diff
@@ -130,3 +145,6 @@ def counterparties() -> tuple[CounterpartyDetails, ...]:
             label=CPT_AAVE_V3.capitalize().replace('-v', ' V'),
             image='aave.svg',
         ),)
+
+    def decoding_by_input_data(self) -> dict[bytes, dict[bytes, Callable]]:
```
LefterisJP, line 149: hmmmm I just noticed something really bad. We don't check for address or list of addresses for the input data rules either. That's bad.

This will hit if any transfer to any contract starts with `b'\xa9\x05\x9c\xbb'`. We need to change this and may as well do it here.

I mean perhaps there is some cases where we need to do input data decoding irrespective of contract. But wealso need one that's limited to specific contracts, like here. Right? Or am I missing something?

And what are you doing here also? Need more of an explanation. What is this input data? Why do you need this rule? What are you trying to accomplish?

If at every transfer of aave you need to check for a mint event shouldn't you just do a post decoding if there is a transfer event of a token that's an aave token? I am not sure what you are trying to do with input data here

rotkehlchen/accounting/constants.py
```diff
@@ -35,6 +35,7 @@
         HistoryEventSubType.LIQUIDATE: EventCategory.LIQUIDATION_REWARD,
         HistoryEventSubType.PAYMENT: EventCategory.RECEIVE_PAYMENT,
         HistoryEventSubType.GRANT: EventCategory.RECEIVE_GRANT,
+        HistoryEventSubType.INTEREST: EventCategory.RECEIVE,
```
LefterisJP, line 38: I would create a new EventCategory with its own icon and stuff for this. I am surprised we don't have an interest payment event category for the frontend.

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -1602,9 +1602,9 @@ def test_remove_from_aave_pool(ethereum_transaction_decoder, ethereum_accounts,
             event_type=HistoryEventType.SPEND,
             event_subtype=HistoryEventSubType.RETURN_WRAPPED,
             asset=Asset('eip155:1/erc20:0xFd2a8fA60Abd58Efe3EeE34dd494cD491dC14900'),
-            balance=Balance(amount=FVal('2619.297635390037009904')),
+            balance=Balance(amount=FVal('2904.458947543711333839')),
```
LefterisJP, line 1605: Why did you change this test?

As I told you the changes you made are touching code they should not be (v2). This is wrong.

You are swapping a3cRV which is the v2 aToken returned here for the aDAI amount which is part of the same transaction.

https://etherscan.io/tx/0xb0a45bc41a83b2bdf2e06b9913a2e4c8b0d7f3080030807a0a06f301287424e9

rotkehlchen/chain/evm/decoding/aave/common.py
OjusWiZard: yeah right this is the deposit amount, I'll rename the variable

rotkehlchen/chain/evm/decoding/aave/v3/decoder.py
OjusWiZard: here we also filter them using the MINT event, that only happens in the aave tokens. So, it is already only running for the transfer of only aave tokens. I did not use post-decoding because there we would need to store a mapping for all the token addresses, and that will occupy a lot of memory.

> And what are you doing here also? Need more of an explanation. What is this input data? Why do you need this rule? What are you trying to accomplish?

here basically check that whenever someone transfers the aToken, do they receive some tokens also. If yes, then we set it's subtype and notes appropriately. I found there is a condition missing here, so I'll add it and also comment the explanation.

rotkehlchen/accounting/constants.py
```diff
@@ -232,6 +233,9 @@
     )}, EventCategory.RECEIVE_GRANT: {DEFAULT: EventCategoryDetails(
         label='receive grant',
         icon='hand-coin-line',
+    )}, EventCategory.INTEREST: {DEFAULT: EventCategoryDetails(
+        label='receive interest',
+        icon='funds-line',
```
LefterisJP, line 238: If this is a new icon you need to do something in the frontend I am forgetting. Can you ask @rotki/frontend ?

rotkehlchen/chain/evm/decoding/aave/common.py
```diff
@@ -97,6 +99,17 @@ def _token_is_aave_contract(self, asset: 'Asset') -> bool:
             asset_identifier=asset.identifier,
         ) == self.counterparty
 
+    def _decode_interest(
```
LefterisJP, line 102: Hmmm this is not the right approach. This shoudl not be in the common decoder. It should be in the aave v3 decoder. Since it's only for aave v3. Isn't there a v3 only common class? If not you need to create one.

rotkehlchen/chain/evm/decoding/aave/v3/decoder.py
LefterisJP: I don't think you understand my comment. This won't work here.

What if I have a safe? And I transfer through it? Or any other kind of smart contract wallet? This decoding by input data rule is wrong here.

you will need to figure something else out

rotkehlchen/chain/evm/decoding/aave/v3/decoder.py
```diff
@@ -130,3 +257,12 @@ def counterparties() -> tuple[CounterpartyDetails, ...]:
             label=CPT_AAVE_V3.capitalize().replace('-v', ' V'),
             image='aave.svg',
         ),)
+
+    def addresses_to_counterparties(self) -> dict[ChecksumEvmAddress, str]:
+        return {token.evm_address: CPT_AAVE_V3 for token in GlobalDBHandler.get_evm_tokens(
```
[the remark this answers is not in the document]
OjusWiZard: right, if we add a new aave-v3 token then we'll need to redecode the events of that token

rotkehlchen/chain/evm/decoding/aave/v3/decoder.py
```diff
@@ -116,6 +123,126 @@ def _decode_incentives(self, context: 'DecoderContext') -> DecodingOutput:
...
+                return_event.balance.amount = withdraw_event.balance.amount
+                return_event.event_type = HistoryEventType.SPEND
+                return_event.event_subtype = HistoryEventSubType.RETURN_WRAPPED
+                return_event.notes = f'Return {withdraw_event.balance.amount} {a_token.symbol} to {self.label}'  # noqa: E501
+
+        if earned_event is not None and corrected_amount is not None and return_event is not None:
+            return_event.balance.amount = corrected_amount
+
+        maybe_reshuffle_events(
+            ordered_events=[supply_event, return_event, withdraw_event, receive_event, earned_event],  # noqa: E501
```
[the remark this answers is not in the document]
OjusWiZard: yes, in the first line of `maybe_reshuffle_events` we do `actual_events = [x for x in ordered_events if x is not None]`
so None value are filtered out there, and the rest are ordered.

LefterisJP (approved):
LGTM!

Issue #8397, opened 2024-08-13
Bugfix: full validator exits for eigenlayer


LefterisJP:
If you fully exit a validator that's validating for eigenlayer then rotki will always see it as exiting.

That is since rotki determines between exiting and exited by seeing the withdrawal event.

With eigenlayer restaking the final withdrawal event won't appear in history as the eigenpod is not supposed to be tracked so withdrawals to it are not seen.

LefterisJP:
So I spend some time thinking on how to solve this and I want to split this into 2. The real solution involves adding a column to the validators DB table, which is why we need two fixes.

### For 1.34.3 (bugfixes)

We go with a very simple fix.

Modify `get_exited_validator_indices` to instead of using the exit events, take the `withdrawable_ts` and the current timestamp and just make an (quite safe but still hacky) assumption that that if there has been `9` days since the `withdrawable_ts` we assume the exit happened.

Just by taking random withdrawals of a validator it seems like we have ~8 days between withdrawals now.

### For 1.35.0 (develop)

We need to undo the above.

And then add an `exited_ts` in the `eth2_validators` DB table. This exited ts should be populated with 2 different ways.

1. With the old logic of `get_exited_validator_indices` where basically we checked if the validator had an exit event.

```python
    def get_exited_validator_indices(self, cursor: 'DBCursor') -> set[int]:
        """Returns the indices of the tracked validators that we know have exited

        Does so by using processed events, so will only return a list as up to date
        as the events we got
        """
        cursor.execute(
            'SELECT validator_index FROM eth_staking_events_info WHERE is_exit_or_blocknumber=1',
        )  # checking against literal 1 is safe since block 1 was not mined during PoS
        return {x[0] for x in cursor}
        ```

If it does set the new `exited_ts`in the DB table.

2. For all validators that don't have (1) and are after `withdrawable_ts` then ask beaconcha.in's API to see their exited_ts. IF they have one set it in the DB table.

LefterisJP:
This is now closed via #8415 and #8547

PR #8415, opened 2024-08-17
Temporary fix for Eigenlayer validators exit/exiting

LefterisJP:
This is the bugfixes part of https://github.com/rotki/rotki/issues/8397.

Essentially we no longer rely on history events processing to detect a validator has exited. That is due to the fact we may track a validator but not its exit event due to not tracking the withdrawal address directly. This happens for example in Eigenlayer native restaking.

This temporary fix is a hack considering that if "enough time" has passed since the exit was started (the time of withdrawable_ts) and now then we consider it exited.

Long term solution will happen in develop/1.35.0 as per the issue description


LefterisJP:
Pushed some test fixes/adjustment. Even had to skip one due to not making sense to adjust and it being too complicated.

They would need to be retouched in develop. And at least 2 of them re-recorded.

LefterisJP:
okay all green. But need to handle the skipped/adustes tests in develop.

PR #8416, opened 2024-08-18
Fix new logic that was affected after merging bugfixes

yabirgb:
THis part of the code didn't show in the conflicts but the logic got affected. We noticed thanks to a test that failed in nightlies


LefterisJP (approved):
LGTM

PR #8418, opened 2024-08-19
Do not create an imported event if fee is zero

LefterisJP:
Importing events with generic import will no longer create a fee event if the fee is zero. Before it was not creating one only if fee was omitted.


LefterisJP:
The failing test is a network hiccup.

Issue #8414, opened 2024-08-16
Detected tokens randomly dissapear in 1.34.2


LefterisJP:
## Problem Definition

For addresses that have recently received new tokens the detected tokens randomly disappear and this is only fixed by redetecting tokens for the address. This started happening in 1.34.2. I suspect it's due to https://github.com/rotki/rotki/pull/8340

The `maybe_detect_new_tokens()` method introduced in that PR may be the culprit.

It does exactly that ... it maybe detects new tokens from events, if there is new events for the address.

But ... ehm ... here: https://github.com/rotki/rotki/blob/6634f7154bf670d36c5a05511e82bc14b3fc5e3c/rotkehlchen/tasks/assets.py#L410C9-L411C42

```python
detected_tokens = defaultdict(list)
```

Initializes the detected tokens as an empty list. Then runs the logic to detect IN token events to the given address.

and then .. here: https://github.com/rotki/rotki/blob/6634f7154bf670d36c5a05511e82bc14b3fc5e3c/rotkehlchen/tasks/assets.py#L445

```python
    with database.conn.write_ctx() as write_cursor:
        for (chain, location_label), tokens in detected_tokens.items():
            database.save_tokens_for_address(
                write_cursor=write_cursor,
                address=string_to_evm_address(location_label),
                blockchain=chain,
                tokens=tokens,
            )
```

saves the detected tokens per address

So it saves only whatever tokens may have come in via new IN events for an address since last time the function ran. Replacing whatever previous detected tokens were before.

Which would explain the bug I am seeing exactly.

## Task

Fix it.

Things I can imagine need doing is:

1. Make it only append to the already detected tokens. Not overwrite them. (yes in the future we can be smarter and check both in/out events and also remove tokens but that's out of scope)
2. In the migration, schedule a full redetect of tokens for all evm chains? This will take A LONG time. For me ~30 mins. So I would not want this to be done during the loading time of the migration. So not sure how to handle this. Another solution would be to tell users to manually redetect but asking user action is never a good idea as 90% of them won't see it.

yabirgb:
Done at https://github.com/rotki/rotki/pull/8419

PR #8419, opened 2024-08-19
Fix token detection

yabirgb:
ensures that detecting tokens in the background using history events doesn't overwrite the already existing detected tokens

Related to [#8414](https://github.com/rotki/rotki/issues/8414). Regarding triggering the auto detection I believe we don't have logic to trigger it and having it druing the login is too much due to the slowness + it depends on the number of addresses. I would let it up to the user to detect tokens if they miss any


docs/changelog.rst
```diff
@@ -6,6 +6,7 @@ Changelog
 * :bug:`8350` Users will no longer be able to add duplicate names for an address for all evm chains to the address book.
 * :bug:`-` Eigenlayer native restaking exited balances residing in eigenpod will no longer be double counted.
 * :bug:`8397` Active/exited validators will now be propely displayed and filtered for validators that are tracked but the withdrawal address is not. This applies to protocols such as eigenlayer.
+* :bug:`8414` Users won't see tokens disappear from the list of tracked tokens of an address.
```
LefterisJP, line 9: Little bit more specific

```suggestion
* :bug:`8414` Detected tokens will no longer randomly disappear from a tracked address if new events are processed.
```

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -379,13 +390,22 @@ def test_update_snapshot_balances(rotkehlchen_instance: 'Rotkehlchen'):
...
             assert save_tokens_mock.call_args_list[0].kwargs['blockchain'] == SupportedBlockchain.ETHEREUM  # noqa: E501
-            assert save_tokens_mock.call_args_list[0].kwargs['tokens'] == [A_DAI]
+            assert set(save_tokens_mock.call_args_list[0].kwargs['tokens']) == {A_COMP, A_LUSD, A_DAI}  # noqa: E501
             assert save_tokens_mock.call_args_list[1].kwargs['address'] == accounts[2]
             assert save_tokens_mock.call_args_list[1].kwargs['blockchain'] == SupportedBlockchain.OPTIMISM  # noqa: E501
             assert save_tokens_mock.call_args_list[1].kwargs['tokens'] == [A_USDC]
     except gevent.Timeout as e:
         raise AssertionError(f'Update snapshot balances was not completed within {timeout} seconds') from e  # noqa: E501
 
+    # verify that newly saved tokens respects the list of already existing tokens
```
LefterisJP, line 400: this test failed before your fix right?

yabirgb: Yes, test failed before the change. The only tokens kept where the ones detected form the events and any other token was discarded. After the fix we keep the tokens already there

LefterisJP (approved):
LGTM. Will adjust nitpick in another PR.

PR #8421, opened 2024-08-19
Fix (Nightlies): Cleanups for windows

OjusWiZard:
## Checklist

- [x] Close DB connections before tests cleanup to avoid `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process`

### DIFF
--- a/rotkehlchen/db/dbhandler.py
+++ b/rotkehlchen/db/dbhandler.py
@@ -502,6 +502,7 @@ class DBHandler:
             # switch to WAL mode: https://www.sqlite.org/wal.html
             conn.execute('PRAGMA journal_mode=WAL;')
         except sqlcipher.DatabaseError as e:  # pylint: disable=no-member
+            conn.close()
             raise AuthenticationError(
                 'Wrong password or invalid/corrupt database for user',
             ) from e
--- a/rotkehlchen/tests/db/test_db_upgrades.py
+++ b/rotkehlchen/tests/db/test_db_upgrades.py
@@ -13,6 +13,7 @@ from rotkehlchen.constants.misc import (
     AIRDROPSDIR_NAME,
     APPDIR_NAME,
     DEFAULT_SQL_VM_INSTRUCTIONS_CB,
+    USERDB_NAME,
 )
 from rotkehlchen.data_handler import DataHandler
 from rotkehlchen.db.checks import sanity_check_impl
@@ -178,6 +179,7 @@ def test_upgrade_db_26_to_27(user_data_dir):  # pylint: disable=unused-argument
     # Finally also make sure that we have updated to the target version
     with db.conn.read_ctx() as cursor:
         assert db.get_setting(cursor, 'version') == 27
+    db.logout()
 
 
 @pytest.mark.parametrize('use_clean_caching_directory', [True])
@@ -230,6 +232,7 @@ def test_upgrade_db_27_to_28(user_data_dir):  # pylint: disable=unused-argument
     # Finally also make sure that we have updated to the target version
     with db.conn.read_ctx() as cursor:
         assert db.get_setting(cursor, 'version') == 28
+    db.logout()
 
 
 @pytest.mark.parametrize('use_clean_caching_directory', [True])
@@ -345,6 +348,7 @@ def test_upgrade_db_28_to_29(user_data_dir):  # pylint: disable=unused-argument
     # Finally also make sure that we have updated to the target version
     with db.conn.read_ctx() as cursor:
         assert db.get_setting(cursor, 'version') == 29
+    db.logout()
 
 
 @pytest.mark.parametrize('use_clean_caching_directory', [True])
@@ -368,6 +372,7 @@ def test_upgrade_db_29_to_30(user_data_dir):  # pylint: disable=unused-argument
     # Check that existing balances are not considered as liabilities after migration
     cursor.execute('SELECT category FROM manually_tracked_balances;')
     assert cursor.fetchone() == ('A',)
+    db.logout()
 
 
 @pytest.mark.parametrize('use_clean_caching_directory', [True])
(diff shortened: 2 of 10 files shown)


rotkehlchen/db/dbhandler.py
```diff
@@ -219,9 +220,13 @@ def __init__(
...
-        self._connect()
-        self._check_unfinished_upgrades(resume_from_backup=resume_from_backup)
-        self._run_actions_after_first_connection()
+        try:
+            self._connect()
+            self._check_unfinished_upgrades(resume_from_backup=resume_from_backup)
+            self._run_actions_after_first_connection()
+        except (SystemPermissionError, AuthenticationError, RotkehlchenPermissionError, DBUpgradeError, DBSchemaError):  # noqa: E501
+            self.logout()
+            raise
```
yabirgb, line 229: I guess this errors existed before right? Maybe add a comment for why you need the logout since we might forget in the future.

Question: in windows if there was an error here what happened? Did the app work?

rotkehlchen/db/dbhandler.py
```diff
@@ -502,6 +507,7 @@ def _connect(self, conn_attribute: Literal['conn', 'conn_transient'] = 'conn') -
             # switch to WAL mode: https://www.sqlite.org/wal.html
             conn.execute('PRAGMA journal_mode=WAL;')
         except sqlcipher.DatabaseError as e:  # pylint: disable=no-member
+            conn.close()
```
yabirgb, line 510: can close fail?

rotkehlchen/tests/unit/test_blockchain_balances.py
```diff
@@ -58,7 +58,8 @@ def fixture_blockchain_balances(use_db, data_dir, username, sql_vm_instructions_
     a.optimism[address2].assets[OPTIMISM_OP_TOKEN] = Balance(amount=1, usd_value=1)
     a.optimism[address2].assets[A_ETH] = Balance(amount=1, usd_value=1)
 
-    return a, address1, address2, all_btc_addresses, xpub_data
+    yield a, address1, address2, all_btc_addresses, xpub_data
+    db.logout()
```
yabirgb, line 62: this db doesn't exist always right? Only if use_db is True. You should check here I believe for the same condition

rotkehlchen/db/dbhandler.py
OjusWiZard: no i guess, in other places also we use it safely

rotkehlchen/db/dbhandler.py
OjusWiZard: this is called when we create a new user or log in an existing user. All these error are handled in `user_login` and the response is sent back accordingly

rotkehlchen/db/dbhandler.py
yabirgb: It seems that never fails https://github.com/python/cpython/blob/77133f570dcad599e5b1199c39e999bfac959ae2/Modules/_sqlite/connection.c#L469

rotkehlchen/db/dbhandler.py
LefterisJP: So if it's all happening there why do you need this here?

Can't you just add it in `reset_after_failed_account_creation_or_login()`:

```python
        try:
            self.rotkehlchen.unlock_user(
                user=name,
                password=password,
                create_new=False,
                sync_approval=sync_approval,
                premium_credentials=None,
                resume_from_backup=resume_from_backup,
            )
        # We are not catching PremiumAuthenticationError here since it should not bubble
        # up until here. Even with non valid keys in the DB login should work fine
        except AuthenticationError as e:
            self.rotkehlchen.reset_after_failed_account_creation_or_login()
            return wrap_in_fail_result(
                message=str(e),
                status_code=HTTPStatus.UNAUTHORIZED,
            )
        except RotkehlchenPermissionError as e:
            self.rotkehlchen.reset_after_failed_account_creation_or_login()
            return {
                'result': e.message_payload,
                'message': e.error_message,
                'status_code': HTTPStatus.MULTIPLE_CHOICES,
            }
        except (DBUpgradeError, SystemPermissionError, DBSchemaError) as e:
            self.rotkehlchen.reset_after_failed_account_creation_or_login()
            return wrap_in_fail_result(
                message=str(e),
                status_code=HTTPStatus.CONFLICT,
            )
        except sqlcipher.OperationalError as e:  # pylint: disable=no-member
            self.rotkehlchen.reset_after_failed_account_creation_or_login()
            return wrap_in_fail_result(
                message=f'Unexpected database error: {e!s}',
                status_code=HTTP_STATUS_INTERNAL_DB_ERROR,  # type: ignore  # Is a custom status code, not a member of HTTPStatus
            )
```

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1327,6 +1337,7 @@ def test_upgrade_db_35_to_36(user_data_dir):  # pylint: disable=unused-argument
     res = cursor.execute('SELECT * FROM history_events WHERE identifier=?', (234,)).fetchone()
     assert res is not None
     assert res[9] == 'Edited event'  # event's notes
+    db.logout()
```
LefterisJP, line 1340: instead of doing this in every single test does it not make sense to do this as part of the DB fixture? Or wait ... it's actually there: https://github.com/rotki/rotki/blob/ebf85137354eec299e6bc541efa91ccaa1e943ad/rotkehlchen/tests/fixtures/db.py#L168

So this is not needed at the end of each test, as long as it's the DB intiialized by the fixture

OjusWiZard: yeah right, we only need to add it in the tests that create a db as part of the test

rotkehlchen/db/dbhandler.py
OjusWiZard: so the initial reason was because of the tests. In many unit tests we directly initialize the db without the rest API, so there the fallback cleanup was not being done. But also, I think cleanup related to the db happening in the same class would better because these similar logic are already present here and this would later make it easy when the rest API gets served from a seperate process

LefterisJP (approved):
LGTM

PR #8440, opened 2024-08-22
Improve logic for calling nodes

yabirgb:
- Avoid adding and removing elements from a list to calculated the default order
- return earlier when we get a result from a node. Avoids reading the except clauses by the interpreter
- simplify conditions when deciding if we call the node or not

Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/evm/node_inquirer.py
+++ b/rotkehlchen/chain/evm/node_inquirer.py
@@ -5,7 +5,7 @@ from abc import ABC, abstractmethod
 from collections.abc import Callable, Sequence
 from contextlib import suppress
 from itertools import zip_longest
-from typing import TYPE_CHECKING, Any, Literal
+from typing import TYPE_CHECKING, Any, Literal, TypeVar
 from urllib.parse import urlparse
 
 import requests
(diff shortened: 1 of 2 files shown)


rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -521,20 +546,19 @@ def _query(self, method: Callable, call_order: Sequence[WeightedNode], **kwargs:
...
+                    continue
+
+                web3 = web3node.web3_instance
+            else:
+                web3 = None
 
             try:
-                web3 = web3node.web3_instance if web3node is not None else None
-                result = method(web3, **kwargs)
+                return method(web3, **kwargs)
```
yabirgb, line 561: The intention here is to avoid checking multiple times `if web3node is not None` and also have the return stamenet earlier

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -287,18 +309,21 @@ def default_call_order(self, skip_etherscan: bool = False) -> list[WeightedNode]
...
             selection = [wnode for wnode in open_nodes if wnode.node_info.name != self.etherscan_node_name and wnode.node_info.owned is False]  # noqa: E501
         else:
             selection = [wnode for wnode in open_nodes if wnode.node_info.owned is False]
 
-        ordered_list = []
-        while len(selection) != 0:
-            weights = [float(entry.weight) for entry in selection]
-            node = random.choices(selection, weights, k=1)
-            ordered_list.append(node[0])
-            selection.remove(node[0])
```
yabirgb, line 301: here I wanted to avoid this append + remove. Instead use a list cvomprehension and avoid the while loop

yabirgb:
Closing it as it needs for tests to support the change

PR #8450, opened 2024-08-27
Feat (DB Writer): Use protobuf instead of pickle

OjusWiZard:
## Checklist

- [x] Migrate to protobuf for serialization/deserialization between the backend process and the DB writer process
- [x] Check tests and fix the failed ones


rotkehlchen/db/drivers/db_messages.proto
```diff
@@ -0,0 +1,56 @@
...
+  repeated TypedData array = 1;
+}
+
+message TypedData {
+  string type = 1;
+  oneof data {
+    bytes bytes = 2;
+    TypedArray array = 3;
+  }
+}
```
yabirgb, line 42: add a comment saying that typedArray is needed because `oneof doesn't allow repeated as label for their members`

yabirgb: https://stackoverflow.com/a/62239063

rotkehlchen/db/drivers/server.py
```diff
@@ -117,15 +104,22 @@ def __init__(self, port: int) -> None:
...
+        connect_func = sqlite3.connect if call_data.connection_type == db_messages_pb2.DBConnectionType.GLOBAL else sqlcipher.connect  # type: ignore[attr-defined]  # pylint: disable=no-member  # noqa: E501
+        self.db_connections[call_data.db_path] = connect_func(*args, **kwargs)
 
     def _handle_query(
             self,
             db_connection: UnderlyingConnection,
-            call_data: ZMQCallData,
-    ) -> ZMQReturnData:
+            call_data: db_messages_pb2.ZMQCallData,  # type: ignore[name-defined]  # pylint: disable=no-member
+            args: tuple[Any],
```
yabirgb, line 120: I would use the type of what is allowed for args and kwargs... not anything is valid

rotkehlchen/db/drivers/server.py
```diff
@@ -144,57 +138,64 @@ def _handle_query(
...
                     continue
 
-                result = self._handle_query(db_connection, call_data)
+                result = self._handle_query(db_connection, ZMQCallData, args, kwargs)
 
             except BaseException as e:
                 if self.shutting_down:
                     break  # The server is shutting down, don't try to send an error reply
-                self.zmq_connection.send_pyobj(ZMQReturnData(result=None, error=e).serialize())
+                DBWriterError.Clear()
```
yabirgb, line 180: Same question... Is there any advantage of creating them and then clearing?

rotkehlchen/db/drivers/server.py
```diff
@@ -144,57 +138,64 @@ def _handle_query(
...
+                if hasattr(e, '__module__'):
+                    DBWriterError.module = e.__module__
```
yabirgb, line 182: can't the class name be the module also? so module + classname

rotkehlchen/db/drivers/server.py
```diff
@@ -29,74 +27,63 @@ class DBConnectionType(Enum):
...
+        typed_data.bytes = data
+    elif isinstance(data, str):
+        typed_data.type = 'string'
+        typed_data.bytes = data.encode()
+    elif isinstance(data, bool):
+        typed_data.type = 'bool'
+        typed_data.bytes = data.to_bytes()
+    elif isinstance(data, int):
+        typed_data.type = 'int'
+        typed_data.bytes = data.to_bytes(length=32, byteorder='big', signed=True)
```
yabirgb, line 57: why are you setting the lenght as 32? Why not 64 or any other value? What drove this decision?

rotkehlchen/db/drivers/server.py
```diff
@@ -144,57 +138,64 @@ def _handle_query(
...
-                    self.db_cursors[call_data.db_path][call_data.cursor_name or 'default'] = db_connection.cursor()  # noqa: E501
-                    self.zmq_connection.send_pyobj(None)
+                    db_connection = self.db_connections[ZMQCallData.db_path]
+                except KeyError as e:
+                    raise KeyError(f'No DB connection for {DBConnectionType(ZMQCallData.connection_type)} found') from e  # noqa: E501
+
+                if ZMQCallData.method == DBMethod.OPEN_CURSOR:
+                    if self.db_cursors[ZMQCallData.db_path].get(ZMQCallData.cursor_name or 'default') is not None:  # noqa: E501
+                        raise KeyError(f'Cursor for {ZMQCallData.db_path} already exists')
+                    self.db_cursors[ZMQCallData.db_path][ZMQCallData.cursor_name or 'default'] = db_connection.cursor()  # noqa: E501
```
yabirgb, line 167: about the error message you should include also the name of the cursor since it's the name what is being used

rotkehlchen/db/drivers/db_messages.proto
```diff
@@ -0,0 +1,56 @@
...
+message DBWriteResult {
+  int32 rowcount = 1;
+  uint64 lastrowid = 2;
+}
+
+message DBWriterError {
+  string message = 1;
+  optional string module = 2;
+  string class_name = 3;
+}
```
yabirgb, line 30: Not sure about the errors... As you are writing them you are passing python errors around. It is okey if it is python processes but in the future we will have other type of errors... Don't you think it would be nice to have something that works for all the languages? In the client you can raise a `DBWrittingError` that includes the error type and the message... I guess the complexity and why you didn't do it is because you will need to change the handling of errors around the app. What kind of errors do we have? OperationalError, IntegrityError and then something unexpected right?

rotkehlchen/db/drivers/client.py
```diff
@@ -192,31 +195,43 @@ def __init__(
...
-            kwargs=kwargs,
-        ))
-        response: SerializedZMQReturnData | None = self.zmq_connection.recv_pyobj()
-        if response is None:
+    def _send(self, method: db_messages_pb2.DBMethod, cursor_name: str | None = None, *args: Any, **kwargs: Any) -> None:  # type: ignore[name-defined]  # pylint: disable=no-member  # noqa: E501
+        ZMQCallData.Clear()
+        ZMQCallData.connection_type = self.connection_type.value
+        ZMQCallData.db_path = self.db_path
+        if cursor_name is not None:
+            ZMQCallData.cursor_name = cursor_name
```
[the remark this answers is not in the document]
OjusWiZard: so we can't assign None to an optional protobuf field, that's why we have it like this

rotkehlchen/db/drivers/server.py
```diff
@@ -3,100 +3,133 @@
...
 
-    @classmethod
-    def deserialize(cls, data: SerializedZMQReturnData) -> 'ZMQReturnData':
-        return cls(
-            result=data['result'],
-            error=data['error'],
-        )
+def to_typed_data(data: SerializableTypes) -> db_messages_pb2.TypedData:  # type: ignore[name-defined]  # pylint: disable=no-member
+    """Converts a python object to a protobuf TypedData object
+    Supported types: list, tuple, bytes, str, bool, int, float, and None"""
```
yabirgb, line 85: since you have the type you don't need this comment

rotkehlchen/db/drivers/server.py
```diff
@@ -144,57 +184,64 @@ def _handle_query(
...
+                    if self.db_cursors[call_data.db_path].get(cursor_name) is not None:
+                        raise KeyError(f'Cursor "{cursor_name}" for {call_data.db_path} already exists')  # noqa: E501
+                    self.db_cursors[call_data.db_path][cursor_name] = db_connection.cursor()
+                    self.zmq_connection.send(b'')
                     continue
-                if call_data.method.value == DbMethod.CLOSE_CURSOR.value:
-                    del self.db_cursors[call_data.db_path][call_data.cursor_name or 'default']
-                    self.zmq_connection.send_pyobj(None)
+                if call_data.method == DBMethod.CLOSE_CURSOR:
+                    del self.db_cursors[call_data.db_path][cursor_name]
```
yabirgb, line 219: this can raise keyError if the cursor is not created. I would use `.pop` with a default here and avoid that error

rotkehlchen/db/drivers/client.py
```diff
@@ -192,31 +192,43 @@ def __init__(
...
-        ))
-        response: SerializedZMQReturnData | None = self.zmq_connection.recv_pyobj()
-        if response is None:
+    def _send(self, method: db_messages_pb2.DBMethod, cursor_name: str | None = None, *args: Any, **kwargs: Any) -> None:  # type: ignore[name-defined]  # pylint: disable=no-member  # noqa: E501
+        call_data = db_messages_pb2.ZMQCallData()  # type: ignore[attr-defined]  # pylint: disable=no-member
+        call_data.connection_type = self.connection_type.value
+        call_data.db_path = self.db_path
+        if cursor_name is not None:
+            call_data.cursor_name = cursor_name
+        call_data.method = method
```
yabirgb, line 203:
```suggestion
            call_data.cursor_name = cursor_name

        call_data.method = method
```

rotkehlchen/db/drivers/client.py
```diff
@@ -192,31 +192,43 @@ def __init__(
...
+        call_data.args.extend([to_typed_data(data) for data in args])
+        for key, value in kwargs.items():
+            typed_data = to_typed_data(value)
+            call_data.kwargs[key].type = typed_data.type
+            if typed_data.WhichOneof('data') == 'bytes':
+                call_data.kwargs[key].bytes = typed_data.bytes
+            if typed_data.WhichOneof('data') == 'array':
+                call_data.kwargs[key].array.CopyFrom(typed_data.array)
```
yabirgb, line 211: can it be both of them? shouldn't this be an elif?

rotkehlchen/db/drivers/client.py
```diff
@@ -192,31 +192,43 @@ def __init__(
...
-        if result.error is not None:
-            raise result.error
-        if result.result is not None:
-            self.lastrowid = result.result['lastrowid']
-            self.rowcount = result.result['rowcount']
+        return_data = db_messages_pb2.ZMQReturnData()  # type: ignore[attr-defined]  # pylint: disable=no-member
+        return_data.ParseFromString(response_serialized)
+        if return_data.HasField('error'):
+            if return_data.error.name == DBError.UnknownError:
+                raise ValueError(f'Got unknown DB error with message {return_data.error.message}')
```
yabirgb, line 220: Dont't raise ValueError... Isn't there something better? I would create a custom error for this

rotkehlchen/db/drivers/db_messages.proto
```diff
@@ -0,0 +1,108 @@
...
+
+enum DBError {
+  ArithmeticError = 0;
+  AssertionError = 1;
+  AttributeError = 2;
+  Exception = 3;
+  EOFError = 4;
+  FloatingPointError = 5;
+  GeneratorExit = 6;
+  ImportError = 7;
```
yabirgb, line 29: I believe we shouldn't map 1:1 the python errors. Also lefteris https://discord.com/channels/657906918408585217/1238139584890535966/1279014760914092085 mentioned here. I would have something like:

- SystemError
- sqliteError
- ValueError

and maybe 1 or 2 more if you need to differenciate them. You can also add attributes if they are needed for debugging/processing like the name of the error

PR #8458, opened 2024-08-28
Improve CSV import error reporting

### LINKED ISSUE #7349: CSV Import error reporting improvements
## Abstract

A rotki user in discord gave some feedback on CSV imports which are valid and would greatly improve user experience when faced with problems importing CSVs.

1. Add a notification of the number of entries that were imported via CSV. So if the CSV had 120 rows and all were imported we should say 120/120 entries imported. Meaning complete success. If 2 entries failed to import for whatever reason show 118/120 entries were imported.
2. For each failure to import an entry show what failed. Perhaps like: "Failed to import row 424 due to inability to convert "x434" to a number.

## Task

1. **Backend**: For all CSV imports keep a tally of all processed rows so we can show the x/y entries processed message. Also whenever we failed to decode an entry a log entry is added and we skip it. Now instead of just the log entry, also (optionally turn into a more user friendly message) add this into some message queue on the importer, that is flushed and send back via the API in case of errors to the frontend. With details per entry not processed.
2. **Frontend**: Proccess the new information once import is successfull and show it to the user in a nice and readable way

nicholasyoder:
Closes 7349

### DIFF
--- a/rotkehlchen/data_import/importers/binance.py
+++ b/rotkehlchen/data_import/importers/binance.py
@@ -9,7 +9,7 @@ from rotkehlchen.accounting.structures.balance import Balance
 from rotkehlchen.assets.converters import asset_from_binance
 from rotkehlchen.constants import ZERO
 from rotkehlchen.constants.assets import A_USD
-from rotkehlchen.data_import.utils import BaseExchangeImporter, hash_csv_row
+from rotkehlchen.data_import.utils import BaseExchangeImporter, UnsupportedCSVEntry, hash_csv_row
 from rotkehlchen.db.drivers.gevent import DBCursor
 from rotkehlchen.errors.asset import UnknownAsset, UnsupportedAsset
 from rotkehlchen.errors.misc import InputError
@@ -38,6 +38,7 @@ from rotkehlchen.utils.misc import ts_sec_to_ms
 
 if TYPE_CHECKING:
     from rotkehlchen.assets.asset import AssetWithOracles
+    from rotkehlchen.db.dbhandler import DBHandler
 
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
@@ -45,6 +46,12 @@ log = RotkehlchenLogsAdapter(logger)
 BinanceCsvRow = dict[str, Any]
 BINANCE_TRADE_OPERATIONS = {'Buy', 'Sell', 'Fee'}
 EVENT_IDENTIFIER_PREFIX = 'BNC_'
+INDEX = '_index'
+
+
+def hash_binance_csv_row(csv_row: BinanceCsvRow) -> str:
+    """Hash the CSV row excluding the INDEX"""
+    return hash_csv_row({k: v for k, v in csv_row.items() if k != INDEX})
 
 
 class BinanceEntry(abc.ABC):  # noqa: B024
@@ -126,7 +133,7 @@ class BinanceTransferEntry(BinanceMultipleEntry):
         row = data[0]
         return [
             HistoryEvent(
-                event_identifier=f'{EVENT_IDENTIFIER_PREFIX}{hash_csv_row(row)}',
+                event_identifier=f'{EVENT_IDENTIFIER_PREFIX}{hash_binance_csv_row(row)}',
                 sequence_index=0,
                 timestamp=ts_sec_to_ms(timestamp),
                 location=Location.BINANCE,
@@ -240,7 +247,12 @@ class BinanceTradeEntry(BinanceMultipleEntry):
                     )
                 except NoPriceForGivenTimestamp:
                     # If we can't find price we can't group, so we quit the method
-                    log.warning(f'Couldn\'t find price of {row["Coin"]} on {timestamp}')
+                    importer.send_message(
+                        row_index=row[INDEX],
+                        csv_row=row,
+                        msg=f'Couldn\'t find price of {row["Coin"]} on {timestamp}',
+                        is_error=True,
+                    )
                     return []
 
         # Group rows depending on whether they are fee or not and then sort them by amount
@@ -294,11 +306,13 @@ class BinanceTradeEntry(BinanceMultipleEntry):
                 to_amount is None or to_amount == ZERO or
                 from_amount is None or from_amount == ZERO
             ):
-                log.warning(
-                    f'Skipped binance rows {data} because '
-                    f"it didn't have enough data",
-                )
-                importer.db.msg_aggregator.add_warning("Skipped some rows because couldn't find amounts or it was zero")  # noqa: E501
+                for row in trade_rows:
+                    importer.send_message(
+                        row_index=row[INDEX],
+                        csv_row=row,
+                        msg='Trade associated with row is missing required assets or amounts.',
+                        is_error=True,
+                    )
                 continue
 
             rate = to_amount / from_amount
@@ -406,7 +420,7 @@ class BinanceDistributionEntry(BinanceSingleEntry):
         """
         importer.add_history_events(write_cursor, history_events=[
             HistoryEvent(
-                event_identifier=f'{EVENT_IDENTIFIER_PREFIX}{hash_csv_row(data)}',
+                event_identifier=f'{EVENT_IDENTIFIER_PREFIX}{hash_binance_csv_row(data)}',
                 sequence_index=0,
                 timestamp=ts_sec_to_ms(timestamp),
                 location=Location.BINANCE,
@@ -440,7 +454,7 @@ class BinanceStakingRewardsEntry(BinanceSingleEntry):
             data: BinanceCsvRow,
     ) -> None:
         event = HistoryEvent(
-            event_identifier=f'{EVENT_IDENTIFIER_PREFIX}{hash_csv_row(data)}',
+            event_identifier=f'{EVENT_IDENTIFIER_PREFIX}{hash_binance_csv_row(data)}',
             sequence_index=0,
             timestamp=ts_sec_to_ms(timestamp),
             location=Location.BINANCE,
@@ -482,7 +496,7 @@ class BinanceEarnProgram(BinanceSingleEntry):
         """
         asset = data['Coin']
         amount = abs(data['Change'])
-        event_identifier = f'{EVENT_IDENTIFIER_PREFIX}{hash_csv_row(data)}'
+        event_identifier = f'{EVENT_IDENTIFIER_PREFIX}{hash_binance_csv_row(data)}'
         timestamp_ms = ts_sec_to_ms(timestamp)
         staking_event = None
         if data['Operation'] in {
@@ -520,11 +534,11 @@ class BinanceEarnProgram(BinanceSingleEntry):
                 location_label='CSV import',
                 notes=f'Deposit in {data["Operation"]}',
             )
-        elif data['Operation'] == (
+        elif data['Operation'] in {
             'Simple Earn Locked Redemption',
             'Simple Earn Flexible Redemption',
             'Staking Redemption',
-        ):
+        }:
             staking_event = HistoryEvent(
                 event_identifier=event_identifier,
                 sequence_index=0,
@@ -538,8 +552,7 @@ class BinanceEarnProgram(BinanceSingleEntry):
                 notes=f'Unstake {asset} in {data["Operation"]}',
             )
         if staking_event is None:
-            log.error(f'Could not process Binance CSV entry {data}')
-            return
+            raise UnsupportedCSVEntry(f'Unknown staking event operation: {data["Operation"]}.')
         importer.add_history_events(write_cursor=write_cursor, history_events=[staking_event])
 
 
@@ -575,7 +588,7 @@ class BinanceUSDMProgram(BinanceSingleEntry):
             action = 'profit' if is_profit else 'loss'
             notes = f'{amount} {data["Coin"].symbol} realized {action} on binance USD-MFutures'
         return HistoryEvent(
-            event_identifier=f'{EVENT_IDENTIFIER_PREFIX}{hash_csv_row(data)}',
+            event_identifier=f'{EVENT_IDENTIFIER_PREFIX}{hash_binance_csv_row(data)}',
             sequence_index=0,
             timestamp=ts_sec_to_ms(timestamp),
             location=Location.BINANCE,
@@ -628,7 +641,7 @@ class BinancePOSEntry(BinanceSingleEntry):
             event_subtype = HistoryEventSubType.NONE
 
         event = HistoryEvent(
-            event_identifier=f'{EVENT_IDENTIFIER_PREFIX}{hash_csv_row(data)}',
+            event_identifier=f'{EVENT_IDENTIFIER_PREFIX}{hash_binance_csv_row(data)}',
             sequence_index=0,
             timestamp=ts_sec_to_ms(timestamp),
             location=Location.BINANCE,
@@ -656,41 +669,58 @@ MULTIPLE_BINANCE_ENTRIES: list[BinanceMultipleEntry] = [
 ]
 
 
-def _group_binance_rows(
-        rows: list[BinanceCsvRow],
-        timestamp_format: str = '%Y-%m-%d %H:%M:%S',
-) -> tuple[int, dict[Timestamp, list[BinanceCsvRow]]]:
-    """Groups Binance rows by timestamp and deletes unused columns"""
-    multirows: dict[Timestamp, list[BinanceCsvRow]] = defaultdict(list)
-    skipped_count = 0
-    for csv_row in rows:
-        try:
-            timestamp = deserialize_timestamp_from_date(
-                date=csv_row['UTC_Time'],
-                formatstr=timestamp_format,
-                location='binance',
-            )
-            csv_row['Coin'] = asset_from_binance(csv_row['Coin'])
-            csv_row['Change'] = deserialize_asset_amount(csv_row['Change'])
-            multirows[timestamp].append(csv_row)
-        except (DeserializationError, UnknownAsset, UnsupportedAsset) as e:
-            log.warning(f'Skipped binance csv row {csv_row} because of {e!s}')
-            skipped_count += 1
-        except KeyError as e:
-            log.error(f'Malformed binance csv columns! Broke on row {csv_row}. {e!s}')
-            return len(rows), {}
-
-    return skipped_count, multirows
-
-
 class BinanceImporter(BaseExchangeImporter):
+    """Binance CSV importer"""
+
+    def __init__(self, db: 'DBHandler') -> None:
+        super().__init__(db=db, name='Binance')
+
+    def _group_binance_rows(
+            self,
+            rows: list[BinanceCsvRow],
+            timestamp_format: str = '%Y-%m-%d %H:%M:%S',
+    ) -> tuple[int, dict[Timestamp, list[BinanceCsvRow]]]:
+        """Groups Binance rows by timestamp and deletes unused columns"""
+        multirows: dict[Timestamp, list[BinanceCsvRow]] = defaultdict(list)
+        skipped_count = 0
+        for index, csv_row in enumerate(rows, start=1):
+            try:
+                timestamp = deserialize_timestamp_from_date(
+                    date=csv_row['UTC_Time'],
+                    formatstr=timestamp_format,
+                    location='binance',
+                )
+                csv_row['Coin'] = asset_from_binance(csv_row['Coin'])
+                csv_row['Change'] = deserialize_asset_amount(csv_row['Change'])
+                csv_row[INDEX] = index
+                multirows[timestamp].append(csv_row)
+            except UnknownAsset as e:
+                self.send_message(
+                    row_index=index,
+                    csv_row=csv_row,
+                    msg=f'Unknown asset "{e.identifier}" provided.',
+                    is_error=True,
+                )
+                skipped_count += 1
+            except (DeserializationError, UnsupportedAsset) as e:
+                self.send_message(
+                    row_index=index,
+                    csv_row=csv_row,
+                    msg=str(e),
+                    is_error=True,
+                )
+                skipped_count += 1
+            except KeyError as e:
+                raise InputError(f'Malformed binance csv columns! {e!s}') from e
+
+        return skipped_count, multirows
 
     def _process_single_binance_entries(
             self,
             write_cursor: DBCursor,
             timestamp: Timestamp,
             rows: list[BinanceCsvRow],
-    ) -> tuple[dict[BinanceSingleEntry, int], list[BinanceCsvRow]]:
+    ) -> tuple[int, dict[BinanceSingleEntry, int], list[BinanceCsvRow]]:
         """
         Processes binance entries that are represented with a single row in a csv file.
         May raise:
@@ -698,6 +728,7 @@ class BinanceImporter(BaseExchangeImporter):
         """
         processed: dict[BinanceSingleEntry, int] = defaultdict(int)
         ignored: list[BinanceCsvRow] = []
+        skipped_count: int = 0
         for row in rows:
             for single_entry_class in SINGLE_BINANCE_ENTRIES:
                 try:
@@ -714,6 +745,15 @@ class BinanceImporter(BaseExchangeImporter):
                         )
                         processed[single_entry_class] += 1
                         break
+                except UnsupportedCSVEntry as e:
+                    self.send_message(
+                        row_index=row[INDEX],
+                        csv_row=row,
+                        msg=str(e),
+                        is_error=True,
+                    )
+                    skipped_count += 1
+                    break
                 except KeyError as e:
                     raise InputError(f'Could not read key {e} in binance CSV row {row}') from e
                 except DeserializationError as e:
@@ -722,7 +762,7 @@ class BinanceImporter(BaseExchangeImporter):
                     ) from e
             else:  # there was no break in the for loop above
                 ignored.append(row)
-        return processed, ignored
+        return skipped_count, processed, ignored
 
     def _process_multiple_binance_entries(
             self,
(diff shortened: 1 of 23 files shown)


rotkehlchen/data_import/utils.py
```diff
@@ -75,9 +97,45 @@ def flush_all(self, write_cursor: DBCursor) -> None:
...
+    def log_debug(self, row_index: int, csv_row: dict, msg: str) -> None:
+        log.debug(f'Skipping row {row_index} of {self.name} CSV file. {msg}')
+        log.debug(csv_row)
+        self.append_msg(row_index, msg, False)
+
+    def log_warning(self, row_index: int, csv_row: dict, msg: str) -> None:
+        log.warning(f'Failed to import row {row_index} of {self.name} CSV file. {msg}')
+        log.debug(csv_row)
+        self.append_msg(row_index, msg, True)
+        self.db.msg_aggregator.warnings.appendleft(msg)
```
OjusWiZard, line 130: I guess failure should be logged as error, and skipping should be logged as warning

rotkehlchen/data_import/utils.py
```diff
@@ -75,9 +97,45 @@ def flush_all(self, write_cursor: DBCursor) -> None:
...
+        self.import_msgs.append({
+            'row': row_index,
+            'msg': msg,
+            'warning': warning,
+        })
+
+    def log_debug(self, row_index: int, csv_row: dict, msg: str) -> None:
+        log.debug(f'Skipping row {row_index} of {self.name} CSV file. {msg}')
+        log.debug(csv_row)
+        self.append_msg(row_index, msg, False)
```
OjusWiZard, line 124: similarly in other places where we call them, because reader doesn't know what False is here
```suggestion
        self.append_msg(row_index=row_index, msg=msg, warning=False)
```

rotkehlchen/data_import/importers/binance.py
```diff
@@ -294,11 +304,12 @@ def process_trades(
...
-                log.warning(
-                    f'Skipped binance rows {data} because '
-                    f"it didn't have enough data",
-                )
-                importer.db.msg_aggregator.add_warning("Skipped some rows because couldn't find amounts or it was zero")  # noqa: E501
+                for row in trade_rows:
+                    importer.log_warning(
+                        row_index=row[INDEX],
+                        csv_row=row,
+                        msg='Trade associated with row is missing required data.',
```
OjusWiZard, line 311: previous message told that what is missing, now it only says 'data'. Is this info sent anywhere else?

rotkehlchen/data_import/importers/binance.py
```diff
@@ -771,42 +792,39 @@ def _process_binance_rows(
...
-            )
+
+        return skipped_count
 
     def _import_csv(self, write_cursor: DBCursor, filepath: Path, **kwargs: Any) -> None:
         """
         Group and process binance CSV entries. May raise:
         - InputError
         """
+        self.name = 'Binance'
```
OjusWiZard, line 824: this should be assigned at initialization right? also in other importer too

rotkehlchen/tests/utils/dataimport.py
OjusWiZard, line 1: is there any test for messaged that includes total_entries, ignored_entries, etc.?

rotkehlchen/data_import/utils.py
```diff
@@ -27,12 +35,26 @@ def __init__(self, db: DBHandler) -> None:
...
+        self.total_entries: int = 0
+        self.imported_entries: int = 0
+        self.import_msgs: list[dict] = []
+        self.max_msgs = False
+        self.name: str = ''
 
     def import_csv(self, filepath: Path, **kwargs: Any) -> tuple[bool, str]:
         try:
             with self.db.user_write() as write_cursor:
                 self._import_csv(write_cursor, filepath=filepath, **kwargs)
```
OjusWiZard, line 47: we should reset the `self.total_entries` and others before calling `self._import_csv` again right? because what if someone imports another CSV again?

rotkehlchen/data_import/utils.py
```diff
@@ -75,9 +97,45 @@ def flush_all(self, write_cursor: DBCursor) -> None:
...
+            self.total_entries >= MIN_ENTRIES and
+            len(self.import_msgs) >= self.total_entries * MAX_ERROR_PERCENT
+        ):
+            self.max_msgs = True
+
+        self.import_msgs.append({
+            'row': row_index,
+            'msg': msg,
+            'warning': warning,
+        })
```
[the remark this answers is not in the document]
nicholasyoder: Since `self.total_entries` actually counts up during the import, then the following could happen: if the beginning of the file contains a high percentage of messages, it will stop collecting messages after line `MIN_ENTRIES`, but if the rest of the file has a lower percentage of messages, then eventually it will start collecting again after the number of collected messages is no longer `MAX_ERROR_PERCENT` of `self.total_entries`.

That's why I used `self.max_msgs`, so it can't start collecting errors again later. Generally I expect this would not happen anyway - if the beginning is full of problems the rest probably is too, but this way it won't do anything unexpected.

The comments on lines 105-108 were supposed to clarify this, but maybe they are too confusing?

rotkehlchen/data_import/importers/binance.py
nicholasyoder: Most of the importers do not currently have an `__init__` function. Is it really worth adding all of this just to be able to set the name at initialization?

```
...
if TYPE_CHECKING:
...

    def __init__(self, db: 'DBHandler') -> None:
        super().__init__(db=db)
        self.name = 'Binance'
```
If it really should be done that way I can add an init function to all the importers, but I assumed it would be better to simply define it at the beginning of the `_import_csv` function.

rotkehlchen/data_import/utils.py
```diff
@@ -75,9 +97,45 @@ def flush_all(self, write_cursor: DBCursor) -> None:
...
+        self.db.msg_aggregator.warnings.appendleft(msg)
+
 
 class UnsupportedCSVEntry(Exception):
-    """Thrown for external exchange exported entries we can't import"""
+    """Thrown for csv entries we fail import. Logged as a warning message."""
+
+
+class SkippedCSVEntry(Exception):
+    """Thrown for csv entries that we skip on purpose. Logged as a debug message."""
```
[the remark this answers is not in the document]
nicholasyoder: Its for stuff like [this]( https://github.com/nicholasyoder/rotki/blob/csv-import/rotkehlchen/data_import/importers/bitcoin_tax.py#L198), where its simply a row that should be skipped and the user doesn't need to worry about it. Basically I tried to use `SkippedCSVEntry` wherever there was a `log.debug` and `UnsupportedCSVEntry` wherever there was a `log.warning`.

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -1925,8 +1925,13 @@ def assert_binance_import_results(rotki: Rotkehlchen):
     assert asset_movements == expected_asset_movements
     assert expected_events == events
     expected_warnings = [
-        '2 Binance rows have bad format. Check logs for details.',
-        'Skipped 4 rows during processing binance csv file. Check logs for details',
+        'Failed to deserialize a timestamp from a null entry in binance',
+        'Unknown asset  provided.',
```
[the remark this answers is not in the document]
nicholasyoder: I've updated it here: https://github.com/nicholasyoder/rotki/blob/csv-import/rotkehlchen/data_import/importers/binance.py#L691 to include double quotes around the asset, but this specific message is no longer present in the test because there was actually a mistake in the importer logic that I fixed and that entry is now being imported without error (See https://discord.com/channels/657906918408585217/1080783409640648765/1278772451634970725).

It is quite possible that this error could still happen though, so I think adding the double quotes is still a good idea.

rotkehlchen/tests/utils/dataimport.py
nicholasyoder: There is now: https://github.com/nicholasyoder/rotki/blob/csv-import/rotkehlchen/tests/utils/dataimport.py#L1971

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -74,7 +76,22 @@ def assert_cointracking_import_results(rotki: Rotkehlchen):
...
+        message_type=WSMessageType.CSV_IMPORT_RESULT,
+        data={
+            'source_name': 'Cointracking',
+            'total_entries': 12,
+            'imported_entries': 7,
+            'messages': [
+                {'msg': 'Not importing ETH Transactions from Cointracking. Cointracking does not export enough data for them. Simply enter your ethereum accounts and all your transactions will be auto imported directly from the chain', 'rows': [1, 2], 'warning': True},  # noqa: E501
+                {'msg': 'Not importing BTC Transactions from Cointracking. Cointracking does not export enough data for them. Simply enter your BTC accounts and all your transactions will be auto imported directly from the chain', 'rows': [5], 'warning': True},  # noqa: E501
+                {'msg': 'Unknown asset ADS.', 'rows': [9], 'warning': True},
+                {'msg': 'Staking event for eip155:1/erc20:0xBB0E17EF65F82Ab018d8EDd776e8DD940327B28b(Axie Infinity Shard) at 1641386280 already exists in the DB', 'rows': [12], 'warning': False},  # noqa: E501
```
OjusWiZard, line 91: so total is 12, we imported 7, and we see 4 errors. This creates a confusion that what happened with 1. Was it skipped? If yes, then maybe include a message for that as well?

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -1924,11 +1963,20 @@ def assert_binance_import_results(rotki: Rotkehlchen):
     assert trades == expected_trades
     assert asset_movements == expected_asset_movements
     assert expected_events == events
-    expected_warnings = [
-        '2 Binance rows have bad format. Check logs for details.',
-        'Skipped 4 rows during processing binance csv file. Check logs for details',
-    ]
-    assert warnings == expected_warnings
+    mocked_notifier = cast(MockRotkiNotifier, rotki.msg_aggregator.rotki_notifier)
+    assert mocked_notifier.pop_message() == MockedWsMessage(
```
OjusWiZard, line 1967: can you also assert that after poppping this message there is no other message left? To check we are not flooding it

rotkehlchen/tests/api/test_data_import.py
OjusWiZard, line 1: so I found `legacy_messages_via_websockets` fixture from `rotkehlchen/tests/websocketsapi/test_legacy.py::test_query_legacy_message`. can you try that one? I guess that should solve this problem of reassigning the notifier everywhere.

rotkehlchen/data_import/utils.py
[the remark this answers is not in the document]
OjusWiZard: hmm then why not just set self.total_entries at the starting? like in binance.

rotkehlchen/data_import/importers/binance.py
OjusWiZard: I mean... we should keep it consistent, and also the holy bible of OOPs and DRY says that lol
I suggested a cleaner way do to it in another comment

rotkehlchen/data_import/importers/binance.py
```diff
@@ -656,48 +666,55 @@ def process_entry(
...
-            multirows[timestamp].append(csv_row)
-        except (DeserializationError, UnknownAsset, UnsupportedAsset) as e:
-            log.warning(f'Skipped binance csv row {csv_row} because of {e!s}')
-            skipped_count += 1
-        except KeyError as e:
-            log.error(f'Malformed binance csv columns! Broke on row {csv_row}. {e!s}')
-            return len(rows), {}
-
-    return skipped_count, multirows
+class BinanceImporter(BaseExchangeImporter):
```
OjusWiZard, line 670:
```suggestion
class BinanceImporter(BaseExchangeImporter):
    """Binance CSV importer"""
    name = 'Binance'
```

rotkehlchen/data_import/utils.py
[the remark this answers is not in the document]
nicholasyoder: If I understand correctly, the way the binance importer does it, the entire csv file is loaded into memory when it converts the csv file's contents to a list like this:
```
input_rows = list(csv.DictReader(csvfile))
self.total_entries = len(input_rows)
```
That seems like a bad idea, as its possible the user could try to load a very large file.

I [asked about this in discord](https://discord.com/channels/657906918408585217/1277696195883700344/1278072814200688721) and yabir seemed to think it would be better to not load it into memory.

rotkehlchen/tests/utils/dataimport.py
nicholasyoder: Notice that the first entry in `messages` has `'rows': [1, 2]`. So there are in fact 5 errors reported.

rotkehlchen/data_import/utils.py
```diff
@@ -75,9 +118,44 @@ def flush_all(self, write_cursor: DBCursor) -> None:
         self._asset_movements = []
         self._history_events = []
 
+    def append_msg(self, row_index: int, msg: str, warning: bool) -> None:
+        """Append message to queue to be sent to frontend"""
+        if self.max_msgs:
+            return
+
+        # Stop collecting msgs if more than MAX_ERROR_PERCENT have errors.
+        # Note that total_entries actually only contains the number of rows processed do far.
```
LefterisJP, line 127:
```suggestion
        # Note that total_entries actually only contains the number of rows processed so far.
```

rotkehlchen/data_import/utils.py
```diff
@@ -13,26 +15,67 @@
...
+        self.imported_entries: int = 0
+        self.import_msgs: list[dict] = []
+        self.max_msgs = False
 
     def import_csv(self, filepath: Path, **kwargs: Any) -> tuple[bool, str]:
+        # Reset variables in case of multiple csv imports
+        self.total_entries = 0
+        self.imported_entries = 0
+        self.import_msgs = []
+        self.max_msgs = False
```
LefterisJP, line 50:
```suggestion
        self.reset()
        
   def reset(self):
       "[Re]Initialize variables used by the CSV imports"""
        self.total_entries = 0
        self.imported_entries = 0
        self.import_msgs = []
        self.max_msgs = False

    def import_csv(self, filepath: Path, **kwargs: Any) -> tuple[bool, str]:
        self.reset()
```

rotkehlchen/data_import/utils.py
```diff
@@ -75,9 +118,44 @@ def flush_all(self, write_cursor: DBCursor) -> None:
...
+        ):
+            self.max_msgs = True
+
+        self.import_msgs.append({
+            'row': row_index,
+            'msg': msg,
+            'warning': warning,
+        })
+
+    def log_debug(self, row_index: int, csv_row: dict, msg: str) -> None:
```
LefterisJP, line 142: Why do you need log debug? Do we have both errors and warnings? Shouldn't we then have different functions?

rotkehlchen/data_import/utils.py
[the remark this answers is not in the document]
LefterisJP: If the user does not need to worry about something then you just use the normal log and don't send a websocket message.

rotkehlchen/data_import/utils.py
```diff
@@ -75,9 +118,44 @@ def flush_all(self, write_cursor: DBCursor) -> None:
         self._asset_movements = []
         self._history_events = []
 
+    def append_msg(self, row_index: int, msg: str, warning: bool) -> None:
+        """Append message to queue to be sent to frontend"""
```
LefterisJP, line 122: Move the comment description of the MAX_ERROR_PERCENT and explain the logic of what you are doing here in the docstring

rotkehlchen/tests/api/test_data_import.py
```diff
@@ -88,7 +89,7 @@ def test_data_import_cointracking(rotkehlchen_api_server, file_upload):
     result = assert_proper_sync_response_with_result(response)
     assert result is True
     # And also assert data was imported successfully
-    assert_cointracking_import_results(rotki)
+    assert_cointracking_import_results(rotki, websocket_connection)
```
LefterisJP, line 92: nitpick: But here you can leave as it's test and obvious. If more than 1 argument, use kwargs

```suggestion
    assert_cointracking_import_results(rotki, websocket_connection)
```

rotkehlchen/data_import/importers/uphold_transactions.py
```diff
@@ -24,14 +22,14 @@
...
-logger = logging.getLogger(__name__)
-log = RotkehlchenLogsAdapter(logger)
-
 
 UPHOLD_PREFIX = 'UPH_'
 
 
 class UpholdTransactionsImporter(BaseExchangeImporter):
+    """Uphold CSV importer"""
+    name = 'Uphold'
```
LefterisJP, line 31: Why not do it in the constructor?

This way it's less code, but having a constructor argument at `BaseExchangeImporter` properly creates the type and lets readers of the code and type checkers know that all importers have a name and it should be set at creation time of an importer.

What you do now is hacky.

rotkehlchen/data_import/importers/shapeshift_trades.py
```diff
@@ -111,21 +105,19 @@ def _import_csv(self, write_cursor: DBCursor, filepath: Path, **kwargs: Any) ->
...
                     continue
                 except DeserializationError as e:
-                    self.db.msg_aggregator.add_warning(
-                        f'Deserialization error during ShapeShift CSV import. '
-                        f'{e!s}. Ignoring entry',
-                    )
+                    self.log_warning(index, row, f'Deserialization error: {e!s}.')
+                    continue
+                except SkippedCSVEntry as e:
+                    self.log_debug(index, row, str(e))
```
LefterisJP, line 120: So yeah as I said before. If the user should not care or need to see something do not send it via websocket. I am not sure here. What is this? Should the user see it? Is it information he can act on?

rotkehlchen/data_import/importers/cointracking.py
```diff
@@ -91,9 +86,11 @@ def exchange_row_to_location(entry: str) -> Location:
 
 
 class CointrackingImporter(BaseExchangeImporter):
+    """Cointracking CSV importer"""
 
     def __init__(self, db: 'DBHandler') -> None:
         super().__init__(db=db)
+        self.name = 'Cointracking'
```
LefterisJP, line 93: so this should go to super

rotkehlchen/data_import/importers/bitmex.py
```diff
@@ -126,10 +129,12 @@ def _import_csv(self, write_cursor: DBCursor, filepath: Path, **kwargs: Any) ->
...
                 except DeserializationError as e:
-                    self.db.msg_aggregator.add_warning(
-                        f'Deserialization error during BitMEX CSV import. '
-                        f'{e!s}. Ignoring entry',
-                    )
+                    self.log_warning(index, row, f'Deserialization error: {e!s}.')
+                    continue
+                except UnsupportedCSVEntry as e:
+                    self.log_warning(index, row, str(e))
+                    continue
```
LefterisJP, line 138: why add continue here and below? Same question for almost all importers.

rotkehlchen/data_import/importers/binance.py
```diff
@@ -742,21 +765,26 @@ def _process_multiple_binance_entries(
...
+        for row in rows:
+            self.log_warning(row[INDEX], row, 'Could not process CSV entry')
         return None, 0
 
     def _process_binance_rows(
             self,
             write_cursor: DBCursor,
             multi: dict[Timestamp, list[BinanceCsvRow]],
-    ) -> None:
+    ) -> int:
```
LefterisJP, line 777: Add a short docstring explaining what this does and what it returns

LefterisJP (commented):
Left various comments

rotkehlchen/data_import/utils.py
nicholasyoder: Its only supposed to be using debug and warning, not error. The basic idea is that warnings are for rows that we couldn't import but may have had valuable data in them, while debug is for rows that we skip importing because we can tell they don't have any data worth importing (for example, [here](https://github.com/rotki/rotki/blob/21ca43e13af1e84248ab3f32f4e74ee32abe6c50/rotkehlchen/data_import/importers/cryptocom.py#L284) it is logged via debug that an entry has been skipped because is doesn't affect the wallet balance).

As was mentioned [here](https://github.com/rotki/rotki/pull/8458#discussion_r1736971529), I used `log.debug` and `log.warning` since that is what was being used in the importers already, also it would seem to me that if something is logged with `log.error` that should mean that the entire csv import operation failed instead of just a single entry failing to import.

rotkehlchen/data_import/importers/shapeshift_trades.py
nicholasyoder: Since we are sending the number of total and successfully imported rows to the frontend for the user to see, shouldn't we try to give some type of explanatory message for every row that is skipped? So it doesn't say something like "Imported 10/15 rows" and then only show 3 messages about rows that had problems and leave the user wondering what happened on the other skipped rows?

You can see the messages being logged here by looking at where the `SkippedCSVEntry` exception is raised on lines 73, 75, and 83 [here](https://github.com/rotki/rotki/blob/21ca43e13af1e84248ab3f32f4e74ee32abe6c50/rotkehlchen/data_import/importers/shapeshift_trades.py#L73). Actually, in this particular case, these should maybe be changed to `UnsupportedCSVEntry` so they get logged as warnings?

rotkehlchen/data_import/utils.py
```diff
@@ -13,26 +15,66 @@
...
+MAX_ERROR_PERCENT = 0.2  # max percent of messages to total entries
+MIN_ENTRIES = 50  # mininmum number of entries before checking MAX_ERROR_PERCENT
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
 
 
 class BaseExchangeImporter(ABC):
-    def __init__(self, db: DBHandler) -> None:
+    def __init__(self, db: DBHandler, name: str = '') -> None:
```
LefterisJP, line 31: Why give it an empty string default value. Shouldn't it always be given? Please make it a positional argument.

rotkehlchen/data_import/utils.py
```diff
@@ -75,9 +117,45 @@ def flush_all(self, write_cursor: DBCursor) -> None:
         self._asset_movements = []
         self._history_events = []
 
+    def append_msg(self, row_index: int, msg: str, warning: bool) -> None:
+        """Append message to queue to be sent to frontend
```
LefterisJP, line 121: Format the docstrings better please.

Also expplain what the arguments are. What is `warning` more specifically
```suggestion
        """Append message to queue to be sent to frontend.

```

rotkehlchen/data_import/importers/bitstamp.py
```diff
@@ -140,25 +149,31 @@ def _import_csv(self, write_cursor: DBCursor, filepath: Path, **kwargs: Any) ->
...
+                    self.send_message(
+                        row_index=index,
+                        csv_row=row,
+                        msg=f'Deserialization error: {e!s}.',
+                        is_error=True,
                     )
                 except ValueError as e:
-                    self.db.msg_aggregator.add_warning(
-                        f'Could not parse some values {e!s} in csv row {row!s}',
+                    self.send_message(
```
LefterisJP, line 172: here same change to reduce copy pasta. Wont' say it in other places but most probably every importer can improve like this

rotkehlchen/data_import/importers/cryptocom.py
```diff
@@ -374,13 +399,19 @@ def _import_cryptocom_associated_entries(
...
-                )
+                for row in debited_rows + [credited_row]:
+                    if row:
+                        self.send_message(
+                            row_index=row[INDEX],
+                            csv_row=row,
+                            msg=f'Failed to get {e!s} event at timestamp {timestamp}.',
+                            is_error=True,
+                        )
                 continue
```
LefterisJP, line 415: Readability. Very hard to understand so many indentation levels with a visual separator

```suggestion
                        )

                continue

```

rotkehlchen/data_import/importers/cryptocom.py
```diff
@@ -441,6 +479,8 @@ def _import_cryptocom_associated_entries(
                     )
                     self.add_trade(write_cursor, trade)
 
+                self.imported_entries += 1 + len(debited_rows)
```
LefterisJP, line 482: add a small comment why this is so. I can't understand why by quickly looking at it.

rotkehlchen/data_import/utils.py
```diff
@@ -75,9 +117,53 @@ def flush_all(self, write_cursor: DBCursor) -> None:
...
+            - is_error (bool): Indicates if the row failed to import (True)
+                or was skipped on purpose (False).
+        """
+        log.debug(f'Failed to import row {row_index} {csv_row} of {self.name} CSV file. {msg}.')
+        self.append_msg(row_index=row_index, msg=msg, is_error=is_error)
+
 
 class UnsupportedCSVEntry(Exception):
-    """Thrown for external exchange exported entries we can't import"""
+    """Thrown for csv entries we fail import."""
```
LefterisJP, line 162:
```suggestion
    """Raised for csv entries we fail to import."""
```

rotkehlchen/data_import/utils.py
```diff
@@ -75,9 +117,53 @@ def flush_all(self, write_cursor: DBCursor) -> None:
...
+
+
+class SkippedCSVEntry(Exception):
+    """Thrown for csv entries we skip on purpose."""
```
LefterisJP, line 166:
```suggestion
    """Raised for csv entries we skip on purpose."""
```

rotkehlchen/data_import/manager.py
```diff
@@ -94,6 +94,6 @@ def import_csv(
         """Imports csv data from `filepath`.`source` determines the format of the file.
         Returns (True, '') if imported successfully and (False, message) otherwise."""
         importer_type = source.get_importer_type()
-        importer = importer_type(db=self.db)
+        importer = importer_type(db=self.db)  # type: ignore[call-arg]
```
LefterisJP, line 97: Why is this needed? I see no new arguments here.

LefterisJP (approved):
LGTM

PR #8460, opened 2024-08-29
Remove un-needed magic fixtures and bump packages

LefterisJP:
Get rid of need_writable_cursor and need_cursor fixtures

They were remnants of before and needed to die


yabirgb (approved):
LGTM

Issue #8452, opened 2024-08-27
Handle monerium v2 contract upgrade


LefterisJP:
## Problem Definition

Monerium has made a contract upgrade: https://monerium.dev/docs/tokens. This has introduced many improvements but has also created some problems.

The way that it works now is that:
1. Approvals have been reset
2. For each v1 transfer both v1 and v2 token events are emitted.
3. For each v2 transfer only v2 token events are emitted.
4. Balances of both are mirrored.

The problem for rotki is:
- Which token to show as balance and which oracle to use.
- How to stop showing duplicate events in history
- How to do proper accounting

## Task

To handle the problems probably the best way would be:

1. Always use the v2 contract for balance queries. Ignore v1. Handle it as ignored asset even for balances so as to not double count balances or show more tokens owned
2. For history events let's say `T`  is the timestamp of the upgrade, `X` the old contract and `Y`  the new.  For every event involving either `X` or `Y` contracts after `T` count only `Y` and completely ignore `X` events and state changes.
3. For accounting do same as you would do for display problems in (2)
4. We also have to edit our minting/burning logic which searches for the mint/burn transactions and decorates them with the SEPA data. To search for mint/burn of v2 after the `T` time.

gislik:
Hi @LefterisJP 

With regards to proposed the solution, I have a few comments that I hope are helpful.

1. You can use either [V1](https://gnosisscan.io/token/0xcB444e90D8198415266c6a2724b7900fb12FC56E) or [V2](https://gnosisscan.io/token/0x420CA0f9B9b604cE0fd9C18EF134C705e5Fa3430) to query balances.  They will always show the same balance. V2 will be more future proof as we'll put all our effort to migrate wallets, explorers, etc to use that version.
2. `X` will contain all the event logs to reconstruct the state of EURE up until `T` and `Y` will contain the events after `T`. Note that before `T`, `Y` only has "checkpoint", i.e. one event per address representing the state at `T`. 

`T` = [35656951](https://gnosisscan.io/block/35656951)  for EURe on Gnosischain
`T` = [60733237](https://polygonscan.com/block/60733237) for EURe on Polygon

yabirgb:
Example of v2 token being sent and not emitting the v1 event https://gnosisscan.io/tx/0x67df15bb6abe97817548dcf2591c6015e4a7e35c143a90099718bab91421fa0e

yabirgb:
Fixed on https://github.com/rotki/rotki/pull/8496

PR #8466, opened 2024-08-30
Monerium fixes

yabirgb:
PR closing #8452

- [x] Always use the v2 contract for balance queries. Ignore v1. Handle it as ignored asset even for balances so as to not double count balances or show more tokens owned
- [x] For history events let's say T is the timestamp of the upgrade, X the old contract and Y the new. For every event involving either X or Y contracts after T count only Y and completely ignore X events and state changes.
- [ ] For accounting do same as you would do for display problems in (2) - 
 **For this since we will redecode the events it will work? Do we need to convert somehow from one contract to the other?**
- [x] We also have to edit our minting/burning logic which searches for the mint/burn transactions and decorates them with the SEPA data. To search for mint/burn of v2 after the T time.

Tokens taken from: https://monerium.dev/docs/tokens

### DIFF
--- a/rotkehlchen/chain/gnosis/decoding/decoder.py
+++ b/rotkehlchen/chain/gnosis/decoding/decoder.py
@@ -1,8 +1,9 @@
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Final
 
 from rotkehlchen.chain.evm.decoding.base import BaseDecoderTools
 from rotkehlchen.chain.evm.decoding.decoder import EVMTransactionDecoder
+from rotkehlchen.chain.gnosis.tokens import GNOSIS_MONERIUM_LEGACY_ADDRESSES
 from rotkehlchen.constants.assets import A_XDAI
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.types import ChecksumEvmAddress
@@ -14,6 +15,8 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
+
+MONERIUM_V2_CONTRACTS_BLOCK: Final = 35656951
 
 
 class GnosisTransactionDecoder(EVMTransactionDecoder):
@@ -37,6 +40,7 @@ class GnosisTransactionDecoder(EVMTransactionDecoder):
                 is_non_conformant_erc721_fn=self._is_non_conformant_erc721,
                 address_is_exchange_fn=self._address_is_exchange,
             ),
+            addresses_exceptions=dict.fromkeys(GNOSIS_MONERIUM_LEGACY_ADDRESSES, MONERIUM_V2_CONTRACTS_BLOCK),  # noqa: E501
         )
 
     # -- methods that need to be implemented by child classes --
--- a/rotkehlchen/chain/polygon_pos/decoding/decoder.py
+++ b/rotkehlchen/chain/polygon_pos/decoding/decoder.py
@@ -1,8 +1,9 @@
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Final
 
 from rotkehlchen.chain.evm.decoding.base import BaseDecoderTools
 from rotkehlchen.chain.evm.decoding.decoder import EVMTransactionDecoder
+from rotkehlchen.chain.polygon_pos.tokens import POLYGON_MONERIUM_LEGACY_ADDRESSES
 from rotkehlchen.constants.assets import A_POLYGON_POS_MATIC
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.types import ChecksumEvmAddress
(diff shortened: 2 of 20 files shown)


rotkehlchen/chain/gnosis/decoding/decoder.py
```diff
@@ -15,6 +16,8 @@
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
+MONERIUM_V2_CONTRACTS_DEPLOYMENT: Final = 35656951
```
LefterisJP, line 19:
```suggestion
MONERIUM_V2_CONTRACTS_BLOCK: Final = 35656951
```

rotkehlchen/chain/polygon_pos/decoding/decoder.py
```diff
@@ -15,6 +16,8 @@
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
+MONERIUM_V2_CONTRACTS_DEPLOYMENT: Final = 60733237
```
LefterisJP, line 19: same name change here

rotkehlchen/db/dbhandler.py
```diff
@@ -1246,15 +1246,19 @@ def get_tokens_for_address(
...
 
         If not, or if there is no saved entry, return None.
         """
-        ignored_asset_ids = self.get_ignored_asset_ids(cursor)
         last_queried_ts = None
-        querystr = 'SELECT key, value FROM evm_accounts_details WHERE account=? AND chain_id=? AND (key=? OR key=?)'  # noqa: E501
+        querystr = (
+            'SELECT key, value FROM evm_accounts_details WHERE account=? AND chain_id=? '
+            'AND (key=? OR key=?) AND value NOT IN '
+            '(SELECT value FROM multisettings WHERE name="ignored_asset")'
```
LefterisJP, line 1260: why add this here? Is it specific to this PR? It's not right? You seemto move the query from `ignored_asset_ids` here in the DB query. It's probably better/more performant right? That's why you are doing this?

But again not something specific for monerium?

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -2463,3 +2463,60 @@ def test_liquidity_withdrawal(database, gnosis_inquirer, gnosis_accounts, load_g
         ),
     ]
     assert events == expected_events
+
+
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('load_global_caches', [[CPT_CURVE]])
+@pytest.mark.parametrize('gnosis_accounts', [['0xc37b40ABdB939635068d3c5f13E7faF686F03B65']])
+def test_monerium_eure(gnosis_inquirer, gnosis_accounts, load_global_caches):
```
LefterisJP, line 2471:
```suggestion
def test_monerium_eure_v2(gnosis_inquirer, gnosis_accounts, load_global_caches):
```

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -2463,3 +2463,60 @@ def test_liquidity_withdrawal(database, gnosis_inquirer, gnosis_accounts, load_g
...
+    """Regression test for https://github.com/rotki/rotki/issues/8452"""
```
LefterisJP, line 2472: Add a tiny bit more details of what you are testing here. I guess you test that the events of v1 do not appear?

rotkehlchen/db/dbhandler.py
```diff
@@ -1246,15 +1246,19 @@ def get_tokens_for_address(
             cursor: 'DBCursor',
             address: ChecksumEvmAddress,
             blockchain: SupportedBlockchain,
+            token_exceptions: set[ChecksumEvmAddress],
```
LefterisJP, line 1249: I wonder ... would it make sense to turn this into a default argument with `None` as the default value?

Would need an extra if check at the end of thefunction but would make you stop needing to add this new `token_exceptions=set()` in many places.

Not sure tbh ... probably fine like that? What do you think?

LefterisJP (commented):
some comments

rotkehlchen/db/dbhandler.py
yabirgb: is not related to the PR. I made the change because I added a variable token_exceptions and then I saw that we were querying the ignored assets for each address in a loop and then I said... Okey then the ignored assets should be also in the argument and I should just combine the ignored assets and the exceptions... but then I looked at it and said... fuck we can do it better in sqlite and don't load to many things into memory. So is not related but I believe it's much better both for readibility and performance

rotkehlchen/db/dbhandler.py
yabirgb: I believe it is better to leave as it is because the only place where I call it with empty set is... in tests. In the code I don't. Also I have changed it in task/assets.py ... there Initially I thought it was not needed but reading it now I believe it can introduce errors so adjusted it.
image

LefterisJP (approved):
LGTM

PR #8468, opened 2024-09-01
Disable scheduled workflows + cassette merges in repo forks

dbast:
This disables scheduled workflows and cassette merges in repository forks:
* to save resources in forks (the workflows make not that much sense there)
* to reduce the github notification noise of failed workflows as those workflows anyways fail as forks by default are missing credentials to notify / push artifacts

Notes:
* The workflows can still be triggered by pushing commits to the `build` and `tests` branch in forks.
* the `always()` calls has to remain otherwise jobs would not run despite `github.repository == 'rotki/rotki'` evaluating to true (tested that).

This allows new OSS developers:
* to clone the repo without burning personal action resources via not working scheduled flows or the need to manually disable those
* to create/merge/close PRs **within** the fork to work against the ci workflow without triggering the cassette flow (before opening already tested PRs against the real parent repo)

Thanks!

## Checklist

### DIFF
--- a/.github/workflows/rotki_cassette_merge.yml
+++ b/.github/workflows/rotki_cassette_merge.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   merge:
     runs-on: ubuntu-latest
+    if: ${{ github.repository == 'rotki/rotki' }}
     environment: cassette-merge
     steps:
       - uses: rotki/action-cassette-deck@v2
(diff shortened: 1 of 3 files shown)


LefterisJP (approved):
Hey @dbast thank you very much for the PR and the nice explanation.

Indeed you are correct and this is a much needed change. Also consulted with @kelsos who is handling most of the CI and agrees.

So LGTM and I am merging.

Again thank you :bow:

PR #8470, opened 2024-09-02
Checkpoint log queries of thegraph delegations

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/ethereum/transactions.py
+++ b/rotkehlchen/chain/ethereum/transactions.py
@@ -1,9 +1,9 @@
 import logging
 from collections.abc import Sequence
 from contextlib import suppress
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
-from rotkehlchen.chain.ethereum.constants import LAST_GRAPH_DELGATIONS
+from rotkehlchen.chain.ethereum.constants import LAST_GRAPH_DELEGATIONS
 from rotkehlchen.chain.ethereum.modules.thegraph.constants import CONTRACT_STAKING
 from rotkehlchen.chain.evm.decoding.thegraph.constants import (
     CPT_THEGRAPH,
@@ -11,6 +11,7 @@ from rotkehlchen.chain.evm.decoding.thegraph.constants import (
 )
 from rotkehlchen.chain.evm.transactions import EvmTransactions
 from rotkehlchen.db.cache import DBCacheDynamic, DBCacheStatic
+from rotkehlchen.db.evmtx import DBEvmTx
 from rotkehlchen.db.filtering import EvmEventFilterQuery
 from rotkehlchen.db.history_events import DBHistoryEvents
 from rotkehlchen.errors.misc import AlreadyExists
@@ -37,10 +38,30 @@ class EthereumTransactions(EvmTransactions):
     ) -> None:
         super().__init__(evm_inquirer=ethereum_inquirer, database=database)
 
+    def _graph_delegation_callback(
+            self,
+            last_block_queried: int,
+            filters: dict[str, Any],
+    ) -> None:
+        """Callback used in query_for_graph_delegation_txns when querying logs.
+        Ensures that when iterating the logs we keep the progress of the queried
+        ranges.
+        """
+        with self.database.user_write() as write_cursor:
+            self.database.set_dynamic_cache(
+                write_cursor=write_cursor,
+                name=DBCacheDynamic.LAST_BLOCK_ID,
+                value=last_block_queried,
+                location=self.evm_inquirer.chain_name,
+                location_name=LAST_GRAPH_DELEGATIONS,
+                account_id=filters['l2Delegator'],
+            )
+
     def query_for_graph_delegation_txns(self, addresses: Sequence[ChecksumEvmAddress]) -> None:
         """This function is used to query for DelegationTransferredToL2 events from thegraph
         contract, if the events have delegated to any of the accounts. Then it will add the
         transactions to the database."""
+        log.debug('Started query of thegraph delegations')
         if len(addresses) == 0:
             return  # do nothing for no addresses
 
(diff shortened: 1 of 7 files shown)


rotkehlchen/chain/ethereum/transactions.py
```diff
@@ -27,6 +27,8 @@
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
+LOGS_QUERY_BLOCKS_RANGE: Final = 900_000
```
LefterisJP, line 30: We already have a range.Check EVM inquirer. For web3 nodes we have `WEB3_LOGQUERY_BLOCK_RANGE = 250000` and for etherscan `300000`

Internally the get_log function splits it.

Would it not make more sense to change the get_logs function and add an optional callback to do in between log queries? And then move the cache writing code into that callback?

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -1035,6 +1039,9 @@ def _get_logs(
                             'Couldnt decode an etherscan event due to {str(e)}}',
                         ) from e
 
+                if callback is not None:
```
LefterisJP, line 1042: 1. This is not enough. You are only doing this for the etherscan case. All other nodes go through `_query_web3_get_logs` which you can see in the very start.
2. This function is huge and hard to read. Probably why you missed it. Perhaps break it down in another commit?

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -918,6 +921,7 @@ def _get_logs(
             argument_filters: dict[str, Any],
             from_block: int,
             to_block: int | Literal['latest'] = 'latest',
+            callback: LogCallback | None = None,
```
LefterisJP, line 924: callback is too generic. It's a callback that will run at each log event query iteration. So `log_iteration_callback`. Can also do `log_iteration_cb` as `_cb` suffix is commonly used for callbacks, and you also have the type.

Rename for all functions.

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -1135,3 +1139,52 @@ def test_snapshots_dont_happen_always(rotkehlchen_api_server: 'APIServer') -> No
...
         assert cursor.execute(query).fetchone()[0] == 2
+
+
+@pytest.mark.parametrize('ethereum_accounts', [['0xc37b40ABdB939635068d3c5f13E7faF686F03B65', '0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12']])  # noqa: E501
+def test_graph_query_query_delegations(
+        eth_transactions: 'EthereumTransactions',
+        ethereum_accounts: list['ChecksumEvmAddress'],
+):
+    """Check that log events for graph delegations are queried in chunks
+    and we save correctly the range queried. Also ensures that only addresses
```
LefterisJP, line 1150:
```suggestion
    and we save the range queried correctly. Also ensures that only addresses
```

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -1135,3 +1139,52 @@ def test_snapshots_dont_happen_always(rotkehlchen_api_server: 'APIServer') -> No
...
+        assert callback_patch.call_count == 31  # math.ceil((target_block - from_block) / query_size)  # noqa: E501
+        with eth_transactions.database.conn.read_ctx() as cursor:
+            assert eth_transactions.database.get_dynamic_cache(
+                cursor=cursor,
+                name=DBCacheDynamic.LAST_BLOCK_ID,
+                location=eth_transactions.evm_inquirer.chain_name,
+                location_name=LAST_GRAPH_DELGATIONS,
+                account_id=ethereum_accounts[0],
+            ) == target_block
+            # we expect only one address and not the both tracked
```
LefterisJP, line 1189:
```suggestion
            # we expect only one address and not both tracked ones
```

rotkehlchen/chain/evm/node_inquirer.py
yabirgb: I didn't abstract this logic now. I prefer to do it in another moment

rotkehlchen/chain/ethereum/transactions.py
```diff
@@ -52,41 +73,57 @@ def query_for_graph_delegation_txns(self, addresses: Sequence[ChecksumEvmAddress
...
+                    filter_query=db_filter,
+                    has_premium=True,
+                )
+                if len(events) == 0:
                     continue
 
-            with self.database.conn.read_ctx() as cursor:
-                from_block = staking_contract.deployed_block
+                # find the earliest delegation to avoid querying a wide range
+                earliest_event = min(events, key=lambda event: event.timestamp)
```
LefterisJP, line 89: Can't you do that directly from the filter above? Use the `order_by` argument and make sure it returns events order by ascending timestamp? So then getting `events[0]` is the `earliest_event` and as such you don't need this at all.

rotkehlchen/chain/ethereum/transactions.py
```diff
@@ -52,41 +73,57 @@ def query_for_graph_delegation_txns(self, addresses: Sequence[ChecksumEvmAddress
...
+                log_iteration_cb=self._graph_delegation_callback,
             )
 
-            if len(events) == 0:
-                log.info(f'No graph delegation events found for {address} from block {from_block} to {to_block}')  # noqa: E501
+            if len(log_events) == 0:
+                log.info(f'No graph delegation events found for {address} from block {from_block} to {target_block}')  # noqa: E501
             else:
-                for event in events:
+                log.info(
```
LefterisJP, line 122: So now you are mixing info and debug in this function. Either turn them all to debug or all info. I would say all debug.

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -945,6 +953,8 @@ def _get_logs(
             filter_args['topics'] = filter_args['topics'][1:]
         events: list[dict[str, Any]] = []
         start_block = from_block
+
+        log.debug(f'Ready to query logs for {filter_args} at {contract_address} from {start_block} to {to_block}')  # noqa: E501
```
LefterisJP, line 957: this also needs a chain name

LefterisJP (approved):
LGTM

PR #8480, opened 2024-09-03
fix: matching logic in autocomplete

cleanerzkp:
Closes
## Checklist

Fix autocomplete functionality and improve filtering logic
- Add exactMatchMatcher computed property to handle exact matches
- Update applySuggestion function to handle exact matches
- Modify watch on search to update selectedMatcher
- Emit 'update:keyword' event to update parent component


docs/changelog.rst
```diff
@@ -16,6 +16,7 @@ Changelog
 * :feature:`8101` Failed transactions will now be displayed as failed and not just show "burned gas" event only.
 * :feature:`-` Transactions claiming SAFE tokens from vesting will now be properly decoded.
 * :feature:`-` Transactions locking, unlocking and withdrawing SAFE tokens will now be properly decoded. Also any locked SAFE tokens will be automatically detected and their balance counted.  
+* :bug:`` Improve autocomplete and filtering logic in FilterDropdown component.
```
kelsos, line 19:
```suggestion
* :bug:`-` Improve autocomplete and filtering logic in FilterDropdown component.
```

When we don't have an issue number you should add the dash in between.

kelsos:
So the fix for `tx_hash` seems to work fine, but we probably should add a hint for assets since now it appears weird before you start typing.

I also get the following warning that we need to figure out.
```
[Vue warn] Write operation failed: computed value is readonly
```

Also since you are already touching can you re-order the components based on the [docs](https://docs.rotki.com/contribution-guides/vue-typescript.html#setup-script-order)

also a small note, in the project we use [vueuse](https://vueuse.org/) and auto-imports for [get](https://vueuse.org/shared/get/#get) and [set](https://vueuse.org/shared/set/#set) that we use instead of accessing `ref.value directly`

docs/changelog.rst
```diff
@@ -16,11 +17,7 @@ Changelog
 * :feature:`4457` Rotki will show links to CoinGecko and CryptoCompare on the asset detail page if they exist.
 * :feature:`8101` Failed transactions will now be displayed as failed and not just show "burned gas" event only.
 * :feature:`-` Transactions claiming SAFE tokens from vesting will now be properly decoded.
-* :feature:`-` Transactions locking, unlocking and withdrawing SAFE tokens will now be properly decoded. Also any locked SAFE tokens will be automatically detected and their balance counted.
-* :bug:`-` rotki will now decode the swaps done on velodrome v2 in the right order.
-* :bug:`-` ZKSync Era tokens will now have prices queried properly by defillama.
-* :bug:`-` rotki will now query TheGraph delegations only for the addresses that interacted with the protocol.
-* :bug:`8452` Fix Monerium integration after the v2 contracts upgrade.
```
lukicenturi, line 23: Revert this

docs/changelog.rst
```diff
@@ -16,11 +17,7 @@ Changelog
...
+* :feature:`-` Transactions locking, unlocking and withdrawing SAFE tokens will now be properly decoded. Also any locked SAFE tokens will be automatically detected and their balance counted.
```
lukicenturi, line 20: Revert this, just keep from develop

frontend/app/src/components/table-filter/FilterDropdown.vue
```diff
@@ -1,4 +1,6 @@
 <script setup lang="ts">
+import { computed, ref, watch } from 'vue';
+import { get, set } from '@vueuse/core';
```
lukicenturi, line 3: We have auto import mechanism, so these things should be able to removed

frontend/app/src/locales/fr.json
```diff
@@ -4028,7 +4028,8 @@
     },
     "hint": {
       "description": "* tapez le filtre puis appuyez sur entrée, par exemple.",
-      "example": "start = 26/05/2022"
+      "example": "start = 26/05/2022",
+      "edit_note": "Double-click a filter to edit"
```
lukicenturi, line 4032: Nitpick, if you're not gonna translate this in other languages, just leave it in other locale files, and just fill the `en.json`

frontend/app/src/locales/en.json
```diff
@@ -4307,7 +4307,9 @@
     },
     "hint": {
       "description": "* type the filter and then press enter e.g.",
-      "example": "start = 26/05/2022"
+      "example": "start = 26/05/2022",
+      "edit_note": "Double-click a filter to edit"
```
lukicenturi, line 4311: Okay as I said this is not triggered by Double-click, but only single click, so:
```
Click applied filter to edit the value
```

frontend/app/src/components/table-filter/FilterDropdown.vue
```diff
@@ -204,12 +253,12 @@ const highlightedTextClasses = 'text-subtitle-2 text-rui-text-secondary';
     </div>
     <div
       v-else-if="keyword && matchers.length === 0"
-      class="pb-2"
+      class="pb-0"
```
lukicenturi, line 256: just remove it and you won't have any padding (?)

frontend/app/src/locales/fr.json
cleanerzkp: I saw they had some English in them hence

frontend/app/src/components/table-filter/FilterDropdown.vue
```diff
@@ -227,8 +272,11 @@ const highlightedTextClasses = 'text-subtitle-2 text-rui-text-secondary';
           :active="index === selectedSuggestion"
           :matcher="matcher"
           :class="{ highlightedMatcher: index === selectedSuggestion }"
-          @click="click($event)"
-        />
+          @click="click(matcher)"
+          @keydown.enter="handleEnter()"
+        >
+          <span class="font-bold text-rui-primary-main">{{ matcher.key }}:</span> {{ t(`table_filter.filter.${matcher.key}`) }}
```
lukicenturi, line 278: Does it do anything to you? Because FilterEntry doesn't even have a slot.

lukicenturi:
> So the fix for `tx_hash` seems to work fine, but we probably should add a hint for assets since now it appears weird before you start typing.
> 
> I also get the following warning that we need to figure out.
> 
> ```
> [Vue warn] Write operation failed: computed value is readonly
> ```
> 
> Also since you are already touching can you re-order the components based on the [docs](https://docs.rotki.com/contribution-guides/vue-typescript.html#setup-script-order)
> 
> also a small note, in the project we use [vueuse](https://vueuse.org/) and auto-imports for [get](https://vueuse.org/shared/get/#get) and [set](https://vueuse.org/shared/set/#set) that we use instead of accessing `ref.value directly`

frontend/app/src/components/table-filter/FilterDropdown.vue
```diff
@@ -136,11 +144,47 @@ watch(selectedSuggestion, async () => {
 });
 
 const highlightedTextClasses = 'text-subtitle-2 text-rui-text-secondary';
+
+function handleEnter() {
```
lukicenturi, line 148: what is this `handleEnter` does?

lukicenturi:
image
weird UI if asset is selected

lukicenturi:
image
hide this when no filter is selected

lukicenturi:
This should not be here at this state

lukicenturi:
the exclude one doesn't work: 

This is the develop
image

Your PR

lukicenturi:
When I hit enter on "protocol: eas", it fill search input with "location="
Uploading Screen Recording 2024-09-05 at 15.51.36.mov…

frontend/app/src/components/table-filter/FilterDropdown.vue
lukicenturi: I'm not sure where is this called, I put console.log here, it never gets called, by pressing enter while any button is highlighted

lukicenturi:
In general I think you overcomplicate thing, especially for the issue when we set exact matcher for keys that have empty suggestions.
I was thinking the approach would as simple as, dont set it as "selectedMatcher" until we have any symbol :, !=, = set

Also, I see that you show that exactMatcher with the template that supposed to be used to show the suggestion value.

cleanerzkp (commented):
LGTM!

lukicenturi (approved):

PR #8486, opened 2024-09-03
Add db migration for addressbook

### LINKED ISSUE #8350: Prevent addressbook duplication addition

yabirgb:
Closes #8350

## Checklist


rotkehlchen/db/upgrades/utils.py
```diff
@@ -0,0 +1,40 @@
...
+from rotkehlchen.types import ANY_BLOCKCHAIN_ADDRESSBOOK_VALUE
+
+if TYPE_CHECKING:
+    from rotkehlchen.db.drivers.gevent import DBCursor
+
+
+def fix_address_book_duplications(write_cursor: 'DBCursor') -> None:
+    """
+    Ensure that there are no invalid addresses when upgrading the table
+    code taken from migration 17. It could happen that migration 17 wasn't
```
LefterisJP, line 13:
```suggestion
    Ensure that there are no invalid addresses when upgrading the table.
    This code is moved here from migration 17. It could happen that migration 17 wasn't
```

rotkehlchen/db/upgrades/utils.py
```diff
@@ -0,0 +1,40 @@
...
+    executed when running this upgrade. After than upgrade the address_book
```
LefterisJP, line 14:
```suggestion
    executed when running this upgrade. After this fix we upgrade the address book table
```

LefterisJP (approved):
LGTM

PR #8490, opened 2024-09-04
Refactor (TaskManager): Run in a `bg_worker` process

OjusWiZard:
## Checklist

- [x] Move the execution of the scheduled tasks in a different process named `bg_worker`


rotkehlchen/tasks/server.py
```diff
@@ -0,0 +1,168 @@
...
+            premium_credentials=PremiumCredentials(
+                given_api_key=call_data.premium_credentials.api_key,
+                given_api_secret=call_data.premium_credentials.api_secret,
+            ) if call_data.HasField('premium_credentials') else None,
+            resume_from_backup=call_data.resume_from_backup,
+            initial_settings=ModifiableDBSettings.deserialize({
+                key: from_typed_data(data)
+                for key, data in call_data.initial_settings.items()
+            }),
+            sync_database=call_data.sync_database,
```
yabirgb, line 64: Do you need to send all those fields? Can't those be read form the db?

yabirgb: I expect that if you change the premium credentials you would need some kind of refresh

rotkehlchen/tasks/client.py
```diff
@@ -0,0 +1,104 @@
...
+            if return_data.error.name == ErrorType.UnknownError:
+                raise UnknownBGWorkerError(f'Got unknown DB error with message {return_data.error.message}')  # noqa: E501
+            # raise the same error that is raised by the DBWriter process
+            raise ENUM_TO_PROCESS_ERROR[return_data.error.name](return_data.error.message)
+
+    def unlock(
+            self,
+            user: str,
+            password: str,
+            sync_approval: Literal['yes', 'no', 'unknown'],
```
yabirgb, line 62: how do you handle the premium crederentials changing? I suspect you should read them from the db and not having to send them

rotkehlchen/tasks/manager.py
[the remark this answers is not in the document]
yabirgb: As mentioned in discord we would keep them duplicated but we should optimize as much as possible removing stuff that is not needed. Probably there are steps of the unlock that you need only in the main class

yabirgb (approved):
lgtm! Thanks

PR #8496, opened 2024-09-05
Convert monerium v1 events to v2

yabirgb:
When processing the decoded events if there are v1 events decoded after the v2 deployment ignore the v1 events and use v2 tokens

### DIFF
--- a/rotkehlchen/chain/evm/decoding/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/decoder.py
@@ -14,7 +14,6 @@ from gevent.lock import Semaphore
 from rotkehlchen.accounting.structures.balance import Balance
 from rotkehlchen.accounting.structures.types import ActionType
 from rotkehlchen.api.websockets.typedefs import WSMessageType
-from rotkehlchen.assets.asset import AssetWithOracles, EvmToken
 from rotkehlchen.assets.utils import TokenEncounterInfo, get_or_create_evm_token
 from rotkehlchen.chain.ethereum.utils import token_normalized_value
 from rotkehlchen.chain.evm.decoding.interfaces import ReloadableDecoderMixin
@@ -49,7 +48,14 @@ from rotkehlchen.history.events.structures.evm_event import EvmProduct
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.tasks.assets import maybe_detect_new_tokens
-from rotkehlchen.types import ChecksumEvmAddress, EvmTokenKind, EvmTransaction, EVMTxHash, Location
+from rotkehlchen.types import (
+    ChainID,
+    ChecksumEvmAddress,
+    EvmTokenKind,
+    EvmTransaction,
+    EVMTxHash,
+    Location,
+)
 from rotkehlchen.utils.misc import from_wei, hex_or_bytes_to_address, hex_or_bytes_to_int
 from rotkehlchen.utils.mixins.customizable_date import CustomizableDateMixin
 
@@ -67,6 +73,7 @@ from .structures import (
 from .utils import maybe_reshuffle_events
 
 if TYPE_CHECKING:
+    from rotkehlchen.assets.asset import Asset, AssetWithOracles, EvmToken
     from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer, EvmNodeInquirerWithDSProxy
     from rotkehlchen.chain.evm.transactions import EvmTransactions
     from rotkehlchen.db.dbhandler import DBHandler
@@ -83,7 +90,7 @@ class EventDecoderFunction(Protocol):
 
     def __call__(
             self,
-            token: EvmToken | None,
+            token: 'EvmToken | None',
             tx_log: EvmTxReceiptLog,
             transaction: EvmTransaction,
             decoded_events: list['EvmEvent'],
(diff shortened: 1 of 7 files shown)


rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -182,6 +190,7 @@ def __init__(
         self.value_asset = value_asset
         self.decoders: dict[str, DecoderInterface] = {}
         self.addresses_exceptions = addresses_exceptions or {}
+        self.exceptions_mappings = exceptions_mappings or {}
```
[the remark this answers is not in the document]
LefterisJP: you need to add this to the constructor docstring and explain what it is.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -434,9 +443,17 @@ def _decode_transaction(
         # Check if any rules should run due to the 4bytes signature of the input data
         fourbytes = transaction.input_data[:4]
         input_data_rules = self.rules.input_data_rules.get(fourbytes)
+        monerium_special_handling_event = False
 
         # decode transaction logs from the receipt
         for tx_log in tx_receipt.logs:
+            if (
```
LefterisJP, line 450: why not just move this to line 513 and not have an extra if for nothing?

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -493,6 +510,19 @@ def _decode_transaction(
             counterparties=counterparties,
         )
 
+        if monerium_special_handling_event is True:
+            new_events = []  # we create a new list to avoid remove operations and modifying while iterating  # noqa: E501
+            replacements = self.exceptions_mappings.values()
+            for event in events:
```
LefterisJP, line 516: you need to explain better what this does. I don't totally get it.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -493,6 +510,19 @@ def _decode_transaction(
...
+                if (replacement := self.exceptions_mappings.get(event.asset.identifier)) is not None:  # noqa: E501
+                    event.asset = Asset(replacement)
```
LefterisJP, line 518: Since you turn it into an asset shouldn't the mapping contain Assets instead?

rotkehlchen/chain/gnosis/modules/monerium/constants.py
```diff
@@ -14,3 +14,10 @@
     string_to_evm_address('0x50D1A74F4b6dcaCddD97fd442C0e22a4c97F2b7f'),  # USDe
     string_to_evm_address('0x614Bd419D3735C9eb51542C06e5Acc09a9953f61'),  # ISKe
 }
+
+LEGACY_TO_NEW_MONERIUM_MAPPINGS: Final = {
```
LefterisJP, line 18: rename all to v1_to_v2_.... 

you can't know if there won't be a v3

rotkehlchen/chain/evm/decoding/decoder.py
yabirgb: Didn't want to store the wrapper in memory for a case that might not ever exist for the user but it might not be that much.

rotkehlchen/chain/evm/decoding/decoder.py
yabirgb: This is here to avoid an extra loop if it isn't needed. What it does is checking if there is a monerium event and if that is the case set the `monerium_special_handling_event` flag. In case we have this flag set to true then later we iterate over all the events to find those that need to be updated. The condition is here to avoid looping twice all the events/logs

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -493,6 +519,23 @@ def _decode_transaction(
             counterparties=counterparties,
         )
 
+        if monerium_special_handling_event is True:
+            # When events that need special handling exist iterate over the decoded events an
```
LefterisJP, line 523:
```suggestion
            # When events that need special handling exist iterate over the decoded events and
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -493,6 +519,23 @@ def _decode_transaction(
...
+            # exchange the legacy assets by the v2 assets. Also delete v2 events to
+            # avoid duplications. V2 events exist only for the case of
+            # interacting with the v1 tokens since interacting with v2 doesn't emit v1 transfers.
```
LefterisJP, line 526:
```suggestion
            # avoid duplications. In the case of this exception handling, v2 events exist only for the case of
            # interacting with the v1 tokens since interacting with v2 doesn't emit v1 transfers.
```

rotkehlchen/chain/gnosis/modules/monerium/constants.py
```diff
@@ -14,3 +15,10 @@
     string_to_evm_address('0x50D1A74F4b6dcaCddD97fd442C0e22a4c97F2b7f'),  # USDe
     string_to_evm_address('0x614Bd419D3735C9eb51542C06e5Acc09a9953f61'),  # ISKe
 }
+
+V1_TO_V2_MONERIUM_MAPPINGS: Final = {
```
LefterisJP, line 19: You have checked those right? That the mappings are all correct I mean

yabirgb: Yes but I have double checked it

LefterisJP (approved):
Lgtm

Issue #8497, opened 2024-09-05
AutoComplete selection not showing


lukicenturi:
<img width="1158" alt="image" src="https://github.com/user-attachments/assets/5f98c8d0-79f2-4ed6-b5d8-f2005771c38f">

For some autocomplete, when selected, it's not showing what inside slot #selection
Perhaps it's missing props, maybe the "key-attr"

Identify also other autocompletes on other pages that may have this issue too.

PR #8498, opened 2024-09-05
fix: add missing key-attr on import page and history events

cleanerzkp:
This PR fixes issue #8497 by adding the missing key attribute on import page and history events filter by account.


lukicenturi (approved):
LGTM!

PR #8501, opened 2024-09-06
Improve Levenshtein search

yabirgb:
- Allows to search by address
- When filtering by evm chain results without any chain will also appear

### DIFF
--- a/rotkehlchen/db/filtering.py
+++ b/rotkehlchen/db/filtering.py
@@ -561,16 +561,24 @@ class DBTypeFilter(DBFilter):
 
 @dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=False)
 class DBEqualsFilter(DBFilter):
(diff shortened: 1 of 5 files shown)


lukicenturi (commented):
Funtionality okay

rotkehlchen/db/filtering.py
```diff
@@ -561,16 +561,24 @@ def prepare(self) -> tuple[list[str], list[Any]]:
 
 @dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=False)
 class DBEqualsFilter(DBFilter):
-    """Filter a column by comparing its column to its value for equality."""
+    """Filter a column by comparing its column to its value for equality.
+    For nullable columns that where we want to include the null rows set
```
LefterisJP, line 565:
```suggestion
    For nullable columns that's where we want to include the null rows set
```

LefterisJP (approved):
LGTM

PR #8503, opened 2024-09-06
Limit linting to our own directories

LefterisJP:
Explicit > implicit.

Plus the change of enforcing it all over messed with the local directories for multiple of our devs who also had other testing/python files in rotki's root directory.


yabirgb (approved):
LGTM

PR #8504, opened 2024-09-06
Add uniswap historical price oracle

### LINKED ISSUE #2217: Add new historical price oracle - Uniswap
## Abstract

For ethereum tokens we can query their historical price directly from uniswap itself by reading the chain. To get data for a pair from the chain we would need either a graph query or an archive node or a cache made by an archive node.

Related: #2216 

## Task

- [ ] Figure out how to do it in the most user friendly way. Since it involves graph queries and archive nodes it would most probably be premium only.
- [ ] Add a new historical price oracle, Uniswap so that users can customize its priority.

nicholasyoder:
Closes #2217

### DIFF
--- a/rotkehlchen/chain/ethereum/oracles/uniswap.py
+++ b/rotkehlchen/chain/ethereum/oracles/uniswap.py
@@ -2,35 +2,46 @@ import abc
 import logging
 from functools import reduce
 from operator import mul
-from typing import TYPE_CHECKING, NamedTuple
+from typing import TYPE_CHECKING, Final, NamedTuple
 
 from eth_utils import to_checksum_address
 from web3.types import BlockIdentifier
 
-from rotkehlchen.assets.asset import AssetWithOracles, EvmToken
+from rotkehlchen.assets.asset import Asset, AssetWithOracles, EvmToken
 from rotkehlchen.chain.ethereum.utils import token_normalized_value
 from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
 from rotkehlchen.chain.evm.contracts import EvmContract
 from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.constants import ONE, ZERO
-from rotkehlchen.constants.assets import A_DAI, A_ETH, A_USD, A_USDC, A_USDT, A_WETH
+from rotkehlchen.constants.assets import (
+    A_DAI,
+    A_ETH,
+    A_ETH_EURE,
+    A_EUR,
+    A_USD,
+    A_USDC,
+    A_USDT,
+    A_WETH,
+)
 from rotkehlchen.constants.prices import ZERO_PRICE
 from rotkehlchen.constants.resolver import ChainID, ethaddress_to_identifier
 from rotkehlchen.errors.asset import UnknownAsset, WrongAssetType
 from rotkehlchen.errors.defi import DefiPoolError
-from rotkehlchen.errors.price import PriceQueryUnsupportedAsset
+from rotkehlchen.errors.price import NoPriceForGivenTimestamp, PriceQueryUnsupportedAsset
 from rotkehlchen.fval import FVal
 from rotkehlchen.inquirer import Inquirer
-from rotkehlchen.interfaces import CurrentPriceOracleInterface
+from rotkehlchen.interfaces import HistoricalPriceOracleInterface
 from rotkehlchen.logging import RotkehlchenLogsAdapter
-from rotkehlchen.types import ChecksumEvmAddress, EvmTokenKind, Price
+from rotkehlchen.types import ChecksumEvmAddress, EvmTokenKind, Price, Timestamp
 from rotkehlchen.utils.mixins.cacheable import CacheableMixIn, cache_response_timewise
 
 if TYPE_CHECKING:
     from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
 
-UNISWAP_FACTORY_DEPLOYED_BLOCK = 12369621
-SINGLE_SIDE_USD_POOL_LIMIT = 5000
+UNISWAPV3_FACTORY_DEPLOYED_BLOCK: Final = 12369621
+UNISWAPV2_FACTORY_DEPLOYED_BLOCK: Final = 10000835
+MULTICALL_DEPLOYED_BLOCK: Final = 14353601
+SINGLE_SIDE_USD_POOL_LIMIT: Final = 5000
 
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
@@ -49,13 +60,13 @@ class PoolPrice(NamedTuple):
         )
 
 
-class UniswapOracle(CurrentPriceOracleInterface, CacheableMixIn):
+class UniswapOracle(HistoricalPriceOracleInterface, CacheableMixIn):
     """
     Provides shared logic between Uniswap V2 and Uniswap V3 to use them as price oracles.
     """
     def __init__(self, ethereum_inquirer: 'EthereumInquirer', version: int):
         CacheableMixIn.__init__(self)
-        CurrentPriceOracleInterface.__init__(self, oracle_name=f'Uniswap V{version} oracle')
+        HistoricalPriceOracleInterface.__init__(self, oracle_name=f'Uniswap V{version} oracle')
         self.ethereum = ethereum_inquirer
         self.weth = A_WETH.resolve_to_evm_token()
         self.routing_assets = [
@@ -88,6 +99,13 @@ class UniswapOracle(CurrentPriceOracleInterface, CacheableMixIn):
         token1 of the pool.
         May raise:
         - DefiPoolError
+        """
+
+    @abc.abstractmethod
+    def is_before_contract_creation(self, block_identifier: BlockIdentifier) -> bool:
+        """Check if block_identifier is before the creation of the uniswap version's contract.
+        When querying current prices block_identifier will be 'latest' instead of a block number,
+        so return False if block_identifier is not an int.
         """
 
     def _find_pool_for(
@@ -183,9 +201,11 @@ class UniswapOracle(CurrentPriceOracleInterface, CacheableMixIn):
             from_asset: AssetWithOracles,
             to_asset: AssetWithOracles,
             block_identifier: BlockIdentifier,
+            timestamp: Timestamp | None = None,
     ) -> Price:
         """
         Return the price of from_asset to to_asset at the block block_identifier.
+        Timestamp enables properly raising NoPriceForGivenTimestamp during historical price queries
 
         Can raise:
         - DefiPoolError
@@ -195,6 +215,13 @@ class UniswapOracle(CurrentPriceOracleInterface, CacheableMixIn):
             f'Searching price for {from_asset} to {to_asset} at '
             f'{block_identifier!r} with {self.name}',
         )
+
+        if timestamp is not None and self.is_before_contract_creation(block_identifier):
+            raise NoPriceForGivenTimestamp(
+                from_asset=from_asset,
+                to_asset=to_asset,
+                time=timestamp,
+            )
 
         # Uniswap V2 and V3 use in their contracts WETH instead of ETH
         if from_asset == A_ETH:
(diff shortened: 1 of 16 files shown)


rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -252,6 +269,21 @@ def get_price(
         price = FVal(reduce(mul, [item.price for item in prices_and_tokens], 1))
         return Price(price)
 
+    def _resolve_assets(self, assets: list[Asset | AssetWithOracles]) -> list[AssetWithOracles]:
```
yabirgb, line 272: can you make this a tuple `(AssetWithOracles | Asset, AssetWithOracles | Asset)`

rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -280,6 +307,69 @@ def query_current_price(
...
+            pool_contract: EvmContract,
+            methods: list[str],
+            block_identifier: BlockIdentifier,
+            timestamp: Timestamp | None,
+    ) -> dict:
+        """Call methods on the pool contract using multicall if timestamp is None (for current)
+        or is after multicall contract creation, otherwise use individual calls.
+        Returns a dict that maps method names to responses.
+        """
+        if timestamp is None or timestamp > 1620277200:
```
yabirgb, line 321: Move this timestamp to a final constant

rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -280,6 +307,69 @@ def query_current_price(
         )
         return price, False
 
+    def _call_methods(
+            self,
+            pool_contract: EvmContract,
+            methods: list[str],
+            block_identifier: BlockIdentifier,
+            timestamp: Timestamp | None,
```
yabirgb, line 315: why do you provide both block_identifier and timestamp? It should be stated in the docstring what those mean and what is the flow for when one is missing. I don't totally get it

rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -280,6 +307,69 @@ def query_current_price(
...
+        return self.is_before_contract_creation(timestamp)
+
+    def query_historical_price(
+            self,
+            from_asset: Asset,
+            to_asset: Asset,
+            timestamp: Timestamp,
+    ) -> Price:
+        to_asset, from_asset = self._resolve_assets([to_asset, from_asset])
+
```
yabirgb, line 365:
```suggestion
```

rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -280,6 +307,69 @@ def query_current_price(
...
+            }
+        return output
+
+    def can_query_history(
+            self,
+            from_asset: Asset,
+            to_asset: Asset,
+            timestamp: Timestamp,
+            seconds: int | None = None,
+    ) -> bool:
```
yabirgb, line 355: add a docstring stating that this is to comply with the interface for price historian

rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -334,6 +424,7 @@ def get_pool_price(
             self,
             pool_addr: ChecksumEvmAddress,
             block_identifier: BlockIdentifier = 'latest',
+            timestamp: Timestamp | None = None,
```
yabirgb, line 427: this is what I don't get completly. You can go from timestamp to block identifier. I would expect that calling this with timesatmp = None means that you query `block_identifier=latest` in the contract. It's confusing to provide both arguments. Also this function should have explanations in the docstring

rotkehlchen/rotkehlchen.py
```diff
@@ -430,6 +422,18 @@ def unlock_user(
 
         uniswap_v2_oracle = UniswapV2Oracle(ethereum_inquirer)
         uniswap_v3_oracle = UniswapV3Oracle(ethereum_inquirer)
+
+        # Initialize the price historian singleton
+        PriceHistorian(
```
yabirgb, line 427:
```suggestion
        PriceHistorian(  # Initialize the price historian singleton
```

rotkehlchen/tests/fixtures/history.py
```diff
@@ -24,6 +25,16 @@ def fixture_defillama():
     return Defillama()
 
 
+@pytest.fixture(name='uniswapv2')
+def fixture_uniswapv2(ethereum_inquirer):
```
yabirgb, line 29: same for the other and the function name
```suggestion
@pytest.fixture(name='uniswapv2_inquirer')
def fixture_uniswapv2(ethereum_inquirer):
```

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -1030,7 +1030,7 @@ def try_insert_mapping(cur):
         oracles_after_upgrade = cursor.execute(
             'SELECT value FROM settings WHERE name="historical_price_oracles"',
         ).fetchone()[0]
-        assert oracles_after_upgrade == '["manual", "cryptocompare", "coingecko", "defillama"]'
+        assert oracles_after_upgrade == json.dumps([oracle.serialize() for oracle in DEFAULT_HISTORICAL_PRICE_ORACLES_ORDER])  # noqa: E501
```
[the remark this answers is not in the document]
nicholasyoder: `DEFAULT_HISTORICAL_PRICE_ORACLES_ORDER` is used to set the value in the [old upgrade](https://github.com/rotki/rotki/blob/develop/rotkehlchen/db/upgrades/v34_v35.py#L362). 
Now we've changed `DEFAULT_HISTORICAL_PRICE_ORACLES_ORDER` to include the uniswap oracles and so this test fails, since it was hardcoded in the test and not the upgrade.

rotkehlchen/chain/ethereum/oracles/uniswap.py
nicholasyoder: Yes, I realize this was a bit confusing and unnecessary. The block_identifier can be used for most of what I was using timestamps for. I've removed the timestamp argument from most of these functions, except for `get_price` where it is needed to raise `NoPriceForGivenTimestamp` correctly.

rotkehlchen/chain/ethereum/oracles/uniswap.py
nicholasyoder: See [this comment](https://github.com/rotki/rotki/pull/8504#discussion_r1750877812)

rotkehlchen/db/upgrades/v43_v44.py
```diff
@@ -96,6 +98,25 @@ def _addressbook_schema_update(write_cursor: 'DBCursor') -> None:
...
 
 
+@enter_exit_debug_log()
+def _add_uniswap_to_historical_oracles(cursor: 'DBCursor') -> None:
+    """Add uniswapv2 and uniswapv3 to the list of historical price oracles."""
+    cursor.execute('SELECT value FROM settings WHERE name=?', ('historical_price_oracles',))
+    price_oracles = cursor.fetchone()
+
+    try:
+        oracles = json.loads(price_oracles[0] if price_oracles else '')
```
yabirgb, line 108: I just notice that this will fail. You are doing an append later so it should be a list

rotkehlchen/db/upgrades/v43_v44.py
```diff
@@ -96,6 +98,25 @@ def _addressbook_schema_update(write_cursor: 'DBCursor') -> None:
     fix_address_book_duplications(write_cursor=write_cursor)
 
 
+@enter_exit_debug_log()
+def _add_uniswap_to_historical_oracles(cursor: 'DBCursor') -> None:
+    """Add uniswapv2 and uniswapv3 to the list of historical price oracles."""
+    cursor.execute('SELECT value FROM settings WHERE name=?', ('historical_price_oracles',))
+    price_oracles = cursor.fetchone()
```
yabirgb, line 105: If the variable isn't set I believe you don't need to do anything since it will use the default. I believe that if price_oracles is None you just exist the logic

rotkehlchen/tests/db/test_db_upgrades.py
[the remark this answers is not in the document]
yabirgb: can you modify the old upgrade to have the values hardocoded? It can have unexpected behavior  in old upgrades to have new values

rotkehlchen/db/upgrades/v43_v44.py
```diff
@@ -96,6 +98,27 @@ def _addressbook_schema_update(write_cursor: 'DBCursor') -> None:
...
+    if (price_oracles := cursor.fetchone()) is None:
+        return
+
+    try:
+        oracles = json.loads(price_oracles[0])
+    except json.JSONDecodeError:
+        return
+
+    oracles.append(HistoricalPriceOracle.UNISWAPV3.serialize())
+    oracles.append(HistoricalPriceOracle.UNISWAPV2.serialize())
```
yabirgb, line 114: nitpick: here is maybe better to use `.extend[...]`

LefterisJP: yeah + 1. Use extend

rotkehlchen/db/upgrades/v43_v44.py
```diff
@@ -96,6 +98,27 @@ def _addressbook_schema_update(write_cursor: 'DBCursor') -> None:
...
+def _add_uniswap_to_historical_oracles(cursor: 'DBCursor') -> None:
+    """Add uniswapv2 and uniswapv3 to the list of historical price oracles."""
+    cursor.execute('SELECT value FROM settings WHERE name=?', ('historical_price_oracles',))
+    if (price_oracles := cursor.fetchone()) is None:
+        return
+
+    try:
+        oracles = json.loads(price_oracles[0])
+    except json.JSONDecodeError:
+        return
```
LefterisJP, line 111: this would be a pretty bad error to happen. Log an error here

rotkehlchen/globaldb/upgrades/v8_v9.py
```diff
@@ -13,6 +13,12 @@ def _addressbook_schema_update(write_cursor: 'DBCursor') -> None:
     fix_address_book_duplications(write_cursor=write_cursor)
 
 
+@enter_exit_debug_log()
+def _add_uniswap_to_price_history_source_types(write_cursor: 'DBCursor') -> None:
+    """Add entries for Uniswap V2 and V3 to price_history_source_types table"""
+    write_cursor.execute("INSERT OR IGNORE INTO price_history_source_types(type, seq) VALUES ('G', 7), ('H', 8);")  # noqa: E501
```
LefterisJP, line 19: nitpick: why not do an executemany and use `?.?` and bindings? I did not know what you wrote actually works :smile:

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2492,6 +2492,10 @@ def test_upgrade_db_43_to_44(user_data_dir, messages_aggregator):
             (bad_address, None, 'yabirgb.eth'),
             (bad_address, None, 'yabir.eth'),
         }
+        cursor.execute(
```
LefterisJP, line 2495: hmmm even if it's a test please let's keep clean practises. We never open a cursor for writing longer than needed. Undo the previous change and open another write cursor only for this here.

rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -252,6 +279,20 @@ def get_price(
         price = FVal(reduce(mul, [item.price for item in prices_and_tokens], 1))
         return Price(price)
 
+    def _resolve_assets(self, assets: tuple[Asset, Asset]) -> list[AssetWithOracles]:
+        """Ensure assets are resolved to AssetWithOracles."""
+        return [asset.resolve_to_asset_with_oracles() for asset in assets]
```
LefterisJP, line 284: nitpick: Hmm okay so this is now used in only 1 place and it's just an 1liner. Just remove the function.

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2493,6 +2493,12 @@ def test_upgrade_db_43_to_44(user_data_dir, messages_aggregator):
             (bad_address, None, 'yabir.eth'),
         }
 
+    with db_v43.conn.write_ctx() as cursor:
+        cursor.execute(
```
LefterisJP, line 2497: nitpick: Name write cursors accurately

```suggestion
    with db_v43.conn.write_ctx() as write_cursor:
        write_cursor.execute(
```

LefterisJP (approved):
Left 2 nitpicks but will merge

PR #8511, opened 2024-09-09
Reduce etherscan calls in web3 node sync check

nicholasyoder:


rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -92,7 +92,7 @@ def _is_synchronized(current_block: int, latest_block: int) -> tuple[bool, str]:
             - message: A message containing information on what the status is.
     """
     message = ''
-    if current_block < (latest_block - 20):
+    if current_block < (latest_block - 20) or current_block > (latest_block + 20):
```
yabirgb, line 95: Don't do this since for example arbitrum publishes a lot of blocks per second

```suggestion
    if current_block < (latest_block - 20)
```

if it's after the latest block it's fine

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -214,6 +214,7 @@ def __init__(
         self.database = database
         self.blockchain = blockchain
         self.etherscan = etherscan
+        self.etherscan_block: BlockNumber = BlockNumber(0)
```
yabirgb, line 217: Add a comment explaining what this constant is

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -460,13 +461,12 @@ def attempt_connect(
...
                         if isinstance(current_block, int) is False:  # Check for https://github.com/rotki/rotki/issues/6350  # TODO: Check if web3.py v6 has a check for this # noqa: E501
                             raise RemoteError(f'Found non-int current block:{current_block}')
-                        latest_block = self.query_highest_block()
                     except (requests.exceptions.RequestException, RemoteError) as e:
                         msg = f'Could not query latest block due to {e!s}'
                         log.warning(msg)
                         synchronized = False
                     else:
-                        synchronized, msg = _is_synchronized(current_block, latest_block)
+                        synchronized, msg = _is_synchronized(current_block, self.etherscan_block)
```
yabirgb, line 469: add an assertion here that the etherscan_block is not 0. Assertions are removed for production

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -504,10 +504,14 @@ def attempt_connect(
...
-            if weighted_node.node_info.name == self.etherscan_node_name:
-                continue
 
+        # remove etherscan nodes
+        nodes = [node for node in nodes if node.node_info.name != self.etherscan_node_name]
+        if not nodes:
+            return
+
+        self.etherscan_block = self.query_highest_block()
+        for weighted_node in nodes:
```
yabirgb, line 514: as you can see you are looping twice here. I guess you have it to remove the etherscan node and then query the highest block... mmm this should be stated in the code why you do it to avoid any future person from chaging the code and breaking your assumptions but not sure if we can do better

```suggestion
        self.etherscan_block = self.query_highest_block()
        for weighted_node in nodes:
            if weighted_node.node_info.name == self.etherscan_node_name:
                continue
            if len(self.web3_mapping) == 0:
                  self.etherscan_block = self.query_highest_block()
```

It's one extra check that is annoying but you avoid looping twice

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -504,10 +510,16 @@ def attempt_connect(
...
     def connect_to_multiple_nodes(self, nodes: Sequence[WeightedNode]) -> None:
         self.web3_mapping = {}
-        for weighted_node in nodes:
-            if weighted_node.node_info.name == self.etherscan_node_name:
-                continue
 
+        # Remove etherscan nodes and return if all nodes use etherscan,
+        # so we don't query the highest block unnecessarily.
+        nodes = [node for node in nodes if node.node_info.name != self.etherscan_node_name]
+        if not nodes:
```
yabirgb, line 517: It's better to do 

```suggestion
        if len(nodes) == 0:
```

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -214,6 +214,7 @@ def __init__(
         self.database = database
         self.blockchain = blockchain
         self.etherscan = etherscan
+        self.etherscan_block: BlockNumber = BlockNumber(0)  # caches highest block for checking sync while connecting to nodes  # noqa: E501
```
LefterisJP, line 217: nitpick: Doesn't it understand the type from the assignment?

```suggestion
        self.etherscan_block = BlockNumber(0)  # caches highest block for checking sync while connecting to nodes  # noqa: E501
```

LefterisJP (approved):
LGTM with just one nit

PR #8515, opened 2024-09-10
fix: hide empty account balance table

lukicenturi:
- [x] Hide empty account balances table
- [x] Persist "Aggregated assets" and "Per chain" selected tab.


cleanerzkp (approved):
lgtm

PR #8517, opened 2024-09-10
feat: disclaimer for other browser wallet imports

lukicenturi:
Add disclaimer in browser wallet page, that only metamask support the multiple address addition.


cleanerzkp (approved):
Screenshot 2024-09-10 at 11 35 33

lgtm

PR #8523, opened 2024-09-10
Fix multichain account label editing error

nicholasyoder:


rotkehlchen/db/addressbook.py
```diff
@@ -109,9 +109,28 @@ def update_addressbook_entries(
...
+                    ).fetchone()[0] > 0
+                ):
+                    try:
+                        # Use add_addressbook_entries to delete blockchain specific entries
+                        # and replace them with a single entry.
+                        self.add_addressbook_entries(write_cursor, [entry])
+                        continue  # skip update query, we just added a new entry instead
+                    except InputError:
+                        # Single entry already exists, it must be updated using query below
+                        pass
```
yabirgb, line 132: use supress instead. We use it at some places. Keep the comment to explain why this is being used

yabirgb: Can you also create a test that reproduces the error that luki has and that works after your change?

rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -839,3 +840,33 @@ def test_insert_into_addressbook_no_blockchain(
...
+            chain_type=ChainType.EVM.serialize(),
+        ),
+        json={
+            'accounts': [{
+                'address': test_address,
+                'label': 'new name',
+            }],
+        },
+    )
+    assert_proper_sync_response_with_result(response)
```
LefterisJP, line 872: this test should also fetch the label and test that it's set properly

LefterisJP (approved):
LGTM if tests pass

PR #8532, opened 2024-09-11
Avoid querying polkadot node if setting is empty

yabirgb:
Since we query the own node for polkadot from settings it can happen that the user resets it and it gets saved as an empty string. In that case the app was trying to connect to the node

Closes #(issue_number)

## Checklist


rotkehlchen/tests/fixtures/greenlets.py
```diff
@@ -3,6 +3,6 @@
 from rotkehlchen.greenlets.manager import GreenletManager
 
 
-@pytest.fixture(scope='session')
+@pytest.fixture
```
LefterisJP, line 6: Put this out of the PR. And probably what would make sense here is as you also hinted to in discord:

to just yield and then call greenlet_manager.clear(). As this is what clears the tasks.

PR #8533, opened 2024-09-11
Remove argument not needed

yabirgb:
image


LefterisJP (approved):
LGTM. Possibly we first gave DM, then evm inquirer and that's how it was built. Having clear dependencies is also not bad, but anyway having less args is also nice

PR #8537, opened 2024-09-12
Fix (API): Pagination in `/assets/all`

OjusWiZard:
## Checklist

- [x] Fix pagination in the all assets retrieval function to apply offset after ignoring the ignored/spam assets.


LefterisJP (approved):
LGTM!

PR #8545, opened 2024-09-13
Avoid storing in memory events discarded

yabirgb:
I saw that we only use the decoding function without tx_hash in tests so I removed it. Also added a conditional check to discard the generated events if we aren't going to consume them later

### DIFF
--- a/rotkehlchen/chain/evm/decoding/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/decoder.py
@@ -591,17 +591,15 @@ class EVMTransactionDecoder(ABC):
                 )
             log.debug(f'Finished task to process undecoded transactions for {self.evm_inquirer.chain_name} with {limit=}')  # noqa: E501
 
-    def decode_transaction_hashes(
+    def decode_and_get_transaction_hashes(
             self,
             ignore_cache: bool,
-            tx_hashes: list[EVMTxHash] | None,
+            tx_hashes: list[EVMTxHash],
             send_ws_notifications: bool = False,
             delete_customized: bool = False,
     ) -> list['EvmEvent']:
-        """Make sure that receipts are pulled + events decoded for the given transaction hashes.
-        If delete_customized is True then also customized events are deleted before redecoding.
-
-        The transaction hashes must exist in the DB at the time of the call
+        """
+        Thin wrapper around _decode_transaction_hashes that returns the decoded events.
 
         May raise:
         - DeserializationError if there is a problem with contacting a remote to get receipts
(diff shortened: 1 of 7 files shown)


rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -588,20 +588,24 @@ def get_and_decode_undecoded_transactions(
...
             delete_customized: bool = False,
+            return_events: bool = True,
     ) -> list['EvmEvent']:
         """Make sure that receipts are pulled + events decoded for the given transaction hashes.
         If delete_customized is True then also customized events are deleted before redecoding.
 
-        The transaction hashes must exist in the DB at the time of the call
+        The transaction hashes must exist in the DB at the time of the call.
+
+        If return_events is set to True whe return the decoded events for the given tx hashes
```
LefterisJP, line 608:
```suggestion
        If return_events is set to True, then we return the decoded events for the given tx hashes
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -649,7 +646,9 @@ def decode_transaction_hashes(
                 ignore_cache=ignore_cache,
                 delete_customized=delete_customized,
             )
-            events.extend(new_events)
+
+            if return_events:
```
LefterisJP, line 650: I would just create a different function that returns the events. No point in having it return empty list.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -612,13 +616,6 @@ def decode_transaction_hashes(
         refresh_balances = False
         with self.database.conn.read_ctx() as cursor:
             self.reload_data(cursor)
-            # If no transaction hashes are passed, decode all transactions.
```
LefterisJP, line 615: Hmmm you removed this functionality. It was not used anywhere apart from that test you deleted?

rotkehlchen/tests/unit/test_evm_tx_decoding.py
```diff
@@ -224,16 +224,6 @@ def test_query_and_decode_transactions_works_with_different_chains(
     hashes = dbevmtx.get_transaction_hashes_not_decoded(chain_id=ChainID.ETHEREUM, limit=None)
     assert len(hashes) == 1
 
-    # decode evm transactions using the optimism decoder and without providing any tx hash
-    optimism_transaction_decoder.decode_transaction_hashes(ignore_cache=False, tx_hashes=None)
-
-    # verify that the optimism transactions got decoded but not the
-    # ethereum one (would raise an error if tried)
-    hashes = dbl2withl1feestx.get_transaction_hashes_not_decoded(chain_id=ChainID.OPTIMISM, limit=None)  # noqa: E501
```
LefterisJP, line 232: why remove those? Were they not testing something?

rotkehlchen/chain/evm/decoding/decoder.py
yabirgb: Yes, not used anywhere else

rotkehlchen/tests/unit/test_evm_tx_decoding.py
yabirgb: I removed them becuase I also removed the functionality. The lines I removed as for what I understood all tested the removed functionality. They check that only the OP transactions get decoded when tx_hashes is None

LefterisJP (approved):
LGTM

PR #8546, opened 2024-09-13
Limit validators queried for balances without premium

### LINKED ISSUE #8528: ETH Staking Balance different in Dashboard than Blockchain Balances
## Problem Definition

I have more than 4 validators running, so without a premium membership the Blockchain Balances view shows me 4 of my validators:
image
This totals correctly to 32 ETH (I only own 25% of each validator => RocketPool).

But when I check the "Assets" section in the dashboard I can see that there it accounts for all 5 validators while I cannot change the ownership share of the 5th validator because of the license restriction.
So there it totals to 64 ETH.

While in "Blockchain Balances" only 32 ETH are used, the total in the dashboard uses 64 ETH.
image

So I suggest that the "Assets" section and the dashboard total also only use the balances of 4 validators and not all.

### System Description

Operating system: 
Rotki version: 1.34.3

nicholasyoder:
Closes #8528

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/eth2/beacon.py
+++ b/rotkehlchen/chain/ethereum/modules/eth2/beacon.py
@@ -24,7 +24,7 @@ from rotkehlchen.types import ChecksumEvmAddress, Eth2PubKey, Timestamp
 from rotkehlchen.utils.misc import from_gwei
 from rotkehlchen.utils.serialization import jsonloads_dict
 
-from .constants import BEACONCHAIN_MAX_EPOCH, DEFAULT_VALIDATOR_CHUNK_SIZE
+from .constants import BEACONCHAIN_MAX_EPOCH, DEFAULT_VALIDATOR_CHUNK_SIZE, FREE_VALIDATORS_LIMIT
 from .structures import ValidatorDailyStats, ValidatorDetails, ValidatorID
 from .utils import calculate_query_chunks, epoch_to_timestamp
 
@@ -146,6 +146,7 @@ class BeaconInquirer:
     def get_balances(
             self,
             indices_or_pubkeys: Sequence[int | Eth2PubKey],
+            has_premium: bool,
     ) -> dict[Eth2PubKey, Balance]:
         """Returns a mapping of validator public key to eth balance.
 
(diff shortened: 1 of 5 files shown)


rotkehlchen/chain/ethereum/modules/eth2/beacon.py
```diff
@@ -146,6 +147,7 @@ def set_rpc_endpoint(self, rpc_endpoint: str) -> None:
     def get_balances(
             self,
             indices_or_pubkeys: Sequence[int | Eth2PubKey],
+            has_premium: Premium | None,
```
LefterisJP, line 150:
```suggestion
            has_premium: bool
```

rotkehlchen/chain/ethereum/modules/eth2/eth2.py
```diff
@@ -122,7 +122,10 @@ def get_balances(
         if len(pubkey_to_ownership) == []:
             return {}  # nothing detected
 
-        balances = self.beacon_inquirer.get_balances(indices_or_pubkeys=list(pubkey_to_ownership.keys()))  # noqa: E501
+        balances = self.beacon_inquirer.get_balances(
+            indices_or_pubkeys=list(pubkey_to_ownership.keys()),
+            has_premium=self.premium,
```
LefterisJP, line 127:
```suggestion
            has_premium=self.premium is not None,
```

rotkehlchen/chain/ethereum/modules/eth2/eth2.py
```diff
@@ -247,6 +250,7 @@ def get_performance(
         if now - to_ts <= DAY_IN_SECONDS:
             balances = self.beacon_inquirer.get_balances(
                 indices_or_pubkeys=list(to_query_indices),
+                has_premium=self.premium,
```
LefterisJP, line 253:
```suggestion
                has_premium=self.premium is not None,
```

LefterisJP (commented):
Looks okay but needs a small change on the boolean check. Since you are just performing a boolean test, pass a boolean, not the entire premium object.

LefterisJP (approved):
LGTM! Thank you Nicholas.

PR #8547, opened 2024-09-13
Introduce exited_timestamp for each validator

LefterisJP:
This is the develop part of https://github.com/rotki/rotki/issues/8397


rotkehlchen/chain/ethereum/modules/eth2/beacon.py
```diff
@@ -234,14 +238,28 @@ def get_validator_data(
                 except DeserializationError:
                     log.error(f'Could not deserialize 0x01 withdrawal credentials for {entry}')
 
+            if queried_beaconchain:
+                if withdrawable_ts is not None and (exit_epoch := entry.get('exit_epoch', 0)) != 0:
+                    exited_ts = epoch_to_timestamp(exit_epoch)
+
+            elif withdrawable_ts is not None:  # query this index from beaconchain to see if exited
+                indices_mapping_to_query_beaconchain[entry[index_key]] = idx
```
yabirgb, line 246: can't you change the order and check first `withdrawable_ts`?

```suggestion
            if withdrawable_ts is not None:
                if  queried_beaconchain and (exit_epoch := entry.get('exit_epoch', 0)) != 0:
                    exited_ts = epoch_to_timestamp(exit_epoch)
                else:  # query this index from beaconchain to see if exited
                    indices_mapping_to_query_beaconchain[entry[index_key]] = idx
```

unless I'm mistaken is the same condition with one check less

rotkehlchen/db/upgrades/v43_v44.py
```diff
@@ -143,6 +150,8 @@ def upgrade_v43_to_v44(db: 'DBHandler', progress_handler: 'DBUpgradeProgressHand
         progress_handler.new_step()
         _add_uniswap_to_historical_oracles(write_cursor)
         progress_handler.new_step()
+        _add_exited_timestamp(write_cursor)
+        progress_handler.new_step()
```
yabirgb, line 154: you didn't increase the number of steps

rotkehlchen/db/upgrades/v43_v44.py
```diff
@@ -122,6 +122,12 @@ def _add_uniswap_to_historical_oracles(cursor: 'DBCursor') -> None:
     )
 
 
+@enter_exit_debug_log()
+def _add_exited_timestamp(cursor: 'DBCursor') -> None:
+    """Add exited_timestamp column to the eth2_validators table"""
+    cursor.execute('ALTER TABLE eth2_validators ADD COLUMN exited_timestamp INTEGER')
```
yabirgb, line 128: cuestion! Does sqlite assing NULL by default to an `INTEGER` nullable column? I would expect so but I'm not sure if it does something weid like giving a default value

rotkehlchen/db/eth2.py
```diff
@@ -185,38 +181,17 @@ def get_validators_with_status(
...
-        """
+        """Returns the indices of the tracked validators that we know have exite"""
         cursor.execute(
-            'SELECT validator_index FROM eth2_validators WHERE withdrawable_timestamp '
-            'IS NOT NULL AND ? - withdrawable_timestamp > ?',
-            (ts_now(), TEMP_MAX_WITHDRAWAL_WAIT),
+            'SELECT validator_index FROM eth2_validators WHERE exited timestamp IS NOT NULL',
         )
         result = {x[0] for x in cursor}
         return result
```
yabirgb, line 197:
```suggestion
            'SELECT validator_index FROM eth2_validators WHERE exited timestamp IS NOT NULL',
        )
       return {x[0] for x in cursor}
```

yabirgb: actually... it was revealed to my while praying that you forgot to do a `SELECT DISTINCT` in this query for better performance

rotkehlchen/db/upgrades/v43_v44.py
LefterisJP: it's NULL yes

rotkehlchen/db/eth2.py
LefterisJP: this query has an actual mistake ;)

There is quite a few things missing from the PR which I am also fixing.

But why would distinct make a difference? validator_index is always unique.

PR #8549, opened 2024-09-16
Forece WETH icon

yabirgb:
Since we have WETH and ETH in the same collection the icon of weth was being replaced by the ETH icon in the app. This PR adds an exception in the logic for this asset

image


LefterisJP (approved):
LGTM. We probably should have some other rule here though as this exception is a bit weird.

PR #8552, opened 2024-09-16
Eth2 cache and premium check

LefterisJP:
### [Introduce a has_premium_check function](https://github.com/rotki/rotki/commit/a0d6256224fd11f840d1443f25770c8debeb69f4) 

This is just to combine the premium is not None check with the active
premium check. In theory this way if a key expires and after few hours
the `Premium` object detects it in all the places where this is
checked they will switch to non premium.

While before this change even if you have a `Premium` object with
expired keys and `status = NOT_ACTIVE` it would count them as premium
being active

### DIFF
--- a/rotkehlchen/chain/aggregator.py
+++ b/rotkehlchen/chain/aggregator.py
@@ -342,6 +342,14 @@ class ChainsAggregator(CacheableMixIn, LockableQueryMixIn):
             if hasattr(module, 'premium') is True and module.premium is not None:  # type: ignore
                 module.premium = None  # type: ignore
 
+        # Also flush the cache of anything that is touched by eth2 validators since
+        # without premium we have a limit
+        self.flush_cache('query_eth2_balances')
+        self.flush_cache('query_balances')
+        self.flush_cache('query_balances', blockchain=SupportedBlockchain.ETHEREUM_BEACONCHAIN)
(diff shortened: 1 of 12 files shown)


rotkehlchen/chain/aggregator.py
```diff
@@ -342,6 +342,14 @@ def deactivate_premium_status(self) -> None:
             if hasattr(module, 'premium') is True and module.premium is not None:  # type: ignore
                 module.premium = None  # type: ignore
 
+        # Also flush the cache of anything that is touched by eth2 validators since
+        # without premium we have a limit
+        self.flush_cache('query_eth2_balances')
+        self.flush_cache('query_balances')
+        self.flush_cache('query_balances', blockchain=SupportedBlockchain.ETHEREUM_BEACONCHAIN)
+        self.flush_cache('query_balances', blockchain=None, ignore_cache=False)
+        self.flush_cache('query_balances', blockchain=SupportedBlockchain.ETHEREUM_BEACONCHAIN, ignore_cache=False)  # noqa: E501
```
yabirgb, line 351: is it needed to clear that last one? ignore cache should ignore the cache 😓

LefterisJP: we already do this in other places we clean the cache. You basically clean all possible argument combinations. It's dirty I know.

yabirgb (approved):
lgtm

PR #8553, opened 2024-09-16
Adjust logic for NFTs balances

### LINKED ISSUE #8535: NFTs counted twice?
## Problem Definition

Since some time I've noticed that many NFTs now show up both under "Assets" *and* in the separate NFT section. I'm not entirely sure what is happening here, but it looks like this is inflating the total asset balance.

### Example

Under NFTs:

CleanShot 2024-09-12 at 11 23 29@2x

Under Assets:

CleanShot 2024-09-12 at 11 23 06@2x

### System Description
Operating system: macOS 13.6.9
Rotki version: 1.34.3

yabirgb:
If a user added a NFT token to the list of assets it was getting queried as any other token and getting double counted by also the NFT module.

WHat I did in this PR is:

- Ignore a token from the balance query if it is also in the nfts tables. The idea is to allow people tracking NFTs in chains where we don't support Opensea
- In the snapshots ignore NFTs if they are already in the list of balances

Closes #8535

### DIFF
--- a/rotkehlchen/chain/evm/tokens.py
+++ b/rotkehlchen/chain/evm/tokens.py
@@ -4,14 +4,14 @@ from collections import defaultdict
 from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
-from rotkehlchen.assets.asset import EvmToken
+from rotkehlchen.assets.asset import EvmToken, Nft
 from rotkehlchen.chain.ethereum.utils import token_normalized_value
 from rotkehlchen.chain.evm.types import WeightedNode, asset_id_is_evm_token
 from rotkehlchen.fval import FVal
 from rotkehlchen.globaldb.handler import GlobalDBHandler
 from rotkehlchen.inquirer import Inquirer
 from rotkehlchen.logging import RotkehlchenLogsAdapter
-from rotkehlchen.types import ChecksumEvmAddress, Price, Timestamp
+from rotkehlchen.types import ChecksumEvmAddress, Price, SupportedBlockchain, Timestamp
 from rotkehlchen.utils.misc import combine_dicts, get_chunks
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


rotkehlchen/chain/evm/tokens.py
```diff
@@ -344,8 +344,19 @@ def query_tokens_for_addresses(
                 )
                 if saved_list is None:
                     continue  # Do not query if we know the address has no tokens
-                all_tokens.update(saved_list)
-                addresses_to_tokens[address] = saved_list
+
+                # get NFT tokens for address and ignores them from the query to avoid duplicates
```
LefterisJP, line 348:
```suggestion
                # get NFT tokens for address and ignore them from the query to avoid duplicates
```

rotkehlchen/chain/evm/tokens.py
```diff
@@ -344,8 +344,19 @@ def query_tokens_for_addresses(
...
+                cursor.execute(
+                    'SELECT identifier, blockchain FROM nfts WHERE owner_address=?',
+                    (address,),
+                )
+                excluded_addresses = {
+                    Nft(row[0]).evm_address for row in cursor
+                    if SupportedBlockchain.deserialize(row[1]) == self.evm_inquirer.blockchain
+                }
+                token_list = [x for x in saved_list if x.evm_address not in excluded_addresses]
```
LefterisJP, line 357: CAn't you somehow do this query in sqlite? I guess it would need substring manipulation which is gonna suck. Not sure if it's more or less performant..

rotkehlchen/rotkehlchen.py
```diff
@@ -1050,17 +1050,31 @@ def query_balances(
...
+                        # in the chain balances we update the price and continue
+                        blockchain_balances = balances[blockchain_location]
+                        nft = Nft(balance_entry['id'])
+                        nft_as_token = GlobalDBHandler.get_evm_token(
+                            address=nft.evm_address,
+                            chain_id=nft.chain_id,
+                        )  # we need the eip155 identifier instead of the _nft_ one.
+
+                        if nft_as_token in blockchain_balances:
+                            blockchain_balances[nft_as_token].usd_value = balance_entry['usd_price']  # noqa: E501
```
LefterisJP, line 1072: Why update the price here? They had no price before? What if previous price was more correct?

rotkehlchen/tests/unit/test_nfts.py
```diff
@@ -79,3 +82,61 @@ def test_sorting_nfts(blockchain: ChainsAggregator):
...
+@pytest.mark.parametrize('start_with_valid_premium', [True])
+@pytest.mark.parametrize('ethereum_modules', [['nfts']])
+def test_duplicate_balances(
+        rotkehlchen_api_server: APIServer,
+        ethereum_accounts,
+        gnosis_accounts,
+):
+    """Checks that we don't have duplicate balances for NFTs in balances.
+
+    If a NFT is tracked only as a token and not in the NFT table we query it
```
LefterisJP, line 99:
```suggestion
    If an NFT is tracked only as a token and not in the NFT table we query it
```

LefterisJP (approved):
LGTM

PR #8557, opened 2024-09-16
Address deprecation warning

yabirgb:
```
DeprecationWarning: connections() is deprecated and will be removed; use net_connections() instead
```


LefterisJP (approved):
LGTM if tests pass

PR #8558, opened 2024-09-17
Added a loss event type and a hack subtype

LefterisJP:
Moved spend/liquidate to loss/liquidate.

And created the loss/hack event type.

Will also need to add an accounting rule for this. And also modify the spend/liquidate rule to be for loss/liquidate


lukicenturi (approved):
frontend LGTM!

PR #8562, opened 2024-09-17
Use new WS message type for exchange unknown asset messages

### LINKED ISSUE #7817: Direct user to manage CEX mapping page, in "exchange unknown asset" notification
image

In this case we can add action to manage CEX mapping, so the token can be recognized.
But frontend doesn't know the type of ws message, because it's generic legacy warning message from backend

nicholasyoder:
Closes #7817

Backend changes only

### DIFF
--- a/docs/websockets.rst
+++ b/docs/websockets.rst
@@ -368,3 +368,27 @@ Whenever a protocol cache is being updated for protocols like Curve, Convex, Gea
 
 
 - ``data``: Contains the information of the progress of the process of updating the cache, that is: protocol, chain, processed, and total.
+
+
+Unknown asset on exchange
+============================
+
+If an unknown asset is encountered on an exchange we emit a message with the following format.
+
+::
+
+    {
+        "type": "exchange_unknown_asset",
+        "data": {
+            "location": "bybit",
+            "name": "Bybit 1",
+            "identifier": "ENA",
+            "details": "balance query",
+        }
+    }
+
+
+- ``location``: Exchange where the asset was found.
+- ``name``: Differentiates between multiple instances of the same location.
+- ``identifier``: Asset identifier of the unknown asset.
+- ``details``: Details about what type of event was being processed when the unknown asset was encountered.
--- a/rotkehlchen/exchanges/exchange.py
+++ b/rotkehlchen/exchanges/exchange.py
@@ -75,6 +75,7 @@ class ExchangeInterface(CacheableMixIn, LockableQueryMixIn):
             api_key: ApiKey,
             secret: ApiSecret,
             database: 'DBHandler',
+            msg_aggregator: 'MessagesAggregator',
     ):
         assert isinstance(api_key, T_ApiKey), (
             f'api key for {name} should be a string'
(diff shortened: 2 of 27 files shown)


rotkehlchen/exchanges/exchange.py
```diff
@@ -507,3 +509,19 @@ def query_history_with_callbacks(
...
+            self,
+            msg_aggregator: 'MessagesAggregator',
+            asset_identifier: str,
+            details: str,
+    ) -> None:
+        log.warning(f'Found unknown {self.name} asset {asset_identifier} in {details}.')
+        msg_aggregator.add_message(
+            message_type=WSMessageType.EXCHANGE_UNKNOWN_ASSET,
+            data={
+                'location': self.location.serialize(),
```
LefterisJP, line 523: Also add the name here and in the websocket and its api docs. What is the name? The name is the name the user can give to the specific api key instance.

What does that mean? It's useless in most cases as the name is default (such as kraken-1) but it can be used to differentiate between two different api keys for two different accounts on the same exchange. Say you have two coinbase accounts. 

```suggestion
                'location': self.location.serialize(),
                'name': self.name
```

rotkehlchen/exchanges/exchange.py
```diff
@@ -507,3 +509,19 @@ def query_history_with_callbacks(
...
         except RemoteError as e:
             fail_callback(str(e))
+
+    def send_unknown_asset_message(
+            self,
+            msg_aggregator: 'MessagesAggregator',
+            asset_identifier: str,
+            details: str,
+    ) -> None:
+        log.warning(f'Found unknown {self.name} asset {asset_identifier} in {details}.')
```
LefterisJP, line 519:
```suggestion
        log.warning(f'Found unknown {self.location.serialize()} {self.name} asset {asset_identifier} in {details}.')
```

rotkehlchen/exchanges/exchange.py
```diff
@@ -507,3 +509,19 @@ def query_history_with_callbacks(
             )
         except RemoteError as e:
             fail_callback(str(e))
+
+    def send_unknown_asset_message(
+            self,
+            msg_aggregator: 'MessagesAggregator',
```
LefterisJP, line 515: Why pass it as argument? Interesting I see that it's not part of the `exchange.py` common class. Why is that? Have you found any exchange it's not passed? Would it make sense to move it to the common class? That would simplify things, no?

You should probably do this in a separate commit as this is gonna be a bit bigger and more easily separable change.

rotkehlchen/exchanges/exchange.py
```diff
@@ -507,3 +509,19 @@ def query_history_with_callbacks(
...
+            asset_identifier: str,
+            details: str,
+    ) -> None:
```
LefterisJP, line 518: Add a docstring, explainign what the function does, what its arguments are and what errors if any it can raise.

docs/websockets.rst
```diff
@@ -367,4 +367,28 @@ Whenever a protocol cache is being updated for protocols like Curve, Convex, Gea
...
+            "location": "bybit",
+            "name": "Bybit 1",
+            "identifier": "ENA",
+            "details": "balance query",
+        }
+    }
+
+
+- ``location``: Exchange where the asset was found.
+- ``name``: Differentiates between multiple instances of the same location.
```
LefterisJP, line 392: nitpick

```suggestion
- ``name``: Differentiates between multiple instances of the same exchange.
```

LefterisJP (approved):
LGTM. Tiny wording nitpick which can be fixed later

PR #8566, opened 2024-09-18
Gnosis pay payments and api

LefterisJP:
This adds a hacky first integration with gnosis pay API to pull
merchant data and overlay it over gnosis pay transactions.

- No tests written yet for the api data. Should mock some things.
- Not querying them periodically at all. May need to do it and not
  just on demand.

Making this PR first now so the frontend can also do their part of the integration. Frontend should:

- [ ] Add a new external service API insertion only for premium users. This should be to add the `__Secure-authjs.session-token` for Gnosis Pay API as the "key" of the Gnosis PAY api. Much like monerium should have a warning that this is experimental. No risks warning, as it's read only for sure.
- [ ] Explain how to add it in a docs entry like we do with monerium. The way do do it is to login gnosis pay and then open the console and go to cookies that the app keeps and take the value of `__Secure-authjs.session-token`

- [ ] For all events that are spend/pay and counterparty `gnosis pay` have a special render rule to render the country flag which is part of the notes as: `:country:alpha2:`. For example for Germany `:country:DE:`.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -469,7 +469,7 @@ Getting or modifying external services API credentials
          "message":""
       }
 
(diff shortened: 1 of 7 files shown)


LefterisJP:
decoders test failed since I had not pushed the cassette. Done now. And hopefully if you rerun it will pass

PR #8569, opened 2024-09-18
Query offchain cowswap order data from API

### LINKED ISSUE #8428: Deal with the cowswap fee changes
## Abstract

Cowswap has changed how their fees behave. 

So our logic of checking the onchain event for fees is valid but only for older cowswap trades. For newer ones it seems the fees information is completely offchain and need api queries

API docs here: https://api.cow.fi/docs/#/

For full conversation follow here: 
https://discord.com/channels/869166959739170836/935460632818516048/1275219104831115335
https://discord.com/channels/869166959739170836/1275398510555955220/1275402907583250473

Market orders were deprecated after 2024-03-19 00:00Z when https://github.com/cowprotocol/infrastructure/pull/1288/files was merge. After that all order were limit orders and had feeAmount = 0 in the logs.

Prior to that timestamp if feeAmount = 0 then it's a limit order and we should have queried the API in order to get the fee data. If not it's a market order.
After that timestamp feeAmount is always going to be zero in the logs as all swaps became limit orders. So we would always need to query the API.

## Task

- Add a DB table that will hold the cowswap order offchain data
- Create an external api handler for CowswapAPI which will first check the local DB for order data and if not there ask the remote api, save the data in the DB and then return it.

For each cowswap swap:

- Get the orderId from the cowswap swap log and if feeAmount = 0, query the API.
- Determine if it's a market or limit order (by the condition described in the abstract) and also write it in the notes.
- Use the fee from the API also in the decoded event.
- Since all events will be redecoded in a release there is no need to go back and fix anything. The logic will rerun.

nicholasyoder:
Closes #8428


rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -228,21 +251,28 @@ def _detect_relevant_trades(
...
-                    event_type=HistoryEventType.TRADE,
-                    event_subtype=HistoryEventSubType.FEE,
-                    asset=swap_data.from_asset,
-                    balance=Balance(amount=swap_data.fee_amount),
-                    location_label=receive_event.location_label,
-                    notes=f'Spend {swap_data.fee_amount} {spend_event.asset.symbol_or_name()} as a cowswap fee',  # noqa: E501
-                    counterparty=CPT_COWSWAP,
-                    address=transaction.to_address,
+            if swap_data.fee_amount == ZERO:
+                raw_fee_amount, swap_data.order_type = self.cowswap_api.query_order_data(swap_data.order_uid)  # noqa: E501
```
LefterisJP, line 255: You are not handling the potential RemoteError here. This means that the entire decoding process will break down the moment the API does not respond for any reason. That's bad.You just have to catch it here and if there is an error just don't add a fee and not change order type. But log the error and state the problem in the logs.

rotkehlchen/externalapis/cowswap.py
```diff
@@ -0,0 +1,94 @@
...
+        self.session = requests.session()
+
+        if chain not in CHAIN_MAPPING:
+            log.warning(f'Encountered unsupported Cowswap API chain {chain}. Falling back to mainnet.')  # noqa: E501
+            chain = SupportedBlockchain.ETHEREUM
+
+        self.api_url = f'https://api.cow.fi/{CHAIN_MAPPING[chain]}/api/v1'
+
+    def _query(self, endpoint: str) -> dict[str, Any]:
+        """Query cowswap API endpoint."""
```
LefterisJP, line 44: you always need to specify which exceptions a function can raise in its docstring

rotkehlchen/externalapis/cowswap.py
```diff
@@ -0,0 +1,94 @@
...
+                    order_type = data['class']
+                else:
+                    raw_fee_amount = data['executedSurplusFee']
+                    order_type = json.loads(data['fullAppData'])['metadata']['orderClass']['orderClass']  # noqa: E501
+            except (KeyError, json.decoder.JSONDecodeError) as e:
+                raise RemoteError(f'Cowswap API response invalid: {data}') from e
+
+            with self.database.conn.write_ctx() as write_cursor:
+                write_cursor.execute(
+                    'INSERT INTO cowswap_orders(identifier, order_type, raw_fee_amount) VALUES (?, ?, ?)',  # noqa: E501
```
LefterisJP, line 90: So that you don't get a DB exception in case you somehow get to inputting the same trade

```suggestion
                    'INSERT OR IGNORE INTO cowswap_orders(identifier, order_type, raw_fee_amount) VALUES (?, ?, ?)',  # noqa: E501
```

rotkehlchen/externalapis/cowswap.py
```diff
@@ -0,0 +1,94 @@
...
+            raise RemoteError(f'Cowswap API request {response.url} failed with HTTP status code {response.status_code} and text {response.text}')  # noqa: E501
+
+        try:
+            json_ret = json.loads(response.text)
+        except json.decoder.JSONDecodeError as e:
+            raise RemoteError(f'Cowswap API returned invalid JSON response: {response.text}') from e  # noqa: E501
+
+        return json_ret
+
+    def query_order_data(self, order_uid: str) -> tuple[str, str]:
```
LefterisJP, line 63: Why return str for the amount?

```suggestion
    def query_order_data(self, order_uid: str) -> tuple[int, str]:
```

rotkehlchen/externalapis/cowswap.py
```diff
@@ -0,0 +1,94 @@
...
+            result = cursor.execute(
+                'SELECT order_type, raw_fee_amount FROM cowswap_orders WHERE identifier = ?',
+                (order_uid,),
+            ).fetchone()
+
+        if result is not None:
+            raw_fee_amount = result[0]
+            order_type = result[1]
+        else:
+            data = self._query(f'orders/{order_uid}')
```
LefterisJP, line 77: In such cases it's much easier to exit early **and** not have indented code.


```suggestion
        if result is not None:
            return int(result[0]), result[1]

        # else, we need to query the API
        data = self._query(f'orders/{order_uid}')
        ....
```

rotkehlchen/externalapis/cowswap.py
```diff
@@ -0,0 +1,94 @@
...
+            try:
+                if iso8601ts_to_timestamp(data['creationDate']) < ORDER_TYPE_CHANGE_TS:
+                    raw_fee_amount = data['executedFeeAmount']
+                    order_type = data['class']
+                else:
+                    raw_fee_amount = data['executedSurplusFee']
+                    order_type = json.loads(data['fullAppData'])['metadata']['orderClass']['orderClass']  # noqa: E501
```
LefterisJP, line 84:
```suggestion
                    raw_fee_amount = int(data['executedFeeAmount'])
                    order_type = data['class']
                else:
                    raw_fee_amount = int(data['executedSurplusFee'])
                    order_type = json.loads(data['fullAppData'])['metadata']['orderClass']['orderClass']  # noqa: E501
```

rotkehlchen/externalapis/cowswap.py
```diff
@@ -0,0 +1,94 @@
...
+            except (KeyError, json.decoder.JSONDecodeError) as e:
```
LefterisJP, line 85: you also need to handle the `ValueError` that doing `int(x)` will raise if `x` is not a stringified int
```suggestion
            except (KeyError, json.decoder.JSONDecodeError, ValueError) as e:
```

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2600,7 +2600,9 @@ def test_latest_upgrade_correctness(user_data_dir):
     assert tables_after_creation - tables_after_upgrade == set()
     assert views_after_creation - views_after_upgrade == set()
     new_tables = tables_after_upgrade - tables_before
-    assert new_tables == set()
+    assert new_tables == {
+        'cowswap_orders',
+    }
```
LefterisJP, line 2605: Why split something that easily fits in 1 line to 3 lines?
```suggestion
    assert new_tables == {'cowswap_orders'}
```

rotkehlchen/tests/unit/decoders/test_cowswap.py
```diff
@@ -50,7 +50,7 @@ def test_swap_token_to_token(ethereum_inquirer, ethereum_accounts):
             asset=A_WBTC,
             balance=Balance(amount=FVal(raw_amount)),
             location_label=user_address,
-            notes=f'Swap {raw_amount} WBTC in cowswap',
+            notes=f'Swap {raw_amount} WBTC in cowswap market order',
```
LefterisJP, line 53: Can you change the decoder to fix the wording here? Shouldn't it have an `a` here?
```suggestion
            notes=f'Swap {raw_amount} WBTC in a cowswap market order',
```

rotkehlchen/tests/unit/decoders/test_cowswap.py
```diff
@@ -63,7 +63,7 @@ def test_swap_token_to_token(ethereum_inquirer, ethereum_accounts):
             asset=Asset('eip155:1/erc20:0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48'),
             balance=Balance(amount=FVal('3800')),
             location_label=user_address,
-            notes='Receive 3800 USDC as the result of a swap in cowswap',
+            notes='Receive 3800 USDC as the result of a cowswap market order swap',
```
LefterisJP, line 66: Change the decoder to fix the wording here. Market order is enough, swap as wording is not needed

```suggestion
            notes='Receive 3800 USDC as the result of a cowswap market order',
```

rotkehlchen/tests/unit/decoders/test_cowswap.py
```diff
@@ -1012,8 +1038,21 @@ def test_gnosis_eure_v2(
         asset=Asset('eip155:100/erc20:0x420CA0f9B9b604cE0fd9C18EF134C705e5Fa3430'),
         balance=Balance(FVal(received_amount)),
         location_label=user_address,
-        notes=f'Receive {received_amount} EURe as the result of a swap in cowswap',
+        notes=f'Receive {received_amount} EURe as the result of a cowswap market order swap',
         tx_hash=tx_hash,
         counterparty=CPT_COWSWAP,
         address=string_to_evm_address('0x9008D19f58AAbD9eD0D60971565AA8510560ab41'),
+    ), EvmEvent(
```
LefterisJP, line 1045: I don't see any limit order tests. We need one. Don't we have any test after 19/03/2024? According to the issue after that everything is a limit order

nicholasyoder: I've added a limit order test, but note that actually the existing `test_swap_token_to_eth_with_other_trade` test is after 19/03/2024, but it is a market order (there are still market orders, the change was only in how they are executed internally).

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -228,6 +256,16 @@ def _detect_relevant_trades(
                 )
                 decoded_events.append(spend_event)
 
+            if swap_data.fee_amount == ZERO:
+                try:
+                    raw_fee_amount, swap_data.order_type = self.cowswap_api.get_order_data(swap_data.order_uid)  # noqa: E501
+                    swap_data.fee_amount = asset_normalized_value(
+                        amount=raw_fee_amount,
+                        asset=swap_data.from_asset.resolve_to_crypto_asset(),
+                    )
```
LefterisJP, line 265: Move this out of the try. This is not supposed to raise anything right?
```suggestion
                    swap_data.fee_amount = asset_normalized_value(
                        amount=raw_fee_amount,
                        asset=swap_data.from_asset.resolve_to_crypto_asset(),
                    )
```

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -228,6 +256,16 @@ def _detect_relevant_trades(
...
+                except RemoteError as e:
+                    log.error(f'Failed to get fee and order type from cowswap API for transaction {transaction} due to {e!s}')  # noqa: E501
```
LefterisJP, line 267:
```suggestion
                    log.error(f'Failed to get fee and order type from cowswap API for transaction {transaction} due to {e!s}. Will proceed without them.')  # noqa: E501
```

rotkehlchen/externalapis/cowswap.py
```diff
@@ -0,0 +1,103 @@
...
+                raw_fee_amount = int(data['executedSurplusFee'])
+                metadata = json.loads(data['fullAppData'])['metadata']
+                order_type = metadata['orderClass']['orderClass'] if 'orderClass' in metadata else 'limit'  # noqa: E501
+        except (KeyError, json.decoder.JSONDecodeError, ValueError) as e:
+            raise RemoteError(f'Cowswap API response invalid: {data}') from e
+
+        with self.database.conn.write_ctx() as write_cursor:
+            write_cursor.execute(
+                'INSERT OR IGNORE INTO cowswap_orders(identifier, order_type, raw_fee_amount) VALUES (?, ?, ?)',  # noqa: E501
+                (order_uid, order_type, str(raw_fee_amount)),
```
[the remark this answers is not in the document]
nicholasyoder: Added comment. See https://discord.com/channels/657906918408585217/1080783409640648765/1286377604407885945 for more details.

LefterisJP (approved):
LGTM. Thank you Nicholas

PR #8573, opened 2024-09-19
Fix re-querying of zksync lite transactions

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=79419328

Changes:
* Retrieve the latest transaction from the database correctly
* Remove use of used query range, and simply check if we were able to retrieve the latest transaction to determine whether to query all transactions or only new transactions.
* When querying new transactions, skip the first transaction if it matches the from_hash since it is already in the database

Note: A fix for the [nitpick from PR 8562](https://github.com/rotki/rotki/pull/8562#discussion_r1765271397) is included in a second commit.

### DIFF
--- a/rotkehlchen/chain/zksync_lite/manager.py
+++ b/rotkehlchen/chain/zksync_lite/manager.py
@@ -19,7 +19,6 @@ from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
 from rotkehlchen.constants import ZERO
 from rotkehlchen.constants.assets import A_ETH
 from rotkehlchen.db.history_events import DBHistoryEvents
-from rotkehlchen.db.ranges import DBQueryRanges
 from rotkehlchen.db.settings import CachedSettings
 from rotkehlchen.errors.asset import UnknownAsset
 from rotkehlchen.errors.misc import NotERC20Conformant, RemoteError
@@ -37,7 +36,6 @@ from rotkehlchen.types import (
     EVMTxHash,
     Fee,
     Location,
-    Timestamp,
     deserialize_evm_tx_hash,
 )
 from rotkehlchen.utils.misc import iso8601ts_to_timestamp, set_user_agent, ts_sec_to_ms
@@ -48,7 +46,7 @@ if TYPE_CHECKING:
     from rotkehlchen.db.dbhandler import DBHandler
     from rotkehlchen.db.drivers.gevent import DBCursor
 
-from .constants import ZKL_IDENTIFIER, ZKSYNCLITE_MAX_LIMIT, ZKSYNCLITE_TX_SAVEPREFIX
+from .constants import ZKL_IDENTIFIER, ZKSYNCLITE_MAX_LIMIT
 from .structures import ZKSyncLiteSwapData, ZKSyncLiteTransaction, ZKSyncLiteTXType
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


rotkehlchen/chain/zksync_lite/manager.py
```diff
@@ -165,26 +165,28 @@ def _query_api(
...
                 address=address,
                 from_hash=from_hash,
                 direction=direction,
         ):
+            if (
+                from_hash != 'latest' and
+                len(new_transactions) != 0 and
+                new_transactions[0].tx_hash.hex() == from_hash
+            ):
+                # Remove first transaction if it matches from_hash
```
LefterisJP, line 186: why would the first transaction match and already be in the DB? This seems like a pretty random thing to check for?

rotkehlchen/chain/zksync_lite/manager.py
```diff
@@ -199,15 +201,6 @@ def _query_and_save_transactions_for_range(
                     write_cursor=write_cursor,
                     transactions=unique_transactions,
                 )
-                if direction == 'older':
-                    current_start_ts = new_transactions[-1].timestamp
-                else:  # direction -> newer
-                    current_end_ts = new_transactions[-1].timestamp
-                ranges.update_used_query_range(
```
LefterisJP, line 206: Why remove this? Shouldn't something be saved here?

rotkehlchen/chain/zksync_lite/manager.py
```diff
@@ -598,6 +577,19 @@ def fetch_transactions(
...
             )
 
+        # Expand queried range to include new start and end timestamps
+        new_queried_range = (start_ts, end_ts) if queried_range is None else (
+            min(queried_range[0], start_ts),
+            max(queried_range[1], end_ts),
+        )
+        ranges = DBQueryRanges(self.database)
+        with self.database.conn.write_ctx() as write_cursor:
+            ranges.update_used_query_range(
```
LefterisJP, line 587: Why update it here? Where do you use this query range? Where is it going to be used?

rotkehlchen/chain/zksync_lite/manager.py
nicholasyoder: Added a comment in the code to explain this.

rotkehlchen/chain/zksync_lite/manager.py
nicholasyoder: Removed the used query range entirely as [discussed in discord](https://discord.com/channels/657906918408585217/1080783409640648765/1286788014533771274)

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2536,6 +2543,9 @@ def test_upgrade_db_43_to_44(user_data_dir, messages_aggregator):
         assert cursor.execute(
             'SELECT value FROM settings WHERE name="historical_price_oracles"',
         ).fetchone()[0] == '["manual", "cryptocompare", "coingecko", "defillama", "uniswapv3", "uniswapv2"]'  # noqa: E501
+        assert cursor.execute(
+            'SELECT * FROM used_query_ranges WHERE name LIKE "zksynclitetxs_%"',
+        ).fetchone() is None
```
LefterisJP, line 2548: nitpick: But for such queries you just want to count.So do this.
```suggestion
        assert cursor.execute(
            'SELECT COUNT(*) FROM used_query_ranges WHERE name LIKE "zksynclitetxs_%"',
        ).fetchone() == 0
```

LefterisJP (approved):
LGTM. Left a nitpick but will merge

PR #8589, opened 2024-09-23
Reset decoded events for DB upgrade and (maybe -- sort of) deal with gnosis pay refunds

LefterisJP:
### [Reset decoded events for 1.35](https://github.com/rotki/rotki/commit/99b4d8f6ef79b915579ebdcadc58359d5e1f5433)

As usual

### [Add untested gnosispay logic for refunds](https://github.com/rotki/rotki/commit/aa667083b8575aff4b60371e9794efb7dafd3c67) 

I am making some assumptions here.

1. There is no onchain logs for these. It's always a transfer from the
spender to the user's gnosis pay safe.
2. Not sure of how the final api response would look like for a refund. At the moment
in my data I have a refund with same tx_hash as payment but this is
probably inconmplete as the refund has not yet happened. I assume
there will be another transaction hash in the transactions entry.

### [Add a simple gnosis pay refund decoder logic](https://github.com/rotki/rotki/commit/b9a3cb01cc48f38edb9a47a3d413ef72fbd93249) 

This has no merchant data but just assumes any transfer from the
payment collector back to your address is a refund.

According to their devs this is the case.

Unfortunately they have no refund related logs


LefterisJP:
why did that single 

test fail .... ? That's why

yabirgb (approved):
lgtm

PR #8591, opened 2024-09-23
Fix oracles that required a database connection

yabirgb:
The oracles weren't removing the singleton and from season to season a reference to the ethereum_manager was kept but it had an old database connection closed after logout. This PR ensures that we clear the singletons during logout


LefterisJP (approved):
LGTM

PR #8594, opened 2024-09-23
Fix retrieving cowswap offchain data from DB

nicholasyoder:
Corrects the indexes used to retrieve the data in the database query response.


LefterisJP (approved):
LGTM

PR #8595, opened 2024-09-23
Fix pagination for assets

yabirgb:
We weren't considering the ignored assets handling logic when returning the total number of entries found. Since information is split between the userdb and the globaldb we need in some cases to iterate manually over the assets.

Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=80755755

The changes move the enum IgnoredAssetsHandling to prevent a circular import

### DIFF
--- a/rotkehlchen/globaldb/handler.py
+++ b/rotkehlchen/globaldb/handler.py
@@ -17,6 +17,7 @@ from rotkehlchen.assets.asset import (
     Nft,
     UnderlyingToken,
 )
+from rotkehlchen.assets.ignored_assets_handling import IgnoredAssetsHandling
 from rotkehlchen.assets.resolver import AssetResolver
 from rotkehlchen.assets.types import AssetData, AssetType
 from rotkehlchen.chain.evm.types import string_to_evm_address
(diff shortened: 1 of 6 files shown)


rotkehlchen/globaldb/handler.py
```diff
@@ -439,10 +440,27 @@ def retrieve_assets(userdb: 'DBHandler', filter_query: 'AssetsFilterQuery') -> t
...
+            # get `entries_found`. In the case of handling the ignored assets we need to manually
+            # count the assets found
             query, bindings = filter_query.prepare(with_pagination=False)
-            total_found_query = f'SELECT COUNT(*) FROM ({parent_query}) ' + query
-            entries_found = cursor.execute(total_found_query, bindings).fetchone()[0]
+            if filter_query.ignored_assets_handling == IgnoredAssetsHandling.NONE:
+                total_found_query = f'SELECT COUNT(*) FROM ({parent_query}) ' + query
+                entries_found = cursor.execute(total_found_query, bindings).fetchone()[0]
+            else:
+                total_found_query = f'SELECT identifier FROM ({parent_query}) ' + query
```
LefterisJP, line 450: Why do it like this? Why not just do modify the sqlite query?

rotkehlchen/globaldb/handler.py
```diff
@@ -439,10 +440,27 @@ def retrieve_assets(userdb: 'DBHandler', filter_query: 'AssetsFilterQuery') -> t
...
+                entries_found = 0
+                cursor.execute(total_found_query, bindings)
+                for (identifier,) in cursor:
+                    if (
+                        filter_query.ignored_assets_handling.value == 2 and
```
LefterisJP, line 455: This makes no sense.

1. filter_query is outside the loop which means it will be the same for each check. Move this outside the loop
2. `2`and `3` are not values that explain what this is. Use the name that explains what they are and serialize, or add a comment.

Same for the or part

rotkehlchen/tests/unit/test_assets.py
```diff
@@ -897,3 +897,33 @@ def test_all_assets_pagination(globaldb: 'GlobalDBHandler', database: 'DBHandler
     )
     assert page1[0] != page2[0]
     assert page1[0] + page2[0] == both_pages[0]
+
+    # test that we calculate correctly the entries found
```
LefterisJP, line 901:
```suggestion
    # test that we calculate the entries found correctly
```

rotkehlchen/globaldb/handler.py
yabirgb: We don't do it purely in sqlite because we need the ignored assets that are in the user db

LefterisJP (approved):
LGTM

PR #8597, opened 2024-09-23
Fix slow login by querying highest block in separate greenlet

nicholasyoder:


rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -533,6 +525,23 @@ def connect_to_multiple_nodes(self, nodes: Sequence[WeightedNode]) -> None:
...
+    def connect_to_multiple_nodes(self, nodes: Sequence[WeightedNode]) -> None:
+        self.web3_mapping = {}
+
+        # Remove etherscan nodes and return if all nodes use etherscan,
+        # so we don't query the highest block unnecessarily.
+        nodes = [node for node in nodes if node.node_info.name != self.etherscan_node_name]
+        if len(nodes) == 0:
+            return
+
+        self.greenlet_manager.spawn_and_track(
```
LefterisJP, line 537: Why do this like that? You are not doing what we asked. 

What was asked was to make the etherscan query for the highest block on its own greenlet. Not what you did.

PR #8601, opened 2024-09-24
fix: notes format when asset is multi-word

lukicenturi:
Fix issue where if the asset symbol is multi-word, then the amount in the notes won't be formatted, because we compare the asset, word 1 by 1, so it wouldn't match

### DIFF
--- a/frontend/app/src/composables/history/events/notes.ts
+++ b/frontend/app/src/composables/history/events/notes.ts
@@ -119,10 +119,22 @@ export function useHistoryEventNote(): UseHistoryEventsNoteReturn {
     if (get(scrambleData) && counterpartyVal === 'monerium')
       notesVal = findAndScrambleIBAN(notesVal);
 
-    // label each word from notes whether it is an address or not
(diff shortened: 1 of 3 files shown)


frontend/app/src/composables/history/events/notes.ts
```diff
@@ -119,10 +119,22 @@ export function useHistoryEventNote(): UseHistoryEventsNoteReturn {
     if (get(scrambleData) && counterpartyVal === 'monerium')
       notesVal = findAndScrambleIBAN(notesVal);
 
-    // label each word from notes whether it is an address or not
     const words = notesVal.split(/[\s,]+/);
+    const assetWords = asset ? asset.split(/\s+/) : [];
+    const processedWords: string[] = [];
+
+    for (let i = 0; i < words.length; i++) {
```
kelsos, line 126: Can we add a test for this @lukicenturi ? just to make sure we don't accidentally break it in the future?

kelsos (approved):
lgtm

PR #8606, opened 2024-09-24
Allow purging of cowswap and gnosispay data

LefterisJP:
Add them as potential module options only in the purge data api endpoint for a module.


yabirgb (approved):
lgtm

PR #8615, opened 2024-09-24
fix: replace note word spans with whitespace

kelsos:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/history/events/HistoryEventNote.vue
+++ b/frontend/app/src/components/history/events/HistoryEventNote.vue
@@ -64,7 +64,7 @@ function isLinkType(t: any): t is keyof ExplorerUrls {
         />
       </template>
       <template v-else-if="note.type === NoteType.WORD && note.word">
-        <span :key="`span-${index}`">{{ note.word }}</span>
+        {{ note.word }}
       </template>
       <HashLink
         v-else-if="note.showHashLink && isLinkType(note.type)"
--- a/frontend/app/src/composables/history/events/notes.ts
+++ b/frontend/app/src/composables/history/events/notes.ts
@@ -95,6 +95,35 @@ export function useHistoryEventNote(): UseHistoryEventsNoteReturn {
     return notes;
   }
 
+  /**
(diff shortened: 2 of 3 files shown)


frontend/app/src/components/history/events/HistoryEventNote.vue
```diff
@@ -64,7 +64,7 @@ function isLinkType(t: any): t is keyof ExplorerUrls {
         />
       </template>
       <template v-else-if="note.type === NoteType.WORD && note.word">
-        <span :key="`span-${index}`">{{ note.word }}</span>
+        {{ note.word }}
```
kelsos, line 67: This is to reduce the number of DOM elements since these spans don't have an actual reason to exist.

frontend/app/src/composables/history/events/notes.ts
```diff
@@ -278,7 +307,7 @@ export function useHistoryEventNote(): UseHistoryEventsNoteReturn {
       formats.push({ type: NoteType.WORD, word: splitted.join('') });
     });
 
-    return formats;
+    return formats.reduce(mergeSequentialWords, new Array<NoteFormat>());
```
kelsos, line 310: With the previous approach it seemed like each word was a new text node, with this unless there is something in between the words the should occupy a single one.

frontend/app/tests/unit/specs/composables/history/notes.spec.ts
```diff
@@ -314,27 +302,25 @@ describe('composables::history/notes', () => {
...
     store.update({ scrambleData: true });
     formatted = get(notesData);
     notesToString = formatted
       .filter(item => item.type === NoteType.WORD)
       .map(item => item.word)
       .join('');
-    included = notesToString.includes(iban.split(' ').join(''));
 
-    expect(included).toBeFalsy();
+    expect(notesToString).not.toContain(iban);
```
kelsos, line 323: I changed the test here so that if something goes wrong we have a more clear message.

PR #8623, opened 2024-09-25
Fix globaldb caches

yabirgb:
- Curve was using the wrong key to check for freshness of the cache
- Now query method of decoders also save to database and update last queried ts. This reduces
complexity in the flow and bugs
- Fixes a bug where if no new pool was detected in the protocols with cache the last_queried_ts
wasn't being updated. Related to the previous point

### DIFF
--- a/rotkehlchen/chain/evm/decoding/curve/curve_cache.py
+++ b/rotkehlchen/chain/evm/decoding/curve/curve_cache.py
@@ -33,6 +33,7 @@ from rotkehlchen.globaldb.cache import (
     globaldb_get_unique_cache_value,
     globaldb_set_general_cache_values,
     globaldb_set_unique_cache_value,
+    globaldb_update_cache_last_ts,
     read_curve_pool_tokens,
 )
 from rotkehlchen.globaldb.handler import GlobalDBHandler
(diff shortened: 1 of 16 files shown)


rotkehlchen/chain/evm/decoding/curve/curve_cache.py
```diff
@@ -489,10 +494,10 @@ def _query_curve_data_from_chain(
 
 def query_curve_data(
         inquirer: 'EvmNodeInquirer',
-        cache_type: Literal[CacheType.CURVE_POOL_ADDRESS],
+        cache_type: Literal[CacheType.CURVE_LP_TOKENS],
         msg_aggregator: 'MessagesAggregator',
 ) -> list[CurvePoolData] | None:
-    """Query curve lp tokens, curve pools and curve gauges.
+    """Query curve lp tokens, curve pools and curve gauges and saves them in the database.
```
LefterisJP, line 500:
```suggestion
    """Query curve lp tokens, curve pools and curve gauges and save them in the database.
```

rotkehlchen/chain/evm/decoding/curve/curve_cache.py
```diff
@@ -526,14 +531,27 @@ def query_curve_data(
             log.error(f'Could not query chain for curve pools due to: {err}')
             return None
 
-    if pools_data is None:
+    if len(pools_data) == 0:
+        # if no new pools update the last_queried_ts of db entries
```
LefterisJP, line 535:
```suggestion
        # if no new pools, update the last_queried_ts of db entries
```

LefterisJP (commented):
2 nitpicks. The rest of the logic I will have to trust you. I am not very much aware of it.

LefterisJP (approved):
LGTM

PR #8624, opened 2024-09-25
Separate SPAM from DECODED as transaction attributes

LefterisJP:
Also need to (in another commit) figure out the usages of those flags and at least rename them to avoid confusion.

They look like they are the same enum while one applies to history events and another to transactions

### DIFF
--- a/rotkehlchen/db/filtering.py
+++ b/rotkehlchen/db/filtering.py
@@ -18,6 +18,7 @@ from rotkehlchen.db.constants import (
     HISTORY_BASE_ENTRY_FIELDS,
     HISTORY_MAPPING_KEY_STATE,
     HISTORY_MAPPING_STATE_CUSTOMIZED,
+    HISTORY_MAPPING_STATE_DECODED,
 )
 from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.fval import FVal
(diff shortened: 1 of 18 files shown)


rotkehlchen/tests/unit/test_evm_tx_decoding.py
```diff
@@ -194,7 +195,7 @@ def test_tx_decode(ethereum_transaction_decoder, database):
         assert write_cursor.execute('SELECT COUNT(*) from evm_tx_mappings').fetchone()[0] == 0
 
 
-@pytest.mark.vcr(filter_query_parameters=['apikey'])
+# @pytest.mark.vcr(filter_query_parameters=['apikey'])
```
yabirgb, line 198: you left this uncommented

rotkehlchen/db/filtering.py
```diff
@@ -195,12 +196,15 @@ class DBTransactionsPendingDecodingFilter(DBFilter):
...
+    Due to the joining we need to check for either value being NULL (meaning no entry
+    in evm_tx_mappings) or not having the DECODED value.
     """
     chain_id: ChainID | None
 
     def prepare(self) -> tuple[list[str], list[Any]]:
-        query_filters = ['B.tx_id is NULL']
-        bindings: list[int] = []
+        query_filters = ['(B.tx_id IS NULL OR B.tx_id NOT IN (SELECT tx_id FROM evm_tx_mappings WHERE value=?))']  # noqa: E501
+        bindings: list[int] = [HISTORY_MAPPING_STATE_DECODED]
```
yabirgb, line 207: Can we also move this `HISTORY_MAPPING_STATE_DECODED` and rename it? Is not related to history events but transactions and is part of what lead to the confusion yesterday

rotkehlchen/tests/unit/test_evm_tx_decoding.py
LefterisJP: ARGH FFSZOR

rotkehlchen/db/filtering.py
LefterisJP: From the PR comments: 

>Also need to (in another commit) figure out the usages of those flags and at least rename them to avoid confusion.

yabirgb (approved):
Code changes look good.

About changing the value from 0 to 1 I'm not sure. What is the positive of having it start at 0 vs 1? If we change it to an enum the value starts at 1 with auto() but I don't have any preference of 1 vs 0

LefterisJP:
> About changing the value from 0 to 1 I'm not sure. What is the positive of having it start at 0 vs 1? If we change it to an enum the value starts at 1 with auto() but I don't have any preference of 1 vs 0

I said change 1 to 0 for that because atm seems like both evmt and historyevents use the same "enum" from the values. But perhaps the comments are enough

PR #8625, opened 2024-09-25
Allow add/delete via addressbook update endpoint

nicholasyoder:
Fixes problem with editing a label that wasn't previously set. [Reported in discord.](https://discord.com/channels/657906918408585217/1080783409640648765/1288371098278428706)

### DIFF
--- a/rotkehlchen/db/addressbook.py
+++ b/rotkehlchen/db/addressbook.py
@@ -110,11 +110,13 @@ class DBAddressbook:
             entries: list[AddressbookEntry],
     ) -> None:
         """Updates names of addressbook entries.
+        Add entry if it doesn't exist. Delete entry if the name is blank
         If blockchain is None then make sure that the same address doesn't appear in combination
         with other blockchain values.
         """
         with self.write_ctx(book_type) as write_cursor:
             for entry in entries:
+                entry_bockchain_value = entry.blockchain.value if entry.blockchain else ANY_BLOCKCHAIN_ADDRESSBOOK_VALUE  # noqa: E501
                 if (
                     entry.blockchain is None and
                     write_cursor.execute(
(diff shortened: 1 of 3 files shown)


rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -405,40 +405,57 @@ def test_update_addressbook(
...
+    with db_addressbook.read_ctx(book_type) as cursor:
+        entries_after_update = db_addressbook.get_addressbook_entries(cursor, filter_query=AddressbookFilterQuery.make())[0]  # noqa: E501
+    assert all(entry in entries_after_update for entry in new_entries)
+
+    # test that updating an entry with a blank label deletes the entry
+    entry_to_delete = AddressbookEntry(
+        address=ADDRESS_ETH,
+        name='',
+        blockchain=SupportedBlockchain.ETHEREUM,
+    )
```
yabirgb, line 445: this one is not sent in the request below... how did it work? I suspect the assertion that you have below doesn't work as you expect

rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -405,40 +405,57 @@ def test_update_addressbook(
...
-        assert db_addressbook.get_addressbook_entries(cursor, filter_query=AddressbookFilterQuery.make())[0] == entries_in_db_before_bad_patch  # noqa: E501
+    response = requests.patch(
+        api_url_for(
+            rotkehlchen_api_server,
+            'addressbookresource',
+            book_type=book_type,
+        ),
+        json={
+            'entries': [entry.serialize() for entry in new_entries],
+        },
```
yabirgb, line 433:
```suggestion
        json={'entries': [entry.serialize() for entry in new_entries]},
```

rotkehlchen/db/addressbook.py
```diff
@@ -128,9 +130,22 @@ def update_addressbook_entries(
...
                         continue  # skip update query, we just added a new entry instead
                     # InputError: single entry already exists, it must be updated using query below
+                elif (
+                    write_cursor.execute(
+                        'SELECT COUNT(*) FROM address_book WHERE address = ? AND blockchain = ?',
+                        (entry.address, entry_bockchain_value),
+                    ).fetchone()[0] == 0
+                ):
+                    # Add entry if it doesn't exist
+                    self.add_addressbook_entries(write_cursor, [entry])
```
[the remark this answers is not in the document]
nicholasyoder: It has just checked here that this entry doesn't exist in the database, so if the insert fails, something strange is happening, and it should probably just raise the InputError without trying to handle it in any special way, so the user sees an error.

rotkehlchen/tests/api/test_addressbook.py
nicholasyoder: I think it was working due to `entry_to_delete` having a blank name, so it didn't equal the event in the database. It should be updated to work correctly now.

rotkehlchen/tests/api/test_addressbook.py
```diff
@@ -405,40 +405,55 @@ def test_update_addressbook(
...
+        ),
+        json={'entries': [entry.serialize() for entry in new_entries]},
+    )
+    assert_proper_sync_response_with_result(response)
+    with db_addressbook.read_ctx(book_type) as cursor:
+        entries_after_update = db_addressbook.get_addressbook_entries(cursor, filter_query=AddressbookFilterQuery.make())[0]  # noqa: E501
+    assert all(entry in entries_after_update for entry in new_entries)
+
+    # test that updating an entry with a blank label deletes the entry
+
```
yabirgb, line 439: This should be removed

LefterisJP (approved):
LGTM

PR #8626, opened 2024-09-25
Raise error for etherscan max daily rate limit

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=81095856

### DIFF
--- a/rotkehlchen/externalapis/etherscan.py
+++ b/rotkehlchen/externalapis/etherscan.py
@@ -315,6 +315,8 @@ class Etherscan(ExternalServiceWithApiKey, ABC):
                     if status == 0:
                         if result == 'Contract source code not verified':
                             return None
+                        if 'Max daily rate limit reached' in result:
+                            raise RemoteError('Etherscan max daily rate limit reached.')
                         if 'rate limit reached' in result:
                             log.debug(
                                 f'Got response: {response.text} from {self.chain} etherscan.'
--- a/rotkehlchen/externalapis/etherscan.py
+++ b/rotkehlchen/externalapis/etherscan.py
@@ -258,8 +258,9 @@ class Etherscan(ExternalServiceWithApiKey, ABC):
 
         query_str += f'&apikey={api_key}'
         backoff = 1
-        backoff_limit = 33
-        timeout = timeout or CachedSettings().get_timeout_tuple()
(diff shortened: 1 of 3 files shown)


rotkehlchen/externalapis/etherscan.py
```diff
@@ -315,6 +315,8 @@ def _query(
                     if status == 0:
                         if result == 'Contract source code not verified':
                             return None
+                        if 'Max daily rate limit reached' in result:
+                            raise RemoteError('Etherscan max daily rate limit reached.')
```
LefterisJP, line 319: How does the normal rate limit reached message look like? Can you find out? As we would want to back off a bit for that. But right now `rate limit reached` is a substring of `Max daily rate limit reached` so if at any point someone moves things around here this will break and make it unreachable.

rotkehlchen/tests/external_apis/test_etherscan.py
```diff
@@ -82,19 +80,34 @@ def test_maximum_rate_limit_reached(temp_etherscan, **kwargs):  # pylint: disabl
 
     Regression test for https://github.com/rotki/rotki/issues/772"
     """
-    etherscan = temp_etherscan
-
-    etherscan_patch = patch_etherscan(etherscan)
+    etherscan_patch = patch_etherscan(
+        etherscan=temp_etherscan,
+        response_msg='Max rate limit reached, please use API Key for higher rate limit',
```
LefterisJP, line 85: is this the exact message etherscan currently sends?

LefterisJP (commented):
Two comments/questions

rotkehlchen/externalapis/etherscan.py
yabirgb: The normal rate limit message is 

```
{"status":"0","message":"NOTOK","result":"Max calls per sec rate limit reached (5/sec)"}
```

rotkehlchen/tests/external_apis/test_etherscan.py
yabirgb: Kelsos shared it yesterday

```
{"status":"0","message":"NOTOK","result":"Max daily rate limit reached. 110000 (100%) of 100000 day/limit"}**
```

PR #8641, opened 2024-09-27
Add extrafi claim for lending pools

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/evm/decoding/extrafi/cache.py
+++ b/rotkehlchen/chain/evm/decoding/extrafi/cache.py
@@ -1,0 +1,127 @@
+import logging
+from typing import TYPE_CHECKING, Literal
+
+from rotkehlchen.chain.evm.contracts import EvmContract
+from rotkehlchen.chain.evm.decoding.extrafi.constants import EXTRAFI_POOL_CONTRACT
+from rotkehlchen.chain.evm.types import string_to_evm_address
+from rotkehlchen.errors.misc import RemoteError
+from rotkehlchen.errors.serialization import DeserializationError
+from rotkehlchen.globaldb.cache import (
+    globaldb_get_general_cache_values,
+    globaldb_get_unique_cache_value,
+    globaldb_set_general_cache_values_at_ts,
+    globaldb_set_unique_cache_value_at_ts,
+)
+from rotkehlchen.globaldb.handler import GlobalDBHandler
+from rotkehlchen.logging import RotkehlchenLogsAdapter
+from rotkehlchen.serialization.deserialize import deserialize_evm_address
+from rotkehlchen.types import CacheType, ChainID, ChecksumEvmAddress
+from rotkehlchen.utils.misc import get_chunks, ts_now
+
+if TYPE_CHECKING:
+    from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
+    from rotkehlchen.user_messages import MessagesAggregator
+
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+MAX_RESERVES_QUERIED_PER_CHUNK = 20  # the maximum number of reserves that we can query without etherscan failing  # noqa: E501
+
+
+def get_existing_reward_pools(chain_id: ChainID) -> tuple[set[ChecksumEvmAddress]]:
+    """Get extrafi rewards pools for extrafi lending"""
+    with GlobalDBHandler().conn.read_ctx() as cursor:
+        return ({
+            string_to_evm_address(address) for address in globaldb_get_general_cache_values(
+                cursor=cursor,
+                key_parts=(CacheType.EXTRAFI_REWARD_CONTRACTS, str(chain_id.serialize())),
+            )
+        },)
+
+
+def query_extrafi_data(
+        inquirer: 'EvmNodeInquirer',
+        cache_type: Literal[CacheType.EXTRAFI_NEXT_RESERVE_ID],
+        msg_aggregator: 'MessagesAggregator',
+) -> list[ChecksumEvmAddress] | None:
+    """Query and store information of rewards pools.
+    Updates the last queried ts when executed.
+
+    May raise:
+    - RemoteError
+    """
+    lending_contract = EvmContract(
+        address=EXTRAFI_POOL_CONTRACT,
+        abi=inquirer.contracts.abi('EXTRAFI_LENDING'),
+        deployed_block=96265067,
+    )
+    cache_key = (cache_type, str(inquirer.chain_id.serialize_for_db()))
+    next_reserve_id = lending_contract.call(
+        node_inquirer=inquirer,
+        method_name='nextReserveId',
+    )
+
+    now = ts_now()
+    with GlobalDBHandler().conn.read_ctx() as cursor:
+        if (saved_next_reserve_id_str := globaldb_get_unique_cache_value(
+            cursor=cursor,
+            key_parts=cache_key,
+        )) is None:
+            saved_next_reserve_id = 1  # used later to start querying on reserve id 1
+        else:
+            saved_next_reserve_id = int(saved_next_reserve_id_str)
+
+        if next_reserve_id == saved_next_reserve_id:
+            with GlobalDBHandler().conn.write_ctx() as write_cursor:
+                globaldb_set_unique_cache_value_at_ts(
+                    write_cursor=write_cursor,
+                    key_parts=cache_key,
+                    value=str(next_reserve_id),
+                    timestamp=now,
+                )
+                return None
+
+        reward_addresses = []
+        for chunk in get_chunks(
+            lst=range(saved_next_reserve_id, next_reserve_id),
+            n=MAX_RESERVES_QUERIED_PER_CHUNK,
+        ):
+            calls = [(
+                EXTRAFI_POOL_CONTRACT,
+                lending_contract.encode(method_name='getStakingAddress', arguments=[idx]),
+            ) for idx in chunk]
+
+            try:
+                outputs = inquirer.multicall_2(require_success=False, calls=calls)
+            except RemoteError:
+                outputs = [(False, b'') for _ in chunk]
+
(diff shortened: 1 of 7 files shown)


rotkehlchen/chain/evm/decoding/extrafi/cache.py
```diff
@@ -0,0 +1,115 @@
...
+from rotkehlchen.types import CacheType, ChainID, ChecksumEvmAddress
+from rotkehlchen.utils.misc import ts_now
+
+if TYPE_CHECKING:
+    from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
+    from rotkehlchen.user_messages import MessagesAggregator
+
+
+def chunk_range(iterable: Sequence, size: int) -> Iterator:
+    """Split into chunks of size `size` the iterable provided"""
```
LefterisJP, line 27:
```suggestion
    """Split the provided iterable into chunks of size `size` """
```

rotkehlchen/chain/evm/decoding/extrafi/cache.py
```diff
@@ -0,0 +1,115 @@
...
+    it = iter(iterable)
+    return iter(lambda: tuple(islice(it, size)), ())
```
LefterisJP, line 29:
```suggestion
    return iter(lambda: tuple(islice(iter(iterable), size)), ())
```

rotkehlchen/chain/evm/decoding/extrafi/cache.py
```diff
@@ -0,0 +1,115 @@
...
+from rotkehlchen.globaldb.handler import GlobalDBHandler
+from rotkehlchen.types import CacheType, ChainID, ChecksumEvmAddress
+from rotkehlchen.utils.misc import ts_now
+
+if TYPE_CHECKING:
+    from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
+    from rotkehlchen.user_messages import MessagesAggregator
+
+
+def chunk_range(iterable: Sequence, size: int) -> Iterator:
```
LefterisJP, line 26: Question: Wouldn't `get_chunks` from utils/misc.py already work for this?

rotkehlchen/chain/evm/decoding/extrafi/cache.py
```diff
@@ -0,0 +1,115 @@
...
+    """Split into chunks of size `size` the iterable provided"""
+    it = iter(iterable)
+    return iter(lambda: tuple(islice(it, size)), ())
+
+
+def get_existing_reward_pools(chain_id: ChainID) -> tuple[set[ChecksumEvmAddress]]:
+    """Get extrafi rewards pools for extrafi lending"""
+    with GlobalDBHandler().conn.read_ctx() as cursor:
```
LefterisJP, line 34: pass the cursor to this function instead. At one place where you use it you initialize a cursor right before calling. That's the cases where you know you should pass the cursor as an argument.

rotkehlchen/chain/evm/decoding/extrafi/cache.py
```diff
@@ -0,0 +1,115 @@
...
+            )
+        },)
+
+
+def query_extrafi_data(
+        inquirer: 'EvmNodeInquirer',
+        cache_type: Literal[CacheType.EXTRAFI_NEXT_RESERVE_ID],
+        msg_aggregator: 'MessagesAggregator',
+) -> list[ChecksumEvmAddress] | None:
+    """Query and store information of rewards pools
```
LefterisJP, line 48:
```suggestion
    """Query and store information of rewards pools.
```

rotkehlchen/chain/evm/decoding/extrafi/cache.py
```diff
@@ -0,0 +1,115 @@
...
+    Updates the last queried ts when executed.
```
LefterisJP, line 49: Write what exceptions this can raise. As I see also contract calls

rotkehlchen/chain/evm/decoding/extrafi/cache.py
```diff
@@ -0,0 +1,115 @@
...
+                globaldb_set_unique_cache_value_at_ts(
+                    write_cursor=write_cursor,
+                    key_parts=cache_key,
+                    value=saved_next_reserve_id,
+                    timestamp=ts_now(),
+                )
+                return None
+
+        reward_addresses = []
+        for chunk in chunk_range(range(1, next_reserve_id), 20):
```
LefterisJP, line 78: use kwargs. Why 20 here? Why start from 1. Need a comment explaining.

rotkehlchen/chain/evm/decoding/extrafi/cache.py
```diff
@@ -0,0 +1,115 @@
...
+                outputs = inquirer.multicall_2(require_success=False, calls=calls)
+            except RemoteError:
+                outputs = [(False, b'') for _ in chunk]
+
+            for idx, output in enumerate(outputs):
+                status, encoded_data = output
+                if status is False:
+                    continue
+
+                reward_addresses.append(to_checksum_address(lending_contract.decode(
```
LefterisJP, line 94: jsut use deserialize_evm_Address. And catch the potential error which is `DeserializationError`

rotkehlchen/chain/evm/decoding/extrafi/cache.py
```diff
@@ -0,0 +1,115 @@
...
+                    result=encoded_data,
+                    method_name='getStakingAddress',
+                    arguments=[chunk[idx]],
```
LefterisJP, line 97: What if idx is out of range? Can that happen if any contract returns something wrong/weird or a remote call fails? If yes you need to guard for it, log an error and continue

rotkehlchen/chain/evm/decoding/extrafi/cache.py
```diff
@@ -0,0 +1,115 @@
...
+                )[0]))
+
+        now = ts_now()
```
LefterisJP, line 100: you already did a `ts_now()` in line 68 above. Same function. Turn that into a variable with walrus operator and use it below too

rotkehlchen/chain/evm/decoding/extrafi/decoder.py
```diff
@@ -400,6 +414,35 @@ def _handle_farm_events(self, context: DecoderContext) -> DecodingOutput:
             return self._handle_farm_repayment(context)
         return DEFAULT_DECODING_OUTPUT
 
+    def _decode_lending_claim(self, context: DecoderContext) -> DecodingOutput:
```
LefterisJP, line 417: Lending claim? Isn't it reward? so decode_lending_claim_reward or something?

rotkehlchen/chain/evm/decoding/extrafi/decoder.py
```diff
@@ -400,6 +414,35 @@ def _handle_farm_events(self, context: DecoderContext) -> DecodingOutput:
...
+
+        for event in context.decoded_events:
+            if (
+                event.event_type == HistoryEventType.RECEIVE and
+                event.asset == token and
+                token_normalized_value(token_amount=claimed, token=token) == event.balance.amount
+            ):
+                event.event_subtype = HistoryEventSubType.REWARD
+                event.counterparty = CPT_EXTRAFI
+                event.notes = f'Claim {event.balance.amount} {token.symbol_or_name()} from Extrafi lending'  # noqa: E501
```
LefterisJP, line 438: do we know the pool here?

rotkehlchen/tests/unit/test_protocol_balances.py
```diff
@@ -969,3 +977,46 @@ def test_extrafi_farm_balances(
             usd_value=FVal('10012.6978598973939729841870'),
         )},
     )}
+
+
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+def test_extrafi_cache(optimism_inquirer: 'OptimismInquirer'):
+    """Check that cache gets populated and timestamp update if
```
LefterisJP, line 984:
```suggestion
    """Check that the cache gets populated and timestamp updated if
```

rotkehlchen/tests/unit/test_protocol_balances.py
```diff
@@ -969,3 +977,46 @@ def test_extrafi_farm_balances(
...
+        last_queried_ts = globaldb_get_unique_cache_last_queried_ts_by_key(
+            cursor=cursor,
+            key_parts=(CacheType.EXTRAFI_NEXT_RESERVE_ID, chain),
+        )
+        assert should_update_protocol_cache(
+            cache_key=CacheType.EXTRAFI_NEXT_RESERVE_ID,
+            args=(chain,),
+        ) is False
+
+    time.sleep(1)  # give some time to avoid querying to fast and see the timestamp modified
```
LefterisJP, line 1012: 1. typo
2. Why? `time.sleep()` is generally a bad idea in tests and anywhere. Shouldn't you instead use freezetime and move the current timestamp? This should work too and not depend on randomness like time.sleep. Also is probably faster.

```suggestion
    time.sleep(1)  # give some time to avoid querying too fast and see the timestamp modified
```

LefterisJP (commented):
Lots of comments?

Also what exactly is the globalDB diff? Cache entries? Why?

rotkehlchen/chain/evm/decoding/extrafi/cache.py
yabirgb: It can't happen that multicall_2 returns False when the query fails and always the length of the output is the same as in the input. the result is tuples of true/false for the status and the actual output of the call

rotkehlchen/chain/evm/decoding/extrafi/cache.py
yabirgb: It is manually called only in the tests. In the actual code it is called at `reload_data` that doesn't reuse the cursor

yabirgb:
The DB diff is the cache used for decoding

LefterisJP (approved):
LGTM

PR #8643, opened 2024-09-27
Clean up temp dir after download

### LINKED ISSUE #1853: Find a way to cleanup the temporary folder after the user downloads the report.zip
## Problem Definition

Followup to PR #1848 

For the download of the CSV reports a temporary folder is created. Then the CSV reports are added into a temporary zip and the temporary zip is downloaded. The temporary zip is never cleanup in the context of the request/response because at this point the file is not yet sent to the client.

We need a way to clean up the temporary zip and remove the folder after the file has been served to the user.

#### For context:
- https://stackoverflow.com/questions/13344538/how-to-clean-up-temporary-file-used-with-send-file

nicholasyoder:
Closes #1853

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -13,7 +13,7 @@ from typing import TYPE_CHECKING, Any, Literal, Optional, cast, get_args, overlo
 from zipfile import BadZipFile, ZipFile
 
 import gevent
-from flask import Response, make_response, send_file
+from flask import Response, after_this_request, make_response, send_file
 from gevent.event import Event
 from gevent.lock import Semaphore
 from marshmallow.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


rotkehlchen/api/rest.py
```diff
@@ -377,6 +377,18 @@ def inner(rest_api: 'RestAPI', **kwargs: Any) -> Response:
     return wrapper
 
 
+def register_post_download_cleanup(temp_file: str | Path) -> None:
```
LefterisJP, line 380: you should alway pass a `Path` here. The caller knows best if it's a string or a path. Most places should be a Path. This way you remove the if

rotkehlchen/api/rest.py
```diff
@@ -377,6 +377,18 @@ def inner(rest_api: 'RestAPI', **kwargs: Any) -> Response:
...
+    @after_this_request
+    def do_cleanup(response: Response) -> Response:
+        try:
+            temp_path = Path(temp_file) if isinstance(temp_file, str) else temp_file
```
LefterisJP, line 384: This is can be removed

LefterisJP (approved):
LGTM. Failed test is failed in develop and will be fixed by yabir's PR.

PR #8650, opened 2024-09-30
Tweak external services api response

LefterisJP:
It will now always return all possible blockscout and etherscan api key values with null value if not existing, so that the frontend can use them to show what blockscout/etherscan keys exist


rotkehlchen/externalapis/blockscout.py
```diff
@@ -46,9 +46,10 @@ def __init__(
             msg_aggregator: MessagesAggregator,
     ) -> None:
         self.chain_id = blockchain.to_chain_id()
+        chainid_to_blockscout = {v: k for k, v in BLOCKSCOUT_TO_CHAINID.items()}
```
yabirgb, line 49: this is kept into memory and not used anymore right? Why not do it like

```suggestion
        service_name={v: k for k, v in BLOCKSCOUT_TO_CHAINID.items()}[self.chain_id],
```

PR #8652, opened 2024-09-30
Etherscan API query log when fail due to backoff

LefterisJP:


yabirgb (approved):
This PR brought such goodness to my life that now I can query on chain information using rotki. Much blessing such wow

LefterisJP:
> This PR brought such goodness to my life that now I can query on chain information using rotki. Much blessing such wow

lol wut? It's just nitpicks and a tiny fix :stuck_out_tongue:

rotkehlchen/externalapis/etherscan.py
```diff
@@ -353,6 +353,10 @@ def _query(
             # success, break out of the loop and return result
             return result
 
+        else:  # will only run if while condition fails # noqa: PLW0120
```
LefterisJP, line 356: Hmmm .... now that I think about it the noqa: PLW0120 maybe wrong. Isn't this equivalent to just not putting an else and having it after the loop?

PR #8657, opened 2024-10-01
Fixes for 1.35

yabirgb:
- Prevent None from propagating with network erros at etherscan
- Avoid marking eure as spam
- Add eigen and beigen to the db

### DIFF
--- a/rotkehlchen/externalapis/etherscan.py
+++ b/rotkehlchen/externalapis/etherscan.py
@@ -231,9 +231,12 @@ class Etherscan(ExternalServiceWithApiKey, ABC):
     ) -> list[dict[str, Any]] | (str | (list[EvmTransaction] | (dict[str, Any] | None))):
         """Queries etherscan
 
+        None is a valid result for this function when the requested information doesn't exist.
+        Happens when asking for the code of a contract, transaction by hash, receipt...
+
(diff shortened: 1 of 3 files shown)


rotkehlchen/externalapis/etherscan.py
```diff
@@ -323,6 +323,7 @@ def _query(
                                     f' Will backoff for {backoff} seconds.',
                                 )
                                 gevent.sleep(backoff)
+                                result = None  # reset result to it's None default value
```
LefterisJP, line 326: why do you need this?

rotkehlchen/externalapis/etherscan.py
```diff
@@ -233,7 +233,7 @@ def _query(
 
         May raise:
         - RemoteError if there are any problems with reaching Etherscan or if
-        an unexpected response is returned
+        an unexpected response is returned. Also the case of exhausting the backoff time.
```
LefterisJP, line 236:
```suggestion
        an unexpected response is returned. Also in the case of exhausting the backoff time.
```

LefterisJP (commented):
GlobalDB diff is for what we just merged?

LefterisJP (approved):
LGTM

PR #8660, opened 2024-10-01
Reload decoder data on demand during event decoding

LefterisJP:
If a new event adds information that any other decoder needs to use then this system will allow it to happen.

Probably need to use it in more places other than just the eigenlayer eigenpod deployments

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,7 +2,7 @@
 Changelog
 =========
 
-* :release:`1.35.0 <2024-10-01>`
+* :release:`1.35.0 <2024-10-02>`
 * :feature:`8428` Rotki will now properly decode cowswap fees and order types after 2024-03-19 by querying the cowswap API for offchain data.
 * :feature:`7817` Users will be able to add CEX mapping for unknown assets from exchange notifications.
 * :feature:`8528` Rotki will now limit the number of validators queried for balances for non-premium users.
--- a/rotkehlchen/chain/evm/decoding/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/decoder.py
@@ -326,17 +326,34 @@ class EVMTransactionDecoder(ABC):
 
         return possible_products
 
+    def _reload_single_decoder(self, cursor: 'DBCursor', decoder: 'DecoderInterface') -> None:
+        """Reload data for a single decoder"""
+        if isinstance(decoder, CustomizableDateMixin):
+            decoder.reload_settings(cursor)
+        if isinstance(decoder, ReloadableDecoderMixin):
+            new_mappings = decoder.reload_data()
+            if new_mappings is not None:
+                self.rules.address_mappings.update(new_mappings)
+
     def reload_data(self, cursor: 'DBCursor') -> None:
         """Reload all related settings from DB and data that any decoder may require from the chain
-        so that decoding happens with latest data"""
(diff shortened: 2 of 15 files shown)


rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -326,11 +326,14 @@ def get_decoders_products(self) -> dict[str, list[EvmProduct]]:
...
 
-    def reload_data(self, cursor: 'DBCursor') -> None:
+    def reload_data(self, cursor: 'DBCursor', reload_decoders: set[str] | None = None) -> None:
         """Reload all related settings from DB and data that any decoder may require from the chain
         so that decoding happens with latest data"""
         self.base.refresh_tracked_accounts(cursor)
-        for decoder in self.decoders.values():
+        for decoder_name, decoder in self.decoders.items():
+            if reload_decoders and decoder_name not in reload_decoders:
+                continue
```
yabirgb, line 335: maybe something smarter to avoid iterating over all of them would be better like iterate over the set and do a get in self.decoders and then reload those only. Otherwise you have a lot of potential decoders and you basically continue in the loop

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -326,11 +326,14 @@ def get_decoders_products(self) -> dict[str, list[EvmProduct]]:
 
         return possible_products
 
-    def reload_data(self, cursor: 'DBCursor') -> None:
+    def reload_data(self, cursor: 'DBCursor', reload_decoders: set[str] | None = None) -> None:
```
yabirgb, line 329: update docstring?

rotkehlchen/chain/evm/decoding/structures.py
```diff
@@ -76,11 +76,13 @@ class DecodingOutput:
     and is used in post-decoding rules like in the case of balancer
     - refresh_balances may be set to True if the user's on-chain balances in some protocols has
     changed (for example if the user has deposited / withdrawn funds from a curve gauge).
-    """
+    - reload_decoders can be None in which case nothing happens. Or a set of decoders names for which to reload data
+    """  # noqa: E501
```
yabirgb, line 80: the set is of the class name right? This would be good info to state here. I see myself using the counterparty instead

docs/changelog.rst
```diff
@@ -2,7 +2,7 @@
 Changelog
 =========
 
-* :release:`1.35.0 <2024-10-01>`
+* :release:`1.35.0 <2024-10-02>`
```
yabirgb, line 5: imagen

rotkehlchen/chain/evm/decoding/decoder.py
LefterisJP: yeah but this way I can reuse the same function and this happens very rarely. But yes this is an improvement to do

docs/changelog.rst
LefterisJP: hehe yeah not happening today. Multiple open things

rotkehlchen/chain/evm/decoding/structures.py
LefterisJP: will do yeah

yabirgb (approved):

PR #8670, opened 2024-10-02
Fix query of exit ts of eth validators

yabirgb:
If the user had its own beacon node and we couldn't obtain the exit timestamp we make a query to beaconchain but at the moment of processing the results we were using the wrong key for the validator index because it was being set previously for the beacon node.

Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/eth2/beacon.py
+++ b/rotkehlchen/chain/ethereum/modules/eth2/beacon.py
@@ -197,7 +197,8 @@ class BeaconInquirer:
           deserialized due to unexpected format
         """
         # Beaconcha.in only keys
-        index_key = 'validatorindex'
+        beacon_chain_index_key = 'validatorindex'
+        index_key = beacon_chain_index_key
         valuegetter = operator.getitem
         withdrawal_credentials_key = 'withdrawalcredentials'
         activation_epoch_key = 'activationepoch'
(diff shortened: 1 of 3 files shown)


LefterisJP (commented):
So the traceback is

```
[02/10/2024 19:15:59 W. Europe Summer Time] ERROR rotkehlchen.greenlets.manager Greenlet with id 1436866503168: Periodically detect withdrawal exits died with exception: 'index'.
Exception Name: <class 'KeyError'>
Exception Info: 'index'
Traceback:
   File "src\\gevent\\greenlet.py", line 908, in gevent._gevent_cgreenlet.Greenlet.run
  File "rotkehlchen\chain\ethereum\modules\eth2\eth2.py", line 623, in detect_exited_validators
  File "rotkehlchen\chain\ethereum\modules\eth2\beacon.py", line 264, in get_validator_data

```

Any way to unit test that path? Just to see this does not happen. Just this single function with these conditions? As a future-proofing error

rotkehlchen/chain/ethereum/modules/eth2/beacon.py
```diff
@@ -246,7 +247,7 @@ def get_validator_data(
                 if queried_beaconchain and (exit_epoch := entry.get('exitepoch', 0)) != 0:
                     exited_ts = epoch_to_timestamp(exit_epoch)
                 else:  # query this index from beaconchain to see if exited
-                    indices_mapping_to_query_beaconchain[entry[index_key]] = idx
+                    indices_mapping_to_query_beaconchain[deserialize_int(entry[index_key])] = idx
```
LefterisJP, line 250: Yes I think this is required here too. Notice that in line 253 this is also done.

In line 264 it's not done (I assume by accident) but it's fine since it's beaconchain website that's queried only there

LefterisJP (approved):
LGTM

PR #8672, opened 2024-10-02
fix: remove property that was causing manual balance price double conversion

### LINKED ISSUE #8669: Manual Balances Table should treat price as already in user selected currency.
## Problem Definition

`usdPrice` property is already converted in the user selected currency but the `AmountInput` component in the ManualBalancesTable treats it as a value in `USD` and displays it after performing a double conversion of the value.

## TODO

Change the `<AmountInput/>` that displays the usdPrice in the `ManualbalancesTable.vue` so that it does not perform a conversion and shows the value as it is, since it's already in the target currecy

### TLDR

Removing the fiat-currency="USD" should display the proper value:

https://github.com/rotki/rotki/blob/e986ea1cebee2a91c634a68d4ccace7073d26855/frontend/app/src/components/accounts/manual-balances/ManualBalanceTable.vue#L235

To confirm compare with dashboard as it should show the proper value

kelsos:
Closes #8669

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,7 +2,9 @@
 Changelog
 =========
 
+
 * :bug:`-` Fix an error querying the exit timestamp for the ethereum validators.
+* :bug:`8669` Fixes double conversion for displayed price in manual balances when not using USD as the selected currency.
 
 * :release:`1.35.0 <2024-10-02>`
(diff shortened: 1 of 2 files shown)


LefterisJP:
rebased on bugfixes and pushed

LefterisJP (approved):
LGTM. Tested it and it works

PR #8679, opened 2024-10-04
fix: double conversion in edit snapshot form

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/dashboard/edit-snapshot/EditBalancesSnapshotAssetPriceForm.vue
+++ b/frontend/app/src/components/dashboard/edit-snapshot/EditBalancesSnapshotAssetPriceForm.vue
@@ -1,0 +1,309 @@
+<script setup lang="ts">
+import { helpers, required } from '@vuelidate/validators';
+import useVuelidate from '@vuelidate/core';
+import { toMessages } from '@/utils/validation';
+import { CURRENCY_USD } from '@/types/currencies';
+import { TaskType } from '@/types/task-type';
+import type { BigNumber } from '@rotki/common';
+import type { HistoricalPriceFormPayload } from '@/types/prices';
+
+const amount = defineModel<string>('amount', { required: true });
+
+const usdValue = defineModel<string>('usdValue', { required: true });
+
+const asset = defineModel<string>('asset', { required: false });
+
+// eslint-disable-next-line vue/define-macros-order
+const props = withDefaults(
+  defineProps<{
+    timestamp: number;
+    disableAsset?: boolean;
+    nft?: boolean;
+  }>(),
+  {
+    disableAsset: false,
+    nft: false,
+  },
+);
+
+const { timestamp, disableAsset } = toRefs(props);
+
+const { t } = useI18n();
+
+const fiatValue = ref<string>('');
+const assetToUsdPrice = ref<string>('');
+const assetToFiatPrice = ref<string>('');
+
+const fiatValueFocused = ref<boolean>(false);
+const fetchedAssetToUsdPrice = ref<string>('');
+const fetchedAssetToFiatPrice = ref<string>('');
+
+const { resetHistoricalPricesData } = useHistoricCachePriceStore();
+const { currencySymbol } = storeToRefs(useGeneralSettingsStore());
+const { isTaskRunning } = useTaskStore();
+
+const { getHistoricPrice } = useBalancePricesStore();
+const { addHistoricalPrice } = useAssetPricesApi();
+
+const isCurrentCurrencyUsd = computed<boolean>(() => get(currencySymbol) === CURRENCY_USD);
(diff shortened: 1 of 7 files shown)


frontend/app/src/components/dashboard/edit-snapshot/EditBalancesSnapshotAssetPriceForm.vue
```diff
@@ -0,0 +1,313 @@
...
+import type { HistoricalPriceFormPayload } from '@/types/prices';
+import type { NewHistoryEventPayload } from '@/types/history/events';
+import type { ActionStatus } from '@/types/action';
+
+const props = withDefaults(
+  defineProps<{
+    timestamp: number;
+    asset: string;
+    disableAsset?: boolean;
+    v$: Validation;
```
kelsos, line 17: Is this the proper way to do this by passing the instance?

There is this entry that talks about [nested validations](https://vuelidate-next.netlify.app/advanced_usage.html#nested-validations) that doesn't seem to pass the instance. Can you have a look to see if this is possible?

frontend/app/src/components/dashboard/edit-snapshot/EditBalancesSnapshotAssetPriceForm.vue
```diff
@@ -0,0 +1,313 @@
...
+  get() {
+    return get(asset);
+  },
+  set(asset: string) {
+    if (get(disableAsset))
+      return;
+
+    emit('update:asset', asset);
+  },
+});
```
kelsos, line 42: two questions here:

1. is there any reason to not use `defineModel` for asset
2. would it make sense to extract the logic around assetModel to a util function (I tried searching for something but I couldn't find anything existing.)? Also is there truly a case where it would update the value when asset is disabled?

kelsos (changes requested):
I am not sure if this is related to the PR changes or not, but I was trying to edit a snapshot where the asset price was `0`, and after editing the price it was still `0`

Who is supposed to save the historic price there? us or the backend?

Also a few smaller comments can we make sure that the `EditBalancesSnapshotAssetPriceForm.vue` component is properly structured as per the contribution guide?

And secondarily a small nitpick is that it would be nice if we can structure the defines there in the order from `develop` to reduce the merge friction.

frontend/app/src/components/dashboard/edit-snapshot/EditBalancesSnapshotAssetPriceForm.vue
lukicenturi: Im just need the error message from the `v$`.

lukicenturi:
> And secondarily a small nitpick is that it would be nice if we can structure the defines there in the order from `develop` to reduce the merge friction.
I dont think I'm 100% sure what you ask here. 😅

kelsos:
> > And secondarily a small nitpick is that it would be nice if we can structure the defines there in the order from `develop` to reduce the merge friction.
> 
> I dont think I'm 100% sure what you ask here. 😅

https://eslint.vuejs.org/rules/define-macros-order.html

https://github.com/rotki/eslint-config/blob/6b7d5330e9dda5a0445c5ca1bd0b1223cb22e03e/src/configs/vue.ts#L208

and the rule by default also adds a whitespace in between each `define`.

lukicenturi:
> > > And secondarily a small nitpick is that it would be nice if we can structure the defines there in the order from `develop` to reduce the merge friction.
> > 
> > 
> > I dont think I'm 100% sure what you ask here. 😅
> 
> [eslint.vuejs.org/rules/define-macros-order.html](https://eslint.vuejs.org/rules/define-macros-order.html)
> 
> [rotki/eslint-config@`6b7d533`/src/configs/vue.ts#L208](https://github.com/rotki/eslint-config/blob/6b7d5330e9dda5a0445c5ca1bd0b1223cb22e03e/src/configs/vue.ts#L208)
> 
> and the rule by default also adds a whitespace in between each `define`.

Apparently for these files, the order didn't change at all (compared to develop), so I think we're fine

frontend/app/src/components/dashboard/edit-snapshot/EditBalancesSnapshotAssetPriceForm.vue
kelsos: Yes this happens because the validation for this sub-form is external to the component. My question is, do we have any specific reason that blocks us from moving the relevant parts of the validation to the component itself instead of passing it externally?

Do we gain something by passing the validation instead of specifying it inside the component?

lukicenturi: I see, yes I think this can be moved to the children instead, let me move it

kelsos (approved):
lgtm thank you @lukicenturi

PR #8682, opened 2024-10-05
Query logs for also vested thegraph accounts

LefterisJP:
For a vesting account the l2Delegator of the event will not be a user owned account, but will be the vesting contract.

Example here:
https://etherscan.io/tx/0x48321bb00e5c5b67f080991864606dbc493051d20712735a579d7ae31eca3d78#eventlog for rotki

This was seen before, as we had a log query that saved all transactions on the staking contract. But now that we properly limited the query to use args again we need to also query for DelegationTransferredToL2 with delegator being the vesting locked wallet.

The good thing is we can find this by the approval query as it's the to address of the approvals that are tagged with graph protocol.


LefterisJP:
should have made this on bugfixes. Cherry picked it and pushed it there too

PR #8687, opened 2024-10-07
feat: redetect token button for particular chains

lukicenturi:
Closes #(issue_number)

## Checklist


kelsos:
OK so in general this looks great.

We have only two small changes to do and we are good to merge.
1. We should dismiss the dialog when pressing the `Re-detect` button
2. We should reset the selection (so that individual re-detect entries are visible` and disable until it finishes.

kelsos (approved):
lgtm thank you @lukicenturi

PR #8697, opened 2024-10-08
feat: redecode page button

cleanerzkp:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/history/events/HistoryEventsTable.vue
+++ b/frontend/app/src/components/history/events/HistoryEventsTable.vue
@@ -25,6 +25,7 @@ const emit = defineEmits<{
   'show:form': [payload: ShowEventHistoryForm];
   'set-page': [page: number];
   'refresh': [data?: EvmChainAndTxHash];
+  'current-page-events': [events: HistoryEventEntry[]];
 }>();
 
 const { groups, groupLoading } = toRefs(props);
@@ -233,6 +234,10 @@ function forceRedecode(): void {
   set(deleteCustom, false);
   set(redecodePayload, undefined);
 }
+
+watch(() => props.groups.data, (newRows) => {
+  emit('current-page-events', newRows);
+}, { immediate: true });
 </script>
 
 <template>
--- a/frontend/app/src/components/history/events/HistoryEventsTable.vue
+++ b/frontend/app/src/components/history/events/HistoryEventsTable.vue
@@ -25,7 +25,6 @@ const emit = defineEmits<{
   'show:form': [payload: ShowEventHistoryForm];
   'set-page': [page: number];
   'refresh': [data?: EvmChainAndTxHash];
-  'current-page-events': [events: HistoryEventEntry[]];
 }>();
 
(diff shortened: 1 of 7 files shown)


frontend/app/src/components/history/events/HistoryEventsTable.vue
```diff
@@ -233,6 +234,10 @@ function forceRedecode(): void {
   set(deleteCustom, false);
   set(redecodePayload, undefined);
 }
+
+watch(() => props.groups.data, (newRows) => {
+  emit('current-page-events', newRows);
+}, { immediate: true });
```
lukicenturi, line 240: Why do you need to emit it from here, while the data is passed from parent, meaning you can get it directly in parent

frontend/app/src/components/history/events/HistoryEventsView.vue
```diff
@@ -330,6 +331,16 @@ function onShowDialog(type: 'decode' | 'protocol-refresh'): void {
     set(protocolCacheStatusDialogOpen, true);
 }
 
+async function redecodePageTransactions(): Promise<void> {
+  const evmEvents = currentPageEvents.value.filter(isEvmEvent);
+  const { pullAndRedecodeTransaction, fetchUndecodedTransactionsStatus } = useHistoryTransactionDecoding();
+
+  for (const event of evmEvents) {
+    await pullAndRedecodeTransaction(toEvmChainAndTxHash(event));
+  }
```
lukicenturi, line 340: with this, the endpoint will be called many times.
Instead, make `pullAndRedecodeTransaction` accept array, (because the endpoint accept array too), so we don't need to call this many times

frontend/app/src/components/history/events/HistoryEventsView.vue
```diff
@@ -330,6 +331,16 @@ function onShowDialog(type: 'decode' | 'protocol-refresh'): void {
     set(protocolCacheStatusDialogOpen, true);
 }
 
+async function redecodePageTransactions(): Promise<void> {
+  const evmEvents = currentPageEvents.value.filter(isEvmEvent);
```
lukicenturi, line 335: use `get()`

frontend/app/src/components/history/events/HistoryEventsView.vue
```diff
@@ -330,6 +331,16 @@ function onShowDialog(type: 'decode' | 'protocol-refresh'): void {
...
+  const { pullAndRedecodeTransaction, fetchUndecodedTransactionsStatus } = useHistoryTransactionDecoding();
```
lukicenturi, line 336: never call composable inside function like this, as this will be called everytime the function called. Move it outside

lukicenturi: Also this is redundant, we have called it outside

lukicenturi (approved):
@cleanerzkp I fix and merge it, please check out the comment so this kind of mistakes wont happen again

PR #8699, opened 2024-10-09
fix: make pinned section persistent for better UX

lukicenturi:
image

Make the pinned section persistent, so user can see history events and the pinned section simultaneously


kelsos (approved):
lgtm @lukicenturi

I rebased it on top of develop and re-installed dependencies/build to revert some accidental changes in the auto-import file.

PR #8701, opened 2024-10-09
Save userDB backups in user directory and not temp

LefterisJP:


rotkehlchen/db/upgrade_manager.py
```diff
@@ -212,46 +211,38 @@ def _perform_single_upgrade(
...
-                self.db.set_setting(
-                    write_cursor=write_cursor,
-                    name='ongoing_upgrade_from_version',
-                    value=upgrade.from_version,
-                )
+        tmp_db_filename = f'{ts_now()}_rotkehlchen_db_v{upgrade.from_version}.backup'
+        shutil.copyfile(
+            os.path.join(self.db.user_data_dir, USERDB_NAME),
+            os.path.join(self.db.user_data_dir, tmp_db_filename),
+        )
```
yabirgb, line 218: one thing that I don't know if it happens before creating the backup... do we consume the wal file before copying the db?

LefterisJP: Good questions. Not sure. What can we do to make sure it is? 

We either need to do this or copy/backup all files. Whcih imo is senseless. Same question for the globalDB upgrade too. Since you dealt with the WAL can you find out?

Since this is at th every very start of the app I guess it's fine? But if there is multipel upgrades one after the other then we need to be more careful?

LefterisJP: done

PR #8702, opened 2024-10-09
Handle yearn api query errors better

LefterisJP:
The function that queried and merged data from yearn API was pretty bad. The bug was that it just continued with the response as an unbound variable if the api query failed.

But also error handling and reporting was bad.
Added error logging everywhere and also refactored to exit early for better readability


yabirgb (approved):
lgtm

PR #8719, opened 2024-10-11
Remove spam logic from transactions

yabirgb:
- Remove spam augmented logic
- Remove spam decoded flag from transactions
- Remove logic to avoid decoding spam transactions Since we can have false positives in the db ensure that all transactions get decoded


rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -522,16 +463,6 @@ def _decode_transaction(
         with self.database.conn.read_ctx() as read_cursor:
             tx_id = transaction.get_or_query_db_id(read_cursor)
 
-        if self._is_spam_airdrop(transaction, tx_receipt):
```
LefterisJP, line 525: This we may need to re-introduce right? Once our spam tokens are more solid?

rotkehlchen/data_migrations/migrations/migrations_18.py
```diff
@@ -168,6 +169,19 @@ def whitelist_monerium_assets(rotki: 'Rotkehlchen') -> None:
             )
 
 
+@enter_exit_debug_log()
+def remove_spam_detection_on_transactions(rotki: 'Rotkehlchen') -> None:
+    """We will remove any marked transaction as spam from the database so they can be decoded
+    in case something went wrong with the autodetection of spam.
```
LefterisJP, line 175: add a link to the issue which explain the problem

LefterisJP (approved):
LGTM

PR #8724, opened 2024-10-11
Yearn: Skip retrieving vault inception and add reloading of vaults

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=82966110

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/yearn/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/yearn/decoder.py
@@ -10,7 +10,7 @@ from rotkehlchen.chain.ethereum.modules.yearn.constants import (
     YEARN_LABEL_V2,
 )
 from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
-from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface
+from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface, ReloadableDecoderMixin
 from rotkehlchen.chain.evm.decoding.structures import (
     FAILED_ENRICHMENT_OUTPUT,
     EnricherContext,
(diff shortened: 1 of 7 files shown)


rotkehlchen/chain/ethereum/modules/yearn/decoder.py
```diff
@@ -58,14 +62,18 @@ def __init__(
             base_tools=base_tools,
             msg_aggregator=msg_aggregator,
         )
+        self.vaults = EMPTY_VAULTS
 
+    def reload_data(self) -> None:
+        """Reload vaults from DB."""
+        self.vaults = EMPTY_VAULTS
```
yabirgb, line 69: have you checked if this works? I'm under the impression that the one in the init

```
        self.vaults = EMPTY_VAULTS
```

will create a reference to `EMPTY_VAULTS` and when you populate it you will be actually modifying `EMPTY_VAULTS` so when you do this you will still have everythin in memory.

It's not a problem because it is a set and you are doing .add to the set. In fact I believe here you don't need to reset so in the init method you can do

```
self.vaults = {
    YEARN_VAULTS_V1_PROTOCOL: set(),
    YEARN_VAULTS_V2_PROTOCOL: set(),
}
```

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
```diff
@@ -58,14 +65,33 @@ def __init__(
...
-            self.vaults_v1 = {string_to_evm_address(row[0]) for row in cursor}
-
-            cursor.execute(db_query, (YEARN_VAULTS_V2_PROTOCOL, ChainID.ETHEREUM.serialize_for_db()))  # noqa: E501
-            self.vaults_v2 = {string_to_evm_address(row[0]) for row in cursor}
+            bindings = (
+                YEARN_VAULTS_V1_PROTOCOL,
+                YEARN_VAULTS_V2_PROTOCOL,
+                ChainID.ETHEREUM.serialize_for_db(),
+            )
+            query_body = 'FROM evm_tokens WHERE protocol IN (?, ?) AND chain=?'
```
LefterisJP, line 86: first query, then bindings. This is how it's all over the code

```suggestion
           query_body = 'FROM evm_tokens WHERE protocol IN (?, ?) AND chain=?'
            bindings = (
                YEARN_VAULTS_V1_PROTOCOL,
                YEARN_VAULTS_V2_PROTOCOL,
                ChainID.ETHEREUM.serialize_for_db(),
            )
```

LefterisJP (approved):
LGTM

PR #8731, opened 2024-10-14
EURe == EUR price for current and historical

LefterisJP:


rotkehlchen/inquirer.py
```diff
@@ -643,6 +644,16 @@ def _find_price(
...
             return price, CurrentPriceOracle.MANUALCURRENT, used_main_currency
 
+        if GlobalDBHandler.asset_in_collection(collection_id=240, asset_id=from_asset.identifier):  # part of the EURe collection # noqa: E501  # todo: Super hacky. Figure out a way to generalize
+            return Inquirer._find_price(
+                from_asset=A_EUR,
+                to_asset=to_asset,
+                ignore_cache=ignore_cache,
+                skip_onchain=skip_onchain,
+                coming_from_latest_price=coming_from_latest_price,
+                match_main_currency=match_main_currency,
```
lukicenturi, line 654: If `fromAsset` = `EURe`, and currentSelectedCurrency = `EUR`, The `match_main_currency` should be `true` in order for frontend to use it directly without double conversion.

lukicenturi: In this part, I believe `match_main_currency` still `False` here, because on the caller function, `EURe` !== `EUR`

Issue #842, opened 2020-04-05
Enable strict typing for all tests


LefterisJP:
## Problem Definition

For backend tests we have an exception in setup.cfg here:
https://github.com/rotki/rotki/blob/b07063d35164f59b438a7385b3c72955a90e71d9/setup.cfg#L28-L32

We don't have strict typing check for all tests.

## Task

- Remove the lines shown above
- Properly type everything in tests so that linting passes again

yashnirmal:
I am working on it

yashnirmal:
Hi @LefterisJP can you check the PR once

LefterisJP:
> Hi @LefterisJP can you check the PR once

As @yabirgb mentioned in the PR this is not the solution. One would need to actually type all tests properly

yashnirmal:
Guys, can you help me with how I can do that?
Currently I am getting this error
![Screenshot from 2022-10-20 11-48-49](https://user-images.githubusercontent.com/78169683/196871371-31839f00-dc8e-4a59-a278-274cef87908b.png)

yabirgb:
Hello @yashnirmal ! Did you modify the code in develop? If not make sure to do `pip install -r requirements.dev` to install the latest version of the linting packages. In develop you shouldn't get any error

yashnirmal:
No. This error was also coming before I made any changes to the code.

yabirgb:
@yashnirmal Then make sure that the version of flake8 is the same as in the `requirements_lint.txt` file.

```
(rotki-main) ➜  rotki git:(develop) ✗ pip freeze | grep flake8
flake8==5.0.4
flake8-bugbear==22.9.23
flake8-commas==2.1.0
flake8-comprehensions==3.10.0
flake8-debugger==4.1.2
flake8-executable==2.1.2
flake8-mutable==1.2.0
flake8-tuple==0.4.1
```

AlexandrSozin:
Hi! I’d like to incrementally enable strict typing for a few small unit-test modules, starting with test_etherscan.py. Is this scope welcome?

AlexandrSozin:
Hi! I’d like to incrementally enable strict typing for a few small unit-test modules, starting with test_etherscan.py. Is this scope welcome?

PR #8738, opened 2024-10-14
Fix typing in tests/api/test_a*

nicholasyoder:
Related: #842

### DIFF
--- a/rotkehlchen/tests/api/test_accounting_rules.py
+++ b/rotkehlchen/tests/api/test_accounting_rules.py
@@ -3,7 +3,7 @@ import random
 from http import HTTPStatus
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from typing import Literal, get_args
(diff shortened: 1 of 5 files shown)


rotkehlchen/tests/api/test_accounting_rules.py
```diff
@@ -218,9 +218,9 @@ def test_manage_rules(rotkehlchen_api_server, db_settings):
     ]
 
     # update rule 2 to a combination that already exists
-    rule_2['event_type'] = rule_1['event_type']
-    rule_2['event_subtype'] = rule_1['event_subtype']
-    rule_2['counterparty'] = rule_1['counterparty']
+    rule_2['event_type'] = rule_1['event_type']  # type: ignore
```
LefterisJP, line 221: why are you using type ignore around here and above? This should only be used in the absence of any other solution

LefterisJP (approved):
LGTM

PR #8739, opened 2024-10-14
improvements to timeouts while decoding

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/ethereum/transactions.py
+++ b/rotkehlchen/chain/ethereum/transactions.py
@@ -21,6 +21,7 @@ from rotkehlchen.types import ChecksumEvmAddress, Location, deserialize_evm_tx_h
 from rotkehlchen.utils.misc import ts_now
 
 if TYPE_CHECKING:
+    from rotkehlchen.chain.gnosis.transactions import GnosisWithdrawalsQueryParameters
     from rotkehlchen.db.dbhandler import DBHandler
 
     from .node_inquirer import EthereumInquirer
@@ -42,6 +43,8 @@ class EthereumTransactions(EvmTransactions):
             self,
             last_block_queried: int,
             filters: dict[str, Any],
+            new_events: list[dict[str, Any]],  # pylint: disable=unused-argument
+            cb_arguments: 'GnosisWithdrawalsQueryParameters | None',  # pylint: disable=unused-argument
     ) -> None:
         """Callback used in query_for_graph_delegation_txns when querying logs.
         Ensures that when iterating the logs we keep the progress of the queried
--- a/rotkehlchen/chain/ethereum/types.py
+++ b/rotkehlchen/chain/ethereum/types.py
@@ -1,4 +1,7 @@
-from typing import TYPE_CHECKING, Any, Literal, Protocol
+from typing import TYPE_CHECKING, Any, Literal, Protocol, TypeVar
+
+if TYPE_CHECKING:
+    from rotkehlchen.chain.gnosis.transactions import GnosisWithdrawalsQueryParameters
 
 if TYPE_CHECKING:
     # This is a list of all ethereum abi that are guaranteed to be returned
@@ -33,9 +36,27 @@ if TYPE_CHECKING:
     ]
 
 
-class LogIterationCallback(Protocol):
+T_contra = TypeVar('T_contra', bound='GnosisWithdrawalsQueryParameters | None', contravariant=True)
+
+
+class LogIterationCallback(Protocol[T_contra]):
     """Callback executed when querying onchain log events after finishing a range
     of blocks
     """
-    def __call__(self, last_block_queried: int, filters: dict[str, Any]) -> None:
-        ...
+    def __call__(
+            self,
+            last_block_queried: int,
+            filters: dict[str, Any],
+            new_events: list[dict[str, Any]],
+            cb_arguments: T_contra,
+    ) -> None:
+        """
+        - last_block_queried: number of the last block queried
+        - filters: filters used when querying log events
+        - new_events: New events found in the current iteration of the log search
+        - cb_arguments: An instance of a dataclass containing additional context and mutable state
+        for the callback. This object facilitates bidirectional communication, allowing the
+        callback to both receive and update query-related information. For example, in
+        GnosisWithdrawalsQueryParameters, it can be used to track and update the last queried block
+        across multiple callback invocations.
+        """
--- a/rotkehlchen/chain/evm/decoding/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/decoder.py
@@ -10,6 +10,7 @@ from dataclasses import dataclass
 from types import ModuleType
 from typing import TYPE_CHECKING, Any, Optional, Protocol
 
+import gevent
 from gevent.lock import Semaphore
 
 from rotkehlchen.accounting.structures.balance import Balance
@@ -85,6 +86,7 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
+MIN_LOGS_PROCESSED_TO_SLEEP = 100
 
 
 class EventDecoderFunction(Protocol):
@@ -477,7 +479,7 @@ class EVMTransactionDecoder(ABC):
         monerium_special_handling_event = False
 
         # decode transaction logs from the receipt
-        for tx_log in tx_receipt.logs:
+        for idx, tx_log in enumerate(tx_receipt.logs):
             if (
                 monerium_special_handling_event is False and
                 self.evm_inquirer.chain_id in {ChainID.GNOSIS, ChainID.POLYGON_POS} and
@@ -489,6 +491,10 @@ class EVMTransactionDecoder(ABC):
                 # We use this flag to avoid iterating twice over the events searching
                 # for legacy transfers.
                 monerium_special_handling_event = True
+
+            if idx + 1 % MIN_LOGS_PROCESSED_TO_SLEEP == 0:
+                log.debug(f'Context switching out of the {idx + 1} log event of {transaction}')
+                gevent.sleep(0)
 
             context = DecoderContext(
                 tx_log=tx_log,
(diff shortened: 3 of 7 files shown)


rotkehlchen/api/rest.py
```diff
@@ -3036,9 +3036,15 @@ def decode_evm_transactions(
             chain_manager.transactions.get_receipts_for_transactions_missing_them()
             amount_of_tx_to_decode = dbevmtx.count_hashes_not_decoded(chain_id=evm_chain)
             if amount_of_tx_to_decode > 0:
-                chain_manager.transactions_decoder.get_and_decode_undecoded_transactions(
-                    send_ws_notifications=True,
-                )
+                # disallow scheduling of new tasks while decoding transactions
+                self.rotkehlchen.task_manager.should_schedule = False  # type: ignore  # task manager is not none here
```
LefterisJP, line 3040: But this is not enough here. This is only if you ask for it directly. In every upgrade and migration this is done from the periodic task. It needs to happen there too. So better abstract it differently

rotkehlchen/chain/gnosis/transactions.py
```diff
@@ -70,7 +70,7 @@ def get_chain_specific_multiaddress_data(
             call_order=self.evm_inquirer.default_call_order(),
         )
 
-        expected_topics = ['0x000000000000000000000000' + x.lower()[2:] for x in addresses]
+        expected_topics = {'0x000000000000000000000000' + x.lower()[2:] for x in addresses}
```
LefterisJP, line 73: didn't you say you would move the logic around here into the callback?

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -490,6 +491,9 @@ def _decode_transaction(
                 # for legacy transfers.
                 monerium_special_handling_event = True
 
+            if idx + 1 % 100 == 0:
```
LefterisJP, line 494: make this into a constant and define it somewhere in the decoder constants explaining what it does.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -490,6 +491,9 @@ def _decode_transaction(
...
+                gevent.sleep(0)
```
LefterisJP, line 495: pseudocode for log

```suggestion
                log.debug(f'Context switching out of the {idx + 1} log event of {tx_hash}')
                gevent.sleep(0)
```

rotkehlchen/chain/ethereum/types.py
```diff
@@ -37,5 +37,10 @@ class LogIterationCallback(Protocol):
     """Callback executed when querying onchain log events after finishing a range
     of blocks
     """
-    def __call__(self, last_block_queried: int, filters: dict[str, Any]) -> None:
+    def __call__(
```
LefterisJP, line 40: would need to explain all arguments in the docstring

rotkehlchen/chain/gnosis/transactions.py
```diff
@@ -59,32 +61,57 @@ def get_chain_specific_multiaddress_data(
...
+                last_block_queried: int,
+                filters: dict[str, Any],  # pylint: disable=unused-argument
+                new_events: list[dict[str, Any]],
+        ) -> None:
+            """Callback that processes new events
+            This function also keeps track of the last block number processed using
+            the variable nonlocal variable last_block_processed. It is later used
+            in case of error to save the queried range.
+            """
+            nonlocal last_block_processed
```
LefterisJP, line 77: no this is ugly and super error prone. What you made is a closure, but I don't think it's required in this case, plus capturing the surrounding scope with nonlocal is less performant. Why implement a closure when you can just add add optional arguments to the callback function?

It's a bit more performatn and less ugly

I do notice you modify the last_block_processed so you can also do that with optional argument but it would need to be an object you modify I suppsoe.

rotkehlchen/chain/gnosis/transactions.py
```diff
@@ -59,32 +61,57 @@ def get_chain_specific_multiaddress_data(
...
-                )
+        last_block_processed = None
+        expected_topics = {'0x000000000000000000000000' + x.lower()[2:] for x in addresses}
+
+        def _process_events_cb(
+                last_block_queried: int,
+                filters: dict[str, Any],  # pylint: disable=unused-argument
+                new_events: list[dict[str, Any]],
+        ) -> None:
+            """Callback that processes new events
```
LefterisJP, line 72:
```suggestion
            """Callback that processes new events.
            
            Here also expain what this particular callback is suppsoed to do. I don't totally get it what id does for gnosis.
```

rotkehlchen/chain/ethereum/transactions.py
```diff
@@ -42,6 +43,8 @@ def _graph_delegation_callback(
             self,
             last_block_queried: int,
             filters: dict[str, Any],
+            new_events: list[dict[str, Any]],  # pylint: disable=unused-argument
+            cb_arguments: 'LOG_ITERATION_CALLBACK_ARGUMENTS_TYPE',  # pylint: disable=unused-argument
```
LefterisJP, line 47: don't do it like this. Please have it explicit here. `'WithdrawalsQueryParameters | None`. Otherwise I have to jump around to understand this is optional. Using yet another "alias" makes it harder to read and understand

rotkehlchen/chain/ethereum/types.py
```diff
@@ -33,9 +36,28 @@
...
+            filters: dict[str, Any],
+            new_events: list[dict[str, Any]],
+            cb_arguments: LOG_ITERATION_CALLBACK_ARGUMENTS_TYPE,
+    ) -> None:
+        """
+        - last_block_queried: number of the last block queried
+        - filters: filters used when querying log events
+        - new_events: New events found in the current iteration of the search
+        - cb_arguments: An instance of a dataclass (one of the types defined in
+        LOG_ITERATION_CALLBACK_ARGUMENTS_TYPE) containing additional context and mutable state
```
LefterisJP, line 58: also modify the variable name here

rotkehlchen/chain/ethereum/types.py
```diff
@@ -33,9 +36,28 @@
...
+            self,
+            last_block_queried: int,
+            filters: dict[str, Any],
+            new_events: list[dict[str, Any]],
+            cb_arguments: LOG_ITERATION_CALLBACK_ARGUMENTS_TYPE,
+    ) -> None:
+        """
+        - last_block_queried: number of the last block queried
+        - filters: filters used when querying log events
+        - new_events: New events found in the current iteration of the search
```
LefterisJP, line 56:
```suggestion
        - new_events: New events found in the current iteration of the log search
```

rotkehlchen/chain/gnosis/transactions.py
```diff
@@ -11,13 +14,45 @@
...
 if TYPE_CHECKING:
+    from rotkehlchen.chain.ethereum.types import LOG_ITERATION_CALLBACK_ARGUMENTS_TYPE
     from rotkehlchen.db.dbhandler import DBHandler
 
     from .node_inquirer import GnosisInquirer
 
 ADDED_RECEIVER_ABI: Final = [{'anonymous': False, 'inputs': [{'indexed': False, 'name': 'amount', 'type': 'uint256'}, {'indexed': True, 'name': 'receiver', 'type': 'address'}, {'indexed': True, 'name': 'bridge', 'type': 'address'}], 'name': 'AddedReceiver', 'type': 'event'}]  # noqa: E501
 DEPLOYED_BLOCK: Final = 9053325
 DEPLOYED_TS: Final = 1539027985
+NO_BLOCK_PROCESSED_VALUE = 0
```
LefterisJP, line 25: 0 is a valid block value

```suggestion
NO_BLOCK_PROCESSED_VALUE: Final = -1
```

rotkehlchen/chain/gnosis/transactions.py
```diff
@@ -11,13 +14,45 @@
...
+
+def _process_withdrawals_events_cb(
+        last_block_queried: int,
+        filters: dict[str, Any],  # pylint: disable=unused-argument
+        new_events: list[dict[str, Any]],
+        cb_arguments: 'LOG_ITERATION_CALLBACK_ARGUMENTS_TYPE',
+) -> None:
+    """Callback that processes new events
+    This function also keeps track of the last block number processed using
+    the variable nonlocal variable last_block_processed. It is later used
```
LefterisJP, line 43: Adjust docstring. This is no longer that hacky closure

rotkehlchen/chain/gnosis/transactions.py
```diff
@@ -11,13 +14,45 @@
...
+    last_block_processed: int
+    expected_topics: set[str]
+    gnosis_transactions: 'GnosisTransactions'
+
+
+def _process_withdrawals_events_cb(
+        last_block_queried: int,
+        filters: dict[str, Any],  # pylint: disable=unused-argument
+        new_events: list[dict[str, Any]],
+        cb_arguments: 'LOG_ITERATION_CALLBACK_ARGUMENTS_TYPE',
```
LefterisJP, line 39: This callback is actually **never** called with `None`. So you need just the `WithdrawalsQueryParameters`. Also the assertion below needs to go away

rotkehlchen/chain/gnosis/transactions.py
```diff
@@ -11,13 +14,45 @@
...
+) -> None:
+    """Callback that processes new events
+    This function also keeps track of the last block number processed using
+    the variable nonlocal variable last_block_processed. It is later used
+    in case of error to save the queried range.
+    """
+    assert cb_arguments is not None
+    cb_arguments.last_block_processed = last_block_queried
```
LefterisJP, line 47: Shouldn't this move at the end of the function? Since only then that's when the block is processed.

rotkehlchen/chain/gnosis/transactions.py
yabirgb: It is called as None in the case of thegraph

LefterisJP: But not the gnosis one, right?

LefterisJP (approved):
LGTM

PR #8740, opened 2024-10-14
Add support for puffer and puffer airdrop claim decoding

LefterisJP:
Also for the PufferXEigen campaign they had


LefterisJP:
huh weird. Linting failed here and not locally.

LefterisJP:
Tests failed due to eth-typing fixing the issue I reported, and since we don't ping eth-typing CI failed. Will fail in develop too. Will push as another commit here.

PR #8746, opened 2024-10-15
fix(frontend): remove manual latest option

lukicenturi:
Merge after https://github.com/rotki/rotki/pull/8742


kelsos (approved):
lgtm, I reverted the auto-imports.d.ts change (because there are differences between develop and bugfixes) and also fixed a small build error in the edit snapshot form.

PR #8754, opened 2024-10-15
Add descriptions to db upgrade and migration steps

nicholasyoder:
Closes: https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=83477669


LefterisJP:
All right so I added a few more commits here and the diff is now huge.

The TL;DR is:

1. Created a uniform mechanism for DB upgrades and migrations. There is some small differences between global and user DB which need a different function, but we can probably improve there in the future.
2. The mechanism is based on introspection. All the step functions are decorated with the progress_step decorator giving them a description. All steps are moved into the main DB upgrade or migration function for the version in question. The code detects them, and executes them in order. The DB upgrades take a write cursor as argument, but can also capture any arguments (if needed) from the enclosing scope of the DB upgrade/migration function. The user DB migrations take the rotki object. The steps are executed in the order of definition of the functions.
3. The steps approach has been extended for global DB upgrades now
4. Global DB migrations, of which we have only a single very old one, have been skipped for now.
5. Removed some unneeded user DB migrations. We need to also get rid of older DB/globalDB upgrades soon.

LefterisJP:
Probably need to add (2) from above in comment in the code.Probably utils/progress.py

yabirgb (approved):
Checked a bit and seems fine but didn't do an exhaustive review. LGTM

PR #8759, opened 2024-10-16
Deprecated Yearn API removal

yabirgb:
- Remove deprecated yearn api
- fail earlier for uniswap oracle


rotkehlchen/chain/ethereum/modules/yearn/utils.py
```diff
@@ -138,6 +117,10 @@ def query_yearn_vaults(db: 'DBHandler') -> None:
             log.error(f'Could not identify the yearn vault type for {vault}. Skipping...')
             continue
 
+        if (version := vault.get('version')) is not None and version.startswith('3.'):
```
LefterisJP, line 120: this should go away in develop merge no?

rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -211,18 +211,6 @@ def get_price(
         - DefiPoolError
         - RemoteError
         """
-        log.debug(
```
LefterisJP, line 214: why did you move this in another place in the function?

rotkehlchen/data_migrations/migrations/migrations_18.py
```diff
@@ -221,4 +221,9 @@ def _remove_manualcurrent_oracle(rotki: 'Rotkehlchen') -> None:
                 ),
             )
 
+    @progress_step(description='Cleaning up Yearn cache.')
+    def _remove_yearn_cache(rotki: 'Rotkehlchen') -> None:
+        with GlobalDBHandler().conn.write_ctx() as write_cursor:
+            write_cursor.execute('DELETE FROM unique_cache WHERE key=?', ('YEARN_VAULTS',))
```
LefterisJP, line 227: why clean up the cache for yearn? You just removed an api source.

rotkehlchen/chain/ethereum/modules/yearn/utils.py
yabirgb: yes

rotkehlchen/data_migrations/migrations/migrations_18.py
yabirgb: Because we store the total amount of entries and this was `len(old_api) + len(new_api)`. To prevent any issue we force a refresh but sicne the tokens should already be in the globaldb it will be fast. Just loop the vaults

LefterisJP (approved):
LGTM

Issue #7622, opened 2024-03-17
cmUMAMI price not displaying


Sovereignllama:
## Problem Definition

cmUMAMI price not displaying properly showing zero

LefterisJP:
This should come in 1.36.0

PR #8762, opened 2024-10-16
Add Umami Finance decoder

nicholasyoder:
Closes: #7622


rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -0,0 +1,175 @@
...
+                event.notes = f'Deposit {amount} {event.asset.symbol_or_name()} into Umami'
+            else:
+                event.event_type = HistoryEventType.SPEND
+                event.event_subtype = HistoryEventSubType.RETURN_WRAPPED
+                event.notes = f'Return {amount} {event.asset.symbol_or_name()} to Umami'
+
+        elif event.balance.amount == asset_normalized_value(
+                amount=context.transaction.value,
+                asset=event.asset.resolve_to_crypto_asset(),
+        ):  # Execution cost, balance will match transaction value
```
LefterisJP, line 72: What exactly is this? What are you checking here? Why are you checking against the transaction value itself? That's a fee right? It's also in case of withdrawals? I have definitely seen it for deposits.

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -0,0 +1,175 @@
...
+        and must be transformed later via an ActionItem.
+        """
+        found_relevant_events = False
+        for event in context.decoded_events:
+            if (
+                event.event_type == HistoryEventType.SPEND and
+                event.event_subtype == HistoryEventSubType.NONE and
+                context.tx_log.topics[0] in {DEPOSIT_TOPIC, WITHDRAW_TOPIC}
+            ):  # deposit or withdraw request
+                self._decode_deposit_withdraw_request(context, event)
```
LefterisJP, line 91: kwargs please

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -0,0 +1,175 @@
...
+        A deposit/withdraw consists of two transactions, a request tx and an execution tx.
+        The event for deposit execution is not available in decoded_events here,
+        and must be transformed later via an ActionItem.
+        """
+        found_relevant_events = False
+        for event in context.decoded_events:
+            if (
+                event.event_type == HistoryEventType.SPEND and
+                event.event_subtype == HistoryEventSubType.NONE and
+                context.tx_log.topics[0] in {DEPOSIT_TOPIC, WITHDRAW_TOPIC}
```
LefterisJP, line 89: for example here you are already checking and getting the topic type and then later in `._decode_deposit_withdraw_request` you again check against it multiple times. Best abstract it beter, check once and route it in the appropriate function somewhere around here

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -0,0 +1,175 @@
...
+            ):  # deposit or withdraw request
+                self._decode_deposit_withdraw_request(context, event)
+                found_relevant_events = True
+            elif (
+                event.event_type == HistoryEventType.RECEIVE and
+                event.event_subtype == HistoryEventSubType.NONE and
+                context.tx_log.topics[0] == TRANSFER_TOPIC
+            ):  # withdraw execution
+                event.event_subtype = HistoryEventSubType.RECEIVE_WRAPPED
+                event.notes = f'Receive {event.balance.amount} {event.asset.symbol_or_name()} from Umami'  # noqa: E501
```
LefterisJP, line 99: You need to resolve he asset first. Otherwise it will fail. And you need symbol, not name. Do this in all other place in the PR. Will not comment on them individually

```suggestion
                event.notes = f'Receive {event.balance.amount} {event.asset.resolve_to_asset_with_symbol().symbol} from Umami'  # noqa: E501
```

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -0,0 +1,175 @@
...
+                found_relevant_events = True
+
+        if found_relevant_events:
+            return DEFAULT_DECODING_OUTPUT
+
+        if (
+            context.tx_log.topics[0] == TRANSFER_TOPIC and
+            context.transaction.input_data.startswith(DEPOSIT_EXECUTION_FOUR_BYTES)
+        ):  # deposit execution
+            vault_asset = Asset(evm_address_to_identifier(
```
LefterisJP, line 110: no no no this is not how you initialize an asset in rotki. use `self.base.get_or_create_evm_asset`

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -0,0 +1,175 @@
...
+                address=context.tx_log.address,
+                chain_id=self.evm_inquirer.chain_id,
+                token_type=EvmTokenKind.ERC20,
+            ))
+            receive_amount = asset_normalized_value(
+                amount=int.from_bytes(context.tx_log.data[0:32]),
+                asset=vault_asset.resolve_to_crypto_asset(),
```
LefterisJP, line 117: also resolve can happen above and then use resolved asset also in the action item

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -0,0 +1,175 @@
...
+                if context.tx_log.topics[0] == REWARD_TOPIC:
+                    event.event_type = HistoryEventType.STAKING
+                    event.event_subtype = HistoryEventSubType.REWARD
+                    event.counterparty = CPT_UMAMI
+                    event.notes = f'Receive staking reward of {event.balance.amount} {event.asset.symbol_or_name()} from Umami'  # noqa: E501
+                elif context.tx_log.topics[0] == UNSTAKE_TOPIC:
+                    event.event_type = HistoryEventType.STAKING
+                    event.event_subtype = HistoryEventSubType.REMOVE_ASSET
+                    event.counterparty = CPT_UMAMI
+                    event.notes = f'Unstake {event.balance.amount} {event.asset.symbol_or_name()} from Umami'  # noqa: E501
```
LefterisJP, line 158: all of those here need changing for resolution as I said above

rotkehlchen/tests/unit/decoders/test_umami.py
```diff
@@ -0,0 +1,318 @@
...
+        ), EvmEvent(
+            sequence_index=1,
+            timestamp=timestamp,
+            location=Location.ARBITRUM_ONE,
+            event_type=HistoryEventType.DEPOSIT,
+            event_subtype=HistoryEventSubType.FEE,
+            asset=A_ETH,
+            balance=Balance(amount=FVal(fee_amount)),
+            location_label=user_address,
+            notes=f'Execution cost of {fee_amount} ETH',
```
LefterisJP, line 60: Since this is a fee for the deposit can you make sure it appears after the deposit? Also add the Umami keyword. "Umami execution cost of ..."

rotkehlchen/tests/unit/decoders/test_umami.py
```diff
@@ -0,0 +1,318 @@
...
+        EvmEvent(
+            sequence_index=8,
+            timestamp=timestamp,
+            location=Location.ARBITRUM_ONE,
+            event_type=HistoryEventType.RECEIVE,
+            event_subtype=HistoryEventSubType.RECEIVE_WRAPPED,
+            asset=A_GM_USDC_WBTC,
+            balance=Balance(amount=FVal(receive_amount)),
+            location_label=user_address,
+            notes=f'Receive {receive_amount} gmUSDC_WBTC from Umami',
```
LefterisJP, line 101: hmmm what exactly is this? Why do you call it deposit_execution when it's just a receive? Should't the notes also say " for a deposit" or something similar?

rotkehlchen/tests/unit/decoders/test_umami.py
```diff
@@ -0,0 +1,318 @@
...
+        EvmEvent(
+            sequence_index=6,
+            timestamp=timestamp,
+            location=Location.ARBITRUM_ONE,
+            event_type=HistoryEventType.RECEIVE,
+            event_subtype=HistoryEventSubType.RECEIVE_WRAPPED,
+            asset=Asset('eip155:42161/erc20:0xaf88d065e77c8cC2239327C5EDb3A432268e5831'),
+            balance=Balance(amount=FVal(receive_amount)),
+            location_label=user_address,
+            notes=f'Receive {receive_amount} USDC from Umami',
```
LefterisJP, line 180: A few things. What is this? If this is a withdrawal then it shoudl say withdraw and type/subtyep shoudl be a withdrawal.

RECEIVE_WRAPPED means to receive a wrapped token. Here you receive USDC. So this is 100% wrong.

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
nicholasyoder: The only meaningful difference between the execution cost event and the deposit/withdraw event is the balance. Currently the asset is also different, with the execution cost in eth and the deposit in some other asset, but if a vault is ever added that uses eth, then the assets on the two events would be the same. So I think checking the event balance against values recorded elsewhere in the context is the only reliable way to differentiate between them.

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`7622` Rotki will now support Umami Finance vaults.
```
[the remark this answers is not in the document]
nicholasyoder: Added the changelog entry back, since we're including the fix for cmUMAMI in this PR, and it closes the issue after all.

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -0,0 +1,236 @@
...
+        )
+
+    def _decode_umami_vault_events(self, context: DecoderContext) -> DecodingOutput:
+        """Decode umami vault deposit/withdraw events.
+
+        A deposit/withdraw consists of two transactions, a request tx and an execution tx.
+        The event for deposit execution is not available in decoded_events here,
+        and must be transformed later via an ActionItem.
+        """
+        main_event = fee_event = None
```
LefterisJP, line 136: Never do that in python. It works in some cases, but not for mutable objects. As if you modify one you end up modifying the other too.

This is how you modify multiple objects explicitly and safely

```suggestion
        main_event, fee_event = None, None
```

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -0,0 +1,236 @@
...
+    from rotkehlchen.user_messages import MessagesAggregator
+
+DEPOSIT_TOPIC: Final = b'\xdc\xbc\x1c\x05$\x0f1\xff:\xd0g\xef\x1e\xe3\\\xe4\x99wbu.:\tR\x84uED\xf4\xc7\t\xd7'  # noqa: E501
+TRANSFER_TOPIC: Final = b'\xdd\xf2R\xad\x1b\xe2\xc8\x9bi\xc2\xb0h\xfc7\x8d\xaa\x95+\xa7\xf1c\xc4\xa1\x16(\xf5ZM\xf5#\xb3\xef'  # noqa: E501
+WITHDRAW_TOPIC: Final = b'\xfb\xdey} \x1ch\x1b\x91\x05e)\x11\x9e\x0b\x02@|{\xb9jJ,u\xc0\x1f\xc9fr2\xc8\xdb'  # noqa: E501
+STAKE_TOPIC: Final = b'\x90\x89\x08\t\xc6T\xf1\x1dnr\xa2\x8f\xa6\x01Iw\n\r\x11\xecl\x921\x9dl\xeb+\xb0\xa4\xea\x1a\x15'  # noqa: E501
+UNSTAKE_TOPIC: Final = b'\xf2y\xe6\xa1\xf5\xe3 \xcc\xa9\x115gm\x9c\xb6\xe4L\xa8\xa0\x8c\x0b\x884+\xcd\xb1\x14Oe\x11\xb5h'  # noqa: E501
+REWARD_TOPIC: Final = b'q\xba\xb6\\\xed.WPwZ\x06\x13\xbe\x06}\xf4\x8e\xf0l\xf9*In\xbfvc\xae\x06`\x92IT'  # noqa: E501
+DEPOSIT_EXECUTION_FOUR_BYTES: Final = b'\xdb\x10\xc3\xb9'
+REQUEST_EVENT_TYPE: TypeAlias = Literal['MAIN', 'FEE']
```
LefterisJP, line 35: Easier here would be to create an enum. More explicit on what you use it for

```suggestion
class FoundEventType(Enum):
    MAIN = auto()
    FEE = auto()
    NONE = auto()
```

This way no need to do `REQUEST_EVENT_TYPE | None` for typing too

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -0,0 +1,236 @@
...
+    def _decode_deposit_withdraw_request(
+            self,
+            event: 'EvmEvent',
+            match_amount: 'FVal',
+            event_type: 'HistoryEventType',
+            event_subtype: 'HistoryEventSubType',
+            event_notes: str,
+            fee_match_amount: 'FVal',
+            fee_event_type: 'HistoryEventType',
+    ) -> REQUEST_EVENT_TYPE | None:
```
LefterisJP, line 60: Change docstring and return types here and elsewhere when you use the enum

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -0,0 +1,236 @@
...
+                amount=context.transaction.value,
+                asset=resolved_asset,
+            ),
+            fee_event_type=HistoryEventType.DEPOSIT,
+        )
+
+    def _decode_withdraw_request(self, context: DecoderContext, event: 'EvmEvent') -> REQUEST_EVENT_TYPE | None:  # noqa: E501
+        """Decode withdraw request events.
+        Returns REQUEST_EVENT_TYPE identifying the type of event found
+        to allow event reshuffling later, or None if no event is found.
```
LefterisJP, line 109: same changes here

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -0,0 +1,236 @@
...
+        if event.balance.amount == match_amount:
+            event.event_type = event_type
+            event.event_subtype = event_subtype
+            event.notes = event_notes
+            event.counterparty = CPT_UMAMI
+            return 'MAIN'
+        elif event.balance.amount == fee_match_amount:
+            event.event_type = fee_event_type
+            event.event_subtype = HistoryEventSubType.FEE
+            event.notes = f'Umami execution cost of {event.balance.amount} ETH'
```
LefterisJP, line 77:
```suggestion
            event.notes = f'Spend {event.balance.amount} ETH as Umami execution fee'
```

tools/assets_database/utils.py
```diff
@@ -123,7 +124,11 @@ def prepare_globaldb(args: argparse.Namespace) -> tuple[GlobalDBHandler, Path]:
         path=args.start_db_path,
         branch=args.assets_branch,
     )
-    return GlobalDBHandler(data_dir=target_directory, sql_vm_instructions_cb=0), target_directory
+    return GlobalDBHandler(
```
LefterisJP, line 127: this actualy needs to go in bugfixes too.

LefterisJP (approved):
LGTM

PR #8765, opened 2024-10-17
Ensure that requests doesn't block

yabirgb:
As per https://docs.python-requests.org/en/latest/user/advanced/#blocking-or-non-blocking it seems that requests always blocks and gevent doesn't patch it. This PR adds grequests that is made by the requests team to work with gevent.

With this changes I managed to not have any 30000ms error while adding a new address and I was able to work at all the time with the app.


yabirgb:
The approach should be to patch httplib

PR #8766, opened 2024-10-17
test: change createAccount command to an async query

kelsos:
Closes #(issue_number)

## Checklist

## Notes
It converts the create account call to use async_query: true, and introduces polling/increased timeout. 

This should provide a fix for the cases where all the tests fail in the before all hook due to 60seconds of timeout during account creation


cleanerzkp (approved):
lgtm

PR #8768, opened 2024-10-17
Fix context switch and add more logs to check progress of decoding

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/assets/utils.py
+++ b/rotkehlchen/assets/utils.py
@@ -211,7 +211,7 @@ def get_token(
 ) -> EvmToken | None:
     """
     Query a token from the cache of the AssetResolver or the GlobalDB if
-    it is not in the cache. If the token doesn't exists this function returns
+    it is not in the cache. If the token doesn't exist this function returns
     None.
     """
     identifier = evm_address_to_identifier(
--- a/rotkehlchen/assets/utils.py
+++ b/rotkehlchen/assets/utils.py
@@ -4,7 +4,14 @@ from typing import TYPE_CHECKING, Any, Final, NamedTuple, Optional
 import regex
 
 from rotkehlchen.api.websockets.typedefs import WSMessageType
-from rotkehlchen.assets.asset import Asset, AssetWithOracles, EvmToken, UnderlyingToken
+from rotkehlchen.assets.asset import (
+    Asset,
+    AssetWithOracles,
+    EvmToken,
+    UnderlyingToken,
+    WrongAssetType,
+)
+from rotkehlchen.assets.resolver import AssetResolver
 from rotkehlchen.assets.types import AssetType
 from rotkehlchen.constants.assets import (
     A_ETH,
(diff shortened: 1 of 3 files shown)


rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -527,10 +531,13 @@ def _decode_transaction(
...
-            token = GlobalDBHandler.get_evm_token(
-                address=tx_log.address,
-                chain_id=self.evm_inquirer.chain_id,
-            )
+            if (token := tokens_cache.get(tx_log.address)) is None:
+                token = GlobalDBHandler.get_evm_token(
+                    address=tx_log.address,
+                    chain_id=self.evm_inquirer.chain_id,
+                )
+                tokens_cache[tx_log.address] = token
```
LefterisJP, line 539: instead of a temporary cache like this here, would it not be better to do `get_or_create_evm_token`? Don't we have a cache somewhere there already in the resolver? This is why we want to be using that function over anything else.

I guess your cache for this particular case is a lot more straighforward but don't we already do that in get_or_create() with an in memory LRU cache? I remember we did. And if we do not then we absolutely should as this would solve similar problems in smaller scale over the rest of the codebase.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -697,8 +704,11 @@ def _decode_transaction_hashes(
 
         refresh_balances = False
         total_transactions = len(tx_hashes)
+        log.debug(f'Started logic to decode {total_transactions} from {self.evm_inquirer.chain_id}')  # noqa: E501
```
LefterisJP, line 707:
```suggestion
        log.debug(f'Started logic to decode {total_transactions=} from {self.evm_inquirer.chain_name}')  # noqa: E501
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -697,8 +704,11 @@ def _decode_transaction_hashes(
...
         for tx_index, tx_hash in enumerate(tx_hashes):
+            log.debug(f'Decoding logic started for {tx_hash.hex()}({self.evm_inquirer.chain_name})')  # noqa: E501
             if send_ws_notifications and tx_index % 10 == 0:
+                log.debug(f'Processed {tx_index} out of {total_transactions} from {self.evm_inquirer.chain_id}')  # noqa: E501
```
LefterisJP, line 711: Please be consistent in the logs

```suggestion
                log.debug(f'Processed {tx_index} out of {total_transactions} transactions from {self.evm_inquirer.chain_name}')  # noqa: E501
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -492,7 +496,7 @@ def _decode_transaction(
                 # for legacy transfers.
                 monerium_special_handling_event = True
 
-            if idx + 1 % MIN_LOGS_PROCESSED_TO_SLEEP == 0:
+            if (idx + 1) % MIN_LOGS_PROCESSED_TO_SLEEP == 0:
                 log.debug(f'Context switching out of the {idx + 1} log event of {transaction}')
```
LefterisJP, line 500:
```suggestion
                log.debug(f'Context switching out of the log event nr. {idx + 1} of {chain_name} tx_hash: {transaction}')
```

rotkehlchen/assets/utils.py
```diff
@@ -197,6 +204,30 @@ class TokenEncounterInfo(NamedTuple):
...
 
 
+def get_token(
+        evm_address: ChecksumEvmAddress,
+        chain_id: ChainID,
+        token_kind: EvmTokenKind = EvmTokenKind.ERC20,
+) -> EvmToken | None:
+    """
+    Query a token from the cache of the AssetResolver or the GlobalDB if
+    it is not in the cache. If the token doesn't exists this function returns
```
LefterisJP, line 214:
```suggestion
    it is not in the cache. If the token doesn't exist this function returns
```

LefterisJP (commented):
nitpick on docstring. Since I am still awake will push and then merge

LefterisJP (approved):
LGTM

PR #8769, opened 2024-10-17
Fix typing in tests/api/test_b*

nicholasyoder:
Related #842

### DIFF
--- a/rotkehlchen/tests/api/test_bitcoin.py
+++ b/rotkehlchen/tests/api/test_bitcoin.py
@@ -1,4 +1,5 @@
 from http import HTTPStatus
+from typing import TYPE_CHECKING, Any
 
 import gevent
 import pytest
@@ -15,6 +16,10 @@ from rotkehlchen.tests.utils.api import (
 )
 from rotkehlchen.tests.utils.factories import UNIT_BTC_ADDRESS1, UNIT_BTC_ADDRESS2
 from rotkehlchen.types import SupportedBlockchain
+
+if TYPE_CHECKING:
+    from rotkehlchen.api.server import APIServer
+    from rotkehlchen.tests.fixtures.networking import ConfigurableSession
 
 EXPECTED_XPUB_ADDESSES = [
     '125yorj66rmk82tMAPG7x97iY8D7cashiA',
@@ -62,7 +67,7 @@ TEST_BITCOIN_XPUB_1 = 'xpub6DCi5iJ57ZPd5qPzvTm5hUt6X23TJdh9H4NjNsNbt7t7UuTMJfawQ
 TEST_BITCOIN_XPUB_2 = 'xpub68V4ZQQ62mea7ZUKn2urQu47Bdn2Wr7SxrBxBDDwE3kjytj361YBGSKDT4WoBrE5htrSB8eAMe59NPnKrcAbiv2veN5GQUmfdjRddD1Hxrk'  # noqa: E501
 
 
-def _check_xpub_addition_outcome(outcome, xpub):
+def _check_xpub_addition_outcome(outcome: dict[str, Any], xpub: str) -> None:
     """Checks the outcome of the xpub additions for the following test.
     Both results should be the same since the 2nd xpub derives no mainnet addresses
     """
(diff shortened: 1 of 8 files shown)


rotkehlchen/tests/api/test_bitcoin.py
```diff
@@ -62,7 +68,7 @@
 TEST_BITCOIN_XPUB_2 = 'xpub68V4ZQQ62mea7ZUKn2urQu47Bdn2Wr7SxrBxBDDwE3kjytj361YBGSKDT4WoBrE5htrSB8eAMe59NPnKrcAbiv2veN5GQUmfdjRddD1Hxrk'  # noqa: E501
 
 
-def _check_xpub_addition_outcome(outcome, xpub):
+def _check_xpub_addition_outcome(outcome: Any, xpub: str) -> None:
```
LefterisJP, line 71: Again same here. Don't use `Any`. What is `result` I see `assert_proper_sync_response_with_result` also uses it, which is wrong. It should be most probably a dict right? Or can it also be a list? In which case you need a union.

If you need both then you would need to create different `assert_` functions which would check the result internally and return it. Since this is tests should be fine

rotkehlchen/tests/api/test_bitcoin.py
```diff
@@ -320,22 +327,23 @@ def test_add_delete_xpub_multiple_chains(rotkehlchen_api_server):
     # Check that addresses that have balances on both bitcoin and bitcoin cash are stored properly
     with rotki.data.db.conn.read_ctx() as cursor:
         mutual_in_chains_aggregator = 0
-        for btc_addr in rotki.chains_aggregator.accounts.btc:
+        btc_addesses: tuple[BTCAddress, ...] = rotki.chains_aggregator.accounts.btc
```
LefterisJP, line 330: 1. you have a typo. `btc_addesses`
2. This should not be needed. This is the type of accounts.btc. Something else is off.

rotkehlchen/tests/api/test_bitcoin.py
```diff
@@ -320,22 +327,23 @@ def test_add_delete_xpub_multiple_chains(rotkehlchen_api_server):
...
-            addresses=rotki.chains_aggregator.accounts.bch,
+            addresses=list(rotki.chains_aggregator.accounts.bch),
         ))
         # get mutual by querying BTC accounts from known BCH accounts
         # this checks that bitcoin cash mappings in the db are not broken
         mutual_in_db_2 = len(rotki.data.db.get_addresses_to_xpub_mapping(
             cursor=cursor,
             blockchain=SupportedBlockchain.BITCOIN_CASH,
-            addresses=rotki.chains_aggregator.accounts.btc,
+            addresses=list(rotki.chains_aggregator.accounts.btc),
```
LefterisJP, line 346: why is the `list()` needed? Does that not mean that the original type requirement is off? Why not make `get_addresses_to_xpub_mapping` accept a  `Sequence[BTCAddress]` instead?

rotkehlchen/tests/api/test_bitcoin.py
nicholasyoder: I was able to make this simpler by just going `btc_addr: BTCAddress`, but I haven't been able to find any more elegant solution.

LefterisJP (commented):
Looks okay. Will approve when I check the btc_addr thing myself to see if I can figure out why

rotkehlchen/tests/api/test_bitcoin.py
LefterisJP: I changed this by:

```python
    reveal_type(rotki.chains_aggregator.accounts.btc)
    assert len(rotki.chains_aggregator.accounts.btc) == 0
    reveal_type(rotki.chains_aggregator.accounts.btc)
```

I was looking at the types as seen by mypy. You can do reveal_type() and run mypy at any point. Will tell you what type mypy thinks it has at that line. I surrounded this line: https://github.com/rotki/rotki/blob/57f687e1192c8d3c6e66616e0e81e1093d5c0736/rotkehlchen/tests/api/test_bitcoin.py#L292

Result is this:

rotkehlchen/tests/api/test_bitcoin.py:299: note: Revealed type is "builtins.tuple[rotkehlchen.types.BTCAddress, ...]"
rotkehlchen/tests/api/test_bitcoin.py:301: note: Revealed type is "tuple[()]"


So ... before the check the type inference is correct. After the length == 0 type inference is that now it's an empty tuple. Which is also correct. But it does not understand the type has changed later down where you needed it.

But this uncovers the problem. I think the original types are off. They should not be tuples. Tuples are immutable

LefterisJP (approved):
LGTM

PR #8771, opened 2024-10-18
Add ability to vacuum at migrations. Enable for migration 18.

LefterisJP:
Reason is that this migration gets rid of a lot of un-needed data for users in some special cases and vacuuming is the only way to reclaim that DB space.


yabirgb (approved):
lgtm

PR #8778, opened 2024-10-18
Add msg_aggregator to GlobalDBHandler in assets script

nicholasyoder:
This change was added to develop in https://github.com/rotki/rotki/pull/8762, but also needs to be present in bugfixes.


LefterisJP (approved):
LGTM

PR #8787, opened 2024-10-21
Fix cursor reuse in zksync lite transaction loading

### LINKED ISSUE #8777: Missing zksync lite transactions
## Problem Definition

The latest zksync lite transactions for 0xc10fcf82f3b870cbb2b0136a0c891f6b410497c8 do not show in History Events, although it seems that they have been retrieved correctly from the zksync api, and are present in the database.

### Steps to reproduce:
1. Create a new Rotki profile
2. Add a Blockchain Address on ZKSync Lite for 0xc10fcf82f3b870cbb2b0136a0c891f6b410497c8
3. Go to History Events and the last few events will be missing. Refreshing / re-decoding can sometimes make a few more txs appear.

Reported in [discord](https://discord.com/channels/657906918408585217/745172610240872508/1285143781703811106).

nicholasyoder:
Closes #8777

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* :bug:`8777` ZKSync lite transactions should no longer be skipped under special circumstances.
+
 * :release:`1.35.1 <2024-10-18>`
 * :bug:`-` yearn vaults data will be properly queried only once and error handling is improved.
 * :bug:`8715` Fix the issue where the historical price doesn't refresh after being updated on the history events page.
--- a/rotkehlchen/chain/zksync_lite/manager.py
+++ b/rotkehlchen/chain/zksync_lite/manager.py
@@ -495,6 +495,14 @@ class ZksyncLiteManager:
         transactions = []
         with self.database.conn.read_ctx() as cursor:
             cursor.execute(
+                f'SELECT tx_id, from_asset, from_amount, to_asset, to_amount '
+                f'FROM zksynclite_swaps WHERE tx_id IN '
+                f'(SELECT identifier FROM zksynclite_transactions{queryfilter})',
+                bindings,
+            )
+            swap_data = {row[0]: row[1:] for row in cursor}
+
+            cursor.execute(
                 f'SELECT identifier, tx_hash, type, timestamp, block_number, from_address, '
                 f'to_address, asset, amount, fee FROM zksynclite_transactions{queryfilter}',
                 bindings,
@@ -504,17 +512,11 @@ class ZksyncLiteManager:
                 try:
                     tx = ZKSyncLiteTransaction.deserialize_from_db(entry[1:])
                     if tx.tx_type == ZKSyncLiteTXType.SWAP:
-                        cursor.execute(
-                            'SELECT from_asset, from_amount, to_asset, to_amount '
-                            'FROM zksynclite_swaps WHERE tx_id=?', (entry[0],),
-                        )
-                        if (swap_result := cursor.fetchone()) is None:
-                            log.error(
-                                f'Could not deserialize zksync lite transaction from the DB due to not finding swap data for tx_id: {entry[0]}',  # noqa: E501
-                            )
+                        if (tx_data := swap_data.get(entry[0])) is None:
+                            log.error(f'Could not deserialize zksync lite transaction from the DB due to not finding swap data for tx_id: {entry[0]}')  # noqa: E501
                             continue
 
-                        tx.swap_data = ZKSyncLiteSwapData.deserialize_from_db(swap_result)
+                        tx.swap_data = ZKSyncLiteSwapData.deserialize_from_db(tx_data)
 
                     transactions.append(tx)
                 except (DeserializationError, UnknownAsset) as e:
--- a/rotkehlchen/tests/integration/test_zksynclite.py
+++ b/rotkehlchen/tests/integration/test_zksynclite.py
@@ -1,7 +1,9 @@
+from typing import TYPE_CHECKING
+
 import pytest
 
 from rotkehlchen.accounting.structures.balance import Balance
-from rotkehlchen.assets.asset import EvmToken
+from rotkehlchen.assets.asset import Asset, EvmToken
 from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.chain.zksync_lite.constants import ZKL_IDENTIFIER
 from rotkehlchen.chain.zksync_lite.structures import (
@@ -21,7 +23,7 @@ from rotkehlchen.constants.assets import (
     A_USDT,
     A_WBTC,
 )
-from rotkehlchen.constants.misc import ZERO
+from rotkehlchen.constants.misc import ONE, ZERO
 from rotkehlchen.db.filtering import EvmEventFilterQuery
 from rotkehlchen.db.history_events import DBHistoryEvents
 from rotkehlchen.fval import FVal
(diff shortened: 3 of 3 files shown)


rotkehlchen/chain/zksync_lite/manager.py
```diff
@@ -494,34 +494,37 @@ def get_db_transactions(
         """Gets any zksynclite transactions from the DB depending on the given filter"""
         transactions = []
         with self.database.conn.read_ctx() as cursor:
-            cursor.execute(
+            zksync_lite_data = cursor.execute(
                 f'SELECT identifier, tx_hash, type, timestamp, block_number, from_address, '
                 f'to_address, asset, amount, fee FROM zksynclite_transactions{queryfilter}',
                 bindings,
+            ).fetchall()
```
yabirgb, line 501: You can avoid executing this `fetchall` if you do this query after saving `swap_data = {row[0]: row[1:] for row in cursor}`.

Fetchall is bad because it is a for loop that consumes all the entries into memory so you process twice the list

rotkehlchen/chain/zksync_lite/manager.py
```diff
@@ -494,34 +494,37 @@ def get_db_transactions(
...
 
-                        tx.swap_data = ZKSyncLiteSwapData.deserialize_from_db(swap_result)
+        for entry in zksync_lite_data:
+            swap_result = None
+            try:
+                tx = ZKSyncLiteTransaction.deserialize_from_db(entry[1:])
+                if tx.tx_type == ZKSyncLiteTXType.SWAP:
+                    if (tx_id := entry[0]) not in swap_data:
+                        log.error(f'Could not deserialize zksync lite transaction from the DB due to not finding swap data for tx_id: {tx_id}')  # noqa: E501
+                        continue
```
yabirgb, line 518: It is cleaner if you do

```suggestion
                    if (tx_data := swap_data.get(entry[0]) is None:
                        log.error(f'Could not deserialize zksync lite transaction from the DB due to not finding swap data for tx_id: {entry[0]}')  # noqa: E501
                        continue
```

then later you don't need to do `swap_data[tx_id]`, just `tx_data`

rotkehlchen/tests/integration/test_zksynclite.py
```diff
@@ -341,3 +346,61 @@ def test_decode_swap(zksync_lite_manager, inquirer):  # pylint: disable=unused-a
...
+                ),
+                ZKSyncLiteTransaction(
+                    tx_hash=deserialize_evm_tx_hash('0x34f4c91b42657bddd4698a7c61152f2a35613c4096f48f554945c4301e08464e'),
+                    tx_type=ZKSyncLiteTXType.TRANSFER,
+                    timestamp=Timestamp(1673340336),
+                    block_number=147251,
+                    from_address=address,
+                    to_address=address,
+                    asset=Asset('eip155:1/erc20:0x6B175474E89094C44Da98b954EedeAC495271d0F'),
+                    amount=FVal(0),
```
yabirgb, line 395:
```suggestion
                    amount=ZERO,
```

rotkehlchen/tests/integration/test_zksynclite.py
```diff
@@ -341,3 +346,61 @@ def test_decode_swap(zksync_lite_manager, inquirer):  # pylint: disable=unused-a
...
+            transactions=[
+                ZKSyncLiteTransaction(
+                    tx_hash=deserialize_evm_tx_hash('0x9922304b069ed8405edcc3c7c4a8c8cc9c8f1edb6a67809f57543e8fe0fa9875'),
+                    tx_type=ZKSyncLiteTXType.TRANSFER,
+                    timestamp=Timestamp(1663539042),
+                    block_number=105088,
+                    from_address=address,
+                    to_address=string_to_evm_address('0x10E87b05fe0EDE0BbB0a52aFa96c08618A3E02F0'),
+                    asset=Asset('eip155:1/erc20:0x6B175474E89094C44Da98b954EedeAC495271d0F'),
+                    amount=FVal(1),
```
yabirgb, line 366:
```suggestion
                    amount=ONE,
```

rotkehlchen/tests/integration/test_zksynclite.py
```diff
@@ -341,3 +346,61 @@ def test_decode_swap(zksync_lite_manager, inquirer):  # pylint: disable=unused-a
...
+                    tx_hash=deserialize_evm_tx_hash('0x34f4c91b42657bddd4698a7c61152f2a35613c4096f48f554945c4301e08464e'),
+                    tx_type=ZKSyncLiteTXType.TRANSFER,
+                    timestamp=Timestamp(1673340336),
+                    block_number=147251,
+                    from_address=address,
+                    to_address=address,
+                    asset=Asset('eip155:1/erc20:0x6B175474E89094C44Da98b954EedeAC495271d0F'),
+                    amount=FVal(0),
+                    fee=None,
+                    swap_data=None,
```
yabirgb, line 397: swap_data by default is None so you don't need to explictly set it

rotkehlchen/chain/zksync_lite/manager.py
```diff
@@ -494,34 +494,37 @@ def get_db_transactions(
         """Gets any zksynclite transactions from the DB depending on the given filter"""
         transactions = []
         with self.database.conn.read_ctx() as cursor:
-            cursor.execute(
```
yabirgb, line 497: another approach that you can take is to use two cursors:

```
self.database.conn.read_ctx() as transactions_read_cursor,
self.database.conn.read_ctx() as swaps_read_cursor
```

and use both of them in different places. Can you measure the difference? I believe creating one extra cursor is not expensive and you avoid storing in memory `swap_data`. Also probably the query is more performant than using the IN operator with the subquery

yabirgb (commented):
This change is missing a changelog entry right? ALso the target should be bugfixes since you are fixing something broken right?

rotkehlchen/chain/zksync_lite/manager.py
nicholasyoder: As mentioned in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1297916464845557811), doing a query per transaction would have an significant hit on performance.

docs/changelog.rst
```diff
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* :bug:`8777` Fix database cursor reuse causing zksync lite transactions to be skipped.
```
LefterisJP, line 5: The changelog text is user facing. Don't mention details like this. it's greek to them.

"ZKSync lite transactions should no longer be skipped under special circumstances"

rotkehlchen/tests/integration/test_zksynclite.py
```diff
@@ -341,3 +346,59 @@ def test_decode_swap(zksync_lite_manager, inquirer):  # pylint: disable=unused-a
         notes='Swap fee of 0.1659 DAI',
         address=address,
     )]
+
+
+def test_get_db_transactions(zksync_lite_manager: 'ZksyncLiteManager'):
+    """Test that all zksync lite transactions are loaded from the database."""
```
LefterisJP, line 352: You need to rename the test and mention in the docstring this is a regression test for the issue and point to the issue with a link.

This is not about getting transactions but decoding and showing them.

LefterisJP (commented):
Shouldn't we also redecode all affected ZKSync lite transactions?

LefterisJP (approved):
LGTM

PR #8795, opened 2024-10-22
Improve yearn decoder

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=84143837


docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`-` Rotki will now support more Yearn vaults due to improved decoding logic.
```
LefterisJP, line 5: Don't add this. You already added yearn v3 earlier. If anything augment that entry.

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
```diff
@@ -76,12 +90,12 @@ def __init__(
             YEARN_VAULTS_V3_PROTOCOL: set(),
         }
 
-    def reload_data(self) -> None:
+    def reload_data(self) -> Mapping['ChecksumEvmAddress', tuple[Any, ...]] | None:
```
LefterisJP, line 93: adjust the docstring to explain what this returns now.

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
```diff
@@ -93,97 +107,131 @@ def reload_data(self) -> None:
...
+            protocol = YEARN_VAULTS_V1_PROTOCOL
+        elif context.transaction.to_address not in self.vaults[YEARN_VAULTS_V2_PROTOCOL]:
             return FAILED_ENRICHMENT_OUTPUT
 
-        is_deposit = False
         if (
             context.transaction.input_data.startswith(YEARN_DEPOSIT_4_BYTES) or
-            context.transaction.input_data.startswith(YEARN_DEPOSIT_AMOUNT_4_BYTES) or
-            context.transaction.input_data.startswith(YEARN_DEPOSIT_4_BYTES_V3)
+            context.transaction.input_data.startswith(YEARN_DEPOSIT_AMOUNT_4_BYTES)
```
LefterisJP, line 146: why are you checking this here? Input data decoding is not reliable. And even if you want to use this then you should use `decoding_by_input_data`, not here.

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
```diff
@@ -93,97 +107,131 @@ def reload_data(self) -> None:
...
+
+    def _decode_deposit_withdraw(self, context: DecoderContext) -> DecodingOutput:
+        """Decode deposit/withdraw events for yearn v2 and v3."""
+        for event in context.decoded_events:
+            if context.tx_log.topics[0] == YEARN_V2_WITHDRAW_TOPIC:
+                self._handle_withdraw_event(event=event, protocol=YEARN_VAULTS_V2_PROTOCOL)
+            elif context.tx_log.topics[0] == YEARN_V3_WITHDRAW_TOPIC:
+                self._handle_withdraw_event(event=event, protocol=YEARN_VAULTS_V3_PROTOCOL)
+            elif context.tx_log.topics[0] == YEARN_V2_DEPOSIT_TOPIC:
+                self._handle_deposit_event(event=event, protocol=YEARN_VAULTS_V2_PROTOCOL)
```
LefterisJP, line 213: Why do you need these `YEARN_VAULTS_V3_PROTOCOL`?And also you make a mapping of these to counterparties? Wouldn't just using the counterparty everywhere by sufficient?

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
```diff
@@ -93,97 +107,131 @@ def reload_data(self) -> None:
...
+            elif context.tx_log.topics[0] == YEARN_V3_DEPOSIT_TOPIC:
+                self._handle_deposit_event(event=event, protocol=YEARN_VAULTS_V3_PROTOCOL)
+
+        return DEFAULT_DECODING_OUTPUT
+
+    def _decode_v2_increase_deposit(self, context: DecoderContext) -> DecodingOutput:
+        """Decode increase deposit events for yearn v2."""
+        if context.tx_log.topics[0] == YEARN_V2_INCREASE_DEPOSIT_TOPIC:
+            for event in context.decoded_events:
```
LefterisJP, line 222: Why do it for all decoded events? This does not make sense.

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
```diff
@@ -93,97 +107,131 @@ def reload_data(self) -> None:
...
-        else:
-            # in this case we failed to find a valid transfer event. Inform about the failure
-            return FAILED_ENRICHMENT_OUTPUT
-
-        return TransferEnrichmentOutput(matched_counterparty=protocol)
+            event.event_subtype = HistoryEventSubType.RECEIVE_WRAPPED
+            event.counterparty = counterparty
+            event.notes = f'Receive {event.balance.amount} {event.asset.resolve_to_asset_with_symbol().symbol} after deposit in a {counterparty} vault'  # noqa: E501
+
+    def _decode_deposit_withdraw(self, context: DecoderContext) -> DecodingOutput:
```
LefterisJP, line 205: This should be called _decode_vault_event

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
```diff
@@ -93,97 +107,131 @@ def reload_data(self) -> None:
...
+        """Decode deposit/withdraw events for yearn v2 and v3."""
+        for event in context.decoded_events:
```
LefterisJP, line 207: What are you doing here? This function will get here only once per log event. Why iterate all already decoded events and redecode again? Do you understand how the decoders work?!?!

rotkehlchen/chain/ethereum/modules/yearn/utils.py
```diff
@@ -114,13 +111,11 @@ def query_yearn_vaults(db: 'DBHandler') -> None:
 
     assert data is not None, 'data exists. Checked by _maybe_reset_yearn_cache_timestamp'
     for vault in data:
-        if 'type' not in vault or (vault['type'] == '' and 'version' not in vault):
+        if 'version' not in vault:
             log.error(f'Could not identify the yearn vault type for {vault}. Skipping...')
             continue
```
LefterisJP, line 116:
```suggestion
        if (version := vault.get('version')) is not None:
            log.error(f'Could not identify the yearn vault type for {vault}. Skipping...')
            continue
```

rotkehlchen/chain/ethereum/modules/yearn/utils.py
```diff
@@ -114,13 +111,11 @@ def query_yearn_vaults(db: 'DBHandler') -> None:
...
 
-        if vault['type'] == 'v1':
-            vault_type = YEARN_VAULTS_V1_PROTOCOL
-        elif vault['type'] == 'v2' or vault['version'].startswith('0.'):  # version '0.x.x' happens in ydemon and is always a v2 vault  # noqa: E501
+        if vault['version'].startswith('0.'):  # version '0.x.x' happens in ydemon and is always a v2 vault  # noqa: E501
```
LefterisJP, line 118:
```suggestion
        if version.startswith('0.'):  # version '0.x.x' happens in ydemon and is always a v2 vault  # noqa: E501
```

rotkehlchen/chain/ethereum/modules/yearn/utils.py
```diff
@@ -114,13 +111,11 @@ def query_yearn_vaults(db: 'DBHandler') -> None:
...
     assert data is not None, 'data exists. Checked by _maybe_reset_yearn_cache_timestamp'
     for vault in data:
-        if 'type' not in vault or (vault['type'] == '' and 'version' not in vault):
+        if 'version' not in vault:
             log.error(f'Could not identify the yearn vault type for {vault}. Skipping...')
             continue
 
-        if vault['type'] == 'v1':
-            vault_type = YEARN_VAULTS_V1_PROTOCOL
-        elif vault['type'] == 'v2' or vault['version'].startswith('0.'):  # version '0.x.x' happens in ydemon and is always a v2 vault  # noqa: E501
```
LefterisJP, line 123: why did you remove the vault['type'] check here?

rotkehlchen/tests/unit/decoders/test_yearn.py
```diff
@@ -115,7 +117,7 @@ def test_deposit_yearn_v3(
             asset=Asset('eip155:1/erc20:0xBF319dDC2Edc1Eb6FDf9910E39b37Be221C8805F'),
             balance=Balance(amount=FVal(receive_amount)),
             location_label=user_address,
-            notes=f'Receive {receive_amount} crvUSD-2 yVault after deposit in a yearn-v3 vault',
+            notes=f'Receive {receive_amount} yvcrvUSD-2 after deposit in a yearn-v3 vault',
```
LefterisJP, line 120: this is wrong. The received token is crvUSD-2 yVault. Did you check the transaction that this is decoding? Why would you think changing this is fine?

rotkehlchen/tests/unit/decoders/test_yearn.py
```diff
@@ -191,8 +193,227 @@ def test_withdraw_yearn_v3(
...
+        ), EvmEvent(
+            tx_hash=tx_hash,
+            sequence_index=365,
+            timestamp=timestamp,
+            location=Location.ETHEREUM,
+            event_type=HistoryEventType.INFORMATIONAL,
+            event_subtype=HistoryEventSubType.APPROVE,
+            asset=Asset('eip155:1/erc20:0xef484de8C07B6e2d732A92B5F78e81B38f99f95E'),
+            balance=Balance(amount=FVal(approve_amount)),
+            location_label=user_address,
```
LefterisJP, line 257: The order here is off. Deposit shoudd be first, followed by the token receival. OUT, then IN. Approval should be pushed out. You need to use maybe_reshuffle_Events in the decoder

rotkehlchen/tests/unit/decoders/test_yearn.py
```diff
@@ -191,8 +193,227 @@ def test_withdraw_yearn_v3(
...
 
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('ethereum_accounts', [['0x34A7a276eD77c6FE866c75Bbc8d79127c4E14a09']])
+@pytest.mark.parametrize('use_clean_caching_directory', [True])
+def test_deposit_yearn_v2(
+        ethereum_inquirer: 'EthereumInquirer',
+        ethereum_accounts: list['ChecksumEvmAddress'],
+) -> None:
+    tx_hash = deserialize_evm_tx_hash('0x7e6493bb2e0c09b9986ab43a562409fd2cf99caca3d1b65fa4b0d46ef5588081')  # noqa: E501
+    get_or_create_evm_token(
```
LefterisJP, line 204: why do you need this? Does the decoder not do get_or_create? Samefor other tests.

rotkehlchen/tests/unit/decoders/test_yearn.py
```diff
@@ -191,8 +193,227 @@ def test_withdraw_yearn_v3(
...
+            notes=f'Withdraw {withdraw_amount} YFIETH-f from yearn-v2 vault yvCurve-YFIETH 0.4.3',
+            counterparty=CPT_YEARN_V2,
+            address=string_to_evm_address('0x790a60024bC3aea28385b60480f15a0771f26D09'),
+        ),
+    ]
+
+
 @pytest.mark.parametrize('ethereum_accounts', [['0xb524c787669185E11d01C645D1910631e04Fa5Eb']])
-def test_deposit_yearn_v2(database, ethereum_inquirer, eth_transactions):
+def test_deposit_yearn_v2_enrich(database, ethereum_inquirer, eth_transactions):
```
LefterisJP, line 416: enrich is an implementation detail. Say without logs or something.

rotkehlchen/tests/unit/decoders/test_yearn.py
```diff
@@ -303,15 +524,15 @@ def test_deposit_yearn_v2(database, ethereum_inquirer, eth_transactions):
             asset=Asset('eip155:1/erc20:0xdb25cA703181E7484a155DD612b06f57E12Be5F0'),
             balance=Balance(amount=FVal('0.02164738945170483'), usd_value=ZERO),
             location_label=user_address,
-            notes='Receive 0.02164738945170483 YFI yVault after deposit in a yearn-v2 vault',
```
LefterisJP, line 306: rewrite this test to use VCR. It's the old type that hard coded stuff.

rotkehlchen/tests/unit/decoders/test_yearn.py
```diff
@@ -303,15 +524,15 @@ def test_deposit_yearn_v2(database, ethereum_inquirer, eth_transactions):
...
+            notes='Receive 0.02164738945170483 yvYFI after deposit in a yearn-v2 vault',
             counterparty=CPT_YEARN_V2,
             address=ZERO_ADDRESS,
         )]
     assert events == expected_events
 
 
 @pytest.mark.parametrize('ethereum_accounts', [['0x7b33b34D45A395518a4143846Ac40dA78CbcAA91']])
-def test_withdraw_yearn_v2(database, ethereum_inquirer, eth_transactions):
+def test_withdraw_yearn_v2_enrich(database, ethereum_inquirer, eth_transactions):
```
LefterisJP, line 535: same comments here as above

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
```diff
@@ -93,97 +107,131 @@ def reload_data(self) -> None:
...
-            is_deposit is False and
-            context.event.event_type == HistoryEventType.RECEIVE and
-            context.event.address == context.transaction.to_address
+            event.event_type = HistoryEventType.WITHDRAWAL
+            event.event_subtype = HistoryEventSubType.REMOVE_ASSET
+            event.counterparty = counterparty
+            vault_token_name = _get_vault_token_name(event.address)
+            event.notes = f'Withdraw {event.balance.amount} {event.asset.resolve_to_asset_with_symbol().symbol} from {counterparty} vault {vault_token_name}'  # noqa: E501
+
+    def _handle_deposit_event(self, event: 'EvmEvent', protocol: str) -> None:
```
LefterisJP, line 181: you can limit thetype in the protocol too with `Literal`

```suggestion
    def _handle_deposit_event(self, event: 'EvmEvent', protocol: str) -> None:
```

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
nicholasyoder: Unfortunately it seems that the input data is the only way to identify these transfer events. I've reworked it to use `decoding_by_input_data` instead of enriching.

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
nicholasyoder: I've changed it to use counterparty most places now, but it does still need the protocol to counterparty mapping for when the vaults are loaded from the database.

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
nicholasyoder: Moved the loop into a subfunction as discussed in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1298386709339308153).

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
nicholasyoder: Moved loop into subfunction, see https://github.com/rotki/rotki/pull/8795#discussion_r1813412944

rotkehlchen/chain/ethereum/modules/yearn/utils.py
nicholasyoder: This was apparently something from the old api that was removed here: https://github.com/rotki/rotki/commit/5ec2771ebbef85b8750cad5a31eaaa90f444b197. In the data from the ydaemon API, `type` is either `Yearn Vault`, `Automated Yearn Vault`, or something similar, and can't be used to determine the vault version.

rotkehlchen/tests/unit/decoders/test_yearn.py
nicholasyoder: For this [token](https://etherscan.io/address/0xbf319ddc2edc1eb6fdf9910e39b37be221c8805f), `crvUSD-2 yVault` is the token name, and `yvcrvUSD-2` is the token symbol. The symbol is generally what is used directly after the amount in event notes like this.

rotkehlchen/tests/unit/decoders/test_yearn.py
nicholasyoder: Normally the vault tokens are all created in `query_yearn_vaults` using the data from the API, but in the tests it doesn't do that api call, so we have to manually add the token needed for the test, unless its a token that is already in the packaged globaldb, as is the case for some of the older tests.

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
```diff
@@ -92,102 +110,264 @@ def reload_data(self) -> None:
...
+
+    def _maybe_get_other_transfer_event(
+            self,
+            context: DecoderContext,
+            event_type: HistoryEventType,
+            event_subtype: HistoryEventSubType,
+    ) -> 'EvmEvent | None':
+        """"""
+        for event in context.decoded_events:
+            if event.event_type == event_type and event.event_subtype == event_subtype:
```
LefterisJP, line 247: Isn't this too wide a matching? Just type/subtype? Shouldn't you narrow it down further?

nicholasyoder: I added a check for counterparty, which seems to be pretty much the only other thing that can be checked for easily, since asset and amount for the event it's matching are unknown at the point where this is being used. I think checking counterparty and event type/subtype should be sufficient.

nicholasyoder: Also added check for `location_label`, which should always match `context.transaction.from_address`

nicholasyoder: There simply does not seem to be enough data available in the log to match this any better.
I've removed the `location_label` check again and put a note in the docstring as discussed in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1298994891011133521)

rotkehlchen/chain/ethereum/modules/yearn/decoder.py
```diff
@@ -92,102 +110,278 @@ def reload_data(self) -> None:
...
+            self,
+            context: DecoderContext,
+            event_type: HistoryEventType,
+            event_subtype: HistoryEventSubType,
+            counterparty: str,
+    ) -> 'EvmEvent | None':
+        """Find matching event in decoded_events, to be used for event reordering.
+        Returns the event or None if no matching event is found.
+
+        The matching is too open here, but there isn't enough data available in the log
```
LefterisJP, line 249: We need to handle this better as this is dangerous, but for now will merge

LefterisJP (approved):
LGTM though the matching on _maybe_get_other_transfer() is too open. That needs to be handled and thought of separately

Failing test is unrelated and will be fixed elsewhere.

Issue #8773, opened 2024-10-18
Don't show saved filters when a table does not support this combination


kelsos:
## Problem Definition

It seems that you are able to select saved filters in a table that does not support these filters.
You shouldn't since that leads to confusion.

PR #8804, opened 2024-10-24
fix: saved filter consistency

lukicenturi:
Closes https://github.com/rotki/rotki/issues/8773

- [x] Remove unsupported filters on the table.
- [x] Remove duplicates filters


kelsos (approved):
@lukicenturi PR looks good so we can merged but the question is would it be possible to enable filter saving on the top filter? would that be much work?

kelsos:
also might need a change log entry for this

PR #8805, opened 2024-10-24
Fix tests in macos

yabirgb:
There was a rounding error in macos because the precission of sqlite is different. Also the history price tests where failing in the CI because we use a cc key there and it wasn't in the cassette


LefterisJP (approved):
LGTM

PR #8813, opened 2024-10-24
Protocol balances fixes

LefterisJP:
[Fix locked SAFE token detection and add Balances usage test](https://github.com/rotki/rotki/commit/cd67561c04af5f5346bcef263ed03fe9f9498610)

Adds the test to detect any unused Balance classes

[Pending AAVE staking rewards should now appear as balances](https://github.com/rotki/rotki/commit/91006af8f11db4c79d477a6940f6ccd29a00c69f) 

This was implemented and tested but not used ...

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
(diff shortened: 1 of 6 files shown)


LefterisJP:
was green but re-pushed since I found some redundant changes

PR #8817, opened 2024-10-25
Update binance CSV import logic

### LINKED ISSUE #8807: Binance CSV not imported correctly.
User @ jesse274523 reported in discord that the Binance CSV that we have to test doesn't work when imported in the app

> Hi Everyone. I am testing out Rotki and trying to import binance CSV files however the import seems to fail on lines will sell, buy or fee operations. I have tried testing with lines from https://github.com/rotki/rotki/blob/develop/rotkehlchen/tests/data/binance_history.csv on version Rotki 1.35.1 rather than my file and the issue still occurs. Is this a known issue?

The logs show

```
DEBUG rotkehlchen.data_import.utils Greenlet-1: Failed to import row 1 {'User_ID': '1', 'UTC_Time': '2020-10-30 14:21:20', 'Account': 'Spot', 'Operation': 'Buy', 'Coin': <Asset identifier:ETH name:Ethereum symbol:ETH>, 'Change': FVal(0.03605), 'Remark': '', '_index': 1} of Binance CSV file. Could not process CSV entry.
DEBUG rotkehlchen.data_import.importers.binance Greenlet-1: Skipped Binance trade rows: [[1604067680, [{'User_ID': '1', 'UTC_Time': '2020-10-30 14:21:20', 'Account': 'Spot', 'Operation': 'Buy', 'Coin': <Asset identifier:ETH name:Ethereum symbol:ETH>, 'Change': FVal(0.03605), 'Remark': '', '_index': 1}]]]
DEBUG rotkehlchen.data_import.importers.binance Greenlet-1: Skipped Binance non-trade rows []
DEBUG rotkehlchen.data_import.importers.binance Greenlet-1: Total found Binance entries: 0
DEBUG rotkehlchen.data_import.importers.binance Greenlet-1: Total skipped Binance csv rows: 1
DEBUG rotkehlchen.data_import.importers.binance Greenlet-1: Binance import stats: []
DEBUG rotkehlchen.data_import.utils Greenlet-1: Imported 0/1 entries.
```

and he also changed the timestamp format `Yes sorry my bad, using  "%Y-%m-%d %H:%M:%S" produced the same results as default`

## Tasks
1. We need to check the logs, why we don't get information about what is going wrong. We need to have logs about the error
2. Check if our CSV can be imported
3. Contact the user if we can't reproduce the error
4. FInd possible reasons to not have the CSV imported correct

He added
imagen

nicholasyoder:
Closes #8807


rotkehlchen/data_import/importers/binance.py
```diff
@@ -191,6 +191,7 @@ def are_entries(self, requested_operations: list) -> bool:
             (keys == {'Buy', 'Sell'} and counted['Buy'] % 2 == 0 and counted['Sell'] % 2 == 0) or
             (keys == {'Buy'} and counted['Buy'] % 2 == 0) or
             (keys == {'Sell'} and counted['Sell'] % 2 == 0) or
+            (keys == {'Buy', 'Sell'} and counted['Buy'] - counted['Sell'] == 0) or
```
yabirgb, line 194: can you add a note in the ones that got deprecated that they aren't valid anymore? I believe we might want to keep them for compatibility but is good to note that is not the latest format

rotkehlchen/data_import/importers/binance.py
```diff
@@ -773,8 +774,9 @@ def _process_multiple_binance_entries(
         """Processes binance entries that are represented with 2+ rows in a csv file.
         Returns Entry type and entries count if any entries were processed. Otherwise, None and 0.
         """
+        operations = [row['Operation'] for row in rows]
         for multiple_entry_class in MULTIPLE_BINANCE_ENTRIES:
-            if multiple_entry_class.are_entries([row['Operation'] for row in rows]):
+            if multiple_entry_class.are_entries(operations):
```
yabirgb, line 779: we were creating the list multiple times here right?

nicholasyoder: Yes, it was, but it was unnecessary since `rows` wasn't changing at all. This way it only creates the list once, plus the list is available to put in the error message.

LefterisJP (approved):
LGTM

PR #8824, opened 2024-10-28
feat: link to history event from missing acquisition

### LINKED ISSUE #8690: Include a link to transactions on the "issues found with generated report" screen
## Abstract

Hello, I am a first-time Rotki user attempting to get it accurately tracking my portfolio so I can do taxes. I've years and years of cruft to get sorted first so naturally, when I went to generate a tax report, I have many issues that I need to manually resolve. In this screen there is a Date/Time reference for presumably when I received the asset with the missing price, but it would also be very nice to have a link to the transaction being referenced for that time as well.

2024-10-07-223338_980x330_scrot

## Motivation

Having a convenient link to the block explorer for these transactions would make it much easier for the user to remember what they were doing and may make it easier to figure out the price they should input, particularly if they were swapping stablecoins for the asset in question.

## Specification

Presumably the Date/Time of the asset is being pulled from the time of a particular transaction and Rotki could insert the link here, like it does on the "newly detected tokens" screen.

lukicenturi:
Closes #8690

image

### DIFF
--- a/frontend/app/src/components/history/events/HistoryEventsView.vue
+++ b/frontend/app/src/components/history/events/HistoryEventsView.vue
@@ -124,6 +124,12 @@ const includes = computed<{ evmEvents: boolean; onlineEvents: boolean }>(() => {
   };
 });
 
+const identifiers = computed<string[] | undefined>(() => {
+  const { identifiers } = get(route).query;
+
+  return identifiers ? [identifiers as string] : undefined;
+});
+
 const { editableItem } = useCommonTableProps<HistoryEventEntry>();
 
 const {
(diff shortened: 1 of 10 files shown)


frontend/app/src/components/history/events/HistoryEventsView.vue
```diff
@@ -332,6 +339,13 @@ async function redecodePageTransactions(): Promise<void> {
   await fetchUndecodedTransactionsStatus();
   await fetchData();
 }
+
+function removeIdentifierParam() {
+  const query = { ...route.query };
+  delete query.identifiers;
+  router.replace({ query });
```
kelsos, line 346: Like I mentioned on discord it would make sense to use `push` here just to keep it in history in case you want to navigate back.

kelsos (approved):
lgtm, if we address the small comment we can merge.

PR #8833, opened 2024-10-30
fix: lazy loader issue where it's not showing on load

lukicenturi:
(Hopefully) fix the lazy loader issue


yabirgb (approved):
Can't confirm it works 100% but I didn't see the error in multiple timest that I tried to reproduce it

PR #8854, opened 2024-11-02
feat: Add assets filtering based on thresold_usd_value

### LINKED ISSUE #7498: Add functionality/checkbox for "Hide assets <1 USD"
## Abstract

Heavy users normally have a lot of dust tokens in their accounts, from reputable currencies that they don't want to blacklist, such as:
<0.01 BNB
<0.01 WETH
<0.01 stETH
etc.

I'd be nice if there is a checkbox on Rotki, similar to the one found on many exchanges, that says "Hide assets <1 USD", so that one doesn't have 300 tokens under Assets.

## Motivation

Nobody cares at a given point in time that they have 0.01 from a certain asset, so hiding these assets would clean up Rotki interface and make it nicer.

## Nice to have

Ideally make the `hide threshold value` configurable, i.e. some folks might want 1 USD, some might want 10 USD or 100 USD, etc.

0saurabh0:
backend part of #7498 

## Checklist

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -733,7 +733,11 @@ class RestAPI:
             status_code = HTTPStatus.CONFLICT
         return api_response(_wrap_in_result(result, message), status_code=status_code)
 
-    def _query_all_exchange_balances(self, ignore_cache: bool) -> dict[str, Any]:
+    def _query_all_exchange_balances(
+            self,
+            ignore_cache: bool,
+            usd_value_threshold: FVal | None = None,
+    ) -> dict[str, Any]:
         final_balances = {}
         error_msg = ''
         for exchange_obj in self.rotkehlchen.exchange_manager.iterate_exchanges():
@@ -754,16 +758,34 @@ class RestAPI:
             result = None
             status_code = HTTPStatus.CONFLICT
         else:
-            result = final_balances
+            # Filter balances by threshold for each exchange
+            if usd_value_threshold is not None:
+                filtered_balances = {}
+                for location, balances in final_balances.items():
+                    filtered_balances[location] = {
+                        asset: balance for asset, balance in balances.items()
+                        if balance.usd_value > usd_value_threshold
+                    }
+                result = filtered_balances
+            else:
+                result = final_balances
             status_code = HTTPStatus.OK
 
         return {'result': result, 'message': error_msg, 'status_code': status_code}
 
     @async_api_call()
-    def query_exchange_balances(self, location: Location | None, ignore_cache: bool) -> dict[str, Any]:  # noqa: E501
+    def query_exchange_balances(
+            self,
+            location: Location | None,
+            ignore_cache: bool,
+            usd_value_threshold: FVal | None = None,
+    ) -> dict[str, Any]:
         if location is None:
             # Query all exchanges
-            return self._query_all_exchange_balances(ignore_cache=ignore_cache)
+            return self._query_all_exchange_balances(
+                ignore_cache=ignore_cache,
+                usd_value_threshold=usd_value_threshold,
+            )
 
         # else query only the specific exchange
         exchanges_list = self.rotkehlchen.exchange_manager.connected_exchanges.get(location)
@@ -784,6 +806,13 @@ class RestAPI:
                     'status_code': HTTPStatus.CONFLICT,
                 }
             balances = combine_dicts(balances, result)
+
+        # Filter balances by threshold for single exchange
+        if usd_value_threshold is not None:
+            balances = {
+                asset: balance for asset, balance in balances.items()
+                if balance.usd_value > usd_value_threshold
+            }
 
         return {
             'result': balances,
@@ -812,6 +841,7 @@ class RestAPI:
             self,
             blockchain: SupportedBlockchain | None,
             ignore_cache: bool,
+            usd_value_threshold: FVal | None = None,
     ) -> dict[str, Any]:
         msg = ''
         status_code = HTTPStatus.OK
@@ -829,6 +859,14 @@ class RestAPI:
             status_code = HTTPStatus.BAD_GATEWAY
         else:
             result = balances.serialize()
+            # Filter balances by threshold after serialization
+            if usd_value_threshold is not None and result is not None:
+                # Filter per chain balances
+                for chain, chain_balances in result.items():
+                    result[chain] = {
+                        asset: balance for asset, balance in chain_balances.items()
+                        if balance['usd_value'] > usd_value_threshold
+                    }
 
         return {'result': result, 'message': msg, 'status_code': status_code}
 
@@ -2049,19 +2087,25 @@ class RestAPI:
 
         return OK_RESULT
 
-    def _get_manually_tracked_balances(self) -> dict[str, Any]:
+    def _get_manually_tracked_balances(self, usd_value_threshold: FVal | None) -> dict[str, Any]:
         db_entries = get_manually_tracked_balances(db=self.rotkehlchen.data.db, balance_type=None)
         balances = process_result(
             {
                 'balances': db_entries,
             },
-
         )
+        # Filter balances if threshold is set
+        if usd_value_threshold is not None:
+            balances['balances'] = [
+                balance for balance in balances['balances']
+                if balance['usd_value'] > usd_value_threshold
+            ]
+
         return _wrap_in_ok_result(balances)
 
     @async_api_call()
-    def get_manually_tracked_balances(self) -> dict[str, Any]:
-        return self._get_manually_tracked_balances()
+    def get_manually_tracked_balances(self, usd_value_threshold: FVal | None) -> dict[str, Any]:
+        return self._get_manually_tracked_balances(usd_value_threshold=usd_value_threshold)
 
     @overload
     def _modify_manually_tracked_balances(  # pylint: disable=unused-argument
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -865,7 +865,8 @@ class RestAPI:
                 for chain, chain_balances in result.items():
                     result[chain] = {
                         asset: balance for asset, balance in chain_balances.items()
-                        if balance['usd_value'] > usd_value_threshold
+                        if 'usd_value' in balance and FVal(balance['usd_value']) >
+                        usd_value_threshold
                     }
 
         return {'result': result, 'message': msg, 'status_code': status_code}
@@ -2098,7 +2099,8 @@ class RestAPI:
         if usd_value_threshold is not None:
             balances['balances'] = [
                 balance for balance in balances['balances']
-                if balance['usd_value'] > usd_value_threshold
+                if 'usd_value' in balance and FVal(balance['usd_value']) >
+                  usd_value_threshold
             ]
 
         return _wrap_in_ok_result(balances)
@@ -2138,7 +2140,7 @@ class RestAPI:
         except TagConstraintError as e:
             return wrap_in_fail_result(str(e), status_code=HTTPStatus.CONFLICT)
 
-        return self._get_manually_tracked_balances()
+        return self._get_manually_tracked_balances(usd_value_threshold=None)
 
     @async_api_call()
     def add_manually_tracked_balances(
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -37,7 +37,7 @@ from rotkehlchen.accounting.export.csv import (
     dict_to_csv_file,
 )
 from rotkehlchen.accounting.pot import AccountingPot
-from rotkehlchen.accounting.structures.balance import Balance, BalanceType
+from rotkehlchen.accounting.structures.balance import Balance, BalanceSheet, BalanceType
 from rotkehlchen.accounting.structures.processed_event import AccountingEventExportType
 from rotkehlchen.accounting.structures.types import ActionType
 from rotkehlchen.api.v1.schemas import TradeSchema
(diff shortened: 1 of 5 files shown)


0saurabh0:
still requires some modifications i.e. to the API

rotkehlchen/tests/unit/test_assets.py
```diff
@@ -935,3 +939,118 @@ def test_all_assets_pagination(globaldb: 'GlobalDBHandler', database: 'DBHandler
...
+        database.add_to_ignored_assets(write_cursor, Asset('XDAI'))
+
+    timestamp = ts_now()
+    balances = []
+    for asset_symbol, usd_value in assets.items():
+        balances.append(DBAssetBalance(
+            category=BalanceType.ASSET,
+            time=timestamp,
+            asset=Asset(asset_symbol),
+            amount=FVal('1'),
```
yabirgb, line 1039: just as a note instead of using `FVal('1')` or `FVal(0)` we use `ONE` and `ZERO` that are predefined constants

rotkehlchen/db/filtering.py
```diff
@@ -1537,6 +1537,7 @@ def make(
 
 class AssetsFilterQuery(DBFilterQuery):
     ignored_assets_handling: IgnoredAssetsHandling = IgnoredAssetsHandling.NONE
+    hide_assets_threshold: FVal | None = None
```
yabirgb, line 1540: this filter is used in contexts where the price doesn't make sense. IT should be a different filter but the approach is not correct

rotkehlchen/globaldb/handler.py
```diff
@@ -372,16 +373,32 @@ def retrieve_assets(userdb: 'DBHandler', filter_query: 'AssetsFilterQuery') -> t
         LEFT JOIN evm_tokens as B ON B.identifier = A.identifier
         LEFT JOIN custom_assets as D ON D.identifier = A.identifier
         """
+        query_to_get_asset_usd_value = """
+        SELECT currency AS id, usd_value FROM timed_balances
+        WHERE timestamp=(SELECT MAX(timestamp) from timed_balances)
+        """
+
```
yabirgb, line 380: I believe this isn't the right approach. This is modifying af unction that is used only in the search of assets when you want to find them in the list of all assets.

Undo this change

yabirgb (dismissed):
The approach here isn't correct. Let me give you some feedback that can help:

The issue tries to solve the problem of assets displayed in the dashboard with low value but that isn't zero. The dashboard takes the balances from 3 endpoints

```
/api/1/balances/blockchains/eth (and variants for the different chains)
/balances/manual
/exchanges/balances
```

you can find what resources answer those endpoints in the server.py file.

- You would need to create a new Schema `AssetValueThreshold` and the schemas on those resources shouls inherit from it.
- Then the different functions that process the logic will iterate over the balances that they return filtering the results with this amount. For example in `query_exchange_balances` instead of 

```python
            result, msg = exchange.query_balances(ignore_cache=ignore_cache)
            if result is None:
                return {
                    'result': result,
                    'message': msg,
                    'status_code': HTTPStatus.CONFLICT,
                }
            balances = combine_dicts(balances, result)
```

you will do something like

```python
            result, msg = exchange.query_balances(ignore_cache=ignore_cache)
            if result is None:
                return {
                    'result': result,
                    'message': msg,
                    'status_code': HTTPStatus.CONFLICT,
                }
            balances = combine_dicts(balances, {asset: balance for asset, value in result.items() if balance.usd_value > threshold})
```

A similar approach should be used by the other endpoints. This should have a test that aggregates manual balances, exchange balances and blockchain balances from different chains and you get the expected assets with enouigh USD value. The `Balance` structure holds both the usd value and the asset amount

0saurabh0:
Thanks for the review!
 > that aggregates manual balances, exchange balances and blockchain balances from different chains and you get the expected assets with enough USD value. 
 
 Need clarification on how to apply the USD value threshold across different balance types (manual, exchange, blockchain). Should the threshold filtering be applied individually to each balance type before aggregation, or should we aggregate all balances first and then apply the threshold to the total?

For example, with a threshold of $10:

   Manual balance: $4
    Exchange balance: $6
    Blockchain balance: $20

 - Option 1 (Aggregate first): Add all balances first: $4 + $6 + $20 = $30. Since $30 > $10, all balances would count towards the total.

 - Option 2 (Filter first): Filter each balance type individually by the $10 threshold. Here, only the blockchain balance ($20) meets the threshold, so only it would contribute to the total.

yabirgb:
Good question @0saurabh0 . I believe it doesn´t make sense to calculate it aggregated since the frontend aggregates the information from different endpoints. You are right that it can affect the final balances but I image this filter used to filter out things like < .10 $ so if you are missing a few cents of a coin in a exchange it doesn´t really matter.  i would go with the second approach. I'll ask other too but 2 is what I would do

0saurabh0:
>it can affect the final balances 

So, If we go with approach 2 then cant we modify the `query_balances` function to consider the `threshold value`, that way the final/total Balance will be more accurate too?

yabirgb:
> > it can affect the final balances
> 
> So, If we go with approach 2 then cant we modify the [query_balances](https://github.com/rotki/rotki/blob/397130c084872c56adcdc2a8624f4415641fca55/rotkehlchen/rotkehlchen.py#L967) function to consider the `threshold value`, that way the final/total Balance will be more accurate too?

I haven´t checked exhaustively the code around since I'm traveling but there are other functions that are called in the different endpoints for the balances that I pointed before. The one you point is only used for snapshots, there you shouldn´t be interested in the threshold. I would create a function filter that reads the dict that will be returned and removes from it values that are under the threshold. The threshold value should be in the schemas for the api endpoint and mentioned before. Let us know in discord if you need help with it

rotkehlchen/tests/api/test_balances.py
```diff
@@ -973,3 +983,108 @@ def query_blockchain_balance(num: int) -> Any:
...
+) -> None:
+    """Test that balance filtering by USD value threshold works for all balance types"""
+    rotki = rotkehlchen_api_server_with_exchanges.rest_api.rotkehlchen
+
+    manually_tracked_balances = [
+        ManuallyTrackedBalance(
+            identifier=-1,
+            asset=A_EUR,
+            label='Small EUR',
+            amount=FVal('1'),
```
yabirgb, line 1004: use `ONE` (we have this constant) instead of calling FVal

```suggestion
            amount=ONE,
```

rotkehlchen/tests/api/test_balances.py
```diff
@@ -973,3 +983,108 @@ def query_blockchain_balance(num: int) -> Any:
...
+            amount=FVal('10'),
+            location=Location.BANKS,
+            tags=None,
+            balance_type=BalanceType.ASSET,
+        ),
+        ManuallyTrackedBalance(
+            identifier=-3,
+            asset=A_BTC,
+            label='Large BTC',
+            amount=FVal('1'),
```
yabirgb, line 1022: same here

rotkehlchen/tests/api/test_balances.py
```diff
@@ -973,3 +983,108 @@ def query_blockchain_balance(num: int) -> Any:
...
+            params={'usd_value_threshold': threshold},
+        )
+        manual_result = assert_proper_sync_response_with_result(response)
+
+        # Assert blockchain balances
+        for chain in blockchain_result['per_account']:
+            for account in blockchain_result['per_account'][chain]:
+                assets = blockchain_result['per_account'][chain][account]['assets']
+                for balance in assets.values():
+                    assert FVal(balance['usd_value']) > FVal(threshold)
```
yabirgb, line 1081: can you add asserts to ensure that the loops are not empty? It could be that your test passes because no data is iterated. You can assert on len(data)

rotkehlchen/tests/api/test_balances.py
```diff
@@ -973,3 +983,108 @@ def query_blockchain_balance(num: int) -> Any:
...
+        manually_tracked_balances=manually_tracked_balances,
+        eth_balances=['100000000000000000', '200000000000000000'],  # 0.1 and 0.2 ETH
+        token_balances={
+            A_DAI.resolve_to_evm_token(): ['5000000000000000000', '0'],  # 5 DAI and 0 DAI
+            A_USDC.resolve_to_evm_token(): ['1000000', '2000000'],  # 1 USDC and 2 USDC
+        },
+    )
+
+    threshold = 10  # Set threshold to 10 USD
+
```
yabirgb, line 1042:
```suggestion
```

remove the empty line

rotkehlchen/api/v1/schemas.py
```diff
@@ -1489,6 +1495,10 @@ class AllBalancesQuerySchema(AsyncQueryArgumentSchema):
     ignore_cache = fields.Boolean(load_default=False)
 
 
+class ManualBalanceQuerySchema(AsyncQueryArgumentSchema, AssetValueThresholdSchema):
+    """Schema for querying manual balances with optional USD threshold filtering"""
```
yabirgb, line 1499: nice well done

rotkehlchen/api/rest.py
```diff
@@ -829,6 +859,15 @@ def query_blockchain_balances(
             status_code = HTTPStatus.BAD_GATEWAY
         else:
             result = balances.serialize()
+            # Filter balances by threshold after serialization
+            if usd_value_threshold is not None and result is not None:
+                # Filter per chain balances
+                for chain, chain_balances in result.items():
+                    result[chain] = {
+                        asset: balance for asset, balance in chain_balances.items()
+                        if 'usd_value' in balance and FVal(balance['usd_value']) >
```
yabirgb, line 868: You don't need `'usd_value' in balance` since the serialization always adds it. Otherwise is a programming error but correct me if something is different here

```suggestion
                        if FVal(balance['usd_value']) > usd_value_threshold
```

rotkehlchen/api/rest.py
```diff
@@ -2049,19 +2088,26 @@ def remove_chain_type_accounts(
...
                 'balances': db_entries,
             },
-
         )
+        # Filter balances if threshold is set
+        if usd_value_threshold is not None:
+            balances['balances'] = [
+                balance for balance in balances['balances']
+                if 'usd_value' in balance and FVal(balance['usd_value']) >
+                  usd_value_threshold
```
yabirgb, line 2103: same comment. The usd_value should be there so no need to check for it. Maybe you can change this whole logic and do it before the call to `balances = process_result(`. The `db_entries` is a list of `ManuallyTrackedBalanceWithValue` and you can check for `ManuallyTrackedBalanceWithValue.value.usd_value`

rotkehlchen/api/rest.py
yabirgb: I just realized. Why not do this for loop before calling to serialize? You avoid serializing entries that you will discard later

rotkehlchen/tests/api/test_balances.py
```diff
@@ -973,3 +983,108 @@ def query_blockchain_balance(num: int) -> Any:
...
+            params={'usd_value_threshold': threshold},
+        )
+        manual_result = assert_proper_sync_response_with_result(response)
+
+        # Assert blockchain balances
+        for chain in blockchain_result['per_account']:
+            for account in blockchain_result['per_account'][chain]:
+                assets = blockchain_result['per_account'][chain][account]['assets']
+                for balance in assets.values():
+                    assert FVal(balance['usd_value']) > FVal(threshold)
```
yabirgb, line 1081: What I meant here is that it could happen that the assert passed because you were testing a thing that was empty and therefore the assert never gets executed

```suggestion
        assert len(blockchain_result['per_account']) != 0
        for chain in blockchain_result['per_account']:
            assert len(blockchain_result['per_account'][chain]) != 0
            for account in blockchain_result['per_account'][chain]:
                assets = blockchain_result['per_account'][chain][account]['assets']
                assert len(assets) != 0
                for balance in assets.values():
                    assert FVal(balance['usd_value']) > FVal(threshold)
```

rotkehlchen/tests/api/test_balances.py
```diff
@@ -973,3 +983,108 @@ def query_blockchain_balance(num: int) -> Any:
...
+
+        # Assert exchange balances
+        for exchange in exchange_result:
+            for balance in exchange_result[exchange].values():
+                assert FVal(balance['usd_value']) > FVal(threshold)
+
+        # Assert manual balances
+        for balance in manual_result['balances']:
+            assert FVal(balance['usd_value']) > FVal(threshold)
```
yabirgb, line 1090:
```suggestion
        # Assert manual balances
        assert len(manual_result['balances']) != 0
        for balance in manual_result['balances']:
            assert FVal(balance['usd_value']) > FVal(threshold)
```

rotkehlchen/tests/api/test_balances.py
```diff
@@ -973,3 +983,108 @@ def query_blockchain_balance(num: int) -> Any:
...
+        for chain in blockchain_result['per_account']:
+            for account in blockchain_result['per_account'][chain]:
+                assets = blockchain_result['per_account'][chain][account]['assets']
+                for balance in assets.values():
+                    assert FVal(balance['usd_value']) > FVal(threshold)
+
+        # Assert exchange balances
+        for exchange in exchange_result:
+            for balance in exchange_result[exchange].values():
+                assert FVal(balance['usd_value']) > FVal(threshold)
```
yabirgb, line 1086:
```suggestion
        assert len(exchange_result) != 0
        for exchange in exchange_result:
            assert len(exchange_result[exchange]) != 0
            for balance in exchange_result[exchange].values():
                assert FVal(balance['usd_value']) > FVal(threshold)
```

rotkehlchen/tests/api/test_balances.py
```diff
@@ -973,3 +983,108 @@ def query_blockchain_balance(num: int) -> Any:
...
+
+        # Test manual balances
+        response = requests.get(
+            api_url_for(
+                rotkehlchen_api_server_with_exchanges,
+                'manuallytrackedbalancesresource',
+            ),
+            params={'usd_value_threshold': threshold},
+        )
+        manual_result = assert_proper_sync_response_with_result(response)
```
yabirgb, line 1074: Maybe you can do this cleaner with a for loop?

```suggestion
        results = []
        for endpoint in (
            'blockchainbalancesresource',
            'exchangebalancesresource',
            'manuallytrackedbalancesresource'
        ):
            response = requests.get(
                api_url_for(
                     rotkehlchen_api_server_with_exchanges,
                    endpoint,
               ),
               params={'usd_value_threshold': threshold},
           )
           results.append(assert_proper_sync_response_with_result(response))
       
       blockchain_result, exchange_result, manual_result = results
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -177,6 +177,12 @@
 log = RotkehlchenLogsAdapter(logger)
 
 
+class AssetValueThresholdSchema(Schema):
+    usd_value_threshold = AmountField(
+        load_default=None,
+    )
```
yabirgb, line 183: We have those endpoints documented in `docs/api.rst`. Can you also update the documentation there? I forgot to tell you this in the previous review. Since you added a new field it should be documented

```suggestion
class AssetValueThresholdSchema(Schema):
    usd_value_threshold = AmountField(load_default=None)
```

rotkehlchen/api/rest.py
```diff
@@ -820,14 +850,36 @@ def query_blockchain_balances(
...
+                                asset: balance for asset, balance in account_data.assets.items()
+                                if balance.usd_value > usd_value_threshold
+                            }
+                            if filtered_assets:
+                                new_balance_sheet = BalanceSheet()
+                                new_balance_sheet.assets = filtered_assets
+                                filtered_balances[account] = new_balance_sheet
+                        elif isinstance(account_data, Balance):
+                            if account_data.usd_value > usd_value_threshold:
+                                filtered_balances[account] = account_data
```
yabirgb, line 870: note: this is the case for btc and bch

yabirgb: can you add it in a comment?

LefterisJP (approved):
LGTM. My questions are mostly for our own team and typos/nitpicks I can handle in my next PR.

This PR has been open for quite long so I will merge it.

Thank you.

PR #8856, opened 2024-11-03
Fix extrafi balance query for multiple pools of same token

LefterisJP:
If a user had interacted with multiple pools of the same token only the last pool they ever interacted with would be considered for balance query.


yabirgb (approved):
lgtm

PR #8863, opened 2024-11-04
Fix addressbook entry deletion for blank label

nicholasyoder:
Fixes problem mentioned by @lukicenturi in [discord](https://discord.com/channels/657906918408585217/683245846070165504/1302985079110500414)

### DIFF
--- a/rotkehlchen/db/dbhandler.py
+++ b/rotkehlchen/db/dbhandler.py
@@ -1145,21 +1145,20 @@ class DBHandler:
             account_data: list[BlockchainAccountData],
     ) -> None:
         # Insert the blockchain account addresses and labels to the DB
-        blockchain_accounts_query = []
-        address_book_query = []
+        blockchain_accounts_query, bindings_to_update = [], []
         for entry in account_data:
             blockchain_accounts_query.append((entry.chain.value, entry.address))
-            if entry.label is not None:
-                address_book_query.append((entry.address, entry.chain.value, entry.label))
+            if entry.label:
+                bindings_to_update.append((entry.address, entry.chain.value, entry.label))
         try:
             write_cursor.executemany(
                 'INSERT INTO blockchain_accounts(blockchain, account) VALUES (?, ?)',
                 blockchain_accounts_query,
             )
-            if len(address_book_query) > 0:
+            if len(bindings_to_update) > 0:
                 write_cursor.executemany(
                     'INSERT OR REPLACE INTO address_book(address, blockchain, name) VALUES (?, ?, ?)',  # noqa: E501
-                    address_book_query,
+                    bindings_to_update,
                 )
         except sqlcipher.IntegrityError as e:  # pylint: disable=no-member
             raise InputError(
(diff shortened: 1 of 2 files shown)


rotkehlchen/db/addressbook.py
```diff
@@ -94,10 +94,11 @@ def add_addressbook_entries(
                         (entry.address, ANY_BLOCKCHAIN_ADDRESSBOOK_VALUE),
                     )
 
-                write_cursor.execute(
-                    'INSERT INTO address_book (address, name, blockchain) VALUES (?, ?, ?)',
-                    entry.serialize_for_db(),
-                )
+                if entry.name != '':  # accounts with no name should not have an addressbook entry
```
prettyirrelevant, line 97: i don't think this is the best approach. looking at the type of `Addressbook` and where it's used(schemas, etc.), i don't think empty strings are allowed.

rotkehlchen/db/dbhandler.py
```diff
@@ -1139,33 +1140,44 @@ def add_multiple_location_data(self, write_cursor: 'DBCursor', location_data: li
...
 
+    def _get_addressbook_entries_from_account_data(
+            self,
+            account_data: list[BlockchainAccountData],
+    ) -> list[AddressbookEntry]:
+        """Get list of addressbook entries from the accounts in account_data"""
+        return [
+            AddressbookEntry(
+                address=entry.address,
+                name=entry.label if entry.label is not None else '',
```
prettyirrelevant, line 1151: this is an anti-pattern of how AddressbookEntry is used. `name` is never null. just return entries that have label or am i missing something?

prettyirrelevant (commented):
thanks. @nicholasyoder left some comments.

this is missing a regression test to ensure this bug does not reoccur.

rotkehlchen/db/addressbook.py
nicholasyoder: As discussed in discord, this has been updated to use a different approach that doesn't involve using the functions from addressbook.

rotkehlchen/db/dbhandler.py
nicholasyoder: Same as https://github.com/rotki/rotki/pull/8863#discussion_r1828358195

rotkehlchen/db/dbhandler.py
```diff
@@ -1145,21 +1145,27 @@ def add_blockchain_accounts(
             account_data: list[BlockchainAccountData],
     ) -> None:
         # Insert the blockchain account addresses and labels to the DB
-        blockchain_accounts_query = []
-        address_book_query = []
+        blockchain_accounts_query, bindings_to_update, bindings_to_delete = [], [], []
```
prettyirrelevant, line 1148: you don't need bindings_to_update/delete for this method as it is used to **only** add an account.

plus, you only add to addressbook iff a label is provided i.e it's not falsy.

rotkehlchen/db/dbhandler.py
```diff
@@ -1145,21 +1145,27 @@ def add_blockchain_accounts(
...
         for entry in account_data:
             blockchain_accounts_query.append((entry.chain.value, entry.address))
-            if entry.label is not None:
-                address_book_query.append((entry.address, entry.chain.value, entry.label))
+            if entry.label not in [None, '']:
```
prettyirrelevant, line 1151:
```suggestion
            if entry.label:
```

this way, you don't need the else branch.

rotkehlchen/db/dbhandler.py
```diff
@@ -1180,22 +1186,35 @@ def edit_blockchain_accounts(
...
-            (entry.address, entry.chain.value, entry.label)
-            for entry in account_data if entry.label is not None
-        ]
-        if len(bindings) > 0:
+        bindings_to_update, bindings_to_delete = [], []
+        for entry in account_data:
+            if entry.label not in [None, '']:
+                bindings_to_update.append((entry.address, entry.chain.value, entry.label))
+            else:
+                bindings_to_delete.append((entry.address, entry.chain.value))
```
prettyirrelevant, line 1194: you can use a list comprehension instead.

rotkehlchen/db/dbhandler.py
```diff
@@ -1180,22 +1186,35 @@ def edit_blockchain_accounts(
...
+
+        modified_count = 0
```
prettyirrelevant, line 1196: it might be better to use a deleted_count & modified_count to make debugging easier.

prettyirrelevant (commented):
minor comments & questions. thanks.

rotkehlchen/db/dbhandler.py
nicholasyoder: As mentioned in discord, using list comprehension would actually make it run two loops so this is probably better as it is.

rotkehlchen/db/dbhandler.py
```diff
@@ -1145,21 +1145,20 @@ def add_blockchain_accounts(
...
         # Insert the blockchain account addresses and labels to the DB
-        blockchain_accounts_query = []
-        address_book_query = []
+        blockchain_accounts_query, bindings_to_update = [], []
         for entry in account_data:
             blockchain_accounts_query.append((entry.chain.value, entry.address))
-            if entry.label is not None:
-                address_book_query.append((entry.address, entry.chain.value, entry.label))
+            if entry.label:
+                bindings_to_update.append((entry.address, entry.chain.value, entry.label))
```
prettyirrelevant, line 1152: nitpick: maybe `bindings_to_update` isn't the best variable name here.

rotkehlchen/db/dbhandler.py
```diff
@@ -1180,22 +1179,36 @@ def edit_blockchain_accounts(
...
+            )
+            deleted_count += write_cursor.rowcount
+
+        if modified_count != len(bindings_to_update) or deleted_count != len(bindings_to_delete):
+            msg = (
+                f'When updating blockchain accounts expected {len(bindings_to_update)}'
+                f'modified and {len(bindings_to_delete)} deleted, but instead there were'
+                f'{modified_count} modified and {deleted_count} deleted. Should not happen.'
+            )
+            log.error(msg)
```
prettyirrelevant, line 1210: nitpick: combining the logs doesn't seem appropriate.

prettyirrelevant (commented):
mostly nitpicks. if test suite passes, i'll ping one of yabir/lefteris for a final review.

lukicenturi:
This fix works for me

rotkehlchen/db/dbhandler.py
LefterisJP: + 1 on this

LefterisJP (approved):
LGTM as we don't want to waste time now and the failing test shold be handled when assets are pulled in master

PR #8867, opened 2024-11-04
Change logic for connecting nodes

yabirgb:
This PR changes the approach for connecting to web3 nodes. Instead of connecting all the nodes for a chain at the start now it connects to nodes on demand. If for chain OP we have nodes `A, B, C` and we call a query with `default_order`
it will create a random order `[B, A, C]`. Then it will iterate and connect to `B` first. Then in the next call the order will be `[C, A, B]` and it will call C first making the connection and having `B and C` as connected nodes for now. A will be offline at this point because it hasn't been requested yet.

This PR also delays the task for sumbmitting analytics stuff

### DIFF
--- a/rotkehlchen/chain/aggregator.py
+++ b/rotkehlchen/chain/aggregator.py
@@ -91,7 +91,6 @@ from rotkehlchen.types import (
     CHAIN_IDS_WITH_BALANCE_PROTOCOLS,
     CHAINS_WITH_CHAIN_MANAGER,
     EVM_CHAIN_IDS_WITH_TRANSACTIONS,
-    EVM_CHAINS_WITH_TRANSACTIONS_TYPE,
     SUPPORTED_CHAIN_IDS,
     SUPPORTED_EVM_CHAINS_TYPE,
     SUPPORTED_EVM_EVMLIKE_CHAINS,
@@ -315,12 +314,6 @@ class ChainsAggregator(CacheableMixIn, LockableQueryMixIn):
         }
         self.chain_modify_append: dict[SupportedBlockchain, Callable[[SupportedBlockchain, BlockchainAddress], None]] = {  # noqa: E501
             SupportedBlockchain.ETHEREUM: self._append_eth_account_modification,  # type:ignore
-            SupportedBlockchain.OPTIMISM: self._append_evm_account_modification,  # type:ignore
-            SupportedBlockchain.POLYGON_POS: self._append_evm_account_modification,  # type:ignore
-            SupportedBlockchain.ARBITRUM_ONE: self._append_evm_account_modification,  # type:ignore
-            SupportedBlockchain.BASE: self._append_evm_account_modification,  # type:ignore
-            SupportedBlockchain.GNOSIS: self._append_evm_account_modification,  # type:ignore
-            SupportedBlockchain.SCROLL: self._append_evm_account_modification,  # type:ignore
         }
         self.chain_modify_remove: dict[SupportedBlockchain, Callable[[SupportedBlockchain, BlockchainAddress], None]] = {  # noqa: E501
             SupportedBlockchain.ETHEREUM: self._remove_eth_account_modification,  # type:ignore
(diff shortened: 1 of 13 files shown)


rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -415,6 +415,7 @@ def attempt_connect(
         except requests.exceptions.RequestException:
             message = f'Failed to connect to {self.chain_name} node {node} at endpoint {rpc_endpoint}'  # noqa: E501
             log.warning(message)
+            self.failed_nodes.add(node.name)
```
LefterisJP, line 418: Also add a small comment explaining what this set is. Would you ever be able to reset this set/ remove somehting out of this so long as the app runs?

```suggestion
            self.failed_to_connect_nodes.add(node.name)
```

rotkehlchen/chain/aggregator.py
```diff
@@ -808,9 +808,7 @@ def _append_evm_account_modification(
             address: ChecksumEvmAddress,  # pylint: disable=unused-argument
     ) -> None:
         """Extra code to run when a non-eth but evm account addition happens"""
-        # If this is the first account added, connect to all relevant nodes
-        chain_manager = self.get_chain_manager(blockchain)
-        chain_manager.node_inquirer.maybe_connect_to_nodes(when_tracked_accounts=False)
+        return
```
LefterisJP, line 811: Then shouldn't this funciton and whenever this function is also used be removed?

We have this code:

```python
        self.chain_modify_append: dict[SupportedBlockchain, Callable[[SupportedBlockchain, BlockchainAddress], None]] = {  # noqa: E501
            SupportedBlockchain.ETHEREUM: self._append_eth_account_modification,  # type:ignore
            SupportedBlockchain.OPTIMISM: self._append_evm_account_modification,  # type:ignore
            SupportedBlockchain.POLYGON_POS: self._append_evm_account_modification,  # type:ignore
            SupportedBlockchain.ARBITRUM_ONE: self._append_evm_account_modification,  # type:ignore
            SupportedBlockchain.BASE: self._append_evm_account_modification,  # type:ignore
            SupportedBlockchain.GNOSIS: self._append_evm_account_modification,  # type:ignore
            SupportedBlockchain.SCROLL: self._append_evm_account_modification,  # type:ignore
        }
```

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -509,8 +510,18 @@ def _query(self, method: Callable, call_order: Sequence[WeightedNode], **kwargs:
...
             web3node = self.web3_mapping.get(node_info, None)
-            if web3node is None and node_info.name != self.etherscan_node_name:
-                continue
+            if (
+                web3node is None and
+                node_info.name != self.etherscan_node_name and
+                node_info.name not in self.failed_nodes
+            ):
+                success, _ = self.attempt_connect(node=node_info)
+                if success is False:
```
LefterisJP, line 519: shouldn't it be added to failed to connect nodes here?

rotkehlchen/rotkehlchen.py
```diff
@@ -350,7 +350,7 @@ def unlock_user(
             settings = self.get_settings(cursor)
             CachedSettings().initialize(settings)  # initialize with saved DB settings
             self.greenlet_manager.spawn_and_track(
-                after_seconds=None,
+                after_seconds=100,
```
LefterisJP, line 353: Why this? You wuld definitely need a comment here too after you explain to me

rotkehlchen/chain/evm/node_inquirer.py
yabirgb: At the moment no, we would need to restart. We can add additional logic maybe to re-check

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -225,7 +225,7 @@ def __init__(
         # A cache for erc20 and erc721 contract info to not requery the info
         self.contract_info_erc20_cache: LRUCacheWithRemove[ChecksumEvmAddress, dict[str, Any]] = LRUCacheWithRemove(maxsize=1024)  # noqa: E501
         self.contract_info_erc721_cache: LRUCacheWithRemove[ChecksumEvmAddress, dict[str, Any]] = LRUCacheWithRemove(maxsize=512)  # noqa: E501
-        self.maybe_connect_to_nodes(when_tracked_accounts=True)
+        self.failed_to_connect_nodes: set[str] = set()
```
LefterisJP, line 228: please add a description either above this or in the docstring of the constructor what this is for and underline the answer to my question from before. Specifically state that if a node is added here only a restart takes it out. No reconnect attempt happens.

And probably we would need a note to handle this as perhaps it makes sense to try again once in an hour or something?

yabirgb: https://github.com/orgs/rotki/projects/11?pane=issue&itemId=86405753

LefterisJP (approved):
lgtm

PR #8872, opened 2024-11-05
Solve circular import

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/assets/resolver.py
+++ b/rotkehlchen/assets/resolver.py
@@ -8,6 +8,7 @@ from rotkehlchen.utils.data_structures import LRUCacheLowerKey
 
 if TYPE_CHECKING:
     from rotkehlchen.assets.asset import (
+        Asset,
         AssetWithNameAndType,
         AssetWithOracles,
         AssetWithSymbol,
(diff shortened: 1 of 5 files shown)


LefterisJP (commented):
why not put this in master?

rotkehlchen/assets/resolver.py
```diff
@@ -25,20 +27,30 @@
...
         It always uses the GlobalDB to resolve assets
         """
         if AssetResolver.__instance is not None:
             return AssetResolver.__instance
 
+        assert globaldb is not None
+        assert constant_assets is not None
         AssetResolver.__instance = object.__new__(cls)
+        AssetResolver._globaldb = globaldb
+        AssetResolver._constant_assets = constant_assets
```
LefterisJP, line 53: this is duplicating constant assets in memory right? Shouldn't we somehow make sure the set is only kept here as it's kept here only?

LefterisJP: it's not as it's a mutable type and as such is just passed by reference

LefterisJP (approved):
LGTM

PR #8875, opened 2024-11-06
Avoid loop querying logs from gnosis with 1rpc

yabirgb:
Closes #(issue_number)

## Checklist


rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -133,7 +133,12 @@ def _query_web3_get_logs(
                 msg = 'query returned more than 10000 results'
 
             # errors from: https://infura.io/docs/ethereum/json-rpc/eth-getLogs
-            if msg in {'query returned more than 10000 results', 'query timeout exceeded'}:
+            if msg in {
+                'query returned more than 10000 results',
+                'eth_getLogs is limited to a 1000 blocks range',  # seen in https://1rpc.io/gnosis
+            }:
+                block_range = 999
```
LefterisJP, line 140: Why force 999? If the node can handle less than 10k then let it return 9999. If the node can handle less than 1k then do 999. You do the conditional check anyway. Use it.

rotkehlchen/tests/unit/test_inquirer.py
```diff
@@ -943,3 +948,51 @@ def test_fiat_to_fiat(inquirer):
             base=A_USD.resolve_to_fiat_asset(),
             quote=A_EUR.resolve_to_fiat_asset(),
         )
+
+
+@pytest.mark.vcr
+def test_errors_web3_logs():
+    """
+    1rpc.io/gnosis has a limit of 1000 logs per query. This test
+    ensure that the amount of queries made to the node is the minimum
```
LefterisJP, line 957:
```suggestion
    ensures that the amount of queries made to the node is the minimum
```

LefterisJP (approved):
typo nitpick but lgtm otherwise

PR #8877, opened 2024-11-06
Message for invalid gnosispay token

LefterisJP:


rotkehlchen/chain/arbitrum_one/modules/umami/balances.py
```diff
@@ -158,10 +158,9 @@ def query_balances(self) -> 'BalancesSheetType':
                     if balance == 0:
                         continue
 
-                    gm_vault_token_address = gm_vault_addresses[idx]
                     if (gm_vault_token := GlobalDBHandler().get_evm_token(
-                        address=gm_vault_token_address,
-                        chain_id=self.evm_inquirer.chain_id,
+                            address=(gm_vault_token_address := gm_vault_addresses[idx]),
+                            chain_id=self.evm_inquirer.chain_id,
```
yabirgb, line 163: now that I see this maybe is better to use the EvmToken class (uses the asset resolver and doesn't go to the db)

yabirgb (approved):
lgtm

PR #8894, opened 2024-11-13
add balancer v1 pools for gnosis and other supported evm chains.

### LINKED ISSUE #7405: Support balancer pools in gnosis chain and more
## Abstract

Support balancer pool decoding in gnosis chain and all other chains we don't support yet. A related issue is to automatically detect LP positions and their value: https://github.com/rotki/rotki/issues/5208

## Motivation

Gnosis chain is getting a lot of traction right now and we should support more protocols there.

prettyirrelevant:
closes #7405

implements balancer v1 protocol support across ethereum, arbitrum and gnosis chains

> [!NOTE]
> lp token balances are fetched via balancer api after multiple attempts at pure on-chain integration proved unreliable 

## implementation details
- [x] integrated balancer api for pool data querying
- [x] implemented common balancer v1 decoder for multi-chain support
- [x] enhanced join/exit pool transaction decoding
- [x] removed balancer ethereum module implementation and tests

## api integration
**endpoint**: https://api-v3.balancer.fi

> [!IMPORTANT]
> - balancer v2 protocol support scheduled for follow-up pr
> - frontend requires updates to remove balancer ethereum module dependencies
> - please flag any lingering unused types from experimenting :(


rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,223 @@
...
+                    'version': version,
+                },
+                'query': GET_POOLS_COUNT_QUERY,
+            },
+            timeout=CachedSettings().get_timeout_tuple(),
+        )
+        response.raise_for_status()
+        data = response.json()
+    except (requests.exceptions.RequestException, JSONDecodeError) as e:
+        raise RemoteError(f'Balancer pools count query failed due to {e!s}') from e
```
yabirgb, line 124: can you check what I'm going to say but:

- JSONDecodeError is different from requests.exception.JSONDecodeError
- you can use requests.RequestException
- requests.RequestException is a parent class of requests.JSONDEcoderError

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,223 @@
...
+
+    return data['data']['poolGetPoolsCount']
```
yabirgb, line 126: this can raise keyError

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,223 @@
...
+            value=str(latest_pools_count),
+            timestamp=now,
+        )
+
+
+def _query_balancer_pools_count(chain: 'ChainID', version: Literal[1, 2]) -> int:
+    """Fetch the total number of pools for the specified chain and protoco
+    May raise:
+        - RemoteError
+        - DeserializationError
```
yabirgb, line 107: Where?

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,223 @@
...
+        database: 'DBHandler',
+        version: Literal[1, 2],
+) -> list[dict[str, Any]]:
+    """Fetches and processes Balancer pool data from API, storing token info in database.
+    Returns list of pool addresses. Supports V1 and V2 Balancer protocols.
+
+    May raise:
+        - RemoteError
+    """
+
```
yabirgb, line 140:
```suggestion
```

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,223 @@
...
+        if 'errors' in data:
+            raise RemoteError(f"Balancer pools query failed: {data['errors']}")
+
+        pools_chunk = data['data']['poolGetPools']
+        all_pools.extend(pools_chunk)
+
+        if len(pools_chunk) < BALANCER_API_CHUNK_SIZE:
+            break
+
+        skip += BALANCER_API_CHUNK_SIZE
```
yabirgb, line 172: can you use as varaible offset instead of skip?

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,223 @@
...
+    return all_pools
+
+
+def _create_or_update_pool_token(
+        db: 'DBHandler',
+        chain_id: 'ChainID',
+        protocol: str,
+        pool: dict,
+) -> 'EvmToken':
+    underlying_tokens = [
```
yabirgb, line 183: in this function you have several key Errors. Also can you use `deserialize_int` for the decimals?

rotkehlchen/chain/evm/decoding/balancer/v1/balances.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+
+    def query_balances(self) -> 'BalancesSheetType':
+        balances: BalancesSheetType = defaultdict(BalanceSheet)
+        if len(addresses_with_deposits := list(self.addresses_with_activity(event_types={
+            (HistoryEventType.DEPOSIT, HistoryEventSubType.DEPOSIT_ASSET),
+        }))) == 0:
+            return balances
+
+        pool_tokens = set(GlobalDBHandler().get_addresses_by_protocol(
```
yabirgb, line 56:
```suggestion
        pool_tokens = set(GlobalDBHandler.get_addresses_by_protocol(
```

rotkehlchen/chain/evm/decoding/balancer/v1/balances.py
```diff
@@ -0,0 +1,99 @@
...
+                    token = Asset(evm_address_to_identifier(
+                        address=token_address,
+                        token_type=EvmTokenKind.ERC20,
+                        chain_id=self.evm_inquirer.chain_id,
+                    ))
+                    balances[address].assets[token] += Balance(
+                        amount=FVal(lp_balance['totalBalance']),
+                        usd_value=FVal(lp_balance['totalBalance']) * FVal(lp_balance['tokenPrice']),  # noqa: E501
+                    )
+            except (RemoteError, JSONDecodeError) as e:
```
yabirgb, line 95: afaik nothing can raise RemoteError here. it is json.RequestException

rotkehlchen/globaldb/cache.py
```diff
@@ -16,6 +18,9 @@
 )
 from rotkehlchen.utils.misc import ts_now
 
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
```
yabirgb, line 23: why? not used

rotkehlchen/tests/unit/decoders/test_balancer.py
```diff
@@ -331,3 +332,310 @@ def test_balancer_trade(ethereum_inquirer, ethereum_accounts):
...
+            notes=f'Receive {receive_amt} BCoW-80QQQ-20WETH from a Balancer v1 pool',
+            address=balancer_qqq_weth_pool_token.evm_address,
+            extra_data={'deposit_events_num': 2},
+        ), EvmEvent(
+            tx_hash=tx_hash,
+            sequence_index=166,
+            timestamp=timestamp,
+            location=Location.ETHEREUM,
+            event_type=HistoryEventType.DEPOSIT,
+            event_subtype=HistoryEventSubType.DEPOSIT_ASSET,
```
yabirgb, line 405: the order should be the other way. First deposit and then receive wrapped. THe logic is always out first and then in

rotkehlchen/chain/evm/decoding/balancer/v1/decoder.py
```diff
@@ -0,0 +1,391 @@
...
+                CPT_BALANCER_V1,
+                self.evm_inquirer.chain_id.serialize_for_db(),
+            )
+            cursor.execute(f'SELECT COUNT(et.identifier) {query_body}', bindings)
+            if cursor.fetchone()[0] == len(self.pools):
+                return None  # up to date
+
+            cursor.execute(f'SELECT et.protocol, et.address {query_body}', bindings)
+            for row in cursor:
+                self.pools.add(string_to_evm_address(row[1]))
```
yabirgb, line 360: you reset it always right? you cna use set comprehnesion and self.pools = ...

yabirgb (commented):
This changes affects the frontend. You need to remove from the docs the endpoints and the info related to balancer. ALso we need to remote from the user db any possible event (I believe we don´t have any) and sync with the frontend on removing the view

rotkehlchen/chain/evm/decoding/balancer/utils.py
prettyirrelevant: okay, thanks

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,201 @@
...
+                    )
+                    for token in pool['displayTokens']
+                ],
+            )
+            if pool_token.protocol != CPT_BALANCER_V1:
+                log.debug(f'Updating protocol for balancer asset {pool_token}')
+                GlobalDBHandler.set_token_protocol_if_missing(
+                    token=pool_token,
+                    new_protocol=CPT_BALANCER_V1,
+                )
```
yabirgb, line 117: this doesn´t raise any erro rright? move it out of the try/catch

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,201 @@
...
+        except (KeyError, ValueError, TypeError, DeserializationError) as e:
+            log.error(
+                f'Failed to process balancer v1 pool due to {e!s}. '
+                f'Pool: {pool}. Skipping...',
```
yabirgb, line 121: have you checked if the poll string version containes enought information?

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,201 @@
...
+                },
+                'query': GET_POOLS_COUNT_QUERY,
+            },
+            timeout=CachedSettings().get_timeout_tuple(),
+        )
+        response.raise_for_status()
+        data = response.json()
+        return data['data']['poolGetPoolsCount']
+    except (requests.exceptions.RequestException, KeyError) as e:
+        raise RemoteError(f'Balancer pools count query failed due to {e!s}') from e
```
yabirgb, line 155: here and in the other cases where you catch keyError you have to be careful. Check other parts of the code where we do it. the string of KeyError is the missing key and not a message that tells you that is a key error. Please adjust those

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,201 @@
...
+            json={
+                'variables': {
+                    'chain': CHAIN_ID_TO_BALANCER_API_MAPPINGS[chain],
+                    'version': version,
+                },
+                'query': GET_POOLS_COUNT_QUERY,
+            },
+            timeout=CachedSettings().get_timeout_tuple(),
+        )
+        response.raise_for_status()
```
yabirgb, line 151: mmmm that is not what you want I think? Does this raise if you receive a 404 error? have you checked it? also what are the responses from the api?

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,201 @@
...
+        globaldb_set_unique_cache_value_at_ts(
+            write_cursor=write_cursor,
+            key_parts=cache_key_parts,
+            value=str(latest_pools_count),
+            timestamp=now,
+        )
+
+
+def _query_balancer_pools_count(chain: 'ChainID', version: Literal[1, 2]) -> int:
+    """Fetch the total number of pools for the specified chain and protoco
```
yabirgb, line 135:
```suggestion
    """Fetch the total number of pools for the specified chain and protocol version
```

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,201 @@
...
+                        'skip': offset,
+                        'version': version,
+                        'first': BALANCER_API_CHUNK_SIZE,
+                        'chain': CHAIN_ID_TO_BALANCER_API_MAPPINGS[chain],
+                    },
+                    'query': GET_POOLS_QUERY,
+                },
+                timeout=CachedSettings().get_timeout_tuple(),
+            )
+            response.raise_for_status()
```
yabirgb, line 185: check again if that is what you want

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,201 @@
...
+        except (requests.exceptions.RequestException, JSONDecodeError) as e:
+            raise RemoteError(f'Balancer pools query failed due to {e!s}') from e
+
+        if 'errors' in data:
+            raise RemoteError(f"Balancer pools query failed: {data['errors']}")
+
+        pools_chunk = data['data']['poolGetPools']
+        all_pools.extend(pools_chunk)
+
+        if len(pools_chunk) < BALANCER_API_CHUNK_SIZE:
```
yabirgb, line 196:
```suggestion
        all_pools.extend(pools_chunk := data['data']['poolGetPools'])

        if len(pools_chunk) < BALANCER_API_CHUNK_SIZE:
```

also this can raise KeyError

rotkehlchen/chain/evm/decoding/balancer/v1/accountant.py
```diff
@@ -0,0 +1,21 @@
...
+
+from rotkehlchen.chain.evm.accounting.interfaces import DepositableAccountantInterface
+from rotkehlchen.chain.evm.accounting.structures import EventsAccountantCallback
+from rotkehlchen.chain.evm.decoding.balancer.constants import CPT_BALANCER_V1
+from rotkehlchen.history.events.structures.base import get_event_type_identifier
+from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.logging import RotkehlchenLogsAdapter
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
```
yabirgb, line 11: why do we need the log here?

yabirgb (commented):
Still a few things off. Also lefteris uses vulture to check for dead code. Can you check if you left any? I think there might be some data structures https://discord.com/channels/657906918408585217/1080783409640648765/1230463935917068359

rotkehlchen/chain/evm/decoding/balancer/v1/balances.py
```diff
@@ -0,0 +1,85 @@
...
+                        token_type=EvmTokenKind.ERC20,
+                        chain_id=self.evm_inquirer.chain_id,
+                    ))
+                    balances[address].assets[token] += Balance(
+                        amount=FVal(lp_balance['totalBalance']),
+                        usd_value=FVal(lp_balance['totalBalance']) * FVal(lp_balance['tokenPrice']),  # noqa: E501
+                    )
+            except (RemoteError, KeyError, ValueError) as e:
+                msg = f'missing key {e!s}' if isinstance(e, KeyError) else str(e)
+                log.error(f'Failed to query Balancer v1 lp balances for {address} on {self.evm_inquirer.chain_id} due to {msg}')  # noqa: E501
```
LefterisJP, line 82:
```suggestion
                log.error(f'Failed to query Balancer v1 lp balances for {address} on {self.evm_inquirer.chain_name} due to {msg}')  # noqa: E501
```

rotkehlchen/chain/evm/decoding/balancer/v1/balances.py
```diff
@@ -0,0 +1,85 @@
...
+                        'address': address,
+                        'first': BALANCER_API_CHUNK_SIZE,
+                        'chain': CHAIN_ID_TO_BALANCER_API_MAPPINGS[self.evm_inquirer.chain_id],
+                    },
+                )
+                for lp_balance in response['userGetPoolBalances']:
+                    if (token_address := to_checksum_address(lp_balance['tokenAddress'])) not in pool_tokens:  # noqa: E501
+                        continue
+
+                    token = Asset(evm_address_to_identifier(
```
LefterisJP, line 71: Is this token guaranteed to be in the DB? IF not then this is a fatal mistake and should be `get_or_create_evm_token`, or just `get_token`, depending on what you want to achieve.

rotkehlchen/chain/evm/decoding/balancer/v1/decoder.py
```diff
@@ -0,0 +1,392 @@
...
+        )
+        self.pools: set[ChecksumEvmAddress] = set()
+
+    def _decode_v1_pool_event(self, all_logs: list[EvmTxReceiptLog]) -> list[dict[str, Any]] | None:  # noqa: E501
+        """Read the list of logs in search for a Balancer v1 event and return the information
+        needed to decode the transfers made in the transaction to/from the ds proxy
+        """
+        # The transfer event appears after the debt generation event, so we need to transform it
+        target_logs = [x for x in all_logs if x.topics[0] in (JOIN_V1, EXIT_V1)]
+        if len(target_logs) == 0:
```
LefterisJP, line 64:
```suggestion
        if len((target_logs := [x for x in all_logs if x.topics[0] in (JOIN_V1, EXIT_V1)])) == 0:
```

rotkehlchen/chain/evm/decoding/balancer/v1/decoder.py
```diff
@@ -0,0 +1,392 @@
...
+        """
+        Enrich balancer v1 transfer to read pool events of:
+        - Depositing in the pool
+        - Withdrawing from the pool
+
+        In balancer v1 pools are managed using a DSProxy so the account doesn't interact
+        directly with the pools.
+        """
+        events_information = self._decode_v1_pool_event(all_logs=context.all_logs)
+        if events_information is None:
```
LefterisJP, line 101:
```suggestion
        if (events_information := self._decode_v1_pool_event(all_logs=context.all_logs)) is None:
```

rotkehlchen/chain/evm/decoding/balancer/v1/decoder.py
```diff
@@ -0,0 +1,392 @@
...
+            return FAILED_ENRICHMENT_OUTPUT
+
+        for proxied_event in events_information:
+            if proxied_event['ds_address'] != context.event.address:
```
LefterisJP, line 105: what about key error here? Shouldn' it be handled? Or at least do a .get() to not have exceptions? Here and elsewhere
```suggestion
            if proxied_event.get('ds_address') != context.event.address:
```

rotkehlchen/chain/evm/decoding/balancer/v1/decoder.py
```diff
@@ -0,0 +1,392 @@
...
+                from_event_subtype=HistoryEventSubType.NONE,
+                to_event_type=HistoryEventType.WITHDRAWAL,
+                location_label=user_address,
+                to_event_subtype=HistoryEventSubType.REMOVE_ASSET,
+                to_counterparty=CPT_BALANCER_V1,
+                to_notes=f'Receive {withdrawn_amount} {token_out.symbol} after removing liquidity from a Balancer v1 pool',  # noqa: E501
+                asset=token_out,
+            )])
+
+        if context.tx_log.topics[0] == TRANSFER_TOPIC:
```
LefterisJP, line 187: can't you do all this have a common exit point?

Define the common attributes of all topics for the arguments of `ActionItem`. And then do an if/elif/else populating the per topic attributes but instead of returning as you do here in every if, return only once at the very end

rotkehlchen/chain/evm/decoding/balancer/v1/decoder.py
```diff
@@ -0,0 +1,392 @@
...
+            decoded_events: list['EvmEvent'],
+            all_logs: list[EvmTxReceiptLog],  # pylint: disable=unused-argument
+    ) -> list['EvmEvent']:
+        """
+        It can happen that after sending tokens to the DSProxy in balancer V1 the amount of tokens
+        required for the deposit is lower than the amount sent and then those tokens are returned
+        to the DSProxy and then to the user.
+        """
+        deposited_assets = set()
+
```
LefterisJP, line 223:
```suggestion
        deposited_assets = set()
```

rotkehlchen/chain/evm/decoding/balancer/v1/decoder.py
```diff
@@ -0,0 +1,392 @@
...
+                event.event_type == HistoryEventType.DEPOSIT and
+                event.event_subtype == HistoryEventSubType.DEPOSIT_ASSET
+            ):
+                deposited_assets.add(event.asset)
+            elif (
+                event.event_type == HistoryEventType.RECEIVE and
+                event.event_subtype == HistoryEventSubType.RECEIVE_WRAPPED and
+                event.asset in deposited_assets
+            ):
+                # in this case we got refunded one of the assets deposited
```
LefterisJP, line 238:
```suggestion
            ):  # in this case we got refunded one of the assets deposited
```

rotkehlchen/chain/evm/decoding/balancer/v1/decoder.py
```diff
@@ -0,0 +1,392 @@
...
+            elif (
+                (
+                    event.event_type == HistoryEventType.DEPOSIT and
+                    event.event_subtype == HistoryEventSubType.DEPOSIT_ASSET
+                ) or
+                (
+                    event.event_type == HistoryEventType.WITHDRAWAL and
+                    event.event_subtype == HistoryEventSubType.REFUND
+                )
+            ):
```
LefterisJP, line 296:
```suggestion
            elif ((
                    event.event_type == HistoryEventType.DEPOSIT and
                    event.event_subtype == HistoryEventSubType.DEPOSIT_ASSET
                ) or (
                    event.event_type == HistoryEventType.WITHDRAWAL and
                    event.event_subtype == HistoryEventSubType.REFUND
            )):
```

rotkehlchen/chain/evm/decoding/balancer/v2/decoder.py
```diff
@@ -0,0 +1,173 @@
...
+            base_tools: 'BaseDecoderTools',
+            msg_aggregator: 'MessagesAggregator',
+    ) -> None:
+        super().__init__(
+            evm_inquirer=evm_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+        )
+        self.eth = A_ETH.resolve_to_crypto_asset()
+        self.weth = A_WETH.resolve_to_evm_token()
```
LefterisJP, line 50: These don't make sense.

```suggestion
```

1. self.eth is used only as the asset and symbol which are known and constant. No need to initialize them again here.
2. self.weth is wrong. There is a different WETH per chain and you are using the ethereum one here. Again not needed at all. You only use the address. And the place where you use the address you should just use the `CHAIN_TO_WRAPPED_TOKEN` dict.

rotkehlchen/tests/unit/decoders/test_balancer.py
```diff
@@ -139,42 +140,42 @@ def test_balancer_v1_exit(ethereum_inquirer, ethereum_accounts):
...
-            event_subtype=HistoryEventSubType.REMOVE_ASSET,
-            asset=A_WETH,
-            balance=Balance(amount=FVal('0.010687148200906598')),
+            event_type=HistoryEventType.SPEND,
+            event_subtype=HistoryEventSubType.RETURN_WRAPPED,
+            asset=A_BPT,
+            balance=Balance(amount=FVal('0.042569019597126949')),
             location_label=user_address,
-            notes='Receive 0.010687148200906598 WETH after removing liquidity from a Balancer v1 pool',  # noqa: E501
+            notes='Return 0.042569019597126949 BPT to a Balancer v1 pool',
```
LefterisJP, line 174: So this seems wrong. The IN event comes first and the out later. rotki's events should always be OUT first, in afterwards

rotkehlchen/tests/unit/decoders/test_balancer.py
```diff
@@ -331,3 +332,310 @@ def test_balancer_trade(ethereum_inquirer, ethereum_accounts):
...
+            timestamp=timestamp,
+            location=Location.ETHEREUM,
+            event_type=HistoryEventType.SPEND,
+            event_subtype=HistoryEventSubType.FEE,
+            asset=A_ETH,
+            balance=Balance(amount=FVal(gas_str)),
+            location_label=user_address,
+            notes=f'Burn {gas_str} ETH for gas',
+            counterparty=CPT_GAS,
+        ), EvmEvent(
```
LefterisJP, line 449: Again same here. This seems wrong. Withdrawal (IN) event comes first in this test. Should always be OUT first, IN later

rotkehlchen/tests/unit/decoders/test_balancer.py
```diff
@@ -331,3 +332,310 @@ def test_balancer_trade(ethereum_inquirer, ethereum_accounts):
...
+            tx_hash=tx_hash,
+            sequence_index=1,
+            timestamp=timestamp,
+            location=Location.ARBITRUM_ONE,
+            event_type=HistoryEventType.WITHDRAWAL,
+            event_subtype=HistoryEventSubType.REMOVE_ASSET,
+            asset=Asset('eip155:42161/erc20:0xaf88d065e77c8cC2239327C5EDb3A432268e5831'),
+            balance=Balance(amount=FVal(withdrawn_usdc_amt)),
+            location_label=user_address,
+            notes=f'Receive {withdrawn_usdc_amt} USDC after removing liquidity from a Balancer v1 pool',  # noqa: E501
```
LefterisJP, line 523: Same here. OUT first, IN later

LefterisJP (changes requested):
Lots of mistakes and things that are not done well and need work

rotkehlchen/chain/evm/decoding/balancer/v1/balances.py
prettyirrelevant: yeah, it is. there's a check.
```py
pool_tokens = set(GlobalDBHandler.get_addresses_by_protocol(
            protocol=CPT_BALANCER_V1,
            chain_id=self.evm_inquirer.chain_id,
        ))
```

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,232 @@
...
+        )
+        if response.status_code != 200:
+            raise RemoteError(f'Balancer API request failed with status {response.status_code}')
+
+        data = response.json()
+        if 'errors' in data:
+            raise RemoteError(f"Balancer API query failed: {data['errors']}")
+
+        return data['data']
+    except (requests.exceptions.RequestException, KeyError) as e:
```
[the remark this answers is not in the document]
prettyirrelevant: it can't throw `JSONDecodeError` as its handled in the method.
```py
    def json(self, **kwargs):
        r"""Returns the json-encoded content of a response, if any.

        :param \*\*kwargs: Optional arguments that ``json.loads`` takes.
        :raises requests.exceptions.JSONDecodeError: If the response body does not
            contain valid json.
        """

        if not self.encoding and self.content and len(self.content) > 3:
            # No encoding set. JSON RFC 4627 section 3 states we should expect
            # UTF-8, -16 or -32. Detect which one to use; If the detection or
            # decoding fails, fall back to `self.text` (using charset_normalizer to make
            # a best guess).
            encoding = guess_json_utf(self.content)
            if encoding is not None:
                try:
                    return complexjson.loads(self.content.decode(encoding), **kwargs)
                except UnicodeDecodeError:
                    # Wrong UTF codec detected; usually because it's not UTF-8
                    # but some other 8-bit codec.  This is an RFC violation,
                    # and the server didn't bother to tell us what codec *was*
                    # used.
                    pass
                except JSONDecodeError as e:
                    raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)

        try:
            return complexjson.loads(self.text, **kwargs)
        except JSONDecodeError as e:
            # Catch JSON-related errors and raise as requests.JSONDecodeError
            # This aliases json.JSONDecodeError and simplejson.JSONDecodeError
            raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
```

rotkehlchen/chain/evm/decoding/balancer/utils.py
```diff
@@ -0,0 +1,232 @@
...
+                            ),
+                        ).evm_address,
+                        token_kind=EvmTokenKind.ERC20,
+                        weight=FVal(token['weight']),
+                    )
+                    for token in pool['displayTokens']
+                ],
+            )
+        except (KeyError, ValueError, TypeError, DeserializationError) as e:
+            log.error(
```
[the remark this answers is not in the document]
prettyirrelevant: only `KeyError` lacks context, every other one contains needed information iiuc.

rotkehlchen/chain/evm/decoding/balancer/utils.py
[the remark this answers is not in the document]
prettyirrelevant: per the discussion in discord, i'm thinking of writing a pylint plugin. what do you think -- an overkill?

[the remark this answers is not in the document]
LefterisJP: > per the discussion in discord, i'm thinking of writing a pylint plugin. what do you think -- an overkill?

These comments are not a place for discussion. Move to disord.

rotkehlchen/chain/evm/decoding/balancer/v2/decoder.py
```diff
@@ -0,0 +1,179 @@
...
+            native_asset: 'Asset',
+            native_asset_symbol: Literal['ETH', 'XDAI', 'POL'],
+            wrapped_native_asset_address: ChecksumEvmAddress,
+    ) -> None:
+        super().__init__(
+            evm_inquirer=evm_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+        )
+        self.native_asset = native_asset
```
LefterisJP, line 56: Where is this used? You don't need this

rotkehlchen/chain/evm/decoding/balancer/v2/decoder.py
```diff
@@ -0,0 +1,179 @@
...
+        self.native_asset_symbol = native_asset_symbol
```
LefterisJP, line 57: you also don't need this. `self.node_inquirer.native_token` exists. And that is already resolved. So you can use its symbol.

rotkehlchen/chain/evm/decoding/balancer/v2/decoder.py
```diff
@@ -0,0 +1,179 @@
...
+        self.wrapped_native_asset_address = wrapped_native_asset_address
```
LefterisJP, line 58: Also don't need this, since you have the wrapped native token mapping I said in the previous PR

rotkehlchen/chain/evm/decoding/balancer/v2/decoder.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+
+        # For native asset swaps, it's wrapped and the native transfer happens before SWAP.
+        # Detect such event if not already found.
+        if len(context.action_items) == 0 and from_token_address == self.wrapped_native_asset_address:  # noqa: E501
+            # When swapping native asset, transfer precedes V2_SWAP.
+            # Check if native asset was swapped
+            amount_of_eth = token_normalized_value_decimals(
+                token_amount=amount_in,
+                token_decimals=18,
```
LefterisJP, line 105: Use `token_decimals=DEFAULT_TOKEN_DECIMALS`

LefterisJP (approved):
LGTM

PR #8912, opened 2024-11-20
chore: update vanilla-jsoneditor

kelsos:
Closes #(issue_number)

## Checklist


kelsos:
It turns out that vanilla-jsoneditor was pulling the dependency as a production dependency in v1.x and that it was removed in v2

PR #8918, opened 2024-11-21
Improve speed of token detection

yabirgb:
We were hitting the database to query more information than what we needed with heavy requests. This PR creates a new minimal data structure with a much simpler query to improve performance on token detection.

### DIFF
--- a/rotkehlchen/chain/evm/tokens.py
+++ b/rotkehlchen/chain/evm/tokens.py
@@ -1,12 +1,16 @@
 import logging
 from abc import ABC, abstractmethod
 from collections import defaultdict
-from collections.abc import Sequence
-from typing import TYPE_CHECKING
+from collections.abc import Mapping, Sequence
+from typing import TYPE_CHECKING, TypeVar
 
-from rotkehlchen.assets.asset import EvmToken, Nft
-from rotkehlchen.chain.ethereum.utils import token_normalized_value
+from rotkehlchen.assets.asset import Asset, EvmToken, Nft
+from rotkehlchen.chain.ethereum.utils import (
+    token_normalized_value,
+    token_normalized_value_decimals,
+)
 from rotkehlchen.chain.evm.types import WeightedNode, asset_id_is_evm_token
+from rotkehlchen.chain.structures import EvmTokenDetectionData
 from rotkehlchen.fval import FVal
 from rotkehlchen.globaldb.handler import GlobalDBHandler
 from rotkehlchen.inquirer import Inquirer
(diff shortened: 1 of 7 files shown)


rotkehlchen/chain/evm/tokens.py
```diff
@@ -127,9 +133,9 @@ def __init__(
     def get_token_balances(
             self,
             address: ChecksumEvmAddress,
-            tokens: list[EvmToken],
+            tokens: list[EvmTokenDetectionData],
             call_order: Sequence[WeightedNode] | None,
-    ) -> dict[EvmToken, FVal]:
+    ) -> dict[Asset, FVal]:
```
LefterisJP, line 138: Can you specify in the docstring why you return an `Asset` instead of an `EvmToken`? I know why -- this is why you made this PR, but should also be explained in the docstring.

rotkehlchen/chain/evm/tokens.py
```diff
@@ -146,22 +152,22 @@ def get_token_balances(
...
-        balances: dict[EvmToken, FVal] = defaultdict(FVal)
+        balances: dict[Asset, FVal] = defaultdict(FVal)
 
         try:
             for token_balance, token in zip(result, tokens, strict=True):
                 if token_balance == 0:
                     continue
 
-                normalized_balance = token_normalized_value(token_balance, token)
+                normalized_balance = token_normalized_value_decimals(token_balance, token.decimals)
```
LefterisJP, line 165: plox add kwargs

rotkehlchen/chain/evm/tokens.py
```diff
@@ -219,11 +225,11 @@ def _get_multicall_token_balances(
     def _query_chunks(
             self,
             address: ChecksumEvmAddress,
-            tokens: list[EvmToken],
+            tokens: list[EvmTokenDetectionData],
             chunk_size: int,
             call_order: list[WeightedNode],
-    ) -> dict[EvmToken, FVal]:
-        total_token_balances: dict[EvmToken, FVal] = defaultdict(FVal)
+    ) -> dict[Asset, FVal]:
```
LefterisJP, line 231: same comment here for the docstring. But since this is a function down in the call line probablyjust needs a small 1-liner

rotkehlchen/globaldb/handler.py
```diff
@@ -944,6 +946,33 @@ def get_evm_tokens(
 
         return tokens
 
+    @staticmethod
+    def get_token_detection_data(
+            chain_id: ChainID,
+            exceptions: set[ChecksumEvmAddress],
+            protocol: str | None = None,
+    ) -> list[EvmTokenDetectionData]:
```
LefterisJP, line 954: Please add a small docstring on what this fucntion does and why it was made

rotkehlchen/globaldb/handler.py
```diff
@@ -944,6 +946,33 @@ def get_evm_tokens(
...
+        with GlobalDBHandler().conn.read_ctx() as cursor:
+            cursor.execute(query, bindings)
+            for identifier, address, decimals in cursor:
+                if address in exceptions:
+                    continue
+
+                result.append(EvmTokenDetectionData(
+                    identifier=identifier,
+                    address=address,
+                    decimals=decimals if decimals is not None else DEFAULT_TOKEN_DECIMALS,  # TODO: at least two tokens are missing the decimals in my DB and also the EvmToken class allows decimals to be None. We need to fix it.  # noqa: E501
```
LefterisJP, line 971: what would the fix be here?

yabirgb: I was thinking that we would need to enforce the decimals for the EvmTokens but now I'm not 100% sure. We have the NFTs as child of the EvmTokens. In theory all the EvmTokens should have decimals unless I'm missing something.

LefterisJP (approved):
LGTM

PR #8922, opened 2024-11-21
Fix history event export/download api endpoints

nicholasyoder:
Changes the history events CSV download into a two step process with an asynchronous request first to export the events and then a separate request to download the generated CSV file.

Related to the problem with history events CSV downloads when running in docker reported in [discord](https://discord.com/channels/657906918408585217/745172610240872508/1308910435935719425).

Backend changes only.

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -4792,7 +4792,7 @@ class RestAPI:
             filter_query: HistoryBaseEntryFilterQuery,
             directory_path: Path | None,
     ) -> dict[str, Any] | Response:
-        """Export or Download history events data to a CSV file."""
+        """Export history events data to a CSV file."""
         dbevents = DBHistoryEvents(self.rotkehlchen.data.db)
         with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
             history_events, _, _ = dbevents.get_history_events_and_limit_info(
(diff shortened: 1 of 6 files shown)


rotkehlchen/api/rest.py
```diff
@@ -4842,35 +4842,22 @@ def export_history_events(
...
-                        message=str(e),
-                        status_code=HTTPStatus.CONFLICT,
-                    )
-                except FileNotFoundError:
-                    return wrap_in_fail_result(
-                        message='No file was found',
-                        status_code=HTTPStatus.NOT_FOUND,
-                    )
+        try:
+            if directory_path is None:  # file will be downloaded later
```
LefterisJP, line 4846: fill in the method of download

```suggestion
            if directory_path is None:  # file will be downloaded later via XXX endpoint
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -3434,6 +3434,12 @@ def generate_fields_post_validation(self, data: dict[str, Any]) -> dict[str, Any
         return extra_fields
 
 
+class ExportHistoryDownloadSchema(Schema):
+    """Schema for downloading history events CSVs."""
+
+    file_path = fields.String(required=True)
```
LefterisJP, line 3440:
```suggestion
    file_path = fields.String(required=True)
```

LefterisJP (commented):
Left some small comments. Then also.

1. Someone from the frontend should check that this is what you guys discussed you need
2. Do we need this for any other endpoint to work fine in docker? Was the history events CSV export the only problematic one?

lukicenturi (approved):
LGTM! Works fine for me

nicholasyoder:
> Do we need this for any other endpoint to work fine in docker? Was the history events CSV export the only problematic one?

This was the only asynchronous endpoint using `register_post_download_cleanup` and `send_file`, so this should be the only problematic one. To confirm I tried downloading a userdb snapshot (which uses the same basic logic, just in a normal syncronous endpoint) and it worked fine.

LefterisJP (approved):
LGTM

PR #8923, opened 2024-11-22
fix(frontend): adjust history events CSV download endpoint call

lukicenturi:
Needs the backend part: https://github.com/rotki/rotki/pull/8922 to be merged first

### DIFF
--- a/frontend/app/src/composables/api/history/events/index.ts
+++ b/frontend/app/src/composables/api/history/events/index.ts
@@ -25,7 +25,7 @@ import { type HistoryEventProductData, HistoryEventTypeData } from '@/types/hist
 import type { CollectionResponse } from '@/types/collection';
 import type { PendingTask } from '@/types/task';
 import type { ActionResult } from '@rotki/common';
-import type { ActionDataEntry } from '@/types/action';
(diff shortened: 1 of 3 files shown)


frontend/app/src/composables/api/history/events/index.ts
```diff
@@ -253,6 +254,28 @@ export function useHistoryEventsApi(): UseHistoryEventsApiReturn {
     return handleResponse(response);
   };
 
+  const downloadHistoryEventsCSV = async (filePath: string): Promise<ActionStatus> => {
+    try {
+      const response = await api.instance.post('/history/events/export/download', snakeCaseTransformer({ filePath }), {
```
kelsos, line 259: So does the endpoint only work with `POST`? no `GET`?

Also wouldn't the Ajax/Blob approach load the file in memory first before downloading?
I would like to avoid this.

lukicenturi: Okay, I'll turn it into `GET` so we don't need to use blob

lukicenturi:
Needs https://github.com/rotki/rotki/pull/8938 to be merged first

cleanerzkp (approved):
LGTM

PR #8938, opened 2024-11-23
fix: turn history events download to [GET]

lukicenturi:
Turn endpoint `/history/events/export/download` to use [GET] instead of [POST]
So frontend can download it directly without having to turn it into blob, so we don't need to store it in memory.
https://github.com/rotki/rotki/pull/8923#discussion_r1854203922


LefterisJP (approved):
LGTM

PR #8943, opened 2024-11-24
Remove dead and test only code in production

LefterisJP:


LefterisJP:
okay finally all green.

PR #8944, opened 2024-11-25
added balancer v2 accountant

prettyirrelevant:


rotkehlchen/chain/evm/decoding/balancer/v2/accountant.py
LefterisJP, line 1: Should this be an empty file, or deleted?

LefterisJP (commented):
Shuldn't other accountant modules also use the same interface? Or they already do?

LefterisJP (approved):
LGTM

PR #8958, opened 2024-11-27
fix vcr

prettyirrelevant:
the vcr step now tries to check out an existing local branch before creating a new one from the remote branch.


LefterisJP (approved):
LGTM

PR #8965, opened 2024-11-27
Add Curve lending balances logic

### LINKED ISSUE #7621: Add Curve Finance - crvUSD lending
## Abstract

No way to track crvUSD loans and interest rates in rotki

## Motivation

Accurate tracking of portfolio

nicholasyoder:
Closes #7621


rotkehlchen/chain/evm/decoding/curve_lend/balances.py
```diff
@@ -0,0 +1,164 @@
...
+                address=token_contract.decode(
+                    result=call_output[1],
+                    method_name='borrowed_token',
+                )[0],
+                chain_id=self.evm_inquirer.chain_id,
+                token_type=EvmTokenKind.ERC20,
+            ))
+
+            # Add collateral to assets balances
+            for token, amount in [
```
LefterisJP, line 153: A generic question here. How common is it to have the same tokens here? As your current approach will do the price queries again and again for each token if they are part of different vault or different address balances.

A better approach would be to split the price queries off and query them later when you know the individual tokens.

Depends also on how they are structured and if they can repeat.

rotkehlchen/chain/evm/decoding/morpho/utils.py
```diff
@@ -124,32 +121,13 @@ def get_morpho_vault_token_price(
...
-        token=underlying_token,
+    """Gets the token price for a Morpho vault."""
+    return get_vault_price(
+        inquirer=inquirer,
+        vault_token=vault_token,
+        evm_inquirer=evm_inquirer,
+        display_name='Morpho',
+        vault_abi=MORPHO_VAULT_ABI,
+        pps_method='convertToAssets',
+        pps_method_args=[int(1e18)],  # 1 share with 18 decimals
```
LefterisJP, line 132: I think we got this as a constant in the app. `EXP18`

rotkehlchen/chain/evm/decoding/utils.py
```diff
@@ -188,3 +196,50 @@ def update_cached_vaults(
...
+        price_per_share = evm_inquirer.call_contract(
+            contract_address=vault_token.evm_address,
+            abi=vault_abi,
+            method_name=pps_method,
+            arguments=pps_method_args,
+        )
+    except RemoteError as e:
+        log.error(
+            f'Failed to get price per share for {display_name} '
+            f'vault {vault_token} on {evm_inquirer.chain_id}: {e}',
```
LefterisJP, line 224: Please remember to use chain name for all logging and display purposes.

```suggestion
            f'vault {vault_token} on {evm_inquirer.chain_name}: {e}',
```

rotkehlchen/chain/evm/decoding/utils.py
```diff
@@ -188,3 +196,50 @@ def update_cached_vaults(
...
+    if (
+        len(vault_token.underlying_tokens) == 0 or
+        (underlying_token := get_token(
+            evm_address=vault_token.underlying_tokens[0].address,
+            chain_id=evm_inquirer.chain_id,
+        )) is None
+    ):
+        log.error(
+            f'Failed to get underlying token for {display_name} '
+            f'vault {vault_token} on {evm_inquirer.chain_id}',
```
LefterisJP, line 237:
```suggestion
            f'vault {vault_token} on {evm_inquirer.chain_name}',
```

rotkehlchen/chain/evm/decoding/utils.py
```diff
@@ -188,3 +196,50 @@ def update_cached_vaults(
...
             )
+
+
+def get_vault_price(
+        inquirer: 'Inquirer',
+        vault_token: 'EvmToken',
+        evm_inquirer: 'EvmNodeInquirer',
+        display_name: str,
+        vault_abi: ABI,
+        pps_method: str,
```
LefterisJP, line 207: You can narrow the type down here by `Literal[]` of possible combinations

rotkehlchen/inquirer.py
```diff
@@ -224,6 +226,12 @@ def get_underlying_asset_price(token: EvmToken) -> tuple[Price | None, CurrentPr
             inquirer=Inquirer(),  # Initialize here to avoid a circular import
             evm_inquirer=Inquirer.get_evm_manager(chain_id=token.chain_id).node_inquirer,
         )
+    elif token.protocol == CURVE_LENDING_VAULTS_PROTOCOL:
+        price = get_curve_vault_token_price(
```
LefterisJP, line 230:
```suggestion
        price = get_curve_lending_vault_token_price(
```

LefterisJP (commented):
Left some comments. Also needs rebase and conflict resolution

rotkehlchen/chain/evm/decoding/curve_lend/balances.py
```diff
@@ -0,0 +1,164 @@
...
+    def __init__(
+            self,
+            evm_inquirer: 'EvmNodeInquirer',
+            tx_decoder: 'EVMTransactionDecoder',
+    ):
+        super().__init__(
+            evm_inquirer=evm_inquirer,
+            tx_decoder=tx_decoder,
+            counterparty=CPT_CURVE,
+            deposit_event_types=set(),
```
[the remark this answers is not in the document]
nicholasyoder: hmm... yes, I guess I simply assumed deposit_event_types didn't apply here since it was using event types that weren't actually deposits. I've changed it to use deposit_event_types now, but I do wonder if perhaps that name should be changed to something else.

rotkehlchen/chain/evm/decoding/curve_lend/balances.py
```diff
@@ -0,0 +1,164 @@
...
+        """Get addresses of controllers for vaults that the user may have balances on.
+        Returns a dict of controller addresses -> set of user addresses with possible balances.
+        """
+        controllers: dict[ChecksumEvmAddress, set[ChecksumEvmAddress]] = defaultdict(set)
+        for address, events in self.addresses_with_activity(event_types={
+            (HistoryEventType.WITHDRAWAL, HistoryEventSubType.GENERATE_DEBT),
+            (HistoryEventType.DEPOSIT, HistoryEventSubType.DEPOSIT_ASSET),
+        }).items():
+            for event in events:
+                if event.extra_data is None or 'vault_controller' not in event.extra_data:
```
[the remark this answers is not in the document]
nicholasyoder: Actually it is possible for the DEPOSIT/DEPOSIT_ASSET events to not have extra data here when there are deposits to a lending vault (see [here](https://github.com/rotki/rotki/blob/6bc3b6d20f6a929d727ba9581b24edc7218219fa/rotkehlchen/chain/evm/decoding/curve_lend/decoder.py#L166)). These simply need to be skipped without any error, since they do not need processed for balances, so I think it is correct as it is.

rotkehlchen/chain/evm/decoding/curve_lend/balances.py
```diff
@@ -0,0 +1,164 @@
...
+
+            # Add collateral to assets balances
+            for token, amount in [
+                (collateral_token, collateral_amount),
+                (borrowed_token, borrowable_collateral_amount),
+            ]:
+                if (balance := self._get_token_balance(token=token, amount=amount)) is not None:
+                    balances[user_address].assets[token] += balance
+
+            # Add debt to liabilities balances
```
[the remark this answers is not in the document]
nicholasyoder: I'm not sure if I see any problem here, but this code has now been heavily modified to fix the multiple price queries, and I think it should be correct now whatever the case.

rotkehlchen/chain/evm/decoding/curve_lend/balances.py
nicholasyoder: Yes, actually with this approach the borrowed_token was getting its price queried twice already - and if the user has balances in multiple vaults, and/or mutiple addresses with balances in the same vaults the same tokens will come through this logic multiple times.

I've modified the logic here so that it should only query prices once for each token.

LefterisJP (approved):
LGTM

LefterisJP (approved):
LGTM

PR #8967, opened 2024-11-28
Detect staked WCT balances and more WCT transactions

LefterisJP:
- Detect staked WCT balances
- Decode WCT withdrawals, stake lock increase, amount updates

### DIFF
--- a/rotkehlchen/chain/optimism/modules/walletconnect/balances.py
+++ b/rotkehlchen/chain/optimism/modules/walletconnect/balances.py
@@ -32,7 +32,10 @@ class WalletconnectBalances(ProtocolWithBalance):
             evm_inquirer=evm_inquirer,
             tx_decoder=tx_decoder,
             counterparty=CPT_WALLETCONNECT,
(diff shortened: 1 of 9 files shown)


rotkehlchen/chain/optimism/modules/walletconnect/balances.py
```diff
@@ -0,0 +1,60 @@
...
+                (HistoryEventType.STAKING, HistoryEventSubType.DEPOSIT_ASSET),
+                (HistoryEventType.STAKING, HistoryEventSubType.REMOVE_ASSET),
+            },
+        )
+
+    def query_balances(self) -> 'BalancesSheetType':
+        """Query balances of staked WalletConnect"""
+        balances: BalancesSheetType = defaultdict(BalanceSheet)
+        wct_token = Asset(WCT_TOKEN_ID)
+        wct_price = Inquirer.find_usd_price(Asset(WCT_TOKEN_ID))
```
yabirgb, line 45: Wouldn't it make sense to query the price after you check that there are accounts with balances?

LefterisJP: yes it would

PR #8955, opened 2024-11-26
Fix typing in tests/api/test_i

azerv1:
Related #842.
tried working like @nicholasyoder and skipped _h in case he has already worked on it

PR #8979, opened 2024-11-29
pyproject.toml: add test/api/test_[i,l] to mypy.overrides

azerv1:
Updated the pyproject.toml because I hadn't done it in #8963 #8955


LefterisJP (approved):
LGTM

PR #8987, opened 2024-11-30
resolve .argent.xyz domains

### LINKED ISSUE #8981: Unable to resolve ENS for argent.xyz account
## Problem Definition

When I try to add my argent account `mysticryuujin.argent.xyz` Rotki says "is not an evm address"

So, I guess there's something preventing Rotki from attempting to lookup this address?

[It's valid](https://etherscan.io/name-lookup-search?id=mysticryuujin.argent.xyz)

### System Description

Operating system: Linux
Rotki version: 1.36.1

yabirgb:
allow the api to use argent domains in addtion to .eth domains.

Closes #8981

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,7 @@ Changelog
 * :bug:`8982` Fix the issue where Cmd / Alt + Arrow Left / Arrow Right forces navigation, as it should interact with user selection in text input.
 * :bug:`8983` The asset amount will be shown if only one asset is detected in an account.
 * :feature:`8991` Add direct navigation to asset details when clicking on small asset icons in the blockchain balances table.
+* :feature:`8981` rotki will now resolve any non .eth domains supported by ENS to addresses.
 
 * :release:`1.36.1 <2024-11-29>`
 * :bug:`-` Coinbase deposits that appeared as withdrawals should now appear as deposits again. Would need to purge and requery coinbase to fix this.
--- a/rotkehlchen/api/v1/schemas.py
+++ b/rotkehlchen/api/v1/schemas.py
@@ -35,6 +35,7 @@ from rotkehlchen.chain.ethereum.modules.eth2.structures import PerformanceStatus
 from rotkehlchen.chain.ethereum.modules.nft.structures import NftLpHandling
 from rotkehlchen.chain.ethereum.node_inquirer import EthereumInquirer
 from rotkehlchen.chain.evm.accounting.structures import BaseEventSettings, TxAccountingTreatment
+from rotkehlchen.chain.evm.decoding.ens.utils import is_potential_ens_name
 from rotkehlchen.chain.evm.types import EvmAccount, EvmlikeAccount
 from rotkehlchen.chain.gnosis.constants import GNOSIS_ETHERSCAN_NODE_NAME
 from rotkehlchen.chain.optimism.constants import OPTIMISM_ETHERSCAN_NODE_NAME
@@ -1803,7 +1804,7 @@ def _validate_blockchain_account_schemas(
     if chain.is_evm():
         for account_data in data['accounts']:
             address_string = address_getter(account_data)
-            if not address_string.endswith('.eth'):
+            if not is_potential_ens_name(address_string):
                 # Make sure that given value is an ethereum address
                 try:
                     address = to_checksum_address(address_string)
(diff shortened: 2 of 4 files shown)


rotkehlchen/chain/evm/constants.py
```diff
@@ -39,3 +39,4 @@
 
 # uniswap like merkle distributor claimed signature
 MERKLE_CLAIM: Final = b'N\xc9\x0e\x96U\x19\xd9&\x81&tg\xf7u\xad\xa5\xbd!J\xa9,\r\xc9=\x90\xa5\xe8\x80\xce\x9e\xd0&'  # noqa: E501
+VALID_ENS_EXTENSIONS: Final = ('.eth', '.argent.xyz')  # extensions compatible with the ENS resolution contract  # noqa: E501
```
LefterisJP, line 42: Why only those? Isn't literally almost any TLD okay? https://docs.ens.domains/dns/tlds

docs/changelog.rst
```diff
@@ -9,6 +9,7 @@ Changelog
 * :feature:`8929` Blockchain aggregated balances will now display the percentage of the net value and can also be filtered by chain.
 * :feature:`-` WalletConnect airdrop claiming and staking transactions are now properly recognized. WCT staked balances will also be autodetected.
 * :feature:`-` Page for inputting external API keys has been revamped.
+* :feature:`8981` rotki will now resolve argent.xyz domains to addresses.
```
LefterisJP, line 12:
```suggestion
* :feature:`8981` rotki will now resolve any non .eth domains supported by ENS to addresses.
```

rotkehlchen/chain/evm/decoding/ens/name_utils.py
```diff
@@ -0,0 +1,18 @@
...
+    If .eth is found we exit fast but for other domains like .argent.xyz
+    we use a regex that detects valid TLDs. The developer has always to check
+    if indeed the address is valid or not.
+
+    The full list of TLDs supported by ENS is available at https://docs.ens.domains/dns/tlds
+    """
+    if text.endswith('.eth'):
+        return True
+
+    return bool(TLD_PATTERN.search(text))
```
LefterisJP, line 18: not worth the extra if check. The regex is simple and regex matching is fast.

```suggestion
    return bool(TLD_PATTERN.search(text))
```

rotkehlchen/chain/evm/decoding/ens/name_utils.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def is_potential_ens_name(text: str) -> bool:
+    """Function used to tell if a string could be a potential ENS domain.
+    If .eth is found we exit fast but for other domains like .argent.xyz
+    we use a regex that detects valid TLDs. The developer has always to check
+    if indeed the address is valid or not.
+
+    The full list of TLDs supported by ENS is available at https://docs.ens.domains/dns/tlds
+    """
```
LefterisJP, line 14: Simplification of the text. Also question. What do you mean by "The developer has to check
    if indeed the address is valid or not."

```suggestion
    """Function used to tell if a string could be a potential ENS domain.
    We use a regex that detects valid TLDs. The developer has to check
    if indeed the address is valid or not.

    The full list of TLDs supported by ENS is available at https://docs.ens.domains/dns/tlds
    """
```

rotkehlchen/tests/conftest.py
```diff
@@ -33,6 +33,7 @@
 
 TESTS_ROOT_DIR = Path(__file__).parent
 SUBPROCESS_TIMEOUT = 30
+os.environ['POLARS_ALLOW_FORKING_THREAD'] = '1'  # ignore warning about fork and polars
```
LefterisJP, line 36: why did this pop out right now? Oh this is develop. Okay then this is another problem.

rotkehlchen/chain/evm/decoding/ens/name_utils.py
yabirgb: I meant that we don't try to resolve it so it might be that the TLD is not supported. But I believe is already clear by the name of the function

LefterisJP: > But I believe is already clear by the name of the function

yes, so you can remove it

LefterisJP (approved):
LGTM

PR #8993, opened 2024-12-01
typing:/tests/api/test_m-n

azerv1:
Related #842 

Note: Changed some lines apart from type hinting because mypy wouldn't accept them. [assert ... is not None]


LefterisJP (approved):
LGTM

Issue #8992, opened 2024-12-01
mouseover asset icons shows confusing tooltips


aspiers:
## Problem Definition

In `Accounts & Balances / Blockchain Balances`, if I mouse over the ETH icon, I see this:

![image](https://github.com/user-attachments/assets/f4dfb14b-3853-4c9a-852a-d616161ad71f)

There are two problems here:

1. One tooltip repeats "ETH" three times, which makes no real sense to the user.  Even if it showed "ETH" just once, this would not be telling the user anything helpful since they can already see the ETH icon.
2. Another tooltip says "Blockchain Balances per asset", which also seems more confusing than helpful.  (Also mentioned in #8990).

If we mouse over a different asset, it becomes clearer why ETH is repeated three times:

![image](https://github.com/user-attachments/assets/4ee5bb99-a933-4360-944f-4276bf03af29)

So presumably one is the ticker symbol, one is a human-readable description, and I guess the third one is the onchain identifier? i.e. either the token's smart contract address or in the case of the chain's native coin, just the symbol again.  But the code which generates this tooltip should only generate information which is

1. useful
2. additional on top of what the user already knows

So it seems this needs a bit of a rethink.

## Logs

Available on request but presumably not necessary as easy to reproduce.

### System Description

Operating system: openSUSE Tumbleweed Linux
Rotki version: v1.36.1 AppImage

aspiers:
I traced this `title` attribute all the way back to 2019:

https://github.com/rotki/rotki/commit/bf968ed788aba#diff-4ecba2a6e01bbc90865ffe6ffe70b9531e2ca61f462c53cf0de66753cb36dd91R27

@kelsos Perhaps you can comment on whether there's still a reason to keep it?

kelsos:
> I traced this `title` attribute all the way back to 2019:
> 
> [bf968ed788aba#diff-4ecba2a6e01bbc90865ffe6ffe70b9531e2ca61f462c53cf0de66753cb36dd91R27](https://github.com/rotki/rotki/commit/bf968ed788aba#diff-4ecba2a6e01bbc90865ffe6ffe70b9531e2ca61f462c53cf0de66753cb36dd91R27)
> 
> @kelsos Perhaps you can comment on whether there's still a reason to keep it?

I would say probably not, that commit seems to be part of the migration from JQuery to Vue. So my guess is that I just moved it over from the previous implementation.

aspiers:
[...]

PR #8994, opened 2024-12-01
fix: remove "Blockchain balances per asset" title attribute

aspiers:
Remove a "Blockchain Balances per Asset" tooltip which was appearing on mouse-over anywhere inside the "Blockchain Balances per asset" section.

Partial fix for #8992.

## Side note

This seems like an improvement to me, but no one on the core team has had a chance to comment on #8992 yet, so I appreciate it may be contentious.

Nevertheless, this is my first PR contribution in a long time, so even if it's not the best approach, it still serves as good practice for getting familiar with hacking on Rotki again :)

## Checklist

### DIFF
--- a/frontend/app/src/locales/cn.json
+++ b/frontend/app/src/locales/cn.json
@@ -1272,9 +1272,6 @@
       "edit_subtitle": "修改账户信息",
       "edit_title": "编辑区块链账户"
     },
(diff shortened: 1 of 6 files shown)


lukicenturi (approved):
LGTM!

PR #8995, opened 2024-12-02
fix: remove electron deprecated warning for browser navigation

lukicenturi:
Closes #(issue_number)

## Checklist


cleanerzkp (approved):
LGTM

frontend/app/electron/main/index.ts
```diff
@@ -71,17 +71,42 @@ async function onReady(): Promise<void> {
 
   getWindow().webContents.on('before-input-event', (event, input) => {
```
kelsos, line 72: What does event look like here? does it have a target?

frontend/app/electron/main/index.ts
```diff
@@ -71,17 +71,42 @@ async function onReady(): Promise<void> {
...
-    if (isMac ? input.meta : input.control) {
-      if (['ArrowLeft', '['].includes(input.key) && win.webContents.canGoBack()) {
-        win.webContents.goBack();
-        event.preventDefault();
-      }
 
-      if (['ArrowRight', ']'].includes(input.key) && win.webContents.canGoForward()) {
-        win.webContents.goForward();
-        event.preventDefault();
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
kelsos, line 75: Instead of suppressing the warning can we instead add a .catch that logs any error?
It should fix the problem and we should have some information in the logs if something goes wrong.

kelsos (approved):

frontend/app/electron/main/index.ts
kelsos: Great, if you can make it run only on the valid key combo that would be great, thanks.

lukicenturi:
ffs forgot changelog

PR #8996, opened 2024-12-02
fix: adjust label for coinbase prime api key form

lukicenturi:
Closes #(issue_number)

## Checklist


yabirgb (approved):
looks okey to me

imagen

frontend/app/src/components/settings/api-keys/exchange/ExchangeKeysForm.vue
```diff
@@ -185,8 +197,19 @@ function onExchangeChange(exchange?: string) {
...
+  if (location === 'coinbase') {
+    return markRaw(defineAsyncComponent(() => import('@/components/settings/api-keys/exchange/CoinbaseKeyForm.vue')));
+  }
+
+  if (location === 'coinbaseprime') {
+    return markRaw(defineAsyncComponent(() => import('@/components/settings/api-keys/exchange/CoinbaseProKeyForm.vue')));
+  }
+
+  return markRaw(defineAsyncComponent(() => import('@/components/settings/api-keys/exchange/RegularKeyForm.vue')));
+});
```
kelsos, line 212: Do we really need 3 components and all this complicated logic? can't we do a single one that takes the location?

lukicenturi: Some of them have a different order of input and also change the label of the input.
 I just thought this way was more readable.

kelsos: I wonder if we could also do the order in a dynamic way as part of the component/config?

kelsos: let me think a bit about it.

kelsos: Would it work if we use defineSlots to define the slots and payload and also create a structure that contains 

```typescript
type SlotProps = {label: string, hint: string}
const slots = defineSlots<{
 apiKey(props: SlotProps): any,
 apiSecret(props: SlotProps): any,
 passphrase(): any
}>();

const data = computed(() => [{
 slot: slots.apiKeys,
 bindings: { hint, label },
}, ...]
```

And in the template we do something with the syntax
```vue
  <template v-for="(_, name) in slots" v-slot:[name]="slotData"><slot :name="name" v-bind="slotData" /></template>
 ```
 
The how of the the template probably needs a bit of figuring out but this way we can keep the single component and also provide custom ordering of the slots.

lukicenturi: I pushed again, not exactly like this, but it's 1 component now, can you check again? @kelsos

kelsos: The implementation is basically what I had in mind.

Thank you :1st_place_medal:

PR #8997, opened 2024-12-02
Do not return zero WETH in base for extrafi balance amounts

LefterisJP:


rotkehlchen/chain/evm/decoding/extrafi/balances.py
```diff
@@ -286,6 +286,9 @@ def query_lending_reserves(
                 )
                 continue
 
+            if result[-1] <= 0:
+                continue
```
yabirgb, line 290: shouldn't this come before the try/except above?

yabirgb (approved):
lgtm

PR #8905, opened 2024-11-18
Add EFP decoder

nicholasyoder:
Closes #8620

PR #8878, opened 2024-11-06
Skip airdrop claim reminders for untracked chains

nicholasyoder:
Fixes FOREIGN KEY constraint problem reported by @lukicenturi in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1303706912369410048)

Issue #7007, opened 2023-11-27
Replace asset movements with history events


LefterisJP:
## Abstract

As part of our push towards having everything in the history events view, the asset movements (deposits & withdrawals) should be removed and consolidated into the history event view.

## Specification

Do it

LefterisJP:
Related to this and part of touching asset movements, deposit withdawals could be https://github.com/rotki/rotki/issues/6206

LefterisJP:
I have a 500+ lines diff branch for this in my local. It's outdated and probably useless by now. Will have to restart :sob:

nicholasyoder:
This should be included in 1.37

PR #9002, opened 2024-12-02
Move extra_data to HistoryBaseEntry

nicholasyoder:
Related #7007

Move extra_data to HistoryBaseEntry in preparation for creating a new AssetMovements class based on HistoryBaseEntry.

Also includes a commit fixing nitpicks from PRs #8863, #8878, and #8905.


rotkehlchen/db/upgrades/v45_v46.py
```diff
@@ -53,4 +53,14 @@ def _refresh_icons(write_cursor: 'DBCursor') -> None:
...
             icon_path.unlink(missing_ok=True)
 
+    @progress_step(description='Moving EVM event extra data to the history_events table.')
+    def _move_extra_data(write_cursor: 'DBCursor') -> None:
+        write_cursor.execute('ALTER TABLE history_events ADD COLUMN extra_data TEXT;')
+        write_cursor.execute(
+            'UPDATE history_events SET extra_data = '
+            '(SELECT extra_data FROM evm_events_info '
+            'WHERE evm_events_info.identifier = history_events.identifier);',
+        )
```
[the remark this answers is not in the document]
nicholasyoder: I've added a check for the contents of an evm event that already is present in the db, and also a check for a plain history entry. As noted in the comments in the code, the history entry has to be inserted manually instead of creating a HistoryEvent or EthStakingEvent class, since those classes now expect there to be an `extra_data` column on the `history_events` table, and would not work correctly before the upgrade is run.

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2620,6 +2622,31 @@ def test_upgrade_db_45_to_46(user_data_dir: 'Path', messages_aggregator):
...
+        existing_evm_event = cursor.execute('SELECT * FROM history_events WHERE identifier = "35"').fetchone()  # noqa: E501
+        existing_evm_event_extra_data = cursor.execute('SELECT extra_data FROM evm_events_info WHERE identifier = "35"').fetchone()[0]  # noqa: E501
+        assert existing_evm_event_extra_data == '{"airdrop_identifier": "elfi"}'
+
+    # Add a plain history event to the db to be checked after upgrade that it wasn't modified
+    # Note that it has to be manually inserted here since the functions for creating
+    # history events now expect there to be an extra_data column in history_events
+    history_event_bindings = [HistoryBaseEntryType.HISTORY_EVENT.value, 'TEST1', 0, 1, Location.KRAKEN.serialize_for_db(), 'Somewhere', A_ETH.identifier, 1, 3000, 'Just a test event', HistoryEventType.INFORMATIONAL.value, HistoryEventSubType.NONE.value]  # noqa: E501
+    with db_v5.conn.write_ctx() as write_cursor:
+        write_cursor.execute(
```
yabirgb, line 2643: nitpick: not related to this PR but the v5 I believe is a typo. Should be 45

rotkehlchen/chain/ethereum/modules/eth2/eth2.py
```diff
@@ -710,7 +709,7 @@ def refresh_activated_validators_deposits(self) -> None:
                 history_changes,
             )
             write_cursor.executemany(
-                'UPDATE evm_events_info SET extra_data=null WHERE identifier=?',
+                'UPDATE history_events SET extra_data=null WHERE identifier=?',
```
LefterisJP, line 712: You should change this. As you can see in this function `history_events` are already edited at line 707. Change the logic of this function to put the extra data change in `history_changes` and do only a single sqlite query to change the table. Not two.

rotkehlchen/history/events/structures/base.py
```diff
@@ -196,8 +205,28 @@ def _serialize_base_tuple_for_db(self) -> tuple[str, str, HISTORY_EVENT_DB_TUPLE
...
+            entry (tuple): event entry from the db to be logged on error
+            extra_data (str | None): extra data to be deserialized
+        Returns the extra data in a dict or None on error."""
+        if extra_data is None:
+            return None
+
+        try:
+            return json.loads(extra_data)
+        except json.JSONDecodeError as e:
+            log.debug(
```
LefterisJP, line 224: That's a pretty bad error. Put it as error.

```suggestion
            log.error(
```

rotkehlchen/history/events/structures/types.py
```diff
@@ -25,11 +24,11 @@
     str | None,  # notes
     str,            # type
     str,            # subtype
+    str | None,  # extra_data
```
LefterisJP, line 27: SUPER DUPER nitpick: Rest seem to be aligned? Can you align this with the one above and one below.

Sorry I know ... but my OCD triggered :rofl: 

```suggestion
    str | None,      # extra_data
```

rotkehlchen/tests/db/test_db_upgrades.py
LefterisJP: +1

LefterisJP (commented):
Some comments. But looks good overall and should be ready soon.

LefterisJP (approved):
LGTM

PR #9008, opened 2024-12-03
Gemini reward handling & events

### LINKED ISSUE #9006: Gemini Credit Card Reward Support
## Abstract

Rotki doesn't currently recognize the "Reward" asset movement type with Gemini. 

```
Error processing a gemini deposit_withdrawal asset_movement={'type': 'Reward', 'status': 'Complete', 'timestampms': 1733168176629, 'eid': 261710778017, 'currency': 'ETH', 'amount': '0.000161', 'method': 'CreditCard'}, error=Failed to deserialize asset movement category symbol. Unknown Reward
```

I would love it if these were recognized and the history event was marked as type deposit subtype reward.

## Motivation

It's a bit niche, but it also shouldn't be too difficult to implement. Currently trying to put together a PR myself.

vincenttaglia:
Closes #9006 

## Checklist

- [x] The PR modified gemini asset movement deserialization and added a related asset movement in the gemini tests

## Specification

Taking inspiration from the coinbase implementation, this PR adds a `GEMINI_EVENTS_PREFIX` and sets it as `GMNI_`. Please let me know if you have some preference related to that.

Also taking (half) inspiration from the coinbase implementation, this PR accesses the DB in the same way the coinbase implementation does, but instead of having all of the default methods point to `_query_transactions`, I just handle history events and writing to the db within `query_online_deposits_withdrawals`, and then return the asset movements to let that be handled as it is currently. Let me know if you'd like me to take it all the way and implement `_query_transactions` like the coinbase implementation.

Also, this is my first PR for Rotki, so let me know if I did anything else wrong. Hopefully I read the docs correctly and don't end up wasting your time.

### DIFF
--- a/rotkehlchen/exchanges/gemini.py
+++ b/rotkehlchen/exchanges/gemini.py
@@ -16,6 +16,8 @@ from rotkehlchen.assets.asset import AssetWithOracles
 from rotkehlchen.assets.converters import asset_from_gemini
 from rotkehlchen.constants import ZERO
 from rotkehlchen.constants.timing import GLOBAL_REQUESTS_TIMEOUT
+from rotkehlchen.db.cache import DBCacheDynamic
+from rotkehlchen.db.history_events import DBHistoryEvents
 from rotkehlchen.db.settings import CachedSettings
 from rotkehlchen.errors.asset import UnknownAsset, UnprocessableTradePair, UnsupportedAsset
 from rotkehlchen.errors.misc import RemoteError
@@ -29,7 +31,7 @@ from rotkehlchen.exchanges.utils import (
 )
 from rotkehlchen.history.deserialization import deserialize_price
 from rotkehlchen.history.events.structures.base import HistoryEvent
-from rotkehlchen.history.events.structures.types import HistoryEventType
+from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.inquirer import Inquirer
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.serialization.deserialize import (
@@ -49,7 +51,7 @@ from rotkehlchen.types import (
     TradeType,
 )
 from rotkehlchen.user_messages import MessagesAggregator
-from rotkehlchen.utils.misc import ts_now_in_ms
+from rotkehlchen.utils.misc import ts_now, ts_now_in_ms
 from rotkehlchen.utils.mixins.cacheable import cache_response_timewise
 from rotkehlchen.utils.mixins.lockable import protect_with_lock
 from rotkehlchen.utils.serialization import jsonloads_dict, jsonloads_list
@@ -62,6 +64,8 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
+
+GEMINI_EVENTS_PREFIX = 'GMNI_'
 
 
 class GeminiPermissionError(Exception):
@@ -539,12 +543,16 @@ class Gemini(ExchangeInterface):
             end_ts=end_ts,
         )
         movements = []
+        history_events: list[HistoryEvent] = []
         for entry in result:
             try:
                 timestamp = deserialize_timestamp(entry['timestampms'])
                 timestamp = Timestamp(int(timestamp / 1000))
                 asset = asset_from_gemini(entry['currency'])
                 category = self._deserialize_asset_movement_category(entry['type'])
+                if (history_event := self._deserialize_history_event(entry, timestamp,
+                                                    asset, category)) is not None:
+                    history_events.append(history_event)
                 movement = AssetMovement(
                     location=Location.GEMINI,
                     category=category,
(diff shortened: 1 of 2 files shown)


rotkehlchen/exchanges/gemini.py
```diff
@@ -507,15 +543,19 @@ def query_online_deposits_withdrawals(
             end_ts=end_ts,
         )
         movements = []
+        history_events: list[HistoryEvent] = []
```
LefterisJP, line 546: So I see what you are doing here but there is multiple small problems which I believe are fixable.

1. This will only fire when you navigate to exchange withdrawals page or rotki periodically decides to try and refresh them.
2. It mixes the processing and query of history events and of asset movements. I guess the reason is since they both come from the same endpoint. So it makes sense to not query it twice but then it would need to be reorganized.
3. To reorganize them perhaps move the logic of this into a `_query_transfers` function much like we do in coinbase for `_query_transactions`. And instead of returning the movements, inside the function just like coinbase write them to the DB.

Now some problems will come out of that. You will:

1. Need to specify in the docstring that even though start_ts/end_ts are for asset movements they may also include some history events.
2. We need whenever this patch is released to think of how to clean Gemini data. As we need to reset this query to the start. And by doing so duplications will happen. So would need to probably purge and requery all gemini data.

rotkehlchen/exchanges/gemini.py
```diff
@@ -554,9 +594,79 @@ def query_online_deposits_withdrawals(
...
+                if len(history_events) != 0:
+                    db = DBHistoryEvents(self.db)
+                    db.add_history_events(write_cursor=write_cursor, history=history_events)
+                self.db.set_dynamic_cache(
+                    write_cursor=write_cursor,
+                    name=DBCacheDynamic.LAST_QUERY_TS,
+                    value=ts_now(),
+                    location=self.location.serialize(),
+                    location_name=self.name,
+                )
```
LefterisJP, line 607: Though the function needs to change a bit more, this is not needed.

1. Stop earlier. Once you know nothing is there, no need to open a DB cursor.
2. The cache is not needed here. In coinbase, it was to remember the last transfer id of the successfully pulled transaction/transfer. That's because the api queries from last transfer. Here that's not the case and the api remembers in a generic way start/end ts for asset movements. These are the arguments to this function.
```suggestion
            if len(history_events) == 0:
                return movements

            with self.db.user_write() as write_cursor:
                    db = DBHistoryEvents(self.db)
                    db.add_history_events(write_cursor=write_cursor, history=history_events)
```

rotkehlchen/exchanges/gemini.py
```diff
@@ -554,9 +594,79 @@ def query_online_deposits_withdrawals(
...
         return movements
 
+    def _deserialize_history_event(
+            self,
+            raw_data: dict[str, Any],
+            timestamp: Timestamp,
+            asset: AssetWithOracles,
+            event_type: AssetMovementCategory,
+        ) -> HistoryEvent | None:
+        """Processes a single transaction from coinbase and deserializes it
```
LefterisJP, line 617:
```suggestion
        """Processes a single transfer from Gemini and deserializes it
```

rotkehlchen/exchanges/gemini.py
```diff
@@ -554,9 +594,79 @@ def query_online_deposits_withdrawals(
...
+
+        Can log error/warning and return None if something went wrong at deserialization
+        """
+        try:
+            if raw_data.get('status').lower() != 'complete':
+                return None
+            if raw_data.get('type').lower() != 'reward':
+                return None
```
LefterisJP, line 625: Since you are actually catching KeyError for all of these you can just keep it simple. The previous code has a bug in that if it's not found it does not raise KeyError but returns None. And `None.lower()` will raise an exception and go Boom Boom

```suggestion
            if raw_data['status'].lower() != 'complete':
                return None
            if raw_data.['type'].lower() != 'reward':
                return None
```

rotkehlchen/exchanges/gemini.py
```diff
@@ -554,9 +594,79 @@ def query_online_deposits_withdrawals(
...
+
+            amount = deserialize_asset_amount(raw_data.get('amount'))
```
LefterisJP, line 627: Let the KeyError be raised. You catch it. And it's better than forwarding None to that function.


```suggestion
            amount = deserialize_asset_amount(raw_data[amount'])
```

rotkehlchen/exchanges/gemini.py
```diff
@@ -554,9 +594,79 @@ def query_online_deposits_withdrawals(
...
+            event_subtype = HistoryEventSubType.REWARD
```
LefterisJP, line 628: As I asked you in the other issue, could this actually be Cashback?

```suggestion
            event_subtype = HistoryEventSubType.CASHBACK
```

rotkehlchen/exchanges/gemini.py
```diff
@@ -554,9 +594,79 @@ def query_online_deposits_withdrawals(
...
+                notes = ''
+            return HistoryEvent(
+                event_identifier=f'{GEMINI_EVENTS_PREFIX}{raw_data["eid"]!s}',
+                sequence_index=0,
+                timestamp=timestamp,
+                location=Location.GEMINI,
+                event_type=event_type,
+                event_subtype=event_subtype,
+                asset=asset,
+                balance=Balance(amount=amount, usd_value=ZERO),
```
LefterisJP, line 641: Default value is ZERO

```suggestion
                balance=Balance(amount=amount),
```

LefterisJP (commented):
Added a review with some questions. We can dicsuss them also in discord:  https://discord.rotki.com/

I think I have seen you there right?

Also important: Nobody in our team has Gemini. You are the only one who can test these changes. So make sure to do so and confirm it does what you need it to do. If you don't know how we are here to help in discord.

LefterisJP:
Closing due to inactivity

PR #9009, opened 2024-12-03
feat(frontend): hide small balances

### LINKED ISSUE #7498: Add functionality/checkbox for "Hide assets <1 USD"

lukicenturi:
Closes #7498

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,7 +2,7 @@
 Changelog
 =========
 
-
+* :feature: `7498` Adds the option for users to hide assets where the balance is smaller than a threshold amount.
 * :feature:`7406` rotki will now support Aura Finance staking for Balancer pools on Gnosis Chain and all supported EVM chains.
 * :feature:`7405` rotki will now support Balancer pools on Gnosis Chain and all supported EVM chains.
 * :feature:`7621` rotki will now support Curve lending, with related historical events properly decoded and correct prices and balances shown.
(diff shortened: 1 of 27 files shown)


kelsos (commented):
So the setting is unified here, it affects all the balances at once.

When we discussed about it didn't we say something about keeping it separate? @lukicenturi @yabirgb @LefterisJP.

Should there be one setting for all or per category (blockchain/manual/exchange)?

yabirgb:
We mentioned irc that it makes sense to have the same value for all of them and that is a value that affects to the table so it sshould be near it to modify it easily

kelsos:
> We mentioned irc that it makes sense to have the same value for all of them and that is a value that affects to the table so it sshould be near it to modify it easily

OK then I am gonna merged it as it is and we can tweak later

kelsos (approved):
lgtm @lukicenturi, can you please also add a change log entry before merging?

kelsos:
I am gonna adjust the .ts thing and add the changelog so I can rebase the utils PR on top in case there need to be some adjustments

PR #9012, opened 2024-12-03
style: update styling of auto-update notification

kelsos:
Closes #(issue_number)

## Checklist

## Notes 
To test, set to always true, 
https://github.com/rotki/rotki/blob/9a7b56b241203e897212e40140f30def4dd2f5bc/frontend/app/src/store/session/index.ts#L185

and then go through the vue-dev-tools and play with the downloadReady, downloading properties to see the different states

## Changes
Adds an RuiCard for the notification because the previous style didn't look that good and also makes sure that the icons don't shrink based on the text.

image
image
image


lukicenturi (approved):
LGTM!

PR #9014, opened 2024-12-03
fix: refactor message dialog to avoid state flickering

kelsos:
With the previous approach, you would see a small flicker when it reset to success from error state.

Closes #(issue_number)

## Checklist


cleanerzkp (approved):
lgtm!

PR #9017, opened 2024-12-03
improve collections' main asset identification for price queries

prettyirrelevant:
main(or group head) assets are now identified in collections by adding an explicit column `main_asset` to the `asset_collections` table.

this replaces the previous approach of using lexicographical ordering to determine the main asset.

it now looks like this:
- most collections use lexicographical ordering -- pretty dumped a csv file to back up decision.
- collection with id, 23 specifically sets DAI as its main asset.

this change is the **first step** in the `Inquirer()` refactor by enabling price queries to use a collection's main asset price for all tokens in that collection, reducing api calls.

here's the csv file https://gist.github.com/prettyirrelevant/ee78d3cbdad95eab2ca490afa43171d5


rotkehlchen/globaldb/upgrades/v9_v10.py
```diff
@@ -0,0 +1,70 @@
...
+            SET is_main_asset = 1
+            WHERE collection_id != 23
+            AND asset = (
+                SELECT MIN(asset)
+                FROM multiasset_mappings m2
+                WHERE m2.collection_id = multiasset_mappings.collection_id
+            );
+        """)
+
+    perform_globaldb_upgrade_steps(connection, progress_handler)
```
LefterisJP, line 70: use kwargs. Also make sure vacuum is True.

rotkehlchen/inquirer.py
```diff
@@ -656,15 +655,8 @@ def _find_price(
...
-        if GlobalDBHandler.asset_in_collection(collection_id=240, asset_id=from_asset.identifier):  # part of the EURe collection # noqa: E501  # todo: Super hacky. Figure out a way to generalize
-            return Inquirer._find_price(
-                from_asset=A_EUR,
-                to_asset=to_asset,
-                ignore_cache=ignore_cache,
-                skip_onchain=skip_onchain,
-                coming_from_latest_price=coming_from_latest_price,
-                match_main_currency=match_main_currency,
-            )
+        if (main_from_asset := GlobalDBHandler.get_collection_main_asset(from_asset.identifier)) is not None:  # noqa: E501
```
LefterisJP, line 658:
```suggestion
        if (collection_main_asset := GlobalDBHandler.get_collection_main_asset(from_asset.identifier)) is not None:  # noqa: E501
```

rotkehlchen/tests/unit/globaldb/test_upgrades.py
```diff
@@ -856,6 +857,50 @@ def test_upgrade_v8_v9(globaldb: GlobalDBHandler, messages_aggregator):
...
+    """Test upgrade from v9 to v10 which adds is_main_asset field to multiasset_mappings"""
+    with globaldb.conn.read_ctx() as cursor:
+        assert not column_exists(
+            cursor=cursor,
+            table_name='multiasset_mappings',
+            column_name='is_main_asset',
+        )
+        mappings_count = cursor.execute('SELECT COUNT(*) FROM multiasset_mappings').fetchone()[0]
+        # get lexicographically first asset for collection 40 before upgrade
+        collection_40_first = cursor.execute('SELECT asset FROM multiasset_mappings WHERE collection_id=40 ORDER BY asset LIMIT 1').fetchone()[0]  # noqa: E501
```
LefterisJP, line 873: Hmm the check I would want to see in this test is:

1. Before upgrade take all groups
2. After upgrade make sure the same amount of groups is there
3. Adn that also for every group the first asset is main asset
4. Except for the dAI one.

Also the schema changes but that you already did.

rotkehlchen/globaldb/updates.py
```diff
@@ -159,6 +159,8 @@ def __init__(self, msg_aggregator: 'MessagesAggregator') -> None:
         self.common_asset_details_re = re.compile(r'.*INSERT +INTO +common_asset_details\( *identifier *, *symbol *, *coingecko *, *cryptocompare *, *forked *, *started *, *swapped_for *\) *VALUES\((.*?),(.*?),(.*?),(.*?),(.*?),([^,]*?),([^,]*?)\).*')  # noqa: E501
         self.assets_collection_re = re.compile(r'.*INSERT +INTO +asset_collections\( *id *, *name *, *symbol *\) *VALUES +\(([^,]*?),([^,]*?),([^,]*?)\).*?')  # noqa: E501
         self.multiasset_mappings_re = re.compile(r'.*INSERT +INTO +multiasset_mappings\( *collection_id *, *asset *\) *VALUES +\(([^,]*?), *([\'"])([^,]+?)\2\).*?')  # noqa: E501
+        if self.local_assets_version > 31:  # multiasset_mappings table adds is_main_asset column to identify main token in collections  # noqa: E501
```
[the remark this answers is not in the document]
prettyirrelevant: > Shouldn't this be an if/else? with the code above?

oh yeah, it should. fixed that.

> Also do we do this? Don't we just keep the version of the code in sync with the latest asset update and then just make sure the asset update does not work for versions less than a specific global DB version?

there's a need for both regex patterns.

```py
re.compile(r'.*INSERT +INTO +multiasset_mappings\( *collection_id *, *asset *\) *VALUES +\(([^,]*?), *([\'"])([^,]+?)\2\).*?')  # noqa: E501
```
is needed to add assets from version 16 -> 30 and the new regex pattern cannot be used as the default since it'll break old updates.

rotkehlchen/globaldb/schema.py
```diff
@@ -164,12 +164,14 @@
...
-    FOREIGN KEY(asset) REFERENCES assets(identifier) ON UPDATE CASCADE ON DELETE CASCADE,
-    PRIMARY KEY(collection_id, asset)
+   collection_id INTEGER NOT NULL,
+   asset TEXT NOT NULL,
+   is_main_asset INTEGER NOT NULL DEFAULT 0 CHECK (is_main_asset IN (0, 1)),
+   FOREIGN KEY(collection_id) REFERENCES asset_collections(id) ON UPDATE CASCADE ON DELETE CASCADE,
+   FOREIGN KEY(asset) REFERENCES assets(identifier) ON UPDATE CASCADE ON DELETE CASCADE,
+   PRIMARY KEY(collection_id, asset)
 );
+CREATE UNIQUE INDEX IF NOT EXISTS idx_main_asset ON multiasset_mappings (collection_id) WHERE is_main_asset = 1;
```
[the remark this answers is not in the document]
LefterisJP: As discussed in the call you shoudl achieve the same with a `UNIQUE()` constraint on the table itself

rotkehlchen/globaldb/updates.py
```diff
@@ -399,22 +398,35 @@ def _process_multiasset_mapping(
...
+        identify the main token in collections. This changes both the table structure and
+        the expected number of matched groups in the regex pattern (3 groups before v30,
+        4 groups after).
+
         May raise:
         - DeserializationError
         - UnknownAsset
         """
-        mapping_match = self.multiasset_mappings_re.match(full_insert)
+        if version > 30:
```
LefterisJP, line 414: as discussed in the call let's get rid of the if/else check and let's enforce the non-backwards compatibility for assets updates.

But to not have users lose assets updates, as Yabir suggested, whenever a global DB upgrade happens for which the asset update will needs its min/max schema change, then force a remote asset update by automatically choosing "select all remote"

rotkehlchen/globaldb/upgrades/v9_v10.py
```diff
@@ -0,0 +1,73 @@
...
+    """This globalDB upgrade does the following:
+
+    1. Adds main_asset column to asset_collections table.
+
+    This upgrade took place in v1.37
+    """
+
+    @progress_step('Adding main_asset column to asset_collections')
+    def add_main_asset_column(write_cursor: 'DBCursor') -> None:
+        write_cursor.execute('CREATE TABLE multiasset_mappings_backup AS SELECT * FROM multiasset_mappings')  # noqa: E501
```
LefterisJP, line 22: Don't tell me you forgot this from the previous PR?
```suggestion
```

rotkehlchen/globaldb/upgrades/v9_v10.py
```diff
@@ -0,0 +1,73 @@
...
+            CREATE TABLE multiasset_mappings(
+                collection_id INTEGER NOT NULL,
+                asset TEXT NOT NULL,
+                FOREIGN KEY(collection_id) REFERENCES asset_collections(id) ON UPDATE CASCADE ON DELETE CASCADE,
+                FOREIGN KEY(asset) REFERENCES assets(identifier) ON UPDATE CASCADE ON DELETE CASCADE,
+                PRIMARY KEY(collection_id, asset)
+            )
+        """)  # noqa: E501
+        write_cursor.execute('INSERT INTO multiasset_mappings SELECT * FROM multiasset_mappings_backup')  # noqa: E501
+        write_cursor.execute('DROP TABLE multiasset_mappings_backup')
```
LefterisJP, line 67: As far as I can understand you just wrote the new upgrade over the old? Wtf? There is no longer any change to 
multiasset_mappings. Why are you doing this?
```suggestion
```

rotkehlchen/globaldb/upgrades/v9_v10.py
```diff
@@ -0,0 +1,73 @@
...
+            CREATE TABLE asset_collections_new(
+                id INTEGER PRIMARY KEY,
+                name TEXT NOT NULL,
+                symbol TEXT NOT NULL,
+                main_asset TEXT NOT NULL UNIQUE,
+                FOREIGN KEY(main_asset) REFERENCES assets(identifier) ON UPDATE CASCADE ON DELETE CASCADE,
+                UNIQUE(name, symbol)
+            );
+        """)  # noqa: E501
+        write_cursor.execute("""
```
LefterisJP, line 33: This seems unnecessarily complicated as an sql statmenent. Would it not be simpler and easier to read if you used python and memory too? There is no need to do it all in one sqlite execute.

If it works good, but it's complicated. And that means error prone.

rotkehlchen/tests/unit/globaldb/test_globaldb_consistency.py
```diff
@@ -86,6 +86,7 @@ def as_dict(self) -> dict[str, Any]:
         }
 
 
+@pytest.mark.skipif('CI' in os.environ, reason='skipped in CI until remote asset updates are fixed')  # noqa: E501
```
LefterisJP, line 89: make a note for that and link it here. What exactly do you need to fix? Will you fix it? If not then who will?

rotkehlchen/db/checks.py
```diff
@@ -33,7 +33,7 @@ def replacer(match: Any) -> str:
         # else
         return s
 
-    return WHITESPACE_RE.sub(replacer, text).replace(' ', '').replace('\n', '').replace('"', "'").lower()  # noqa: E501
+    return WHITESPACE_RE.sub(replacer, text).replace(' ', '').replace('\n', '').replace('"', "'").replace('\t', '').lower()  # noqa: E501
```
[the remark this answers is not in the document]
prettyirrelevant: schema comparison failed during sanity checks because the parser only handled spaces and newlines characters, whereas my db gui created the tables with tab characters(which is also valid).

rotkehlchen/tests/unit/globaldb/test_globaldb_consistency.py
prettyirrelevant: https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=89977797

rotkehlchen/globaldb/upgrades/v9_v10.py
```diff
@@ -0,0 +1,55 @@
...
+    """This globalDB upgrade does the following:
+
+    1. Adds main_asset column to asset_collections table.
+
+    This upgrade took place in v1.37
+    """
+
+    @progress_step('Adding main_asset column to asset_collections')
+    def add_main_asset_column(write_cursor: 'DBCursor') -> None:
+        write_cursor.executescript('PRAGMA foreign_keys = OFF;')
```
LefterisJP, line 22: please add a comment explaining why you do the foreign keys off and on

LefterisJP (approved):
LGTM

PR #9018, opened 2024-12-03
feat: subscript amount

cleanerzkp:
## Checklist

### DIFF
--- a/frontend/app/src/components/display/amount/AmountDisplay.vue
+++ b/frontend/app/src/components/display/amount/AmountDisplay.vue
@@ -94,6 +94,7 @@ const {
   currencyLocation,
   decimalSeparator,
   minimumDigitToBeAbbreviated,
+  subscriptDecimals,
   thousandSeparator,
   valueRoundingMode,
 } = storeToRefs(useFrontendSettingsStore());
@@ -196,6 +197,7 @@ const isNaN = computed<boolean>(() => get(displayValue).isNaN());
 // Decimal place of `realValue`
 const decimalPlaces = computed<number>(() => get(displayValue).decimalPlaces() ?? 0);
 
+const zeroDisplay = '0';
 // Set exponential notation when the `realValue` is too big
 const showExponential = computed<boolean>(() => get(displayValue).gt(1e18));
 
@@ -218,7 +220,7 @@ const renderedValue = computed<string>(() => {
 
   if (get(showExponential) && !get(abbreviate)) {
     return fixExponentialSeparators(
-      get(displayValue).toExponential(floatingPrecisionUsed, get(rounding)),
+      get(displayValue).toExponential(floatingPrecisionUsed, get(subscriptDecimals) ? undefined : get(rounding)),
       get(thousandSeparator),
       get(decimalSeparator),
     );
@@ -229,7 +231,7 @@ const renderedValue = computed<string>(() => {
     floatingPrecisionUsed,
     get(thousandSeparator),
     get(decimalSeparator),
-    get(rounding),
+    get(subscriptDecimals) ? undefined : get(rounding),
     get(abbreviate),
   );
 });
@@ -251,6 +253,10 @@ const displayCurrency = computed<Currency>(() => {
 });
 
 const comparisonSymbol = computed<'' | '<' | '>'>(() => {
+  if (get(subscriptDecimals)) {
+    return '';
+  }
+
   const value = get(displayValue);
   const floatingPrecisionUsed = get(integer) ? 0 : get(floatingPrecision);
   const decimals = get(decimalPlaces);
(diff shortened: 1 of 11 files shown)


frontend/app/src/store/settings/frontend.ts
```diff
@@ -128,8 +129,9 @@ export const useFrontendSettingsStore = defineStore('settings/frontend', () => {
     explorers,
     graphZeroBased,
     itemsPerPage,
-    language,
```
kelsos, line 131: this seems like an accidental removal that causes build errors

frontend/app/src/store/settings/frontend.ts
```diff
@@ -138,7 +140,7 @@ export const useFrontendSettingsStore = defineStore('settings/frontend', () => {
     refreshPeriod,
     renderAllNftImages,
     savedFilters,
-    selectedTheme,
+    subscriptDecimals,
```
kelsos, line 143: you also need to run `pnpm run lint:fix` after the rebase here the only changes should be this line and the one adding the const

frontend/app/tests/unit/specs/store/settings/frontend.spec.ts
```diff
@@ -64,6 +64,7 @@ describe('settings:frontend', () => {
       itemsPerPage: 10,
       amountRoundingMode: BigNumber.ROUND_UP,
       valueRoundingMode: BigNumber.ROUND_DOWN,
+      subscript_decimals: false,
```
kelsos, line 67: this line should also be removed here the other should be enough,
the undefined error is probably because of the accidental removal of the `language` from the store.

frontend/app/src/components/settings/general/amount/SubscriptToggle.vue
```diff
@@ -0,0 +1,63 @@
+<script setup lang="ts">
+import { useFrontendSettingsStore } from '@/store/settings/frontend';
+
```
kelsos, line 3: here you would need to import `SettingsOption` and `AmountInput` explicitly.

kelsos:
The PR seems to be still missing some changes from develop btw related to explicit component imports so my comments make sense after rebasing on top of latest develop

frontend/app/src/components/settings/general/AskUserUponSizeDiscrepancySetting.vue
```diff
@@ -20,42 +20,41 @@ const { t } = useI18n();
 </script>
 
 <template>
-  <SettingsOption
-    #default="{ error, success, updateImmediate }"
-    setting="askUserUponSizeDiscrepancy"
+  <SettingsItem
```
kelsos, line 23: After rebasing make sure to also import `SettingsItem`.

frontend/app/src/components/settings/AmountSettingsCategory.vue
```diff
@@ -35,16 +35,12 @@ const amountExample = bigNumberify(123456.789);
...
-    </SettingsItem>
-
-    <SettingsItem>
-      <template #title>
-        {{ t('general_settings.amount.label.currency_location') }}
+      <template #subtitle>
+        {{ t('rounding_settings.subscript.subtitle') }}
       </template>
-      <CurrencyLocationSetting />
+      <SubscriptToggle />
```
kelsos, line 43: Here after rebasing make sure to import SubscriptToggle

frontend/app/src/components/display/amount/AmountDisplay.vue
```diff
@@ -195,6 +196,8 @@ const isNaN = computed<boolean>(() => get(displayValue).isNaN());
 // Decimal place of `realValue`
 const decimalPlaces = computed<number>(() => get(displayValue).decimalPlaces() ?? 0);
 
+const zeroDisplay = computed(() => '0');
```
kelsos, line 199: Isn't a computed that always returns zero a bit redundant? a simple const should suffice

```suggestion
const zeroDisplay = '0';
```

frontend/app/src/components/display/amount/AmountDisplay.vue
```diff
@@ -264,11 +270,77 @@ const comparisonSymbol = computed<'' | '<' | '>'>(() => {
...
 
+const shouldUseSubscript = computed(() => {
+  if (!get(subscriptDecimals) || get(showExponential) || get(abbreviate)) {
+    return false;
+  }
+
+  const value = get(displayValue);
+
+  // only apply to decimal numbers between 0 and 1
+  if (!value.lt(1) || !value.gt(0) || value.isZero() || value.isNaN()) {
```
lukicenturi, line 281: isn't `!value.lt(1) || !value.gt(0)` just the same with `value.isZero()`?

frontend/app/tests/unit/specs/data/subscript-number.spec.ts
```diff
@@ -0,0 +1,196 @@
+import { mount } from '@vue/test-utils';
+import { createPinia, setActivePinia } from 'pinia';
+import { beforeEach, describe, expect, it } from 'vitest';
+import { BigNumber, bigNumberify } from '@rotki/common';
+import AmountDisplay from '@/components/display/amount/AmountDisplay.vue';
+import { useFrontendSettingsStore } from '@/store/settings/frontend';
+import { useGeneralSettingsStore } from '@/store/settings/general';
+
+describe('amountDisplay subscript', () => {
```
kelsos, line 9: Shouldn't these tests be part of amount-display.spec.ts? you can do an extra `describe` for subscript related test cases.

frontend/app/tests/unit/specs/data/subscript-number.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+    setActivePinia(pinia);
+    const frontendStore = useFrontendSettingsStore();
+    frontendStore.update({
+      ...frontendStore.settings,
+      subscriptDecimals: true,
+      thousandSeparator: ',',
+      decimalSeparator: '.',
+      abbreviateNumber: false,
+      amountRoundingMode: BigNumber.ROUND_HALF_UP,
+      valueRoundingMode: BigNumber.ROUND_HALF_UP,
```
kelsos, line 23: Is there a problem with the default values here?

frontend/app/tests/unit/specs/data/subscript-number.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+        forceCurrency: true,
+      },
+      global: {
+        plugins: [pinia],
+        stubs: {
+          CopyTooltip: { template: '<div><slot /></div>' },
+          RuiTooltip: { template: '<div><slot /></div>' },
+        },
+      },
+    });
```
kelsos, line 50: For the wrapper, let's avoid copy-pasting the same setup everywhere.

frontend/app/tests/unit/specs/data/subscript-number.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+
+    const wrapper = mount(AmountDisplay, {
+      props: {
+        value: bigNumberify('0.0000000815'),
+        noScramble: true,
+        noTruncate: true,
+        fiatCurrency: null,
+        showCurrency: 'none',
+        forceCurrency: true,
```
kelsos, line 41: I have been trying for a while but it seems that without some of these options the tests fail. 

The problem is that these are mostly non-default options, the simplest case will be the one where you only supply the value. But from what I can see it fails. We need to figure out why, the test should be able to pass with only passing value to the wrapper.

It might be that there is some promise or something that needs to complete so things calling `await flushPromises() and `await nextTick()` might be required after mounting the component.

frontend/app/tests/unit/specs/data/subscript-number.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+    const subscriptElement = wrapper.find('[data-cy="amount-display-subscript"]');
+    expect(subscriptElement.exists()).toBe(false);
+  });
+
+  it('shows no subscript for number with only 1 leading zero', () => {
+    const generalStore = useGeneralSettingsStore();
+    generalStore.update({
+      ...generalStore.settings,
+      uiFloatingPrecision: 10,
+    });
```
kelsos, line 144: the ones that have the same initialization of the floatingPrecision could be grouped with a `describe` and their own `beforeEach` just to avoid repeating common setup.

kelsos (approved):
lgtm @cleanerzkp thank you :)

Issue #3010, opened 2021-05-30
Trades, asset_movements, margin position identifiers are problematic


LefterisJP:
## Problem Definition

The trades, asset movements and margin positions data structures formulate their own unique id based on some fields that becomes the primary key in the DB. This identifier is also used to identify the data structure around the code. Example for trades:

https://github.com/rotki/rotki/blob/9fd79fa453260a57c5f9ca244a38ec44b7f55825/rotkehlchen/exchanges/data_structures.py#L117-L131

The reasoning for having a field that depends on other fields of the event is to have it unique. So that if someone tries to input an identical trade it's rejected.

The problem here is that this is brittle and can break at any point if the database is altered without also changing the identifier that is saved in the DB.

Despite that we were always taking care of that so this problem never manifested.
But now with the addition of [asset merging](https://rotki.readthedocs.io/en/latest/usage_guide.html#merging-two-assets) and generally the ability for fields of the DB to change due to cascading updates we have a problem.

A field may change, resulting in a different identifier calculation but the original identifier in the DB won't be changed. This manifests in many problems most prominent of which would be the inability to edit the trade from the frontend.

## Task

Figure out a way out of this mess. Perhaps it should be solved in 2 steps or in an either/or way.

1. When querying for any of those events from the DB perform a check the the DB identifier is correct and matches the fields. And if it does not replace the id in the DB. 
2. Figure out a better way of uniquely identifying the events and siwtch to a different system.

(1) could be a short term solution to fix this problem but would make DB querying for them a bit slower.
(2) would need thinking but would result in a better long term solution?

PR #9020, opened 2024-12-03
Implement new AssetMovement class

nicholasyoder:
Related #7007

### DIFF
--- a/rotkehlchen/history/events/structures/asset_movement.py
+++ b/rotkehlchen/history/events/structures/asset_movement.py
@@ -1,0 +1,203 @@
+import logging
+from typing import TYPE_CHECKING, Any, Literal
+
+from rotkehlchen.accounting.mixins.event import AccountingEventType
+from rotkehlchen.accounting.structures.balance import Balance
+from rotkehlchen.assets.asset import Asset
+from rotkehlchen.constants import ZERO
+from rotkehlchen.constants.location_details import get_formatted_location_name
+from rotkehlchen.errors.serialization import DeserializationError
+from rotkehlchen.exchanges.data_structures import hash_id
+from rotkehlchen.history.events.structures.base import (
+    HISTORY_EVENT_DB_TUPLE_WRITE,
+    HistoryBaseEntry,
+    HistoryBaseEntryType,
+)
+from rotkehlchen.history.events.structures.types import (
+    HistoryEventSubType,
+    HistoryEventType,
+)
+from rotkehlchen.logging import RotkehlchenLogsAdapter
+from rotkehlchen.serialization.deserialize import deserialize_fval
+from rotkehlchen.types import (
+    Location,
+    TimestampMS,
+)
+from rotkehlchen.utils.misc import ts_ms_to_sec
+
+if TYPE_CHECKING:
+    from more_itertools import peekable
+
+    from rotkehlchen.accounting.mixins.event import AccountingEventMixin
+    from rotkehlchen.accounting.pot import AccountingPot
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+class AssetMovement(HistoryBaseEntry):
+    """Asset movement event representing deposits and withdrawals on exchanges."""
+
+    def __init__(
+            self,
+            timestamp: TimestampMS,
+            location: Location,
+            event_type: Literal[HistoryEventType.DEPOSIT, HistoryEventType.WITHDRAWAL],
+            asset: Asset,
+            balance: Balance,
+            identifier: int | None = None,
+            event_identifier: str | None = None,
+            unique_id: str | None = None,
+            extra_data: dict[str, Any] | None = None,
+            is_fee: bool = False,
+    ) -> None:
+        """
+        `unique_id`: Unique identifier for this asset movement.
+            Either the exchange transaction id or the associated onchain transaction hash.
+            Used in conjunction with location to generate the event identifier.
+        `is_fee`: Controls whether this represents a fee event.
+        """
+        location_name = get_formatted_location_name(location)
+        asset_symbol = asset.resolve_to_asset_with_symbol().symbol
+        if is_fee:
+            sequence_index = 1
+            event_subtype = HistoryEventSubType.FEE
+            notes = f'Pay {balance.amount} {asset_symbol} as {location_name} {str(event_type).lower()} fee'  # noqa: E501
+        else:
+            sequence_index = 0
+            if event_type == HistoryEventType.DEPOSIT:
+                event_subtype = HistoryEventSubType.DEPOSIT_ASSET
+                notes = f'Deposit {balance.amount} {asset_symbol} to {location_name}'
+            else:
+                event_subtype = HistoryEventSubType.REMOVE_ASSET
+                notes = f'Withdraw {balance.amount} {asset_symbol} from {location_name}'
+
+        super().__init__(
+            event_identifier=event_identifier if event_identifier is not None else self._create_event_identifier(  # noqa: E501
+                location=location,
+                timestamp=timestamp,
+                asset=asset,
+                balance=balance,
+                unique_id=unique_id,
+            ),
+            sequence_index=sequence_index,
+            timestamp=timestamp,
+            location=location,
+            event_type=event_type,
+            event_subtype=event_subtype,
+            asset=asset,
+            balance=balance,
+            notes=notes,
+            identifier=identifier,
+            extra_data=extra_data,
+        )
+
+    @staticmethod
+    def _create_event_identifier(
+            location: Location,
+            timestamp: TimestampMS,
+            asset: Asset,
+            balance: Balance,
+            unique_id: str | None,
+    ) -> str:
+        """Create a unique event identifier from the given parameters.
+        `unique_id` is a transaction id from the exchange, which in combination with the
+        location makes a unique event identifier. If this is not available, the location,
+        timestamp, asset, and balance must all be combined to ensure a unique identifier.
+        """
+        if unique_id is not None:
+            return hash_id(str(location) + unique_id)
+
+        return hash_id(
+            str(location) +
+            str(timestamp) +
+            asset.identifier +
+            str(balance.amount),
+        )
+
+    @property
+    def entry_type(self) -> HistoryBaseEntryType:
+        return HistoryBaseEntryType.ASSET_MOVEMENT_EVENT
+
+    def serialize_for_db(self) -> tuple[tuple[str, str, HISTORY_EVENT_DB_TUPLE_WRITE]]:
+        return (self._serialize_base_tuple_for_db(),)
+
+    @classmethod
+    def deserialize_from_db(cls: type['AssetMovement'], entry: tuple) -> 'AssetMovement':
+        """Deserialize an AssetMovement DB tuple.
+        May raise:
+        - DeserializationError
+        - UnknownAsset
+        """
+        amount = deserialize_fval(entry[7], 'amount', 'evm event')
+        usd_value = deserialize_fval(entry[8], 'usd_value', 'evm event')
+        return cls(
+            identifier=entry[0],
+            event_identifier=entry[1],
+            timestamp=TimestampMS(entry[3]),
+            location=Location.deserialize_from_db(entry[4]),
+            event_type=HistoryEventType.deserialize(entry[10]),  # type: ignore  # should always be correct from the DB
+            is_fee=(HistoryEventSubType.deserialize(entry[11]) == HistoryEventSubType.FEE),
+            asset=Asset(entry[6]).check_existence(),
+            balance=Balance(amount, usd_value),
+            extra_data=cls.deserialize_extra_data(entry=entry, extra_data=entry[12]),
+        )
+
+    @classmethod
+    def deserialize(cls: type['AssetMovement'], data: dict[str, Any]) -> 'AssetMovement':
+        base_data = cls._deserialize_base_history_data(data)
+        if (event_type := base_data['event_type']) not in {
+            HistoryEventType.DEPOSIT,
+            HistoryEventType.WITHDRAWAL,
+        }:
+            raise DeserializationError(
+                f'Unsupported asset movement event type {event_type}. '
+                f'Expected DEPOSIT or WITHDRAWAL',
+            )
+
+        return cls(
+            identifier=base_data['identifier'],
+            event_identifier=base_data['event_identifier'],
+            timestamp=base_data['timestamp'],
+            location=base_data['location'],
+            event_type=event_type,  # type: ignore  # just confirmed it's a DEPOSIT or WITHDRAWAL above
+            is_fee=(base_data['event_subtype'] == HistoryEventSubType.FEE),
+            asset=base_data['asset'],
+            balance=base_data['balance'],
+            extra_data=base_data['extra_data'],
+        )
+
+    def __repr__(self) -> str:
+        return f'AssetMovement({", ".join(self._history_base_entry_repr_fields())})'
+
+    # -- Methods of AccountingEventMixin
+
+    @staticmethod
+    def get_accounting_event_type() -> AccountingEventType:
+        return AccountingEventType.ASSET_MOVEMENT
+
+    def process(
+            self,
+            accounting: 'AccountingPot',
+            events_iterator: "peekable['AccountingEventMixin']",  # pylint: disable=unused-argument
+    ) -> int:
+        if self.asset.identifier == 'KFEE' or self.balance.amount == ZERO:
+            # There is no reason to process deposits of KFEE for kraken as it has only value
+            # internal to kraken and KFEE has no value and will error at cryptocompare price query.
+            return 1
+
+        if self.event_subtype == HistoryEventSubType.FEE:
+            accounting.add_out_event(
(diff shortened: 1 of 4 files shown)


rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,200 @@
...
+            extra_data: dict[str, Any] | None = None,
+            address: str | None = None,
+            transaction_id: str | None = None,
+            link: str | None = None,
+    ) -> None:
+        """
+        Optional args stored in extra_data:
+        `address`: The source address for deposits and the destination address for withdrawals.
+        `transaction_id`: Transaction id on the exchange or the associated onchain transaction hash.
+        `link`: For exchange asset movements this should be the exchange unique identifier.
```
LefterisJP, line 63: The `link` name was a misnomer. Call it `unique_id`

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,200 @@
...
+            if value is None:
+                continue
+
+            if extra_data is None:
+                extra_data = {}
+
+            extra_data[key] = value
+
+        super().__init__(
+            event_identifier=event_identifier if event_identifier is not None else hash_id(
```
LefterisJP, line 82: so no this is not gonna work. This will re-create the same problem I posted about in the morning that we want to avoid: https://github.com/rotki/rotki/issues/3010

We can't have such unpredictable identifiers. We need to make sure that events never get dupicated and uniqueidentifiers are unique.

The how is a problem which you need to think about.

If there is a `unique_id` from the exchange use exchangename + uniqueid.
If not then you need to get creative.

And that creativeness should definitely be encoded in a function so it can be reused in other places where the unique id would need to be recreated.

Perhaps enough would be: location | timestamp | asset | amount.

Only problem then would be if you have somehow two duplicate deposit/withdrawals at the same milisecond for the same asset and the same amount. Which perhaps should indeed never happen?

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,200 @@
...
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+class AssetMovement(HistoryBaseEntry):
+    """Asset movement event representing deposits and withdrawals on exchanges."""
+
+    def __init__(
+            self,
+            sequence_index: int,
```
LefterisJP, line 43: Why do you need the sequence index here? Shouldn't it always be 0 or 1? So 0 if it's a deposit or withdrawal. And 1 for the fee? So that sequence index would actually depend on other constructor arguments and as such should not be an argument?

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,200 @@
...
+            timestamp: TimestampMS,
+            location: Location,
+            event_type: HistoryEventType,
+            event_subtype: HistoryEventSubType,
```
LefterisJP, line 47: In a similar fashion here, shouldn't it just be three options here? Which would translate to specific type/subtypes? Deposit, withdrawal or fee of any of the two?


I mean it's not bad to have them as arguments. But depending on where you will use this there will need to be some logic on what to use. So it coudl all be in the constructor and simplify the amount of arguments.

Really depends on where/how this is used.
```suggestion
```

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,200 @@
...
+            asset: Asset,
+            balance: Balance,
+            location_label: str | None = None,
```
LefterisJP, line 50: what is this used for here? If you don't know then you don't need to provide it as argument. One thing we used this for was the particular exchange name. So kraken1, coinbase2 etc. since user can have multiple api keys for exchange. If so then use that name in the argument.

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,200 @@
...
+            notes: str | None = None,
```
LefterisJP, line 51: Where would specific notes be given at construction time? Do you need this argument and the extra check in the constructor? If not then remove it. Add it only if you ever end up using it. Don't complicate stuff trying to anticipate things that may never be needed

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,200 @@
...
+            identifier: int | None = None,
+            event_identifier: str | None = None,
+            extra_data: dict[str, Any] | None = None,
+            address: str | None = None,
+            transaction_id: str | None = None,
+            link: str | None = None,
```
LefterisJP, line 57: rename to exchange_unique_id or something like that

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,200 @@
...
+            location: Location,
+            event_type: HistoryEventType,
+            event_subtype: HistoryEventSubType,
+            asset: Asset,
+            balance: Balance,
+            location_label: str | None = None,
+            notes: str | None = None,
+            identifier: int | None = None,
+            event_identifier: str | None = None,
+            extra_data: dict[str, Any] | None = None,
```
LefterisJP, line 54: Same here. Would you need to provide extra data at construction time? Only thing I can think of is the potential memo some exchange deposit/withdrawals have for chains like monero. 

But then what's the point of giving some args as part of extra data and some as normal args in the constructor (address, transaction_id and exchange_uniqueid)? Do one or the other. This way the logic inside the constructor also becomes simpler.

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,200 @@
...
+                return f'Deposit {amount} {asset_symbol} to {location_name}'
+            elif event_subtype == HistoryEventSubType.FEE:
+                return f'Deposit fee of {amount} {asset_symbol} on {location_name}'
+        elif event_type == HistoryEventType.WITHDRAWAL:
+            if event_subtype == HistoryEventSubType.REMOVE_ASSET:
+                return f'Withdraw {amount} {asset_symbol} from {location_name}'
+            elif event_subtype == HistoryEventSubType.FEE:
+                return f'Withdrawal fee of {amount} {asset_symbol} on {location_name}'
+
+        log.error('Unknown asset movement type/subtype encountered, leaving event notes blank.')
```
LefterisJP, line 130: But this should actually never happen. Since we control how those are initialized. Right?

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,200 @@
...
+        if event_type == HistoryEventType.DEPOSIT:
+            if event_subtype == HistoryEventSubType.DEPOSIT_ASSET:
+                return f'Deposit {amount} {asset_symbol} to {location_name}'
+            elif event_subtype == HistoryEventSubType.FEE:
+                return f'Deposit fee of {amount} {asset_symbol} on {location_name}'
+        elif event_type == HistoryEventType.WITHDRAWAL:
+            if event_subtype == HistoryEventSubType.REMOVE_ASSET:
+                return f'Withdraw {amount} {asset_symbol} from {location_name}'
+            elif event_subtype == HistoryEventSubType.FEE:
+                return f'Withdrawal fee of {amount} {asset_symbol} on {location_name}'
```
LefterisJP, line 128:
```suggestion
                return f'Pay {amount} {asset_symbol} as {location_name} withdrawal fee'
```

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,200 @@
...
+            asset_symbol: str,
+            amount: 'FVal',
+    ) -> str | None:
+        """Create event notes from the provided data.
+        Returns the generated notes or None on error."""
+        if event_type == HistoryEventType.DEPOSIT:
+            if event_subtype == HistoryEventSubType.DEPOSIT_ASSET:
+                return f'Deposit {amount} {asset_symbol} to {location_name}'
+            elif event_subtype == HistoryEventSubType.FEE:
+                return f'Deposit fee of {amount} {asset_symbol} on {location_name}'
```
LefterisJP, line 123:
```suggestion
                return f'Pay {amount} {asset_symbol} as {location_name} deposit fee'
```

LefterisJP (commented):
some initial comments

rotkehlchen/history/events/structures/asset_movement.py
yabirgb: Yes, we only have deposit/withdrawal/fee right?

rotkehlchen/history/events/structures/asset_movement.py
yabirgb: here it would be unique_id, maybe is good idea to have `location + unique_id` to prevent conflicts. Then I'm not sure what exchanges don't provide an id. Would need to check. Only thing that might be dangerous is imports from csvs where we need to check if we have ids for everything.

rotkehlchen/history/events/structures/asset_movement.py
nicholasyoder: Added a `memo` argument and removed the `extra_data` argument. `memo` can be used as you mentioned and also for stuff like [this](https://github.com/rotki/rotki/blob/develop/rotkehlchen/data_import/importers/binance.py#L397) that was being put in `link` a few places.

rotkehlchen/history/events/structures/asset_movement.py
nicholasyoder: Changed this to use location + unique_id when unique_id is present, and otherwise create it from location, timestamp, asset and amount as suggested.

rotkehlchen/history/events/structures/asset_movement.py
nicholasyoder: Yes, should only ever be deposit, deposit fee, withdrawal, withdrawal fee here. I've changed it to use AssetMovementCategory which was used in the old AssetMovement class, with added values for deposit fees and withdrawal fees.

rotkehlchen/history/events/structures/asset_movement.py
nicholasyoder: As discussed in discord, I've changed this to have the `extra_data` argument again, and removed most of the other args that were being put in `extra_data`. I've kept the `unique_id` arg, since it is needed for generating the event identifier, but its not being added to `extra_data` anymore. If the value being used for `unique_id` is needed in `extra_data` then it can be passed in there as well, likely under a different key such as `transaction_hash` or something.

LefterisJP (commented):
Left a review.

Give special note to the comment about duplication.

Also tests failed.

rotkehlchen/db/schema.py
```diff
@@ -138,6 +138,10 @@
 INSERT OR IGNORE INTO asset_movement_category(category, seq) VALUES ('A', 1);
 /* Withdrawal Category */
 INSERT OR IGNORE INTO asset_movement_category(category, seq) VALUES ('B', 2);
+/* Deposit Fee Category */
```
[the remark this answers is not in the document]
nicholasyoder: Yes, this is unneeded. Not sure why I didn't think about that. Removed now.

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,207 @@
...
+
+    @staticmethod
+    def _create_event_identifier(
+            location: Location,
+            timestamp: TimestampMS,
+            asset: Asset,
+            balance: Balance,
+            unique_id: str | None,
+    ) -> str:
+        """Create a unique event identifier from the given parameters."""
```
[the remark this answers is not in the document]
nicholasyoder: Updated the docstring there. In regard to the duplication problem - yes I realize this doesn't fix that. I haven't been able to come up with a better way to do it yet though, so just doing it like this for now. Will keep thinking about how this should be done.

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,210 @@
...
+    from more_itertools import peekable
+
+    from rotkehlchen.accounting.mixins.event import AccountingEventMixin
+    from rotkehlchen.accounting.pot import AccountingPot
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+class AssetMovementCategory(Enum):
```
LefterisJP, line 38: Why do you re-declare this? This already exists at rotkehlchen/types.py.  Though you don't really need his here at all

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+class AssetMovement(HistoryBaseEntry):
+    """Asset movement event representing deposits and withdrawals on exchanges."""
+
+    def __init__(
+            self,
+            timestamp: TimestampMS,
+            location: Location,
+            category: AssetMovementCategory,
```
LefterisJP, line 51: Again just like previous review I am repeating. Why do you need this? Just pass `event_type: Literal[HistoryEventType.DEPOSIT, HistoryEventType.WITHDRAWAL]` here.

You won't need this articial new ENUM and also save the if checks

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,210 @@
...
+            asset: Asset,
+            balance: Balance,
+            identifier: int | None = None,
+            event_identifier: str | None = None,
+            unique_id: str | None = None,
+            extra_data: dict[str, Any] | None = None,
+            is_fee: bool = False,
+    ) -> None:
+        """
+        `category`: Type of asset movement
```
LefterisJP, line 61: will need to change the docstring again

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,210 @@
...
+        `unique_id`: Unique identifier for this asset movement.
+            Either the exchange transaction id or the associated onchain transaction hash.
+            Used in conjunction with location to generate the event identifier.
+        `is_fee`: Controls whether this represents a fee event.
+        """
+        location_name = get_formatted_location_name(location)
+        asset_symbol = asset.resolve_to_asset_with_symbol().symbol
+        if is_fee:
+            sequence_index = 1
+            event_type = HistoryEventType.DEPOSIT if category == AssetMovementCategory.DEPOSIT else HistoryEventType.WITHDRAWAL  # noqa: E501
```
LefterisJP, line 71: won't need this if check

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,210 @@
...
+            event_subtype = HistoryEventSubType.FEE
+            notes = f'Pay {balance.amount} {asset_symbol} as {location_name} {str(category).lower()} fee'  # noqa: E501
+        else:
```
LefterisJP, line 74: With what I suggested just set event_type to the given event type and then depending on its value and the `is_fee` value set subtype, and notes.

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,210 @@
...
+
+        return hash_id(
+            str(location) +
+            str(timestamp) +
+            asset.identifier +
+            str(balance.amount),
+        )
+
+    @staticmethod
+    def _get_category_from_event_type(event_type: HistoryEventType) -> AssetMovementCategory:
```
LefterisJP, line 129: you don't need this

rotkehlchen/history/events/structures/asset_movement.py
nicholasyoder: Removed this, so its kind of irrelevant now, but just to explain my reason - I thought it would be better to use a plain enum rather than using the existing DBCharEnumMixIn... but I suppose that would have been converted to a plain enum later when the db part gets removed anyway - so yeah, I guess this was unnecessary either way.

rotkehlchen/history/events/structures/asset_movement.py
nicholasyoder: Yes, using Literal with HistoryEventType is much better. I should have thought of this.

Note that I also added a check for the event type in the deserialization functions, partly because mypy complained about the type now that its restricted with Literal, but also so that it raises an error if somehow it does get the wrong event type during deserialization.

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -0,0 +1,208 @@
...
+
+    @property
+    def entry_type(self) -> HistoryBaseEntryType:
+        return HistoryBaseEntryType.ASSET_MOVEMENT_EVENT
+
+    def serialize_for_db(self) -> tuple[tuple[str, str, HISTORY_EVENT_DB_TUPLE_WRITE]]:
+        return (self._serialize_base_tuple_for_db(),)
+
+    @staticmethod
+    def _check_deserialized_event_type(
```
LefterisJP, line 126: So don't have this into a separate function. You don't need this in many places.

1. Deserializing from the outside, so `deserialize` yes you would need to check. But just add it as a simple if check.
2. Deserializing from the DB you should not need to do that as the DB data have already been checked at addition. We don't recheck DB data when reading it out of the DB.

LefterisJP (approved):
LGTM

PR #9028, opened 2024-12-05
Fix serialization

yabirgb:
- The PR adding the threshold to balances made a change that broke the logic
for serialization that we have since we have checks for the type of chain.

In addition the test didn't catch it because the balances for bitcoin weren't set

- Acknowledge polars warning about fork (doesn't affect us afaik)

```
Using fork() can cause Polars to deadlock in the child process.
  In addition, using fork() with Python in general is a recipe for mysterious
  deadlocks and crashes.

  The most likely reason you are seeing this error is because you are using the
  multiprocessing module on Linux, which uses fork() by default. This will be
  fixed in Python 3.14. Until then, you want to use the "spawn" context instead.

  See https://docs.pola.rs/user-guide/misc/multiprocessing/ for details.
```

### DIFF
--- a/rotkehlchen/tests/conftest.py
+++ b/rotkehlchen/tests/conftest.py
@@ -28,6 +28,7 @@ if TYPE_CHECKING:
     from vcr import VCR
 
 
+os.environ['POLARS_ALLOW_FORKING_THREAD'] = '1'
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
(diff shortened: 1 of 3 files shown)


rotkehlchen/tests/conftest.py
```diff
@@ -28,6 +28,7 @@
     from vcr import VCR
 
 
+os.environ['POLARS_ALLOW_FORKING_THREAD'] = '1'
```
LefterisJP, line 31: Why do you need this?

yabirgb: It appears every time that you run a test and since it is big it's annoying. We are aware of it so it is fine to silence it 
imagen

LefterisJP: I will add a comment here in my PR

LefterisJP (approved):
LGTM

PR #9042, opened 2024-12-06
Update ExchangeInterface and Bitmex for new AssetMovement

nicholasyoder:
Related #7007

* Updates ExchangeInterface to better support querying history events, replacing the old `query_deposits_withdrawals` and `query_income_loss_expense` with `query_history_events`.
Note that `query_income_loss_expense` is hardly in use at all currently. Apparently it was added for something that never got implemented, or what was using it has been removed.
Also note that there are still some of the old functions present, to prevent errors from the exchanges that have not been updated yet.
* Converts only the bitmex exchange to use the new AssetMovement class and history events query function. I chose this exchange simply because it was a smaller, relatively simple one.
* Does also touch some files for other exchanges due to renaming of the existing `query_history_events` function to `query_other_history_events`. I think renaming this was the right approach to take, but not 100% sure, so looking for feedback on that. 
I had initially thought that perhaps I should use this existing function to query the new AssetMovements, but decided against it as it seems to be used for more edge cases, and doesn't support start_ts/end_ts/only_cache like the other query functions.


rotkehlchen/exchanges/exchange.py
```diff
@@ -214,10 +214,10 @@ def query_online_deposits_withdrawals(
         """Queries the exchange's API for the asset movements of the user
 
         Should be implemented in subclasses.
+
+        Deprecated, will be replaced by query_online_history_events
```
LefterisJP, line 218: Why do you want to deprecate this?

Some exchanges query "all transactions".
Some exchanges have different endpoints for each type such as trade, deposit/withdrawal etc.

This function is eventually called by the frontend at asset movements view to refresh the user's asset movements. So by removing that frontend view you also think we should conslidate those in the exchange interface?

rotkehlchen/exchanges/exchange.py
```diff
@@ -230,14 +230,31 @@ def query_online_income_loss_expense(
         Has to be implemented by exchanges if they have anything exchange specific
 
         For example coinbase
+
+        Deprecated, will be replaced by query_online_history_events
```
LefterisJP, line 234: Hm ... ehm ... hmm ... who uses this? I did a really quick check and it seems unused? If so you can remove the entire thing in a different commit too.

rotkehlchen/exchanges/exchange.py
```diff
@@ -230,14 +230,31 @@ def query_online_income_loss_expense(
...
+            self,
+            start_ts: Timestamp,
+            end_ts: Timestamp,
+    ) -> Sequence['HistoryBaseEntry']:
+        """Queries the exchange's API for history events of the user
+
+        Should be implemented in subclasses.
+        Returns events based on HistoryBaseEntry, such as HistoryEvent, AssetMovement, etc.
+        """
+        # Should raise NotImplementedError here after all exchanges are updated
```
LefterisJP, line 248: What's this?

rotkehlchen/exchanges/exchange.py
```diff
@@ -230,14 +230,31 @@ def query_online_income_loss_expense(
...
+        return []
+
+    def query_other_history_events(self) -> None:
```
LefterisJP, line 251: I saw what you wrote in the PR description but you should not rename old function and use a new function.

Should be combined.

This will now be called for each exchange the moment you enter the history view. It needs to be combined.

rotkehlchen/exchanges/exchange.py
```diff
@@ -457,6 +478,52 @@ def query_income_loss_expense(
...
+                )
+
+            for query_start_ts, query_end_ts in ranges_to_query:
+                new_events = self.query_online_history_events(
+                    start_ts=query_start_ts,
+                    end_ts=query_end_ts,
+                )
+                with self.db.user_write() as write_cursor:
+                    if len(new_events) != 0:
+                        db.add_history_events(write_cursor, new_events)
```
LefterisJP, line 512: kwargs

rotkehlchen/exchanges/exchange.py
nicholasyoder: > So by removing that frontend view you also think we should conslidate those in the exchange interface?

Yes, I assumed that since that view is going away there would be no point in using different functions for AssetMovements versus other history events, since they will all need queried at the same time anyway. And this function can be used for other things we may convert to history events in the future as well, such as Trades.

rotkehlchen/exchanges/exchange.py
nicholasyoder: Coinbase was the only exchange using this function at all, and it was easily converted to use `query_online_history_events` instead, so I've added a commit removing `query_income_loss_expense` and `query_online_income_loss_expense` completely.

rotkehlchen/exchanges/exchange.py
nicholasyoder: Combined it. Now exchanges that need this will simply reimplement `query_history_events` with custom logic.

rotkehlchen/exchanges/exchange.py
nicholasyoder: This is no longer applicable. As mentioned here: https://github.com/rotki/rotki/pull/9042#discussion_r1876303157 `query_history_events` may be reimplemeted with custom logic that won't necessarily use this function, which means this function may not need implemented in all subclasses after all.

But to explain what it was supposed to mean - this comment was simply to remind us that once all exchanges have been updated and implement this function, it should do this here:
```
raise NotImplementedError(
    'query_online_history_events() should only be implemented by subclasses',
)
```
Like is done in `query_online_margin_history` for example, to prevent the addition of exchanges without implementing this function.

rotkehlchen/api/rest.py
```diff
@@ -3831,11 +3831,7 @@ def get_associated_locations(self) -> Response:
     def query_online_events(self, query_type: HistoryEventQueryType) -> dict[str, Any]:
         """Queries the specified event type for any new events and saves them in the DB"""
         try:
-            if query_type == HistoryEventQueryType.EXCHANGES:
-                self.rotkehlchen.exchange_manager.query_history_events()
-                return OK_RESULT
-
-            # else we query eth staking events
+            # query eth staking events
```
LefterisJP, line 3834: As I asked here something does not make sense: https://discord.com/channels/657906918408585217/1080783409640648765/1315760941425496074

We need to decide how events for exchange deposit/withdrawals (and later other types) of stuff will be queried by the frontend.

Deleting this I am not sure if it's the right or wrong approach.

What you need to make sure is that there is a way for the frontend to refresh the deposit/withdrawals of each exchange (query latest) per exchange via an endpoint.

Not just history events, as that's too generic. Same we would have for trades etc.

I asked them for confirmation in discord as per the link

rotkehlchen/exchanges/exchange.py
LefterisJP: As I wrote in the other comment, the view may go away but the way we query stuff does not mean that it will be so condensed as one api query for all history events. You need to have the ability to be a bit more flexible.

For example you would need to be able to at least refresh/query latest events of a single exchange at least. When you go to the exchange specific view, or when you add a new api key for that exchange for example.

rotkehlchen/tests/api/test_history.py
```diff
@@ -208,7 +208,7 @@ def test_query_history_remote_errors(rotkehlchen_api_server_with_exchanges: 'API
     warnings = rotki.msg_aggregator.consume_warnings()
     assert len(warnings) == 0
     errors = rotki.msg_aggregator.consume_errors()
-    assert len(errors) == 2
+    assert len(errors) == 3
```
LefterisJP, line 211: why this change?

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -596,43 +592,3 @@ def mock_check_asset_existence(identifier: str, query_packaged_db: bool = True)
         'Could not deserialize one or more history event(s). '
         'Try redecoding the event(s) or check the logs for more details.',
     ]
-
-
-@pytest.mark.parametrize('number_of_eth_accounts', [0])
```
LefterisJP, line 601: we will need an endpoint like this I think. Don't think it's a good idea to remove it. Unless you have a way to satisfy the requirements I explained in the other comments

rotkehlchen/api/rest.py
nicholasyoder: I've added a new endpoint for querying history events for specific exchanges.

> Not just history events, as that's too generic. Same we would have for trades etc.

It would seem you're saying here that we need ways to query specific types of history events, but you mentioned in discord that 'having one generic per exchange is fine', so the new endpoint simply queries all history events from a specific exchange via the `query_history_events` function in the exchange interface.

rotkehlchen/tests/api/test_history.py
nicholasyoder: The error being checked here is this:
```
Failed to query kraken ledger between 0 and 1733842067. Kraken remote error
```
Which happens in `query_kraken_ledgers` which is called in these functions:
* `query_online_trade_history`
* `query_online_deposits_withdrawals`
* `query_history_events`

There were two of these errors originally because the `query_history_with_callbacks` function was being used, which queried both trades and deposit/withdrawals. But now `query_history_events` has also been added to `query_history_with_callbacks` and now this error appears three times, due to all three of the above functions being called.

After the kraken exchange interface is updated to use the new AssetMovements, `query_online_deposits_withdrawals` will be merged with `query_history_events` and this should go back to only two errors here.

rotkehlchen/tests/api/test_history_base_entry.py
nicholasyoder: Added this test back in, but modified it to use the new exchange specific endpoint.

rotkehlchen/exchanges/exchange.py
nicholasyoder: As mentioned here: https://github.com/rotki/rotki/pull/9042#discussion_r1878299682 I've added an endpoint for querying latest events from a single exchange.

rotkehlchen/exchanges/exchange.py
```diff
@@ -214,32 +214,23 @@ def query_online_deposits_withdrawals(
...
-    ) -> list['HistoryEvent']:
-        """Queries the exchange's API for simple history events of the user
+    ) -> Sequence['HistoryBaseEntry']:
+        """Queries the exchange's API for history events of the user
 
-        Should be implemented in subclasses.
-        Has to be implemented by exchanges if they have anything exchange specific
-
-        For example coinbase
+        Should be implemented in subclasses, unless query_history_events is reimplemented with
```
LefterisJP, line 229: what do you mean by this?

rotkehlchen/exchanges/manager.py
```diff
@@ -306,11 +308,21 @@ def get_user_binance_pairs(self, name: str, location: Location) -> list[str]:
...
-            exchange.query_history_events()
+        if (exchange := self.get_exchange(name=name, location=location)) is None:
+            log.error(
+                'Failed to query history events for unknown exchange. '
+                f'Location: {location!s}, Name: {name}',
+            )
+            return
+
+        exchange.query_history_events(
+            start_ts=Timestamp(0),
```
LefterisJP, line 325: Hmmm shouldn't it query from latest event until now? Or is this logic implemented inside the function? If so then why do you need the start_ts passed as an argument here? I am a bit confused. Doesn't make much sense to me.

Would you not want to query just query_online_history_events here?

rotkehlchen/tests/api/test_history.py
LefterisJP: okay can you add a comment then explaining that this will be reverted soon and why. And delete it when making it go back to two errors.

rotkehlchen/exchanges/exchange.py
nicholasyoder: `query_online_history_events` normally needs implemented in subclasses since its called by `query_history_events`. The exception to this is when the subclass reimplements `query_history_events` with custom logic (for instance in kraken) that may not necessarily use the `query_online_history_events` function.

rotkehlchen/exchanges/manager.py
nicholasyoder: Simplified the `query_history_events` function, so now no arguments are passed here. The other functions similar to `query_history_events` in exchange interface should be simplified as well, but leaving that for now as discussed in discord.

Note that a couple tests had a check for the ability to query specific ranges via query_history_events (or a function it replaces). So these checks have been removed.

> Would you not want to query just query_online_history_events here?

It does need to use the query ranges logic in query_history_events so no, directly calling query_online_history_events here wouldn't work.

rotkehlchen/tests/api/test_history.py
nicholasyoder: added a comment

rotkehlchen/api/v1/schemas.py
```diff
@@ -1577,6 +1577,11 @@ class ExchangesDataResourceSchema(Schema):
     location = LocationField(limit_to=ALL_SUPPORTED_EXCHANGES, load_default=None)
 
 
+class ExchangeEventsQuerySchema(AsyncQueryArgumentSchema):
+    name = fields.String()
```
LefterisJP, line 1581: **Always** specify these stuff. Do not rely on default arguments which may change one release to the next.

```suggestion
    name = fields.String(required=False)
```

rotkehlchen/exchanges/exchange.py
```diff
@@ -412,50 +404,32 @@ def query_deposits_withdrawals(
...
-                ranges_to_query = ranges.get_location_query_ranges(
-                    cursor=cursor,
-                    location_string=location_string,
-                    start_ts=start_ts,
-                    end_ts=end_ts,
-                )
+        ranges = DBQueryRanges(self.db)
+        location_string = f'{self.location!s}_history_events_{self.name}'
+        with self.db.conn.read_ctx() as cursor:
+            ranges_to_query = ranges.get_location_query_ranges(
```
LefterisJP, line 413:
```suggestion
        location_string = f'{self.location!s}_history_events_{self.name}'
        with self.db.conn.read_ctx() as cursor:
            ranges_to_query = (ranges := DBQueryRanges(self.db)).get_location_query_ranges(
```

rotkehlchen/exchanges/manager.py
```diff
@@ -306,11 +306,29 @@ def get_user_binance_pairs(self, name: str, location: Location) -> list[str]:
             return self.database.get_binance_pairs(name, location)
         return []
 
-    def query_history_events(self) -> None:
-        """Queries all history events for exchanges that need it
+    def query_history_events(self, location: Location, name: str | None) -> None:
```
LefterisJP, line 309: note that this changes the basic naming rules of this class. As every type of event was separate to query_online_XXX and query_XXX. This new thing you are making here breaks this. How would you document this for the interim? Moving everything to history events will take time.

LefterisJP (approved):
LGTM for now. We will need adjustments of the leftovers and probably a lot of touchup and renames in other PR.

PR #9045, opened 2024-12-07
typing:api/tests_t-u-y

azerv1:
Related #842

### DIFF
--- a/rotkehlchen/tests/api/test_user_assets.py
+++ b/rotkehlchen/tests/api/test_user_assets.py
@@ -43,7 +43,11 @@ from rotkehlchen.types import EvmTokenKind, Location
     'internet-computer': {'symbol': 'ICP', 'name': 'Internet computer'},
 }])
 @pytest.mark.parametrize('cryptocompare_cache_coinlist', [{'ICP': {}}])
-def test_add_user_assets(rotkehlchen_api_server, globaldb, cache_coinlist):  # pylint: disable=unused-argument
+def test_add_user_assets(
+        rotkehlchen_api_server: APIServer,
+        globaldb: GlobalDBHandler,
+        cache_coinlist: list[dict[str, dict]],
+) -> None:  # pylint: disable=unused-argument
     """Test that the endpoint for adding a user asset works"""
 
     user_asset1 = {
@@ -63,6 +67,7 @@ def test_add_user_assets(rotkehlchen_api_server, globaldb, cache_coinlist):  # p
     user_asset1_id = result['identifier']
 
     data = globaldb.get_asset_data(identifier=user_asset1_id, form_with_incomplete_data=False)
+    assert data is not None
     assert data.identifier == user_asset1_id
     assert data.asset_type == AssetType.OWN_CHAIN
     assert data.name == user_asset1['name']
@@ -90,6 +95,7 @@ def test_add_user_assets(rotkehlchen_api_server, globaldb, cache_coinlist):  # p
     user_asset2_id = result['identifier']
 
     data = globaldb.get_asset_data(identifier=user_asset2_id, form_with_incomplete_data=False)
+    assert data is not None
     assert data.identifier == user_asset2_id
     assert data.asset_type == AssetType.STELLAR_TOKEN
     assert data.name == user_asset2['name']
@@ -235,7 +241,11 @@ def test_add_user_assets(rotkehlchen_api_server, globaldb, cache_coinlist):  # p
     'internet-computer': {'symbol': 'ICP', 'name': 'Internet computer'},
 }])
 @pytest.mark.parametrize('cryptocompare_cache_coinlist', [{'ICP': {}}])
-def test_editing_user_assets(rotkehlchen_api_server, globaldb, cache_coinlist):  # pylint: disable=unused-argument
+def test_editing_user_assets(
+        rotkehlchen_api_server: APIServer,
+        globaldb: GlobalDBHandler,
+        cache_coinlist: list[dict[str, dict]],
+) -> None:  # pylint: disable=unused-argument
     """Test that the endpoint for editing user assets works"""
 
     user_asset1 = {
@@ -255,6 +265,7 @@ def test_editing_user_assets(rotkehlchen_api_server, globaldb, cache_coinlist): 
     user_asset1_id = result['identifier']
 
     data = globaldb.get_asset_data(identifier=user_asset1_id, form_with_incomplete_data=False)
+    assert data is not None
     assert data.identifier == user_asset1_id
     assert data.asset_type == AssetType.OWN_CHAIN
     assert data.name == user_asset1['name']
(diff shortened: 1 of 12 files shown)


rotkehlchen/tests/api/test_user_assets.py
```diff
@@ -504,7 +520,9 @@ def test_deleting_user_assets(rotkehlchen_api_server, globaldb, cache_coinlist):
     assert_proper_response(response)
     assert globaldb.get_asset_data(identifier=user_asset1_id, form_with_incomplete_data=False) is None  # noqa: E501
     # Also check that `forked` was updated
-    assert globaldb.get_asset_data(identifier=user_asset2_id, form_with_incomplete_data=False).forked is None  # noqa: E501
+
+    if globaldb.get_asset_data(identifier=user_asset2_id, form_with_incomplete_data=False) is not None:  # noqa: E501
+        assert globaldb.get_asset_data(identifier=user_asset2_id, form_with_incomplete_data=False).forked is None  # type: ignore # mypy complains about None of Assetdata | None even after checking in the previous line. # noqa: E501
```
LefterisJP, line 525: It can't see it since you don't save it in a variable and there is no concept of a pure function. So there is no guarantee the function will always return the same value for the same input.

So this should probably work.

```suggestion
    assert (asset_data := globaldb.get_asset_data(identifier=user_asset2_id, form_with_incomplete_data=False)) is not None
    assert asset_data.forked is None
```

rotkehlchen/tests/api/test_user_evm_tokens.py
```diff
@@ -216,8 +225,8 @@ def test_adding_user_tokens(rotkehlchen_api_server, cache_coinlist):  # pylint:
         ],
     )
     serialized_token = bad_token.to_dict()
-    del serialized_token['identifier']
-    del serialized_token['forked']
+    serialized_token.pop('identifier', None)
```
LefterisJP, line 228: why this change here and below?

rotkehlchen/tests/api/test_users.py
```diff
@@ -179,9 +194,12 @@ def test_user_creation_with_no_analytics(rotkehlchen_api_server, data_dir):
...
 def test_user_creation_permission_error(
-        mock_path_mkdir, rotkehlchen_api_server,  # pylint:disable=unused-argument
-):
-    """Test that creating a user when data directory permissions are wrong is handled"""
+        mock_path_mkdir: mock.MagicMock,
+        rotkehlchen_api_server: 'APIServer',
+        use_clean_caching_directory: bool,
+        start_with_logged_in_user: bool,
+) -> None:
+    """Test that creati: Path,ng a user when data directory permissions are wrong is handled"""
```
LefterisJP, line 202: Why did you do here?
```suggestion
    """Test that creating a user when data directory permissions are wrong is handled"""
```

rotkehlchen/tests/api/test_users.py
```diff
@@ -215,8 +237,8 @@ def test_user_creation_with_premium_credentials(rotkehlchen_api_server, data_dir
         PremiumCredentials(VALID_PREMIUM_KEY, VALID_PREMIUM_SECRET),
         username=username,
         patch_get=True,
-        metadata_last_modify_ts=0,
-        metadata_data_hash=b'',
+        metadata_last_modify_ts=Timestamp(0),
+        metadata_data_hash=str(b''),
```
LefterisJP, line 241: If this should be a string then this is what you need to do

```suggestion
        metadata_data_hash='',
```

rotkehlchen/tests/api/test_users.py
```diff
@@ -549,16 +582,16 @@ def test_user_password_change(rotkehlchen_api_server, username, db_password):
         check_proper_unlock_result(result)
 
     assert rotki.user_is_logged_in is True
-    users_data = check_user_status(rotkehlchen_api_server)
-    assert len(users_data) == 1
-    assert users_data[username] == 'loggedin'
+    users_data_2 = check_user_status(rotkehlchen_api_server)  # type: ignore
```
LefterisJP, line 585: why rename it here? I don't see this used elsewhere. Also what is the type ignore for? You need to be stating these things when adding a type ignore.

rotkehlchen/tests/api/test_users.py
```diff
@@ -669,7 +707,8 @@ def test_user_login(rotkehlchen_api_server, username, db_password, data_dir):
         check_proper_unlock_result(result)
 
     assert rotki.user_is_logged_in is True
-    users_data = check_user_status(rotkehlchen_api_server)
+    users_data = check_user_status(rotkehlchen_api_server)  # type: ignore  # TODO: mypy for some reason says that this statement is unreachable. Not sure why.
```
LefterisJP, line 710: you would have to find out. Probably cause it thinks `assert rotki.user_is_logged_in is True` is always False for some reason?

rotkehlchen/tests/api/test_users.py
```diff
@@ -887,22 +932,26 @@ def test_user_del_premium_credentials(rotkehlchen_api_server, username):
...
         assert rotki.premium.is_active()
 
-    # Delete premium credentials for current user
     response = requests.delete(api_url_for(rotkehlchen_api_server, 'userpremiumkeyresource',
                                            name=username))
     assert_simple_ok_response(response)
+    # Delete premium credentials for current user
     assert rotki.premium is None
-    assert rotki.premium_sync_manager.premium is None
+    assert rotki.premium_sync_manager.premium is None  # type: ignore  # TODO: mypy for some reason says that this statement is unreachable. Not sure why.
```
LefterisJP, line 943: again here, this is not an answer you would have to figure out why.

rotkehlchen/tests/api/test_users.py
```diff
@@ -887,22 +932,26 @@ def test_user_del_premium_credentials(rotkehlchen_api_server, username):
             json=data,
         )
     with patched_get:
+        assert rotki.premium is not None
         assert rotki.premium.is_active()
 
-    # Delete premium credentials for current user
```
LefterisJP, line 892: why did you remove this and move it down? This should stay here, above the deletion.

rotkehlchen/tests/api/test_users.py
```diff
@@ -1022,7 +1081,8 @@ def test_users_by_name_endpoint_errors(rotkehlchen_api_server, username, db_pass
     assert rotki.user_is_logged_in is False
 
     # Now let's try to login with an invalid password
-    data = {'password': 'wrong-password', 'sync_approval': 'unknown', 'async_query': True}
+    data = {'password': 'wrong-password', 'sync_approval': 'unknown', 'async_query': True}  # type: ignore  # TODO: mypy for some reason says that this statement is unreachable. Not sure why.
```
LefterisJP, line 1084: same here. And I suspect the reason is always the same. Something to do with `assert rotki.user_is_logged_in is False` right above

LefterisJP (approved):
LGTM but justt needs a rebase and conflict resolution. I can probably do that now.

PR #9046, opened 2024-12-08
Bump production, develop and lint

LefterisJP:


LefterisJP:
Removed these bumps since something in them was failing `test_backend` only in the CI.

```
pytest==8.3.4
coverage==7.6.9
psutil==6.1.0
```

PR #9047, opened 2024-12-09
Only take monerium mint/burn events that belong to the user

LefterisJP:


rotkehlchen/chain/evm/decoding/monerium/decoder.py
```diff
@@ -68,7 +68,7 @@ def _decode_mint_and_burn(self, context: DecoderContext) -> DecodingOutput:
         # Not iterating over the events because the logic for decoding erc_20 transfers
         # has not been executed before calling this decoding logic
         event = None
-        if from_address == ZERO_ADDRESS:  # Create a mint event
+        if from_address == ZERO_ADDRESS and self.base.is_tracked(to_address):  # Create mint event
```
prettyirrelevant, line 71: isn't it better to have the `is_tracked()` check at the beginning of this method since it's repeated both here and below.

LefterisJP: We track different addresses depending on which is from and which is to and other conditions

prettyirrelevant (approved):
lgtm

PR #9048, opened 2024-12-09
refactor: some icons on the app

lukicenturi:
Needs update from @rotki/ui-library


frontend/app/src/plugins/rui/index.ts
```diff
@@ -443,6 +264,165 @@ export function createRuiPlugin(defaults: Partial<RuiOptions['defaults']>): RuiP
...
+        RiRestartLine,
+        RiRefreshLine,
+        RiAddLine,
+        RiSubtractLine,
+        RiSendPlane2Line,
+        RiPaletteLine,
+        RiExternalLinkLine,
+        RiCornerLeftUpLine,
+        RiArrowUpCircleLine,
+        RiInformationLine,
```
kelsos, line 425: Ok so for this we need a breaking alpha release before we remove otherwise components will be broken in the meanwhile.

kelsos (approved):
lgtm, initial check looks ok, let's check the console for any unregistered icons and in the meanwhile I will try to setup the eslint rule for checking for that (as much as possible)

kelsos (approved):

yabirgb (approved):
LGTM

PR #9054, opened 2024-12-10
fix: refactor conditional check for improved readability rules.py

mdqst:
# Description

In the following line:

```python
if isinstance(event, EvmEvent) is False or rule is not None:
```

The conditional check can be simplified to:

```python
if not isinstance(event, EvmEvent) or rule is not None:
```

This change improves readability and aligns with Python’s preferred idiomatic style. By removing the unnecessary `is False` comparison, the code becomes more concise and easier to understand. It also reduces cognitive load for other developers reading or maintaining the code, making it a small but valuable improvement.

## Checklist

### DIFF
--- a/rotkehlchen/accounting/rules.py
+++ b/rotkehlchen/accounting/rules.py
@@ -57,7 +57,7 @@ class AccountingRulesManager:
         else:
             callback = None
 
-        if isinstance(event, EvmEvent) is False or rule is not None:
+        if not isinstance(event, EvmEvent) or rule is not None:
             return rule, callback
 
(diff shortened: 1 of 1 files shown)


LefterisJP:
> This change improves readability and aligns with Python’s preferred idiomatic style. By removing the unnecessary `is False` comparison, the code becomes more concise and easier to understand. It also reduces cognitive load for other developers reading or maintaining the code, making it a small but valuable improvement.

Hello @mdqst thanks a lot for the PR.

Unfortunately I do not agree with that style and we use `is False` quite a lot in our codebase. We also use `not` in some cases such as clearcut boolean checks. But anyway won't accept a PR trying to change one from the other.

PR #9056, opened 2024-12-10
feat: show double confirmation when ignoring asset

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/asset-manager/managed/ManagedAssetTable.vue
+++ b/frontend/app/src/components/asset-manager/managed/ManagedAssetTable.vue
@@ -19,6 +19,7 @@ import CollectionHandler from '@/components/helper/CollectionHandler.vue';
 import TableFilter from '@/components/table-filter/TableFilter.vue';
 import AssetStatusFilter from '@/components/asset-manager/AssetStatusFilter.vue';
 import IgnoreButtons from '@/components/history/IgnoreButtons.vue';
+import { useConfirmStore } from '@/store/confirm';
 import type { ActionStatus } from '@/types/action';
 import type { SupportedAsset } from '@rotki/common';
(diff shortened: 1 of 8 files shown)


frontend/app/src/composables/blockchain/use-account-addition-notifications.ts
```diff
@@ -82,7 +82,7 @@ export function useAccountAdditionNotifications(): UseAccountAdditionNotificatio
         address: account.address,
         list: listOfFailureText.join('\n'),
       }),
-      severity: Severity.INFO,
+      severity: Severity.WARNING,
```
kelsos, line 85: I am thinking that the severity in this actually depends on the data.

We probably need to identify the different cases and change the severity based on the status.

frontend/app/src/components/asset-manager/managed/ManagedAssetTable.vue
```diff
@@ -136,15 +137,28 @@ function getAsset(item: SupportedAsset) {
...
+    show({
+      message: t('ignore.confirm.message', {
+        asset: symbol || name,
+      }),
+      title: t('ignore.confirm.title'),
+      type: 'warning',
+    }, async () => {
+      await ignoreAsset(identifier);
+    });
+  }
```
kelsos, line 161: Would it make sense to move this to the `useIgnoreAssetStore` and reuse it to the other component too?

frontend/app/src/composables/blockchain/use-account-addition-notifications.ts
lukicenturi: okay gonna revert this and do this in different PR then

cleanerzkp (approved):
LGTM

PR #9060, opened 2024-12-11
New AssetMovement in binance bitfinex and bitpanda

nicholasyoder:
Related #7007

### DIFF
--- a/rotkehlchen/exchanges/binance.py
+++ b/rotkehlchen/exchanges/binance.py
@@ -4,6 +4,7 @@ import json
 import logging
 import operator
 from collections import defaultdict
+from collections.abc import Sequence
 from contextlib import suppress
 from json.decoder import JSONDecodeError
 from sqlite3 import IntegrityError
@@ -18,6 +19,7 @@ from rotkehlchen.assets.asset import AssetWithOracles
 from rotkehlchen.assets.converters import asset_from_binance
 from rotkehlchen.constants import ZERO
 from rotkehlchen.constants.assets import A_USD
+from rotkehlchen.data_import.utils import maybe_set_transaction_extra_data
 from rotkehlchen.db.constants import BINANCE_MARKETS_KEY
 from rotkehlchen.db.history_events import DBHistoryEvents
 from rotkehlchen.db.ranges import DBQueryRanges
@@ -27,7 +29,6 @@ from rotkehlchen.errors.misc import InputError, RemoteError
 from rotkehlchen.errors.price import NoPriceForGivenTimestamp
 from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.exchanges.data_structures import (
-    AssetMovement,
     BinancePair,
     MarginPosition,
     Trade,
@@ -45,7 +46,11 @@ from rotkehlchen.exchanges.utils import (
 )
 from rotkehlchen.fval import FVal
 from rotkehlchen.history.deserialization import deserialize_price
-from rotkehlchen.history.events.structures.base import HistoryEvent
+from rotkehlchen.history.events.structures.asset_movement import (
+    AssetMovement,
+    create_asset_movement_with_fee,
+)
+from rotkehlchen.history.events.structures.base import HistoryBaseEntry, HistoryEvent
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.history.price import PriceHistorian
 from rotkehlchen.inquirer import Inquirer
@@ -60,7 +65,6 @@ from rotkehlchen.serialization.deserialize import (
 from rotkehlchen.types import (
     ApiKey,
     ApiSecret,
-    AssetMovementCategory,
     ExchangeAuthCredentials,
     Fee,
     Location,
@@ -68,7 +72,7 @@ from rotkehlchen.types import (
     TimestampMS,
 )
 from rotkehlchen.user_messages import MessagesAggregator
-from rotkehlchen.utils.misc import ts_ms_to_sec, ts_now_in_ms
+from rotkehlchen.utils.misc import ts_ms_to_sec, ts_now_in_ms, ts_sec_to_ms
 from rotkehlchen.utils.mixins.cacheable import cache_response_timewise
 from rotkehlchen.utils.mixins.lockable import protect_with_lock
 
@@ -1356,8 +1360,8 @@ class Binance(ExchangeInterface, ExchangeWithExtras):
     def _deserialize_fiat_movement(
             self,
             raw_data: dict[str, Any],
-            category: AssetMovementCategory,
-    ) -> AssetMovement | None:
+            event_type: Literal[HistoryEventType.DEPOSIT, HistoryEventType.WITHDRAWAL],
+    ) -> list[AssetMovement] | None:
         """Processes a single deposit/withdrawal from binance and deserializes it
 
         Can log error/warning and return None if something went wrong at deserialization
@@ -1371,9 +1375,8 @@ class Binance(ExchangeInterface, ExchangeWithExtras):
 
             asset = asset_from_binance(raw_data['fiatCurrency'])
             tx_id = get_key_if_has_val(raw_data, 'orderNo')
-            timestamp = deserialize_timestamp_from_intms(raw_data['createTime'])
+            timestamp = ts_sec_to_ms(deserialize_timestamp_from_intms(raw_data['createTime']))
             fee = Fee(deserialize_asset_amount(raw_data['totalFee']))
-            link_str = str(tx_id) if tx_id else ''
             amount = deserialize_asset_amount_force_positive(raw_data['amount'])
             address = deserialize_asset_movement_address(raw_data, 'address', asset)
         except UnknownAsset as e:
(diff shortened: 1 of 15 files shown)


rotkehlchen/exchanges/binance.py
```diff
@@ -1400,33 +1400,46 @@ def _deserialize_fiat_movement(
...
             if 'insertTime' in raw_data:
-                category = AssetMovementCategory.DEPOSIT
+                event_type = HistoryEventType.DEPOSIT
                 timestamp = deserialize_timestamp_from_intms(raw_data['insertTime'])
                 fee = Fee(ZERO)
             else:
-                category = AssetMovementCategory.WITHDRAWAL
+                event_type = HistoryEventType.WITHDRAWAL
                 timestamp = deserialize_timestamp_from_date(
                     date=raw_data['applyTime'],
```
yabirgb, line 1444: you can do `ts_sec_to_ms` here and avoid doing it both in the movement and the fee

rotkehlchen/exchanges/bitfinex.py
```diff
@@ -324,16 +328,31 @@ def _api_query_paginated(
...
             processed_result_ids = processed_result_ids.copy()
-            # type ignore is due to always having a trade link for bitfinex trades
-            processed_result_ids.update({int(result.link) for result in results_})  # type: ignore
+            newly_processed_results = set()
+            for result in results_:
+                if isinstance(result, AssetMovement):
+                    if result.unique_id is None:
+                        continue  # fees don't have a unique id
+
+                    newly_processed_results.add(int(result.unique_id))
```
yabirgb, line 337: are those guaranteed to be integers? I'm not sure what casting to int does

rotkehlchen/exchanges/bitfinex.py
```diff
@@ -324,16 +328,31 @@ def _api_query_paginated(
...
+                else:
+                    # type ignore is due to always having a trade link for bitfinex trades
+                    newly_processed_results.add(int(result.link))  # type: ignore
```
yabirgb, line 340: same here. What does casting to int mean?

rotkehlchen/exchanges/bitfinex.py
```diff
@@ -324,16 +328,31 @@ def _api_query_paginated(
...
             # Update pagination params per endpoint
             # NB: Copying the dict before updating it prevents losing the call args values
             call_options = call_options.copy()
+            last_item = results[-1]
             call_options.update({
-                'start': results[-1].timestamp * 1000,
+                'start': (
+                    last_item.timestamp
+                    if isinstance(last_item, AssetMovement)
+                    else last_item.timestamp * 1000
```
yabirgb, line 354: use ts_to_ms here instead of the *1000

rotkehlchen/exchanges/bitfinex.py
```diff
@@ -482,31 +501,43 @@ def _deserialize_asset_movement(self, raw_result: list[Any]) -> AssetMovement:
...
-        return AssetMovement(
-            timestamp=Timestamp(int(raw_result[5] / 1000)),
-            location=Location.BITFINEX,
-            category=category,
-            address=address,
-            transaction_id=transaction_id,
+        events = [AssetMovement(
+            location=self.location,
+            event_type=event_type,
+            timestamp=(timestamp := TimestampMS(int(raw_result[5]))),
```
yabirgb, line 518:
```suggestion
            timestamp=(timestamp := TimestampMS(deserialize_timestamp(raw_result[5]))),
```

rotkehlchen/serialization/deserialize.py
```diff
@@ -325,6 +327,27 @@ def deserialize_asset_movement_category(
...
+    Can throw DeserializationError if value is not as expected
+    """
+    if isinstance(value, str):
+        if value.lower() == 'deposit':
+            return HistoryEventType.DEPOSIT
+        if value.lower() in {'withdraw', 'withdrawal'}:
+            return HistoryEventType.WITHDRAWAL
+        raise DeserializationError(
+            f'Failed to deserialize asset movement category symbol. Unknown {value}',
+        )
```
yabirgb, line 344:
```suggestion
        if (lowered_value := value.lower()) == 'deposit':
            return HistoryEventType.DEPOSIT
        if lowered_value in {'withdraw', 'withdrawal'}:
            return HistoryEventType.WITHDRAWAL
        raise DeserializationError(
            f'Failed to deserialize asset movement category symbol. Unknown {value}',
        )
```

yabirgb: We already have `deserialize_asset_movement_category`. I believe you took it from there right?

rotkehlchen/exchanges/bitfinex.py
yabirgb: Do you need unique_id here? Isn't it enought with the event_identifier?

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -92,6 +92,10 @@ def __init__(
             extra_data=extra_data,
         )
 
+        # Note that unique_id is not stored in the database, so this is useful only when working
+        # with newly created movements. This will be None in movements loaded from the database.
+        self.unique_id = unique_id
```
yabirgb, line 97: Do you only need this for bitfinex? It looks like a hack since entries for db don't have it and will cause trouble in the future

rotkehlchen/exchanges/bitfinex.py
nicholasyoder: > I'm not sure what casting to int does

These were being cast to int so that they would match here: https://github.com/rotki/rotki/blob/a282502d7ce3b365f4df8d3e1004cd38b190ff58/rotkehlchen/exchanges/bitfinex.py#L429 since the raw data is an int (at least in our test data).

> are those guaranteed to be integers?

In the test data they are integers but actually it looks like in the docs it says this (index 0 of the movement array) is a string: https://docs.bitfinex.com/reference/rest-auth-movements

I've changed this to handle these as strings, and instead of using a new unique_id property on the AssetMovement, I'm just putting it in extra_data. Not sure why I didn't think of doing it that way to begin with.

> Isn't it enought with the event_identifier?

Well, no that wouldn't really work here, since the `processed_result_ids` are being checked against the id in the raw data from the api to prevent processing the same event twice. The event_identifier is a hash of the location and the unique_id, so it can't be matched against anything in the raw data.

rotkehlchen/exchanges/bitfinex.py
nicholasyoder: As mentioned in the other comment, this was cast to int so it would match the raw data. Handling this as a string now.

rotkehlchen/serialization/deserialize.py
nicholasyoder: Yes. `deserialize_asset_movement_category` will be removed after all the exchange interfaces are converted. I added a line to its doc string mentioning that it is deprecated.

rotkehlchen/history/events/structures/asset_movement.py
nicholasyoder: Yes, this was only for bitfinex, although I thought it might be needed elsewhere as well.

But I've removed it again, since it should simply be put in extra_data if it needs be available. That way it will get saved to the db as well.

rotkehlchen/exchanges/bitpanda.py
```diff
@@ -208,18 +210,30 @@ def _deserialize_wallettx(
...
-            link=tx_id,
-        )
+            balance=Balance(abs(amount)),
+            unique_id=f'{tx_id}{transaction_id}',  # Use both here as tx_id is not always unique (at least in the test data)  # noqa: E501
+            extra_data=maybe_set_transaction_extra_data(
+                address=address,
+                transaction_id=transaction_id,
+            ),
+        )]
+        if fee != ZERO:
```
LefterisJP, line 225: same as in bitfinex and here this would be simplified and reduce code with the abstracted function

rotkehlchen/tests/exchanges/test_bitfinex.py
```diff
@@ -867,18 +869,27 @@ def test_deserialize_asset_movement_deposit(mock_bitfinex):
...
+            'address': 'DESTINATION_ADDRESS',
+            'transaction_id': 'TRANSACTION_ID',
+        },
+    ), AssetMovement(
+        timestamp=TimestampMS(1569348774000),
+        location=Location.BITFINEX,
+        event_type=HistoryEventType.DEPOSIT,
+        asset=fee_asset,
+        balance=Balance(FVal('0.00135')),
+        unique_id=str(13105603),
```
LefterisJP, line 890: here and in other places please fix. These are just string literals

```suggestion
        unique_id='13105603',
```

rotkehlchen/exchanges/bitfinex.py
```diff
@@ -324,16 +332,30 @@ def _api_query_paginated(
             results.extend(cast('Iterable', results_))
             # NB: Copying the set before updating it prevents losing the call args values
             processed_result_ids = processed_result_ids.copy()
-            # type ignore is due to always having a trade link for bitfinex trades
-            processed_result_ids.update({int(result.link) for result in results_})  # type: ignore
+            newly_processed_results = set()
+            for result in results_:
```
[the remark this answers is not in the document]
nicholasyoder: Yes, once trades are also based on history events this should be able to be simplified. Added a comment mentioning that some of this is temporary until trades are converted.

rotkehlchen/exchanges/bitfinex.py
```diff
@@ -324,16 +332,30 @@ def _api_query_paginated(
...
                 break
             # Update pagination params per endpoint
             # NB: Copying the dict before updating it prevents losing the call args values
             call_options = call_options.copy()
+            last_item = results[-1]
             call_options.update({
-                'start': results[-1].timestamp * 1000,
+                'start': (
+                    last_item.timestamp
+                    if isinstance(last_item, AssetMovement)
```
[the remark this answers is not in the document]
nicholasyoder: also added a comment here.

LefterisJP (approved):
LGTM

PR #9061, opened 2024-12-11
update assets before schema-breaking DB upgrades

prettyirrelevant:
currently, when the global db schema is upgraded, we may miss asset updates that were compatible with the old schema but incompatible with the new one. this pr ensures we apply pending asset updates before upgrading the db schema.

key changes:
- pass globaldbhandler to db upgrade functions for asset updates.
- add `assets_updater.apply_pending_compatible_updates()`
- refactor parsing logic into dedicated parser classes.

### DIFF
--- a/rotkehlchen/globaldb/asset_updates/manager.py
+++ b/rotkehlchen/globaldb/asset_updates/manager.py
@@ -1,35 +1,34 @@
-import json
 import logging
-import re
 import sqlite3
+from collections import defaultdict
 from contextlib import suppress
-from enum import Enum, auto
 from http import HTTPStatus
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from typing import TYPE_CHECKING, Any, Literal, NamedTuple
+from typing import TYPE_CHECKING, Any, Literal
 
 import requests
 
 from rotkehlchen.assets.asset import Asset
 from rotkehlchen.assets.resolver import AssetResolver
-from rotkehlchen.assets.types import AssetData, AssetType
+from rotkehlchen.assets.types import AssetData
 from rotkehlchen.constants.misc import GLOBALDB_NAME, GLOBALDIR_NAME
 from rotkehlchen.db.drivers.gevent import DBCursor
 from rotkehlchen.db.settings import CachedSettings
 from rotkehlchen.errors.asset import UnknownAsset
 from rotkehlchen.errors.misc import RemoteError
 from rotkehlchen.errors.serialization import DeserializationError
+from rotkehlchen.globaldb.utils import initialize_globaldb
 from rotkehlchen.logging import RotkehlchenLogsAdapter
-from rotkehlchen.serialization.deserialize import deserialize_evm_address
-from rotkehlchen.types import ChainID, ChecksumEvmAddress, EvmTokenKind, Timestamp
 from rotkehlchen.utils.misc import is_production
 from rotkehlchen.utils.network import query_file
 
-from .handler import GlobalDBHandler, initialize_globaldb
+from .parsers import AssetCollectionParser, AssetParser, MultiAssetMappingsParser
+from .types import UpdateFileType
 
 if TYPE_CHECKING:
     from rotkehlchen.db.drivers.gevent import DBConnection
+    from rotkehlchen.globaldb.handler import GlobalDBHandler
     from rotkehlchen.user_messages import MessagesAggregator
 
 logger = logging.getLogger(__name__)
@@ -40,12 +39,7 @@ ASSETS_UPDATES_URL = 'https://raw.githubusercontent.com/rotki/assets/{branch}/up
 ASSET_COLLECTIONS_UPDATES_URL = 'https://raw.githubusercontent.com/rotki/assets/{branch}/updates/{version}/asset_collections_updates.sql'
 ASSET_COLLECTIONS_MAPPINGS_UPDATES_URL = 'https://raw.githubusercontent.com/rotki/assets/{branch}/updates/{version}/asset_collections_mappings_updates.sql'
 FIRST_VERSION_WITH_COLLECTIONS = 16
-
-
-class UpdateFileType(Enum):
-    ASSETS = auto()
-    ASSET_COLLECTIONS = auto()
-    ASSET_COLLECTIONS_MAPPINGS = auto()
+FIRST_GLOBAL_DB_VERSION_WITH_COLLECTIONS = 4
 
 
 def executeall(cursor: DBCursor, statements: str) -> None:
@@ -64,28 +58,60 @@ def _replace_assets_from_db(
         connection: 'DBConnection',
         sourcedb_path: Path,
 ) -> None:
+    """Replace asset-related tables with data from source database.
+
+    Handles: token_kinds, asset_types, assets, evm_tokens, underlying_tokens_list,
+    common_asset_details, asset_collections, multiasset_mappings.
+    """
+    # First handle token_kinds & asset_types since other tables reference it
+    required_tables = [
+        'token_kinds',
+        'asset_types',
+        'assets',
+        'evm_tokens',
+        'underlying_tokens_list',
+        'common_asset_details',
+        'asset_collections',
+        'multiasset_mappings',
+        'settings',
+    ]
+
+    script_parts = ['PRAGMA foreign_keys = OFF;']
     with connection.write_ctx() as cursor:
-        cursor.executescript(f"""
-        ATTACH DATABASE '{sourcedb_path}' AS other_db;
-        PRAGMA foreign_keys = OFF;
-        DELETE FROM assets;
-        DELETE FROM evm_tokens;
-        DELETE FROM underlying_tokens_list;
-        DELETE FROM common_asset_details;
-        DELETE FROM asset_collections;
-        DELETE FROM multiasset_mappings;
-        INSERT INTO assets SELECT * FROM other_db.assets;
-        INSERT INTO evm_tokens SELECT * FROM other_db.evm_tokens;
-        INSERT INTO underlying_tokens_list SELECT * FROM other_db.underlying_tokens_list;
-        INSERT INTO common_asset_details SELECT * FROM other_db.common_asset_details;
-        INSERT INTO asset_collections SELECT * FROM other_db.asset_collections;
-        INSERT INTO multiasset_mappings SELECT * FROM other_db.multiasset_mappings;
-        INSERT OR REPLACE INTO settings(name, value) VALUES('{ASSETS_VERSION_KEY}',
-        (SELECT value FROM other_db.settings WHERE name='{ASSETS_VERSION_KEY}')
-        );
-        PRAGMA foreign_keys = ON;
-        DETACH DATABASE 'other_db';
-        """)
+        cursor.execute(f"ATTACH DATABASE '{sourcedb_path}' AS other_db;")
+        cursor.execute(
+            """SELECT COUNT(*) FROM sqlite_master
+            WHERE type='table' AND name IN ({})
+            """.format(','.join('?' * len(required_tables))),
+            required_tables,
+        )
+        if cursor.fetchone()[0] < len(required_tables):
+            schemas = {}
+            for table in required_tables:
+                if (schema := cursor.execute(
+                    "SELECT sql FROM other_db.sqlite_master WHERE type='table' AND name=?",
+                    (table,),
+                ).fetchone()):
+                    raw_schema = schema[0]
+                    schema_def = raw_schema.split('CREATE TABLE')[1].split('(', 1)[1]
+                    schemas[table] = schema_def
+
+            for table, schema in schemas.items():
+                script_parts.append(f'CREATE TABLE IF NOT EXISTS {table} ({schema};')
+
+        # Always do the delete and insert
+        for table in required_tables:
+            script_parts.extend([
+                f'DELETE FROM {table};',
+                f'INSERT INTO {table} SELECT * FROM other_db.{table};',
+            ])
+        script_parts.extend([
+            f"INSERT OR REPLACE INTO settings(name, value) VALUES('{ASSETS_VERSION_KEY}',"
+            f"(SELECT value FROM other_db.settings WHERE name='{ASSETS_VERSION_KEY}'));",
+            'PRAGMA foreign_keys = ON;',
+            "DETACH DATABASE 'other_db';",
+        ])
+        cursor.executescript('\n'.join(script_parts))
 
 
 def _force_remote_asset(cursor: DBCursor, local_asset: Asset, full_insert: str) -> None:
@@ -104,20 +130,27 @@ def _force_remote_asset(cursor: DBCursor, local_asset: Asset, full_insert: str) 
         'WHERE identifier=? OR parent_token_entry=?;',
         (local_asset.identifier, local_asset.identifier),
     ).fetchall()
-    collection = cursor.execute(
-        'SELECT id, name, symbol, main_asset FROM asset_collections WHERE main_asset=?;',
-        (local_asset.identifier,),
-    ).fetchone()
+
+    try:
+        collection = cursor.execute(
+            'SELECT * FROM asset_collections WHERE main_asset=?;',
+            (local_asset.identifier,),
+        ).fetchone()
+    except sqlite3.Error:
+        # If query fails due to missing main_asset
+        # column (pre-v10 schema), set collection to None
+        collection = None
+
     cursor.execute(
         'DELETE FROM assets WHERE identifier=?;',
         (local_asset.identifier,),
     )
 
-    # Insert new entry. Since identifiers are the same, no foreign key constrains should break
+    # Insert new entry. Since identifiers are the same, no foreign key constraints should break
     executeall(cursor, full_insert)
     if collection:
         cursor.execute(  # reinsert the collection since it was cascade-deleted
-            'INSERT OR IGNORE INTO asset_collections (id, name, symbol, main_asset) VALUES (?, ?, ?, ?);',  # noqa: E501
+            'INSERT OR IGNORE INTO asset_collections VALUES (?, ?, ?, ?)',
             collection,
         )
     # now add the old mappings back into the db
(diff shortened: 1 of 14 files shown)


rotkehlchen/rotkehlchen.py
```diff
@@ -510,7 +511,10 @@ def unlock_user(
         )
 
         self.migration_manager.maybe_migrate_data()
-        self.assets_updater = AssetsUpdater(self.msg_aggregator)
+        self.assets_updater = AssetsUpdater(
+            msg_aggregator=self.msg_aggregator,
+            globaldb=GlobalDBHandler(),
+        )
```
yabirgb, line 517:
```suggestion
        self.assets_updater = AssetsUpdater(
            msg_aggregator=self.msg_aggregator,
            globaldb=globaldb,
        )
```

yabirgb (commented):
Do we have a simple test for the logic? Is the one for the integrity executing this logic? Having a test that easily test this logic would be great since it will help us to debug future possible issues

rotkehlchen/globaldb/updates/manager.py
```diff
@@ -706,7 +544,13 @@ def perform_update(
...
                 # now move the data to the actual global DB
                 log.info('Finishing assets update. Replacing users globaldb with the updated information')  # noqa: E501
-                _replace_assets_from_db(GlobalDBHandler().conn, tmpdir / temp_db_name)
+                _replace_assets_from_db(self.globaldb.conn, tmpdir / temp_db_name)
+
+                # foreign keys are turned OFF since DB upgrades may need to modify tables
+                # in an order that would violate foreign key constraints.
+                # this only happens when assets are updated just before a global db upgrade happens.  # noqa: E501
+                with self.globaldb.conn.read_ctx() as cursor:
+                    cursor.executescript('PRAGMA foreign_keys = OFF;')
```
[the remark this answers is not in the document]
prettyirrelevant: > Were we doing it before? 

no but previously we didn't run asset updates before a global db upgrade happens

>  Also is not read_ctx but write_ctx what you would use

thanks

rotkehlchen/globaldb/upgrades/manager.py
```diff
@@ -171,3 +192,42 @@ def _perform_single_upgrade(
...
+        - DBSchemaError if the database schema is invalid.
+    """
+    is_fresh_db = maybe_upgrade_globaldb(
+        globaldb=globaldb,
+        connection=connection,
+        global_dir=global_dir,
+        db_filename=db_filename,
+        msg_aggregator=msg_aggregator,
+    )
+    connection.executescript('PRAGMA foreign_keys=on;')
```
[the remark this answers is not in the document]
prettyirrelevant: nah nah. i didn't set it now -- i was just splitting the implementation of the funtion below into `initialize_globaldb()` and `configure_globaldb()` https://github.com/rotki/rotki/blob/67b1dbfc379ecd0bd9a7d348ea3dedbd9c6528d6/rotkehlchen/globaldb/handler.py#L145-L183

rotkehlchen/globaldb/utils.py
```diff
@@ -51,3 +59,64 @@ def set_token_spam_protocol(
...
+        sql_vm_instructions_cb=sql_vm_instructions_cb,
+    )
+    try:
+        with connection.read_ctx() as cursor:
+            ongoing_upgrade_from_version = globaldb_get_setting_value(
+                cursor=cursor,
+                name='ongoing_upgrade_from_version',
+                default_value=-1,
+            )
+    except sqlite3.OperationalError:  # pylint: disable=no-member
```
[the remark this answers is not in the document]
prettyirrelevant: yeah, it does. a quick search and i found multiple occurrences.

rotkehlchen/rotkehlchen.py
prettyirrelevant: `globaldb` as a variable is not available here since this is called at `unlock()` and not at `__init__()`

prettyirrelevant:
> Is the one for the integrity executing this logic? 

yeah, it goes from global db v3 -> v10.

rotkehlchen/globaldb/updates/parsers.py
```diff
@@ -0,0 +1,325 @@
...
+            return f"'{content}'"
+
+        return self._double_quotes_re.sub(replace_double_quotes, text)
+
+
+class AssetParser(BaseParser):
+    def __init__(self) -> None:
+        super().__init__()
+        self._version_parsers = [
+            (VersionRange(1, None), self._parse),
```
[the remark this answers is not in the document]
prettyirrelevant: `evm_tokens` was introduced v15 iirc.

rotkehlchen/globaldb/updates/manager.py
```diff
@@ -64,28 +58,62 @@ def _replace_assets_from_db(
...
-        """)
+        cursor.execute(f"ATTACH DATABASE '{sourcedb_path}' AS other_db;")
+        cursor.execute(
+            """SELECT COUNT(*) FROM sqlite_master
+            WHERE type='table' AND name IN ({})
+            """.format(','.join('?' * len(required_tables))),
+            required_tables,
+        )
+        existing_tables = cursor.fetchone()[0]
+        if existing_tables < len(required_tables):
```
LefterisJP, line 89:
```suggestion 
        if (existing_tables := cursor.fetchone()[0]) < len(required_tables):
```

rotkehlchen/globaldb/updates/manager.py
```diff
@@ -64,28 +58,62 @@ def _replace_assets_from_db(
...
+            schemas = {}
+            for table in required_tables:
+                schema = cursor.execute(
+                    "SELECT sql FROM other_db.sqlite_master WHERE type='table' AND name=?",
+                    (table,),
+                ).fetchone()
+                if schema:
```
LefterisJP, line 96:
```suggestion
                if (schema := cursor.execute(
                    "SELECT sql FROM other_db.sqlite_master WHERE type='table' AND name=?",
                    (table,),
                ).fetchone():
```

rotkehlchen/globaldb/updates/manager.py
```diff
@@ -676,19 +513,26 @@ def perform_update(
...
             # use a critical section to avoid modifications in the globaldb during the update
             # process since we ignore any possible change in the user globaldb once we started
             # the update.
-            with GlobalDBHandler().conn.critical_section():
+            with self.globaldb.conn.critical_section():
                 log.info('Starting assets update. Copying content from the user globaldb')
                 _replace_assets_from_db(temp_db_connection, global_db_path)
+
+                # foreign keys are turned ON to ensure that force_remote_asset() properly cascade-deletes  # noqa: E501
+                # related entries in multiasset_mappings, underlying_tokens_list, etc. when removing assets  # noqa: E501
```
LefterisJP, line 532: Don't use noqa in multiline comments like this. Just split it. Something like this.
```suggestion
                # foreign keys are turned ON to ensure that force_remote_asset() properly
                # cascade-deletes related entries in multiasset_mappings, underlying_tokens_list
                # , etc. when removing assets 
```

rotkehlchen/globaldb/updates/manager.py
```diff
@@ -706,7 +550,13 @@ def perform_update(
                 # otherwise we are sure the DB will work without conflicts so let's
                 # now move the data to the actual global DB
                 log.info('Finishing assets update. Replacing users globaldb with the updated information')  # noqa: E501
-                _replace_assets_from_db(GlobalDBHandler().conn, tmpdir / temp_db_name)
+                _replace_assets_from_db(self.globaldb.conn, tmpdir / temp_db_name)
+
+                # foreign keys are turned OFF since DB upgrades may need to modify tables
+                # in an order that would violate foreign key constraints.
+                # this only happens when assets are updated just before a global db upgrade happens.  # noqa: E501
```
LefterisJP, line 557: same here for the noqa

rotkehlchen/globaldb/updates/manager.py
```diff
@@ -837,3 +687,52 @@ def _retrieve_update_files(
...
+    def apply_pending_compatible_updates(self) -> None:
+        """Apply any pending asset updates that are compatible with the current DB version
+        before upgrading to target_db_version.
+
+        This ensures we don't miss any asset updates that would become incompatible
+        after the DB upgrade.
+        """
+        max_compatible_version = None
+        info_json = self._get_remote_info_json()
+        latest_assets_version = info_json['latest']
```
LefterisJP, line 700: This entire function seems to not check for any key errors. How do you guarantee all keys exist? Does it not depend on a remote call?

rotkehlchen/globaldb/updates/parsers.py
```diff
@@ -0,0 +1,334 @@
...
+
+    from rotkehlchen.db.drivers.gevent import DBConnection
+
+T = TypeVar('T')
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+class BaseParser(abc.ABC, Generic[T]):
```
LefterisJP, line 27: Specify it's a parser for assets.

```suggestion
class BaseAssetParser(abc.ABC, Generic[T]):
```

rotkehlchen/globaldb/updates/parsers.py
```diff
@@ -0,0 +1,334 @@
...
+    """Base parser with common parsing functionality.
```
LefterisJP, line 28:
```suggestion
    """Base assets parser with common parsing functionality.
```

rotkehlchen/globaldb/updates/parsers.py
```diff
@@ -0,0 +1,334 @@
...
+        self._double_quotes_re = re.compile(r'\"(.*?)\"')
+        self._string_re = re.compile(r'.*([\'"])(.*?)\1.*')
+        self._version_parsers: list[tuple[VersionRange, Callable]] = []
+
+    def parse(self, insert_text: str, version: int, connection: 'DBConnection') -> T:
+        for version_range, parser in self._version_parsers:
+            if version_range.contains(version):
+                return parser(insert_text=insert_text, connection=connection)
+
+        raise ValueError(f'No parser found for version {version}')
```
LefterisJP, line 45: We never ever use ValueErrors.

Use something else. Who handles it? Who should catch it? What should happen if this happens? When can this happen?

rotkehlchen/globaldb/updates/parsers.py
```diff
@@ -0,0 +1,334 @@
...
+        come with double quotes, we convert them here.
+        """
+        def replace_double_quotes(match_result: re.Match[str]) -> str:
+            content = match_result.group(1).replace("'", "''")
+            return f"'{content}'"
+
+        return self._double_quotes_re.sub(replace_double_quotes, text)
+
+
+class AssetParser(BaseParser[AssetData]):
```
LefterisJP, line 98: small docstring here.

Also question: Shouldn't this specify which versions its' for? I guess for all known versions until now? At least in the docstring specify it.

rotkehlchen/globaldb/updates/parsers.py
```diff
@@ -0,0 +1,334 @@
...
+        return (
+            deserialize_evm_address(self._parse_str(match.group(4), 'address', insert_text)),
+            self._parse_optional_int(match.group(5), 'decimals', insert_text),
+            self._parse_optional_str(match.group(6), 'protocol', insert_text),
+            chain_id,
+            token_kind,
+        )
+
+
+class AssetCollectionParser(BaseParser[tuple[int, str, str] | tuple[int, str, str, str]]):
```
LefterisJP, line 227: Same here for docstring and parsing versions/limitations

rotkehlchen/globaldb/updates/parsers.py
```diff
@@ -0,0 +1,334 @@
...
+            raise DeserializationError(
+                f'At asset DB update invalid collection ID {collection_id} from {insert_text}',
+            )
+
+        name = self._parse_str(collection_match.group(2), 'name', insert_text)
+        symbol = self._parse_str(collection_match.group(3), 'symbol', insert_text)
+        return collection_id, name, symbol
+
+
+class MultiAssetMappingsParser(BaseParser[tuple[int, str]]):
```
LefterisJP, line 289: same here

rotkehlchen/globaldb/updates/types.py
```diff
@@ -0,0 +1,36 @@
+from enum import Enum, auto
+from typing import NamedTuple
+
+from rotkehlchen.assets.types import AssetType
```
LefterisJP, line 4: Can you rename the directory from `/updates` to `asset_updates`? This will make it clear what the difference to the `upgrades/` directory is.

rotkehlchen/globaldb/updates/types.py
```diff
@@ -0,0 +1,36 @@
...
+from rotkehlchen.types import Timestamp
+
+
+class ParsedAssetData(NamedTuple):
```
LefterisJP, line 8: We already have `AssetData` in assets/types.py. Why create yet another data structure here which is almost the same?

rotkehlchen/globaldb/upgrades/manager.py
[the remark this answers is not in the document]
LefterisJP: Why is this here? The fact that it was here before does not answer much. Is it something that always needs to be specified at initialization?

Keep in mind it's also done in `DB_SCRIPT_CREATE_TABLES` if it's a `fresh_db`

rotkehlchen/globaldb/utils.py
```diff
@@ -51,3 +59,64 @@ def set_token_spam_protocol(
...
+        ongoing_upgrade_from_version = -1  # Fresh DB
+
+    if ongoing_upgrade_from_version == -1:
+        return connection, False  # We are all good
+
+    # Otherwise replace the db with a backup and relogin
+    connection.close()
```
LefterisJP, line 100: do we have a test for this code? I can easily see the splitting of this function breaking something.

rotkehlchen/tests/api/test_assets_updates.py
```diff
@@ -722,20 +722,20 @@ def test_update_from_early_clean_db(
     version key set we still upgrade properly and set the assets version properly.
     """
     rotki = rotkehlchen_api_server.rest_api.rotkehlchen
-    update_1 = """INSERT INTO assets(identifier, name, type) VALUES('121-ada-FADS-as', 'A name', 'F'); INSERT INTO common_asset_details(identifier, symbol, coingecko, cryptocompare, forked, started, swapped_for) VALUES('121-ada-FADS-as', 'SYMBOL', '', '', 'BTC', NULL, NULL);
+    update_15 = """INSERT INTO assets(identifier, name, type) VALUES('121-ada-FADS-as', 'A name', 'F'); INSERT INTO common_asset_details(identifier, symbol, coingecko, cryptocompare, forked, started, swapped_for) VALUES('121-ada-FADS-as', 'SYMBOL', '', '', 'BTC', NULL, NULL);
```
LefterisJP, line 725: why did you have to rename here?

rotkehlchen/tests/unit/globaldb/test_asset_updates.py
```diff
@@ -530,23 +542,23 @@ def test_conflict_updates(assets_updater: AssetsUpdater, globaldb: GlobalDBHandl
 *"""  # noqa: E501
     update_patch = mock_asset_updates(
         original_requests_get=requests.get,
-        latest=2,
-        updates={'1': {
+        latest=16,
```
LefterisJP, line 545: same as before why the rename?

LefterisJP (commented):
So left a few comments.

For the code itself, it's a lot and hard to review so the best way to confirm corrrectness is exhaustive tests.

How is this entire functionality tested?

Assets updates from start until now, have their own test which confirms they work right? Waht about this new functionality of forcing a remote asset update during a breaking global DB schema update?

rotkehlchen/globaldb/updates/types.py
yabirgb: It is moved from other file. 
imagen

We should create an issue to get rid of it

yabirgb:
Looks as what I would expect for the change to be. The thing that concerns me the most is not having a good test for the functionality since I believe is key here. Have you tested it with one old globaldb to see how it works? We have some at rotki/assets/databases

rotkehlchen/globaldb/utils.py
prettyirrelevant: every test in the codebase that uses the globaldb calls this method.

rotkehlchen/tests/api/test_assets_updates.py
prettyirrelevant: the test was using an arbitrary update version but now that we're enforcing versions they need to be changed.

answers similar question below.

rotkehlchen/tests/unit/globaldb/test_asset_updates.py
prettyirrelevant: the test was using an arbitrary update version but now that we're enforcing versions they need to be changed.

rotkehlchen/globaldb/upgrades/manager.py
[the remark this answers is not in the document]
prettyirrelevant: >  Is it something that always needs to be specified at initialization?

yeah since its not a fresh database and foreign keys are not turned on by default when a db is created.

rotkehlchen/globaldb/asset_updates/manager.py
```diff
@@ -837,3 +674,55 @@ def _retrieve_update_files(
...
+                version_str = str(version)
+                if version_str not in info_json['updates']:
+                    continue
+
+                update_info = info_json['updates'][version_str]
+                if update_info['min_schema_version'] <= current_db_version <= update_info['max_schema_version']:  # noqa: E501
+                    max_compatible_version = version
+                elif update_info['min_schema_version'] > current_db_version:
+                    # Stop at first incompatible version
+                    break
```
LefterisJP, line 705:
```suggestion
                    break  # Stop at first incompatible version
```

rotkehlchen/globaldb/asset_updates/manager.py
```diff
@@ -837,3 +674,55 @@ def _retrieve_update_files(
...
+        except KeyError as e:
+            raise RemoteError(f'Didnt find expected key {e!s} in github assets json during update') from e  # noqa: E501
```
LefterisJP, line 707:
```suggestion
            raise RemoteError(f"Didn't find expected key {e!s} in github assets json during update") from e  # noqa: E501
```

rotkehlchen/globaldb/asset_updates/manager.py
```diff
@@ -837,3 +674,55 @@ def _retrieve_update_files(
...
+            latest_assets_version = info_json['latest']
+            current_db_version = self.globaldb.get_schema_version()
+            if current_db_version < FIRST_GLOBAL_DB_VERSION_WITH_COLLECTIONS:
+                return
+
+            start_version = max(FIRST_VERSION_WITH_COLLECTIONS, self.local_assets_version + 1)
+            # Find the highest compatible assets version for current DB version
+            for version in range(start_version, latest_assets_version + 1):
+                version_str = str(version)
+                if version_str not in info_json['updates']:
```
LefterisJP, line 697: As discussed in discord this should probably be let to raise a key error.

rotkehlchen/globaldb/asset_updates/manager.py
```diff
@@ -837,3 +674,51 @@ def _retrieve_update_files(
...
+                    to_version=max_compatible_version,
+                    global_db_version=current_db_version,
+                )
+
+                # get all the assets in the db and choose remote.
+                with self.globaldb.conn.read_ctx() as cursor:
+                    conflicts: dict[Asset, Literal['remote', 'local']] = {
+                        Asset(row[0]): 'remote'
+                        for row in cursor.execute('SELECT identifier FROM assets')
+                    }
```
yabirgb, line 716: wait this is quite heave. Can't you do `defaultdict(lambda: 'remote')` here/

rotkehlchen/globaldb/asset_updates/manager.py
```diff
@@ -837,3 +674,51 @@ def _retrieve_update_files(
...
+    def apply_pending_compatible_updates(self) -> None:
+        """Apply any pending asset updates that are compatible with the current DB version
+        before upgrading to target_db_version.
+
+        This ensures we don't miss any asset updates that would become incompatible
+        after the DB upgrade.
+        """
+        try:
+            max_compatible_version = None
+            info_json = self._get_remote_info_json()
```
yabirgb, line 687: Can we go with smaller try/except statements? This is the only thing that can raise RemoteError. So you can unindent the rest and catch the RemoteError. Log it properly and return

rotkehlchen/globaldb/asset_updates/manager.py
```diff
@@ -837,3 +674,51 @@ def _retrieve_update_files(
...
+                return
+
+            start_version = max(FIRST_VERSION_WITH_COLLECTIONS, self.local_assets_version + 1)
+            # Find the highest compatible assets version for current DB version
+            for version in range(start_version, latest_assets_version + 1):
+                update_info = info_json['updates'][str(version)]
+                if update_info['min_schema_version'] <= current_db_version <= update_info['max_schema_version']:  # noqa: E501
+                    max_compatible_version = version
+                elif update_info['min_schema_version'] > current_db_version:
+                    break  # Stop at first incompatible version
```
yabirgb, line 700: can you do a try/except here for keyError in this block? If one version is bad you can apply until that one here. Can't you?

prettyirrelevant: `._perform_update()` can also raise `RemoteError`

rotkehlchen/globaldb/asset_updates/manager.py
prettyirrelevant: thanks. i had to update its usage since `.get()` does not use `__getitem__`

rotkehlchen/globaldb/asset_updates/manager.py
```diff
@@ -837,3 +674,51 @@ def _retrieve_update_files(
...
+                with self.globaldb.conn.read_ctx() as cursor:
+                    conflicts: dict[Asset, Literal['remote', 'local']] = {
+                        Asset(row[0]): 'remote'
+                        for row in cursor.execute('SELECT identifier FROM assets')
+                    }
+
+                self.perform_update(
+                    up_to_version=max_compatible_version,
+                    conflicts=conflicts,
+                )
```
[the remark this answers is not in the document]
prettyirrelevant: yeah.

LefterisJP (approved):
LGTM

PR #9062, opened 2024-12-11
Ensure addresses are checksummed

yabirgb:
The API returns some addresses checksummed and others aren't. To ensure that we don't introduce bad ids in the DB we checksumm before creating the token


LefterisJP (approved):
LGTM

PR #9064, opened 2024-12-11
New AssetMovement in bitstamp bybit and coinbase

nicholasyoder:
Related #7007 

Note that there is some logic in bitstamp's `query_online_history_events` function for retrieving the address and transaction id for asset movements which isn't covered by the tests. I converted it to using the new history event based asset movements as well as I could, but I can't guarantee that it still works correctly since there aren't tests for this.

Also, one potentially confusing change is in `test_account_for_coinbase_income_expense`, where `expected_total_actions` was incremented from 7 to 8 - this is due to there being one asset movement with a fee being processed here, and asset movement fees are now in separate history events, which made this count go up one for the new event. But the other counts checked here remain the same since this fee is still handled in the same way as before by the actual accounting logic.

### DIFF
--- a/rotkehlchen/exchanges/bitstamp.py
+++ b/rotkehlchen/exchanges/bitstamp.py
@@ -2,6 +2,7 @@ import hashlib
 import hmac
 import logging
 import uuid
+from collections.abc import Sequence
 from http import HTTPStatus
 from json.decoder import JSONDecodeError
 from typing import TYPE_CHECKING, Any, Literal, NamedTuple, overload
@@ -14,19 +15,20 @@ from rotkehlchen.accounting.structures.balance import Balance
 from rotkehlchen.assets.asset import AssetWithOracles
 from rotkehlchen.assets.converters import asset_from_bitstamp
 from rotkehlchen.constants import ZERO
+from rotkehlchen.data_import.utils import maybe_set_transaction_extra_data
 from rotkehlchen.db.cache import DBCacheDynamic
+from rotkehlchen.db.history_events import DBHistoryEvents
 from rotkehlchen.errors.asset import UnknownAsset, UnsupportedAsset
 from rotkehlchen.errors.misc import RemoteError
 from rotkehlchen.errors.serialization import DeserializationError
-from rotkehlchen.exchanges.data_structures import (
-    AssetMovement,
-    AssetMovementCategory,
-    MarginPosition,
-    Trade,
-    TradeType,
-)
+from rotkehlchen.exchanges.data_structures import MarginPosition, Trade, TradeType
 from rotkehlchen.exchanges.exchange import ExchangeInterface, ExchangeQueryBalances
 from rotkehlchen.history.deserialization import deserialize_price
+from rotkehlchen.history.events.structures.asset_movement import (
+    AssetMovement,
+    create_asset_movement_with_fee,
+)
+from rotkehlchen.history.events.structures.types import HistoryEventType
 from rotkehlchen.inquirer import Inquirer
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.serialization.deserialize import (
@@ -40,12 +42,11 @@ from rotkehlchen.types import (
     ApiSecret,
     AssetAmount,
     ExchangeAuthCredentials,
-    Fee,
     Location,
     Timestamp,
 )
 from rotkehlchen.user_messages import MessagesAggregator
-from rotkehlchen.utils.misc import ts_now_in_ms
+from rotkehlchen.utils.misc import ts_now_in_ms, ts_sec_to_ms
 from rotkehlchen.utils.mixins.cacheable import cache_response_timewise
 from rotkehlchen.utils.mixins.lockable import protect_with_lock
 from rotkehlchen.utils.serialization import jsonloads_dict, jsonloads_list
@@ -228,11 +229,11 @@ class Bitstamp(ExchangeInterface):
 
         return assets_balance, ''
 
-    def query_online_deposits_withdrawals(
+    def query_online_history_events(
             self,
             start_ts: Timestamp,
             end_ts: Timestamp,
-    ) -> list[AssetMovement]:
+    ) -> Sequence[AssetMovement]:
         """Return the account asset movements on Bitstamp.
 
         NB: when `since_id` is used, the Bitstamp API v2 will return by default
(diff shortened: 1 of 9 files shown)


rotkehlchen/exchanges/bitstamp.py
```diff
@@ -246,15 +248,26 @@ def query_online_deposits_withdrawals(
...
             with self.db.conn.read_ctx() as cursor:
-                query_result = cursor.execute(
-                    'SELECT link FROM asset_movements WHERE location=? AND timestamp <= ? ORDER BY timestamp DESC LIMIT 1',  # noqa: E501
-                    (Location.BITSTAMP.serialize_for_db(), start_ts),
-                ).fetchone()
-                if query_result is not None:
-                    since_id = int(query_result[0]) + 1
+                query_result = history_db.get_history_events(
+                    cursor=cursor,
+                    filter_query=HistoryEventFilterQuery.make(
```
LefterisJP, line 257: If you want to use our own classes then you would need to also create an AssetMovementFilterQuery class, which would explicitly query asset movements. As this would return all types of events.

You can also simply mimic the approach they had before and do a direct DB query. Perhaps that's easier and it's definitely more performant.

rotkehlchen/exchanges/bitstamp.py
```diff
@@ -282,10 +295,17 @@ def query_online_deposits_withdrawals(
         # they correspond to so we can also have the fee taken into account
         indices_to_delete = []
         for asset_movement in asset_movements:
+            assert asset_movement.extra_data is not None  # should always be set here
```
LefterisJP, line 298: We try to avoid assertions in the code. Prefer type ignore.

But this shows something weird. Isn't `extra_data` always set in one way or another for asset movements? If so the class should have it as non optional so all these type checks should succeed and not need type ignore or assets. It will be much better if you can make it so that it's set always if this is what happens.

Same for all other places (like right below) where you do this in this PR and probably in other parts of the code.

rotkehlchen/exchanges/bitstamp.py
```diff
@@ -295,18 +315,31 @@ def query_online_deposits_withdrawals(
...
         # may end up with some crypto asset movements here that would need to
         # check for corresponding asset movement in the DB.
-        serialized_location = Location.BITSTAMP.serialize_for_db()
         indices_to_delete = []
         with self.db.user_write() as write_cursor:
             for idx, crypto_movement in enumerate(crypto_asset_movements):
-                write_cursor.execute(
-                    'SELECT id from asset_movements WHERE location=? AND category=? AND timestamp=? AND asset=?',  # noqa: E501
-                    (serialized_location, crypto_movement.category.serialize_for_db(), crypto_movement.timestamp, crypto_movement.asset.identifier),  # noqa: E501
+                query_result = history_db.get_history_events(
```
LefterisJP, line 321: same here for the query you are doing as I said above. Also isn't this logic repeating from above?

rotkehlchen/exchanges/bitstamp.py
```diff
@@ -295,18 +315,31 @@ def query_online_deposits_withdrawals(
...
                 )
-                if (result := write_cursor.fetchone()) is not None:
-                    write_cursor.execute(
-                        'UPDATE asset_movements SET address=?, transaction_id=? WHERE id=?',
-                        (crypto_movement.address, crypto_movement.transaction_id, result[0]),
+                if len(query_result) != 0:
+                    matched_movement = query_result[0]
+                    extra_data = matched_movement.extra_data if matched_movement.extra_data is not None else {}  # noqa: E501
+                    assert crypto_movement.extra_data is not None  # should always be set here
+                    extra_data.update(crypto_movement.extra_data)
```
LefterisJP, line 338: same comments as above here. Though the direct query here looks worse than in the filter/get. the edit with the function is probably fine

rotkehlchen/exchanges/bitstamp.py
```diff
@@ -689,23 +722,25 @@ def _deserialize_asset_movement_from_user_transaction(
...
             asset=fee_asset,
             amount=abs(amount),
             fee_asset=fee_asset,
-            fee=deserialize_fee(raw_movement['fee']),
-            link=str(raw_movement['id']),
+            fee=(fee := deserialize_fee(raw_movement['fee'])),
+            unique_id=(movement_id := str(raw_movement['id'])),
+            extra_data={
+                'movement_id': movement_id,
+                'fee': fee,  # Used for matching the corresponding crypto_transaction. Removed before being saved to the DB.  # noqa: E501
```
LefterisJP, line 736: This is super hacky. I would leave a note here and in the project notes as a TODO to rethink and fix this. This entire part of the code is really hacky

rotkehlchen/exchanges/coinbase.py
```diff
@@ -607,28 +610,26 @@ def _query_single_account_transactions(
             account_id: str,
             account_last_id_name: str,
             last_tx_id: str | None,
-    ) -> tuple[list[Trade], list[AssetMovement], list[HistoryEvent]]:
+    ) -> tuple[list[Trade], Sequence['HistoryBaseEntry']]:
```
LefterisJP, line 613: Why not be more specific? Do types not work then? `HistoryEvent|AssetMovement`

rotkehlchen/exchanges/coinbase.py
```diff
@@ -898,16 +899,15 @@ def _deserialize_asset_movement(self, raw_data: dict[str, Any]) -> AssetMovement
 
             # Only get address/transaction id for "send" type of transactions
             address = None
-            transaction_id = None
-            transaction_url = raw_data.get('id', '')
+            transaction_id = raw_data.get('id')
+            transaction_url = None
```
LefterisJP, line 903: you are switching them from before. This changes the implementation. your change appears to be correct but that still means the implementation changes.

You understand this means that after this PR half your asset movements will have swapped transactionid/url for coinbase for things before running this version.

So if this is the right approach this needs to also be swapped during the DB upgrade. Right?

rotkehlchen/exchanges/bitstamp.py
nicholasyoder: > Isn't extra_data always set in one way or another for asset movements?

No, `extra_data` is not always set. For instance fees generally have no `extra_data` set, and even normal deposit/withdrawals are not guaranteed to have anything in `extra_data`. Which actually means that it needs to be doing an if check on `asset_movement.extra_data` here to skip the fee events, so I've added that.  `crypto_movement.extra_data` should always be set though, so I changed that to use type ignore.

rotkehlchen/exchanges/bitstamp.py
nicholasyoder: Also changed this to do a direct DB query, although it can't just retrieve the identifier as it did before since we need to actually deserialize the event for use with the `edit_event_extra_data` function.

> isn't this logic repeating from above?

not exactly - here its trying to combine the crypto_asset_movements with ones already saved to the database, while above its combining with the ones it just queried.

rotkehlchen/exchanges/bitstamp.py
nicholasyoder: still using the edit function as suggested, but changed the assert to a type ignore.

rotkehlchen/exchanges/bitstamp.py
nicholasyoder: Yes, this isn't a very elegant solution. Added a todo comment in the code and created a note: https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=90720824

rotkehlchen/exchanges/coinbase.py
nicholasyoder: These were being used to set the `transaction_id` and `link` properties of the AssetMovement, neither of which exist in the new AssetMovement. Now `transaction_id` and `transaction_url` are simply set in `extra_data`, as well as `transaction_id` being used for the `unique_id` (to create the `event_identifier`).

As discussed with yabir in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1316794835830706207), the `link` (which is what may contain this `id` value) will be added to `extra_data`. So while it may not have the same key name, it should be present in the upgraded events' `extra_data`.

LefterisJP (approved):
LGTM. Thanks

PR #9067, opened 2024-12-12
fix: do truncate address better

lukicenturi:
Increase the limit (3 chars), because the `...` takes up 3 spaces, so instead of showing the `...`, just show the original string.

Before: 
image

After:
image


kelsos (approved):
lgtm

PR #9074, opened 2024-12-12
Move asset movements to history events

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/db/drivers/gevent.py
+++ b/rotkehlchen/db/drivers/gevent.py
@@ -92,7 +92,11 @@ class DBCursor:
             logger.trace(f'FINISH EXECUTE {statement}')
         return self
 
-    def executemany(self, statement: str, *bindings: Sequence[Sequence]) -> 'DBCursor':
+    def executemany(
+            self,
+            statement: str,
+            *bindings: Sequence[Sequence] | Generator[Sequence, None, None],
+    ) -> 'DBCursor':
         if __debug__:
             logger.trace(f'EXECUTEMANY {statement}')
         self._cursor.executemany(statement, *bindings)
--- a/rotkehlchen/db/upgrades/v45_v46.py
+++ b/rotkehlchen/db/upgrades/v45_v46.py
@@ -3,9 +3,21 @@ import logging
 import urllib.parse
 from typing import TYPE_CHECKING
 
+from rotkehlchen.assets.asset import Asset
 from rotkehlchen.constants import ALLASSETIMAGESDIR_NAME, ASSETIMAGESDIR_NAME, IMAGESDIR_NAME
 from rotkehlchen.constants.assets import A_COW, A_GNOSIS_COW, A_LQTY
+from rotkehlchen.data_import.utils import maybe_set_transaction_extra_data
+from rotkehlchen.errors.asset import UnknownAsset
+from rotkehlchen.errors.serialization import DeserializationError
+from rotkehlchen.fval import FVal
+from rotkehlchen.history.events.structures.asset_movement import (
+    AssetMovement,
+    create_asset_movement_with_fee,
+)
+from rotkehlchen.history.events.structures.types import HistoryEventType
 from rotkehlchen.logging import RotkehlchenLogsAdapter, enter_exit_debug_log
+from rotkehlchen.types import Location
+from rotkehlchen.utils.misc import ts_sec_to_ms
 from rotkehlchen.utils.progress import perform_userdb_upgrade_steps, progress_step
 
 if TYPE_CHECKING:
@@ -62,5 +74,59 @@ def upgrade_v45_to_v46(db: 'DBHandler', progress_handler: 'DBUpgradeProgressHand
             'WHERE evm_events_info.identifier = history_events.identifier);',
         )
         write_cursor.execute('ALTER TABLE evm_events_info DROP COLUMN extra_data;')
+
+    @progress_step(description='Converting asset movements to history events')
+    def move_asset_movements(write_cursor: 'DBCursor') -> None:
+        new_events: list[AssetMovement] = []
+        write_cursor.execute(
+            'SELECT id, location, category, address, transaction_id, timestamp, asset, '
+            'amount, fee_asset, fee, link FROM asset_movements',
+        )
+
+        for row in write_cursor:
+            try:
+                new_events.extend(create_asset_movement_with_fee(
+                    timestamp=ts_sec_to_ms(row[5]),
+                    location=Location.deserialize_from_db(row[1]),
+                    event_type=HistoryEventType.DEPOSIT if row[2] == 'A' else HistoryEventType.WITHDRAWAL,  # noqa: E501
+                    asset=Asset(row[6]),
+                    amount=FVal(row[7]),
+                    fee_asset=Asset(row[8]),
+                    fee=FVal(row[9]),
+                    extra_data=maybe_set_transaction_extra_data(
+                        address=row[3],
+                        transaction_id=row[4],
+                        extra_data={'reference': row[10]} if row[10] is not None and row[10] not in ('', 'None') else None,  # in coinbase we were storing 'None' as string.  # noqa: E501
+                    ),
+                ))
+            except (DeserializationError, UnknownAsset) as e:
+                log.error(f'Failed to deserialize row {row} due to {e}. Skipping.')
+
+        binding_tuples = ((
+            event.entry_type.value,
+            event.event_identifier,
+            event.sequence_index,
+            event.timestamp,
+            event.location.serialize_for_db(),
+            event.location_label,
+            event.asset.identifier,
+            str(event.balance.amount),
+            str(event.balance.usd_value),
+            event.notes,
+            event.event_type.serialize(),
+            event.event_subtype.serialize(),
+            json.dumps(event.extra_data) if event.extra_data else None,
+        ) for event in new_events)
+        write_cursor.executemany(
+            (
+                'INSERT OR IGNORE INTO history_events(entry_type, event_identifier, '
+                'sequence_index, timestamp, location, location_label, asset, amount, usd_value, '
+                'notes, type, subtype, extra_data) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)'
+            ),
+            binding_tuples,
+        )
+        # TODO: Uncomment this two lines when we are ready to drop the tables
+        # write_cursor.execute('DROP TABLE asset_movements')  # noqa: ERA001
+        # write_cursor.execute('DROP TABLE asset_movement_category')  # noqa: ERA001
 
     perform_userdb_upgrade_steps(db=db, progress_handler=progress_handler, should_vacuum=True)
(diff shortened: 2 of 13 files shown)


rotkehlchen/db/drivers/gevent.py
```diff
@@ -92,7 +92,10 @@ def execute(self, statement: str, *bindings: Sequence) -> 'DBCursor':
             logger.trace(f'FINISH EXECUTE {statement}')
         return self
 
-    def executemany(self, statement: str, *bindings: Sequence[Sequence]) -> 'DBCursor':
+    def executemany(
+            self,
+            statement: str, *bindings: Sequence[Sequence] | Generator[Sequence, None, None],
```
nicholasyoder, line 97:
```suggestion
            self,
            statement: str, 
            *bindings: Sequence[Sequence] | Generator[Sequence, None, None],
```

rotkehlchen/db/upgrades/v45_v46.py
```diff
@@ -63,4 +74,58 @@ def _move_extra_data(write_cursor: 'DBCursor') -> None:
...
+            event.location.serialize_for_db(),
+            event.location_label,
+            event.asset.identifier,
+            str(event.balance.amount),
+            str(event.balance.usd_value),
+            event.notes,
+            event.event_type.serialize(),
+            event.event_subtype.serialize(),
+            json.dumps(event.extra_data) if event.extra_data else None,
+        ) for event in new_events)
```
nicholasyoder, line 118: Wouldn't `event.serialize_for_db()` work here?

rotkehlchen/db/upgrades/v45_v46.py
```diff
@@ -63,4 +74,58 @@ def _move_extra_data(write_cursor: 'DBCursor') -> None:
...
+                    location=Location.deserialize_from_db(row[1]),
+                    event_type=HistoryEventType.DEPOSIT if row[2] == 'A' else HistoryEventType.WITHDRAWAL,  # noqa: E501
+                    asset=Asset(row[6]),
+                    amount=FVal(row[7]),
+                    fee_asset=Asset(row[8]),
+                    fee=FVal(row[9]),
+                    extra_data=maybe_set_transaction_extra_data(
+                        address=row[3],
+                        transaction_id=row[4],
+                        extra_data={'url': row[10]} if row[10] is not None and row[10] not in ('', 'None') else None,  # in coinbase we were storing 'None' as string.  # noqa: E501
```
nicholasyoder, line 98: Note that `link` isn't always a url. Its often an id of some kind. But I guess anything we call this here will not really apply for some of the events, so I suppose using `url` will work.

yabirgb: I used url because is what you had in your PR. Intitially I had link that is what we had in the origins structure. Would it make sense to rename it? Maybe something better than link? If changed it needs to be changed also were you used url

rotkehlchen/db/upgrades/v45_v46.py
yabirgb: yes but I did it like this in case we change the serialization logic due to changes in the schema

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -31,14 +31,26 @@
...
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
 
-class AssetMovement(HistoryBaseEntry):
+class AssetMovementExtraData(TypedDict):
+    address: NotRequired[str]
+    transaction_id: NotRequired[str]
+    reference: NotRequired[str]
+    movement_id: NotRequired[str]
```
yabirgb, line 44: where have movement_id and reference. Should we just use reference? (reference is how I named what was previously link and url by you)

rotkehlchen/exchanges/bitstamp.py
```diff
@@ -729,7 +729,7 @@ def _deserialize_asset_movement_from_user_transaction(
             unique_id=(movement_id := str(raw_movement['id'])),
             extra_data={
                 'movement_id': movement_id,
-                'fee': fee,  # Used for matching the corresponding crypto_transaction. Removed before being saved to the DB.  # noqa: E501
+                'fee': fee,
```
nicholasyoder, line 732: This is used here: https://github.com/rotki/rotki/blob/27d664cae869c24e721a36060805292149aaead7/rotkehlchen/exchanges/bitstamp.py#L300 to match the events from the crypto transaction endpoint, which include the fee in their amount. There not really any point in saving this to the database since this is the only place its used, so then I have it being removed just below: https://github.com/rotki/rotki/blob/27d664cae869c24e721a36060805292149aaead7/rotkehlchen/exchanges/bitstamp.py#L304 But this is rather hacky and there's a note about the logic in this exchange needing improvement: https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=90720824

rotkehlchen/history/events/structures/asset_movement.py
nicholasyoder: Yes, `movement_id` should probably be replaced with `reference`. No need to have more different keys than necessary.

rotkehlchen/db/upgrades/v45_v46.py
nicholasyoder: ok, makes sense.

rotkehlchen/db/upgrades/v45_v46.py
```diff
@@ -63,4 +75,58 @@ def _move_extra_data(write_cursor: 'DBCursor') -> None:
...
+                    ),
+                ))
+            except (DeserializationError, UnknownAsset) as e:
+                log.error(f'Failed to deserialize row {row} due to {e}. Skipping.')
+
+        binding_tuples = ((
+            event.entry_type.value,
+            event.event_identifier,
+            event.sequence_index,
+            int(event.timestamp),
```
LefterisJP, line 109: why do you need int here?

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -31,14 +31,25 @@
...
     from rotkehlchen.accounting.pot import AccountingPot
     from rotkehlchen.fval import FVal
+    from rotkehlchen.types import Fee
 
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
 
-class AssetMovement(HistoryBaseEntry):
+class AssetMovementExtraData(TypedDict):
```
LefterisJP, line 40: please add a small docstring here

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -31,14 +31,25 @@
...
+    address: NotRequired[str]
+    transaction_id: NotRequired[str]
+    reference: NotRequired[str]
```
LefterisJP, line 43: Wherever you prefer, either in the docstring or on theright or on top of each with a `#` add a small description of what each field is. Especially useful for reference and transaction id

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2683,6 +2684,15 @@ def test_upgrade_db_45_to_46(user_data_dir: 'Path', messages_aggregator):
             history_event_bindings,
         ).fetchone()[0] == 1
 
+        assert cursor.execute(
```
LefterisJP, line 2687: should you also add a COUNT for asset movements table being zero? With a comment to change it to check the 2 tables are deleted after all is done?

rotkehlchen/db/upgrades/v45_v46.py
yabirgb: It was a copy paste but I removed it in the history events serialization function

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -31,14 +31,29 @@
...
 
 
-class AssetMovement(HistoryBaseEntry):
+class AssetMovementExtraData(TypedDict):
+    """Typed dict with all the valid fields used in extra_data for AssetMovements"""
+    # Address receiving or sending funds in the asset movement.
+    address: NotRequired[str]
+    # Transaction if it is available where funds were moved on chain.
+    transaction_id: NotRequired[str]
+    # Internal reference used in exchanges. Can be either an id or a url.
```
LefterisJP, line 46: what do you mean by either an id or a url? This needs to be the unique exchange id reference. Not a link

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -31,14 +31,29 @@
...
+    reference: NotRequired[str]
+    # Used for matching the corresponding crypto_transaction. Removed before being saved to the DB.  # noqa: E501
```
LefterisJP, line 48:
```suggestion
    # Internal use only. Used for matching the corresponding crypto_transaction. Removed before being saved to the DB.  # noqa: E501
```

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2683,6 +2684,18 @@ def test_upgrade_db_45_to_46(user_data_dir: 'Path', messages_aggregator):
...
+            'timestamp, location, location_label, asset, amount, usd_value, notes, '
+            'type, subtype, extra_data FROM history_events WHERE entry_type=6',
+        ).fetchall() == [
+            (6, '20522c693bcda4ef646682c6a58bb0349b01f4d7b9168a62ce94b2c8dd1fe639', 0, 1644244023000, 'B', None, 'ATOM', '3.40000000', '0', 'Withdraw 3.40000000 ATOM from Kraken', 'withdrawal', 'remove asset', '{"reference": "KRAKEN-XXX"}'),  # noqa: E501
+            (6, '20522c693bcda4ef646682c6a58bb0349b01f4d7b9168a62ce94b2c8dd1fe639', 1, 1644244023000, 'B', None, 'ATOM', '0.10000000', '0', 'Pay 0.10000000 ATOM as Kraken withdrawal fee', 'withdrawal', 'fee', None),  # noqa: E501
+            (6, '1cf3fc675d4835efe532d18730db35163b07fb5f44956b888d7aaf852b04fd1c', 0, 1727542684000, 'G', None, 'ETH', '0.07126231', '0', 'Deposit 0.07126231 ETH to Coinbase', 'deposit', 'deposit asset', '{"address": "0xc37b40ABdB939635068d3c5f13E7faF686F03B65", "transaction_id": "0xc8aa60c9f8a93692b66fbc3f57c64f2b1c4afd92f370e12dc7bb23acfa303dbb"}'),  # noqa: E501
+        ]
+        # check that asset movements table got removed. Uncommented when we address the tables removal  # noqa: E501
```
LefterisJP, line 2696:
```suggestion
        # TODO: Remove this when table is removed
        assert cursor.execute('SELECT COUNT(*) FROM asset_movements').fetchone()[0] == 0
        # check that asset movements table got removed. TODO: Uncomment when we address the tables removal  # noqa: E501
```

rotkehlchen/history/events/structures/asset_movement.py
yabirgb: Fixed it in coinbase where we were using a url here

rotkehlchen/tests/db/test_db_upgrades.py
LefterisJP: you did not do this. You just added the comment to remove the assertion, without the assertion itself.

`assert cursor.execute('SELECT COUNT(*) FROM asset_movements').fetchone()[0] == 0`

LefterisJP (approved):
LGTM, with the addendum you forgot something

PR #9075, opened 2024-12-12
Include liabilities when filtering usd value

yabirgb:
Include liabilities when filtering assets. They weren't being added and completly ignored so they didn't appear in the response to the frontend


LefterisJP (approved):
LGTM

PR #9076, opened 2024-12-12
New AssetMovement in coinbaseprime gemini and htx

nicholasyoder:
Related #7007


rotkehlchen/exchanges/htx.py
```diff
@@ -298,17 +305,19 @@ def _query_deposits_withdrawals(
...
+                    event_type=event_type,
                     asset=coin,
                     amount=deserialize_asset_amount(movement['amount']),
                     fee_asset=coin,
                     fee=deserialize_fee(movement.get('fee', '0')),
-                    link=str(movement['id']),
+                    unique_id=str(movement['id']),
+                    extra_data=maybe_set_transaction_extra_data(
+                        address=movement['address'],
+                        transaction_id=movement['tx-hash'],
```
prettyirrelevant, line 319: can't you combine this with the `id` and use that as the `unique_id`?

rotkehlchen/exchanges/gemini.py
```diff
@@ -511,19 +513,18 @@ def query_online_deposits_withdrawals(
...
-                    amount=deserialize_asset_amount_force_positive(entry['amount']),
-                    fee_asset=asset,
-                    # Gemini does not include withdrawal fees neither in the API nor in their UI
-                    fee=Fee(ZERO),
-                    link=str(entry['eid']),
+                    balance=Balance(deserialize_asset_amount_force_positive(entry['amount'])),
+                    unique_id=str(entry['eid']),
+                    extra_data=maybe_set_transaction_extra_data(
+                        address=deserialize_asset_movement_address(entry, 'destination', asset),
+                        transaction_id=get_key_if_has_val(entry, 'txHash'),
```
prettyirrelevant, line 526: same question about combining the tx hash with `eid` in this case.

rotkehlchen/exchanges/htx.py
prettyirrelevant, line 319: why not use `get_key_if_has_val()` here?

prettyirrelevant (commented):
left some comments.

rotkehlchen/exchanges/htx.py
nicholasyoder: I assumed that since it wasn't being used here before that these were guaranteed to be present, but I checked the [htx docs](https://www.htx.com/en-us/opend/newApiPages/?id=7ec4f050-7773-11ed-9966-0242ac110003) and it does say that at least `tx-hash` may not always be present, so I've changed these to use `get_key_if_has_val()`.

rotkehlchen/exchanges/gemini.py
nicholasyoder: same here, I think `eid` should be unique on its own. Also, since `txHash` is being retrieved with `get_key_if_has_val` it would seem that it may not always be present as well.

rotkehlchen/exchanges/htx.py
nicholasyoder: unique_id is used in combination with the location to create the event_identitfier, so it simply needs to be unique on this exchange. I think this `id` should be unique on this exchange, so I'm not sure there's really any point in combining it with `tx-hash`, expecially since `tx-hash` may not even be present some of the time as mentioned in my reply to your comment about `get_key_if_has_val`.

LefterisJP (approved):
LGTM

PR #9077, opened 2024-12-13
fix: empty ETH native breakdown table in blockchain balances

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/store/blockchain/index.ts
+++ b/frontend/app/src/store/blockchain/index.ts
@@ -259,12 +259,12 @@ export const useBlockchainStore = defineStore('blockchain', () => {
 
   const { getAssetAssociationIdentifiers, getAssociatedAssetIdentifier } = useAssetInfoRetrieval();
 
-  const getBreakdown = (asset: string, isLiability = false, chains?: string[], groupId?: string): AssetBreakdown[] => {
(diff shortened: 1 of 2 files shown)


frontend/app/src/store/blockchain/index.ts
```diff
@@ -295,11 +295,11 @@ export const useBlockchainStore = defineStore('blockchain', () => {
     }
 
     return breakdown;
-  };
+  });
 
-  const assetBreakdown = (asset: string, chains?: string[], groupId?: string): AssetBreakdown[] => getBreakdown(asset, false, chains, groupId);
+  const assetBreakdown = (asset: string, chains?: string[], groupId?: string): ComputedRef<AssetBreakdown[]> => getBreakdown(asset, false, chains, groupId);
```
kelsos, line 300: do we expose the `getBreakdown` also to non-reactive context?

I mostly wonder if it would work if we kept the `getBreakdown` as a normal function and used `reactify` for the asset/liability breakdown functions

kelsos (approved):
lgtm thanks.

frontend/app/src/store/blockchain/index.ts
kelsos: I checked and we don't see to use it in a non-reactive context so it's ok

PR #9078, opened 2024-12-13
add create asset movement api schema

prettyirrelevant:


rotkehlchen/api/v1/schemas.py
```diff
@@ -945,12 +946,37 @@ def make_history_base_entry(
...
+            enum_class=HistoryEventType,
+            exclude_types=[
+                event_type for event_type in HistoryEventType
+                if event_type not in (HistoryEventType.DEPOSIT, HistoryEventType.WITHDRAWAL)
+            ],
+            required=True,
+        )
+        asset = AssetField(required=True, expected_type=Asset, form_with_incomplete_data=True)
+        event_identifier = fields.String(required=False, load_default=None)
+        unique_id = fields.String(required=False, load_default=None)
```
LefterisJP, line 960: you have extra_data and this, while this is supposed to be part of extra data. This seems wrong.

rotkehlchen/api/v1/schemas.py
```diff
@@ -945,12 +946,37 @@ def make_history_base_entry(
...
+        extra_data = fields.Dict(load_default=None)
```
LefterisJP, line 961: Since this is something coming from the API, thus the user we don't need to pass the extra data directly. Just have fields for each possible value and type check each field on its own merit. We can't have any kind of extra data there, can we?

rotkehlchen/api/v1/schemas.py
```diff
@@ -945,12 +946,37 @@ def make_history_base_entry(
...
+        location = LocationField(required=True)
+        is_fee = fields.Boolean(required=True)
```
LefterisJP, line 963: again since this is from the user it makes no sense for them to repeat same things for fee. Instead add an optional fee amount (andif asset can be different, optional fee asset) and we can turn it into the second event in the backend.

rotkehlchen/api/v1/schemas.py
nicholasyoder: `unique_id` is only used to create the event_identifier - it doesn't actually get put in `extra_data` automatically. Of course sometimes the value used for `unique_id` also gets added to `extra_data`, for instance if its an onchain tx hash or something, but I think there should be be the option of specifying a unique_id here.

rotkehlchen/api/v1/schemas.py
```diff
@@ -943,14 +947,67 @@ def make_history_base_entry(
...
+        asset = AssetField(required=True, expected_type=Asset, form_with_incomplete_data=True)
+        fee_asset = AssetField(load_default=None, required=False, expected_type=Asset, form_with_incomplete_data=True)  # noqa: E501
+
+        @post_load
+        def make_history_base_entry(
+                self,
+                data: dict[str, Any],
+                **_kwargs: Any,
+        ) -> dict[str, Any]:
+            if (fee := data['fee']) is not None and data['fee_asset'] is None:
```
yabirgb, line 971: maybe you want something like

`(fee := data['fee']) is None ^ data['fee_asset'] is None`

using the XOR operator here. You want both of them or none

yabirgb: right below you check again `if fee is not None and fee > ZERO:` so maybe can be combined properly.

rotkehlchen/api/v1/schemas.py
```diff
@@ -943,14 +947,67 @@ def make_history_base_entry(
...
+                main_event, *related_events = create_asset_movement_with_fee(
+                    fee=fee,
+                    asset=data['asset'],
+                    location=data['location'],
+                    timestamp=data['timestamp'],
+                    fee_asset=data['fee_asset'],
+                    event_type=data['event_type'],
+                    identifier=data.get('identifier'),
+                    amount=data['balance'].amount,
+                )
```
yabirgb, line 987:
```
            if (fee := data['fee']) is not None:
                if data['fee_asset'] is None:
                    raise ValidationError(
                        message='fee_asset must be provided if fee is set',
                        field_name='fee',
                    )

                if fee > ZERO:
                    main_event, *related_events = create_asset_movement_with_fee(
                        fee=fee,
                        asset=data['asset'],
                        location=data['location'],
                        timestamp=data['timestamp'],
                        fee_asset=data['fee_asset'],
                        event_type=data['event_type'],
                        identifier=data.get('identifier'),
                        amount=data['balance'].amount,
                    )
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -943,14 +947,67 @@ def make_history_base_entry(
...
+                    location=data['location'],
+                    unique_id=data['unique_id'],
+                    timestamp=data['timestamp'],
+                    identifier=data.get('identifier'),
+                    event_type=data['event_type'],
+                    event_identifier=data['event_identifier'],
+                )
+                related_events = None
+
+            return {'main_event': main_event, 'related_events': related_events}
```
yabirgb, line 1002: I believe there is no reason to add two fields. Make it return a list and main_event be included inside the list

rotkehlchen/api/rest.py
```diff
@@ -1100,12 +1110,16 @@ def add_history_event(self, event: 'HistoryBaseEntry') -> Response:
...
+    def edit_history_event(self, main_event: 'HistoryBaseEntry', related_events: list['HistoryBaseEntry'] | None) -> Response:  # noqa: E501
         db = DBHistoryEvents(self.rotkehlchen.data.db)
-        result, msg = db.edit_history_event(event)
+        result, msg = db.edit_history_event(main_event)
         if result is False:
             return api_response(wrap_in_fail_result(msg), status_code=HTTPStatus.CONFLICT)
 
+        if related_events is not None and len(related_events) > 0:
+            for event in related_events:
+                db.edit_history_event(event)
```
yabirgb, line 1121: if you check in both methods you edited to:

1. duplicate the same code that you use in the main_event
2. Add a for loop and here for example it doesn't even have proper handling of the events. You shouldn't differienciate between main and related

rotkehlchen/api/rest.py
```diff
@@ -1077,17 +1077,27 @@ def get_asset_movements(
...
                         HISTORY_MAPPING_KEY_STATE: HISTORY_MAPPING_STATE_CUSTOMIZED,
                     },
                 )
+
+                if related_events is not None and len(related_events) > 0:
+                    for event in related_events:
+                        db.add_history_event(
+                            write_cursor=cursor,
+                            event=event,
+                            mapping_values={
```
yabirgb, line 1097: can you do it in a transaction? If you for example two events you don't want one being added and the other missing in the db. If the transaction fails you want to revert any already added event

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -635,3 +635,76 @@ def test_query_new_events(rotkehlchen_api_server_with_exchanges: 'APIServer') ->
...
+        assert 'identifier' in result
+        entry['identifier'] = result['identifier']
+
+    with rotki.data.db.conn.read_ctx() as cursor:
+        saved_events = db.get_history_events(
+            cursor=cursor,
+            filter_query=HistoryEventFilterQuery.make(),
+            has_premium=True,
+            group_by_event_ids=False,
+        )
```
yabirgb, line 680:
```suggestion
       assert len(db.get_history_events(
            cursor=cursor,
            filter_query=HistoryEventFilterQuery.make(),
            has_premium=True,
            group_by_event_ids=False,
        )) == 3
```

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -635,3 +635,76 @@ def test_query_new_events(rotkehlchen_api_server_with_exchanges: 'APIServer') ->
...
+    assert len(saved_events) == 3  # including the fee event.
+
```
yabirgb, line 682: shouldn't you check that the events are what you expect and the fields are populated as you expect?

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -635,3 +635,105 @@ def test_query_new_events(rotkehlchen_api_server_with_exchanges: 'APIServer') ->
...
+            group_by_event_ids=False,
+        )
+        assert len(saved_events) == 3
+        for idx, event in enumerate(saved_events):
+            serialized_event = event.serialize()
+            fields_to_exclude = {
+                'notes', 'extra_data', 'event_subtype', 'sequence_index',
+                'location_label', 'unique_id', 'event_identifier',
+                'fee', 'fee_asset', 'timestamp',
+            }
```
yabirgb, line 719: nitpick: this can be outside the for loop

rotkehlchen/tests/api/test_history_base_entry.py
yabirgb: This wasn't done but was more a nitpick

rotkehlchen/db/history_events.py
```diff
@@ -148,36 +149,35 @@ def add_history_events(
                 event=event,
             )
 
-    def edit_history_event(self, event: HistoryBaseEntry) -> tuple[bool, str]:
+    def edit_history_event(self, write_cursor: 'DBCursor', event: HistoryBaseEntry) -> None:
```
yabirgb, line 152: this method seems to be used only in the rest_api.py file and there the error is handled now.

The reason to raise the exception is so we can revert when editint multiple events

rotkehlchen/api/rest.py
```diff
@@ -1077,34 +1077,47 @@ def get_asset_movements(
...
+        result_dict = _wrap_in_ok_result({'identifier': main_identifier})
         return api_response(result_dict, status_code=HTTPStatus.OK)
 
-    def edit_history_event(self, event: 'HistoryBaseEntry') -> Response:
+    def edit_history_events(self, events: list['HistoryBaseEntry']) -> Response:
         db = DBHistoryEvents(self.rotkehlchen.data.db)
-        result, msg = db.edit_history_event(event)
-        if result is False:
-            return api_response(wrap_in_fail_result(msg), status_code=HTTPStatus.CONFLICT)
+        try:
```
LefterisJP, line 1112: both here and above why is the try above the `with:` ?

`self.rotkehlchen.data.db.user_write()` can't raise can it?

Is it so that any in-between add/edits are undone thanks to the exception uwinding?

rotkehlchen/api/v1/schemas.py
```diff
@@ -943,14 +947,63 @@ def make_history_base_entry(
...
+    class CreateAssetMovementEventSchema(BaseSchema):
+        event_type = SerializableEnumField(
+            enum_class=HistoryEventType,
+            allow_only=[HistoryEventType.DEPOSIT, HistoryEventType.WITHDRAWAL],
+            required=True,
+        )
+        fee = FeeField(load_default=None, validate=validate.Range(min=ZERO, min_inclusive=False))
+        location = LocationField(required=True)
+        unique_id = fields.String(required=False, load_default=None)
+        event_identifier = fields.String(required=False, load_default=None)
```
LefterisJP, line 961: shouldn't you also have an optional address and transaction hash?

rotkehlchen/db/history_events.py
LefterisJP: this comment is addressed I guess?

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -228,6 +228,7 @@ def create_asset_movement_with_fee(
         fee_asset: Asset,
         fee: 'FVal',
         unique_id: str | None = None,
+        identifier: int | None = None,
```
LefterisJP, line 231: Where do you need this?

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -240,10 +241,13 @@ def create_asset_movement_with_fee(
         asset=asset,
         balance=Balance(amount),
         unique_id=unique_id,
+        identifier=identifier,
         extra_data=extra_data,
     )]
     if fee != ZERO:
         events.append(AssetMovement(
+            # we can increment here since it's added after the main event.
+            identifier=identifier + 1 if identifier is not None else None,
```
LefterisJP, line 250: This is very hacky. If at any point for whatever reason we somehow context switch between the creation of the main asset movement and the asset movement fee they would end up getting non-consecutive identifiers. Making this assumptions is dangerous.

rotkehlchen/tests/api/test_history_base_entry.py
LefterisJP: +1. Do it.

rotkehlchen/tests/api/test_history_base_entry.py
LefterisJP: +1. Do it.

LefterisJP (commented):
Left some comments.

Also please do not leave comments from previews reviews hanging. The following comments from @yabirgb have not been answered or resolved. Just left there ignored. Do not do that.

https://github.com/rotki/rotki/pull/9078/files#r1884118528
https://github.com/rotki/rotki/pull/9078/files#r1887511341
https://github.com/rotki/rotki/pull/9078/files#r1887037358
https://github.com/rotki/rotki/pull/9078/files#r1887507567

rotkehlchen/api/rest.py
prettyirrelevant: yeah, so the context manager rolls back the transaction.

rotkehlchen/history/events/structures/asset_movement.py
prettyirrelevant: this is used when creating the `AssetMovement` instances inside the api schema

rotkehlchen/api/v1/schemas.py
```diff
@@ -945,12 +946,37 @@ def make_history_base_entry(
         ) -> dict[str, Any]:
             return {'event': EthWithdrawalEvent(**data)}
 
+    class CreateAssetMovementEventSchema(BaseSchema):
+        event_type = SerializableEnumField(
```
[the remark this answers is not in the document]
prettyirrelevant: nothing more is needed. this handles addition & editing. deleting is simply just passing the identifiers.

rotkehlchen/history/events/structures/asset_movement.py
prettyirrelevant: so i went with the approach of passing the `database` to the schema and querying the `identifier` from the db instead of making such an assumption.

PR #9082, opened 2024-12-13
New AssetMovement in independentreserve kraken and kucoin

nicholasyoder:
Related #7007 

Note that the independentreserve exchange doesn't have any tests for asset movements. The changes there were relatively simple, so I expect it should be fine. There is already a todo comment about adding a test as well: https://github.com/rotki/rotki/blob/42767ed33f64c8af07eeaa5a73afe7b095355233/rotkehlchen/tests/exchanges/test_independentreserve.py#L226

Also, I removed the `test_query_asset_movements` test. That api endpoint is going away, so I think most of what it was testing is being removed. It was using kraken, which is why I'm removing it in this PR.

### DIFF
--- a/rotkehlchen/exchanges/independentreserve.py
+++ b/rotkehlchen/exchanges/independentreserve.py
@@ -4,6 +4,7 @@ import json
 import logging
 import time
 from collections import OrderedDict
+from collections.abc import Sequence
 from json.decoder import JSONDecodeError
 from typing import TYPE_CHECKING, Any, Literal
 
@@ -49,38 +50,35 @@ from rotkehlchen.constants.assets import (
     A_YFI,
     A_ZRX,
 )
+from rotkehlchen.data_import.utils import maybe_set_transaction_extra_data
 from rotkehlchen.db.settings import CachedSettings
 from rotkehlchen.errors.asset import UnknownAsset
 from rotkehlchen.errors.misc import RemoteError
 from rotkehlchen.errors.serialization import DeserializationError
-from rotkehlchen.exchanges.data_structures import (
-    AssetMovement,
-    Location,
-    MarginPosition,
-    Price,
-    Trade,
-)
+from rotkehlchen.exchanges.data_structures import Location, MarginPosition, Price, Trade
 from rotkehlchen.exchanges.exchange import ExchangeInterface, ExchangeQueryBalances
 from rotkehlchen.fval import FVal
+from rotkehlchen.history.events.structures.asset_movement import AssetMovement
+from rotkehlchen.history.events.structures.base import HistoryBaseEntry
+from rotkehlchen.history.events.structures.types import HistoryEventType
 from rotkehlchen.inquirer import Inquirer
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.serialization.deserialize import (
     deserialize_asset_amount,
-    deserialize_asset_movement_category,
+    deserialize_asset_movement_event_type,
     deserialize_timestamp_from_date,
 )
 from rotkehlchen.types import (
     ApiKey,
     ApiSecret,
     AssetAmount,
-    AssetMovementCategory,
     ExchangeAuthCredentials,
     Fee,
     Timestamp,
     TradeType,
 )
 from rotkehlchen.user_messages import MessagesAggregator
-from rotkehlchen.utils.misc import timestamp_to_iso8601
+from rotkehlchen.utils.misc import timestamp_to_iso8601, ts_sec_to_ms
 
 if TYPE_CHECKING:
     from rotkehlchen.db.dbhandler import DBHandler
@@ -184,7 +182,7 @@ def _asset_movement_from_independentreserve(raw_tx: dict) -> AssetMovement | Non
     - KeyError
     """
     log.debug(f'Processing raw IndependentReserve transaction: {raw_tx}')
-    movement_type = deserialize_asset_movement_category(raw_tx['Type'])
+    movement_type = deserialize_asset_movement_event_type(raw_tx['Type'])
     asset = independentreserve_asset(raw_tx['CurrencyCode'])
     bitcoin_tx_id = raw_tx.get('BitcoinTransactionId')
     eth_tx_id = raw_tx.get('EthereumTransactionId')
@@ -195,18 +193,12 @@ def _asset_movement_from_independentreserve(raw_tx: dict) -> AssetMovement | Non
     else:
         transaction_id = None
 
-    timestamp = deserialize_timestamp_from_date(
-        date=raw_tx['CreatedTimestampUtc'],
-        formatstr='iso8601',
-        location='IndependentReserve',
-    )
-
     comment = raw_tx.get('Comment')
     address = None
     if comment is not None and comment.startswith('Withdrawing to'):
         address = comment.rsplit()[-1]
 
-    raw_amount = raw_tx.get('Credit') if movement_type == AssetMovementCategory.DEPOSIT else raw_tx.get('Debit')  # noqa: E501
+    raw_amount = raw_tx.get('Credit') if movement_type == HistoryEventType.DEPOSIT else raw_tx.get('Debit')  # noqa: E501
 
     if raw_amount is None:  # skip
         return None   # Can end up being None for some things like this: 'Comment': 'Initial balance after Bitcoin fork'  # noqa: E501
@@ -214,15 +206,19 @@ def _asset_movement_from_independentreserve(raw_tx: dict) -> AssetMovement | Non
 
     return AssetMovement(
         location=Location.INDEPENDENTRESERVE,
-        category=movement_type,
-        address=address,
-        transaction_id=transaction_id,
-        timestamp=timestamp,
+        event_type=movement_type,
+        timestamp=ts_sec_to_ms(deserialize_timestamp_from_date(
+            date=raw_tx['CreatedTimestampUtc'],
+            formatstr='iso8601',
+            location='IndependentReserve',
+        )),
         asset=asset,
-        amount=amount,
-        fee_asset=asset,  # whatever -- no fee
-        fee=Fee(ZERO),  # we can't get fee from this exchange
-        link=raw_tx['CreatedTimestampUtc'] + str(amount) + str(movement_type) + asset.identifier,
+        balance=Balance(amount),
+        unique_id=raw_tx.get('TransactionGuid'),
+        extra_data=maybe_set_transaction_extra_data(
+            address=address,
+            transaction_id=transaction_id,
+        ),
     )
 
 
@@ -460,11 +456,11 @@ class Independentreserve(ExchangeInterface):
 
         return trades, (start_ts, end_ts)
 
-    def query_online_deposits_withdrawals(
+    def query_online_history_events(
             self,
             start_ts: Timestamp,  # pylint: disable=unused-argument
             end_ts: Timestamp,
-    ) -> list[AssetMovement]:
+    ) -> Sequence[HistoryBaseEntry]:
         if self.account_guids is None:
             self.query_balances()  # do a balance query to populate the account guids
         movements = []
--- a/rotkehlchen/exchanges/kraken.py
+++ b/rotkehlchen/exchanges/kraken.py
@@ -35,11 +35,15 @@ from rotkehlchen.db.settings import CachedSettings
 from rotkehlchen.errors.asset import UnknownAsset
 from rotkehlchen.errors.misc import InputError, RemoteError
 from rotkehlchen.errors.serialization import DeserializationError
-from rotkehlchen.exchanges.data_structures import AssetMovement, MarginPosition, Trade
+from rotkehlchen.exchanges.data_structures import MarginPosition, Trade
 from rotkehlchen.exchanges.exchange import (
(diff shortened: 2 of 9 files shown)


rotkehlchen/exchanges/independentreserve.py
```diff
@@ -206,23 +204,23 @@ def _asset_movement_from_independentreserve(raw_tx: dict) -> AssetMovement | Non
...
     amount = deserialize_asset_amount(raw_amount)
 
     return AssetMovement(
         location=Location.INDEPENDENTRESERVE,
-        category=movement_type,
-        address=address,
-        transaction_id=transaction_id,
-        timestamp=timestamp,
+        event_type=movement_type,
+        timestamp=ts_sec_to_ms(timestamp),
```
prettyirrelevant, line 216: you can just do `ts_sec_to_ms(deserialize_timestamp_from_date(
        date=raw_tx['CreatedTimestampUtc'],
        formatstr='iso8601',
        location='IndependentReserve',
    ))`

rotkehlchen/exchanges/kucoin.py
```diff
@@ -523,47 +533,49 @@ def _deserialize_accounts_balances(
...
-            category=category,
-            address=address,
-            transaction_id=transaction_id,
+            event_type=event_type,
             asset=fee_asset,
             amount=amount,
             fee_asset=fee_asset,
             fee=fee,
-            link=link_id,
+            unique_id=unique_id if unique_id is not None else transaction_id,
```
prettyirrelevant, line 574:
```suggestion
            unique_id=unique_id or transaction_id,
```

rotkehlchen/exchanges/kucoin.py
```diff
@@ -423,7 +430,10 @@ def _api_query_paginated(
                     )
                     continue
 
-                results.append(result)  # type: ignore
+                if isinstance(result, list):
+                    results.extend(result)
+                else:
+                    results.append(result)
```
prettyirrelevant, line 436: why is this needed? i'm guessing because `deserialize_asset_movement()` returns a list now?

rotkehlchen/exchanges/kraken.py
```diff
@@ -582,90 +582,6 @@ def _query_endpoint_for_period(
             request.update(extra_dict)
         return self.api_query(endpoint, request)
 
-    def query_online_deposits_withdrawals(
-            self,
-            start_ts: Timestamp,
-            end_ts: Timestamp,
-    ) -> list[AssetMovement]:
-        with self.db.conn.read_ctx() as cursor:
-            self.query_kraken_ledgers(cursor, start_ts=start_ts, end_ts=end_ts)
```
prettyirrelevant, line 591: what now happens to this query?

does it now happen in `query_online_margin_history()`?

prettyirrelevant (commented):
some comments

rotkehlchen/exchanges/kucoin.py
nicholasyoder: Yes, its due to asset movements now being in a list here. I expect eventually when we convert trades to being based on history events as well they will also need to be in a list due to having fees as separate events and then this will be changed to just `results.extend(result)`

rotkehlchen/exchanges/kraken.py
nicholasyoder: > does it now happen in query_online_margin_history()?

No, it happens in `query_history_events()`

The flow here is a bit confusing. `query_kraken_ledgers` is called from several different places, and simply loads all events from the kraken api. These events were then being reprocessed later to convert them to AssetMovements, Trades, etc. But now AssetMovements, since they are actually history events, are simply created  in `query_kraken_ledgers` the first time around, and there's no need for any reprocessing.

So where before it was doing something like this to get asset movements into the database: 
`query_deposits_withdrawals` -> `query_online_deposits_withdrawals` -> `query_kraken_ledgers` -> `reprocessing logic in query_online_deposits_withdrawals`

its now simply doing this: 
`query_history_events` -> `query_kraken_ledgers`

rotkehlchen/exchanges/kraken.py
```diff
@@ -1146,7 +1062,17 @@ def history_event_from_kraken(
 
                 # Make sure to not generate an event for KFEES that is not of type FEE
                 if asset != A_KFEE:
-                    event = HistoryEvent(
+                    event = AssetMovement(
+                        timestamp=timestamp,
+                        location=Location.KRAKEN,
+                        event_type=event_type,  # type: ignore  # will be deposit or withdrawal
```
LefterisJP, line 1068: Hmmm I think that using the `x = y if z else k` ternary operator usage here is confusing mypy and probably also the reader. Is it enforced by linting? Or can't you just do

```python
if event_type in {HistoryEventType.DEPOSIT,  HistoryEventType.WITHDRAWAL}:
    event = AssetMovement()
else:
    event = historyEvent()
```

rotkehlchen/exchanges/kraken.py
```diff
@@ -1162,11 +1088,22 @@ def history_event_from_kraken(
...
+                    group_events.append((idx, AssetMovement(
+                        timestamp=timestamp,
+                        location=Location.KRAKEN,
+                        event_type=event_type,  # type: ignore  # will be deposit or withdrawal
+                        asset=asset,
+                        balance=Balance(abs(fee_amount)),
+                        unique_id=identifier,
+                        is_fee=True,
+                    ) if event_type in {
+                        HistoryEventType.DEPOSIT,
```
LefterisJP, line 1104: you are doing same thing as above here. I think this function is getting too complicated and needs to be rethought. Lots of code is repeating

rotkehlchen/exchanges/kucoin.py
LefterisJP: can you then add a comment with a TODO here for this explaining it?

rotkehlchen/exchanges/kucoin.py
```diff
@@ -523,47 +533,49 @@ def _deserialize_accounts_balances(
...
-            category=category,
-            address=address,
-            transaction_id=transaction_id,
+            event_type=event_type,
             asset=fee_asset,
             amount=amount,
             fee_asset=fee_asset,
             fee=fee,
-            link=link_id,
+            unique_id=unique_id or transaction_id,
```
LefterisJP, line 574: unique_id as per its name should have the  guarantee of uniqueness for the given exchange. Transaction id should probably also have the same uniqueness guarantee but are you sure?

rotkehlchen/tests/api/test_exchanges.py
```diff
@@ -736,212 +733,6 @@ def assert_okay(response: 'Response') -> None:
     assert len(result['entries']) == 0
 
 
-@pytest.mark.parametrize('number_of_eth_accounts', [0])
-@pytest.mark.parametrize('added_exchanges', [(Location.KRAKEN, Location.POLONIEX)])
-@pytest.mark.parametrize('start_with_valid_premium', [False, True])
-def test_query_asset_movements(
```
LefterisJP, line 742: There is no other test to write instead of this? Just removing it as the endpoint no longer exists right?

rotkehlchen/tests/exchanges/test_kucoin.py
```diff
@@ -355,18 +357,26 @@ def test_deserialize_asset_movement_deposit(mock_kucoin):
...
-        address='0x5bedb060b8eb8d823e2414d82acce78d38be7fe9',
-        transaction_id='3e2414d82acce78d38be7fe9',
+        event_type=HistoryEventType.DEPOSIT,
         asset=A_ETH,
-        amount=AssetAmount(FVal('1')),
-        fee_asset=A_ETH,
-        fee=Fee(FVal('0.01')),
-        link='',
-    )
+        balance=Balance(FVal('1')),
```
LefterisJP, line 365:
```suggestion
        balance=Balance(ONE),
```

rotkehlchen/exchanges/kraken.py
yabirgb: Also I believe you are not converting them properly. In line 1045 we have

```python
                elif event_type == HistoryEventType.DEPOSIT:
                    event_subtype = HistoryEventSubType.DEPOSIT_ASSET
                elif event_type == HistoryEventType.WITHDRAWAL:
                    event_subtype = HistoryEventSubType.REMOVE_ASSET
``` 

I believe you should catch those two and convert them to assetMovements instead of history events

rotkehlchen/tests/exchanges/test_kucoin.py
```diff
@@ -389,18 +399,26 @@ def test_deserialize_asset_movement_withdrawal(mock_kucoin):
...
-        address='0x5bedb060b8eb8d823e2414d82acce78d38be7fe9',
-        transaction_id='3e2414d82acce78d38be7fe9',
+        event_type=HistoryEventType.WITHDRAWAL,
         asset=A_ETH,
-        amount=AssetAmount(FVal('1')),
-        fee_asset=A_ETH,
-        fee=Fee(FVal('0.01')),
-        link='5c2dc64e03aa675aa263f1ac',
-    )
+        balance=Balance(FVal('1')),
```
yabirgb, line 407:
```suggestion
        balance=Balance(ONE),
```

rotkehlchen/tests/exchanges/test_kucoin.py
```diff
@@ -687,44 +705,64 @@ def test_query_asset_movements_sandbox(
...
-            fee=Fee(FVal('0.25')),
-            link='5c2dc64e03aa675aa263f1a4',
-        ),
-    ]
+    expected_asset_movements = [AssetMovement(
+        location=Location.KUCOIN,
+        event_type=HistoryEventType.DEPOSIT,
+        timestamp=TimestampMS(1612556652000),
+        asset=A_KCS,
+        balance=Balance(FVal('1')),
```
yabirgb, line 713:
```suggestion
        balance=Balance(ONE),
```

rotkehlchen/exchanges/kraken.py
nicholasyoder: Changed it to use a normal `if` like this, but had to rename `event` to `history_event` several places so that mypy wouldn't think that `event` always had to be a HistoryEvent when that variable name is reused later in the function. Also simplified things further by using the `create_asset_movement_with_fee` function.

rotkehlchen/exchanges/kucoin.py
nicholasyoder: As far as I can tell from the docs (https://www.kucoin.com/docs/rest/funding/deposit/get-deposit-list), the value in `transaction_id` (from `walletTxId`) will be an onchain tx hash, which I expect should always be unique.

rotkehlchen/tests/api/test_exchanges.py
nicholasyoder: Yes, removing this since the endpoint is being removed. The stuff like order by, time range, pagination, etc that was being tested here should already be in place and properly tested for history events, so I don't think any new tests need made for the new history event based asset movements.

rotkehlchen/exchanges/kraken.py
```diff
@@ -1162,11 +1088,22 @@ def history_event_from_kraken(
...
-                        receive_spend_events[event.event_identifier].append((idx, event))
+                        receive_spend_events[event.event_identifier].append((idx, event))  # type: ignore  # will only be HistoryEvent for receive/spend
                     else:
                         group_events.append((idx, event))
                 if event_type != HistoryEventType.INFORMATIONAL and fee_amount != ZERO:  # avoid processing ignored events with fees that were converted to informational  # noqa: E501
-                    group_events.append((idx, HistoryEvent(
+                    group_events.append((idx, AssetMovement(
+                        timestamp=timestamp,
+                        location=Location.KRAKEN,
+                        event_type=event_type,  # type: ignore  # will be deposit or withdrawal
```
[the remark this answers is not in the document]
nicholasyoder: As mentioned in the other comments this has been simplified by using `create_asset_movement_with_fee`.

rotkehlchen/exchanges/kraken.py
```diff
@@ -1146,7 +1061,23 @@ def history_event_from_kraken(
 
                 # Make sure to not generate an event for KFEES that is not of type FEE
                 if asset != A_KFEE:
-                    event = HistoryEvent(
+                    # Process asset movements
+                    if event_type in {HistoryEventType.DEPOSIT, HistoryEventType.WITHDRAWAL}:
```
yabirgb, line 1065: Add to the comment that there aren't any deposit/withdrawal events of KFEEs. I've checked the tests that we have and seems to be the case. It it wasn't like this this logic would need adjusment so it is one of the assumptions

rotkehlchen/exchanges/kraken.py
```diff
@@ -1146,7 +1061,23 @@ def history_event_from_kraken(
...
+                        group_events.extend(
+                            (idx, event) for event in create_asset_movement_with_fee(
+                                timestamp=timestamp,
+                                location=Location.KRAKEN,
+                                event_type=event_type,  # type: ignore  # will be deposit or withdrawal
```
yabirgb, line 1070: I've checked and type casting seems to be the only way to avoid this so the ignore is okey

LefterisJP (approved):
LGTM

PR #9084, opened 2024-12-13
New AssetMovement in okx poloniex and woo

nicholasyoder:
Related #7007 

Also removed `test_query_asset_movements_over_limit` which was another test for the old api endpoint similar to the one removed in https://github.com/rotki/rotki/pull/9082. Seems to be focused on testing that the FREE_ASSET_MOVEMENTS_LIMIT is being respected, so should be fine to just remove it.


LefterisJP (approved):
LGTM

PR #9092, opened 2024-12-16
add direction to history export

prettyirrelevant:


rotkehlchen/history/events/structures/base.py
```diff
@@ -311,9 +311,11 @@ def serialize_for_csv(self, fiat_value: FVal) -> dict[str, Any]:
         right after the 'asset' in the serialized dictionary. Note that
         'fiat_value' is not in USD but in the user-selected currency.
         """
+        new_dict: dict[str, Any] = {
+            'direction': str(EVENT_CATEGORY_MAPPINGS[self.event_type][self.event_subtype]['default'].direction),  # noqa: E501
```
LefterisJP, line 315: What are you doing here? There is a function for this and that function also handles the case where type/subtype is invalid. Please don't copy paste stuff. Just use the function: `maybe_get_direction`

LefterisJP (approved):
LGTM

PR #2179, opened 2021-01-25
Handle rare cases of bittrex deposit/withdrawals with an integer as transaction id

LefterisJP:
Fix #2175 

[Bittrex's api](https://bittrex.github.io/api/v3#definition-Order) states that a deposit/withdrawal's transaction id should always be a string.

But as seen by a user in #2175 that can in some rare cases be an integer which goes against their docs.

This PR handles that.

PR #9094, opened 2024-12-17
Remove old asset movement stuff

nicholasyoder:
Related #7007 

Mostly removes code associated with the old asset movements.

A couple things to note:

`test_int_overflow_at_tuple_insertion` was testing that `OverflowError` from writing really long integers via `write_tuples` is handled correctly here: https://github.com/rotki/rotki/blob/develop/rotkehlchen/db/dbhandler.py#L1956. Unfortunately it was using asset movement for this, so I converted `test_int_overflow_at_tuple_insertion` to use a MarginPosition instead, so that it is still testing the `write_tuples` function. But since the original issue when this was added was apparently related to asset movements (https://github.com/rotki/rotki/pull/2179) I also added a `test_int_overflow_at_history_event_insertion` test for adding a new history event based asset movement - and it was not being handled correctly there. So I also added a try except for `OverflowError` in `add_history_event`.

Also the `test_assets_movements_not_accounted_for` test has been removed since the new AssetMovements are not using the `account_for_assets_movements` setting. I'll remove this setting in another PR as discussed in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1318589282050379828).


rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2693,9 +2693,10 @@ def test_upgrade_db_45_to_46(user_data_dir: 'Path', messages_aggregator):
             (6, '20522c693bcda4ef646682c6a58bb0349b01f4d7b9168a62ce94b2c8dd1fe639', 1, 1644244023000, 'B', None, 'ATOM', '0.10000000', '0', 'Pay 0.10000000 ATOM as Kraken withdrawal fee', 'withdrawal', 'fee', None),  # noqa: E501
             (6, '1cf3fc675d4835efe532d18730db35163b07fb5f44956b888d7aaf852b04fd1c', 0, 1727542684000, 'G', None, 'ETH', '0.07126231', '0', 'Deposit 0.07126231 ETH to Coinbase', 'deposit', 'deposit asset', '{"address": "0xc37b40ABdB939635068d3c5f13E7faF686F03B65", "transaction_id": "0xc8aa60c9f8a93692b66fbc3f57c64f2b1c4afd92f370e12dc7bb23acfa303dbb"}'),  # noqa: E501
         ]
-        # TODO: Remove this when table is removed
-        # assert table_exists(cursor, 'asset_movements') is False  # noqa: ERA001
-        # assert table_exists(cursor, 'asset_movement_category') is False  # noqa: ERA001
+
+        assert table_exists(cursor, 'asset_movements') is False
+        assert table_exists(cursor, 'asset_movement_category') is False
```
yabirgb, line 2698: there is a test below this one that checks the removed tables. You will need to update it too

rotkehlchen/db/dbhandler.py
```diff
@@ -1742,7 +1738,7 @@ def edit_exchange(
                 raise InputError(f'Could not update DB user_credentials_mappings due to {e!s}') from e  # noqa: E501
 
         if new_name is not None:
-            exchange_re = re.compile(r'(.*?)_(trades|margins|asset_movements).*')
+            exchange_re = re.compile(r'(.*?)_(trades|margins|history_events).*')
```
LefterisJP, line 1741: hmmm this was also not happening before which was a bug right?

rotkehlchen/db/history_events.py
```diff
@@ -117,20 +117,31 @@ def add_history_event(
...
-                'VALUES(?, ?, ?)',
-                [(identifier, k, v) for k, v in mapping_values.items()],
+            if mapping_values is not None:
+                write_cursor.executemany(
+                    'INSERT OR IGNORE INTO history_events_mappings(parent_identifier, name, value) '  # noqa: E501
+                    'VALUES(?, ?, ?)',
+                    [(identifier, k, v) for k, v in mapping_values.items()],
+                )
+        except OverflowError:
+            self.db.msg_aggregator.add_error(
```
LefterisJP, line 138: I don't think this is a good idea.

Because before the error was bubbling up and at the destruction of the write context also reverts the DB changes. The way you do this won't revert them. But just log an error.

rotkehlchen/db/dbhandler.py
nicholasyoder: Yes, I think this must have been a bug. Although not many exchanges were actually using history_events ranges, so it probably didn't affect much.

yabirgb (approved):
lgtm

LefterisJP (approved):
LGTM

PR #9097, opened 2024-12-17
Add more properties to the endpoint for asset movements

yabirgb:
- The endpoints accepts now a location label. So the user can add names of the exchanges like `Kraken 1`
- When deleting an asset movement the fee gets deleted
- Editing an asset movement to add/remove the fee information will delete the second event of the group


rotkehlchen/api/rest.py
```diff
@@ -1108,11 +1109,23 @@ def add_history_events(self, events: list['HistoryBaseEntry']) -> Response:
         return api_response(result_dict, status_code=HTTPStatus.OK)
 
     def edit_history_events(self, events: list['HistoryBaseEntry']) -> Response:
-        db = DBHistoryEvents(self.rotkehlchen.data.db)
-        try:
+        events_db = DBHistoryEvents(self.rotkehlchen.data.db)
+        if events[0].entry_type == HistoryBaseEntryType.ASSET_MOVEMENT_EVENT:
+            try:
```
LefterisJP, line 1114: EDIT: Probably should just be removed as per what I wrote in the other part.

Two things.

1. Explain why you do this here.
2. Why not just move this to `edit_history_event`? To save conditionals?
```suggestion
            try:  # Explain here
```

rotkehlchen/db/history_events.py
```diff
@@ -198,9 +198,23 @@ def delete_history_events_by_identifier(
 
         If any identifier is missing the entire call fails and an error message
         is returned. Otherwise None is returned.
+
+        For asset movements the fee and the main event are linked. Deleting one of them
```
LefterisJP, line 202: Why do you want to add this logic? I would say it's a can of worms. Don't do this. As you can see adds weird conditionals everywhere and does not scale.

Do we delete the entire eth staking event if wejust delete the mev event?
Do we delete the entire swap if we just remove the send?
Do we delete the entire  safe transaction if we just delete the info event of safe execution completion?

etc.

I guess you get the problem. This approach does not scale. Handle this in the frontend, or in the api conditional where  asset movements are deleted, to alway delete both.

Please undo this

rotkehlchen/db/history_events.py
```diff
@@ -873,3 +889,66 @@ def query_wrap_stats(self, from_ts: Timestamp, to_ts: Timestamp) -> dict[str, An
...
         }
+
+    def count_number_of_events_by_event_identifier(self, event_identifier: str) -> int:
+        with self.db.conn.read_ctx() as cursor:
+            return cursor.execute(
+                'SELECT COUNT(*) FROM history_events WHERE event_identifier=?',
+                (event_identifier,),
+            ).fetchone()[0]
+
+    def edit_asset_movements(
```
LefterisJP, line 900: yeah sorry no undo all of this.

rotkehlchen/api/rest.py
yabirgb: As discussed in discord moved this to `rest_helpers/history_events.py`

LefterisJP (approved):
LGTM

PR #9112, opened 2024-12-19
Fix special case in balancer with reth

yabirgb:
Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/chain/evm/decoding/balancer/v2/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/balancer/v2/decoder.py
@@ -23,6 +23,7 @@ from rotkehlchen.chain.evm.decoding.structures import (
     TransferEnrichmentOutput,
 )
 from rotkehlchen.chain.evm.decoding.types import CounterpartyDetails
+from rotkehlchen.chain.evm.decoding.utils import maybe_reshuffle_events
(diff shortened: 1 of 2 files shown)


rotkehlchen/chain/evm/decoding/balancer/v2/decoder.py
```diff
@@ -105,7 +110,16 @@ def _decode_join_or_exit(self, context: DecoderContext) -> DecodingOutput:
...
+                send_events.append(event)
+
+        # in _check_deposits_withdrawals we expect the receive to be the last event.
+        # This happens almost always but there are some cases like rETH on arb where it doesn't.
+        # This reshuffle ensures that the receive event is always the last one before grouping
+        # them in _check_deposits_withdrawals.
+        maybe_reshuffle_events(
+            ordered_events=send_events + receive_events,
+            events_list=context.decoded_events,
+        )
```
prettyirrelevant, line 122: i was thinking of refactoring the `self._check_deposits_withdrawals()` function but this is simpler.

yabirgb: That functions calculates the number of events per balancer pool in the case of multiple deposits. SInce the logic is shared with v1 and I'm not sure if this happens in v1 I believe is just easier to have this split

LefterisJP (approved):
LGTM.

PR #9113, opened 2024-12-19
Properly set the receive address of gitcoin v2 donation events

LefterisJP:
For receive events we tend to add the receiver as the "address" field. This was omitted in gitcoin v2 and made it hard to do analysis on events.

In my analysis scripts I am forced to do this now:

```python
def get_from_address(event) -> ChecksumEvmAddress:
    """Ideally from_address in a receive should always be the "address" field, but for some gitcoin
    donations it's not set properly so let's try to get it from the notes"""
    from_address = event.address
    if ' from ' in event.notes and event.event_subtype == HistoryEventSubType.DONATE and (maybe_address := event.notes[-42:]).startswith('0x'):
        from_address = maybe_address

    return from_address
```


yabirgb (approved):
lgtm

PR #9116, opened 2024-12-19
Add evm event redecode trigger to db upgrade

nicholasyoder:


rotkehlchen/db/upgrades/v45_v46.py
```diff
@@ -68,11 +69,7 @@ def _refresh_icons(write_cursor: 'DBCursor') -> None:
     @progress_step(description='Moving EVM event extra data to the history_events table.')
     def _move_extra_data(write_cursor: 'DBCursor') -> None:
         write_cursor.execute('ALTER TABLE history_events ADD COLUMN extra_data TEXT;')
-        write_cursor.execute(
-            'UPDATE history_events SET extra_data = '
-            '(SELECT extra_data FROM evm_events_info '
-            'WHERE evm_events_info.identifier = history_events.identifier);',
-        )
+        # no need to migrate the data since evm events are to be redecoded after this upgrade.
```
LefterisJP, line 72: I see why you would think this, but no you can't do this. Not all events are redecoded. All customized events are kept as they are. So at least those would need to be migrated. So you lose nothing by leaving this in as is, and removing it introduces bugs.

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2627,12 +2627,6 @@ def test_upgrade_db_45_to_46(user_data_dir: 'Path', messages_aggregator):
         assert cursor.execute(
             "SELECT COUNT(*) FROM pragma_table_info('evm_events_info') WHERE name='extra_data'",
         ).fetchone()[0] == 1
-        events_with_extra_data = [row[0] for row in cursor.execute(
```
LefterisJP, line 2630: Keep this one in actually. And turn the event to a custom one. To see that the logic works.

nicholasyoder: still can't check the `events_with_extra_data` since that looks at all the existing evm events with extra data. But now it converts the event its checking with `existing_evm_event` and `existing_evm_event_extra_data` to a custom event, so it can still check that one.

PR #9129, opened 2024-12-20
Fix ENS registration with refund decoding

nicholasyoder:
Fixes problem with decoding of ENS registrations that include a refund on the new ENS registrar contract.
Also fixes a small bug where it was displaying a blank name instead of falling back to `an ENS name`.


LefterisJP (approved):
LGTM

PR #9134, opened 2024-12-23
Fix decoding of 0x swap with amount returned

### LINKED ISSUE #9122: Optimism 0x swap with amount returned not seen correctly
## Problem Definition

There is an extra receival of OP, perhaps some kind of return of remaining from the swap?

https://optimistic.etherscan.io/tx/0x0745615163e01c8a446f2520d5fa008dd69f308f24bcc3d2fec2466c1a2bc25c

Reminder to self: Fix/adjust the customized event once it's fixed in production rotki.

## TODO

Fix it, use the transaction hash as a test.

The fix would be to make it so that if there is a return, disregard the return event and reduce the original amount by the returned amount. Looking like this basically

The reason for this is for easier accounting and cost basic.

nicholasyoder:
Closes #9122


LefterisJP (approved):
LGTM but should have a changelog

PR #9136, opened 2024-12-23
fix balancer pool token price inquiry

prettyirrelevant:


LefterisJP (approved):
Can't test this, but if this fixes the issue then should be fine. You also wrote a test so this is reassuring as a fix.

Will confirm when testing too.

PR #9140, opened 2024-12-23
Claiming morpho from multiple distributors should work now

LefterisJP:
In the future we can change from hardcoding distributors to asking the api about them: https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=91840111

### DIFF
--- a/rotkehlchen/chain/evm/decoding/morpho/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/morpho/decoder.py
@@ -50,7 +50,7 @@ class MorphoCommonDecoder(DecoderInterface, ReloadableDecoderMixin):
             base_tools: 'BaseDecoderTools',
             msg_aggregator: 'MessagesAggregator',
             bundlers: set['ChecksumEvmAddress'],
-            rewards_distributor: 'ChecksumEvmAddress',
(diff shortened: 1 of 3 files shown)


rotkehlchen/chain/evm/decoding/morpho/decoder.py
```diff
@@ -258,14 +258,15 @@ def _decode_reward_claim(self, context: DecoderContext) -> DecodingOutput:
...
             if (
-                event.event_type == HistoryEventType.RECEIVE and
-                event.event_subtype == HistoryEventSubType.NONE and
-                event.address == self.rewards_distributor and
-                event.asset == self.base.get_or_create_evm_asset(
-                    address=bytes_to_address(context.tx_log.topics[2]),
-                )
+                    event.event_type == HistoryEventType.RECEIVE and
+                    event.event_subtype == HistoryEventSubType.NONE and
+                    event.address == context.tx_log.address and
```
yabirgb, line 268: Before we had `event.address == self.rewards_distributor`. Was it wrong?

LefterisJP: now there is multiple distributors. And this event is always fired by the distributor.

PR #9158, opened 2024-12-26
Add aave lido tokens

yabirgb:
This PR:

- Tracks lido and etherFI aave reserve tokens
- Improves performance of task for querying aave tokens

The performance has been improved mostly by reducing the number of remote calls but also by reducing the number of function calls.

The worst case that we had before is when we had to add new reserve tokens. This would mean that for each token we will make a multicall to retrieve name, symbol and decimals in one query. We had `num_aave_providers * num_reservers * num_debt_tokens` queries. In the case of ethereum it meant to do around 100 rpc calls.

What I've done is to check what tokens need to have their details queried and then do bigger multicalls since I query name, symbol and decimals of all of them in 3 multicalls (instead of num_debt_tokens). This allowed me to avoid recreating classes from web3.py like the contract class that happened every time we had to query details and encode/decode serveral times the same information.

Now for the ethereum data we do ~30 rpc calls, ~1/3 of what we were doing before. Also the cassette file went from 15K lines to 5K lines.

### DIFF
--- a/rotkehlchen/chain/gnosis/modules/aave/v3/constants.py
+++ b/rotkehlchen/chain/gnosis/modules/aave/v3/constants.py
@@ -2,4 +2,4 @@ from typing import Final
 
 from rotkehlchen.chain.evm.types import string_to_evm_address
 
-AAVE_V3_DATA_PROVIDER: Final = string_to_evm_address('0x501B4c19dd9C2e06E94dA7b6D5Ed4ddA013EC741')
+AAVE_V3_DATA_PROVIDER: Final = string_to_evm_address('0x57038C3e3Fe0a170BB72DE2fD56E98e4d1a69717')
--- a/rotkehlchen/tasks/assets.py
+++ b/rotkehlchen/tasks/assets.py
@@ -14,9 +14,13 @@ from rotkehlchen.chain.base.modules.aave.v3.constants import (
 )
 from rotkehlchen.chain.ethereum.modules.aave.v3.constants import (
     AAVE_V3_DATA_PROVIDER as AAVE_V3_DATA_PROVIDER_ETH,
+    AAVE_V3_DATA_PROVIDER_OLD as AAVE_V3_DATA_PROVIDER_ETH_OLD,
+    ETHERFI_AAVE_V3_DATA_PROVIDER as ETHERFI_AAVE_V3_DATA_PROVIDER_ETH,
+    LIDO_AAVE_V3_DATA_PROVIDER as LIDO_AAVE_V3_DATA_PROVIDER_ETH,
 )
 from rotkehlchen.chain.ethereum.utils import MULTICALL_CHUNKS
-from rotkehlchen.chain.evm.constants import EVM_ADDRESS_REGEX
+from rotkehlchen.chain.evm.constants import EVM_ADDRESS_REGEX, ZERO_ADDRESS
+from rotkehlchen.chain.evm.contracts import EvmContract
 from rotkehlchen.chain.evm.decoding.aave.constants import CPT_AAVE_V3
 from rotkehlchen.chain.evm.decoding.aave.v3.constants import (
     AAVE_V3_DATA_PROVIDER as AAVE_V3_DATA_PROVIDER_EVM,
@@ -70,6 +74,8 @@ from rotkehlchen.utils.misc import ts_now, ts_sec_to_ms
 
 if TYPE_CHECKING:
     from rotkehlchen.chain.aggregator import ChainsAggregator
+    from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
+    from rotkehlchen.types import ChecksumEvmAddress
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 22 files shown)


rotkehlchen/tasks/assets.py
```diff
@@ -196,10 +202,88 @@ def update_owned_assets(user_db: DBHandler) -> None:
         )
 
 
+def _aave_tokens_are_correct(
```
LefterisJP, line 205:
```suggestion
def _find_missing_tokens(
```

rotkehlchen/tasks/assets.py
```diff
@@ -196,10 +202,88 @@ def update_owned_assets(user_db: DBHandler) -> None:
...
+        token_to_underlying: dict['ChecksumEvmAddress', 'ChecksumEvmAddress'],
+        chain_id: ChainID,
+) -> list['ChecksumEvmAddress']:
+    """Check if aave tokens are stored in the globaldb.
+    We consider a token as stored when it has the aave v3 protocol set. This means that
+    it was not added by decoders but by the logic of update_aave_v3_underlying_assets.
```
LefterisJP, line 211:
```suggestion
    it was not added by decoders but by the logic of update_aave_v3_underlying_assets.
    
    Returns all tokens that are not stored in the global db.
```

rotkehlchen/tasks/assets.py
```diff
@@ -196,10 +202,88 @@ def update_owned_assets(user_db: DBHandler) -> None:
...
+            'AND chain=?',
+            tuple(token_to_underlying.keys()) + (chain_id.serialize_for_db(),),
+        )
+
+        missing_tokens = set(token_to_underlying.keys())
+        for row in cursor:
+            if row[1] == CPT_AAVE_V3:
+                missing_tokens.remove(row[0])
+
+    output = list(missing_tokens)  # we return a sorted list to avoid issues with cassettes
```
LefterisJP, line 226: Doing more work in production just to make VCR easier is not something we should do.

How about abstracting this in a way that can be mocked in the tests and adds the sorting only for tests?

rotkehlchen/tasks/assets.py
```diff
@@ -196,10 +202,88 @@ def update_owned_assets(user_db: DBHandler) -> None:
...
+    output.sort()
+    return output
+
+
+def _batch_query_properties(
+        node_inquirer: 'EvmNodeInquirer',
+        addresses: list['ChecksumEvmAddress'],
+) -> dict['ChecksumEvmAddress', tuple[str, str, int]]:
+    """Query details for the addresses provided.
```
LefterisJP, line 235:
```suggestion
    """Query details for the provided addresses.
```

rotkehlchen/tasks/assets.py
```diff
@@ -196,10 +202,88 @@ def update_owned_assets(user_db: DBHandler) -> None:
...
 def update_aave_v3_underlying_assets(chains_aggregator: 'ChainsAggregator') -> None:
-    """Fetch the Aave v3 underlying assets and populate `underlying_tokens_list` in globaldb"""
+    """Fetch the Aave v3 underlying assets and populate `underlying_tokens_list` in globaldb
+
+    This function is heavy since it makes several queries to external nodes. The logic has
+    been refactored to do the following:
+
+    1. Query all the reserver tokens (getAllReservesTokens).
+    2. For each reserve token query the debt tokens (getReserveTokensAddresses).
+    3. Check if we have in the db the addresses for the debt tokens.
```
LefterisJP, line 276:
```suggestion
    3. Check if we have the addresses for the debt tokens in the db.
```

rotkehlchen/tasks/assets.py
```diff
@@ -196,10 +202,88 @@ def update_owned_assets(user_db: DBHandler) -> None:
...
+    4. For those addresses that we don't have in the db query
```
LefterisJP, line 277:
```suggestion
    4. For the addresses we don't have in the db, query
```

rotkehlchen/tasks/assets.py
```diff
@@ -196,10 +202,88 @@ def update_owned_assets(user_db: DBHandler) -> None:
...
+    their properties (name, symbol, decimals).
+    5. For each reserve token create ir or ensure that the protocol is aave-v3 and the
```
LefterisJP, line 279:
```suggestion
    5. For each reserve token create it or ensure that the protocol is aave-v3 and the
```

rotkehlchen/tasks/assets.py
```diff
@@ -239,41 +323,62 @@ def update_aave_v3_underlying_assets(chains_aggregator: 'ChainsAggregator') -> N
...
-                    log.error(
-                        f'Failed to add underlying token {underlying_token_address} for '
-                        f'{decoded_reserve_token.identifier} due to {e!s}',
-                    )
+                if (deserialized_reserve := deserialize_evm_address(reserve)) != ZERO_ADDRESS:
+                    if deserialized_reserve in reserve_to_underlying:
+                        continue  # the same token can appear multiple times
+                    reserve_to_underlying[deserialized_reserve] = underlying_token_address
+
+        # check the tokens missing and query their properties
```
LefterisJP, line 339:
```suggestion
        # check the missing tokens and query their properties
```

rotkehlchen/tasks/assets.py
```diff
@@ -239,41 +323,62 @@ def update_aave_v3_underlying_assets(chains_aggregator: 'ChainsAggregator') -> N
...
+                    token_kind=EvmTokenKind.ERC20,
+                    underlying_tokens=[UnderlyingToken(address=underlying_address, token_kind=EvmTokenKind.ERC20, weight=ONE)],  # noqa: E501
+                    symbol=symbol,
+                    name=name,
+                    decimals=decimals,
+                    encounter=encounter,
+                )
+            except DeserializationError as e:
+                log.error(
+                    'Failed to deserialize Aave v3 reserve token address: '
```
LefterisJP, line 374: shouldn't you also add chain name in both this and the other error log beneat this one?

rotkehlchen/tests/unit/decoders/test_aave_v3.py
```diff
@@ -1373,3 +1383,69 @@ def test_aave_v3_interest_on_transfer(ethereum_inquirer, ethereum_accounts) -> N
...
+            tx_hash=tx_hash,
+            sequence_index=185,
+            timestamp=timestamp,
+            location=Location.ETHEREUM,
+            event_type=HistoryEventType.RECEIVE,
+            event_subtype=HistoryEventSubType.RECEIVE_WRAPPED,
+            asset=EvmToken('eip155:1/erc20:0x18577F0f4A0B2Ee6F4048dB51c7acd8699F97DB8'),
+            balance=Balance(amount=FVal(borrow_amount)),
+            location_label=ethereum_accounts[0],
+            notes=f'Receive {borrow_amount} variableDebtEthLidoGHO from AAVE v3',
```
LefterisJP, line 1433: What is this example exactly? Borrowing 2 different tokens?

yabirgb: yes, the new ones. Added it to verify that the lido tokens work as expected

rotkehlchen/chain/gnosis/modules/aave/v3/constants.py
```diff
@@ -2,4 +2,4 @@
 
 from rotkehlchen.chain.evm.types import string_to_evm_address
 
-AAVE_V3_DATA_PROVIDER: Final = string_to_evm_address('0x501B4c19dd9C2e06E94dA7b6D5Ed4ddA013EC741')
+AAVE_V3_DATA_PROVIDER: Final = string_to_evm_address('0x57038C3e3Fe0a170BB72DE2fD56E98e4d1a69717')
```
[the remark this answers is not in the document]
yabirgb: is the new one. It has a different abi but the info is the same. I updated it to have all of them up to date whith what is available at the aave listings. This one is the mantained afaik

LefterisJP (approved):
LGTM

PR #9159, opened 2024-12-26
Support unified etherscan api

nicholasyoder:
Adds support for the new unified etherscan api. Backend changes only - the new setting controlling which api version to use will need added to the frontend.


rotkehlchen/externalapis/etherscan.py
```diff
@@ -238,7 +239,15 @@ def _query(
         an unexpected response is returned. Also in the case of exhausting the backoff time.
         """
         result = None
-        query_str = f'https://{self.prefix_url}{self.base_url}/api?module={module}&action={action}'
+        cached_settings = CachedSettings()
+        if cached_settings.get_settings().use_unified_etherscan_api:
+            self.service_name = ExternalService.ETHERSCAN
+            query_str = f'https://api.etherscan.io/v2/api?chainid={self.chain.to_chain_id().serialize()}&'
```
yabirgb, line 245: maybe you can save as an attribute this and save 2 function calls per etherscan call?

rotkehlchen/externalapis/etherscan.py
```diff
@@ -238,7 +239,15 @@ def _query(
         an unexpected response is returned. Also in the case of exhausting the backoff time.
         """
         result = None
-        query_str = f'https://{self.prefix_url}{self.base_url}/api?module={module}&action={action}'
+        cached_settings = CachedSettings()
```
yabirgb, line 242: could cached_settings be in the attributes? that avoids creating it every time

rotkehlchen/externalapis/etherscan.py
```diff
@@ -238,7 +241,16 @@ def _query(
         an unexpected response is returned. Also in the case of exhausting the backoff time.
         """
         result = None
-        query_str = f'https://{self.prefix_url}{self.base_url}/api?module={module}&action={action}'
+        if self.cached_settings.get_settings().use_unified_etherscan_api:
```
LefterisJP, line 244: I don't get what this is. I mean I get the sort of general idea but the details are very hard to understand from the code and very obscure.

Waht's more it adds multiple if conditions to each and every etherscan api call.

That's a no-go. A setting like that should just setup etherscan queries in a specific way but not add multiple conditionals to every call

nicholasyoder: Changed this to use an approach like what was suggested in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1322179131550732370), which removes the if conditions from the query function.

rotkehlchen/rotkehlchen.py
```diff
@@ -1205,6 +1207,13 @@ def set_settings(self, settings: ModifiableDBSettings) -> tuple[bool, str]:
 
         with self.data.db.user_write() as cursor:
             self.data.db.set_settings(cursor, settings)
+
+        for chain in SUPPORTED_EVM_CHAINS:
```
LefterisJP, line 1211: add this under an if. If the unified etherscan api key is given/changed.

rotkehlchen/externalapis/etherscan.py
```diff
@@ -108,7 +108,10 @@ def __init__(
     ) -> None:
         super().__init__(database=database, service_name=service)
         self.msg_aggregator = msg_aggregator
+        self.original_service = service
+        self.cached_settings = CachedSettings()
         self.chain = chain
+        self.chain_id = self.chain.to_chain_id().serialize()
```
LefterisJP, line 114: Don't do this. Why would you save this in memory of each etherscan object? You only use it a single if condition of a single function. Just do it there.

The only thing that makes sense to be saved in memory along with each object is something that's used everywhere and a lot

rotkehlchen/externalapis/etherscan.py
```diff
@@ -118,6 +121,9 @@ def __init__(
             SupportedBlockchain.SCROLL,
         ) else 'api-'
         self.base_url = base_url
+        self.full_url: str = ''
```
LefterisJP, line 124: why keep both base and full url? Is this required?

rotkehlchen/externalapis/etherscan.py
nicholasyoder: Removed this. It was [suggested](https://github.com/rotki/rotki/pull/9159#discussion_r1898032819) to do it like this earlier when the chain id was being referenced for every api call.

rotkehlchen/externalapis/etherscan.py
nicholasyoder: I've reworked this a bit. Now its called `legacy_api_domain` and `api_url` - which should hopefully be a bit clearer, and also removes the `prefix_url` attribute. Now `api_url` holds the url currently in use, and `legacy_api_domain` holds the etherscan domain used when creating the url for the old api version.

rotkehlchen/tests/api/test_makerdao_dsr.py
```diff
@@ -144,14 +144,14 @@ def mock_etherscan_for_dsr(
     proxy1_contents = proxy1[2:].lower()
     proxy2_contents = proxy2[2:].lower()
 
-    def mock_requests_get(url: str, *args: Any, **kwargs) -> _patch | MockResponse:
-        if 'etherscan.io/api?module=proxy&action=eth_blockNumber' in url:
+    dsr_params = params
+
+    def mock_requests_get(url: str, params: dict[str, str], *args: Any, **kwargs) -> _patch | MockResponse:  # noqa: E501
+        if 'etherscan.io/api' in url and {'module': 'proxy', 'action': 'eth_blockNumber'}.items() <= params.items():  # noqa: E501
```
LefterisJP, line 150: minor suggestion here would be to add this in a constant in the same file to not repeat

```suggestion
        if ETHERSCAN_API in url and {'module': 'proxy', 'action': 'eth_blockNumber'}.items() <= params.items():  # noqa: E501
```

rotkehlchen/tests/api/test_settings.py
```diff
@@ -669,3 +671,53 @@ def test_excluded_exchanges_settings(rotkehlchen_api_server: 'APIServer') -> Non
...
+            assert_proper_response(response)
+
+        request_count = 0
+        with patch.object(etherscan.session, 'get', wraps=mock_request):
+            assert etherscan.has_activity(
+                account=string_to_evm_address('0x95222290DD7278Aa3Ddd389Cc1E1d165CC4BAfe5'),
+            ) == HasChainActivity.TRANSACTIONS
+            assert request_count == 1  # confirm the asserts in mock_request were hit
+
+        # Check it works for a L2 chain as well (Gnosis)
```
LefterisJP, line 717:
```suggestion
        # Check it works for an L2 chain as well (Gnosis)
```

rotkehlchen/tests/utils/mock.py
```diff
@@ -179,19 +185,18 @@ def patch_etherscan_request(etherscan, mock_data: dict[str, Any]):
...
     counter = 0
 
-    def mock_etherscan_query(url, **kwargs):  # pylint: disable=unused-argument
+    def mock_etherscan_query(url, params, **kwargs):  # pylint: disable=unused-argument
         nonlocal counter
-        match = ETHERSCAN_ACTION_RE.search(url)
-        if match is None:
-            raise AssertionError(f'Could not parse etherscan query: {url}')
+        action = params.get('action')
+        if action is None:
```
LefterisJP, line 191:
```suggestion
        if (action := params.get('action')) is None:
```

rotkehlchen/tests/utils/blockchain.py
```diff
@@ -221,21 +221,38 @@ def mock_etherscan_query(
...
+                        return False
+
+            if keys is not None:
+                for key in keys:
+                    if key not in params:
+                        return False
+
+            return True
+
+        if 'etherscan.io/api' in url and check_params(values={'module': 'account', 'action': 'balance'}, keys=['address']):  # noqa: E501
```
LefterisJP, line 242: same here for etherscan api. Use the same constant all across the file/s.

rotkehlchen/tests/api/test_makerdao_dsr.py
nicholasyoder: I actually put the constant in tests/utils/bockchain.py since its also used there and seems like a better place for it to be defined than in the makerdao test.

rotkehlchen/tests/utils/blockchain.py
nicholasyoder: Using constant now, but actually moved the url check into the `check_params` function to cut down on repetition even more.

LefterisJP (approved):
LGTM!

PR #9161, opened 2024-12-26
Add OpenOcean swap decoder

### LINKED ISSUE #7254: Support OpenOcean Swaps in all EVM chains that support it.
## Abstract

We should support openocean swaps in ethereum mainnet and all other supported EVM chains.  The mainnet openocean v2 contract is here: https://etherscan.io/address/0x6352a56caadc4f1e25cd6c75970fa768a3304e64

We had an issue for arbitrum: https://github.com/rotki/rotki/issues/7103 but merged it with this one.

 https://arbiscan.io/tx/0xf882a6a323bdc7ac05e78e2c346fddb489249afc04e68f09b1858ab357c8ede0

Reminder to self: When implemented edit and redecode my custom event/s. We need to get a query to the DB or rotki itself which will tell you which transactions need to be touched.

## Specification

Support them properly in the decoders. Check all example swap types and decode them.

nicholasyoder:
Closes #7254


rotkehlchen/chain/ethereum/modules/open_ocean/decoder.py
```diff
@@ -0,0 +1,5 @@
+from rotkehlchen.chain.evm.decoding.open_ocean.decoder import OpenOceanCommonDecoder
+
+
+class OpenOceanDecoder(OpenOceanCommonDecoder):
+    pass
```
yabirgb, line 5: Can you do

```suggestion
class OpenOceanDecoder(OpenOceanCommonDecoder):
    ...
```

instead of pass here and in the others?

rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
```diff
@@ -0,0 +1,188 @@
...
+    from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
+    from rotkehlchen.chain.evm.structures import EvmTxReceiptLog
+    from rotkehlchen.fval import FVal
+    from rotkehlchen.history.events.structures.evm_event import EvmEvent
+    from rotkehlchen.user_messages import MessagesAggregator
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+SWAPPED_TOPIC: Final = b'v\xaf"J\x148e\xa5\x0bAIn\x1asb&\x98i,V\\\x12\x14\xbc\x86/\x18\xe2-\x82\x9c^'  # noqa: E501
```
yabirgb, line 41: can you move this one to the constant file?

rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
```diff
@@ -0,0 +1,188 @@
...
+            self,
+            evm_inquirer: 'EvmNodeInquirer',
+            base_tools: 'BaseDecoderTools',
+            msg_aggregator: 'MessagesAggregator',
+    ) -> None:
+        super().__init__(
+            evm_inquirer=evm_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+        )
```
yabirgb, line 56: no need for init if you just call super

rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
```diff
@@ -0,0 +1,188 @@
...
+    from rotkehlchen.user_messages import MessagesAggregator
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+SWAPPED_TOPIC: Final = b'v\xaf"J\x148e\xa5\x0bAIn\x1asb&\x98i,V\\\x12\x14\xbc\x86/\x18\xe2-\x82\x9c^'  # noqa: E501
+
+
+class OpenOceanCommonDecoder(DecoderInterface, ABC):
+
```
yabirgb, line 45: I read that there is difference between this and `metaclass=ABC`. Maybe worth reading since I don't have all the details

rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
```diff
@@ -0,0 +1,188 @@
...
+    def _get_asset_and_amount(
+            self,
+            asset_address: 'ChecksumEvmAddress',
+            raw_amount: int,
+    ) -> tuple['CryptoAsset', 'FVal']:
+        """Get asset and normalized amount from asset address and raw amount.
+        Specifically handles when asset is native rather than erc20 token.
+        """
+        asset = self.base.get_or_create_evm_asset(address=asset_address)
+        if asset == self.evm_inquirer.native_token:
```
yabirgb, line 67: this only happens on polygon and gnosis right?

rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
```diff
@@ -0,0 +1,188 @@
...
+            amount = token_normalized_value_decimals(
+                token_amount=raw_amount,
+                token_decimals=18,
+            )
+        else:
+            amount = token_normalized_value(
+                token_amount=raw_amount,
+                token=asset.resolve_to_evm_token(),
```
yabirgb, line 75: it is already an evm token right? No need to resolve? You can also do 

```
            amount = token_normalized_value_decimals(
                token_amount=raw_amount,
                token_decimals=18 if asset == self.evm_inquirer.native token else asset.decimals,
            )
```

rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
```diff
@@ -0,0 +1,188 @@
...
+                (receive_amount is None or event.balance.amount == receive_amount)
+            ):
+                event.event_type = HistoryEventType.TRADE
+                event.event_subtype = HistoryEventSubType.RECEIVE
+                event.notes = f'Receive {event.balance.amount} {event.asset.resolve_to_asset_with_symbol().symbol} from {OPENOCEAN_LABEL} swap'  # noqa: E501
+                event.counterparty = CPT_OPENOCEAN
+                in_event = event
+
+        if in_event is None or out_event is None:
+            log.error(f'Failed to find both out and in events for {OPENOCEAN_LABEL} swap transaction {transaction}')  # noqa: E501
```
yabirgb, line 144: no need to reshufle here?

rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
```diff
@@ -0,0 +1,188 @@
...
+            self,
+            transaction: EvmTransaction,
+            decoded_events: list['EvmEvent'],
+            all_logs: list['EvmTxReceiptLog'],
+    ) -> list['EvmEvent']:
+        """Handle post decoding for OpenOcean.
+        Decodes swaps that used uniswap and have no SWAPPED_TOPIC tx_log.
+        """
+        topics = {tx_log.topics[0] for tx_log in all_logs}
+        if SWAPPED_TOPIC in topics:
```
yabirgb, line 161:
```suggestion
        if SWAPPED_TOPIC in (topics := {tx_log.topics[0] for tx_log in all_logs}):
```

rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
nicholasyoder: As far as I can tell simply using `ABC` is correct. That is the first way its shown here: https://docs.python.org/3/library/abc.html, although it does also show `metaclass=ABCMeta` as being valid as well.

rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
nicholasyoder: No actually I think I encountered this on ethereum (the native_token was ETH anyway).

rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
nicholasyoder: No its only a CryptoAsset - notice that I'm using `get_or_create_evm_asset` not `get_or_create_evm_token`, so that it doesn't throw a NotErc20Compliant error when the asset is native ETH. With that said, I can simplify it a bit and only use `token_normalized_value_decimals`, but I do still have to resolve to evm token to get decimals when its not the native asset.

rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
nicholasyoder: No - if one of the events isn't found there's no reshuffling that can be done is there?

BTW, I changed this error to only be a warning, since in rare cases this does happen, for instance when meta mask uses openocean to do a swap.

rotkehlchen/chain/evm/decoding/metamask/decoder.py
```diff
@@ -81,7 +81,10 @@ def _decode_swap(self, context: DecoderContext) -> DecodingOutput:
                 event.notes = f'Swap {event.balance.amount} {event.asset.resolve_to_asset_with_symbol().symbol} in metamask'  # noqa: E501
                 event.address = self.router_address
                 out_event = event
-            elif event.event_type == HistoryEventType.RECEIVE and event.location_label == sender:  # find the receive event  # noqa: E501
+            elif (
+                    (event.event_type == HistoryEventType.TRADE and event.event_subtype == HistoryEventSubType.RECEIVE) or  # noqa: E501
+                    (event.event_type == HistoryEventType.RECEIVE and event.event_subtype == HistoryEventSubType.NONE and event.location_label == sender)  # noqa: E501
+            ):  # find the receive event
```
nicholasyoder, line 87: Added this to fix a metamask test. Metamask apparently routes through openocean sometimes, in which case the event has already been modified to trade/receive here by the openocean decoder. It's already checking for trade/spend, so no need to change that.

LefterisJP: Can you expain this in a comment too?

LefterisJP (approved):
LGTM

PR #9164, opened 2024-12-27
Fix cc price queries for special assets

yabirgb:
In a refactor the return type of a function was modified and it broke the assumtpion about types in other part of the code.

It went unnoticed since we used getattr. I've changed it to a case so we can see by searching for the function name where it is used and help us in future refactors. Also added a test

Closes #(issue_number)

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* :bug:`9163` Cryptocompare will handle correctly price queries again.
+
 * :release:`1.37.0 <2024-12-24>`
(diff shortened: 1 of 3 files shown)


docs/changelog.rst
```diff
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* :bug:`9163` Cryptocompare will handle correctly price queries again.
```
```suggestion
* :bug:`9163` Cryptocompare price queries will be handled correctly again. Fixes "the FVal issue"
```

LefterisJP (approved):
LGTM. The changelog fix can go to another PR

PR #9166, opened 2024-12-29
Fix bisq price queries

yabirgb:
The api that we used to query BSQ market price isn't available anymore so fallback to the definition of 100 satoshis to calculate the price


LefterisJP (approved):
LGTM

PR #9167, opened 2024-12-29
Increase window for bybit

yabirgb:
Balance query was failing for me due to

```
'invalid request, please check your server timestamp or recv_window param. req_timestamp[1735289540738],server_timestamp[1735289550958],recv_window[5000]'
```

increased the window from 5s to 10 seconds. Since we perform only read events it should be fine


LefterisJP (approved):
LGTM

PR #9168, opened 2024-12-29
Address connection errors

yabirgb:
When the app is left running for a long time and I come back to it I see connectionError: connection clossed by remote server. This happens to me mostly with kraken.

After some investigation the cause seems to be the keep alive property of the connection. The session will use the connection that has expired and get this error from the server.

It is not a behaviour from requests but internal to urllib. The solution seems to be to use the httpAdapter class and enforce a retry, by default it doesn't happen. From the rquests docs:

https://docs.python-requests.org/en/latest/api/#requests.adapters.HTTPAdapter

```
max_retries – The maximum number of retries each connection should attempt.
Note, this applies only to failed DNS lookups, socket connections and
connection timeouts, never to requests where data has made it to the server.
By default, Requests does not retry failed connections.
```

### DIFF
--- a/rotkehlchen/utils/network.py
+++ b/rotkehlchen/utils/network.py
@@ -6,6 +6,7 @@ from typing import Any, Literal, overload
 
 import gevent
 import requests
+from requests.adapters import HTTPAdapter
 
 from rotkehlchen.constants import GLOBAL_REQUESTS_TIMEOUT
 from rotkehlchen.db.settings import CachedSettings
@@ -14,6 +15,22 @@ from rotkehlchen.logging import RotkehlchenLogsAdapter
 
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
+
+
+def create_session() -> requests.Session:
+    """Create a requests session with the configuration to retry a maximum
+    of 3 times on connection errors or DNS resolution errors.
+
+    From the requests docs about max_retries:
+    The maximum number of retries each connection should attempt. Note, this applies only
+    to failed DNS lookups, socket connections and connection timeouts, never to requests
+    where data has made it to the server.
+    """
+    session = requests.Session()
+    adapter = HTTPAdapter(max_retries=5)
+    session.mount('http://', adapter)
+    session.mount('https://', adapter)
+    return session
 
 
 def request_get(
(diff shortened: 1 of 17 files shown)


rotkehlchen/utils/network.py
```diff
@@ -16,6 +17,21 @@
...
 
+def create_session() -> requests.Session:
+    """Create a requests session with the configuration to retry a maximum
+    of 3 times on connection errors or DNS resolution errors.
+    """
+    session = requests.Session()
+    # from the requests docs about max_retries:
+    # The maximum number of retries each connection should attempt. Note, this applies only
+    # to failed DNS lookups, socket connections and connection timeouts, never to requests
+    # where data has made it to the server.
```
LefterisJP, line 28: Put this domment in the main docstring.

rotkehlchen/utils/network.py
```diff
@@ -16,6 +17,21 @@
...
+    adapter = HTTPAdapter(max_retries=3)
```
LefterisJP, line 29: We do the same but with more details and more args set in premium/premium.py

I think the extra arguments are needed.

```python
        self.session = requests.session()
        # Make sure to have 3 retries on read/connect/other errors for all requests
        # The reason for this is that we have noticed that in unstable/slow connections
        # rotki.com server will close/cause the connection to result to a read timeout
        # At the moment this only happens for the backup upload endpoint. More info:
        # https://github.com/rotki/rotki/pull/6423
        adapter = requests.adapters.HTTPAdapter()
        # Have to set retries like this, since this granularity is not given by HTTPAdapter ctor
        adapter.max_retries = Retry(
            total=3,  # have to set each one individually as the code checks them all
            read=3,
            connect=3,
            other=3,
            allowed_methods=False,  # retry on all method verbs
        )
        self.session.mount('https://', adapter)
```

And it should also be abstracted together with that code and combined.

LefterisJP (commented):
Good approach. Left a comment. But also worried / note sure if any request sessions were missed.

rotkehlchen/utils/network.py
yabirgb: Since we couldn't find references for why to set the different counts individually in the Retry object we agreed on going with `HTTPAdapter` and adjust later if needed. Also checked the code for possible sessions again and didn't find any

LefterisJP (approved):
LGTM

PR #9173, opened 2024-12-30
Query protocol balances in polygon gnosis and scroll

nicholasyoder:
Fixes bug where protocol balances weren't being queried for polygon, gnosis, and scroll.


rotkehlchen/tests/api/test_balances.py
```diff
@@ -512,6 +522,38 @@ def test_query_all_balances_errors(rotkehlchen_api_server: 'APIServer') -> None:
     )
 
 
+def test_protocol_balances_all_chains(rotkehlchen_api_server: 'APIServer') -> None:
```
LefterisJP, line 525: Don;t we already have `test_all_balance_classes_used`? They test different things?

nicholasyoder: Yes, they test different things I think. `test_all_balance_classes_used` is checking that `CHAIN_TO_BALANCE_PROTOCOLS` includes all the existing classes based on `ProtocolWithBalance`. The new `test_protocol_balances_all_chains` is checking that `_query_protocols_with_balance` actually gets called for all the chains in `CHAIN_TO_BALANCE_PROTOCOLS`.

LefterisJP (approved):
LGTM

PR #9176, opened 2024-12-30
Ignore txs with ignored assets in Wrapped

nicholasyoder:


rotkehlchen/tests/api/test_statistics.py
```diff
@@ -650,7 +662,7 @@ def test_query_wrap(
         'BASE': 1,
     }
     assert result['top_days_by_number_of_transactions'] == [
-        {'timestamp': 1718496000, 'amount': '2'},
+        {'timestamp': 1734307200, 'amount': '4'},
```
nicholasyoder, line 665: This number went from 2 to 4 because it is now counting based on `history_events` rather than `evm_transactions`, and the `history_events`/`evm_transactions` added in this test are not correlated the way they would be in real world data. I also added another EvmEvent using the same tx_hash as another event to confirm that it is still counting by transactions even though its not using the `evm_transactions` table.

rotkehlchen/db/history_events.py
```diff
@@ -861,11 +861,13 @@ def query_wrap_stats(self, from_ts: Timestamp, to_ts: Timestamp) -> dict[str, An
...
             cursor.execute(
-                "SELECT unixepoch(date(datetime(timestamp, 'unixepoch'), 'localtime'), 'utc'), COUNT(*) as tx_count FROM "  # noqa: E501
-                'evm_transactions WHERE timestamp >= ? AND timestamp <= ? '
-                "GROUP BY date(datetime(timestamp, 'unixepoch'), 'localtime') ORDER BY "
+                "SELECT unixepoch(date(datetime(timestamp/1000, 'unixepoch'), 'localtime'), 'utc'), COUNT(DISTINCT event_identifier) as tx_count "  # noqa: E501
+                'FROM evm_events_info JOIN history_events ON evm_events_info.identifier = history_events.identifier '  # noqa: E501
+                'WHERE timestamp >= ? AND timestamp <= ? AND history_events.asset NOT IN '
+                "(SELECT value FROM multisettings WHERE name = 'ignored_asset') "
+                "GROUP BY date(datetime(timestamp/1000, 'unixepoch'), 'localtime') ORDER BY "
                 'tx_count DESC LIMIT 10',
```
yabirgb, line 869: SEems that something is still off. 
image

This is what it show to me but If I go to the history events and filter
image

if Instead I include the ignored assets 
image

the number matches. I thought that also one case if for transactions that have more than one event with spam assets. Is that one handled?

rotkehlchen/tests/api/test_statistics.py
yabirgb: Did you add any ignored asset in this test?

yabirgb (approved):
Can you also add a changelog entry here?

yabirgb (changes requested):
Approved by mistake I believe

rotkehlchen/db/history_events.py
nicholasyoder: Transactions per chain should skip spam txs now as well. As mentioned in discord I had only modified the query for days with most transactions.

> for transactions that have more than one event with spam assets. Is that one handled?

Its checking history_events with spam assets, so if a transaction has multiple events, it will still count it as long as there is one event with non spam asset, but if all the transaction's events have ignored assets then the entire transaction will be skipped.

rotkehlchen/tests/api/test_statistics.py
nicholasyoder: I've added a transaction and event with an ignored asset now. I also reworked the tx_hashes in the test so that the history_events correctly correspond to evm_transactions since the per chain tx count query now does a join on these two tables and needs this to be correct.

yabirgb (approved):
This PR fixed all the issues that I had in my life

LefterisJP (approved):
LGTM

PR #9179, opened 2024-12-31
Fix incorrect order in curve deposit & stake

yabirgb:
It was possible that if multiple deposit events happened they weren't correctly ordered and other events could appear in between such approvals.

In the test case the order was:

```
- deposit
- deposit
- approval
- receive wrapped
```

and after this fix it is:

```
- approval
- deposit
- deposit
- receive wrapped
```

Closes #(issue_number)

## Checklist


LefterisJP (approved):
LGTM

PR #9181, opened 2024-12-31
Add Paraswap v6 decoder

### LINKED ISSUE #9083: Support Paraswap v6
Example mainnet transaction: https://etherscan.io/tx/0x04df701a75c3dc3f1283c6a26e668bec4bd92cf7f02e9963c3c887f5376c590e

This is not properly decoded.

We need to support paraswap v6 in all chains it's been deployed. Same as the previous paraswap is supported

Addendum: Seems paraswap is now in gnosis and we should also support it. So adding this as a part of this issue.

Note to self: Make sure to reset customized events for this

nicholasyoder:
Closes #9083


rotkehlchen/chain/evm/decoding/paraswap/v6/decoder.py
```diff
@@ -0,0 +1,74 @@
...
+    ) -> list['EvmEvent']:
+        """Decode Paraswap v6 swaps in post decoding
+        since they don't have any relevant log event.
+        """
+        if transaction.input_data[:4] not in PARASWAP_METHODS:
+            return decoded_events
+
+        self._decode_swap(
+            context=DecoderContext(
+                tx_log=all_logs[0],  # not used
```
prettyirrelevant, line 56: instead of doing this, why not define the parameters `_decode_swap()` needs? i.e. `transaction` and `decoded_events` then pass those directly.

rotkehlchen/chain/evm/decoding/paraswap/v5/decoder.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def _decode_paraswap_swap(self, context: DecoderContext) -> DecodingOutput:
+        """This decodes the following types of trades:
+        - Simple Buy
+        - Simple Swap
+        - Multi Swap
+        - Mega Swap
+        - Direct Swap on Uniswap V3
+        - Direct Swap on Curve V1 and V2
+        - Direct Swap on Balancer V2"""
```
prettyirrelevant, line 36:
```suggestion
        - Direct Swap on Balancer V2
        """
```

rotkehlchen/chain/evm/decoding/paraswap/v6/decoder.py
nicholasyoder: I went to do this, and realized that `tx_log` is actually being referenced when it creates a fee event. As far as I can tell though its only used to determine the sequence index. The only attribute here that isn't being used is `action_items`. So since most of `DecoderContext`'s attributes are actually referenced in `_decode_swap` I think keeping it like this is probably best. Also this way its simpler in the v5 decoder where there it directly passes the context.

I've updated the comment here and also changed it to use the last log event instead of the first so it for sure uses a higher sequence index for fee events.

prettyirrelevant: I didn't know `tx_log` was used in `_decode_swap()`. 

> I've updated the comment here and also changed it to use the last log event instead of the first so it for sure uses a higher sequence index for fee events.

yeah, it's okay in its current form.

rotkehlchen/tests/unit/decoders/test_paraswap_v6.py
```diff
@@ -0,0 +1,789 @@
...
+        location_label=user_address,
+        notes=f'Receive {receive_amount} PSP as the result of a swap in paraswap',
+        counterparty=CPT_PARASWAP,
+        address=PARASWAP_AUGUSTUS_V6_ROUTER,
+    ), EvmEvent(
+        tx_hash=tx_hash,
+        sequence_index=3,
+        timestamp=timestamp,
+        location=Location.ETHEREUM,
+        event_type=HistoryEventType.SPEND,
```
LefterisJP, line 615: The fee event for the swap should be TRADE/FEE

rotkehlchen/chain/evm/decoding/paraswap/v6/decoder.py
```diff
@@ -0,0 +1,74 @@
...
+
+        self._decode_swap(
+            context=DecoderContext(
+                tx_log=all_logs[-1],  # only referenced to set sequence index for new fee events
+                transaction=transaction,
+                action_items=[],
+                decoded_events=decoded_events,
+                all_logs=all_logs,
+            ),
+            receiver=ZERO_ADDRESS,
```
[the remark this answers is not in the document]
nicholasyoder: zero address was just a place holder. I've changed `receiver` to be an optional argument, and adjusted the docstring as discussed in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1324076724270403635)

LefterisJP (approved):
LGTM

PR #9189, opened 2025-01-02
Update bitmex

yabirgb:
This PR:

- brings the fixes of bitmex to bugfixes
- changes the base url from `bitmex.com` to `www.bitmex.com`. I tested it myself and without the `www.` it fails.
- handles errors in the api response
- fixes wrong value decoded in the asset movements. We were applying decimals only to bitcoin and other assets got huge values
- eth uses the special symbol `gwei` so mapped it

### DIFF
--- a/rotkehlchen/chain/ethereum/utils.py
+++ b/rotkehlchen/chain/ethereum/utils.py
@@ -48,6 +48,10 @@ def token_normalized_value_decimals(token_amount: int, token_decimals: int | Non
     return token_amount / (FVal(10) ** FVal(token_decimals))
 
 
+def normalized_fval_value_decimals(amount: FVal, decimals: int) -> FVal:
+    return amount / (FVal(10) ** FVal(decimals))
+
+
 def token_raw_value_decimals(token_amount: FVal, token_decimals: int | None) -> int:
     if token_decimals is None:  # if somehow no info on decimals ends up here assume 18
         token_decimals = 18
--- a/rotkehlchen/exchanges/bitmex.py
+++ b/rotkehlchen/exchanges/bitmex.py
@@ -2,16 +2,17 @@ import hashlib
 import hmac
 import json
 import logging
-import time
 from collections.abc import Sequence
 from json.decoder import JSONDecodeError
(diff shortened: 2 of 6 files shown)


rotkehlchen/chain/ethereum/utils.py
```diff
@@ -48,6 +48,10 @@ def token_normalized_value_decimals(token_amount: int, token_decimals: int | Non
     return token_amount / (FVal(10) ** FVal(token_decimals))
 
 
+def normalized_fval_value_decimals(amount: FVal, decimals: int) -> FVal:
```
LefterisJP, line 51: Don"t reinvent the wheel and make same functions. 

There is already one doing this `token_normalized_value_decimals`

rotkehlchen/exchanges/bitmex.py
```diff
@@ -129,11 +149,33 @@ def edit_exchange_credentials(self, credentials: ExchangeAuthCredentials) -> boo
         return changed
 
     def first_connection(self) -> None:
+        if self.first_connection_made:
+            return
+        try:
```
LefterisJP, line 154:
```suggestion

        try:
```

rotkehlchen/exchanges/bitmex.py
```diff
@@ -307,12 +351,19 @@ def query_online_history_events(
...
 
         log.debug('Bitmex deposit/withdrawals query', results_num=len(resp))
         movements = []
         for movement in resp:
             try:
+                timestamp = iso8601ts_to_timestamp(movement['timestamp'])
+                if timestamp < start_ts:
+                    continue
+                if timestamp > end_ts:
+                    continue
```
LefterisJP, line 365: Can't this be combined?

```suggestion
                if end_ts < timestamp < start_ts:
                    continue
```

rotkehlchen/chain/ethereum/utils.py
LefterisJP: we discussed this will go in another PR

LefterisJP (approved):
LGTM

PR #9193, opened 2025-01-02
Query OKX funding account balances

nicholasyoder:


rotkehlchen/exchanges/okx.py
```diff
@@ -50,6 +50,8 @@
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
+OKX_ENDPOINTS: TypeAlias = Literal['currencies', 'trading balance', 'funding balance', 'trades', 'deposits', 'withdrawals']  # noqa: E501
+
```
yabirgb, line 54: Would it make sense to convert this to an enum and use the endpoint as value?. Since you compare against it I believe it is better and less error prone

rotkehlchen/exchanges/okx.py
```diff
@@ -228,30 +232,36 @@ def _api_query_list_paginated(
...
-            raise RemoteError(
-                f'{self.name} balance API request failed due to unexpected response {msg}',
-            ) from e
+        currencies_data: list[dict] = []
+        for endpoint in (
+            'trading balance',
+            'funding balance',
+        ):
+            data = self._api_query_list(endpoint=endpoint)
+            if not (len(data) > 0 and isinstance(data[0], dict)):
```
yabirgb, line 254: in the case of empty balances do you get `len(data) == 0`?

nicholasyoder: Yes, `data` will be an empty list when there's no futures balances. I've reworked this logic, and it should handle it being empty correctly now.

rotkehlchen/exchanges/okx.py
```diff
@@ -51,6 +52,15 @@
 log = RotkehlchenLogsAdapter(logger)
 
 
+class OkxEndpoint(Enum):
+    CURRENCIES = 'currencies'
```
yabirgb, line 56: I meant to do this

```suggestion
    CURRENCIES = '/api/v5/asset/currencies'
```

since afaik you etierh compare to the key or use the value

rotkehlchen/exchanges/okx.py
```diff
@@ -120,11 +130,13 @@ def _api_query(
         options = options.copy() if options else {}
 
         params = {}
-        if endpoint == 'currencies':
+        if endpoint == OkxEndpoint.CURRENCIES:
             path = '/api/v5/asset/currencies'
-        elif endpoint == 'balance':
+        elif endpoint == OkxEndpoint.TRADING_BALANCE:
```
yabirgb, line 135: it will allow you to remove this if/elif thing

rotkehlchen/exchanges/okx.py
```diff
@@ -228,31 +240,35 @@ def _api_query_list_paginated(
...
+        https://www.okx.com/docs-v5/en/#trading-account-rest-api-get-balance
+        https://www.okx.com/docs-v5/en/#funding-account-rest-api-get-balance
 
         May raise
         - RemoteError if the OKX API or price oracle returns an unexpected response
         """
-        data = self._api_query_list(endpoint='balance')
+        currencies_data: list[dict] = []
+        data = self._api_query_list(endpoint=OkxEndpoint.TRADING_BALANCE)
         if not (len(data) == 1 and isinstance(data[0], dict)):
```
yabirgb, line 258: I still fail to see how this works. What happens when there is more than 1 element. Is there a test case for this?

rotkehlchen/exchanges/okx.py
nicholasyoder: ah, yes this simplifies it.

rotkehlchen/exchanges/okx.py
nicholasyoder: explained in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1324395011382444062)

LefterisJP (approved):
LGTM

PR #9192, opened 2025-01-02
test: restore original 60 second wait for price fetching

kelsos:
Closes #(issue_number)

## Checklist

- [ ] The PR modified the frontend, and updated the [user guide](https://github.com/rotki/rotki/blob/develop/docs/usage_guide.rst) to reflect the changes.

PR #9194, opened 2025-01-02
Adjust retry on session

yabirgb:
urllib3 by default retries on 3 status codes that include too_many_requests. Since we handle this manually and in different ways for each consumer this commit sets the list `status_forcelist` to an empty list.

Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/utils/network.py
+++ b/rotkehlchen/utils/network.py
@@ -7,6 +7,7 @@ from typing import Any, Literal, overload
 import gevent
 import requests
 from requests.adapters import HTTPAdapter
+from urllib3.util import Retry
 
 from rotkehlchen.constants import GLOBAL_REQUESTS_TIMEOUT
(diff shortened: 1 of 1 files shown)


yabirgb:
Tested that it works here https://github.com/rotki/rotki/pull/9192

rotkehlchen/utils/network.py
```diff
@@ -27,7 +28,15 @@ def create_session() -> requests.Session:
     where data has made it to the server.
     """
     session = requests.Session()
-    adapter = HTTPAdapter(max_retries=5)
+    # we don't use total in the adapter because it seems to triggers on certain bad status codes
```
LefterisJP, line 31:
```suggestion
    # we don't use total in the adapter because it seems to trigger on certain bad status codes
```

LefterisJP (approved):
LGTM

PR #9195, opened 2025-01-02
add alchemy price oracle

prettyirrelevant:
image


rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,424 @@
...
+            raise NoPriceForGivenTimestamp(
+                from_asset=from_asset,
+                to_asset=to_asset,
+                time=timestamp,
+            ) from e
+
+        payload: dict[str, str | int] = {
+            'startTime': (start_of_day_timestamp := Timestamp(timestamp - (timestamp % DAY_IN_SECONDS))),  # noqa: E501
+            'endTime': start_of_day_timestamp + DAY_IN_SECONDS,
+        }
```
yabirgb, line 141: this will query several points right? Can't you set the frequency to 1d? But also, do we want to get the average of prices? Why not set a 1h window around the timestamp that we want? one day seems to much

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,424 @@
...
+
+            if 'error' in entry:
+                log.warning(
+                    f'Queried Alchemy Price API from {from_asset.identifier} to {to_asset.identifier}. '  # noqa: E501
+                    f'API returned error: {entry["error"]}',
+                )
+                return ZERO_PRICE
+
+            if not (prices := entry['prices']):
+                log.warning(
```
yabirgb, line 275: this can raise keyError. In general different things in this function

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,424 @@
...
+                    f'Queried Alchemy Price API from {from_asset.identifier} to {to_asset.identifier}, '  # noqa: E501
+                    f'but received no prices in response',
+                )
+                return ZERO_PRICE
+
+            for price in prices:
+                if price['currency'] != 'usd':
+                    continue
```
yabirgb, line 283: what other currency is in there?

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,424 @@
...
+                    url=url,
+                    params=payload,
+                    timeout=CachedSettings().get_timeout_tuple(),
+                )
+            else:  # POST
+                response = self.session.post(
+                    url=url,
+                    json=payload,
+                    timeout=CachedSettings().get_timeout_tuple(),
+                )
```
yabirgb, line 400: This can be simplified I believe

``` 
self.session.request(
    method=verb,
    url=url,
    **{'params' if verb == 'GET' else 'json': payload},
    timeout=CachedSettings().get_timeout_tuple(),
)
```

rotkehlchen/tests/external_apis/test_alchemy.py
```diff
@@ -0,0 +1,79 @@
...
+from rotkehlchen.errors.misc import RemoteError
+from rotkehlchen.externalapis.alchemy import Alchemy
+from rotkehlchen.fval import FVal
+from rotkehlchen.types import ApiKey, ExternalService, ExternalServiceApiCredentials, Price
+
+
+@pytest.mark.vcr
+@pytest.mark.parametrize('should_mock_price_queries', [False])
+def test_alchemy_historical_prices(price_historian, session_alchemy):  # pylint: disable=unused-argument
+    eur = A_EUR.resolve()
```
yabirgb, line 13: since you are at it can you type the tests also?

rotkehlchen/tests/external_apis/test_alchemy.py
```diff
@@ -0,0 +1,79 @@
+import pytest
+
+from rotkehlchen.constants.assets import A_DAI, A_ETH, A_EUR, A_USD
+from rotkehlchen.errors.misc import RemoteError
+from rotkehlchen.externalapis.alchemy import Alchemy
+from rotkehlchen.fval import FVal
+from rotkehlchen.types import ApiKey, ExternalService, ExternalServiceApiCredentials, Price
+
+
+@pytest.mark.vcr
```
yabirgb, line 10: can you skip the api key? In case we need to change it in the future

rotkehlchen/tests/external_apis/test_alchemy.py
```diff
@@ -0,0 +1,79 @@
...
+        from_asset=dai,
+        to_asset=usd,
+    )
+    assert price == Price(FVal('0.9990162355'))
+
+
+def test_alchemy_invalid_api_key(price_historian, database):  # pylint: disable=unused-argument
+    with database.user_write() as write_cursor:  # add the api key to the DB
+        database.add_external_service_credentials(
+            write_cursor=write_cursor,
```
yabirgb, line 66: can you also test some error cases? Like assets without oracles and so on

rotkehlchen/globaldb/upgrades/manager.py
```diff
@@ -73,6 +74,10 @@
         from_version=9,
         function=migrate_to_v10,
     ),
+    UpgradeRecord(
+        from_version=10,
+        function=migrate_to_v11,
+    ),
```
yabirgb, line 80: How does introducing a new version affect to the thing you did for the ranges in the updates before the globaldb start?

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,424 @@
...
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+class Alchemy(
+    ExternalServiceWithApiKeyOptionalDB,
+    HistoricalPriceOracleInterface,
+    PenalizablePriceOracleMixin,
+):
```
yabirgb, line 39: Do you need any change related to the api schema to accept this one? Have you checked deleting prices from the cache?

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,424 @@
...
+    def _query(
+            self,
+            verb: Literal['GET', 'POST'],
+            payload: dict[str, Any],
+            endpoint: Literal['by-symbol', 'by-address', 'historical'],
+    ) -> dict[str, Any]:
+        if (api_key := self._get_api_key()) is not None:
+            base_url = f'https://api.g.alchemy.com/prices/v1/{api_key}/tokens'
+        else:
+            base_url = 'https://api.g.alchemy.com/prices/v1/Y_nFGh5UCi30NzRWOAam0gsVpQrjDXZX/tokens'
```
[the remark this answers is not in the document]
prettyirrelevant: yeah, it's intentional -- it's scoped to just their prices api.

> Maybe make it a class attribute so you can do

okay.

rotkehlchen/globaldb/upgrades/manager.py
prettyirrelevant: it does not affect it. since there's no breaking change in the assets schema.

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,424 @@
...
+        if isinstance(asset_identifier, str):
+            payload['symbol'] = asset_identifier
+        else:  # can only be dict
+            payload.update(asset_identifier)
+
+        result = self._query(
+            verb='POST',
+            endpoint='historical',
+            payload=payload,
+        )
```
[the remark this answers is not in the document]
prettyirrelevant: remote error is handled in `PriceHistorian.query_historical_price()`

rotkehlchen/externalapis/alchemy.py
prettyirrelevant: only usd. however, it's an array of prices so doing a `prices[0]['value']` increases the risk of surprise if they add a new fiat equivalent

rotkehlchen/externalapis/alchemy.py
prettyirrelevant: the default frequency is 1d. that's why i didn't include in the payload.

> But also, do we want to get the average of prices? Why not set a 1h window around the timestamp that we want? one day seems to much

yeah, sure.

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,424 @@
...
+            from_asset = from_asset.resolve_to_asset_with_oracles()
+            to_asset = to_asset.resolve_to_asset_with_oracles()
+        except UnknownAsset as e:
+            raise PriceQueryUnsupportedAsset(e.identifier) from e
+
+        price_cache_entry = GlobalDBHandler.get_historical_price(
+            from_asset=from_asset,
+            to_asset=to_asset,
+            timestamp=timestamp,
+            max_seconds_distance=DAY_IN_SECONDS,
```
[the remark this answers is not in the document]
prettyirrelevant: i think it's because of how the get_historical_price works? 
if `max_seconds_distance` meant the price within that day, that's acceptable right?

or should i reduce it to 6 hours?

rotkehlchen/externalapis/alchemy.py
prettyirrelevant: not at all, the schemas use `HistoricalPriceOracleField` so that handles serialisation and deserialisation.

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,422 @@
...
+
+        if 'symbol' in asset_identifier:
+            result = self._query(
+                verb='GET',
+                payload={'symbols': asset_identifier['symbol']},
+                endpoint='by-symbol',
+            )
+        else:
+            result = self._query(
+                verb='POST',
```
yabirgb, line 86: you never call the same endpoint both with get and post right? If so make the verb a variable in self._query depending on the endpoint argument

rotkehlchen/externalapis/alchemy.py
yabirgb: This can raise KeyError. Can you state it in the docstring and handle where this is called?

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,422 @@
...
+                'address' in asset_identifier and
+                result['network'] != asset_identifier['network'] and
+                result['address'] != asset_identifier['address']
+            )
+        ):
+            log.warning(
+                f'Queried Alchemy API from {from_asset.identifier} to {to_asset.identifier}, '
+                f'but received mismatched symbol in response',
+            )
+            return ZERO_PRICE
```
yabirgb, line 322: this logic seems very similar to the one in the other function. Can't this be abstracted somehow?

rotkehlchen/tests/conftest.py
```diff
@@ -218,6 +218,20 @@ def beaconchain_matcher(r1, r2):
...
+        if (
+            (base_url := 'https://api.g.alchemy.com/prices/v1/') and
+            r1.uri.startswith(base_url) and
+            r2.uri.startswith(base_url)
+        ):
+            # Extract everything after API key by finding second '/' after base_url
+            path1 = r1.uri[r1.uri.find('/', len(base_url)) + 1:]
+            path2 = r2.uri[r2.uri.find('/', len(base_url)) + 1:]
+            return path1 == path2 and r1.method == r2.method
+        return r1.uri == r2.uri and r1.method == r2.method
```
yabirgb, line 232:
```suggestion
    def alchemy_api_matcher(r1, r2):
        """Match Alchemy price API paths, ignoring API key."""

        if (
            r1.uri.startswith('https://api.g.alchemy.com/prices/v1/') and
            r2.uri.startswith('https://api.g.alchemy.com/prices/v1/')
        ):
            # Extract everything after API key by finding second '/' after base_url
            path1 = r1.uri[r1.uri.find('/', len(base_url)) + 1:]
            path2 = r2.uri[r2.uri.find('/', len(base_url)) + 1:]
            return path1 == path2 and r1.method == r2.method
        return r1.uri == r2.uri and r1.method == r2.method
```

hardcode it since it is faster than assigning a variable

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,422 @@
...
+
+    @overload
+    def _query(
+            self,
+            verb: Literal['GET'],
+            payload: dict[str, Any],
+            endpoint: Literal['by-symbol'],
+    ) -> dict[str, Any]: ...
```
yabirgb, line 358: making the verb depend on endpoint you can avoid those overloads

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,422 @@
...
+    def _get_asset_identifier(asset: AssetWithOracles) -> dict[str, str]:
+        """
+        May raise:
+        - UnsupportedAsset
+        """
+        if asset.is_evm_token() is True:
+            asset = asset.resolve_to_evm_token()
+            if asset.chain_id == ChainID.POLYGON_POS:
+                network = 'polygon-mainnet'
+            elif asset.chain_id == ChainID.ARBITRUM_ONE:
```
yabirgb, line 209: can you check in a small example if using `match`  warns you when you don't check all the cases of an enum?

rotkehlchen/tests/conftest.py
prettyirrelevant: it's used in the lines below, to define `path1` and `path2`.

rotkehlchen/externalapis/alchemy.py
prettyirrelevant:
```py
from enum import Enum

class MyEnum(Enum):
    A = 1
    B = 2


def is_a(x: MyEnum):
    match x:
        case MyEnum.A:
            return True

a = is_a(MyEnum.B)
print(f'{a=}'). # a=None
```

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,396 @@
...
+        url = f'https://api.g.alchemy.com/prices/v1/{api_key}/tokens/{endpoint}'
+        log.debug(f'Querying alchemy api: {url=} with {payload=}')
+        try:
+            response = self.session.request(
+                method=(verb := 'GET' if endpoint == 'by-symbol' else 'POST'),
+                url=url,
+                params=payload if verb == 'GET' else None,
+                json=payload if verb == 'POST' else None,
+                timeout=CachedSettings().get_timeout_tuple(),
+            )
```
yabirgb, line 372: your are checking 3 times already the verb. Instead do 

```
if endpoint == 'by-symbol':
    verb = 'GET'
    params, json_body = payload, None
else:
    verb = 'POST' 
    params, json_body = None, payload

File: rotkehlchen/externalapis/alchemy.py [hunk 19]
```diff
@@ -0,0 +1,400 @@
...
+            self,
+            from_asset: Asset,  # pylint: disable=unused-argument
+            to_asset: Asset,
+            timestamp: Timestamp,
+            seconds: int | None = None,
+    ) -> bool:
+        return not self.is_penalized()
+
+    @staticmethod
+    def _get_asset_identifier(asset: AssetWithOracles) -> dict[str, str]:
```
LefterisJP, line 196: Please rename this method. In our code `asset_identifier` is something else. Also why return a dict? This is a named tuple that this function returns. Network / address. As such I would name this something like `_alchemy_asset_data`

Moreover with a named_tuple you can have a function that implements this logic on the named tuple itself.

```python
        if 'symbol' in asset_identifier:
            result = self._query(
                payload={'symbols': asset_identifier['symbol']},
                endpoint='by-symbol',
            )
        else:
            result = self._query(
                payload={'addresses': [asset_identifier]},
                endpoint='by-address',
            )
```

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,400 @@
...
+            ) from e
+
+        payload: dict[str, str | int] = {
+            'startTime': timestamp - HOUR_IN_SECONDS,
+            'endTime': timestamp + HOUR_IN_SECONDS,
+            'interval': '1h',
+        }
+        result = self._query(
+            endpoint='historical',
+            payload={**payload, **asset_identifier},
```
LefterisJP, line 150: I guess if you go with the named tuple approach you would need to implement the ** operator, or do a to_dict() function.

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,400 @@
...
+        except (DeserializationError, KeyError, ValueError) as e:
+            error_msg = f'missing key {e!s}' if isinstance(e, KeyError) else str(e)
+            log.warning(
+                f'Queried Alchemy Price API from {from_asset.identifier} to {to_asset.identifier}. '  # noqa: E501
+                f'Failed to deserialize historical price data due to: {error_msg}',
+            )
+            return ZERO_PRICE
+
+    @staticmethod
+    def _is_identifier_match(
```
LefterisJP, line 335: This should also be a function of the `AlchemyAssetData` named tuple / data class I said you should make.

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,400 @@
...
+            to_asset: Asset,
+    ) -> Price:
+        """Deserializes current price data from Alchemy API response.
+
+        If price data is missing, invalid, or there's an error, ZERO_PRICE is returned.
+        Otherwise, the deserialized price is returned.
+        """
+        try:
+            for entry in result:
+                if self._is_identifier_match(entry, asset_identifier) is False:
```
LefterisJP, line 250: For boolean stuff keep it simple. I won't repeat further down, but please do it in all other places

```suggestion
                if not self._is_identifier_match(entry, asset_identifier):
```

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,400 @@
...
+                    continue
+
+                if 'error' in entry:
+                    log.warning(
+                        f'Queried Alchemy Price API from {from_asset.identifier} to {to_asset.identifier}. '  # noqa: E501
+                        f'API returned error: {entry["error"]}',
+                    )
+                    return ZERO_PRICE
+
+                if not (prices := entry['prices']):
```
LefterisJP, line 260: So this is a null check? Then here be explicit. Or can it be either `''` or `null` ?

```suggestion
                if not (prices := entry['prices']):
```

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,400 @@
...
+                        f'but received no prices in response',
+                    )
+                    return ZERO_PRICE
+
+                for price in prices:
+                    if price['currency'] != 'usd':
+                        continue
+
+                    return deserialize_price(price['value'])
+        except (DeserializationError, KeyError) as e:
```
LefterisJP, line 272: More readable and logs error if usd is not found.

Also what other currencies are there? Could there be 'eur' for example? If so we would need a smarter matching.

```suggestion
                for price in prices:
                    if price['currency'] == 'usd':
                        return deserialize_price(price['value'])
                else:  # usd price not found
                    log.error('bla bla')

        except (DeserializationError, KeyError) as e:
```

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,400 @@
...
+                    f'but received no historical price data in response',
+                )
+                return ZERO_PRICE
+
+            closest_price = min(
+                data,
+                key=lambda x: abs(timestamp - int(datetime.fromisoformat(x['timestamp']).timestamp())),  # noqa: E501
+            )
+            return deserialize_price(closest_price['value'])
+        except (DeserializationError, KeyError, ValueError) as e:
```
LefterisJP, line 326:
```suggestion

        except (DeserializationError, KeyError, ValueError) as e:
```

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,400 @@
...
+                return ZERO_PRICE
+
+            if not (data := result.get('data', [])):
+                log.warning(
+                    f'Queried Alchemy API from {from_asset.identifier} to {to_asset.identifier}, '
+                    f'but received no historical price data in response',
+                )
+                return ZERO_PRICE
+
+            closest_price = min(
```
LefterisJP, line 321: Hmm isn't there an api that can return a specific price for a time and do this calculation on the server?

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,400 @@
...
+                data,
+                key=lambda x: abs(timestamp - int(datetime.fromisoformat(x['timestamp']).timestamp())),  # noqa: E501
```
LefterisJP, line 323: Can't you use our own `iso8601ts_to_timestamp` function here?

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,400 @@
...
+        May raise:
+        - RemoteError
+        """
+        if (api_key := self._get_api_key()) is None:
+            api_key = self.default_api_key
+
+        url = f'https://api.g.alchemy.com/prices/v1/{api_key}/tokens/{endpoint}'
+        log.debug(f'Querying alchemy api: {url=} with {payload=}')
+        try:
+            if endpoint == 'by-symbol':
```
LefterisJP, line 364: no need to have this if check in every query. Just put the verb and the params as part of the `_query()` function. It's not used in that many places anyway and will save you an if check in every single API call

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,400 @@
...
+            self.penalty_info.note_failure_or_penalize()
+            raise RemoteError(f'Alchemy API request failed due to {e!s}') from e
+
+        if response.status_code == HTTPStatus.TOO_MANY_REQUESTS:
+            self.last_rate_limit = ts_now()
+            msg = f'Got rate limited by Alchemy API querying {url}'
+            log.warning(msg)
+            raise RemoteError(message=msg, error_code=HTTPStatus.TOO_MANY_REQUESTS)
+
+        if response.status_code != 200:
```
LefterisJP, line 387:
```suggestion
        if response.status_code != HTTPStatus.OK:
```

rotkehlchen/tests/external_apis/test_alchemy.py
```diff
@@ -0,0 +1,124 @@
...
+
+    price = session_alchemy.query_current_price(
+        from_asset=dai,
+        to_asset=usd,
+    )
+    assert price == Price(FVal('0.9990162355'))
+
+
+@pytest.mark.parametrize('should_mock_price_queries', [False])
+def test_alchemy_api_errors(
```
LefterisJP, line 80: this and the next test seem to be querying the live API. Is this intended? They will be slow and could end up being flaky

LefterisJP (commented):
Left some comments

rotkehlchen/externalapis/alchemy.py
prettyirrelevant: no, it's to check if the array is empty

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,400 @@
...
+            elif asset.chain_id == ChainID.FANTOM:
+                network = 'fantom-mainnet'
+            elif asset.chain_id == ChainID.ARBITRUM_NOVA:
+                network = 'arbnova-mainnet'
+            else:
+                raise UnsupportedAsset(asset.identifier)
+
+            return {'network': network, 'address': asset.evm_address}
+
+        # Alchemy team recommends using standard ticker symbols (e.g. ETH)
```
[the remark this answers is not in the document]
prettyirrelevant: so i checked again and i got two responses.
- from alchemy's official support on discord.
image

- from their public support channel on discord
image


as a result, i just removed the comment so as not to confuse anyone.

rotkehlchen/externalapis/alchemy.py
prettyirrelevant: > Also what other currencies are there? 

Just usd, but it's an array meaning they could add other fiat assets in future. that's why the guard was added.

rotkehlchen/externalapis/alchemy.py
prettyirrelevant: not at all, they just do a dump of prices within the timestamp range.

example response.
```json
{
  "symbol": "ETH",
  "currency": "usd",
  "data": [
    {
      "value": "390.7589072038",
      "timestamp": "2020-08-10T00:07:04.612Z"
    },
    {
      "value": "391.3763433072",
      "timestamp": "2020-08-10T01:01:49.650Z"
    },
    {
      "value": "394.3572783286",
      "timestamp": "2020-08-10T02:01:06.004Z"
    },
    {
      "value": "395.9425126734",
      "timestamp": "2020-08-10T03:05:15.241Z"
    },
    {
      "value": "396.6322602948",
      "timestamp": "2020-08-10T03:09:35.556Z"
    },
    {
      "value": "398.2669178678",
      "timestamp": "2020-08-10T04:07:46.502Z"
    },
    {
      "value": "395.6788589001",
      "timestamp": "2020-08-10T05:03:25.204Z"
    },
    {
      "value": "394.8208471427",
      "timestamp": "2020-08-10T06:01:41.227Z"
    },
    {
      "value": "396.6927635678",
      "timestamp": "2020-08-10T07:07:38.701Z"
    },
    {
      "value": "395.0491919377",
      "timestamp": "2020-08-10T08:03:13.376Z"
    },
    {
      "value": "395.4452934078",
      "timestamp": "2020-08-10T09:05:20.881Z"
    },
    {
      "value": "394.0033493067",
      "timestamp": "2020-08-10T10:04:06.128Z"
    },
    {
      "value": "390.2548575938",
      "timestamp": "2020-08-10T11:05:00.035Z"
    },
    {
      "value": "391.3692533564",
      "timestamp": "2020-08-10T12:04:39.017Z"
    },
    {
      "value": "394.6345715121",
      "timestamp": "2020-08-10T13:01:21.466Z"
    },
    {
      "value": "395.5329727927",
      "timestamp": "2020-08-10T14:04:20.446Z"
    },
    {
      "value": "394.6658622322",
      "timestamp": "2020-08-10T15:04:11.781Z"
    },
    {
      "value": "396.3504971348",
      "timestamp": "2020-08-10T16:02:21.482Z"
    },
    {
      "value": "394.1603213833",
      "timestamp": "2020-08-10T17:04:41.934Z"
    },
    {
      "value": "396.7690749991",
      "timestamp": "2020-08-10T18:03:26.964Z"
    },
    {
      "value": "396.2437119379",
      "timestamp": "2020-08-10T19:03:39.084Z"
    },
    {
      "value": "395.1280805008",
      "timestamp": "2020-08-10T20:03:43.052Z"
    },
    {
      "value": "395.2599481976",
      "timestamp": "2020-08-10T21:07:11.619Z"
    },
    {
      "value": "393.7527804345",
      "timestamp": "2020-08-10T22:01:29.623Z"
    },
    {
      "value": "394.1691444011",
      "timestamp": "2020-08-10T23:04:17.614Z"
    }
  ]
}```

File: rotkehlchen/tests/external_apis/test_alchemy.py (hunk 30 above)
Comment (prettyirrelevant) [reply rotkehlchen/tests/external_apis/test_alchemy.py:80]: yeah, i'd have vcr-ed it but vcr does not store 4xx responses

plus, to have an heads-up in case their schema changes.

File: rotkehlchen/externalapis/alchemy.py [hunk 32]
```diff
@@ -0,0 +1,421 @@
...
+from rotkehlchen.utils.network import create_session
+
+if TYPE_CHECKING:
+    from rotkehlchen.db.dbhandler import DBHandler
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+class AlchemyAssetData(NamedTuple):
```
LefterisJP, line 36: add a docstring.

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,421 @@
...
+        )) == ZERO_PRICE:
+            raise NoPriceForGivenTimestamp(
+                from_asset=from_asset,
+                to_asset=to_asset,
+                time=timestamp,
+                rate_limited=False,
+            )
+
+        if to_asset != A_USD:
+            # we need to query intermediate price in this case.
```
LefterisJP, line 204:
```suggestion
        if to_asset != A_USD:  # we need to query intermediate price in this case.
```

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,421 @@
...
+                        f'Queried Alchemy Price API from {from_asset.identifier} to {to_asset.identifier}, '  # noqa: E501
+                        f'but received no prices in response',
+                    )
+                    return usd_price
+
+                for price in prices:
+                    if price['currency'] == 'usd':
+                        usd_price = deserialize_price(price['value'])
+                        break
+                else:  # usd price not found
```
LefterisJP, line 310:
```suggestion
                        break

                else:  # usd price not found
```

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,421 @@
...
+                    log.error(
+                        f'Queried Alchemy Price API from {from_asset.identifier} to '
+                        f'{to_asset.identifier}, but received non-USD price response',
+                    )
```
LefterisJP, line 314:
```suggestion
                    )

```

rotkehlchen/externalapis/alchemy.py
```diff
@@ -0,0 +1,421 @@
...
+        except (DeserializationError, KeyError) as e:
+            error_msg = f'missing key {e!s}' if isinstance(e, KeyError) else str(e)
+            log.warning(
+                f'Queried Alchemy Price API from {from_asset.identifier} to {to_asset.identifier}. '  # noqa: E501
+                f'Failed to deserialize price data due to: {error_msg}',
+            )
+            return usd_price
+        else:
+            return usd_price
```
LefterisJP, line 323: Don't you return usd_price always? Which is equivalent to this?

But isn't there a bug here?

```suggestion
        finally:
            return usd_price
```

rotkehlchen/tests/external_apis/test_alchemy.py
LefterisJP: The heads up if the schema changes in error responses is not important.

You can definitely modify vcr through a fixture to record 4xx responses if it does not do it by default. Though we do have cassettes with 404 response for example: https://github.com/rotki/test-caching/blob/08df6cdbbad13ebd63d0af58b7fc335749f7eb94/cassettes/unit/test_calendar_reminders/test_ens_expiry_calendar_reminders%5Bens_data2-base_accounts0-ethereum_accounts0%5D.yaml#L1081

rotkehlchen/tests/external_apis/test_alchemy.py
```diff
@@ -0,0 +1,127 @@
...
+    )
+
+    # fetch the current price of a token that is not supported.
+    price = session_alchemy.query_current_price(
+        from_asset=sample_token,
+        to_asset=usd,
+    )
+    assert price == ZERO_PRICE
+
+    # fetch the historical price of a token that is not supported.
```
LefterisJP, line 100: seems this test is not vcred properly or not pushed into the vcr branch

LefterisJP (approved):
LGTM

PR #9196, opened 2025-01-03
Refactor normalize function

yabirgb:
Closes #(issue_number)

## Checklist


rotkehlchen/chain/ethereum/utils.py
```diff
@@ -41,29 +41,59 @@
...
+        amount: int | FVal,
+        decimals: int | None,
+        as_int: Literal[True],
+) -> int:
+    ...
 
-    return token_amount / (FVal(10) ** FVal(token_decimals))
 
+@overload
+def _normalize_value(
```
LefterisJP, line 54: Would it not make more sense to have two different functions to avoid the extra if check? Since this is a function deep in our call stack that's used a lot. Some something like normalize_fval and normalize_int

rotkehlchen/chain/ethereum/utils.py
```diff
@@ -41,29 +41,59 @@
...
-    return token_normalized_value_decimals(token_amount, token.decimals)
+def token_normalized_value_decimals(token_amount: int, token_decimals: int | None) -> FVal:
+    return _normalize_value(amount=token_amount, decimals=token_decimals, as_int=False)
+
+
+def normalized_fval_value_decimals(amount: FVal, decimals: int) -> FVal:
+    return _normalize_value(amount=amount, decimals=decimals, as_int=False)
+
+
+def asset_normalized_value(amount: int, asset: CryptoAsset | EvmToken) -> FVal:
```
LefterisJP, line 85: Again here perhaps since this is a function used multiple times many times deep in the logic why not make two functions? asset_normalized_value and token_normalized_value as it was before to avoid the if?

rotkehlchen/chain/ethereum/modules/zksync/decoder.py
```diff
@@ -53,10 +53,10 @@ def _decode_deposit(self, context: DecoderContext, user_address: ChecksumEvmAddr
         for event in context.decoded_events:
             if event.event_type == HistoryEventType.SPEND and event.location_label == user_address:
                 resolved_event_asset = event.asset.resolve_to_crypto_asset()
-                event_raw_amount = asset_raw_value(
+                event_raw_amount = normalized_fval_value_decimals(
                     amount=event.balance.amount,
-                    asset=resolved_event_asset,
-                )
+                    decimals=-get_decimals(resolved_event_asset),
```
LefterisJP, line 58: why is there a `-` here?

rotkehlchen/chain/evm/decoding/extrafi/balances.py
```diff
@@ -151,7 +151,7 @@ def _query_farm_positions(
                 if debt_amount == 0:
                     continue
 
-                amount = token_normalized_value(debt_amount, debt_token)
+                amount = asset_normalized_value(debt_amount, debt_token)
```
LefterisJP, line 154: since you touched so many places how about introducing kwargs in places they are missing it?

rotkehlchen/chain/ethereum/modules/zksync/decoder.py
yabirgb: this does the opposite. instead of `/ 10**18` it does `/10**-18 = * 10**-18`. What the function wants is to check that `1800000 == 1.8 * 10**6`

LefterisJP: I don't get it :sweat_smile:

yabirgb:
Closing as abandoned. Might reopen if it is worth

PR #9205, opened 2025-01-05
Fix an always true elif in node_inquirer logs query handling

LefterisJP:
The bug was introduced here: https://github.com/rotki/rotki/commit/1a3d7536ab94d7d470cdd816d1cd861b99042771


yabirgb (approved):
lgtm

PR #9209, opened 2025-01-06
more nexo importer fixes

prettyirrelevant:


rotkehlchen/data_import/importers/nexo.py
```diff
@@ -80,16 +84,20 @@ def _consume_nexo(
...
-        asset = asset_from_nexo(csv_row['Output Currency'])
-        amount = deserialize_asset_amount_force_positive(csv_row['Output Amount'])
         entry_type = csv_row['Type']
         transaction = csv_row['Transaction']
+        amount = deserialize_asset_amount_force_positive(csv_row['Output Amount'])
+        try:
+            asset = asset_from_nexo(csv_row['Output Currency'])
+        except UnknownAsset:
+            # from the csv a user shared, some entries with type of
+            # interest and referral bonus did not contain an output currency
```
LefterisJP, line 94: if they did not contain an output currency isn't iut best to do csv_row.get('Output Currency') and compare with None? Instead of calling the entire function?

rotkehlchen/data_import/importers/nexo.py
```diff
@@ -128,6 +136,12 @@ def _consume_nexo(
                 log.debug(f'Ignoring nexo entry {csv_row} with negative interest')
                 return
 
+            # from the csv a user shared, some entries with type of
```
LefterisJP, line 139: this seems like a copy pasta comment from above. There is no output amount here?

rotkehlchen/data_import/importers/nexo.py
```diff
@@ -157,6 +171,35 @@ def _consume_nexo(
                 notes=f'{entry_type} from Nexo',
             )
             self.add_history_events(write_cursor, [event])
+        elif entry_type == 'Exchange':
+            input_asset = asset_from_nexo(csv_row['Input Currency'])
```
LefterisJP, line 175: All of these seems to also be able to raise KeyError all over the function right? The docstring does not mention it may raise it. Should it not mention it? Is it handled anywhere?

rotkehlchen/tests/api/test_data_import.py
```diff
@@ -215,6 +219,12 @@ def test_data_import_nexo(
     dir_path = Path(__file__).resolve().parent.parent
     filepath = dir_path / 'data' / 'nexo.csv'
 
+    # add location mapping for luna.
```
LefterisJP, line 222: what's this?

prettyirrelevant: so from the entry, unknown asset is raised due to a missing mapping -- that's what i added it here.

there's a pr to update location mappings too.

rotkehlchen/data_import/importers/nexo.py
prettyirrelevant: key error is handled at here.
https://github.com/rotki/rotki/blob/5d9704524cdf87965c988868ea64e7632a0b5ead/rotkehlchen/data_import/importers/nexo.py#L168-L207

rotkehlchen/data_import/importers/nexo.py
prettyirrelevant: yeah, from the csv. output amount & output currency are always empty for such cases.

rotkehlchen/data_import/importers/nexo.py
prettyirrelevant: ah yes, thanks.

rotkehlchen/data_import/importers/nexo.py
```diff
@@ -128,6 +133,12 @@ def _consume_nexo(
                 log.debug(f'Ignoring nexo entry {csv_row} with negative interest')
                 return
 
+            # from the csv a user shared, some entries with type of
+            # interest and referral bonus did not contain an output amount
+            # so in such cases, fall back to the input amount.
+            if entry_type in {'Interest', 'Referral Bonus'} and amount == ZERO:
```
LefterisJP, line 139: Apart from the comment what else makes no sense is this entire elif code.

The comment says in some cases the OutputAmount is empty. And goes to mention again this `'Interest'` and `'Referral Bonus'` and in the case of which you should take and force positive the `Input Amount`.

But ... the comment said in these cases OutputAmount is missing. You later said, and the code you wrote assumes it's not missing but it's empty. In both cases `deserialize_asset_amount(csv_row['Output Amount'])` will raise a ValueError and we will never get here.

So this seems completely broken.

Did you by any chance mean here that in those cases the output amount is zero (and not missing, '', or "did not contain an output amount) ?

rotkehlchen/tests/api/test_data_import.py
LefterisJP: The entry? What do you mean? What entry? Please can you discuss things in discord and communicate faster?

If you mean the test data you are using, given by the user? Isn't the solution to add the mapping to the global DB and not manually in a single test?

rotkehlchen/tests/api/test_location_asset_mappings.py
```diff
@@ -12,7 +12,7 @@
 if TYPE_CHECKING:
     from rotkehlchen.api.server import APIServer
 
-NUM_ASSETS_MAPPINGS_V1_37: Final = 2023
+NUM_ASSETS_MAPPINGS_V1_37: Final = 2024
```
LefterisJP, line 15: Hmmm this variable name is now wrong. Is it supposed to be `NUM_SHIPPED_ASSETS_MAPPING`, so the number of the included global.db? Or the one in 1.37?

LefterisJP (approved):
LGTM

PR #9211, opened 2025-01-06
Save calls for free coingecko history api call more than 1 year ago

LefterisJP:
Because this is what happens. They stop it :(


yabirgb (approved):
lgtm

PR #9215, opened 2025-01-07
refactor addressbook to handle duplicate entries via upsert

prettyirrelevant:
Resolves 
image


lukicenturi (approved):
Works as expected for me

PR #9220, opened 2025-01-08
Introduce pyright with only reportUndefinedVariable rule

LefterisJP:
This rule alone has found multiple bugs in the past and even now, among the fixes is multiple bugs. Almost always with paths that are not going to be visited often.

### DIFF
--- a/rotkehlchen/tests/utils/ethereum.py
+++ b/rotkehlchen/tests/utils/ethereum.py
@@ -228,7 +228,7 @@ def setup_ethereum_transactions_test(
         block_number=13142218,
         from_address=TEST_ADDR1,
         to_address=string_to_evm_address('0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D'),
-        value=int(10 * 10**18),
+        value=10 * 10**18,
         gas=194928,
         gas_price=int(0.000000204 * 10**18),
         gas_used=136675,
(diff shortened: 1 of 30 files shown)


rotkehlchen/tests/utils/ethereum.py
```diff
@@ -228,7 +228,7 @@ def setup_ethereum_transactions_test(
         block_number=13142218,
         from_address=TEST_ADDR1,
         to_address=string_to_evm_address('0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D'),
-        value=int(10 * 10**18),
+        value=(10 * 10**18),
```
yabirgb, line 231: is the parenthesis needed?

rotkehlchen/usage_analytics.py
```diff
@@ -25,6 +25,15 @@ class GeolocationData(NamedTuple):
     country_code: str
 
 
+def _handle_retrieval_error(geoip_dir: Path, message: str) -> Path | None:
+    log.error(message)
+    old_files = list(geoip_dir.glob('*.mmdb'))
+    if len(old_files) == 0:
```
yabirgb, line 31:
```suggestion
    if len(old_files := list(geoip_dir.glob('*.mmdb'))) == 0:
```

rotkehlchen/tests/utils/ethereum.py
LefterisJP: no :rofl:

yabirgb (approved):
lgtm

PR #9223, opened 2025-01-08
add historical balance query endpoints

prettyirrelevant:
## Overview
Adds new API endpoints to query historical asset balances at specific timestamps. 

## Endpoints
- `GET /balances/historical/{timestamp}`: Returns balances for all assets at timestamp
- `POST /balances/historical`: Returns balance for a specific asset at timestamp

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -13866,3 +13866,88 @@ Managing calendar reminders
   :statuscode 200: Data queried correctly.
   :statuscode 401: No user is currently logged in.
   :statuscode 500: Internal rotki error.
+
+Historical Balance Queries
+==============================
+
+  .. http:post:: /api/(version)/balances/historical
+
+    Gets historical balance data at a specific timestamp, calculated from processing of historical events.
+    If asset is provided, returns balance for that specific asset. Otherwise returns balances for all assets.
+
+    .. note::
+      This endpoint can also be queried asynchronously by using ``"async_query": true``.
+
+    **Example Request (All Assets):**
+
+      .. http:example:: curl wget httpie python-requests
+
+        PUT /api/(version)/balances/historical HTTP/1.1
+        Host: localhost:5042
+        Content-Type: application/json;charset=UTF-8
+
+        { "timestamp": 1672531200 }
+
+    **Example Request (Single Asset):**
+
+      .. http:example:: curl wget httpie python-requests
+
+      PUT /api/(version)/balances/historical HTTP/1.1
+      Host: localhost:5042
+      Content-Type: application/json;charset=UTF-8
+
+      {
+        "timestamp": 1672531200,
+        "asset": "BTC"
+      }
+
+      :reqjsonarr integer timestamp: The timestamp to query the balance for
+      :reqjsonarr string asset: (Optional) The asset identifier to query balance for. If not provided, returns balances for all assets.
+
+    **Example Response (All Assets):**
+
+      .. sourcecode:: http
+
+        HTTP/1.1 200 OK
+        Content-Type: application/json
+
+        {
+          "message": "",
+          "result": {
+            "BTC": {
+              "amount": "2.0",
+              "price": "20000"
+            },
+            "ETH": {
+              "amount": "10.0",
+              "price": "1500"
+            }
+          },
+          "status_code": 200
+        }
+
+    **Example Response (Single Asset):**
+
+      .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+        "message": "",
+        "result": {
+          "amount": "2.0",
+          "price": "20000"
+        },
+        "status_code": 200
+      }
+
+      :resjson object result: Either a mapping of all assets to their balance info, or single asset balance info if asset was provided
+      :resjson string amount: The amount of the asset held at the timestamp
+      :resjson string price: The price of the asset at the timestamp in the user's profit currency
+      :statuscode 200: Historical balances returned
+      :statuscode 400: Malformed query
+      :statuscode 404: No historical data found for the timestamp
+      :statuscode 403: User does not have premium access
+      :statuscode 409: User is not logged in
+      :statuscode 500: Internal Rotki error
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -54,6 +54,7 @@ from rotkehlchen.assets.asset import (
 )
 from rotkehlchen.assets.resolver import AssetResolver
 from rotkehlchen.assets.types import ASSET_TYPES_EXCLUDED_FOR_USERS, AssetType
+from rotkehlchen.balances.historical import HistoricalBalancesManager
 from rotkehlchen.balances.manual import (
     ManuallyTrackedBalance,
     add_manually_tracked_balances,
@@ -170,6 +171,7 @@ from rotkehlchen.errors.misc import (
     GreenletKilledError,
     InputError,
     ModuleInactive,
+    NotFoundError,
     RemoteError,
     SystemPermissionError,
     TagConstraintError,
@@ -5353,3 +5355,33 @@ class RestAPI:
         )
         stats['score'] = score
         return api_response(_wrap_in_ok_result(result=stats, status_code=HTTPStatus.OK))
+
+    @async_api_call()
+    def get_historical_balance(self, timestamp: Timestamp) -> dict[str, Any]:
+        """Query historical balances for all assets at a given timestamp
+        by processing historical events
+        """
+        try:
+            balances = HistoricalBalancesManager(self.rotkehlchen.data.db).get_balances(timestamp)
+        except NotFoundError as e:
+            return wrap_in_fail_result(str(e), status_code=HTTPStatus.NOT_FOUND)
+
+        return _wrap_in_ok_result(
+            result=process_result(balances),
+            status_code=HTTPStatus.OK,
+        )
+
+    @async_api_call()
+    def get_historical_asset_balance(self, asset: Asset, timestamp: Timestamp) -> dict[str, Any]:
+        """Query historical balance of a specific asset at a given timestamp
+        by processing historical events
+        """
+        try:
+            balance = HistoricalBalancesManager(self.rotkehlchen.data.db).get_asset_balance(
+                asset=asset,
+                timestamp=timestamp,
+            )
+        except NotFoundError as e:
+            return wrap_in_fail_result(str(e), status_code=HTTPStatus.NOT_FOUND)
+
+        return _wrap_in_ok_result(result=balance)
--- a/rotkehlchen/api/v1/resources.py
+++ b/rotkehlchen/api/v1/resources.py
@@ -103,6 +103,7 @@ from rotkehlchen.api.v1.schemas import (
     ExternalServicesResourceDeleteSchema,
     FileListSchema,
     HistoricalAssetsPriceSchema,
+    HistoricalPerAssetBalanceSchema,
     HistoryEventSchema,
     HistoryEventsDeletionSchema,
     HistoryExportingSchema,
@@ -3377,4 +3378,29 @@ class StatsWrapResource(BaseMethodView):
         return self.rest_api.query_wrap_stats(
             from_ts=from_timestamp,
             to_ts=to_timestamp,
+        )
+
+
+class TimestampHistoricalBalanceResource(BaseMethodView):
+
+    post_schema = HistoricalPerAssetBalanceSchema()
+
+    @require_premium_user(active_check=False)
+    @use_kwargs(post_schema, location='json')
+    def post(
+            self,
+            async_query: bool,
+            timestamp: Timestamp,
+            asset: Asset | None = None,
+    ) -> Response:
+        if asset is None:
+            return self.rest_api.get_historical_balance(
+                timestamp=timestamp,
+                async_query=async_query,
+            )
+
+        return self.rest_api.get_historical_asset_balance(
+            asset=asset,
+            timestamp=timestamp,
+            async_query=async_query,
         )
--- a/rotkehlchen/balances/historical.py
+++ b/rotkehlchen/balances/historical.py
@@ -1,0 +1,114 @@
+from collections import defaultdict
+from typing import TYPE_CHECKING, TypedDict
+
+from rotkehlchen.assets.asset import Asset
+from rotkehlchen.constants.prices import ZERO_PRICE
+from rotkehlchen.db.filtering import HistoryEventFilterQuery
+from rotkehlchen.db.history_events import DBHistoryEvents
+from rotkehlchen.errors.misc import NotFoundError, RemoteError
+from rotkehlchen.errors.price import NoPriceForGivenTimestamp
+from rotkehlchen.fval import FVal
+from rotkehlchen.history.events.structures.types import EventDirection
+from rotkehlchen.history.price import PriceHistorian
+from rotkehlchen.types import Timestamp
+
+if TYPE_CHECKING:
+    from rotkehlchen.db.dbhandler import DBHandler
+
+
+class HistoricalBalance(TypedDict):
+    amount: FVal
+    price: FVal
+
+
+class HistoricalBalancesManager:
+    """Processes historical events and calculates balances"""
+
+    def __init__(self, db: 'DBHandler') -> None:
+        self.db = db
+
+    def get_balances(self, timestamp: Timestamp) -> dict[Asset, HistoricalBalance]:
+        """Get historical balances for all assets at a given timestamp
+
+        May raise:
+        - NotFoundError if balance for the given timestamp does not exist.
+        """
+        events, main_currency = self._get_events_and_currency(timestamp)
+        if len(events) == 0:
+            raise NotFoundError('No historical data found until the given timestamp.')
+
+        amounts = self._calculate_balance_from_events(events)
+        result: dict[Asset, HistoricalBalance] = {}
+        for asset, amount in amounts.items():
+            try:
+                price = PriceHistorian().query_historical_price(
+                    from_asset=asset,
+                    to_asset=main_currency,
+                    timestamp=timestamp,
+                )
+            except (RemoteError, NoPriceForGivenTimestamp):
+                price = ZERO_PRICE
+
+            result[asset] = {'amount': amount, 'price': price}
+
+        return result
+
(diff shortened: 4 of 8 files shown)


rotkehlchen/balances/historical_balances.py
```diff
@@ -0,0 +1,300 @@
...
+        self.db = database
+
+    def get_balances(self) -> tuple[list[Timestamp], list[str]]:
+        """Retrieve historical usd balances until now, calculating cumulative sums.
+
+        When a timestamp has zero balance, we use the last non-zero value but don't
+        include it in the cumulative sum calculation.
+        """
+        if not (daily_values := self._get_daily_values(
+            query='SELECT timestamp, usd_value, usd_value FROM historical_balance_per_asset',
```
LefterisJP, line 52: So I see you created a new DB table. Didn't we have a discussion in the standup to not do exactly that? Please if we discuss something stick to it, or ask earlier about the approach to make sure we are in the same page.

Communicate, communicate, communicate.

rotkehlchen/balances/historical_balances.py
```diff
@@ -0,0 +1,300 @@
...
+                ),
+                has_premium=True,
+            )
+
+            balances: dict[Timestamp, dict[Asset, Balance]] = defaultdict(
+                lambda: defaultdict(lambda: Balance()),  # noqa: PLW0108
+            )
+            for event in events:
+                if (direction := event.maybe_get_direction()) is None or direction == EventDirection.NEUTRAL:  # noqa: E501
+                    log.debug(
```
LefterisJP, line 194: Isn't it a bit of an overkill to log this? This is expected to happen.

rotkehlchen/balances/historical_balances.py
```diff
@@ -0,0 +1,300 @@
...
+
+            bindings = []
+            for timestamp, asset_balances in balances.items():
+                for asset, balance in asset_balances.items():
+                    if balance.amount == ZERO:
+                        continue
+
+                    if balance.usd_value == ZERO:
+                        try:
+                            usd_price = PriceHistorian().query_historical_price(
```
LefterisJP, line 215: no no no. This should absolutely not happen. You are now turning a DB-local-only operation to one that will be querying historical prices for all assets, for all days. This will never finish. Price should not go in here.

rotkehlchen/db/cache.py
```diff
@@ -26,6 +26,7 @@ class DBCacheStatic(Enum):
     LAST_CREATE_REMINDER_CHECK_TS: Final = 'last_create_reminder_check_ts'
     LAST_GRAPH_DELEGATIONS_CHECK_TS: Final = 'last_graph_delegations_check_ts'
     LAST_GNOSISPAY_QUERY_TS: Final = 'last_gnosispay_query_ts'
+    LAST_HISTORICAL_BALANCES_RECREATION_TS: Final = 'last_historical_balances_recreation_ts'
```
LefterisJP, line 29: So for caching anything we should do it per asset. Not just on a total. Though not sure if the approach of having another table is requird.

rotkehlchen/tasks/manager.py
```diff
@@ -976,6 +978,17 @@ def _maybe_query_graph_delegated_tokens(self) -> Optional[list[gevent.Greenlet]]
             addresses=self.chains_aggregator.accounts.eth,
         )]
 
+    def _maybe_recreate_historical_balances(self) -> Optional[list[gevent.Greenlet]]:
+        if should_run_periodic_task(self.database, DBCacheStatic.LAST_HISTORICAL_BALANCES_RECREATION_TS, HOUR_IN_SECONDS * 3) is False:  # noqa: E501
```
LefterisJP, line 982: This should not be a periodic task that runs every 3 hours. It's way too frequent. And if nothing has changed there is no reason to rerun.
Plus for exmple for rotki's account the way you wrote it right now it would take probably 4-6 weeks to run the task :rofl: 

This should be split into per asset tasks. And only if a new event for a specific asset has been added only balances **after** that event for that asset and for the total should be recalculated.

We already discussed this.

LefterisJP (commented):
Some first comments. I only looked at the `rotkehlchen/balances/historical_balances.py` file.

1. Please communicate the planned approach early on so we can iterate on the feedback. Now you ended up doing something you will need to change. And made tests, DB upgrade, and API on top. A lot of that is now wasted work since you did it before getting any feedback. I have already explained this in multiple previous PRs. Please listen and understand, don't just nod your head say you will keep it in mind and ignore it. You should have done a small PoC of the `rotkehlchen/balances/historical_balances.py` and pinged for feedback. Or write the approach out and pign for feedback.

2. I see you added an extra DB table. As we had discussed this is probably not needed as it just caches values that can be calculated on the fly. Perhaps some form of cache will be needed, but your entire logic is based around that table which seems wrong.
3. You are querying prices. That's a big no no. Imagine how that would end up working. Ever for one asset querying its price for every day would (1) not work for many days (2) take forever, even with paid oracles.
4. The periodic task should not run like this. This can't run periodically so often and in a complete rerun like this. As discussed during planning this, only when a new event is added (can be added in the past), all balances of that event's asset **after** the event and all totals should be recalculated.

rotkehlchen/balances/historical.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def get_asset_balance(self, asset: Asset, timestamp: Timestamp) -> HistoricalBalance:
+        """Get historical balance for a single asset at a given timestamp
+
+        May raise:
+        - InputError if balance for the asset at the given timestamp does not exist.
+        """
+        events, main_currency = self._get_events_and_currency(timestamp, (asset,))
+        if len(events) == 0:
+            raise InputError(f'No historical data found for {asset} at given timestamp.')
```
LefterisJP, line 63: How about introducing a NotFoundError? A new exception? As this is not really an input error. There is just no events present for the asset until the given timestamp

```suggestion
            raise NotFoundError(f'No historical data found for {asset} until the given timestamp.')
```

rotkehlchen/balances/historical.py
```diff
@@ -0,0 +1,113 @@
...
+    amount: FVal
+    price: FVal
+
+
+class HistoricalBalancesManager:
+
+    def __init__(self, db: 'DBHandler') -> None:
+        self.db = db
+
+    def get_balance(self, timestamp: Timestamp) -> dict[Asset, HistoricalBalance]:
```
LefterisJP, line 29:
```suggestion
    def get_balances(self, timestamp: Timestamp) -> dict[Asset, HistoricalBalance]:
```

rotkehlchen/balances/historical.py
```diff
@@ -0,0 +1,113 @@
...
+        """Get historical balances for all assets at a given timestamp
+
+        May raise:
+        - InputError if balance for the given timestamp does not exist.
+        """
+        events, main_currency = self._get_events_and_currency(timestamp)
+        if len(events) == 0:
+            raise InputError('No historical data found for given timestamp.')
```
LefterisJP, line 37:
```suggestion
            raise NotFoundError('No historical data found until the given timestamp.')
```

rotkehlchen/balances/historical.py
```diff
@@ -0,0 +1,113 @@
...
+class HistoricalBalancesManager:
+
+    def __init__(self, db: 'DBHandler') -> None:
+        self.db = db
+
+    def get_balance(self, timestamp: Timestamp) -> dict[Asset, HistoricalBalance]:
+        """Get historical balances for all assets at a given timestamp
+
+        May raise:
+        - InputError if balance for the given timestamp does not exist.
```
LefterisJP, line 33: change it also in the docstrings

rotkehlchen/api/v1/resources.py
```diff
@@ -3378,3 +3379,23 @@ def post(self, from_timestamp: Timestamp, to_timestamp: Timestamp) -> Response:
...
         )
+
+
+class TimestampHistoricalBalanceResource(BaseMethodView):
+
+    query_schema = SnapshotTimestampQuerySchema()
+    post_schema = HistoricalPerAssetBalanceSchema()
+
+    @require_premium_user(active_check=False)
+    @use_kwargs(query_schema, location='view_args')
```
LefterisJP, line 3390: as I wrote in the docs I would keep it json for both

rotkehlchen/api/v1/resources.py
```diff
@@ -3378,3 +3379,23 @@ def post(self, from_timestamp: Timestamp, to_timestamp: Timestamp) -> Response:
...
+    def get(self, timestamp: Timestamp) -> Response:
```
LefterisJP, line 3391: since this also has remote calls it should also be able to be done async. So it's missing the async query argument

rotkehlchen/api/rest.py
```diff
@@ -5353,3 +5354,28 @@ def query_wrap_stats(self, from_ts: Timestamp, to_ts: Timestamp) -> Response:
         )
         stats['score'] = score
         return api_response(_wrap_in_ok_result(result=stats, status_code=HTTPStatus.OK))
+
+    def get_historical_balance(self, timestamp: Timestamp) -> Response:
```
LefterisJP, line 5358: this should also be asyncable

rotkehlchen/api/rest.py
```diff
@@ -5353,3 +5354,28 @@ def query_wrap_stats(self, from_ts: Timestamp, to_ts: Timestamp) -> Response:
...
+        """Query historical balances for all assets at a given timestamp"""
```
LefterisJP, line 5359:
```suggestion
        """Query historical balances for all assets at a given timestamp by processing historical events"""
```

rotkehlchen/api/rest.py
```diff
@@ -5353,3 +5354,28 @@ def query_wrap_stats(self, from_ts: Timestamp, to_ts: Timestamp) -> Response:
...
+            return api_response(wrap_in_fail_result(str(e)), status_code=HTTPStatus.BAD_REQUEST)
+
+        return api_response(
+            result=_wrap_in_ok_result(result=process_result(balances)),
+            status_code=HTTPStatus.OK,
+        )
+
+    @async_api_call()
+    def get_historical_asset_balance(self, asset: Asset, timestamp: Timestamp) -> dict[str, Any]:
+        """Query historical balance of a specific asset at a given timestamp"""
```
LefterisJP, line 5372:
```suggestion
        """Query historical balance of a specific asset at a given timestamp by processing historical events"""
```

rotkehlchen/tests/api/test_historical_balances.py
```diff
@@ -0,0 +1,208 @@
...
+from rotkehlchen.utils.misc import ts_sec_to_ms
+
+if TYPE_CHECKING:
+    from rotkehlchen.api.server import APIServer
+    from rotkehlchen.db.dbhandler import DBHandler
+
+START_TS = Timestamp(1672531200)
+
+
+def create_test_events(db: 'DBHandler') -> None:
```
LefterisJP, line 32:
```suggestion
def _create_test_events(db: 'DBHandler') -> None:
```

rotkehlchen/tests/api/test_historical_balances.py
```diff
@@ -0,0 +1,208 @@
...
+            write_cursor=write_cursor,
+            history=events,
+        )
+
+
+@pytest.fixture(name='setup_historical_data')
+def fixture_setup_historical_data(rotkehlchen_api_server: 'APIServer') -> None:
+    """Fixture to setup historical test data"""
+    db = rotkehlchen_api_server.rest_api.rotkehlchen.data.db
+    create_test_events(db=db)
```
LefterisJP, line 80: Ah wait if this is only used here, just move the function contents here in the fixture

docs/api.rst
```diff
@@ -13866,3 +13866,94 @@ Managing calendar reminders
...
+        }
+      },
+      "status_code": 200
+    }
+
+  :resjson object result: Object mapping asset identifiers to their balance info
+  :resjson string amount: The amount of the asset held at the timestamp
+  :resjson string price: The price of the asset at the timestamp in the user's preferred currency
+  :statuscode 200: Historical balances returned
+  :statuscode 400: Invalid timestamp or malformed query
```
LefterisJP, line 13911: Malformed query yes. But timestamp having no events, is not a bad request. It should be a 404. What we use for no data found.

docs/api.rst
```diff
@@ -13866,3 +13866,94 @@ Managing calendar reminders
...
+        "amount": "2.0",
+        "price": "20000"
+      },
+      "status_code": 200
+    }
+
+  :resjson string amount: The amount of the asset held at the timestamp
+  :resjson string price: The price of the asset at the timestamp in the user's preferred currency
+  :statuscode 200: Historical balance returned
+  :statuscode 400: Invalid timestamp/asset or malformed query
```
LefterisJP, line 13956: Same as above

rotkehlchen/tests/api/test_historical_balances.py
```diff
@@ -0,0 +1,208 @@
...
+    assert result['BTC']['amount'] == '2'
+    assert result['BTC']['price'] == '15519.8178641018'
+    assert result['ETH']['amount'] == '10'
+    assert result['ETH']['price'] == '1120.71447101836'
+
+
+@pytest.mark.vcr
+@pytest.mark.parametrize('start_with_valid_premium', [True])
+@pytest.mark.parametrize('should_mock_price_queries', [False])
+def test_get_historical_balance_skipped_day(
```
LefterisJP, line 108: What's the point of this test? Why not just do a second query on the test above? This way you can combine them.

rotkehlchen/tests/api/test_historical_balances.py
```diff
@@ -0,0 +1,208 @@
...
+    )
+
+    result = assert_proper_sync_response_with_result(response)
+    assert result['BTC']['amount'] == '2'
+    assert result['BTC']['price'] == '15519.8178641018'
+    assert result['ETH']['amount'] == '10'
+    assert result['ETH']['price'] == '1120.71447101836'
+
+
+@pytest.mark.vcr
```
LefterisJP, line 105: Why are you vcring them all? For the price queries only right?

rotkehlchen/api/rest.py
```diff
@@ -5353,3 +5354,28 @@ def query_wrap_stats(self, from_ts: Timestamp, to_ts: Timestamp) -> Response:
         )
         stats['score'] = score
         return api_response(_wrap_in_ok_result(result=stats, status_code=HTTPStatus.OK))
+
+    def get_historical_balance(self, timestamp: Timestamp) -> Response:
+        """Query historical balances for all assets at a given timestamp"""
+        try:
+            balances = HistoricalBalancesManager(self.rotkehlchen.data.db).get_balance(timestamp)
+        except InputError as e:
+            return api_response(wrap_in_fail_result(str(e)), status_code=HTTPStatus.BAD_REQUEST)
```
LefterisJP, line 5363: as I wrote this should not be BAD_REQUEST, but NOT_FOUND

rotkehlchen/api/rest.py
```diff
@@ -5353,3 +5354,28 @@ def query_wrap_stats(self, from_ts: Timestamp, to_ts: Timestamp) -> Response:
...
+    @async_api_call()
+    def get_historical_asset_balance(self, asset: Asset, timestamp: Timestamp) -> dict[str, Any]:
+        """Query historical balance of a specific asset at a given timestamp"""
+        try:
+            balance = HistoricalBalancesManager(self.rotkehlchen.data.db).get_asset_balance(
+                asset=asset,
+                timestamp=timestamp,
+            )
+        except InputError as e:
+            return wrap_in_fail_result(str(e), status_code=HTTPStatus.BAD_REQUEST)
```
LefterisJP, line 5379: as I wrote this should not be BAD_REQUEST, but NOT_FOUND

rotkehlchen/tests/api/test_historical_balances.py
prettyirrelevant: > Why are you vcring them all? 

yeah, because of historical price queries

rotkehlchen/tests/api/test_historical_balances.py
prettyirrelevant: > What's the point of this test?

to check a behaviour that calling retrieving the historical balance at a timestamp that does have event uses the last known event -- added a comment now.

> Why not just do a second query on the test above? This way you can combine them.

sure.

docs/api.rst
```diff
@@ -13866,3 +13866,98 @@ Managing calendar reminders
...
+  :resjson string price: The price of the asset at the timestamp in the user's profit currency
+  :statuscode 200: Historical balances returned
+  :statuscode 400: Malformed query
+  :statuscode 404: No historical data found for the timestamp
+  :statuscode 403: User does not have premium access
+  :statuscode 409: User is not logged in
+  :statuscode 500: Internal Rotki error
+
+
+.. http:post:: /api/(version)/balances/historical
```
LefterisJP, line 13920: Looking at this ... would it not make sense to combine those too then? If asset is missing give snapshot of all, if asset is given only of that asset? And keep it in the `POST` verb?

rotkehlchen/api/v1/resources.py
```diff
@@ -3378,3 +3380,26 @@ def post(self, from_timestamp: Timestamp, to_timestamp: Timestamp) -> Response:
...
         )
+
+
+class TimestampHistoricalBalanceResource(BaseMethodView):
+
+    query_schema = HistoricalBalanceSchema()
+    post_schema = HistoricalPerAssetBalanceSchema()
+
+    @require_premium_user(active_check=False)
+    @use_kwargs(query_schema, location='json')
```
LefterisJP, line 3391: Though with the new request this would go away

```suggestion
    @use_kwargs(put_schema, location='json')
```

rotkehlchen/tests/api/test_historical_balances.py
```diff
@@ -0,0 +1,198 @@
...
+from rotkehlchen.utils.misc import ts_sec_to_ms
+
+if TYPE_CHECKING:
+    from rotkehlchen.api.server import APIServer
+    from rotkehlchen.db.dbhandler import DBHandler
+
+START_TS = Timestamp(1672531200)
+
+
+def _create_test_events(db: 'DBHandler') -> None:
```
LefterisJP, line 34: No I did not mean this. I meant to remove this function and instead keep the fixture. Just to move this function's code in the fixture, and use the fixture in the tests

prettyirrelevant: okay, reverted that.

LefterisJP (approved):
LGTM

PR #9224, opened 2025-01-08
Enable pyright operator issue rule

LefterisJP:


LefterisJP:
rerunning one suite cause I did not push a cassette

Issue #1573, opened 2020-10-07
BC/BSC support for Blockchain Balances


mariodigital:
Please add support for Binance Chain / Binance Smart Chain Blockchain addresses to automatically accumulate your balances under the dashboard.

petermllrr:
🙏

qqq-jump:
+1

mariodigital:
Any day now 😅

LefterisJP:
Hey guys. Thanks for +1 on this issue. We don't really know when developer resources can be dedicated to do this issue right.

Rotki is opensource and PRs are greatly appreciated.

The only open API I could find is etherscan's bscscan: https://bscscan.com/apis#accounts
But that's only for BNB balances.

gufertum:
> But that's only for BNB balances.

AFAIK it's a clone of etherscan.io and therefore the apis look nearly the same. You can check BEP-20 Tokens just like ERC-20 on etherscan...

just my 2 cents

LefterisJP:
> You can check BEP-20 Tokens just like ERC-20 on etherscan.

Yes but you need to know the token addresses to query them. We don't use etherscan's api to query ethereum token addresses in rotki. That would not work as there is over 10k tokens and the etherscan api accepts 1 token per request.

I guess would need a curated list of BSC token addresses etc, and probably figure out a nice way to do batch queries to not get rate limited by that api.

Or find a better api.

But as with any chain a proper solution would be to be able to connect to your own BSC node if you got one. But ofc step by step. This issue should just be for the first iteration.

I will repeat my call to action from above. We need help here. If you want this done soon consider opening a PR or finding someone who can do that is indeed the fastest way to see this in rotki.

If not we still want to see it in the app ofc. But we are a team of 2-3 people with very limited resources and a huge issue backlog. So no guarantees as to when this will happen.

a7e8:
A good first step would be to just support BNB on the Binance Chain (not Smart Chain) using the explorer or by querying a node like the `bnbcli` does.

MarcBerneman:
> Hey guys. Thanks for +1 on this issue. We don't really know when developer resources can be dedicated to do this issue right.
> 
> Rotki is opensource and PRs are greatly appreciated.
> 
> The only open API I could find is etherscan's bscscan: https://bscscan.com/apis#accounts
> But that's only for BNB balances.

[...]

PR #9237, opened 2025-01-10
Add openocean firebird and paraswap decoders on BSC

nicholasyoder:
Related: #1573 

Since openocean is used on all evm chains and doesn't need any chain specific stuff, I've added it as a built-in decoder, like is done for Safemultisig, Oneinch, etc.

Also removed the unused callable import as requested in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1327338317687033997)

### DIFF
--- a/rotkehlchen/tests/unit/decoders/test_openocean.py
+++ b/rotkehlchen/tests/unit/decoders/test_openocean.py
@@ -11,7 +11,7 @@ from rotkehlchen.chain.evm.decoding.open_ocean.constants import (
     OPENOCEAN_LABEL,
 )
 from rotkehlchen.chain.evm.types import string_to_evm_address
-from rotkehlchen.constants.assets import A_ARB, A_ETH, A_POLYGON_POS_MATIC, A_WETH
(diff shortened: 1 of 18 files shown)


yabirgb (approved):
lgtm

rotkehlchen/tests/unit/decoders/test_openocean.py
```diff
@@ -375,3 +376,66 @@ def test_openocean_swap_uniswapv3(
...
+        notes=f'Burn {gas_amount} BNB for gas',
+        counterparty=CPT_GAS,
+    ), EvmEvent(
+        tx_hash=tx_hash,
+        sequence_index=338,
+        timestamp=timestamp,
+        location=Location.BINANCE_SC,
+        event_type=HistoryEventType.INFORMATIONAL,
+        event_subtype=HistoryEventSubType.APPROVE,
+        asset=a_bsc_eth,
```
LefterisJP, line 410: both here and in the previous test you could use the `:=` operator to initialize `a_bsc_eth` here and then use it below.

LefterisJP (approved):
Small nitpick but will merge

PR #9238, opened 2025-01-10
Fix curve on BSC

nicholasyoder:
Add a couple things that were missed in #9235

Should fix KeyError from `CURVE_CHAIN_ID`.


LefterisJP (approved):
LGTM

PR #9241, opened 2025-01-12
Hardening of the aave-v3 return event decoding

gianluca-pub-dev:
Fixed an issue in Aave V3 decoder where it could erroneously decode non-aave return wrapped events as aave ones (eg. weth unwrapping in the same transaction)

### DIFF
--- a/rotkehlchen/tests/unit/decoders/test_aave_v3.py
+++ b/rotkehlchen/tests/unit/decoders/test_aave_v3.py
@@ -16,6 +16,7 @@ from rotkehlchen.constants.assets import (
     A_USDT,
     A_WBTC,
     A_WETH,
+    A_WETH_ARB,
     A_XDAI,
 )
 from rotkehlchen.fval import FVal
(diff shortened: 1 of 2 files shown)


LefterisJP (commented):
Hey @gianluca-pub-dev looks like a correct change. Thanks!

But can you also add a test that would fail without the change? Just try to decode such a transaction that you mentioned with weth.

gianluca-pub-dev:
> Hey @gianluca-pub-dev looks like a correct change. Thanks!
> 
> But can you also add a test that would fail without the change? Just try to decode such a transaction that you mentioned with weth.

no problem, I detected this with my transactions and verified the fix, but I would prefer to not add them for privacy reasons. I will look if I can find some similar transaction on blockchain.

LefterisJP:
> > Hey @gianluca-pub-dev looks like a correct change. Thanks!
> > But can you also add a test that would fail without the change? Just try to decode such a transaction that you mentioned with weth.
> 
> no problem, I detected this with my transactions and verified the fix, but I would prefer to not add them for privacy reasons. I will look if I can find some similar transaction on blockchain.

yeah we always use some random transaction that's similar to the one we see it when writing a test. Both when we see it ourselves or when a user shows us a transaction.

gianluca-pub-dev:
Backen tests (others) failed on something that doesn't sound related to this PR

rotkehlchen/tests/unit/decoders/test_aave_v3.py
```diff
@@ -1453,3 +1454,223 @@ def test_aave_v3_interest_on_transfer(ethereum_inquirer, ethereum_accounts) -> N
...
+            tx_hash=tx_hash,
+            sequence_index=2,
+            timestamp=timestamp,
+            location=Location.ARBITRUM_ONE,
+            event_type=HistoryEventType.RECEIVE,
+            event_subtype=HistoryEventSubType.NONE,
+            asset=A_ETH,
+            balance=Balance(amount=FVal(eth_withdraw_amount) + FVal(eth_interest_amount)),
+            location_label=user_safe_proxy,
+            # New bug, this event should detected as WETH unwrapping and not as a receive
```
gianluca-pub-dev, line 1508: Hello @LefterisJP ,

I added a test case as you suggested, picking a transaction run on Arbitrum One containing both aave-v3 operations and weth unwrapping. Problem is that I found a new bug it seems, because the WETH unwrapping is not detected in this transaction.

I added the test case in any case, ready to be updated when the bug would be fixed. Is this a known bug? I don't have the time now for investigating the root cause of this issue, but if you know where is the issue, I could also fix it in the same PR if needed.

gianluca-pub-dev: Some month ago I detected this similar WETH issue https://github.com/gianluca-pub-dev/rotki/pull/1/commits/eada01325fc1bd67603271ec8891fb062dc9b417 but I don't remember if I discussed it with you: https://github.com/gianluca-pub-dev/rotki/pull/1/commits/eada01325fc1bd67603271ec8891fb062dc9b417 . Maybe it is related to this.

rotkehlchen/tests/unit/decoders/test_aave_v3.py
```diff
@@ -1453,3 +1454,223 @@ def test_aave_v3_interest_on_transfer(ethereum_inquirer, ethereum_accounts) -> N
...
+@pytest.mark.parametrize('arbitrum_one_accounts', [['0x57D4bEb0fD4438b8910fEf03a961131742D845E2', '0x1cdC859a9685103A0791075B6c365e2D583BC236']])   # noqa: E501
+def test_aave_v3_close_position_with_arbitrum_one(arbitrum_one_inquirer, arbitrum_one_accounts) -> None:    # noqa: E501
+    tx_hash = deserialize_evm_tx_hash('0x56c57b83508f46d45af6b3230ee882f99df315c089af3250fbe5494ea59e5624')  # noqa: E501
+    events, _ = get_decoded_events_of_transaction(
+        evm_inquirer=arbitrum_one_inquirer,
+        tx_hash=tx_hash,
+    )
+    timestamp = TimestampMS(1736028280000)
+    # Note for the reviewer: I would prefer splitting this into multiple lines
+    # but I was told in another PR to keep it in one line. Please let me know
```
LefterisJP, line 1469: Keep them in one line for those tests
```suggestion
```

rotkehlchen/tests/unit/decoders/test_aave_v3.py
LefterisJP: I think this is because it involves safe and aave. These are getting quite complicated to detect properly especially when they involve multiple transfers like this.

What exactly is this transaction supposed to be doing? That's the first thing that should be explained in order to determine next steps.

If we know what it's supposed to do we can try to see how to adjust.

- User via the safe pays back `2,353.18136` USDT to Aave
- User via the safe withdraws 3.038166066433018795 WETH from Aave and unwraps to ETH and withdraws directly to the user controller address

Is that right?

gianluca-pub-dev: yes @LefterisJP right, that is the flow, based on my analysis because I am not the author of the transaction.

I was looking now at the WETH contract on arbitrum one, and it seems to me that the contract is not emitting event logs, and this could be the reason why rotki is not detecting, because looking at the code is looking at the withdraw topic for decoding the unwrapping. I think that current test cases for the weth unwrapping are working because etherscan is providing the event log analyzing the input_data when the transaction is simple, only direct call to WETH arb contract: https://arbiscan.io/tx/0x8289389dd5493d56db582a68d5d14b82e38369a7d7b5eede630a764240e50941 . Is my guess reasonable?

If this is the case, I think weth decoder should be updated for checking the transfers from/to WETH contract address, so wrapping and unwrapping would be decoded even if the event log is missing.

LefterisJP: no no WETH in general does not emit event logs. it's a common issue with weth. Etherscan does not inject its own logs in the transaction results. That would have been a security issue.

This has something to do with the safe + aave combination. Someone from the team would have to take a look and I suppose in this PR since you have a test case ready even.

gianluca-pub-dev: well, in ethereum L1 it seems it does: https://etherscan.io/tx/0xdff9b85093c6af369d197815c42930e23b0dfdcd37f8c3e614ca3483cf120af9#eventlog#20

LefterisJP: Not all chains are the same. Don't ask me why some chains used different contract implementations of the WETH9 interface. No idea. We have it noted:
https://github.com/rotki/rotki/blob/f114a82d25c2d4e67606c1e8901532b972292f21/rotkehlchen/chain/evm/decoding/weth/decoder.py#L158

- In most chains (including mainnet) there is only deposit/withdraw events but no transfer events emitted.
- In arbitrum it's only transfer events.
- In scroll it's both transfer and deposit/withdrawal.

And I actually think that's where the problem lies. The way the code seems to currently solve this in arbitrum is by adding extra logic:
https://github.com/rotki/rotki/blob/f114a82d25c2d4e67606c1e8901532b972292f21/rotkehlchen/chain/arbitrum_one/modules/weth/decoder.py#L18

But that extra logic seem to check the transaction input data. So this would only work for an EoA and not for safes. Which is always a wrong approach to take, relying on input data. As that excludes safes and other smart contracts.

In scroll it seems we also do something different but only since we have all events properly: https://github.com/rotki/rotki/blob/develop/rotkehlchen/chain/scroll/modules/weth/decoder.py 
And that approach should work fine for safes.

So I think the right way forward here is to fix arbitrum WETH where I linked it above. And your suggestion to decode it when there is transfer from/to WETH contract if you do only for arbitrum makes sense. But again here I woudl go with address rule and not post-decoding rule as it's more performant.

Can you do that? We can also do the extra fix on top of your PR.

gianluca-pub-dev: oh, thanks for the analysis and links to the code @LefterisJP . So I was right that it was relying on input_data on arbitrum, but I was wrong on the fact that is was coming from the explorer. 

I will check if I will have some time in the next days for working on the agreed solution on arbitrum only: address rule for the WETH_ARB address, checking for the transfers of ETH/WETH to/from it. It seems pretty straighforward, so I am confident I will find the time.

LefterisJP (approved):
LGTM. I rebased and adjusted the PR.

There is still a TODO here which is taken by one of our devs: https://github.com/orgs/rotki/projects/11?pane=issue&itemId=95697078

PR #9242, opened 2025-01-12
Enforce git locale to english in conftests.py

gianluca-pub-dev:
Enforce git locale to english in conftests, since python script is looking for git output string, expecting them in english


LefterisJP (commented):
There is a linting error here. Also question, what exactly do you achieve by this? What is this fixing?

gianluca-pub-dev:
> There is a linting error here. Also question, what exactly do you achieve by this? What is this fixing?

linting error now fixed. The fix is for problem I encountered where  RECORD_CASSETTES=true python pytestgeventwrapper.py fails when the git locale on development system is not not en_us, this because conftests.py code is checking for english strings at line: [313](https://github.com/rotki/rotki/pull/9242/files#diff-730e9c839e2cd24c3c868effb76a60b2e044b48c16fb32889944e4f2bb6d9bd4R320)

LefterisJP (approved):
LGTM

PR #9248, opened 2025-01-14
Use the correct gnosis pay API url

LefterisJP:
The proper error codes will be only returned by api.gnosispay. The fact app.gnosispay worked for so long is just luck.


bobanm (approved):

yabirgb (approved):
lgtm

PR #9251, opened 2025-01-14
feat: integrate add to evm checkbox to chain select

lukicenturi:
Closes #(issue_number)

## Checklist


kelsos (approved):
lgtm @lukicenturi thank you

We are missing a changelog entry for this.
Also I don't remember if we have any for splitting the accounts and balances can you make sure we have one for it?

Also a small thing I noticed is that for the evm chain addition if you try to close the dialog when it's loading you get the confirmation. 

Do you think we need that after you clicked save?

lukicenturi:
> lgtm @lukicenturi thank you
> 
> We are missing a changelog entry for this. Also I don't remember if we have any for splitting the accounts and balances can you make sure we have one for it?
> 
> Also a small thing I noticed is that for the evm chain addition if you try to close the dialog when it's loading you get the confirmation.
> 
> Do you think we need that after you clicked save?

Yeah I think we need to remove that confirmation after save

PR #9252, opened 2025-01-14
handle cases of negative amounts in historical balances processing

prettyirrelevant:


rotkehlchen/balances/historical.py
```diff
@@ -111,6 +111,8 @@ def get_asset_amounts(
                 total_amount -= event.balance.amount
 
             amounts[ts_ms_to_sec(event.timestamp)] = total_amount
+            if total_amount < ZERO:
```
LefterisJP, line 114: i don't think this is enough.

The frontend also needs information about the event that caused this.

rotkehlchen/api/rest.py
```diff
@@ -5404,4 +5404,5 @@ def get_historical_asset_amounts(
         return api_response(_wrap_in_ok_result(result={
             'times': list(balances),
             'values': [str(x) for x in balances.values()],
+            'last_event_identifier': last_event_identifier,
```
LefterisJP, line 5407: don't return it if it's None

LefterisJP (approved):
LGTM

Issue #9203, opened 2025-01-05
Support uniswap v3 balance detection in other evm chains except mainnet.


LefterisJP:
## Abstract

It seems we do not detect uniswap v3 balances in other evm chain except mainnet.

## Specification

That's bad. We need to detect them.

PR #9254, opened 2025-01-14
Extend uniswap oracle to support evm chains

nicholasyoder:
Related #9203

### DIFF
--- a/rotkehlchen/chain/ethereum/oracles/constants.py
+++ b/rotkehlchen/chain/ethereum/oracles/constants.py
@@ -1,0 +1,87 @@
+from typing import Final
+
+from rotkehlchen.assets.asset import Asset
+from rotkehlchen.chain.evm.types import string_to_evm_address
+from rotkehlchen.constants.assets import (
+    A_BSC_BNB,
+    A_DAI,
+    A_ETH,
+    A_ETH_EURE,
+    A_EUR,
+    A_OP,
+    A_POLYGON_POS_MATIC,
+    A_USD,
+    A_USDC,
+    A_USDT,
+    A_WBNB,
+    A_WETH,
+    A_WETH_ARB,
+    A_WETH_BASE,
+    A_WETH_OPT,
+    A_WETH_POLYGON,
+    A_WMATIC,
+)
+from rotkehlchen.types import ChainID
+
+SINGLE_SIDE_USD_POOL_LIMIT: Final = 5000
+UNISWAP_ASSET_TO_EVM_ASSET: Final = {
+    A_USD: A_USDC,
+    A_EUR: A_ETH_EURE,
+    A_ETH: A_WETH,
+    A_BSC_BNB: A_WBNB,
+    A_POLYGON_POS_MATIC: A_WMATIC,
+}
+
+A_POLYGON_USDC: Final = Asset('eip155:137/erc20:0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174')
+A_POLYGON_POS_USDT: Final = Asset('eip155:137/erc20:0xc2132D05D31c914a87C6611C10748AEb04B58e8F')
+A_ARBITRUM_USDC: Final = Asset('eip155:42161/erc20:0xaf88d065e77c8cC2239327C5EDb3A432268e5831')
+A_ARBITRUM_USDT: Final = Asset('eip155:42161/erc20:0xFd086bC7CD5C481DCC9C85ebE478A1C0b69FCbb9')
+A_OPTIMISM_USDC: Final = Asset('eip155:10/erc20:0x0b2C639c533813f4Aa9D7837CAf62653d097Ff85')
+A_OPTIMISM_USDT: Final = Asset('eip155:10/erc20:0x94b008aA00579c1307B0EF2c499aD98a8ce58e58')
+A_BASE_USDC: Final = Asset('eip155:8453/erc20:0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913')
+A_BSC_USDC: Final = Asset('eip155:56/erc20:0x8AC76a51cc950d9822D68b83fE1Ad97B32Cd580d')
+A_BSC_USDT: Final = Asset('eip155:56/erc20:0x55d398326f99059fF775485246999027B3197955')
+A_BSC_ETH: Final = Asset('eip155:56/erc20:0x2170Ed0880ac9A755fd29B2688956BD959F933F8')
+
+UNISWAP_ETH_ASSETS: Final = {
+    ChainID.ETHEREUM: A_WETH,
+    ChainID.POLYGON_POS: A_WETH_POLYGON,
+    ChainID.ARBITRUM_ONE: A_WETH_ARB,
+    ChainID.OPTIMISM: A_WETH_OPT,
+    ChainID.BASE: A_WETH_BASE,
+    ChainID.BINANCE_SC: A_BSC_ETH,
+}
+UNISWAP_ROUTING_ASSETS: Final = {
+    ChainID.ETHEREUM: [A_WETH, A_DAI, A_USDT],
+    ChainID.POLYGON_POS: [A_POLYGON_POS_MATIC, A_WETH_POLYGON, A_POLYGON_USDC, A_POLYGON_POS_USDT],
+    ChainID.ARBITRUM_ONE: [A_WETH_ARB, A_ARBITRUM_USDC, A_ARBITRUM_USDT],
+    ChainID.OPTIMISM: [A_OP, A_WETH_OPT, A_OPTIMISM_USDC, A_OPTIMISM_USDT],
+    ChainID.BASE: [A_WETH_BASE, A_BASE_USDC],
+    ChainID.BINANCE_SC: [A_WBNB, A_BSC_USDC, A_BSC_USDT],
+}
+UNISWAP_FACTORY_ADDRESSES: Final = {
+    2: {  # V2 factories
+        ChainID.ETHEREUM: string_to_evm_address('0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f'),
+        ChainID.POLYGON_POS: string_to_evm_address('0x9e5A52f57b3038F1B8EeE45F28b3C1967e22799C'),
+        ChainID.ARBITRUM_ONE: string_to_evm_address('0xf1D7CC64Fb4452F05c498126312eBE29f30Fbcf9'),
+        ChainID.OPTIMISM: string_to_evm_address('0x0c3c1c532F1e39EdF36BE9Fe0bE1410313E074Bf'),
+        ChainID.BASE: string_to_evm_address('0x8909Dc15e40173Ff4699343b6eB8132c65e18eC6'),
+        ChainID.BINANCE_SC: string_to_evm_address('0x8909Dc15e40173Ff4699343b6eB8132c65e18eC6'),
+    },
+    3: {  # V3 factories
+        ChainID.ETHEREUM: string_to_evm_address('0x1F98431c8aD98523631AE4a59f267346ea31F984'),
+        ChainID.POLYGON_POS: string_to_evm_address('0x1F98431c8aD98523631AE4a59f267346ea31F984'),
+        ChainID.ARBITRUM_ONE: string_to_evm_address('0x1F98431c8aD98523631AE4a59f267346ea31F984'),
+        ChainID.OPTIMISM: string_to_evm_address('0x1F98431c8aD98523631AE4a59f267346ea31F984'),
+        ChainID.BASE: string_to_evm_address('0x33128a8fC17869897dcE68Ed026d694621f6FDfD'),
+        ChainID.BINANCE_SC: string_to_evm_address('0xdB1d10011AD0Ff90774D0C6Bb92e5C5c8b4461F7'),
+    },
+}
+UNISWAP_SUPPORTED_CHAINS: Final = {
+    ChainID.ETHEREUM,
+    ChainID.POLYGON_POS,
+    ChainID.ARBITRUM_ONE,
+    ChainID.OPTIMISM,
+    ChainID.BASE,
+    ChainID.BINANCE_SC,
+}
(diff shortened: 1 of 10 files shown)


rotkehlchen/chain/ethereum/oracles/constants.py
```diff
@@ -0,0 +1,68 @@
...
+A_ARBITRUM_USDC = Asset('eip155:42161/erc20:0xaf88d065e77c8cC2239327C5EDb3A432268e5831')
+A_ARBITRUM_USDT = Asset('eip155:42161/erc20:0xFd086bC7CD5C481DCC9C85ebE478A1C0b69FCbb9')
+A_OPTIMISM_USDC = Asset('eip155:10/erc20:0x0b2C639c533813f4Aa9D7837CAf62653d097Ff85')
+A_OPTIMISM_USDT = Asset('eip155:10/erc20:0x94b008aA00579c1307B0EF2c499aD98a8ce58e58')
+A_BASE_USDC: Final = Asset('eip155:8453/erc20:0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913')
+A_BSC_USDC: Final = Asset('eip155:56/erc20:0x8AC76a51cc950d9822D68b83fE1Ad97B32Cd580d')
+A_BSC_USDT: Final = Asset('eip155:56/erc20:0x55d398326f99059fF775485246999027B3197955')
+
+UNISWAP_ROUTING_ASSETS: Final = {
+    ChainID.ETHEREUM: [A_WETH, A_DAI, A_USDT],
```
nicholasyoder, line 46: @yabirgb, You mentioned in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1328744693084721226) to replace DAI, but when I changed it to USDC it made the `test_uniswap_oracles_asset_to_asset` test fail due to it trying to use a pool with low liquidity instead of whatever route it was using before. I think the `find_route` logic does need some rethinking, since currently if it finds a tiny pool with one of the routing assets it will try to use it and then fail due to low reserves, while it could have found a higher liquidity route by going through a couple other pools. I'll see about working on this logic in another PR.

rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -357,40 +367,47 @@ def can_query_history(
...
-        super().__init__(ethereum_inquirer=ethereum_inquirer, version=3)
-        self.uniswap_v3_pool_abi = self.ethereum.contracts.abi('UNISWAP_V3_POOL')
-        self.uniswap_v3_factory = self.ethereum.contracts.contract(string_to_evm_address('0x1F98431c8aD98523631AE4a59f267346ea31F984'))  # noqa: E501
+    def __init__(self, evm_inquirer: 'EvmNodeInquirer'):
+        super().__init__(evm_inquirer=evm_inquirer, version=3)
+        self.uniswap_v3_pool_abi = self.evm_inquirer.contracts.abi('UNISWAP_V3_POOL')
+        self.update_factory_contract()
+
+    def update_factory_contract(self) -> None:
+        self.uniswap_v3_factory = self.evm_inquirer.contracts.contract(UNISWAPV3_FACTORY_ADDRESSES[self.evm_inquirer.chain_id])  # noqa: E501
```
yabirgb, line 402: no need to load it from the database each time. make it a dict with key the chain and it should work

rotkehlchen/chain/ethereum/oracles/constants.py
yabirgb: yes. It is a known issue that we need to query the liquidity. Not sure how to properly do it because it at least is 1 extra blockchain check and I'm not sure if we also need to know the price of the asset. But let's see if we can improve it in another PR

rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -64,23 +60,63 @@ class UniswapOracle(HistoricalPriceOracleInterface, CacheableMixIn):
...
+                token.token_kind != EvmTokenKind.ERC20 or
+                token.chain_id not in {ChainID.ETHEREUM, ChainID.POLYGON_POS, ChainID.ARBITRUM_ONE, ChainID.OPTIMISM, ChainID.BASE, ChainID.BINANCE_SC}  # noqa: E501
+            ):
+                raise PriceQueryUnsupportedAsset(f'{self.name} oracle: {token} is not an ERC20 token in an EVM chain supported by Uniswap')  # noqa: E501
+
+            tokens.append(token)
+
+        # Convert A_WETH to the correct weth token from the chain of the other token
+        from_token, to_token = tokens
+        if from_token.chain_id != to_token.chain_id and A_WETH.resolve_to_evm_token() in tokens:
```
yabirgb, line 97: here you are checking for `A_WETH`. Shouldn't you also take care of weth/wrapped on other chains

rotkehlchen/chain/ethereum/oracles/constants.py
```diff
@@ -0,0 +1,68 @@
...
+
+A_POLYGON_USDC = Asset('eip155:137/erc20:0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174')
+A_POLYGON_POS_USDT: Final = Asset('eip155:137/erc20:0xc2132D05D31c914a87C6611C10748AEb04B58e8F')
+A_ARBITRUM_USDC = Asset('eip155:42161/erc20:0xaf88d065e77c8cC2239327C5EDb3A432268e5831')
+A_ARBITRUM_USDT = Asset('eip155:42161/erc20:0xFd086bC7CD5C481DCC9C85ebE478A1C0b69FCbb9')
+A_OPTIMISM_USDC = Asset('eip155:10/erc20:0x0b2C639c533813f4Aa9D7837CAf62653d097Ff85')
+A_OPTIMISM_USDT = Asset('eip155:10/erc20:0x94b008aA00579c1307B0EF2c499aD98a8ce58e58')
+A_BASE_USDC: Final = Asset('eip155:8453/erc20:0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913')
+A_BSC_USDC: Final = Asset('eip155:56/erc20:0x8AC76a51cc950d9822D68b83fE1Ad97B32Cd580d')
+A_BSC_USDT: Final = Asset('eip155:56/erc20:0x55d398326f99059fF775485246999027B3197955')
```
yabirgb, line 43: Whjy make some final and other not?

yabirgb:
Also this is to have prices with uniswap v3 but is this needed to have the Uniswap V3 balances on other chains detected?

rotkehlchen/chain/ethereum/oracles/uniswap.py
nicholasyoder: This is handling when plain `A_ETH` is used in chains other than mainnet. In the `UNISWAP_ASSET_TO_EVM_ASSET` dict it maps `A_ETH` to `A_WETH` since at that point it doesn't know the chain of the other token, so here it will be `A_WETH` no matter the chain of the other token. Then the correct weth for the other token's chain is retrieved using `CHAIN_TO_WRAPPED_TOKEN` in this `if` block.

nicholasyoder:
> Also this is to have prices with uniswap v3 but is this needed to have the Uniswap V3 balances on other chains detected?

So to simply detect that the user has a uniswap balance, no this isn't needed, but the Uniswap oracle is used to retrieve the price/value of that balance. See here: https://github.com/rotki/rotki/blob/develop/rotkehlchen/chain/ethereum/modules/uniswap/v3/utils.py#L534

rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -64,16 +60,55 @@ class UniswapOracle(HistoricalPriceOracleInterface, CacheableMixIn):
...
+            self,
+            from_asset: Asset,
+            to_asset: Asset,
+    ) -> Sequence[EvmToken]:
+        """Resolve assets to EvmTokens. Returns a tuple containing the from token and to token.
+        May raise:
+            - PriceQueryUnsupportedAsset
+        """
+        tokens: list[EvmToken] = []
+        for asset in [from_asset, to_asset]:
```
LefterisJP, line 89:
```suggestion
        for asset in (from_asset, to_asset):
```

rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -64,16 +60,55 @@ class UniswapOracle(HistoricalPriceOracleInterface, CacheableMixIn):
...
+            ):
+                raise PriceQueryUnsupportedAsset(f'{self.name} oracle: {token} is not an ERC20 token in an EVM chain supported by Uniswap')  # noqa: E501
+
+            tokens.append(token)
+
+        # Convert A_WETH to the correct weth token from the chain of the other token
+        from_token, to_token = tokens
+        if from_token.chain_id != to_token.chain_id and A_WETH.resolve_to_evm_token() in tokens:
+            if from_asset == A_ETH:
+                from_token = CHAIN_TO_WRAPPED_TOKEN[to_token.chain_id.to_blockchain()].resolve_to_evm_token()  # noqa: E501
```
LefterisJP, line 107: Hmm this seems wrong. `CHAIN_TO_WRAPPED_TOKEN` returns the native wrapped token of the chain. Not the wraped eth of the chain. So for gnosis chain this would be WXDAI. Just look at the definition of the mapping you are using

```
CHAIN_TO_WRAPPED_TOKEN: Final = {
    SupportedBlockchain.ETHEREUM: A_WETH,
    SupportedBlockchain.ARBITRUM_ONE: A_WETH_ARB,
    SupportedBlockchain.OPTIMISM: A_WETH_OPT,
    SupportedBlockchain.BASE: A_WETH_BASE,
    SupportedBlockchain.GNOSIS: A_WXDAI,
    SupportedBlockchain.POLYGON_POS: A_WETH_POLYGON,
    SupportedBlockchain.BINANCE_SC: A_WBNB,
}
```

I would say we need a different definition for chain to WRAPPED_ETH. But also similarly logic for the native token. Say if it's gnosis and either from/to is xDAI it shoudl use wXDAI. Or polygon and POL/MATIC it should use wMATIC or wPOL whatever it's called now.

rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -357,41 +373,41 @@ def can_query_history(
             seconds: int | None = None,
     ) -> bool:
         """Here to comply with price historian interface."""
-        return self.is_before_contract_creation(self.ethereum.get_blocknumber_by_time(timestamp))
+        token, _ = self.resolve_assets(from_asset=from_asset, to_asset=to_asset)
```
LefterisJP, line 376: What exactly is this? What contract are you checking? The docstring seem the one you would use for a no-op function. Populate the docstring with a proper explanation

rotkehlchen/tests/unit/test_defi_oracles.py
```diff
@@ -94,6 +111,49 @@ def test_uniswap_oracles_historic_price(inquirer_defi, socket_enabled):  # pylin
...
+    assert inquirer_defi._uniswapv3.query_current_price(
+        from_asset=A_ETH,
+        to_asset=A_OPTIMISM_USDT,
+    ) == Price(FVal('3214.5282531850976'))
+
+    # v3 current on bsc
+    assert inquirer_defi._uniswapv3.query_current_price(
+        from_asset=A_BSC_USDT,
+        to_asset=A_BSC_BNB,
+    ) == Price(FVal('0.001433422881060733'))
```
LefterisJP, line 154: From the comment in the implementation I think we should shoudl see polygon pos matic and gnosis xDAI and BSC BSC token. To make sure native tokens are handled properly.

rotkehlchen/chain/ethereum/oracles/uniswap.py
```diff
@@ -64,16 +60,55 @@ class UniswapOracle(HistoricalPriceOracleInterface, CacheableMixIn):
...
-        self.routing_assets = [
-            self.weth,
-            A_DAI.resolve_to_evm_token(),
-            A_USDT.resolve_to_evm_token(),
-        ]
+        self.routing_assets = {
+            chain_id: [asset.resolve_to_evm_token() for asset in assets]
+            for chain_id, assets in UNISWAP_ROUTING_ASSETS.items()
+        }
+        self.evm_inquirers = {
```
LefterisJP, line 70: do you need this mapping? Can't you call this on demand? Cause you are recreating this mapping for the entire lifetime of the app, while the data is already in the Inquirer()

rotkehlchen/chain/ethereum/oracles/uniswap.py
nicholasyoder: When I used `CHAIN_TO_WRAPPED_TOKEN` I was assuming that the only chains where it would encounter plain `A_ETH` and trigger this logic are those chains where ETH is the native token, which would make using that mapping viable, since those chains are all mapped to WETH there. But I suppose that is an unsafe assumption.

I've changed it to use a new mapping: `UNISWAP_ETH_ASSETS`. Note that for BSC it uses `A_BSC_ETH` which is not actually WETH - its something called [Binance-Peg Ethereum token](https://bscscan.com/token/0x2170ed0880ac9a755fd29b2688956bd959f933f8), but seems to generally be what is used for ETH on BSC.

In regard to other native tokens, that is being handled by `UNISWAP_ASSET_TO_EVM_ASSET`, where it maps native BNB to WBNB and POL/MATIC to WMATIC. Note that Gnosis is not supported by Uniswap, so it doesn't include xDAI.

rotkehlchen/chain/ethereum/oracles/uniswap.py
nicholasyoder: updated the docstring, renamed the `is_before_contract_creation` function to be a bit more descriptive, and also added a try except for `PriceQueryUnsupportedAsset` since it should return false if that happens as well.

rotkehlchen/chain/ethereum/oracles/uniswap.py
nicholasyoder: hmm... no I guess we don't need this mapping. Removed it.

LefterisJP (approved):
LGTM

PR #9263, opened 2025-01-16
allow querying of historical balance amounts for assets in a collection

prettyirrelevant:
image
Improvement as requested by @lukicenturi

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -5388,16 +5388,29 @@ class RestAPI:
 
     def get_historical_asset_amounts(
             self,
-            asset: Asset,
+            asset: Asset | None,
+            collection_id: int | None,
             from_timestamp: Timestamp,
             to_timestamp: Timestamp,
     ) -> Response:
+        assets: tuple[Asset, ...]
         try:
-            balances, last_event_identifier = HistoricalBalancesManager(self.rotkehlchen.data.db).get_asset_amounts(  # noqa: E501
-                asset=asset,
+            if asset is not None:
+                assets = (asset,)
+            else:  # collection_id is present due to validation.
+                with GlobalDBHandler().conn.read_ctx() as cursor:
+                    cursor.execute(
+                        'SELECT asset FROM multiasset_mappings WHERE collection_id=?',
+                        (collection_id,),
+                    )
+                    assets = tuple(Asset(row[0]) for row in cursor)
+
+            balances, last_event_identifier = HistoricalBalancesManager(self.rotkehlchen.data.db).get_assets_amounts(  # noqa: E501
+                assets=assets,
                 from_ts=from_timestamp,
                 to_ts=to_timestamp,
             )
+
         except NotFoundError as e:
             return api_response(wrap_in_fail_result(str(e), status_code=HTTPStatus.NOT_FOUND))
 
--- a/rotkehlchen/balances/historical.py
+++ b/rotkehlchen/balances/historical.py
@@ -78,13 +78,13 @@ class HistoricalBalancesManager:
 
         return {'amount': amounts[asset], 'price': price}
 
-    def get_asset_amounts(
+    def get_assets_amounts(
             self,
-            asset: Asset,
+            assets: tuple[Asset, ...],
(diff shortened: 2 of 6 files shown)


lukicenturi (approved):
Tested, works fine to me

rotkehlchen/api/rest.py
```diff
@@ -5388,16 +5388,24 @@ def get_historical_asset_balance(self, asset: Asset, timestamp: Timestamp) -> di
...
-                to_ts=to_timestamp,
-            )
+            if asset is not None:
+                balances, last_event_identifier = HistoricalBalancesManager(self.rotkehlchen.data.db).get_asset_amounts(  # noqa: E501
+                    asset=asset,
+                    from_ts=from_timestamp,
+                    to_ts=to_timestamp,
+                )
+            else:
+                balances, last_event_identifier = HistoricalBalancesManager(self.rotkehlchen.data.db).get_collection_amounts(  # noqa: E501
```
yabirgb, line 5404: can you do `elif collection_id is not None` and have an else that is a bad input? It shouldn't happen since you validate it in the schema but just to be protected

rotkehlchen/balances/historical.py
```diff
@@ -78,6 +79,33 @@ def get_asset_balance(self, asset: Asset, timestamp: Timestamp) -> HistoricalBal
...
+        with GlobalDBHandler().conn.read_ctx() as cursor:
+            cursor.execute(
+                'SELECT asset FROM multiasset_mappings WHERE collection_id=?',
+                (collection_id,),
+            )
+            assets = tuple(Asset(row[0]) for row in cursor)
+
+        events, _ = self._get_events_and_currency(from_ts=from_ts, to_ts=to_ts, assets=assets)
+        if len(events) == 0:
+            raise NotFoundError(f'No historical data found for collection {collection_id} between {from_ts=} and {to_ts=}.')  # noqa: E501
```
yabirgb, line 105: looks like this part can also be moved to the common logic? The only thing is that for the string you will have some check of `if len(assets) > 1`. I would change `get_asset_amounts` to `get_assets_amounts` and when you call it in rest.py just to `assets=(asset,)`. 

Then `get_collection_amounts` is just a wrapper that gets the assets by collection id

rotkehlchen/tests/api/test_historical_balances.py
```diff
@@ -228,6 +229,66 @@ def test_get_historical_asset_amounts_over_time_with_negative_amount(
...
+    db = rotkehlchen_api_server.rest_api.rotkehlchen.data.db
+    events = [
+        HistoryEvent(
+            event_identifier='btc_spend_2',
+            sequence_index=3,
+            timestamp=ts_sec_to_ms(three_days_after_start := Timestamp(START_TS + DAY_IN_SECONDS * 3)),  # noqa: E501
+            event_type=HistoryEventType.SPEND,
+            event_subtype=HistoryEventSubType.NONE,
+            location=Location.BLOCKCHAIN,
+            asset=Asset('eip155:1/erc20:0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599'),
```
LefterisJP, line 246: Can you write name of asset in a coment?

rotkehlchen/tests/api/test_historical_balances.py
```diff
@@ -228,6 +229,66 @@ def test_get_historical_asset_amounts_over_time_with_negative_amount(
...
+            notes='BTC first overspend attempt',
+        ),
+        HistoryEvent(
+            event_identifier='btc_spend_3',
+            sequence_index=4,
+            timestamp=ts_sec_to_ms(four_days_after_start := Timestamp(START_TS + DAY_IN_SECONDS * 4)),  # noqa: E501
+            event_type=HistoryEventType.SPEND,
+            event_subtype=HistoryEventSubType.NONE,
+            location=Location.BLOCKCHAIN,
+            asset=Asset('eip155:100/erc20:0x8e5bBbb09Ed1ebdE8674Cda39A0c169401db4252'),
```
LefterisJP, line 257: Can you write name of asset in a coment?

PR #9265, opened 2025-01-16
feat: use historical balances to show asset amount and balance over time

lukicenturi:
https://github.com/rotki/rotki-premium/pull/16

More things to do:
- [x] Throttle the data (for example if there is 1000 thousand data, I can't fetch historic prices for all of them, I should throttle them in "All view", and then I can show the detail when they are zoomed in)
- [x] Handle case if there is negative amount because of an problematic tx, and redirect to the history event.


kelsos (approved):
lgtm thanks

frontend/app/src/premium/premium-apis.ts
```diff
@@ -14,6 +14,9 @@ import { useBalancesBreakdown } from '@/composables/balances/breakdown';
 import { useAssetManagementApi } from '@/composables/api/assets/management';
 import { useStatisticsApi } from '@/composables/api/statistics/statistics-api';
 import { useAssetInfoRetrieval } from '@/composables/assets/retrieval';
+import { useHistoricCachePriceStore } from '@/store/prices/historic';
+import { usePriceApi } from '@/composables/api/balances/price';
+import type BigNumber from 'bignumber.js';
```
kelsos, line 19: this should be from `@rotki/common`

frontend/app/src/premium/register-components.ts
```diff
@@ -126,6 +127,7 @@ export function registerComponents(app: App): void {
   app.component('UniswapPoolAssetBalance', UniswapPoolAssetBalance);
   // Version 26 - 1.34
   app.component('RefreshButton', RefreshButton);
+  app.component('AssetBalanceStatisticSourceSetting', AssetBalanceStatisticSourceSetting);
```
kelsos, line 130: Do you think we could use a slot to pass this to the component instead?

kelsos (approved):
So small comments but overall looks good.

frontend/app/src/premium/register-components.ts
lukicenturi: Probably we can, but we use <PremiumStatistics /> directly, so this slot will be really specific

kelsos: Hmm true, it would work fine for the details page but it would have to be passed through in the PremiumStatistics. 

Let's keep the PR's approach then thanks.

PR #9268, opened 2025-01-16
Add Uniswap V3 balances to non-mainnet chains

nicholasyoder:
Closes: #9203


rotkehlchen/chain/evm/decoding/uniswap/v3/utils.py
```diff
@@ -0,0 +1,211 @@
...
+
+        if asset_usd_price == ZERO_PRICE:
+            try:
+                asset_usd_price = inquirer.query_uniswapv3_current_price(asset, A_USDC.resolve_to_asset_with_oracles())  # noqa: E501
+            except (PriceQueryUnsupportedAsset, RemoteError) as e:
+                log.debug(
+                    f'Failed to find price for {asset!s}/{A_USDC!s} LP using '
+                    f'Uniswap V3 oracle due to: {e!s}.',
+                )
+                continue
```
prettyirrelevant, line 207: is this really needed? doesn't the `get_uniswap_v3_position_price()` suffice?

rotkehlchen/inquirer.py
```diff
@@ -473,6 +481,22 @@ def add_defi_oracles(
...
+        Used to avoid a circular import of the Inquirer when finding Uniswap V3 LP position prices.
+        """
+        if (uniswap := Inquirer()._uniswapv3) is not None:
+            return uniswap.query_current_price(
+                from_asset=from_asset,
+                to_asset=to_asset,
+            )
+
+        return Price(ZERO)
+
```
prettyirrelevant, line 499: as a result of removing the diff attached in the image, this should go away i guess.
image

or if i'm wrong, what exactly does this method achieve?

prettyirrelevant (commented):
from what i understand, `usd_value` in the NFTs section is now obsolete and the accurate LP worth is shown in balances?

rotkehlchen/chain/evm/decoding/uniswap/v3/utils.py
nicholasyoder: So this is part of `get_uniswap_v3_position_price`'s logic. Its getting the prices of the assets within a uniswap position. If the asset has oracles it simply uses the normal `find_usd_price`:
```
        if asset.has_oracle():
            asset_usd_price = inquirer.find_usd_price(asset)
```

But when that isn't the case it needs to request the price directly from the UniswapV3Oracle. [Previously](https://github.com/rotki/rotki/blob/973b43d1633de5318b90934ca6cc975ec87c1716/rotkehlchen/chain/ethereum/modules/uniswap/v3/utils.py#L531) the UniswapV3Oracle was imported directly, but that isn't possible here due to circular import issues with the inquirer. Also the uniswap oracle is already initialized in the inquirer but is a protected member, so I added the `query_uniswapv3_current_price` function to the inquirer in order to directly query the uniswap oracle here.

nicholasyoder:
> from what i understand, `usd_value` in the NFTs section is now obsolete and the accurate LP worth is shown in balances?

yes, the NFTs section will be removed eventually as its part of the deprecated ethereum module structure. The correct position values should be shown in the asset balances.

rotkehlchen/chain/evm/tokens.py
```diff
@@ -410,6 +411,12 @@ def _get_token_exceptions(self) -> set[ChecksumEvmAddress]:
             if (evm_details := asset_id_is_evm_token(asset_id)) is not None and evm_details[0] == self.evm_inquirer.chain_id:  # noqa: E501
                 exceptions.add(evm_details[1])
 
+        # Skip Uniswap V3 position tokens. A balance corresponding to the number
+        # of positions owned by the user is found for each position token otherwise.
+        # Actual balances are handled by the UniswapV3Balances class.
+        if self.evm_inquirer.chain_id in UNISWAP_V3_NFT_MANAGER_ADDRESSES:
+            exceptions.add(UNISWAP_V3_NFT_MANAGER_ADDRESSES[self.evm_inquirer.chain_id])
```
prettyirrelevant, line 418: this comment is quite confusing especially this part
> A balance corresponding to the number of positions owned by the user is found for each position token otherwise.

rotkehlchen/chain/evm/decoding/uniswap/v3/utils.py
prettyirrelevant: oh okay. i understand now but there's no need for `asset.has_oracle()` here as it prevents defillama from being queried.

defillama does not use `coingecko` or `cryptocompare` identifiers for evm tokens price queries https://github.com/rotki/rotki/blob/125cb90e7c0305678154bf657d471987f90608c8/rotkehlchen/externalapis/defillama.py#L109

as such, you won't need to call the uniswap oracle directly as defillama should have prices for most evm tokens. the code was added when we didn't have defillama as an oracle

prettyirrelevant (commented):
two comments. 
also, tested it with two accounts and it works as expected.

rotkehlchen/chain/evm/decoding/uniswap/v3/utils.py
nicholasyoder: hmm... yes I guess you are right. I've removed this direct call to the uniswap oracle and also removed the `query_uniswapv3_current_price` function.

Also when reVCRing the tests after this change, I ran into some trouble with the multicall chunk size, so as recommended in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1330902164012404913) I've made it skip etherscan when doing multicalls in the uniswap balances logic.

rotkehlchen/inquirer.py
nicholasyoder: Removed this now.

rotkehlchen/chain/evm/tokens.py
nicholasyoder: Tried to rewrite this to be clearer.

rotkehlchen/assets/asset.py
```diff
@@ -473,6 +473,7 @@ class EvmToken(CryptoAsset):
     decimals: int | None = field(init=False)
     protocol: str | None = field(init=False)
     underlying_tokens: list[UnderlyingToken] = field(init=False)
+    collectible_id: str | None = field(init=False)
```
LefterisJP, line 476: No no no. Do not give a new field to all EvmTokens, just for uniswap's sake. That makes no sense and wastes a ton of memory as `EvmToken` is widely used across the entire codebase.

If you really need it then make a superclass just for this case and use it only in uniswap

rotkehlchen/chain/evm/decoding/uniswap/v3/balances.py
```diff
@@ -0,0 +1,164 @@
...
+        nft_token_ids: list[int] = []
+        try:
+            amount_of_positions = self.uniswap_v3_nft_manager.call(
+                node_inquirer=self.evm_inquirer,
+                method_name='balanceOf',
+                arguments=[address],
+            )
+        except RemoteError as e:
+            raise RemoteError(
+                f'Error calling nft manager contract to fetch of LP positions count for '
```
LefterisJP, line 67:
```suggestion
                f'Error calling nft manager contract to fetch LP positions count for '
```

rotkehlchen/chain/evm/decoding/uniswap/v3/balances.py
```diff
@@ -0,0 +1,164 @@
...
+                self.uniswap_v3_nft_manager.decode(
+                    result=data[1],
+                    method_name='tokenOfOwnerByIndex',
+                    arguments=[address, index],
+                )[0]
+                for index, data in enumerate(tokens_ids_multicall) if data[0] is True
+            ]
+
+            try:
+                # Get the user liquidity position using the token ID retrieved.
```
LefterisJP, line 104:
```suggestion
            try:  # Get the user liquidity position using the token ID retrieved.
```

rotkehlchen/chain/evm/decoding/uniswap/v3/utils.py
```diff
@@ -0,0 +1,199 @@
...
+        evm_inquirer: 'EvmNodeInquirer',
+        inquirer: 'Inquirer',
+        token: 'EvmToken',
+) -> Price:
+    """Get the USD price of a Uniswap V3 LP position identified by the token's collectible id."""
+    try:
+        token_id = int(token.collectible_id or '')  # also raise ValueError if id is None
+    except ValueError:
+        log.error(f'Failed to find Uniswap V3 position price for {token} due to invalid token ID.')
+        return ZERO_PRICE
```
LefterisJP, line 113: This seems wrong.If `collectible_id` is None it will raise `TypeError`.

```suggestion
    try:
        token_id = int(token.collectible_id)
    except TypeError:
        log.error(f'Failed to find Uniswap V3 position price for {token} due to invalid token ID.')
        return ZERO_PRICE
```

rotkehlchen/chain/evm/decoding/uniswap/v3/utils.py
```diff
@@ -0,0 +1,199 @@
...
+            token0_address_raw=position[2],
+            token1_address_raw=position[3],
+            fee=position[4],
+        )
+    except DeserializationError as e:
+        log.error(f'Failed to compute Uniswap V3 pool address for {token} due to {e!s}')
+        return ZERO_PRICE
+
+    # Get the liquidity pool's state i.e `slot0`
+    try:
```
LefterisJP, line 146:
```suggestion
    try:    # Get the liquidity pool's state i.e `slot0`
```

LefterisJP (commented):
I left some comments.

As for the logic itself is quite complicated and the only way to attest correctness is test it in the app and compare to other trackers that the positions are both detected and valued correctly. Did you do that?

rotkehlchen/assets/asset.py
nicholasyoder: Removed this, and added a small `identifier_to_collectible_id` function to retrieve the collectible id from the identifier when it is needed.

rotkehlchen/chain/evm/decoding/uniswap/v3/utils.py
nicholasyoder: Now using `identifier_to_collectible_id` here with an if check for `None`. The `token_id` is now being converted to `int` where it actually makes the contract call and in the unexpected event of the id containing non-integer characters I added `ValueError` to the try block there to handle it.

rotkehlchen/chain/evm/decoding/uniswap/v3/balances.py
```diff
@@ -0,0 +1,163 @@
...
+                positions_multicall = self.evm_inquirer.multicall_2(
+                    require_success=False,
+                    calls=[
+                        (
+                            self.uniswap_v3_nft_manager.address,
+                            self.uniswap_v3_nft_manager.encode('positions', [token_id]),
+                        )
+                        for token_id in tokens_ids
+                    ],
+                    call_order=self.evm_inquirer.default_call_order(skip_etherscan=True),  # too many calls per chunk can fail on etherscan  # noqa: E501
```
LefterisJP, line 113: isn;t the right solution to have different chunk limit per etherscan and per other nodes? I mean each node also may have different gas limit

nicholasyoder: Changed this to use different chunk sizes depending on which type of node will be used as discussed in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1330997197365317713). Unfortunately the chunk sizes used in `get_chunk_size_call_order` are much larger than what works here, so can't use that directly, but implemented similar logic here.

rotkehlchen/chain/evm/decoding/uniswap/v3/balances.py
```diff
@@ -0,0 +1,173 @@
...
+        except RemoteError as e:
+            raise RemoteError(
+                f'Error calling nft manager contract to fetch LP positions count for '
+                f'an address with properties: {e!s}',
+            ) from e
+
+        if amount_of_positions == 0:
+            return nft_token_ids
+
+        # Limit chunk size similarly to get_chunk_size_call_order but use smaller sizes
```
LefterisJP, line 76: This is software development. Not copy pasta.

why not just make `get_chunk_size_call_order` parametrizable with a default argument? and use it both here and in he other places it's used with its default chunk values while here set one?

nicholasyoder: yes, that is better - I should have thought of that. Changed it.

LefterisJP (approved):
LGTM

PR #9279, opened 2025-01-19
Deleting an ethereum address should now purge its withdrawal cache

LefterisJP:
Addresses a bug reported by my friend @gnidan

### DIFF
--- a/rotkehlchen/tests/db/test_db.py
+++ b/rotkehlchen/tests/db/test_db.py
@@ -330,6 +330,7 @@ def test_writing_fetching_data(data_dir, username, sql_vm_instructions_cb):
     data = DataHandler(data_dir, msg_aggregator, sql_vm_instructions_cb)
     data.unlock(username, '123', create_new=True, resume_from_backup=False)
 
+    existing_address = string_to_evm_address('0xd36029d76af6fE4A356528e4Dc66B2C18123597D')
     with data.db.user_write() as write_cursor:
         data.db.add_blockchain_accounts(
             write_cursor,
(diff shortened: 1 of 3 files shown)


rotkehlchen/tests/db/test_db.py
```diff
@@ -365,28 +366,24 @@ def test_writing_fetching_data(data_dir, username, sql_vm_instructions_cb):
...
-        data.db.set_dynamic_cache(
-            write_cursor=write_cursor,
-            name=DBCacheDynamic.EXTRA_INTERNAL_TX,
-            value=string_to_evm_address('0xd36029d76af6fE4A356528e4Dc66B2C18123597D'),
-            tx_hash=random_tx_hash_in_cache.hex(),  # pylint: disable=no-member
-            receiver=string_to_evm_address('0xd36029d76af6fE4A356528e4Dc66B2C18123597D'),
+        random_tx_hash_in_cache = make_evm_tx_hash().hex()  # pylint: disable=no-member
+
+        cache_data = (
+            (DBCacheDynamic.EXTRA_INTERNAL_TX, {'tx_hash': random_tx_hash_in_cache, 'receiver': existing_address}, existing_address),  # noqa: E501
```
LefterisJP, line 372: question to self: This `EXTRA_INTERNAL_TX` cache key:

1. It gets deleted at EVM address deletion in a very broad way for any chain's evm address. 

```python
        write_cursor.execute(
            f"DELETE FROM key_value_cache WHERE name LIKE '{EXTRAINTERNALTXPREFIX}_%' AND value = ?",  # noqa: E501
            (address,),
        )
```

That seems wrong. So if my   main address gets added to BSC and I delete the spam BSC account, my main EVM account on mainnet will get this cache purged.

2. What does it do? Where is it used? I only find it getting deleted. It most probably is no longer used? If so I will remove it in develop after I merge bugfixes to it.

PR #9282, opened 2025-01-21
fix: history event decoding status

lukicenturi:
- [ ] Fix issue with history event decoding status
- [ ] Rename `Non Fungible Balances` to `NFT Balances`


kelsos (approved):
lgtm thanks

PR #9285, opened 2025-01-21
accurately track prices of spark.fi token balances

prettyirrelevant:


rotkehlchen/chain/evm/decoding/spark/utils.py
```diff
@@ -0,0 +1,24 @@
...
+        inquirer: 'Inquirer',
+        evm_inquirer: 'EvmNodeInquirer',
+) -> Price:
+    assert token.underlying_tokens is not None and len(token.underlying_tokens) == 1, 'Spark token must have exactly one underlying token.'  # noqa: E501
+    return inquirer.find_usd_price(
+        asset=Asset(evm_address_to_identifier(
+            address=token.underlying_tokens[0].address,
+            chain_id=evm_inquirer.chain_id,
+        )),
+    )
```
yabirgb, line 24: is this just querying the price of the underlying token?

rotkehlchen/inquirer.py
```diff
@@ -251,6 +253,12 @@ def get_underlying_asset_price(token: EvmToken) -> tuple[Price | None, CurrentPr
             inquirer=Inquirer(),
             evm_inquirer=Inquirer.get_evm_manager(chain_id=token.chain_id).node_inquirer,
         )
+    elif token.protocol == CPT_SPARK:
+        price = get_spark_token_price(
+            token=token,
+            inquirer=Inquirer(),
+            evm_inquirer=Inquirer.get_evm_manager(chain_id=token.chain_id).node_inquirer,
+        )
```
yabirgb, line 261: The logic in the inquirer already call this

```python
if is_known_protocol is True or underlying_tokens is not None:
                if (
                    underlying_tokens is not None and
                    asset.evm_address in (x.address for x in underlying_tokens)
                ):
                    Inquirer._msg_aggregator.add_error(
                        f'Token {asset} has itself as underlying token. Please edit the '
                        'asset to fix it. Price queries will not work until this is done.',
                    )
                else:
                    result, oracle = get_underlying_asset_price(asset)
                    if result is not None:
                        Inquirer.set_cached_price(
                            cache_key=cache_key,
                            cached_price=CachedPriceEntry(
                                price=(usd_price := Price(result)),
                                time=ts_now(),
                                oracle=oracle,
                            ),
                        )
                        return usd_price, oracle
                # else known protocol on-chain query failed. Continue to external oracles
```

that is this function and at the end it has

```python
    if token.underlying_tokens is not None:
        usd_price = ZERO
        for underlying_token in token.underlying_tokens:
            token = EvmToken(underlying_token.get_identifier(parent_chain=token.chain_id))
            underlying_asset_price, oracle = Inquirer.find_usd_price_and_oracle(token)
            usd_price += underlying_asset_price * underlying_token.weight

        if usd_price != ZERO_PRICE:
            price = Price(usd_price)
```

Isn't it enought with having the underlying token price query?

PR #9286, opened 2025-01-21
Improve decoding of erc721 approval and transfer

nicholasyoder:
Fixes an issue I noticed where uniswap transactions dealing with erc721 position tokens was not being decoded correctly. 

Improves the handling of erc721 tokens in the decoding of both approval events and transfer events. Also, when new tokens are encountered here, it will include the collectible id in the identifier when it creates the token.


rotkehlchen/tests/unit/decoders/test_uniswapv3.py
```diff
@@ -1030,3 +1031,65 @@ def test_swap_on_binance_sc(
...
     ]
+
+
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('arbitrum_one_accounts', [['0x706A70067BE19BdadBea3600Db0626859Ff25D74']])
+def test_receive_position_token_on_arbitrum(
+        arbitrum_one_inquirer: 'ArbitrumOneInquirer',
+        arbitrum_one_accounts: list['ChecksumEvmAddress'],
+) -> None:
+    tx_hash = deserialize_evm_tx_hash('0xbdfe4bfba8932fafed46a0cae44a945088c2496e1ef3c643d315159f900c338b')  # noqa: E501
```
yabirgb, line 1042: as pointed in discord the order of the events is wrong

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -880,6 +885,16 @@ def _get_eth_transfer_event(self, tx: EvmTransaction) -> Optional['EvmEvent']:
             counterparty=counterparty,
         )
 
+    def _get_transfer_or_approval_token_kind(self, tx_log: EvmTxReceiptLog) -> EvmTokenKind | None:
+        """Determine if a transfer or approval event is for an erc20 or erc721 token."""
+        if self._is_non_conformant_erc721(tx_log.address) or len(tx_log.topics) == 4:  # typical ERC721 has 3 indexed args  # noqa: E501
+            return EvmTokenKind.ERC721
+        elif len(tx_log.topics) == 3:  # typical ERC20 has 2 indexed args
+            return EvmTokenKind.ERC20
```
yabirgb, line 893: is there a reason to not put the erc20 before? I guess is the most common type of token and you save 2 checks for any erc20

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -889,25 +904,18 @@ def _maybe_decode_erc20_approve(
...
-            except NotERC20Conformant:
-                return DEFAULT_DECODING_OUTPUT  # ignore non-ERC20 approval for now
-
-        if len(tx_log.topics) == 3:
+        collectible_id = None
+        if len(tx_log.topics) in (3, 4):
             owner_address = bytes_to_address(tx_log.topics[1])
             spender_address = bytes_to_address(tx_log.topics[2])
+            if len(tx_log.topics) == 4:
+                collectible_id = str(int.from_bytes(tx_log.topics[3]))
```
yabirgb, line 918:
```suggestion
            collectible_id = str(int.from_bytes(tx_log.topics[3])) if len(tx_log.topics) == 4 else None
```
I believe is not possible to avoid checking twice unless you unroll and duplicate... what doesn't sound bad to me tbh but and if check is not too much

rotkehlchen/chain/evm/decoding/uniswap/v3/decoder.py
```diff
@@ -531,7 +531,7 @@ def _maybe_enrich_liquidity_pool_creation(
     ) -> TransferEnrichmentOutput:
         """This method enriches Uniswap V3 LP creation transactions."""
         if (
-            context.event.asset == self.uniswap_v3_nft and
+            self.uniswap_v3_nft in context.event.asset.identifier and  # asset identifier will also include the token id  # noqa: E501
```
yabirgb, line 534: This is hard to follow because there is so much implicit knowledge here 🙈 

I would create a function where we have all the `evm_address_to_identifier` functions with

```
def identifier_belongs_to_collection(asset_identifier, collection_identifier):
      # some good docstring explaining that the initial part is the same and the only difference is the nft id
      return asset_identifier.startswith(collection_identifier)
```

rotkehlchen/tests/unit/decoders/test_juicebox.py
```diff
@@ -56,7 +56,7 @@ def test_donation(ethereum_inquirer, ethereum_accounts):
         location=Location.ETHEREUM,
         event_type=HistoryEventType.RECEIVE,
         event_subtype=HistoryEventSubType.NONE,
-        asset=Asset('eip155:1/erc721:0x723932B58a7c6AEf036d1Fe17654E845d0f0fae5'),
+        asset=Asset('eip155:1/erc721:0x723932B58a7c6AEf036d1Fe17654E845d0f0fae5/4000000011'),
```
yabirgb, line 59: !!!! UPS this is bad isn't it? Because we will have tokens that have the wrong id in the database. This needs more thinking I guess or create a db migration to fix the ones that have the wrong id. But this also affects the user db since the events are there

rotkehlchen/tests/unit/decoders/test_uniswapv3.py
nicholasyoder: As discussed in discord, the tx used in this test is actually interacting with a protocol we don't support, so the events are simply decoded as transfers and are not reordered at all. I've added a docstring to the test explaining this.

rotkehlchen/tests/unit/decoders/test_juicebox.py
nicholasyoder: As discussed in discord, I've added a db upgrade for this.

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -30,4 +32,29 @@ def _add_new_locations(write_cursor: 'DBCursor') -> None:
     def _clean_cache(write_cursor: 'DBCursor') -> None:
         write_cursor.execute("DELETE FROM key_value_cache WHERE name LIKE 'extrainternaltx_%'")
 
+    @progress_step(description='Updating ERC721 token identifiers.')
+    def _update_erc721_token_identifiers(write_cursor: 'DBCursor') -> None:
+        """Update erc721 token identifiers to include the token id
+        for all evm events that have the token id in extra_data.
+        """
+        write_cursor = write_cursor.execute("SELECT asset, extra_data FROM history_events WHERE entry_type = 2 AND extra_data != ''")  # noqa: E501
```
yabirgb, line 40: mmm this won't work the moment you chain it with a previous upgrade that deletes the history events. Need to think it

rotkehlchen/tests/unit/decoders/test_basenames.py
```diff
@@ -402,16 +402,13 @@ def test_basenames_transfer_name(database, base_inquirer, action, base_accounts,
...
         balance=Balance(amount=ONE),
         location_label=from_address,
         notes=notes,
         counterparty=CPT_BASENAMES,
         address=string_to_evm_address(to_address),
-        extra_data={
-            'token_id': 112426549028048856546593988202926666418642845280196262619695088491431122056723,  # noqa: E501
-            'token_name': 'Basenames',
-        },
+        extra_data={'token_id': token_id, 'token_name': 'Basenames'},
```
yabirgb, line 411: also you are storing redundant information now?

rotkehlchen/tests/unit/decoders/test_basenames.py
```diff
@@ -134,13 +134,13 @@ def test_basenames_register(
...
             event_subtype=HistoryEventSubType.RECEIVE,
-            asset=Asset('eip155:8453/erc721:0x03c4738Ee98aE44591e1A4A4F3CaB6641d95DD9a'),
+            asset=Asset(f'eip155:8453/erc721:0x03c4738Ee98aE44591e1A4A4F3CaB6641d95DD9a/{token_id}'),
             balance=Balance(amount=ONE),
             location_label=user_address,
             notes=f'Receive Basenames name yabir.base.eth from {ZERO_ADDRESS} to {user_address}',
             counterparty=CPT_BASENAMES,
             address=ZERO_ADDRESS,
-            extra_data={'token_id': 26612040215479394739615825115912800930061094786769410446114278812336794170041, 'token_name': 'Basenames'},  # noqa: E501
+            extra_data={'token_id': token_id, 'token_name': 'Basenames'},
```
yabirgb, line 143: mmm isn't storing the token_id duplicating information? Since you have it in the asset identifier

rotkehlchen/db/upgrades/v46_v47.py
nicholasyoder: Leaving this as it is as suggested in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1331609707110207578)

rotkehlchen/tests/unit/decoders/test_basenames.py
nicholasyoder: Yes, both token_id and token_name are stored in the asset itself now. Removed both from extra_data.

When I was removing this from `evm/decoding/base.py` I realized that it was also retrieving the token id for non-conformant tokens there (https://github.com/rotki/rotki/pull/9286/files#diff-e349e4a4c646f70f1d1776d7bb16b8c42ba2eac19dcb86b76ed8107adfa84fa7L182), so I reworked the code in `evm/decoding/decoder.py` to also retrieve that, and combined retrieving the token kind and token id into `_get_transfer_or_approval_token_kind_and_id` to avoid repeating code.

rotkehlchen/tests/unit/decoders/test_basenames.py
nicholasyoder: yes - see response to other comment.

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -30,4 +32,35 @@ def _add_new_locations(write_cursor: 'DBCursor') -> None:
...
+                assets_updates.append((f'{asset_identifier}/{token_id}', asset_identifier))
+                # remove the token id and name from extra_data to avoid storing redundant data,
+                extra_data.pop('token_id', '')
+                extra_data.pop('token_name', '')
+                extra_data_updates.append((json.dumps(extra_data), entry[0]))
+
+        if len(assets_updates) > 0:
+            assets_query = 'UPDATE assets SET identifier=? WHERE identifier=?'
+            with GlobalDBHandler().conn.write_ctx() as global_db_write_cursor:
+                global_db_write_cursor.executemany(assets_query, assets_updates)
```
LefterisJP, line 61: do we also need to do any changes in the packaged global DB?

rotkehlchen/tests/unit/decoders/test_uniswapv3.py
```diff
@@ -1030,3 +1025,71 @@ def test_swap_on_binance_sc(
...
+    The approval event is on an untracked address, but is what creates the token initially.
+    Regression test for https://github.com/rotki/rotki/pull/9286
+    Since this transaction is interacting with the currently unsupported Krystal ALM protocol's
+    V3Utils contract (https://alm-docs.krystal.app/resources/smart-contracts), these events are
+    simply decoded as plain transfers and are not ordered correctly.
+    """
+    tx_hash = deserialize_evm_tx_hash('0xbdfe4bfba8932fafed46a0cae44a945088c2496e1ef3c643d315159f900c338b')  # noqa: E501
+    events, _ = get_decoded_events_of_transaction(evm_inquirer=arbitrum_one_inquirer, tx_hash=tx_hash)  # noqa: E501
+    user_address, timestamp, gas_amount, spend_amount, receive_amount = arbitrum_one_accounts[0], TimestampMS(1718052436000), '0.00002126616', '100.768463', '0.000044879191744967'  # noqa: E501
+    v3utils_contract = string_to_evm_address('0x3991bA795fb13a7646a8745F90b0F24Ed2443b03')
```
LefterisJP, line 1046: combine this with the above line

rotkehlchen/chain/evm/decoding/base.py
```diff
@@ -8,6 +8,7 @@
 from rotkehlchen.chain.evm.constants import ETH_SPECIAL_ADDRESS, ZERO_ADDRESS
 from rotkehlchen.chain.evm.decoding.constants import OUTGOING_EVENT_TYPES
 from rotkehlchen.constants import ONE, ZERO
+from rotkehlchen.constants.resolver import identifier_to_collectible_id
```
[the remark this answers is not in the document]
nicholasyoder: Renamed it. There were also a couple places where the result from this was called `token_id` so I renamed those to `collectible_id` to avoid confusion since `tokenid` is now used in the function name to refer to the the token's asset identifier.

rotkehlchen/chain/evm/decoding/ens/decoder.py
```diff
@@ -85,7 +86,7 @@ def _decode_name_transfer(self, context: DecoderContext) -> DecodingOutput:
         if transfer_event is None:  # Can happen if neither from/to is tracked
             return DEFAULT_DECODING_OUTPUT
 
-        label_hash = '0x{:064x}'.format(transfer_event.extra_data['token_id'])  # type: ignore[index]  # ERC721 transfer always has extra data. This code is to transform the int token id to a 32 bytes hex label hash
+        label_hash = f'0x{collectible_id:064x}'  # Transform the int token id to a 32 bytes hex label hash  # noqa: E501
```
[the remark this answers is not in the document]
nicholasyoder: ruff complains without it:
```
rotkehlchen/chain/evm/decoding/ens/decoder.py:89:100: E501 Line too long (106 > 99)
```
I'm not sure why it was fine before when the line was much longer... maybe when there's a type ignore ruff also ignores line length?

rotkehlchen/db/upgrades/v46_v47.py
nicholasyoder: No. As far as I can tell we don't have any erc721 tokens there - `SELECT * FROM assets WHERE identifier LIKE '%erc721%';` returns nothing.

rotkehlchen/chain/evm/decoding/ens/decoder.py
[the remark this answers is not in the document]
LefterisJP: > I'm not sure why it was fine before when the line was much longer... maybe when there's a type ignore ruff also ignores line length?

yes I think that is the case.

But the reason I thought it's fine to remove, is it's shorter now and looked ~96 to me. If it complaines it means it's more than 99 so it's fine to stay

LefterisJP (approved):
LGTM

PR #9288, opened 2025-01-22
add historical net worth graph data endpoint

prettyirrelevant:


rotkehlchen/balances/historical.py
```diff
@@ -120,6 +121,67 @@ def get_assets_amounts(
...
+        the total net worth in the user's main currency for each day.
+
+        May raise:
+            - NotFoundError if no events exist in the specified time period.
+
+        Returns:
+            - A mapping of timestamps to net worth in main currency for each day
+            - A list of (asset id, timestamp) tuples where price data was missing
+        """
+        events, main_currency = self._get_events_and_currency(from_ts=from_ts, to_ts=to_ts)
```
yabirgb, line 141: Regarding this function it is very expensive since it loads all the events into memery and initializes a lot of stuff that we ignore. It would be good to have a lighter version but I would add it as a todo here

rotkehlchen/balances/historical.py
```diff
@@ -120,6 +121,67 @@ def get_assets_amounts(
...
+        if len(events) == 0:
+            raise NotFoundError(f'No historical data found within {from_ts=} and {to_ts=}.')
+
+        current_balances: dict[Asset, FVal] = defaultdict(FVal)
+        daily_balances: dict[Timestamp, dict[Asset, FVal]] = defaultdict(lambda: defaultdict(FVal))
```
yabirgb, line 146: this doesn't need to be a defaultdict as you use it right now unless I'm missing something

rotkehlchen/balances/historical.py
```diff
@@ -120,6 +121,67 @@ def get_assets_amounts(
...
+                direction == EventDirection.NEUTRAL
+            ):
+                continue
+
+            if direction == EventDirection.IN:
+                current_balances[event.asset] += event.balance.amount
+            elif direction == EventDirection.OUT:
+                current_balances[event.asset] -= event.balance.amount
+
+            daily_balances[timestamp_to_daystart_timestamp(ts_ms_to_sec(event.timestamp))] = current_balances.copy()  # noqa: E501
```
yabirgb, line 159: you are copying a dict that can be big several times. Can't you keep track of the last day, current day and 

```
if current_day < this_event_day:
     daily_balances[timestamp_to_daystart_timestamp(ts_ms_to_sec(event.timestamp))] = current_balances.copy()
```

you don't need to copy always, only when you change from one day to the next

rotkehlchen/balances/historical.py
```diff
@@ -120,6 +121,67 @@ def get_assets_amounts(
...
+            if (
+                (direction := event.maybe_get_direction()) is None or
+                direction == EventDirection.NEUTRAL
+            ):
+                continue
+
+            if direction == EventDirection.IN:
+                current_balances[event.asset] += event.balance.amount
+            elif direction == EventDirection.OUT:
+                current_balances[event.asset] -= event.balance.amount
```
yabirgb, line 157: when you have ZERO here for one asset (say that you spend all what you had) don't you want to remove it from the dict? Less stuff to copy from day to day

rotkehlchen/balances/historical.py
```diff
@@ -120,6 +121,67 @@ def get_assets_amounts(
...
+                    continue
+
+                if (result := GlobalDBHandler.get_historical_price(
+                    from_asset=asset,
+                    timestamp=day_ts,
+                    to_asset=main_currency,
+                    max_seconds_distance=DAY_IN_SECONDS,
+                )) is None:  # Skip this asset's value for the day's total if we can't price it # noqa: E501
+                    missing_price_points.append((asset.identifier, day_ts))
+                    continue
```
yabirgb, line 177: each one of those is a trip to the globaldb... I fear this can be really slow. Can't we batch it somehow for a whole day? So create a function that queries several assets at once. Imagine you have 100 tokens and you do this for 720 days. That is 720 * 120 queries to the db when you can do only 720.

rotkehlchen/balances/historical.py
```diff
@@ -120,6 +121,67 @@ def get_assets_amounts(
 
         return amounts, last_event_id
 
+    def get_historical_netvalue(
+            self,
+            from_ts: Timestamp,
+            to_ts: Timestamp,
```
yabirgb, line 127: this logic processes only history events. What about trades and asset movements? Those also affect the net worth

rotkehlchen/balances/historical.py
prettyirrelevant: added a comment here

rotkehlchen/balances/historical.py
prettyirrelevant: i didn't find a clean way to do this and trying different variants resulted in wrong calculations.
```py
 timestamp = ts_ms_to_sec(event.timestamp) if isinstance(event, HistoryBaseEntry) else event.timestamp  # noqa: E501
  current_day = timestamp_to_daystart_timestamp(timestamp)
  if last_day is not None and current_day > last_day:
      daily_balances[last_day] = current_balances.copy()

  last_day = current_day

if last_day is not None:
  daily_balances[last_day] = current_balances.copy()
```

yabirgb (approved):

rotkehlchen/balances/historical.py
```diff
@@ -1,29 +1,131 @@
...
+            if current_balances[self.quote_asset] == ZERO:
+                del current_balances[self.quote_asset]
+        else:
+            if current_balances[self.base_asset] - self.amount < ZERO:
+                return self.identifier
+
+            current_balances[self.base_asset] -= self.amount
+            current_balances[self.quote_asset] += self.amount * self.rate
+            if current_balances[self.base_asset] == ZERO:
+                del current_balances[self.base_asset]
```
yabirgb, line 72: this two look really similar. Isn't it possible to make it a single function method that you call passing the arguments in the right order?

rotkehlchen/balances/historical.py
```diff
@@ -1,29 +1,131 @@
...
+        """
+        if self.trade_type in {TradeType.BUY, TradeType.SETTLEMENT_BUY}:
+            quote_amount = self.amount * self.rate
+            if current_balances[self.quote_asset] - quote_amount < ZERO:
+                return self.identifier
+
+            current_balances[self.base_asset] += self.amount
+            current_balances[self.quote_asset] -= quote_amount
+            if current_balances[self.quote_asset] == ZERO:
+                del current_balances[self.quote_asset]
```
yabirgb, line 64: I believe that since it is a defaultdict the `del` is not dangerous

rotkehlchen/tests/api/test_historical_balances.py
```diff
@@ -347,3 +346,239 @@ def test_get_historical_asset_balance_without_premium(rotkehlchen_api_server: 'A
...
+    assert result['times'][0] == day1_ts
+    assert FVal(result['values'][0]) == FVal('10') * eth_price
+
+
+@pytest.mark.parametrize('start_with_valid_premium', [True])
+def test_get_historical_netvalue_with_negative_amount(
+        rotkehlchen_api_server: 'APIServer',
+        setup_historical_data: None,
+        globaldb: 'GlobalDBHandler',
+) -> None:
```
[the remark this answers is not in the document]
prettyirrelevant: > Does that case have any overhead?

not at all.

rotkehlchen/balances/historical.py
prettyirrelevant: yeah yeah.

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +209,250 @@ def get_assets_amounts(
...
+                trade_filters.append(DBNestedFilter(
+                    and_op=False,
+                    filters=[
+                        DBMultiStringFilter(and_op=False, column='base_asset', values=asset_identifiers, operator='IN'),  # noqa: E501
+                        DBMultiStringFilter(and_op=False, column='quote_asset', values=asset_identifiers, operator='IN'),  # noqa: E501
+                    ],
+                ))
+
+            trade_where_clauses = []
+            trade_bindings: Any = []
```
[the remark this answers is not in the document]
prettyirrelevant: went with that because mypy raises an error

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +209,250 @@ def get_assets_amounts(
...
+            )]
+            if assets is not None:
+                event_filters.append(DBMultiStringFilter(
+                    and_op=True,
+                    column='asset',
+                    values=[asset.identifier for asset in assets],
+                    operator='IN',
+                ))
+            event_where_clauses = []
+            event_bindings: Any = []
```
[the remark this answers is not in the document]
prettyirrelevant: mypy complains 
`rotkehlchen/balances/historical.py:385: error: Need type annotation for "trade_bindings" (hint: "trade_bindings: list[<type>] = ...")  [var-annotated]
rotkehlchen/balances/historical.py:397: error: Need type annotation for "event_bindings" (hint: "event_bindings: list[<type>] = ...")  [var-annotated]
`

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +209,250 @@ def get_assets_amounts(
...
+            trades_iterator = cursor1.execute(
+                f'SELECT timestamp, id, type, base_asset, quote_asset, amount, rate '
+                f'FROM trades WHERE {" AND ".join(trade_where_clauses)} ORDER BY timestamp',
+                trade_bindings,
+            )
+
+            event_filters: list[DBFilter] = [DBTimestampFilter(
+                and_op=True,
+                to_ts=to_ts,
+                from_ts=from_ts,
```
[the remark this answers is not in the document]
prettyirrelevant: no need for the check, removed it.

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +260,227 @@ def get_assets_amounts(
...
+
+                    # Compare timestamps and process the earlier event
+                    if current_event[0] <= current_trade[0]:
+                        events.append(MinimalHistoryEvent.deserialize(current_event))
+                        current_event = next(events_iterator, None)
+                    else:
+                        events.append(MinimalTrade.deserialize(current_trade))
+                        current_trade = next(trades_iterator, None)
+                except (UnknownAsset, DeserializationError) as e:
+                    log.error(f'Failed to deserialize event due to {e!s}')
```
yabirgb, line 437: This should log also in what part of the logic you are, the name of the function or something. If you read this you wouldn't know

rotkehlchen/balances/historical.py
```diff
@@ -1,29 +1,182 @@
...
+        Returns the event_identifier if a negative balance would occur, else None.
+        For trades:
+        - BUY means we receive base_asset and spend quote_asset
+        - SELL means we spend base_asset and receive quote_asset
+        """
+        is_buy = self.trade_type in {TradeType.BUY, TradeType.SETTLEMENT_BUY}
+        spend_asset = self.quote_asset if is_buy else self.base_asset
+        receive_asset = self.base_asset if is_buy else self.quote_asset
+        spend_amount = self.amount * self.rate if is_buy else self.amount
+        receive_amount = self.amount if is_buy else self.amount * self.rate
```
yabirgb, line 94: donyou can use a normal if/else. Otherwise you are performing the same check 4 times

rotkehlchen/balances/historical.py
```diff
@@ -1,29 +1,187 @@
...
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
 
 
 class HistoricalBalance(TypedDict):
     amount: FVal
     price: FVal
 
 
+@dataclass(init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=True)
```
LefterisJP, line 46: why introduce and maintain another data structure for trades and for history events? Just to save memory?

It does not make sense to me.

Also if you ever wanted to do that that would mean you would need to introduce this at the source. And have Trades inherit this. And history events inherit the other one, no ? So you at least have a common data struct for both? And then inherit from minimal and inheirt from normal?

Otherwise you are duplicating structs here.

Please explain the logic. Do not just go about making changes in response to this comment before discussing this.

But in general this whole approach seems hacky.

rotkehlchen/balances/historical.py
```diff
@@ -1,29 +1,187 @@
...
+        )
+
+    @staticmethod
+    def _get_direction_from_event(
+            event_type: HistoryEventType,
+            event_subtype: HistoryEventSubType,
+    ) -> EventDirection | None:
+        """Determine the direction of a history event based on its type and subtype.
+
+        This was copied from rotkehlchen/history/events/structures/base.py
```
LefterisJP, line 149: Yep ... as I said. Hacky as fuck. you don't copy code like this. This alone should indicate that this approach is completely wrong.

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +265,230 @@ def get_assets_amounts(
...
+            if (day_ts := timestamp_to_daystart_timestamp(event.timestamp)) > current_day:
+                daily_balances[current_day] = current_balances.copy()
+                current_day = day_ts
+
+            if (negative_balance_id := event.update_balances(current_balances)) is not None:
+                negative_balance_event = negative_balance_id
                 break
 
-        return amounts, last_event_id
+        if current_day is not None and negative_balance_event is None:
```
LefterisJP, line 317: when can `current_day` be None here?

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +265,230 @@ def get_assets_amounts(
...
+            daily_balances[current_day] = current_balances.copy()
+
+        net_worth_per_day: dict[Timestamp, FVal] = {}
+        missing_price_points: list[tuple[str, Timestamp]] = []
+        for day_ts, asset_balances in daily_balances.items():
```
LefterisJP, line 322: add a comment for what you do in this loop. it's hard to read

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +265,230 @@ def get_assets_amounts(
...
+            assets_to_price = [
```
LefterisJP, line 323: Why are you calling this `assets_to_prices`? This is just a list of assets. What name were you trying to give here?

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +265,230 @@ def get_assets_amounts(
...
+                main_currency=main_currency,
+            )
+            missing_price_points.extend(missing)
+
+            day_total = sum(
+                balance * prices[asset]
+                for asset, balance in asset_balances.items()
+                if asset in prices and balance != ZERO
+            )
+            net_worth_per_day[day_ts] = FVal(day_total)
```
LefterisJP, line 342: Why do you need to do `FVal()` here? It should be FVal * Fval, so the result is already an FVal. If mypy can't find it that's mypy's problem. You should not call the constructor for no reason. You should figure out and fix the type error.

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +265,230 @@ def get_assets_amounts(
...
+            main_currency = self.db.get_setting(cursor=cursor1, name='main_currency')
+            trade_filters: list[DBFilter] = [
+                DBNotEqualFilter(and_op=True, column='location', value='B'),
+                DBTimestampFilter(and_op=True, from_ts=from_ts, to_ts=to_ts),
+            ]
+            event_filters: list[DBFilter] = [DBTimestampFilter(
+                and_op=True,
+                to_ts=to_ts,
+                from_ts=from_ts,
+                scaling_factor=FVal(1000),
```
LefterisJP, line 372: I know why this is there, but nothing to lose to explain in a super short comment

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +265,230 @@ def get_assets_amounts(
...
+            )]
+
+            if assets is not None:
```
LefterisJP, line 375: Why is assets optional in this function? Add a comment explaining what you do in this codeblock

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +265,230 @@ def get_assets_amounts(
...
+                f'SELECT CAST(timestamp / 1000 AS INTEGER) AS timestamp, event_identifier, asset, amount, type, subtype '  # noqa: E501
+                f'FROM history_events WHERE {" AND ".join(event_where_clauses)} ORDER BY timestamp',  # noqa: E501
+                event_bindings,
+            )
+
+            current_event = next(events_iterator, None)
+            current_trade = next(trades_iterator, None)
+            while current_event is not None or current_trade is not None:
+                try:
+                    # If there are no more trades, process the remaining events
```
LefterisJP, line 421:
```suggestion
                try:  # If there are no more trades, process the remaining events
```

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +265,230 @@ def get_assets_amounts(
...
+                    if current_trade is None:
+                        assert current_event is not None, 'current_event should never be None here'
```
LefterisJP, line 423: Why? What stops it? Explain.

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +265,230 @@ def get_assets_amounts(
...
+                event_where_clauses.extend(clauses)
+                event_bindings.extend(bindings)
+
+            events_iterator = cursor2.execute(
+                f'SELECT CAST(timestamp / 1000 AS INTEGER) AS timestamp, event_identifier, asset, amount, type, subtype '  # noqa: E501
+                f'FROM history_events WHERE {" AND ".join(event_where_clauses)} ORDER BY timestamp',  # noqa: E501
+                event_bindings,
+            )
+
+            current_event = next(events_iterator, None)
```
LefterisJP, line 417: So this gets events and trades one by one from the DB? To not hate it all in memory, right? Makes sense ... but would it not be better to have a rolling window of `N` events.

This way you won't have 99,999 DB queries for 99,999 eventsd. Just get them in batches.

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +265,230 @@ def get_assets_amounts(
...
+
+                    # Compare timestamps and process the earlier event
+                    if current_event[0] <= current_trade[0]:
+                        events.append(MinimalHistoryEvent.deserialize(current_event))
+                        current_event = next(events_iterator, None)
+                    else:
+                        events.append(MinimalTrade.deserialize(current_trade))
+                        current_trade = next(trades_iterator, None)
+                except (UnknownAsset, DeserializationError) as e:
+                    log.error(
```
LefterisJP, line 442: This should also be pushed upwards to the user as the final result of the graph would be off. If you can't deserialize an event then your result will be off. And if you just log it then 95% of the user's wont see it unless they get something in the  api response, or as part of a websocket message.

So this should stop the processing and propagate the error the api response.

rotkehlchen/balances/historical.py
```diff
@@ -99,63 +265,230 @@ def get_assets_amounts(
...
-                (direction := event.maybe_get_direction()) is None or
-                direction == EventDirection.NEUTRAL
-            ):
-                continue
+    def _get_prices(
+            assets: list[Asset],
+            main_currency: Asset,
+            timestamp: Timestamp,
+    ) -> tuple[dict[Asset, FVal], list[tuple[str, Timestamp]]]:
+        """Gets historical prices for multiple assets at once.
```
LefterisJP, line 456:
```suggestion
        """Gets cached historical prices for multiple assets at once.
```

LefterisJP (changes requested):
requesting changes since for some reason yabir put an approval which makes it look mergeable when looking at the list of PR reviews

rotkehlchen/balances/historical.py
prettyirrelevant: it cannot be `None` and is no-op. leftover from yabir's review. removed it now.

rotkehlchen/balances/historical.py
prettyirrelevant: > Why are you calling this assets_to_prices? This is just a list of assets. What name were you trying to give here?

yeah, my bad. renamed it to `assets`

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +121,318 @@ def get_assets_amounts(
...
+                        process_event = False
+                    else:  # compare timestamps
+                        trade_ts = trade_row[1]
+                        event_ts = ts_ms_to_sec(event_row[4])
+                        process_event = event_ts <= trade_ts
+
+                    if process_event:
+                        # we do not need the `entry_type` so we remove it from the tuple
+                        event = HistoryEvent.deserialize_from_db(events_buffer.popleft()[1:])
+                        events.append(event)
```
LefterisJP, line 323:
```suggestion
                        events.append(HistoryEvent.deserialize_from_db(events_buffer.popleft()[1:]))
```

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +121,318 @@ def get_assets_amounts(
...
+                        if len(events_buffer) == 0:  # refill trades buffer if empty
+                            events_buffer.extend(cursor2.fetchmany(BATCH_SIZE))
+                    else:
+                        trade = Trade.deserialize_from_db(trades_buffer.popleft())
+                        events.append(trade)
```
LefterisJP, line 328:
```suggestion
                        events.append(Trade.deserialize_from_db(trades_buffer.popleft()))
```

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +121,318 @@ def get_assets_amounts(
...
+                        if len(trades_buffer) == 0:  # refill trades buffer if empty
+                            trades_buffer.extend(cursor1.fetchmany(BATCH_SIZE))
```
LefterisJP, line 330:
```suggestion
                            trades_buffer.extend(cursor1.fetchmany(BATCH_SIZE))

```

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +121,318 @@ def get_assets_amounts(
...
+            (direction := event.maybe_get_direction()) is None or
+            direction == EventDirection.NEUTRAL
+        ):
+            return None
+
+        if direction == EventDirection.IN:
+            current_balances[event.asset] += event.balance.amount
+        else:
+            if current_balances[event.asset] - event.balance.amount < ZERO:
+                return event.event_identifier
```
LefterisJP, line 432: So I just realized it's not `event.identifier` but it's `event.event_identifier`.

Can you make sure that this is what frontend will need to link it?

We need to change this naming. It's really confusing. One is for the group (the one you use) and the other for the single individual event.

i would guess you would need for the single event no?

rotkehlchen/balances/historical.py
```diff
@@ -35,16 +53,21 @@ def get_balances(self, timestamp: Timestamp) -> dict[Asset, HistoricalBalance]:
...
+        - DeserializationError if there is a problem deserializing an event or trade from DB.
         """
         events, main_currency = self._get_events_and_currency(to_ts=timestamp)
         if len(events) == 0:
             raise NotFoundError('No historical data found until the given timestamp.')
 
-        amounts = self._calculate_balance_from_events(events)
+        current_balances: dict[Asset, FVal] = defaultdict(FVal)
+        for event in events:
+            if self._update_balances(event, current_balances) is not None:
```
LefterisJP, line 64: you did not use kwargs here and marked [the comment](https://github.com/rotki/rotki/pull/9288#discussion_r1929153243) as resolved. Please do not do this.

rotkehlchen/balances/historical.py
```diff
@@ -61,29 +84,34 @@ def get_asset_balance(self, asset: Asset, timestamp: Timestamp) -> HistoricalBal
...
-        return {'amount': amounts[asset], 'price': price}
+        return {'amount': current_balances[asset], 'price': price}
 
     def get_assets_amounts(
             self,
             assets: tuple[Asset, ...],
             from_ts: Timestamp,
             to_ts: Timestamp,
-    ) -> tuple[dict[Timestamp, FVal], str | None]:
+    ) -> tuple[dict[Timestamp, FVal], tuple[str | int | None, str | None] | None]:
```
LefterisJP, line 114: You have to adjust the docstring explaining what the return value is now. As you changed it but the docstring is not updated.

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +122,341 @@ def get_assets_amounts(
...
         May raise:
         - NotFoundError if no events exist for the assets in the specified time period.
+        - DeserializationError if there is a problem deserializing an event or trade from DB.
         """
         events, _ = self._get_events_and_currency(from_ts=from_ts, to_ts=to_ts, assets=assets)
         if len(events) == 0:
             raise NotFoundError(f'No historical data found for {assets} within {from_ts=} and {to_ts=}.')  # noqa: E501
 
-        total_amount, last_event_id = ZERO, None
+        negative_balance_event = None
```
LefterisJP, line 131: It's not just an event. But it's a tuple with data

```suggestion
        negative_balance_data= None
```

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +122,341 @@ def get_assets_amounts(
...
+        current_balances: dict[Asset, FVal] = defaultdict(FVal)
         amounts: dict[Timestamp, FVal] = defaultdict(lambda: ZERO)
         for event in events:
-            if (
-                (direction := event.maybe_get_direction()) is None or
-                direction == EventDirection.NEUTRAL
-            ):
-                continue
+            if (event_ids := self._update_balances(event=event, current_balances=current_balances)) is not None:  # noqa: E501
+                negative_balance_event = event_ids
```
LefterisJP, line 136: Why have an extra variable here?

```suggestion
            if (negative_balance_data := self._update_balances(event=event, current_balances=current_balances)) is not None:  # noqa: E501
```

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +122,341 @@ def get_assets_amounts(
...
-            if total_amount < ZERO:
-                last_event_id = event.event_identifier
+        May raise:
+            - NotFoundError if no events exist in the specified time period.
+            - DeserializationError if there is a problem deserializing an event or trade from DB.
+
+        Returns:
+            - A mapping of timestamps to net worth in main currency for each day
+            - A list of (asset id, timestamp) tuples where price data was missing
+            - The event_identifier that caused a negative balance if any, else None
```
LefterisJP, line 167:
```suggestion
            - UPDATE ME
```

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +122,341 @@ def get_assets_amounts(
...
+        """
+        events, main_currency = self._get_events_and_currency(from_ts=from_ts, to_ts=to_ts)
+        if len(events) == 0:
+            raise NotFoundError(f'No historical data found within {from_ts=} and {to_ts=}.')
+
+        negative_balance_event = None
```
LefterisJP, line 173:
```suggestion
        negative_balance_data = None
```

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +122,341 @@ def get_assets_amounts(
...
+            if (day_ts := timestamp_to_daystart_timestamp(
+                event.timestamp
+                if isinstance(event, Trade)
+                else ts_ms_to_sec(event.timestamp),
+            )) > current_day:
+                daily_balances[current_day] = current_balances.copy()
+                current_day = day_ts
+
+            if (event_ids := self._update_balances(event=event, current_balances=current_balances)) is not None:  # noqa: E501
+                negative_balance_event = event_ids
```
LefterisJP, line 192: same as above

```suggestion
            if (negative_balance_data := self._update_balances(event=event, current_balances=current_balances)) is not None:  # noqa: E501
```

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +122,341 @@ def get_assets_amounts(
...
                 break
 
-        return amounts, last_event_id
+        if negative_balance_event is None:
```
LefterisJP, line 195: add this as an else to the above for loop. With a comment explaining that no negative balance was encountered

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +122,341 @@ def get_assets_amounts(
...
+                    from_ts=from_ts,
+                    scaling_factor=FVal(1000),  # timestamp of events are stored in ms
+                ), DBIgnoredAssetsFilter(
+                    and_op=True,
+                    asset_key='asset',
+                    operator='NOT IN',
                 ),
-                has_premium=True,
+            ]
+            if assets is not None:
```
LefterisJP, line 285:
```suggestion
            if assets is not None:  # filter for the specific assets
```

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +122,341 @@ def get_assets_amounts(
...
+                ))
+                event_filters.append(DBMultiStringFilter(
+                    and_op=True,
+                    column='asset',
+                    values=asset_identifiers,
+                    operator='IN',
+                ))
+
+            trade_where_clauses = []
+            trade_bindings: list[Any] = []
```
LefterisJP, line 302: combine in one line

```suggestion
            trade_where_clauses = []
            trade_bindings: list[Any] = []
```

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +122,341 @@ def get_assets_amounts(
...
+            for filter_ in trade_filters:
+                clauses, bindings = filter_.prepare()
+                trade_where_clauses.extend(clauses)
+                trade_bindings.extend(bindings)
+
+            event_where_clauses = []
+            event_bindings: list[Any] = []
```
LefterisJP, line 309: combine in one line

```suggestion
            event_where_clauses = []
            event_bindings: list[Any] = []
```

rotkehlchen/balances/historical.py
```diff
@@ -94,68 +122,341 @@ def get_assets_amounts(
...
+
+            current_balances[spend_asset] -= spend_amount
+            current_balances[receive_asset] += receive_amount
+            if current_balances[spend_asset] == ZERO:
+                del current_balances[spend_asset]
+
+            return None
+
+        # history event
+        if (
```
LefterisJP, line 446:
```suggestion
        if (  # history event logic from here and down
```

LefterisJP (approved):
LGTM

PR #9292, opened 2025-01-22
fix: check for ping before starting colibri

kelsos:
This is to avoid an error in colibri if the global db is not yet initialized.

Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/electron/main/subprocess-handler.ts
+++ b/frontend/app/electron/main/subprocess-handler.ts
@@ -1,5 +1,7 @@
 import * as os from 'node:os';
 import process from 'node:process';
+import { URL } from 'node:url';
+import http from 'node:http';
 import { app } from 'electron';
 import { psList } from '@electron/main/ps-list';
 import { getPortAndUrl } from '@electron/main/port-utils';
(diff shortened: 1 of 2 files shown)


frontend/app/electron/main/subprocess-handler.ts
```diff
@@ -167,6 +176,68 @@ export class SubprocessHandler {
...
+   * @param {number} [retries=10] - The maximum number of ping attempts before giving up.
+   * @param {number} [waitSeconds=6] - The delay, in seconds, between consecutive ping attempts.
+   * @return {Promise<boolean>} A promise that resolves to true if the API responds successfully within the given attempts, otherwise false.
+   */
+  private async checkCoreApiAvailability(
+    url: string,
+    retries: number = 10,
+    waitSeconds: number = 6,
+  ): Promise<boolean> {
+    await wait(3000);
```
yabirgb, line 195: why this wait of 3 secs? Let the app enough time for the app to run? Shouldn't it be less?

kelsos: Yes so if the code runs immediately the backend won't give a proper response because it is close to spawning, so it would have to wait a bit. 

When I was trying locally it was around 2ish seconds so 3 just to be in time.

That being said, there is a problem that we need to think about both this and the docker and this is when do we decide to bail out.

PR #9294, opened 2025-01-22
Detect safes when adding account

yabirgb:
This PR allows for a user to add a safe account to all the supported EVM chains at once.

Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=94775586

### DIFF
--- a/rotkehlchen/chain/aggregator.py
+++ b/rotkehlchen/chain/aggregator.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING, Any, Literal, Optional, TypeVar, cast, get_arg
 
 import requests
 from gevent.lock import Semaphore
+from web3 import Web3
 from web3.exceptions import BadFunctionCallOutput, Web3Exception
 
 from rotkehlchen.accounting.structures.balance import Balance, BalanceSheet
@@ -32,6 +33,7 @@ from rotkehlchen.chain.bitcoin import get_bitcoin_addresses_balances
 from rotkehlchen.chain.bitcoin.bch import get_bitcoin_cash_addresses_balances
 from rotkehlchen.chain.bitcoin.bch.utils import force_address_to_legacy_address
 from rotkehlchen.chain.bitcoin.xpub import XpubManager
+from rotkehlchen.chain.constants import SAFE_BASIC_ABI
 from rotkehlchen.chain.ethereum.defi.chad import DefiChad
 from rotkehlchen.chain.ethereum.defi.structures import DefiProtocolBalances
 from rotkehlchen.chain.ethereum.modules import (
@@ -139,6 +141,7 @@ if TYPE_CHECKING:
     from rotkehlchen.chain.ethereum.modules.uniswap.uniswap import Uniswap
     from rotkehlchen.chain.evm.decoding.types import CounterpartyDetails
     from rotkehlchen.chain.evm.manager import EvmManager
+    from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
     from rotkehlchen.chain.gnosis.manager import GnosisManager
     from rotkehlchen.chain.optimism.manager import OptimismManager
     from rotkehlchen.chain.polygon_pos.manager import PolygonPOSManager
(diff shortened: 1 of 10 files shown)


rotkehlchen/chain/aggregator.py
```diff
@@ -1665,7 +1692,10 @@ def add_accounts_to_all_evm(
         for account in accounts:
             existed_accounts += [(chain, account) for chain in SUPPORTED_EVM_EVMLIKE_CHAINS if account in self.accounts.get(chain)]  # noqa: E501
             # Distinguish between contracts and EOAs
-            if self.is_contract(account, SupportedBlockchain.ETHEREUM):
+            if (
+                self.is_contract(account, SupportedBlockchain.ETHEREUM) and
```
LefterisJP, line 1696: the problem with this is that the safe may have been created only in Gnosis and say Optimism. We have to check against the given EVM chain here.

rotkehlchen/chain/aggregator.py
```diff
@@ -1497,6 +1499,31 @@ def get_evm_manager(
     def is_contract(self, address: ChecksumEvmAddress, chain: SUPPORTED_EVM_CHAINS_TYPE) -> bool:
         return self.get_chain_manager(chain).node_inquirer.get_code(address) != '0x'
 
+    def is_safe_proxy(self, address: ChecksumEvmAddress, chain: SUPPORTED_EVM_CHAINS_TYPE) -> bool:
+        """Check if an address is a SAFE contract by checking the getThreshold, VERSION
+        and getChainId methods. We say that if a contract has the same methods as a safe
```
LefterisJP, line 1504:
```suggestion
        and getChainId methods. We assume that if a contract has the same methods as a safe
```

rotkehlchen/chain/aggregator.py
```diff
@@ -1497,6 +1499,35 @@ def get_evm_manager(
...
+            (address, contract.encode_abi(method_name))
+            for method_name in ('getThreshold', 'VERSION', 'getChainId')
+        ]
+        try:
+            outputs = manager.multicall_2(
+                calls=calls,
+                require_success=False,
+            )
+        except RemoteError as e:
+            log.error(f'Failed to check SAFE properties for {address} due to {e}. Skipping')
```
LefterisJP, line 1526:
```suggestion
            log.error(f'Failed to check SAFE properties for {address} in {chain} due to {e}. Skipping')
```

rotkehlchen/chain/aggregator.py
```diff
@@ -1497,6 +1499,35 @@ def get_evm_manager(
...
         return self.get_chain_manager(chain).node_inquirer.get_code(address) != '0x'
 
+    def is_safe_proxy_or_eoa(
+            self,
+            address: ChecksumEvmAddress,
+            chain: SUPPORTED_EVM_CHAINS_TYPE,
+    ) -> bool:
+        """Check if an address is a SAFE contract by checking the getThreshold, VERSION
+        and getChainId methods. We assume that if a contract has the same methods as a safe
+        then it is a safe.
```
LefterisJP, line 1509:
```suggestion
        """Check if an address is a SAFE contract or an EoA. We do this by checking the getThreshold, VERSION
        and getChainId methods. We assume that if a contract has the same methods as a safe
        then it is a safe. Also EoAs return XXXX for any method so this function will also return True.
```

rotkehlchen/chain/aggregator.py
```diff
@@ -1660,35 +1691,40 @@ def add_accounts_to_all_evm(
...
+            elif had_activity is False and len(chains_with_valid_addresses) != 0:
+                no_activity_accounts += [(chain, account) for chain in chains_with_valid_addresses]
 
         return (
             added_accounts,
             existed_accounts,
             failed_accounts,
             no_activity_accounts,
-            eth_contract_addresses,
+            evm_contract_addresses,
```
LefterisJP, line 1727: you need to modify the docstring in `detect_evm_accounts` right below my comment

rotkehlchen/chain/aggregator.py
```diff
@@ -1497,6 +1499,35 @@ def get_evm_manager(
...
+            chain: SUPPORTED_EVM_CHAINS_TYPE,
+    ) -> bool:
+        """Check if an address is a SAFE contract by checking the getThreshold, VERSION
+        and getChainId methods. We assume that if a contract has the same methods as a safe
+        then it is a safe.
+        """
+        manager: EvmNodeInquirer = self.get_chain_manager(chain).node_inquirer
+        contract = Web3().eth.contract(
+            address=address,
+            abi='[{"inputs":[],"name":"getChainId","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[],"name":"VERSION","outputs":[{"internalType":"string","name":"","type":"string"}],"stateMutability":"view","type":"function"},{"inputs":[],"name":"getThreshold","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"}]',
```
LefterisJP, line 1514: Can you put this as a constant somewhere? Also mark it with the appropriate type? When you do you will see that all `internalType` keys can go away.

rotkehlchen/rotkehlchen.py
```diff
@@ -675,7 +675,7 @@ def add_evm_accounts(
         list[tuple[SUPPORTED_EVM_EVMLIKE_CHAINS_TYPE, ChecksumEvmAddress]],
         list[tuple[SUPPORTED_EVM_EVMLIKE_CHAINS_TYPE, ChecksumEvmAddress]],
         list[tuple[SUPPORTED_EVM_EVMLIKE_CHAINS_TYPE, ChecksumEvmAddress]],
-        list[ChecksumEvmAddress],
```
LefterisJP, line 678: need to modify the docstring here

LefterisJP (approved):
LGTM

PR #9297, opened 2025-01-23
Fix yearn v2 deposit/withdrawal in special cases

LefterisJP:
The part of the decoder for yearn v2 that was based only on transfer was too open. It was matching on all transfer of the vault and its underlying token and as such the logic was breaking for more complicated transactions.


prettyirrelevant (approved):
lgtm

Issue #8978, opened 2024-11-29
Batch queries of prices for crypto assets


yabirgb:
At the moment the frontend queries prices from the backend using the prices endopoint. The prices endpoint (`api/1/assets/prices/latest`) takes a list of prices but it then does

```python
 for asset in assets:
            if asset != target_asset:
                if asset.asset_type == AssetType.NFT:
                    nft_price_data = self._eth_module_query(
                        module_name='nfts',
                        method='get_nfts_with_price',
                        query_specific_balances_before=None,
                    )
                    oracle = CurrentPriceOracle.MANUALCURRENT if nft_price_data['manually_input'] is True else CurrentPriceOracle.BLOCKCHAIN  # noqa: E501
                    assets_price[asset] = [Price(nft_price_data['usd_price']), oracle.value, False]
                else:
                    price, oracle, used_main_currency = Inquirer.find_price_and_oracle(
                        from_asset=asset,
                        to_asset=target_asset,
                        ignore_cache=ignore_cache,
                        match_main_currency=True,
                    )
                    assets_price[asset] = [price, oracle.value, used_main_currency]  # type: ignore  # mypy detects here the CurrentPriceOracle as nullable but is not by the definition of find_price_and_oracle
            else:
                assets_price[asset] = [Price(ONE), CurrentPriceOracle.BLOCKCHAIN.value, False]
```

so we are sequential with the prices. Since the oracles (coingecko, cryptocompare, defillama) allow us to query multiple assets we need:

1. Split the assets in NFTs and normal assets and have them queried following different paths
2. For the assets that we query using standard oracles:

Let's say that we have [manual, coingecko, defillama, uniswap]

- We would try to find price using manual. We create two lists `([assets that we could find their price], [assets that we failed])`. We save `[assets that we could find their price]` and
- Call coingecko with the assets in `[assets that we failed]` from the prev step. Notice that if the oracle uses the ids in the url we might need to split in batches to not be over the limit of urls. Likewise we would have two lists as output `([assets that we could find their price], [assets that we failed])`
- As we did for coingecko we would call defillama now with ` [assets that we failed]` from the prev step
- For uniswap we have to be careful. We do several queries looking for the routes but it might be possible to do a multicall. This needs research. Worst case it would be sequential over `[assets that we failed]`.

After iterating the oracles in the order that the user has selected we would end up with a set of assets and their prices and a set of assets that we couldn't retrieve a price for. Those will have price 0.

[...]

PR #9301, opened 2025-01-23
Add support for batch current price queries

nicholasyoder:
Related #8978

Implements support for batch current price queries in the inquirer, and coingecko (other oracles are simply queried sequentially for now, and will need modified in later PRs).

The bulk of the changes are in the inquirer's price querying logic. In order to avoid repeated code, single asset price queries and batch queries are handled by the same functions with a single asset query simply using a list with one element.

Also I had to rework the price mocking stuff in `rotkehlchen/tests/fixtures/accounting.py`, as its now the batch querying functions that call each other within the inquirer logic.

### DIFF
--- a/rotkehlchen/inquirer.py
+++ b/rotkehlchen/inquirer.py
@@ -286,7 +286,7 @@ def get_underlying_asset_price(token: EvmToken) -> tuple[Price | None, CurrentPr
 T = TypeVar('T', bound=Callable[..., Any])
 
 
-def handle_recursion_error(return_price_only: bool = False) -> Callable:
+def handle_recursion_error(return_price_only: bool = False, return_dict: bool = False) -> Callable:
     """
     In the app we saw that having a wrongly configured token which had itself
     as underlying token created a RecursionError. That recursion error was
@@ -294,10 +294,11 @@ def handle_recursion_error(return_price_only: bool = False) -> Callable:
     type from sqlite. The recursion error was not only seen here but also in the past
     so we've decided to catch both of them here.
 
-    This decorator is used with methods that return a tuple with the price or only the price.
-    To keep the decorator compatible with the expected returned type if `return_price_only` is
-    set to True we return only `Price(ZERO)` and if it's False we return a tuple in case of
-    error.
+    This decorator is used with methods that return either a single price, a tuple containing
+    a price and oracle, or a dict that maps assets to either prices or price and oracle tuples.
+    When there is an error, to keep the decorator compatible with the expected returned type,
+    if `return_dict` is True we simply return an empty dict, otherwise if `return_price_only`
+    is True we return only `Price(ZERO)` and if it's False we return a price and oracle tuple.
     """
     def decorator(func: T) -> T:
         @wraps(func)
@@ -305,8 +306,16 @@ def handle_recursion_error(return_price_only: bool = False) -> Callable:
             try:
                 result = func(*args, **kwargs)
             except (RecursionError, sqlite3.OperationalError) as e:
-                from_asset = kwargs.get('from_asset') or kwargs.get('asset')
                 to_asset = kwargs.get('to_asset')
+                if return_dict is True:
+                    from_assets = kwargs.get('from_assets') or kwargs.get('assets')
+                    log.error(
+                        f'Failed to query prices {from_assets=} {to_asset=} '
+                        f'due to a recursion error: {e}.',
+                    )
+                    return {}
+
+                from_asset = kwargs.get('from_asset') or kwargs.get('asset')
                 log.error(
                     f'Failed to query price {from_asset=} {to_asset=} due to a recursion error: '
                     f'{e}. Using zero as price.',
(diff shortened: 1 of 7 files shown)


rotkehlchen/inquirer.py
```diff
@@ -619,89 +633,109 @@ def _query_oracle_instances(
...
+            return found_prices
 
         if ignore_cache is False:
-            cache = Inquirer.get_cached_current_price_entry(cache_key=(from_asset, to_asset))
-            if cache is not None:
-                return cache.price, cache.oracle
+            for from_asset in unknown_prices.copy():
+                if (cache := Inquirer.get_cached_current_price_entry(cache_key=(from_asset, to_asset))) is not None:  # noqa: E501
+                    found_prices[from_asset] = cache.price, cache.oracle
+                    unknown_prices.remove(from_asset)
```
yabirgb, line 702: you an avoid creating a copy using another list but not sure what is best 🤔 remove is also costly irc

rotkehlchen/inquirer.py
```diff
@@ -619,89 +633,109 @@ def _query_oracle_instances(
...
 
         # check manual prices
-        if (price_result := Inquirer._try_oracle_price_query(
+        prices, unknown_prices = Inquirer._try_oracle_price_query(  # type: ignore[assignment]  # Manual current oracle uses Asset type
```
yabirgb, line 705: I don't get the message about the type ignore. Could you explain?

rotkehlchen/inquirer.py
```diff
@@ -619,89 +633,109 @@ def _query_oracle_instances(
...
             oracle=CurrentPriceOracle.MANUALCURRENT,
             oracle_instance=Inquirer._manualcurrent,
-            from_asset=from_asset,  # type: ignore[arg-type]  # Manual current oracle can works with Asset type
+            from_assets=unknown_prices,  # type: ignore[arg-type]  # Manual current oracle can works with Asset type
```
yabirgb, line 708:
```suggestion
            from_assets=unknown_prices,  # type: ignore[arg-type]  # Manual current oracle can work with Asset type
```

yabirgb: here unknown_price is maybe confusing? Something like references that this is a list of assets would be great but any name that I can think of is large. Maybe lefteris has a suggestion?

rotkehlchen/inquirer.py
```diff
@@ -619,89 +633,109 @@ def _query_oracle_instances(
...
+        if to_asset.is_fiat():
+            for from_asset in unknown_prices.copy():
+                if from_asset.is_fiat():
+                    with suppress(RemoteError):
+                        price, oracle = Inquirer._query_fiat_pair(
+                            base=from_asset.resolve_to_fiat_asset(),
+                            quote=to_asset.resolve_to_fiat_asset(),
+                        )
+                        found_prices[from_asset] = price, oracle
+                        unknown_prices.remove(from_asset)
```
yabirgb, line 723: here again there is a copy... could we avoid it somehow?

rotkehlchen/inquirer.py
```diff
@@ -766,136 +832,184 @@ def find_usd_price_and_oracle(
...
-            log.error(f'Tried to ask for {asset.identifier} price but asset is missing from the DB')  # noqa: E501
-            return ZERO_PRICE, CurrentPriceOracle.FIAT
+            if ignore_cache is False:
+                cache = Inquirer.get_cached_current_price_entry(cache_key=(asset, A_USD))
+                if cache is not None:
+                    found_prices[asset] = cache.price, cache.oracle
+                    continue
+
+            try:
+                asset = asset.resolve()  # noqa: PLW2901  # we do want to overwrite the loop variable  # noqa: E501
```
yabirgb, line 906: it happened before but why do we want to resolve here? Also maybe just use a variable resolved_asset? Is also more readable imo

nicholasyoder: Its using `isinstance` right after this to check if its a `FiatAsset` and also checks later if its an `EvmToken`. This doesn't work if the asset hasn't had `resolve()` called.

Changed the variable name as suggested.

rotkehlchen/inquirer.py
nicholasyoder: yeah `unknown_price` isn't a very good name here. I've replaced it with several other names, since its creating new lists a couple places instead of using `.copy()`. Not sure if the new names are terribly good either, but I think they're a bit of an improvement anyway.

rotkehlchen/inquirer.py
nicholasyoder: The manual price oracle can handle `Asset` instead of `AssetWithOracles`, but since all other oracles use `AssetWithOracles` that is what is used in  `_try_oracle_price_query`, so it has to do type ignores since here were using just `Asset` in this case.

With that said I'm now using a new `assets_without_manual_price` variable here instead of reusing `unknown_prices` from above so this type ignore isn't needed on this line anymore, but the ones for `from_assets` and `to_asset` still are needed and I also had to add one below for `non_fiat_assets = assets_without_manual_price`

rotkehlchen/inquirer.py
nicholasyoder: Using a new list called `assets_not_in_cache` here.

rotkehlchen/tests/unit/test_inquirer.py
```diff
@@ -1187,3 +1191,37 @@ def test_bsq_price(inquirer: 'Inquirer') -> None:
...
+    assert inquirer.find_usd_prices_and_oracles(
+        assets=[
+            A_ETC,
+            A_AAVE,
+            (a_air := Asset('eip155:1/erc20:0x27Dce1eC4d3f72C3E457Cc50354f1F975dDEf488')),
+        ],
+    ) == {
+        A_ETC: (FVal('26.38'), CurrentPriceOracle.COINGECKO),
+        A_AAVE: (FVal('315.63'), CurrentPriceOracle.COINGECKO),
+        a_air: (FVal('0.01301'), CurrentPriceOracle.CRYPTOCOMPARE),
```
nicholasyoder, line 1226: you mentioned in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1334566877443264594) to add an asset without a coingecko id that would use another oracle like defillama. I couldn't seem to find any right off that would go to defilama rather than cryptocompare - so just added one that uses cryptocompare.

rotkehlchen/inquirer.py
```diff
@@ -595,20 +602,27 @@ def _query_oracle_instances(
...
-            to_asset = to_asset.resolve_to_asset_with_oracles()
-            if skip_onchain:
-                oracles = instance._oracles_not_onchain
-                oracle_instances = instance._oracle_instances_not_onchain
+
+        found_prices: dict[Asset, tuple[Price, CurrentPriceOracle]] = {}
+        unknown_prices: list[AssetWithOracles] = []
+
+        for from_asset in from_assets:
+            if from_asset.is_asset_with_oracles() is True:
```
LefterisJP, line 610:
```suggestion
            if from_asset.is_asset_with_oracles():
```

rotkehlchen/inquirer.py
```diff
@@ -595,20 +602,27 @@ def _query_oracle_instances(
...
-            from_asset = from_asset.resolve_to_asset_with_oracles()
-            to_asset = to_asset.resolve_to_asset_with_oracles()
-            if skip_onchain:
-                oracles = instance._oracles_not_onchain
-                oracle_instances = instance._oracle_instances_not_onchain
+
+        found_prices: dict[Asset, tuple[Price, CurrentPriceOracle]] = {}
+        unknown_prices: list[AssetWithOracles] = []
+
+        for from_asset in from_assets:
```
LefterisJP, line 609: what are you doing in this loop? Why do you need to initialize those? At least add some comments

rotkehlchen/inquirer.py
```diff
@@ -766,136 +839,185 @@ def find_usd_price_and_oracle(
...
-        elif asset == A_ETH2:
-            asset = A_ETH
+        found_prices: dict[Asset, tuple[Price, CurrentPriceOracle]] = {}
 
-        cache_key = (asset, A_USD)
-        if ignore_cache is False:
-            cache = Inquirer.get_cached_current_price_entry(cache_key=cache_key)
-            if cache is not None:
-                return cache.price, cache.oracle
+        # Retrieve prices from the cache, and set
```
LefterisJP, line 897: and set what?

rotkehlchen/inquirer.py
```diff
@@ -766,136 +839,185 @@ def find_usd_price_and_oracle(
...
-                    Inquirer._msg_aggregator.add_error(
-                        f'Token {asset} has itself as underlying token. Please edit the '
-                        'asset to fix it. Price queries will not work until this is done.',
+        )
+        for asset, price in prices.items():
+            found_prices[asset] = price, CurrentPriceOracle.MANUALCURRENT
+
+        for asset in assets_without_price:
+            # Try and check if it is an ethereum token with specified protocol or underlying tokens
+            is_known_protocol = False
```
LefterisJP, line 944: things like these should be in another subfunction to be called in the for loop.  The code in this PR has become unreadable. 

The inquirer code was already hard to read and needed improvement. This PR makes it a lot worse.

rotkehlchen/interfaces.py
```diff
@@ -55,6 +56,29 @@ def query_current_price(
             Current price between from_asset and to_asset using this oracle's data
         """
 
+    def query_multiple_current_price(
```
LefterisJP, line 59:
```suggestion
    def query_multiple_current_prices(
```

rotkehlchen/tests/external_apis/test_coingecko.py
```diff
@@ -93,3 +93,15 @@ def test_coingecko_with_api_key(database):
 
     )
     assert result == FVal('3295.1477375227337')
+
+
+@pytest.mark.vcr
+def test_coingecko_query_multiple_current_price(session_coingecko: 'Coingecko'):
```
LefterisJP, line 99:
```suggestion
def test_coingecko_query_multiple_current_prices(session_coingecko: 'Coingecko'):
```

rotkehlchen/inquirer.py
nicholasyoder: Added a comment for this loop. Those variables need initialized since assets and prices are appended to them in the loops below. I did realize now that they don't actually need the type hints here, so removed that.

rotkehlchen/inquirer.py
```diff
@@ -498,91 +458,81 @@ def set_cached_price(cache_key: tuple[Asset, Asset], cached_price: CachedPriceEn
...
-            )
-        return price, is_error
+        failed_assets: list[AssetWithOracles] = []
+        for from_asset in from_assets:
+            if (price := prices.get(from_asset, ZERO_PRICE)) != ZERO_PRICE:
+                Inquirer.set_cached_price(
+                    cache_key=(from_asset, to_asset),
+                    cached_price=CachedPriceEntry(
+                        price=price,
+                        time=ts_now(),
```
LefterisJP, line 488: ts_now should get out of the for loop

rotkehlchen/inquirer.py
```diff
@@ -593,283 +543,344 @@ def _query_oracle_instances(
...
+    def _find_prices(
+            from_assets: list[Asset],
+            to_asset: Asset,
+            ignore_cache: bool = False,
+            skip_onchain: bool = False,
+    ) -> dict[Asset, tuple[Price, CurrentPriceOracle]]:
+        """Returns a dict mapping from_assets to tuples containing the current price of the
+        from_asset in to_asset valuation and the oracle that was used to get that price.
+
+        NB: prices for special symbols in any currency but USD are not supported.
```
LefterisJP, line 729: what does this comment want to say? not sure if it was there before but reading it I don't get it

rotkehlchen/inquirer.py
```diff
@@ -1313,6 +1324,32 @@ def query_historical_fiat_exchange_rates(
...
+        """If to_asset is fiat, query the current price for any fiat assets in from_assets.
+        Returns a tuple containing a list of non fiat assets and a dict of prices found.
+        """
+        if not to_asset.is_fiat():
+            return from_assets, {}
+
+        non_fiat_assets, found_prices = [], {}
+        for from_asset in from_assets:
+            if from_asset.is_fiat():
+                with suppress(RemoteError):
```
LefterisJP, line 1341: is `Inquirer._query_fiat_pair` already logging something? If so then supressing is fine, but I would add a comment about it.

Otherwise we would need to log the failure somewhere.

rotkehlchen/inquirer.py
nicholasyoder: Yeah, this was here before: https://github.com/rotki/rotki/blob/develop/rotkehlchen/inquirer.py#L629

According to claude `"NB" stands for "Nota Bene," which is Latin for "note well" or "take special note."`. I've reworded it to hopefully make a little more sense.

rotkehlchen/inquirer.py
nicholasyoder: It doesn't seem that `_query_fiat_pair` was logging anything. I've reworked it so it does log something and returns None on failure instead of raising a remote error that must be suppressed like this.

LefterisJP (approved):
I will merge this for now, but as i mentioned in the standup I don't feel very comfortable with the change in complexity of the code. We should simplify and try to keep it simple.

Also these all needs extensive testing

PR #9303, opened 2025-01-24
Colibri icon collection main and custom icons path

LefterisJP:
Also purge unused code from python icons.py

Unfortunately still a lot used here.

- All the image/cache functionality also used by ENS icons.
- Retrieve a custom icon functionality missing from colibri
- Delete icon functionality missing from colibri


rotkehlchen/tests/external_apis/test_coingecko.py
```diff
@@ -75,74 +70,6 @@ def test_coingecko_historical_price(session_coingecko):
     assert price == Price(FVal('2065.603754353392'))
 
 
-@pytest.mark.parametrize('use_clean_caching_directory', [True])
-def test_asset_icons_for_collections(icon_manager: IconManager) -> None:
```
LefterisJP, line 79: we need to be able to move relevant tests to rust. As this is no longer testable in python

yabirgb (approved):
lgtm

PR #9304, opened 2025-01-24
Splits docker build to multi-stage to use the native arm runner and updates the dockerfile

kelsos:
Closes #(issue_number)

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # build stage
-FROM --platform=$BUILDPLATFORM node:20-buster AS frontend-build-stage
+FROM --platform=$BUILDPLATFORM node:20-bookworm AS frontend-build-stage
 
 ARG BUILDARCH
 ENV COREPACK_ENABLE_STRICT=0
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -8,11 +8,21 @@ ENV NODE_OPTIONS="--max-old-space-size=4096"
...
+RUN if [ "$BUILDARCH" != "amd64" ]; then \
+      apt-get update && \
+      apt-get install -y build-essential python3 --no-install-recommends; \
+    fi && \
+    npm install -g pnpm@9 && \
+    pnpm install --frozen-lockfile && \
+    pnpm run docker:build
 
-FROM python:3.11-buster AS backend-build-stage
+FROM rust:1-bookworm AS colibri-build-stage
```
yabirgb, line 19: what is bookworm?

Dockerfile
```diff
@@ -62,12 +67,13 @@ RUN apt-get update && \
...
 
 COPY --from=backend-build-stage /tmp/dist /opt/rotki
+COPY --from=colibri-build-stage /tmp/dist/colibri/release/colibri /opt/rotki
 COPY --from=frontend-build-stage /app/app/dist /opt/rotki/frontend
 
 RUN APP=$(find "/opt/rotki" -name "rotki-core-*-linux"  | head -n 1) && \
     echo "${APP}" && \
     ln -s "${APP}" /usr/sbin/rotki && \
-    ln -s /opt/rotki/colibri/release/colibri /usr/sbin/colibri
+    ln -s /opt/rotki/colibri /usr/sbin/colibri
```
yabirgb, line 76: this is what was copying more stuff?

yabirgb (approved):
lgtm

Dockerfile
kelsos: It's my favorite light novel https://en.wikipedia.org/wiki/Ascendance_of_a_Bookworm !!!, no, just kidding :stuck_out_tongue: it's not this bookworm.

So **buster** is **Debian 10**: 

> *(Security updates have been discontinued as of June 30th, 2022)*
- https://www.debian.org/releases/buster/

**bookworm** is **Debian 12**:

> *(full Debian support, until June 10th, 2026, and two years of 
Long Term Support (LTS), until June 30th, 2028)*

- https://www.debian.org/releases/bookworm/

In the docker image context it just means that the docker image uses the specific **Debian** version as it's base. And I wanted to update the build stages because they are outdated.

The final image we ship was safe since **nginx:1.26** is already based on **bookworm**

Dockerfile
kelsos: This part is just linking the binary to `/usr/sbin` the problem was with the copy command that was copying the whole colibri release directory from the build stage to the final image stage.

PR #9305, opened 2025-01-24
Fix issue merging assets

yabirgb:
When two assets were selected to be merged we weren't merging their timed balances. This triggers a constraint error (in addition to a bug) since we need to sum both the entries for both assets if they are in the same timestamp and share the same category.

Closes #(issue_number)

## Checklist


LefterisJP (approved):
LGTM

PR #9306, opened 2025-01-24
Fix erc721 identifier db upgrade

nicholasyoder:
Fixes issue with missing asset errors after the upgrade. Removes all old erc721 tokens (have no collectible id), and also removes any invalid erc20 assets with the same address as an erc721.

### DIFF
--- a/rotkehlchen/db/upgrades/v46_v47.py
+++ b/rotkehlchen/db/upgrades/v46_v47.py
@@ -32,35 +32,84 @@ def upgrade_v46_to_v47(db: 'DBHandler', progress_handler: 'DBUpgradeProgressHand
     def _clean_cache(write_cursor: 'DBCursor') -> None:
         write_cursor.execute("DELETE FROM key_value_cache WHERE name LIKE 'extrainternaltx_%'")
 
-    @progress_step(description='Updating ERC721 token identifiers.')
-    def _update_erc721_token_identifiers(write_cursor: 'DBCursor') -> None:
-        """Update erc721 token identifiers to include the token id
-        for all evm events that have the token id in extra_data.
+    @progress_step(description='Remove unneeded nft collection assets (may take some time).')
+    def _remove_nft_collection_assets(write_cursor: 'DBCursor') -> None:
+        """Remove erc721 assets that have no collectible id, and also any erc20 assets
+        with the same address that were incorrectly added.
         """
-        write_cursor = write_cursor.execute("SELECT identifier, asset, extra_data FROM history_events WHERE entry_type = 2 AND extra_data != ''")  # noqa: E501
-        assets_updates = []
-        extra_data_updates = []
-        for entry in write_cursor:
-            asset_identifier = entry[1]
-            try:
-                extra_data: dict = json.loads(entry[2])
-            except json.JSONDecodeError as e:
-                log.error(f'Failed to parse {entry[2]} as JSON due to {e}')
-                continue
+        with (globaldb_conn := GlobalDBHandler().conn).read_ctx() as global_db_cursor:
+            identifiers_to_remove = tuple(entry[0] for entry in global_db_cursor.execute(
+                'SELECT identifier FROM assets WHERE identifier IN ('
+                "SELECT identifier FROM evm_tokens WHERE token_kind = 'B' "
+                "UNION SELECT REPLACE(identifier, 'erc721', 'erc20') "
+                "FROM evm_tokens WHERE token_kind = 'B');",
+            ))
 
-            if (token_id := extra_data.get('token_id')) is not None:
-                assets_updates.append((f'{asset_identifier}/{token_id}', asset_identifier))
-                # remove the token id and name from extra_data to avoid storing redundant data,
-                extra_data.pop('token_id', '')
-                extra_data.pop('token_name', '')
-                extra_data_updates.append((json.dumps(extra_data), entry[0]))
+        if (num_to_remove := len(identifiers_to_remove)) == 0:
+            return
 
-        if len(assets_updates) > 0:
-            assets_query = 'UPDATE assets SET identifier=? WHERE identifier=?'
-            with GlobalDBHandler().conn.write_ctx() as global_db_write_cursor:
-                global_db_write_cursor.executemany(assets_query, assets_updates)
+        log.debug(f'Deleting {num_to_remove} assets...')
+        placeholders = ','.join(['?'] * len(identifiers_to_remove))
+        with globaldb_conn.write_ctx() as global_db_write_cursor:
+            log.debug('Deleting from globaldb...')
+            global_db_write_cursor.execute(f'DELETE FROM assets WHERE identifier IN ({placeholders})', identifiers_to_remove)  # noqa: E501
 
-            write_cursor.executemany(assets_query, assets_updates)
-            write_cursor.executemany('UPDATE history_events SET extra_data=? WHERE identifier=?', extra_data_updates)  # noqa: E501
+        log.debug('Deleting from user db...')
+        # Load any data associated with these identifiers that might actually be valuable and
+        # save it in a temporary table that will need to be dealt with manually.
+        # For 99% of users no data should be found here.
+        selected_data = {}
+        tables_to_delete_from = []
+        values_placeholders = ','.join(['(?)'] * len(identifiers_to_remove))
+        for table, columns in (
+            ('history_events', ('asset',)),
+            ('manually_tracked_balances', ('asset',)),
+            ('trades', ('base_asset', 'quote_asset', 'fee_currency')),
+            ('margin_positions', ('pl_currency', 'fee_currency')),
+            ('zksynclite_transactions', ('asset',)),
+            ('zksynclite_swaps', ('from_asset', 'to_asset')),
+            ('nfts', ('identifier', 'last_price_asset')),
+        ):
+            column_str = ','.join([f'{table}.{column}' for column in columns])
+            result = write_cursor.execute(
+                f'WITH asset_list(value) AS (VALUES {values_placeholders}) '
+                f'SELECT * FROM {table} '
+                f'WHERE EXISTS (SELECT 1 FROM asset_list WHERE value IN ({column_str}))',
+                identifiers_to_remove,
+            ).fetchall()
+            if len(result) > 0:
+                selected_data[table] = result
+                tables_to_delete_from.append((table, columns))
+
+        if len(selected_data) > 0:
+            log.error('Found data associated with old erc721 tokens. Saving it in the temp_erc721_data table.')  # noqa: E501
+            write_cursor.execute('CREATE TABLE temp_erc721_data (table_name TEXT, data TEXT)')
+            for table, data in selected_data.items():
+                try:
+                    write_cursor.execute(
+                        'INSERT INTO temp_erc721_data(table_name, data) VALUES(?, ?)',
+                        (table, json.dumps(data)),
+                    )
+                except (TypeError, OverflowError, UnicodeEncodeError) as e:
+                    log.error(f'Failed to serialize {data!s} as json due to {e!s}')
+
+        # Delete the old erc721 assets and any associated data.
+        # The data in timed_balances and evm_accounts_details will likely be present and
+        # has no value so it is not included in the select queries above.
+        for table, columns in tuple(tables_to_delete_from) + (
+            ('timed_balances', ('currency',)),
+            ('evm_accounts_details', ('value',)),
+            ('assets', ('identifier',)),
+        ):
+            column_str = ','.join([f'{table}.{column}' for column in columns])
+            write_cursor.execute(
+                f'WITH asset_list(value) AS (VALUES {values_placeholders}) '
+                f'DELETE FROM {table} '
+                f'WHERE EXISTS (SELECT 1 FROM asset_list WHERE value IN ({column_str}))',
+                identifiers_to_remove,
+            )
+
+    # TODO: Put the history events reset step BEFORE _remove_nft_collection_assets so that
+    # most history events are gone when it checks for data associated with old erc721 assets.
 
     perform_userdb_upgrade_steps(db=db, progress_handler=progress_handler, should_vacuum=True)
--- a/rotkehlchen/rotkehlchen.py
+++ b/rotkehlchen/rotkehlchen.py
@@ -518,6 +518,16 @@ class Rotkehlchen:
         self.user_is_logged_in = True
         log.debug('User unlocking complete')
 
+        # Send a notification to the user if data associated with
+        # old erc721 tokens has been saved during the db upgrade.
+        # TODO: Remove this after a couple versions (added in version 1.38).
+        with self.data.db.conn.read_ctx() as cursor:
+            if cursor.execute("SELECT COUNT(*) FROM sqlite_master WHERE type='table' AND name='temp_erc721_data'").fetchone()[0] != 0:  # noqa: E501
+                self.msg_aggregator.add_warning(
+                    'Data associated with invalid ERC721 assets is present in your database. '
+                    'Please contact rotki support via our discord to resolve this issue.',
+                )
+
     def _logout(self) -> None:
         if not self.user_is_logged_in:
             return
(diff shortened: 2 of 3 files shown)


rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -32,35 +31,33 @@ def _add_new_locations(write_cursor: 'DBCursor') -> None:
...
-        if len(assets_updates) > 0:
-            assets_query = 'UPDATE assets SET identifier=? WHERE identifier=?'
-            with GlobalDBHandler().conn.write_ctx() as global_db_write_cursor:
-                global_db_write_cursor.executemany(assets_query, assets_updates)
-
-            write_cursor.executemany(assets_query, assets_updates)
-            write_cursor.executemany('UPDATE history_events SET extra_data=? WHERE identifier=?', extra_data_updates)  # noqa: E501
+        identifiers_to_remove = []
+        with GlobalDBHandler().conn.read_ctx() as global_db_cursor:
+            cursor = global_db_cursor.execute("SELECT identifier FROM assets WHERE identifier LIKE '%erc721%'")  # noqa: E501
```
yabirgb, line 41: I believe this is slow because you check all the assets. You can join with the evm tokens table and fitler only then. Even better... The token_kind column (I don't recall if that is the name) is not accurate? You can obtain the identifiers of the column where the kind is erc721

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -32,35 +31,33 @@ def _add_new_locations(write_cursor: 'DBCursor') -> None:
...
+            for entry in cursor:
+                identifiers_to_remove.extend([
+                    entry[0],
+                    entry[0].replace('erc721', 'erc20'),
```
yabirgb, line 45: I wonder if this would be faster if you use a regex instead of replace. Not sure how replace works inside but I can imagine it kind of slow because it searches in all the string

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -32,35 +31,33 @@ def _add_new_locations(write_cursor: 'DBCursor') -> None:
...
+                ])
+
+        if len(identifiers_to_remove) == 0:
+            return
+
+        placeholders = ','.join(['?'] * len(identifiers_to_remove))
+        with GlobalDBHandler().conn.write_ctx() as global_db_write_cursor:
```
yabirgb, line 52: just a nitpick but you can store the conn using the walrus operator and avoid invoking the new method on the globaldb

rotkehlchen/db/upgrades/v46_v47.py
nicholasyoder: Actually I can simply query the evm_tokens table by itself without doing any join.

rotkehlchen/db/upgrades/v46_v47.py
nicholasyoder: simply using replace as discussed in discord.

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -32,35 +31,35 @@ def _add_new_locations(write_cursor: 'DBCursor') -> None:
...
-            except json.JSONDecodeError as e:
-                log.error(f'Failed to parse {entry[2]} as JSON due to {e}')
-                continue
+        with (globaldb_conn := GlobalDBHandler().conn).read_ctx() as global_db_cursor:
+            identifiers_to_remove = [entry[0] for entry in global_db_cursor.execute(
+                'SELECT identifier FROM assets WHERE identifier IN ('
+                "SELECT identifier FROM evm_tokens WHERE token_kind = 'B' "
+                "UNION SELECT REPLACE(identifier, 'erc721', 'erc20') "
+                "FROM evm_tokens WHERE token_kind = 'B');",
+            ).fetchall()]
```
yabirgb, line 45: You don't need to call fetchall since when you consume the cursor it does it internally. If you do fetchall like you did you are reading twice the data. Also if you don't modify it can you make it a tuple?

```suggestion
            identifiers_to_remove = tuple(entry[0] for entry in global_db_cursor.execute(
                'SELECT identifier FROM assets WHERE identifier IN ('
                "SELECT identifier FROM evm_tokens WHERE token_kind = 'B' "
                "UNION SELECT REPLACE(identifier, 'erc721', 'erc20') "
                "FROM evm_tokens WHERE token_kind = 'B');",
            ))
```

nicholasyoder: yep, tuple works

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -32,35 +31,35 @@ def _add_new_locations(write_cursor: 'DBCursor') -> None:
...
-            write_cursor.executemany(assets_query, assets_updates)
-            write_cursor.executemany('UPDATE history_events SET extra_data=? WHERE identifier=?', extra_data_updates)  # noqa: E501
+        log.debug('Deleting from user db...')
+        # Delete from other tables where they might be referenced first to avoid FOREIGN KEY constraint errors when deleting from assets.  # noqa: E501
+        write_cursor.execute(f'DELETE FROM history_events WHERE asset IN ({placeholders})', identifiers_to_remove)  # noqa: E501
+        write_cursor.execute(f'DELETE FROM timed_balances WHERE currency IN ({placeholders})', identifiers_to_remove)  # noqa: E501
+        write_cursor.execute(f'DELETE FROM evm_accounts_details WHERE value IN ({placeholders})', identifiers_to_remove)  # noqa: E501
+        write_cursor.execute(f'DELETE FROM assets WHERE identifier IN ({placeholders})', identifiers_to_remove)  # noqa: E501
+
+    # TODO: Reset events to trigger re-decoding so erc721 assets are added properly.
```
LefterisJP, line 63: No need for this todo right? We always add the events redecoding during a DB upgrade. It's more probable you will forget to remove this

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -32,35 +31,35 @@ def _add_new_locations(write_cursor: 'DBCursor') -> None:
...
+        placeholders = ','.join(['?'] * len(identifiers_to_remove))
+        with globaldb_conn.write_ctx() as global_db_write_cursor:
+            log.debug('Deleting from globaldb...')
+            global_db_write_cursor.execute(f'DELETE FROM assets WHERE identifier IN ({placeholders})', identifiers_to_remove)  # noqa: E501
 
-            write_cursor.executemany(assets_query, assets_updates)
-            write_cursor.executemany('UPDATE history_events SET extra_data=? WHERE identifier=?', extra_data_updates)  # noqa: E501
+        log.debug('Deleting from user db...')
+        # Delete from other tables where they might be referenced first to avoid FOREIGN KEY constraint errors when deleting from assets.  # noqa: E501
+        write_cursor.execute(f'DELETE FROM history_events WHERE asset IN ({placeholders})', identifiers_to_remove)  # noqa: E501
```
LefterisJP, line 58: if you want all tables with a FK you need to also add manually_tracked_balances here, trades, margin_positions, zksynclite_transactions, zksynclite_swaps, nfts

But do you want to delete the entries? Or you want only specific ones to be deleted and the others be cascaded with an update to None?

If you only want to clean specific tables the comment needs to be updated.

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -32,35 +31,35 @@ def _add_new_locations(write_cursor: 'DBCursor') -> None:
...
+        log.debug(f'Deleting {num_to_remove} assets...')
+        placeholders = ','.join(['?'] * len(identifiers_to_remove))
+        with globaldb_conn.write_ctx() as global_db_write_cursor:
+            log.debug('Deleting from globaldb...')
+            global_db_write_cursor.execute(f'DELETE FROM assets WHERE identifier IN ({placeholders})', identifiers_to_remove)  # noqa: E501
 
-            write_cursor.executemany(assets_query, assets_updates)
-            write_cursor.executemany('UPDATE history_events SET extra_data=? WHERE identifier=?', extra_data_updates)  # noqa: E501
+        log.debug('Deleting from user db...')
+        # Delete from other tables where they might be referenced first to avoid FOREIGN KEY constraint errors when deleting from assets.  # noqa: E501
```
LefterisJP, line 57: Why do this exactly? Don't you rely on cascading? Or you are doing it only on tables where no FK is there? If so the only one correct is `evm_accounts_details`

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -32,35 +31,35 @@ def _add_new_locations(write_cursor: 'DBCursor') -> None:
     def _clean_cache(write_cursor: 'DBCursor') -> None:
         write_cursor.execute("DELETE FROM key_value_cache WHERE name LIKE 'extrainternaltx_%'")
 
-    @progress_step(description='Updating ERC721 token identifiers.')
-    def _update_erc721_token_identifiers(write_cursor: 'DBCursor') -> None:
-        """Update erc721 token identifiers to include the token id
-        for all evm events that have the token id in extra_data.
+    @progress_step(description='Remove unneeded nft collection assets (may take some time).')
```
LefterisJP, line 34: Is it possible to rewrite this so the steps are the aount of tokens deleted? So it counts slowly towards the end and shows a progress of deletion?

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2744,13 +2751,23 @@ def test_upgrade_db_46_to_47(user_data_dir, messages_aggregator):
...
+            (HistoryBaseEntryType.EVM_EVENT.value, 'TEST1', 0, 1, Location.ETHEREUM.serialize_for_db(), (user_address := '0x706A70067BE19BdadBea3600Db0626859Ff25D74'), uniswap_erc721_token.identifier, 1, 0, '', HistoryEventType.DEPLOY.value, HistoryEventSubType.NFT.value, ''),  # noqa: E501
+        )
+        write_cursor.executemany(
+            'INSERT INTO evm_accounts_details(account, chain_id, key, value) VALUES(?, ?, ?, ?)',
+            [
+                (user_address, 1, 'tokens', uniswap_erc721_token.identifier),
+                (user_address, 1, 'testkey', 'testdata'),
+            ],
+        )
+        write_cursor.execute(
```
LefterisJP, line 2768: can you also assert on their existence here? Not just adding?

rotkehlchen/db/upgrades/v46_v47.py
nicholasyoder: Removed it. Didn't realize this was something we always did in upgrades.

rotkehlchen/db/upgrades/v46_v47.py
nicholasyoder: >  the others be cascaded with an update to None?

Unfortunately this isn't how it works. When it doesn't have `ON DELETE CASCADE` it does a `FOREIGN KEY` error if you try to delete from assets when its still being referenced in another table.

I was figuring that erc721 tokens would only ever be in the tables I had it deleting from here, but I guess that isn't a very safe assumption. I've added all the tables that reference assets.identifier and updated the comment to hopefully be a little clearer.

rotkehlchen/db/upgrades/v46_v47.py
nicholasyoder: See the answer to the other comment.

rotkehlchen/db/upgrades/v46_v47.py
nicholasyoder: Skipping this as discussed in discord to avoid doing more queries and also since when we add idices it should speed this up majorly.

rotkehlchen/tests/db/test_db_upgrades.py
nicholasyoder: put the assertions in a function with an `expect_removed` parameter and call it both before and after the upgrade to avoid repeating code.

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -32,35 +32,75 @@ def _add_new_locations(write_cursor: 'DBCursor') -> None:
...
+            if len(result) > 0:
+                selected_data[table] = result
+                tables_to_delete_from.append((table, columns))
+
+        if len(selected_data) > 0:
+            log.error('Found data associated with old erc721 tokens. Saving it in the temp_erc721_data table.')  # noqa: E501
+            write_cursor.execute('CREATE TABLE temp_erc721_data (table_name TEXT, data TEXT)')
+            write_cursor.executemany(
+                'INSERT INTO temp_erc721_data(table_name, data) VALUES(?, ?)',
+                [(table, json.dumps(data)) for table, data in selected_data.items()],
```
LefterisJP, line 86: why json dumps here? Will this work for all tables in question? Is there any value in any of these tables that is not serializable in json? In any case if there is you would need to handle the json serialization error and not let the DB upgrade die.

Which is why I would not do an executemany here. I would instead do it  table by table and log an error and skip if we got any error here in serialization.

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -32,35 +32,75 @@ def _add_new_locations(write_cursor: 'DBCursor') -> None:
...
+        # The data in timed_balances and evm_accounts_details will likely be present and
+        # has no value so it is not included in the select queries above.
+        for table, columns in tuple(tables_to_delete_from) + (
+            ('timed_balances', ('currency',)),
+            ('evm_accounts_details', ('value',)),
+            ('assets', ('identifier',)),
+        ):
+            where_clauses = ' OR '.join([f'{column} IN ({placeholders})' for column in columns])
+            write_cursor.execute(
+                f'DELETE FROM {table} WHERE {where_clauses}',
```
LefterisJP, line 99: probably same comment on the multiplication here as above

rotkehlchen/rotkehlchen.py
```diff
@@ -518,6 +518,16 @@ def unlock_user(
         self.user_is_logged_in = True
         log.debug('User unlocking complete')
 
+        # Send a notification to the user if data associated with
+        # old erc721 tokens has been saved during the db upgrade.
+        # TODO: Remove this after a couple versions (added in version 1.38).
+        with self.data.db.conn.read_ctx() as cursor:
+            if cursor.execute("SELECT COUNT(*) FROM sqlite_master WHERE type='table' AND name='temp_erc721_data'").fetchone()[0] != 0:  # noqa: E501
+                self.msg_aggregator.add_error(
```
LefterisJP, line 526:
```suggestion
                self.msg_aggregator.add_warning(
```

rotkehlchen/rotkehlchen.py
```diff
@@ -518,6 +518,16 @@ def unlock_user(
...
+                    'Data associated with invalid ERC721 assets is present in your database. '
+                    'Please contact support to resolve this issue.',
```
LefterisJP, line 528:
```suggestion
                    'Please contact rotki support via our discord to resolve this issue.',
```

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2718,6 +2722,35 @@ def test_upgrade_db_45_to_46(user_data_dir: 'Path', messages_aggregator):
...
+            ).fetchone()[0] == 0 if expect_removed else count
+
+        with GlobalDBHandler().conn.read_ctx() as global_db_cursor:
+            assert global_db_cursor.execute(
+                'SELECT COUNT(*) FROM assets WHERE identifier IN (?, ?)',
+                (uniswap_erc20_token.identifier, uniswap_erc721_token.identifier),
+            ).fetchone()[0] == 0 if expect_removed else 2
+
+        if expect_removed:
+            # TODO: the history_events entry will need removed from this assertion
```
LefterisJP, line 2747: mark it as custom. Then the test will be even more accurate and you won't need to fix this test after resetting of events.

rotkehlchen/db/upgrades/v46_v47.py
nicholasyoder: I think all the data here should be json serializable. I've changed it to not use `executemany` though as suggested, with a try except to log any errors.

LefterisJP (approved):
LGTM with a nitpick that can be done in another PR

PR #9312, opened 2025-01-26
Bump pytest related packaged and vcrpy

LefterisJP:


LefterisJP:
Trying to figure out why upgrading to pytest 8.3.4 makes `rotkehlchen/tests/integration/test_backend.py::test_backend` fails and does so only in the CI.

LefterisJP:
Thank you @prettyirrelevant

PR #9317, opened 2025-01-27
feat(frontend): setting to show net worth in tray for Mac

lukicenturi:
image

Only possible in mac, so I disabled it for other OSes


frontend/app/electron/main/tray-manager.ts
```diff
@@ -41,9 +42,23 @@ export class TrayManager {
...
+
+          this.update();
+        },
       },
+      ...(this.config.isMac
+        ? [{
+            label: visible ? 'Minimize to tray' : 'Restore from tray',
+            click: () => this.toggleWindowVisibility(),
+          }]
+        : []),
```
kelsos, line 61: why limit this only on mac? was it originally only on mac?

LefterisJP: The tray icon also works on linux and windows. Please don't remove it.

lukicenturi: LOL my bad, I placed the condition in the wrong spot. 
I wanted to set the condition for the new setting "Display net worth on the tray" to apply only on Mac and Linux, as Windows won't be showing it at all.

lukicenturi: Updated now @kelsos

frontend/app/public/rotki-trayTemplate.png
kelsos, line 1: I am assuming this is fixing the icon in dark mode right?
Can you check what is needed for both dark mode and dark mode at the same time?

White does not look that good in light mode.

lukicenturi: can you screenshot it how it looks like?
rotki-trayTemplate supposed to be used on mac only.
And from what I observe in mac, in dark mode or light mode, all the tray icons are always white

lukicenturi: Perhaps I should use https://developer.apple.com/documentation/appkit/nsimage/istemplate instead to make it adapt automatically

frontend/app/electron/main/tray-manager.ts
```diff
@@ -52,10 +67,18 @@ export class TrayManager {
...
       return;
 
+    const trayData = tray ?? this.trayData;
+
+    if (!trayData)
+      return;
+
+    const { currency, delta, percentage, up, period, netWorth } = trayData;
+    this.trayData = trayData;
+
```
kelsos, line 81: So I noticed an interesting bug on windows in a new account. It should be probably easy to replicate in a new account on macOS too.

`up` is undefined because it's a new account but networth is not empty. 

Do you think we can do a small change here? if `netWorth` is defined/non-empty then show `Net worth ${netWorth} ${currency}.` Otherwise show the `rotki is running` text

kelsos: This is for line 85 below

frontend/app/electron/main/tray-manager.ts
```diff
@@ -40,6 +41,20 @@ export class TrayManager {
           ]
         : []),
       { type: 'separator' },
+      ...(this.config.isMac || this.config.isLinux
```
kelsos, line 44: So the `isLinux` is redundant here.

For the feature we update the [title](https://www.electronjs.org/docs/latest/api/tray#traysettitletitle-options-macos) which works only works for macOS.

frontend/app/public/rotki-trayTemplate.png
kelsos: this is the before for me image and this is the after 
image

If you can use the native image to make it adaptable that would be perfect

lukicenturi:
@kelsos updated can you check again

kelsos (approved):
lgtm

PR #9322, opened 2025-01-28
Add endpoint for fetching historical prices per asset in profit currency

prettyirrelevant:


rotkehlchen/api/rest.py
```diff
@@ -5431,3 +5431,35 @@ def get_historical_netvalue(
...
+            except NoPriceForGivenTimestamp:
+                no_prices_ts.append(current_ts)
+            else:
+                prices[current_ts] = str(price)
+
+            current_ts += interval  # type: ignore[assignment]  # they're both integers
+
+        return _wrap_in_ok_result(result={
+            'prices': prices,
+            'no_prices_timestamps': no_prices_ts,
```
LefterisJP, line 5464: I don't say it's wrong but why choose this approach? Did you talk with the frontend? They prefer this to a single dict with `None` values if missing?

rotkehlchen/api/v1/schemas.py
```diff
@@ -3861,3 +3861,32 @@ def make_calendar_entry(self, data: dict[str, Any], **_kwargs: dict[str, Any]) -
...
+            data: dict[str, Any],
+            **_kwargs: Any,
+    ) -> None:
+        if (to_ts := data['to_timestamp']) <= (from_ts := data['from_timestamp']):
+            raise ValidationError(
+                message='from_timestamp must be smaller than to_timestamp',
+                field_name='from_timestamp',
+            )
+
+        if from_ts % (interval := data['interval']) != 0:
```
LefterisJP, line 3882: why the multiples? Can't you just get to the closest one and return that? Both from and for to?

rotkehlchen/api/v1/schemas.py
```diff
@@ -3861,3 +3861,32 @@ def make_calendar_entry(self, data: dict[str, Any], **_kwargs: dict[str, Any]) -
 
 class HistoricalPerAssetBalanceSchema(SnapshotTimestampQuerySchema, AsyncQueryArgumentSchema):
     asset = AssetField(expected_type=Asset, load_default=None)
+
+
+class HistoricalPricesPerAssetSchema(AsyncQueryArgumentSchema, TimestampRangeSchema):
+    interval = fields.Integer(strict=True, required=True)
```
LefterisJP, line 3867: This should be a positive Integer. Add a validate keyword

```
validate=webargs.validate.Range(
                min=0,
                error='Identifier must be an integer >= 0',
            )
```

rotkehlchen/api/rest.py
```diff
@@ -5431,3 +5432,49 @@ def get_historical_netvalue(
...
+            from_timestamp: Timestamp,
+    ) -> dict[str, Any]:
+        prices, no_prices_ts = {}, []
+        with (db := self.rotkehlchen.data.db).conn.read_ctx() as cursor:
+            main_currency = db.get_setting(cursor=cursor, name='main_currency')
+
+        processed_count = 0
+        total_intervals = (to_timestamp - from_timestamp) // interval + 1
+        # Send status update every 5% progress
+        update_threshold = max(total_intervals // 20, 1)
```
LefterisJP, line 5451: no no no.

The updates should be done every `N` remote calls where N is configurable. Let's say every 5 or 10 calls?

LefterisJP (commented):
Shouldn't the websockets api also be adjusted? Other than that it's the frontend that will need to adjust now too right?

LefterisJP (approved):
merging it for speed and the websocket docs can be adjusted in another PR

PR #9323, opened 2025-01-28
Fix aerodrome/velodrome pool querying

nicholasyoder:
Its seems that at present there are exactly 2000 aerodrome pools, which triggers a bug in our logic where it tries to query a chunk of pools that doesn't exist, which resulted in a remote error. Now the remote error should be handled gracefully, and it will continue with adding the pools it has queried instead of exiting the logic with an error.

Also I adjusted where the pool querying status messages were being sent in this logic, to better reflect what is actually happening. Previously the messages indicating how many pools were processed were on a loop that finished in several seconds, while the other loop (where I've moved it to now) took a number of minutes to complete.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` Correctly handle a remote error in the Aerodrome/Velodrome pool querying logic.
 * :bug:`-` Some mainnet to optimism bridging transactions that were not seen correctly will now be properly decoded.
 * :bug:`9308` Fix some cases of Coinbase events that were not properly understood in rotki.
 * :bug:`-` Some specific cases of yearn v2 vault deposit/withdrawals that had problems will now be decoded properly.
--- a/rotkehlchen/chain/evm/decoding/velodrome/velodrome_cache.py
+++ b/rotkehlchen/chain/evm/decoding/velodrome/velodrome_cache.py
@@ -181,6 +181,22 @@ def query_velodrome_data_from_chain_and_maybe_create_tokens(
     pool_data_chunk: list[dict] = []
     offset, limit, last_notified_ts = 0, 200, Timestamp(0)
     while len(pool_data_chunk) == limit or (len(pool_data_chunk) == 0 and offset == 0):
+        try:
+            pool_data_chunk = data_contract.call(
+                node_inquirer=inquirer,
+                method_name='all',
+                arguments=[limit, offset],
+            )
+        except RemoteError as e:
+            log.warning(f'Failed to query velodrome pool data chunk due to {e!s}.')
+            # If the total count of existing pools is a multiple of 200 it will try to query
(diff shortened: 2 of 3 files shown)


rotkehlchen/chain/evm/decoding/velodrome/velodrome_cache.py
```diff
@@ -283,6 +283,14 @@ def query_velodrome_data_from_chain_and_maybe_create_tokens(
             fallback_symbol=pool_name,
         )
         returned_pools.append(entry)
+        last_notified_ts = maybe_notify_cache_query_status(
```
LefterisJP, line 286: Hmm seeing that you also had to change the test I think that this makes no sense here. If it's too fast it means it's also useless. Before from what I see it was inside the while loop and after each remote call. Here it's in another loop but that loop has no remote calls so you don't need progress here.

The progress is required if something takes a long time. And that something is up there, in the previous loop. So can you put that back? And in that case re-adjust the test

nicholasyoder: > from what I see it was inside the while loop and after each remote call.

No, no, that's where the total pools message is being sent. Notice that it uses `maybe_notify_new_pools_status` in the first loop to update how many total pools it needs to process, then later it uses `maybe_notify_cache_query_status` in another loop to notify of how many pools have been processed.

I've moved both of these:
* `maybe_notify_new_pools_status` I simply moved from the beginning to the end of the first loop so it doesn't skip notifying of the last chunk of pools that it finds.
* `maybe_notify_cache_query_status` I moved from the second loop in the function to the third loop in the function. The loop this was in previously simply does deserialization of the data it has already queried - it has no remote calls or database queries and completes almost instantly. The loop I've moved it to is where its actually adding the tokens (making db queries) and takes a little time to complete. Also, in between these two loops it calls `get_multiple_erc20_contract_info`, which also takes a little time. 

So what was happening before is that it would send messages updating the total number of pools, then the deserialization loop would run, and none of the messages updating the number of pools processed would even be sent since that loop finishes so fast. Then it would take a couple minutes to call `get_multiple_erc20_contract_info` and do the last loop, while still saying that it had processed 0 pools.

I think it makes sense to do it the way I've changed it. Perhaps you misread the code? Or am I misunderstanding something?

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` Correctly handle a remote error in the Aerodrome/Velodrome pool querying logic.
```
LefterisJP, line 5: Please don't make me repeat myself every time. Messages in the changelog should be user facing. What is a remote error, what is a pool querying logic? These are implementation details the user does not need to know.

```suggestion
* :bug:`-` A rare error with velodrome events decoding should now be fixed and the relevants events properly decoded.
```

LefterisJP (approved):
LGTM

PR #9326, opened 2025-01-29
fix: changes how usd value threshold works

kelsos:
Closes #(issue_number)

## Checklist

## Notes
- Adds a migration to change the schema for the threshold.
  - New schema treats `0` as a valid value and undefined as deactivated
- Adds validation to the component
- Adds hints to the component
  - To clarify what 0 or non zero value means
- Fixes a bug that would reset the value during the first load if the user used any currency different than `USD`


lukicenturi (approved):
LGTM! Thanks

PR #9328, opened 2025-01-29
Fix aerodrome lp token prices

nicholasyoder:
Apparently prices for aerodrome LP tokens was never implemented correctly, which prevented aerodrome balances from showing. Now the prices should be found correctly. Since aerodrome is very similar to velodrome, `AERODROME_POOL_PROTOCOL` just needed added several places to make it query the price in the same way that it does for tokens with `VELODROME_POOL_PROTOCOL`.

Also fixes the changelog nitpick from https://github.com/rotki/rotki/pull/9323

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,7 +2,8 @@
 Changelog
 =========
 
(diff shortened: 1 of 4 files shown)


LefterisJP (approved):
LGTM

PR #9330, opened 2025-01-30
fix: improve error message UI

### LINKED ISSUE #9321: Improve error message UI to show more of the error message
## Problem Definition

When Rotki has a problem, in my case a startup problem with a stuck process blocking the port, an error screen is shown. The error screen does not make good use of the screen space and could be improved.
E.g. the scrollable text area showing the error and log only covers 50% of the framed area vertically (see screenshot) and should use as much space as possible to convey the error. Also, the framed area has no margin on the bottom and could be centered, but styling is not the main issue.

### System Description

Operating system: MacOS Sonoma 14.6.1
Rotki version: 1.37.1

Image

lukicenturi:
Closes #9321 

image

### DIFF
--- a/frontend/app/src/components/error/ErrorScreen.vue
+++ b/frontend/app/src/components/error/ErrorScreen.vue
@@ -32,7 +32,7 @@ const errorText = computed(() => {
 </script>
 
 <template>
-  <div class="error-screen">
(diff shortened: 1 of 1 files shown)


kelsos (approved):
lgtm thank you @lukicenturi

PR #9335, opened 2025-01-30
Speed up db upgrade

yabirgb:
- Create indexes for faster look ups
- Disable foreign key and manually delete assets from tables where they could be
- deletes history events

### DIFF
--- a/rotkehlchen/db/upgrades/v46_v47.py
+++ b/rotkehlchen/db/upgrades/v46_v47.py
@@ -2,6 +2,7 @@ import json
 import logging
 from typing import TYPE_CHECKING
 
+from rotkehlchen.db.constants import HISTORY_MAPPING_KEY_STATE, HISTORY_MAPPING_STATE_CUSTOMIZED
 from rotkehlchen.globaldb.handler import GlobalDBHandler
 from rotkehlchen.logging import RotkehlchenLogsAdapter, enter_exit_debug_log
 from rotkehlchen.utils.progress import perform_userdb_upgrade_steps, progress_step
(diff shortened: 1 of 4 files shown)


rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -52,7 +80,24 @@ def _remove_nft_collection_assets(write_cursor: 'DBCursor') -> None:
...
+                ('evm_tokens', 'identifier'),
+                ('common_asset_details', 'identifier'),
+                ('multiasset_mappings', 'asset'),
+                ('price_history', 'from_asset'),
+                ('price_history', 'to_asset'),
+                ('underlying_tokens_list', 'identifier'),
+                ('underlying_tokens_list', 'parent_token_entry'),
+                ('user_owned_assets', 'asset_id'),
+            )
+            total = len(tables_with_assets)
```
LefterisJP, line 95: You can move this out of the cursor. Same with log.debug('Deleting from globaldb...')

```suggestion
```

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -52,7 +80,24 @@ def _remove_nft_collection_assets(write_cursor: 'DBCursor') -> None:
         placeholders = ','.join(['?'] * len(identifiers_to_remove))
         with globaldb_conn.write_ctx() as global_db_write_cursor:
             log.debug('Deleting from globaldb...')
-            global_db_write_cursor.execute(f'DELETE FROM assets WHERE identifier IN ({placeholders})', identifiers_to_remove)  # noqa: E501
+            global_db_write_cursor.executescript('PRAGMA foreign_keys = OFF;')
```
LefterisJP, line 83: Explain why you are doing it like this. Say that cascading deletion with Foreign keys ON took `X` mins for your DB. While this approach, essentially manually managing the cascading deletion took `X` mins for your DB.

LefterisJP (approved):
LGTM

PR #9344, opened 2025-01-31
feat: asset update advanced mode

lukicenturi:
Screenshot at Jan 31 20-11-46
Screenshot at Jan 31 20-11-40


kelsos (approved):
lgtm thanks @lukicenturi

PR #9338, opened 2025-01-30
Improve arbitrum weth decoder

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=95697078

PR #9347, opened 2025-01-31
Fix duplicate sequence index bug

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=95879616

### DIFF
--- a/rotkehlchen/chain/evm/decoding/base.py
+++ b/rotkehlchen/chain/evm/decoding/base.py
@@ -7,6 +7,7 @@ from rotkehlchen.assets.asset import CryptoAsset, EvmToken
 from rotkehlchen.assets.utils import get_or_create_evm_token, get_token
 from rotkehlchen.chain.evm.constants import ETH_SPECIAL_ADDRESS, ZERO_ADDRESS
 from rotkehlchen.chain.evm.decoding.constants import OUTGOING_EVENT_TYPES
+from rotkehlchen.chain.evm.structures import EvmTxReceipt
 from rotkehlchen.constants import ONE, ZERO
 from rotkehlchen.constants.resolver import tokenid_to_collectible_id
 from rotkehlchen.fval import FVal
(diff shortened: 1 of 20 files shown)


rotkehlchen/chain/evm/decoding/base.py
```diff
@@ -51,24 +52,46 @@ def __init__(
...
 
-    def get_next_sequence_counter(self) -> int:
-        """Returns current counter and also increases it.
-        Meant to be called for all transaction events that do not have a corresponding log index"""
+    def get_next_sequence_index_pre_decoding(self) -> int:
+        """Get a sequence index for a new event created prior to running the decoding rules.
+        Used for gas events, eth transfers, etc. Must never be used after `get_sequence_index`
+        or `get_next_sequence_index` has been used to prevent sequence index collisions.
+        Returns the current counter and increments it.
+        """
```
LefterisJP, line 71: >         Must never be used after `get_sequence_index`
>        or `get_next_sequence_index` has been used to prevent sequence index collisions.

You can actually check for this in develop time only. Will help us to make sure we are fine for a while. And then we can remove it after a few release cycles. This should not be in production binary as we do `-OO` thanks to pyinstaller's python optimize: https://github.com/rotki/rotki/blob/67c9c0f5166d74611bc6cba65cfeec37708c107b/package.py#L750

```suggestion
        """
        if __debug__:   # develop only test that sequence index was not called
            assert not self.get_sequence_index_called  # Perhaps remove after some time.
```

rotkehlchen/chain/evm/decoding/base.py
```diff
@@ -51,24 +52,46 @@ def __init__(
...
 
-    def reset_sequence_counter(self) -> None:
+    def reset_sequence_counter(self, tx_receipt: 'EvmTxReceipt') -> None:
+        """Reset the sequence index counter before decoding a transaction.
+        `sequence_offset` is set to one more than the highest tx log index, and is used in
+        `get_next_sequence_index` to add new events whose sequence index will not collide with
+        the sequence index of any events that have associated tx logs.
+        """
         self.sequence_counter = 0
+        self.sequence_offset = tx_receipt.logs[-1].log_index + 1 if len(tx_receipt.logs) else 0
```
LefterisJP, line 64:
```suggestion
        self.sequence_offset = tx_receipt.logs[-1].log_index + 1 if len(tx_receipt.logs) else 0
        if __debug__:
            self.get_sequence_index_called = False
```

rotkehlchen/chain/evm/decoding/base.py
```diff
@@ -51,24 +52,46 @@ def __init__(
...
-        """Get the value that should go for this event's sequence index
-
-        This function exists to calculate the index bases on the pre-calculated
-        sequence index and the event's log index"""
+        """Get the sequence index for an event associated with a specific tx log.
+        Used for token transfers, approvals, and misc events created by the protocol decoders.
+        Returns the current counter added to the log index, placing these events after those
+        created using `get_next_sequence_index_pre_decoding`.
+        """
         return self.sequence_counter + tx_log.log_index
```
LefterisJP, line 82:
```suggestion
        if __debug__:
            self.get_sequence_index_called = True

        return self.sequence_counter + tx_log.log_index
```

rotkehlchen/chain/evm/decoding/base.py
```diff
@@ -51,24 +52,46 @@ def __init__(
...
 
+    def get_next_sequence_index(self) -> int:
+        """Get a sequence index for a new event with no associated tx log.
+        Used during protocol decoding for things like informational or fee events that are not
+        directly associated with any specific tx log.
+        Returns the current counter added to the sequence offset, placing these events after
+        any events created using `get_sequence_index`.
+        """
+        value = self.sequence_counter
```
LefterisJP, line 91:
```suggestion
        if __debug__:
            self.get_sequence_index_called = True

        value = self.sequence_counter
```

rotkehlchen/chain/evm/decoding/base.py
```diff
@@ -257,7 +279,9 @@ def make_event_from_transaction(
             address: ChecksumEvmAddress | None = None,
             extra_data: dict[str, Any] | None = None,
     ) -> 'EvmEvent':
-        """Convenience function on top of make_event to use transaction and ReceiptLog"""
+        """Convenience function on top of make_event to use transaction and ReceiptLog
```
LefterisJP, line 282:
```suggestion
        """Convenience function on top of make_event to use the transaction and a given log.
```

rotkehlchen/tests/unit/decoders/test_aave_v3.py
```diff
@@ -1784,7 +1771,20 @@ def test_aave_v3_close_position_with_safe(arbitrum_one_inquirer, arbitrum_one_ac
             address=ZERO_ADDRESS,
         ), EvmEvent(
             tx_hash=tx_hash,
-            sequence_index=23,
+            sequence_index=25,
```
LefterisJP, line 1774: When https://github.com/orgs/rotki/projects/11?pane=issue&itemId=95697078 is fixed this won't come in between the wrap and unwrap of the WETH, right? Cause right now it's in-between them and breaks the logic. If so also mention that we know this is broken and link to same note

nicholasyoder: These WETH events are already fixed in bugfixes: https://github.com/rotki/rotki/pull/9338

But it doesn't seem that there is actually any wrap event here, only an unwrap and the corresponding ETH recieve event. This safe informational event is not between those, either as it is now or with the weth fix from bugfixes, so it should be fine.

I've updated the comments on the weth events to indicate that they have already been fixed in bugfixes. This test will need updated when bugfixes are merged to develop.

LefterisJP (approved):
LGTM

PR #9355, opened 2025-02-03
Fix repeated morpho reward distributors task

nicholasyoder:
Fixes a problem I noticed during testing that the morpho reward distributors update task runs over and over. It wasn't including the chain_id in the key parts when checking `should_update_protocol_cache`, so the task just kept getting rescheduled.


yabirgb (commented):
should be okey

LefterisJP (commented):
looks okay but can you include a small unit test for this scheduling? We should be able to test against this

PR #9356, opened 2025-02-03
Fix ENS token name and symbol

nicholasyoder:
Hardcodes the ENS erc721 token name and symbol where the token is created in the decoder, since the [ENS contract](https://etherscan.io/token/0x57f1887a8bf19b14fc0df6fd9b2acc9af147ea85) doesn't implement the functions for retrieving this automatically.


yabirgb (approved):
lgtm

LefterisJP (approved):
LGTM

PR #9362, opened 2025-02-04
Add head endpoint for checking assets status

yabirgb:
- configures the logging in the app.
- Add options to set the debug level and if we want to log to the terminal or file.
- adds new endpoint for icons with verb head that queries the icons if they are needed or returns if they exists/can't be found
- verb get on icons only returns 200 or 404
- add unit tests for icons
- add .gitignore

### DIFF
--- a/colibri/Cargo.toml
+++ b/colibri/Cargo.toml
@@ -24,3 +24,19 @@ clap = { version = "4.5.26", features = ["string"] }
 # for globalDB and other sqlite operations
 rusqlite = { version = "0.32.1", features = ["bundled", "bundled-sqlcipher-vendored-openssl"] }
 async-sqlite = {version = "0.4.0" }
+# for cors and network
+tower-http = {version = "0.6.2", features = ["cors", "trace"]}
+http = "1.0"
+# for detailed logs when requests happen. Tracing in maintained by the tokio team and it helps
+# to log events in multithread environments. It is used by axum (also by tokio) to signal when
+# a request happens and keeps information about it. We use tracing to extract those events and
+# log them.
+tracing = "0.1.37"
+tracing-subscriber = { version = "0.3.16", features = ["env-filter"]}
+# Used to keep log files that rotate when they reach certain size. Is not available in tracing
+# directly (they have a PR but not merged)
+file-rotate = "0.7.6"
+
+[dev-dependencies]
+mockito = "1.6.1"
+tokio = { version = "*", features = ["test-util"]}
--- a/colibri/src/api/icons.rs
+++ b/colibri/src/api/icons.rs
@@ -1,13 +1,31 @@
-use crate::api::AppState;
+use crate::api::{constants::QUERY_ICONS_TASK_PREFIX, AppState};
 use crate::icons;
 use axum::{extract::Query, extract::State, response::IntoResponse};
+use log::error;
+use reqwest::StatusCode;
 use serde::Deserialize;
-use std::sync::Arc;
+use std::{sync::Arc, time::SystemTime};
+use tokio::fs;
 
+const HOUR_IN_SECS: u64 = 60 * 60;
+const MAX_ICON_RECHECK_PERIOD: u64 = HOUR_IN_SECS * 12;
+
+/// Used when requesting an asset locally
 #[derive(Deserialize)]
 pub struct AssetIconRequest {
+    // id of the asset to be queried
     asset_id: String,
+    // hash used to inform the consumer if the file has changed locally or not
     match_header: Option<String>,
+}
+
+/// Used when checking the state of an icon locally
+#[derive(Deserialize)]
+pub struct AssetIconCheck {
+    // id of the asset to be queried
+    asset_id: String,
+    // true if we should delete the local file and pull it again
+    force_refresh: Option<bool>,
 }
 
 /// The handler for the get icon endpoint
@@ -19,12 +37,18 @@ pub async fn get_icon(
     State(state): State<Arc<AppState>>,
     Query(payload): Query<AssetIconRequest>,
 ) -> impl IntoResponse {
-    match icons::get_or_query_icon(
-        state.globaldb.clone(),
-        state.coingecko.clone(),
+    let path = icons::get_asset_path(
+        &payload.asset_id,
+        state.data_dir.as_path(),
+        state.globaldb.as_ref(),
+    )
+    .await;
+
+    match icons::get_icon(
         state.data_dir.clone(),
         &payload.asset_id,
         payload.match_header,
+        path,
     )
     .await
     {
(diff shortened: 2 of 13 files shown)


colibri/Cargo.toml
```diff
@@ -9,7 +9,7 @@ log = "0.4.22"
 simplelog = "0.12.2"
 # For the asynchronous Rest API
 reqwest = { version = "0.12.2", features = ["json"]}
-tokio = { version = "1.43.0", features = ["rt-multi-thread", "fs"]}
+tokio = { version = "1.43.0", features = ["rt-multi-thread", "fs", "test-util"]}
```
LefterisJP, line 12: what is test-util tokio feature for? We probably need some more comments here ot be understanding why we have stuff.

colibri/Cargo.toml
```diff
@@ -24,3 +24,13 @@ clap = { version = "4.5.26", features = ["string"] }
 # for globalDB and other sqlite operations
 rusqlite = { version = "0.32.1", features = ["bundled", "bundled-sqlcipher-vendored-openssl"] }
 async-sqlite = {version = "0.4.0" }
+# for cors and network
+tower-http = {version = "0.6.2", features = ["cors", "trace"]}
+http = "1.0"
+# for detailed logs when requests happen
+tracing = "0.1.37"
```
LefterisJP, line 31: so this is yet a different library? https://crates.io/crates/tracing

Is this close to what lighthouse uses? Is it a heavily used solution by many other projects? I am not familiar nor do I have time to research so I need to trust you here.

Just make sure to use something that is not heavy and does not affect performance, does everything we need and is used by many other projects. The last one is important, because if it's something small, niche and new there is a big chance of it just dissapearing down the line

colibri/src/api/icons.rs
```diff
@@ -33,3 +57,117 @@ pub async fn get_icon(
         (status, _, _) => status.into_response(),
     }
 }
+
+/// The handler for the HEAD icon endpoint
+///
+/// First check if the file exists locally. If the file is not empty it means
+/// that we have the image locally and we can server it. Otherwise if the file
```
LefterisJP, line 64:
```suggestion
/// that we have the image locally and we can serve it. Otherwise if the file
```

colibri/src/api/icons.rs
```diff
@@ -33,3 +57,117 @@ pub async fn get_icon(
...
+/// has 0 size it means that the last time it was queried the icon couldn't be
+/// found remotely and if it was more than 12 hours that we tried we retry. If we
```
LefterisJP, line 66:
```suggestion
/// found remotely. Additionally if that query was more than MAX_ICON_RECHECK_PERIOD hours ago we will retry. If on the other hand we
```

colibri/src/api/icons.rs
```diff
@@ -33,3 +57,117 @@ pub async fn get_icon(
...
+/// queried it less than MAX_ICON_RECHECK_PERIOD hours ago then we say that the
+/// icon doesn't exists.
```
LefterisJP, line 68:
```suggestion
/// queried it less than MAX_ICON_RECHECK_PERIOD hours ago then we treat it as if 
/// the icon doesn't exist.
```

colibri/src/api/icons.rs
```diff
@@ -33,3 +57,117 @@ pub async fn get_icon(
...
+                            }
+                            None | Some(false) => return StatusCode::OK.into_response(),
+                        }
+                    }
+
+                    // check when was the last time that the file got updated
+                    if let Ok(time) = metadata.modified() {
+                        if let Ok(duration) = SystemTime::now().duration_since(time) {
+                            if duration.as_secs() < MAX_ICON_RECHECK_PERIOD {
+                                // It was queried less than 12 hours ago so don't try again
```
LefterisJP, line 112:
```suggestion
                                // It was queried recently so don't try again
```

colibri/src/api/icons.rs
```diff
@@ -33,3 +57,117 @@ pub async fn get_icon(
...
+                    }
+                }
+                Err(e) => {
+                    error!("Failed to query icon for {} due to {}", payload.asset_id, e);
+                    StatusCode::NOT_FOUND.into_response()
+                }
+            }
+        }
+        None => {
+            // There is no reference locally to the file, query it. Ensure that if it is requested
```
LefterisJP, line 141:
```suggestion
            // There is no local reference to the file, query it. Ensure that if it is requested
```

colibri/src/api/icons.rs
```diff
@@ -33,3 +57,117 @@ pub async fn get_icon(
...
+    state: Arc<AppState>,
+    payload: AssetIconCheck,
+    path: std::path::PathBuf,
+) -> StatusCode {
+    let task_name = format!("{}_{}", QUERY_ICONS_TASK_PREFIX, payload.asset_id);
+    let mut tasks = state.active_tasks.lock().await;
+    if !tasks.insert(task_name.clone()) {
+        return StatusCode::ACCEPTED;
+    };
+    drop(tasks);
```
LefterisJP, line 162: this code here looks suspicious. Why get a mutable reference to the tasks, insert the new task and then put `tasks` out of scope with `drop`?

colibri/src/args.rs
```diff
@@ -73,6 +78,27 @@ pub fn parse_args() -> Args {
...
+            Arg::new("max-logfiles-num")
+                .long("max-logfiles-num")
+                .value_parser(clap::value_parser!(usize))
+                .default_value("5")
+                .help("Max number of log files to keep"),
+        )
+        .arg(
+            Arg::new("log-level")
+                .long("log-level")
+                .value_parser(clap::value_parser!(LevelFilter))
```
LefterisJP, line 98: What are the valid levels? The ones here? https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.LevelFilter.html

Will they appear in `--help`?

colibri/src/coingecko.rs
```diff
@@ -54,6 +58,50 @@ impl Coingecko {
             Err(e) => error!("Failed to query coingecko for {} due to {}", asset_id, e),
         }
 
+        debug!("Icon not found in coingecko for {}", asset_id);
         None
     }
 }
+
+#[cfg(test)]
```
LefterisJP, line 66: Hmmm I know this is possible in rust but it's ugly to have tests in the same file as implementation. Please use the separate tests directory approach.

- Create a tests directory at the same level as src. 
- Create test files there (e.g., tests/coingecko_test.rs)

And in that test you would use coingecko and test it

colibri/src/coingecko.rs
```diff
@@ -54,6 +58,50 @@ impl Coingecko {
...
+            server.url(),
+        );
+        let json = format!(
+            r#"{{"image": {{"small": "{}/coins/images/279/thumb/ethereum.png"}}}}"#,
+            server.url()
+        );
+
+        // mock successful query
+        server
+            .mock("GET", format!("/api/v3/coins/ethereum").as_str())
```
LefterisJP, line 90: shouldn't we have one for HEAD too?

colibri/src/globaldb.rs
```diff
@@ -63,3 +63,58 @@ impl GlobalDB {
         })
     }
 }
+
+/// macro that creates a copy of the globaldb in the rotkehlchen data folder
+/// and stores it in a temp folder
+#[cfg(test)]
+#[macro_export]
+macro_rules! create_globaldb {
```
LefterisJP, line 71: same concerns for the test location here. I guess we would have this in a utils or just globaldb file but under the tests directory

colibri/src/icons.rs
```diff
@@ -81,8 +84,8 @@ async fn query_token_icon_and_extension(
                     }
                 } else if let Ok(text) = response.text().await {
                     error!(
-                        "Got non success response status {} with text: {} ",
-                        status, text
+                        "Got non success response status when querying smolDapp for the icon of {} with format {}. {} with text: {} ",
```
LefterisJP, line 87:
```suggestion
                        "Got non success response status when querying SmolDapp for the icon of {} with format {}. {} with text: {} ",
```

colibri/src/icons.rs
```diff
@@ -109,50 +112,49 @@ fn url_encode_identifier(input: &str) -> String {
...
     None
 }
 
-/// Gets the given icon from the user's system if it's already
-/// downloaded or asks for it from the icon sources. Returns it if found
-pub async fn get_or_query_icon(
-    globaldb: Arc<globaldb::GlobalDB>,
-    coingecko: Arc<coingecko::Coingecko>,
-    data_dir: PathBuf,
+/// Given the path for an icon it returns its bytes and the extension of the file
```
LefterisJP, line 140:
```suggestion
/// Given the path for an icon return its bytes and the extension of the file
```

colibri/src/icons.rs
```diff
@@ -164,67 +166,139 @@ pub async fn get_or_query_icon(
...
-                            None => return (StatusCode::NOT_FOUND, None, None),
-                        },
+            _ => (StatusCode::NOT_FOUND, None, None),
+        },
+        _ => (StatusCode::NOT_FOUND, None, None),
+    }
+}
+
+// Writes a zero bytes file to mark that we tried to query this icon already
+// and was not possible to find it in our sources for icons.
```
LefterisJP, line 208:
```suggestion
// Writes a zero bytes file to mark that we already tried to query this icon
// and that it was not possible to find it in our sources for icons.
```

colibri/src/icons.rs
```diff
@@ -164,67 +166,139 @@ pub async fn get_or_query_icon(
...
+                "Unable to write zero bytes file {} due to {}",
+                path.display(),
+                e
+            );
+        });
+    debug!("Wrote zero bytes file {}", path.display());
+}
+
+/// query icon remotely from SmolDapp and coingecko. If the image couldn't
+/// be found this function returns false otherwise it returns true.
```
LefterisJP, line 223:
```suggestion
/// query icon remotely from SmolDapp and coingecko. Returns if the image was found (boolean)
```

colibri/src/icons.rs
```diff
@@ -164,67 +166,139 @@ pub async fn get_or_query_icon(
...
+        .map_err(|e| {
+            error!(
+                "Unable to write {} to the file system due to {}",
+                path.display(),
+                e
+            );
+        });
+}
+
+#[cfg(test)]
```
LefterisJP, line 274: same comment for tests as before

colibri/src/logging.rs
```diff
@@ -0,0 +1,44 @@
+use crate::args::Args;
+
+use tracing_appender::rolling::{RollingFileAppender, Rotation};
+use tracing_subscriber::filter::EnvFilter;
+
+// Config logging for the app. We allow to log to a system file
+// or to the stdout. If logs are stored in files they are rotated
+// daily and there is a max of `max_logfiles_num` files saved.
```
LefterisJP, line 8:
```suggestion
// Configure logging for the app. We allow logging to a system file
// or to the stdout. If logs are stored in files they are rotated
// daily and there is a max of `max_logfiles_num` files saved.
```

colibri/src/logging.rs
```diff
@@ -0,0 +1,44 @@
...
+        let filename: String = args
+            .logfile_path
+            .file_name()
+            .unwrap()
+            .to_str()
+            .unwrap()
+            .to_string();
+        let file_appender = RollingFileAppender::builder()
+            .max_log_files(args.max_logfiles_num)
+            .rotation(Rotation::DAILY)
```
LefterisJP, line 25: Do we need to rotate by day? I don't think so. It's just the size rotation we care about. With no usage at all we don't need to rotate every 24 hours. Just put a specific size for rotation. Same as the one in python

Also adjust the docstring when you do that

colibri/src/args.rs
yabirgb: yes, those in that list are valid. Unfortunatly help doesn't list hem. Only shows 

``` 
      --log-level <log-level>
          Log level for the app [default: info]
``` 

maybe I can adjust it. Need to check

colibri/Cargo.toml
yabirgb: This is a library created by tokio and maintained. Is the same one that lighthouse uses https://github.com/sigp/lighthouse/blob/029b4f21047c37d1ffde51c554737b1aa0880f97/Cargo.toml#L194 and I don't expect it to get unmaintained from one day to another.

colibri/src/coingecko.rs
yabirgb: this one is not testing the api but the logic for `query_asset_image`. I need to add integration tests for the API but don't want to keep this PR for longer

colibri/src/coingecko.rs
yabirgb: As mentioned in discord the tests/ folder doesn't behave like this cfg test. That folder doesn't have access to private functions and also macros aren't accessible. There are a couple of details. The tests folder is meant to be used when testing libraries that is what I believe we should do. Implement as a library and then use it in the rotki binary. But we can see in the future, for reference https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests-for-binary-crates

I need to add tests for the api endpoints also and I'll do that in the tests folder

colibri/src/args.rs
yabirgb: Now it looks like this

```
Usage: colibri [OPTIONS]

Options:
      --data-directory <data-directory>
          Sets the rotki data directory [default: "/Users/yabirgb/Library/Application Support/rotki/develop_data"]
      --port <port>
          Sets the colibri api port number [default: 6969]
      --logfile-path <logfile-path>
          Sets the path for the colibri logfile [default: colibri.log]
      --log-to-stdout
          Log to the stdout instead of the logfile
      --max-logfiles-num <max-logfiles-num>
          Max number of log files to keep [default: 5]
      --log-level <log-level>
          Log level for the app: ["error", "warn", "info", "debug", "trace"] [default: info]
      --max-size-in-mb <max-size-in-mb>
          Max size in MB for each one of the log files [default: 50]
  -h, --help
          Print help
  -V, --version
          Print 
```

colibri/Cargo.toml
yabirgb: This one is for using tokio in tests, tests that use the async runtime. they use `#[tokio::test]`

colibri/src/api/icons.rs
LefterisJP: This still looks suspiciously complicated. Can you at least add a small comment on the right of the drop `// BLA BLA` explaining why.

I don't have time to improve it unfortunately at this point.

LefterisJP (approved):
LGTM

PR #9368, opened 2025-02-04
Correctly count trades in asset balance graphs

nicholasyoder:
Related: https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=96493730

Fixes two things:
* Trades were not being queries correctly from the db - it was doing an `AND` when checking `base_asset` and `quote_asset` instead of an `OR`.
* In the logic for processing the queried events, it was finding a negative amount and failing when processing trades, since it was trying to process the amounts for both the spend and receive assets of the trade. Now it only processes the amount for the asset(s) whose balances are actually being queried.

### DIFF
--- a/rotkehlchen/balances/historical.py
+++ b/rotkehlchen/balances/historical.py
@@ -92,7 +92,11 @@ class HistoricalBalancesManager:
 
         current_balances: dict[Asset, FVal] = defaultdict(FVal)
         for event in events:
-            if self._update_balances(event=event, current_balances=current_balances) is not None:
+            if self._update_balances(
+                event=event,
+                current_balances=current_balances,
+                assets=(asset,),
+            ) is not None:
                 break
 
         try:
@@ -133,7 +137,11 @@ class HistoricalBalancesManager:
         current_balances: dict[Asset, FVal] = defaultdict(FVal)
         amounts: dict[Timestamp, FVal] = defaultdict(lambda: ZERO)
         for event in events:
-            if (negative_balance_data := self._update_balances(event=event, current_balances=current_balances)) is not None:  # noqa: E501
+            if (negative_balance_data := self._update_balances(
+                event=event,
+                current_balances=current_balances,
+                assets=assets,
+            )) is not None:
                 break
 
             for asset in assets:
@@ -143,7 +151,6 @@ class HistoricalBalancesManager:
                     else ts_ms_to_sec(event.timestamp)
                 )
                 amounts[event_ts] += current_balances[asset]
-
         return amounts, negative_balance_data
 
     def get_historical_netvalue(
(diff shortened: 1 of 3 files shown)


rotkehlchen/balances/historical.py
```diff
@@ -301,8 +308,11 @@ def _get_events_and_currency(
             trade_bindings: list[Any] = []
             for filter_ in trade_filters:
                 clauses, bindings = filter_.prepare()
-                trade_where_clauses.extend(clauses)
                 trade_bindings.extend(bindings)
+                if not filter_.and_op:
+                    trade_where_clauses.append(f'({" OR ".join(clauses)})')
+                else:
+                    trade_where_clauses.extend(clauses)
```
prettyirrelevant, line 315: ah, nice catch! thanks. instead of doing this, why not try fixing the filter class itself?

1. create a new `DBNestedOrFilter` class specifically for this case.
2. enhance `DBNestedFilter` to just combine all clauses and set the operator based on `and_op` of the `DBNestedFiter`. would need to update its docstring too.

i lean toward option 2 since it keeps things simpler. thoughts?

rotkehlchen/balances/historical.py
```diff
@@ -431,6 +444,12 @@ def _update_balances(
                 spend_amount = event.amount
                 receive_amount = event.amount * event.rate  # type: ignore[assignment]  # AssetAmount is the same as fval
 
+            if assets is not None:  # Only include amounts for the specified assets.
+                if spend_asset not in assets:
+                    spend_amount = ZERO
+                if receive_asset not in assets:
+                    receive_amount = ZERO  # type: ignore[assignment]  # AssetAmount is the same as fval
+
```
prettyirrelevant, line 452:
```suggestion
           if isinstance(event, Trade):
        if event.trade_type in {TradeType.BUY, TradeType.SETTLEMENT_BUY}:
            spend_asset = event.quote_asset
            receive_asset = event.base_asset
            spend_amount = event.amount * event.rate
            receive_amount = event.amount
        else:
            spend_asset = event.base_asset
            receive_asset = event.quote_asset
            spend_amount = event.amount
            receive_amount = event.amount * event.rate  # type: ignore[assignment]  # AssetAmount is the same as fval

        # Only process amounts for tracked assets
        spend_amount = ZERO if (assets is not None and spend_asset not in assets) else spend_amount
        receive_amount = ZERO if (assets is not None and receive_asset not in assets) else receive_amount  # type: ignore[assignment]
```
formatting is a bit off in my suggestion.

rotkehlchen/balances/historical.py
nicholasyoder: Yes, went with option 2. At first I thought this might break where `DBNestedFiter` is already being used in the `LevenshteinFilterQuery`, but as far as I can tell it seems to work correctly.

LefterisJP (approved):
LGTM

PR #9376, opened 2025-02-05
Fix coinbase advanced trade market orders

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=96520463

Fixes a bug where there are multiple events returned from the api for a single order. Now it should correctly count the amounts from all of the events into one trade and calculate a weighted average rate.

### DIFF
--- a/rotkehlchen/exchanges/coinbase.py
+++ b/rotkehlchen/exchanges/coinbase.py
@@ -229,8 +229,9 @@ class Coinbase(ExchangeInterface):
         self.apiversion = 'v2'
         self.base_uri = 'https://api.coinbase.com'
         self.host = 'api.coinbase.com'
-        # keep orderids of advanced trades while parsing. As some appear two times showing both debit and credit part of the trade.  # noqa: E501
-        self.advanced_trade_orders: set[str] = set()
+        # Maps advanced trade order ids to the trade currency so unneeded events can be
+        # skipped when both the debit and credit part of the trade is present.
+        self.advanced_orders_to_currency: dict[str, str] = {}
 
     def is_legacy_key(self, api_key: str) -> bool:
         if LEGACY_RE.match(api_key):
@@ -905,12 +906,15 @@ class Coinbase(ExchangeInterface):
                 log.error(f'Got non buy/sell order side in Coinbase advanced trade fill: {event}')
                 return None
 
-            if (order_id := event['advanced_trade_fill'].get('order_id')):
-                if order_id in self.advanced_trade_orders:
-                    log.debug('Ignoring already seen coinbase advanced trade other side', other_side=event)  # noqa: E501
+            if (
+                (order_id := event['advanced_trade_fill'].get('order_id')) is not None and
+                (event_currency := event['amount'].get('currency')) is not None
+            ):
+                if (existing_trade_currency := self.advanced_orders_to_currency.get(order_id)) is None:  # noqa: E501
+                    self.advanced_orders_to_currency[order_id] = event_currency
+                elif existing_trade_currency != event_currency:
+                    log.debug('Ignoring other side of already seen coinbase advanced trade', other_side=event)  # noqa: E501
                     return None
-
-                self.advanced_trade_orders.add(order_id)
 
         else:
             trade_type = TradeType.BUY if amount < 0 else TradeType.SELL
--- a/rotkehlchen/tests/exchanges/test_coinbase.py
+++ b/rotkehlchen/tests/exchanges/test_coinbase.py
@@ -5,6 +5,7 @@ import pytest
 import requests
 
 from rotkehlchen.accounting.structures.balance import Balance
+from rotkehlchen.assets.asset import Asset
 from rotkehlchen.assets.converters import asset_from_coinbase
 from rotkehlchen.constants import ONE, ZERO
 from rotkehlchen.constants.assets import A_1INCH, A_BTC, A_ETH, A_EUR, A_SOL, A_USD, A_USDC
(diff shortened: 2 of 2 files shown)


docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` Fix a bug where some Coinbase advanced market order trades were showing the wrong amount.
```
yabirgb, line 5: We already have 

```
 :bug:`9308` Fix some cases of Coinbase events that were not properly understood in rotki.
 ``` 
 
 So I believe it falls there

File: rotkehlchen/exchanges/coinbase.py [hunk 2]
```diff
@@ -905,37 +907,58 @@ def _process_advanced_trade(self, event: dict[str, Any], timestamp: Timestamp) -
...
+                    fee=fee_amount,  # type: ignore  # abs() doesn't propagate Fee type
+                    fee_currency=fee_asset,
+                    link=str(event['id']),
+                )),
+                [(amount, rate)],
+            )
+            return new_trade
+        else:  # there are multiple events for this order - update the existing trade.
+            existing_trade = existing_order_data[0]
+            existing_trade.amount += amount  # type: ignore  # both are AssetAmount, but mypy sees it as FVal after adding them
```
yabirgb, line 951: I believe they get added as FVal and not assetAmount. YOu would need to overwrite the __iadd__ method on assetAmount if I'm not mistaken

rotkehlchen/exchanges/coinbase.py
```diff
@@ -905,37 +907,58 @@ def _process_advanced_trade(self, event: dict[str, Any], timestamp: Timestamp) -
...
+            existing_trade.fee += fee_amount  # type: ignore  # fee will not be None here
```
yabirgb, line 952: same for fee

rotkehlchen/tests/exchanges/test_coinbase.py
```diff
@@ -1118,7 +1119,7 @@ def test_coinbase_query_trade_history_advanced_fill(function_scope_coinbase):
                 "product_id": "ETH-USDC",
                 "order_id": "orderid3",
                 "commission": "0.0000005",
-                "order_side": "sell"
+                "order_side": "buy"
```
nicholasyoder, line 1122: I changed this, as it would seem to be what is intended here. This event was being skipped with the 'ignoring already seen trade' message, but now it only skips if the trade is actually of the other type. The other trade in this test (`id3`) is also a sell, so I think this one was intended to be a buy, especially since the sell trade has negative amounts and this one is positive.

rotkehlchen/exchanges/coinbase.py
```diff
@@ -905,37 +907,58 @@ def _process_advanced_trade(self, event: dict[str, Any], timestamp: Timestamp) -
...
+
+            # find the weighted average rate using the amounts and rates from all the trades in the order  # noqa: E501
+            total_cost = FVal(0)
+            existing_order_data[1].append((amount, rate))
+            for trade_amount, trade_rate in existing_order_data[1]:
+                total_cost += trade_amount * trade_rate
```
yabirgb, line 958: question. I believe you can avoid the for loop and this is

```
new_rate = (previus_rate * new total)/previous total + (new_trade_amount * new_trade_rate)/new total
```

can you check it? I might have made a mistake but I believe it saves you the whole loop

rotkehlchen/exchanges/coinbase.py
```diff
@@ -905,37 +907,58 @@ def _process_advanced_trade(self, event: dict[str, Any], timestamp: Timestamp) -
...
+                [(amount, rate)],
+            )
+            return new_trade
+        else:  # there are multiple events for this order - update the existing trade.
+            existing_trade = existing_order_data[0]
+            existing_trade.amount += amount  # type: ignore  # both are AssetAmount, but mypy sees it as FVal after adding them
+            existing_trade.fee += fee_amount  # type: ignore  # fee will not be None here
+
+            # find the weighted average rate using the amounts and rates from all the trades in the order  # noqa: E501
+            total_cost = FVal(0)
```
yabirgb, line 955: We have the ZERO constant

rotkehlchen/tests/exchanges/test_coinbase.py
```diff
@@ -1189,6 +1236,17 @@ def test_coinbase_query_trade_history_advanced_fill(function_scope_coinbase):
...
+    ), Trade(
+        timestamp=1711371791,
+        location=Location.COINBASE,
+        base_asset=Asset('NEAR'),
+        quote_asset=A_USD,
+        trade_type=TradeType.BUY,
+        amount=FVal('27.8100000000000000'),
+        rate=FVal('7.40892808342322905429701546206400575332614167565623876303487953973390866594750'),
+        fee=FVal('2.0604229'),
+        fee_currency=A_USD,
```
yabirgb, line 1248: does this match what you see in the coinbase UI?

rotkehlchen/exchanges/coinbase.py
```diff
@@ -905,37 +907,58 @@ def _process_advanced_trade(self, event: dict[str, Any], timestamp: Timestamp) -
...
-
-                self.advanced_trade_orders.add(order_id)
-
         else:
             trade_type = TradeType.BUY if amount < 0 else TradeType.SELL
 
+        order_id = event['advanced_trade_fill']['order_id']
+        if (
+            (existing_order_data := self.advanced_trade_orders.get(order_id)) is not None and
+            existing_order_data[0].trade_type != trade_type
```
yabirgb, line 916: shouldn't you ignore if the trade type is the same? I'm not entirely sure if I follow this part

rotkehlchen/exchanges/coinbase.py
nicholasyoder: So I'm doing something a little different than this, but yes, that loop was unneeded. It just keeps track of the total_cost of each trade and then divides that by the total amount traded to get the rate.

rotkehlchen/tests/exchanges/test_coinbase.py
nicholasyoder: so there were actually more trades than this in the order in my real data, so no the numbers here don't match, but when I run it in the app and it processes all the trades, then yes it matches.

rotkehlchen/exchanges/coinbase.py
nicholasyoder: This is so that it ignores when both sides of a trade have events in the api result. So like if you bought 0.5 ETH with USDC, there could be both a buy event of eth and a sell event of USDC. So it needs to skip the other side of the event if it is present.

This was added here: https://github.com/rotki/rotki/commit/61884a24cbc5e46280fc0538e6a8b0a0cbf449d5

rotkehlchen/exchanges/coinbase.py
```diff
@@ -82,6 +84,12 @@
 )
 
 
+@dataclass
+class AdvancedTradeOrderData:
+    trade: 'Trade'
```
yabirgb, line 89: nitpick: this should have a docstring

rotkehlchen/tests/exchanges/test_coinbase.py
nicholasyoder: I'm actually completely removing this event (was `id5` before my changes) from the test data, as I think it is incorrect.

Notice that `id3` and `id5` were identical except that the values were switched from negative to positive. This doesn't make sense - why would there be two events at the same timestamp, for the same asset, one going negative and one positive the same absolute amount?

In my own data when there is an event for the "other side" of a trade, it is for the other asset. For example, I have a buy of 4.13 NEAR with 30.61 USD which has the following two events in the coinbase api results:

```
{'advanced_trade_fill': {... 'order_side': 'buy', 'product_id': 'NEAR-USD'}, 'amount': {'amount': '4.1340000000000000', 'currency': 'NEAR'}, 'created_at': '2024-03-25T13:03:11Z', ... 'native_amount': {'amount': '30.61', 'currency': 'USD'}, ...}
{'advanced_trade_fill': {... 'order_side': 'buy', 'product_id': 'NEAR-USD'}, 'amount': {'amount': '-30.61', 'currency': 'USD'}, 'created_at': '2024-03-25T13:03:11Z', ... 'native_amount': {'amount': '-30.62', 'currency': 'USD'}, ...}
```

Looking at the way this was originally implemented, I think this `id5` event was constructed manually by copy/pasting `id3` and then making the amount negative - since it was using negative/positive amounts to determine which event to ignore:
https://github.com/rotki/rotki/blob/3e39f48b7b150b37f312454a03cccab32b5097f9/rotkehlchen/exchanges/coinbase.py#L727

This relied on already knowing the trade type though, which is something that apparently is not always present, as it was reworked here: https://github.com/rotki/rotki/commit/61884a24cbc5e46280fc0538e6a8b0a0cbf449d5
But this change breaks multiple fills per order, as it only processes one event per order and then ignores the rest.

So now I've changed it to only ignore events if they have a differing currency (base asset) - so in the above example it will process the events with `'currency': 'NEAR'` and skip the ones with `'currency': 'USD'`.

rotkehlchen/exchanges/coinbase.py
```diff
@@ -229,8 +229,9 @@ def __init__(
         self.apiversion = 'v2'
         self.base_uri = 'https://api.coinbase.com'
         self.host = 'api.coinbase.com'
-        # keep orderids of advanced trades while parsing. As some appear two times showing both debit and credit part of the trade.  # noqa: E501
-        self.advanced_trade_orders: set[str] = set()
+        # Maps advanced trade order ids to the trade currency so unneeded events can be
+        # skipped when both the debit and credit part of the trade is present.
+        self.advanced_trade_orders: dict[str, str] = {}
```
LefterisJP, line 234:
```suggestion
        self.advanced_orders_to_currency: dict[str, str] = {}
```

rotkehlchen/exchanges/coinbase.py
```diff
@@ -905,13 +906,16 @@ def _process_advanced_trade(self, event: dict[str, Any], timestamp: Timestamp) -
...
-                if order_id in self.advanced_trade_orders:
-                    log.debug('Ignoring already seen coinbase advanced trade other side', other_side=event)  # noqa: E501
+            if (
+                (order_id := event['advanced_trade_fill'].get('order_id')) is not None and
+                (event_currency := event['amount'].get('currency')) is not None
+            ):
+                if (existing_trade_currency := self.advanced_trade_orders.get(order_id)) is None:
+                    self.advanced_trade_orders[order_id] = event_currency
+                elif existing_trade_currency != event_currency:
+                    log.debug('Ignoring other side of already seen coinbase advanced trade', other_side=event)  # noqa: E501
```
LefterisJP, line 916: this is not an error right? If it's the same order if and different currency it's the other side? But don't you also need that side?

nicholasyoder: Yes. This isn't an error here. It just needs to ignore these events from the other side of the trade, since they are showing the same thing just in the other currency.

LefterisJP (approved):
LGTM. Let's merge this for now. And we can check with all of our data

PR #9382, opened 2025-02-06
Multimev txs block

LefterisJP:
- All block events with sequence index 0 represent the block
production event and its block proposal fee. If the fee recipient is
not tracked then it's an info event
- All block events with sequence index 1, if existing, represent the
amount the MEV relayer claims the mev fee recipient earned. This is
always an info event as it does not represent actual transactions.
- All receive transactions in produced blocks with mev fee recipient
as receiver will be moved at sequence_index 2++ of the block event
identifier.
- We also need a DB upgrade that correctly sets info type to non
tracked fee recipient block events and also resets any sequence index > 1 events so the new logic can run and combine them properly.
- In the performance endpoint of eth staking the "execution" key of each individual validator and in the sums no longer exists. It's now split into two. "execution_blocks" and "execution_mev".
- When adding or removing an address we need to re-evaluate the
informational type of block production events.
- Update the accounting rules fixture logic for tests
- [[Hacky] Accounting rules updates are now forced.](https://github.com/rotki/rotki/pull/9382/commits/a2eac86c3182bec60253d21d6a13877f10489337) 
a2eac86
Related: https://github.com/rotki/data/pull/162
This forced update is only there to allow us editing the accounting
rules. With the current methodology there is no other way to do so.
Related TODO:
https://github.com/orgs/rotki/projects/11?pane=issue&itemId=96831912

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -8340,7 +8340,8 @@ Getting eth2 staking performance
           "result": {
               "sums": {
                   "apr": "0.0597652039949379861158979362586657031468575710584143257025001370766265371913491",
-                  "execution": "0.951964836013963505",
+                  "execution_blocks": "0.451964836013963505",
+                  "execution_mev": "0.31964836013963505",
                   "exits": "0.0014143880000005993",
                   "outstanding_consensus_pnl": "0.000829238",
                   "sum": "3.2351487110139639043",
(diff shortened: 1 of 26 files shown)


LefterisJP:
Fucking finally all green. This took way too long and it's a complicated PR. Wrote all changes in the subject of the PR.

Also needs frontend changes for the performance endpoint of ETH staking.

Note: test_balances_caching_mixup failed with 502 from etherscan once here and locally and I am not sure why. Trying to investigate it passed properly.

We have it in the [flaky/TO VCR tests](https://github.com/orgs/rotki/projects/11?pane=issue&itemId=71130135) already. It needs some investigation I guess.

PR #9383, opened 2025-02-06
simplify gemini exchange symbols validation

prettyirrelevant:
simplify the Gemini exchange's symbol parsing based on current trading pairs from:
- https://api.gemini.com/v1/symbols and
- https://support.gemini.com/hc/en-us/articles/115005868106-What-cryptos-are-supported-on-the-Gemini-Exchange

### DIFF
--- a/rotkehlchen/exchanges/gemini.py
+++ b/rotkehlchen/exchanges/gemini.py
@@ -26,7 +26,6 @@ from rotkehlchen.exchanges.exchange import ExchangeInterface, ExchangeQueryBalan
 from rotkehlchen.exchanges.utils import (
     deserialize_asset_movement_address,
     get_key_if_has_val,
-    pair_symbol_to_base_quote,
 )
 from rotkehlchen.history.deserialization import deserialize_price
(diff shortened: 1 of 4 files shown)


rotkehlchen/exchanges/gemini.py
```diff
@@ -71,17 +70,27 @@ class GeminiPermissionError(Exception):
...
+        raise UnprocessableTradePair(symbol)
+
+    special_cases = {'moodengusd': ('MOODENG', 'USD')}
+    if symbol in special_cases:
+        base, quote = special_cases[symbol]
+        return asset_from_gemini(base), asset_from_gemini(quote)
+
+    # from gemini api, quote assets are either 4 chars ('gusd', 'usdt')
+    # or 3 chars ('usd', 'btc', etc). Try 4-char quotes first.
+    if len(symbol) >= 4 and symbol[-4:] in {'gusd', 'usdt', 'paxg'}:
```
LefterisJP, line 87: If the `gemini_symbol_to_base_quote` was broken that badly that means that gemini trades were also not read for all of these pairs. So wouldn't resetting data just like we did with coinbase make sense here?

PR #9386, opened 2025-02-07
simplify bybit exchange symbols validation

prettyirrelevant:
similar to #9383

### DIFF
--- a/rotkehlchen/db/upgrades/v46_v47.py
+++ b/rotkehlchen/db/upgrades/v46_v47.py
@@ -62,30 +62,6 @@ def upgrade_v46_to_v47(db: 'DBHandler', progress_handler: 'DBUpgradeProgressHand
                 (0,),  # decoded tx state
             )
 
-    @progress_step(description='Reset coinbase events and cache')
-    def _reset_coinbase_events(write_cursor: 'DBCursor') -> None:
-        write_cursor.execute(
-            'DELETE FROM history_events WHERE location=? AND event_identifier NOT LIKE ?',
-            ((coinbase_db_loc := Location.COINBASE.serialize_for_db()), f'{ROTKI_EVENT_PREFIX}_%'),
-        )
-        write_cursor.execute(  # only deleting trades that were added via the coinbase api
-            "DELETE FROM trades WHERE location=? AND link != ''",
-            (coinbase_db_loc,),
-        )
-        write_cursor.execute(
-            'DELETE FROM key_value_cache WHERE name LIKE ? OR name LIKE ?',
-            (f'{(coinbase_loc := Location.COINBASE.serialize())}_%_last_query_ts', f'{coinbase_loc}_%_last_query_id'),  # noqa: E501
-        )
-        write_cursor.execute(
-            'DELETE FROM used_query_ranges WHERE name LIKE ? OR name LIKE ? OR name LIKE ? OR name LIKE ?',  # noqa: E501
-            (
-                f'{coinbase_loc}_history_events_%',
-                f'{coinbase_loc}_trades_%',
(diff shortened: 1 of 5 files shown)


rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -206,4 +206,24 @@ def _reset_gemini_events(write_cursor: 'DBCursor') -> None:
             ),
         )
 
+    @progress_step(description='Reset bybit events and cache')
+    def _reset_bybit_events(write_cursor: 'DBCursor') -> None:
+        write_cursor.execute(
```
yabirgb, line 211: can't we just combine it with the previous step? You are repeting code here that could be a for loop unless I'm missing something

rotkehlchen/exchanges/bybit.py
```diff
@@ -88,13 +81,11 @@ def bybit_symbol_to_base_quote(
     if split_symbol is not None and len(split_symbol) == 2:
         base_asset = asset_from_bybit(split_symbol[0])
         quote_asset = asset_from_bybit(split_symbol[1])
+    # If more four-letter quote suffixes are discovered, update the tuple below accordingly.
+    elif len(symbol) >= 4 and symbol[-4:] in ('USDT', 'USDC', 'USDE'):
```
yabirgb, line 85: why not read the from the mappings? I believe is better than hardcoding them since it allows the user to fix issues without us needed to make a new release. something like what `six_letter_assets` was doing before.

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -186,24 +186,25 @@ def _remove_nft_collection_assets(write_cursor: 'DBCursor') -> None:
...
-        write_cursor.execute(
-            'DELETE FROM used_query_ranges WHERE name LIKE ? OR name LIKE ? OR name LIKE ? OR name LIKE ?',  # noqa: E501
-            (
-                f'{(gemini_loc := Location.GEMINI.serialize())}_history_events_%',
-                f'{gemini_loc}_trades_%',
-                f'{gemini_loc}_margins_%',
-                f'{gemini_loc}_asset_movements_%',
-            ),
-        )
+    @progress_step(description='Reset exchanges events and cache')
```
LefterisJP, line 189: can't you also put coinbase here? Coinbase has only one extra call which you can just add right? or am I missing something?

rotkehlchen/exchanges/bybit.py
```diff
@@ -370,10 +359,9 @@ def query_online_trade_history(
                 try:
                     base_asset, quote_asset = bybit_symbol_to_base_quote(
                         symbol=raw_trade['symbol'],
-                        five_letter_assets=self.five_letter_assets,
```
LefterisJP, line 373: hmmm I don't think this PR has been thought of or checked enough. @yabirgb please check this more carefully. You have data for this exchange. I don't.

Why are you no longer using the five and six ones? They are still initialized in the constructor and have not been touched.

```
        self.five_letter_assets = set()
        self.six_letter_assets = set()

            cursor.execute(
                'SELECT exchange_symbol FROM location_asset_mappings WHERE (location IS ? OR location IS NULL) AND LENGTH(exchange_symbol) >= 5;',  # noqa: E501
                (Location.BYBIT.serialize_for_db(),),
            )
            for symbol in cursor:
                if len(symbol[0]) == 5:
                    self.five_letter_assets.add(symbol[0])
                elif len(symbol[0]) == 6:
                    self.six_letter_assets.add(symbol[0])
```

LefterisJP (approved):
LGTM. This is also gonna conflict with my work. Kill me.

PR #9391, opened 2025-02-07
Rework deposit/withdraw event subtypes

nicholasyoder:
Introduces new `DEPOSIT_FOR_WRAPPED` and `REDEEM_WRAPPED` history event subtypes replacing `DEPOSIT_ASSET` and `REMOVE_ASSET` where ever there is a corresponding `RECEIVE_WRAPPED` or `RETURN_WRAPPED` as discussed in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1337395621874630676)


rotkehlchen/history/events/structures/types.py
```diff
@@ -91,6 +91,8 @@ class HistoryEventSubType(SerializableEnumNameMixin):
     CASHBACK = auto()
     HACK = auto()
     CLAWBACK = auto()
+    DEPOSIT_FOR_WRAPPED = auto()
+    REDEEM_WRAPPED = auto()
```
yabirgb, line 95: this two lines forced everything else 🤣

rotkehlchen/chain/ethereum/modules/convex/decoder.py
```diff
@@ -176,7 +177,7 @@ def _decode_gauge_events(self, context: DecoderContext) -> DecodingOutput:
             ):
                 if context.tx_log.topics[0] in WITHDRAWAL_TOPICS:
                     event.event_type = HistoryEventType.WITHDRAWAL
-                    event.event_subtype = HistoryEventSubType.REMOVE_ASSET
+                    event.event_subtype = HistoryEventSubType.REDEEM_WRAPPED if found_return_wrapped else HistoryEventSubType.REMOVE_ASSET  # noqa: E501
```
yabirgb, line 180: this one I've checked and indeed you can lock cvxCRV and you don't receive anything in return

LefterisJP (approved):
LGTM.

As discussed in discord the failing test will be fixed once the global DB is repackaged.

There may still be not properly decoded events but we would have to find them on an individual basis from here and on

PR #9393, opened 2025-02-07
Ignore assets with collectible ids in db upgrade

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=96870115

### DIFF
--- a/rotkehlchen/db/upgrades/v46_v47.py
+++ b/rotkehlchen/db/upgrades/v46_v47.py
@@ -72,7 +72,8 @@ def upgrade_v46_to_v47(db: 'DBHandler', progress_handler: 'DBUpgradeProgressHand
                 'SELECT identifier FROM assets WHERE identifier IN ('
                 "SELECT identifier FROM evm_tokens WHERE token_kind = 'B' "
                 "UNION SELECT REPLACE(identifier, 'erc721', 'erc20') "
-                "FROM evm_tokens WHERE token_kind = 'B');",
+                "FROM evm_tokens WHERE token_kind = 'B') AND "
+                "LENGTH(identifier) <= 62;",  # Skip identifiers with collectible ids (will be longer than 62)  # noqa: E501
             ))
 
         if (num_to_remove := len(identifiers_to_remove)) == 0:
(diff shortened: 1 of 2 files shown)


rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -72,7 +72,8 @@ def _remove_nft_collection_assets(write_cursor: 'DBCursor') -> None:
                 'SELECT identifier FROM assets WHERE identifier IN ('
                 "SELECT identifier FROM evm_tokens WHERE token_kind = 'B' "
                 "UNION SELECT REPLACE(identifier, 'erc721', 'erc20') "
-                "FROM evm_tokens WHERE token_kind = 'B');",
+                "FROM evm_tokens WHERE token_kind = 'B') AND "
+                "LENGTH(identifier) - LENGTH(REPLACE(identifier, '/', '')) = 1;",
```
LefterisJP, line 76: Isn't the length standard? I mean the length of the token identifiers for NFTs ... unless I recall wrong aren't they always the same? So can't you check if it's bigger than this standard length? (along with a comment explaining what you are doing here.

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2744,6 +2744,10 @@ def _check_tokens_existence(user_db_cursor: 'DBCursor', expect_removed: bool = F
                 'SELECT COUNT(*) FROM assets WHERE identifier IN (?, ?)',
                 (uniswap_erc20_token.identifier, uniswap_erc721_token.identifier),
             ).fetchone()[0] == 0 if expect_removed else 2
+            assert global_db_cursor.execute(
+                'SELECT COUNT(*) FROM assets WHERE identifier = ?',
+                (erc721_token_with_id.identifier,),
+            ).fetchone()[0] == 1  # tokens with collectible ids shouldn't be modified
```
LefterisJP, line 2750: did you add it yourself in the DB or was it added before?

rotkehlchen/db/upgrades/v46_v47.py
nicholasyoder: yes, that will make it simpler.

rotkehlchen/tests/db/test_db_upgrades.py
nicholasyoder: Its added here along with the others used in the test: https://github.com/rotki/rotki/blob/6bfb21bfb74c2548a646e301d32cb9fc161b69fe/rotkehlchen/tests/db/test_db_upgrades.py#L2780

We have to add these within the test as none of these are assets that we want to actually have in the packaged globaldb.

LefterisJP (approved):
LGTM

PR #9399, opened 2025-02-10
Fix withdrawal from aura

yabirgb:
We were lacking logic to process withdrawals from aura. I checked and this one seems to be the only method being call.

Closes #(issue_number)

## Checklist


rotkehlchen/chain/evm/decoding/aura_finance/decoder.py
```diff
@@ -125,34 +134,30 @@ def _decode_lock_aura_bridged(self, context: DecoderContext) -> DecodingOutput:
...
+                    event.balance.amount == asset_normalized_value(
+                        amount=amount_paid,
+                        asset=(event.asset.resolve_to_crypto_asset()),
+                    ) or
                     event.asset in self.base_reward_tokens
-                ) and
-                event.location_label == recipient
+                )
             ):
+                crypto_asset = event.asset.resolve_to_crypto_asset()
```
nicholasyoder, line 160:
```suggestion
```
asset should always be getting resolved to a crypto asset in the `if` above

nicholasyoder (approved):

LefterisJP (approved):
LGTM

PR #9402, opened 2025-02-10
Add db upgrade removing old accounting rules

nicholasyoder:
Remove old accounting rules via a db upgrade since the accounting rule updates don't support deletion yet as noted here: https://github.com/orgs/rotki/projects/11?pane=issue&itemId=96831912

Only removes the counterparty specific rules - the base rules are still needed since `deposit asset` and `remove asset` are still being used sometimes.

Also adds the `STAKING/REDEEM_WRAPPED` combination. `STAKING/DEPOSIT_FOR_WRAPPED` is [being used for gmx](https://github.com/rotki/rotki/blob/develop/rotkehlchen/tests/unit/decoders/test_gmx.py#L234) - but not `REDEEM_WRAPPED`, so this wasn't added initially. But now we've added an [accounting rule for staking / redeem wrapped](https://github.com/rotki/data/blob/develop/updates/accounting_rules/v5.json#L59), which shows in the app with an unknown `Resulting combination` unless its defined here.

### DIFF
--- a/rotkehlchen/db/upgrades/v46_v47.py
+++ b/rotkehlchen/db/upgrades/v46_v47.py
@@ -227,4 +227,23 @@ def upgrade_v46_to_v47(db: 'DBHandler', progress_handler: 'DBUpgradeProgressHand
             (f'{(coinbase_loc := Location.COINBASE.serialize())}_%_last_query_ts', f'{coinbase_loc}_%_last_query_id'),  # noqa: E501
         )
 
+    @progress_step(description='Remove old accounting rules')
+    def _remove_old_accounting_rules(write_cursor: 'DBCursor') -> None:
+        """Remove unused `deposit asset` and `remove asset` accounting rules.
+        This must be done here as the accounting rule updates don't currently support deletion.
+        See https://github.com/orgs/rotki/projects/11?pane=issue&itemId=96831912
+        """
+        counterparties = ('aave-v1', 'aave-v2', 'aave-v3', 'compound', 'compound-v3', 'Locked GNO', 'yearn-v1', 'yearn-v2')  # noqa: E501
+        write_cursor.executemany(
+            'DELETE FROM accounting_rules WHERE type=? AND subtype=? AND counterparty=?',
+            [
+                (event_type, event_subtype, counterparty)
+                for counterparty in counterparties
+                for event_type, event_subtype in (
+                    ('deposit', 'deposit asset'),
+                    ('withdrawal', 'remove asset'),
+                )
+            ],
+        )
+
     perform_userdb_upgrade_steps(db=db, progress_handler=progress_handler, should_vacuum=True)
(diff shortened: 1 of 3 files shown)


yabirgb (approved):
LGTM

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -227,4 +227,23 @@ def _reset_exchanges_events_and_cache(write_cursor: 'DBCursor') -> None:
             (f'{(coinbase_loc := Location.COINBASE.serialize())}_%_last_query_ts', f'{coinbase_loc}_%_last_query_id'),  # noqa: E501
         )
 
+    @progress_step(description='Remove old accounting rules')
+    def _remove_old_accounting_rules(write_cursor: 'DBCursor') -> None:
+        """Remove unused `deposit asset` and `remove asset` accounting rules.
+        This must be done here as the accounting rule updates don't currently support deletion.
+        See https://github.com/orgs/rotki/projects/11?pane=issue&itemId=96831912
+        """
+        counterparties = ('aave-v1', 'aave-v2', 'aave-v3', 'compound', 'compound-v3', 'Locked GNO', 'yearn-v1', 'yearn-v2')  # noqa: E501
```
LefterisJP, line 236: Questions: What is the defautl rule for deposit/deposit and withdraw/remove asset? it's a no_op for sure right? Because it will fall back to it.

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -227,4 +227,23 @@ def _reset_exchanges_events_and_cache(write_cursor: 'DBCursor') -> None:
...
+        write_cursor.executemany(
+            'DELETE FROM accounting_rules WHERE type=? AND subtype=? AND counterparty=?',
```
LefterisJP, line 238: Also shouldn't we then undo this: `https://github.com/rotki/data/commit/73c71608fb736f63ec9b697132ea1ecf7b0ec470`
and remove those from the new accounting rules? They were only added there since i needed to somehow "edit/delete"  those old rules? 

Can't you just edit them here?

rotkehlchen/db/upgrades/v46_v47.py
nicholasyoder: the default rules for these have these options:
```
"taxable": false,
"count_entire_amount_spend": false,
"count_cost_basis_pnl": true,
"accounting_treatment": null
```
Which should be correct I think.

rotkehlchen/db/upgrades/v46_v47.py
nicholasyoder: As mentioned in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1338909245654958162) editing existing rules here won't affect what rules someone running rotki for the first time gets. So I think that edit needs to stay as is in the accounting rule upgrade in the data repo.

LefterisJP (approved):
LGTM as discussed in discord

PR #9404, opened 2025-02-11
refactor historical price oracle

prettyirrelevant:


rotkehlchen/interfaces.py
```diff
@@ -70,20 +72,73 @@ def can_query_history(
...
         - RemoteError
         """
 
+    @abc.abstractmethod
+    def oracle_source(self) -> HistoricalPriceOracle:
+        """Source enum used as key for storing and retrieving cached prices"""
+
+    @abc.abstractmethod
+    def max_cache_seconds(self) -> int:
+        """Maximum age for cached prices before they must be refetched"""
```
yabirgb, line 95: for what I see this is the max time difference for which a price is valid. At least is the usage you do so this docstring is misleading. Also maybe a better name?

rotkehlchen/interfaces.py
```diff
@@ -70,20 +72,73 @@ def can_query_history(
...
+
+    def query_historical_price(
+            self,
+            from_asset: Asset,
+            to_asset: Asset,
+            timestamp: Timestamp,
+    ) -> Price:
+        """Fetch historical price with caching.
+        Uses cached price if fresh enough, otherwise queries and caches new price.
+
```
yabirgb, line 105: mention that price is persisted in the db

prettyirrelevant: > Uses cached price if fresh enough, otherwise queries and "caches new price".

it's mentioned already.

rotkehlchen/interfaces.py
prettyirrelevant: done.

rotkehlchen/interfaces.py
```diff
@@ -70,20 +72,73 @@ def can_query_history(
...
+        """Implementation specific historical price query logic.
 
         If no price can be found may raise:
         - PriceQueryUnsupportedAsset
         - NoPriceForGivenTimestamp
         - RemoteError
         """
 
+    @abc.abstractmethod
+    def oracle_source(self) -> HistoricalPriceOracle:
```
LefterisJP, line 90: So for these new ones can't you turn them into a constant?

```python
class Base(ABC):
    @abstractmethod
    def CONSTANT(self) -> Final[int]: ...

class Child(Base):
    CONSTANT: Final[int] = 42
```

Essentially you would need to add the `Final` typing and initialize at all Children oracle class definitions.
This is more performant as it's stored directly in the class's namespace rather than being computed via property/method calls.

rotkehlchen/interfaces.py
```diff
@@ -70,20 +72,73 @@ def can_query_history(
...
+        """Source enum used as key for storing and retrieving cached prices"""
+
+    @abc.abstractmethod
+    def max_price_age_in_seconds(self) -> int:
+        """Maximum time difference in seconds for which a historical price is considered valid"""
```
LefterisJP, line 95: This is wrong. The name suggests something that it's not and it's misleading. This is not the price age. These are historical prices. It's how far from the requested timestamp in either direction a price is found.

```suggestion
    def max_seconds_distance(self) -> int:
        """Maximum distance in seconds around a given timestamp for which if a price is found it will be considered valid and returned"""
```

rotkehlchen/interfaces.py
```diff
@@ -70,20 +72,73 @@ def can_query_history(
...
+
+    def query_historical_price(
+            self,
+            from_asset: Asset,
+            to_asset: Asset,
+            timestamp: Timestamp,
+    ) -> Price:
+        """Fetch historical price with caching.
+        Uses cached price if fresh enough, otherwise queries and caches new price.
```
LefterisJP, line 104:
```suggestion
        Uses a cached price if one exists close enough to given timestamp. Otherwise queries and caches the new price.
```

rotkehlchen/interfaces.py
```diff
@@ -70,20 +72,73 @@ def can_query_history(
...
+
+        May raise:
+        - PriceQueryUnsupportedAsset
+        - NoPriceForGivenTimestamp
+        - RemoteError
+        """
+        price_cache_entry = GlobalDBHandler.get_historical_price(
```
LefterisJP, line 111: Why do you need to check the global DB for the given oracle? This makes no sense.

So this is called also in the for loop for each oracle? So you would check per oracle if a price is cached? That's useless.

There should be only a single check like this. And that shoudl be at the very start of the loop or if it's not called in the oracle loop at the start. But it should not matter from which oracle the price is saved in the DB, should it?

rotkehlchen/externalapis/cryptocompare.py
```diff
@@ -795,21 +794,9 @@ def query_historical_price(
...
-            source=HistoricalPriceOracle.CRYPTOCOMPARE,
-        )
-        if price_cache_entry and price_cache_entry.price != ZERO_PRICE:
-            log.debug('Got historical price from cryptocompare', from_asset=from_asset, to_asset=to_asset, timestamp=timestamp, price=price_cache_entry.price)  # noqa: E501
-            return price_cache_entry.price
 
-        # else
         log.debug(
-            f"Couldn't find historical price from {from_asset} to "
+            f'No cached historical price from {from_asset} to '
```
LefterisJP, line 799: you don't need this log. It's already on the calling function no?

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -246,4 +246,20 @@ def _remove_old_accounting_rules(write_cursor: 'DBCursor') -> None:
...
+    def _remove_manual_historical_price_oracle(write_cursor: 'DBCursor') -> None:
+        write_cursor.execute(
+            'SELECT value FROM settings WHERE name=?',
+            ('historical_price_oracles',),
+        )
+        result = write_cursor.fetchone()
+        if result is None:
+            return
+
+        oracles = [x for x in json.loads(result[0]) if x != 'manual']
```
LefterisJP, line 259: you need to handle jsondecode error here.

docs/api.rst
```diff
@@ -5806,7 +5806,7 @@ Export PnL report debug data
                     "uniswapv2",
                     "uniswapv3",
                 ],
-                "historical_price_oracles": ["manual", "cryptocompare", "coingecko"],
+                "historical_price_oracles": ["cryptocompare", "coingecko"],
```
LefterisJP, line 5809: add some more of the new ones in the example. Such as defillama, alchemy, uniswapv2 and uniswapv3

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2869,6 +2869,10 @@ def _check_tokens_existence(user_db_cursor: 'DBCursor', expect_removed: bool = F
                 ('bybit_history_events_cb1', 1641386301, 1641386400),
             ],
         )
+        write_cursor.execute(
```
LefterisJP, line 2872: do that directly in the test DB instead of here. And then in the part of the test before the upgrade, check manual is there. The part after the upgrade can stay the same.

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -246,4 +247,24 @@ def _remove_old_accounting_rules(write_cursor: 'DBCursor') -> None:
...
+            ('historical_price_oracles',),
+        )
+        result = write_cursor.fetchone()
+        if result is None:
+            return
+
+        try:
+            oracles = [x for x in json.loads(result[0]) if x != 'manual']
+        except json.JSONDecodeError:
+            oracles = [x.serialize() for x in DEFAULT_HISTORICAL_PRICE_ORACLES_ORDER]
```
LefterisJP, line 263: but you should also log an error here. So that by looking at the logs one can know the this happened.

rotkehlchen/db/upgrades/v46_v47.py
```diff
@@ -246,4 +247,24 @@ def _remove_old_accounting_rules(write_cursor: 'DBCursor') -> None:
...
         )
 
+    @progress_step(description='Remove manual historical oracle')
+    def _remove_manual_historical_price_oracle(write_cursor: 'DBCursor') -> None:
+        write_cursor.execute(
+            'SELECT value FROM settings WHERE name=?',
+            ('historical_price_oracles',),
+        )
+        result = write_cursor.fetchone()
+        if result is None:
```
LefterisJP, line 257:
```suggestion
        if (result := write_cursor.fetchone()) is None:
```

LefterisJP (approved):
LGTM

PR #9409, opened 2025-02-11
Fix multiple events with same timestamp in asset graph

nicholasyoder:


rotkehlchen/balances/historical.py
```diff
@@ -148,13 +148,17 @@ def get_assets_amounts(
...
 
+            combined_amounts = ZERO
             for asset in assets:
-                event_ts = (
-                    event.timestamp
-                    if isinstance(event, Trade)
-                    else ts_ms_to_sec(event.timestamp)
-                )
-                amounts[event_ts] += current_balances[asset]
+                combined_amounts += current_balances[asset]
```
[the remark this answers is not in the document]
nicholasyoder: Yes, that makes it simpler. Changed it.

rotkehlchen/balances/historical.py
```diff
@@ -148,13 +148,13 @@ def get_assets_amounts(
...
-                    if isinstance(event, Trade)
-                    else ts_ms_to_sec(event.timestamp)
-                )
-                amounts[event_ts] += current_balances[asset]
+            event_ts = (
+                event.timestamp
+                if isinstance(event, Trade)
+                else ts_ms_to_sec(event.timestamp)
+            )
+            amounts[event_ts] = sum((current_balances[asset] for asset in assets), ZERO)
```
LefterisJP, line 156: Why do you need the starting value? Doesn't sum start from ZERO? Is the extra prentheses needed also?


Also took me a long time to see it but the difference between this and the previous code is that for same `event_ts`  you add it multiple times? Can you maybe add a comment here? As this is tricky

LefterisJP (approved):
LGTM

rotkehlchen/balances/historical.py
nicholasyoder: Starting value of ZERO was for typing, since it needs to be FVal - but that's not very self obvious, so I've changed it to simply wrap `sum` with `FVal()`. Also now that there's only one argument to `sum` the extra parentheses can be skipped as well.

Also added a comment to try and make it clearer what's happening.

PR #9413, opened 2025-02-12
Fix balance query task retries

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=97572435

### DIFF
--- a/rotkehlchen/db/dbhandler.py
+++ b/rotkehlchen/db/dbhandler.py
@@ -3338,16 +3338,23 @@ class DBHandler:
             )
             return {Location.deserialize_from_db(loc[0]) for loc in cursor}
 
-    def should_save_balances(self, cursor: 'DBCursor') -> bool:
+    def should_save_balances(
+            self,
+            cursor: 'DBCursor',
+            last_query_ts: Timestamp | None = None,
+    ) -> bool:
         """
-        Returns whether or not we can save data to the database depending on
-        the balance data saving frequency setting
+        Returns whether we should save a balance snapshot depending on whether the last snapshot
+        and last query timestamps are older than the period defined by the save frequency setting.
         """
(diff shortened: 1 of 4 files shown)


rotkehlchen/db/dbhandler.py
```diff
@@ -3338,16 +3338,23 @@ def get_associated_locations(self) -> set[Location]:
...
-    def should_save_balances(self, cursor: 'DBCursor') -> bool:
+    def should_save_balances(
+            self,
+            cursor: 'DBCursor',
+            last_query_ts: Timestamp | None = None,
+    ) -> bool:
         """
-        Returns whether or not we can save data to the database depending on
-        the balance data saving frequency setting
+        Returns whether we should save a balance snapshot depending on whether the last snapshot
```
LefterisJP, line 3347: this approach seems to only take into account the query of all balances during the periodic task right?

But it's also done directly via the api and can probably de done elsewhere no? The function is `rotkehlchen::query_balances` so shouldn't the last_query_ts be set in there?

LefterisJP (approved):
LGTM.

rotkehlchen/db/dbhandler.py
nicholasyoder: > The function is rotkehlchen::query_balances so shouldn't the last_query_ts be set in there?

Yes, `task_manager.last_balance_query_ts` **is** being set from `rotkehlchen::query_balances`: https://github.com/nicholasyoder/rotki/blob/c5fdb988eae7b73612d252089c2cd16e1537d882/rotkehlchen/rotkehlchen.py#L1002
We're only checking this timestamp before running the periodic task though, as `rotkehlchen::query_balances` should be able to be run via the api, etc regardless of how recent `last_balance_query_ts` is.

PR #9414, opened 2025-02-12
exclude exited validators from balances query

prettyirrelevant:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=97572198

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` rotki will now skip balance queries for exited validators, reducing API rate limits and improving performance.
 * :feature:`-` Support for compound v3 protocol on OP mainnet has now been added. Also USDS in mainnet and AERO in BASE for Compound v3 should be seen properly now. Finally compound interactions that wrap ETH to WETH through the Compound bulker and vice versa should be seen properly now.
 * :feature:`1379` For the asset amount and value graph, users can now choose to use historical events and prices as the source, instead of snapshots.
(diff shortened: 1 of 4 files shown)


rotkehlchen/chain/ethereum/modules/eth2/beacon.py
```diff
@@ -113,7 +126,10 @@ def query_chunked(
         chunks = calculate_query_chunks(indices_or_pubkeys, chunk_size=chunk_size)
         data = []
         for chunk in chunks:
-            data.extend(self.query(querystr.format(chunk=','.join(str(x) for x in chunk))))
+            data.extend(self.query(
+                endpoint=endpoint,
+                data={'ids': [str(x) for x in chunk]},
+            ))
```
yabirgb, line 132: checking where it is used `calculate_query_chunks` seems to always be used as string so:

1. Confirm that we always do `str(x) for x in calculate_query_chunks`
2. make `calculate_query_chunks` return a list of str directly. In fact inside `calculate_query_chunks` we use `list()` so we are iterating always twice

rotkehlchen/externalapis/beaconchain/service.py
```diff
@@ -238,7 +244,7 @@ def get_validator_data(
         """Returns data for the given validators
 
         Essentially calls:
-        https://beaconcha.in/api/v1/docs/index.html#/Validator/get_api_v1_validator__indexOrPubkey_
+        https://beaconcha.in/api/v1/docs/index.html#/Validator/post_api_v1_validator
```
yabirgb, line 247: please add a small test to confirm that exited validators aren't queried anymore and also as a regression test

PR #9416, opened 2025-02-12
fix: fix incorrect usage of datetime.UTC

mdqst:
I noticed that in the code, there was an incorrect reference to `datetime.UTC` which doesn’t exist.
Instead, I’ve updated it to `datetime.timezone.utc`, which is the correct usage.

### DIFF
--- a/tools/profiling/trace.py
+++ b/tools/profiling/trace.py
@@ -22,7 +22,7 @@ class TraceProfiler:
         self.datadir = datadir
         self.profiling = True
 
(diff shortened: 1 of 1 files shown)


LefterisJP:
Hey thanks for the interedt in rotki!

`datetime.UTC` is an alias for that. 

source: https://docs.python.org/3/library/datetime.html#datetime.UTC

And it's used in more places in our code than this. It works fine so will leave it as is.

PR #9428, opened 2025-02-13
Micro optimizations

yabirgb:
- Avoid looping twice when processing trades
- Avoid calling deserialize multiple times in a for loop

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -931,34 +931,35 @@ class RestAPI:
         except RemoteError as e:
             return {'result': None, 'message': str(e), 'status_code': HTTPStatus.BAD_GATEWAY}
 
-        trades_result = []
+        with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
+            mapping = self.rotkehlchen.data.db.get_ignored_action_ids(cursor, ActionType.TRADE)
+            total_entries = self.rotkehlchen.data.db.get_entries_count(
+                cursor=cursor,
+                entries_table='trades',
+            )
+
+        ignored_ids = mapping.get(ActionType.TRADE, set())
+        entries_result = []
         for trade in trades:
             serialized_trade = self.trade_schema.dump(trade)
             serialized_trade['trade_id'] = trade.identifier
-            trades_result.append(serialized_trade)
+            if (
+                (is_trade_ignored := trade.identifier in ignored_ids) and
+                include_ignored_trades is False
+            ):
+                continue
 
-        with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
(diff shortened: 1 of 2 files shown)


rotkehlchen/api/rest.py
```diff
@@ -935,34 +935,35 @@ def get_trades(
         except RemoteError as e:
             return {'result': None, 'message': str(e), 'status_code': HTTPStatus.BAD_GATEWAY}
 
-        trades_result = []
+        with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
```
LefterisJP, line 938: what is the optimization here?  To just close the cursor early?

rotkehlchen/db/dbhandler.py
```diff
@@ -1009,7 +1009,14 @@ def get_ignored_action_ids(
         cursor.execute(query, tuples)
         mapping = defaultdict(set)
         for entry in cursor:
-            mapping[ActionType.deserialize_from_db(entry[0])].add(entry[1])
+            mapping[entry[0]].add(entry[1])
+
+        for map_key in list(mapping.keys()):
+            # To avoid doing a deserialization for each trade we use the string key
+            # and deserialize it at the end. This `pointing + deleting` doesn't allocate
+            # new memory so doesn't affect to the memory usage.
```
LefterisJP, line 1017: Some comment suggestion and below questions.

```suggestion
            # To avoid doing a deserialization for each trade we use the string key
            # of the ActionType and deserialize it at the end. This `pointing + deleting` 
            # doesn't allocate new memory so doesn't affect memory usage.
```

1. What do you mean by "This pointing + deleting doesn't allocate new memory so doesn't affect memory usage." What is pointing and deleting?
2. The optimization here is just not doing this function as many times as there is trades but just do it 1 or 2 times?
```python
    def deserialize_from_db(cls: type[D], value: str) -> D:
        """May raise a DeserializationError if something is wrong with the DB data"""
        if not isinstance(value, str):
            raise DeserializationError(
                f'Failed to deserialize {cls.__name__} DB value from non string value: {value}',
            )

        number = ord(value)
        if number < 65 or number > list(cls)[-1].value + 64:
            raise DeserializationError(f'Failed to deserialize {cls.__name__} DB value {value}')
        return cls(number - 64)
```

yabirgb: 1. What I mean is that reasigning the key of the dict doesn't copy the elements, just adds a pointer.
2. Yes, I saw that we were calling several times the `deserialize_from_db` in the for loop but the set of values is not that big so we can just do it once at the end. Will adjust the docstring
3.

rotkehlchen/api/rest.py
yabirgb: no, here we were looping twice the trades. If you notice there are two for loops in the current implementation, one to convert to dict and other to exclude/add the ignore flag

PR #9429, opened 2025-02-13
Only run eth2 tasks when active validators are present

nicholasyoder:


rotkehlchen/tasks/manager.py
```diff
@@ -574,15 +575,18 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
 
     def _maybe_query_produced_blocks(self) -> Optional[list[gevent.Greenlet]]:
         """Schedules the blocks production query if enough time has passed"""
+        eth2 = self.chains_aggregator.get_module('eth2')
+        if eth2 is None:
```
LefterisJP, line 579:
```suggestion
        if (eth2 := self.chains_aggregator.get_module('eth2')) is None:
```

rotkehlchen/tests/unit/test_tasks_manager.py
```diff
@@ -609,18 +626,49 @@ def test_maybe_query_ethereum_withdrawals(task_manager, ethereum_accounts):
...
+                )
             task_manager.schedule()
             gevent.sleep(0)  # context switch for execution of task
             assert query_mock.call_count == expected_call_count, msg
 
 
+@pytest.mark.parametrize('ethereum_accounts', [['0x2B888954421b424C5D3D9Ce9bB67c9bD47537d12']])
+@pytest.mark.parametrize('ethereum_modules', [['eth2']])
+@pytest.mark.parametrize('max_tasks_num', [5])
+def test_maybe_query_produced_blocks(task_manager, ethereum_accounts):
```
LefterisJP, line 643: don't we have other block production query tests? I am pretty sure I have written many of them.

LefterisJP (commented):
Okay so I have some concerns which I realized while looking at this PR.

Assuming you start tracking withdrawal addresses and validators from the time they are created then all should be good. You should have all events and the two periodic tasks should only run for events that may exist and not bother with exited validators.

---

But let's look at the scenario where you have 4 validators. Two of which are exited already. And you just download rotki and want to try it. You create you rotki account **after exit** of those validators. So the first condition is exited and want to track all of them and all their history. In that case you want:

1. All events to first be pulled both for block productions and for withdrawals for those validators.
2. From here and on don't bother with the exited validators. But only after all events until exit have been pulled.


I don't think (1) will happen with this change. Will it? Do you understand my concern?

rotkehlchen/tests/unit/test_tasks_manager.py
nicholasyoder: Yes, there's at least this one: https://github.com/rotki/rotki/blob/develop/rotkehlchen/tests/integration/test_eth2.py#L152

But none that tested the periodic task as far as I can tell.

rotkehlchen/chain/ethereum/modules/eth2/eth2.py
```diff
@@ -389,14 +389,25 @@ def query_services_for_validator_withdrawals(
...
-                address=address,
-            )
+            # Get the last query timestamp for this address and a count of this address's
+            # validators that are either active, exited but never queried, or exited and
+            # queried but exited_timestamp is after last query timestamp.
+            key_name = DBCacheDynamic.WITHDRAWALS_TS.get_db_key(address=address)
+            last_query, validator_count = cursor.execute(
+                'SELECT kv.value, COUNT(*) FROM eth2_validators ev '
+                'LEFT JOIN key_value_cache kv ON kv.name=? WHERE ev.withdrawal_address=? AND '
+                '(ev.exited_timestamp IS NULL OR kv.name IS NULL OR ev.exited_timestamp > kv.value)',  # noqa: E501
```
LefterisJP, line 399: what is this for `OR kv.name IS NULL`? It can never be null according to the schema. Does this mean it does not exist?

rotkehlchen/externalapis/beaconchain/service.py
```diff
@@ -250,14 +250,28 @@ def get_validator_data(
...
-            indices_or_pubkeys: list[int | Eth2PubKey],
+            indices: list[int],
     ) -> None:
         with self.produced_blocks_lock:
-            return self._get_and_store_produced_blocks(indices_or_pubkeys)
+            # Get indices for validators that are either active, exited but never queried, or
+            # exited and queried but exited timestamp is after last query timestamp.
+            with self.db.conn.read_ctx() as cursor:
+                placeholders = ', '.join(['?'] * len(indices))
+                key_name = DBCacheDynamic.LAST_PRODUCED_BLOCKS_QUERY_TS.value[0].replace('{index}', '')  # noqa: E501
```
LefterisJP, line 260: isn't this a constant value? Essentially `last_produced_blocks_query_ts_` ? Why not just do something like `DBCacheDynamic.LAST_PRODUCED_BLOCKS_QUERY_TS[:30]` and then you get `last_produced_blocks_query_ts_` without any other extra logic?

rotkehlchen/externalapis/beaconchain/service.py
```diff
@@ -338,11 +352,13 @@ def _get_and_store_produced_blocks(
...
-                    name=DBCacheStatic.LAST_PRODUCED_BLOCKS_QUERY_TS,
-                    value=ts_now(),
-                )
+                for index in indices:
+                    self.db.set_dynamic_cache(
+                        write_cursor=write_cursor,
+                        name=DBCacheDynamic.LAST_PRODUCED_BLOCKS_QUERY_TS,
+                        value=ts_now(),
+                        index=index,
+                    )
```
LefterisJP, line 361:
```suggestion
                now = ts_now()
                for index in indices:
                    self.db.set_dynamic_cache(
                        write_cursor=write_cursor,
                        name=DBCacheDynamic.LAST_PRODUCED_BLOCKS_QUERY_TS,
                        value=now,
                        index=index,
                    )
```

rotkehlchen/tasks/manager.py
```diff
@@ -550,59 +550,58 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
...
+
+        if self.chains_aggregator.beaconchain.produced_blocks_lock.locked():
+            return None  # task is already running, either api or periodic
+
+        # Get indices for validators that haven't been queried yet,
+        # or were last queried long enough ago that they need queried again
         with self.database.conn.read_ctx() as cursor:
-            result = self.database.get_static_cache(
-                cursor=cursor, name=DBCacheStatic.LAST_PRODUCED_BLOCKS_QUERY_TS,
+            key_name = DBCacheDynamic.LAST_PRODUCED_BLOCKS_QUERY_TS.value[0].replace('{index}', '')
```
LefterisJP, line 562: so this is very similar to what you have in beaconchain/service.py `get_and_store_produce_blocks`. Can you abstract this out into a function?

rotkehlchen/tasks/manager.py
```diff
@@ -550,59 +550,58 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
...
     def _maybe_query_produced_blocks(self) -> Optional[list[gevent.Greenlet]]:
         """Schedules the blocks production query if enough time has passed"""
+        if self.chains_aggregator.get_module('eth2') is None:
+            return None
+
+        if self.chains_aggregator.beaconchain.produced_blocks_lock.locked():
+            return None  # task is already running, either api or periodic
+
+        # Get indices for validators that haven't been queried yet,
+        # or were last queried long enough ago that they need queried again
```
LefterisJP, line 560:
```suggestion
        # Get indices for validators that haven't been queried yet, or were last
        # queried long enough ago that they need to be queried again
```

rotkehlchen/tasks/manager.py
```diff
@@ -550,59 +550,58 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
...
-        should_query = False
-        if len(end_timestamps) != len(addresses):
-            should_query = True
-        else:
-            for entry in end_timestamps:
-                if now - int(entry[0]) >= HOUR_IN_SECONDS * 3:
-                    should_query = True
-                    break
+            # Get user addresses that have validators that may need queried
+            key_name = DBCacheDynamic.WITHDRAWALS_TS.value[0].replace('{address}', '')
```
LefterisJP, line 595: similar here. Both the replace logic should be changed as I mentioned above and perhaps this abstracted if used anywhere else.

rotkehlchen/tasks/manager.py
```diff
@@ -550,59 +550,58 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
...
+            cursor.execute(
+                'SELECT DISTINCT ev.withdrawal_address FROM eth2_validators ev '
+                f"LEFT JOIN key_value_cache kv ON kv.name = '{key_name}' || ev.withdrawal_address "
+                'WHERE kv.value <= ? OR kv.name IS NULL',
+                (ts_now() - HOUR_IN_SECONDS * 3,),
+            )
+            addresses = [row[0] for row in cursor]
 
-        if not should_query:
+        if len(addresses) == 0:
```
LefterisJP, line 604: can combine this with line 602 using `:=` right?

rotkehlchen/tests/integration/test_eth2.py
```diff
@@ -386,9 +395,16 @@ def test_withdrawals_detect_exit(eth2: 'Eth2', database):
 @pytest.mark.freeze_time('2023-11-20 07:07:55 GMT')
 def test_query_no_withdrawals(
         eth2: 'Eth2',
+        database: 'DBHandler',
         ethereum_accounts: list[ChecksumEvmAddress],
 ) -> None:
     """Test that if an address has no withdrawals we correctly handle it"""
+    with database.user_write() as write_cursor:
+        # Add a validator associated with the address so that withdrawals get queried.
```
LefterisJP, line 403: Hmm this makes things tricky I just realized. How hard would it be to modify the logic to also query withdrawals and remember the exited or not if validator is not trcked but just withdrawal address is. Do not perform this change yet in this PR.

I am just thinking what it would look like. As for example it's possible to set as the withdrawal address of a validator an address of someone else and that someone else just getting withdrawals. For whatever reason. I am not sure how common that would be but it's in the realm of possibility.

rotkehlchen/tasks/manager.py
nicholasyoder: Changed the replace logic, but this isn't really used elsewhere so not abstracting it. The query in `query_single_address_withdrawals` while slightly similar works quite a bit different than this query, and doesn't really make sense to try to combine with this.

rotkehlchen/chain/ethereum/modules/eth2/eth2.py
nicholasyoder: yes, its doing a `LEFT JOIN` here, so if there's nothing in `key_value_cache` for this key then `kv.name` will be null, in which case it means we haven't queried for this address yet.

nicholasyoder:
Added cache cleaning when deleting validators / addresses as mentioned in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1340256066595979264)

rotkehlchen/externalapis/beaconchain/service.py
```diff
@@ -262,16 +263,49 @@ def get_validator_data(
             endpoint=None,
         )
 
+    def get_validators_to_query_for_blocks(self, indices: list[int] | None = None) -> list[int]:
+        """Get indices for validators that need queried for produced blocks.
```
LefterisJP, line 267:
```suggestion
        """Get indices for validators that need to be queried for produced blocks.
```

rotkehlchen/tasks/manager.py
```diff
@@ -550,60 +550,43 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
...
-                return None
-
-            cursor.execute('SELECT validator_index FROM eth2_validators')
-            indices = [row[0] for row in cursor]
-
-        if len(indices) == 0:
+        if (
+            self.chains_aggregator.get_module('eth2') is None or
+            self.chains_aggregator.beaconchain.produced_blocks_lock.locked() or
+            len(indices := self.chains_aggregator.beaconchain.get_validators_to_query_for_blocks()) == 0  # noqa: E501
```
LefterisJP, line 556: Why are you doing it like this here? I don't understand. Why did you add the option of indices being zero and use it in the periodic task. That's exactly where the problem lies. You will also query exited validators here.

Why?

rotkehlchen/externalapis/beaconchain/service.py
```diff
@@ -262,16 +263,49 @@ def get_validator_data(
...
+        If `indices` is None, any validator that hasn't been queried in the past day
```
LefterisJP, line 268: I don't understand why you put this option here. See my other question.

rotkehlchen/tasks/manager.py
```diff
@@ -550,60 +550,43 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
...
-            should_query = True
-        else:
-            for entry in end_timestamps:
-                if now - int(entry[0]) >= HOUR_IN_SECONDS * 3:
-                    should_query = True
-                    break
-
-        if not should_query:
-            return None
+            # Get user addresses that have validators that may need queried
```
LefterisJP, line 580:
```suggestion
            # Get user addresses that have validators that may need to be queried
```

rotkehlchen/tasks/manager.py
nicholasyoder: So, `get_validators_to_query_for_blocks` was actually getting called twice in this logic - The first time here, to just get validators needing to be queried based on time only, and the second time in `get_and_store_produced_blocks` to filter out exited validators. This is a bit confusing and inefficient though. 

I've combined those two queries and removed the call to `get_validators_to_query_for_blocks` from `get_and_store_produced_blocks`. Now its called only once here in the periodic task logic. It must also now be called in `rest.py::query_online_events`, since its not called from within `get_and_store_produced_blocks` itself anymore.

rotkehlchen/externalapis/beaconchain/service.py
nicholasyoder: Reworked this as mentioned in response to other comment. Now the option here is `include_recently_queried` to allow querying via the api no matter how recently they were previously queried, while still only querying once per day in the periodic task.

rotkehlchen/externalapis/beaconchain/service.py
```diff
@@ -262,16 +263,43 @@ def get_validator_data(
             endpoint=None,
         )
 
+    def get_validators_to_query_for_blocks(
+            self,
+            include_recently_queried: bool = True,
```
LefterisJP, line 268: This is still a confusing approach. You are combining two functions that do a different thing and mixing their implementation while at the same time adding an unecessary if conditional.

Move the main logic of this function into a common `_get_validators_to_query_for_blocks`. And do not pass this argument there. Instead pass the `where` and the `bindings`. This way you save the conditional.

And then make the `get_validators_to_query_for_blocks` which is just like include_recently_queried=True, the main logic of what this should do.

And then the include_recently_queried=False rename to `get_outdated_validators_to_query_for_blocks`. Adjust docstrings accordingly.

rotkehlchen/api/rest.py
```diff
@@ -3878,14 +3878,10 @@ def query_online_events(self, query_type: HistoryEventQueryType) -> dict[str, An
...
                 )
-            else:  # block production
-                with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
-                    cursor.execute('SELECT validator_index FROM eth2_validators')
-                    indices = [row[0] for row in cursor]
-                if len(indices) != 0:
-                    log.debug(f'Querying information for validator indices {indices}')
-                    eth2.beacon_inquirer.beaconchain.get_and_store_produced_blocks(indices)
-                    eth2.combine_block_with_tx_events()
+            elif len(indices := eth2.beacon_inquirer.beaconchain.get_validators_to_query_for_blocks()) != 0:  # block production  # noqa: E501
```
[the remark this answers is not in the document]
nicholasyoder: ruff complains that it thinks this needs to be an elif, but it does make the logic more confusing to have it that way, especially with logging added for when indices are zero, so I've added a `noqa` for the `elif` warning.

rotkehlchen/externalapis/beaconchain/service.py
nicholasyoder: Implemented this.

LefterisJP (approved):
LGTM

PR #9384, opened 2025-02-06
Ignore deposit/withdrawal events in asset graphs

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=96697911
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=96493730

Two commits:
* Ignore deposit/withdrawal events in asset graphs
* Fix double count of coinbase withdrawal fees

PR #9430, opened 2025-02-13
Ignore staking deposit/withdrawals in asset graph

nicholasyoder:
In https://github.com/rotki/rotki/pull/9384 we made the asset graph ignore deposit/deposit asset and withdraw/remove asset - but actually we also need to ignore staking deposit/withdraw as well. This actually means we can just ignore all events with deposit asset or remove asset subtypes.

### DIFF
--- a/rotkehlchen/db/filtering.py
+++ b/rotkehlchen/db/filtering.py
@@ -605,13 +605,12 @@ class DBMultiValueFilter(DBFilter, Generic[T]):
     operator: Literal['IN', 'NOT IN'] = 'IN'
 
     def prepare(self) -> tuple[list[str], Sequence[T]]:
-        suffix = ''  # for NOT IN comparison remember NULL is a special case
-        if self.operator == 'NOT IN':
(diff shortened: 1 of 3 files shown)


prettyirrelevant (approved):

rotkehlchen/db/filtering.py
```diff
@@ -605,13 +605,12 @@ class DBMultiValueFilter(DBFilter, Generic[T]):
...
 
     def prepare(self) -> tuple[list[str], Sequence[T]]:
-        suffix = ''  # for NOT IN comparison remember NULL is a special case
-        if self.operator == 'NOT IN':
-            suffix = f' OR {self.column} IS NULL'
-        return (
-            [f'{self.column} {self.operator} ({", ".join(["?"] * len(self.values))}){suffix}'],
-            self.values,
-        )
+        placeholders = ', '.join(['?'] * len(self.values))
```
LefterisJP, line 608: why the change here?

LefterisJP (approved):
LGTM

PR #9433, opened 2025-02-14
Ensure that anon event logs don't break the decoding

yabirgb:
Post decoding logic on several decoders wasn't handling the case of anon events
and this was making the decode logic fail. This commit:

1. Ensures that indexError is handled so the decoding logic doesn't break stopping the decoding of other transactions
2. Ensures that anon events aren't processed by the decoders to prevent indexing errors since we don't use them

Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=97803880

### DIFF
--- a/rotkehlchen/tests/unit/decoders/test_zerox.py
+++ b/rotkehlchen/tests/unit/decoders/test_zerox.py
@@ -1,11 +1,13 @@
-from typing import Final
+from typing import TYPE_CHECKING, Final
 
 import pytest
 
 from rotkehlchen.assets.asset import Asset
 from rotkehlchen.chain.ethereum.modules.zerox.constants import ZEROX_ROUTER
 from rotkehlchen.chain.evm.decoding.constants import CPT_GAS
+from rotkehlchen.chain.evm.decoding.cowswap.constants import CPT_COWSWAP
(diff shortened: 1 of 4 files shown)


rotkehlchen/tests/unit/decoders/test_zerox.py
```diff
@@ -1189,3 +1195,40 @@ def test_swap_optimism_with_return(optimism_inquirer, optimism_accounts):
         counterparty=CPT_ZEROX,
         address=OP_ZEROX_ROUTER,
     )]
+
+
+@pytest.mark.vcr
+@pytest.mark.parametrize('ethereum_accounts', [['0x43F9A40200310CE535EdF5EA0eb71afB53779BA4']])
+def test_swap_anon_event(ethereum_inquirer: 'EthereumInquirer', ethereum_accounts):
+    """zerox has a special contract for swaps that emits an anonymous event that makes fail
+    the logic of some post decoding rules. This is a regression test for this contract.
```
LefterisJP, line 1204:
```suggestion
    """zerox has a special contract for swaps that emits an anonymous event that causes the
     logic of some post decoding rules to fail. This is a regression test for this contract.
```

LefterisJP (commented):
Failing test: `rotkehlchen/tests/unit/decoders/test_eth2.py::test_deposit_with_anonymous_event`

PR #9438, opened 2025-02-17
Adjust the collateral detection in compound balance query

LefterisJP:
The error in the logs was not really an error. It was going through all compound events and if it found something that's not a token logged the error. But you can deposit ETH as I have seen in my transaction or another test transaction here:
https://arbiscan.io/tx/0x6e3e791c6a8f2261ebe79265ca6cdb01678b1763dc34574e1b9113f8c1941bbc

And that will raise the problem. Ignoring it with continue is fine since ETH is not a collateral to query via the comet contract, so it is not queried via this logic, rather via the underlying token of cWETH.

Fixes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=98186700


yabirgb (approved):
lgtm

PR #9442, opened 2025-02-17
Improvements on binance

yabirgb:
- This PR tracks the last trade id pulled from binance and doesn't query all the trades every time
~- Also reads the list of markets to query from the database instead of using an attribute. The reason is that it was being incorrectly processed by the logic and causing errors when querying some pairs only. The reason of the error is that each letter of the trade pair was being processed as single string instead of considering the whole string making the intersection that we check empty.~

```
[17/02/2025 17:11:57 CET] DEBUG rotkehlchen.exchanges.binance Greenlet-7: ["'['", '\'"\'', "'G'", "'M'", "'X'", "'U'", "'S'", "'D'", "'T'", '\'"\'', "']'"]
[17/02/2025 17:11:57 CET] DEBUG rotkehlchen.exchanges.binance Greenlet-7: ["'ETHBTC'"
```

### DIFF
--- a/rotkehlchen/exchanges/binance.py
+++ b/rotkehlchen/exchanges/binance.py
@@ -19,6 +19,7 @@ from rotkehlchen.assets.asset import AssetWithOracles
 from rotkehlchen.assets.converters import asset_from_binance
 from rotkehlchen.constants import ZERO
 from rotkehlchen.data_import.utils import maybe_set_transaction_extra_data
+from rotkehlchen.db.cache import DBCacheDynamic
 from rotkehlchen.db.constants import BINANCE_MARKETS_KEY
 from rotkehlchen.db.history_events import DBHistoryEvents
 from rotkehlchen.db.ranges import DBQueryRanges
@@ -1125,6 +1126,11 @@ class Binance(ExchangeInterface, ExchangeWithExtras):
             end_ts: Timestamp,
     ) -> tuple[list[Trade], tuple[Timestamp, Timestamp]]:
         """
+        For trades coming from api/myTrades this function won't respect the provided range and
+        will always query all the trades until now. The reason is that binance forces us to query
+        all the pairs and we use the cache at BINANCE_PAIR_LAST_ID to remember which one was the
+        last trade queried on each market speeding up the queries. For fiat payments the time
+        range is respected.
 
         May raise due to api query and unexpected id:
         - RemoteError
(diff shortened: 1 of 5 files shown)


rotkehlchen/exchanges/binance.py
```diff
@@ -209,7 +209,7 @@ def __init__(
             database: 'DBHandler',
             msg_aggregator: MessagesAggregator,
             uri: str = BINANCE_BASE_URL,
-            binance_selected_trade_pairs: list[str] | None = None,
+            **_kwargs: dict[str, Any],
```
LefterisJP, line 212: why are you adding these kwargs if they are not used?

rotkehlchen/exchanges/binance.py
```diff
@@ -1125,22 +1118,40 @@ def query_online_trade_history(
...
+                last_trade_id = self.db.get_dynamic_cache(  # api returns trades with id >= last_trade_id  # noqa: E501
+                    cursor=cursor,
+                    name=DBCacheDynamic.BINANCE_PAIR_LAST_ID,
+                    location=self.location.serialize(),
+                    location_name=self.name,
+                    queried_pair=symbol,
+                ) or 0
+
+            log.debug(
+                f'Will query binance trades on {self.name} for {symbol} after {last_trade_id}',
```
LefterisJP, line 1153:
```suggestion
                f'Will query binance trades on {self.name} for {symbol=} after {last_trade_id=}',
```

rotkehlchen/exchanges/binance.py
```diff
@@ -1205,14 +1218,24 @@ def query_online_trade_history(
...
+            # trades are ordered in asc order by us
+            last_trade_by_pair[trade_pair] = trade
+            trades.append(trade)
 
-            if trade.timestamp > end_ts:
-                break
+        with self.db.conn.write_ctx() as write_cursor:
+            for symbol, trade in last_trade_by_pair.items():
+                if trade.link is None:
+                    log.error(f'Unexpected none link in binance trade {trade}')
```
LefterisJP, line 1228:
```suggestion
                    log.error(f'Missing link field in binance trade {trade}')
```

rotkehlchen/tests/exchanges/test_binance.py
```diff
@@ -849,7 +870,14 @@ def test_binance_query_trade_history_custom_markets(function_scope_binance):
     binance = function_scope_binance
 
     markets = ['ETHBTC', 'BNBBTC', 'BTCUSDC']
-    binance.edit_exchange_extras({BINANCE_MARKETS_KEY: markets})
+    with binance.db.user_write() as write_cursor:
```
LefterisJP, line 873: why did you have to change this?

rotkehlchen/tests/utils/history.py
```diff
@@ -293,34 +293,35 @@ def mock_binance_api_queries(url, params, *args, **kwargs):  # pylint: disable=u
...
-                "price": "0.0063213",
-                "qty": "5.0",
-                "commission": "0.005",
-                "commissionAsset": "RDN",
-                "time": 1512561942000,
-                "isBuyer": false,
-                "isMaker": false,
-                "isBestMatch": true
-                }]"""
+            if params['fromId'] == 0:
```
LefterisJP, line 296: why this extra check here?

yabirgb: because the test queries it twice and the mock was desiged to always return a result. In the case of the test it expects that the second time nothing gets returned and for that we need to check the fromID

yabirgb:
We have CI green again. Merging now since left said the text was okey

PR #9445, opened 2025-02-18
pass correct path when retrying icon downloads

prettyirrelevant:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=97907567


yabirgb (approved):
lgtm

PR #9449, opened 2025-02-18
Accounting adjustments

LefterisJP:
Fixes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=98186898

- [Improve evm event and base event str() function](https://github.com/rotki/rotki/commit/8ed7b51c5efcc297f77d9068c9cd401371904c9f)

- [Remove DepositableAccountantInterface and its surrounding logic](https://github.com/rotki/rotki/commit/0e7e8fa97659101bcd42e671dd3186f8b5316d90) 

That logic was all callback based and was not doing anything regarding
amounts. Just consuming events. Which already happens by the
accounting rules. So it's all unnecessary.

I also removed related tests which seemed useless and just tested the now removed callback logic. With the callback removed there is no danger of consuming events the wrong way if the order of events is somehow not correct during decoding.


rotkehlchen/history/events/structures/evm_event.py
```diff
@@ -259,6 +260,13 @@ def __repr__(self) -> str:
         ]
         return f'EvmEvent({", ".join(fields)})'
 
+    def __str__(self) -> str:
+        return (
+            f'{self.event_type} / {self.event_subtype} EvmEvent in {self.location} with '
+            f'tx_hash={self.tx_hash.hex()} and time '
+            f'{timestamp_to_date(ts_ms_to_sec(self.timestamp))} using {self.asset}'
+        )
```
yabirgb, line 268: will this override the `__repr__` in tests errors? It might make more difficult to compare events in tests

LefterisJP: Why would it? I am only touching the __str__ function here

yabirgb (approved):
LGTM

PR #9450, opened 2025-02-18
improve bybit asset symbol resolver

prettyirrelevant:
Fixes
CleanShot 2025-02-18 at 15 40 34@2x

Bybit has leverage pairs such as:
- BTC3L/USDT
- BTC3S/USDT

So this pr handles resolving them to BTC instead of throwing an error of no mapping found.


yabirgb (approved):
looks okey

PR #9454, opened 2025-02-18
Adjust USD value in history events

yabirgb:
Values in kraken test come from https://github.com/rotki/rotki/commit/afd52ba645f64d44a117e9d8c377b90f3fb03a04#diff-95ba79e91c47b53369f3910b9804237ed1334e01ef4239216d73151c5f5e3896

### DIFF
--- a/docs/websockets.rst
+++ b/docs/websockets.rst
@@ -463,3 +463,47 @@ During historical price queries, progress updates are sent to track the status o
 - ``total``: Total number of time intervals to query
 - ``processed``: Number of intervals that have been processed so far
 - ``subtype``: Set to "historical_price_query_status" to identify historical price query progress
+
+
+Liquity staking status
+=======================
+
+Liquity staking has several products whose price needs to be remotely queried. The progress of these queries can be followed by
+
+::
+
+    {
+        "type": "progress_updates",
+        "data": {
+            "total": 4,
+            "processed": 1,
+            "subtype": "liquity_statking_query"
+        }
+    }
+
+
+- ``total``: Total number of products that the staking logic has to check.
+- ``processed``: Number of products queried.
+- ``subtype``: Set to "liquity_statking_query" to identify liquity staking queries.
+
+
+Async prices queries
+======================
+
+Kraken and Liquity query prices for historical events and they report the status using the following structure sent after processing some number of events:
+
+::
+
+    {
+        "type": "progress_updates",
+        "data": {
+            "total": 220,
+            "processed": 190,
+            "subtype": "stats_price_query"
+        }
+    }
+
+
+- ``total``: Total number of history events that need their price queried.
+- ``processed``: Number of events processed.
+- ``subtype``: Set to "stats_price_query" to identify prices queries for statistics.
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -66,6 +66,7 @@ from rotkehlchen.balances.manual import (
 from rotkehlchen.chain.accounts import SingleBlockchainAccountData
 from rotkehlchen.chain.bitcoin.xpub import XpubManager
 from rotkehlchen.chain.ethereum.airdrops import check_airdrops, fetch_airdrops_metadata
+from rotkehlchen.chain.ethereum.constants import CPT_KRAKEN
 from rotkehlchen.chain.ethereum.defi.protocols import DEFI_PROTOCOLS
 from rotkehlchen.chain.ethereum.modules.convex.convex_cache import (
     query_convex_data,
@@ -4547,17 +4548,18 @@ class RestAPI:
                 query_filter=table_filter,
             )
             value_query_filters, value_bindings = value_filter.prepare(with_pagination=False, with_order=False)  # noqa: E501
-            usd_value, amounts = history_events_db.get_value_stats(
+            asset_amounts_and_value, total_usd = history_events_db.get_amount_and_value_stats(
                 cursor=cursor,
                 query_filters=value_query_filters,
                 bindings=value_bindings,
+                counterparty=CPT_KRAKEN,
             )
             result = {
                 'entries': events,
                 'entries_found': entries_found,
                 'entries_limit': entries_limit,
                 'entries_total': entries_total,
-                'total_usd_value': usd_value,
+                'total_usd_value': total_usd,
                 'assets': history_events_db.get_entries_assets_history_events(
                     cursor=cursor,
                     query_filter=table_filter,
@@ -4567,7 +4569,7 @@ class RestAPI:
                         'asset': entry[0],
                         'amount': entry[1],
                         'usd_value': entry[2],
-                    } for entry in amounts
+                    } for entry in asset_amounts_and_value
                 ],
             }
 
--- a/rotkehlchen/db/history_events.py
+++ b/rotkehlchen/db/history_events.py
@@ -1,11 +1,13 @@
 import copy
 import json
 import logging
+from collections import defaultdict
 from collections.abc import Mapping, Sequence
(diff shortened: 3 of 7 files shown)


rotkehlchen/db/history_events.py
```diff
@@ -752,17 +747,76 @@ def get_value_stats(
...
+            cursor: 'DBCursor',
+            query_filters: str,
+            bindings: list[Any],
+            counterparty: str,
+    ) -> list[tuple[str, FVal, FVal]]:
+        """Returns the sum of the amounts received by asset and the sum of USD value
+        at the time of the events
+        """
+        cursor.execute(f'SELECT COUNT(*) FROM history_events {query_filters}', bindings)
+        total_events = cursor.fetchone()[0]
```
prettyirrelevant, line 768: this can be a single line.

rotkehlchen/db/history_events.py
```diff
@@ -752,17 +747,76 @@ def get_value_stats(
...
+                        'counterparty': counterparty,
+                    },
+                )
+
+            try:
+                asset = row[0]  # existence is guaranteed due the foreign key relation
+                amount = deserialize_fval(
+                    value=row[1],
+                    name='total amount in history events stats',
+                    location='get_amount_stats',
```
prettyirrelevant, line 792: you're not using the variable defined above `query_location`

rotkehlchen/db/history_events.py
```diff
@@ -752,17 +747,76 @@ def get_value_stats(
...
+        """
+        cursor.execute(f'SELECT COUNT(*) FROM history_events {query_filters}', bindings)
+        total_events = cursor.fetchone()[0]
+        query = f'SELECT asset, amount, timestamp FROM history_events {query_filters};'
+        cursor.execute(query, bindings)
+        assets_amounts: dict[str, FVal] = defaultdict(FVal)
+        assets_value: dict[str, FVal] = defaultdict(FVal)
+        query_location: str = 'get_amount_stat'
+        log.debug(f'Will process stats for {total_events} events.')
+        for idx, row in enumerate(cursor):
```
prettyirrelevant, line 775: the cursor.execute can be added here also instead of defining it above.

rotkehlchen/api/rest.py
```diff
@@ -4547,11 +4548,13 @@ def _get_exchange_staking_or_savings_history(
...
             )
             value_query_filters, value_bindings = value_filter.prepare(with_pagination=False, with_order=False)  # noqa: E501
-            usd_value, amounts = history_events_db.get_value_stats(
+            asset_amounts_and_value = history_events_db.get_amount_and_value_stats(
                 cursor=cursor,
                 query_filters=value_query_filters,
                 bindings=value_bindings,
+                counterparty=CPT_KRAKEN,
             )
+            usd_value = sum(x[2] for x in asset_amounts_and_value)
```
prettyirrelevant, line 4557: you can just return the total usd value from the method call. that way you don't need to loop twice.

rotkehlchen/db/history_events.py
```diff
@@ -752,17 +747,82 @@ def get_value_stats(
...
+
+    def get_amount_and_value_stats(
+            self,
+            cursor: 'DBCursor',
+            query_filters: str,
+            bindings: list[Any],
+            counterparty: str,
+    ) -> tuple[list[tuple[str, FVal, FVal]], FVal]:
+        """Returns the sum of the amounts received by asset and the sum of USD value
+        at the time of the events and the total USD value of all the assets queried.
```
LefterisJP, line 765: why USD value and not user's selected currency? Since now we can do the actual currency and it ends up taking the same amount of effort.

LefterisJP (commented):
Left some questions. Also you need to adjust the websockets api rst docs

docs/websockets.rst
```diff
@@ -462,4 +462,48 @@ During historical price queries, progress updates are sent to track the status o
...
 - ``processed``: Number of intervals that have been processed so far
-- ``subtype``: Set to "historical_price_query_status" to identify historical price query progress
\ No newline at end of file
+- ``subtype``: Set to "historical_price_query_status" to identify historical price query progress
+
+
+Liquity staking status
+=======================
+
+Liquity staking has several sources to check that need to query prices remotely. The process can be followed by
```
LefterisJP, line 471:
```suggestion
Liquity staking has several products whose price needs to be remotely queried. The progress of these queries can be followed by
```

docs/websockets.rst
```diff
@@ -462,4 +462,48 @@ During historical price queries, progress updates are sent to track the status o
...
+        "type": "progress_updates",
+        "data": {
+            "total": 4,
+            "processed": 1,
+            "subtype": "liquity_statking_query"
+        }
+    }
+
+
+- ``total``: Total number of sources that the staking logic has to check.
```
LefterisJP, line 485:
```suggestion
- ``total``: Total number of products that the staking logic has to check.
```

docs/websockets.rst
```diff
@@ -462,4 +462,48 @@ During historical price queries, progress updates are sent to track the status o
...
+- ``processed``: Number of sources queried.
```
LefterisJP, line 486:
```suggestion
- ``processed``: Number of products queried.
```

docs/websockets.rst
```diff
@@ -462,4 +462,48 @@ During historical price queries, progress updates are sent to track the status o
...
+- ``subtype``: Set to "liquity_statking_query" to identify liquity staking queries.
+
+
+Async prices queries
+======================
+
+Kraken and Liquity query prices for historical events and they report the status using the following structure sent every 10 events processed:
```
LefterisJP, line 493:
```suggestion
Kraken and Liquity query prices for historical events and they report the status using the following structure sent every 10 processed events:
```

rotkehlchen/db/history_events.py
```diff
@@ -752,17 +748,84 @@ def get_value_stats(
...
+            bindings,
+        ).fetchone()[0]
+
+        assets_amounts: dict[str, FVal] = defaultdict(FVal)
+        assets_value: dict[str, FVal] = defaultdict(FVal)
+        total_usd_value: FVal = ZERO
+        query_location: str = 'get_amount_stats'
+        log.debug(f'Will process {counterparty} stats for {total_events} events')
+        # the max here is to avoid spamming with ws messages accounts with few events
+        send_ws_every_events = max(50, total_events // NUMBER_OF_WS_MSG_PER_PRICE_QUERY)
```
LefterisJP, line 779: The math here does not make sense as I said yesterday. It's always 50.

```python
>>> NUMBER_OF_WS_MSG_PER_PRICE_QUERY = 10
>>> total_events = 200
>>> max(50, total_events // NUMBER_OF_WS_MSG_PER_PRICE_QUERY)
50
>>> total_events = 30
>>> max(50, total_events // NUMBER_OF_WS_MSG_PER_PRICE_QUERY)
50
>>> total_events = 10
>>> max(50, total_events // NUMBER_OF_WS_MSG_PER_PRICE_QUERY)
50
>>> 
```

What you should is is rename `NUMBER_OF_WS_MSG_PER_PRICE_QUERY` to something like `EVENTS_SCALE_FACTOR = 20`

What we want to achieve? Let's say for total_Events up to 50 to always be `5` events. For anything above that scale linearly up to 50 events per ws when total_events is 1000 or more.

So this:

```python
min(50, max(5, 5 + max(0, total_events - 50) // EVENTS_SCALE_FACTOR))
```


Finally do you even need this scale factor as a constant? Is it used anywhere else? If not just hardcode the numbers here and explain in a comment above:
`# Scales the number of events needed to send a WS message. Start from 5 and scale up to 50 linearly if total events >= 1000`

LefterisJP (approved):
LGTM

PR #9459, opened 2025-02-19
feat: table filter pill edit directly

lukicenturi:
https://github.com/user-attachments/assets/716b51f6-9714-4495-8c73-c80e83d4b091

### DIFF
--- a/frontend/app/src/components/table-filter/SuggestedItem.vue
+++ b/frontend/app/src/components/table-filter/SuggestedItem.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { useTemplateRef } from 'vue';
 import { truncateAddress } from '@/utils/truncate';
 import { useSupportedChains } from '@/composables/info/chains';
 import { useAssetInfoRetrieval } from '@/composables/assets/retrieval';
@@ -9,13 +10,28 @@ const props = withDefaults(
   defineProps<{
     suggestion: Suggestion;
     chip?: boolean;
+    editMode?: boolean;
   }>(),
   {
     chip: false,
+    editMode: false,
   },
 );
 
-const { suggestion } = toRefs(props);
+const emit = defineEmits<{
+  'cancel-edit': [];
+  'update:search': [value: string];
+}>();
+
+const { editMode, suggestion } = toRefs(props);
+
+const search = ref<string>('');
+
+const inputWidth = computed(() => Math.min(get(search).length + 2, 25));
+const editInput = useTemplateRef<InstanceType<typeof HTMLInputElement>>('editInput');
+
+const { assetInfo } = useAssetInfoRetrieval();
+const { getChainName } = useSupportedChains();
 
 const isBoolean = computed(() => {
   const item = get(suggestion);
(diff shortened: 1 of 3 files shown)


frontend/app/src/components/table-filter/SuggestedItem.vue
```diff
@@ -9,13 +10,28 @@ const props = withDefaults(
...
     chip: false,
+    editMode: false,
   },
 );
 
-const { suggestion } = toRefs(props);
+const emit = defineEmits<{
+  (e: 'cancel-edit'): void;
+  (e: 'update:search', value: string): void;
+}>();
```
kelsos, line 24:
```suggestion
const emit = defineEmits<{
  'cancel-edit': [];
  'update:search': [value: string];
}>();
```

this is most a nitpick to use the literal format instead, unfortunately it does not have an autofix so we have to slowly do it by hand

frontend/app/src/components/table-filter/SuggestedItem.vue
```diff
@@ -68,6 +81,40 @@ const displayValue = computed(() => {
 
   return value;
 });
+
+function onKeyDown(e: KeyboardEvent) {
+  const excludes = ['Enter', 'Esc', 'ArrowUp', 'ArrowDown'];
```
kelsos, line 86: should we not make this as const?

frontend/app/src/components/table-filter/TableFilter.vue
```diff
@@ -372,6 +410,9 @@ const { t } = useI18n();
           ref="input"
           v-model:search-input="search"
           :model-value="selection"
+          :class="{
+            '[&_input:not([data-cy=edit-input])]:hidden': !!suggestionBeingEdited,
```
kelsos, line 414: I would rather not use the `[data-cy]` property in styling can we instead add a class and use that?

frontend/app/src/components/table-filter/TableFilter.vue
```diff
@@ -65,12 +65,37 @@ const selectedMatcher = computed(() => {
...
+
+  return getSuggestionText(suggestion).text === getSuggestionText(edited).text;
+}
+
 function clickItem(item: Suggestion) {
   if (typeof item.value !== 'boolean') {
-    selectItem(item);
+    set(suggestionBeingEdited, item);
+    set(search, `${item.key}${item.exclude ? '!=' : '='}`);
+    // selectItem(item);
```
kelsos, line 82: can we also remove the comment here?

frontend/app/src/components/table-filter/TableFilter.vue
```diff
@@ -65,12 +65,37 @@ const selectedMatcher = computed(() => {
 
 const usedKeys = computed(() => get(selection).map(entry => entry.key));
 
+const suggestionBeingEdited = ref<Suggestion | null>(null);
```
kelsos, line 68:
```suggestion
const suggestionBeingEdited = ref<Suggestion>();
```

any reason to default to null?

kelsos:
@lukicenturi I left you some small comments there other than that it looks good

lukicenturi:
@kelsos updated again

kelsos (approved):
lgtm thank you :)

PR #9464, opened 2025-02-19
process manual balances with missing assets

yabirgb:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=98191028

Needs frontend work


nicholasyoder (approved):
lgtm

rotkehlchen/balances/manual.py
```diff
@@ -32,24 +37,37 @@ class ManuallyTrackedBalanceWithValue(NamedTuple):
     location: Location
     tags: list[str] | None
     balance_type: BalanceType
+    asset_is_missing: bool = False
```
LefterisJP, line 40: What are you doing here? Why add it here and add it also in `ManuallyTrackedBalanceWithFlags`. That should be an attribute only of `ManuallyTrackedBalanceWithFlags` not of this. And you should just extend the serialize over there, not touch it here. 

Basic inheritance concepts.


Otherwise if you want to add the `asset_is_missing` to the basic class then just do that. You then don't need the other class.

rotkehlchen/balances/manual.py
```diff
@@ -32,24 +37,37 @@ class ManuallyTrackedBalanceWithValue(NamedTuple):
...
 
     def serialize(self) -> dict[str, Any]:
         result = self._asdict()  # pylint: disable=no-member
         del result['value']
+        if self.asset_is_missing is False:
```
LefterisJP, line 45: Is this extra logic needed? It's to save space in the response right? But you also then add an extra if at every serialization

rotkehlchen/db/dbhandler.py
```diff
@@ -1546,18 +1547,26 @@ def get_manually_tracked_balances(
         data = []
         for entry in query:
             tags = deserialize_tags_from_db(entry[4])
+            if (asset_is_missing := not Asset(entry[0]).exists()) is True:
+                self.msg_aggregator.add_warning(
```
LefterisJP, line 1551: Shouldn't the frontend handle this? Why add a warning that will be shown to the frontend here?

rotkehlchen/tests/api/test_manually_tracked_balances.py
```diff
@@ -242,6 +247,25 @@ def test_add_and_query_manually_tracked_balances(
...
+                'manuallytrackedbalancesresource',
+            ),
+        )
+
+    result = assert_proper_sync_response_with_result(response)
+    for entry in result['balances']:
+        if entry['asset'] == A_ETH.identifier:
+            assert result['balances'][0]['asset_is_missing'] is True
+        else:
+            assert 'asset_is_missing' not in result
```
LefterisJP, line 267: Shouldn't this be like that?

```suggestion
            assert 'asset_is_missing' not in result['balances'][0]
```

rotkehlchen/balances/manual.py
```diff
@@ -14,42 +15,52 @@
...
-class ManuallyTrackedBalance:
+class _BaseManualBalance:
     identifier: int
     asset: Asset
     label: str
-    amount: FVal
     location: Location
     tags: list[str] | None
     balance_type: BalanceType
+    asset_is_missing: bool = field(default=False, metadata={'description': 'set to true when the asset points to an asset that is unknown to the db'})  # noqa: E501
```
LefterisJP, line 25: What is this metadata? I tried to find out more info about it but seems to say it's not part of dataclasses but just consumed by third party libraries. Why add this here? Why not just add it as a comment? Does it get used anywhere?

PR #9471, opened 2025-02-20
fix kraken assets test

prettyirrelevant:
this pr updates error handling in Kraken asset resolution test by removing `.symmetric_difference()` check since exchange symbols serve as fallback identifiers

### DIFF
--- a/rotkehlchen/tests/utils/exchanges.py
+++ b/rotkehlchen/tests/utils/exchanges.py
@@ -1442,14 +1442,18 @@ def mock_normal_coinbase_query(url, **kwargs):  # pylint: disable=unused-argumen
     raise AssertionError(f'Unexpected url {url} for test')
 
 
-def get_exchange_asset_symbols(exchange: Location) -> set[str]:
-    """Queries and returns all the asset symbols for a given exchange from the globalDB."""
+def get_exchange_asset_symbols(
+        exchange: Location,
+        query_suffix: Literal[' OR location IS NULL;', ';'] = ' OR location IS NULL;',
+) -> set[str]:
+    """Get all asset symbols for an exchange from the global database.
+
+    Using ';' returns only symbols specific to the exchange, while the default
+    ' OR location IS NULL;' includes exchange-specific and generic symbols.
+    """
     with GlobalDBHandler().conn.read_ctx() as cursor:
-        cursor.execute(
-            'SELECT exchange_symbol FROM location_asset_mappings WHERE location IS ? OR location IS NULL;',  # noqa: E501
-            (exchange.serialize_for_db(),),
-        )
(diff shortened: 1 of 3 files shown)


nicholasyoder (approved):
lgtm

LefterisJP (commented):
I do not understand what this PR is fixing.

Why are you ignoring all symbols from the `location_asset_mappings` now for Kraken?

Right now the code in this PR will no longer check if our DB has any extra garbage symbols in location_asset_mappings for kraken that are not needed.

I don't think this is a good change and I do not understand what it is fixing.

rotkehlchen/tests/utils/exchanges.py
```diff
@@ -1442,14 +1442,19 @@ def mock_normal_coinbase_query(url, **kwargs):  # pylint: disable=unused-argumen
     raise AssertionError(f'Unexpected url {url} for test')
 
 
-def get_exchange_asset_symbols(exchange: Location) -> set[str]:
+def get_exchange_asset_symbols(
+        exchange: Location,
+        include_null_locations: bool = True,
+) -> set[str]:
     """Queries and returns all the asset symbols for a given exchange from the globalDB."""
```
LefterisJP, line 1449: the docstring needs an update

rotkehlchen/tests/utils/exchanges.py
```diff
@@ -1442,14 +1442,19 @@ def mock_normal_coinbase_query(url, **kwargs):  # pylint: disable=unused-argumen
     raise AssertionError(f'Unexpected url {url} for test')
 
 
-def get_exchange_asset_symbols(exchange: Location) -> set[str]:
+def get_exchange_asset_symbols(
+        exchange: Location,
+        include_null_locations: bool = True,
```
LefterisJP, line 1447: again just like many other places I have told you in the past, you should avoid extra if checks like this.

Change the argument to query_suffix with default vaule being `' OR location IS NULL;'` and in the place you want to not include_null_locations call it with `query_suffix=';'`

Simpler and doesn't waste if checks. Adjust the docstring accordingly though

rotkehlchen/tests/utils/kraken.py
```diff
@@ -19,7 +19,21 @@
...
+    'ZKRW',
+    'XNMC',
+    'BSV',
+    'XICN',
+    'SNT',
+    'RNDR',
+    'BNB',
+    'PLA',
+    'WAVES',
+    'DEGEN',
```
LefterisJP, line 34: BNB, DEGEN, EURC ... when were they (1) listed and then (2) delisted from Kraken? When did that happen?

LefterisJP (approved):
LGTM

PR #9472, opened 2025-02-20
Skip simple earn history query in Binance US

nicholasyoder:


rotkehlchen/tests/exchanges/test_binance.py
```diff
@@ -944,17 +944,22 @@ def mock_my_lendings(url, *args, **kwargs):  # pylint: disable=unused-argument
 
     with (
         patch.object(binance.session, 'request', side_effect=mock_my_lendings),
-        binance.db.conn.cursor() as cursor,
+        binance.db.conn.read_ctx() as cursor,
```
nicholasyoder, line 947: Using `binance.db.conn.cursor()` here was having a strange effect: it hid failing asserts in this `with` block - i.e, the with block exited but the test wouldn't fail. I'm not sure why it behaves like this, but finally narrowed it down to that line of code after realizing my asserts were failing without showing an error.

Apparently the `... fetchone()[0] == 4` assert has been failing quietly for some time - it actually returns a count of `2`. I'll need to look into this and figure out if the test was simply written wrong, or if this is something the broke at some point and needs fixed.

nicholasyoder: This test should be fixed and working correctly now. The mocking was being done wrong, which was causing it to only see two events. This was missed whenever it broke initially due to the failing asserts being hidden as mentioned above.

rotkehlchen/exchanges/binance.py
```diff
@@ -701,6 +701,10 @@ def query_lending_interests_history(
         - RemoteError
         - BinancePermissionError
         """
+        if self.location == Location.BINANCEUS:
+            log.debug('Skipping query of simple earn history as Binance US does not support it.')
+            return False
```
yabirgb, line 706: if you return false here it will be interpreted an error and I'm not sure if it will also fail the snapshot. Shouldn't this be true?

rotkehlchen/tests/exchanges/test_binance.py
```diff
@@ -894,9 +894,9 @@ def test_binance_query_lending_interests_history(
     )
     binance = function_scope_binance
 
-    def mock_my_lendings(url, *args, **kwargs):  # pylint: disable=unused-argument
+    def mock_my_lendings(url, params, *args, **kwargs):  # pylint: disable=unused-argument
         if 'simple-earn/flexible/history/rewardsRecord' in url:
-            if 'BONUS' in url:
+            if (request_type := params.get('type')) == 'BONUS':
```
yabirgb, line 899: any reason to change this? Was it wrong because the bonus part wasn't ever in the url?

rotkehlchen/tests/exchanges/test_binance.py
```diff
@@ -944,17 +944,22 @@ def mock_my_lendings(url, *args, **kwargs):  # pylint: disable=unused-argument
...
+            binance.location = location
+            assert binance.query_lending_interests_history(
+                cursor=binance.db.conn.cursor(),
+                start_ts=Timestamp(0),
+                end_ts=Timestamp(API_TIME_INTERVAL_CONSTRAINT_TS),
+            ) is False
+
+            assert cursor.execute(
+                'SELECT COUNT(*) FROM history_events WHERE subtype="reward";',
+            ).fetchone()[0] == count
```
yabirgb, line 962: this should have a location filter. If you change the order in the list I guess this one would fail

rotkehlchen/tests/exchanges/test_binance.py
yabirgb: mmm in theory read_ctx just takes care of closing the cursor. It shouldn't prevent any error 🤔  could it be that before it was counting 2 + 2?

rotkehlchen/exchanges/binance.py
nicholasyoder: Actually returning True is when it gets interpreted as an error. Notice where it handles a remote error below, and also here where this function gets called: https://github.com/rotki/rotki/blob/develop/rotkehlchen/history/manager.py#L197

rotkehlchen/tests/exchanges/test_binance.py
nicholasyoder: I think initially when this test was written BONUS probably was in the url, but at some point that changed - but the test didn't fail due to the way assertions were behaving in the test, so it didn't get updated here.

rotkehlchen/tests/exchanges/test_binance.py
nicholasyoder: Its really weird, but this test:
```
def test_asserts(database):
    with database.conn.cursor():
        print('now you see me')
        assert False
        print('now you dont')

    print('now you see me again')
```
Makes this output, and the test finishes successfully
```
now you see me
now you see me again
```

nicholasyoder: Made a note about this: https://github.com/orgs/rotki/projects/11?pane=issue&itemId=98922841 - will look into it further.

LefterisJP: Isn't it because `assert False` raises an exception, it gets caught by the open cursor, the DB gets reset and then continues further?

PR #9475, opened 2025-02-21
Allow to request icons for collections conditionally

yabirgb:
This PR allows the caller of the API to decide if they want to use the collection icon or the asset icon by passing the boolean argument `use_collection_icon`

### DIFF
--- a/colibri/src/api/icons.rs
+++ b/colibri/src/api/icons.rs
@@ -17,6 +17,8 @@ pub struct AssetIconRequest {
     asset_id: String,
     // hash used to inform the consumer if the file has changed locally or not
     match_header: Option<String>,
+    #[serde(default)]
+    use_collection_icon: bool, // default is false
 }
 
 /// Used when checking the state of an icon locally
@@ -26,6 +28,8 @@ pub struct AssetIconCheck {
     asset_id: String,
     // true if we should delete the local file and pull it again
     force_refresh: Option<bool>,
+    #[serde(default)]
+    use_collection_icon: bool,
 }
 
 /// The handler for the get icon endpoint
@@ -40,6 +44,7 @@ pub async fn get_icon(
     let path = icons::get_asset_path(
         &payload.asset_id,
         state.data_dir.as_path(),
+        payload.use_collection_icon,
         state.globaldb.as_ref(),
     )
     .await;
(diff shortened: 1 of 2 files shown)


colibri/src/api/icons.rs
```diff
@@ -10,13 +10,19 @@ use tokio::fs;
 const HOUR_IN_SECS: u64 = 60 * 60;
 const MAX_ICON_RECHECK_PERIOD: u64 = HOUR_IN_SECS * 12;
 
+fn default_use_collection_icon() -> bool {
```
LefterisJP, line 13: Do you really need this? wtf? A function always returning true? Just use true wherever you call it.

LefterisJP: It can also be the default for the type. Which for booleans is false. So can't you just use default being false

with something like this

```rust
    #[serde(default)]
    use_collection_icon: bool, // will be false
```
    

Another option is to define the default as true for the struct.

```rust
#[derive(Deserialize, Default)]
struct Config {
    #[serde(default)]
    use_collection_icon: bool // still false, needs override
}

// Or implement Default manually
impl Default for Config {
    fn default() -> Self {
        Self {
            use_collection_icon: true
        }
    }
}
```

But this is also verbose like the function you chose. 

So if possible I would just use the default as being false and adjust the API accordingly. Defautl being false also makes sense.

yabirgb: This is what I wasn't sure. If default false was a good default. I guess both are good as how it is used the frontend will need to use both

colibri/src/icons.rs
```diff
@@ -86,7 +86,9 @@ async fn smaldapp_image_query(
 
 // extract the bytes for an image from the provided CDN
 async fn query_image_from_cdn(url: &str) -> Option<Bytes> {
-    smaldapp_image_query(&Client::new(), url, "").await.map(|(bytes, _)| bytes)
+    smaldapp_image_query(&Client::new(), url, "")
```
LefterisJP, line 89: smaldapp?

```suggestion
    smoldapp_image_query(&Client::new(), url, "")
```

LefterisJP (approved):
LGTM. Try to add bit more comments on the tests

PR #9480, opened 2025-02-21
Fix DBCursor exception suppression

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=98922841

As mentioned in the [python docs](https://docs.python.org/3/library/stdtypes.html#contextmanager.__exit__), if the `__exit__` function returns True then any exception that happened within the `with` block is suppressed.

### DIFF
--- a/rotkehlchen/tests/unit/globaldb/test_upgrades.py
+++ b/rotkehlchen/tests/unit/globaldb/test_upgrades.py
@@ -998,19 +998,13 @@ def test_applying_all_upgrade(globaldb: GlobalDBHandler, messages_aggregator):
     """Test globalDB upgrade from v2 to latest"""
     # Check the state before upgrading
     assert globaldb.get_setting_value('version', 0) == 2
-    with globaldb.conn.cursor() as cursor:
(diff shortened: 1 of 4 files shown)


rotkehlchen/tests/unit/globaldb/test_upgrades.py
```diff
@@ -998,19 +998,13 @@ def test_applying_all_upgrade(globaldb: GlobalDBHandler, messages_aggregator):
...
     maybe_upgrade_globaldb(
         connection=globaldb.conn,
         global_dir=globaldb._data_directory / GLOBALDIR_NAME,  # type: ignore
         db_filename=GLOBALDB_NAME,
         msg_aggregator=messages_aggregator,
     )
-
     assert globaldb.get_setting_value('version', 0) == GLOBAL_DB_VERSION
-    with globaldb.conn.cursor() as cursor:
-        assert cursor.execute('SELECT COUNT(*) from assets WHERE identifier=?', ('eip155:/erc20:0x32c6fcC9bC912C4A30cd53D2E606461e44B77AF2',)).fetchone()[0] == 1  # noqa: E501
```
nicholasyoder, line 1013: Removed the assertions here as discussed in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1342572555382292633), since they seem to be testing for something that is no longer present.

LefterisJP (approved):
LGTM

PR #9481, opened 2025-02-21
feat: adjustment for charts

lukicenturi:
https://github.com/rotki/rotki-premium/pull/28

### DIFF
--- a/frontend/app/src/store/tasks/index.ts
+++ b/frontend/app/src/store/tasks/index.ts
@@ -14,6 +14,7 @@ import {
 import { removeKey } from '@/utils/data';
 import { logger } from '@/utils/logging';
 import { useTaskApi } from '@/composables/api/task';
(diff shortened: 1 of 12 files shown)


frontend/app/src/store/tasks/index.ts
```diff
@@ -184,6 +185,14 @@ export const useTaskStore = defineStore('tasks', () => {
     }
   };
 
+  const cancelTaskByTaskType = async (taskTypes: TaskType | TaskType[]): Promise<void> => {
```
kelsos, line 188: I wonder if we could do the change in a way that we can use abort signals in the request chain

lukicenturi: for the historic price it would be more complicated, since we have many layer of request (the cache)

kelsos (approved):
thanks @lukicenturi merging it

PR #9483, opened 2025-02-21
Fix aave v3 receive interest events

nicholasyoder:
Fixes an issue I noticed where aave v3 receive interest events had the wrong asset and subsequently broke the asset graph.


rotkehlchen/tests/unit/decoders/test_aave_v3.py
```diff
@@ -1219,40 +1219,40 @@ def test_aave_v3_withdraw_eth(scroll_inquirer, scroll_accounts) -> None:
             sequence_index=63,
             timestamp=timestamp,
             location=Location.SCROLL,
-            event_type=HistoryEventType.SPEND,
-            event_subtype=HistoryEventSubType.RETURN_WRAPPED,
+            event_type=HistoryEventType.RECEIVE,
```
LefterisJP, line 1222: interest always comes at the end. So this will require changes in the decoder to make it happen

LefterisJP (commented):
I think this is not an error per se. Interest always come after the event. This is how we have the things structured.

The asset graph **should add up** if you use ETH as a collection, which the graph should use by default.

So I don't think this should be modified.

In general asset graphs will not add up if you try to do graphs for assets that are used in collections like ETH/WETH and target only one, since in all of our logic we count it as one and we do not necessarilly have events that turn one to another every time it's needed.

Simplest example here is you go in MORPHO with ETH, and then you withdraw WETH a few weeks later. Everything is off if you count them as different assets since the conversion is not visible anywhere.

nicholasyoder:
I've decided to have the interest amount first as I did initially. Two main reasons:

* It makes more sense this way when the user withdraws a specific amount. For example, in the transaction in [this test](https://github.com/rotki/rotki/blob/develop/rotkehlchen/tests/unit/decoders/test_aave_v3.py#L1263),
the user withdraws `4000` POL, but only `3999.05657246584020169` aPolWMATIC actually leaves their wallet since the `0.94342753415979831` received as interest is included to make the full `4000`. If we were to modify the return and withdrawal amounts so that the interest event could go last, this event would look confusing - the user knows he withdrew exactly `4000`, but here in rotki the withdrawal event will only show that he withdrew `3999.05657246584020169` of course the user should hopefully notice the `0.94342753415979831` received as interest and realize that completes his `4000` withdrawal, but I think this could cause confusion.
* Receiving the interest first more accurately represents what really happened on chain. Not a huge deal, but I like it better when our decoded events and what you see in etherscan are aligned. Also, I expect there are other types of transactions where an IN event happens first, rather than an OUT, and in some cases it likely will not be possible to represent it with the OUT first. So even though we could do that here, its probably best to show it in the order it actually happened.

PR #9500, opened 2025-02-24
fix: fix CLI command declaration typo

mdqst:
Noticed a typo in the CLI command declaration—square brackets `[]` were likely unintentional since `cac()` doesn't use this syntax. Updated it to an empty string `''`, which should be the correct way to define a command without arguments.  

**This fix prevents potential parsing issues or unexpected behavior.**


kelsos (approved):
Hey @mdqst thanks.

PR #9501, opened 2025-02-24
Remove code used only in the deprecated and now removed Defi view

LefterisJP:
Anything that's also used elsewhere should not be touched (hopefully)

---

A lot of tests will probably fail. This PR needs care. I discovered as I suspected that parts of the modules I thought could just be removed were used in the eth protocol balance queries.


rotkehlchen/chain/ethereum/modules/yearn/vaults.py
```diff
@@ -1,340 +1,10 @@
...
-            'yvdusd3CRV': YearnVault(
-                name='curve.fi/dusd Vault',
-                contract=self.ethereum.contracts.contract(string_to_evm_address('0x8e6741b456a074F0Bc45B8b82A755d4aF7E965dF')),
-                token=A_YV1_DUSD_3CRV.resolve_to_evm_token(),
-            ),
-            'yva3CRV': YearnVault(
-                name='curve.fi/aave Vault',
-                contract=self.ethereum.contracts.contract(string_to_evm_address('0x03403154afc09Ce8e44C3B185C82C6aD5f86b9ab')),
-                token=A_YV1_A3CRV.resolve_to_evm_token(),
-            ),
```
yabirgb, line 240: Some of those constant assets were used only here so they can be removed too

rotkehlchen/chain/ethereum/modules/compound/compound.py
```diff
@@ -1,233 +1,23 @@
...
-    from rotkehlchen.user_messages import MessagesAggregator
 
 ADDRESS_TO_ASSETS = dict[ChecksumEvmAddress, dict[Asset, Balance]]
 
 logger = logging.getLogger(__name__)
 log = RotkehlchenLogsAdapter(logger)
 
 
-class Compound(EthereumModule, CompoundV2, CompoundV3):
+class Compound(EthereumModule):
```
yabirgb, line 15: This module is now empty (?) Should we remove it from the settings in the db upgrade? So remove it completly

rotkehlchen/tests/unit/test_stats.py
```diff
@@ -2,21 +2,12 @@
 
 import pytest
 
-from rotkehlchen.chain.ethereum.modules.compound.compound import Compound
 from rotkehlchen.chain.ethereum.modules.liquity.trove import Liquity
-from rotkehlchen.constants import ONE
-from rotkehlchen.constants.assets import A_COMP, A_CUSDC, A_DAI, A_USDC, A_WBTC
```
yabirgb, line 8: `A_CUSDC` is used only on loopring and tests. Wish we could remove it from the constants :/

rotkehlchen/tests/unit/uniswap/test_calculate_events_balances.py
```diff
@@ -1,130 +0,0 @@
...
-@pytest.mark.parametrize('ethereum_modules', [['uniswap']])
-def test_no_events_no_balances(rotkehlchen_api_server: 'APIServer') -> None:
-    rotki = rotkehlchen_api_server.rest_api.rotkehlchen
-    uniswap = rotki.chains_aggregator.get_module('uniswap')
-    assert uniswap is not None
-    events_balances = uniswap.get_balances_chain(addresses=[TEST_ADDRESS_1])
-    assert events_balances == {}
-
-
-@pytest.mark.vcr(filter_query_parameters=['apikey'])
```
yabirgb, line 32: Also we need to make sure that the cassettess are deleted too. Don't leave the files in the cassette repo. At some point we need to figure out if we have files for tests that no longer exist

LefterisJP: I am removing cassettes too yes. There is a branch there

rotkehlchen/chain/ethereum/modules/yearn/vaults.py
LefterisJP: I plan to do a few rounds of dead code linting after the PR to find stuff like this

PR #9503, opened 2025-02-25
Prioritize manual prices for the same timestamp

yabirgb:
Make sure that if we have multiple prices in the same timestamp we prioritize the manual over the rest


rotkehlchen/globaldb/handler.py
```diff
@@ -1202,22 +1202,22 @@ def get_historical_price(
...
+            'WHERE from_asset=? AND to_asset=? AND timestamp between ? AND ?'
         )
-        querylist = [timestamp, from_asset.identifier, to_asset.identifier, timestamp - max_seconds_distance, timestamp + max_seconds_distance]  # noqa: E501
+        querylist = [from_asset.identifier, to_asset.identifier, timestamp - max_seconds_distance, timestamp + max_seconds_distance]  # noqa: E501
         if source is not None:
             querystr += ' AND source_type=? '
             querylist.append(source.serialize_for_db())
 
+        # prioritize manual price if it exists and choose the timestamp closest to the target
+        querystr += ' ORDER BY CASE WHEN source_type=? THEN 0 ELSE 1 END, ABS(timestamp - ?)'
```
LefterisJP, line 1214: this logic and the query logic is duplicated in this and the get_historical_prices. Can you maybe abstract what you can out?

yabirgb:
The failing test `test_balances_caching_mixup` is a remote error

rotkehlchen/globaldb/handler.py
```diff
@@ -84,6 +84,16 @@
 )
 
 
+def _prioritize_manual_balances_query() -> tuple[str, str]:
+    """Prioritize manual price if it exists and choose the timestamp closest to the target.
+    Return the order statement for the query and the type to prioritize already serialized.
+    """
+    return (
+        ' ORDER BY CASE WHEN source_type=? THEN 0 ELSE 1 END, ABS(timestamp - ?)',
+        HistoricalPriceOracle.MANUAL.serialize_for_db(),
```
LefterisJP, line 93: Can you:

1. Change this to a constant? Since it's always the same? 
2. Make the type of the function also return literal?
3. Write a unit test calling for this function checking against constant being equal to `HistoricalPriceOracle.MANUAL.serialize_for_db()` so that if that ever changes we catch it.

LefterisJP (approved):
LGTM

PR #9507, opened 2025-02-25
fix error in `index.ts`  /  `prices.ts`

Fallengirl:
frontend/app/src/composables/api/assets/index.ts
frontend/app/src/composables/api/assets/prices.ts
`@/services/axios-tranformers` - `@/services/axios-transformers`

### DIFF
--- a/frontend/app/tests/unit/specs/types/user.spec.ts
+++ b/frontend/app/tests/unit/specs/types/user.spec.ts
@@ -10,7 +10,7 @@ import {
 } from '@/types/settings/frontend-settings';
 import { TableColumn } from '@/types/table-column';
 import { DateFormat } from '@/types/date-format';
-import { snakeCaseTransformer } from '@/services/axios-tranformers';
+import { snakeCaseTransformer } from '@/services/axios-transformers';
 import { OtherSettings } from '@/types/user';
 
 describe('types/user', () => {


kelsos:
Hey @Fallengirl 

Thank you for the contribution, but I cannot accept the PR in it's current state as it would break the app.

What this PR does is that it renames a valid import that has a typo to an invalid import that does not have a typo. This is not a fix.

A proper fix would adjust the package name along with the imports in the related files. 
Also I don't see the point of having different commits for each file changed, is this done through some automated platform? 

Also assuming that there is a proper PR to fix the typo, you need to also accept the CLA.

kelsos:
> @kelsos sorry I didn't combine the commits, and thanks for pointing out the issue

Can you please combine all the commit to one?

kelsos:
> strange question, but still can I ask for help? It’s literally 5 minutes, my friend will do 

I am not sure what is considered a strange question here. 

You can check our contribution guide [here](https://docs.rotki.com/contribution-guides/contribute-as-developer.html#committing-rules) of what is expected when contributing to this project.

The issue at hand is a single typo which means that it should be a single commit fixing the typo.

What happened instead is that not only you did one commit per file, making it a total of of 54 commits but you also pushed after each commit, which amounts to spam.

This whole PR should have been a single commit and you should only have pushed after finishing the changes.

We really love to have contributions from external contributors, but this attempt feels like really low effort, and I don't want to encourage such behavior.

I am gonna close this PR for the reasons stated above. 

You are welcome to create a proper PR that follows the contribution guide and I will gladly accept it, but this low effort spam is not acceptable at all.

PR #9508, opened 2025-02-25
feat: wording for chart

lukicenturi:
https://github.com/rotki/rotki-premium/pull/30


kelsos:
Let's wait with merging it because it will break the current deployment of the components

PR #9510, opened 2025-02-26
ci: add concurrency setting to sqldiff workflow

kelsos:
Closes #(issue_number)

## Checklist

### DIFF
--- a/.github/workflows/rotki_sqldiff.yml
+++ b/.github/workflows/rotki_sqldiff.yml
@@ -7,6 +7,10 @@ on:
       - bugfixes
       - master
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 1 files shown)


.github/workflows/rotki_sqldiff.yml
```diff
@@ -7,6 +7,10 @@ on:
       - bugfixes
       - master
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
yabirgb, line 11: github.ref in the case of this PR is `kelsos:add_concurrency_option` right?

kelsos: I might need to add head let me check the action

yabirgb (approved):
If I understood correctly this should work. I feared that if two people had a PR the CI would block but since it is based in the ref it should work

PR #9520, opened 2025-02-26
chore: update chunk strategy and remove deprecated split

kelsos:
Closes #(issue_number)

## Checklist


kelsos:
I did a build and tested it with docker, and it seems to be a bit better than the previous one.
we still have the vanilla-json-editor that is quite big but there is not much we can do there and then there is a big bundle of our code, we would need to figure which part and optimize it later.

lukicenturi (approved):
LGTM!

PR #9540, opened 2025-02-28
ci: update tag to not include the version when building

kelsos:
Closes #(issue_number)

## Checklist

https://docs.docker.com/build/ci/github-actions/multi-platform/


LefterisJP (approved):
Only way to test is in the real run. So just doing a LGTM

PR #9550, opened 2025-03-01
Bitcoin balance queries should work for linux binary again

### LINKED ISSUE #9547: In 1.38 BTC query failed for ripemd160
## Problem Definition

In 1.38 at querying onchain BTC balances we now get an error

`unsupported hash type ripemd160`

This did not happen when testing in develop and is only in the binary, so must be a problem with the way some dependency was packaged?

EDIT: 

Seems like it's a **linux only binary issue** and due to switching to the ubuntu 22.04 runner https://github.com/actions/runner-images/issues/6676

That one includes openssl 3.0.2 which has ripemd as not available. Big discussion as to why (I also think it's silly they did that) here: https://github.com/openssl/openssl/issues/16994

This is fixed in openssl 3.0.7 but that does not come with the ubuntu 22.04 runner by default.

For now ... probably easiest way is for linux to switch to a python implementation of ripemd as it's also not used that often, only when calculating a bitcoin address. But probably should revert it when we can use a different openSSL version.

## Logs

```
[01/03/2025 00:14:33 CET] DEBUG rotkehlchen.api.server Greenlet-0: start rotki api GET /api/1/blockchains/btc/accounts view_args={'blockchain': 'btc'}, query_string=b''
[01/03/2025 00:14:33 CET] CRITICAL rotkehlchen.api.server Greenlet-0: Unhandled exception when processing endpoint request exc_info=True, exception=unsupported hash type ripemd160
[01/03/2025 00:14:33 CET] DEBUG rotkehlchen.api.server Greenlet-0: end rotki api GET /api/1/blockchains/btc/accounts view_args={'blockchain': 'btc'}, query_string=b'', status_code=500, result={'result': None, 'message': 'unsupported hash type ripemd160'}
```

### System Description

Operating system: Archlinux
Rotki version: 1.38

LefterisJP:
Fix #9547

For now simply using a python only implementation by Pieter Wuille from bitcoin's repository:
https://github.com/bitcoin/bitcoin/blob/ad3e9e1f214d739e098c6ebbd300da5df1026a44/test/functional/test_framework/ripemd160.py

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* :bug:`9547` Bitcoin balance query should work again for users of the linux binary.
+
 * :release:`1.38.0 <2025-02-28>`
 * :bug:`-` Fix selected binance trading pairs not being properly loaded when restarting rotki.
 * :bug:`-` rotki will now skip balance queries for exited validators, reducing API rate limits and improving performance.
--- a/rotkehlchen/chain/bitcoin/ripemd160.py
+++ b/rotkehlchen/chain/bitcoin/ripemd160.py
@@ -1,0 +1,113 @@
+# Copyright (c) 2021 Pieter Wuille
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""pure Python RIPEMD160 implementation.
+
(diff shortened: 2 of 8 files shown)


yabirgb (approved):
lgtm

LefterisJP:
okay so the tests passed. Will revert the CI changes so all tests run again.

LefterisJP:
ah this should be against bugfixes

PR #9555, opened 2025-03-03
Fix rate limiting handling for beaconchain

yabirgb:
There were two things wrong:

- the retry logic was raising exception.RetryError for rate limits and 500 errors. Since those are errors that we handle manually I made it to not raise an exception in those cases.
- In the case of beaconchain to prevent requests to their API when we are rate limit I store the next timestamp when we can query and if we are before that moment I don't even try the request.
- The background task that checks for produced blocks won't fire if beaconchain is rate limited

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` Prevents too many error notifications when beaconcha.in rate limits the app for long periods of time.
 * :bug:`-` Uniswap v3 swaps in ethereum using the universal router 2 will now be decoded properly.
 * :bug:`9547` Bitcoin balance query should work again for users of the linux binary.
 
--- a/rotkehlchen/externalapis/beaconchain/service.py
+++ b/rotkehlchen/externalapis/beaconchain/service.py
@@ -38,6 +38,7 @@ from rotkehlchen.utils.misc import (
     from_gwei,
     from_wei,
     set_user_agent,
+    timestamp_to_iso8601,
     ts_now,
     ts_sec_to_ms,
 )
@@ -64,6 +65,10 @@ class BeaconChain(ExternalServiceWithApiKey):
         set_user_agent(self.session)
         self.url = f'{BEACONCHAIN_ROOT_URL}/api/v1/'
         self.produced_blocks_lock = Semaphore()
+        self.ratelimited_until = Timestamp(0)
+
+    def is_rate_limited(self) -> bool:
+        return self.ratelimited_until > ts_now()
 
     def _query(
             self,
@@ -78,6 +83,16 @@ class BeaconChain(ExternalServiceWithApiKey):
         May raise:
         - RemoteError due to problems querying beaconcha.in API
         """
+        if self.is_rate_limited():
+            log.error(
+                f'Beaconcha.in is rate limited until {self.ratelimited_until} when processing '
+                f'{module=} {endpoint=} {encoded_args=} with {data=}',
+            )
+            raise RemoteError(
+                'Beaconcha.in is rate limited until '
+                f'{timestamp_to_iso8601(self.ratelimited_until)}. Check logs for more details',
+            )
+
         if endpoint is None:  # for now only validator data
             query_str = f'{self.url}{module}'
         elif endpoint in ('eth1', 'stats'):
@@ -144,6 +159,7 @@ class BeaconChain(ExternalServiceWithApiKey):
                             f'daily limit: {user_daily_rate_limit}/{daily_rate_limit}, '
                             f'monthly limit: {user_month_rate_limit}/{month_rate_limit}',
                         )
+                        self.ratelimited_until = Timestamp(ts_now() + retry_after_secs)
                         raise RemoteError(msg)
                     # else
                     sleep_seconds = retry_after_secs
--- a/rotkehlchen/tasks/manager.py
+++ b/rotkehlchen/tasks/manager.py
@@ -552,6 +552,7 @@ class TaskManager:
         """Schedules the blocks production query if enough time has passed"""
         if (
             self.chains_aggregator.get_module('eth2') is None or
+            self.chains_aggregator.beaconchain.is_rate_limited() or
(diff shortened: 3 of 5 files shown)


rotkehlchen/externalapis/beaconchain/service.py
```diff
@@ -78,6 +83,16 @@ def _query(
...
+        if self.is_rate_limited():
+            log.error(
+                f'Beaconchain is rate limited until {self.ratelimited_until} when processing '
+                f'{module=} {endpoint=} {encoded_args=} with {data=}',
+            )
+            raise RemoteError(
+                'Beaconchain is rate limited until '
+                f'{timestamp_to_iso8601(self.ratelimited_until)}. Check logs for more details',
+            )
+
```
prettyirrelevant, line 95: i think you're duplicating logs here since `RemoteError` is still handled and the error is logged.

rotkehlchen/tasks/manager.py
```diff
@@ -550,10 +550,13 @@ def _maybe_update_snapshot_balances(self) -> Optional[list[gevent.Greenlet]]:
 
     def _maybe_query_produced_blocks(self) -> Optional[list[gevent.Greenlet]]:
         """Schedules the blocks production query if enough time has passed"""
+        if (beeaconchain := self.chains_aggregator.beaconchain).is_rate_limited():
+            return None
+
```
prettyirrelevant, line 555: this if check can be added to the one below, immediately after `self.chains_aggregator.get_module('eth2') is None or`

prettyirrelevant (commented):
smol comments.

rotkehlchen/externalapis/beaconchain/service.py
yabirgb: The remoteError is more human readable since it can get to the user. It happens in the background task for querying produced blocks for example. The log above is there to give more context if we need it in the logs

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`-` Prevents too many error notifications when beaconchain rate limits the app for long periods of time.
```
```suggestion
* :bug:`-` Prevents too many error notifications when beaconcha.in rate limits the app for long periods of time.
```

rotkehlchen/externalapis/beaconchain/service.py
```diff
@@ -78,6 +83,16 @@ def _query(
         May raise:
         - RemoteError due to problems querying beaconcha.in API
         """
+        if self.is_rate_limited():
+            log.error(
+                f'Beaconchain is rate limited until {self.ratelimited_until} when processing '
+                f'{module=} {endpoint=} {encoded_args=} with {data=}',
+            )
+            raise RemoteError(
+                'Beaconchain is rate limited until '
```
LefterisJP, line 92:
```suggestion
                f'Beaconcha.in is rate limited until {self.ratelimited_until} when processing '
                f'{module=} {endpoint=} {encoded_args=} with {data=}',
            )
            raise RemoteError(
                'Beaconcha.in is rate limited until '
```

rotkehlchen/externalapis/beaconchain/service.py
LefterisJP: But won't this spam the user if the RemotError is consumed all the time and shown in the frontend?

rotkehlchen/tasks/manager.py
```diff
@@ -552,8 +552,9 @@ def _maybe_query_produced_blocks(self) -> Optional[list[gevent.Greenlet]]:
         """Schedules the blocks production query if enough time has passed"""
         if (
             self.chains_aggregator.get_module('eth2') is None or
-            self.chains_aggregator.beaconchain.produced_blocks_lock.locked() or
-            len(indices := self.chains_aggregator.beaconchain.get_outdated_validators_to_query_for_blocks()) == 0  # noqa: E501
+            (beeaconchain := self.chains_aggregator.beaconchain).is_rate_limited() or
```
LefterisJP, line 555: why add another variable here? Just to make the line smaller? Don't do that. Also it's `beaconchain`. Not `beeaconchain`

rotkehlchen/utils/network.py
```diff
@@ -35,6 +35,9 @@ def create_session() -> requests.Session:
         connect=3,
         read=2,
         status=0,
+        # in the case of 500 errors or rate limits don't raise error and return the response.
+        # We handle manually error status codes.
+        raise_on_status=False,
```
LefterisJP, line 40: 1. Why are you making this change? To not raise on 429? Are you 100% sure this does not affect any other part of the code that uses `create_session? This sounds quite dangerous.
2. If you want to configure this properly then add all the required arguments along with an explanation in proper Engilish of what is it you are trying to achieve through all arguments.

The relevant arguments are

```

    :param Collection allowed_methods:
        Set of uppercased HTTP method verbs that we should retry on.

        By default, we only retry on methods which are considered to be
        idempotent (multiple requests with the same parameters end with the
        same state). See :attr:`Retry.DEFAULT_ALLOWED_METHODS`.

        Set to a ``None`` value to retry on any verb.

    :param Collection status_forcelist:
        A set of integer HTTP status codes that we should force a retry on.
        A retry is initiated if the request method is in ``allowed_methods``
        and the response status code is in ``status_forcelist``.

        By default, this is disabled with ``None``.


    :param bool raise_on_redirect: Whether, if the number of redirects is
        exhausted, to raise a MaxRetryError, or to return a response with a
        response code in the 3xx range.

    :param bool raise_on_status: Similar meaning to ``raise_on_redirect``:
        whether we should raise an exception, or return a response,
        if status falls in ``status_forcelist`` range and retries have
        been exhausted.

```

and don't rely on defaults. Defaults can change. For more arguments read the `Retry` docs.

yabirgb: 1. The change is because as it is right now in the code 429 raises `RetryError` from requests instead of returning the response with the status code 429 and our custom logic to read the headers and retry is not used. I've checked and where we use this session we check the status code of the responses for errors so this change shouldn't break anything.

PR #9557, opened 2025-03-03
Fix unknown token error when reading tokens from the logs.

yabirgb:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=100117079

Closes #(issue_number)

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,7 @@ Changelog
 
 * :bug:`-` Prevents too many error notifications when beaconcha.in rate limits the app for long periods of time.
 * :bug:`-` Uniswap v3 swaps in ethereum using the universal router 2 will now be decoded properly.
+* :bug:`-` Batched cowswap trades will no longer raise unknown token error in some rare cases.
 * :bug:`9547` Bitcoin balance query should work again for users of the linux binary.
 
 * :release:`1.38.0 <2025-02-28>`
--- a/rotkehlchen/chain/evm/decoding/cowswap/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/cowswap/decoder.py
@@ -5,6 +5,7 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Final, Literal, Optional, cast
 
 from rotkehlchen.assets.asset import Asset, EvmToken
+from rotkehlchen.assets.utils import TokenEncounterInfo
 from rotkehlchen.chain.ethereum.utils import asset_normalized_value, token_normalized_value
 from rotkehlchen.chain.evm.constants import ETH_SPECIAL_ADDRESS
 from rotkehlchen.chain.evm.decoding.airdrops import match_airdrop_claim
@@ -30,7 +31,7 @@ from rotkehlchen.history.events.structures.types import (
     HistoryEventType,
 )
 from rotkehlchen.logging import RotkehlchenLogsAdapter
-from rotkehlchen.types import ChainID, ChecksumEvmAddress, EvmTokenKind, EvmTransaction
+from rotkehlchen.types import ChainID, ChecksumEvmAddress, EvmTokenKind, EvmTransaction, EVMTxHash
 from rotkehlchen.utils.misc import bytes_to_address
 
 if TYPE_CHECKING:
@@ -132,7 +133,11 @@ class CowswapCommonDecoder(DecoderInterface, abc.ABC):
 
     # --- Aggregator methods ---
 
-    def _find_trades(self, all_logs: list[EvmTxReceiptLog]) -> list[CowswapSwapData]:
+    def _find_trades(
+            self,
+            tx_hash: EVMTxHash,
+            all_logs: list[EvmTxReceiptLog],
+    ) -> list[CowswapSwapData]:
         """
         Finds the emitted Trade events and decodes them.
         Also handles special cases when native asset is swapped.
@@ -140,8 +145,7 @@ class CowswapCommonDecoder(DecoderInterface, abc.ABC):
         Returns a list of swap data. Each entry in the list contains basic information about a
         swap made in the transaction.
         """
-        trades = []
-
+        trades, tx_info = [], TokenEncounterInfo(tx_hash=tx_hash, description='CowSwap trade')
         for tx_log in all_logs:
             if tx_log.topics[0] != TRADE_SIGNATURE:
                 continue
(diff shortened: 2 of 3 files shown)


nicholasyoder (commented):
Maybe you should also add a regression test for this? Looks good other than that.

yabirgb:
I wasn't able to reproduce in a test because what happened to the user is an network error and the token wasn't being created so it didn't exist at that point

docs/changelog.rst
```diff
@@ -3,6 +3,7 @@ Changelog
 =========
 
 * :bug:`-` Uniswap v3 swaps in ethereum using the universal router 2 will now be decoded properly.
+* :bug:`-` Fixes an error that might appear for unknown tokens in batched cowswap trades while decoding transactions.
```
LefterisJP, line 6: User facing changelog please.

```suggestion
* :bug:`-`  Batched cowswap trades will no longer raise unknown token error in some rare cases.
```

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -141,7 +146,7 @@ def _find_trades(self, all_logs: list[EvmTxReceiptLog]) -> list[CowswapSwapData]
         swap made in the transaction.
         """
         trades = []
-
+        tx_info = TokenEncounterInfo(tx_hash=tx_hash, description='CowSwap trade')
```
LefterisJP, line 149: Put this in the line above and combine with `trades` in the same initialization line

rotkehlchen/chain/evm/decoding/cowswap/decoder.py
```diff
@@ -159,18 +164,14 @@ def _find_trades(self, all_logs: list[EvmTxReceiptLog]) -> list[CowswapSwapData]
                 owner_address == self.native_asset_flow_address
             ):
                 from_asset = self.native_asset  # native asset swaps are made by eth flow contract in cowswap  # noqa: E501
-            else:  # token should exist since this is the post-processing stage
-                from_asset = EvmToken(evm_address_to_identifier(
+            else:
```
LefterisJP, line 167: Add a comment here, negating the previous comment. Explain why this token may not exist here.
```suggestion
            else:  # need get_or_create_evm_token because token may not exist WHEN?
```

Is the answer batched trades? I don't really get the situation. So an explanation is needed here

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -93,6 +94,41 @@
...
+def decoding_wrapper(
+        msg_aggregator: 'MessagesAggregator | None',
+        transaction_for_logs: EvmTransaction,
+        func: Callable,
+        *args: tuple[Any],
+        **kwargs: Any,
+) -> tuple[Any, bool]:
+    """
+    Wrapper for methods that execute logic from decoders. It handles handles the errors logging
+    them and optionally sending them to the user.
```
LefterisJP, line 106: Note: If you end up removing the if change the docstring here

```suggestion
    Wrapper for methods that execute logic from decoders. It handles all known errors
    by logging them and optionally sending them to the user.
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -93,6 +94,41 @@
 MIN_LOGS_PROCESSED_TO_SLEEP = 1000
 
 
+def decoding_wrapper(
+        msg_aggregator: 'MessagesAggregator | None',
+        transaction_for_logs: EvmTransaction,
```
LefterisJP, line 99: Why are you passing this? The name `transaction_for_logs` implies it's only for logs. But then why the entire class? Even if it's just for logs give it a proper name. Just call it `transaction`.

```suggestion
        transaction_for_logs: EvmTransaction,
```

Why not just pass the transaction hash and chain id? Since this passes a lot more info too. Though this may be just a nitpick.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -93,6 +94,41 @@
...
+        UnknownAsset,
+        WrongAssetType,
+        DeserializationError,
+        IndexError,
+        ValueError,
+        ConversionError,
+    ) as e:
+        log.error(traceback.format_exc())
+        log.error(
+            f'Decoding of transactions {transaction_for_logs} failed due to {e} when calling '
```
LefterisJP, line 123:
```suggestion
            f'Decoding of transaction {transaction_for_logs} failed due to {e} when calling '
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -93,6 +94,41 @@
...
+            f'{func.__name__} with {args} {kwargs}',
```
LefterisJP, line 124:
```suggestion
            f'{func.__name__} with {args=} {kwargs=}',
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -93,6 +94,41 @@
...
+        )
+        if msg_aggregator is not None:
```
LefterisJP, line 126: So the code you wrote does not use this at all. You always set it to `None`. So this also changes current behaviour where it's done 2/4 places you touched.

But here we need to be consistent. We either always send a message to the user or never. So I would get rid of the if or of it alltogether.

Need to think when we need this to tell the users something went wrong. I think the answer is always? The code before your change did it in 2/4 places from what I can see. But perhaps best is to always do it? So get rid of the if. What do you think?

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -93,6 +94,41 @@
 MIN_LOGS_PROCESSED_TO_SLEEP = 1000
 
 
+def decoding_wrapper(
```
LefterisJP, line 97: Perhaprs a better name?

```suggestion
def decode_safely(
```

rotkehlchen/chain/evm/decoding/decoder.py
yabirgb: the reason to not use `transaction` as name is because if it is used as kwarg in the `func` being wrapped it will appear twice and python complains

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -93,6 +94,42 @@
...
+        msg_aggregator: 'MessagesAggregator',
+        tx_hash: EVMTxHash,
+        chain_id: ChainID,
+        func: Callable,
+        *args: tuple[Any],
+        **kwargs: Any,
+) -> tuple[Any, bool]:
+    """
+    Wrapper for methods that execute logic from decoders. It handles all known errors
+    by logging them and optionally sending them to the user.
```
LefterisJP, line 107: I said that if you change the optionality you needed to ofc also change the docstring

```suggestion
    by logging them and sending them to the user.
```

LefterisJP (approved):
LGTM

PR #9561, opened 2025-03-05
style: fix padding in event form

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/components/profitloss/ReportGenerator.vue
+++ b/frontend/app/src/components/profitloss/ReportGenerator.vue
@@ -1,5 +1,4 @@
 <script setup lang="ts">
-import { checkIfDevelopment } from '@shared/utils';
 import { Routes } from '@/router/routes';
 import { convertToTimestamp } from '@/utils/date';
 import RangeSelector from '@/components/helper/date/RangeSelector.vue';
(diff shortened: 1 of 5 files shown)


kelsos (approved):
lgtm

frontend/app/src/components/profitloss/ReportGenerator.vue
```diff
@@ -90,34 +87,22 @@ const accountSettingsRoute = Routes.SETTINGS_ACCOUNTING;
...
-                      {{ t('profit_loss_reports.debug.title') }}
-                    </span>
-                  </RuiButton>
+                <template #prepend>
+                  <RuiIcon name="lu-bug" />
                 </template>
-
                 {{ t('profit_loss_reports.debug.title') }}
-              </RuiTooltip>
+              </RuiButton>
```
kelsos, line 105: The mode with the tooltip was what we did in production and the one with the label is what we do in development, should we just keep the production mode for everything instead?

lukicenturi: Yeah, I think it doesn't make sense to differentiate.

kelsos: With that I agree but what we use now was previously the develop version right? not the production on? why not keep the production mode instead?

PR #9565, opened 2025-03-05
aero/velo locked balances and reminders

### LINKED ISSUE #9267: Support (velo/aero)drome locks
## Abstract

In aerodrome you can create a lock by depositing VELO and setting an amount of time. You receive a NFT that represents your lock. The locks can be extended and I guess that claimed. Locked tokens are able to vote and receive bribes and can claim fees from the pools they voted for.

## Specification
 
- Example of transaction creating lock: https://basescan.org/tx/0xe7b5a3f11dbae89fa04dbf9513d338b098f0db026fd932fd575e1c20baabe29c
- Example of claiming bribes: https://basescan.org/tx/0xc7e900481f02ace531e620327dce80dd7aa14e0c6f1e4624c83a1123fb19f0f8
- Example claiming fees: https://basescan.org/tx/0x114fc97de168f0f054b9ae9157dec6aad090ebe884c0292779b06d42fddf47ca
- Example voting: https://basescan.org/tx/0x7ea170f787a6de81952aa0ea5955757239b906bf825b28b3bb4689fae844d8ed

## Task

- Decode events
- Ideally we would be able to track the locked tokens

prettyirrelevant:
Closes #9267 

- Adds protocol balances for locked AERO/VELO
- Adds reminders for when locks expire


rotkehlchen/chain/evm/decoding/velodrome/balances.py
```diff
@@ -34,6 +52,63 @@ def __init__(
...
         return event.address
+
+    def query_balances(self) -> BalancesSheetType:
+        balances = super().query_balances()
+        if len(addresses_with_deposits := self.addresses_with_activity(event_types=self.deposit_event_types)) == 0:  # noqa: E501
+            return balances
+
+        addresses_to_token_ids = {}
+        for address, events in addresses_with_deposits.items():
+            addresses_to_token_ids[address] = [event.extra_data['token_id'] for event in events if event.extra_data is not None]  # noqa: E501
```
nicholasyoder, line 68:
```suggestion
        addresses_to_token_ids = {
            address: [event.extra_data['token_id'] for event in events if event.extra_data is not None]
            for address, events in addresses_with_deposits.items()
        }
```
Also, will `extra_data['token_id']` always be set here, or could this possibly result in a key error?

rotkehlchen/chain/evm/decoding/velodrome/balances.py
```diff
@@ -34,6 +52,63 @@ def __init__(
...
+                    arguments=[token_ids[idx]],
+                )[0]
+                if balance == 0:
+                    continue
+
+                if (price := Inquirer().find_usd_price(self.protocol_token)) == ZERO_PRICE:
+                    log.error(
+                        f'Failed to request the USD price of {self.protocol_token.evm_address}. '
+                        f"{self.counterparty} locked balances won't be accurate.",
+                    )
```
nicholasyoder, line 104: This should be moved out of the loops. No need to do this if and rerun the find price logic on each loop iteration.

rotkehlchen/chain/evm/decoding/velodrome/balances.py
```diff
@@ -34,6 +52,63 @@ def __init__(
...
+
+                balances[user_address].assets[self.protocol_token] = Balance(
```
nicholasyoder, line 106:
```suggestion
                balances[user_address].assets[self.protocol_token] += Balance(
```
Shouldn't this be adding to the existing balance, not replacing it? maybe I'm misunderstanding how this logic works, but if there are multiple token_ids it looks like it would only get the balance for the last one as the code is currently.

rotkehlchen/tasks/calendar.py
```diff
@@ -365,6 +367,44 @@ def maybe_create_locked_crv_reminders(self) -> None:
...
+                (HistoryEventType.INFORMATIONAL, HistoryEventSubType.NONE),  # unlock time updates
+            ],
+            counterparties=[CPT_VELODROME, CPT_AERODROME],
+        )) == 0:
+            return
+
+        lock_entries: list[int] = []
+        for event in events:
+            if not (
+                    (extra_data := event.extra_data or {}) and
```
nicholasyoder, line 384:
```suggestion
                    (extra_data := event.extra_data) is not None and
```
If extra_data is None the other conditions in the if won't be evaluated so I don't think theres any need to make sure extra_data is set to an empty dict.

rotkehlchen/chain/base/modules/aerodrome/balances.py
```diff
@@ -18,4 +21,6 @@ def __init__(
             evm_inquirer=evm_inquirer,
             tx_decoder=tx_decoder,
             counterparty=CPT_AERODROME,
+            voting_escrow_address=VOTING_ESCROW_CONTRACT_ADDRESS,
+            protocol_token=Asset('eip155:8453/erc20:0x940181a94A35A4569E4529A3CDfB74e38FD98631').resolve_to_evm_token(),
```
nicholasyoder, line 25: Should probably have a comment saying what token this is. Or define it as a constant if its used elsewhere.

rotkehlchen/chain/optimism/modules/velodrome/balances.py
```diff
@@ -18,4 +21,6 @@ def __init__(
             evm_inquirer=evm_inquirer,
             tx_decoder=tx_decoder,
             counterparty=CPT_VELODROME,
+            voting_escrow_address=VOTING_ESCROW_CONTRACT_ADDRESS,
+            protocol_token=Asset('eip155:10/erc20:0x9560e827aF36c94D2Ac33a39bCE1Fe78631088Db').resolve_to_evm_token(),
```
nicholasyoder, line 25: Same here, should have a comment saying what token it is or be defined as a constant.

rotkehlchen/chain/evm/decoding/velodrome/balances.py
nicholasyoder, line 106: hmm... `self.protocol_token` is the same token that is deposited in a `DEPOSIT_FOR_WRAPPED` event for these locked balances? Won't this mess with accounting / the asset graphs, since we count a `DEPOSIT_FOR_WRAPPED` as an out event, while this will make it show as if its still part of the users balance?

Not sure exactly what the best solution is. Perhaps the events actually need decoded as a `DEPOSIT_ASSET` event (which is treated as a neutral event), and then a plain receive or a deploy nft or something instead of the receive wrapped - since it is an nft that you get back, rather than a normal wrapped token.

rotkehlchen/chain/evm/decoding/velodrome/balances.py
prettyirrelevant: > Also, will extra_data['token_id'] always be set here, or could this possibly result in a key error?

yeah, it is always present.

rotkehlchen/chain/evm/decoding/velodrome/balances.py
prettyirrelevant: i agree with your approach. initially considered assigning the locked tokens' value to the NFT, but your suggestion is better:

- use `DEPOSIT_ASSET` instead of `DEPOSIT_FOR_WRAPPED` for the protocol token (AERO/VELO)
- then handle the NFT as probably `INFORMATIONAL` / `DEPLOY_NFT` since we're not attaching value to it.

rotkehlchen/chain/evm/decoding/velodrome/balances.py
```diff
@@ -34,6 +52,63 @@ def __init__(
             deposit_event_types={(HistoryEventType.DEPOSIT, HistoryEventSubType.DEPOSIT_FOR_WRAPPED)},  # noqa: E501
             gauge_deposit_event_types={(HistoryEventType.DEPOSIT, HistoryEventSubType.DEPOSIT_ASSET)},  # noqa: E501
```
nicholasyoder, line 53: Is the `deposit_event_types` being used at all anymore? I see its still set to `DEPOSIT_FOR_WRAPPED`. I think this could be confusing if its not actually being used - it makes it look like this balances logic uses `DEPOSIT_FOR_WRAPPED` somewhere when it actually doesn't.

rotkehlchen/chain/evm/decoding/velodrome/balances.py
```diff
@@ -34,6 +52,63 @@ def __init__(
...
         )
+        self.protocol_token = protocol_token
+        self.voting_escrow_address = voting_escrow_address
 
     def get_gauge_address(self, event: 'EvmEvent') -> ChecksumEvmAddress | None:
         return event.address
+
+    def query_balances(self) -> BalancesSheetType:
+        balances = super().query_balances()
+        if len(addresses_with_deposits := self.addresses_with_activity(event_types=self.gauge_deposit_event_types)) == 0:  # noqa: E501
```
nicholasyoder, line 63:
```suggestion
        if len(addresses_with_deposits := self.addresses_with_deposits(products=None)) == 0:  # noqa: E501
```
Maybe should set `deposit_event_types` to `DEPOSIT_ASSET` and then just use the `addresses_with_deposits` function here. Or at least use the `addresses_with_gauge_deposits` function rather than `addresses_with_activity`

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`9267` rotki now supports decoding of Aerodrome and Velodrome lock transactions, tracking locked balances, and setting reminders for lock expirations.
```
nicholasyoder, line 5:
```suggestion
* :feature:`9267` rotki now supports decoding of Aerodrome and Velodrome lock transactions, tracking locked balances, and automatic reminders for lock expirations.
```
Not sure if this is the best wording either, but it should probably be more clear that the reminders are created automatically, rather than something the user has to set.

rotkehlchen/chain/evm/decoding/velodrome/balances.py
```diff
@@ -25,15 +41,74 @@ def __init__(
...
+        voting_escrow_contract = EvmContract(
+            address=self.voting_escrow_address,
+            abi=VOTING_ESCROW_ABI,
+            deployed_block=0,
+        )
+        chunk_size, call_order = get_chunk_size_call_order(self.evm_inquirer)
+        if (price := Inquirer().find_usd_price(self.protocol_token)) == ZERO_PRICE:
+            log.error(
+                f'Failed to request the USD price of {self.protocol_token.evm_address}. '
+                f"{self.counterparty} locked balances won't be accurate.",
```
LefterisJP, line 79:
```suggestion
                f"{self.counterparty} locked balances value won't be accurate.",
```

rotkehlchen/chain/evm/decoding/velodrome/balances.py
```diff
@@ -25,15 +41,74 @@ def __init__(
...
+                    method_name='locked',
+                    arguments=[token_ids[idx]],
+                )[0]
+                if balance == 0:
+                    continue
+
+                balances[user_address].assets[self.protocol_token] += Balance(
+                    amount=(amount := token_normalized_value_decimals(
+                        token_amount=balance,
+                        token_decimals=DEFAULT_TOKEN_DECIMALS,
```
LefterisJP, line 109:
```suggestion
                        token_decimals=DEFAULT_TOKEN_DECIMALS,  # both AERO and VELo have 18 decimals
```

rotkehlchen/tests/unit/decoders/test_aerodrome.py
```diff
@@ -311,8 +311,8 @@ def test_unlock_aero(base_accounts, base_transaction_decoder):
             sequence_index=140,
             timestamp=timestamp,
             location=Location.BASE,
-            event_type=HistoryEventType.SPEND,
-            event_subtype=HistoryEventSubType.RETURN_WRAPPED,
+            event_type=HistoryEventType.INFORMATIONAL,
```
LefterisJP, line 314: Why make the burning informational? That means the NFT won't be trackable anymore

rotkehlchen/accounting/constants.py
```diff
@@ -284,6 +290,12 @@
     )}, EventCategory.CLAWBACK: {DEFAULT: EventCategoryDetails(
         label='Clawback',
         icon='lu-undo-2',
+    )}, EventCategory.MINT_NFT: {DEFAULT: EventCategoryDetails(
```
LefterisJP, line 293: there is also a place in the frontend may need to add them. I think it's rui/index.ts but ask them.

rotkehlchen/chain/evm/decoding/velodrome/balances.py
```diff
@@ -25,15 +41,74 @@ def __init__(
...
+                    method_name='locked',
+                    arguments=[token_ids[idx]],
+                )[0]
+                if balance == 0:
+                    continue
+
+                balances[user_address].assets[self.protocol_token] += Balance(
+                    amount=(amount := token_normalized_value_decimals(
+                        token_amount=balance,
+                        token_decimals=DEFAULT_TOKEN_DECIMALS,  # both AERO and VELo have 18 decimals  # noqa: E501
```
LefterisJP, line 109: That was a typo in my suggestion. Use common sense and at least ask if something seems off

```suggestion
                        token_decimals=DEFAULT_TOKEN_DECIMALS,  # both AERO and VELO have 18 decimals  # noqa: E501
```

rotkehlchen/accounting/constants.py
prettyirrelevant: there's no place in the frontend that needs updating (they retrieve all the info from the backend).

rotkehlchen/chain/evm/decoding/velodrome/balances.py
prettyirrelevant: oops. my bad. adjusted it.

rotkehlchen/accounting/constants.py
LefterisJP: there is but you did it

PR #9572, opened 2025-03-06
Add cryptocompare and defillama batch query support

nicholasyoder:
Related: https://github.com/rotki/rotki/issues/8978

### DIFF
--- a/rotkehlchen/externalapis/cryptocompare.py
+++ b/rotkehlchen/externalapis/cryptocompare.py
@@ -1,5 +1,6 @@
 import logging
 from collections import deque
+from enum import StrEnum
 from json.decoder import JSONDecodeError
 from typing import TYPE_CHECKING, Any, Final, Literal, Optional, overload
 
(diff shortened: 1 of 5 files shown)


rotkehlchen/externalapis/cryptocompare.py
```diff
@@ -164,6 +165,15 @@
 }
 CRYPTOCOMPARE_SPECIAL_CASES = CRYPTOCOMPARE_SPECIAL_CASES_MAPPING.keys()
 CRYPTOCOMPARE_HOURQUERYLIMIT = 2000
+MAX_FSYMS_CHARS: Final = 300
+
+
+class CCApiUrl(Enum):
```
LefterisJP, line 171: Use a StrEnum. https://docs.python.org/3/library/enum.html#enum.StrEnum.

You don't need to use `.value` everywhere then to get the string.


Also question: Why an Enum? Why do you need this?

rotkehlchen/externalapis/cryptocompare.py
```diff
@@ -164,6 +165,15 @@
 }
 CRYPTOCOMPARE_SPECIAL_CASES = CRYPTOCOMPARE_SPECIAL_CASES_MAPPING.keys()
 CRYPTOCOMPARE_HOURQUERYLIMIT = 2000
+MAX_FSYMS_CHARS: Final = 300
```
LefterisJP, line 168: add a comment explaining what this is

rotkehlchen/externalapis/defillama.py
```diff
@@ -254,8 +276,13 @@ def query_historical_price(
...
 
-        usd_price = self._deserialize_price(result, coin_id, from_asset, to_asset)
-        if usd_price == ZERO:
+        usd_price = self._deserialize_prices(
+            result=result,
+            coin_id_mapping={coin_id: from_asset},
+            from_assets=[from_asset],
+            to_asset=to_asset,
+        ).get(from_asset, ZERO_PRICE)
+        if usd_price == ZERO_PRICE:
```
LefterisJP, line 285:
```suggestion
        if (usd_price := self._deserialize_prices(
            result=result,
            coin_id_mapping={coin_id: from_asset},
            from_assets=[from_asset],
            to_asset=to_asset,
        ).get(from_asset, ZERO_PRICE)) == ZERO_PRICE:
```

rotkehlchen/externalapis/cryptocompare.py
nicholasyoder: > Why do you need this?

It's not strictly necessary - but it seems better to only define these urls in one place, rather than using the full url strings both in the Literal hints and when calling the function.

LefterisJP: The issue with this though then becomes memory. It's probably negligible here but for no apparent reason to me you are adding Enums somewhere where they don't need to be. I mean even simple constants would be both more readable and take less memory. 

But here it's probably negligible, though still I would ask you to be careful about stuff like this and not add stuff that are unnecessary

LefterisJP (approved):
LGTM

PR #9573, opened 2025-03-06
Use batch queries in the current prices endpoint

nicholasyoder:
Related: https://github.com/rotki/rotki/issues/8978

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -3118,33 +3118,36 @@ class RestAPI:
             target_asset: Asset,
             ignore_cache: bool,
     ) -> dict[str, Any]:
-        """Return the current price of the assets in the target asset currency.
-        """
+        """Return the current price of the assets in the target asset currency."""
         log.debug(
             f'Querying the current {target_asset.identifier} price of these assets: '
             f'{", ".join([asset.identifier for asset in assets])}',
         )
         # Type is list instead of tuple here because you can't serialize a tuple
         assets_price: dict[Asset, list[Price | (int | None)]] = {}
+        non_nft_assets = []
         for asset in assets:
-            if asset != target_asset:
-                if asset.asset_type == AssetType.NFT:
-                    nft_price_data = self._eth_module_query(
-                        module_name='nfts',
-                        method='get_nfts_with_price',
-                        query_specific_balances_before=None,
-                    )
-                    oracle = CurrentPriceOracle.MANUALCURRENT if nft_price_data['manually_input'] is True else CurrentPriceOracle.BLOCKCHAIN  # noqa: E501
-                    assets_price[asset] = [Price(nft_price_data['usd_price']), oracle.value]
-                else:
-                    price, oracle = Inquirer.find_price_and_oracle(
-                        from_asset=asset,
-                        to_asset=target_asset,
-                        ignore_cache=ignore_cache,
-                    )
-                    assets_price[asset] = [price, oracle.value]
+            if asset.asset_type == AssetType.NFT:
+                nft_price_data = self._eth_module_query(
+                    module_name='nfts',
+                    method='get_nfts_with_price',
+                    query_specific_balances_before=None,
+                )
+                oracle = CurrentPriceOracle.MANUALCURRENT if nft_price_data['manually_input'] is True else CurrentPriceOracle.BLOCKCHAIN  # noqa: E501
+                assets_price[asset] = [Price(nft_price_data['usd_price']), oracle.value]
             else:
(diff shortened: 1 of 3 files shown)


rotkehlchen/api/rest.py
```diff
@@ -3118,33 +3118,36 @@ def get_current_assets_price(
...
+                nft_price_data = self._eth_module_query(
+                    module_name='nfts',
+                    method='get_nfts_with_price',
+                    query_specific_balances_before=None,
+                )
+                oracle = CurrentPriceOracle.MANUALCURRENT if nft_price_data['manually_input'] is True else CurrentPriceOracle.BLOCKCHAIN  # noqa: E501
+                assets_price[asset] = [Price(nft_price_data['usd_price']), oracle.value]
             else:
-                assets_price[asset] = [Price(ONE), CurrentPriceOracle.BLOCKCHAIN.value]
+                unpriced_assets.append(asset)
```
LefterisJP, line 3139: this is really confusing. You go through the list of assets and if it's an nft you query it and then call the rest unpriced assets. In reality they are the non-nft assets. Use proper naming.

yabirgb:
I get this error

```
[07/03/2025 10:33:55 CET] ERROR rotkehlchen.api.rest Greenlet with id 4847351104: Greenlet for task 30 dies with exception: <Asset identifier:eip155:1/erc20:0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48>.
Exception Name: <class 'KeyError'>
Exception Info: <Asset identifier:eip155:1/erc20:0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48>
Traceback:
   File "src/gevent/greenlet.py", line 900, in gevent._gevent_cgreenlet.Greenlet.run
  File "/Users/yabirgb/work/rotki/rotkehlchen/api/rest.py", line 461, in _do_query_async
    result = command(self, **kwargs)
    ^^^^^^^^^^^^^^^^^
  File "/Users/yabirgb/work/rotki/rotkehlchen/api/rest.py", line 3142, in get_current_assets_price
    found_prices = Inquirer.find_prices_and_oracles(
    ^^^^^^^^^^^^^^^^^
  File "/Users/yabirgb/work/rotki/rotkehlchen/inquirer.py", line 824, in find_prices_and_oracles
    return Inquirer._find_prices(
    ^^^^^^^^^^^^^^^^^
  File "/Users/yabirgb/work/rotki/rotkehlchen/inquirer.py", line 762, in _find_prices
    found_prices[original_asset] = found_prices[replacement_asset]
    ^^^^^^^^^^^^^^^^^
```

so it doesn't work

yabirgb:
It worked after the changes that nicholas did

rotkehlchen/inquirer.py
```diff
@@ -758,10 +758,14 @@ def _find_prices(
...
 
         # Only include the assets that were originally requested.
+        unneeded_replacements = set()
         for original_asset, replacement_asset in replaced_assets.items():
             found_prices[original_asset] = found_prices[replacement_asset]
             if replacement_asset not in from_assets:
-                del found_prices[replacement_asset]
+                unneeded_replacements.add(replacement_asset)
+
+        for asset in unneeded_replacements:
```
LefterisJP, line 767: It's a set. No need to iterate. Just get the set differene of `set(found_prices) - unneeded_replacements`

rotkehlchen/inquirer.py
```diff
@@ -758,10 +758,14 @@ def _find_prices(
                 ))
 
         # Only include the assets that were originally requested.
+        unneeded_replacements = set()
         for original_asset, replacement_asset in replaced_assets.items():
             found_prices[original_asset] = found_prices[replacement_asset]
             if replacement_asset not in from_assets:
-                del found_prices[replacement_asset]
+                unneeded_replacements.add(replacement_asset)
```
LefterisJP, line 765: also why are you doing it like this. What was wrong before?

rotkehlchen/inquirer.py
nicholasyoder: still have to iterate to build the new dict, but I suppose that's better than repeatedly calling `del`

rotkehlchen/inquirer.py
nicholasyoder: If several assets from the same collection were queried, but not the main asset, then the main asset would get removed from found_prices here when it processed the first asset and then when it tried to process the second asset from the same collection, `found_prices[replacement_asset]` would result in a key error since the main asset had already been removed. So it has to wait and remove the unneeded main assets until after this loop is finished.

LefterisJP (approved):
LGTM

PR #9575, opened 2025-03-07
chore: fix token symbol

fragchain:
Noticed that the wrong token symbol was used—**WBTC** instead of **WSBTC**.
Fixed it, everything should be correct now.

## Checklist

- [x] The PR modified the frontend, and updated the [user guide](https://github.com/rotki/docs/blob/main/usage-guides/index.md) to reflect the changes.


LefterisJP:
Hello @fragchain. Thank you for your PR but this is unfortunately violating our [policy](https://github.com/rotki/rotki/blob/develop/CONTRIBUTING.md) of low effort PRs .

What's more there is no mistake there. And if there was you would have had to also change the usage of the variable where it's used.

PR #9580, opened 2025-03-07
add icon for uniswap v3 nft

prettyirrelevant:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=100293085

CleanShot 2025-03-10 at 16 31 23@2x

### DIFF
--- a/colibri/src/blockchain/node_inquirer.rs
+++ b/colibri/src/blockchain/node_inquirer.rs
@@ -1,0 +1,112 @@
+use std::sync::Arc;
+use std::collections::HashMap;
+use alloy::providers::DynProvider;
+use alloy::providers::ProviderBuilder;
+use log::{debug, error};
+use tokio::sync::RwLock;
+
+use crate::blockchain::{SupportedBlockchain, RpcNode};
+use crate::globaldb::GlobalDB;
+
+
+/// A struct that manages connections to EVM RPC nodes for a specific blockchain
+pub struct EvmNodeInquirer {
+    globaldb: Arc<GlobalDB>,
+    pub rpc_nodes: RwLock<Vec<RpcNode>>,
+    pub blockchain: SupportedBlockchain,
+    // Connected nodes
+    provider_mapping: RwLock<HashMap<RpcNode, Box<DynProvider>>>,
+}
+
+impl EvmNodeInquirer {
+    pub fn new(blockchain: SupportedBlockchain, globaldb: Arc<GlobalDB>) -> Self {
+        let instance = Self {
+            blockchain,
+            globaldb,
+            rpc_nodes: RwLock::new(Vec::new()),
+            provider_mapping: RwLock::new(HashMap::new()),
+        };
+
+        debug!("created EvmNodeInquirer for {}", blockchain.as_str());
+        instance
+    }
+
+    pub async fn update_rpc_nodes(&self) -> Result<(), String> {
+        let nodes = self.globaldb.get_rpc_nodes(self.blockchain)
+            .await
+            .map_err(|e| format!("Failed to get RPC nodes: {}", e))?;
+
+        *self.rpc_nodes.write().await = nodes;
+        Ok(())
+    }
+
+    /// Gets an existing RPC node connection or creates a new one.
+    ///
+    /// Checks if a connection to the specified RPC node already exists in the cache.
+    /// If found, returns the existing provider; otherwise creates a new connection,
+    /// stores it in the cache, and returns it.
+    pub async fn get_or_create_node_connection(&self, node: &RpcNode) -> Result<Arc<DynProvider>, String> {
+        if let Some(provider) = self.provider_mapping.read().await.get(node) {
+            return Ok(Arc::from(provider.clone()));
+        }
+
+        let endpoint = node.endpoint.parse()
+            .map_err(|e| format!("Invalid endpoint URL: {}", e))?;
+        let provider = DynProvider::new(ProviderBuilder::new().on_http(endpoint));
+        let mut mapping = self.provider_mapping.write().await;
+        mapping.insert(node.clone(), Box::new(provider.clone()));
+
+        Ok(Arc::new(provider))
+    }
+}
+
+// A simple manager that stores EVM node inquirers for different chains
+pub struct EvmInquirerManager {
+    inquirers: RwLock<HashMap<SupportedBlockchain, Arc<EvmNodeInquirer>>>,
+    pub globaldb: Arc<GlobalDB>,
+}
+
+impl EvmInquirerManager {
+    pub fn new(globaldb: Arc<GlobalDB>) -> Self {
+        Self {
+            inquirers: RwLock::new(HashMap::new()),
+            globaldb,
+        }
+    }
+
+    pub async fn initialize_rpc_nodes(&self) {
+        let mut inquirers = self.inquirers.write().await;
+        for blockchain in [
+            SupportedBlockchain::Ethereum,
+            SupportedBlockchain::Optimism,
+            SupportedBlockchain::PolygonPos,
+            SupportedBlockchain::ArbitrumOne,
+            SupportedBlockchain::Base,
+            SupportedBlockchain::Gnosis,
+            SupportedBlockchain::BinanceSc,
+        ] {
+            let inquirer = inquirers.entry(blockchain)
+                .or_insert_with(|| Arc::new(EvmNodeInquirer::new(blockchain, self.globaldb.clone())));
+
+            if let Err(e) = inquirer.update_rpc_nodes().await {
+                error!("Failed to update RPC nodes for {}: {}", blockchain.as_str(), e);
+            }
+        }
+    }
+
+    pub async fn get_or_init_inquirer(&self, blockchain: SupportedBlockchain) -> Arc<EvmNodeInquirer> {
+        if let Some(inquirer) = self.inquirers.read().await.get(&blockchain) {
+            return inquirer.clone();
+        }
+
+        let new_inquirer = Arc::new(EvmNodeInquirer::new(blockchain, self.globaldb.clone()));
+        if let Err(e) = new_inquirer.update_rpc_nodes().await {
+            error!("Failed to initialize RPC nodes for {}: {}", blockchain.as_str(), e);
+        }
+
+        let mut inquirers = self.inquirers.write().await;
+        inquirers.entry(blockchain)
+            .or_insert(new_inquirer)
+            .clone()
+    }
+}
--- a/colibri/src/blockchain/types.rs
+++ b/colibri/src/blockchain/types.rs
@@ -1,0 +1,51 @@
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+/// Represents supported EVM-compatible blockchains
+/// 
+/// Note: This implementation only includes EVM chains and is not identical to Python's implementation
+/// which may include additional non-EVM blockchains.
+pub enum SupportedBlockchain {
+    Ethereum,
+    Optimism,
+    PolygonPos,
+    ArbitrumOne,
+    Base,
+    Gnosis,
+    BinanceSc,
+}
+
+impl SupportedBlockchain {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::Ethereum => "ETH",
+            Self::Optimism => "OPTIMISM",
+            Self::PolygonPos => "POLYGON_POS",
+            Self::ArbitrumOne => "ARBITRUM_ONE",
+            Self::Base => "BASE",
+            Self::Gnosis => "GNOSIS",
+            Self::BinanceSc => "BINANCE_SC",
+        }
+    }
+
+    pub fn from_chain_id(chain_id: u64) -> Option<Self> {
+        // Maximum chain ID value is floor(MAX_UINT64 / 2) - 36 as per EIP-2294
+        // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2294.md
+        match chain_id {
+            1 => Some(Self::Ethereum),
+            10 => Some(Self::Optimism),
+            42161 => Some(Self::ArbitrumOne),
+            8453 => Some(Self::Base),
+            56 => Some(Self::BinanceSc),
+            100 => Some(Self::Gnosis),
+            137 => Some(Self::PolygonPos),
+            _ => None,
+        }
+    }
+}
+
+/// Information about an RPC node
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct RpcNode {
+    pub name: String,
+    pub endpoint: String,
+    pub blockchain: SupportedBlockchain,
+}
--- a/colibri/src/globaldb.rs
+++ b/colibri/src/globaldb.rs
@@ -1,3 +1,4 @@
+use crate::blockchain::{RpcNode, SupportedBlockchain};
 use rusqlite::{Connection, Result};
 use std::path::PathBuf;
 use std::sync::Arc;
@@ -62,6 +63,38 @@ impl GlobalDB {
                 })
         })
     }
+
+    /// Get all active RPC endpoints for a specific blockchain.
+    pub async fn get_rpc_nodes(&self, blockchain: SupportedBlockchain) -> Result<Vec<RpcNode>> {
+        let conn = self.conn.lock().await;
+        conn.prepare("SELECT name, endpoint FROM default_rpc_nodes WHERE blockchain=? AND name NOT LIKE '%etherscan%' AND active=1 AND (CAST(weight as decimal) != 0 OR owned == 1) ORDER BY name;")
+            .and_then(|mut stmt| {
+                let mut rows = stmt.query(rusqlite::params![blockchain.as_str()])?;
+                let mut nodes = Vec::new();
+                while let Some(row) = rows.next()? {
+                    let name: String = row.get(0)?;
+                    let endpoint: String = row.get(1)?;
+                    nodes.push(RpcNode {
+                        name,
+                        endpoint,
+                        blockchain,
+                    });
+                }
+                Ok(nodes)
+            })
+    }
+
+    pub async fn is_uniswap_v3_position(&self, asset_id: &str) -> Result<bool> {
+        let conn = self.conn.lock().await;
+        let mut stmt = conn.prepare(
+            "SELECT 1 FROM evm_tokens
+             WHERE identifier = ?
+             AND protocol = 'UNI-V3'
+             LIMIT 1"
+        )?;
+        let result = stmt.exists(rusqlite::params![asset_id])?;
+        Ok(result)
+    }
 }
 
 /// macro that creates a copy of the globaldb in the rotkehlchen data folder
@@ -71,10 +104,16 @@ impl GlobalDB {
 macro_rules! create_globaldb {
     () => {{
         use crate::globaldb::GlobalDB;
-        use std::{env, fs, path::PathBuf};
+        use std::{env, path::PathBuf, time::SystemTime};
+        use tokio::fs;
 
-        let tmp_dir = env::temp_dir().join("global");
-        fs::create_dir_all(tmp_dir.clone()).expect("Failed to create temp folder for globaldb");
+        let timestamp = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("Time went backwards")
+            .as_nanos();
+        let pid = std::process::id();
+        let tmp_dir = env::temp_dir().join(format!("global_{}_{}", timestamp, pid));
+        fs::create_dir_all(tmp_dir.clone()).await.expect("Failed to create temp folder for globaldb");
         let root_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
         fs::copy(
             (root_path.parent())
@@ -84,7 +123,9 @@ macro_rules! create_globaldb {
                 .join("global.db"),
             tmp_dir.join("global.db"),
         )
+        .await
         .expect("Failed to copy globaldb in create_globaldb macro");
+
         GlobalDB::new(tmp_dir.join("global.db"))
     }};
 }
--- a/colibri/src/icons.rs
+++ b/colibri/src/icons.rs
@@ -1,19 +1,31 @@
+use crate::blockchain::{parse_asset_identifier, EvmInquirerManager, EvmNodeInquirer, SupportedBlockchain};
+use crate::coingecko;
+use crate::globaldb;
+use alloy::{
+    primitives::{Address, U256},
+    sol,
+};
 use axum::body::Bytes;
 use axum::http::StatusCode;
+use base64::{engine::general_purpose::STANDARD, Engine as _};
 use log::{debug, error};
 use reqwest::Client;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::Duration;
 
-use crate::coingecko;
-use crate::globaldb;
-
 const SMOLDAPP_BASE_URL: &str =
     "https://raw.githubusercontent.com/SmolDapp/tokenAssets/refs/heads/main/tokens";
 
 pub enum FileTypeError {
     UnsupportedFileType,
+}
+
+
+sol! {
+    #[sol(rpc)]
+    UniswapV3NFTManager,
+    "src/blockchain/abis/UniswapV3NFTManager.json"
 }
 
 // Create the response headers from a path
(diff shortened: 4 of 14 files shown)


colibri/src/icons.rs
```diff
@@ -119,6 +138,86 @@ async fn query_token_icon_and_extension(
...
+        Ok(address) => address,
+        Err(_e) => {
+            return None;
+        }
+    };
+    let token_id: U256 = token_id.parse().unwrap();
+
+    // Try all RPC endpoints until one succeeds
+    for rpc_url in rpc_endpoints {
+        let provider = match rpc_url.parse().map(|url| ProviderBuilder::new().on_http(url)) {
```
yabirgb, line 159: you are destroying and creating the rpcs all the time. Ideally you would have them created and reused in some structure like we have in python the Web3 node manager. What does alloy in the builder?

colibri/src/icons.rs
```diff
@@ -119,6 +138,86 @@ async fn query_token_icon_and_extension(
...
+                continue;
+            }
+        };
+
+        let contract = IERC721Metadata::new(contract_address, provider);
+        let token_uri = match contract.tokenURI(token_id).call().await {
+            Ok(result) => result._0,
+            Err(e) => {
+                error!("RPC call failed on {}: {}", rpc_url, e);
+                continue;
```
yabirgb, line 172: we need more info like the contract called and the chain

colibri/src/icons.rs
```diff
@@ -119,6 +138,86 @@ async fn query_token_icon_and_extension(
...
+            }
+        };
+        let base64_str = match token_uri.strip_prefix("data:application/json;base64,") {
+            Some(data) => data,
+            None => {
+                error!("Invalid token URI format from {}", rpc_url);
+                continue;
```
yabirgb, line 179: do you want to continue here? This means that you got a response but the contract is badly formated. Any other node would return valid information?

colibri/src/icons.rs
```diff
@@ -119,6 +138,86 @@ async fn query_token_icon_and_extension(
...
+            }
+        };
+        let image_uri = match json_data.get("image").and_then(|v| v.as_str()) {
+            Some(uri) => uri,
+            None => {
+                error!("No 'image' field in JSON from {}", rpc_url);
+                continue;
+            }
+        };
```
yabirgb, line 202: is something like this possible?

```rust
if let Ok(json_bytes) = BASE64_STANDARD.decode(base64_str) {
    if let Ok(json_data) = serde_json::from_slice(&json_bytes) {
        if let Some(image_uri) = json_data.get("image").and_then(|v| v.as_str()) {
            // Use image_uri here
        } else {
            error!("No 'image' field in JSON from {}", rpc_url);
        }
    } else {
        error!("Failed to parse JSON from {}: {}", rpc_url, e);
    }
} else {
    error!("Failed to decode base64 JSON from {}: {}", rpc_url, e);
    continue;
}
```

I find it easier to read and uses less lines

colibri/src/icons.rs
```diff
@@ -119,6 +138,86 @@ async fn query_token_icon_and_extension(
...
+        match BASE64_STANDARD.decode(image_base64) {
+            Ok(image_data) => return Some((Bytes::from(image_data), "svg")),
+            Err(e) => {
+                error!("Failed to decode uniswap v3 base64 image from {}: {}", rpc_url, e);
+                continue;
+            }
+        }
+    }
```
yabirgb, line 217: this whole function is very confusing. I believe it could be written in a more clear way and with comments. Also I believe that you continue for errors that are fatal

colibri/src/icons.rs
```diff
@@ -16,6 +22,13 @@ pub enum FileTypeError {
     UnsupportedFileType,
 }
 
+sol! {
+    #[sol(rpc)]
+    interface IERC721Metadata {
+        function tokenURI(uint256 tokenId) external view returns (string memory);
+    }
+}
```
yabirgb, line 30: should we have them in a contract file? I believe having them disperse across the code is not a good idea

colibri/src/icons.rs
```diff
@@ -5,7 +5,13 @@ use reqwest::Client;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::Duration;
-
+use alloy::{
+    primitives::{Address, U256},
+    providers::{ProviderBuilder},
+    sol,
+};
+use base64::prelude::*;
```
yabirgb, line 13: do you need to use all?

colibri/src/blockchain.rs
```diff
@@ -0,0 +1,64 @@
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
yabirgb: I would prefer to have them mapped and avoid dead code

colibri/src/icons.rs
```diff
@@ -270,8 +370,33 @@ pub async fn query_icon_remotely(
...
+                        &contract_address,
+                        &globaldb,
+                    )
+                        .await
+                    {
+                        let _ = tokio::fs::write(path.with_extension(extension), &icon_bytes)
+                            .await
+                            .map_err(|e| {
+                                error!("Failed to write icon: {}", e);
+                            });
```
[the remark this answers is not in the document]
prettyirrelevant: no, we don't. created one and reused that.

colibri/src/blockchain/types.rs
```diff
@@ -0,0 +1,48 @@
...
+            100 => Some(Self::Gnosis),
+            137 => Some(Self::PolygonPos),
+            _ => None,
+        }
+    }
+}
+
+/// Information about an RPC node
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Node {
```
yabirgb, line 43: rather than Node that is quite generic call it RPCNode?

colibri/src/database/user_db.rs
```diff
@@ -2,7 +2,6 @@ use async_sqlite::{Client, ClientBuilder};
 use std::collections::HashSet;
 use std::path::PathBuf;
 use std::sync::Arc;
-
 use crate::database::errors::DBError;
```
yabirgb, line 5: why this change?

colibri/src/globaldb.rs
```diff
@@ -62,6 +63,28 @@ impl GlobalDB {
                 })
         })
     }
+
+    /// Get all active RPC endpoints for a specific blockchain.
+    pub async fn get_rpc_nodes(&self, blockchain: SupportedBlockchain) -> Result<Vec<Node>> {
+        let conn = self.conn.lock().await;
```
yabirgb, line 69: why return Result rather than the vec itself?

colibri/src/icons.rs
```diff
@@ -119,6 +136,98 @@ async fn query_token_icon_and_extension(
...
+        return None;
+    }
+
+    let token_id: U256 = match token_id.parse() {
+        Ok(id) => id,
+        Err(e) => {
+            error!("Invalid token ID '{}' for NFT position on chain ID {} ({}): {}", token_id, chain_id, inquirer.blockchain.as_str(), e);
+            return None;
+        }
+    };
```
yabirgb, line 156: this is the same
```suggestion
    let token_id: U256 = token_id.parse()
    .map_err(|e| {
        error!("Invalid token ID '{}' for NFT position on chain ID {} ({}): {}", 
            token_id, chain_id, inquirer.blockchain.as_str(), e)
    })
    .ok()?;
```

and avoids the Ok(id) => id part

colibri/src/icons.rs
```diff
@@ -119,6 +136,98 @@ async fn query_token_icon_and_extension(
...
+
+        // process the base64 data from the rpc call
+        let base64_str = match token_uri.strip_prefix("data:application/json;base64,") {
+            Some(data) => data,
+            None => {
+                error!("Invalid token URI format from node '{}' for token ID {} on contract {}: URI does not start with 'data:application/json;base64,'",
+                       node_name.name, token_id, contract_address);
+                break;
+            }
+        };
```
yabirgb, line 179:
```suggestion
        let Some(base64_str) = token_uri.strip_prefix("data:application/json;base64,") else {
            error!("Invalid token URI format from node '{}' for token ID {} on contract {}: URI does not start with 'data:application/json;base64,'",
                node_name.name, token_id, contract_address);
            break;
        };
```

this should also work

colibri/src/icons.rs
```diff
@@ -119,6 +136,98 @@ async fn query_token_icon_and_extension(
...
+            },
+            Err(e) => {
+                error!("Failed to decode base64 JSON from node '{}' for token ID {} on contract {}: {}",
+                       node_name.name, token_id, contract_address, e);
+                break;
+            }
+        };
+
+        // retrieve the base64 image from the json data
+        let image_base64 = match json_data.get("image").and_then(|v| v.as_str()) {
```
yabirgb, line 199: what happens when the key is not found in rust?

colibri/src/icons.rs
```diff
@@ -261,68 +384,73 @@ pub async fn query_icon_remotely(
...
-                            asset_id
-                        );
-                        write_zero_bytes_file(path.as_path()).await;
-                        return;
-                    }
-                },
+    if let Some((chain_id, contract_address, token_id)) = parse_asset_identifier(&asset_id) {
+        let blockchain = match SupportedBlockchain::from_chain_id(chain_id) {
+            Some(blockchain) => blockchain,
+            None => {
```
yabirgb, line 394: Unless I'm mistaken here you are checking always for on chain nft info but you need only to do this in the case of NFTs at most

colibri/src/icons.rs
```diff
@@ -261,68 +384,73 @@ pub async fn query_icon_remotely(
...
+                debug!("Unsupported chain ID: {} for asset {}", chain_id, asset_id);
+                return
```
yabirgb, line 396: if you do this you won't even try checking smoldapp or coingecko right?

colibri/src/icons.rs
```diff
@@ -30,23 +41,29 @@ fn get_headers(extension: &str) -> Result<[(&'static str, &'static str); 2], Fil
     Ok([("Content-Type", content_type), ("mimetype", content_type)])
 }
 
-fn extract_chain_id_and_address(identifier: &str) -> Option<(u64, String)> {
+fn parse_asset_identifier(identifier: &str) -> Option<(u64, String, Option<String>)> {
```
yabirgb, line 44: add a docstring to this function

colibri/src/icons.rs
```diff
@@ -30,23 +41,29 @@ fn get_headers(extension: &str) -> Result<[(&'static str, &'static str); 2], Fil
...
     let chain_id = chain_parts[1].parse::<u64>().ok()?;
-    let address = address_parts[1].to_string();
 
-    Some((chain_id, address))
+    // Parse asset type and contract address
+    let asset_parts: Vec<&str> = parts[1].splitn(2, ':').collect();
+    if asset_parts.len() != 2 {
+        return None;
+    }
+    let contract_address = asset_parts[1].to_string();
```
yabirgb, line 62: can't this break if somehow the id is wrong and it is missing parts? Is it possible to access with a get method that allows you to catch errors?

colibri/src/icons.rs
```diff
@@ -261,68 +384,73 @@ pub async fn query_icon_remotely(
...
+    if let Some(icon_bytes) = coingecko.query_asset_image(&asset_id).await {
+        let _ = tokio::fs::write(path.with_extension("png"), &icon_bytes)
+            .await
+            .map_err(|e| {
+                error!(
+                    "Unable to write {} to the file system due to {}",
+                    path.display(),
+                    e
+                );
+            });
```
yabirgb, line 434: you can use `write_icon_to_file` here right?

colibri/src/icons.rs
```diff
@@ -261,68 +384,73 @@ pub async fn query_icon_remotely(
...
-                    write_zero_bytes_file(path.as_path()).await;
-                    return;
+        };
+
+        // first, check if it's uniswap v3 lp nft
+        if let Some(manager_address) = get_uniswap_nft_manager(chain_id) {
+            if manager_address.to_string() == contract_address && token_id.is_some() {
+                if let Some((icon_bytes, extension)) = query_uniswap_v3_position_icon(
+                    chain_id,
+                    &token_id.unwrap(),
```
yabirgb, line 405: unwrap here is bad 🙈 won't this fail for every non nft token?

colibri/src/icons.rs
```diff
@@ -261,68 +384,73 @@ pub async fn query_icon_remotely(
...
-                        "Icon not found in coingecko for asset {}. Writing zero bytes file.",
-                        asset_id
-                    );
-                    write_zero_bytes_file(path.as_path()).await;
-                    return;
+        };
+
+        // first, check if it's uniswap v3 lp nft
+        if let Some(manager_address) = get_uniswap_nft_manager(chain_id) {
+            if manager_address.to_string() == contract_address && token_id.is_some() {
```
yabirgb, line 402: ah won't fail because of this but this whole function is cursed and can be better written:

1. You don't need always to check the chain, only when it is a NFT and in the case of NFT only when it is a uniswap position. Do tokens keep the unisaw-v3 protocol? We can use it since we have access to the globaldb
2. If it is uniswap most probably you don't need to check other sources so you can exit
3. It is important that if you don't have uniswap LPs colibri doesn't connect ever to the nodes. No need in that case
4. I see some unswaps in the code like `evm_inquirer_manager.get_or_init_inquirer(blockchain).await.unwrap(),` please don't do that. Handle the errors. unwrap is good while prototyping but shouldn't be used in prod. It will kill the app
5. I believe `parse_asset_identifier` is complicated an error prone. Do we have something similar in python to use as template? Maybe return a struct instead of 3 items if it makes thing simpler and more redeable

colibri/src/icons.rs
```diff
@@ -359,4 +487,39 @@ mod tests {
...
+        let globaldb = create_globaldb!().await.unwrap();
```
yabirgb, line 495: unwrap in tests good 👍 . Won't kill the app

colibri/src/main.rs
```diff
@@ -58,17 +63,17 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-                }
+        .filter_map(|pattern| match Pattern::new(pattern) {
+            Ok(glob) => Some(glob),
+            Err(e) => {
+                error!(
+                    "Found pattern {} not valid for API CORS due to {}",
+                    pattern, e
+                );
+                None
             }
```
yabirgb, line 77: why changes in the style here?

colibri/src/globaldb.rs
```diff
@@ -62,6 +63,28 @@ impl GlobalDB {
...
+        conn.prepare("SELECT name, endpoint, owned FROM default_rpc_nodes WHERE blockchain=? AND name NOT LIKE '%etherscan%' AND active=1 AND (CAST(weight as decimal) != 0 OR owned == 1) ORDER BY name;")
+            .and_then(|mut stmt| {
```
yabirgb, line 71: in the globaldb we don't have weight 0 afaik or owned

colibri/src/blockchain/node_inquirer.rs
```diff
@@ -0,0 +1,140 @@
+use std::sync::Arc;
+use std::collections::{HashMap};
```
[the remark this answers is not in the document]
prettyirrelevant: it does not but i adjusted it nonetheless

colibri/src/main.rs
prettyirrelevant: most likely an issue from my ide. reverting it.

colibri/src/globaldb.rs
prettyirrelevant: unless i'm wrong but isn't `Result<>` or `Option<>` the way rust handles return values or am i missing something?

yabirgb: Yes but you can also return an empty vec. It is not mandatory to wrap everything on result/option

colibri/src/icons.rs
prettyirrelevant: errr... not really, it can be other reasons like the rpc failing and not the contract.

colibri/src/icons.rs
```diff
@@ -30,23 +48,39 @@ fn get_headers(extension: &str) -> Result<[(&'static str, &'static str); 2], Fil
...
 
-fn extract_chain_id_and_address(identifier: &str) -> Option<(u64, String)> {
+/// Parses an asset identifier in the format "eip155:{chain_id}/{asset_type}:{contract_address}[/{token_id}]"
+///
+/// Examples:
+///   - "eip155:1/erc20:0x6B175474E89094C44Da98b954EedeAC495271d0F" (ERC-20 token)
+///   - "eip155:1/erc721:0xBC4CA0EdA7647A8aB7C2061c2E118A18a936f13D/1" (ERC-721 NFT with token ID)
+/// 
+/// Returns None if the format is invalid or any required component is missing.
+fn parse_asset_identifier(identifier: &str) -> Option<AssetIdentifier> {
```
yabirgb, line 58: can you add a test for this one?

colibri/src/icons.rs
```diff
@@ -261,68 +397,67 @@ pub async fn query_icon_remotely(
...
-                            asset_id
-                        );
-                        write_zero_bytes_file(path.as_path()).await;
-                        return;
+    // Parse asset identifier for EVM-based assets
+    if let Some(asset_info) = parse_asset_identifier(&asset_id) {
+        // Handle NFTs - only check for Uniswap V3 if we have a token ID
+        if let Some(token_id) = &asset_info.token_id {
+            if let Some(manager_address) = get_uniswap_nft_manager(asset_info.chain_id) {
+                if manager_address.to_string() == asset_info.contract_address {
```
yabirgb, line 409: isn't it better to query the gloaldb here before checking the chain? This will trigger to every possible nft

colibri/src/blockchain/node_inquirer.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    /// Get inquirer for a specific blockchain, initializing it if it doesn't exist
+    pub async fn get_or_init_inquirer(&self, blockchain: SupportedBlockchain) -> Option<Arc<EvmNodeInquirer>> {
+        // First check if it already exists
+        {
+            let inquirers = self.inquirers.read().await;
+            if let Some(inquirer) = inquirers.get(&blockchain) {
+                return Some(inquirer.clone());
+            }
+        }
```
yabirgb, line 136: why this closure here?

colibri/src/blockchain/node_inquirer.rs
```diff
@@ -0,0 +1,141 @@
...
+        let mut inquirers = self.inquirers.write().await;
+        if let std::collections::hash_map::Entry::Vacant(e) = inquirers.entry(blockchain) {
+            let inquirer = Arc::new(EvmNodeInquirer::new(blockchain, self.globaldb.clone()));
+            inquirer.connect_to_all_nodes().await;
+            e.insert(inquirer.clone());
+            Some(inquirer)
+        } else {
+            inquirers.get(&blockchain).cloned()
+        }
+    }
```
yabirgb, line 125: but again this will connect to all the nodes in the list right? Why would you want this?

in python what we do is have the list of names and then iterate trough it, if the node is connected get it from the mapping and if not connect to it. Why did you decide to init all of them? doesn't it make sense to just init one and if it fails you init the next one? If there is some technical detail that I don't see that makes it dificult you can leave a TODO and we can see it in another PR but I'm not sure why you would want to init all of them.

colibri/src/blockchain/types.rs
```diff
@@ -0,0 +1,47 @@
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum SupportedBlockchain {
```
yabirgb, line 2: this might be confusing since you don't list all of them. Can you point in the docstring that this isn't the same as rotki since you only use the evm chains? Maybe is worth renaming it to SupportedEvmChains?

colibri/src/blockchain/node_inquirer.rs
prettyirrelevant: not needed. removed.

colibri/src/icons.rs
```diff
@@ -1,21 +1,39 @@
...
 pub enum FileTypeError {
     UnsupportedFileType,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AssetIdentifier {
+    pub chain_id: u64,
+    pub contract_address: String,
+    pub token_id: Option<String>,
+}
```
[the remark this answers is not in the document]
prettyirrelevant: moved it (the function & type to `blockchain/utils.rs`)

colibri/src/blockchain/types.rs
prettyirrelevant: added a docstring.

colibri/src/blockchain/node_inquirer.rs
prettyirrelevant: found a way around it. updated it.

colibri/src/globaldb.rs
```diff
@@ -62,6 +63,38 @@ impl GlobalDB {
...
+    }
+
+    pub async fn is_uniswap_v3_position(&self, asset_id: &str) -> Result<bool> {
+        let conn = self.conn.lock().await;
+        let mut stmt = conn.prepare(
+            "SELECT 1 FROM evm_tokens
+             WHERE identifier = ?
+             AND protocol = 'UNI-V3'
+             LIMIT 1"
+        )?;
```
yabirgb, line 94: question: we do label them right? I didn't find any in my db itself

yabirgb: Isaac confirmed it 
image

colibri/src/icons.rs
```diff
@@ -119,6 +112,96 @@ async fn query_token_icon_and_extension(
     None
 }
 
+async fn query_uniswap_v3_position_icon(
```
LefterisJP, line 115: add a docstring above here

colibri/src/icons.rs
```diff
@@ -119,6 +112,96 @@ async fn query_token_icon_and_extension(
...
+) -> Option<(Bytes, &'static str)> {
+    let token_id: U256 = token_id.parse()
+        .map_err(|e| {
+            error!("Invalid token ID '{}' for NFT position on chain ID {} ({}): {}", 
+            token_id, chain_id, inquirer.blockchain.as_str(), e)
+        })
+        .ok()?;
+
+    for node in inquirer.rpc_nodes.read().await.clone() {
+        let provider = match inquirer.connect_to_node(&node).await {
```
LefterisJP, line 129: So wait ... you have to redo th econnection to a node each time you want to query an icon? Shouldnt' connections be cached somewhere?

colibri/src/icons.rs
LefterisJP: Isnt' tthere a way to distinguish between the error types? There should be.

If it's a badly formatted contract then stop. If the rpc node is failing then continue and try another one

colibri/src/blockchain/node_inquirer.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub async fn update_rpc_nodes(&self) -> Result<(), String> {
+        let nodes = self.globaldb.get_rpc_nodes(self.blockchain)
+            .await
+            .map_err(|e| format!("Failed to get RPC nodes: {}", e))?;
+
+        *self.rpc_nodes.write().await = nodes;
+        Ok(())
+    }
+
+    pub async fn connect_to_node(&self, node: &RpcNode) -> Result<Arc<DynProvider>, String> {
```
LefterisJP, line 43: Add a docstring and rename this to explain what it does. A sugested name.
```suggestion
    pub async fn get_or_create_node_connection(&self, node: &RpcNode) -> Result<Arc<DynProvider>, String> {
```

colibri/src/icons.rs
```diff
@@ -119,6 +112,112 @@ async fn query_token_icon_and_extension(
...
+        })
+        .ok()?;
+
+    for node in inquirer.rpc_nodes.read().await.clone() {
+        let provider = match inquirer.get_or_create_node_connection(&node).await {
+            Ok(p) => p,
+            Err(e) => {
+                // Check if this is a contract-related error
+                let error_message = e.to_string();
+                if error_message.contains("function selector was not recognized") ||
```
LefterisJP, line 135: you should use contains_any. Example code.

```rust
// Using a slice of patterns
let error_message = e.to_string();
let error_patterns = ["function selector was not recognized", 
                      "invalid function signature", 
                      "contract code"];

if error_patterns.iter().any(|&pattern| error_message.contains(pattern)) {
    error!(
        "Contract appears to be malformed or not a valid UniswapV3NFTManager: {} - token ID {} on contract {}",
        e, token_id, contract_address
    );
    break;
} else {
    error!(
        "RPC call to tokenURI failed on node '{}' (endpoint: {}) for token ID {} on contract {}: {}",
        node.name, node.endpoint, token_id, contract_address, e
    );
    continue;
}
```

colibri/src/icons.rs
```diff
@@ -119,6 +112,112 @@ async fn query_token_icon_and_extension(
...
+                    );
+                    continue;
+                }
+            }
+        };
+        let contract = UniswapV3NFTManager::new(contract_address, provider);
+
+        // try to get the token URI
+        let token_uri = match contract.tokenURI(token_id).call().await {
+            Ok(result) => result._0,
```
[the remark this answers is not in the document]
prettyirrelevant: oh, mistake. i added it to the wrong place. nothing contract call is made above. this is where it happens. fixed.

LefterisJP (approved):
LGTM

PR #9582, opened 2025-03-08
Upgrade `coincurve` to version 21.0.0

ofek:
This upgrades `coincurve` to [v21.0.0](https://github.com/ofek/coincurve/releases/tag/v21.0.0) which brings performance improvements and removes all runtime dependencies!


LefterisJP (approved):
Thank you @ofek. Keep up the great work with coincurve!

PR #9585, opened 2025-03-09
chore: fix isinstance() type check syntax error

zkpepe:
Fixed a mistake in `isinstance()` where the `|` operator was used for type union:  

```python
if not isinstance(other, (int | FVal)):  # Incorrect
```  

The `|` operator is valid in type annotations starting from Python 3.9, but `isinstance()` requires a tuple of types. This caused a `SyntaxError` in Python <3.10 and a `TypeError` in newer versions.  

Now it's correct:  

```python
if not isinstance(other, (int, FVal)):  # Fixed
```

### DIFF
--- a/rotkehlchen/accounting/structures/balance.py
+++ b/rotkehlchen/accounting/structures/balance.py
@@ -54,7 +54,7 @@ class Balance:
         )
 
     def __mul__(self, other: Any) -> 'Balance':
-        if not isinstance(other, (int | FVal)):
+        if not isinstance(other, (int, FVal)):
             raise InputError(f'Tried to multiply balance with {type(other)}')
 
         return Balance(


LefterisJP (commented):
Hey @zkpepe that's actually an interesting find.

Can you please:

1. Tell me which tool did you use to detect this error? Which linter?
2. Can you add a small unit test about the multiplication of balances just like the addition one here: https://github.com/rotki/rotki/blob/b8bc648109babfbd48cf8d6ab6294ea763d64cb8/rotkehlchen/tests/unit/test_structures.py#L10-L19

Keep it simple. Make sure to test it work for multiplying a balance by both an FVal and an int.

LefterisJP:
hmm wait that was wrong. From python 3.10 you can use `|` in instance: https://docs.python.org/3.10/whatsnew/3.10.html#pep-604-new-type-union-operator

Issue #9583, opened 2025-03-09
Make P&L Report columns sortable


Barabazs:
## Abstract

The P&L Report table columns are not sortable.

![Image](https://github.com/user-attachments/assets/432d15d5-596d-4699-ab76-6d7aaf53f490)

## Motivation

Consistent behavior with other tables

LefterisJP:
This is the table where the existing PL reports exist right? Not the report table itself.

PR #9591, opened 2025-03-10
feat: allow sorting for reports table

lukicenturi:
Closes https://github.com/rotki/rotki/issues/9583


kelsos (approved):
looks good, we just need the change log entry and we can merge.

PR #9594, opened 2025-03-10
Remove outdated content from api docs

yabirgb:
removes docs for endpoint that we have removed:

- removes compound balances
- removes aave balances
- yearn v1
- yearn v2

Also updates the balances -> amount field transition in history events


prettyirrelevant (commented):
you're missing uniswap/sushiswap events docs too.

prettyirrelevant (approved):

PR #9596, opened 2025-03-10
Correctly handle zero address in openocean

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=100801423

On chains where ETH is the native token the token address in the openocean swap event is `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE`, which is handled correctly by `get_or_create_evm_asset`, but on other chains, the address is simply a zero address when its the native token, which wasn't being handled.

### DIFF
--- a/rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
@@ -4,7 +4,7 @@ from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
 
 from rotkehlchen.chain.ethereum.utils import token_normalized_value_decimals
-from rotkehlchen.chain.evm.constants import DEFAULT_TOKEN_DECIMALS
+from rotkehlchen.chain.evm.constants import DEFAULT_TOKEN_DECIMALS, ZERO_ADDRESS
 from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface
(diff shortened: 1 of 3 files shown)


rotkehlchen/chain/evm/decoding/open_ocean/decoder.py
```diff
@@ -42,9 +42,10 @@ def _get_asset_and_amount(
             raw_amount: int,
     ) -> tuple['CryptoAsset', 'FVal']:
         """Get asset and normalized amount from asset address and raw amount.
-        Specifically handles when asset is native rather than erc20 token.
+        Handles decimals when asset is native rather than erc20 token, and assumes asset is
+        native token if address is ZERO_ADDRESS.
```
LefterisJP, line 46: Is there anywhere in the docs or in the contracts where you can link to instead of saying "Assumes"

nicholasyoder: Unfortunately they don't seem to cover anything about this in their docs, so I've added a link to the relevant contract code.

LefterisJP (approved):
LGTM

LefterisJP:
The failing test should have been fixed on its own PR but has not .... so I am going to just merge now but for the love of god when the build is broken FIX IT. Do not yolo mix the fix in another PR

PR #9599, opened 2025-03-10
feat: added support in backend to sort pnl reports

prodesert22:
## Checklist
Added support in backend to sort pnl reports by pnl amount and asset symbol or name


rotkehlchen/api/v1/schemas.py
```diff
@@ -1717,7 +1717,13 @@ def validate_report_schema(
             data: dict[str, Any],
             **_kwargs: Any,
     ) -> None:
-        valid_ordering_attr = {None, 'timestamp'}
+        valid_ordering_attr = {
+            None,
+            'timestamp',
+            'pnl_amount',
+            'asset',
+        }
```
nicholasyoder, line 1725: Keep this all one line. We usually try to keep stuff fairly compact. For something like this I would break it into multiple lines only after it gets longer than 100 characters and starts triggering the `E501` ruff error during lint.

Also, the backend lint CI step is failing - notice the "Some checks were not successful" message here on your PR. Its good to run `make lint` locally before pushing changes: https://docs.rotki.com/contribution-guides/contribute-as-developer.html#linting-formatting. It looks like there's a number of things that will need adjusted to fix the linting.

rotkehlchen/tests/db/test_reports.py
```diff
@@ -54,3 +111,181 @@ def test_report_settings(database):
...
+    )
+
+    # Using existing EVENTS 
+    for event in EVENTS:
+        dbreport.add_report_data(
+            report_id=report_id,
+            time=event.timestamp,
+            ts_converter=lambda x: str(x),
+            event=event,
+        )
```
nicholasyoder, line 142: Some of the code here in the tests is being repeated in each test. This should probably be refactored to avoid repeating yourself.

prodesert22: Ok fixed

rotkehlchen/api/v1/schemas.py
prodesert22: I noticed it after I send to review, I fixed this error, my mistake

rotkehlchen/tests/db/test_reports.py
```diff
@@ -1,10 +1,76 @@
...
 from rotkehlchen.tests.utils.constants import A_GBP
+from rotkehlchen.types import Location, Price, Timestamp
+from rotkehlchen.utils.misc import create_order_by_rules_list, timestamp_to_date
 
+TIMESTAMP_1_SECS = Timestamp(1741634066)
+TIMESTAMP_2_SECS = Timestamp(1741634100)
+ETH_PRICE_TS_1 = FVal('2000')
+ETH_PRICE_TS_2 = FVal('2200')
+
+EVENTS = [
```
nicholasyoder, line 19:
```suggestion
EVENTS: Final = [
```
Use the `Final` type annotation for global constants like this. Same for the timestamps and prices defined above.

Actually though, if you combine the tests into a single test, this should probably get defined within the test itself. Its always good to avoid global variables when possible.

rotkehlchen/db/reports.py
```diff
@@ -226,14 +227,21 @@ def add_report_data(
         - InputError if the event can not be written to the DB. Probably report id does not exist.
         """
         data = event.serialize_for_db(ts_converter)
+
+        pnl_taxable = str(event.pnl.taxable)
+        try:
+            asset = event.asset.symbol_or_name()
```
yabirgb, line 233: when sorting do we send the symbol? make sure that the schema sorts by symbol and not id

rotkehlchen/tests/db/test_reports.py
```diff
@@ -1,10 +1,26 @@
...
 from rotkehlchen.tests.utils.constants import A_GBP
+from rotkehlchen.types import Location, Price, Timestamp
+from rotkehlchen.utils.misc import create_order_by_rules_list, timestamp_to_date
 
 
-def test_report_settings(database):
+def convert_timestamp(timestamp: Timestamp):
+    """Converts the timestamp to string"""
+
+    return timestamp_to_date(timestamp)
```
yabirgb, line 18: why do you need this wrapper? Just call `timestamp_to_date`

rotkehlchen/db/reports.py
yabirgb, line 233:
```suggestion
            asset_symbol = event.asset.symbol_or_name()
```

rotkehlchen/db/reports.py
```diff
@@ -226,14 +227,21 @@ def add_report_data(
...
+        except WrongAssetType:
+            asset = 'UNKNOWN'
```
yabirgb, line 235: why not just store it as empty string?

yabirgb: Also why not make the schema allow nullable and have them null if we have assets without symbol?

rotkehlchen/tests/db/test_reports.py
```diff
@@ -1,10 +1,26 @@
...
+
+
+def setup_db_account_settings(database):
+    """Setup accounting reports and settings db"""
+
     dbreport = DBAccountingReports(database)
```
yabirgb, line 24: no need from empty lines here

```suggestion
    """Setup accounting reports and settings db"""
    dbreport = DBAccountingReports(database)
```

same between dbreport and settings

rotkehlchen/tests/db/test_reports.py
```diff
@@ -54,3 +77,115 @@ def test_report_settings(database):
...
+                        'order_by_attributes': [test_case['column']],
+                        'ascending': [is_ascending],
+                    },
+                    default_order_by_fields=['timestamp'],
+                    default_ascending=[False],
+                ),
+                report_id=report_id,
+            )
+
+            results, _ = dbreport.get_report_data(filter_query, True)
```
yabirgb, line 183: use kwargs here too

rotkehlchen/db/reports.py
prodesert22: Makes sense store as null, I will change the schema to accept it

rotkehlchen/db/reports.py
prodesert22: Actually is only the [timestamp column](https://github.com/rotki/rotki/blob/develop/frontend/app/src/components/profitloss/ProfitLossEvents.vue#L50) used in frontend, I don't see a [column of asset in the table headers ](https://github.com/rotki/rotki/blob/develop/frontend/app/src/components/profitloss/ProfitLossEvents.vue#L57)but the [api return the asset id](https://github.com/rotki/rotki/blob/develop/rotkehlchen/accounting/structures/processed_event.py#L127) and [the symbol](https://github.com/rotki/rotki/blob/develop/docs/api.rst#get-saved-events-of-a-pnl-report) for my not have problem in leaving the symbol because we can sort is more accurate to the use for example have the token ABC with the id 2 and the token ACB with the id so sorting will appear that it's incorrect

prodesert22: Actually is only the[ timestamp column used](https://github.com/rotki/rotki/blob/develop/frontend/app/src/components/profitloss/ProfitLossEvents.vue#L50) in frontend for sort, and I think the column of asset not is used . Also the [api only returns the asset id](https://github.com/rotki/rotki/blob/develop/rotkehlchen/accounting/structures/processed_event.py#L127) but it will sort by the column and save the id on the asset column can cause the issue eg asset with symbol "acb" with ID 1 and another with symbol "abc" with ID 2

rotkehlchen/tests/db/test_reports.py
```diff
@@ -54,3 +70,122 @@ def test_report_settings(database):
...
+def test_report_events_sort_by_columns(database):
+    """Test that sorting by asset, pnl_taxable and timestamp works correctly"""
+    timestamp_1_secs = Timestamp(1741634066)
+    timestamp_2_secs = Timestamp(1741634100)
+    eth_price_ts_1 = FVal('2000')
+    eth_price_ts_2 = FVal('2200')
+    half_amount = FVal(0.5)
+    hundred = FVal('100')
+
+    dbreport, settings = setup_db_account_settings(database)
```
LefterisJP, line 84: For tests we can do this to save space

```suggestion
    timestamp_1_secs, timestamp_2_secs, eth_price_ts_1, eth_price_ts_2, half_amount, hundred  = Timestamp(1741634066), Timestamp(1741634100), FVal('2000'), FVal('2200'), FVal(0.5), FVal('100')  # noqa: E501
    dbreport, settings = setup_db_account_settings(database)
```

rotkehlchen/tests/db/test_reports.py
```diff
@@ -54,3 +70,122 @@ def test_report_settings(database):
...
+        {
+            'column': 'pnl_taxable',
+            'expected_values': [ZERO, half_amount, eth_price_ts_1],
+            'check_field': lambda event: event.pnl.taxable,
+        },
+        {
+            'column': 'timestamp',
+            'expected_values': [timestamp_1_secs, timestamp_2_secs, timestamp_2_secs],
+            'check_field': lambda event: event.timestamp,
+        },
```
LefterisJP, line 166: styling

```suggestion
        {
            'column': 'asset',
            'expected_values': [A_DAI, A_ETH, A_ETH],
            'check_field': lambda event: event.asset,
        }, {
            'column': 'pnl_free',
            'expected_values': [ZERO, ONE, hundred],
            'check_field': lambda event: event.pnl.free,
        }, {
            'column': 'pnl_taxable',
            'expected_values': [ZERO, half_amount, eth_price_ts_1],
            'check_field': lambda event: event.pnl.taxable,
        }, {
            'column': 'timestamp',
            'expected_values': [timestamp_1_secs, timestamp_2_secs, timestamp_2_secs],
            'check_field': lambda event: event.timestamp,
        },
```

prodesert22:
> Mostly styling stuff. Changelog will go in the frontend part?

Yes, need to change some things on frontend to support this feature

PR #9607, opened 2025-03-11
Add new SwapEvent class and api schema

### LINKED ISSUE #6097: Move the trades fully in history events view

nicholasyoder:
Closes #6097


docs/api.rst
```diff
@@ -5250,15 +5250,50 @@ Dealing with History Events
...
+            }
+
+         :reqjson string location: The location/exchange where the swap occurred
+         :reqjson string spend_amount: The amount being spent
+         :reqjson string spend_asset: The asset being spent (e.g. "USD", "BTC")
+         :reqjson string receive_amount: The amount being received
+         :reqjson string receive_asset: The asset being received (e.g. "USD", "BTC")
+         :reqjson string[optional] fee: The fee amount charged for the swap. If provided, fee_asset must also be provided
+         :reqjson string[optional] fee_asset: The asset in which the fee was paid. If provided, fee must also be provided
+         :reqjson string[optional] unique_id: A unique identifier for this swap
```
LefterisJP, line 5295: what's the difference between unique_id and event_identifier?

rotkehlchen/history/events/structures/swap.py
```diff
@@ -0,0 +1,201 @@
...
+if TYPE_CHECKING:
+    from more_itertools import peekable
+
+    from rotkehlchen.accounting.mixins.event import AccountingEventMixin
+    from rotkehlchen.accounting.pot import AccountingPot
+    from rotkehlchen.fval import FVal
+
+
+class SwapEvent(HistoryBaseEntry):
+    """Swap event representing trades on exchanges."""
```
LefterisJP, line 25: Is this only exchanges?

```suggestion
    """Swap event representing trades on exchanges, defi and more."""
```

docs/api.rst
nicholasyoder: Added a bit more to the descriptions here to hopefully make it clearer what the difference is. Also updated this in the asset movement section.

rotkehlchen/api/rest_helpers/history_events.py
```diff
@@ -34,34 +35,39 @@ def edit_asset_movements(
...
-        )
+        for idx in range(existing_events_num):
+            events_db.edit_history_event(
+                event=events[idx],
+                write_cursor=write_cursor,
+            )
         events_db.add_history_event(
             write_cursor=write_cursor,
-            event=events[1],
+            event=events[-1],
```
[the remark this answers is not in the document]
nicholasyoder: This is when the user manually adds a fee to the asset movement/swap, and when a user adds a new event manually it should be customized.

rotkehlchen/api/v1/schemas.py
```diff
@@ -1012,22 +1024,86 @@ def make_history_base_entry(
...
+                    field_name='fee_amount',
+                )
+
+            if (notes := data.get('notes')) is None:
+                spend_notes, receive_notes, fee_notes = None, None, None
+            elif len(notes) == 2:
+                spend_notes, receive_notes = notes
+                fee_notes = None
+            else:  # len == 3, enforced by validate.Length above
+                spend_notes, receive_notes, fee_notes = notes
```
LefterisJP, line 1061: Two Questions here. 

1.Why not just forward a list of notes to the create_swap_Events()? Same for the asset movement code, as it's the same

2. I think we need some more checks here. If the third node is not None then we also need to make sure that fee exists. Otherwise it's incosistent data.  Same for the assets movement code.

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -115,37 +103,14 @@ def __init__(
             location=location,
             event_type=event_type,
             event_subtype=event_subtype,
-            asset=asset,
+            asset=asset.check_existence(),
```
LefterisJP, line 106: why is this needed? How do you handle errors? This function can raise `UnknownAsset`. It's not caught here. And that means that whenever you call AssetMovement's constryctor you would need to check for this. This is not a good idea and even if you did this, you would have to mention this in the docstring. Explain why this is needed here in a comment too, not only to me in the review.

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -172,9 +137,29 @@ def deserialize_from_db(cls: type['AssetMovement'], entry: tuple) -> 'AssetMovem
...
+            notes=entry[8],
         )
 
+    def serialize(self) -> dict[str, Any]:
+        """Serialize the event for api.
+        Autogenerates the event notes, appending any notes added by the user in a second sentence.
+        """
+        serialized_data = super().serialize()
+        location_name = get_formatted_location_name(self.location)
+        asset_symbol = self.asset.resolve_to_asset_with_symbol().symbol
```
LefterisJP, line 149: Maybe the asset has only name and not symbol?

```suggestion
        asset_symbol = self.asset.symbol_or_name()
```

But both of them can raise `UnknownAsset`. In theory. Can this too? Or since it's an already initialized AssetMovement it should not. If so then specify in a comment to make sure it's understood by the reader.

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -172,9 +137,29 @@ def deserialize_from_db(cls: type['AssetMovement'], entry: tuple) -> 'AssetMovem
...
+        if self.event_subtype == HistoryEventSubType.FEE:
+            notes = f'Pay {self.amount} {asset_symbol} as {location_name} {str(self.event_type).lower()} fee.'  # noqa: E501
+        elif self.event_type == HistoryEventType.DEPOSIT:
+            notes = f'Deposit {self.amount} {asset_symbol} to {location_name}.'
+        else:
```
LefterisJP, line 154:
```suggestion
        else:  # withdrawal
```

rotkehlchen/history/events/structures/swap.py
```diff
@@ -0,0 +1,222 @@
...
+            sequence_index={
+                HistoryEventSubType.SPEND: 0,
+                HistoryEventSubType.RECEIVE: 1,
+                HistoryEventSubType.FEE: 2,
+            }[event_subtype],
+            timestamp=timestamp,
+            location=location,
+            event_type=HistoryEventType.TRADE,
+            event_subtype=event_subtype,
+            asset=asset.check_existence(),
```
LefterisJP, line 68: same comment for check existence as in the asset movement code

rotkehlchen/history/events/structures/swap.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        amount = deserialize_fval(entry[7], 'amount', 'swap event')
+        return cls(
+            identifier=entry[0],
+            event_identifier=entry[1],
+            timestamp=TimestampMS(entry[3]),
+            location=Location.deserialize_from_db(entry[4]),
+            location_label=entry[5],
+            event_subtype=HistoryEventSubType.deserialize(entry[10]),  # type: ignore  # should always be correct from the DB
+            asset=Asset(entry[6]).check_existence(),
```
LefterisJP, line 97: same comment as before for this check existence here. It can raise UnknownAsset in theory and here it even is on the docstring.

But ... since it's deserializing from DB it should not really ever be unknown asset right? Otherwise the DB consistency is broken. So perhaps here it's fine but then needs to be in a comment and maybe also mentioned in the docstring that UnknownAsset and DeseriazliationError should never really happen due to coming straight from the DB.

rotkehlchen/history/events/structures/swap.py
```diff
@@ -0,0 +1,222 @@
...
+            notes=entry[8] or None,
+        )
+
+    def serialize(self) -> dict[str, Any]:
+        """Serialize the event for api.
+        Autogenerates the event notes, appending any notes added by the user in a second sentence.
+        """
+        serialized_data = super().serialize()
+        location_name = get_formatted_location_name(self.location)
+        asset_symbol = self.asset.resolve_to_asset_with_symbol().symbol
```
LefterisJP, line 108: same comment I had for asset movements here for maybe using symbol or name and think on possible errors. If the event does not come from the DB or API then it can be invalid?

rotkehlchen/history/events/structures/swap.py
```diff
@@ -0,0 +1,222 @@
...
+        location: Location,
+        spend_asset: Asset,
+        spend_amount: 'FVal',
+        receive_asset: Asset,
+        receive_amount: 'FVal',
+        fee_asset: Asset,
+        fee_amount: 'FVal',
+        location_label: str | None = None,
+        unique_id: str | None = None,
+        spend_notes: str | None = None,
```
LefterisJP, line 174: as I mentioned in the other part of the code perhaps all notes can be a list. Same as for asset movements

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -673,12 +675,16 @@ def test_add_edit_asset_movements(rotkehlchen_api_server: 'APIServer') -> None:
...
+        assert len(events := db.get_history_events(
             cursor=cursor,
             filter_query=HistoryEventFilterQuery.make(),
             has_premium=True,
             group_by_event_ids=False,
         )) == 3  # including the fee event.
+        assert events[1].event_subtype == HistoryEventSubType.REMOVE_ASSET
+        assert events[1].notes == 'Main event note'
+        assert events[2].event_subtype == HistoryEventSubType.FEE
+        assert events[2].notes == 'Fee event note'
```
LefterisJP, line 687: Check on how this serializes for the api and that it's properly appended.

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -756,3 +762,112 @@ def test_add_edit_asset_movements(rotkehlchen_api_server: 'APIServer') -> None:
...
+            unique_id='TRADE2',
+        ), SwapEvent(
+            identifier=5,
+            timestamp=TimestampMS(1569924576000),
+            location=Location.BITFINEX,
+            event_subtype=HistoryEventSubType.FEE,
+            asset=A_ETH,
+            amount=FVal('0.000004'),
+            unique_id='TRADE2',
+        )]
```
LefterisJP, line 873: check the api serialization and check notes get appended here too

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -2700,14 +2700,14 @@ def test_upgrade_db_45_to_46(user_data_dir: 'Path', messages_aggregator):
...
-            (6, '1cf3fc675d4835efe532d18730db35163b07fb5f44956b888d7aaf852b04fd1c', 0, 1727542684000, 'G', None, 'ETH', '0.07126231', '0', 'Deposit 0.07126231 ETH to Coinbase', 'deposit', 'deposit asset', '{"address": "0xc37b40ABdB939635068d3c5f13E7faF686F03B65", "transaction_id": "0xc8aa60c9f8a93692b66fbc3f57c64f2b1c4afd92f370e12dc7bb23acfa303dbb"}'),  # noqa: E501
-            (6, 'dd253e55b9e3bd672d98f49062a27d6d9946427061b7d36083d78146e6dcce73', 0, 1676524754000, 'B', 'kraken', 'eip155:1/erc20:0x7D1AfA7B718fb893dB30A3aBc0Cfc608AaCfeBB0', '10', '0', 'Withdraw 10 MATIC from Kraken', 'withdrawal', 'remove asset', '{"reference": "KRAKEN-XXX"}'),  # noqa: E501
-            (6, 'dd253e55b9e3bd672d98f49062a27d6d9946427061b7d36083d78146e6dcce73', 1, 1676524754000, 'B', 'kraken', 'eip155:1/erc20:0x7D1AfA7B718fb893dB30A3aBc0Cfc608AaCfeBB0', '0.0100000000', '0', 'Pay 0.0100000000 MATIC as Kraken withdrawal fee', 'withdrawal', 'fee', None),  # noqa: E501
+        ).fetchall() == [  # Don't check notes here since the way asset movements handle notes changes after this upgrade.  # noqa: E501
```
LefterisJP, line 2705: That's for an old upgrade. Don't touch the code here. Old upgrades and old DBs are not affected.

LefterisJP (commented):
Left some comments. Also make sure you have a list of stuff remaining to do later in other PRs.

For example the DB upgrade for the notes we mentioned in our call today.

rotkehlchen/api/v1/schemas.py
nicholasyoder: 1. keeping this as is as mentioned in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1349767797135118437)
2. Changed it to raise a validation error if the fee note is present without a fee amount being present for both swaps and asset movements.

rotkehlchen/history/events/structures/asset_movement.py
nicholasyoder: So I think this makes sense to have here, since the asset needs to be present to prevent unknown asset errors during serialization, and foreign key errors when writing the event to the db. In the exchanges where new asset movements/swaps are initialized, they already check the asset and handle unknown assets, so it should be fine most places this is being used.

The reason I thought of adding this though is actually that after removing `resolve_to_asset_with_symbol` from the init function, the `test_try_premium_at_start_new_account_pull_old_data` test failed like this:
```
rotkehlchen.errors.misc.DBUpgradeError: Failed at database upgrade from version 45 to 46: FOREIGN KEY constraint failed
```
It was skipping an asset movement here: https://github.com/rotki/rotki/blob/develop/rotkehlchen/db/upgrades/v45_v46.py#L125 due to an unknown asset. If no unknown asset error is raised when initializing the asset movement it ends up with a foreign key error when it writes to the db (the foreign key being the asset).

I've added 'may raise' to the docstring and added a comment after `check_existence` for both swaps and asset movements.

rotkehlchen/history/events/structures/asset_movement.py
nicholasyoder: Added a note to the docstring mentioning that it can raise UnknownAsset if for some reason the asset is changed after the event is initialized (bypassing the `check_existence` in the init function)

rotkehlchen/history/events/structures/swap.py
nicholasyoder: see response to other comment.

rotkehlchen/history/events/structures/swap.py
nicholasyoder: see response to other comment.

rotkehlchen/tests/db/test_db_upgrades.py
nicholasyoder: In the v45_to_v46 upgrade, it uses `create_asset_movement_with_fee` to create the asset movements from the old data. So now when we change the way the AssetMovement class handles notes it does in fact affect the old upgrade.
I added a bit more to this comment to try and make clear what happened.

rotkehlchen/history/events/structures/asset_movement.py
```diff
@@ -115,37 +105,14 @@ def __init__(
             location=location,
             event_type=event_type,
             event_subtype=event_subtype,
-            asset=asset,
+            asset=asset.check_existence(),  # Ensure asset exists to prevent errors during serialization and when writing the event to the db.  # noqa: E501
```
LefterisJP, line 108: Hmm then this would mean you now need to do a try/except **every time** you call the `AssetMovement` constructor. I don't think this is a good idea. It's not something we generally do in our code.

And very easy to fuck up. Raising from constructors is generally very tricky as that means every time you initialize something you need to do it in try/except.

Do you currently do it everywhere all over the code? Both for this and the other class constructor you do this? I doubt it.

The way we normally handle those is make the type that's passed to the constructor be something that needs to exist in the DB. So basically not the base `Asset`.

Or am I missing something? Because with the logic you are using we would need to do this **Everywhere** we pass an asset to all these classes and then also do try/except everywhere.

This does not seem right.

rotkehlchen/history/events/structures/swap.py
```diff
@@ -0,0 +1,227 @@
...
+            sequence_index={
+                HistoryEventSubType.SPEND: 0,
+                HistoryEventSubType.RECEIVE: 1,
+                HistoryEventSubType.FEE: 2,
+            }[event_subtype],
+            timestamp=timestamp,
+            location=location,
+            event_type=HistoryEventType.TRADE,
+            event_subtype=event_subtype,
+            asset=asset.check_existence(),  # Ensure asset exists to prevent errors during serialization and when writing the event to the db.  # noqa: E501
```
LefterisJP, line 70: same here. It's not okay

rotkehlchen/history/events/structures/swap.py
LefterisJP: but you still have not removed it. I don't think you should be doign this. This goes against the design of our code.

nicholasyoder: mmm... so `check_existence` is being used in `deserialize_from_db` in the other types of events. See in [EvmEvent](https://github.com/rotki/rotki/blob/develop/rotkehlchen/history/events/structures/evm_event.py#L196), and [HistoryEvent](https://github.com/rotki/rotki/blob/develop/rotkehlchen/history/events/structures/base.py#L500)

In `get_history_events` we handle this UnknownAsset error (https://github.com/rotki/rotki/blob/develop/rotkehlchen/db/history_events.py#L515), log an error, and set a flag that then sends the `Could not deserialize one or more history event(s)` error to the frontend.

So I'm pretty sure this is as it should be.

rotkehlchen/history/events/structures/asset_movement.py
nicholasyoder: ok, so as discussed in discord, I've removed the `check_existence` from the init, and to prevent the test from failing that I mentioned here: https://github.com/rotki/rotki/pull/9607#discussion_r1994004534 I had to add a `check_existence` in the v45_to_v46 db upgrade so it would mimic the way the upgrade was working before this change.

rotkehlchen/history/events/structures/swap.py
nicholasyoder: removed this check

rotkehlchen/history/events/structures/swap.py
nicholasyoder: Leaving this as is as mentioned in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1349831891343118369)

LefterisJP (approved):
LGTM.

The failing tests are due to wrongly VCRed tests by @prettyirrelevant : https://github.com/rotki/test-caching/commit/41c8cea85ab85db5a0142837f93e8721c70e1c0a

Which need to be fixed but not in this PR.

PR #9613, opened 2025-03-12
feat: update PnL report events to include entries_total

prodesert22:
Summary
The entries_found field returns all the events of a report id, this should return all the events that the query requested, e.g. if there is a limit its value should be the same as the limit. And the entries_total field should return the total number of events that this report id has.
Also added tests to the endpoint: `/reports/<int:report_id>/data` .

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -6359,7 +6359,8 @@ Get saved events of a PnL Report
                 "type": "trade"
         }],
         "entries_found": 2,
-        "entries_limit": 1000
+        "entries_total": 10,
+        "entries_limit": 2
        },
        "message": ""
       }
@@ -6377,6 +6378,9 @@ Get saved events of a PnL Report
    :resjson int timestamp: The timestamp this event took place in.
    :resjson str type: The type of event. Can be any of the possible accounting event types.
    :resjson str group_id: Optional. Can be missing. An id signifying events that should be grouped together in the frontend. If missing no grouping needs to happen.
+   :resjson int entries_found: The number of entries found in the current query. This limited by the "limit" field and for free this is limited by FREE_PNL_EVENTS_LIMIT.
+   :resjson int entries_total: The total number of entries in the database for the requested report id, ignoring any filters.
+   :resjson int entries_limit: The limit of entries that can be returned.
 
    :statuscode 200: Report event data was successfully queried.
    :statuscode 400: Report id does not exist.
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -3663,9 +3663,10 @@ class RestAPI:
         if self.rotkehlchen.premium is None:
             with_limit = True
             entries_limit = FREE_PNL_EVENTS_LIMIT
+
         dbreports = DBAccountingReports(self.rotkehlchen.data.db)
         try:
-            report_data, entries_found = dbreports.get_report_data(
+            report_data, entries_found, entries_total = dbreports.get_report_data(
                 filter_=filter_query,
                 with_limit=with_limit,
             )
@@ -3678,6 +3679,7 @@ class RestAPI:
                 export_type=AccountingEventExportType.API,
             ) for x in report_data],
             'entries_found': entries_found,
+            'entries_total': entries_total,
             'entries_limit': entries_limit,
         }
         result_dict = _wrap_in_result(result, '')
--- a/rotkehlchen/db/reports.py
+++ b/rotkehlchen/db/reports.py
@@ -28,9 +28,21 @@ if TYPE_CHECKING:
 def _get_reports_or_events_maybe_limit(
         entry_type: Literal['events'],
         entries_found: int,
+        entries_total: int,
         entries: list[ProcessedAccountingEvent],
         with_limit: bool,
-) -> tuple[list[ProcessedAccountingEvent], int]:
+) -> tuple[list[ProcessedAccountingEvent], int, int]:
+    ...
+
+
+@overload
+def _get_reports_or_events_maybe_limit(
+        entry_type: Literal['events'],
+        entries_found: int,
+        entries_total: None,
+        entries: list[ProcessedAccountingEvent],
+        with_limit: bool,
+) -> tuple[list[ProcessedAccountingEvent], int, None]:
     ...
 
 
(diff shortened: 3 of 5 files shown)


rotkehlchen/api/rest.py
```diff
@@ -3663,9 +3663,13 @@ def get_report_data(self, filter_query: ReportDataFilterQuery) -> Response:
         if self.rotkehlchen.premium is None:
             with_limit = True
             entries_limit = FREE_PNL_EVENTS_LIMIT
+
+        if filter_query.pagination is not None and filter_query.pagination.limit is not None:
+            entries_limit = filter_query.pagination.limit
+
```
yabirgb, line 3669: this one is not needed. Check other cases but this refers to the limit that the user has regarding the amount of entries that can be displayed. This is fine being -1 for premium

rotkehlchen/tests/api/test_reports.py
```diff
@@ -0,0 +1,196 @@
+
```
yabirgb, line 1: remove this first empty line

rotkehlchen/tests/api/test_reports.py
```diff
@@ -0,0 +1,196 @@
...
+from rotkehlchen.types import Location, Price, Timestamp
+from rotkehlchen.utils.misc import timestamp_to_date
+
+
+def setup_report_events(database) -> tuple[int, list[ProcessedAccountingEvent]]:
+    """Input events to the DB for testing"""
+    timestamp_1_secs, timestamp_2_secs, eth_price_ts_1, eth_price_ts_2, half_amount, hundred = Timestamp(1741634066), Timestamp(1741634100), FVal('2000'), FVal('2200'), FVal(0.5), FVal('100')  # noqa: E501
+
+    dbreport = DBAccountingReports(database)
+
```
yabirgb, line 33:
```suggestion
```

remove empty line

rotkehlchen/tests/api/test_reports.py
```diff
@@ -0,0 +1,196 @@
...
+from rotkehlchen.tests.utils.constants import A_GBP
+from rotkehlchen.tests.utils.history import prices
+from rotkehlchen.types import Location, Price, Timestamp
+from rotkehlchen.utils.misc import timestamp_to_date
+
+
+def setup_report_events(database) -> tuple[int, list[ProcessedAccountingEvent]]:
+    """Input events to the DB for testing"""
+    timestamp_1_secs, timestamp_2_secs, eth_price_ts_1, eth_price_ts_2, half_amount, hundred = Timestamp(1741634066), Timestamp(1741634100), FVal('2000'), FVal('2200'), FVal(0.5), FVal('100')  # noqa: E501
+
```
yabirgb, line 31: same remove empty line

rotkehlchen/tests/api/test_reports.py
```diff
@@ -0,0 +1,196 @@
...
+        rotkehlchen_api_server: APIServer,
+) -> None:
+    """Test that getting report data works correctly"""
+    # First create a report and added events
+    report_id, events = setup_report_events(rotkehlchen_api_server.rest_api.rotkehlchen.data.db)
+
+    # Query report data with no filters
+    response = requests.post(
+        api_url_for(rotkehlchen_api_server, 'per_report_data_resource', report_id=report_id),
+        json={},
```
yabirgb, line 117: I believe that you don't need the json key if it is empty

rotkehlchen/tests/api/test_reports.py
```diff
@@ -0,0 +1,196 @@
...
+    assert len(data['result']['entries']) == 1
+    assert data['result']['entries_found'] == 1
+    assert data['result']['entries_total'] == len(events)
+    assert data['result']['entries_limit'] == 1
+
+
+@pytest.mark.parametrize('start_with_valid_premium', [True])
+def test_get_report_data_with_premium(
+        rotkehlchen_api_server: APIServer,
+) -> None:
```
yabirgb, line 147: unless I'm missing something you can use

```suggestion
@pytest.mark.parametrize('start_with_valid_premium', [True, False])
def test_get_report_data_with_premium(
        rotkehlchen_api_server: APIServer,
        start_with_valid_premium: bool,
) -> None:
```

and avoid repeating the test

rotkehlchen/tests/api/test_reports.py
```diff
@@ -0,0 +1,196 @@
...
+    )
+    assert_proper_response(response)
+    data = response.json()
+    assert len(data['result']['entries']) == 1
+    assert data['result']['entries_found'] == 1
+    assert data['result']['entries_total'] == len(events)
+    assert data['result']['entries_limit'] == 1
+
+
+@pytest.mark.parametrize('have_decoders', [True])
```
yabirgb, line 182: do you need to load the decoders here?

rotkehlchen/tests/api/test_reports.py
yabirgb, line 1: I believe we test the reports in another endpoint, can you check if that is the case? It would make sense to keep them together so maybe moving the test that you have there. If we don't have any I guess it is fine to leave them here

rotkehlchen/db/reports.py
```diff
@@ -302,12 +331,15 @@ def get_report_data(
             query = f'SELECT COUNT(*) FROM pnl_events {query}'
             results = cursor.execute(query, bindings).fetchone()
             total_filter_count = results[0]
+            entries_found = len(records)
         else:
             total_filter_count = len(records)
+            entries_found = total_filter_count
```
yabirgb, line 337: entries found in both cases is len(records) right? you can define it outside the if/else and then in this else branch total_filter_count = entries_found

rotkehlchen/tests/api/test_reports.py
prodesert22: no, I forgot to remove

rotkehlchen/tests/api/test_reports.py
prodesert22: Ah ok thanks

docs/api.rst
```diff
@@ -6377,6 +6378,9 @@ Get saved events of a PnL Report
    :resjson int timestamp: The timestamp this event took place in.
    :resjson str type: The type of event. Can be any of the possible accounting event types.
    :resjson str group_id: Optional. Can be missing. An id signifying events that should be grouped together in the frontend. If missing no grouping needs to happen.
+   :resjson int entries_found: The number of entries found in the current query. This limited by the "limit" field and for free this is limited by FREE_PNL_EVENTS_LIMIT.
```
LefterisJP, line 6381:
```suggestion
   :resjson int entries_found: The number of entries found in the current query. This is limited by the "limit" field. Additionally for free users this is limited by FREE_PNL_EVENTS_LIMIT.
```

rotkehlchen/tests/api/test_reports.py
```diff
@@ -0,0 +1,149 @@
...
+    return report_id, events
+
+
+@pytest.mark.parametrize('start_with_valid_premium', [True, False])
+def test_get_report_data_with_premium(
+        rotkehlchen_api_server: APIServer,
+        start_with_valid_premium: bool,
+) -> None:
+    """Test that getting report data works correctly with premium subscription active"""
+    # First create a report and added events
```
LefterisJP, line 106:
```suggestion
    # First create a report and add events
```

LefterisJP (approved):
LGTM

PR #9619, opened 2025-03-13
fix: convert pnl_taxable and pn_free to real to fix the sort

prodesert22:
Need to conver the pnl_taxable and pn_free to real to fix the sort
image


LefterisJP (approved):
LGTM

PR #9620, opened 2025-03-13
feat: add validator label in CSV export

### LINKED ISSUE #9385: Add "label" and "tags" support for Ethereum validators
## Abstract

In Rotki v1.37.1, when an `export accounts` is executed 
(as described in Rotki help at : [https://docs.rotki.com/usage-guides/accounts-and-balances.html#import-and-export-blockchain-accounts-csv ]()), all Ethereum accnts are exported with their `label` and `tags` data intact.  
However, Ethereum validators, which are included in the exported .csv file indexed by their validator hash (and not the withdrawl addr, or other Ethereum address) are null in those fields.

## Motivation

I use the new/recent `export accounts` feature to export all my blockchain accnts for my tax professional to use in tax filing.  All Ethereum accnts come out w label (which is the name I've assigned in Rotki to each accnt) and tag.  The tax person and I use those label names to refer to my accnts, rather than obtuse hexadecimal numbers.  
But the Ethereum validators come with nothing helpful, other than a very long unreadable hexadecimal value.  

## Specification

Allow each hex validator number to DEFAULT the label to the Ethereum validator number and a bit of prose:
  for example:  label = `Ethereum validator no. 342xxx`    OR   `Ethereum validator no. 165xxxx  (32 ETH balance)`
NOTE:  Rotki already knows the validator no. and surfaces it in the Staking menu, calling it the more esoteric term: "index".

Alternative:  just allow the Rotki end user to set a name for each validator, like they can an Ethereum accnt, and add a tag to that validator.  In this case, it would not be necessary for Rotki to find the accnt no. and add the text to make it human readable.

lukicenturi:
Closes #9385

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`9385` Users will see the default label using the validator index when exporting validators.
 * :bug:`-` The "Ignore Errors" setting in the snapshot menu will persist after the user moves to another page or logs out.
 * :bug:`-` Some Gnosis pay transactions that were missing merchant data should now redecode again properly.
 * :bug:`-` "Add transaction by tx hash" is now reusable again after errors.
--- a/frontend/app/auto-imports.d.ts
+++ b/frontend/app/auto-imports.d.ts
@@ -526,7 +526,6 @@ declare module 'vue' {
     readonly createPinia: UnwrapRef<typeof import('pinia')['createPinia']>
     readonly createProjection: UnwrapRef<typeof import('@vueuse/math')['createProjection']>
     readonly createReactiveFn: UnwrapRef<typeof import('@vueuse/core')['createReactiveFn']>
-    readonly createRef: UnwrapRef<typeof import('@vueuse/core')['createRef']>
     readonly createReusableTemplate: UnwrapRef<typeof import('@vueuse/core')['createReusableTemplate']>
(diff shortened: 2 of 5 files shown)


docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`-` Users will see the default label using the validator index when exporting validators.
```
kelsos, line 5:
```suggestion
* :feature:`9385` Users will see the default label using the validator index when exporting validators.
```

frontend/app/auto-imports.d.ts
```diff
@@ -526,7 +526,6 @@ declare module 'vue' {
     readonly createPinia: UnwrapRef<typeof import('pinia')['createPinia']>
     readonly createProjection: UnwrapRef<typeof import('@vueuse/math')['createProjection']>
     readonly createReactiveFn: UnwrapRef<typeof import('@vueuse/core')['createReactiveFn']>
-    readonly createRef: UnwrapRef<typeof import('@vueuse/core')['createRef']>
```
kelsos, line 529: I need to think the best way to avoid this happening, the removal here is correct, but the original addition was mistaken because this is from a newer version that was not in `bugfixes` I think since we removed almost all of our code from the auto-imports, it should be possible to add a check here, so that this happens only if `pnpm-lock.yaml` is modified

PR #9625, opened 2025-03-13
fix: distinguish USD, CAD, NZD, AUD symbols

### LINKED ISSUE #9624: Make it obvious when I have CAD vs. USD as display currency
Currently there is nothing visually differentiating between CAD and USD in the UI. I have to click on the currency setting to see what I'm using (and even that doesn't make it obvious.

Suggestion: 

Show CAD price with a C$ instead of $ prefix.

kelsos:
Closes #9624

## Checklist


LefterisJP (approved):
LGTM. So we only have those 3 with $ except for the US dollar. The wikipedia page lists a ton more but we don't have them selectable yet :)

https://en.wikipedia.org/wiki/Currency_symbol

PR #9631, opened 2025-03-14
feat: added rainbow decoder

prodesert22:
## Summary
- Added rainbow decoder, it works enricher the transfers to rainbow router and post deconding the events to create the fee event.
For now the decoder only decode transactions from **ethereum mainnet**.

### DIFF
--- a/rotkehlchen/chain/evm/decoding/rainbow/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/rainbow/decoder.py
@@ -1,0 +1,408 @@
+import logging
+from collections.abc import Callable
+from typing import TYPE_CHECKING, Optional
+
+from rotkehlchen.assets.asset import CryptoAsset
+from rotkehlchen.chain.ethereum.utils import asset_normalized_value, asset_raw_value
+from rotkehlchen.chain.evm.decoding.constants import ERC20_OR_ERC721_TRANSFER
+from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface
+from rotkehlchen.chain.evm.decoding.structures import (
+    FAILED_ENRICHMENT_OUTPUT,
+    EnricherContext,
+    TransferEnrichmentOutput,
+)
+from rotkehlchen.chain.evm.decoding.types import CounterpartyDetails
+from rotkehlchen.chain.evm.decoding.utils import maybe_reshuffle_events
+from rotkehlchen.chain.evm.decoding.weth.decoder import WETH_DEPOSIT_TOPIC, WETH_WITHDRAW_TOPIC
+from rotkehlchen.chain.evm.structures import EvmTxReceiptLog
+from rotkehlchen.errors.asset import UnsupportedAsset, WrongAssetType
+from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.logging import RotkehlchenLogsAdapter
+from rotkehlchen.types import ChecksumEvmAddress, EvmTransaction
+from rotkehlchen.utils.misc import bytes_to_address
+
+from .constants import CPT_RAINBOW_SWAPS
+
+if TYPE_CHECKING:
+    from rotkehlchen.chain.evm.decoding.base import BaseDecoderTools
+    from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
+    from rotkehlchen.history.events.structures.evm_event import EvmEvent
+    from rotkehlchen.user_messages import MessagesAggregator
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+class RainbowCommonDecoder(DecoderInterface):
+
+    def __init__(
+            self,
+            evm_inquirer: 'EvmNodeInquirer',
+            base_tools: 'BaseDecoderTools',
+            msg_aggregator: 'MessagesAggregator',
+            router_address: ChecksumEvmAddress,
+    ) -> None:
+        super().__init__(evm_inquirer, base_tools, msg_aggregator)
+        self.router_address = router_address
+
+    def _maybe_set_event_properties(
+            self,
+            event: 'EvmEvent',
+            sender: ChecksumEvmAddress,
+    ) -> None:
+        """Set properties of the event to a rainbow swap considering the event direction"""
+        if (
+            event.location_label != sender or
+            event.address != self.router_address or
+            event.counterparty is not None
+        ):
+            return
+        if event.event_type == HistoryEventType.RECEIVE and event.event_subtype == HistoryEventSubType.NONE:  # noqa: E501
+            # Find the receive event
+            event.event_type = HistoryEventType.TRADE
+            event.event_subtype = HistoryEventSubType.RECEIVE
+            event.counterparty = CPT_RAINBOW_SWAPS
+            event.notes = f'Receive {event.amount} {event.asset.resolve_to_asset_with_symbol().symbol} as the result of a swap in Rainbow'  # noqa: E501
+        elif event.event_type == HistoryEventType.SPEND and event.event_subtype == HistoryEventSubType.NONE:  # noqa: E501
+            # Find the spend event
+            event.event_type = HistoryEventType.TRADE
+            event.event_subtype = HistoryEventSubType.SPEND
+            event.notes = f'Swap {event.amount} {event.asset.resolve_to_asset_with_symbol().symbol} in Rainbow'  # noqa: E501
+            event.counterparty = CPT_RAINBOW_SWAPS
+            event.address = self.router_address
+
+    def _find_event_by_log(
+            self,
+            logs: list['EvmTxReceiptLog'],
+            event_signature: bytes,
+    ) -> EvmTxReceiptLog | None:
+        """
+        Find the log that matches  the provided event signature.
+
+        Returns the log event if found or None otherwise.
+        """
+        for tx_log in logs:
+            if tx_log.topics[0] == event_signature:
+                return tx_log
+
+        return None
+
+    def _get_in_out_events(
+            self,
+            events: list['EvmEvent'],
+            sender: ChecksumEvmAddress,
+    ) -> tuple[Optional['EvmEvent'], Optional['EvmEvent']]:
+        """
+        Extracts 'in' and 'out' events for a sender from a list of events.
+
+        Args:
+            events: List of EVM events.
+            sender: Sender address to filter events by.
+
+        Returns:
+            Tuple of (in_event, out_event), representing token receive and spend
+            events respectively.
+        """
+        in_event = out_event = None
+        for event in events:
+            if (
+                event.event_subtype == HistoryEventSubType.NONE and
+                event.asset == self.evm_inquirer.native_token
+            ):
+                # The "send eth" event is not enriched by enricher_rules,
+                # so it needs to be enriched here
+                self._maybe_set_event_properties(event, sender)
+
+            if (
+                event.event_type == HistoryEventType.TRADE and
+                event.event_subtype == HistoryEventSubType.RECEIVE and
+                event.counterparty == CPT_RAINBOW_SWAPS
+            ):
+                in_event = event
+            elif (
+                event.event_type == HistoryEventType.TRADE and
+                event.event_subtype == HistoryEventSubType.SPEND and
+                event.counterparty == CPT_RAINBOW_SWAPS
+            ):
+                out_event = event
+
+        return in_event, out_event
+
+    def _create_and_append_fee_event(
+            self,
+            transaction: EvmTransaction,
+            fee_raw: int,
+            fee_asset: CryptoAsset,
+            sender: ChecksumEvmAddress,
+            decoded_events: list['EvmEvent'],
+            in_event: Optional['EvmEvent'],
+            out_event: Optional['EvmEvent'],
+    ) -> None:
+        """Create and append a fee event to the decoded events list."""
+        fee_event = self.base.make_event(
+            tx_hash=transaction.tx_hash,
+            sequence_index=self.base.get_next_sequence_index(),
+            timestamp=transaction.timestamp,
+            event_type=HistoryEventType.SPEND,
+            event_subtype=HistoryEventSubType.FEE,
+            asset=fee_asset,
+            amount=(fee_amount := asset_normalized_value(amount=fee_raw, asset=fee_asset)),
+            location_label=sender,
+            notes=f'Spend {fee_amount} {fee_asset.resolve_to_asset_with_symbol().symbol} a Rainbow fee',  # noqa: E501
+            counterparty=CPT_RAINBOW_SWAPS,
+            address=self.router_address,
+        )
+        decoded_events.append(fee_event)
+        maybe_reshuffle_events(
+            ordered_events=[out_event, in_event, fee_event],
+            events_list=decoded_events,
+        )
+
+    def _maybe_decode_fee_from_swap_native_to_token(
+            self,
+            transaction: EvmTransaction,
+            out_event: 'EvmEvent',
+            in_event: 'EvmEvent',
+            decoded_events: list['EvmEvent'],
+            sender: ChecksumEvmAddress,
+            wrap_event: EvmTxReceiptLog,
+    ) -> None:
+        """
+        Decodes the fee from a swap transaction involving native token to another token.
+
+        The fee is calculated as the difference between the amount of native token sent
+        and the amount of native token wrapped into the token.
+        """
+        try:
+            swap_amount = asset_raw_value(out_event.amount, self.evm_inquirer.native_token)
+        except UnsupportedAsset:
+            log.error(
+                f'Failed to decode rainbow swap from {self.evm_inquirer.native_token} to token'
+                f' for transaction {transaction} '
+                f'due to unsupported asset {out_event.asset.identifier}.',
+            )
+            return
+
+        self._create_and_append_fee_event(
+            transaction=transaction,
+            fee_raw=swap_amount - int.from_bytes(wrap_event.data),
+            fee_asset=self.evm_inquirer.native_token,
+            sender=sender,
+            decoded_events=decoded_events,
+            in_event=in_event,
+            out_event=out_event,
+        )
+
+    def _maybe_decode_fee_from_swap_token_to_native(
+            self,
+            transaction: EvmTransaction,
+            decoded_events: list['EvmEvent'],
+            in_event: Optional['EvmEvent'],
+            out_event: Optional['EvmEvent'],
+            sender: ChecksumEvmAddress,
+            unwrap_event: EvmTxReceiptLog,
+    ) -> None:
+        """
+        Decodes the fee from a swap transaction involving token to native token.
+
+        The fee is calculated as the difference between the amount unwrapped to native token
+        and the amount of native token received by the user.
+        """
+        if in_event is None or out_event is None:
+            log.error(
+                f'Failed to decode fee from rainbow swap token to '
+                f'{self.evm_inquirer.native_token} for transaction {transaction}.',
+            )
+            return
+
+        try:
+            in_amount = asset_raw_value(in_event.amount, self.evm_inquirer.native_token)
+        except UnsupportedAsset:
+            log.error(
+                f'Failed to decode fee from rainbow swap token to '
+                f'{self.evm_inquirer.native_token} for transaction {transaction} '
+                f'due to unsupported asset {in_event.asset.identifier}.',
+            )
+            in_amount = 0
+
+        self._create_and_append_fee_event(
+            transaction=transaction,
+            fee_raw=int.from_bytes(unwrap_event.data) - in_amount,
+            fee_asset=self.evm_inquirer.native_token,
+            sender=sender,
+            decoded_events=decoded_events,
+            in_event=in_event,
+            out_event=out_event,
+        )
+
+    def _maybe_decode_fee_from_swap_token_to_token(
+            self,
+            transaction: EvmTransaction,
+            decoded_events: list['EvmEvent'],
+            all_logs: list[EvmTxReceiptLog],
+            sender: ChecksumEvmAddress,
+            out_event: 'EvmEvent',
+            in_event: 'EvmEvent',
+    ) -> None:
+        """
+        Decodes the fee from a swap transaction involving token to token.
+
+        The function identifies the fee by analyzing the transfer logs:
+        - It looks for the transfer log from the user's wallet to the router
+        to determine the sell amount.
+        - It looks for the transfer log from the router to the swap provider
+        to determine the swap amount.
+        - The fee is calculated as the difference between the sell amount and the swap amount.
+        """
+        try:
+            token_address = out_event.asset.resolve_to_evm_token().evm_address
+        except WrongAssetType:
+            log.error(
+                f'Failed to decode fee from rainbow swap token to token '
+                f'for transaction {transaction} due to token_address being null.',
+            )
+            return
+
+        sell_amount = swap_amount = None
+        for tx_log in all_logs:
+            if (
+                tx_log.topics[0] == ERC20_OR_ERC721_TRANSFER and
+                tx_log.address == token_address
+            ):
+                if bytes_to_address(tx_log.topics[2]) == self.router_address:
+                    sell_amount = int.from_bytes(tx_log.data)
+                elif bytes_to_address(tx_log.topics[1]) == self.router_address:
+                    swap_amount = int.from_bytes(tx_log.data)
+
+                if swap_amount and sell_amount:
+                    break
+        else:
+            log.error(
+                f'Failed to decode fee from rainbow swap token to token '
+                f'for transaction {transaction} due to sell_amount '
+                'or swap_amount being null.',
+            )
+            return
+
+        self._create_and_append_fee_event(
+            transaction=transaction,
+            fee_raw=sell_amount - swap_amount,
+            fee_asset=out_event.asset.resolve_to_evm_token(),
+            sender=sender,
+            decoded_events=decoded_events,
+            in_event=in_event,
+            out_event=out_event,
+        )
+
+    def _process_swap(self,
+            transaction: 'EvmTransaction',
+            decoded_events: list['EvmEvent'],
+            all_logs: list['EvmTxReceiptLog'],
+    ) -> list['EvmEvent']:
+        """
+        Process a swap transaction by decoding the relevant events and logs.
+
+        The function handles different types of swaps:
+        - Native token to EVM token swap, looking for the output asset is
+        eth and input asset is token
+        - EVM token to native token swap, looking for the output asset is
+        token and input asset is eth
+        - EVM token to EVM token swap, looking for input and output assets are tokens
+
+        It attempts to find the necessary logs for wrapping and unwrapping tokens
+        and decodes the swap accordingly.
+        If the required logs are missing, it logs an error and returns
+        the original decoded events.
+        """
+        in_event, out_event = self._get_in_out_events(
+            events=decoded_events,
+            sender=(sender := transaction.from_address),
+        )
+        if in_event is None or out_event is None:
+            log.error(
+                f'Failed to post decode rainbow {transaction} '
+                f'due to missing in_event or out_event.',
+            )
+            return decoded_events
+
+        out_asset_is_evm_token = out_event.asset.is_evm_token()
+        in_asset_is_evm_token = in_event.asset.is_evm_token()
+
+        if out_event.asset == self.evm_inquirer.native_token and in_asset_is_evm_token:
+            # See if we have a eth into weth log event
+            # TODO: Change this logic to get the amount of eth wrapped
+            if (wrap_event := self._find_event_by_log(all_logs, WETH_DEPOSIT_TOPIC)) is None:
+                log.error(
+                    f'Failed to decode fee from rainbow swap involving '
+                    f'{self.evm_inquirer.native_token} to token for transaction '
+                    f'{transaction} due to missing wrap log.',
+                )
+                return decoded_events
+
+            self._maybe_decode_fee_from_swap_native_to_token(
+                transaction=transaction,
+                out_event=out_event,
+                in_event=in_event,
+                decoded_events=decoded_events,
+                sender=sender,
+                wrap_event=wrap_event,
+            )
+        elif out_asset_is_evm_token and in_event.asset == self.evm_inquirer.native_token:
+            # TODO: Change this logic to get the amount of eth unwrapped
+            if (unwrap_event := self._find_event_by_log(all_logs, WETH_WITHDRAW_TOPIC)) is None:
+                log.error(
+                    f'Failed to decode fee from rainbow swap involving'
+                    f'token to {self.evm_inquirer.native_token} for transaction '
+                    f'{transaction} due to missing unwrap log.',
+                )
+                return decoded_events
+
+            self._maybe_decode_fee_from_swap_token_to_native(
+                transaction=transaction,
+                decoded_events=decoded_events,
+                in_event=in_event,
+                out_event=out_event,
+                sender=sender,
+                unwrap_event=unwrap_event,
+            )
+        elif out_asset_is_evm_token and in_asset_is_evm_token:
+            self._maybe_decode_fee_from_swap_token_to_token(
+                transaction=transaction,
+                decoded_events=decoded_events,
+                all_logs=all_logs,
+                sender=sender,
+                out_event=out_event,
+                in_event=in_event,
+            )
+
+        return decoded_events
+
+    def _maybe_enrich_rainbow_transfer(
+            self,
+            context: EnricherContext,
+    ) -> TransferEnrichmentOutput:
+        """
+        Enricher transfer events to rainbow router. The contract doesn't emit events for swaps.
+        """
+        if context.event.address != self.router_address:
+            return FAILED_ENRICHMENT_OUTPUT
+
+        self._maybe_set_event_properties(context.event, context.transaction.from_address)
+        if context.event.counterparty == CPT_RAINBOW_SWAPS:
+            return TransferEnrichmentOutput(matched_counterparty=CPT_RAINBOW_SWAPS)
+
+        return FAILED_ENRICHMENT_OUTPUT
+
+    def enricher_rules(self) -> list[Callable]:
+        return [self._maybe_enrich_rainbow_transfer]
+
+    def post_decoding_rules(self) -> dict[str, list[tuple[int, Callable]]]:
+        return {CPT_RAINBOW_SWAPS: [(0, self._process_swap)]}
+
+    @staticmethod
+    def counterparties() -> tuple[CounterpartyDetails, ...]:
+        return (CounterpartyDetails(
+            identifier=CPT_RAINBOW_SWAPS,
+            label='rainbow swaps',
+            image='rainbow.svg',
+        ),)
--- a/rotkehlchen/tests/unit/decoders/test_rainbow.py
+++ b/rotkehlchen/tests/unit/decoders/test_rainbow.py
@@ -1,0 +1,219 @@
+from typing import Final
+
+import pytest
+
+from rotkehlchen.assets.asset import Asset
+from rotkehlchen.chain.evm.decoding.constants import CPT_GAS
+from rotkehlchen.chain.evm.decoding.rainbow.constants import (
+    CPT_RAINBOW_SWAPS,
+    RAINBOW_ROUTER_CONTRACT,
+)
+from rotkehlchen.constants.assets import A_ETH
+from rotkehlchen.fval import FVal
+from rotkehlchen.history.events.structures.evm_event import EvmEvent
+from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
+from rotkehlchen.tests.utils.ethereum import get_decoded_events_of_transaction
+from rotkehlchen.types import Location, TimestampMS, deserialize_evm_tx_hash
+
+A_AIT: Final = Asset('eip155:1/erc20:0x89d584A1EDB3A70B3B07963F9A3eA5399E38b136')
+A_BUOY: Final = Asset('eip155:1/erc20:0x289Ff00235D2b98b0145ff5D4435d3e92f9540a6')
+A_AXGT: Final = Asset('eip155:1/erc20:0xDd66781D0E9a08D4FBb5eC7BAc80B691BE27F21D')
+A_ZIG: Final = Asset('eip155:1/erc20:0xb2617246d0c6c0087f18703d576831899ca94f01')
+
+
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('ethereum_accounts', [['0xfe25d6300D33e19E15AeeFEFD0Aafb319Dd61ae1']])
+def test_rainbow_swap_eth_to_token(ethereum_inquirer, ethereum_accounts):
+    tx_hash = deserialize_evm_tx_hash('0xcd6afacf1efce38186b6cf9164792c4034d33d3f071c5e28c0b79ce5ab0223a9')  # noqa: E501
+    events, _ = get_decoded_events_of_transaction(evm_inquirer=ethereum_inquirer, tx_hash=tx_hash)
+    swap_amount, received_amount, gas_fees, fee_amount, timestamp, user_address = '0.2', '28827.267041421686554081', '0.000114360530468618', '0.00199745', TimestampMS(1741808351000), ethereum_accounts[0]  # noqa: E501
+    expected_events = [EvmEvent(
+        tx_hash=tx_hash,
+        sequence_index=0,
+        timestamp=timestamp,
+        location=Location.ETHEREUM,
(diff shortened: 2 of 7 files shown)


rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,261 @@
...
+
+        return self._decode_swap(
+            context=context,
+            fee_raw=fee_raw,
+            in_asset=buy_asset,
+            in_amount=buy_amount,
+        )
+
+    def _decode_swap_token_to_native(self, context: DecoderContext) -> DecodingOutput:
+        tx_log = context.tx_log
```
yabirgb, line 167: this needs a docstring explaining what it does, also don't assing a variable only for tx_log

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,261 @@
...
+                    event.event_type = HistoryEventType.TRADE
+                    event.event_subtype = HistoryEventSubType.RECEIVE
+                    event.counterparty = CPT_RAINBOW_SWAPS
+                    event.notes = f'Receive {event.amount} {event.asset.resolve_to_asset_with_symbol().symbol} as the result of a Rainbow Swap'  # noqa: E501
+                    event.address = self.router_address
+                    in_event = event
+
+                    # Find the log withdraw the weth token
+                    for log in context.all_logs:
+                        if log.topics[0] == WETH_WITHDRAW_TOPIC:
```
yabirgb, line 185: as a rule if you are looping and trying to find something when you find it break the loop so you don't waste time. Also in python loops have an `else` statement where you can log if you didn't find what you were loooking for

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,261 @@
...
+                            # The fee will be the withdraw data subtract the in_amount
+                            fee_raw = int.from_bytes(log.data) - in_amount
+
+            if not (fee_raw and in_event):
+                return DEFAULT_DECODING_OUTPUT
+
+            sell_amount_raw = int.from_bytes(tx_log.data)
+            sell_asset = self.base.get_or_create_evm_token(tx_log.address)
+            amount = asset_normalized_value(sell_amount_raw, sell_asset)
+            tx_hash = context.transaction.tx_hash
```
yabirgb, line 199: don't create a variable for this

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,261 @@
...
+                event_subtype=HistoryEventSubType.SPEND,
+                asset=sell_asset,
+                amount=amount,
+                location_label=sender,
+                notes=f'Swap {amount} {sell_asset.resolve_to_asset_with_symbol().symbol} in Rainbow Swap',  # noqa: E501
+                counterparty=CPT_RAINBOW_SWAPS,
+                address=self.router_address,
+            )
+            # Need to change the index of receive to respect the events order
+            in_event.sequence_index += 1
```
yabirgb, line 215: instead of doing this check the maybe_reshuffle, not sure if it is what you need but just in case. Also we have a function `get_next_sequence_index` that you can use. This one looks chaotic and error prone

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,336 @@
...
+    def _resolve_asset(self, asset: Asset) -> EvmToken | CryptoAsset:
+        """Resolve a asset into evmtoken or cryptoasset
+
+        Args:
+            asset (Asset): Asset to be resolve
+
+        Returns:
+            EvmToken | CryptoAsset: Asset resolved
+        """
+        return self.evm_inquirer.native_token if asset == A_ETH else asset.resolve_to_evm_token()
```
yabirgb, line 52: remove this function. I believe is wrong and not needed

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,336 @@
...
+        Returns:
+            in_event (EvmEvent | None): input event, the receive the tokens
+            out_event (EvmEvent | None): ouput event, the spend the tokens
+        """
+        in_event = out_event = None
+        for event in decoded_events:
+            if (event.event_type == HistoryEventType.INFORMATIONAL and event.event_subtype == HistoryEventSubType.APPROVE):
+                # Remove the approve from rainbow router to aggretor
+                # to avoid show unnecessary events 
+                decoded_events.remove(event)
```
yabirgb: don't do this. We always leave them as they are useful to identify infinite approvals and perform other analytics

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,336 @@
...
+            if (event.event_type == HistoryEventType.RECEIVE and event.event_subtype == HistoryEventSubType.NONE and event.location_label == sender):  # noqa: E501
+                event.event_type = HistoryEventType.TRADE
+                event.event_subtype = HistoryEventSubType.RECEIVE
+                event.counterparty = CPT_RAINBOW_SWAPS
+                event.notes = f'Receive {event.amount} {event.asset.resolve_to_asset_with_symbol().symbol} as the result of a Rainbow Swap'  # noqa: E501
+                event.address = self.router_address
+                in_event = event
+
+            # Find the spend event
+            if (event.event_type == HistoryEventType.SPEND and event.event_subtype == HistoryEventSubType.NONE and event.location_label == sender):  # noqa: E501
```
yabirgb, line 81: this should be an elif

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,336 @@
...
+                event.event_type = HistoryEventType.TRADE
+                event.event_subtype = HistoryEventSubType.SPEND
+                event.notes = f'Swap {event.amount} {event.asset.resolve_to_asset_with_symbol().symbol} in Rainbow Swap'  # noqa: E501
+                event.counterparty = CPT_RAINBOW_SWAPS
+                event.address = self.router_address
+                out_event = event
+
```
yabirgb, line 88: what if any of them is None? You should log if something is off

rotkehlchen/data/eth_abi.json
yabirgb, line 1: This file shouldn't have any change. Remember to use the globaldb

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,336 @@
...
+        sender = transaction.from_address
+
+        # Check if is calling the rainbow router
+        if transaction.to_address != self.router_address:
+            return DEFAULT_DECODING_OUTPUT
+
+        # Check if the first log is wrap the native token
+        # and the last log is the transfer the token to the user
+        wrap_log = all_logs[0]
+        transfer_log = all_logs[-1]
```
yabirgb, line 145: don't trust this because when using a safe or calling other contracts this can be off. For example the first event in safe transactions is the safe itself executing the call

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,336 @@
...
+    ) -> DecodingOutput:
+        """
+        Decodes a swap transaction from native token to another token.
+        Check if the first log is deposit and the last is transfer log
+        """
+        sender = transaction.from_address
+
+        # Check if is calling the rainbow router
+        if transaction.to_address != self.router_address:
+            return DEFAULT_DECODING_OUTPUT
```
yabirgb: with safes the to is the safe address so don't make assumptions on the target of the transaction

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,336 @@
...
+        ):
+            return DEFAULT_DECODING_OUTPUT
+
+        _, out_event = self._tansform_decoded_events(decoded_events, sender)
+
+        if out_event is None:
+            return DEFAULT_DECODING_OUTPUT
+
+        # The fee amount is the difference of msg.value and the wrapped token amount
+        fee_raw = transaction.value - int.from_bytes(wrap_log.data)
```
yabirgb, line 151: this is not 100% safe since a single transaction can do several things

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,388 @@
...
+        # to create a new event with the right sequence index
+        remove_action = ActionItem(
+            action='skip',
+            from_event_type=HistoryEventType.RECEIVE,
+            from_event_subtype=HistoryEventSubType.NONE,
+            asset=buy_asset,
+        )
+
+        tx_hash = transaction.tx_hash
+        timestamp = transaction.timestamp
```
yabirgb, line 168: don't make variables for this

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,388 @@
...
+
+        # Remove the transfer event
+        # of the purchased token to the user's wallet
+        # to create a new event with the right sequence index
+        remove_action = ActionItem(
+            action='skip',
+            from_event_type=HistoryEventType.RECEIVE,
+            from_event_subtype=HistoryEventSubType.NONE,
+            asset=buy_asset,
+        )
```
yabirgb, line 165: instead of this you can use the reshuffle function that we have

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,388 @@
...
+            sequence_index (int): event sequence index
+            timestamp (int): transaction timestamp
+            fee_raw (int): amount of fee charged in wei
+            fee_asset (Asset): fee asset
+            sender (ChecksumEvmAddress): sender address
+        """
+        fee_amount = asset_normalized_value(amount=fee_raw, asset=self._resolve_asset(fee_asset))
+        return self.base.make_event(
+            tx_hash=tx_hash,
+            sequence_index=sequence_index,
```
yabirgb, line 118: instead of this check `make_event_next_index`. In general try to avoid messing with the sequence index manually since you can reach weird states easily

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,388 @@
...
+        ):
+            return DEFAULT_DECODING_OUTPUT
+
+        if not (in_event and out_event):
+            return DEFAULT_DECODING_OUTPUT
+
+        try:
+            in_amount = asset_raw_value(in_event.amount, native_token)
+        except UnsupportedAsset:
+            in_amount = 0
```
yabirgb, line 223: this error can happen in theory but not here. In any case you should log an error if it happens

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,388 @@
...
+        # The fee will be the withdraw data subtract the in_amount
+        fee_raw = int.from_bytes(unwrap_log.data) - in_amount
```
yabirgb, line 225: you would need to find a reliable way to find the unwrap log event. I believe you can check for transfers to the router contract as those are the initiators of the logic

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,388 @@
...
+        decoded_events: list['EvmEvent'],
+        all_logs: list[EvmTxReceiptLog],
+        sender: ChecksumEvmAddress,
+        out_event: Optional['EvmEvent'],
+    ) -> DecodingOutput:
+        """Decodes a swap transaction from token to native token.
+        Check if the first and last logs is transfer
+        """
+        if out_event is None:
+            return DEFAULT_DECODING_OUTPUT
```
yabirgb, line 255: this is an error, log as an error

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,388 @@
...
+
+        sell_amount = swap_amount = None
+        for log in all_logs:
+            if (
+                log.address == out_event.asset.resolve_to_evm_token().evm_address and
```
yabirgb, line 260: what haoppens if it is eth?

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,388 @@
...
+                log.topics[0] == ERC20_OR_ERC721_TRANSFER
```
yabirgb, line 261: you should probably check this first since if the condition doesn't hold you avoid resolving that is a potential trip to the db. Also don't resolve in every iteration of the loop, resolve once outside the loop

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,388 @@
...
+            ):
+                # Find the transfer log from the user wallet to router
+                if bytes_to_address(log.topics[2]) == self.router_address:
+                    sell_amount = int.from_bytes(log.data)
+
+                # Find the transfer log from the router to the swap
+                if bytes_to_address(log.topics[1]) == self.router_address:
```
yabirgb, line 268: this should be an elif. Otherwise you are checking twice per loop iteration

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,388 @@
...
+                    swap_amount = int.from_bytes(log.data)
+
+                if swap_amount and sell_amount:
+                    break
+
+        if not (sell_amount and swap_amount):
+            return DEFAULT_DECODING_OUTPUT
```
yabirgb, line 275: log error

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,388 @@
...
+
+        # Remove the transfer event
+        # of the purchased token to the user's wallet
+        # to create a new event with the right sequence index
+        transfer_log = all_logs[-1]
```
yabirgb, line 280: don't rely on this as it is error prone. Identify the events by checking that is a transfer and the to/from

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,388 @@
...
+        decoded_events: list['EvmEvent'],
+        action_items: list[ActionItem],   # pylint: disable=unused-argument
+        all_logs: list[EvmTxReceiptLog],
+    ) -> DecodingOutput:
+        sender = transaction.from_address
+        first_log = all_logs[0]
+        last_log = all_logs[-1]
+        # Check if is calling the rainbow router
+        if transaction.to_address != self.router_address:
+            return DEFAULT_DECODING_OUTPUT
```
yabirgb, line 332: we said that you can't trust the input data but neither you can trust the to_address since when using smart wallets it is the address of the wallet

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
prodesert22: not every transaction will have the receiving event in decode_events, so in_event can be none, sometimes this event is added at the end of the process, that's why you can see that in some functions I create an actionitem to delete and create a new event for "receive"

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,395 @@
...
+                all_logs=context.all_logs,
+                action_items=context.action_items,
+                sender=sender,
+                out_event=context.event,
+            )
+
+        return FAILED_ENRICHMENT_OUTPUT
+
+    def enricher_rules(self) -> list[Callable]:
+        return [self._maybe_enricher_rainbow_transfer]
```
nicholasyoder, line 387: I *think* a better approach would be to use `addresses_to_counterparties` and `post_decoding_rules` instead of `enricher_rules`. Something like this:

```python
    def _maybe_decode_rainbow_swap(
            self,
            transaction: 'EvmTransaction',
            all_logs: list['EvmTxReceiptLog'],
    ) -> list['EvmEvent']:
        # your decoding logic here
        return decoded_events

    def addresses_to_counterparties(self) -> dict[ChecksumEvmAddress, str]:
        return {self.router_address: CPT_RAINBOW_SWAPS}

    def post_decoding_rules(self) -> dict[str, list[tuple[int, Callable]]]:
        return {CPT_RAINBOW_SWAPS: [(0, self._maybe_decode_rainbow_swap)]}
```
I experimented with this a little and as far as I can tell it looks like it should work.

I should have thought of suggesting this earlier when I recommended using input data (sorry for the bad advice there - I see the others have explained that using input data is usually something to avoid).

I would recommend discussing this with @yabirgb though before trying to implement it - he suggested the current `enricher_rules` approach, and maybe there is something with `post_decoding_rules` that I'm not thinking of that would make it the wrong approach.

rotkehlchen/tests/unit/decoders/test_rainbow.py
```diff
@@ -0,0 +1,222 @@
...
+from rotkehlchen.types import Location, TimestampMS, deserialize_evm_tx_hash
+
+A_AIT: Final = Asset('eip155:1/erc20:0x89d584A1EDB3A70B3B07963F9A3eA5399E38b136')
+A_BUOY: Final = Asset('eip155:1/erc20:0x289Ff00235D2b98b0145ff5D4435d3e92f9540a6')
+A_AXGT: Final = Asset('eip155:1/erc20:0xDd66781D0E9a08D4FBb5eC7BAc80B691BE27F21D')
+A_ZIG: Final = Asset('eip155:1/erc20:0xb2617246d0c6c0087f18703d576831899ca94f01')
+
+
+@pytest.mark.parametrize('ethereum_accounts', [['0xfe25d6300D33e19E15AeeFEFD0Aafb319Dd61ae1']])
+def test_rainbow_swap_eth_to_token(ethereum_inquirer, ethereum_accounts):
```
nicholasyoder, line 25:
```suggestion
@pytest.mark.vcr(filter_query_parameters=['apikey'])
@pytest.mark.parametrize('ethereum_accounts', [['0xfe25d6300D33e19E15AeeFEFD0Aafb319Dd61ae1']])
def test_rainbow_swap_eth_to_token(ethereum_inquirer, ethereum_accounts):
```
You will need to VCR these tests to record the network calls and avoid making queries to etherscan here every time the CI runs. See https://docs.rotki.com/contribution-guides/python-testing.html#using-vcr

Feel free to ask questions if you need help with VCRing the tests.

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,395 @@
...
+            if (
+                event.event_type == HistoryEventType.TRADE and
+                event.event_subtype == HistoryEventSubType.RECEIVE
+            ):
+                in_event = event
+            elif (
+                event.event_type == HistoryEventType.TRADE and
+                event.event_subtype == HistoryEventSubType.SPEND
+            ):
+                out_event = event
```
nicholasyoder, line 86: the code in `_enricher_event` should just be here in `_get_in_out_events` so you only have to do one `if ... elif ...` block.

nicholasyoder (commented):
So I'm afraid the enricher approach may be wrong. There are a very large number of tests failing (especially decoder tests). I think the enricher logic here is modifying events for other protocols that it should not be touching.

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
prodesert22: I will invert this logic, so checking if the event is er20 transfer should resolve this problem
```python
tx_log.topics[0] == ERC20_OR_ERC721_TRANSFER and
tx_log.address == out_event.asset.resolve_to_evm_token().evm_address
```

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,380 @@
...
+            timestamp=transaction.timestamp,
+            fee_raw=fee_raw,
+            fee_asset=native_token,
+            sender=sender,
+        )
+        decoded_events.append(fee_event)
+
+        return
+
+    def _maybe_decode_swap_token_to_token(
```
yabirgb, line 225: this isn't decoding anything right? Just calculating the fee? Rename it correctly

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,380 @@
...
+                tx_log.topics[0] == ERC20_OR_ERC721_TRANSFER and
+                tx_log.address == token_address
+            ):
+                # Find the transfer log from the user wallet to router
+                if bytes_to_address(tx_log.topics[2]) == self.router_address:
+                    sell_amount = int.from_bytes(tx_log.data)
+
+                # Find the transfer log from the router to the swap provider
+                elif bytes_to_address(tx_log.topics[1]) == self.router_address:
+                    swap_amount = int.from_bytes(tx_log.data)
```
yabirgb, line 260:
```suggestion
                if bytes_to_address(tx_log.topics[2]) == self.router_address:
                    # Find the transfer log from the user wallet to router
                    sell_amount = int.from_bytes(tx_log.data)
                elif bytes_to_address(tx_log.topics[1]) == self.router_address:
                    # Find the transfer log from the router to the swap provider
                    swap_amount = int.from_bytes(tx_log.data)
```

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,380 @@
...
+
+                if swap_amount and sell_amount:
+                    break
+
+        if not (sell_amount and swap_amount):
+            log.error(
```
yabirgb, line 266: this can be the `else:` of the for loop

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,380 @@
...
+
+        return
+
+    def _process_swap(self,
+            transaction: 'EvmTransaction',
+            decoded_events: list['EvmEvent'],
+            all_logs: list['EvmTxReceiptLog'],
+    ) -> list['EvmEvent']:
+        sender = transaction.from_address
+        in_event, out_event = self._get_in_out_events(decoded_events, sender)
```
yabirgb, line 290: explain the docstring what you do here and what you are looking in the logs for the wrapping so the reader can follow the code easily

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,380 @@
...
+            for tx_log in all_logs:
+                if tx_log.topics[0] == WETH_DEPOSIT_TOPIC:
+                    wrap_log = tx_log
+                    break
+            else:
+                log.error(
+                    f'Failed to decode swap from native token to token '
+                    f'for transaction {transaction.tx_hash.hex()} due to missing wrap log.',
+                )
+                return decoded_events
```
yabirgb, line 310: this looks really similar to the other branch of the code. Can you try to abstract it? Maybe with a funtion that returns the event or None and you do something like

```
if (wrap_event := your_function(...)) is None:
     return decoded_events
```

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,380 @@
...
+            return decoded_events
+
+        if out_event.asset == self.evm_inquirer.native_token and in_event.asset.is_evm_token():
+            # Try to find if have wrap eth into weth log
+            wrap_log = None
+            for tx_log in all_logs:
+                if tx_log.topics[0] == WETH_DEPOSIT_TOPIC:
+                    wrap_log = tx_log
+                    break
+            else:
```
yabirgb, line 305: mmmm note that in arbitrum (irc it was arbitrum) and other chains the implementation of WETH is not the same and the events aren't emmited or they are different. You might have issues there with this approach. Isn't it easier to check the amounts of (eth sen't by the router to the rainbow wallet vs the the eth sent by the user to the rainbow router). You can do something like checking both ETH and weth if that reduces the complexity. Just asking since I'm not aware of the details

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,380 @@
...
+                # Try to find if have unwrap weth into eth log
+                if tx_log.topics[0] == WETH_WITHDRAW_TOPIC:
+                    unwrap_log = tx_log
+                    break
+            else:
+                log.error(
+                    f'Failed to decode swap from native token to token '
+                    f'for transaction {transaction.tx_hash.hex()} due to missing unwrap log.',
+                )
+                return decoded_events
```
yabirgb, line 331: this should be abstracted

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
prodesert22: Hmm, ok, so I will create 2 functions _find_weth_deposit_event and _find_weth_withdrawn_event, so in the arbitrum we ca override with the correct logic for this chain

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+
+        # Create the event for the fee
+        fee_event = self._create_fee_event(
+            tx_hash=transaction.tx_hash,
+            timestamp=transaction.timestamp,
+            fee_raw=fee_raw,
+            fee_asset=native_token,
+            sender=sender,
+        )
+        decoded_events.append(fee_event)
```
yabirgb, line 250: have the fee also properly sorted

```suggestion
        # The fee will be the withdraw data subtract the in_amount
        fee_raw = int.from_bytes(unwrap_event.data) - in_amount

        # Create the event for the fee
        fee_event = self._create_fee_event(
            tx_hash=transaction.tx_hash,
            timestamp=transaction.timestamp,
            fee_raw=fee_raw,
            fee_asset=native_token,
            sender=sender,
        )
        decoded_events.append(fee_event)
        maybe_reshuffle_events(
            ordered_events=[out_event, in_event, fee_event],
            events_list=decoded_events,
        )
```

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+            sender: ChecksumEvmAddress,
+    ) -> tuple[Optional['EvmEvent'], Optional['EvmEvent']]:
+        """
+        Gets the in and out event in the list of events
+        Args:
+            events (list['EvmEvent']): A list of EVM events to be processed.
+            sender (ChecksumEvmAddress): The address of the sender involved in the events.
+        Returns:
+            in_event: Represents the receive the tokens from the swap
+            ou_event: Represents the spend the tokens to swap
```
yabirgb, line 107: Typo in out_events but also we don't follow this docstring style. I agree that in this one it gives good context but keep it simple

```python
"""
        Extracts 'in' and 'out' events for a sender from a list of events.

        Args:
            events: List of EVM events.
            sender: Sender address to filter events by.

        Returns:
            Tuple of (in_event, out_event), representing token receive and spend events respectively.
        """
```

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+        super().__init__(evm_inquirer, base_tools, msg_aggregator)
+        self.router_address = router_address
+
+    def _enricher_event(
+            self,
+            event: 'EvmEvent',
+            sender: ChecksumEvmAddress,
+    ) -> None:
+        """Set the event to rainbow event
+        """
```
yabirgb, line 54: Rename the function to `_set_event_properties` but also keeps docstrings in one line when possible 

```python
"""Set properties of the event to a rainbow swap considering the event direction"""

File: rotkehlchen/chain/evm/decoding/rainbow/decoder.py [hunk 37]
```diff
@@ -0,0 +1,417 @@
...
+        Find the log by the event signature.
+
+        Args:
+            logs (list[EvmTxReceiptLog]): List of logs.
+            event_signature (bytes): Event signature to be searched.
+
+        Returns:
+            Optional[EvmTxReceiptLog]: Log if found, else None.
+        """
```
yabirgb, line 89:
```suggestion
        """
        Find the log that matches  the provided event signature.

        returns the log event if found or None otherwise.
        """
```

btw what happens if there are multiple wraps? I believe the logic can fail there, do you have anything else like the address to match against?

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+
+        out_asset_is_evm_token = out_event.asset.is_evm_token()
+        in_asset_is_evm_token = in_event.asset.is_evm_token()
+
+        if out_event.asset == self.evm_inquirer.native_token and in_asset_is_evm_token:
+            # Try to find if have wrap eth into weth log
+            if (wrap_event := self._find_event_by_log(all_logs, WETH_DEPOSIT_TOPIC)) is None:
+                log.error(
+                    f'Failed to decode fee from swap involving native token to token '
+                    f'for transaction {transaction.tx_hash.hex()} due to missing wrap log.',
```
yabirgb, line 350:
```suggestion
                    f'for transaction {transaction} due to missing wrap log.',
```

check it but the string version of the transaction includes the hash and important the chain too

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+    ) -> 'EvmEvent':
+        """Create a new fee event for rainbow swap
+
+        Args:
+            tx_hash (EVMTxHash): transaction hash
+            timestamp (int): transaction timestamp
+            fee_raw (int): amount of fee charged in wei
+            fee_asset (Asset): fee asset
+            sender (ChecksumEvmAddress): sender address
+        """
```
yabirgb, line 150: the arguments are explicit I believe you can simplify this docstring

```suggestion
        """Create a new fee event for a rainbow swap"""
```

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+            self,
+            transaction: EvmTransaction,
+            event: 'EvmEvent',
+            decoded_events: list['EvmEvent'],
+            sender: ChecksumEvmAddress,
+            wrap_event: EvmTxReceiptLog,
+    ) -> None:
+        """
+        Decodes a swap transaction from native token to another token.
+        Check if the first log is deposit and the last is transfer log
```
yabirgb, line 175: You didn't adjust this one. You don't seem to check on the transfer logs so it is wrong

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+        """
+        # The fee amount is the difference of msg.value and the wrapped token amount
+        try:
+            swap_amount = asset_raw_value(event.amount, self.evm_inquirer.native_token)
+        except UnsupportedAsset:
+            log.error(
+                f'Failed to decode swap from native token to token'
+                f' for transaction {transaction.tx_hash.hex()} '
+                f'due to unsupported asset {event.asset.identifier}.',
+            )
```
yabirgb, line 185: instead of `native token` use `{self.self.evm_inquirer.native_token}` and also in this file use `{transaction}` instead of just the hash

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+            return
+
+        fee_raw = swap_amount - int.from_bytes(wrap_event.data)
+
+        # Create fee event
```
yabirgb, line 190: this comment is not needed. It is obvious by the function name

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+            decoded_events: list['EvmEvent'],
+            in_event: Optional['EvmEvent'],
+            out_event: Optional['EvmEvent'],
+            sender: ChecksumEvmAddress,
+            unwrap_event: EvmTxReceiptLog,
+    ) -> None:
+        """
+        Decodes the fee from a swap transaction involving token to native token.
+        Via the first transfer log transfer and the last withdrawn log
+        """
```
yabirgb, line 214: this docstring seems to be a copy paste

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+        native_token = self.evm_inquirer.native_token
```
yabirgb, line 215: don't assign the alias, use it directly

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+
+        if in_event is None or out_event is None:
+            log.error(
+                f'Failed to decode fee from swap token to native token '
+                f'for transaction {transaction.tx_hash.hex()}.',
+            )
```
yabirgb, line 221: use the str version of transaction, not just the hash. The reason as I said is that it displays the chain that is important to find the transaction

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+            fee_asset=self.evm_inquirer.native_token,
+            sender=sender,
+        )
+        decoded_events.append(fee_event)
+
+        return
+
+    def _maybe_decode_fee_from_swap_token_to_native(
+            self,
+            transaction: EvmTransaction,
```
yabirgb, line 204: this and the case above look familiar. Is it possible to abstract the logic or part of it in a single function?

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+            decoded_events: list['EvmEvent'],
+            all_logs: list[EvmTxReceiptLog],
+            sender: ChecksumEvmAddress,
+            out_event: 'EvmEvent',
+    ) -> None:
+        """
+        Decodes the fee from a swap transaction involving token to token.
+        Via the first and last transfers logs
+        """
```
yabirgb, line 265: again a copy paste on the docstring, please pay attention to this

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+        Process a swap transaction by decoding the relevant events and logs.
+        Args:
+            transaction (EvmTransaction): The transaction object containing details of the swap.
+            decoded_events (list[EvmEvent]): A list of decoded events related to the transaction.
+            all_logs (list[EvmTxReceiptLog]): A list of all logs from the transaction receipt.
+        Returns:
+            list[EvmEvent]: The list of decoded events after processing the swap.
+        The function handles different types of swaps:
+        - Native token to EVM token swap, looking for the output asset is
+        eth and input asset is token
```
yabirgb, line 324:
```suggestion
        Process a swap transaction by decoding the relevant events and logs.

        The function handles different types of swaps:
        - Native token to EVM token swap, looking for the output asset is
        eth and input asset is token
```

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,417 @@
...
+    ) -> TransferEnrichmentOutput:
+        """
+        Enricher the event to rainbow event
+        """
+        sender = context.transaction.from_address
+        event = context.event
+        self._enricher_event(event, sender)
+        if event.counterparty == CPT_RAINBOW_SWAPS:
+            return TransferEnrichmentOutput(matched_counterparty=CPT_RAINBOW_SWAPS)
+        return FAILED_ENRICHMENT_OUTPUT
```
yabirgb, line 400:
```suggestion
        """
        Enricher transfer events to rainbow router. The contract doesn't emit events for swaps.
        """
        if event.address != self.router_address:
             return FAILED_ENRICHMENT_OUTPUT
             
        self._enricher_event(event=context.event, context.transaction.from_address)
        if event.counterparty == CPT_RAINBOW_SWAPS:
            return TransferEnrichmentOutput(matched_counterparty=CPT_RAINBOW_SWAPS)
        return FAILED_ENRICHMENT_OUTPUT
```

always exit as soon as possible. You don't need a extra call or read properties if you can exit fast. This is important for enrichers

rotkehlchen/tests/unit/decoders/test_rainbow.py
```diff
@@ -0,0 +1,225 @@
...
+    )]
+
+    assert expected_events == events
+
+
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('ethereum_accounts', [['0xB80177697e160eFC91d4B1ec295ABE6Ce0c0Fe1f']])
+def test_rainbow_swap_token_to_token(ethereum_inquirer, ethereum_accounts):
+    tx_hash = deserialize_evm_tx_hash('0x4f85ee11bbc401240755b083106e0811e6b60ef7972063a51596934cbb6ed43f')  # noqa: E501
+    events, _ = get_decoded_events_of_transaction(evm_inquirer=ethereum_inquirer, tx_hash=tx_hash)
```
yabirgb, line 156: this transaction is not decoded properly. You have an approval event in between the trade events. It has to be `spend -> receive -> fee` without other events between them

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
prodesert22: It get the first log, I check the some transactions the transactions and only have one wrap event, theoretically we will only have this event per swap. It's not possible to check by address because this log has the same value as the "target" that is present in input_data and since we cannot use it, there is no way to check since this address is dynamic.

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
prodesert22: I will check

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,402 @@
...
+
+    def enricher_rules(self) -> list[Callable]:
+        return [self._maybe_enricher_rainbow_transfer]
+
+    def post_decoding_rules(self) -> dict[str, list[tuple[int, Callable]]]:
+        return {CPT_RAINBOW_SWAPS: [(0, self._process_swap)]}
+
+    def addresses_to_counterparties(self) -> dict[ChecksumEvmAddress, str]:
+        return {self.router_address: CPT_RAINBOW_SWAPS}
```
nicholasyoder, line 394: So `addresses_to_counterparties` can be removed I think. Since `_maybe_enricher_rainbow_transfer` is setting the counterparty on the events there's no need for this.

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,402 @@
...
+                decoded_events=decoded_events,
+                all_logs=all_logs,
+                sender=sender,
+                out_event=out_event,
+                in_event=in_event,
+            )
+
+        return decoded_events
+
+    def _maybe_enricher_rainbow_transfer(
```
nicholasyoder, line 372:
```suggestion
    def _maybe_enrich_rainbow_transfer(
```
Should use the verb `enrich` here instead of the noun `enricher`.

LefterisJP (commented):
Also add a changelog entry.

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -0,0 +1,408 @@
...
+                f'Failed to post decode rainbow {transaction} '
+                f'due to missing in_event or out_event.',
+            )
+            return decoded_events
+
+        out_asset_is_evm_token = out_event.asset.is_evm_token()
+        in_asset_is_evm_token = in_event.asset.is_evm_token()
+
+        if out_event.asset == self.evm_inquirer.native_token and in_asset_is_evm_token:
+            # See if we have a eth into weth log event
```
LefterisJP, line 332:
```suggestion
            # See if we have an eth wrapping into weth log event
```

LefterisJP (approved):
Two things that should be done in the next PR, but for now this looks okay for the ethereum mainnet part. Will need to do the rest of the evm chains though and the logic needs changes for this.

PR #9637, opened 2025-03-17
Use SwapEvents in poloniex

nicholasyoder:
Related #6097


rotkehlchen/tests/exchanges/test_poloniex.py
```diff
@@ -234,24 +263,6 @@ def mock_api_return(url, **kwargs):  # pylint: disable=unused-argument
     mock_poloniex_and_query(given_input, expected_warnings_num=0, expected_errors_num=0)
 
 
-@pytest.mark.parametrize('function_scope_initialize_mock_rotki_notifier', [True])
-def test_query_trade_history_unknown_asset(poloniex):
-    """Test that poloniex trade history querying returning unknown asset is handled correctly"""
```
nicholasyoder, line 239: Removing this test since unknown asset handling is already being tested in `test_query_trade_history_unexpected_data`.

prettyirrelevant (approved):

rotkehlchen/exchanges/poloniex.py
```diff
@@ -74,45 +79,42 @@ def trade_from_poloniex(poloniex_trade: dict[str, Any]) -> Trade:
...
+        )
     except KeyError as e:
         raise DeserializationError(
             f'Poloniex trade deserialization error. Missing key entry for {e!s} in trade dict',
         ) from e
 
     log.debug(
-        'Processing poloniex Trade',
+        'Processing poloniex Swap',
         timestamp=timestamp,
```
LefterisJP, line 99:
```suggestion
        timestamp=(timestamp := deserialize_timestamp_ms_from_intms(poloniex_trade['createTime']))
```

rotkehlchen/exchanges/poloniex.py
```diff
@@ -74,45 +79,42 @@ def trade_from_poloniex(poloniex_trade: dict[str, Any]) -> Trade:
...
-        trade_type = TradeType.deserialize(poloniex_trade['side'])
-        # quantity is the base units of the trade
-        amount = deserialize_asset_amount(poloniex_trade['quantity'])
-        rate = deserialize_price(poloniex_trade['price'])
         fee = deserialize_fee(poloniex_trade['feeAmount'])
         fee_currency = asset_from_poloniex(poloniex_trade['feeCurrency'])
-        base_currency = asset_from_poloniex(get_pair_position_str(pair, 'first'))
-        quote_currency = asset_from_poloniex(get_pair_position_str(pair, 'second'))
-        timestamp = deserialize_timestamp_from_intms(poloniex_trade['createTime'])
+        timestamp = deserialize_timestamp_ms_from_intms(poloniex_trade['createTime'])
```
LefterisJP, line 84:
```suggestion
```

rotkehlchen/exchanges/poloniex.py
```diff
@@ -448,65 +450,51 @@ def query_balances(self) -> ExchangeQueryBalances:
...
+    ) -> list[SwapEvent]:
+        """Processes a single trade from poloniex and deserializes it via trade_from_poloniex.
+        Returns the deserialized SwapEvents in a list.
+        Logs error/warning and returns an empty list if unable to deserialize.
+        """
+        account_type = trade_data.get('accountType')
+        try:
+            if account_type == 'SPOT':
+                timestamp = deserialize_timestamp_from_intms(trade_data['createTime'])
+                if start_ts < timestamp < end_ts:
```
LefterisJP, line 467:
```suggestion
                if start_ts < (timestamp := deserialize_timestamp_from_intms(trade_data['createTime'])) < end_ts:
```

rotkehlchen/serialization/deserialize.py
```diff
@@ -194,18 +195,23 @@ def deserialize_timestamp_from_floatstr(time: str | (FVal | float)) -> Timestamp
     )
 
 
-def deserialize_timestamp_from_intms(time: int) -> Timestamp:
-    """Deserializes a timestamp an integer in milliseconds
-
-
-    Can throw DeserializationError if the data is not as expected
+def deserialize_timestamp_ms_from_intms(time: int) -> TimestampMS:
```
LefterisJP, line 198: What is this doing? This seems to just be a typecast. Why do you want this? Please respond in discord.

1. If you wanted a dynamic type check then this should pass a `value: Any` and just check it's an int and then return the proper type. But this means it's dynamic typechecking.
2. If you don't want dynamic type checking then this is useless. Just use cast or type ignore. A TimestampMS is an int.

rotkehlchen/tests/api/test_trades.py
```diff
@@ -1,1319 +0,0 @@
...
-    assert_poloniex_trades_result,
-    mock_history_processing_and_exchanges,
-)
-from rotkehlchen.types import AssetAmount, Fee, Location, Price, Timestamp, TradeType
-from rotkehlchen.utils.misc import ts_now
-
-
-@pytest.mark.parametrize('added_exchanges', [(Location.BINANCE, Location.POLONIEX)])
-@pytest.mark.parametrize('start_with_valid_premium', [False, True])
-def test_query_trades(
```
LefterisJP, line 43: This is a lot of test code that is being remove. Both here and in other files.

We are not removing exchange trades. Just turning them into a HistoryEvent derived structure. They are all still there. Are they still tested properly and do we still get proper coverage of all exchanges and trades query, deserialization and saving in the DB?

rotkehlchen/exchanges/poloniex.py
nicholasyoder: So this would have moved `poloniex_trade['createTime']` out of the try block allowing a potential uncaught KeyError, so I've actually moved the `log.debug(` within the try block and am now assigning fee and fee_currency with `:=` as well.

rotkehlchen/serialization/deserialize.py
nicholasyoder: Changed the argument to `value: Any` as discussed in discord.

rotkehlchen/tests/api/test_trades.py
nicholasyoder: As far as I can tell the things I removed here are directly related to the trades endpoints that are to be removed, testing stuff like ordering, pagination, filtering, adding/editing trades, etc.

For the new SwapEvents alot of these things like ordering, pagination, etc is now handled by the existing history event logic which we have existing tests for.

In addition here's a couple other tests we have already:
* `test_add_edit_swap_events` [here](https://github.com/rotki/rotki/blob/develop/rotkehlchen/tests/api/test_history_base_entry.py#L785) - tests adding/editing swap events via the normal history event endpoints
* `test_query_new_events` [here](https://github.com/rotki/rotki/blob/develop/rotkehlchen/tests/api/test_history_base_entry.py#L604) - tests the `exchangeeventsqueryresource` endpoint for querying new history events from exchanges, added when we did the asset movements.

With all that said, I think we do need to add SwapEvents and AssetMovements to the `predefined_events_to_insert` [here](https://github.com/rotki/rotki/blob/develop/rotkehlchen/tests/utils/history_base_entry.py#L73) which will include them in several more tests such as `test_get_events` [here](https://github.com/rotki/rotki/blob/develop/rotkehlchen/tests/api/test_history_base_entry.py#L404). This will involve adjusting a number of more tests though I think, so will do this in another PR.

LefterisJP (approved):
LGTM

PR #9638, opened 2025-03-17
allow repulling of missed evm transactions

### LINKED ISSUE #9592: Allow checking/repulling of transactions and history in a given period
## Problem Definition

In rotki it's possible that for some reason a range of transactions/history is not pulled. Be it a bug in rotki, a glitch in the source query (say etherscan re-indexing when querying) etc. it's in the realm of possibility that some transactions are missing.

Right now the only way to see if something is missing is a painstaking manual comparison, and addition of the transaction/event by hash.

To improve on that we need the ability to requery a time period and ignore any existing transactions but add ones missing. The query should be filterable by time period, by chain/location and address. If a filter is missing then it would be more generic.

## Specification

- Add an API endpoint that will alllow requerying transactions for a given time range with the aforementioned filters. The implementation of the logic should add the transactions and its receipt if missing, otherwise completely ignore. Perhaps we can report back to the user how many new transactions were found in the range.
- Create frontend for this

prettyirrelevant:
Backend part of #9592


rotkehlchen/chain/evm/transactions.py
```diff
@@ -428,6 +410,49 @@ def _get_erc20_transfers_for_ranges(
...
+            to_ts=Timestamp(period.to_value),
+        ):
+            for tx_hash in erc20_tx_hashes:
+                with self.database.conn.read_ctx() as cursor:
+                    tx, _ = self.get_or_create_transaction(
+                        cursor=cursor,
+                        tx_hash=tx_hash,
+                        relevant_address=address,
+                    )
+                timestamp = tx.timestamp
```
nicholasyoder, line 433: Should probably just use `tx.timestamp` instead of making a new variable here.

rotkehlchen/chain/evm/transactions.py
```diff
@@ -846,3 +871,70 @@ def get_chain_specific_multiaddress_data(
         """Can be implemented by each chain subclass to add chain-specific data queries
         for all tracked addresses at once"""
         return None
+
+    def refetch_transactions_for_address(
```
nicholasyoder, line 875: It seems that there could be some refactoring that should be done here. For instance `single_address_query_transactions` is quite similar to `refetch_transactions_for_address`, although it calls different functions. Not sure right off if there's a good way to refactor some of this or not. Only refactor if there's a way that actually makes it better though.

nicholasyoder (commented):
I notice you have `Closes` in the PR description - but there is frontend work that needs done after this right? So probably should close the issue with this PR.

rotkehlchen/api/server.py
```diff
@@ -321,6 +322,7 @@
     ('/balances/historical/asset', HistoricalAssetAmountsResource),
     ('/balances/historical/asset/prices', HistoricalPricesPerAssetResource),
     ('/balances/historical/netvalue', HistoricalNetValueResource),
+    ('/blockchains/evm/transactions/refetch', RefetchEvmTransactionsResource),
```
yabirgb, line 325: have it next to the other endpoints in the same path, having them disperse around the list is making ourselves the work harder

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`9592` Users can now force refetch EVM transactions for specific time periods to recover missed transactions.
```
```suggestion
* :feature:`9592` Users can now force a query of EVM transactions for specific time periods to recover missed transactions.
```

rotkehlchen/chain/evm/transactions.py
```diff
@@ -185,13 +186,14 @@ def _query_and_save_transactions_for_range(
...
-                with self.database.user_write() as write_cursor:
-                    # update last queried time for the address
-                    self.dbranges.update_used_query_range(
-                        write_cursor=write_cursor,
-                        location_string=location_string,
-                        queried_ranges=[(period.from_value, new_transactions[-1].timestamp)],  # type: ignore
-                    )
+                if update_ranges:
+                    with self.database.user_write() as write_cursor:
+                        # update last queried time for the address
```
yabirgb, line 191: you open twice the write cursor, is it possible to do it only once?

rotkehlchen/chain/evm/transactions.py
```diff
@@ -428,6 +410,48 @@ def _get_erc20_transfers_for_ranges(
...
+                with self.database.conn.read_ctx() as cursor:
+                    tx, _ = self.get_or_create_transaction(
+                        cursor=cursor,
+                        tx_hash=tx_hash,
+                        relevant_address=address,
+                    )
+
+                if period.range_type == 'timestamps':
+                    log.debug(f'{self.evm_inquirer.chain_name} ERC20 Transfers for {address} -> update range {period.from_value} - {tx.timestamp}')  # noqa: E501
+                    if update_ranges:
```
yabirgb, line 436: this if statement I'm not sure if it is needed. I'm under the impression that you always want to update the ranges queried. Is there a reason to not do it?

rotkehlchen/chain/evm/transactions.py
yabirgb: It feels like it is the same function tbh. Can't you do something like defining a decorator that does

```
def decorator:
     acquire the log
     self message (start query)
     call wrapper
     self message (end querying)
```

does it become ugly?

rotkehlchen/api/v1/schemas.py
```diff
@@ -4024,3 +4024,46 @@ class RefreshProtocolDataSchema(AsyncQueryArgumentSchema):
...
+                    ).fetchall()
+                else:
+                    accounts_count = cursor.execute(
+                        'SELECT blockchain, account FROM blockchain_accounts WHERE account=?',
+                        (address,),
+                    ).fetchall()
+
+                if len(accounts_count) == 0:
+                    raise ValidationError(
+                        message=f'Account {address} is not tracked by rotki',
```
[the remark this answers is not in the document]
prettyirrelevant: done.

rotkehlchen/api/v1/schemas.py
```diff
@@ -4024,3 +4024,46 @@ class RefreshProtocolDataSchema(AsyncQueryArgumentSchema):
...
+                field_name='from_timestamp',
+            )
+
+        if (address := data['address']) is not None:
+            with self.db.conn.read_ctx() as cursor:
+                if (evm_chain := data['evm_chain']) is not None:
+                    accounts_count = cursor.execute(
+                        'SELECT blockchain, account FROM blockchain_accounts WHERE blockchain=? AND account=?',  # noqa: E501
+                        (evm_chain.to_blockchain().value, address),
+                    ).fetchall()
```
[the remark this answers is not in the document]
prettyirrelevant: i was trying to do something else. fixed. thanks.

rotkehlchen/chain/evm/transactions.py
```diff
@@ -846,3 +870,70 @@ def get_chain_specific_multiaddress_data(
...
+    ) -> None:
+        """Force refetch transactions for an address within a time range.
+
+       Directly queries for transactions, internal transactions, and ERC20
+       transfers without updating query ranges, allowing recovery of
+       potentially missed transactions.
+
+       May raise:
+       - RemoteError if any of the remote queries fail.
+       - IntegrityError if the transaction already exists.
```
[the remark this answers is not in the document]
prettyirrelevant: i just checked the code again and we do `INSERT OR IGNORE` so integrity error is never thrown.

the errors that can be thrown are(it comes from `get_or_create_transaction()`):
```
        - pysqlcipher3.dbapi2.OperationalError if the SQL query fails due to invalid
        filtering arguments.
        - RemoteError if there is a problem querying the data source.
        - DeserializationError if the transaction cannot be deserialized from the DB.
```

> So what happens if I query for August 2023 where I made transactions every day and in that range I'm missing transactions on the 3rd and 4th but have transactions for the rest of the month, will this logic fail?

it won't fail.

rotkehlchen/chain/evm/transactions.py
```diff
@@ -292,13 +295,14 @@ def _query_and_save_internal_transactions_for_range_or_parent_hash(
...
-                            queried_ranges=[(period_or_hash.from_value, timestamp)],  # type: ignore
-                        )
+                    if update_ranges:
+                        with self.database.conn.write_ctx() as write_cursor:
+                            # update last queried time for address
+                            self.dbranges.update_used_query_range(
+                                write_cursor=write_cursor,
+                                location_string=location_string,
+                                queried_ranges=[(period_or_hash.from_value, timestamp)],  # type: ignore
+                            )
```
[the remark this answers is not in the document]
prettyirrelevant: the reason `update_ranges=False` is critical for force refetching is simple: database range tracking.

when rotki queries transactions, it records which time ranges have been queried in the database. this is an "insert or replace" operation, marking ranges as "already checked."

during normal operation, this behavior is perfect - we don't waste time re-querying periods we've already processed.

but for a force refetch, this same behavior would be counterproductive. if we set `update_ranges=True` during force refetch, we'd mark that period as "checked" in the database - effectively telling rotki "don't worry about this period anymore" when the whole point was to double-check it.

by using `update_ranges=False`, we query transactions but don't update the database's range tracking. this means:

1. we get any missing transactions from that period
2. the normal query scheduling remains intact
3. if you were rate-limited during previous queries, you can force refetch without corrupting the system's understanding of what still needs to be checked

it's like telling rotki "check this drawer again, but don't cross it off your to-do list."

rotkehlchen/chain/evm/transactions.py
prettyirrelevant: answered here https://github.com/rotki/rotki/pull/9638#discussion_r2001613487

rotkehlchen/api/rest.py
```diff
@@ -5346,3 +5346,66 @@ def get_historical_prices_per_asset(
...
+        chains_to_query: list[EVM_CHAIN_IDS_WITH_TRANSACTIONS_TYPE] = [evm_chain] if evm_chain else list(EVM_CHAIN_IDS_WITH_TRANSACTIONS)  # noqa: E501
+        for chain_id in chains_to_query:
+            chain_manager = self.rotkehlchen.chains_aggregator.get_evm_manager(chain_id)
+            if address:
+                addresses_to_query: tuple[ChecksumEvmAddress, ...] = (address,)
+            else:
+                with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
+                    addresses_to_query = self.rotkehlchen.data.db.get_blockchain_accounts(cursor).get(chain_manager.node_inquirer.blockchain)  # noqa: E501
+
+            if addresses_to_query is None or len(addresses_to_query) == 0:
```
LefterisJP, line 5383: When can `addresses_to_query` be None here? Isn't it always a tuple of addresses and at worst case an empty tuple? Or am I missing something?

```suggestion
            if addresses_to_query is None or len(addresses_to_query) == 0:
```

rotkehlchen/api/rest.py
```diff
@@ -5346,3 +5346,66 @@ def get_historical_prices_per_asset(
...
+            )
+            for addr in addresses_to_query:
+                try:
+                    chain_manager.transactions.refetch_transactions_for_address(
+                        address=addr,
+                        start_ts=from_timestamp,
+                        end_ts=to_timestamp,
+                    )
+                except (sqlcipher.OperationalError, RemoteError, DeserializationError) as e:  # pylint: disable=no-member
+                    log.debug(f'Skipping transaction entry for {addr} on {chain_id} due to: {e!s}')
```
LefterisJP, line 5400:
```suggestion
                    log.debug(f'Skipping transaction refetching for {addr} on {chain_id} due to: {e!s}')
```

rotkehlchen/api/rest.py
```diff
@@ -5346,3 +5346,66 @@ def get_historical_prices_per_asset(
...
+                    continue
+
+            # Get total count after query and add difference
```
LefterisJP, line 5403:
```suggestion
            # Get total count after query and count the difference as addition
```

rotkehlchen/chain/evm/transactions.py
```diff
@@ -166,6 +188,7 @@ def _query_and_save_transactions_for_range(
             address: ChecksumEvmAddress,
             period: TimestampOrBlockRange,
             location_string: str | None = None,
+            update_ranges: bool = True,
```
LefterisJP, line 191: can you explain a bit in the docstring about why `update_ranges` is here and when it's expected to be False?

rotkehlchen/chain/evm/transactions.py
```diff
@@ -257,6 +279,7 @@ def _query_and_save_internal_transactions_for_range_or_parent_hash(
             period_or_hash: TimestampOrBlockRange | EVMTxHash,
             address: ChecksumEvmAddress | None = None,
             location_string: str | None = None,
+            update_ranges: bool = True,
```
LefterisJP, line 282: same for a docstring here

rotkehlchen/chain/evm/transactions.py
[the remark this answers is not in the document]
LefterisJP: very good explanation. Exactly what I was expecting to be done with the issue.

rotkehlchen/chain/evm/transactions.py
```diff
@@ -428,6 +430,48 @@ def _get_erc20_transfers_for_ranges(
...
             )
 
+    def _query_and_save_erc20_transfers_for_range(
+            self,
+            address: ChecksumEvmAddress,
+            period: TimestampOrBlockRange,
+            location_string: str,
+            update_ranges: bool = True,
+    ) -> None:
+        """Helper function to abstract ERC20 transfer querying functionality for different range types"""  # noqa: E501
```
LefterisJP, line 440: same for docstring for update ranges. (perhaps it will end up repeating but not sure what's best)

rotkehlchen/chain/evm/transactions.py
```diff
@@ -428,6 +430,48 @@ def _get_erc20_transfers_for_ranges(
...
+                        cursor=cursor,
+                        tx_hash=tx_hash,
+                        relevant_address=address,
+                    )
+
+                if period.range_type == 'timestamps':
+                    log.debug(f'{self.evm_inquirer.chain_name} ERC20 Transfers for {address} -> update range {period.from_value} - {tx.timestamp}')  # noqa: E501
+                    if update_ranges:
+                        with self.database.user_write() as write_cursor:
+                            # update last queried time for the address
```
LefterisJP, line 458: save a line

```suggestion
                    if update_ranges:   # update last queried time for the address
                        with self.database.user_write() as write_cursor:   
```

rotkehlchen/chain/evm/transactions.py
```diff
@@ -292,13 +315,14 @@ def _query_and_save_internal_transactions_for_range_or_parent_hash(
...
-                    with self.database.conn.write_ctx() as write_cursor:
-                        # update last queried time for address
-                        self.dbranges.update_used_query_range(
-                            write_cursor=write_cursor,
-                            location_string=location_string,
-                            queried_ranges=[(period_or_hash.from_value, timestamp)],  # type: ignore
-                        )
+                    if update_ranges:
+                        with self.database.conn.write_ctx() as write_cursor:
+                            # update last queried time for address
```
LefterisJP, line 320: save a line

```suggestion
                    if update_ranges:  # update last queried time for address
                        with self.database.conn.write_ctx() as write_cursor:
```

LefterisJP (commented):
Very nicely done. Small comments only.

rotkehlchen/api/rest.py
prettyirrelevant: my bad. i didn't get the return type of `.get()` in this case. assumed it'll return None.

LefterisJP (approved):
LGTM

PR #9645, opened 2025-03-18
backend: Add active management

yabirgb:
- add endpoint to build payload for token transfer
- add endpoint to build payload for native token transfer
- add endpoint to check if two addresses interacted before

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -13642,7 +13642,7 @@ Historical Balance Queries
 
 
 Refetch EVM transactions for a specific time period
-==================================================
+===================================================
 
 .. http:post:: /api/(version)/blockchains/evm/transactions/refetch
 
@@ -13703,7 +13703,7 @@ Active management
 
       .. http:example:: curl wget httpie python-requests
 
-        POST /api/(version)/wallet/transfer/token HTTP/1.1
+        POST /api/1/wallet/transfer/token HTTP/1.1
         Host: localhost:5042
         Content-Type: application/json;charset=UTF-8
 
@@ -13714,10 +13714,10 @@ Active management
           "amount": "0.000000000000000001"
         }
 
-      :reqjsonarr string from_address: The address from which the tokens will be sent
-      :reqjsonarr string to_address: The address to which the tokens will be sent
-      :reqjsonarr string token: The token identifier in the format eip155:{chain_id}/erc20:{contract_address}
-      :reqjsonarr string amount: The amount of tokens to transfer as a string
+    :reqjson string from_address: The address from which the tokens will be sent
+    :reqjson string to_address: The address to which the tokens will be sent
+    :reqjson string token: The token identifier in the format eip155:{chain_id}/erc20:{contract_address}
+    :reqjson string amount: The amount of tokens to transfer as a string
 
     **Example Response:**
 
@@ -13737,15 +13737,15 @@ Active management
           "status_code": 200
         }
 
-      :resjson object result: The transaction data needed for the token transfer
-      :resjson integer chainId: The chain ID of the blockchain network
-      :resjson string from: The sender's address
-      :resjson string to: The token contract address
-      :resjson string data: The encoded transaction data for the token transfer
-      :statuscode 200: Transaction data successfully prepared
-      :statuscode 400: Malformed request
-      :statuscode 409: Error preparing the payload
-      :statuscode 500: Internal Rotki error
+    :resjson object result: The transaction data needed for the token transfer
+    :resjson integer chainId: The chain ID of the blockchain network
+    :resjson string from: The sender's address
+    :resjson string to: The token contract address
+    :resjson string data: The encoded transaction data for the token transfer
+    :statuscode 200: Transaction data successfully prepared
+    :statuscode 400: Malformed request
+    :statuscode 409: Error preparing the payload
+    :statuscode 500: Internal Rotki error
 
   .. http:post:: /api/(version)/wallet/transfer/native
 
@@ -13755,7 +13755,7 @@ Active management
 
       .. http:example:: curl wget httpie python-requests
 
-        POST /api/(version)/wallet/transfer/native HTTP/1.1
+        POST /api/1/wallet/transfer/native HTTP/1.1
         Host: localhost:5042
         Content-Type: application/json;charset=UTF-8
 
@@ -13763,13 +13763,13 @@ Active management
           "from_address": "0xc37b40ABdB939635068d3c5f13E7faF686F03B65",
           "to_address": "0x9531C059098e3d194fF87FebB587aB07B30B1306",
           "amount": "0.0003",
-          "blockchain": "ETH"
+          "chain": "ethereum"
         }
 
-      :reqjsonarr string from_address: The address from which the native cryptocurrency will be sent
-      :reqjsonarr string to_address: The address to which the native cryptocurrency will be sent
-      :reqjsonarr string amount: The amount of native cryptocurrency to transfer as a string
-      :reqjsonarr string blockchain: The blockchain identifier (e.g., "ETH" for Ethereum)
+    :reqjson string from_address: The address from which the native cryptocurrency will be sent
+    :reqjson string to_address: The address to which the native cryptocurrency will be sent
+    :reqjson string amount: The amount of native cryptocurrency to transfer as a string
+    :reqjson string chain: The EVM chain name (e.g., "ethereum" for Ethereum)
 
     **Example Response:**
 
@@ -13784,21 +13784,53 @@ Active management
             "from": "0xc37b40ABdB939635068d3c5f13E7faF686F03B65",
             "to": "0x9531C059098e3d194fF87FebB587aB07B30B1306",
             "value": 300000000000000,
-            "maxFeePerGas": 2000000000,
-            "maxPriorityFeePerGas": 1000000000,
             "nonce": 55
           },
           "status_code": 200
         }
 
-      :resjson object result: The transaction data needed for the native transfer
-      :resjson string from: The sender's address
-      :resjson string to: The recipient's address
-      :resjson integer value: The amount to transfer in the smallest unit of the native cryptocurrency
-      :resjson integer maxFeePerGas: The maximum fee per gas unit the user is willing to pay
-      :resjson integer maxPriorityFeePerGas: The maximum priority fee per gas unit
-      :resjson integer nonce: The transaction nonce
-      :statuscode 200: Transaction data successfully prepared
-      :statuscode 400: Malformed request
-      :statuscode 409: Error preparing the payload
-      :statuscode 500: Internal Rotki error
+    :resjson object result: The transaction data needed for the native transfer
+    :resjson string from: The sender's address
+    :resjson string to: The recipient's address
+    :resjson integer value: The amount to transfer in the smallest unit of the native cryptocurrency
+    :resjson integer nonce: The transaction nonce
+    :statuscode 200: Transaction data successfully prepared
+    :statuscode 400: Malformed request
+    :statuscode 409: Error preparing the payload
+    :statuscode 500: Internal Rotki error
+
+
+  .. http:post:: /api/(version)/wallet/interacted
+
+    Checks if the address in the `from` interacted with the address in the `to`. This interaction check is unidirectional, meaning it only verifies if `from_address` has interacted with `to_address`, not the other way around.
+
+    **Example Request:**
+        .. http:example:: curl wget httpie python-requests
+
+          POST /api/1/wallet/interacted HTTP/1.1
+          Host: localhost:5042
+          Content-Type: application/json;charset=UTF-8
+
+          {
+              "from_address": "0xc37b40ABdB939635068d3c5f13E7faF686F03B65",
+              "to_address": "0x9531C059098e3d194fF87FebB587aB07B30B1306"
+          }
+
+      :reqjson string from_address: The address from which the interaction is being checked.
+      :reqjson string to_address: The address with which the interaction is being checked.
+
+    **Example Response:**
+
+      .. sourcecode:: http
+
+        HTTP/1.1 200 OK
+        Content-Type: application/json
+
+        {
+            "result": true
+        }
+
+    :resjson boolean result: Indicates whether the `from_address` has interacted with the `to_address`.
+    :statuscode 200: Interaction check successful.
+    :statuscode 400: Malformed request.
+    :statuscode 500: Internal Rotki error.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -13690,3 +13690,115 @@ Refetch EVM transactions for a specific time period
    :statuscode 401: User is not logged in.
    :statuscode 400: Invalid parameters such as from_timestamp > to_timestamp. Address not tracked by rotki.
    :statuscode 500: Internal rotki error
+
+
+Active management
+==================
+
+  .. http:post:: /api/(version)/wallet/transfer/token
+
+    Prepares a token transfer transaction without submitting it to the blockchain. This endpoint returns the transaction data needed to perform an ERC20 token transfer.
+
+    **Example Request:**
+
+      .. http:example:: curl wget httpie python-requests
+
+        POST /api/(version)/wallet/transfer/token HTTP/1.1
+        Host: localhost:5042
+        Content-Type: application/json;charset=UTF-8
+
+        {
+          "from_address": "0xc37b40ABdB939635068d3c5f13E7faF686F03B65",
+          "to_address": "0x9531C059098e3d194fF87FebB587aB07B30B1306",
+          "token": "eip155:1/erc20:0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84",
+          "amount": "0.000000000000000001"
+        }
+
+      :reqjsonarr string from_address: The address from which the tokens will be sent
+      :reqjsonarr string to_address: The address to which the tokens will be sent
+      :reqjsonarr string token: The token identifier in the format eip155:{chain_id}/erc20:{contract_address}
+      :reqjsonarr string amount: The amount of tokens to transfer as a string
+
+    **Example Response:**
+
+      .. sourcecode:: http
+
+        HTTP/1.1 200 OK
+        Content-Type: application/json
+
+        {
+          "message": "",
+          "result": {
+            "chainId": 1,
+            "from": "0xc37b40ABdB939635068d3c5f13E7faF686F03B65",
+            "to": "0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84",
+            "data": "0xa9059cbb0000000000000000000000009531c059098e3d194ff87febb587ab07b30b13060000000000000000000000000000000000000000000000000000000000000001"
+          },
+          "status_code": 200
+        }
+
+      :resjson object result: The transaction data needed for the token transfer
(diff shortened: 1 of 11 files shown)


rotkehlchen/api/rest.py
```diff
@@ -5409,3 +5409,62 @@ def force_refetch_evm_transactions(
...
+            self,
+            from_address: ChecksumEvmAddress,
+            to_address: ChecksumEvmAddress,
+    ) -> Response:
+        with self.rotkehlchen.data.db.conn.read_ctx() as cursor:
+            cursor.execute(
+                'SELECT COUNT(*) FROM history_events JOIN evm_events_info ON '
+                'history_events.identifier=evm_events_info.identifier WHERE '
+                'location_label=? AND address=?',
+                (from_address, to_address),
```
prettyirrelevant, line 5423: this check is not bidirectional as it could be the `address` interacting with `location_label`. 
on a second thought, this might not be what you're trying to achieve.

rotkehlchen/api/rest.py
```diff
@@ -5409,3 +5409,62 @@ def force_refetch_evm_transactions(
...
+    @async_api_call()
+    def prepare_token_transfer(
+            self,
+            from_address: ChecksumEvmAddress,
+            to_address: ChecksumEvmAddress,
+            token: EvmToken,
+            amount: FVal,
+    ) -> dict[str, Any]:
+        manager = self.rotkehlchen.chains_aggregator.get_chain_manager(
+            blockchain=token.chain_id.to_blockchain(),  # type: ignore  # checked in the validation
```
prettyirrelevant, line 5436: if it is checked in the validation, you can just use that as a parameter here. something like:
```py
def prepare_token_transfer(
            self,
            blockchain: SUPPORTED_EVM_CHAINS_TYPE,
            from_address: ChecksumEvmAddress,
            to_address: ChecksumEvmAddress,
            token: EvmToken,
            amount: FVal,
    ) -> dict[str, Any]:
```

rotkehlchen/api/rest.py
```diff
@@ -5409,3 +5409,62 @@ def force_refetch_evm_transactions(
...
+            return wrap_in_fail_result(str(e), status_code=HTTPStatus.BAD_REQUEST)
+
+        return _wrap_in_ok_result(result=payload)
+
+    @async_api_call()
+    def prepare_native_transfer(
+            self,
+            from_address: ChecksumEvmAddress,
+            to_address: ChecksumEvmAddress,
+            blockchain: SupportedBlockchain,
```
prettyirrelevant, line 5456: use `SUPPORTED_EVM_CHAINS_TYPE` so there's no need for the `type: ignore`.

rotkehlchen/api/v1/schemas.py
```diff
@@ -4065,3 +4065,32 @@ def validate_schema(
                         message=f'Account {address} with chain {evm_chain.to_name()} is not tracked by rotki',  # noqa: E501
                         field_name='address',
                     )
+
+
+class AddressesInteraction(Schema):
+    from_address = fields.String(required=True)
```
prettyirrelevant, line 4071: we have `EvmAddressField()` so perhaps it makes sense to use that here and below.

rotkehlchen/api/v1/schemas.py
```diff
@@ -4065,3 +4065,32 @@ def validate_schema(
...
+class AssetTransferSchema(AddressesInteraction):
+    amount = PositiveAmountField()
+
+
+class TokenTransfer(AssetTransferSchema):
+    token = AssetField(expected_type=EvmToken)
+
+
+class NativeAssetTransfer(AssetTransferSchema):
+    blockchain = BlockchainField(required=True)
```
prettyirrelevant, line 4084:
```suggestion
    blockchain = BlockchainField(required=True, limit_to=SUPPORTED_EVM_CHAINS)
```

this way you don't need the validation below

rotkehlchen/api/v1/schemas.py
```diff
@@ -4065,3 +4065,32 @@ def validate_schema(
...
                     )
+
+
+class AddressesInteraction(Schema):
+    from_address = fields.String(required=True)
+    to_address = fields.String(required=True)
+
+
+class AssetTransferSchema(AddressesInteraction):
+    amount = PositiveAmountField()
```
prettyirrelevant, line 4076:
```suggestion
    amount = PositiveAmountField(required=True)
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -4065,3 +4065,32 @@ def validate_schema(
...
+
+
+class TokenTransfer(AssetTransferSchema):
+    token = AssetField(expected_type=EvmToken)
```
prettyirrelevant, line 4080:
```suggestion
    token = AssetField(required=True, expected_type=EvmToken)
```

rotkehlchen/api/v1/wallet_resources.py
```diff
@@ -0,0 +1,63 @@
+from flask import Response
+from webargs.flaskparser import use_kwargs
+
+from rotkehlchen.api.v1.common_resources import BaseMethodView
+from rotkehlchen.api.v1.schemas import AddressesInteraction, NativeAssetTransfer, TokenTransfer
+from rotkehlchen.assets.asset import EvmToken
+from rotkehlchen.types import ChecksumEvmAddress, FVal, SupportedBlockchain
+
+
+class PrepareTokenTransferResource(BaseMethodView):
```
prettyirrelevant, line 10: i have nothing against separating resources but any reason in particular you did this? (learning purposes)

rotkehlchen/chain/evm/active_management/manager.py
```diff
@@ -0,0 +1,153 @@
...
+    ) -> bool:
+        """Check if from address interacted before with to_address"""
+        with self.node_inquirer.database.conn.read_ctx() as cursor:
+            cursor.execute(
+                'SELECT COUNT(*) FROM history_events JOIN evm_events_info ON '
+                'history_events.identifier=evm_events_info.identifier WHERE '
+                'location_label=? AND address=?',
+                (from_address, to_address),
+            )
+            return cursor.fetchone()[0] > 0
```
prettyirrelevant, line 41: why not use this method in in `rest.py` to avoid duplication?

rotkehlchen/chain/evm/active_management/manager.py
```diff
@@ -0,0 +1,153 @@
...
+            log.error(
+                'Failed to obtain baseFeePerGas because is not present in the last block for '
+                f'{self.node_inquirer.chain_id}',
+            )
+            base_fee_per_gas = Wei(0)
+
+        max_priority_fee_per_gas = web3.eth.max_priority_fee
+        return {
+            'maxFeePerGas': base_fee_per_gas + max_priority_fee_per_gas,  # Correct formula
+            'maxPriorityFeePerGas': max_priority_fee_per_gas,
```
prettyirrelevant, line 54: why not use snake case for the naming? for convention sake. the frontend handles the conversion if that's the reason.

rotkehlchen/chain/evm/active_management/manager.py
```diff
@@ -0,0 +1,153 @@
...
+        with self.node_inquirer.database.conn.read_ctx() as cursor:
+            cursor.execute(
+                'SELECT COUNT(*) FROM history_events JOIN evm_events_info ON '
+                'history_events.identifier=evm_events_info.identifier WHERE '
+                'location_label=? AND address=?',
+                (from_address, to_address),
+            )
+            return cursor.fetchone()[0] > 0
+
+    def _query_gas(self, web3: Web3) -> dict[str, int]:
```
prettyirrelevant, line 43: etherscan also provides endpoints to query gas 
- https://docs.etherscan.io/api-endpoints/gas-tracker#get-gas-oracle
- https://docs.etherscan.io/api-endpoints/geth-parity-proxy#eth_estimategas

there's no privacy concern with this i guess.

rotkehlchen/chain/evm/active_management/manager.py
```diff
@@ -0,0 +1,153 @@
...
+            from_address: 'ChecksumEvmAddress',
+            to_address: 'ChecksumEvmAddress',
+            token: 'EvmToken',
+            amount: 'FVal',
+    ) -> dict[str, Any]:
+        """Wrapper for _create_token_transfer that tries different nodes to create
+        the transaction handling errors.
+
+        May raise:
+            - RemoteError: If no web3 node is available
```
prettyirrelevant, line 116: etherscan can be used i suppose? this is an `eth_call` and they support it.

rotkehlchen/chain/evm/active_management/manager.py
```diff
@@ -0,0 +1,153 @@
...
+            )
+            base_fee_per_gas = Wei(0)
+
+        max_priority_fee_per_gas = web3.eth.max_priority_fee
+        return {
+            'maxFeePerGas': base_fee_per_gas + max_priority_fee_per_gas,  # Correct formula
+            'maxPriorityFeePerGas': max_priority_fee_per_gas,
+        }
+
+    def _query_with_nodes(self, func: Callable, **kwargs: Any) -> dict[str, Any]:
```
prettyirrelevant, line 57: is this needed? the `_query()` method in the EvmNodeInquirer does exactly this or am i missing something?

rotkehlchen/chain/evm/active_management/manager.py
```diff
@@ -0,0 +1,153 @@
...
+            to_address: 'ChecksumEvmAddress',
+            token: 'EvmToken',
+            amount: 'FVal',
+            web3node: 'Web3Node',
+    ) -> dict[str, Any]:
+        """Build transaction to transfer erc20 tokens"""
+        contract = web3node.web3_instance.eth.contract(
+            address=token.evm_address,
+            abi=self.node_inquirer.contracts.erc20_abi,
+        )
```
prettyirrelevant, line 92: instead of this, we can just improve the `EvmContract` and abstract this away.

rotkehlchen/chain/evm/active_management/manager.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def _transfer_native_token(
+            self,
+            from_address: 'ChecksumEvmAddress',
+            to_address: 'ChecksumEvmAddress',
+            amount: 'FVal',
+            web3node: 'Web3Node',
+    ) -> dict[str, Any]:
+        """Build transaction to transfer erc20 tokens"""
```
prettyirrelevant, line 133:
```suggestion
        """Build transaction to transfer the native asset of the blockchain"""
```

rotkehlchen/chain/evm/active_management/manager.py
yabirgb: this is a left over as before it was used here. Removed.

rotkehlchen/chain/evm/active_management/manager.py
yabirgb: this is the payload that is sent directly to the wallet and expects this format

rotkehlchen/api/rest.py
yabirgb: yes, intentionally is not bidirectional since you want only the check in the direction that you send. Otherwise someone could send you a transaction and then trick you to send back to that address.

rotkehlchen/chain/evm/active_management/manager.py
yabirgb: for this I believe it is better to keep web3 nodes for now and we can check to use etherscan later

docs/api.rst
```diff
@@ -13689,4 +13689,116 @@ Refetch EVM transactions for a specific time period
...
+
+      .. sourcecode:: http
+
+        HTTP/1.1 200 OK
+        Content-Type: application/json
+
+        {
+          "message": "",
+          "result": {
+            "chainId": 1,
```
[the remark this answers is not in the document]
prettyirrelevant: oh you misunderstood me. i meant if you change it to `chain_id`, the frontend will still transform it `chainId`

```ts
export function snakeCaseTransformer(data: any): any {
  return convertKeys(data, false, false);
}

export function camelCaseTransformer(data: any): any {
  return convertKeys(data, true, false);
}

export function noRootCamelCaseTransformer(data: any): any {
  return convertKeys(data, true, true);
}
```

rotkehlchen/chain/evm/active_management/manager.py
prettyirrelevant: same comment as the one above. the camel case transformer.

rotkehlchen/api/v1/wallet_resources.py
```diff
@@ -0,0 +1,64 @@
...
+class PrepareNativeTransferResource(BaseMethodView):
+
+    post_schema = NativeAssetTransfer()
+
+    @use_kwargs(post_schema, location='json_and_view_args')
+    def post(
+            self,
+            from_address: ChecksumEvmAddress,
+            to_address: ChecksumEvmAddress,
+            blockchain: SupportedBlockchain,
```
prettyirrelevant, line 40:
```suggestion
            blockchain: 'SUPPORTED_EVM_CHAINS_TYPE',
```

docs/api.rst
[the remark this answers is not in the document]
yabirgb: image

rotkehlchen/chain/evm/active_management/manager.py
```diff
@@ -0,0 +1,113 @@
...
+            'nonce': web3_instance.eth.get_transaction_count(from_address),
+            **self._query_gas(web3_instance),
+        }
+
+    def transfer_native_token(
+            self,
+            from_address: 'ChecksumEvmAddress',
+            to_address: 'ChecksumEvmAddress',
+            amount: 'FVal',
+    ) -> dict[str, Any]:
```
LefterisJP, line 106: short docstring here to specify it's like the one for create token transfer?

rotkehlchen/chain/evm/node_inquirer.py
```diff
@@ -531,34 +531,29 @@ def _query(self, method: Callable, call_order: Sequence[WeightedNode], **kwargs:
...
-            if (
-                web3node is None and
-                node_info.name != self.etherscan_node_name and
-                node_info.name not in self.failed_to_connect_nodes
-            ):
-                success, _ = self.attempt_connect(node=node_info)
-                if success is False:
-                    self.failed_to_connect_nodes.add(node_info.name)
+            if web3node is None:
+                if node_info.name in self.failed_to_connect_nodes:
```
LefterisJP, line 535: is this a bugfix?

download (45)

rotkehlchen/tests/api/active_management/test_transfers.py
```diff
@@ -0,0 +1,111 @@
...
+    weight=ONE,
+),)])
+def test_transfers(
+        rotkehlchen_api_server: APIServer,
+        ethereum_accounts: list[ChecksumEvmAddress],
+        ethereum_manager_connect_at_start: list[WeightedNode],
+):
+    db = rotkehlchen_api_server.rest_api.rotkehlchen.data.db
+    from_address = ethereum_accounts[0]
+    to_address = string_to_evm_address('0x9531C059098e3d194fF87FebB587aB07B30B1306')
```
LefterisJP, line 37: combine them in oneline as we do in tests

rotkehlchen/chain/evm/active_management/manager.py
```diff
@@ -0,0 +1,113 @@
...
+
+        tx_data = dict(contract.functions.transfer(
+            to_address,
+            asset_raw_value(amount=amount, asset=token),
+        ).build_transaction({
+            'from': from_address,
+            'nonce': web3_instance.eth.get_transaction_count(from_address),
+        }))
+
+        tx_data |= self._query_gas(web3_instance)
```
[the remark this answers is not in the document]
yabirgb: I have removed the gas part since it gave us problems again. Luki is able to create the transaction without it taking default values

docs/api.rst
```diff
@@ -13689,4 +13689,152 @@ Refetch EVM transactions for a specific time period
...
+          "from_address": "0xc37b40ABdB939635068d3c5f13E7faF686F03B65",
+          "to_address": "0x9531C059098e3d194fF87FebB587aB07B30B1306",
+          "amount": "0.0003",
+          "chain_id": "ethereum"
+        }
+
+    :reqjson string from_address: The address from which the native cryptocurrency will be sent
+    :reqjson string to_address: The address to which the native cryptocurrency will be sent
+    :reqjson string amount: The amount of native cryptocurrency to transfer as a string
+    :reqjson string chain_id: The chain id name (e.g., "ethereum" for Ethereum)
```
LefterisJP, line 13772:
```suggestion
    :reqjson string chain: The EVM chain name (e.g., "ethereum" for Ethereum)
```

See how we use this elsewhere
```
The name of the chain for which to delete transaction. ``"eth"``, ``"optimism"``, ``"zksync_lite"`` etc. If not given all transactions for all chains are purged. This is using the backend's SupportedBlockchain with the limitation being only chains for which we save transactions.
```

docs/api.rst
```diff
@@ -13689,4 +13689,152 @@ Refetch EVM transactions for a specific time period
...
+            "nonce": 55
+          },
+          "status_code": 200
+        }
+
+    :resjson object result: The transaction data needed for the native transfer
+    :resjson string from: The sender's address
+    :resjson string to: The recipient's address
+    :resjson integer value: The amount to transfer in the smallest unit of the native cryptocurrency
+    :resjson integer maxFeePerGas: The maximum fee per gas unit the user is willing to pay
```
LefterisJP, line 13798: if you removed gas you need to remove it from here too. I am fine with having it remove for now, but "gave problems is not a good reason". We should be able to handle problems.

On the other hand I don't think gas is needed as it's something the user sets at the wallet anyway so am fine with the change. Having a way to set it already from rotki (the dapp) is not really common, though some dapps do. But in the end the wallet has the final say.

rotkehlchen/chain/evm/active_management/manager.py
```diff
@@ -0,0 +1,100 @@
...
+            to_address: 'ChecksumEvmAddress',
+            token: 'EvmToken',
+            amount: 'FVal',
+    ) -> dict[str, Any]:
+        """Build transaction to transfer erc20 tokens"""
+        contract = web3_instance.eth.contract(
+            address=token.evm_address,
+            abi=self.node_inquirer.contracts.erc20_abi,
+        )
+
```
LefterisJP, line 36:
```suggestion
```

docs/api.rst
yabirgb: was in the frontend side when sending it to the wallet. But as you said the wallet chooses it so no need to configure it. The error seems to come from how it is calculated in some chains

docs/api.rst
```diff
@@ -13689,4 +13689,148 @@ Refetch EVM transactions for a specific time period
...
+
+        POST /api/1/wallet/transfer/native HTTP/1.1
+        Host: localhost:5042
+        Content-Type: application/json;charset=UTF-8
+
+        {
+          "from_address": "0xc37b40ABdB939635068d3c5f13E7faF686F03B65",
+          "to_address": "0x9531C059098e3d194fF87FebB587aB07B30B1306",
+          "amount": "0.0003",
+          "chain_id": "ethereum"
```
LefterisJP, line 13766: you did not change it here

LefterisJP (approved):
LGTM

PR #9652, opened 2025-03-19
Improve rainbow decoder

yabirgb:
Closes #(issue_number)

## Checklist


rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -45,99 +45,16 @@ def __init__(
...
             transaction: EvmTransaction,
-            fee_raw: int,
+            fee_amount: FVal,
             fee_asset: CryptoAsset,
             sender: ChecksumEvmAddress,
             decoded_events: list['EvmEvent'],
             in_event: Optional['EvmEvent'],
             out_event: Optional['EvmEvent'],
-    ) -> None:
+    ) -> list['EvmEvent']:
```
nicholasyoder, line 57: The list of events returned from this function isn't being used anywhere as far as I can tell. So should still just return None.

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -308,74 +90,154 @@ def _process_swap(self,
...
+                    swapped_amount = from_wei(FVal(internal_tx.value))
+                    break
+        # When eth is swapped the fee is taken from the sent ETH. When ETH is received the
+        # free is taken from the received amount.
+        if (
+            out_event.asset == self.evm_inquirer.native_token and
+            (fee_amount := out_event.amount - swapped_amount) > 0
+        ):
+            out_event.amount -= fee_amount
+            decoded_events = self._create_and_append_fee_event(
```
nicholasyoder, line 174:
```suggestion
            self._create_and_append_fee_event(
```
Same for the other places this function is called.

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -308,74 +90,154 @@ def _process_swap(self,
...
+                    internal_tx.from_address == self.router_address and
+                    internal_tx.to_address != out_event.location_label
+                ) or (
+                    internal_tx.from_address != out_event.location_label and
+                    internal_tx.to_address == self.router_address
+                )):
+                    swapped_amount = from_wei(FVal(internal_tx.value))
+                    break
+        # When eth is swapped the fee is taken from the sent ETH. When ETH is received the
+        # free is taken from the received amount.
```
nicholasyoder, line 168:
```suggestion
        # fee is taken from the received amount.
```

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -308,74 +90,154 @@ def _process_swap(self,
...
+        swapped_amount = out_event.amount
+        # if we are dealing with eth swaps check the internal transfers of eth from/to
+        # the rainbow router
+        if self.evm_inquirer.native_token in {out_event.asset, in_event.asset}:
+            for internal_tx in DBEvmTx(self.evm_inquirer.database).get_evm_internal_transactions(
+                parent_tx_hash=transaction.tx_hash,
+                blockchain=self.evm_inquirer.blockchain,
+            ):
+                # the fee is the amount diff with the amount that the proxy sends to
+                # the swap solver
```
nicholasyoder, line 157: mmm... maybe this comment should be at a different place, or reworded or something? Its not actually finding the fee in this loop - its finding the swapped_amount.

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -308,74 +90,154 @@ def _process_swap(self,
...
+                    ) - in_event.amount) > 0
+                ):
+                    decoded_events = self._create_and_append_fee_event(
+                        transaction=transaction,
+                        fee_amount=fee_amount,
+                        fee_asset=in_asset,
+                        sender=out_event.location_label,  # type: ignore
+                        decoded_events=decoded_events,
+                        in_event=in_event,
+                        out_event=out_event,
```
nicholasyoder, line 238: Could `_create_and_append_fee_event` be called only once at the end of the function? It looks like your passing the same variable names for most of the arguments each time, except for the fee_asset. So you could have `fee_asset = None` at the beginning, and then set fee_asset each place you are currently calling this, and then do `if fee_asset is not None` at the end.

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -158,237 +75,165 @@ def _create_and_append_fee_event(
...
+            (possible_fee_amount := swapped_amount - in_event.amount) > 0
+        ):
+            fee_amount = possible_fee_amount
+            fee_asset = self.evm_inquirer.native_token
+        else:  # token transfer
+            # check in the logs the transfers made by the router to find if the fee was
+            # taken from the asset sent or the asset received
+            for log_event in all_logs:
+                if len(log_event.topics) != 3 or log_event.topics[0] == ERC20_OR_ERC721_TRANSFER:
+                    continue  # we look only for transfers
```
LefterisJP, line 188:
```suggestion
                    continue  # we only look for transfers
```

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -158,237 +75,165 @@ def _create_and_append_fee_event(
...
+
+                if (
+                    bytes_to_address(log_event.topics[1]) == self.router_address and
+                    bytes_to_address(log_event.topics[2]) != in_event.location_label and
+                    (possible_fee_amount := (out_event.amount - asset_normalized_value(
+                        amount=int.from_bytes(log_event.data),
+                        asset=(out_asset := out_event.asset.resolve_to_crypto_asset()),
+                    ))) > 0
+                ):
+                    # this is the router sending to the solver the token and it charged a fee
```
LefterisJP, line 198: save lines, plus rewording

```suggestion
                ):  # this is the router sending the token to the solver and charging a fee
```

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -158,237 +75,165 @@ def _create_and_append_fee_event(
...
+                    fee_amount = possible_fee_amount
+                    out_event.amount -= fee_amount
+                    fee_asset = out_asset
+                    break
+                elif (
+                    bytes_to_address(log_event.topics[1]) != in_event.location_label and  # from
```
LefterisJP, line 204: we do this for looping of all events to check for ERC20 transfer in way too many decoders lately. Can some of it be abstracted?

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -158,237 +75,165 @@ def _create_and_append_fee_event(
...
+                    fee_asset = in_asset
                     break
-        else:
-            log.error(
-                f'Failed to decode fee from rainbow swap token to token '
-                f'for transaction {transaction} due to sell_amount '
-                'or swap_amount being null.',
-            )
+
+        if fee_asset is None or fee_amount is None:
```
LefterisJP, line 215: Why not use for / else here? Don't you save the if check if you do so?

rotkehlchen/tests/unit/decoders/test_rainbow.py
```diff
@@ -88,7 +88,7 @@ def test_rainbow_swap_eth_to_token(ethereum_inquirer, ethereum_accounts):
 def test_rainbow_swap_token_to_eth(ethereum_inquirer, ethereum_accounts):
     tx_hash = deserialize_evm_tx_hash('0xd34d7393151a3c0f2d23d0df4d8f0b4a000be7613277e831ef0860191e68f855')  # noqa: E501
     events, _ = get_decoded_events_of_transaction(evm_inquirer=ethereum_inquirer, tx_hash=tx_hash)
-    swap_amount, received_amount, gas_fees, fee_amount, timestamp, user_address = '1010.887928111872496631', '0.089068967427375408', '0.000147805218572058', '0.00089852786738257', TimestampMS(1741892171000), ethereum_accounts[0]  # noqa: E501
+    swap_amount, received_amount, gas_fees, fee_amount, timestamp, user_address = '1010.887928111872496631', '0.089068967427375408', '0.000147805218572058', '0.000763576624440434', TimestampMS(1741892171000), ethereum_accounts[0]  # noqa: E501
```
LefterisJP, line 91: so these were wrong before?

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
yabirgb: I have modified it to info. I made it this way because I'm not sure if there is always a fee for all the swaps

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
yabirgb: I see what you mean. Not sure if there is anything that can be done since they depend on the context and you would need to pass custom logic just to abstract the for loop. not sure what would be a good path here

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
```diff
@@ -158,237 +75,164 @@ def _create_and_append_fee_event(
...
+                    continue  # we only look for transfers
+
+                if (
+                    bytes_to_address(log_event.topics[1]) == self.router_address and
+                    bytes_to_address(log_event.topics[2]) != in_event.location_label and
+                    (possible_fee_amount := (out_event.amount - asset_normalized_value(
+                        amount=int.from_bytes(log_event.data),
+                        asset=(out_asset := out_event.asset.resolve_to_crypto_asset()),
+                    ))) > 0
+                ):  # this is the router sending the token to the solver and charging a
```
LefterisJP, line 197: dude >.<

```suggestion
                ):  # this is the router sending the token to the solver and charging a fee
```

rotkehlchen/chain/evm/decoding/rainbow/decoder.py
LefterisJP: Then this needs to be debug. info means it gets logged always. No matter the log setting.

LefterisJP (approved):
LGTM

PR #9654, opened 2025-03-19
add price inquiry for curve savings token

prettyirrelevant:


rotkehlchen/chain/evm/decoding/curve_lend/utils.py
```diff
@@ -127,17 +127,17 @@ def query_curve_lending_vaults(database: 'DBHandler') -> None:
     )
 
 
-def get_curve_lending_vault_token_price(
+def get_curve_vault_token_price(
```
LefterisJP, line 130: will this now work for all curve vaults?

LefterisJP (commented):
Also I see you added a new token to the packaged global DB.

Is this also in the assets repo in an updates? Was it missing before?

LefterisJP (approved):
LGTM

PR #9656, opened 2025-03-20
Improve test coverage of SwapEvents and AssetMovements

nicholasyoder:
Related #6097 

Improves test coverage, and also fixes a couple things:
* Sets `reference` in asset movements extra_data from the unique_id. Without this, reference would have been lost if an AssetMovement was edited. The one drawback here is that if an AssetMovement that has a unique_id set but does not have reference set, is edited, reference will get set when it wasn't before. Not sure if there's a better way to do this.
* Allows `event_identifier` to be edited on AssetMovements


rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -42,14 +46,59 @@ def pop_multiple_keys(serialized_event: dict[str, Any], entry_type: HistoryBaseE
...
+            (grouped_entry := entries[0]).entry_type == (main_entry := group[0]).entry_type and
+            main_entry.event_type == grouped_entry.event_type and
+            grouped_entry.event_subtype == event_subtype
+        ):
+            group.append(entries.pop(0))
+
+        return group
+
+    try:
+        while True:
```
yabirgb, line 69: so as mentioned in discord this can probably be a group by identifier from itertools

rotkehlchen/api/v1/schemas.py
```diff
@@ -1004,15 +1004,19 @@ def make_history_base_entry(
             if (blockchain := data['blockchain']) is not None:
                 extra_data['blockchain'] = blockchain
 
+            if (unique_id := data['unique_id']) is not None:
+                extra_data['reference'] = unique_id
+
```
yabirgb, line 1009: this sounds like it should be a bugfixes PR with a check in a test (probably existing test)

nicholasyoder: Removed this fix from this PR and made a bugfix PR: https://github.com/rotki/rotki/pull/9659

rotkehlchen/tests/utils/history_base_entry.py
nicholasyoder: Simplified this logic as discussed in [discord](https://discord.com/channels/657906918408585217/1352348479380262973/1352621104866394163)

rotkehlchen/api/v1/schemas.py
```diff
@@ -1013,6 +1013,7 @@ def make_history_base_entry(
                 fee_asset=data['fee_asset'],
                 event_type=data['event_type'],
                 identifier=data.get('identifier'),
+                event_identifier=data.get('event_identifier'),
```
yabirgb, line 1016: in the other branch of the if/else you don't use get but the direct access. Shouldn't it be present always?

rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -42,14 +47,38 @@ def pop_multiple_keys(serialized_event: dict[str, Any], entry_type: HistoryBaseE
             serialized_event.pop(field)
 
 
+def maybe_group_entries(entries: list[HistoryBaseEntry]) -> list[list[HistoryBaseEntry]]:
+    """Group AssetMovements and SwapEvents by their event_identifier so they can be processed
+    as a unit when adding/editing via the api.
+
+    Can't simply group all entry types by the event_identifier since some other entry types also
+    have duplicate event_identifiers (used to test event grouping when querying history events).
+    """
```
yabirgb, line 56: this needs rewording to actually state why you need them grouped and you don't need evm_events to be grouped by identifier

rotkehlchen/api/v1/schemas.py
nicholasyoder: yes, should always be present. Fixed it.

rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -42,14 +47,38 @@ def pop_multiple_keys(serialized_event: dict[str, Any], entry_type: HistoryBaseE
...
+        else:
+            grouped_entries.append([entry])
+
+    grouped_entries.extend([list(g) for k, g in groupby(need_group_by_id, lambda entry: entry.event_identifier)])  # noqa: E501
+    return grouped_entries
+
+
 def entry_to_input_dict(
-        entry: 'HistoryBaseEntry',
+        entries: list[HistoryBaseEntry],
```
LefterisJP, line 72: You then need to rename this function. `entries_to_input_dict`

rotkehlchen/tests/utils/history_base_entry.py
```diff
@@ -165,6 +214,52 @@ def predefined_events_to_insert() -> list['HistoryBaseEntry']:
...
+        location=Location.COINBASE,
+        event_type=HistoryEventType.WITHDRAWAL,
+        asset=A_ETH,
+        amount=FVal('0.0586453'),
+        unique_id='MOVEMENT1',
+        location_label='Coinbase 1',
+        extra_data={
+            'address': '0x6dcD6449dbCa615e40d696328209686eA95327b2',
+            'transaction_id': '0x558bfa4d2a4ef598ddb92233459c00eda9e6c14cda75e6773b90208cb6938169',
+            # TODO: add `'reference': 'MOVEMENT1',` here after the bugfix for editing AssetMovements with references is merged into develop  # noqa: E501
```
LefterisJP, line 228: Why not just merge it now and do it?

LefterisJP (approved):
LGTM

PR #9659, opened 2025-03-20
Fix editing of AssetMovements with reference

nicholasyoder:
Fixing problem noticed in #9656


yabirgb (commented):
nitpick: it is missing a changelog entry

LefterisJP (approved):
LGTM. Pushed a changelog entry. Let's wait till tests are green.

PR #9666, opened 2025-03-21
feat(frontend): repulling transaction

### LINKED ISSUE #9592: Allow checking/repulling of transactions and history in a given period

lukicenturi:
Closes #9592

## Checklist


frontend/app/src/components/history/events/tx/RepullingTransactionFormDialog.vue
```diff
@@ -0,0 +1,107 @@
...
+  defineProps<{
+    loading?: boolean;
+  }>(),
+  {
+    loading: false,
+  },
+);
+
+const emit = defineEmits<{
+  (e: 'refresh'): void;
```
kelsos, line 25:
```suggestion
  refresh: [];
```

kelsos (approved):
I only have two minor comments @lukicenturi but otherwise it looks fine

Issue #3635, opened 2021-10-21
Implement CEX trade query recovery mechanism for all exchanges.


LefterisJP:
## Abstract

In https://github.com/rotki/rotki/pull/3634 a mechanism to recover from rate limiting for trades in a CEX is introduced but utilized properly only in kraken.

## Task

Implement the same for all other exchanges.

PR #9673, opened 2025-03-24
Use SwapEvents in Kraken

nicholasyoder:
Related: #6097

### DIFF
--- a/rotkehlchen/exchanges/kraken.py
+++ b/rotkehlchen/exchanges/kraken.py
@@ -9,6 +9,7 @@ import logging
 import operator
 import time
 from collections import defaultdict
+from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any, Literal
 from urllib.parse import urlencode
 
@@ -17,21 +18,16 @@ import requests
 from requests import Response
 
 from rotkehlchen.accounting.structures.balance import Balance
-from rotkehlchen.api.v1.types import IncludeExcludeFilterData
-from rotkehlchen.api.websockets.typedefs import HistoryEventsStep
 from rotkehlchen.assets.converters import asset_from_kraken
 from rotkehlchen.constants import KRAKEN_API_VERSION, KRAKEN_BASE_URL, ZERO
 from rotkehlchen.constants.assets import A_ETH2, A_KFEE, A_USD
-from rotkehlchen.constants.prices import ZERO_PRICE
 from rotkehlchen.db.constants import KRAKEN_ACCOUNT_TYPE_KEY
-from rotkehlchen.db.filtering import HistoryEventFilterQuery
 from rotkehlchen.db.history_events import DBHistoryEvents
-from rotkehlchen.db.ranges import DBQueryRanges
 from rotkehlchen.db.settings import CachedSettings
 from rotkehlchen.errors.asset import UnknownAsset
-from rotkehlchen.errors.misc import InputError, RemoteError
+from rotkehlchen.errors.misc import RemoteError
 from rotkehlchen.errors.serialization import DeserializationError
-from rotkehlchen.exchanges.data_structures import MarginPosition, Trade
+from rotkehlchen.exchanges.data_structures import MarginPosition
 from rotkehlchen.exchanges.exchange import (
     ExchangeInterface,
     ExchangeQueryBalances,
@@ -42,11 +38,12 @@ from rotkehlchen.history.events.structures.asset_movement import (
     create_asset_movement_with_fee,
 )
 from rotkehlchen.history.events.structures.base import (
-    HistoryBaseEntryType,
+    HistoryBaseEntry,
     HistoryEvent,
     HistoryEventSubType,
     HistoryEventType,
 )
+from rotkehlchen.history.events.structures.swap import SwapEvent, create_swap_events
 from rotkehlchen.inquirer import Inquirer
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.serialization.deserialize import (
@@ -56,14 +53,10 @@ from rotkehlchen.serialization.deserialize import (
 from rotkehlchen.types import (
     ApiKey,
     ApiSecret,
-    AssetAmount,
     ExchangeAuthCredentials,
-    Fee,
     Location,
-    Price,
     Timestamp,
     TimestampMS,
-    TradeType,
 )
 from rotkehlchen.utils.misc import pairwise, ts_ms_to_sec, ts_now
 from rotkehlchen.utils.mixins.cacheable import cache_response_timewise
@@ -72,9 +65,8 @@ from rotkehlchen.utils.mixins.lockable import protect_with_lock
 from rotkehlchen.utils.serialization import jsonloads_dict
 
 if TYPE_CHECKING:
-    from rotkehlchen.assets.asset import Asset, AssetWithOracles
+    from rotkehlchen.assets.asset import AssetWithOracles
     from rotkehlchen.db.dbhandler import DBHandler
-    from rotkehlchen.db.drivers.gevent import DBCursor
     from rotkehlchen.user_messages import MessagesAggregator
 
 
@@ -530,40 +522,6 @@ class Kraken(ExchangeInterface, ExchangeWithExtras):
 
         return result, with_errors
 
-    def query_online_trade_history(
-            self,
-            start_ts: Timestamp,
-            end_ts: Timestamp,
-    ) -> tuple[list[Trade], tuple[Timestamp, Timestamp]]:
-        """
-        Query kraken events from database and create trades from them. May raise:
-        - RemoteError if the kraken pairs couldn't be queried
-        """
-        with self.db.conn.read_ctx() as cursor:
-            with_errors = self.query_kraken_ledgers(cursor, start_ts=start_ts, end_ts=end_ts)
-            filter_query = HistoryEventFilterQuery.make(
-                from_ts=Timestamp(start_ts),
-                to_ts=Timestamp(end_ts),
-                event_types=[
-                    HistoryEventType.TRADE,
-                    HistoryEventType.RECEIVE,
-                    HistoryEventType.SPEND,
-                    HistoryEventType.ADJUSTMENT,
-                ],
-                location=Location.KRAKEN,
-                location_labels=[self.name],
-                entry_types=IncludeExcludeFilterData(values=[HistoryBaseEntryType.HISTORY_EVENT]),
-            )
-            trades_raw = self.history_events_db.get_history_events(
-                cursor=cursor,
-                filter_query=filter_query,
-                has_premium=True,
-            )
-
-        trades, max_ts = self.process_kraken_trades(trades_raw)  # type: ignore[arg-type]  # HistoryBaseEntry vs HistoryEvent
-        queried_range = (start_ts, Timestamp(max_ts)) if with_errors else (start_ts, end_ts)
-        return trades, queried_range
-
     def _query_endpoint_for_period(
             self,
             endpoint: Literal['Ledgers'],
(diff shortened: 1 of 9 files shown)


rotkehlchen/exchanges/kraken.py
```diff
@@ -592,12 +550,12 @@ def process_kraken_events_for_trade(
             self,
             trade_parts: list[HistoryEvent],
             adjustments: list[HistoryEvent],
-    ) -> Trade | None:
+    ) -> list[SwapEvent]:
```
LefterisJP, line 553: you would need to change the docstring here

rotkehlchen/exchanges/kraken.py
```diff
@@ -800,28 +725,25 @@ def process_kraken_trades(
...
-                    fee=None,
-                    fee_currency=None,
-                    link='adjustment' + a1.event_identifier + a2.event_identifier,
-                )
-                trades.append(trade)
+                    spend_asset=spend_event.asset,
+                    spend_amount=spend_event.amount,
+                    receive_asset=receive_event.asset,
+                    receive_amount=receive_event.amount,
+                    fee_asset=A_USD,  # whatever
```
LefterisJP, line 735: I don't understaht this "whatever" thing. If it can be whatever, then it should be `None`. Otherwise the logic makes no sense. Please fix this in the definition of `create_swap_events`. if should check if `fee_asset` is None`. And both `fee_asset` and `fee_amount` should be None/Noneable.

rotkehlchen/exchanges/kraken.py
```diff
@@ -865,86 +787,61 @@ def process_kraken_raw_events(
                 processed_refids.add(events[0]['refid'])
             new_events.extend(group_events)
 
-        if len(new_events) != 0:
```
LefterisJP, line 868: why did you remove this?

rotkehlchen/exchanges/kraken.py
```diff
@@ -865,86 +787,61 @@ def process_kraken_raw_events(
...
-        this function is called internally and not directly from the UI by the user.
+    ) -> tuple[Sequence['HistoryBaseEntry'], Timestamp]:
+        """Query Kraken's ledger to retrieve events and transform them to our
+        internal representation of history events.
+
         May raise:
         - RemoteError if request to kraken fails for whatever reason
 
-        Returns true if any query to the kraken API was not successful
+        Returns the events found and the actual ending timestamp in a tuple. Only returns an
```
LefterisJP, line 804: Explain what do you mean by the actual ending timestamp. Ending timestamp of what?

rotkehlchen/exchanges/kraken.py
```diff
@@ -865,86 +787,61 @@ def process_kraken_raw_events(
...
-                except InputError as e:  # not catching IntegrityError. event asset is resolved
-                    self.msg_aggregator.add_error(
-                        f'Failed to save kraken events from {events_source} in database. {e!s}',
-                    )
-
         return new_events, processed_refids
 
     @protect_with_lock()
-    def query_kraken_ledgers(
+    def maybe_query_online_history_events(
```
LefterisJP, line 793: To be honest I don't understand what you are doing here. Why did you create this new function in the base class and why did you use it here? The name does not explain much.

rotkehlchen/history/manager.py
```diff
@@ -357,9 +350,9 @@ def new_step_cb(state_name: str) -> None:
         history.extend(base_entries)
         self._increase_progress(step, total_steps)
 
-        history.sort(  # sort events first by timestamp and if history base by sequence index
+        history.sort(  # sort events first by timestamp (in milliseconds) and by sequence index if HistoryBaseEntry  # noqa: E501
             key=lambda x: (
-                x.get_timestamp(),
+                x.timestamp if isinstance(x, HistoryBaseEntry) else ts_sec_to_ms(x.get_timestamp()),  # noqa: E501
```
LefterisJP, line 355: you are doing the if check two times here. Do it outside of the sort and don't repeat it

rotkehlchen/history/skipped.py
```diff
@@ -109,11 +110,15 @@ def reprocess_skipped_external_events(rotki: 'Rotkehlchen') -> tuple[int, int]:
...
         exchange = cast('Kraken', exchange)
-        _, processed_refids = exchange.process_kraken_raw_events(
+        new_events, processed_refids = exchange.process_kraken_raw_events(
             events=[x[1] for x in raw_events],
             events_source='processing skipped events',
             save_skipped_events=False,
         )
+        if len(new_events) != 0:
+            with rotki.data.db.user_write() as write_cursor:
+                DBHistoryEvents(rotki.data.db).add_history_events(write_cursor=write_cursor, history=new_events)  # noqa: E501
```
LefterisJP, line 120: what's that? Why did this move here?

rotkehlchen/exchanges/exchange.py
```diff
@@ -383,8 +400,12 @@ def query_history_events(self) -> None:
                 ranges.update_used_query_range(
                     write_cursor=write_cursor,
                     location_string=location_string,
-                    queried_ranges=[(query_start_ts, query_end_ts)],
+                    queried_ranges=[(query_start_ts, actual_end_ts)],
                 )
+
+            if actual_end_ts != query_end_ts:
```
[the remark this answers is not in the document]
nicholasyoder: added a log here

rotkehlchen/exchanges/exchange.py
```diff
@@ -206,12 +206,29 @@ def query_online_history_events(
...
+    def maybe_query_online_history_events(
+            self,
+            start_ts: Timestamp,
+            end_ts: Timestamp,
+    ) -> tuple[Sequence['HistoryBaseEntry'], Timestamp]:
+        """Queries the exchange's API for history events of the user in the specified time range.
+        Returns the events found and the actual ending timestamp of what was queried if unable
+        to query the full range.
+
+        If this is not implemented in the subclass, query_online_history_events is called instead
```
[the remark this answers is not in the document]
nicholasyoder: added to the docstring to hopefully make this clearer.

rotkehlchen/exchanges/kraken.py
nicholasyoder: The way the exchange interface is intended to work, is that `query_history_events` handles adding the events to the db and only `query_online_history_events` or now also `maybe_query_online_history_events` is to be implemented for the exchange itself. Previously Kraken was re-implementing `query_history_events` and instead of saving the events there it was doing it here. Now I've reworked it to use the standard implementation of `query_history_events` from the exchange interface, so the saving of events to the database is now handled there and is not needed here.

rotkehlchen/exchanges/kraken.py
nicholasyoder: Its the last timestamp to be successfully queried and will be set as the ending timestamp of the queried range in `query_history_events`. Reworded the docstring here to make it a bit clearer.

rotkehlchen/exchanges/kraken.py
nicholasyoder: So I've tried to add to its docstring in the base class to explain better how it is used. Basically this is so that if there are errors partway through querying a range, it can still return the events it has gotten up to the point of the error, and then also return the last successfully queried timestamp to be set in the used query ranges so that next time it queries it begins where it left off instead of skipping anything between where the error was and the end_ts of the range. Perhaps there would be a better name for the function... not sure right off of anything better though.

rotkehlchen/history/manager.py
nicholasyoder: it does need to be inside the sort since its a check on the item being sorted - but I've changed it so it only does the check once in the lambda.

rotkehlchen/history/skipped.py
nicholasyoder: As mentioned [here](https://github.com/rotki/rotki/pull/9673#discussion_r2012201635) `query_history_events` now handles adding the events to the db instead of `process_kraken_raw_events`, but since we don't use the normal `query_history_events` logic flow here we have to write the events to the db here after using `process_kraken_raw_events`.

rotkehlchen/history/events/structures/swap.py
```diff
@@ -210,7 +210,7 @@ def create_swap_events(
         identifier=receive_identifier,
         event_identifier=event_identifier,
     )]
-    if fee_amount != ZERO:
+    if fee_asset is not None and fee_amount is not None and fee_amount != ZERO:
```
LefterisJP, line 213: no need to have three checks in this if. Just one. If `fee_asset is not None`. The zero check is also not needed as it's now `None` that matters. 

The rest you can do with python type hints. Essentially explain to the python type checker that the function can either accept both or None of these two. With the usual `@overload` thing.

Quick example by claude (I did not double check it too much):

```python
from typing import Optional, TypedDict, overload

class ConnectionParams(TypedDict, total=False):
    host: str
    port: int

class RelatedArgumentsExample:
    @overload
    def connect(self) -> None:
        """Overload for no arguments case"""
        ...

    @overload
    def connect(self, *, host: str, port: int) -> None:
        """Overload for both arguments case"""
        ...

    def connect(self, *, host: Optional[str] = None, port: Optional[int] = None) -> None:
        """
        Method that requires either both host and port or neither.
        
        Note: Type checking is handled statically by mypy
        """
        # Actual implementation can be minimal
        if host is not None and port is not None:
            print(f"Connecting to {host}:{port}")
        else:
            print("Using default connection")

# Mypy will catch these at type-checking time
def demonstrate_type_checking():
    example = RelatedArgumentsExample()
    
    # These are fine
    example.connect()  # No arguments
    example.connect(host="localhost", port=8000)  # Both arguments

    # These would be type errors caught by mypy
    # Uncomment to see type checking errors
    # example.connect(host="localhost")  # Mypy error
    # example.connect(port=8000)  # Mypy error
```

rotkehlchen/history/skipped.py
```diff
@@ -109,11 +110,15 @@ def reprocess_skipped_external_events(rotki: 'Rotkehlchen') -> tuple[int, int]:
 
         total_num += len(raw_events)
         exchange = cast('Kraken', exchange)
-        _, processed_refids = exchange.process_kraken_raw_events(
+        new_events, processed_refids = exchange.process_kraken_raw_events(
             events=[x[1] for x in raw_events],
             events_source='processing skipped events',
             save_skipped_events=False,
         )
+        if len(new_events) != 0:
```
LefterisJP, line 118: Discussed with Yabir that this logic is only in Kraken. It's ugly and probably does not scale. We have this old issue to implement in other exchanges but I don't see it happening: https://github.com/rotki/rotki/issues/3635


Not an action item for this PR, but perhaps we should try to see what would it mean to remove them.

rotkehlchen/tests/exchanges/test_kraken.py
```diff
@@ -168,21 +168,8 @@ def test_querying_balances(kraken):
...
-    )
-    assert isinstance(result, list)
-    assert len(result) != 0
-
-    for kraken_trade in result:
-        assert isinstance(kraken_trade, Trade)
-
-
+# TODO: Add support for partial range queries to all exchanges and re-enable this test
+@pytest.mark.skip('This functionality has been temporarily removed')
```
LefterisJP, line 172: you need to make a note for this or to handle it immediately after

nicholasyoder: Already added it to this: https://github.com/orgs/rotki/projects/11?pane=issue&itemId=102361655. Will see about implementing it soon.

rotkehlchen/history/events/structures/swap.py
```diff
@@ -171,6 +193,30 @@ def create_swap_events(
         receive_amount: 'FVal',
         fee_asset: Asset,
         fee_amount: 'FVal',
+        *,  # Require keyword arguments after this (so the overload works correctly)
```
LefterisJP, line 196: why not just put the fee_amount and fee_asset and fee_identifier and fee_notes as the last ones?

nicholasyoder: Changed it to 
```
        fee_asset: None = None,
        fee_amount: None = None,
```
on the no fee overload as discussed in discord.

LefterisJP (approved):
LGTM

Issue #9570, opened 2025-03-06
Support Curve crvUSD minting and peg keeping


nicholasyoder:
## Abstract

Need to support loans made via minting crvUSD rather than in the normal lending markets, and also correctly decode crvUSD peg keeper update transactions.

Mint markets: https://curve.fi/crvusd/#/ethereum/markets
Mint docs: https://docs.curve.fi/liquidity-gauges-and-minting-crv/minter/Minter/
Peg Keeper: https://curve.fi/crvusd/#/ethereum/pegkeepers
Peg docs: https://resources.curve.fi/crvusd/faq/#peg-keepers

Example txs:
* Borrow: https://etherscan.io/tx/0x9ddb6882b59b5d03159de7badc672ab186027c17ad724817d27190ca03695a23
* Repay: https://etherscan.io/tx/0x9d6c9cf285a22c9c9bdb03282cb64eaf5464ad68839870b0cff3e70fae02f51e
* Peg Keeper update: https://etherscan.io/tx/0xa76c3e49f9c2afa951ccb860ea092c9f5b710da0892002f221a2fe1989649912

More discussion about this can be found on [discord](https://discord.com/channels/657906918408585217/1347194175782780999)

## Specification

Add decoding of related events and ensure that balances are correctly tracked. 

The mint loan transactions seem to be relatively similar to the normal curve lend loans, but they don't use a vault contract for anything, which breaks most of the existing logic we have for curve loans.

PR #9675, opened 2025-03-24
Add crvusd loan decoder

nicholasyoder:
Related #9570  (still need to do peg keeping decoding and crvusd loan balances)

* Moves the existing curve lend decoder under evm/decoding/curve to keep all the curve stuff grouped under one directory
* Refactors the curve lend decoder so it uses `CurveBorrowRepayCommonDecoder` to handle the borrow/repay/collateral events
* Adds `CurvecrvusdDecoder` based on this common decoder in ethereum to add support for crvusd minting loan markets


rotkehlchen/chain/evm/decoding/curve/lend/common.py
```diff
@@ -0,0 +1,330 @@
...
+            if not (value := self.evm_inquirer.call_contract(
+                contract_address=contract_address,
+                abi=contract_abi,
+                method_name=contract_method,
+            )):
+                log.error(f'Empty response from {contract_method} method on Curve contract {contract_address}.')  # noqa: E501
+                return None
+
+            # Use to_checksum_address to fix lower case addresses
+            value = to_checksum_address(value)
```
prettyirrelevant, line 97: use `deserialize_evm_address()` and handle errors it can throw

rotkehlchen/chain/evm/decoding/curve/lend/common.py
```diff
@@ -0,0 +1,330 @@
...
+            contract_abi = CURVE_VAULT_ABI
+        else:
+            contract_method = 'borrowed_token'
+            contract_abi = CURVE_VAULT_ABI
+
+        if (token_address := self._maybe_get_cached_address_from_contract(
+            cache_type=cache_type,
+            contract_address=contract_address,
+            contract_abi=contract_abi,
+            contract_method=contract_method,  # type: ignore  # hardcoded to the correct methods above
```
prettyirrelevant, line 131: you can just define the variable instead of a type ignore.
`contract_method: Literal['method1', 'method2']`

rotkehlchen/chain/ethereum/modules/curve/crvusd/utils.py
```diff
@@ -0,0 +1,53 @@
...
+log = RotkehlchenLogsAdapter(logger)
+
+
+def query_crvusd_controllers(evm_inquirer: 'EvmNodeInquirer') -> None:
+    """Query crvUSD controller contract addresses from the minter contract and cache them"""
+    number_of_markets = (minter := EvmContract(
+        address=CRVUSD_MINTER,
+        abi=CRVUSD_MINTER_ABI,
+        deployed_block=0,
+    )).call(node_inquirer=evm_inquirer, method_name='n_collaterals')
```
prettyirrelevant, line 27: this can fail or throw a remote error. you should probably handle it.

rotkehlchen/chain/ethereum/modules/curve/crvusd/utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+    controllers = []
+    for idx, result in enumerate(evm_inquirer.multicall(
+        calls=[
+            (minter.address, minter.encode(method_name='controllers', arguments=[idx]))
+            for idx in range(number_of_markets)
+        ],
```
prettyirrelevant, line 34: same thing here.

rotkehlchen/chain/ethereum/modules/curve/crvusd/decoder.py
```diff
@@ -0,0 +1,170 @@
...
+            return None  # we didn't update the globaldb cache, and we have the data already
+
+        with GlobalDBHandler().conn.read_ctx() as cursor:
+            self.controllers = set(globaldb_get_general_cache_values(  # type: ignore[arg-type]  # strings from cache will be valid addresses
+                cursor=cursor,
+                key_parts=(
+                    CacheType.CURVE_CRVUSD_CONTROLLERS,
+                    str(self.evm_inquirer.chain_id.serialize_for_db()),
+                ),
+            ))
```
prettyirrelevant, line 67: since you're doing a `set()`, you can easily use set comprehension and use `string_to_evm_address()` -- no need for the type ignore.

rotkehlchen/chain/ethereum/modules/curve/crvusd/decoder.py
```diff
@@ -0,0 +1,170 @@
...
+        )
+
+    def maybe_decode_leveraged_borrow(self, context: DecoderContext) -> DecodingOutput | None:
+        """Decode events associated with creating a leveraged Curve position."""
+        if (tokens_and_amounts := self._get_controller_event_tokens_and_amounts(
+            controller_address=(controller_address := context.tx_log.address),
+            context=context,
+        )) is None:
+            log.error(
+                f'Failed to find tokens and amounts for Curve borrow transaction {context.transaction}')  # noqa: E501
```
prettyirrelevant, line 104:
```suggestion
            log.error(f'Failed to find tokens and amounts for Curve borrow transaction {context.transaction}')  # noqa: E501
```

rotkehlchen/chain/evm/decoding/curve/lend/common.py
```diff
@@ -0,0 +1,330 @@
...
+                key_parts=(cache_type, str(contract_address)),
+            )) is not None:
+                return string_to_evm_address(value)
+
+        try:
+            if not (value := self.evm_inquirer.call_contract(
+                contract_address=contract_address,
+                abi=contract_abi,
+                method_name=contract_method,
+            )):
```
prettyirrelevant, line 92: you probably want to use `value != ZERO_ADDRESS` instead of what you currently have

rotkehlchen/chain/evm/decoding/curve/lend/common.py
```diff
@@ -0,0 +1,330 @@
...
+                log.error(f'Empty response from {contract_method} method on Curve contract {contract_address}.')  # noqa: E501
```
prettyirrelevant, line 93: this log will needs adjusting in that case.

rotkehlchen/chain/evm/decoding/curve/lend/common.py
```diff
@@ -0,0 +1,330 @@
...
+                asset=token,
+                amount=amount,
+                to_event_subtype=HistoryEventSubType.PAYBACK_DEBT,
+                to_notes=f'Repay {amount} {token.symbol} to Curve',
+                to_counterparty=CPT_CURVE,
+                paired_events_data=([in_event], False) if in_event is not None else None,
+            ) for token, amount in [
+                (collateral_token, collateral_amount),
+                (borrowed_token, borrowed_amount),
+            ]
```
prettyirrelevant, line 284: the indentation here makes it hard to follow.

rotkehlchen/chain/evm/decoding/curve/lend/utils.py
```diff
@@ -101,11 +101,11 @@ def _process_curve_lending_vault(database: 'DBHandler', vault: dict[str, Any]) -
         globaldb_set_unique_cache_value(
             write_cursor=write_cursor,
             key_parts=[CacheType.CURVE_LENDING_VAULT_CONTROLLER, vault_token.evm_address],
-            value=vault['controllerAddress'],
+            value=(controller_address := vault['controllerAddress']),
         )
         globaldb_set_unique_cache_value(
             write_cursor=write_cursor,
-            key_parts=[CacheType.CURVE_LENDING_VAULT_AMM, vault_token.evm_address],
+            key_parts=[CacheType.CURVE_CRVUSD_AMM, controller_address],
```
prettyirrelevant, line 108: you need to clear existing cache.

rotkehlchen/types.py
```diff
@@ -1186,6 +1184,9 @@ class CacheType(Enum):
     AERODROME_GAUGE_FEE_ADDRESS = auto()
     AERODROME_GAUGE_BRIBE_ADDRESS = auto()
     CURVE_LENDING_VAULT_GAUGE = auto()
+    CURVE_CRVUSD_CONTROLLERS = auto()
+    CURVE_CRVUSD_COLLATERAL_TOKEN = auto()
+    CURVE_CRVUSD_AMM = auto()
```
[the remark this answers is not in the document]
nicholasyoder: Added to the globaldb upgrade to clear these caches.

rotkehlchen/chain/evm/decoding/curve/lend/common.py
nicholasyoder: So actually I don't think this function needs to check for ZERO_ADDRESS - the `if` here was checking for a blank response. But this `if not` is removed since now `deserialize_evm_address` will do a `DeserializationError` if it gets an empty response.

rotkehlchen/chain/evm/decoding/curve/lend/common.py
nicholasyoder: mmm... I'm not sure how to do this better - Do you have a suggestion for how it should be formatted? This is how it was before: https://github.com/nicholasyoder/rotki/blob/develop/rotkehlchen/chain/evm/decoding/curve_lend/decoder.py#L584

prettyirrelevant: you can leave it as it is.

rotkehlchen/chain/ethereum/modules/curve/crvusd/utils.py
```diff
@@ -0,0 +1,58 @@
...
+            for idx in range(number_of_markets)
+        ])
+    except RemoteError as e:
+        log.error(f'Failed to query crvUSD controllers from the minter contract due to {e!s}')
+        return
+
+    controllers = []
+    for idx, result in enumerate(controllers_result):
+        try:
+            controllers.append(deserialize_evm_address(minter.decode(
```
prettyirrelevant, line 42: only add it to the list if it's not a zero address

rotkehlchen/chain/evm/decoding/curve/lend/common.py
```diff
@@ -0,0 +1,327 @@
...
+            contract_address: 'ChecksumEvmAddress',
+            contract_abi: ABI,
+            contract_method: Literal['amm', 'controller', 'collateral_token', 'borrowed_token'],
+    ) -> ChecksumEvmAddress | None:
+        """Get address from the cache or the specified vault contract method.
+        Returns the address or None on error."""
+        with GlobalDBHandler().conn.read_ctx() as cursor:
+            if (value := globaldb_get_unique_cache_value(
+                cursor=cursor,
+                key_parts=(cache_type, str(contract_address)),
```
prettyirrelevant, line 84:
```suggestion
                key_parts=(cache_type, contract_address),
```

rotkehlchen/chain/ethereum/modules/curve/crvusd/utils.py
nicholasyoder: Changed it to log an error and skip adding it to the list if its a zero address. I checked and the contract here does return zero address for invalid indexes, so its good you noticed this (although theoretically `n_collaterals` shouldn't ever return invalid indexes).

rotkehlchen/chain/evm/decoding/curve/lend/common.py
```diff
@@ -0,0 +1,327 @@
...
+                return string_to_evm_address(value)
+
+        try:
+            value = deserialize_evm_address(self.evm_inquirer.call_contract(
+                contract_address=contract_address,
+                abi=contract_abi,
+                method_name=contract_method,
+            ))
+        except (RemoteError, DeserializationError) as e:
+            log.error(f'Failed to retrieve an evm address from the {contract_method} method on Curve contract {contract_address} due to {e!s}')  # noqa: E501
```
LefterisJP, line 95: you need the chain here too

rotkehlchen/chain/evm/decoding/curve/lend/common.py
```diff
@@ -0,0 +1,327 @@
...
+            ActionItem(
+                action='transform',
+                from_event_type=HistoryEventType.SPEND,
+                from_event_subtype=HistoryEventSubType.NONE,
+                asset=token,
+                amount=amount,
+                to_event_subtype=HistoryEventSubType.PAYBACK_DEBT,
+                to_notes=f'Repay {amount} {token.symbol} to Curve',
+                to_counterparty=CPT_CURVE,
+                paired_events_data=([in_event], False) if in_event is not None else None,
```
LefterisJP, line 277: this if is done two times right? It should be out of the loop.

rotkehlchen/chain/ethereum/modules/curve/crvusd/decoder.py
```diff
@@ -0,0 +1,170 @@
...
+            cache_type=CacheType.CURVE_CRVUSD_COLLATERAL_TOKEN,
+            contract_address=controller_address,
+        )) is None:
+            return None
+
+        return (
+            collateral_token,
+            self.crvusd,
+            token_normalized_value(
+                token_amount=int.from_bytes(context.tx_log.data[:32]),
```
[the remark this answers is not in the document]
nicholasyoder: yeah this would get handled at the top level of decoding. Added a 'May raise' to the docstring.

rotkehlchen/chain/ethereum/modules/curve/crvusd/utils.py
```diff
@@ -0,0 +1,66 @@
...
+            abi=CRVUSD_MINTER_ABI,
+            deployed_block=0,
+        )).call(node_inquirer=evm_inquirer, method_name='n_collaterals')
+
+        controllers_result = evm_inquirer.multicall(calls=[
+            (minter.address, minter.encode(method_name='controllers', arguments=[idx]))
+            for idx in range(number_of_markets)
+        ])
+    except RemoteError as e:
+        log.error(f'Failed to query crvUSD controllers from the minter contract due to {e!s}')
```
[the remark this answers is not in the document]
nicholasyoder: > only if multiple per chain
There is only one minter contract (the `CRVUSD_MINTER` constant here) and this is only on Ethereum mainnet. So skipping this (and leaving the other log messages as is here as well).

rotkehlchen/chain/evm/decoding/curve/lend/common.py
```diff
@@ -0,0 +1,327 @@
...
+            msg_aggregator: 'MessagesAggregator',
+            leverage_zap: 'ChecksumEvmAddress | None' = None,
+    ) -> None:
+        """Decoder for Curve borrow/repay events."""
+        super().__init__(
+            evm_inquirer=evm_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+        )
+        self.controllers: set[ChecksumEvmAddress] = set()
```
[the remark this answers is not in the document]
nicholasyoder: In the subclasses' reload_data function. Added a comment here noting this.

rotkehlchen/chain/evm/decoding/curve/lend/common.py
nicholasyoder: Added the chain here. In contrast to the logs in crvusd/utils, it can be on multiple chains here since this is used by the lend decoder as well.

PR #9676, opened 2025-03-25
Bugfixes/9157

### LINKED ISSUE #9157: Change datepicker input and output format
`<DateTimePicker />` component in frontend now accept and emit the date in string format: DD/MM/YYYY

It's better to change it to timestamp instead, since we always use timestamp, and we always do the conversion manually on the parent, so we can cut some complexity on the parent.

BRp99:
Closes #9157

## Description

This PR refactors the DateTimePicker to use timestamp and update related components and tests:

- refactored the DateTimePicker to accept and display the date as a timestamp instead of a string, supporting milliseconds and second;

- updated components that utilize DateTimePicker to reflect these changes and ensure compatibility with the new timestamp format;

- change unit tests to expect a number (timestamp) instead of a string.

### DIFF
--- a/frontend/app/src/components/asset-manager/managed/ManagedAssetForm.vue
+++ b/frontend/app/src/components/asset-manager/managed/ManagedAssetForm.vue
@@ -14,7 +14,6 @@ import { useSupportedChains } from '@/composables/info/chains';
 import { useMessageStore } from '@/store/message';
 import { CUSTOM_ASSET, EVM_TOKEN } from '@/types/asset';
 import { evmTokenKindsData } from '@/types/blockchain/chains';
-import { convertFromTimestamp, convertToTimestamp } from '@/utils/date';
 import { refOptional, useRefPropVModel } from '@/utils/model';
 import { toMessages } from '@/utils/validation';
 import { isValidEthAddress, onlyIfTruthy, type SupportedAsset, toSentenceCase, type UnderlyingToken } from '@rotki/common';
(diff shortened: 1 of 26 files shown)


kelsos:
Hey @BRp99 thank you for your PR.

I am not sure if it is your editor or something, but there seem to be a lot of changes in the PR that should not be there.

for example I can see a lot of the following, which makes it harder to review the PR
- changes of single quotes to double quotes
- removal of newlines in code and templates

Can you try `pnpm run lint --fix` to see what can be auto-fixed?
Also it would be nice if you make sure that the PR does not include any unnecessary quote and whitespace changes in general.

Also if possible
```typescript 
// instead of 
ref<number | null>(null); 
// use 
ref<number>();
```

The second is the preferred approach the first is mostly leftover technical debt.

Also if possible could you squash all the commits to a single commit?

BRp99:
Hi Kelsos,

Thanks for the feedback! I’ve already run pnpm run lint --fix, but I’m still seeing the same issues with formatting (single quotes to double quotes). Could you let me know how you format the code on your side so I can align with it and ensure consistency?

Additionally, I’ll make sure to avoid unnecessary quote and whitespace changes in the PR moving forward. Regarding the ref<number>() suggestion, I’ll update the code accordingly as well. Once everything is resolved, I’ll squash all the commits into a single one as requested.

lukicenturi:
Thanks for the help, but this issue will be covered in https://github.com/rotki/rotki/pull/9711

lukicenturi:
I will close this issue.

PR #9678, opened 2025-03-25
chore: improve type safety by changing `catch` error type to `unknown`

fragchain:
In the `catch` block, the type of the `error` variable has been changed from `any` to `unknown`. This is a safer approach, as the error could potentially be an instance of something other than `Error`.


kelsos:
Hey @fragchain thank you for your PR, but this goes against our [policy](https://github.com/rotki/rotki/blob/develop/CONTRIBUTING.md) for low effort PRs.

PR #9681, opened 2025-03-25
feat: decode deposit and withdrawal for StakeDAO

prodesert22:
## Summary
- Decode deposit and with withdrawal for StakeDAO (for ethereum).

Other chains use a different type of contract. that will come in a different PR together with balances(and correct prices)

### DIFF
--- a/rotkehlchen/chain/evm/decoding/stakedao/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/stakedao/decoder.py
@@ -1,0 +1,293 @@
+import logging
+from collections.abc import Mapping
+from typing import TYPE_CHECKING, Any
+
+from rotkehlchen.assets.utils import get_or_create_evm_token
+from rotkehlchen.chain.ethereum.utils import (
+    asset_normalized_value,
+    asset_raw_value,
+    should_update_protocol_cache,
+    token_normalized_value,
+)
+from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
+from rotkehlchen.chain.evm.decoding.constants import ERC20_OR_ERC721_TRANSFER
+from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface, ReloadableDecoderMixin
+from rotkehlchen.chain.evm.decoding.stakedao.utils import (
+    query_stakedao_gauges,
+)
+from rotkehlchen.chain.evm.decoding.structures import (
+    DEFAULT_DECODING_OUTPUT,
+    ActionItem,
+    DecoderContext,
+    DecodingOutput,
+)
+from rotkehlchen.chain.evm.decoding.types import CounterpartyDetails
+from rotkehlchen.chain.evm.decoding.utils import maybe_reshuffle_events
+from rotkehlchen.globaldb.cache import globaldb_get_general_cache_values
+from rotkehlchen.globaldb.handler import GlobalDBHandler
+from rotkehlchen.history.events.structures.evm_event import EvmProduct
+from rotkehlchen.history.events.structures.types import (
+    HistoryEventSubType,
+    HistoryEventType,
+)
+from rotkehlchen.logging import RotkehlchenLogsAdapter
+from rotkehlchen.serialization.deserialize import deserialize_evm_address
+from rotkehlchen.types import CacheType, ChecksumEvmAddress, Timestamp
+from rotkehlchen.utils.misc import bytes_to_address, timestamp_to_date
+
+from .constants import (
+    CLAIMED_WITH_BOUNTY,
+    CLAIMED_WITH_BRIBE,
+    CPT_STAKEDAO,
+    STAKEDAO_DEPOSIT,
+    STAKEDAO_GAUGE_ABI,
+    STAKEDAO_VAULT_ABI,
+    STAKEDAO_WITHDRAW,
+)
+
+if TYPE_CHECKING:
+    from rotkehlchen.chain.evm.decoding.base import BaseDecoderTools
+    from rotkehlchen.chain.evm.node_inquirer import EvmNodeInquirer
+    from rotkehlchen.user_messages import MessagesAggregator
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+
+class StakedaoCommonDecoder(DecoderInterface, ReloadableDecoderMixin):
+    def __init__(
+            self,
+            evm_inquirer: 'EvmNodeInquirer',
+            base_tools: 'BaseDecoderTools',
+            msg_aggregator: 'MessagesAggregator',
+            claim_bribe_addresses: set['ChecksumEvmAddress'] | None = None,
+            claim_bounty_addresses: set['ChecksumEvmAddress'] | None = None,
+    ):
+        super().__init__(evm_inquirer, base_tools, msg_aggregator)
+        self.claim_bribe_addresses = claim_bribe_addresses
+        self.claim_bounty_addresses = claim_bounty_addresses
+        self.gauges: set[ChecksumEvmAddress] = set()
+
+    def reload_data(self) -> Mapping['ChecksumEvmAddress', tuple[Any, ...]] | None:
+        """Check that cache is up to date and refresh cache from db.
+        Returns a fresh addresses to decoders mapping.
+        """
+        if should_update_protocol_cache(
+            cache_key=CacheType.STAKEDAO_GAUGES,
+            args=(str(self.evm_inquirer.chain_id.value),),
+        ) is True:
+            query_stakedao_gauges()
+
+        if len(self.gauges) != 0:
+            return None  # we didn't update the globaldb cache, and we have the data already
+
+        with GlobalDBHandler().conn.read_ctx() as cursor:
+            self.gauges = set(globaldb_get_general_cache_values(  # type: ignore[arg-type]  # addresses are always checksummed
+                cursor=cursor,
+                key_parts=(CacheType.STAKEDAO_GAUGES, str(self.evm_inquirer.chain_id.value)),
+            ))
+
+        return self.addresses_to_decoders()
+
+    def _decode_claim(
+            self,
+            context: DecoderContext,
+            reward_token_address: ChecksumEvmAddress,
+            amount: int,
+            period: Timestamp,
+    ) -> DecodingOutput:
+        """Base functionality for claiming different types of stakedao votemarket bribes
+
+        Note: We don't check the user address in the logs as user is not always
+        the recipient according to the contract.
+        """
+        claimed_token = get_or_create_evm_token(
+            userdb=self.base.database,
+            evm_address=reward_token_address,
+            chain_id=self.evm_inquirer.chain_id,
+            evm_inquirer=self.evm_inquirer,
+        )
+        normalized_amount = token_normalized_value(amount, claimed_token)
+        for event in context.decoded_events:
+            if event.event_type == HistoryEventType.RECEIVE and event.event_subtype == HistoryEventSubType.NONE and event.asset == claimed_token and event.amount == normalized_amount:  # noqa: E501
+                event.event_subtype = HistoryEventSubType.REWARD
+                event.counterparty = CPT_STAKEDAO
+                event.notes = f'Claim {normalized_amount} {claimed_token.symbol} from StakeDAO veCRV bribes for the period starting at {timestamp_to_date(period, formatstr="%d/%m/%Y %H:%M:%S")}'  # noqa: E501
+                event.product = EvmProduct.BRIBE
+                break
+        else:  # not found
+            log.error(f'Stakedao bribe transfer was not found for {context.transaction}')
+
+        return DEFAULT_DECODING_OUTPUT
+
+    def _decode_claim_with_bounty(self, context: DecoderContext) -> DecodingOutput:
+        if context.tx_log.topics[0] != CLAIMED_WITH_BOUNTY:
+            return DEFAULT_DECODING_OUTPUT
+
+        reward_token_address = bytes_to_address(context.tx_log.data[0:32])
+        amount = int.from_bytes(context.tx_log.data[32:64])
+        period = Timestamp(int.from_bytes(context.tx_log.data[96:128]))
+        return self._decode_claim(context=context, reward_token_address=reward_token_address, amount=amount, period=period)  # noqa: E501
+
+    def _decode_claim_with_bribe(self, context: DecoderContext) -> DecodingOutput:
+        if context.tx_log.topics[0] != CLAIMED_WITH_BRIBE:
+            return DEFAULT_DECODING_OUTPUT
+
+        reward_token_address = bytes_to_address(context.tx_log.topics[2])
+        amount = int.from_bytes(context.tx_log.data[0:32])
+        period = Timestamp(int.from_bytes(context.tx_log.data[32:64]))
+        return self._decode_claim(context=context, reward_token_address=reward_token_address, amount=amount, period=period)  # noqa: E501
+
+    def _decode_deposit(self, context: DecoderContext) -> DecodingOutput:
+        deposited_raw_amount = int.from_bytes(context.tx_log.data[:32])
+        received_amount = asset_normalized_value(
+            amount=deposited_raw_amount,
+            asset=(received_asset := self.base.get_or_create_evm_token(context.tx_log.address)),
+        )
+        for event in context.decoded_events:
+            if (
+                    # we compare raw amounts and not the received amount because
+                    # the event does not include the token and normalizing could
+                    # be wrong if the token uses different decimals.
+                    event.event_type == HistoryEventType.SPEND and
+                    event.event_subtype == HistoryEventSubType.NONE and
+                    asset_raw_value(
+                        amount=event.amount,
+                        asset=(deposited_asset := event.asset.resolve_to_crypto_asset()),
+                    ) == deposited_raw_amount
+            ):
+                event.counterparty = CPT_STAKEDAO
+                event.event_type = HistoryEventType.DEPOSIT
+                event.event_subtype = HistoryEventSubType.DEPOSIT_FOR_WRAPPED
+                event.notes = f'Deposit {event.amount} {deposited_asset.symbol} in StakeDAO'
+                break
+        else:
+            log.error(f'Could not find stakedao deposit event for transaction {context.transaction}')  # noqa: E501
+            return DEFAULT_DECODING_OUTPUT
+
+        return DecodingOutput(action_items=[ActionItem(
+            action='transform',
+            from_event_type=HistoryEventType.RECEIVE,
+            from_event_subtype=HistoryEventSubType.NONE,
+            to_event_type=HistoryEventType.RECEIVE,
+            to_event_subtype=HistoryEventSubType.RECEIVE_WRAPPED,
+            amount=received_amount,
+            asset=received_asset,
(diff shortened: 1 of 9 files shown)


rotkehlchen/tests/unit/decoders/test_stakedao.py
```diff
@@ -157,3 +157,120 @@ def test_claim_multiple(ethereum_inquirer, ethereum_accounts):
...
+            location=Location.ETHEREUM,
+            event_type=HistoryEventType.WITHDRAWAL,
+            event_subtype=HistoryEventSubType.REDEEM_WRAPPED,
+            asset=Asset('eip155:1/erc20:0x460638e6F7605B866736e38045C0DE8294d7D87f'),
+            amount=FVal(received_amount),
+            location_label=user_address,
+            notes=f'Withdraw {received_amount} alUSDsDOLA from StakeDAO vault',
+            counterparty=CPT_STAKEDAO,
+            address=vault,
+        ),
```
prettyirrelevant, line 275: aren't you missing the event that returns the gauge token back to the vault?

gauge token goes out ->
-> normal token comes in

prettyirrelevant (commented):
left some comments. 
also, you forgot to extend this to other supported chains (OP, BSC, etc).

rotkehlchen/chain/ethereum/modules/stakedao/utils.py
```diff
@@ -0,0 +1,81 @@
...
+from rotkehlchen.globaldb.handler import GlobalDBHandler
+from rotkehlchen.logging import RotkehlchenLogsAdapter
+from rotkehlchen.serialization.deserialize import deserialize_evm_address
+from rotkehlchen.types import CacheType, ChainID
+
+logger = logging.getLogger(__name__)
+log = RotkehlchenLogsAdapter(logger)
+
+STAKEDAO_API: Final = 'https://api.stakedao.org/api'
+PROTOCOLS: Final = {'curve', 'balancer', 'angle'}
```
prettyirrelevant, line 19: https://github.com/stake-dao/api?tab=readme-ov-file#strategies-data
there are more strategies in the link above. why did you add just three?

rotkehlchen/chain/ethereum/modules/stakedao/utils.py
```diff
@@ -0,0 +1,81 @@
...
+            )
+
+            for item in response_data.json()['deployed']:
+                try:
+                    chain_id = ChainID.deserialize(item['chainId'])
+                except DeserializationError:
+                    log.warning(f"Skipping invalid ChainID value: {item['chainId']}")
+                    continue
+
+                strategies_gauges[chain_id].add(deserialize_evm_address(item['gaugeAddress']))
```
prettyirrelevant, line 69: no no. this is curve's gauge address. use `item["sdGauge"]["address"]` instead.

rotkehlchen/chain/arbitrum_one/modules/stakedao/decoder.py
```diff
@@ -0,0 +1,24 @@
...
+            evm_inquirer: 'EvmNodeInquirer',
+            base_tools: 'BaseDecoderTools',
+            msg_aggregator: 'MessagesAggregator',
+    ):
+        super().__init__(
+            evm_inquirer=evm_inquirer,
+            base_tools=base_tools,
+            msg_aggregator=msg_aggregator,
+            claim_bribe_addresses=[],
+            claim_bounty_addresses=[],
```
nicholasyoder, line 23: Should probably make these optional arguments in `StakedaoCommonDecoder` and then you wouldn't need to define `__init__` here and in the other chains where you just pass empty lists for these.

rotkehlchen/tests/unit/decoders/test_stakedao.py
```diff
@@ -157,3 +178,135 @@ def test_claim_multiple(ethereum_inquirer, ethereum_accounts):
...
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('ethereum_accounts', [['0x76d5eb42A854A1cEAfFD99000341d4E4e7a4a70F']])
+def test_deposit(ethereum_inquirer, ethereum_accounts, stakedao_gauges):
+    tx_hex = deserialize_evm_tx_hash('0x0b98f04aeeaa4068b8c8ae0568ed236537c3573b4c3e6fd6b1924741cd5c9ef5')  # noqa: E501
+    evmhash = deserialize_evm_tx_hash(tx_hex)
+    user_address = ethereum_accounts[0]
+    events, _ = get_decoded_events_of_transaction(evm_inquirer=ethereum_inquirer, tx_hash=tx_hex)
+    gas_amount, deposit_amount, timestamp = '0.000333982538036425', '29741.066052414178579757', TimestampMS(1742908919000)  # noqa: E501
+    expected_events = [
+        EvmEvent(
```
nicholasyoder, line 192:
```suggestion
    expected_events = [EvmEvent(
```
styling nitpick: I prefer doing it like this to avoid having it indented more than needed. Also you could even do the assert on this line as well and avoid the extra line for the assert at the end.

rotkehlchen/globaldb/cache.py
```diff
@@ -94,6 +94,7 @@ def globaldb_get_general_cache_values(
     """Function that reads from the general cache table.
     It returns all the values that are paired with the given key."""
     cursor.execute('SELECT value FROM general_cache WHERE key=?', (compute_cache_key(key_parts),))
+
```
nicholasyoder, line 97: no need for this new blank line.

rotkehlchen/chain/evm/decoding/stakedao/decoder.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+class StakedaoCommonDecoder(DecoderInterface, ReloadableDecoderMixin):
+    def __init__(
+            self,
+            evm_inquirer: 'EvmNodeInquirer',
+            base_tools: 'BaseDecoderTools',
+            msg_aggregator: 'MessagesAggregator',
+            claim_bribe_addresses: list['ChecksumEvmAddress'],
+            claim_bounty_addresses: list['ChecksumEvmAddress'],
```
nicholasyoder, line 64: Should probably use sets instead of lists for these, as well as for `self.gauges` below.

rotkehlchen/chain/evm/decoding/stakedao/decoder.py
```diff
@@ -0,0 +1,279 @@
...
+            asset=(received_asset := self.base.get_or_create_evm_token(context.tx_log.address)),
+        )
+        for event in context.decoded_events:
+            if (
+                    event.event_type == HistoryEventType.SPEND and
+                    event.event_subtype == HistoryEventSubType.NONE and
+                    asset_raw_value(
+                        amount=event.amount,
+                        asset=(deposited_asset := event.asset.resolve_to_crypto_asset()),
+                    ) == deposited_raw_amount
```
nicholasyoder, line 153: So it looks like at least in your test the deposited amount and the received amount are the same. Can't you just check `event.amount` against what you call `received_amount` here? also is there any way to check the event asset?

rotkehlchen/chain/evm/decoding/stakedao/decoder.py
```diff
@@ -0,0 +1,279 @@
...
+            amount=received_amount,
+            asset=received_asset,
+            to_notes=f'Receive {received_amount} {received_asset.symbol} after depositing in StakeDAO',  # noqa: E501
+            to_counterparty=CPT_STAKEDAO,
+        )])
+
+    def _decode_withdraw(self, context: DecoderContext) -> DecodingOutput:
+        # find the vault behind the gauge, then get the underlying token of that vault
+        received_token = self.base.get_or_create_evm_token(deserialize_evm_address(self.evm_inquirer.call_contract(  # noqa: E501
+            contract_address=(vault_address := deserialize_evm_address(self.evm_inquirer.call_contract(  # noqa: E501
```
nicholasyoder, line 179: Would there be any way to do this without making these remote calls? It does look like maybe we do contract calls like this in a couple other decoders but its probably something that would be best if we could avoid.

If this is necessary at least move it to closer to the end of the function. As far as I can tell `received_token` and `received_amount` that are being retrieved here aren't used till the action item at the end. There's no point in making these remote calls at the beginning when it is possible for the function to return before using the data.

rotkehlchen/chain/evm/decoding/stakedao/decoder.py
```diff
@@ -0,0 +1,279 @@
...
+            abi=STAKEDAO_VAULT_ABI,
+            method_name='token',
+        )))
+        received_amount = asset_normalized_value(
+            amount=(removed_raw_amount := int.from_bytes(context.tx_log.data[:32])),
+            asset=received_token,
+        )
+        claim_events = []
+        for event in context.decoded_events:
+            if event.event_type == HistoryEventType.RECEIVE and event.event_subtype == HistoryEventSubType.NONE:  # noqa: E501
```
nicholasyoder, line 193: This can surely check a few more things? We don't want it to match unrelated receive events in complex transactions.

rotkehlchen/chain/evm/decoding/stakedao/decoder.py
```diff
@@ -0,0 +1,279 @@
...
+            return self._decode_withdraw(context)
+        return DEFAULT_DECODING_OUTPUT
+
+    # -- DecoderInterface methods
+
+    @staticmethod
+    def possible_products() -> dict[str, list[EvmProduct]]:
+        return {
+            CPT_STAKEDAO: [EvmProduct.BRIBE],
+        }
```
nicholasyoder, line 264:
```suggestion
        return {CPT_STAKEDAO: [EvmProduct.BRIBE]}
```

rotkehlchen/chain/evm/decoding/stakedao/utils.py
```diff
@@ -0,0 +1,60 @@
...
+log = RotkehlchenLogsAdapter(logger)
+
+STAKEDAO_API: Final = 'https://api.stakedao.org/api'
+# This is gotten from:
+# https://github.com/stake-dao/api?tab=readme-ov-file#strategies-data
+SUPPORTED_STAKEDAO_STRATEGIES: Final = {'curve', 'balancer', 'pendle', 'pancakeswap', 'angle'}
+
+
+def query_stakedao_gauges() -> None:
+    """Query StakeDAO lockers and strategies gauges from the API and cache them."""
```
nicholasyoder, line 24:
```suggestion
    """Query StakeDAO guages for lockers and strategies from the API and cache them."""
```
I think this is better. When I read the original I thought it meant that it was getting `lockers` and `strategy guages`.

rotkehlchen/chain/evm/decoding/stakedao/utils.py
```diff
@@ -0,0 +1,60 @@
...
+            for item in response_data['deployed']:
+                try:
+                    all_gauges[ChainID.deserialize(item['chainId'])].add(deserialize_evm_address(item['sdGauge']['address']))
+                except DeserializationError as e:
+                    log.warning(f'Skipping stakedao strategy entry {item} due to {e!s}')
+                    continue
+
+    except (RemoteError, UnableToDecryptRemoteData, TypeError, KeyError) as e:
+        msg = f'missing key {e!s}' if isinstance(e, KeyError) else f'{e!s}'
+        log.error(f'Failed to retrieve StakeDAO strategies gauges from API due to {msg}')
```
nicholasyoder, line 52: This is the same list of exceptions as above. Maybe it should all be one try block? Would need to change the error message to also say `gauges for lockers and strategies`.

rotkehlchen/chain/evm/decoding/stakedao/decoder.py
prettyirrelevant: >  also is there any way to check the event asset?

no, the `Deposit` event only emits the amount deposited.

> So it looks like at least in your test the deposited amount and the received amount are the same. Can't you just check event.amount against what you call received_amount here

i thought of doing but i was playing it safe since tokens can have different decimals and gauges use `decimals=18`

rotkehlchen/chain/evm/decoding/stakedao/decoder.py
prettyirrelevant: > Would there be any way to do this without making these remote calls? It does look like maybe we do contract calls like this in a couple other decoders but its probably something that would be best if we could avoid.

at all. the event does not give much information except the amount received.

> If this is necessary at least move it to closer to the end of the function. As far as I can tell received_token and received_amount that are being retrieved here aren't used till the action item at the end. There's no point in making these remote calls at the beginning when it is possible for the function to return before using the data.

i use the vault address when matching the tx log. so i moved it there.

rotkehlchen/chain/evm/decoding/stakedao/decoder.py
prettyirrelevant: yeah, it can. i added a check to ensure it is coming from the gauge address and going to the user's address.

rotkehlchen/chain/evm/decoding/stakedao/decoder.py
```diff
@@ -0,0 +1,286 @@
...
+        amount = int.from_bytes(context.tx_log.data[32:64])
+        period = Timestamp(int.from_bytes(context.tx_log.data[96:128]))
+        return self._decode_claim(context=context, reward_token_address=reward_token_address, amount=amount, period=period)  # noqa: E501
+
+    def _decode_claim_with_bribe(self, context: DecoderContext) -> DecodingOutput:
+        if context.tx_log.topics[0] != CLAIMED_WITH_BRIBE:
+            return DEFAULT_DECODING_OUTPUT
+
+        # we are not checking user address in the logs as user is not always
+        # the recipient according to the contract
```
nicholasyoder, line 134: This comment is exactly the same as above in `_decode_claim_with_bounty` - should probably just put it in `_decode_claim` so the same comment isn't repeated here.

rotkehlchen/chain/evm/decoding/stakedao/decoder.py
nicholasyoder: ok. Add a comment saying something about that you have to check raw amounts here since the asset can't be found to use for finding the normalized deposited amount.

rotkehlchen/chain/evm/decoding/stakedao/decoder.py
nicholasyoder: ah, ok I didn't notice that the vault_address was being used there. Maybe just retrieve the vault address by itself, and then move the call that gets the token to after the tx_log loop? That way if the loop's `else` is hit and it returns it will have only done the one remote call that it needed in the loop.

rotkehlchen/chain/evm/decoding/stakedao/utils.py
```diff
@@ -0,0 +1,55 @@
...
+log = RotkehlchenLogsAdapter(logger)
+
+STAKEDAO_API: Final = 'https://api.stakedao.org/api'
+# This is gotten from:
+# https://github.com/stake-dao/api?tab=readme-ov-file#strategies-data
+SUPPORTED_STAKEDAO_STRATEGIES: Final = {'curve', 'balancer', 'pendle', 'pancakeswap', 'angle'}
+
+
+def query_stakedao_gauges() -> None:
+    """Query StakeDAO guages for lockers and strategies from the API and cache them."""
```
nicholasyoder, line 24:
```suggestion
    """Query StakeDAO gauges for lockers and strategies from the API and cache them."""
```
sorry I think I misspelled this in my suggestion. the `a` comes before the `u` in gauges.

rotkehlchen/tests/unit/decoders/test_stakedao.py
```diff
@@ -157,3 +178,129 @@ def test_claim_multiple(ethereum_inquirer, ethereum_accounts):
...
+        location_label=user_address,
+        notes=f'Receive {deposit_amount} sdeUSDUSDC-gauge after depositing in StakeDAO',
+        counterparty=CPT_STAKEDAO,
+        address=ZERO_ADDRESS,
+    )]
+
+
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('ethereum_accounts', [['0x5bAaC7ccda079839C9524b90dF81720834FC039f']])
+def test_withdraw(ethereum_inquirer, ethereum_accounts, stakedao_gauges):
```
nicholasyoder, line 235: Shouldn't one of these tests be for a chain other than ethereum? It looks like all the tests for stakedao are for ethereum currently.

prettyirrelevant: so i just realised that they deploy a different version of the contract on other chains. will need to think a bit on how to approach that, so will come in a separate PR together with balances.

rotkehlchen/chain/evm/decoding/stakedao/decoder.py
```diff
@@ -0,0 +1,292 @@
...
+    def reload_data(self) -> Mapping['ChecksumEvmAddress', tuple[Any, ...]] | None:
+        """Check that cache is up to date and refresh cache from db.
+        Returns a fresh addresses to decoders mapping.
+        """
+        if should_update_protocol_cache(
+            cache_key=CacheType.STAKEDAO_GAUGES,
+            args=(str(self.evm_inquirer.chain_id.value),),
+        ) is True:
+            query_stakedao_gauges()
+        if len(self.gauges) != 0:
```
LefterisJP, line 80:
```suggestion

        if len(self.gauges) != 0:
```

rotkehlchen/chain/evm/decoding/stakedao/decoder.py
```diff
@@ -0,0 +1,292 @@
...
+        normalized_amount = token_normalized_value(amount, claimed_token)
+        for event in context.decoded_events:
+            if event.event_type == HistoryEventType.RECEIVE and event.event_subtype == HistoryEventSubType.NONE and event.asset == claimed_token and event.amount == normalized_amount:  # noqa: E501
+                event.event_subtype = HistoryEventSubType.REWARD
+                event.counterparty = CPT_STAKEDAO
+                event.notes = f'Claim {normalized_amount} {claimed_token.symbol} from StakeDAO veCRV bribes for the period starting at {timestamp_to_date(period, formatstr="%d/%m/%Y %H:%M:%S")}'  # noqa: E501
+                event.product = EvmProduct.BRIBE
+                break
+        else:  # not found
+            log.error(f'Stakedao bribe transfer was not found for {context.transaction.tx_hash.hex()}')  # noqa: E501
```
LefterisJP, line 118: This way also the chain will appear

```suggestion
            log.error(f'Stakedao bribe transfer was not found for {context.transaction}')  # noqa: E501
```

rotkehlchen/chain/evm/decoding/stakedao/utils.py
```diff
@@ -0,0 +1,55 @@
...
+# https://github.com/stake-dao/api?tab=readme-ov-file#strategies-data
+SUPPORTED_STAKEDAO_STRATEGIES: Final = {'curve', 'balancer', 'pendle', 'pancakeswap', 'angle'}
+
+
+def query_stakedao_gauges() -> None:
+    """Query StakeDAO gauges for lockers and strategies from the API and cache them."""
+    all_gauges: dict[ChainID, set] = defaultdict(set)
+
+    try:  # Query lockers
+        response_data = request_get_dict(url=STAKEDAO_API + '/lockers')
```
LefterisJP, line 28: use fstring

LefterisJP (approved):
LGTM

PR #9684, opened 2025-03-26
Remove vendored OpenSSL usage from rusqlite

strmfos:
**Summary**
- Remove vendored OpenSSL usage from `rusqlite` by dropping the `bundled-sqlcipher-vendored-openssl` feature.
- Rely on the system-installed OpenSSL (e.g., via Homebrew) instead.

**Why This Change?**
- Homebrew no longer provides `openssl@1.1`, causing build failures when attempting to compile vendored OpenSSL.
- This switch ensures compatibility with modern macOS systems and keeps the build environment simpler.

**Testing**
- Verified locally with `cargo clippy --all -- -D warnings` and any existing tests to ensure no regressions.

### DIFF
--- a/colibri/Cargo.lock
+++ b/colibri/Cargo.lock
@@ -2364,7 +2364,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2e99fb7a497b1e3339bc746195567ed8d3e24945ecd636e3619d20b9de9e9149"
 dependencies = [
  "cc",
- "openssl-sys",
  "pkg-config",
  "vcpkg",
 ]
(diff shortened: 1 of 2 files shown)


kelsos:
Hey @strmfos I just tried the build on macOS Sequoia 15.3.2 (arm64) and I don't see any build errors. 

Our CI also does not seem to have any build failures on on macOS 13 running on intel-mac and macOS 14 running on (arm64)

Also I think a big issue with this PR is that it's only macOS-centric, and ignores Windows and Linux as supported platforms. This is a change that would need to be checked on all 3 operating systems. The fact is that this change if accepted would break Windows builds.

For that we are not going to accept this PR since there is also no sufficient evidence provided that there is a problem that it fixes.

However, if it is an actual problem you encountered please provide more information so that we can assist you.

Finally,please keep in mind that we have a [contribution policy](https://github.com/rotki/rotki/blob/develop/CONTRIBUTING.md) that does not accept low quality AI generated PRs.

PR #9692, opened 2025-03-27
feat: send token

lukicenturi:
Closes #(issue_number)

## Checklist


frontend/app/.env
```diff
@@ -2,4 +2,5 @@ VITE_BACKEND_URL=http://127.0.0.1:4242
 VITE_COLIBRI_URL=http://127.0.0.1:4343
 VITE_I18N_LOCALE=en
 VITE_I18N_FALLBACK_LOCALE=en
+VITE_WALLET_CONNECT_PROJECT_ID=eca1bd3842e084bfc8ef309b6eb584db
```
lukicenturi, line 5: For the project_id, wallet connect do some endpoint call, that we can see from the network tab, and the project id is exposed anyway, so I think it should be fine (?)

Otherwise, we can let user put their own projectId, but it needs more work

frontend/app/electron/main/http.ts
```diff
@@ -156,3 +160,72 @@ export function stopHttp() {
   if (server?.listening)
     server.close();
 }
+
+export function startWalletConnectBridgeServer(port: number, logger: LogService) {
```
lukicenturi, line 164: I will need your help to check this. This is to open the bridge.
it doesn't open a single html file, but it open a route on the app itself.

frontend/app/src/router/index.ts
```diff
@@ -50,7 +50,7 @@ router.beforeEach((to, from, next) => {
 
     next();
   }
-  else if (to.path.startsWith('/user')) {
+  else if (to.path.startsWith('/user') || to.path === '/wallet-bridge') {
```
lukicenturi, line 53: this one to make if the `/wallet-bridge` route opened in the browser, it bypass the auth

frontend/app/vite.config.ts
```diff
@@ -48,6 +48,12 @@ if (!hmrEnabled)
 
 const enableChecker = !(process.env.CI || isTest || process.env.VITEST);
 
+const builtInModulesToNotExternalized = [
+  'buffer',
+  'crypto',
+  'events',
+];
```
lukicenturi, line 55: Without this, I got this error after build:
```
Uncaught TypeError: Failed to resolve module specifier "buffer". Relative references must start with either "/", "./", or "../".
```

To be honest I'm not 100% sure why it happens.
But 1 things for sure, these 3 built in modules are imported in the `reown` package

frontend/app/electron/main/create-protocol.ts
```diff
@@ -5,6 +5,29 @@ import { type Protocol, protocol } from 'electron';
...
+  else if (extension === '.json')
+    return 'application/json';
+  else if (extension === '.wasm')
+    return 'application/wasm';
+
+  return 'application/octet-stream';
+}
+
 export function createProtocol(scheme: string, customProtocol?: Protocol) {
   (customProtocol || protocol).registerBufferProtocol(scheme, (request, respond) => {
```
kelsos, line 32: did you get some warning about deprecation around here?

frontend/app/src/components/inputs/AmountInput.vue
```diff
@@ -26,12 +27,14 @@ const modelValue = defineModel<string>({ required: true });
 const props = withDefaults(defineProps<AmountInputProps>(), {
   hideDetails: false,
   integer: false,
+  rawInput: false,
```
kelsos, line 30: when and why do we use this raw Input instead of the RuiTextField one?

frontend/app/vite.config.ts
kelsos: Weird, for nuxt in the website we didn't have to do anything similar right?

frontend/app/src/store/trade/wallet.ts
```diff
@@ -0,0 +1,334 @@
...
+
+  const setupAppKitListener = (): void => {
+    assert(appKit);
+
+    appKit.subscribeAccount((account) => {
+      set(connected, account.isConnected);
+      set(connectedAddress, account.isConnected ? account.address : undefined);
+
+      if (account.isConnected) {
+        const provider: any = appKit!.getProvider(EIP155);
```
kelsos, line 111: maybe can we add also an assertion in the subscribe instead of the !

frontend/app/src/store/trade/wallet.ts
```diff
@@ -0,0 +1,334 @@
...
+  bsc,
+  gnosis,
+  polygon,
+  scroll,
+] as const;
+
+function buildAppKit(): AppKit {
+  const projectId = import.meta.env.VITE_WALLET_CONNECT_PROJECT_ID as string;
+
+  const wagmiAdapter = new WagmiAdapter({
```
kelsos, line 43: we don't use this in the website right? what is the difference of this and the ethers adapter? do we need both?

frontend/app/src/store/trade/wallet.ts
```diff
@@ -0,0 +1,334 @@
...
+  const getBrowserProvider = (): BrowserProvider => {
+    const { walletProvider } = useAppKitProvider(EIP155);
+    return new BrowserProvider(walletProvider as any);
+  };
+
+  const getGasFeeForChain = async (): Promise<GasFeeEstimation> => {
+    try {
+      const provider = getBrowserProvider();
+
+      const DEFAULT_GAS_LIMIT = 21000n; // for native transfers
```
kelsos, line 169: small nitpick but maybe we can pull this to top level?

frontend/app/src/store/trade/wallet.ts
```diff
@@ -0,0 +1,334 @@
...
+      else {
+        throw new Error('Failed to load the payload from backend');
+      }
+
+      return tx;
+    }
+    catch (error) {
+      set(preparing, false);
+      set(waitingForWalletConfirmation, false);
+      const txError = error as TransactionError;
```
kelsos, line 305: instead of typecasting can we do an instance check instead?

frontend/app/src/components/wallet-bridge/Pairing.vue
```diff
@@ -0,0 +1,345 @@
...
+const projectId = import.meta.env.VITE_WALLET_CONNECT_PROJECT_ID as string;
+
+const core = new Core({
+  projectId,
+});
+
+const pairUri = ref('');
+const logs = ref<LogEntry[]>([]);
+const isConnecting = ref(false);
+const walletKit = ref<IWalletKit | null>(null);
```
kelsos, line 38: is the default null? if we just manage the state let's keep it `undefined`

frontend/app/src/components/wallet-bridge/Pairing.vue
```diff
@@ -0,0 +1,345 @@
...
+  if (!pairUriVal) {
+    addLog('Please enter a valid Pair URI', 'error');
+    return;
+  }
+
+  set(isConnecting, true);
+
+  try {
+    const kit = get(walletKit);
+    if (!kit) {
```
kelsos, line 76: is there any specific reason we check twice here? also we could do the `isDefined()` and only get if it is initialized right?

frontend/app/src/components/wallet-bridge/Pairing.vue
```diff
@@ -0,0 +1,345 @@
...
+  const kit = get(walletKit);
+  if (!kit)
+    return [];
+
+  const sessionsMap = kit.getActiveSessions() || {};
+  return Object.values(sessionsMap);
+}
+
+async function chainChanged(topic: string, chainId: string) {
+  const kit = get(walletKit);
```
kelsos, line 108: wouldn't it make sense to bail early here if the kit is not defined? and only proceed if it is?

frontend/app/src/components/wallet-bridge/Pairing.vue
```diff
@@ -0,0 +1,345 @@
...
+    event: {
+      data: Number(chainId),
+      name: 'chainChanged',
+    },
+    topic,
+  });
+}
+
+async function accountsChanged(topic: string, chainId: string, address: string) {
+  const kit = get(walletKit);
```
kelsos, line 123: pretty much the same as above here, no need to get the chain id if the kit is not defined

frontend/app/src/components/wallet-bridge/Pairing.vue
```diff
@@ -0,0 +1,345 @@
...
+
+async function onSessionRequest(event: WalletKitTypes.SessionRequest) {
+  const kit = get(walletKit);
+  assert(kit);
+
+  const { id, params, topic } = event;
+  try {
+    const request = params.request.params[0];
+    const result = await triggerTransaction(request);
+    const chainName = get(getChainName(getChainFromChainId(Number(result.chainId))));
```
kelsos, line 194: to many gets here, don't we have any single function to get the chain name from id?

frontend/app/src/components/trade/send/TradeSendCard.vue
```diff
@@ -0,0 +1,432 @@
...
+const amount = ref<string>('');
+const asset = ref<string>('');
+const assetChain = ref<string>(Blockchain.ETH);
+const toAddress = ref<string>('');
+
+const max = ref<string>('0');
+const estimatingGas = ref<boolean>(false);
+const showNeverInteractedWarning = ref<boolean>(false);
+const errorMessage = ref<string>('');
+const currentGasEstimationController = ref<AbortController | null>(null);
```
kelsos, line 38: again here if we are the ones tracking it let's keep it simple and get rid of the null

frontend/app/src/components/inputs/AmountInput.vue
lukicenturi: that's for the amount.
So I want to use the `AmountInput` but I don't want to use the styles for the RuiTextField

frontend/app/src/components/trade/send/TradeAssetSelector.vue
```diff
@@ -0,0 +1,172 @@
...
+import { Blockchain } from '@rotki/common';
+
+const asset = defineModel<string>({ required: true });
+const chain = defineModel<string>('chain', { required: true });
+
+const props = defineProps<{
+  address?: string;
+}>();
+
+const emit = defineEmits(['set-max']);
```
kelsos, line 16: shouldn't this use the defineEmits<{ 'set-max': [] }>() format?

frontend/app/src/components/trade/send/TradeAmountInput.vue
```diff
@@ -0,0 +1,152 @@
...
+const { getAssetDetail } = useTradeAsset(address);
+const assetDetail = getAssetDetail(asset, chain);
+
+// Computed properties for conversion
+const amountModel = computed({
+  get: () => get(model),
+  set: (value: string) => {
+    set(model, value);
+    const price = get(assetDetail)?.price;
+    if (price) {
```
kelsos, line 41: was there any issue with using watchers for this an the other?

lukicenturi: You're right, can be simplified

frontend/app/src/components/trade/send/TradeRecipientAddress.vue
```diff
@@ -0,0 +1,309 @@
...
+                {{ t('trade.to_address.label') }}
+              </span>
+              <span
+                v-else
+                class="text-sm text-rui-error font-medium block"
+              >
+                {{ t('trade.to_address.error') }}
+              </span>
+            </Transition>
+            <input
```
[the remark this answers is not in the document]
lukicenturi: because the RuiTextField, has many element and excess styling I don't need

frontend/app/vite.config.ts
lukicenturi: probably because it's nuxt, there is likely a difference in the building process.

frontend/app/src/store/trade/wallet.ts
lukicenturi: Okay now when I tried again, I think it doesn't work with SPA, but works with nuxt.
I got error `process is not defined` smth.

lukicenturi:
> btw how much effort would it be to move the new code (the app related parts at least not the electron ones) under the `module` directory?
> 
> Since this is a new feature it would make sense to use the new structure instead of following the old one wouldn't it?

Okay now it should be moved to the modules too

kelsos (approved):
Ok, so the ping is still in the packaged version, I will have to check why. But otherwise I think we can merge and keep improving on top.

Some small things I would add as follow-ups here @lukicenturi 

- We should show also some kind of loading indicator to the asset list too, the working is there on top but it took me a bit to figure it out.
- We probably need to filter-out unsupported chains for the address book. I got an error while it was trying to resolve the ens for a bitcoin account.

PR #9699, opened 2025-03-28
Allow exchange history events partial range query

nicholasyoder:
Related: #6097 and [this note](https://github.com/orgs/rotki/projects/11?pane=issue&itemId=102361655)

Makes it possible for an exchange to only return events from part of the range queried if there are errors part way through querying the range. Currently only utilized by Kraken since it was doing something similar before by using custom logic that didn't conform to the exchange interface. Now this is supported by the exchange interface itself so it can be used in any exchange where it might be useful.


LefterisJP (approved):
LGTM

PR #9702, opened 2025-03-28
Add decoding of crvusd peg keeper updates

nicholasyoder:
Related: #9570

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/curve/crvusd/constants.py
+++ b/rotkehlchen/chain/ethereum/modules/curve/crvusd/constants.py
@@ -7,3 +7,11 @@ from rotkehlchen.chain.evm.types import string_to_evm_address
 CRVUSD_MINTER: Final = string_to_evm_address('0xC9332fdCB1C491Dcc683bAe86Fe3cb70360738BC')
 CRVUSD_MINTER_ABI: Final[ABI] = [{'stateMutability': 'view', 'type': 'function', 'name': 'controllers', 'inputs': [{'name': 'arg0', 'type': 'uint256'}], 'outputs': [{'name': '', 'type': 'address'}]}, {'stateMutability': 'view', 'type': 'function', 'name': 'n_collaterals', 'inputs': [], 'outputs': [{'name': '', 'type': 'uint256'}]}]  # noqa: E501
 CURVE_CRVUSD_CONTROLLER_ABI: Final[ABI] = [{'stateMutability': 'view', 'type': 'function', 'name': 'amm', 'inputs': [], 'outputs': [{'name': '', 'type': 'address'}]}, {'stateMutability': 'view', 'type': 'function', 'name': 'collateral_token', 'inputs': [], 'outputs': [{'name': '', 'type': 'address'}]}]  # noqa: E501
+CRVUSD_PEG_KEEPERS_AND_POOLS: Final = {
+    string_to_evm_address('0x9201da0D97CaAAff53f01B2fB56767C7072dE340'): string_to_evm_address('0x4DEcE678ceceb27446b35C672dC7d61F30bAD69E'),  # USDC/crvUSD  # noqa: E501
+    string_to_evm_address('0xFb726F57d251aB5C731E5C64eD4F5F94351eF9F3'): string_to_evm_address('0x390f3595bCa2Df7d23783dFd126427CCeb997BF4'),  # USDT/crvUSD  # noqa: E501
+    string_to_evm_address('0x3fA20eAa107DE08B38a8734063D605d5842fe09C'): string_to_evm_address('0x625E92624Bc2D88619ACCc1788365A69767f6200'),  # PYUSD/crvUSD  # noqa: E501
+    string_to_evm_address('0x503E1Bf274e7a6c64152395aE8eB57ec391F91F8'): string_to_evm_address('0x30cE6E5A75586F0E83bCAc77C9135E980e6bc7A8'),  # USDM/crvUSD  # noqa: E501
+}
+PEG_KEEPER_PROVIDE_TOPIC: Final = b'\x8dh[\xd3\xf4]\x86\x1cu\x9e\xd7\xa4n\xa3\xd3\x0e\xb5\xccl\xe9\xfe\x06\xc5&\x93\x1f\x94\xc9c\xbc\xa7\xd2'  # noqa: E501
+PEG_KEEPER_WITHDRAW_TOPIC: Final = b'[kC\x1dDv\xa2\x11\xbb}A\xc2\r\x1a\xab\x9a\xe22\x1d\xee\xe0\xd2\x0b\xe3\xd9\xfc\x9b\x10\x93\xfan='  # noqa: E501
--- a/rotkehlchen/chain/ethereum/modules/curve/crvusd/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/curve/crvusd/decoder.py
@@ -3,7 +3,12 @@ from collections.abc import Mapping
 from typing import TYPE_CHECKING, Any
 
 from rotkehlchen.assets.asset import EvmToken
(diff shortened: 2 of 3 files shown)


rotkehlchen/chain/ethereum/modules/curve/crvusd/constants.py
```diff
@@ -7,3 +7,10 @@
...
 CURVE_CRVUSD_CONTROLLER_ABI: Final[ABI] = [{'stateMutability': 'view', 'type': 'function', 'name': 'amm', 'inputs': [], 'outputs': [{'name': '', 'type': 'address'}]}, {'stateMutability': 'view', 'type': 'function', 'name': 'collateral_token', 'inputs': [], 'outputs': [{'name': '', 'type': 'address'}]}]  # noqa: E501
+CRVUSD_PEG_KEEPERS: Final = {
+    string_to_evm_address('0x9201da0D97CaAAff53f01B2fB56767C7072dE340'),  # USDC/crvUSD
+    string_to_evm_address('0xFb726F57d251aB5C731E5C64eD4F5F94351eF9F3'),  # USDT/crvUSD
+    string_to_evm_address('0x3fA20eAa107DE08B38a8734063D605d5842fe09C'),  # PYUSD/crvUSD
+    string_to_evm_address('0x503E1Bf274e7a6c64152395aE8eB57ec391F91F8'),  # USDM/crvUSD
+}
+PEG_KEEPER_WITHDRAW_TOPIC: Final = b'[kC\x1dDv\xa2\x11\xbb}A\xc2\r\x1a\xab\x9a\xe22\x1d\xee\xe0\xd2\x0b\xe3\xd9\xfc\x9b\x10\x93\xfan='  # noqa: E501
```
prettyirrelevant, line 16: reading through the curve docs, it seems this can either be a deposit or withdraw.

here's an example tx that's a deposit: https://etherscan.io/tx/0xb557deee6fcf5fcbb03b539f81413a32e5ac40ccbc4915bcab291136b2049461#eventlog

rotkehlchen/chain/ethereum/modules/curve/crvusd/decoder.py
```diff
@@ -170,3 +175,61 @@ def maybe_decode_leveraged_borrow(self, context: DecoderContext) -> DecodingOutp
...
 
         return DEFAULT_DECODING_OUTPUT
+
+    def _decode_peg_keeper_update(self, context: DecoderContext) -> DecodingOutput:
+        if context.tx_log.topics[0] != PEG_KEEPER_WITHDRAW_TOPIC:
+            return DEFAULT_DECODING_OUTPUT
+
+        # The amount in the peg keeper withdraw event is the total amount of crvUSD being traded.
+        # The user is only rewarded with a small portion of that and is rewarded in a different
+        # token. Unfortunately there isn't any information about the actual reward token and amount
```
prettyirrelevant, line 185: from the curve docs, the reward token is the `POOL` variable so you can use that. also, don't forget to handle rewards for deposit & withdrawals.
CleanShot 2025-03-28 at 20 44 42@2x


they also provide a way to calculate the reward but i don't think doing that here is necessary as the transfer happens after the `Provide` or `Withdraw` event.
CleanShot 2025-03-28 at 20 50 20@2x

nicholasyoder: I've hardcoded the pool addresses along with the peg keeper addresses, and removed the part where it retrieved the pool token address from the tx_logs.

>they also provide a way to calculate the reward but i don't think doing that here is necessary

yeah, I think this would be more complicated than its worth. Also its best to avoid a remote call. Keeping this as it was so it finds the transfer log event and gets the amount there.

rotkehlchen/chain/ethereum/modules/curve/crvusd/decoder.py
```diff
@@ -170,3 +176,51 @@ def maybe_decode_leveraged_borrow(self, context: DecoderContext) -> DecodingOutp
...
+                to_notes=f'Receive {reward_amount} {reward_token.symbol} after updating a Curve peg keeper',  # noqa: E501
+            ),
+        ])
+
+    # -- DecoderInterface methods
+
+    def addresses_to_decoders(self) -> dict[ChecksumEvmAddress, tuple[Any, ...]]:
+        return (
+            super().addresses_to_decoders() |
+            dict.fromkeys(CRVUSD_PEG_KEEPERS_AND_POOLS.keys(), (self._decode_peg_keeper_update,))
```
prettyirrelevant, line 225:
```suggestion
            dict.fromkeys(CRVUSD_PEG_KEEPERS_AND_POOLS, (self._decode_peg_keeper_update,))
```

rotkehlchen/chain/ethereum/modules/curve/crvusd/decoder.py
```diff
@@ -170,3 +176,51 @@ def maybe_decode_leveraged_borrow(self, context: DecoderContext) -> DecodingOutp
...
         return DEFAULT_DECODING_OUTPUT
+
+    def _decode_peg_keeper_update(self, context: DecoderContext) -> DecodingOutput:
+        if context.tx_log.topics[0] not in {PEG_KEEPER_PROVIDE_TOPIC, PEG_KEEPER_WITHDRAW_TOPIC}:
+            return DEFAULT_DECODING_OUTPUT
+
+        # Unfortunately the peg keeper provide/withdraw event doesn't give much information,
+        # so we have to find the transfer of the pool token from the peg keeper to the
+        # tracked address and retrieve the raw reward amount from there.
+        pool_token_address = CRVUSD_PEG_KEEPERS_AND_POOLS[context.tx_log.address]
```
LefterisJP, line 187: This can never raise a keyerror right? Can you write why here.

rotkehlchen/chain/ethereum/modules/curve/crvusd/decoder.py
```diff
@@ -170,3 +176,51 @@ def maybe_decode_leveraged_borrow(self, context: DecoderContext) -> DecodingOutp
...
+                reward_raw_amount = int.from_bytes(tx_log.data[:32])
+                break
+        else:
+            log.error(f'Failed to find reward amount for curve peg keeper update transaction {context.transaction!s}')  # noqa: E501
+            return DEFAULT_DECODING_OUTPUT
+
+        reward_token = self.base.get_or_create_evm_token(address=pool_token_address)
+        reward_amount = token_normalized_value(
+            token_amount=reward_raw_amount,
+            token=reward_token,
```
LefterisJP, line 204:
```suggestion
            token=(reward_token := self.base.get_or_create_evm_token(address=pool_token_address)),
```

rotkehlchen/chain/ethereum/modules/curve/crvusd/decoder.py
```diff
@@ -170,3 +176,51 @@ def maybe_decode_leveraged_borrow(self, context: DecoderContext) -> DecodingOutp
...
+                bytes_to_address(tx_log.topics[1]) == context.tx_log.address and
+                self.base.is_tracked(bytes_to_address(tx_log.topics[2]))
+            ):
+                reward_raw_amount = int.from_bytes(tx_log.data[:32])
+                break
+        else:
+            log.error(f'Failed to find reward amount for curve peg keeper update transaction {context.transaction!s}')  # noqa: E501
+            return DEFAULT_DECODING_OUTPUT
+
+        reward_token = self.base.get_or_create_evm_token(address=pool_token_address)
```
LefterisJP, line 201: also remove this if you do the below

rotkehlchen/chain/ethereum/modules/curve/crvusd/decoder.py
nicholasyoder: Actually I've changed this to use `:=` for both `reward_token` and `reward_amount` in the action item below.

LefterisJP (approved):
LGTM

PR #9717, opened 2025-04-01
Add support for crvusd controller balances

### LINKED ISSUE #9570: Support Curve crvUSD minting and peg keeping

nicholasyoder:
Closes #9570

### DIFF
--- a/rotkehlchen/chain/evm/decoding/curve/lend/balances.py
+++ b/rotkehlchen/chain/evm/decoding/curve/lend/balances.py
@@ -1,6 +1,7 @@
 import logging
+from abc import ABC, abstractmethod
 from collections import defaultdict
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
 
 from rotkehlchen.accounting.structures.balance import Balance, BalanceSheet
 from rotkehlchen.assets.utils import get_or_create_evm_token
@@ -13,6 +14,7 @@ from rotkehlchen.chain.evm.decoding.curve.lend.constants import CURVE_VAULT_CONT
 from rotkehlchen.constants import ZERO
 from rotkehlchen.errors.misc import NotERC20Conformant, NotERC721Conformant, RemoteError
 from rotkehlchen.errors.serialization import DeserializationError
+from rotkehlchen.history.events.structures.evm_event import EvmProduct
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.inquirer import Inquirer
 from rotkehlchen.logging import RotkehlchenLogsAdapter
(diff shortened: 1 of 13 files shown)


rotkehlchen/chain/evm/decoding/curve/lend/common.py
```diff
@@ -53,6 +53,7 @@ def __init__(
             evm_inquirer: 'EvmNodeInquirer',  # pylint: disable=unused-argument
             base_tools: 'BaseDecoderTools',
             msg_aggregator: 'MessagesAggregator',
+            controller_extra_data_key: Literal['vault_controller', 'crvusd_controller'],
```
LefterisJP, line 56: is this abstraction needed? Can't you just use a common name for both and get rid of the extra logic and variable?

rotkehlchen/chain/evm/decoding/curve/lend/balances.py
```diff
@@ -29,12 +30,15 @@
...
 
 
-class CurveLendBalances(ProtocolWithBalance):
+class CurveControllerCommonBalances(ProtocolWithBalance, ABC):
+
     def __init__(
             self,
             evm_inquirer: 'EvmNodeInquirer',
             tx_decoder: 'EVMTransactionDecoder',
+            controller_extra_data_key: Literal['vault_controller', 'crvusd_controller'],
```
LefterisJP, line 39: same here. As in the other place. Why add this extra logic just for the name of a dict key? Just find a common name to use that describes both. Perhaps `controller_address`?

rotkehlchen/chain/evm/decoding/curve/lend/common.py
nicholasyoder: `controller_extra_data_key` is used to differentiate between the two types of controllers in `_get_controllers_with_balances` so that `CurveLendBalances` and `CurveCrvusdBalances` only process the correct type of controller.

I've added more explanation about this to the relevant docstrings to hopefully make it clearer.

nicholasyoder: Changed this to use EvmProduct as we discussed in discord.

LefterisJP (approved):
LGTM

PR #9720, opened 2025-04-02
feat: allow theme switch for free user

lukicenturi:
Closes #(issue_number)

## Checklist


kelsos (commented):
Code looks ok, I need to run a check and we can merge

kelsos:
So look good I created a new account, and I can properly use dark mode

There is only one extra thing that needs to be done before we merge though, not sure how much work it is but we should keep the theme going regardless if you are logged in or not.

So login screen should still remember the theme selection.

kelsos:
@lukicenturi ok one last thing I promise and then we can merge.

Can we make it so that the last selected theme is persisted in localstorage too?
So if I had dark theme and restart the application I would still get the dark theme at the login page.

The logout works fine but on restart it resets to white, since we have to load it from the preferences.

kelsos (approved):
lgtm

PR #9721, opened 2025-04-02
Use SwapEvents in bitpanda

nicholasyoder:
Related: #6097 

* Converts bitpanda to use swap events.
* Removes the unused `query_online_trade_history` function from bitmex 
* Fixes a nitpick from #9716


LefterisJP (approved):
LGTM

PR #9724, opened 2025-04-03
Use SwapEvents in bitstamp

nicholasyoder:
Related: #6097 

* Converts bitstamp to use swap events.
* Adds support for swap events to have `reference` in their extra_data like we do for asset movements, since bitstamp relies on this to retrieve the trade ids.

### DIFF
--- a/rotkehlchen/api/v1/schemas.py
+++ b/rotkehlchen/api/v1/schemas.py
@@ -99,7 +99,7 @@ from rotkehlchen.history.events.structures.eth2 import (
     EthWithdrawalEvent,
 )
 from rotkehlchen.history.events.structures.evm_event import EvmEvent, EvmProduct
-from rotkehlchen.history.events.structures.swap import create_swap_events
+from rotkehlchen.history.events.structures.swap import SwapEventExtraData, create_swap_events
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.history.types import HistoricalPriceOracle
 from rotkehlchen.icons import ALLOWED_ICON_EXTENSIONS
@@ -1081,6 +1081,10 @@ class CreateHistoryEventSchema(Schema):
                     field_name='fee_notes',
                 )
 
+            extra_data: SwapEventExtraData = {}
+            if (unique_id := data['unique_id']) is not None:
+                extra_data['reference'] = unique_id
+
             context_schema = CreateHistoryEventSchema.history_event_context.get()['schema']
             events = create_swap_events(
                 timestamp=data['timestamp'],
@@ -1106,6 +1110,7 @@ class CreateHistoryEventSchema(Schema):
                     data=data,
                     subtype=HistoryEventSubType.FEE,
                 ),
+                extra_data=extra_data,
             )
             return {'events': events}
 
--- a/rotkehlchen/exchanges/bitstamp.py
+++ b/rotkehlchen/exchanges/bitstamp.py
@@ -21,12 +21,18 @@ from rotkehlchen.db.history_events import DBHistoryEvents
 from rotkehlchen.errors.asset import UnknownAsset, UnsupportedAsset
 from rotkehlchen.errors.misc import RemoteError
 from rotkehlchen.errors.serialization import DeserializationError
-from rotkehlchen.exchanges.data_structures import MarginPosition, Trade, TradeType
+from rotkehlchen.exchanges.data_structures import MarginPosition
 from rotkehlchen.exchanges.exchange import ExchangeInterface, ExchangeQueryBalances
 from rotkehlchen.history.deserialization import deserialize_price
 from rotkehlchen.history.events.structures.asset_movement import (
     AssetMovement,
     create_asset_movement_with_fee,
+)
+from rotkehlchen.history.events.structures.base import HistoryBaseEntryType
+from rotkehlchen.history.events.structures.swap import (
+    SwapEvent,
+    create_swap_events,
+    get_swap_spend_receive,
 )
 from rotkehlchen.history.events.structures.types import HistoryEventType
 from rotkehlchen.inquirer import Inquirer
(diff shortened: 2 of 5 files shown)


rotkehlchen/history/events/structures/swap.py
```diff
@@ -271,20 +285,56 @@ def create_swap_events(
...
+    ...
+
+
+def get_swap_spend_receive(
         base_asset: Asset,
         quote_asset: Asset,
         amount: 'AssetAmount',
         rate: 'Price',
+        is_buy: bool | None = None,
+        raw_trade_type: str | None = None,
```
prettyirrelevant, line 320: why this? any reason for not sticking to `raw trade type`? i think there's no upside to having both `is_buy` & `raw_trade_type` since we can always differentiate per exchange.

rotkehlchen/exchanges/bitstamp.py
```diff
@@ -774,44 +784,48 @@ def _deserialize_asset_movement_from_crypto_transaction(
...
+            location=self.location,
+            spend_asset=spend_asset,
+            spend_amount=spend_amount,
+            receive_asset=receive_asset,
+            receive_amount=receive_amount,
+            fee_asset=trade_pair_data.quote_asset,
+            fee_amount=deserialize_fee(raw_trade['fee']),
+            location_label=self.name,
+            unique_id=(reference := str(raw_trade['id'])),
+            extra_data={'reference': reference},
```
prettyirrelevant, line 827: since `unique_id` is an existing parameter, there's no need to do this here. it can be done inside the `create_swap_events()` function where you create an extra_data with the `unique_id`

rotkehlchen/api/v1/schemas.py
```diff
@@ -1081,6 +1081,10 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
                     field_name='fee_notes',
                 )
 
+            extra_data: SwapEventExtraData = {}
+            if (unique_id := data['unique_id']) is not None:
+                extra_data['reference'] = unique_id
+
```
prettyirrelevant, line 1087: as mentioned earlier, this can into the `create_swap_events()` function.

rotkehlchen/exchanges/binance.py
```diff
@@ -147,7 +147,7 @@ def trade_from_binance(
         )
 
     spend_asset, spend_amount, receive_asset, receive_amount = get_swap_spend_receive(
-        raw_trade_type='buy' if binance_trade['isBuyer'] else 'sell',
+        is_buy=binance_trade['isBuyer'],
```
prettyirrelevant, line 150: they're pretty much the same thing if i'm not mistaken so there's no reason to have both parameter in the function, `get_swap_spend_receive()`

rotkehlchen/exchanges/bitstamp.py
```diff
@@ -230,11 +236,38 @@ def query_balances(self) -> ExchangeQueryBalances:
...
+            if (
+                query_result is not None and
+                (extra_data := AssetMovement.deserialize_extra_data(
+                    entry=query_result,
+                    extra_data=query_result[0],
+                )) is not None and
+                'reference' in extra_data
+            ):
+                since_id = int(extra_data['reference']) + 1
+                return {'since_id': since_id}
```
prettyirrelevant, line 262:
```suggestion
                (reference := extra_data.get('reference')) is not None
            ):
                return {'since_id': int(reference) + 1}
```

rotkehlchen/exchanges/bitstamp.py
```diff
@@ -660,7 +667,10 @@ def _api_query_paginated(
                 # NB: re-assign dict instead of update, prevent lose call args values
                 call_options = call_options.copy()
                 offset = 0 if has_results else call_options['offset'] + API_MAX_LIMIT
-                since_id = int(results[-1].link) + 1 if has_results else call_options['since_id']  # type: ignore # we always got a link in bitstamp trades
+                since_id = (
+                    int((results[-2].extra_data or results[-3].extra_data)['reference']) + 1  # type: ignore[index]  # The spend event with extra_data will be at either index -2 or -3 depending on if there is a fee event.
+                    if has_results else call_options['since_id']
+                )
```
prettyirrelevant, line 673: this can raise an index error

rotkehlchen/history/events/structures/swap.py
nicholasyoder: With only `raw_trade_type`, when we don't have a raw type from the api and instead need to check some other condition to determine the type, it seems really hacky to do `'buy' if some_condition else 'sell'` - this makes it do the deserialization logic unnecessarily, plus having the `buy` and `sell` strings hardcoded in every exchange where this needs to be done really doesn't seem like good a good thing to do. At the very least those should instead be an enum or something, but I think allowing `get_swap_spend_receive` to accept a bool flag like this is the cleanest solution. Or is there something I'm not considering here?

If it was only one or two places where we had to do this, passing `buy` or `sell` strings would maybe be ok, but we already have three exchanges where we need to do something like this, and likely there will be more.

rotkehlchen/exchanges/bitstamp.py
nicholasyoder: We don't actually want to set 'reference' every time we use `unique_id` to avoid taking up more space in the DB than is needed. See the discussion we had a couple days ago in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1356627435969581189) about this. So we only want to set it in extra_data when we need it for something - in this case its used in `_get_since_id_option`.

rotkehlchen/exchanges/binance.py
nicholasyoder: See my response to the other comment about this. In my opinion its much cleaner to simply pass the bool `isBuyer` value here rather than doing an `if ... else` and passing the hardcoded raw string `buy` and `sell` values.

rotkehlchen/exchanges/bitstamp.py
nicholasyoder: Actually I don't think it should ever be able to encounter an index error here. If you notice, its only doing this if `has_results` is true, which means that `_deserialize_trade` has been successfully executed, which will always return either 2 or 3 events, the first of which will be the spend event which will have extra_data. This is what the comment there is supposed to explain - there will always be an event with extra data at either -2 or -3. I've reworded the comment slightly to maybe make it a bit clearer.

prettyirrelevant: hmmm, i see. thanks for the clarification.

rotkehlchen/exchanges/bitstamp.py
prettyirrelevant: oh i see. thanks.

rotkehlchen/history/events/structures/swap.py
prettyirrelevant: I get where you’re coming from. That said, having multiple ways to represent the same thing in a function can make things harder to reason about.

It’s probably better to stick to just one field — either `is_buy` or `raw_trade_type`. Not every exchange gives you a clean answer, but you can usually infer the direction based on conditions anyway.

> If it was only one or two places where we had to do this, passing `buy` or `sell` strings would maybe be ok, but we already have three exchanges...

Yeah, that’s fair. I don’t have a problem with `is_buy`, but if we go that route, we should probably update the older integrations to use it too — just so we’re consistent.

Not sure an enum adds much here, since we’d still be deriving the same boolean logic for some exchanges.

nicholasyoder: Changed this back to how it was, as discussed in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1357451739111882762)

rotkehlchen/exchanges/bitstamp.py
```diff
@@ -660,7 +666,10 @@ def _api_query_paginated(
                 # NB: re-assign dict instead of update, prevent lose call args values
                 call_options = call_options.copy()
                 offset = 0 if has_results else call_options['offset'] + API_MAX_LIMIT
-                since_id = int(results[-1].link) + 1 if has_results else call_options['since_id']  # type: ignore # we always got a link in bitstamp trades
+                since_id = (  # If has_results is true a spend event with extra_data should always be present at either index -2 or -3 (depending on if there is a fee event).  # noqa: E501
+                    int((results[-2].extra_data or results[-3].extra_data)['reference']) + 1  # type: ignore[index]
```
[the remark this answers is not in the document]
nicholasyoder: Updated this, reworking the logic a bit and adding more comments to hopefully make it clearer what happens here.

PR #9728, opened 2025-04-04
feat: option to persist table sorting

### LINKED ISSUE #8928: Column sorting should be sticky
## Abstract

When clicking a column of a table to change the sorting, this preference is lost as soon as the view is exited and reentered.  

## Motivation

This means any sorting which doesn't suit the user by default has to be changed over and over again.  For example I usually want to see my accounts sorted by name not value.

## Specification

It would be better if the sorting state persisted to disk as soon as it was set, so that it is sticky not just within the session, but as a per-user setting across all settings.

lukicenturi:
Closes #8928

## Checklist


frontend/app/src/modules/table/use-remember-table-sorting.ts
```diff
@@ -0,0 +1,82 @@
...
+        const dataInArray = arrayify(dataVal);
+        const sortableOnly = dataInArray.filter(item => item.column && sortableColumns.includes(item.column));
+
+        if (sortableOnly.length > 0) {
+          set(sort, isArray ? sortableOnly : sortableOnly[0]);
+        }
+      }
+    }
+
+    watch(sort, (sort) => {
```
kelsos, line 73: Is the watcher here intentionally in the `onBeforeMount`?

kelsos:
@lukicenturi it looks good but I have a single question.

frontend/app/src/modules/table/use-remember-table-sorting.ts
lukicenturi: yeah, it's intentional
If I put it outside, the initial definition of the sort on the component will trigger this sort, and it will be forced to be saved and used. (the saved sort on localStorage will always be ignored)

kelsos (approved):
ok so @lukicenturi lgtm but can you maybe before merging add a note to that internal watch explaining why it's needed just for future reference?

PR #9729, opened 2025-04-04
fix: onchain send token bugs

lukicenturi:
Follow up of https://github.com/rotki/rotki/pull/9692
- [x] Show loading alert too in the asset selector dialog
- [x] Filter out connected address, and invalid eth address from the recipient address options


kelsos (approved):
This PR looks, so we can proceed to merge.

To follow up there are a few minor things I noticed.
- If you disconnect the session from the app and then go to the bridge the disconnect button gives an error
- use-wallet-store gives an error
```
use-wallet-store.ts:164 [Vue warn]: onUnmounted is called when there is no active component instance to be associated with. Lifecycle injection APIs can only be used during execution of setup(). If you are using async setup(), make sure to register lifecycle hooks before the first await statement.
```

PR #9734, opened 2025-04-04
Use SwapEvents in coinbase

nicholasyoder:
Related #6097


rotkehlchen/exchanges/coinbase.py
```diff
@@ -84,119 +88,6 @@
 )
 
 
-def trade_from_conversion(tx_a: dict[str, Any], tx_b: dict[str, Any] | None) -> Trade | None:
```
nicholasyoder, line 87: This function was moved to within the `Coinbase` class so `self.name` can be used for the `location_label`.

LefterisJP (approved):
LGTM. I guess it's fine. Can't see anything weird. Just hope the logic does not break. We need to heavily test trades with these new changes across all exchanges.

PR #9741, opened 2025-04-07
feat: show price oracle for asset price

lukicenturi:
Closes #(issue_number)

## Checklist


frontend/app/src/components/helper/CopyTooltip.vue
```diff
@@ -1,15 +1,32 @@
...
+    return undefined;
+
+  const mapping: Record<string, string> = {
+    [PriceOracle.MANUALCURRENT]: t('oracles.manual_current'),
+    [PriceOracle.UNISWAP2]: t('oracles.uniswap_v2'),
+    [PriceOracle.UNISWAP3]: t('oracles.uniswap_v3'),
+  };
+
+  return mapping[oracleVal] || oracleVal;
+});
```
kelsos, line 29: Can we do it in a way where the `CopyTooltip` component is not aware of the oracles?

Maybe we introduce a slot to the component so that if we need custom content we can add it in the slot.

kelsos (approved):
lgtm, I removed the used oracle property and re-pushed

PR #9743, opened 2025-04-07
Fix incorrect variable reference in merge log output

PixelPil0t1:
The original message referenced a variable file3, which is undefined in this context. This could lead to confusion when reading logs or debugging.
By updating it to use file2, the log now accurately reflects the actual merge process and improves clarity.


kelsos:
Hey @PixelPil0t1 thanks for the PR as this PR does not provide any value to the project and looks similar to other AI low effort PR that are farming reputation, which is clearly against our [Contribution Policy](https://github.com/rotki/rotki/blob/develop/CONTRIBUTING.md).

PR #9744, opened 2025-04-07
Use SwapEvents in gemini

nicholasyoder:
Related: #6097


rotkehlchen/exchanges/gemini.py
```diff
@@ -461,20 +464,30 @@ def query_online_trade_history(
                     if timestamp > end_ts:
                         break
 
+                    if (unique_id := str(entry['tid'])) in processed_ids:
+                        continue  # Skip already processed trade
```
nicholasyoder, line 468: Added this due to the `test_gemini_query_all_trades_pagination` test. Apparently before this it was relying on the `INSERT OR IGNORE` when adding events at the db to handle duplicate trades. This would still work here, but I noticed this since now in the test I use `query_online_history_events` which returns the events before they are written to the db, so it was now encountering a duplicate trade. It seems better to skip duplicates here and avoid processing them unnecessarily, even if they would get skipped later at the db.

LefterisJP (approved):
LGTM

PR #9753, opened 2025-04-08
Use mocking in kucoin tests

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=105529855


rotkehlchen/tests/exchanges/test_kucoin.py
```diff
@@ -839,38 +1051,16 @@ def mock_api_query_response(case, options):  # pylint: disable=unused-argument
...
         )
 
     assert asset_movements == expected_asset_movements
-
-
-@pytest.mark.skip('Fails with Kucoin old trades JSON response is missing key: "data"')
-@pytest.mark.parametrize('should_mock_current_price_queries', [True])
-def test_query_old_trades_sandbox(sandbox_kucoin, inquirer):  # pylint: disable=unused-argument
-    """Test that the endpoint for old trades returns valid results from the api
-    in very old timestamps
```
nicholasyoder, line 861: Simply removing this test, since we don't have an account with any old trades to use for example data for mocking. I guess if we ever have trouble with this then we'll be able to make a new test using data from the account that's having the issue.

LefterisJP (approved):
LGTM. I will just merge this, as no need to rerun tests since the lint failure is fixed in develop

PR #9758, opened 2025-04-09
migrate trades to swap events for bisq importer

prettyirrelevant:


rotkehlchen/data_import/importers/bisq_trades.py
```diff
@@ -40,37 +41,23 @@ def _consume_bisq_trade(
...
         assets1_symbol, assets2_symbol = csv_row['Market'].split('/')
-        if offer[0] == 'Sell':
-            trade_type = TradeType.SELL
-            if offer[1] == assets1_symbol:
-                base_asset = symbol_to_asset_or_token(assets1_symbol)
-                quote_asset = symbol_to_asset_or_token(assets2_symbol)
-            else:
-                base_asset = symbol_to_asset_or_token(assets2_symbol)
-                quote_asset = symbol_to_asset_or_token(assets1_symbol)
+        rate = Price(deserialize_asset_amount(csv_row['Price']))
```
nicholasyoder, line 48: Just pass this directly to `get_swap_spend_receive`. It doesn't look like the `rate` variable is used anywhere else.

rotkehlchen/data_import/importers/bisq_trades.py
```diff
@@ -40,37 +41,23 @@ def _consume_bisq_trade(
...
         if csv_row['Status'] == 'Canceled':
             return
-        timestamp = deserialize_timestamp_from_date(
-            date=csv_row['Date/Time'],
-            formatstr=timestamp_format,
-            location='Bisq',
-        )
+
         # Get assets and amounts sold
         offer = csv_row['Offer type'].split()
```
nicholasyoder, line 46:
```suggestion
        trade_type, base_asset_symbol = csv_row['Offer type'].split()
```
I think something like this would be an improvement here. Using `offer[x]` is a bit confusing.

rotkehlchen/data_import/importers/bisq_trades.py
```diff
@@ -79,20 +66,31 @@ def _consume_bisq_trade(
...
+                    date=csv_row['Date/Time'],
+                    formatstr=timestamp_format,
+                    location='Bisq',
+                )),
+                spend_amount=spend_amount,
+                spend_asset=spend_asset,
+                receive_asset=receive_asset,
+                receive_amount=receive_amount,
+                location=Location.BISQ,
+                spend_notes=f'ID: {csv_row["Trade ID"]}',
```
nicholasyoder, line 89: If the trade id is actually present here it should probably be used for the `unique_id`.

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -46,10 +47,11 @@
 from rotkehlchen.tests.utils.constants import (
     A_AXS,
     A_CRO,
+    A_DASH,
     A_GBP,
     A_KCS,
     A_MCO,
-    A_NANO,
+   A_NANO,
```
nicholasyoder, line 54: I think maybe this was a mistake? should keep it indented the same as the others.

PR #9768, opened 2025-04-10
Fix handling of adjustments in Kraken

nicholasyoder:
Fixes a problem that was introduced in #9673 where adjustment events were no longer getting written to the db.


LefterisJP (approved):
LGTM

PR #9773, opened 2025-04-11
Do backups in a new thread

### LINKED ISSUE #9772: Creating a backup doesn't allow gevent to context switch
Right now creating a backup involves compressing and encrypting the data and this is a very CPU intensive logic that doesn't allow gevent to context switch. This prevents the app from answering API calls and in some extreme cases crashes it.

## What we can do

We need to check if it is possible to spawn it in a different thread so it doesn't block the main app or if there isn't any solution see how we can spawn a different process for it

yabirgb:
Closes #9772

The backup logic compresses and encrypts the database before uploading it. This is a very intensive CPU logic and was blocking the app since it doesn't context switch. This commit spawns it in a thread allowing the api to continue answering requests.

It increases the threadpool max size to allow this additional thread for backups.


rotkehlchen/server.py
```diff
@@ -54,7 +54,7 @@ def main(self) -> None:
         hub.exception_stream = None
         # we don't use threadpool much so go to 2 instead of default 10
         hub.threadpool_size = 2
-        hub.threadpool.maxsize = 2
+        hub.threadpool.maxsize = 3
```
LefterisJP, line 57: why do you need the extra thread here? We use all 2 and you need a third for this?

rotkehlchen/premium/sync.py
```diff
@@ -196,7 +198,9 @@ def maybe_upload_data_to_server(
             self.last_upload_attempt_ts = ts_now()
             return False, message
 
-        data, our_hash = self.data.compress_and_encrypt_db()
+        current_pool = gevent.get_hub().threadpool
+        greenlet = current_pool.spawn(self.data.compress_and_encrypt_db)
```
LefterisJP, line 202: nothing to lose by doing this right?

```suggestion
        greenlet = gevent.get_hub().threadpool.spawn(self.data.compress_and_encrypt_db)
```

LefterisJP (commented):
you tested it and it works right?

rotkehlchen/server.py
yabirgb: I thought it was needed but seems that it doesn't affect after checking with the app. 2 seems fine

LefterisJP (approved):
LGTM

PR #9774, opened 2025-04-11
migrate trades to swap events for cryptocom & blockfi trades importer

prettyirrelevant:

### DIFF
--- a/rotkehlchen/tests/utils/dataimport.py
+++ b/rotkehlchen/tests/utils/dataimport.py
@@ -206,11 +206,9 @@ def assert_cointracking_import_results(rotki: Rotkehlchen, websocket_connection:
 def assert_cryptocom_import_results(rotki: Rotkehlchen):
     """A utility function to help assert on correctness of importing data from crypto.com"""
     with rotki.data.db.conn.read_ctx() as cursor:
-        trades = rotki.data.db.get_trades(cursor, filter_query=TradesFilterQuery.make(), has_premium=True)  # noqa: E501
-        events_db = DBHistoryEvents(rotki.data.db)
-        events = events_db.get_history_events(
+        events = DBHistoryEvents(rotki.data.db).get_history_events(
             cursor=cursor,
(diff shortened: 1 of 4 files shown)


rotkehlchen/tests/utils/dataimport.py
```diff
@@ -611,9 +663,17 @@ def assert_cryptocom_special_events_import_results(rotki: Rotkehlchen):
         event_subtype=HistoryEventSubType.NONE,
         amount=FVal('0.5'),
         asset=A_MCO,
-        notes=get_cryptocom_note('MCO Stake Rewards'),
+        notes='MCO Stake Rewards',
+    ), SwapEvent(
```
nicholasyoder, line 667: maybe change the timestamp on this event in the test data or change the ordering in the query here in the test so that the swap spend/receive aren't separated. Its kind of confusing to have them separated like this.

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -200,298 +198,343 @@ def assert_cryptocom_import_results(rotki: Rotkehlchen):
...
+        identifier=30,
         event_identifier='CCM_5450de111ff9e242523f011a42f2b1ed9684110bf2cb1b3c893ea3c5957bc3db',
         sequence_index=0,
         timestamp=TimestampMS(1616669547000),
         location=Location.CRYPTOCOM,
         event_type=HistoryEventType.SPEND,
         event_subtype=HistoryEventSubType.NONE,
-        amount=FVal('15.38'),
         asset=A_CRO,
+        amount=FVal('15.38'),
```
nicholasyoder, line 472: Probably not a big deal, but I'm not sure that its a good idea to move arguments like this for no reason. It will make git blame messy later if someone wants to find if this value was changed at some point or something. It looks like maybe amount was moved on a number of these events.

LefterisJP (approved):
LGTM. The failing test was fixed in https://github.com/rotki/rotki/pull/9788

PR #9775, opened 2025-04-11
Add new EvmSwapEvent class and api schema

nicholasyoder:


rotkehlchen/history/events/structures/evm_swap.py
```diff
@@ -0,0 +1,141 @@
...
+            tx_hash=deserialize_evm_tx_hash(entry[12]),
+            counterparty=entry[13],
+            product=EvmProduct.deserialize(entry[14]) if entry[14] is not None else None,
+            address=deserialize_optional(input_val=entry[15], fn=string_to_evm_address),
+        )
+
+    def serialize(self) -> dict[str, Any]:
+        """Serialize the event for api."""
+        serialized_data = super().serialize()  # Super is EvmEvent.
+        serialized_data['auto_notes'] = self._generate_auto_notes()
```
yabirgb, line 102: I believe this is possible and more explicit

```suggestion
        serialized_data = super(EvmEvent, self).serialize()
        serialized_data['auto_notes'] = self._generate_auto_notes()
```

rotkehlchen/history/events/structures/swap.py
```diff
@@ -28,6 +33,23 @@ class SwapEventExtraData(TypedDict):
     reference: NotRequired[str]
 
 
+class SwapEventEntryData(TypedDict):
+    timestamp: TimestampMS
+    location: Location
```
yabirgb, line 38: can you add a dosctring explaining what is this used for?

rotkehlchen/api/rest_helpers/history_events.py
```diff
@@ -27,18 +28,32 @@ def edit_grouped_events_with_optional_fee(
...
+            # Only count events in this group using the consecutive sequence indexes of the
+            # spend/receive/fee events.
+            existing_events_num = read_cursor.execute(
+                'SELECT COUNT(*) FROM history_events WHERE event_identifier=? AND entry_type=? '
+                'AND ((subtype=? AND sequence_index=?) OR (subtype=? AND sequence_index=?) '
+                'OR (subtype=? AND sequence_index=?))',
+                (event_identifier, entry_type.serialize_for_db(),
+                 HistoryEventSubType.SPEND.serialize(), sequence_index,
+                 HistoryEventSubType.RECEIVE.serialize(), sequence_index + 1,
+                 HistoryEventSubType.FEE.serialize(), sequence_index + 2),
```
yabirgb, line 50: is this something that you has enforced? This impose the restriction that all the sequence indices need to be consecutive but If I recall correctly this is not actually impossed

yabirgb: I have the intuition that you need `select subtype where sequence_index >= ?` and then loop that.

rotkehlchen/api/rest_helpers/history_events.py
```diff
@@ -27,18 +28,32 @@ def edit_grouped_events_with_optional_fee(
     db = events_db.db
     with db.conn.read_ctx() as read_cursor:
         read_cursor.execute(
-            'SELECT event_identifier FROM history_events WHERE identifier=?',
+            'SELECT event_identifier, sequence_index FROM history_events WHERE identifier=?',
```
yabirgb, line 31: by reading this function do you see if we can actually improve it with the new logic for trades? I see it really complicated for some reason

rotkehlchen/api/v1/schemas.py
```diff
@@ -852,6 +853,51 @@ class BaseEventSchema(BaseSchema):
         sequence_index = fields.Integer(required=True)
         location_label = fields.String(load_default=None)
 
+    class BaseEvmEventSchema(Schema):
+        tx_hash = EVMTransactionHashField(required=True)
```
yabirgb, line 857: Add a dosctrring for what it represents

rotkehlchen/api/v1/schemas.py
```diff
@@ -852,6 +853,51 @@ class BaseEventSchema(BaseSchema):
...
+        event_identifier = fields.String(required=False, load_default=None)
+        counterparty = fields.String(load_default=None)
+        product = SerializableEnumField(enum_class=EvmProduct, load_default=None)
+        address = EvmAddressField(load_default=None)
+        extra_data = fields.Dict(load_default=None)
+        location = LocationField(required=True, limit_to=EVM_EVMLIKE_LOCATIONS)
+
+    class BaseSwapEventSchema(Schema):
+        identifier = fields.Integer(required=True)
```
yabirgb, line 866: docstring

rotkehlchen/api/rest.py
```diff
@@ -1088,6 +1088,7 @@ def edit_history_events(self, events: list['HistoryBaseEntry']) -> Response:
         if (events_type := events[0].entry_type) in {
             HistoryBaseEntryType.ASSET_MOVEMENT_EVENT,
             HistoryBaseEntryType.SWAP_EVENT,
+            HistoryBaseEntryType.EVM_SWAP_EVENT,
```
yabirgb, line 1091: this requires adjusments in docs/api.rst right?

rotkehlchen/api/v1/schemas.py
```diff
@@ -1108,15 +1119,79 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
...
+                    data=data,
+                    subtype=HistoryEventSubType.RECEIVE,
+                    sequence_index_offset=1,
+                ),
+                event_identifier=event_identifier,
+            )]
+            if (
+                (fee_asset := data['fee_asset']) is not None and
+                (fee_amount := data['fee_amount']) != ZERO
+            ):
```
yabirgb, line 1175: this is already ensured by the base schema validation right? You only need to check one unless I'm mistaken

rotkehlchen/api/rest_helpers/history_events.py
yabirgb: actually It just came to my mind that we will re-decode everything and do the post decoding so I guess it can be enforced there

rotkehlchen/history/events/structures/evm_swap.py
nicholasyoder: actually `super(EvmEvent, self)` gets the next one back after EvmEvent (SwapEvent), which is apparently the expected behavior of python's method resolution order. 

But I've changed this to `EvmEvent.serialize(self)` instead since that also makes it more explicit.

rotkehlchen/api/rest_helpers/history_events.py
nicholasyoder: > is this something that you has enforced?

Unlike SwapEvents and AssetMovements the sequence index can't be hardcoded in the EvmSwapEvent class itself since there can be multiple groups of swap events within a single transaction (therefore needing different indexes). But we will only be creating these at a limited number of places in the code. Currently they are only created in the `CreateEvmSwapEventSchema` where the consecutive sequence indexes are being enforced - and then also they will get created in the post decoding logic which we still need to add, where we will also need to make sure these indexes are always consecutive.

> I have the intuition that you need select subtype where sequence_index >= ? and then loop that.

Not sure I totally follow. We don't ever want to have other types of events between the events in a swap event group do we? And this would still require all the events in the group to be ordered correctly, so if we need to assume they are in the correct order here (which can't be enforced either except in the code where we create them) we might as well require them to be consecutive as well, right?

rotkehlchen/api/rest_helpers/history_events.py
nicholasyoder: mmm... unfortunately I don't really see any way to fundamentally change how this function works to make it simpler. However, I did make some adjustments to make the code more compact and hopefully more readable.

rotkehlchen/api/rest.py
nicholasyoder: ah, yes. Added a section to the docs for EvmSwapEvents like we have for the other types of events.

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -923,3 +924,156 @@ def test_add_edit_swap_events(rotkehlchen_api_server: 'APIServer') -> None:
         'extra_data': {'reference': 'TRADE2'},
         'auto_notes': 'Swap 0.01 ETH in Bitfinex',
     }
+
+
+def test_add_edit_evm_swap_events(rotkehlchen_api_server: 'APIServer') -> None:
+    rotki = rotkehlchen_api_server.rest_api.rotkehlchen
+    db = DBHistoryEvents(rotki.data.db)
+    entries = [
+        {
```
LefterisJP, line 933:
```suggestion
    entries = [{
```

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -923,3 +924,156 @@ def test_add_edit_swap_events(rotkehlchen_api_server: 'APIServer') -> None:
...
+            'spend_asset': A_USDT.identifier,
+            'receive_amount': '0.026',
+            'receive_asset': 'ETH',
+            'sequence_index': 123,
+            'tx_hash': '0x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f',
+            'counterparty': 'some counterparty',
+            'product': 'pool',
+            'address': '0xb5d85CBf7cB3EE0D56b3bB207D5Fc4B82f43F511',
+        },
+    ]
```
LefterisJP, line 964:
```suggestion
    }]
```

docs/api.rst
```diff
@@ -5555,12 +5556,59 @@ Dealing with History Events
...
+                "fee_asset": "ETH",
+                "user_notes": ["Example note", "", ""],
+                "sequence_index": 0,
+                "tx_hash": "0x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f",
+                "counterparty": "some counterparty",
+                "address": "0xA090e606E30bD747d4E6245a1517EbE430F0057e"
+            }
+
+         :reqjson string tx_hash: This is the transaction hash of the evm event
+         :reqjson int sequence_index: This is an index that tries to provide the order of history entries for a single event_identifier.
```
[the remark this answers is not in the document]
nicholasyoder: > This would be the first sequence index in the events right

yes, since there can be multiple groups of evm swap events in the same transaction they can't have their sequence indexes hardcoded to start at 0 like the plain swap events. 

Added more explanation to the description here.

rotkehlchen/api/rest_helpers/history_events.py
```diff
@@ -26,51 +27,54 @@ def edit_grouped_events_with_optional_fee(
...
             mapping_values={HISTORY_MAPPING_KEY_STATE: HISTORY_MAPPING_STATE_CUSTOMIZED},
         )
     else:
-        assert edit_events_num == existing_events_num and existing_events_num in (no_fee_num, with_fee_num)  # noqa: E501
-        for event in events:
-            events_db.edit_history_event(
-                event=event,
-                write_cursor=write_cursor,
-            )
+        assert new_event_count == existing_event_count and existing_event_count in (no_fee_num, with_fee_num)  # noqa: E501
```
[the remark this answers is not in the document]
nicholasyoder: mmm... I guess because its a programming error if those conditions are false at that point in the code. It looks like it was introduced here: https://github.com/rotki/rotki/commit/1f96f55e81c56e04ba8a9abdf0f19f2032dfd19b#diff-9308b914d18f01109623e57290ddda84ae9491ecb3ade05d6ad515d1c69038a1R63 Perhaps @yabirgb would know better why this is here since he added it originally.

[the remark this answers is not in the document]
yabirgb: I don't recall but most probably is to detect easily possible issues in the logic

rotkehlchen/api/v1/schemas.py
```diff
@@ -996,7 +1009,7 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
                 spend_amount=data['spend_amount'],
                 receive_asset=data['receive_asset'],
                 receive_amount=data['receive_amount'],
-                fee_amount=fee_amount or ZERO,
+                fee_amount=data['fee_amount'] or ZERO,
```
LefterisJP, line 1012: when the other PR gets merged this needs to change too right? to be None

rotkehlchen/api/rest_helpers/history_events.py
```diff
@@ -26,51 +27,54 @@ def edit_grouped_events_with_optional_fee(
     """
     db = events_db.db
     with db.conn.read_ctx() as read_cursor:
-        read_cursor.execute(
-            'SELECT event_identifier FROM history_events WHERE identifier=?',
+        if (result := read_cursor.execute(
+            'SELECT event_identifier, sequence_index FROM history_events WHERE identifier=?',
```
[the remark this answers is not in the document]
LefterisJP: Yes I was referring to the query. That makes sense. I thought it was querying by `event_identifier` ... our naming sucks :sweat_smile: 

Can you maybe add a small comment here or add to the docstring to make this a bit clearer?

rotkehlchen/api/v1/schemas.py
nicholasyoder: yes, will change in the other PR.

LefterisJP (approved):
LGTM

PR #9776, opened 2025-04-11
migrate trades to swap events for rotki generic trades importer

prettyirrelevant:


rotkehlchen/data_import/importers/rotki_trades.py
```diff
@@ -37,30 +34,39 @@ def _consume_rotki_trades(
...
-        asset, fee, fee_currency, location, timestamp = process_rotki_generic_import_csv_fields(csv_row, 'Base Currency')  # noqa: E501
-        if (trade_type := TradeType.deserialize(csv_row['Type'])) == TradeType.BUY:
-            rate = amount_sold / amount_bought
-            amount = amount_bought
+        base_asset, fee, fee_currency, location, timestamp = process_rotki_generic_import_csv_fields(csv_row, 'Base Currency')  # noqa: E501
+        quote_asset = symbol_to_asset_or_token(csv_row['Quote Currency'])
+
+        # For a buy: spend quote asset, receive base asset
+        # For a sell: spend base asset, receive quote asset
+        if csv_row['Type'].strip().lower() == 'buy':
```
yabirgb, line 45: why not use the deserialize as it was before? It's safer than using hardcoded strings and also easier to maintain

rotkehlchen/data_import/importers/rotki_trades.py
```diff
@@ -37,30 +34,39 @@ def _consume_rotki_trades(
...
-            rate=Price(rate),
-            base_asset=asset,
-            quote_asset=symbol_to_asset_or_token(csv_row['Quote Currency']),
-            trade_type=trade_type,
-            amount=amount,
-            notes=csv_row['Description'],
+            spend_asset = base_asset
+            spend_amount = amount_sold
+            receive_asset = quote_asset
+            receive_amount = amount_bought
```
yabirgb, line 54: spend amount and receive amount are the same in both cases so you don't need to duplicate the logic

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -2283,62 +2283,115 @@ def assert_binance_import_results(rotki: Rotkehlchen, websocket_connection: Webs
...
-    assert trades == expected_trades
     websocket_connection.wait_until_messages_num(num=1, timeout=10)
     assert websocket_connection.pop_message() == {
         'type': 'progress_updates',
         'data': {
             'subtype': 'csv_import_result',
             'source_name': 'Rotki generic trades',
             'total': 5,
-            'processed': 3,
+            'processed': 4,
```
yabirgb, line 2293: why is there one more?

rotkehlchen/tests/utils/dataimport.py
```diff
@@ -2283,62 +2283,115 @@ def assert_binance_import_results(rotki: Rotkehlchen, websocket_connection: Webs
...
+        amount=FVal('1875.64'),
+        notes='Trade USDC for ETH',
+    ), SwapEvent(
+        identifier=2,
+        timestamp=TimestampMS(1659085200000),
+        event_identifier='30a4f24e74690eda15c201ad037844e66725e097048459817264bf27675992e3',
+        location=Location.BINANCE,
+        event_subtype=HistoryEventSubType.RECEIVE,
+        asset=A_ETH,
+        amount=FVal('1.0000'),
```
yabirgb, line 2324: use ONE here

rotkehlchen/tests/utils/dataimport.py
prettyirrelevant: previously, we used to handle a division by zero error before but that's not the case anymore because we directly use the bought & sold amount.

rotkehlchen/data_import/importers/rotki_trades.py
prettyirrelevant: because `TradeType.deserialize()` will go away with all the trade structure.

rotkehlchen/tests/utils/dataimport.py
yabirgb: I guess it makes sense to keep it to cover the case and see that it doesn't break the logic

rotkehlchen/data_import/importers/rotki_trades.py
```diff
@@ -37,30 +34,35 @@ def _consume_rotki_trades(
...
         amount_sold = deserialize_asset_amount(csv_row['Sell Amount'])
         amount_bought = deserialize_asset_amount(csv_row['Buy Amount'])
-        asset, fee, fee_currency, location, timestamp = process_rotki_generic_import_csv_fields(csv_row, 'Base Currency')  # noqa: E501
-        if (trade_type := TradeType.deserialize(csv_row['Type'])) == TradeType.BUY:
-            rate = amount_sold / amount_bought
-            amount = amount_bought
+        base_asset, fee, fee_currency, location, timestamp = process_rotki_generic_import_csv_fields(csv_row, 'Base Currency')  # noqa: E501
+        quote_asset = symbol_to_asset_or_token(csv_row['Quote Currency'])
+
+        # For a buy: spend quote asset, receive base asset
```
LefterisJP, line 43: Why not also change the CSV to call it spend asset and receive asset to be more obvious?

LefterisJP (approved):
LGTM. Failed test is being fixed here: https://github.com/rotki/rotki/pull/9784

PR #9785, opened 2025-04-14
migrate trades to swap events for binance importer

prettyirrelevant:
Part of https://github.com/orgs/rotki/projects/11?pane=issue&itemId=102361655


rotkehlchen/data_import/importers/binance.py
```diff
@@ -315,40 +314,49 @@ def process_trades(
...
-            rate = to_amount / from_amount
-            trade = Trade(
-                timestamp=timestamp,
+            all_swap_events.extend(create_swap_events(
+                timestamp=ts_sec_to_ms(timestamp),
                 location=Location.BINANCE,
-                trade_type=trade_type,
-                base_asset=to_asset,
-                quote_asset=from_asset,
-                amount=to_amount,
```
nicholasyoder, line 325: mmm... yeah. This was all wrong. Rate was calculated incorrectly, and even the assets and amount should have been set to different things here depending on if it was a buy or sell I think.

As far as I can see the new SwapEvents in the test line up with the actual test data so I think it the new swap events code is correct.

Its surprising we never noticed that this was importing trades all wrong. I guess most people just connected to binance via the api instead of importing from csv.

rotkehlchen/data_import/importers/binance.py
```diff
@@ -315,40 +317,43 @@ def process_trades(
...
-            raw_trades.append(trade)
-
-        # Sometimes we can get absolutely identical trades (including timestamp) but the database
-        # allows us to add only one of them. So we combine these trades into a huge single trade
-        # First step: group trades
-        grouped_trades: dict[TradeID, list[Trade]] = defaultdict(list)
-        for trade in raw_trades:
-            grouped_trades[trade.identifier].append(trade)
+            for event in swap_events:
+                grouped_events[event.event_identifier].append(event)
```
nicholasyoder, line 333:
```suggestion
            grouped_events[event.event_identifier].extend(swap_events)
```
No need to loop here. Actually don't even use the swap_events variable - put `create_swap_events` inside the extend.

rotkehlchen/data_import/importers/binance.py
```diff
@@ -273,8 +272,11 @@ def process_trades(
                 cur_batch.append(rows_grouped_by_fee[True].pop())
             grouped_trade_rows.append(cur_batch)
 
-        # Creating trades structures based on grouped rows data
-        raw_trades: list[Trade] = []
+        # Sometimes we can get absolutely identical swap events
+        # (including timestamp) but the database allows us to add only one of them.
+        # So we combine these swap events into a huge group of swap events
+        # First step: group swap events
```
nicholasyoder, line 278:
```suggestion
        # First step: create and group swap events
```
we're not just grouping them here, this is also where they get created.

rotkehlchen/data_import/importers/binance.py
prettyirrelevant: sorry, fixed it

rotkehlchen/data_import/importers/binance.py
```diff
@@ -315,40 +313,20 @@ def process_trades(
                     )
                 continue
 
-            rate = to_amount / from_amount
-            trade = Trade(
-                timestamp=timestamp,
+            swap_events.extend(create_swap_events(   # any row works, just using the first to create the event identifier  # noqa: E501
```
LefterisJP, line 316: shouldn't this comment be on line 325?

rotkehlchen/data_import/importers/binance.py
```diff
@@ -315,40 +313,20 @@ def process_trades(
...
-
-        return unique_trades
+                spend_asset=from_asset,
+                spend_amount=from_amount,
+                receive_asset=to_asset,
+                receive_amount=to_amount,
+                fee_amount=fee_amount,  # type: ignore[arg-type]
+                fee_asset=fee_asset,  # type: ignore[arg-type]
+                event_identifier=f'{EVENT_IDENTIFIER_PREFIX}{hash_binance_csv_row(trade_rows[0])}',
+                spend_notes='Imported from binance CSV file. Binance operation: Buy / Sell',
```
LefterisJP, line 326: Question here: Will these be fixed during DB upgrade? You said they were imported completely wrong before. Couldn't you add some logic using the notes here to identify and fix?

LefterisJP: Isaac made a note for this here: https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=106596282

LefterisJP (approved):
LGTM

PR #9792, opened 2025-04-14
Fix typing in abi.py

nicholasyoder:


rotkehlchen/chain/ethereum/abi.py
```diff
@@ -93,7 +93,7 @@ def decode_event_data_abi(
     if duplicate_names:
         raise DeserializationError(
             f'The following argument names are duplicated '
-            f"between event inputs: '{', '.join(duplicate_names)}'",
+            f"between event inputs: '{', '.join({name for name in duplicate_names if name is not None})}'",  # noqa: E501
```
LefterisJP, line 96: Hmm I don't get this. Why do you need the none check? get_abi_input_names returns a list of string. They can't be None right?

LefterisJP: Implementation of 

```python
def get_abi_input_names(abi_element: ABIElement) -> List[str]:
    """
    Return names for each input from the function or event ABI.

    :param abi_element: ABI element.
    :type abi_element: `ABIElement`
    :return: Names for each input in the function or event ABI.
    :rtype: `List[str]`

    .. doctest::

        >>> from eth_utils import get_abi_input_names
        >>> abi = {
        ...   'constant': False,
        ...   'inputs': [
        ...     {
        ...       'name': 's',
        ...       'type': 'uint256'
        ...     }
        ...   ],
        ...   'name': 'f',
        ...   'outputs': [],
        ...   'payable': False,
        ...   'stateMutability': 'nonpayable',
        ...   'type': 'function'
        ... }
        >>> get_abi_input_names(abi)
        ['s']
    """
    _raise_if_fallback_or_receive_abi(abi_element)
    return [
        arg["name"]
        for arg in cast(Sequence[ABIComponent], abi_element.get("inputs", []))
    ]
```

ABIComponent definition

```python
class ABIComponent(TypedDict):
    """
    TypedDict representing an `ABIElement` component.
    """

    name: str
    """Name of the component."""
    type: str
    """Type of the component."""
    components: NotRequired[Sequence["ABIComponent"]]
    """List of nested `ABI` components for ABI types."""
```

Thus the return can never be a list of None. I would just type ignore. This looks liek a mypy bug. Instead of adding a new if check in a for loop like you did

LefterisJP: Actually @nicholasyoder pointed out that they just changed this exact code, which is what caused the failure: https://discord.com/channels/657906918408585217/1080783409640648765/1361456828499890308

Issue #5071, opened 2022-11-01
Change how the backend and frontend deals with prices/amounts (usd_value)


kelsos:
## Abstract

At the moment all the prices/amounts from the backend come in `USD` this can lead to double conversion problems since the frontend doesn't know what the original currency was and it will attempt to reconvert from `USD` to the display currency that the user has selected.

This can lead to something that was supposed to be 60 `EUR` displaying as 59.999.

## Motivation

We need to get rid of this problem at its root and that can happen only by the backend returning the amounts/prices in the `main_currency`

## Specification

- Undo the temporary hack we introduced: https://github.com/rotki/rotki/pull/5069
- Replace the assumption that all prices returned are in `usd_value` with that that they are in `main_currency`
- Backend returns all amounts in `main_currency` and does any necessary conversion, frontend displays these values as it is.
- Massively rename the `usd_value` fields to `value` in the schema.
- Frontend starts requesting prices in `main_currency`
  - Refresh and recalculate prices on currency change 
  - Make sure that all the fiat-currency='USD' fields now use the `main_currency` instead (maybe use a wrapper component to encapsulate this)

LefterisJP:
Woudl also need to undo this commit: https://github.com/rotki/rotki/pull/5081/commits/61b193a2ae39f14bbba6a69d197d13db6dadcb34

LefterisJP:
Adding more thoughts here, after our discussion with @kelsos yesterday.

This is bigger than just the double conversion fix, but also deals in general with the way we save values in the DB. Fixing this will also address all the views in the app where we show historical USD value but converted to the profit currency with the current rate. In the evm transactions view, staking view, history trades view etc. Also the graph statistics.

Relevant issues are:
- https://github.com/rotki/rotki/issues/2253
- https://github.com/rotki/rotki/issues/4031
- https://github.com/rotki/rotki/issues/3477
- https://github.com/rotki/rotki/issues/1790

To address all them we put some thought in the way rotki does things which has not changed since 2017 and is probably finally time to change.

Namely the fact that usd value at the time is saved for each event. We have the `Balance` structure that is amount + value (implicit usd).

And we save that value in the DB events. But that is redundant information. Since we have the `amount` and the `price` is cached in the globalDB's prices. So in a sense we could:

1. Remove `usd_value` from all the tables in the DB.
[...]

PR #9794, opened 2025-04-15
Remove calls to zerion sdk and price queries

yabirgb:
This PR does two things:

- Avoid querying prices of tokens after querying balances. It's a intermediate step until we get #5071 done but speeds up the startup of the app.
- Remove queries for the zerionsdk since they aren't optimized and when several ethereum accounts are tracked it slows down in a considerable manner the app. The sdk contract hasn't been updated in 2 years

I made a comparison running each branch 3 times and counting the time since the first query started for querying balances until all of them were queried

|master | this branch |
|-------| -------------|
|1:49|59|
|1:35|1:20|
|2:20|1:17|


rotkehlchen/tests/api/blockchain/test_base.py
```diff
@@ -1462,13 +1405,8 @@ def _remove_blockchain_accounts_test_start(
             eth_balances=eth_balances_after_removal,
             token_balances=token_balances_after_removal,
             also_btc=True,
-            expected_liabilities=after_liabilities,
         )
 
-    # check that after removing ethereum account defi balances were updated
-    if query_balances_before_first_modification:
```
LefterisJP, line 1469: hmmm don't you still need to check that the balances are no longer there? They are not defi balances anymore but should be normal balances too right? After removing the account they should be deleted right?

yabirgb: unless I'm mistaken is the if above this one. For some reason the condition was being checked twice

rotkehlchen/chain/ethereum/defi/zerionsdk.py
```diff
@@ -1,427 +1,6 @@
...
-        if usd_price is None:
-            return None
-
-        return DefiBalance(
-            token_address=deserialize_evm_address(token_address),
-            token_name=token_name,
-            token_symbol=token_symbol,
-            balance=Balance(normalized_balance, usd_value=normalized_balance * usd_price),
-        )
+ZERION_ADAPTER_ADDRESS: Final = string_to_evm_address('0x06FE76B2f432fdfEcAEf1a7d4f6C3d41B5861672')
```
LefterisJP, line 6: why keep this? Due to `uniswap_lp_token_balances` ? Why do we query this there?

rotkehlchen/tests/api/blockchain/test_base.py
LefterisJP: not sure. I guess maybe this was checking the in-memory defi balances which no longer exist so you can't keep it anyway.

LefterisJP (approved):
LGTM

PR #9796, opened 2025-04-15
refactor historical balances query to remove trades

prettyirrelevant:
Part of https://github.com/orgs/rotki/projects/11?pane=issue&itemId=102361655

### DIFF
--- a/rotkehlchen/balances/historical.py
+++ b/rotkehlchen/balances/historical.py
@@ -1,25 +1,16 @@
 import logging
-from collections import defaultdict, deque
-from typing import TYPE_CHECKING, Any, Final, TypedDict
+from collections import defaultdict
+from typing import TYPE_CHECKING, Final, TypedDict
 
 from rotkehlchen.assets.asset import Asset
 from rotkehlchen.constants import DAY_IN_SECONDS, ONE, ZERO
 from rotkehlchen.constants.prices import ZERO_PRICE
-from rotkehlchen.db.constants import HISTORY_BASE_ENTRY_FIELDS
 from rotkehlchen.db.filtering import (
-    DBFilter,
-    DBIgnoredAssetsFilter,
-    DBMultiStringFilter,
-    DBMultiValueFilter,
-    DBNestedFilter,
-    DBNotEqualFilter,
-    DBTimestampFilter,
+    HistoryEventFilterQuery,
 )
-from rotkehlchen.errors.asset import UnknownAsset
 from rotkehlchen.errors.misc import NotFoundError, RemoteError
 from rotkehlchen.errors.price import NoPriceForGivenTimestamp
 from rotkehlchen.errors.serialization import DeserializationError
-from rotkehlchen.exchanges.data_structures import Trade
 from rotkehlchen.fval import FVal
 from rotkehlchen.globaldb.handler import GlobalDBHandler
 from rotkehlchen.history.events.structures.base import HistoryEvent
@@ -29,7 +20,7 @@ from rotkehlchen.history.events.structures.types import (
 )
 from rotkehlchen.history.price import PriceHistorian
 from rotkehlchen.logging import RotkehlchenLogsAdapter
-from rotkehlchen.types import Timestamp, TradeType
+from rotkehlchen.types import Timestamp
 from rotkehlchen.utils.misc import timestamp_to_daystart_timestamp, ts_ms_to_sec
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


rotkehlchen/balances/historical.py
```diff
@@ -243,135 +218,42 @@ def _get_events_and_currency(
...
+                with_group_by=False,
+                with_pagination=False,
+                without_ignored_asset_filter=False,
             )
-
-            trades_buffer = deque(cursor1.fetchmany(BATCH_SIZE))
-            events_buffer = deque(cursor2.fetchmany(BATCH_SIZE))
-            trade_row, event_row = None, None
-            while trades_buffer or events_buffer:
+            for entry in cursor.execute(f'SELECT {filter_query.get_columns()} FROM history_events {where_clauses}', bindings):  # noqa: E501
```
nicholasyoder, line 251: Why not use `DBHistoryEvents().get_history_events(` here?

rotkehlchen/balances/historical.py
```diff
@@ -425,52 +307,15 @@ def _get_prices(
 
     @staticmethod
     def _update_balances(
-            event: HistoryEvent | Trade,
+            event: HistoryEvent,
             current_balances: dict[Asset, FVal],
-            assets: tuple[Asset, ...] | None = None,
     ) -> tuple[str | int | None, str | None] | None:
         """Updates current balances for a trade or history event, checking for negative balances.
```
nicholasyoder, line 313:
```suggestion
        """Updates current balances for a history event, checking for negative balances.
```

rotkehlchen/balances/historical.py
```diff
@@ -425,52 +307,15 @@ def _get_prices(
...
-                return event.identifier, None
-
-            current_balances[spend_asset] -= spend_amount
-            current_balances[receive_asset] += receive_amount
-            if current_balances[spend_asset] == ZERO:
-                del current_balances[spend_asset]
-
-            return None
-
         if (  # history event logic from here and down
```
nicholasyoder, line 319:
```suggestion
        if (
```

rotkehlchen/tests/api/test_historical_balances.py
```diff
@@ -158,19 +158,16 @@ def test_get_historical_asset_balance(
...
-                ),
-            ],
+            history=[*create_swap_events(
+                timestamp=ts_sec_to_ms(DAY_AFTER_START_TS),
+                location=Location.EXTERNAL,
+                spend_asset=A_EUR,
+                spend_amount=FVal('3200.0'),
+                receive_asset=A_BTC,
+                receive_amount=FVal('0.2'),
+            )],
```
nicholasyoder, line 170:
```suggestion
            history=create_swap_events(
                timestamp=ts_sec_to_ms(DAY_AFTER_START_TS),
                location=Location.EXTERNAL,
                spend_asset=A_EUR,
                spend_amount=FVal('3200.0'),
                receive_asset=A_BTC,
                receive_amount=FVal('0.2'),
            ),
```

rotkehlchen/tests/api/test_historical_balances.py
```diff
@@ -544,18 +528,19 @@ def test_get_historical_netvalue(
...
-    assert len(result['times']) == len(result['values']) == 2
+    assert len(result['times']) == len(result['values']) == 3
     assert all(ts in result['times'] for ts in expected_timestamps[::2])
 
     # Expected values after ignoring BTC
     expected_values_ignored = [
         FVal('32000'),  # Day 1: (10 ETH * 1200) + 20000 EUR
-        FVal('32000'),  # Day 3: (9.8 ETH * 1200) + (20000 + 240) EUR (after ETH sale)
+        FVal('28800.0'),  # Day 2: (10 ETH * 1200) + (20000 - 3200) EUR = 28800
+        FVal('28800.0'),  # Day 3: (9.8 ETH * 1200 = 11760) + (16800 + 240 = 17040) EUR = 28800
```
nicholasyoder, line 538: mmm... why did this change? same for the other things changed in the assertions in this test. Shouldn't this remain as it is if the trades (now swap events) are being processed correctly? or were trades actually being processed wrong before?

rotkehlchen/balances/historical.py
prettyirrelevant: `DBHistoryEvents().get_history_events()` does a lot of processing(resolving to different history event type) and an additional query. both of which aren't needed as the base history event is sufficient.

rotkehlchen/tests/api/test_historical_balances.py
prettyirrelevant: previously we skipped entire trades when either asset(base or quote) was ignored. Now, we only skip the specific ignored asset event in a swap while processing the other asset.

nicholasyoder: hmm... ok. yeah, I guess that is how it should be. Non-ignored assets should get processed even if they are part of a swap that goes to an ignored asset or vice versa.

LefterisJP (approved):
LGTM with 1 big comment on the ignoring functionality that I made in discord about ignoring whole swap or not which we should look out of this PR

PR #9800, opened 2025-04-15
remove asset movements autogenerated notes

prettyirrelevant:
Part of https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=106596282


rotkehlchen/db/upgrades/v47_v48.py
```diff
@@ -166,4 +167,28 @@ def _replace_history_note_locations(write_cursor: 'DBCursor') -> None:
...
+
+        Auto-generated notes were previously stored for asset movements (entry_type=6).
+        Now these are generated on the fly, so we remove them unless the event is customized.
+        """
+        write_cursor.execute(
+            """
+            UPDATE history_events
+            SET notes = NULL
+            WHERE entry_type = 6
+              AND notes IS NOT NULL
```
nicholasyoder, line 182: I think this whole query could be made more compact without hurting its readability - it seems quite spread out as is. Its better to not use more lines than needed.

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -3148,6 +3153,10 @@ def test_upgrade_db_47_to_48(user_data_dir, messages_aggregator):
             (1003, 'HISTORY'),
             (1004, 'DASHBOARD'),
         ]
+        assert cursor.execute('SELECT identifier, notes FROM history_events WHERE entry_type=6').fetchall() == [  # noqa: E501
```
LefterisJP, line 3156: please add the assertion here too.

LefterisJP (approved):
LGTM

PR #9804, opened 2025-04-16
Improve a bit time spent used converting trades

yabirgb:
The most part of the time spent in this db upgrade step is in the inserting part. The changes in this commit reduce the time from ~1:50 to ~1:15 the time converting the trades. The changes are:

- ~~Use fetchmany to avoid hitting the database on each event~~ debunked
- ~~Use a list instead of a tuple since it seems that sqlite processes it better for executemany~~ debunked
- Avoid wrapper function in the comprehension list

Closes #(issue_number)

## Checklist

### DIFF
--- a/rotkehlchen/db/upgrades/v47_v48.py
+++ b/rotkehlchen/db/upgrades/v47_v48.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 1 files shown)


rotkehlchen/db/upgrades/v47_v48.py
```diff
@@ -125,16 +126,17 @@ def _convert_trades_to_swap_events(write_cursor: 'DBCursor') -> None:
...
-            try:
-                new_events.extend(upgrade_trade_to_swap_events(
-                    row=row,
-                    kraken_ids_to_labels=kraken_ids_to_labels,
-                    kraken_adjustments_data=kraken_adjustments_data,
-                    kraken_ids_to_delete=kraken_ids_to_delete,
-                ))
-            except (DeserializationError, ValueError) as e:
-                log.error(f'Failed to deserialize row {row} due to {e}. Skipping.')
+        while (row_chunk := write_cursor.fetchmany(CHUNK_SIZE_TRADE_QUERY)):
```
LefterisJP, line 129: I don't think this is correct. This **should not** improve anything. The previous approach did not hit the DB in every iteration. If that was true all of our code around rotki woudl be wrong.

A cursor already loads everything into memory and you just iterate it. Here what you do is you just add another level of iteration. This can't be helping.

rotkehlchen/db/upgrades/v47_v48.py
```diff
@@ -147,7 +149,7 @@ def _convert_trades_to_swap_events(write_cursor: 'DBCursor') -> None:
             'INSERT OR IGNORE INTO history_events(entry_type, event_identifier, '
             'sequence_index, timestamp, location, location_label, asset, amount, '
             'notes, type, subtype, extra_data) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)',
-            (event.serialize_for_db()[0][2] for event in new_events),
+            [event._serialize_base_tuple_for_db()[2] for event in new_events],
```
LefterisJP, line 152: using the underlying _serialize, which is essentially the same for SwapEvent (for now) seems like a good reason to improve the speed by saving 1 function call per iteration.

LefterisJP (approved):
LGTM

PR #9805, opened 2025-04-16
Add post decoding swap processing

nicholasyoder:

Adds logic in post decoding to process swaps into EvmSwapEvents.

This would affect all the decoding tests involving swaps, so I also added a temporary check, so that it only does this for a few counterparties whose tests I've updated. This check will need removed and the rest of the tests updated in a subsequent PR.

### DIFF
--- a/rotkehlchen/chain/evm/decoding/decoder.py
+++ b/rotkehlchen/chain/evm/decoding/decoder.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING, Any, Optional, Protocol
 
 import gevent
 from gevent.lock import Semaphore
+from more_itertools import peekable
 from web3.exceptions import Web3Exception
 
 from rotkehlchen.accounting.structures.types import ActionType
@@ -57,6 +58,7 @@ from rotkehlchen.errors.misc import (
 from rotkehlchen.errors.serialization import ConversionError, DeserializationError
 from rotkehlchen.fval import FVal
 from rotkehlchen.history.events.structures.evm_event import EvmProduct
+from rotkehlchen.history.events.structures.evm_swap import EvmSwapEvent
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.tasks.assets import maybe_detect_new_tokens
@@ -78,6 +80,7 @@ from .constants import (
     ERC20_OR_ERC721_TRANSFER,
     OUTGOING_EVENT_TYPES,
 )
+from .llamazip.constants import CPT_LLAMAZIP
 from .structures import (
     DEFAULT_DECODING_OUTPUT,
     FAILED_ENRICHMENT_OUTPUT,
@@ -513,7 +516,7 @@ class EVMTransactionDecoder(ABC):
             decoded_events: list['EvmEvent'],
             all_logs: list[EvmTxReceiptLog],
             counterparties: set[str],
-    ) -> list['EvmEvent']:
+    ) -> tuple[list['EvmEvent'], bool]:
         """
         The post-decoding rules list consists of tuples (priority, rule) and must be
         sorted by priority in ascending order. The higher the priority number the later
@@ -521,7 +524,10 @@ class EVMTransactionDecoder(ABC):
         Matches post decoding rules to all matched counterparties propagated for decoding
         from the decoding/enriching rules and also the counterparties associated with the
         transaction to_address field.
+        Returns a tuple containing the list of decoded events and a boolean flag indicating
+        whether any post-decoding rules ran successfully and may have modified the events.
         """
+        maybe_modified = False
         if transaction.to_address is not None:
             address_counterparty = self.rules.addresses_to_counterparties.get(transaction.to_address)  # noqa: E501
             if address_counterparty is not None:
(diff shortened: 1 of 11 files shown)


rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -555,6 +560,78 @@ def run_all_post_decoding_rules(
...
+                next_event.event_subtype not in trade_subtypes
+            ):  # event is not part of a swap - save it and continue
+                processed_events.append(next(events_iterator))
+                continue
+
+            # Temporarily skip counterparties whose tests have not been updated.
+            # TODO: Update all decoding tests involving swaps and remove this counterparty check.
+            if next_event.counterparty not in {CPT_METAMASK_SWAPS, CPT_LLAMAZIP, CPT_PENDLE}:
+                processed_events.append(next(events_iterator))
+                continue
```
prettyirrelevant, line 593: also, this won't be needed. we'll only need to check that the swap has a spend/receive or spend/receive/fee

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -555,6 +560,78 @@ def run_all_post_decoding_rules(
...
+        in the correct spend/receive/fee order with no other events between them (although the
+        indexes do not need to be consecutive). If an incomplete or unordered group of Trade events
+        is encountered an error will be logged and the original EvmEvents saved to the db.
+
+        Returns the list of decoded events ordered by sequence index with any complete groups
+        of trade events replaced with EvmSwapEvents.
+        """
+        processed_events = []
+        trade_subtypes = (HistoryEventSubType.SPEND, HistoryEventSubType.RECEIVE, HistoryEventSubType.FEE)  # noqa: E501
+        events_iterator = peekable(iter(sorted(decoded_events, key=lambda x: x.sequence_index)))
```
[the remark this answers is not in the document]
nicholasyoder: Added a `process_swaps` attribute to DecodingOutput, and TransferEnrichmentOutput. The one problem with this approach is that we need to also process swaps created using post decoding rules. We do have several decoders where this is done (llamazip, rainbow, maybe others). 

Post decoding functions simply return the list of events, so we don't have a good way of passing this `process_swaps` flag. Instead I modified the `run_all_post_decoding_rules` to return a flag showing whether any rules were run. If this is True then we also process swaps, since a swap may have been created in the post decoding. 

This does mean that the swap processing will also run when post decoding is used for something other than swaps. but only some of the decoders use post decoding, so this is still much better than doing swap processing for everything as it was before.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -555,6 +560,78 @@ def run_all_post_decoding_rules(
...
+        while (next_event := events_iterator.peek(None)) is not None:
+            if (
+                next_event.event_type != HistoryEventType.TRADE or
+                next_event.event_subtype not in trade_subtypes
+            ):  # event is not part of a swap - save it and continue
+                processed_events.append(next(events_iterator))
+                continue
```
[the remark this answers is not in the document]
nicholasyoder: It still needs to process all the events in the transaction, so we do still need this. (to handle the gas fee event for instance)

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -555,6 +560,78 @@ def run_all_post_decoding_rules(
...
+                log.error(
+                    f'Encountered incomplete swap event group {spend_receive_fee} '
+                    f'in transaction {transaction!s}',
+                )
+                continue
+
+            for idx, trade_event in enumerate(trade_events):
+                swap_event = EvmSwapEvent(
+                    tx_hash=trade_event.tx_hash,
+                    sequence_index=spend_event.sequence_index + idx,  # Make indexes consecutive.
```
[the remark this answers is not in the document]
nicholasyoder: so, they will be in the right order already here, but they might not be consecutive (counting up one right after the other). For example the spend/receive/fee might have indexes of `1,5,7` - this will turn that into `1,2,3`. This is needed since we assume the sequence indexes will come one after the other in the logic for retrieving the receive and fee events when editing a swap event group via the api.

rotkehlchen/history/events/structures/evm_swap.py
```diff
@@ -99,7 +99,7 @@ def deserialize_from_db(cls: type['EvmSwapEvent'], entry: tuple) -> 'EvmSwapEven
     def serialize(self) -> dict[str, Any]:
         """Serialize the event for api."""
         serialized_data = EvmEvent.serialize(self)
-        serialized_data['auto_notes'] = self._generate_auto_notes()
+        serialized_data['auto_notes'] = self._generate_auto_notes(location=':counterparty:')  # replaced with counterparty label in frontend  # noqa: E501
```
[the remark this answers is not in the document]
nicholasyoder: yes, asked luki about it in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1362063190267007086)

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -547,13 +553,78 @@ def run_all_post_decoding_rules(
...
+        in the correct spend/receive/fee order with no other events between them (although the
+        indexes do not need to be consecutive). If an incomplete or unordered group of Trade events
+        is encountered an error will be logged and the original EvmEvents saved to the db.
+
+        Returns the list of decoded events ordered by sequence index with any complete groups
+        of trade events replaced with EvmSwapEvents.
+        """
+        processed_events = []
+        trade_subtypes = (HistoryEventSubType.SPEND, HistoryEventSubType.RECEIVE, HistoryEventSubType.FEE)  # noqa: E501
+        events_iterator = peekable(iter(sorted(decoded_events, key=lambda x: x.sequence_index)))
```
LefterisJP, line 580: they are not already sorted by sequence index here?

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -547,13 +553,78 @@ def run_all_post_decoding_rules(
...
+                continue
+
+            spend_receive_fee: list[EvmEvent | None] = [None, None, None]
+            for idx, subtype in enumerate(trade_subtypes):
+                if (
+                    (next_event := events_iterator.peek(None)) is not None and
+                    next_event.event_type == HistoryEventType.TRADE and
+                    next_event.event_subtype == subtype
+                ):
+                    spend_receive_fee[idx] = next(events_iterator)
```
LefterisJP, line 596: Why do you need both `spend_receive_fee` list and `trade_events` list? Isn't it doign the same thing and the second is just filtering out the `None`s? Can't you combine this step?

rotkehlchen/chain/evm/decoding/decoder.py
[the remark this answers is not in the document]
LefterisJP: Can you add a small addendum in the comment here as to why you are making them consecutive. You already explained it so add it as a comment.

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -650,13 +724,22 @@ def _decode_transaction(
...
+                    process_swaps = True
 
-        events = self.run_all_post_decoding_rules(
+        events, maybe_modified = self.run_all_post_decoding_rules(
             transaction=transaction,
             decoded_events=events,
             all_logs=tx_receipt.logs,
             counterparties=counterparties,
         )
+        # TODO: update tests for all decoders that process swaps via post decoding rules
```
LefterisJP, line 736: can you make a note for this and add it here?

rotkehlchen/chain/evm/decoding/structures.py
```diff
@@ -81,12 +81,14 @@ class DecodingOutput:
     - refresh_balances may be set to True if the user's on-chain balances in some protocols has
     changed (for example if the user has deposited / withdrawn funds from a curve gauge).
     - reload_decoders can be None in which case nothing happens. Or a set of decoders names for which to reload data. The decoder's name is the class name without the Decoder suffix. For example Eigenlayer for EigenlayerDecoder
+    - process_swaps indicates whether there are swaps that need converted into EvmSwapEvents.
```
LefterisJP, line 84:
```suggestion
    - process_swaps indicates whether there are swaps that need to be converted into EvmSwapEvents.
```

rotkehlchen/chain/evm/decoding/structures.py
```diff
@@ -97,9 +99,11 @@ class TransferEnrichmentOutput(NamedTuple):
     and is used in post-decoding rules like in the case of balancer.
     - refresh_balances may be set to True if the user's on-chain balances in some protocols has
     changed (for example if the user has deposited / withdrawn funds from a curve gauge).
+    - process_swaps indicates whether there are swaps that need converted into EvmSwapEvents.
```
LefterisJP, line 102:
```suggestion
    - process_swaps indicates whether there are swaps that need to be converted into EvmSwapEvents.
```

rotkehlchen/chain/evm/decoding/structures.py
```diff
@@ -81,12 +81,14 @@ class DecodingOutput:
...
     changed (for example if the user has deposited / withdrawn funds from a curve gauge).
     - reload_decoders can be None in which case nothing happens. Or a set of decoders names for which to reload data. The decoder's name is the class name without the Decoder suffix. For example Eigenlayer for EigenlayerDecoder
+    - process_swaps indicates whether there are swaps that need converted into EvmSwapEvents.
     """  # noqa: E501
     event: Optional['EvmEvent'] = None
     action_items: list[ActionItem] = field(default_factory=list)
     matched_counterparty: str | None = None
     refresh_balances: bool = False
     reload_decoders: set[str] | None = None
+    process_swaps: bool = False
```
LefterisJP, line 91: question: Is there any test we can have to make suire that for all decoders that create swaps this is set to True properly? I can easily see this missing in some decoder.

rotkehlchen/history/events/structures/swap.py
```diff
@@ -139,24 +139,25 @@ def deserialize_from_db(cls: type['SwapEvent'], entry: tuple) -> 'SwapEvent':
             notes=entry[8] or None,
         )
 
-    def _generate_auto_notes(self) -> str:
+    def _generate_auto_notes(self, location: str) -> str:
```
LefterisJP, line 142: you need to explain the new argument and what it does in the docstring and why nto always use self.location

rotkehlchen/history/events/structures/evm_swap.py
LefterisJP, line 102: I don't understand what you are doing here. This can't work. Frontend is just one place where notes are used, right? What about CSV export or PnL report?

rotkehlchen/chain/evm/decoding/decoder.py
nicholasyoder: No, they are just in whatever order the logic in the decoders has them in. We were sorting them at the very end of `_decode_transaction` - but I guess that actually makes it sort twice when it processes swaps - so I've moved that up to right before calling `_process_swaps`, and removed the sort from here, so it only does it at one place.

rotkehlchen/chain/evm/decoding/decoder.py
nicholasyoder: Reworked this logic a bit.

rotkehlchen/chain/evm/decoding/decoder.py
nicholasyoder: added it to [this note](https://github.com/orgs/rotki/projects/11?pane=issue&itemId=106071652) and mentioned it in the comment.

rotkehlchen/chain/evm/decoding/structures.py
nicholasyoder: mmm... yeah, good point, this could get missed rather easily. Unfortunately I don't think this is something we can check in a test very well - as far as I can tell any test that would want to check this would need to run all the decoding logic for all protocols and see if they created events with event_type of TRADE... which isn't really feasible.

I think the best thing we can do is put an assertion in `_decode_transaction` where if `process_swaps` is false, it does this:
```
assert len([e for e in events if e.event_type == HistoryEventType.TRADE]) == 0
```
If I recall correctly assertions are removed in production? so this would only run in the tests and when we run the app in develop. If you think this is a good idea I'll add it to [this note](https://github.com/orgs/rotki/projects/11?pane=issue&itemId=106071652) as something we need to add. (can't add it in this PR or all the tests we haven't updated yet would start failing).

rotkehlchen/history/events/structures/evm_swap.py
nicholasyoder: Just using the original notes from the decoder for EvmSwapEvents instead of this as discussed in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1362502580978319551).

LefterisJP (approved):
LGTM

PR #9806, opened 2025-04-16
Remove the Fee type

nicholasyoder:
Removes the Fee type as discussed in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1361748995315204329). Just using FVal now, like we did with the old AssetAmount


LefterisJP (approved):
LGTM

PR #9807, opened 2025-04-16
Sanitizing Rotki Backend Start-Up Arguments in Docker Entrypoint

edelbluth:
Python's `subprocess.Popen` accepts a list of strings and PathLikes as arguments to start a process. The `entrypoint.py` script reads environment variables and settings from a config file, casting numeric-expected values to `int` as a first line of defense against mis-configuration. The casted values are added to the arguments list for `subprocess.Popen`. When such values are configured, the startup fails because of an exception from `Popen`.

This fix introduces a quick sanitize of the arguments list, just for the use with `subprocess.Popen`, keeping the defense line intact.

For a later refactoring, a proper value validation should be considered. This could provide better / more helpful error messages in case of a misconfiguration.

## Checklist

### DIFF
--- a/packaging/docker/entrypoint.py
+++ b/packaging/docker/entrypoint.py
@@ -20,6 +20,20 @@ logging.basicConfig(level=logging.DEBUG)
 DEFAULT_LOG_LEVEL = 'critical'
 
 
+def sanitize_args_for_popen(args: list[Any]) -> list[str]:
+    """
+    Sanitizes the arguments for `subprocess.Popen` by converting them to strings.
+
+    Ratio: `subprocess.Popen` expects a list of strings and PathLikes as arguments.
+    Some configuration values may be of type `int` or other types. Casting them to
+    string ensures that `subprocess.Popen` can handle them correctly.
+
+    :param args: The list of arguments to sanitize.
+    :return: A sanitized list of arguments.
+    """
+    return [str(arg) for arg in args]
+
+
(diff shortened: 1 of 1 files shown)


packaging/docker/entrypoint.py
```diff
@@ -192,9 +206,11 @@ def load_config() -> tuple[list[str], str | None]:
     '0.0.0.0',
 ]
 
-cmd = base_args + config_args
+# Arguments needs to be sanitized for `subprocess.Popen`, as it does not accept `int`
+cmd = sanitize_args_for_popen(base_args + config_args)
```
LefterisJP, line 210: no need for the extra function
```suggestion
# Arguments needs to be sanitized for `subprocess.Popen`, as it does not accept `int`
cmd = [str(arg) for arg in (base_args + config_args)]
```

packaging/docker/entrypoint.py
```diff
@@ -192,9 +206,11 @@ def load_config() -> tuple[list[str], str | None]:
...
 
 logger.info('starting rotki backend')
+logger.debug('backend arguments: %s', cmd)
```
LefterisJP, line 213: Keep a single log entry for this

```suggestion
logger.info(f'starting rotki backend with args: {cmd}')
```

LefterisJP (commented):
hey thanks. Left a review

LefterisJP (approved):
LGTM. Thanks for finding this out! Really appreciate it.

PR #9810, opened 2025-04-17
exclude related events when any contains an ignored asset

prettyirrelevant:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=106711800

### DIFF
--- a/rotkehlchen/db/history_events.py
+++ b/rotkehlchen/db/history_events.py
@@ -684,6 +684,12 @@ class DBHistoryEvents:
         if entries_limit is None:
             return count_without_limit, count_without_limit
 
+        # When we have a limit but the total is already smaller or equal,
+        # just return the total for both counts
+        if count_without_limit <= entries_limit:
+            return count_without_limit, count_without_limit
+
+        # Otherwise, get the limited count
         free_query, free_bindings = self._create_history_events_query(
             has_premium=False,
             filter_query=query_filter,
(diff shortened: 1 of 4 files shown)


rotkehlchen/db/history_events.py
```diff
@@ -684,6 +684,12 @@ def get_history_events_count(
         if entries_limit is None:
             return count_without_limit, count_without_limit
 
+        # When we have a limit but the total is already smaller or equal,
+        # just return the total for both counts
+        if count_without_limit <= entries_limit:
+            return count_without_limit, count_without_limit
+
```
yabirgb, line 691: why this change? Could you point why this?

rotkehlchen/db/history_events.py
```diff
@@ -694,6 +700,12 @@ def get_history_events_count(
             f'SELECT COUNT(*) FROM ({free_query})',
             free_bindings,
         ).fetchone()[0]
+
+        # If we're grouping by event IDs and got 0 results but should have some,
+        # fall back to using the minimum of limit and total
+        if group_by_event_ids and count_with_limit == 0 and entries_limit > 0:
+            count_with_limit = min(entries_limit, count_without_limit)
+
         return count_without_limit, count_with_limit
```
yabirgb, line 709: same, I'm not sure why this

prettyirrelevant: also, the same test as above but this test case is `(1, True, 2, 1)`

this time we're fixing an issue where grouping by event identifiers with a limit was returning 0 results when it should return some. It provides a fallback that ensures we get the correct count even when the database query doesn't behave as expected.

rotkehlchen/db/history_events.py
prettyirrelevant: we have a test, `test_get_events()`
https://github.com/rotki/rotki/blob/8e61ba1b8028ed4e00c8b61990edb417a9b45329/rotkehlchen/tests/api/test_history_base_entry.py#L569-L586

it handles the test case: `(2, True, 2, 2)` when the total count is less than or equal to the limit. In this case, we should just return the total count for both values rather than running an unnecessary limited query

LefterisJP (approved):
LGTM

PR #9812, opened 2025-04-18
Update swaps in tests for decoders using post decoding

nicholasyoder:

* Updates tests to EvmSwapEvents for decoders using post decoding and removes the counterparty check limiting when to process swaps.
* Fixes a problem in the oneinch decoder where the spend event had a different counterparty (uniswap).
* Fixes a problem in the swap processing logic where if an incomplete group was encountered it wasn't breaking/continuing correctly which caused an IndexError at `spend_event = trade_events[0]`.
* Also removes some code in the decoders where they set the address, counterparty, etc on receive and fee events, which is no longer needed.


rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -596,15 +595,17 @@ def _process_swaps(
                 ):  # match events in the order defined in trade_subtypes
                     trade_events.append(next(events_iterator))
                 elif subtype != HistoryEventSubType.FEE:  # spend and receive are required, fee is optional  # noqa: E501
-                    processed_events.extend(trade_events)
-                    processed_events.append(next(events_iterator))
+                    processed_events.extend(trade_events if len(trade_events) > 0 else [next(events_iterator)])  # noqa: E501
```
LefterisJP, line 598: I don't understand what you are doing here. Getting here is an error always? From the comment this is not obvious

rotkehlchen/chain/evm/decoding/zerox/decoder.py
```diff
@@ -83,15 +83,11 @@ def _update_send_receive_fee_events(
             send_event.event_subtype = HistoryEventSubType.SPEND
             send_event.notes = f'Swap {send_event.amount} {send_event.asset.symbol_or_name()} via the 0x protocol'  # noqa: E501
         if receive_event is not None:
-            receive_event.counterparty = CPT_ZEROX
```
LefterisJP, line 86: Hmm didn't we say that you should not remove those? Why removing them from here.  It's the only place you do.

LefterisJP:
Also in general this way of combining events is not great. The goal both for this, and asset movements and other stuff should be to somehow think to combine them, perhaps even in the DB, or if not in the DB at least in memory when consumign them. To avoid this duplication. This needs planning and thinking. (and I don't mean by me -- you guys should be capable of doing so too).

rotkehlchen/chain/evm/decoding/decoder.py
nicholasyoder: yes, getting here is an error. If either the spend or the receive do not match, then it means the group is incomplete or out of order. Since fee is optional, its fine if it doesn't match, so we do `elif subtype != HistoryEventSubType.FEE`. The change here is so we don't do `next(events_iterator)` unless we're on the first event in the group.

Added more comments to the code to better explain what's happening here.

LefterisJP (approved):
LGTM

PR #9813, opened 2025-04-18
Use historical events to detect NFT ownership instead of balanceOf calls

prettyirrelevant:


rotkehlchen/balances/historical.py
```diff
@@ -74,6 +74,32 @@ def get_balances(self, timestamp: Timestamp) -> dict[Asset, HistoricalBalance]:
...
+            address=address,
+        )
+
+        if len(events) == 0:
+            raise NotFoundError(f'No historical data found for {assets} for user address {address}')  # noqa: E501
+
+        current_balances: dict[Asset, FVal] = defaultdict(FVal)
+        for event in events:
+            if self._update_balances(event=event, current_balances=current_balances) is not None:
+                break
```
nicholasyoder, line 99: This needs a comment to make it clear that `_update_balances` returning something other than `None` and it breaking here is when there is an error (negative balance). I thought this break shouldn't be here at first as I assumed `_update_balances` returned None on error rather than success.

prettyirrelevant: i think you're in fact right that `break` should not be present. since this is used for multiple unrelated assets, we want to gather every possible event and at the end filter the negative balances out.

rotkehlchen/chain/evm/tokens.py
```diff
@@ -285,14 +292,84 @@ def _compute_detected_tokens_info(self, addresses: Sequence[ChecksumEvmAddress])
...
+                ):
+                    continue
+
+                calls.append((
+                    collection_address,
+                    erc721_contract.encode('ownerOf', arguments=[int(collectible_id)]),
+                ))
+
+            results = self.evm_inquirer.multicall(calls=calls)
+            for token, result in zip(list(token_balances), results, strict=False):
```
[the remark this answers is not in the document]
prettyirrelevant: oh i see. i changed the function, `get_address_balances()` to return only > 0 balances. 
also, i found an issue with `zip()` when there's a mismatch in the length of `tokens` and `results`.

rotkehlchen/chain/evm/tokens.py
```diff
@@ -285,14 +292,84 @@ def _compute_detected_tokens_info(self, addresses: Sequence[ChecksumEvmAddress])
...
+            valid_tokens, calls = [], []
+            for token, balance in token_balances.items():
+                if not (
+                        balance > ZERO and
+                        (collectible_id := tokenid_to_collectible_id(token.identifier)) is not None and  # noqa: E501
+                        (collection_address := evm_identifier_to_address(token.identifier)) is not None  # noqa: E501
+                ):
+                    continue
+
+                calls.append((
```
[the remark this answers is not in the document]
prettyirrelevant: i could not use a comprehension because i need to populate another list but i removed the **not**

rotkehlchen/balances/historical.py
```diff
@@ -74,6 +74,30 @@ def get_balances(self, timestamp: Timestamp) -> dict[Asset, HistoricalBalance]:
 
         return result
 
+    def get_address_balances(
+            self,
+            address: str,
```
LefterisJP, line 79: Why str?

```suggestion
            address: ChecksumEvmAddress,
```

rotkehlchen/balances/historical.py
```diff
@@ -74,6 +74,30 @@ def get_balances(self, timestamp: Timestamp) -> dict[Asset, HistoricalBalance]:
...
+        events, _ = self._get_events_and_currency(
+            assets=assets,
+            address=address,
+        )
+        if len(events) == 0:
+            raise NotFoundError(f'No historical data found for {assets} for user address {address}')  # noqa: E501
+
+        current_balances: dict[Asset, FVal] = defaultdict(FVal)
+        for event in events:
+            self._update_balances(event=event, current_balances=current_balances)
```
LefterisJP, line 97: can negative balance occur at any point? And how would it be handled?

rotkehlchen/balances/historical.py
```diff
@@ -218,6 +242,7 @@ def _get_events_and_currency(
             from_ts: Timestamp | None = None,
             to_ts: Timestamp | None = None,
             assets: tuple[Asset, ...] | None = None,
+            address: str | None = None,
```
LefterisJP, line 245: same here

rotkehlchen/balances/historical.py
```diff
@@ -74,6 +74,30 @@ def get_balances(self, timestamp: Timestamp) -> dict[Asset, HistoricalBalance]:
 
         return result
 
+    def get_address_balances(
+            self,
+            address: str,
+            assets: tuple[Asset, ...],
+    ) -> dict[Asset, FVal]:
+        """Get current balances for the given assets of a specific address.
```
LefterisJP, line 82: explain that it's from processing historical events.

rotkehlchen/globaldb/handler.py
```diff
@@ -871,33 +871,37 @@ def get_token_detection_data(
             chain_id: ChainID,
             exceptions: set[ChecksumEvmAddress],
             protocol: str | None = None,
-    ) -> list[EvmTokenDetectionData]:
+    ) -> tuple[list[EvmTokenDetectionData], list[EvmTokenDetectionData]]:
```
LefterisJP, line 874: Please whenever you make such changes explain in the docstring what the function is doing. You need to update it.

rotkehlchen/globaldb/handler.py
```diff
@@ -871,33 +871,37 @@ def get_token_detection_data(
...
                     continue
 
-                result.append(EvmTokenDetectionData(
+                details = EvmTokenDetectionData(
                     identifier=identifier,
                     address=address,
                     decimals=decimals if decimals is not None else DEFAULT_TOKEN_DECIMALS,  # TODO: at least two tokens are missing the decimals in my DB and also the EvmToken class allows decimals to be None. We need to think if that is correct and if we should enforce or not for all the erc20s to have decimals.  # noqa: E501
-                ))
+                )
+                if EvmTokenKind.deserialize_from_db(token_kind) == EvmTokenKind.ERC721:
```
LefterisJP, line 899: Don't deserialize it from DB always. Since it's just two get the deserialized value in the start out of the loop and then compare

rotkehlchen/constants/resolver.py
```diff
@@ -49,3 +49,14 @@ def strethaddress_to_identifier(address: str) -> str:
         chain_id=ChainID.ETHEREUM,
         token_type=EvmTokenKind.ERC20,
     )
+
+
+def evm_identifier_to_address(identifier: str) -> ChecksumEvmAddress | None:
```
LefterisJP, line 54: you don't need that

rotkehlchen/chain/evm/tokens.py
```diff
@@ -285,14 +292,81 @@ def _compute_detected_tokens_info(self, addresses: Sequence[ChecksumEvmAddress])
...
+                )
+            except (NotFoundError, DeserializationError) as e:
+                log.error(f'Failed to get erc721 token balances for {address} due to {e}. Skipping.')  # noqa: E501
+                continue
+
+            filtered_tokens, calls = [], []
+            for token in token_balances:
+                if (
+                    (collectible_id := tokenid_to_collectible_id(token.identifier)) is not None and
+                    (collection_address := evm_identifier_to_address(token.identifier)) is not None
```
LefterisJP, line 339: you don't need that. You can get the token address from the token itself.

LefterisJP (commented):
Lots of comments. Also this needs a changelog as it's fixing a bug. Something like: "rotki will no longer count all previously owned NFTs of a collection as balance owned"

rotkehlchen/balances/historical.py
```diff
@@ -74,6 +75,30 @@ def get_balances(self, timestamp: Timestamp) -> dict[Asset, HistoricalBalance]:
...
+            address=address,
+        )
+        if len(events) == 0:
+            raise NotFoundError(f'No historical data found for {assets} for user address {address}')  # noqa: E501
+
+        current_balances: dict[Asset, FVal] = defaultdict(FVal)
+        for event in events:
+            self._update_balances(event=event, current_balances=current_balances)
+
+        return {asset: balance for asset, balance in current_balances.items() if balance > ZERO}
```
LefterisJP, line 100: Since this is always ERC721 tokens please do the resolution here, and adjust the types.

Then you don't need to resolve EVM token in that loop in `_detect_erc721_tokens`. And this way you won't need to worry what if resolution fails, as it will be explicit it can't fail.

LefterisJP (approved):
LGTM

PR #9815, opened 2025-04-19
Optimize query of history events

yabirgb:
We used `get_blockchain_accounts` to retrieve ethereum accounts that were not always used, only when `ETH_BLOCK_EVENT` needed to be processed.

This function is really bad because it queries ALL the accounts known and then filters them IN PYTHON to get the ones that are in Ethereum. AND THIS FOR ALL the queries of history events, even if no block event was getting processed.

This change in the logics queries it only once if there is a block event and avoids querying them in the rest of cases. ALso the query is optimized to do the filtering at the sql level.


LefterisJP (approved):
LGTM

PR #9818, opened 2025-04-21
Update more decoders/tests for evm swaps

nicholasyoder:

### DIFF
--- a/rotkehlchen/tests/unit/decoders/test_basenames.py
+++ b/rotkehlchen/tests/unit/decoders/test_basenames.py
@@ -9,13 +9,13 @@ from rotkehlchen.chain.base.modules.basenames.constants import (
     BASENAMES_REGISTRY,
     CPT_BASENAMES,
 )
-from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
 from rotkehlchen.chain.evm.decoding.constants import CPT_GAS
 from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.constants.assets import A_ETH
 from rotkehlchen.constants.misc import ONE, ZERO
 from rotkehlchen.fval import FVal
 from rotkehlchen.history.events.structures.evm_event import EvmEvent
+from rotkehlchen.history.events.structures.evm_swap import EvmSwapEvent
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.tests.utils.ethereum import get_decoded_events_of_transaction
 from rotkehlchen.types import Location, Timestamp, TimestampMS, deserialize_evm_tx_hash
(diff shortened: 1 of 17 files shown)


rotkehlchen/tests/unit/decoders/test_basenames.py
```diff
@@ -126,19 +126,18 @@ def test_basenames_register(
...
             tx_hash=tx_hash,
             sequence_index=192,
             timestamp=timestamp,
             location=Location.BASE,
-            event_type=HistoryEventType.TRADE,
             event_subtype=HistoryEventSubType.RECEIVE,
             asset=Asset(f'eip155:8453/erc721:0x03c4738Ee98aE44591e1A4A4F3CaB6641d95DD9a/{token_id}'),
             amount=ONE,
             location_label=user_address,
             notes=f'Receive Basenames name yabir.base.eth from {ZERO_ADDRESS} to {user_address}',
```
prettyirrelevant, line 138: there's a mismatch here between the address & the notes

rotkehlchen/tests/unit/decoders/test_curve.py
```diff
@@ -1191,25 +1186,23 @@ def test_curve_swap_router(ethereum_transaction_decoder, ethereum_accounts):
...
             sequence_index=1,
             timestamp=TimestampMS(1679550275000),
             location=Location.ETHEREUM,
-            event_type=HistoryEventType.TRADE,
             event_subtype=HistoryEventSubType.SPEND,
             asset=Asset('ETH'),
             amount=FVal('40'),
             location_label=user_address,
             notes='Swap 40 ETH in curve',
             counterparty='curve',
```
prettyirrelevant, line 1199: switch this to use the `CPT_xxx` and everywhere else.

rotkehlchen/tests/unit/decoders/test_basenames.py
nicholasyoder: The address has to be the same on all events in a swap event group, so we can't have the address on the receive be the zero address. I think its fine to have it say zero address in the note but have the address of the events in the swap group be the `BASENAMES_REGISTRAR_CONTROLLER`. The registrar controller is minting the name nft here, so yes its from the zero address, but the event is also associated with the registrar controller.

LefterisJP: It's never fine to say from ZERO address for something that's actually processed properly by us as a protocol. A transfer from zero is minting. A transfer to zero is burning. But if you have a normal adddress then use that too.

nicholasyoder: Updated the basenames decoder to use the registrar address there instead (keep it consistent with how its done for mainnet ENS)

PR #9829, opened 2025-04-23
account for umami dynamic deposit and withdrawal fee in decoding

prettyirrelevant:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=107503180

### DIFF
--- a/rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
+++ b/rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
@@ -15,6 +15,7 @@ from rotkehlchen.chain.evm.decoding.structures import (
 )
 from rotkehlchen.chain.evm.decoding.types import CounterpartyDetails
 from rotkehlchen.chain.evm.decoding.utils import maybe_reshuffle_events
+from rotkehlchen.fval import FVal
 from rotkehlchen.globaldb.handler import GlobalDBHandler
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.types import ChecksumEvmAddress
@@ -22,7 +23,6 @@ from rotkehlchen.types import ChecksumEvmAddress
 if TYPE_CHECKING:
     from rotkehlchen.chain.arbitrum_one.node_inquirer import ArbitrumOneInquirer
     from rotkehlchen.chain.evm.decoding.base import BaseDecoderTools
-    from rotkehlchen.fval import FVal
     from rotkehlchen.history.events.structures.evm_event import EvmEvent
     from rotkehlchen.user_messages import MessagesAggregator
 
@@ -33,6 +33,7 @@ STAKE_TOPIC: Final = b'\x90\x89\x08\t\xc6T\xf1\x1dnr\xa2\x8f\xa6\x01Iw\n\r\x11\x
 UNSTAKE_TOPIC: Final = b'\xf2y\xe6\xa1\xf5\xe3 \xcc\xa9\x115gm\x9c\xb6\xe4L\xa8\xa0\x8c\x0b\x884+\xcd\xb1\x14Oe\x11\xb5h'  # noqa: E501
 REWARD_TOPIC: Final = b'q\xba\xb6\\\xed.WPwZ\x06\x13\xbe\x06}\xf4\x8e\xf0l\xf9*In\xbfvc\xae\x06`\x92IT'  # noqa: E501
 DEPOSIT_EXECUTION_FOUR_BYTES: Final = b'\xdb\x10\xc3\xb9'
+UMAMI_DEPOSIT_WITHDRAWAL_FEE_PERCENTAGE: Final = FVal('0.0015')  # this is an estimate since the actual percentage is dynamic -- 0.15%  # noqa: E501
 
 
 class FoundEventType(Enum):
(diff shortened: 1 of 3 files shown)


rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -85,21 +92,35 @@ def _decode_deposit_withdraw_request(
...
+        Returns a tuple of the protocol fee event and the event type found for reshuffling later.
         """
-        resolved_asset = event.asset.resolve_to_crypto_asset()
+        deposit_protocol_fee_event, deposit_fee, resolved_asset = None, ZERO, event.asset.resolve_to_crypto_asset()  # noqa: E501
         match_amount = asset_normalized_value(
             amount=int.from_bytes(context.tx_log.data[0:32]),
             asset=resolved_asset,
         )
-        return self._decode_deposit_withdraw_request(
+        if event.asset != self.evm_inquirer.native_token:  # execution fee is paid in native token
```
nicholasyoder, line 104: This is confusing. Am I understanding correctly that this logic is run both for the main deposit event and the execution fee event and this is to make it only add the protocol fee when its processing the main event? This is not clear from the code or comments - I'm not totally sure if I'm understanding correctly myself.

Would there be something else to check here other than just the asset?
Also, maybe the creation of the `deposit_protocol_fee_event` should be in `_decode_umami_vault_events` where the withdrawal one is? Having them at different locations makes it confusing and unless I'm not missing something, there isn't really any reason it needs to be done here in `_decode_deposit_request`

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -157,15 +181,33 @@ def _decode_umami_vault_events(self, context: DecoderContext) -> DecodingOutput:
...
                 event.event_subtype = HistoryEventSubType.REDEEM_WRAPPED
-                event.notes = f'Withdraw {event.amount} {event.asset.resolve_to_asset_with_symbol().symbol} from Umami'  # noqa: E501
+                event.notes = f'Withdraw {event.amount} {token.symbol} from Umami'
                 event.counterparty = CPT_UMAMI
                 main_event = event
 
         if main_event is not None:
+            ordered_events = [main_event, fee_event]
+            if withdrawal_protocol_fee_event is not None or deposit_protocol_fee_event is not None:
+                ordered_events.insert(1, (protocol_fee_event := withdrawal_protocol_fee_event or deposit_protocol_fee_event))  # noqa: E501
```
nicholasyoder, line 206: I think you can just do `(protocol_fee_event := withdrawal_protocol_fee_event or deposit_protocol_fee_event) is not None` in the if.

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -136,15 +142,35 @@ def _decode_umami_vault_events(self, context: DecoderContext) -> DecodingOutput:
...
             if (
                 event.event_type == HistoryEventType.SPEND and
                 event.event_subtype == HistoryEventSubType.NONE
             ):  # deposit or withdraw request
                 type_found = FoundEventType.NONE
                 if context.tx_log.topics[0] == DEPOSIT_TOPIC:
-                    type_found = self._decode_deposit_request(context=context, event=event)
+                    # Skip protocol fees for native tokens like ETH, which are fees being paid.
+                    # We only want to apply protocol fees to the actual tokens being deposited.
+                    if event.asset != self.evm_inquirer.native_token:
```
nicholasyoder, line 155: Couldn't you just check the event type for `DEPOSIT_FOR_WRAPPED` events here instead of checking the asset? Checking the asset seems a bit counter intuitive.

nicholasyoder (commented):
Also this should probably have a changelog entry.

rotkehlchen/chain/arbitrum_one/modules/umami/decoder.py
```diff
@@ -157,15 +163,33 @@ def _decode_umami_vault_events(self, context: DecoderContext) -> DecodingOutput:
...
+                )
                 event.event_type = HistoryEventType.WITHDRAWAL
+                event.amount += withdrawal_fee
                 event.event_subtype = HistoryEventSubType.REDEEM_WRAPPED
-                event.notes = f'Withdraw {event.amount} {event.asset.resolve_to_asset_with_symbol().symbol} from Umami'  # noqa: E501
+                event.notes = f'Withdraw {event.amount} {token.symbol} from Umami'
                 event.counterparty = CPT_UMAMI
                 main_event = event
 
         if main_event is not None:
```
nicholasyoder, line 185: The changes in this PR seem to be making the code here very long and hard to see what's happening. Removing `_decode_deposit_request` and `_decode_withdraw_request` especially make `_decode_umami_vault_events` long, and also the code creating the two fee events is very similar.

I've looked at this a bit and I think there's actually a much simpler approach. Basically revert the code back to what its like in develop currently, and then add a bit of logic right here in `if main_event is not None:` to create the protocol fee event and modify the main_event's amount if the main_event is a `DEPOSIT`/`DEPOSIT_FOR_WRAPPED` or `WITHDRAWAL`/`REDEEM_WRAPPED`

This allows the rest of the decoder to remain as is, keeping the code related to this extra fee all in one spot. I think the flow here will be much easier to follow like this. Or is there something I'm not seeing that would prevent this from working?

LefterisJP (approved):
LGTM

PR #9830, opened 2025-04-23
Add multi trades and event grouping

nicholasyoder:


rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -1124,3 +1126,123 @@ def test_add_edit_evm_swap_events(rotkehlchen_api_server: 'APIServer') -> None:
         'product': 'pool',
         'address': '0xb5d85CBf7cB3EE0D56b3bB207D5Fc4B82f43F511',
     }
+
+
+def test_multi_grouping(rotkehlchen_api_server: 'APIServer') -> None:
+    """Test that events are properly grouped into sub-lists
+    when they are serialized for the api.
+    """
```
yabirgb, line 1134: can you write in the docstring a summary of what are the events? Otherwise I need to go and read them and match them to understand what is a swap and so on.

rotkehlchen/api/rest.py
```diff
@@ -3581,6 +3582,59 @@ def query_online_events(self, query_type: HistoryEventQueryType) -> dict[str, An
...
+            if (
+                event.event_type == HistoryEventType.MULTI or
+                event.entry_type == HistoryBaseEntryType.EVM_SWAP_EVENT
+            ):
+                if (  # Append to current group as long as events are in the order spends/receives/fees  # noqa: E501
+                    len(entries) != 0 and isinstance(entries[-1], list) and (
+                        (current_subtype == HistoryEventSubType.SPEND and event.event_subtype in (HistoryEventSubType.SPEND, HistoryEventSubType.RECEIVE)) or  # noqa: E501
+                        (current_subtype == HistoryEventSubType.RECEIVE and event.event_subtype in (HistoryEventSubType.RECEIVE, HistoryEventSubType.FEE)) or  # noqa: E501
+                        (current_subtype == HistoryEventSubType.FEE and event.event_subtype == HistoryEventSubType.FEE)  # noqa: E501
+                )):
```
yabirgb, line 3626: this will be harder to abstract and is tied to the function. Can you use this one instead https://gist.github.com/yabirgb/614a16a4a6af33ed1bca86b57bbdf57d and also move the order to a constant probably near to where the event types? I believe you can also prepare it and have the mapping something like

```python
{
    HistoryEventSubType.SPEND: {
            HistoryEventSubType.SPEND: 0,
            HistoryEventSubType.RECEIVE: 1,
            HistoryEventSubType.FEE: 2,
        }
}
```

so it can later be expanded to

```python
{
    HistoryEventSubType.SPEND: {
            HistoryEventSubType.SPEND: 0,
            HistoryEventSubType.RECEIVE: 1,
            HistoryEventSubType.FEE: 2,
        },
   HistoryEventSubType.DEPOSIT: {
           DEPOSIT: 0
           WITHDRAWAL: 1
   } 
}
```

In addition you have to be careful and instead of `event_subtype_index = expected_subtype_order[event.event_subtype]` do an `if + := + .get` to catch possible errors in the definition of the event

yabirgb (commented):
This requires still an adjustment in the api.rst file to state what can be the output. Also you should create a note for:

- Adjusting the logic in the accounting part to manage the MULTI events
- Adjusting the docs so the users know how to use this and what it means

rotkehlchen/api/rest.py
```diff
@@ -3581,6 +3582,59 @@ def query_online_events(self, query_type: HistoryEventQueryType) -> dict[str, An
...
+
+        Returns a list of serialized events with grouped events in sub-lists.
+        """
+        entries: list[dict[str, Any] | list[dict[str, Any]]] = []
+        current_subtype = None
+        for (grouped_events_num, event), event_accounting_rule_status in zip(
+            events_result,
+            event_accounting_rule_statuses,
+            strict=True,
+        ):
```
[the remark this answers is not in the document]
nicholasyoder: I reworked this a bit so that now instead of this `events_result` thing, `events`, `event_accounting_rule_statuses`, and `grouped_events_nums` will all be passed as separate lists and then zipped. Also added to the docstring to explain what the arguments are.

rotkehlchen/api/rest.py
nicholasyoder: Changed this as suggested. It got a bit more complicated than I was expecting since the grouping order is categorized by the subtype of the first event in the group, so now it also have to keep track of the `current_group_type`.

nicholasyoder:
> requires still an adjustment in the api.rst

I ended up spending enough time working or the other changes suggested here that I think I'll do this later in another PR. 

Will also create the notes as mentioned

rotkehlchen/api/rest.py
```diff
@@ -3581,6 +3583,88 @@ def query_online_events(self, query_type: HistoryEventQueryType) -> dict[str, An
...
+        """
+        entries: list[dict[str, Any] | list[dict[str, Any]]] = []
+        current_group: list[dict[str, Any]] = []
+        current_group_type = None
+        last_subtype_index: int | None = None
+        for event, event_accounting_rule_status, grouped_events_num in zip(
+            events,
+            event_accounting_rule_statuses,
+            grouped_events_nums,
+            strict=True,
```
LefterisJP, line 3620: you are using `strict = true` which means this can raise a value error if at any point the iterables are seen to not be of equal length. i don't see that mentioned in the docstring or handled anywhere.

rotkehlchen/accounting/constants.py
```diff
@@ -126,6 +126,19 @@
     HistoryEventType.BURN: {
         HistoryEventSubType.NFT: {DEFAULT: EventCategory.BURN_NFT},
     },
+    HistoryEventType.MULTI: {
```
LefterisJP, line 129: We had a convo about the naming here: https://discord.com/channels/657906918408585217/1364225036621316096/1364693201688203316

Probably best to add more subtypes to keep the hierarchy of main events clear.

rotkehlchen/accounting/constants.py
```diff
@@ -126,6 +126,19 @@
...
+        HistoryEventSubType.SPEND: {DEFAULT: EventCategory.SWAP_OUT},
+        HistoryEventSubType.RECEIVE: {DEFAULT: EventCategory.SWAP_IN},
+        HistoryEventSubType.FEE: {DEFAULT: EventCategory.FEE},
+    },
+}
+
+EVENT_GROUPING_ORDER = {
```
LefterisJP, line 136: this will need to change I guess with the type changes but also explain in a short comment here where this is used

rotkehlchen/api/rest.py
```diff
@@ -3581,6 +3583,88 @@ def query_online_events(self, query_type: HistoryEventQueryType) -> dict[str, An
...
+                ignored_ids=ignored_ids,
+                hidden_event_ids=hidden_event_ids,
+                event_accounting_rule_status=event_accounting_rule_status,
+                grouped_events_num=grouped_events_num,
+            )
+            if (
+                event.event_type == HistoryEventType.MULTI or
+                event.entry_type == HistoryBaseEntryType.EVM_SWAP_EVENT
+            ):
+                if current_group_type is None:
```
LefterisJP, line 3633: The logic here is kind of convoluted and will probably also need to be adjusted after the changes.

rotkehlchen/tests/api/test_history_base_entry.py
```diff
@@ -1124,3 +1126,129 @@ def test_add_edit_evm_swap_events(rotkehlchen_api_server: 'APIServer') -> None:
         'product': 'pool',
         'address': '0xb5d85CBf7cB3EE0D56b3bB207D5Fc4B82f43F511',
     }
+
+
+def test_multi_grouping(rotkehlchen_api_server: 'APIServer') -> None:
+    """Test that events are properly grouped into sub-lists
```
LefterisJP, line 1132: adjust test, test name, docstring after your changes

rotkehlchen/api/rest.py
nicholasyoder: The iterables are guaranteed to have the same length here. `events` and `grouped_events_nums` are created together from `events_result`, and `event_accounting_rule_statuses` is created with `query_missing_accounting_rules` which explicitly states in its docstring `For a list of events returns a list of the same length ...`

Added a comment explaining that they will all be the same length.

rotkehlchen/api/rest.py
nicholasyoder: Removed `current_group_type` since EVENT_GROUPING_ORDER is organized by event_type now.

rotkehlchen/api/rest.py
```diff
@@ -3581,6 +3583,77 @@ def query_online_events(self, query_type: HistoryEventQueryType) -> dict[str, An
...
+                event_accounting_rule_status=event_accounting_rule_status,
+                grouped_events_num=grouped_events_num,
+            )
+            if (
+                event.event_type == HistoryEventType.MULTI_TRADE or
+                event.entry_type == HistoryBaseEntryType.EVM_SWAP_EVENT
+            ):
+                if (event_subtype_index := EVENT_GROUPING_ORDER[event.event_type].get(event.event_subtype)) is None:  # noqa: E501
+                    log.error(
+                        f'Unable to determine group order for event type/subtype '
```
yabirgb, line 3630: the f string is not needed here

rotkehlchen/api/rest.py
```diff
@@ -3581,6 +3583,77 @@ def query_online_events(self, query_type: HistoryEventQueryType) -> dict[str, An
...
+        Returns a list of serialized events with grouped events in sub-lists.
+        """
+        entries: list[dict[str, Any] | list[dict[str, Any]]] = []
+        current_group: list[dict[str, Any]] = []
+        last_subtype_index: int | None = None
+        for event, event_accounting_rule_status, grouped_events_num in zip(
+            events,
+            event_accounting_rule_statuses,
+            grouped_events_nums,
+            strict=True,  # guaranteed to have same length. event_accounting_rule_statuses and grouped_events_nums are created directly from the events list.  # noqa: E501
```
LefterisJP, line 3615: Then remove the strict=True. There is no reason to have it. Keep the comment somewhere there though to explain why the strict is missing. I guess the only reason to have strict=True is to check if ever that is not right. And that check is additional overhead that you are not handling here. So there is no reason for it to exist.

rotkehlchen/api/rest.py
LefterisJP: + 1

rotkehlchen/api/rest.py
```diff
@@ -3619,41 +3692,28 @@ def get_history_events(
...
-                    customized_event_ids=customized_event_ids,
-                    ignored_ids=ignored_ids,
-                    hidden_event_ids=hidden_event_ids,
-                    event_accounting_rule_status=event_accounting_rule_status,
-                ) for x, event_accounting_rule_status in zip(events_result, event_accounting_rule_statuses, strict=True)  # noqa: E501
-            ]
+        events: list[HistoryBaseEntry]
+        grouped_events_nums: list[int | None]
+        grouped_events_nums, events = (
+            zip(*events_result, strict=True)  # type: ignore  # mypy doesn't understand significance of boolean check
```
LefterisJP, line 3698: same question about the strict true here. If you have that check you should be checking for the ValueError and handling it. if it's guaranteed to be of same length then you should just explain it. Essentially explaining why you don't need the strict=True check

rotkehlchen/chain/evm/decoding/odos/common.py
```diff
@@ -115,7 +115,7 @@ def _calculate_router_fee(
                 router_fees.append(self.base.make_event_next_index(
                     tx_hash=context.transaction.tx_hash,
                     timestamp=context.transaction.timestamp,
-                    event_type=HistoryEventType.SPEND,
+                    event_type=HistoryEventType.TRADE,
```
LefterisJP, line 118: shouldn't this be multitrade fee or did I misunderstand? Or that's set properly later at the if len check below? I guess the latter ... but just checking

LefterisJP (commented):
Small comment

rotkehlchen/api/rest.py
nicholasyoder: Changed strict to False. If strict is actually removed ruff does this:
```
B905 [*] `zip()` without an explicit `strict=` parameter
```

rotkehlchen/api/rest.py
nicholasyoder: In this case its actually unzipping `events_result` - so there's no way for there to be unequal lengths here. But have to specify strict due to ruff. Changed it to `strict=False`

rotkehlchen/chain/evm/decoding/odos/common.py
nicholasyoder: > that's set properly later at the if len check below
yep it gets changed to MULTI_TRADE later. It needs to be TRADE here since this same logic does produce normal trades some of the time.

LefterisJP (approved):
lgtm

PR #9832, opened 2025-04-23
ignore constant inflow & outflow tokens in optimism

prettyirrelevant:
these tokens return a hardcoded balance of 1 for all addresses, which can appear after interacting with superfluid protocol. excluding them prevents confusion in balance summaries.

https://explorer.superfluid.org/base-mainnet/supertokens -> [Select any token] -> `CONSTANT_INFLOW_NFT` & `CONSTANT_OUTFLOW_NFT` methods

### DIFF
--- a/rotkehlchen/chain/optimism/tokens.py
+++ b/rotkehlchen/chain/optimism/tokens.py
@@ -3,23 +3,20 @@ from typing import TYPE_CHECKING
 from rotkehlchen.chain.evm.tokens import EvmTokens
 from rotkehlchen.chain.optimism.modules.giveth.constants import GIVPOW_ADDRESS
 from rotkehlchen.constants.assets import A_OPTIMISM_ETH
-from rotkehlchen.types import ChecksumEvmAddress
 
 if TYPE_CHECKING:
-    from rotkehlchen.db.dbhandler import DBHandler
-
-    from .node_inquirer import OptimismInquirer
+    from rotkehlchen.types import ChecksumEvmAddress
 
 
 class OptimismTokens(EvmTokens):
 
-    def __init__(self, database: 'DBHandler', optimism_inquirer: 'OptimismInquirer') -> None:
-        super().__init__(database=database, evm_inquirer=optimism_inquirer)
-
     # -- methods that need to be implemented per chain
-    def _per_chain_token_exceptions(self) -> set[ChecksumEvmAddress]:
+    def _per_chain_token_exceptions(self) -> set['ChecksumEvmAddress']:
         """
         Optimism ETH ERC20 token mirrors the user's balance on the chain.
         To avoid double counting, we exclude the token from the balance query.
         """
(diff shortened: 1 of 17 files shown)


rotkehlchen/chain/optimism/tokens.py
```diff
@@ -5,6 +5,8 @@
 from rotkehlchen.constants.assets import A_OPTIMISM_ETH
 from rotkehlchen.types import ChecksumEvmAddress
 
+from ..evm.types import string_to_evm_address
```
yabirgb, line 8: can you use the full path instead of the relative one?

```
from rotkehlchen.chain.evm.types import string_to_evm_address
```

rotkehlchen/chain/optimism/tokens.py
```diff
@@ -22,4 +24,9 @@ def _per_chain_token_exceptions(self) -> set[ChecksumEvmAddress]:
         Optimism ETH ERC20 token mirrors the user's balance on the chain.
         To avoid double counting, we exclude the token from the balance query.
         """
-        return {A_OPTIMISM_ETH.resolve_to_evm_token().evm_address, GIVPOW_ADDRESS}
+        return {
+            A_OPTIMISM_ETH.resolve_to_evm_token().evm_address,
+            GIVPOW_ADDRESS,
+            string_to_evm_address('0x0C6D90a98426bfD572a5c5Be572a7f6Bd1C5ED76'),  # Constant Inflow Token  # noqa: E501
+            string_to_evm_address('0xFb2b126660BE2fdEBa254b1F6e4348644E8482e7'),  # Constant Outflow Token  # noqa: E501
+        }
```
yabirgb, line 32: seems that there is only one per chain of those https://docs.superfluid.org/docs/technical-reference/architecture#overview but I'm not sure 100%. Maybe you can dig further? A search on google points that there might be more

image

but I didn't see even the op one listed on https://explorer.superfluid.org/optimism-mainnet/protocol . If you don't know we have a telegram chat and we can ping them

prettyirrelevant: > but I didn't see even the op one listed on https://explorer.superfluid.org/optimism-mainnet/protocol . If you don't know we have a telegram chat and we can ping them

thanks, the link was very helpful as you just need to select one of the tokens per chain and look for `CONSTANT_INFLOW_NFT` & `CONSTANT_OUTFLOW_NFT` methods

LefterisJP (commented):
Two things:

1. Can you maybe add a test for this somehow?
2. I was thinking that since this is in every chain we support it could somehow be abstracted better, but since the addresses are different, perhaps it makes little sense? Would need it set to constructor and then having it in superclass `_per_chain_token_exceptions` and have super called from all subclasses and combined with them if there is more to add? Would that make sense?

LefterisJP (approved):
Small comment but will merge.

PR #9834, opened 2025-04-24
Improve bitfinex asset handling

nicholasyoder:
* Checks if assets already have a mapping before showing the new asset message
* Adds some assets to BITFINEX_EXCHANGE_TEST_ASSETS


yabirgb (approved):
lgtm

PR #9840, opened 2025-04-25
style: swap events animation and layout

lukicenturi:
Several things to do here:
- [x] Fix the thing here, if the swaps events are trimmed with this "Load more events" thing. The swaps group here should be treated as single event in the calculation.

Here if you click more, it will be correct

image

image

- [x] Adjust the notes, if it has multiple in, out and fee assets

Example tx: https://arbiscan.io/tx/0x087764c60c9d58ba30f2d7dc5f90b0f58cd64fa2e76cb067247bccd577d58d25
just add `lukicenturi.eth`


kelsos (approved):
lgtm @lukicenturi thanks

PR #9851, opened 2025-04-28
Extend EvmSwapEvents to support multi swaps

nicholasyoder:

* Updates the EvmSwapEvent class to support having a type of either TRADE or MULTI_TRADE
* Changes the evm swap api schema and the associated `edit_grouped_events_with_optional_fee` to allow editing multi swaps
* Extends the swap processing logic to also process multi swaps. We had discussed that maybe we would just make the decoders that needed multiswaps directly create EvmSwapEvents, but this actually looked more complicated to implement. The changes required to process multi swaps like other swaps were relatively simple.
* Changed the schema and swap processing to allow `location_label` to be different for each the event in a group, since sometimes the initiator/sender of a swap can be different from the receiver as mentioned in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1366336145562275900)
* Also updated the api docs to reflect the recent changes with grouping and the changes in this PR to the evm swap add/editing.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -5010,6 +5010,86 @@ Dealing with History Events
                   },
                   "grouped_events_num": 3,
                   "event_accounting_rule_status": "processed"
+              }, [{
+                  "entry": {
+                      "timestamp": 1569924575000,
+                      "event_type": "multi trade",
+                      "event_subtype": "spend",
+                      "location": "ethereum",
+                      "location_label": null,
+                      "asset": "ETH",
+                      "amount": "0.123",
+                      "identifier": 10,
+                      "entry_type": "evm swap event",
+                      "event_identifier": "10x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f",
+                      "sequence_index": 5,
+                      "extra_data": null,
+                      "tx_hash": "0x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f",
+                      "counterparty": null,
+                      "product": null,
+                      "address": null
+                  },
+                  "event_accounting_rule_status": "not processed"
+              }, {
+                  "entry": {
+                      "timestamp": 1569924575000,
+                      "event_type": "multi trade",
+                      "event_subtype": "spend",
+                      "location": "ethereum",
+                      "location_label": null,
+                      "asset": "eip155:1/erc20:0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599",
+                      "amount": "0.0032",
+                      "identifier": 7,
+                      "entry_type": "evm swap event",
+                      "event_identifier": "10x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f",
+                      "sequence_index": 6,
+                      "extra_data": null,
+                      "tx_hash": "0x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f",
+                      "counterparty": null,
+                      "product": null,
+                      "address": null
+                  },
+                  "event_accounting_rule_status": "not processed"
+              }, {
+                  "entry": {
+                      "timestamp": 1569924575000,
+                      "event_type": "multi trade",
+                      "event_subtype": "receive",
+                      "location": "ethereum",
+                      "location_label": null,
+                      "asset": "eip155:1/erc20:0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+                      "amount": "120",
+                      "identifier": 8,
+                      "entry_type": "evm swap event",
+                      "event_identifier": "10x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f",
+                      "sequence_index": 7,
+                      "extra_data": null,
+                      "tx_hash": "0x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f",
+                      "counterparty": null,
+                      "product": null,
+                      "address": null
+                  },
+                  "event_accounting_rule_status": "not processed"
+              }, {
+                  "entry": {
+                      "timestamp": 1569924575000,
+                      "event_type": "multi trade",
+                      "event_subtype": "fee",
+                      "location": "ethereum",
+                      "location_label": null,
+                      "asset": "ETH",
+                      "amount": "0.0002",
+                      "identifier": 10,
+                      "entry_type": "evm swap event",
+                      "event_identifier": "10x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f",
+                      "sequence_index": 9,
+                      "extra_data": null,
+                      "tx_hash": "0x8d822b87407698dd869e830699782291155d0276c5a7e5179cb173608554e41f",
+                      "counterparty": null,
+                      "product": null,
+                      "address": null
+                  },
+                  "event_accounting_rule_status": "not processed"
               }],
              "entries_found": 95,
              "entries_limit": 500,
@@ -5018,7 +5098,7 @@ Dealing with History Events
           "message": ""
       }
 
-   :resjson list decoded_events: A list of history events. Each event is an object comprised of the event entry and a boolean denoting if the event has been customized by the user or not. Each entry may also have a `has_details` flag if true. If `has_details` is true, then it is possible to call /history/events/details endpoint to retrieve some extra information about the event. Also each entry may have a `customized` flag set to true. If it does, it means the event has been customized/added by the user. Each entry may also have a `hidden` flag if set to true. If it does then that means it should be hidden in the UI due to consolidation of events. Also if `group_by_event_ids` exist and is true, each entry contains `grouped_events_num` which is an integer with the amount of events under the common event identifier. The consumer has to query this endpoint again with `group_by_event_ids` set to false and with the `event_identifiers` filter set to the identifier of the events having more than 1 event. Finally `ignored_in_accounting` is set to `true` when the user has marked this event as ignored. Following are all possible entries depending on entry type.
+   :resjson list decoded_events: A list of history events, with some events grouped into sub-lists (for instance the spend/receive/fee events making up a swap). Each event is an object comprised of the event entry and a boolean denoting if the event has been customized by the user or not. Each entry may also have a `has_details` flag if true. If `has_details` is true, then it is possible to call /history/events/details endpoint to retrieve some extra information about the event. Also each entry may have a `customized` flag set to true. If it does, it means the event has been customized/added by the user. Each entry may also have a `hidden` flag if set to true. If it does then that means it should be hidden in the UI due to consolidation of events. Also if `group_by_event_ids` exist and is true, each entry contains `grouped_events_num` which is an integer with the amount of events under the common event identifier. The consumer has to query this endpoint again with `group_by_event_ids` set to false and with the `event_identifiers` filter set to the identifier of the events having more than 1 event. Finally `ignored_in_accounting` is set to `true` when the user has marked this event as ignored. Following are all possible entries depending on entry type.
    :resjson string identifier: Common key. This is the identifier of a single event.
    :resjson string entry_type: Common key. This identifies the category of the event and determines the schema. Possible values are: ``"history event"``, ``"evm event"``, ``"eth withdrawal event"``, ``"eth block event"``, ``"eth deposit event"``.
    :resjson string event_identifier: Common key. An event identifier grouping multiple events under a common group. This is how we group transaction events under a transaction, staking related events under block production etc.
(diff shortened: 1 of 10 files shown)


rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -771,6 +779,10 @@ def _decode_transaction(
         if len(events) == 0 and (eth_event := self._get_eth_transfer_event(transaction)) is not None:  # noqa: E501
             events = [eth_event]
 
+        events = sorted(events, key=lambda x: x.sequence_index, reverse=False)
+        if process_swaps:
+            events = self._process_swaps(transaction=transaction, decoded_events=events)
+
```
nicholasyoder, line 785: Moved this to be after the `monerium_special_handling_event` logic. There is a paraswap test where the spend is monerium and so it was getting two spends in the process swaps logic. With the changes in this PR it was now seeing it as a multi trade since there were still both monerium transfers when it tried to process swaps (before, process swaps just ignored the one since it only handled single spend trades). Now the duplicate monerium transfer is removed prior to processing swaps and it sees it as a normal swap again.

rotkehlchen/api/v1/schemas.py
```diff
@@ -1027,72 +1018,111 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
...
+
+            events, is_multi = [], False
+            for subtype in (
+                HistoryEventSubType.SPEND,
+                HistoryEventSubType.RECEIVE,
+                HistoryEventSubType.FEE,
+            ):
+                subtype_identifiers = identifiers.get(subtype_str := subtype.serialize(), [])
+                try:
+                    amounts_and_assets = list(zip(
```
prettyirrelevant, line 1071: is `list()` needed here? i guess you can enumerate over a zip since it's an iterable

rotkehlchen/api/v1/schemas.py
```diff
@@ -988,12 +951,40 @@ def make_history_base_entry(
...
         location = LocationField(required=True)
+        spend_amount = AmountField(required=True, validate=validate.Range(min=ZERO, min_inclusive=False))  # noqa: E501
+        spend_asset = AssetField(required=True, expected_type=Asset, form_with_incomplete_data=True)  # noqa: E501
+        receive_amount = AmountField(required=True, validate=validate.Range(min=ZERO, min_inclusive=False))  # noqa: E501
+        receive_asset = AssetField(required=True, expected_type=Asset, form_with_incomplete_data=True)  # noqa: E501
+        fee_amount = AmountField(required=False, load_default=None, validate=validate.Range(min=ZERO, min_inclusive=False))  # noqa: E501
+        fee_asset = AssetField(required=False, load_default=None, expected_type=Asset, form_with_incomplete_data=True)  # noqa: E501
+        location_label = fields.String(required=False, load_default=None)
+        unique_id = fields.String(required=False, load_default=None)
+        user_notes = fields.List(fields.String(), required=False, validate=validate.Length(min=2, max=3))  # noqa: E501
```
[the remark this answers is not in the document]
nicholasyoder: Note that its using `data.get()` for the user_notes - so it should be fine.

docs/api.rst
```diff
@@ -5248,6 +5333,7 @@ Dealing with History Events
             }
 
          :reqjson string event_type: The type of asset movement event ("deposit" or "withdrawal")
+         :reqjson object amount: The amount of being moved.
```
LefterisJP, line 5336:
```suggestion
         :reqjson object amount: The amount being moved.
```

rotkehlchen/api/rest_helpers/history_events.py
```diff
@@ -78,3 +75,61 @@ def edit_grouped_events_with_optional_fee(
...
+    """
+    edited_identifiers, new_events = defaultdict(list), []
+    for event in events:
+        if event.identifier is None:  # No existing identifier - this is a new event.
+            new_events.append(event)
+        else:  # Already has an identifier - edit existing event.
+            events_db.edit_history_event(write_cursor=write_cursor, event=event)
+            edited_identifiers[event.event_subtype.serialize()].append(event.identifier)
+
+    if identifiers != edited_identifiers:  # There are identifiers with no corresponding events - these events need deleted.  # noqa: E501
```
LefterisJP, line 106:
```suggestion
    if identifiers != edited_identifiers:  # There are identifiers with no corresponding events - these events need to be deleted.  # noqa: E501
```

rotkehlchen/api/v1/schemas.py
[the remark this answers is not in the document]
LefterisJP: yeah but why not keep it consistent and add load_default=None here and remove the `.get()` from `make_history_base_entry`.

Why be incosistent in this one field?

rotkehlchen/api/v1/schemas.py
```diff
@@ -1027,72 +1018,111 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
...
+        receive_amounts = fields.List(AmountField(validate=validate.Range(min=ZERO, min_inclusive=False)), required=True, validate=validate.Length(min=1))  # noqa: E501
+        receive_assets = fields.List(AssetField(expected_type=Asset, form_with_incomplete_data=True), required=True, validate=validate.Length(min=1))  # noqa: E501
+        fee_amounts = fields.List(AmountField(load_default=None, validate=validate.Range(min=ZERO, min_inclusive=False)), required=False)  # noqa: E501
+        fee_assets = fields.List(AssetField(load_default=None, expected_type=Asset, form_with_incomplete_data=True), required=False)  # noqa: E501
+        identifiers = fields.Dict(
+            keys=fields.String(validate=validate.OneOf(['spend', 'receive', 'fee'])),
+            values=fields.List(fields.Integer()),
+            required=True,
+        )
+        location_labels = fields.Dict(
```
LefterisJP, line 1037: same here for the default loading. Just use `load_default=XXX` instead of using `data.get()` in the implementation of a custom `@post_load` function. Use the library we are working with, instead of doing things manually

rotkehlchen/api/v1/schemas.py
```diff
@@ -1027,72 +1018,111 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
...
+            ):
+                subtype_identifiers = identifiers.get(subtype_str := subtype.serialize(), [])
+                try:
+                    amounts_and_assets = zip(
+                        (amounts_list := data.get(f'{subtype_str}_amounts', [])),
+                        data.get(f'{subtype_str}_assets', []),
+                        strict=True,
+                    )
+                except ValueError as e:
+                    raise ValidationError(f'Number of {subtype_str} amounts and assets must correlate.') from e  # noqa: E501
```
LefterisJP, line 1077: you can also add the `field_name` for this validation error right?

rotkehlchen/api/v1/schemas.py
```diff
@@ -1027,72 +1018,111 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
...
+                for subtype_data, field_name in (
+                    ((subtype_notes := user_notes.get(subtype_str, [])), 'user_notes'),
+                    ((subtype_location_labels := location_labels.get(subtype_str, [])), 'location_labels'),  # noqa: E501
+                ):
+                    if len(subtype_data) > len(amounts_list):  # Use amounts_list instead of amounts_and_assets since `len()` doesn't work on a zip object  # noqa: E501
+                        raise ValidationError(
+                            message=f'Number of {subtype_str} {field_name} may not be larger '
+                                    f'than the number of {subtype_str} amounts and assets.',
+                            field_name=field_name,
+                        )
```
LefterisJP, line 1088:
```suggestion
                        raise ValidationError(
                            message=(
                                f'Number of {subtype_str} {field_name} may not be larger '
                                f'than the number of {subtype_str} amounts and assets.'
                             ),
                            field_name=field_name,
                        )
```

rotkehlchen/api/v1/schemas.py
```diff
@@ -1027,72 +1018,111 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
...
+
+                subtype_events = []
+                for idx, (amount, asset) in enumerate(amounts_and_assets):
+                    subtype_events.append(EvmSwapEvent(
+                        identifier=subtype_identifiers[idx] if len(subtype_identifiers) > idx else None,  # noqa: E501
```
LefterisJP, line 1093: I don't get what this does. Comment please

rotkehlchen/api/v1/schemas.py
```diff
@@ -1027,72 +1018,111 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
...
+                        tx_hash=tx_hash,
+                        sequence_index=sequence_index,
+                        timestamp=timestamp,
+                        location=location,
+                        event_type=HistoryEventType.TRADE,
+                        event_subtype=subtype,
+                        asset=asset,
+                        amount=amount,
+                        location_label=subtype_location_labels[idx] if len(subtype_location_labels) > idx else None,  # noqa: E501
```
LefterisJP, line 1102: Isn't this the same check essentially? Or those lists are not supposed to be the same length?

rotkehlchen/api/v1/schemas.py
```diff
@@ -1027,72 +1018,111 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
...
+                        notes=subtype_notes[idx] if len(subtype_notes) > idx else None,
```
LefterisJP, line 1103: same for this

rotkehlchen/api/v1/schemas.py
```diff
@@ -1121,7 +1151,11 @@ def make_history_base_entry(
             **_kwargs: Any,
     ) -> dict[str, Any]:
         entry_type = data.pop('entry_type')  # already used to decide schema
-        exclude = () if self.include_identifier else ('identifier',)
+        exclude = () if self.include_identifier else (
```
LefterisJP, line 1154: explain with a comment what you are doing here and why. I don't get it.

rotkehlchen/chain/evm/decoding/decoder.py
LefterisJP: add a comment here then on why this must be here after the monerium handling.

rotkehlchen/api/v1/schemas.py
nicholasyoder: Yeah these lists should be the same length, although since notes and location_label are optional they can be skipped entirely which is why it was checking like this.

To avoid these checks in the loop here, now it instead sets the default list to be a list of `None` values of the correct length.

rotkehlchen/api/v1/schemas.py
nicholasyoder: Added a comment explaining whats happening here.

rotkehlchen/api/v1/schemas.py
nicholasyoder: Added a comment explaining this.

rotkehlchen/api/v1/schemas.py
```diff
@@ -1027,72 +1018,122 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
...
+        spend_amounts = fields.List(AmountField(validate=validate.Range(min=ZERO, min_inclusive=False)), required=True, validate=validate.Length(min=1))  # noqa: E501
+        spend_assets = fields.List(AssetField(expected_type=Asset, form_with_incomplete_data=True), required=True, validate=validate.Length(min=1))  # noqa: E501
+        receive_amounts = fields.List(AmountField(validate=validate.Range(min=ZERO, min_inclusive=False)), required=True, validate=validate.Length(min=1))  # noqa: E501
+        receive_assets = fields.List(AssetField(expected_type=Asset, form_with_incomplete_data=True), required=True, validate=validate.Length(min=1))  # noqa: E501
+        fee_amounts = fields.List(AmountField(validate=validate.Range(min=ZERO, min_inclusive=False)), required=False, load_default=[])  # noqa: E501
+        fee_assets = fields.List(AssetField(expected_type=Asset, form_with_incomplete_data=True), required=False, load_default=[])  # noqa: E501
+        identifiers = fields.Dict(
+            keys=fields.String(validate=validate.OneOf(['spend', 'receive', 'fee'])),
+            values=fields.List(fields.Integer()),
+            required=True,
```
LefterisJP, line 1035: why not do the load_default here to and avoid the .get() as we said? Any limitation?

nicholasyoder: Notice that `identifiers` is actually a required field. This means that load_default doesn't work - it does this:
```
ValueError: 'load_default' must not be set for required fields.
```
But `identifiers` will be missing some times, since it gets excluded from the schema in CreateHistoryEventSchema's post_load depending on whether `include_identifier` is set - this allows identifiers to be a required field when editing but not when creating.

Added a comment in the code explaining why its using get instead of load_default.

LefterisJP (approved):
LGTM

PR #9852, opened 2025-04-28
Fix problem with empty event query

nicholasyoder:
Fixes a bug where it tried to unzip an empty list when querying history events when no events were returned and group_by_event_ids was set. This problem was introduced recently in https://github.com/rotki/rotki/pull/9830 with the changes to the even grouping.


LefterisJP (approved):
LGTM

PR #9859, opened 2025-04-29
Check tx_hash for existance before adding/editing

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=107754546

### DIFF
--- a/rotkehlchen/api/v1/schemas.py
+++ b/rotkehlchen/api/v1/schemas.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING, Any, Final, Literal, cast, get_args
 import marshmallow
 import webargs
 from eth_utils import to_checksum_address
-from marshmallow import INCLUDE, Schema, fields, post_load, validate, validates_schema
+from marshmallow import INCLUDE, Schema, fields, post_load, validate, validates, validates_schema
 from marshmallow.exceptions import ValidationError
 from werkzeug.datastructures import FileStorage
 
@@ -761,6 +761,21 @@ class CreateHistoryEventSchema(Schema):
         address = EvmAddressField(load_default=None)
         extra_data = fields.Dict(load_default=None)
         location = LocationField(required=True, limit_to=EVM_EVMLIKE_LOCATIONS)
+
+        @validates('tx_hash')
+        def validate_tx_hash(self, tx_hash: str) -> None:
+            """Check if the provided tx_hash is present in the db.
+            Raises ValidationError if tx_hash is missing.
+            """
+            with CreateHistoryEventSchema.history_event_context.get()['schema'].database.conn.read_ctx() as cursor:  # noqa: E501
+                if cursor.execute(
+                    'SELECT COUNT(*) FROM evm_transactions WHERE tx_hash=?',
+                    (tx_hash,),
+                ).fetchone()[0] == 0:
+                    raise ValidationError(
+                        message='The provided transaction hash does not exist in the DB.',
+                        field_name='tx_hash',
+                    )
 
     class CreateBaseHistoryEventSchema(BaseEventSchema):
         event_identifier = fields.String(required=True)
--- a/rotkehlchen/tests/api/test_history_base_entry.py
+++ b/rotkehlchen/tests/api/test_history_base_entry.py
@@ -7,6 +7,7 @@ from unittest.mock import patch
 import pytest
 import requests
 
+from rotkehlchen.chain.evm.constants import ZERO_ADDRESS
 from rotkehlchen.chain.evm.decoding.constants import CPT_GAS
 from rotkehlchen.chain.evm.types import string_to_evm_address
 from rotkehlchen.constants import ZERO
@@ -62,7 +63,9 @@ from rotkehlchen.utils.misc import ts_sec_to_ms
 
 if TYPE_CHECKING:
     from rotkehlchen.api.server import APIServer
+    from rotkehlchen.db.dbhandler import DBHandler
     from rotkehlchen.history.events.structures.base import HistoryBaseEntry
+    from rotkehlchen.types import EVMTxHash
 
 
 def assert_editing_works(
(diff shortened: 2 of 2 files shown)


rotkehlchen/api/rest.py
```diff
@@ -922,17 +922,39 @@ def query_blockchain_balances(
...
     def add_history_events(self, events: list['HistoryBaseEntry']) -> Response:
         """Add list of history events to DB. Returns identifier of first event.
         The first event is the main event, subsequent events are related (e.g. fees).
         """
         db = DBHistoryEvents(self.rotkehlchen.data.db)
         main_identifier = None
         try:
-            with self.rotkehlchen.data.db.user_write() as cursor:
+            with self.rotkehlchen.data.db.user_write() as write_cursor:
+                if self._is_evm_event_tx_hash_missing(write_cursor=write_cursor, event=events[0]):
```
prettyirrelevant, line 949: why not do this at the schemas level? we already pass the `database` to the marshmallow schema.

rotkehlchen/api/rest.py
```diff
@@ -922,17 +922,39 @@ def query_blockchain_balances(
...
+            event: 'HistoryBaseEntry',
+    ) -> bool:
+        if event.entry_type not in (
+            HistoryBaseEntryType.EVM_EVENT,
+            HistoryBaseEntryType.EVM_SWAP_EVENT,
+        ):
+            return False
+
+        return write_cursor.execute(
+            'SELECT COUNT(*) FROM evm_events_info WHERE tx_hash=?',
```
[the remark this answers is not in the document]
nicholasyoder: Two reasons:
* In the app, a user cannot remove the last event from an evm transaction. So if a tx_hash is in evm_transactions it will also be in evm_events_info at least once.
* Only need to patch this validation in the tests when initially adding events. When editing existing events their hashes will be present in evm_events_info

Added this explanation to the docstring of the validation function.

rotkehlchen/api/rest.py
nicholasyoder: Good idea. Moved it there and now it simply raises a validation error for missing tx_hash.

rotkehlchen/api/v1/schemas.py
```diff
@@ -762,6 +762,27 @@ class BaseEvmEventSchema(Schema):
         extra_data = fields.Dict(load_default=None)
         location = LocationField(required=True, limit_to=EVM_EVMLIKE_LOCATIONS)
 
+        @staticmethod
+        def validate_tx_hash(tx_hash: str) -> None:
```
prettyirrelevant, line 766: you can do this instead

```py
@validates("tx_hash")
    def validate_tx_hash(self, data, **kwargs):
        if not data:
            raise ValidationError("There is no data.")

File: rotkehlchen/api/v1/schemas.py [hunk 4]
```diff
@@ -762,6 +762,27 @@ class BaseEvmEventSchema(Schema):
...
+            """Check if the provided tx_hash is present in the db.
+            Check evm_events_info rather than evm_transactions here for two reasons:
+            * In the app, a user cannot remove the last event from an evm transaction. So if a
+               tx_hash is in evm_transactions it will also be in evm_events_info at least once.
+            * Only need to patch this validation in the tests when initially adding events.
+               When editing existing events their hashes will be present in evm_events_info
```
prettyirrelevant, line 772: i don't think the db table will matter much if you're just patching the method. perhaps, this comment is not necessary.

rotkehlchen/api/v1/schemas.py
```diff
@@ -762,6 +762,27 @@ class BaseEvmEventSchema(Schema):
         extra_data = fields.Dict(load_default=None)
         location = LocationField(required=True, limit_to=EVM_EVMLIKE_LOCATIONS)
 
+        @staticmethod
+        def validate_tx_hash(tx_hash: str) -> None:
+            """Check if the provided tx_hash is present in the db.
+            Check evm_events_info rather than evm_transactions here for two reasons:
+            * In the app, a user cannot remove the last event from an evm transaction. So if a
+               tx_hash is in evm_transactions it will also be in evm_events_info at least once.
```
prettyirrelevant, line 770: my only reason for suggesting `evm_transactions` is because `evm_events_info.tx_hash` does not have a foreign key relationship to `evm_transactions`.

the reasoning you gave makes sense so you can leave as is.

rotkehlchen/api/v1/schemas.py
nicholasyoder: ah, nice. I wasn't aware of that decorator.

rotkehlchen/api/v1/schemas.py
nicholasyoder: mmm... so the point here is that if we were checking the other table we would have to patch things in more places since in the tests for this we generally don't add the tx to `evm_transactions` but it will be present in `evm_events_info` when its editing events. Other than this it doesn't really matter which table is checked here.

Reworded the comment to try and make this clearer.

rotkehlchen/api/v1/schemas.py
```diff
@@ -762,6 +762,27 @@ class BaseEvmEventSchema(Schema):
         extra_data = fields.Dict(load_default=None)
         location = LocationField(required=True, limit_to=EVM_EVMLIKE_LOCATIONS)
 
+        @validates('tx_hash')
+        def validate_tx_hash(self, tx_hash: str) -> None:
+            """Check if the provided tx_hash is present in the db.
+            Check evm_events_info rather than evm_transactions here for two reasons:
```
LefterisJP, line 768: no please change it to evm_transaction as Isaac said. The relationship you stated in your first reason can change at any point. And you can add a transaction hash manually from the UI and it can have no event decoded. So it is also a false assumption.

The second one is for testing so it's not relevant

nicholasyoder: > you can add a transaction hash manually from the UI and it can have no event decoded

ah, right. Should have thought about that.

> The second one is for testing so it's not relevant
yeah - I was thinking it made no difference otherwise so we might as well go with the one that was easier in the tests, but that isn't the case after all. Also now instead of patching this in the tests its simpler (and tests this better) to actually add a tx with the right hash to the db.

LefterisJP (approved):
LGTM

PR #9862, opened 2025-04-30
improve events export timestamp and filename

prettyirrelevant:
Related https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=108217361

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -4748,7 +4748,7 @@ class RestAPI:
             headers.update(dict.fromkeys(serialized_event))
 
         try:
-            filename = generate_events_export_filename(filter_query)
+            filename = generate_events_export_filename(filter_query=filter_query, use_localtime=settings.display_date_in_localtime)  # noqa: E501
             if directory_path is None:  # file will be downloaded later via download_history_events_csv endpoint  # noqa: E501
                 file_path = Path(tempfile.mkdtemp()) / filename
                 dict_to_csv_file(
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -32,7 +32,6 @@ from rotkehlchen.accounting.constants import (
 from rotkehlchen.accounting.debugimporter.json import DebugHistoryImporter
 from rotkehlchen.accounting.entry_type_mappings import ENTRY_TYPE_MAPPINGS
 from rotkehlchen.accounting.export.csv import (
-    FILENAME_HISTORY_EVENTS_CSV,
     FILENAME_SKIPPED_EXTERNAL_EVENTS_CSV,
     CSVWriteError,
     dict_to_csv_file,
@@ -206,7 +205,10 @@ from rotkehlchen.history.events.structures.base import (
 )
 from rotkehlchen.history.events.structures.evm_event import EvmProduct
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
-from rotkehlchen.history.events.utils import history_event_to_staking_for_api
+from rotkehlchen.history.events.utils import (
+    generate_events_export_filename,
+    history_event_to_staking_for_api,
+)
 from rotkehlchen.history.price import PriceHistorian
 from rotkehlchen.history.skipped import (
     export_skipped_external_events,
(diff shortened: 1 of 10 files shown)


rotkehlchen/api/rest.py
```diff
@@ -4734,14 +4737,18 @@ def export_history_events(
...
+            )
             serialized_history_events.append(serialized_event)
             # maintain insertion order without storing extra info
             headers.update(dict.fromkeys(serialized_event))
 
         try:
+            filename = generate_events_export_filename(filter_query)
             if directory_path is None:  # file will be downloaded later via download_history_events_csv endpoint  # noqa: E501
-                file_path = Path(tempfile.mkdtemp()) / FILENAME_HISTORY_EVENTS_CSV
+                file_path = Path(tempfile.mkdtemp()) / filename
```
nicholasyoder, line 4751: `download_history_events_csv` is still referencing `FILENAME_HISTORY_EVENTS_CSV` - so that also needs updated. Then you should also be able to completely remove the `FILENAME_HISTORY_EVENTS_CSV` constant as well.

rotkehlchen/history/events/utils.py
```diff
@@ -55,3 +56,22 @@ def create_event_identifier(
...
+def generate_events_export_filename(filter_query: 'HistoryBaseEntryFilterQuery') -> str:
+    """Generate a filename for CSV export based on filter properties."""
+    parts = [f'{filter_query.__class__.__name__.replace("FilterQuery", "").lower()}s']
+    if filter_query.timestamp_filter:
+        if filter_query.timestamp_filter.from_ts and filter_query.timestamp_filter.to_ts:
+            from_date = timestamp_to_date(filter_query.timestamp_filter.from_ts, formatstr='%Y%m%d')  # noqa: E501
+            to_date = timestamp_to_date(filter_query.timestamp_filter.to_ts, formatstr='%Y%m%d')
+            parts.append(f'{from_date}_to_{to_date}')
+        elif filter_query.timestamp_filter.to_ts:
+            parts.append(f'till_{timestamp_to_date(filter_query.timestamp_filter.to_ts, formatstr="%Y%m%d")}')  # noqa: E501
```
nicholasyoder, line 70: maybe its just me, but I don't quite like how `till` sounds here. Maybe should be the full word `until`?

Actually it could simplify the logic to simply use `to` here. so the combinations would be `from_xxx_to_xxx`, `from_xxx` and `to_xxx` - but perhaps `historyevents_to_xxxx.csv` isn't as clear as `historyevents_until_xxxx.csv`

With that said `till` can work as well - Only change this if you think it would be better.

rotkehlchen/history/events/utils.py
```diff
@@ -55,3 +57,22 @@ def create_event_identifier(
...
         str(amount),
     )
+
+
+def generate_events_export_filename(filter_query: 'HistoryBaseEntryFilterQuery', settings: 'DBSettings') -> str:  # noqa: E501
+    """Generate a filename for CSV export based on filter properties."""
+    parts = [f'{filter_query.__class__.__name__.replace("FilterQuery", "").lower()}s']
+    if filter_query.timestamp_filter:
+        if filter_query.timestamp_filter.from_ts and filter_query.timestamp_filter.to_ts:
+            from_date = timestamp_to_date(ts=filter_query.timestamp_filter.from_ts, formatstr='%Y%m%d', treat_as_local=settings.display_date_in_localtime)  # noqa: E501
```
nicholasyoder, line 67: mmm... for this wouldn't it be better to just pass a `date_in_localtime` flag to the `generate_events_export_filename` function instead of passing the whole settings object and then having to do `settings.` every time you reference it?

also, one other thing I just noticed here - probably `formatstr` should just be defined once at the beginning instead of hardcoding the string each time timestamp_to_date is used. Actually wait - shouldn't this actually be using `settings.date_display_format` as well? In that case maybe passing the whole settings object is fine if you use two different settings.

rotkehlchen/tests/api/test_history_events_export.py
```diff
@@ -286,7 +287,7 @@ def test_history_export_csv_free_limit(
         ))
         assert_csv_export_response(
             response=response,
-            csv_dir=csv_dir,
+            csv_path=csv_dir / 'historyevents_till_20250430.csv',
```
LefterisJP, line 290: please change till to until here and everywhere.

rotkehlchen/tests/api/test_history_events_export.py
```diff
@@ -173,7 +173,7 @@ def test_history_export_csv_custom_delimiter(
         api_url_for(rotkehlchen_api_server_with_exchanges, 'exporthistoryeventresource'),
         json={'async_query': False, 'directory_path': str(csv_dir)},
     )
-    assert_csv_export_response(response, csv_dir, csv_delimiter=csv_delimiter)
+    assert_csv_export_response(response, csv_dir / 'historyevents_till_20250430.csv', csv_delimiter=csv_delimiter)  # noqa: E501
```
LefterisJP, line 176: I want to see also a type/subtype query filename test

rotkehlchen/history/events/structures/base.py
```diff
@@ -295,6 +300,11 @@ def serialize_for_csv(self, fiat_value: FVal) -> dict[str, Any]:
         """
         new_dict: dict[str, Any] = {}
         entry = self.serialize()
+        entry['timestamp'] = timestamp_to_date(
```
LefterisJP, line 303: Is this tested anywhere?

rotkehlchen/tests/api/test_history_events_export.py
prettyirrelevant: added one.

LefterisJP (approved):
lgtm

PR #9863, opened 2025-04-30
Improve evm swap add/edit schema

nicholasyoder:
Improve the evm swap add/edit schema as discussed in [discord](https://discord.com/channels/657906918408585217/1366881669356851310)

### DIFF
--- a/rotkehlchen/api/rest_helpers/history_events.py
+++ b/rotkehlchen/api/rest_helpers/history_events.py
@@ -1,4 +1,3 @@
-from collections import defaultdict
 from typing import TYPE_CHECKING
 
 from rotkehlchen.db.constants import HISTORY_MAPPING_KEY_STATE, HISTORY_MAPPING_STATE_CUSTOMIZED
@@ -15,7 +14,7 @@ def edit_grouped_events_with_optional_fee(
         write_cursor: 'DBCursor',
         events: list[HistoryBaseEntry],
         events_type: HistoryBaseEntryType,
-        identifiers: dict[str, list[int]] | None = None,
+        identifiers: list[int] | None = None,
 ) -> None:
     """
     Handle grouped events, including fee-related modifications:
@@ -81,12 +80,12 @@ def edit_grouped_evm_swap_events(
         events_db: 'DBHistoryEvents',
         write_cursor: 'DBCursor',
         events: list[HistoryBaseEntry],
-        identifiers: dict[str, list[int]],
+        identifiers: list[int],
         event_identifier: str,
 ) -> None:
     """Handle editing of grouped evm swap events.
-    Determines which events to add, edit, or remove using the `identifiers` dict
-    and the `events` list as follows:
+    Determines which events to add, edit, or remove using the `identifiers`
+    and `events` lists as follows:
     - Inserts events that have no identifier.
     - Edits events whose identifier is set.
     - Removes events whose identifier is present in `identifiers`,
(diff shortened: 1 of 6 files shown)


rotkehlchen/api/rest_helpers/history_events.py
```diff
@@ -95,25 +94,18 @@ def edit_grouped_evm_swap_events(
...
-        for key, id_list in identifiers.items():
-            if id_list == edited_identifiers[key]:
-                continue
-
-            ids_to_delete.extend(list(set(id_list) - set(edited_identifiers[key])))
-
         write_cursor.executemany(
             'DELETE FROM history_events WHERE identifier=?',
-            [(identifier,) for identifier in ids_to_delete],
+            [(identifier,) for identifier in (set(identifiers) - set(edited_identifiers))],
```
prettyirrelevant, line 108: do you need the extra parentheses?

rotkehlchen/api/v1/schemas.py
```diff
@@ -1019,33 +1019,22 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
             return {'events': events}
 
     class CreateEvmSwapEventSchema(BaseEvmEventSchema):
+
+        class EvmSwapSubEventSchema(Schema):
+            identifier = fields.Integer(required=False, load_default=None)
+            amount = AmountField(validate=validate.Range(min=ZERO, min_inclusive=False))
```
prettyirrelevant, line 1025:
```suggestion
            amount = AmountField(validate=validate.Range(min=ZERO, min_inclusive=False), required=True)
```

don't you need to set required=True here?

rotkehlchen/api/v1/schemas.py
```diff
@@ -1058,62 +1047,32 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
             product = data['product']
             address = data['address']
             extra_data = data['extra_data']
-            # Use get with a default value instead of load_default since identifiers is a required
-            # field that sometimes gets excluded in the post_load of CreateHistoryEventSchema.
-            identifiers = data.get('identifiers', {})
-            location_labels = data['location_labels']
-            user_notes = data['user_notes']
+            # Use get instead of load_default=None since identifiers is a required field
```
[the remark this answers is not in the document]
nicholasyoder: mmm... sort of - The point is that it has to use `get` instead of relying on `identifiers` already being set to None by the `load_default` option in the schema. Reworded the comment slightly to try and make that clearer.

rotkehlchen/api/v1/schemas.py
```diff
@@ -1058,62 +1047,32 @@ def make_history_base_entry(self, data: dict[str, Any], **_kwargs: Any) -> dict[
...
-                        (amounts_list := data[f'{subtype_str}_amounts']),
-                        data[f'{subtype_str}_assets'],
-                        strict=True,
-                    )
-                except ValueError as e:
-                    raise ValidationError(
-                        message=f'Number of {subtype_str} amounts and assets must correlate.',
-                        field_name=f'{subtype_str}_amounts',
-                    ) from e
+                if len(subtype_data_list := data.get(subtype.serialize(), [])) > 1:
```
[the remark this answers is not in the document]
nicholasyoder: right. good catch. Changed it.

rotkehlchen/api/v1/schemas.py
nicholasyoder: ah, yes. Changed this. Also I realized that I hadn't mentioned the identifier field in the api docs so added that as well.

LefterisJP (approved):
LGTM

PR #9869, opened 2025-05-01
improve historical price queries for events export

prettyirrelevant:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=108217361

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -104,7 +104,7 @@ from rotkehlchen.chain.evm.types import (
 )
 from rotkehlchen.chain.gnosis.modules.gnosis_pay.constants import CPT_GNOSIS_PAY
 from rotkehlchen.chain.zksync_lite.constants import ZKL_IDENTIFIER
-from rotkehlchen.constants import ONE
+from rotkehlchen.constants import HOUR_IN_SECONDS, ONE
 from rotkehlchen.constants.limits import (
     FREE_HISTORY_EVENTS_LIMIT,
     FREE_USER_NOTES_LIMIT,
@@ -182,7 +182,7 @@ from rotkehlchen.errors.misc import (
     SystemPermissionError,
     TagConstraintError,
 )
-from rotkehlchen.errors.price import NoPriceForGivenTimestamp, PriceQueryUnsupportedAsset
+from rotkehlchen.errors.price import NoPriceForGivenTimestamp
 from rotkehlchen.errors.serialization import DeserializationError
 from rotkehlchen.exchanges.constants import ALL_SUPPORTED_EXCHANGES
 from rotkehlchen.exchanges.utils import query_binance_exchange_pairs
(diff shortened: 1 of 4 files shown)


rotkehlchen/api/rest.py
```diff
@@ -4719,25 +4719,30 @@ def export_history_events(
 
         serialized_history_events = []
         headers: dict[str, None] = {}
+        prices_from_db = GlobalDBHandler.get_historical_prices(
+            query_data=(query_data := [(event.asset, currency, ts_ms_to_sec(event.timestamp)) for event in history_events]),  # type: ignore[arg-type]  # currency is a subclass of Asset  # noqa: E501
+            max_seconds_distance=HOUR_IN_SECONDS,
+        )
+        missing_prices: list[tuple[Asset, Timestamp]] = []
+        cached_db_prices: defaultdict[Asset, defaultdict[Timestamp, FVal]] = defaultdict(lambda: defaultdict(lambda: ZERO))  # noqa: E501
+        for idx, data in enumerate(query_data):
```
nicholasyoder, line 4728:
```suggestion
        for idx, (asset, _, timestamp) in enumerate(query_data):
```
Using `data[x]` is quite confusing.

rotkehlchen/api/rest.py
```diff
@@ -4719,25 +4719,30 @@ def export_history_events(
...
+            if (db_price := prices_from_db[idx]) is None or (data[0], data[2]) not in missing_prices:  # noqa: E501
+                missing_prices.append((data[0], data[2]))
+            else:
+                cached_db_prices[db_price.from_asset][db_price.timestamp] = db_price.price
```
nicholasyoder, line 4732: mmm... I think somethings wrong with the logic in this `if`. Since it does `or` won't it add all the entries to missing_prices regardless of whether db_price is None?

rotkehlchen/api/rest.py
```diff
@@ -4719,25 +4719,30 @@ def export_history_events(
...
+            if (cached_price := cached_db_prices[event.asset][ts_ms_to_sec(event.timestamp)]) != ZERO:  # noqa: E501
+                fiat_value = cached_price * event.amount
+            elif (queried_price := queried_prices[event.asset][ts_ms_to_sec(event.timestamp)]) != ZERO:  # noqa: E501
+                fiat_value = queried_price * event.amount
             else:
-                fiat_value = event.amount * price
+                fiat_value = ZERO
 
             serialized_event = event.serialize_for_csv(
                 fiat_value=fiat_value,
```
nicholasyoder, line 4748:
```suggestion
            serialized_event = event.serialize_for_csv(
                fiat_value=event.amount * (
                    cached_db_prices[event.asset][ts_ms_to_sec(event.timestamp)] or
                    queried_prices[event.asset][ts_ms_to_sec(event.timestamp)]
                )
```
I think doing something like this should work, and then the ifs above can be removed.

rotkehlchen/tests/api/test_history_events_export.py
```diff
@@ -162,6 +162,7 @@ def test_history_export_download_csv(
 
 @pytest.mark.parametrize('db_settings', [{'csv_export_delimiter': ';'}])
 @pytest.mark.freeze_time('2025-04-30')
+@pytest.mark.parametrize('default_mock_price_value', [ONE])
```
nicholasyoder, line 165: Why this change? There's no changes to the test... I don't understand why this would need to be added.

rotkehlchen/tests/api/test_historical_assets_price.py
```diff
@@ -26,22 +26,7 @@
...
-    'USD': {'USD': {1579543935: FVal('1')}},
-    'GBP': {
-        'USD': {
-            1548007935: FVal('1.25'),
-            1611166335: FVal('1.27'),
-        },
-    },
-    'XRP': {'USD': {1611166335: FVal('0')}},
-}])
+@pytest.mark.vcr
```
nicholasyoder, line 29: mmm... why change these tests to be VCRed instead of mocking prices? In my previous comment, I wasn't meaning that adding `@pytest.mark.parametrize('default_mock_price_value', [ONE])` was the wrong approach - I was just asking why it was needed. I'm still not sure I follow why there are any changes needed to the tests at all.

nicholasyoder: For future readers of this PR - we discussed this in [discord](https://discord.com/channels/657906918408585217/1080783409640648765/1367841704148865026)

LefterisJP (approved):
LGTM

PR #9870, opened 2025-05-01
Remove etherscan from db

yabirgb:
- Removes etherscan related information from the user db
- Removes code to represent etherscan in other chains
- Updates packaged globaldb


rotkehlchen/data_migrations/utils.py
```diff
@@ -31,15 +36,21 @@ def update_data_and_detect_accounts(
...
-    with rotki.data.db.conn.write_ctx() as write_cursor:
-        rotki.data.db.add_external_service_credentials(
-            write_cursor=write_cursor,  # add temporary etherscan key for the given chains
-            credentials=external_service_credentials,
+    with rotki.data.db.conn.read_ctx() as cursor:
+        cursor.execute(
+            'SELECT COUNT(*) FROM external_service_credentials WHERE name=?',
+            (ExternalService.ETHERSCAN.name.lower(),),
         )
+        if (needs_etherscan_key := cursor.fetchone()[0] == 0):
```
prettyirrelevant, line 44:
```suggestion
        if (needs_etherscan_key := cursor.fetchone()[0]) == 0:
```

maybe `needs_etherscan_key` isn't a right variable name as it's an integer even though in this case it's either 0/1

rotkehlchen/data_migrations/utils.py
```diff
@@ -31,15 +36,21 @@ def update_data_and_detect_accounts(
...
+                    )],
+                )
 
     rotki.chains_aggregator.detect_evm_accounts(progress_handler, chains=chains)
-    external_services = []
-    for external_service, _, _ in external_service_credentials:
-        external_services.append(external_service)
-
-    rotki.data.db.delete_external_service_credentials(external_services)
+    if needs_etherscan_key:
```
prettyirrelevant, line 55: shouldn't this be `not needs_etherscan_key`? since you're trying to delete the temporary key added earlier

rotkehlchen/tests/api/test_misc.py
```diff
@@ -360,45 +360,6 @@ def test_manage_nodes(rotkehlchen_api_server: 'APIServer') -> None:
             assert FVal(node['weight']) == 20
             break
 
-    # Try to edit etherscan weight
```
prettyirrelevant, line 363: isn't etherscan still a node? why is this test being deleted?

rotkehlchen/tests/api/test_misc.py
```diff
@@ -220,20 +220,20 @@ def test_manage_nodes(rotkehlchen_api_server: 'APIServer') -> None:
             assert node['blockchain'] == blockchain_key
             break
 
-    # Try to add etherscan as node
+    # Try to add an empty name
     response = requests.put(
         api_url_for(rotkehlchen_api_server, 'rpcnodesresource', blockchain=blockchain_key),
         json={
-            'name': 'etherscan',
```
prettyirrelevant, line 227: same thing here. you should not be able to add "etherscan" as a node.

rotkehlchen/tests/utils/ethereum.py
```diff
@@ -171,7 +174,7 @@ def wait_until_all_nodes_connected(
         timeout: int = NODE_CONNECTION_TIMEOUT,
 ):
     """Wait until all ethereum nodes are connected or until a timeout is hit"""
-    connect_at_start = [x for x in connect_at_start if x.node_info.name != evm_inquirer.etherscan_node_name]  # noqa: E501
+    connect_at_start = [x for x in connect_at_start if x.node_info.name != ETHEREUM_ETHERSCAN_NODE_NAME]  # noqa: E501
```
prettyirrelevant, line 177: i think i'm confused. you deleted every etherscan node so there's no way there's going to be a match for the `if` check.

rotkehlchen/api/v1/schemas.py
```diff
@@ -3294,9 +3286,9 @@ class RpcAddNodeSchema(Schema):
     blockchain = BlockchainField(required=True, exclude_types=(SupportedBlockchain.ETHEREUM_BEACONCHAIN,))  # noqa: E501
     name = fields.String(
         required=True,
-        validate=webargs.validate.NoneOf(
```
prettyirrelevant, line 3297: i don't think this should change.

rotkehlchen/db/upgrades/v47_v48.py
```diff
@@ -202,6 +202,35 @@ def _replace_history_note_locations(write_cursor: 'DBCursor') -> None:
...
+                'base_etherscan',
+                'gnosis_etherscan',
+                'scroll_etherscan',
+                'binance_sc_etherscan',
+            ),
+        )
+        write_cursor.execute(
+            'DELETE FROM rpc_nodes WHERE name IN (?, ?, ?, ?, ?, ?, ?, ?)',
+            (
+                'etherscan',
```
prettyirrelevant, line 222: this should not be deleted right?

rotkehlchen/api/v1/schemas.py
yabirgb: We don't allow touching etherscan anymore. It won't be available in the UI

rotkehlchen/tests/api/test_misc.py
yabirgb: now you can name it as you want since etherscan is not a reserver name anymore. The only condition now is that the endpoint can't be empty

rotkehlchen/tests/utils/ethereum.py
yabirgb: I did the change in two steps first removing the other etherscan endpoints and finally all the etherscan from the db. THis isn't needed

rotkehlchen/data_migrations/utils.py
yabirgb: mmm no, the variable is a boolean from the comparassion. I've renamed it nevertheless

rotkehlchen/api/v1/schemas.py
prettyirrelevant: i see.

rotkehlchen/data_migrations/migrations/migration_10.py
```diff
@@ -22,11 +21,6 @@ def data_migration_10(rotki: 'Rotkehlchen', progress_handler: 'MigrationProgress
     # steps are: ethereum accounts + 4 (potentially write to db + updating spam assets + polygon rpc + new round msg)  # noqa: E501
     progress_handler.set_total_steps(len(rotki.chains_aggregator.accounts.eth) + 4)
 
-    # Check updates for spam assets. This happens before accounts detection to avoid
```
LefterisJP, line 25: This is completely removed without a replacement. It's not in `update_data_and_detect_accounts`

LefterisJP (commented):
This also needs a changelog entry linking to the etherscan v2 announcement explaining we switch everything to it now, and delete all old api keys from the app since they can't be used anymore.

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -3131,6 +3131,17 @@ def test_upgrade_db_47_to_48(user_data_dir, messages_aggregator):
             ('gemini_trades_gemini', 1672531200, 1704067200),
         ]
         assert table_exists(cursor, 'action_type')
+        assert cursor.execute('SELECT COUNT(*) FROM rpc_nodes').fetchone()[0] == 51
```
LefterisJP, line 3134: add test for the unified api setting before and after upgrade too

rotkehlchen/data_migrations/migrations/migration_10.py
LefterisJP: no it's in that functionfrom before. Resolving

LefterisJP (approved):
LGTM

PR #9872, opened 2025-05-02
fix: update value on ignore / unignore assets

lukicenturi:


frontend/app/src/store/monitor.ts
```diff
@@ -124,6 +125,12 @@ export const useMonitorStore = defineStore('monitor', () => {
     }
   });
 
+  const { ignoredAssets } = storeToRefs(useIgnoredAssetsStore());
+
+  watch(ignoredAssets, () => {
+    startPromise(fetchBlockchainBalances());
+  });
+
```
kelsos, line 133: So this will refetch all balances every time someone toggles the ignore/unignore for an asset right? what do we want to achieve here? maybe there is a more optimized way to achieve the result?

lukicenturi: So the ignored assets gets excluded from blockchain balances endpoint.
if we unignore, we need to refresh it.

I will optimize it, by only refresh particular chains, instead all

kelsos: @lukicenturi maybe then can we only make sure that it's called on unignore

lukicenturi: okay yeah we can do that, let me change it

lukicenturi: @kelsos updated

frontend/app/src/modules/balances/use-balances-store.ts
```diff
@@ -42,21 +42,60 @@ export const useBalancesStore = defineStore('balances', () => {
...
     const data = perAccount[camelCase(chain)] ?? {};
-    set(balances, {
-      ...get(balances),
+    set(blockchainBalances, {
+      ...get(blockchainBalances),
       [chain]: data,
     });
   };
 
+  const removeIgnoredAssets = (ignoredAssets: string[]): void => {
```
kelsos, line 53: I did some small changes in the remove method, that should improve performance a bit.

Also added a few tests to make sure that all works fine.

kelsos (approved):
lgtm

PR #9874, opened 2025-05-02
Purge balances when marking as spam

yabirgb:
When the user marks an asset as spam it still appears in the cached response for the balance endpoint.

This change allows the frontend to trigger a balance query without doing remote calls, just querying the local cache of balances, and get the balances updated without the spam asset.

### DIFF
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -4990,6 +4990,25 @@ class RestAPI:
                 )
                 AssetResolver.clean_memory_cache(token.identifier)
 
+                # Finally if the token was in any cached balances, that cache is cleared,
+                # so it can be requeried later without the ignored token.
+                for balances in self.rotkehlchen.chains_aggregator.balances.get(
+                    chain=(blockchain := token.chain_id.to_blockchain()),
+                ).values():
+                    # variables below can only be dicts because we mark as spam only evm tokens
+                    in_assets = balances.assets.pop(token, None)  # type: ignore
+                    in_liabilities = balances.liabilities.pop(token, None)  # type: ignore
+
+                    if in_assets is not None or in_liabilities is not None:
+                        # If the token was found in balances, flush relevant caches so
+                        # subsequent calls reflect its removal without re-querying the chain.
+                        self.rotkehlchen.chains_aggregator.flush_cache('query_balances')
+                        self.rotkehlchen.chains_aggregator.flush_cache(
+                            name='query_balances',
+                            blockchain=blockchain,
+                        )
+                        log.debug(f'Flushed query_balances cache after setting {token} as spam')
+
         # add to ignored assets if it wasn't there
         self.rotkehlchen.data.add_ignored_assets(assets=tokens)
         return api_response(OK_RESULT, status_code=HTTPStatus.OK)
--- a/rotkehlchen/tests/api/test_assets.py
+++ b/rotkehlchen/tests/api/test_assets.py
@@ -7,7 +7,7 @@ import pytest
 import requests
 from polyleven import levenshtein
 
-from rotkehlchen.accounting.structures.balance import BalanceType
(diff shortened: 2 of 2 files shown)


rotkehlchen/api/rest.py
```diff
@@ -4993,6 +4993,22 @@ def add_tokens_to_spam(self, tokens: list[EvmToken]) -> Response:
                 )
                 AssetResolver.clean_memory_cache(token.identifier)
 
+                chain_balances = self.rotkehlchen.chains_aggregator.balances.get(
+                    chain=(blockchain := token.chain_id.to_blockchain()),
+                )
+                for balances in chain_balances.values():
+                    # those can only bee dicts because we mark as spam only evm tokens
```
prettyirrelevant, line 5000:
```suggestion
                    # those can only be dicts because we mark as spam only evm tokens
```

rotkehlchen/api/rest.py
```diff
@@ -4993,6 +4993,22 @@ def add_tokens_to_spam(self, tokens: list[EvmToken]) -> Response:
...
+                    in_assets = balances.assets.pop(token, None)  # type: ignore
+                    in_liabilities = balances.liabilities.pop(token, None)  # type: ignore
+
+                    if in_assets is not None or in_liabilities is not None:
+                        log.debug(f'Flushing cache after setting {token} as spam')
+                        self.rotkehlchen.chains_aggregator.flush_cache('query_balances')
+                        self.rotkehlchen.chains_aggregator.flush_cache(
+                            name='query_balances',
+                            blockchain=blockchain,
+                        )
```
prettyirrelevant, line 5010: is this not missing `self.flush_cache(f'query_{chain_key}_balances')`

rotkehlchen/api/rest.py
```diff
@@ -4993,6 +4993,22 @@ def add_tokens_to_spam(self, tokens: list[EvmToken]) -> Response:
                 )
                 AssetResolver.clean_memory_cache(token.identifier)
 
+                chain_balances = self.rotkehlchen.chains_aggregator.balances.get(
```
prettyirrelevant, line 4996: this variable is not used elsewhere so you can just use directly in the for loop

rotkehlchen/api/rest.py
yabirgb: no, intentionally I don't want to clear it to prevent a query to the balances endpoint from querying the blockchain

rotkehlchen/api/rest.py
```diff
@@ -4984,6 +4984,25 @@ def add_tokens_to_spam(self, tokens: list[EvmToken]) -> Response:
                 )
                 AssetResolver.clean_memory_cache(token.identifier)
 
+                for balances in self.rotkehlchen.chains_aggregator.balances.get(
+                    chain=(blockchain := token.chain_id.to_blockchain()),
+                ).values():
```
LefterisJP, line 4989: explain here what you do. Or in the function's docstring.

I do not understand why you are doing this here without a short explanation.

rotkehlchen/api/rest.py
```diff
@@ -4984,6 +4984,25 @@ def add_tokens_to_spam(self, tokens: list[EvmToken]) -> Response:
...
+                    # those can only be dicts because we mark as spam only evm tokens
+                    in_assets = balances.assets.pop(token, None)  # type: ignore
+                    in_liabilities = balances.liabilities.pop(token, None)  # type: ignore
+
+                    if in_assets is not None or in_liabilities is not None:
+                        # don't flush the cache of query_{blockchain}_balances since
+                        # we don't need to clear it and prevents us from querying the
```
LefterisJP, line 4996: Also here it's unclear: `and prevents us from querying the blockchain again` ... I don't understand what you are trying to say. Please write the comment in discord in English and try to explain to me what you want to say. Then we can just have it put here without going through a PR review cycle.

PR #9884, opened 2025-05-05
Add pytest-icdiff

LefterisJP:
When running tests it's often hard to see what's missing. I tried pytest-clarity which helped but its last commit was more than 4 years ago and it needed to be ran with `-vv` for helpful output.

Then I settled on pytest-icdiff which also has helpful output by default and its last commit is still not recent but closer (12/2023)


yabirgb (approved):
lgtm

PR #9887, opened 2025-05-05
Add new endpoint to query assets in a collection id

yabirgb:
With this endpoint the frontend will be able to retrieve the list of assets in a certain collection. It returns:

- 500 if server error
- 200 if found. The result is an empty list if the id is invalid


colibri/src/blockchain/node_inquirer.rs
kelsos, line 1: these are just stylistic changes right?

kelsos (approved):
looks good

PR #9888, opened 2025-05-05
Remove curve savings decoder

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=108905483

Removing the Curve Savings decoder since it already decodes as a yearn v3 vault. It would still be nice to have the counterparty set to curve on these though, so made a [note](https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=109112670) regarding that


LefterisJP (commented):
Have you also removed the `test_withdraw_from_crvusd_savings` and `test_deposit_into_crvusd_savings` from the cassetes repo?

LefterisJP (approved):
LGTM

PR #9894, opened 2025-05-05
calendar reminders are acknowledged before deletion

prettyirrelevant:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=109086785


rotkehlchen/tasks/calendar.py
```diff
@@ -91,10 +91,15 @@ def notify_reminders(
...
 def delete_past_calendar_entries(database: DBHandler) -> None:
-    """delete old calendar entries that the user has allowed to delete"""
+    """Delete past calendar events that are marked for auto-deletion,
+    but only if all associated reminders (if any) have been acknowledged."""
     now = ts_now()
     with database.conn.write_ctx() as write_cursor:
-        write_cursor.execute('DELETE FROM calendar WHERE timestamp < ? AND auto_delete=1', (now,))
+        write_cursor.execute(
+            'DELETE FROM calendar WHERE timestamp < ? AND auto_delete = 1 '
+            'AND identifier NOT IN (SELECT event_id FROM calendar_reminders WHERE is_acknowledged = 0);',  # noqa: E501
```
nicholasyoder, line 100:
```suggestion
            'AND identifier IN (SELECT event_id FROM calendar_reminders WHERE is_acknowledged = 1);',  # noqa: E501
```
maybe I'm missing something here, but wouldn't it be better to skip the `NOT` and simply check for `is_acknowledged = 1`?

rotkehlchen/tests/db/test_db_upgrades.py
```diff
@@ -3204,6 +3209,11 @@ def test_upgrade_db_47_to_48(user_data_dir, messages_aggregator):
             ('some-deposit-id',),
             ('some-withdrawal-id',),
         ]
+        assert cursor.execute('SELECT * from calendar_reminders').fetchall() == [
+            (1, 1, 3600, 0),
```
nicholasyoder, line 3213: Probably this should first assert on the existence of the `is_acknowledged` column. This is sort of checking that its there I guess since its checking for the extra `0` value, but its not actually checking the name of the column at all. Also I think adding an explicit check for the column name will help make it clearer what is being checked here for someone reading this code later.

rotkehlchen/db/schema.py
```diff
@@ -691,6 +691,7 @@
     identifier INTEGER PRIMARY KEY NOT NULL,
     event_id INTEGER NOT NULL,
     secs_before INTEGER NOT NULL,
+    is_acknowledged INTEGER NOT NULL CHECK (is_acknowledged IN (0, 1)) DEFAULT 0,
```
nicholasyoder, line 694: Actually I think this could just be `acknowledged` rather than `is_acknowledged` - not sure that the `is` really adds any meaning. Same for everywhere else `is_acknowledged` is being used in the PR.

rotkehlchen/tasks/calendar.py
```diff
@@ -63,7 +63,7 @@ class CalendarNotification(BaseReminderData):
     event: CalendarEntry
 
     def serialize(self) -> dict[str, str | int]:
-        return self.event.serialize()
+        return self.event.serialize() | {'reminder_identifier': self.identifier}
```
nicholasyoder, line 66: What is this change for? I'm not seeing where it is used at all, but maybe I'm missing something?

prettyirrelevant: Answered this in discord.
> the frontend will need the identifier of the reminder to update the acknowledged status of the reminder.

https://discord.com/channels/657906918408585217/1080783409640648765/1369294860112105534

LefterisJP (approved):
LGTM

PR #9898, opened 2025-05-06
Improve performance of balance query

yabirgb:
When we query protocol balances the filter was heavy and could be optimized. The main problem is that our filters don't allow to query pairs of [event_type, event_subtype] and we have a filter in python to discard the combinations that we don't need.

To improve the performance of this query the filter now includes all the [event_subtype] but not the combinations. This allows to improve the performance of this section since the for loop needs to process less events later.

Also I've adjusted the joins that are made when querying using EvmFilterQuery. This didn't affect in a clear manner but I believe is a good thing to have


rotkehlchen/chain/ethereum/interfaces/balances.py
```diff
@@ -91,7 +93,9 @@ def addresses_with_activity(
         db_filter = EvmEventFilterQuery.make(
             counterparties=[self.counterparty],
             products=products,
+            event_subtypes=[entry[1] for entry in event_types],
```
yabirgb, line 96: this line in the tests I've done made the query go from 5:20 mins to ~4:30

rotkehlchen/db/history_events.py
```diff
@@ -337,7 +337,12 @@ def _create_history_events_query(
             group_by_event_ids: bool = False,
     ) -> tuple[str, list]:
         """Returns the sql queries and bindings for the history events without pagination."""
-        base_suffix = f'{HISTORY_BASE_ENTRY_FIELDS}, {EVM_EVENT_FIELDS}, {ETH_STAKING_EVENT_FIELDS} {ALL_EVENTS_DATA_JOIN}'  # noqa: E501
+
+        if isinstance(filter_query, EvmEventFilterQuery):
+            base_suffix = f'{HISTORY_BASE_ENTRY_FIELDS}, {EVM_EVENT_FIELDS} {EVM_EVENT_JOIN}'
+        else:
+            base_suffix = f'{HISTORY_BASE_ENTRY_FIELDS}, {EVM_EVENT_FIELDS}, {ETH_STAKING_EVENT_FIELDS} {ALL_EVENTS_DATA_JOIN}'  # noqa: E501
```
yabirgb, line 344: this I don't believe improves the query a lot but it is less joins and columns in memory so I believe is a positive change

rotkehlchen/chain/ethereum/interfaces/balances.py
```diff
@@ -102,11 +106,11 @@ def addresses_with_activity(
 
         addresses_with_activity = defaultdict(list)
         for event in events:  # TODO: HistoryEventFilterquery should allow querying by type/subtype combos to avoid this  # noqa: E501
```
LefterisJP, line 108: Shouldn't you look at doing this TODO also though? Or at least make an issue/note about it? Is it hard?

LefterisJP (approved):
LGTM

PR #9899, opened 2025-05-06
Fix rainbow swap fee decoding

nicholasyoder:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=109177215

Rainbow swap fees are calculated using data from the internal transactions, but the app only queries internal transactions that are linked to user addresses, which omits the transactions needed here. The tests overlooked this since `get_decoded_events_of_transaction` queries all internal transactions for a given tx_hash.

This PR fixes this by making the rainbow decoder query all the internal transactions for the transactions it decodes using `get_and_ensure_internal_txns_of_parent_in_db` which is used somewhat similarly in several other decoders (although there were some changes required to that function as well).

### DIFF
--- a/rotkehlchen/chain/evm/transactions.py
+++ b/rotkehlchen/chain/evm/transactions.py
@@ -633,8 +633,8 @@ class EvmTransactions(ABC):  # noqa: B024
             self,
             tx_hash: 'EVMTxHash',
             chain_id: ChainID,
-            to_address: 'ChecksumEvmAddress',
             user_address: 'ChecksumEvmAddress',
+            to_address: 'ChecksumEvmAddress | None' = None,
             from_address: 'ChecksumEvmAddress | None' = None,
     ) -> list[EvmInternalTransaction]:
         """Queries the internal transactions of a parent tx_hash, saves them in the DB and returns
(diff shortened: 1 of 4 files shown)


rotkehlchen/chain/evm/transactions.py
```diff
@@ -644,15 +644,6 @@ def get_and_ensure_internal_txns_of_parent_in_db(
...
         - RemoteError if there is a problem querying the data sources or transaction hash does
         not exist."""
-        db_internal_txs = self.dbevmtx.get_evm_internal_transactions(
-            parent_tx_hash=tx_hash,
-            blockchain=CHAINID_TO_SUPPORTED_BLOCKCHAIN[self.evm_inquirer.chain_id],
-            from_address=from_address,
-            to_address=to_address,
-        )
-        if len(db_internal_txs) > 0:
-            return db_internal_txs
```
nicholasyoder, line 654: Removed this since in the app there can be some internal transactions present for a given tx_hash without having them all present. So it needs to only skip the query if we have saved in the cache that it has already had all its internal txns pulled.

LefterisJP (approved):
LGTM

PR #9900, opened 2025-05-06
Allow to query with combined filter of type and subtype

yabirgb:
- allows to query pairs of type AND subtype

### DIFF
--- a/rotkehlchen/db/filtering.py
+++ b/rotkehlchen/db/filtering.py
@@ -1,6 +1,6 @@
 import logging
 from abc import ABC, abstractmethod
-from collections.abc import Collection, Sequence
+from collections.abc import Collection, Iterable, Sequence
 from dataclasses import dataclass, field
 from enum import Enum, auto
 from typing import Any, Generic, Literal, NamedTuple, Self, TypeVar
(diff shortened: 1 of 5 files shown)


rotkehlchen/db/filtering.py
```diff
@@ -389,7 +389,7 @@ def create(
             group_by = DBFilterGroupBy(field_name=group_by_field)
 
         order_by = None
-        if order_by_rules is not None:
+        if order_by_rules is not None and len(order_by_rules) != 0:
```
LefterisJP, line 392: why this change? Why not keep it to None or a full list as possible types?

rotkehlchen/chain/evm/decoding/thegraph/balances.py
```diff
@@ -49,6 +49,7 @@ def _get_staking_events(self, location: Location | None = None) -> list[EvmEvent
             counterparties=[self.counterparty],
             products=[EvmProduct.STAKING],
             location=location,
+            order_by_rules=[],
```
LefterisJP, line 52: why this change?

LefterisJP (approved):
LGTM

PR #9901, opened 2025-05-06
Improve pendle event ordering

nicholasyoder:


rotkehlchen/tests/unit/decoders/test_pendle.py
```diff
@@ -225,6 +225,64 @@ def test_buy_pt(ethereum_inquirer, ethereum_accounts):
     assert events == expected_events
 
 
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('ethereum_accounts', [['0xFd83CCCecef02a334e6A86e7eA8D0aa0F61f1Faf']])
+def test_sell_pt(ethereum_inquirer, ethereum_accounts, pendle_cache):
```
LefterisJP, line 230: I know they may use the sell lingo in some places, but it's mostly return/withdraw also in our types so I think the name should be changed. Also remember to adjust the cassete name too.

```suggestion
def test_return_pt(ethereum_inquirer, ethereum_accounts, pendle_cache):
```

LefterisJP (approved):
LGTM

PR #9905, opened 2025-05-07
endpoint to get token balance of an addy

prettyirrelevant:
Closes https://github.com/orgs/rotki/projects/11/views/2?pane=issue&itemId=109454548

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -14025,3 +14025,62 @@ Active management
     :statuscode 200: Interaction check successful.
     :statuscode 400: Malformed request.
     :statuscode 500: Internal Rotki error.
+
+  .. http:post:: /api/(version)/wallet/balance
+
+    Queries the token or native asset balance for a given EVM chain and address.
+
+   .. note::
+      This endpoint can also be queried asynchronously by using ``"async_query": true``.
+
+    **Example Request (Native Balance):**
+
+      .. http:example:: curl wget httpie python-requests
+
+        POST /api/1/wallet/balance HTTP/1.1
+        Host: localhost:5042
+        Content-Type: application/json;charset=UTF-8
+
+        {
+          "evm_chain": "ethereum",
+          "address": "0xc37b40ABdB939635068d3c5f13E7faF686F03B65",
+          "asset": "ETH",
+          "async_query": false
+        }
+
+    **Example Request (ERC20 Balance):**
+
+      .. http:example:: curl wget httpie python-requests
+
+        POST /api/1/wallet/balance HTTP/1.1
+        Host: localhost:5042
+        Content-Type: application/json;charset=UTF-8
+
+        {
+          "evm_chain": "ethereum",
+          "address": "0xc37b40ABdB939635068d3c5f13E7faF686F03B65",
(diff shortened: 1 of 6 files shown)


rotkehlchen/api/rest.py
```diff
@@ -5420,3 +5421,29 @@ def prepare_native_transfer(
             return wrap_in_fail_result(str(e), status_code=HTTPStatus.BAD_REQUEST)
 
         return _wrap_in_ok_result(result=payload)
+
+    @async_api_call()
+    def fetch_token_balance_for_address(
+            self,
+            address: ChecksumEvmAddress,
+            evm_chain: EVM_CHAIN_IDS_WITH_TRANSACTIONS_TYPE,
+            asset: 'CryptoAsset',
```
yabirgb, line 5430: mmmm the chain is needed if you want to query the native token but if the query is for a token should you check that the chains match? or is it an overkill?

rotkehlchen/api/rest.py
```diff
@@ -5420,3 +5421,29 @@ def prepare_native_transfer(
...
+                    asset=asset,
+                    amount=node_inquirer.call_contract(
+                        contract_address=asset.resolve_to_evm_token().evm_address,
+                        abi=node_inquirer.contracts.erc20_abi,
+                        method_name='balanceOf',
+                        arguments=[address],
+                    ),
+                )
+        except RemoteError as e:
+            return wrap_in_fail_result(str(e), status_code=HTTPStatus.INTERNAL_SERVER_ERROR)
```
yabirgb, line 5447: double check it but I believe we don't raise internal server errors for remoteError.It is 409 conflict irc

docs/api.rst
```diff
@@ -14025,3 +14025,61 @@ Active management
...
+    **Example Response:**
+
+      .. sourcecode:: http
+
+        HTTP/1.1 200 OK
+        Content-Type: application/json
+
+        {
+          "result": "0",
+          "message": ""
```
yabirgb, line 14079: the result is with the decimals adjusted right? Point that in the in the doc for result

rotkehlchen/api/rest.py
prettyirrelevant: i don't think it's needed since nothing gets stored in the database. the call will simply fail.

rotkehlchen/api/rest.py
prettyirrelevant: we raise both across the codebase and even `400 BAD REQUEST` in some scenario.

rotkehlchen/api/rest.py
```diff
@@ -5420,3 +5424,31 @@ def prepare_native_transfer(
...
+                    token_amount=node_inquirer.call_contract(
+                        contract_address=token.evm_address,
+                        abi=node_inquirer.contracts.erc20_abi,
+                        method_name='balanceOf',
+                        arguments=[address],
+                    ),
+                    token=token,
+                )
+            else:
+                raise InputError('Token exists on different chain than requested')
```
LefterisJP, line 5450: why raise InputError? Isn't this the place to actually return directly? Never use exceptions for control flow like this

LefterisJP (approved):
LGTM

PR #9907, opened 2025-05-07
feat(frontend): acknowledge reminder

lukicenturi:
Closes #(issue_number)

## Checklist

### DIFF
--- a/frontend/app/src/composables/use-pagination-filter/index.ts
+++ b/frontend/app/src/composables/use-pagination-filter/index.ts
@@ -33,7 +33,7 @@ interface UsePaginationFiltersOptions<
   locationOverview?: Ref<string>;
   filterSchema?: () => FilterSchema<TFilter, TSuggestionMatcher>;
   onUpdateFilters?: (query: LocationQuery) => void;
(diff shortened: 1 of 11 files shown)


frontend/app/src/composables/use-pagination-filter/index.ts
```diff
@@ -33,7 +33,7 @@ interface UsePaginationFiltersOptions<
   locationOverview?: Ref<string>;
   filterSchema?: () => FilterSchema<TFilter, TSuggestionMatcher>;
   onUpdateFilters?: (query: LocationQuery) => void;
-  extraParams?: ComputedRef<RawLocationQuery>;
+  extraParams?: ComputedRef<RawLocationQuery> | Ref<RawLocationQuery>;
```
kelsos, line 36: Any reason for this change?

lukicenturi: I wanted to wrap the extraParams inside refDebounced, so if user click `next month` too fast, it's efficient.
But nevermind, we can remove it, I can refDebounced the dependency instead

PR #9909, opened 2025-05-07
fix(frontend): fetch asset amount separately for onchain feature

lukicenturi:


frontend/app/src/modules/onchain/send/TradeAssetDisplay.vue
```diff
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import type { TradableAsset } from '@/modules/onchain/types';
+import type BigNumber from 'bignumber.js';
```
kelsos, line 3: I think I also introduced a few of these accidentally, due to auto-complete.

I wonder if I can make a custom rule at some point.

lukicenturi: Probably we can use this? https://eslint.org/docs/latest/rules/no-restricted-imports#paths

kelsos (approved):
lgtm thanks :)

Issue #9875, opened 2025-05-02
Make rotki Pectra ready


LefterisJP:
## Abstract

The Pectra hard fork is coming on Wednesday, the 7th of May. We are also close to one of our biggest releases, 1.39.0.

It would make little sense to ignore Pectra changes since they are happening. We should check what's needed to be changed, adjusted or added to rotki and add it for 1.39 too.

## Links

Some links with explanations:

- [Easier explanations](https://www.coinbase.com/en-de/blog/the-ultimate-guide-to-ethereums-pectra-upgrade)
- [Bit more technical](https://mixbytes.io/blog/the-prague-electra-pectra-hardfork-explained)

## Task

Understand what we need to adjust, split it into bullet points and subtasks here in the issue and do them.

- [ ] Task 1 (edit me)
- [ ] Task 2 (edit me .. etc -- add more if required)

LefterisJP:
This is now done.

PR #9911, opened 2025-05-07
Decode eth2 consolidation requests

nicholasyoder:
Related: #9875

### DIFF
--- a/rotkehlchen/chain/ethereum/modules/eth2/decoder.py
+++ b/rotkehlchen/chain/ethereum/modules/eth2/decoder.py
@@ -1,8 +1,6 @@
 import logging
 from typing import Any
 
-from eth_utils import encode_hex
-
 from rotkehlchen.chain.ethereum.constants import ETH2_DEPOSIT_ADDRESS
 from rotkehlchen.chain.evm.decoding.interfaces import DecoderInterface
 from rotkehlchen.chain.evm.decoding.structures import (
@@ -12,6 +10,7 @@ from rotkehlchen.chain.evm.decoding.structures import (
 )
 from rotkehlchen.chain.evm.decoding.types import CounterpartyDetails
 from rotkehlchen.chain.evm.types import string_to_evm_address
+from rotkehlchen.constants import ZERO
 from rotkehlchen.constants.assets import A_ETH
 from rotkehlchen.errors.misc import RemoteError
 from rotkehlchen.externalapis.beaconchain.service import BeaconChain
@@ -19,9 +18,9 @@ from rotkehlchen.history.events.structures.eth2 import EthDepositEvent
 from rotkehlchen.history.events.structures.types import HistoryEventSubType, HistoryEventType
 from rotkehlchen.logging import RotkehlchenLogsAdapter
 from rotkehlchen.types import ChecksumEvmAddress, Eth2PubKey
-from rotkehlchen.utils.misc import from_gwei
+from rotkehlchen.utils.misc import bytes_to_address, bytes_to_hexstr, from_gwei
 
-from .constants import CPT_ETH2, UNKNOWN_VALIDATOR_INDEX
+from .constants import CONSOLIDATION_REQUEST_CONTRACT, CPT_ETH2, UNKNOWN_VALIDATOR_INDEX
 
 DEPOSIT_EVENT = b'd\x9b\xbcb\xd0\xe3\x13B\xaf\xeaN\\\xd8-@I\xe7\xe1\xee\x91/\xc0\x88\x9a\xa7\x90\x80;\xe3\x908\xc5'  # noqa: E501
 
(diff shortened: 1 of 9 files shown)


rotkehlchen/chain/ethereum/modules/eth2/decoder.py
```diff
@@ -94,11 +105,54 @@ def _decode_eth2_deposit_event(self, context: DecoderContext) -> DecodingOutput:
...
 
+    def _decode_eth2_consolidation_request(self, context: DecoderContext) -> DecodingOutput:
+        """Decode a request to consolidate validators.
+        Handles two types of requests:
+        - Self consolidation - changes the withdrawal credentials to 0x02 (accumulating validator)
+          See https://epf.wiki/?ref=blog.obol.org#/wiki/pectra-faq?id=q-i-have-a-validator-with-0x01-credentials-how-do-i-move-to-0x02
+        - Normal consolidation - consolidates the source validator into the target validator
+        """
+        if len(context.tx_log.data) != 116:  # This an anonymous tx log so can't check topic[0]
+            return DEFAULT_DECODING_OUTPUT
```
yabirgb, line 116: can you add a log warning here? Also this only gets executed by the contract right? So there is no need to check again the address

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
```diff
@@ -94,11 +105,54 @@ def _decode_eth2_deposit_event(self, context: DecoderContext) -> DecodingOutput:
...
+                event.event_type == HistoryEventType.SPEND and
+                event.event_subtype == HistoryEventSubType.NONE and
+                event.asset == A_ETH
+            ):
+                event.event_subtype = HistoryEventSubType.FEE
+                event.counterparty = CPT_ETH2
+                event.notes = f'Spend {event.amount} ETH as validator consolidation fee'
+
+        caller_address = bytes_to_address(b'\x00' * 12 + context.tx_log.data[:20])
+        source_public_key = Eth2PubKey(encode_hex(context.tx_log.data[20:68]))
```
yabirgb, line 130: I know we had it before but we have `bytes_to_hexstr` that does the same. Maybe we can get rid of one of them but I would use `bytes_to_hexstr` here because it raises deserializationError and the other I suspect raises ValueError

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
```diff
@@ -94,11 +105,54 @@ def _decode_eth2_deposit_event(self, context: DecoderContext) -> DecodingOutput:
...
+            timestamp=context.transaction.timestamp,
+            event_type=HistoryEventType.INFORMATIONAL,
+            event_subtype=HistoryEventSubType.NONE,
+            asset=A_ETH,
+            amount=ZERO,
+            location_label=caller_address,
+            notes=(
+                f'Request to convert validator {target_idx_or_key} into an accumulating validator'
+                if source_public_key == target_public_key else
+                f'Request to consolidate validator {self._get_index_or_public_key(source_public_key)} into {target_idx_or_key}'  # noqa: E501
```
yabirgb, line 144: question: can you make `_get_index_or_public_key` able to take a list of public keys. If they are in the database you do a single request instead of two and if you need to query beaconchain then `get_validator_data` allows to query using a list instead so you will just need to process a bit the output. In the best scenario you are saving an extra query to the db and in the worst case you save 1 request to beaconchain (that costs money under certain conditions)

yabirgb (commented):
it is missing the changelog entry but I'm not sure if we want to have it when everything for pectra is ready

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
nicholasyoder: > this only gets executed by the contract right?

yep, its only called via the addresses_to_decoders mapping where its mapped to CONSOLIDATION_REQUEST_CONTRACT

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
```diff
@@ -31,39 +30,54 @@
...
+                for result in beaconchain.get_validator_data(unknown_keys):
+                    validator_index = result['validatorindex']
+                    if isinstance(validator_index, int) and validator_index >= 0:
+                        found_indexes[result['pubkey']] = validator_index
+            except RemoteError as e:
+                log.error(f'Failed to query validator index for {unknown_keys} due to {e!s}')
+
+        return [found_indexes.get(key, UNKNOWN_VALIDATOR_INDEX) for key in public_keys]
+
+    def _get_indexes_or_public_keys(self, public_keys: list[Eth2PubKey]) -> list[int | Eth2PubKey]:
```
LefterisJP, line 62: don't we already have this somewhere else?

LefterisJP (commented):
Left some questions.

Also the questions that pop up in my head are:

1. These are request for consolidation. As a user experence when should we delete validators from the DB and stop tracking them? Should we do it later when they are consolidated? How do detect consolidation has happened?
2. Shouldn't we also mark them as distributing (0x1) or accumulating (0x2) in the DB now?
3. Also speaking for what to mark in the DB is the effective balance when you change to 0x2 set to 2048? Or is this something you can configure on your own?

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
nicholasyoder: No, not as far as I can tell.

LefterisJP (approved):
LGTM

PR #9912, opened 2025-05-07
fix: explicitly specify global scope for i18n

kelsos:
By default vue-i18n first checks local scope which we do not use at all.

Closes #(issue_number)

## Checklist


lukicenturi (approved):
LGTM!

PR #9920, opened 2025-05-09
decode eip7702 transactions

prettyirrelevant:
Closes https://github.com/orgs/rotki/projects/11?pane=issue&itemId=109469245


rotkehlchen/tests/unit/test_evm_tx_decoding.py
```diff
@@ -327,3 +329,41 @@ def test_token_detection_after_decoding(
...
+            sequence_index=1,
+            timestamp=timestamp,
+            location=Location.ETHEREUM,
+            event_type=HistoryEventType.INFORMATIONAL,
+            event_subtype=HistoryEventSubType.NONE,
+            asset=A_ETH,
+            amount=ZERO,
+            location_label=user_address,
+            address=string_to_evm_address('0x775c8D470CC8d4530b8F233322480649f4FAb758'),
+            notes='Successfully executed eip-7702 transaction by delegate 0x775c8D470CC8d4530b8F233322480649f4FAb758',  # noqa: E501
```
yabirgb, line 366: I find this confusing. Something better would be

```suggestion
            notes='Executed account delegation to 0x775c8D470CC8d4530b8F233322480649f4FAb758',  # noqa: E501
```

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -793,6 +794,29 @@ def _decode_transaction(
         if process_swaps:
             events = self._process_swaps(transaction=transaction, decoded_events=events)
 
+        if (
+                tx_receipt.tx_type == 4 and
+                transaction.from_address == transaction.to_address and
+                transaction.authorization_list is not None
+        ):  # if the event is an eip-7702 transaction, we need to add an informational event.
+            if (delegated_address := transaction.authorization_list[-1].delegated_address) == ZERO_ADDRESS:  # noqa: E501
+                notes = 'Successfully revoked eip-7702 transaction authorization'
```
yabirgb, line 803: my problem with having eip-7702 is that people forget what those are and also most of the users don't know what it is so I find it confusing to have in the notes. I belive they key information here is that you either revoked or delegatd

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -793,6 +794,29 @@ def _decode_transaction(
         if process_swaps:
             events = self._process_swaps(transaction=transaction, decoded_events=events)
 
+        if (
+                tx_receipt.tx_type == 4 and
+                transaction.from_address == transaction.to_address and
+                transaction.authorization_list is not None
+        ):  # if the event is an eip-7702 transaction, we need to add an informational event.
+            if (delegated_address := transaction.authorization_list[-1].delegated_address) == ZERO_ADDRESS:  # noqa: E501
```
yabirgb, line 802: can it be empty by some kind of error? `-1` would raise index error here. Is it guaranteed somewhere else to not be empty?

rotkehlchen/chain/evm/transactions.py
```diff
@@ -604,12 +604,14 @@ def ensure_tx_data_exists(
...
         not exist.
         """
-        query, bindings = EvmTransactionsFilterQuery.make(tx_hash=tx_hash, chain_id=self.evm_inquirer.chain_id).prepare()  # noqa: E501
-        query, bindings = self.dbevmtx._form_evm_transaction_dbquery(query=query, bindings=bindings, has_premium=True)  # noqa: E501
-        tx_data = cursor.execute(query, bindings).fetchone()
+        transactions = self.dbevmtx.get_evm_transactions(
+            cursor=cursor,
+            filter_=EvmTransactionsFilterQuery.make(tx_hash=tx_hash, chain_id=self.evm_inquirer.chain_id),  # noqa: E501
+            has_premium=True,
+        )
```
yabirgb, line 611: shouldn't you query inside the if statement once you know that the transaction is in the database? You don't seem to use this query result more

rotkehlchen/db/schema.py
```diff
@@ -328,6 +328,16 @@
 );
 """
 
+DB_CREATE_EVM_TRANSACTIONS_AUTHORIZATION = """
+CREATE TABLE IF NOT EXISTS evm_transactions_authorization (
+    tx_hash BLOB NOT NULL,
+    chain_id INTEGER NOT NULL,
+    nonce INTEGER NOT NULL,
+    delegated_address TEXT NOT NULL,
+    FOREIGN KEY(tx_hash, chain_id) REFERENCES evm_transactions(tx_hash, chain_id) ON DELETE CASCADE
```
yabirgb, line 337: shouldn't tx_hash + chain_id be primary keys here?

rotkehlchen/chain/evm/decoding/decoder.py
```diff
@@ -793,6 +794,29 @@ def _decode_transaction(
...
+
+            events.append(self.base.make_event(
+                tx_hash=transaction.tx_hash,
+                sequence_index=self.base.get_next_sequence_index(),
+                timestamp=transaction.timestamp,
+                event_type=HistoryEventType.INFORMATIONAL,
+                event_subtype=HistoryEventSubType.NONE,
+                asset=self.value_asset,
+                amount=ZERO,
+                location_label=transaction.from_address,
```
yabirgb, line 815: mmmm I have a question. Reading about the EIP it says that one can sign the transaction and then anyone can upload it
image

is correct to use the from_address here? https://etherscan.io/tx/0x42402dcf6658abaf2c47593a7ebe1264fb2f331de918239d1717a7a9d2996abf#authorizationlist etherscan seems to have it but I'm not sure if it is the from or they use something else

yabirgb: edit: I believe it is correct. It is the tx.origin but it is always a transaction `from delegator address to delegator address`

rotkehlchen/db/evmtx.py
```diff
@@ -129,6 +130,17 @@ def add_evm_transactions(
...
+            (tx.tx_hash, tx.chain_id.serialize_for_db(), entry.nonce, entry.delegated_address)
+            for tx in evm_transactions if tx.authorization_list is not None
+            for entry in tx.authorization_list
+        ]
+        if len(authorization_list) > 0:
+            write_cursor.executemany(
+                'INSERT INTO evm_transactions_authorization(tx_hash, chain_id, nonce, delegated_address) '  # noqa: E501
+                'VALUES (?, ?, ?, ?)',
+                authorization_list,
+            )
```
yabirgb, line 143: if you change the schema to have the primary key I believe you need to handle possible errors here

rotkehlchen/db/evmtx.py
yabirgb, line 1: you are missing that when we delete a transaction you need to delete it from the new table that you created too

rotkehlchen/db/evmtx.py
```diff
@@ -216,21 +228,29 @@ def get_evm_transactions(
...
+            if auth[0] is not None:  # Only add authorizations if exists
+                tx_map[tx_key][13].append(auth)
 
         evm_transactions = []
-        for result in results:
+        for tx_data in tx_map.values():
             try:
-                tx = self._build_evm_transaction(result)
+                tx = self._build_evm_transaction(tx_data[:-1], tx_data[-1])
+                evm_transactions.append(tx)
```
yabirgb, line 246:
```suggestion
                evm_transactions.append(self._build_evm_transaction(tx_data[:-1], tx_data[-1]))
```

rotkehlchen/db/evmtx.py
```diff
@@ -216,21 +228,29 @@ def get_evm_transactions(
...
+        tx_map = {}  # Group results by transaction (tx_hash + chain_id)
+        for result in cursor.execute(query, bindings):
+            tx_key = (result[0], result[1])  # (tx_hash, chain_id)
+            auth = (result[13], result[14])  # (nonce, delegated_address)
+
+            if tx_key not in tx_map:  # Store base transaction data + empty auth list
+                tx_map[tx_key] = (*result[:13], [])
+
+            if auth[0] is not None:  # Only add authorizations if exists
+                tx_map[tx_key][13].append(auth)
```
yabirgb, line 240: why do you use a map? Why not use a list and avoid accessing twice the mapping by just adding the final thing? Why can transactions appear duplicated?

```suggestion
        tx_map = {}  # Group results by transaction (tx_hash + chain_id)
        for result in cursor.execute(query, bindings):
            if esult[13] is not None:  # Only add authorizations if exists
                auth = (result[13], result[14])  # (nonce, delegated_address)
                tx_map.append(*result[:13], [auth])
            else:
                tx_map.append((*result[:13], []))

```

rotkehlchen/db/evmtx.py
```diff
@@ -686,20 +706,20 @@ def get_or_create_genesis_transaction(
...
-        )
+            sql = f'{base_select} FROM evm_transactions {join_clause} {query}'
+        else:
+            sql = f'{base_select} FROM (SELECT * FROM evm_transactions ORDER BY timestamp DESC LIMIT ?) AS evm_transactions {join_clause} {query}'  # noqa: E501
+            bindings = [FREE_ETH_TX_LIMIT] + bindings
 
-    def _build_evm_transaction(self, result: tuple[Any, ...]) -> EvmTransaction:
+        return sql, bindings
+
+    def _build_evm_transaction(self, result: tuple[Any, ...], authorization_list_result: list[tuple[Any, ...]]) -> EvmTransaction:  # noqa: E501
```
yabirgb, line 722: authorization list is a pair of integer and string, can you properly type it?

rotkehlchen/db/evmtx.py
```diff
@@ -686,20 +706,20 @@ def get_or_create_genesis_transaction(
                 )
         return tx
 
-    def _form_evm_transaction_dbquery(self, query: str, bindings: list[Any], has_premium: bool) -> tuple[str, list[tuple]]:  # noqa: E501
+    def _form_evm_transaction_dbquery(self, query: str, bindings: list[Any], has_premium: bool) -> tuple[str, list]:  # noqa: E501
         """Return query and bindings for the evm_transaction database table"""
+        base_select = 'SELECT evm_transactions.tx_hash, evm_transactions.chain_id, timestamp, block_number, from_address, to_address, value, gas, gas_price, gas_used, input_data, evm_transactions.nonce, identifier, auth.nonce AS auth_nonce, auth.delegated_address'  # noqa: E501
+        join_clause = 'LEFT JOIN evm_transactions_authorization auth ON evm_transactions.tx_hash = auth.tx_hash AND evm_transactions.chain_id = auth.chain_id'  # noqa: E501
```
yabirgb, line 712: what is the change in this function? I don't see it clearly

rotkehlchen/serialization/deserialize.py
```diff
@@ -534,6 +535,16 @@ def deserialize_evm_transaction(
...
 
+        authorization_list = None
+        if (raw_authorization_list := data.get('authorizationList')) is not None:
+            authorization_list = [
+                EvmTransactionAuthorization(
+                    nonce=read_integer(entry, 'nonce', source),
+                    delegated_address=deserialize_evm_address(entry['address']),
+                )
+                for entry in raw_authorization_list
+            ]
```
yabirgb, line 546: those can raise deserializationError, I believe it would be best to make it a normal loop and handle the potential errors

rotkehlchen/types.py
```diff
@@ -354,6 +354,12 @@ class EvmlikeChain(StrEnum):
     ZKSYNC_LITE = auto()
 
 
+@dataclass(frozen=True)
+class EvmTransactionAuthorization:
+    nonce: int
+    delegated_address: ChecksumEvmAddress
```
yabirgb, line 360: why a dataclass and not a named tuple?

yabirgb: add a docstring exmplaining what it is

rotkehlchen/db/arbitrum_one_tx.py
```diff
@@ -42,4 +42,8 @@ def _build_evm_transaction(self, result: tuple[Any, ...]) -> ArbitrumOneTransact
             nonce=result[11],
             tx_type=tx_receipt.tx_type,
             db_id=result[12],
+            authorization_list=None if len(authorization_list_result) == 0 else [
+                EvmTransactionAuthorization(nonce=entry[0], delegated_address=entry[1])
+                for entry in authorization_list_result
+            ],
```
yabirgb, line 48: why receive a list instead of a nullable list? `list | None`. Does it come from the db query? If you do LEFT join doesn't it come as None here?

rotkehlchen/db/evmtx.py
prettyirrelevant: i added a join to include the authorizations for each transaction. broke it into multiple lines to make it more readable.

rotkehlchen/db/arbitrum_one_tx.py
prettyirrelevant: > Does it come from the db query?

yes, it does. it comes as [(None, None), ...] since a transaction can have > 1 authorization.

rotkehlchen/db/evmtx.py
prettyirrelevant: > Why not use a list and avoid accessing twice the mapping by just adding the final thing? 

because a transaction can have more than one authorization.

> Why can transactions appear duplicated?

same reason has above. since authorizations can be >1, the transaction data will be duplicated for each row.



unless i'm missing something here.

rotkehlchen/db/evmtx.py
prettyirrelevant: i don't think that's needed as we defined a foreign key constraint "ON DELETE CASCADE"

rotkehlchen/db/schema.py
prettyirrelevant: it cannot be just tx_hash + chain id as a transaction can have multiple authorizations. so i included the delegated_address.

rotkehlchen/chain/evm/decoding/decoder.py
prettyirrelevant: at every point, if the number of entries is zero, i set it to `None`. i can also add a `len(transaction.authorization_list) > 0` condition here if it makes sense

rotkehlchen/chain/evm/l2_with_l1_fees/transactions.py
```diff
@@ -71,7 +70,21 @@ def ensure_tx_data_exists(
                 'INSERT OR IGNORE INTO optimism_transactions(tx_id, l1_fee) VALUES(?, ?)',
                 (tx_id, str(transaction.l1_fee)),
             )
-        query, bindings = EvmTransactionsFilterQuery.make(tx_hash=tx_hash, chain_id=self.evm_inquirer.chain_id).prepare()  # noqa: E501
```
LefterisJP, line 74: why are you removing this query?

rotkehlchen/db/evmtx.py
```diff
@@ -67,6 +68,8 @@
 
 
 class DBEvmTx:
+    # Index in the SQL result tuple where authorization fields (nonce, delegated_address) begin  # noqa: E501
+    authorization_data_start_index = 13
```
LefterisJP, line 72: This gets created again and again for each class instance. Do not do that. Make it into a file lever constant :Final

rotkehlchen/db/evmtx.py
```diff
@@ -686,20 +708,28 @@ def get_or_create_genesis_transaction(
...
+            sql = f'{base_select} FROM evm_transactions {join_clause} {query}'
+        else:
+            sql = f'{base_select} FROM (SELECT * FROM evm_transactions ORDER BY timestamp DESC LIMIT ?) AS evm_transactions {join_clause} {query}'  # noqa: E501
+            bindings = [FREE_ETH_TX_LIMIT] + bindings
+
+        return sql, bindings
 
-    def _build_evm_transaction(self, result: tuple[Any, ...]) -> EvmTransaction:
+    def _build_evm_transaction(self, result: tuple[Any, ...], authorization_list_result: list[tuple[int, ChecksumEvmAddress]]) -> EvmTransaction:  # noqa: E501
         """Build a transaction object from queried data
```
LefterisJP, line 733: explain a bit here what kind of input data it should get.

rotkehlchen/db/evmtx.py
```diff
@@ -686,20 +708,28 @@ def get_or_create_genesis_transaction(
                 )
         return tx
 
-    def _form_evm_transaction_dbquery(self, query: str, bindings: list[Any], has_premium: bool) -> tuple[str, list[tuple]]:  # noqa: E501
+    def _form_evm_transaction_dbquery(self, query: str, bindings: list[Any], has_premium: bool) -> tuple[str, list]:  # noqa: E501
```
LefterisJP, line 711: update docstring here with the new details since this no longer concerns only one table

rotkehlchen/db/schema.py
```diff
@@ -328,6 +328,17 @@
 );
 """
 
+DB_CREATE_EVM_TRANSACTIONS_AUTHORIZATION = """
+CREATE TABLE IF NOT EXISTS evm_transactions_authorization (
```
LefterisJP, line 332: Just like transactions is plural

```suggestion
DB_CREATE_EVM_TRANSACTION_AUTHORIZATIONS = """
CREATE TABLE IF NOT EXISTS evm_transaction_authorizations (
```

rotkehlchen/db/schema.py
```diff
@@ -328,6 +328,17 @@
...
+    tx_hash BLOB NOT NULL,
+    chain_id INTEGER NOT NULL,
+    nonce INTEGER NOT NULL,
+    delegated_address TEXT NOT NULL,
+    PRIMARY KEY(tx_hash, chain_id, delegated_address),
+    FOREIGN KEY(tx_hash, chain_id) REFERENCES evm_transactions(tx_hash, chain_id) ON DELETE CASCADE
```
LefterisJP, line 338: No no no. I told you it should be referencing the transaction table. Why complicate things with composite primary keys and composite foreign key? Not only is this slower but it takes a lot more space.

```suggestion
    identifier INTEGER NOT NULL PRIMARY KEY,
    tx_id INTEGER
    nonce INTEGER NOT NULL,
    delegated_address TEXT NOT NULL,
    FOREIGN KEY(tx_id) REFERENCES evm_transactions(identifier) ON DELETE CASCADE
```

rotkehlchen/db/evmtx.py
```diff
@@ -216,21 +231,28 @@ def get_evm_transactions(
         """
         query, bindings = filter_.prepare()
         query, bindings = self._form_evm_transaction_dbquery(query, bindings, has_premium)
-        results = cursor.execute(query, bindings)
+        tx_map: dict[tuple[bytes, int], tuple[Any, ...]] = {}  # Group results by transaction (tx_hash + chain_id)  # noqa: E501
```
LefterisJP, line 234: I don't understand what you are doing here and why. You don't need to use chain id / tx_hash as key ... you have the unique transaction identifier.

Please simplify this

rotkehlchen/db/evmtx.py
```diff
@@ -686,20 +708,28 @@ def get_or_create_genesis_transaction(
...
-            return (
-                'SELECT DISTINCT evm_transactions.tx_hash, evm_transactions.chain_id, timestamp, block_number, from_address, to_address, value, gas, gas_price, gas_used, input_data, nonce, identifier FROM evm_transactions ' + query,  # noqa: E501
-                bindings,
-            )
-        # else
-        return (
-            'SELECT DISTINCT evm_transactions.tx_hash, evm_transactions.chain_id, timestamp, block_number, from_address, to_address, value, gas, gas_price, gas_used, input_data, nonce, identifier FROM (SELECT * from evm_transactions ORDER BY timestamp DESC LIMIT ?) AS evm_transactions ' + query,  # noqa: E501
-            [FREE_ETH_TX_LIMIT] + bindings,
+        base_select = (
+            'SELECT evm_transactions.tx_hash, evm_transactions.chain_id, '
```
LefterisJP, line 714: simplify the query using the tx id

rotkehlchen/chain/evm/transactions.py
```diff
@@ -625,9 +626,26 @@ def ensure_tx_data_exists(
                 data=raw_receipt_data,
             )
 
-        tx_data = cursor.execute(query, bindings).fetchone()
```
LefterisJP, line 628: same question here. Why remove this query?

rotkehlchen/db/l2withl1feestx.py
```diff
@@ -43,18 +49,29 @@ def add_evm_transactions(
...
     def _form_evm_transaction_dbquery(self, query: str, bindings: list[Any], has_premium: bool) -> tuple[str, list[tuple]]:  # noqa: E501
+        base_select = (
+            'SELECT evm_transactions.tx_hash, evm_transactions.chain_id, evm_transactions.timestamp, '  # noqa: E501
+            'evm_transactions.block_number, evm_transactions.from_address, evm_transactions.to_address, '  # noqa: E501
+            'evm_transactions.value, evm_transactions.gas, evm_transactions.gas_price, evm_transactions.gas_used, '  # noqa: E501
+            'evm_transactions.input_data, evm_transactions.nonce, OP.l1_fee, evm_transactions.identifier, auth.nonce AS auth_nonce, auth.delegated_address'  # noqa: E501
+        )
+        join_clause = (
+            'LEFT JOIN optimism_transactions AS OP ON evm_transactions.identifier=OP.tx_id '
+            'LEFT JOIN evm_transactions_authorization AS auth ON evm_transactions.tx_hash = auth.tx_hash AND evm_transactions.chain_id = auth.chain_id'  # noqa: E501
```
LefterisJP, line 60: simplify this now with the schema change

rotkehlchen/serialization/deserialize.py
```diff
@@ -534,6 +535,18 @@ def deserialize_evm_transaction(
...
         value = read_integer(data, 'value', source)
 
+        authorization_list = None
+        if (raw_authorization_list := data.get('authorizationList')) is not None:
+            authorization_list = []
+            for entry in raw_authorization_list:
+                try:
+                    authorization_list.append(EvmTransactionAuthorization(
+                        nonce=read_integer(entry, 'nonce', source),
+                        delegated_address=deserialize_evm_address(entry['address']),
```
LefterisJP, line 545: didn't we say to keep more data? Or no other data is there and needed?

rotkehlchen/serialization/deserialize.py
```diff
@@ -534,6 +535,18 @@ def deserialize_evm_transaction(
...
+                    ))
+                except DeserializationError as e:
+                    log.error(f'Unable to deserialize authorization entry due to {e}')
```
LefterisJP, line 548:
```suggestion
                    log.error(f'Unable to deserialize authorization entry {entry} due to {e}')
```

rotkehlchen/serialization/deserialize.py
```diff
@@ -534,6 +535,18 @@ def deserialize_evm_transaction(
         to_address = deserialize_evm_address(data['to']) if is_empty_to_address else None
         value = read_integer(data, 'value', source)
 
+        authorization_list = None
```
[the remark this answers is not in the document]
prettyirrelevant: this won't work as we need to define an initial value for the variable.

rotkehlchen/chain/evm/transactions.py
prettyirrelevant: same response as above. plus the only call site of `ensure_tx_data_exists()`, we're simply reconstructing the `EvmTransaction` object there.

rotkehlchen/db/evmtx.py
```diff
@@ -67,6 +68,8 @@
 
 
 class DBEvmTx:
+    # Index in the SQL result tuple where authorization fields (nonce, delegated_address) begin  # noqa: E501
+    AUTHORIZATION_DATA_START_INDEX = 13
```
LefterisJP, line 72: Typing here
```suggestion
    AUTHORIZATION_DATA_START_INDEX: ClassVar[int] = 13
```

rotkehlchen/db/evmtx.py
```diff
@@ -216,21 +228,25 @@ def get_evm_transactions(
         """
         query, bindings = filter_.prepare()
         query, bindings = self._form_evm_transaction_dbquery(query, bindings, has_premium)
-        results = cursor.execute(query, bindings)
+        grouped_transactions: dict[int, tuple[Any, ...]] = {}  # Group results by transaction identifier  # noqa: E501
+        for result in cursor.execute(query, bindings):
+            if (tx_identifier := result[12]) not in grouped_transactions:  # Store base transaction data + empty auth list  # noqa: E501
+                grouped_transactions[tx_identifier] = (*result[:self.AUTHORIZATION_DATA_START_INDEX], [])  # noqa: E501
+
+            if (nonce := result[self.AUTHORIZATION_DATA_START_INDEX]) is not None:  # Only add authorizations if exists  # noqa: E501
```
LefterisJP, line 236:
```suggestion
            if (nonce := result[self.AUTHORIZATION_DATA_START_INDEX]) is not None:  # add authorizations if they exist  # noqa: E501
```

rotkehlchen/db/l2withl1feestx.py
```diff
@@ -17,6 +22,7 @@
 
 
 class DBL2WithL1FeesTx(DBEvmTx):
+    AUTHORIZATION_DATA_START_INDEX = 14
```
LefterisJP, line 25: same here. Mark it as classvar.
```suggestion
    AUTHORIZATION_DATA_START_INDEX: ClassVar[int] = 14
```

rotkehlchen/serialization/deserialize.py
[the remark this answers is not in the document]
LefterisJP: That will work in combination with the else I put below. And it makes it less confusing as you end up defining twice here before and after the if.

rotkehlchen/serialization/deserialize.py
```diff
@@ -534,6 +535,18 @@ def deserialize_evm_transaction(
...
+            authorization_list = []
+            for entry in raw_authorization_list:
+                try:
+                    authorization_list.append(EvmTransactionAuthorization(
+                        nonce=read_integer(entry, 'nonce', source),
+                        delegated_address=deserialize_evm_address(entry['address']),
+                    ))
+                except DeserializationError as e:
+                    log.error(f'Unable to deserialize authorization entry due to {e}')
+
```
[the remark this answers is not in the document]
LefterisJP: this else

rotkehlchen/tests/unit/test_evm_tx_decoding.py
```diff
@@ -327,3 +329,42 @@ def test_token_detection_after_decoding(
         assert save_tokens_mock.call_args_list[0].kwargs['address'] == ethereum_accounts[0]
         assert save_tokens_mock.call_args_list[0].kwargs['blockchain'] == SupportedBlockchain.ETHEREUM  # noqa: E501
         assert save_tokens_mock.call_args_list[0].kwargs['tokens'] == [Asset('eip155:1/erc20:0x98C23E9d8f34FEFb1B7BD6a91B7FF122F4e16F5c')]  # noqa: E501
+
+
+@pytest.mark.vcr(filter_query_parameters=['apikey'])
+@pytest.mark.parametrize('ethereum_accounts', [['0xf4ae64c5c4fb632D0e0D77097b957941c399d26e']])
```
LefterisJP, line 335: can you find one with revocation?

rotkehlchen/db/evmtx.py
```diff
@@ -215,21 +227,25 @@ def get_evm_transactions(
         """
         query, bindings = filter_.prepare()
         query, bindings = self._form_evm_transaction_dbquery(query, bindings, has_premium)
-        results = cursor.execute(query, bindings)
+        grouped_transactions: dict[int, tuple[Any, ...]] = {}  # Group results by transaction identifier  # noqa: E501
+        for result in cursor.execute(query, bindings):
+            if (tx_identifier := result[12]) not in grouped_transactions:  # Store base transaction data + empty auth list  # noqa: E501
+                grouped_transactions[tx_identifier] = (*result[:self.AUTHORIZATION_DATA_START_INDEX], [])  # noqa: E501
+
+            if (nonce := result[self.AUTHORIZATION_DATA_START_INDEX]) is not None:  # add authorizations if exist  # noqa: E501
```
LefterisJP, line 235: I asked this in the previous PR

```suggestion
            if (nonce := result[self.AUTHORIZATION_DATA_START_INDEX]) is not None:  # add authorizations if they exist  # noqa: E501
```

LefterisJP (approved):
LGTM

PR #9928, opened 2025-05-12
Fix export of events to match what the user sees

yabirgb:
backend part of https://github.com/orgs/rotki/projects/11?pane=issue&itemId=109656418

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -5580,6 +5580,7 @@ Exporting History Events
 
    .. _history_export_schema_section:
 
+   :reqjson bool match_exact_events: If set to true only the events that match the filter exactly will be exported otherwise the whole group of events that match the filter will be exported.
    :reqjson string directory_path: The directory in which to write the exported CSV file
    :reqjson object otherargs: Check the documentation of the remaining arguments `here <filter-request-args-label_>`_.
 
--- a/rotkehlchen/api/rest.py
+++ b/rotkehlchen/api/rest.py
@@ -3672,6 +3672,7 @@ class RestAPI:
                 filter_query=filter_query,
                 has_premium=has_premium,
                 group_by_event_ids=group_by_event_ids,
+                match_exact_events=True,  # set to True since the frontend requests the event_identifiers manually in their second call to this endpoint. https://github.com/orgs/rotki/projects/11?pane=issue&itemId=110464193  # noqa: E501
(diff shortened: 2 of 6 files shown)


rotkehlchen/api/rest.py
```diff
@@ -3672,6 +3672,7 @@ def get_history_events(
                 filter_query=filter_query,
                 has_premium=has_premium,
                 group_by_event_ids=group_by_event_ids,
+                match_exact_events=True,  # set to True since the frontend requests the event_identifiers manually in their second call to this endpoint  # noqa: E501
```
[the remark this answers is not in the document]
yabirgb: As discussed in the standup how it works with the pagination is tricky and there are some cases where it fails still. I have the work here https://github.com/rotki/rotki/compare/develop...yabirgb:rotki:export-fix-follow?expand=1 and it can be done as a follow up

docs/api.rst
```diff
@@ -5580,6 +5580,7 @@ Exporting History Events
 
    .. _history_export_schema_section:
 
+   :reqjson bool match_exact_events: If set to true only the events that match exactly the filter will be exported otherwise the whole group of the events that match the filter will be exported.
```
LefterisJP, line 5583:
```suggestion
   :reqjson bool match_exact_events: If set to true only the events that match the filter exactly will be exported otherwise the whole group of events that match the filter will be exported.
```

rotkehlchen/api/rest.py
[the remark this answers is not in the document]
LefterisJP: Can you add the note in the comment here too?

rotkehlchen/db/history_events.py
```diff
@@ -77,10 +77,11 @@
 def filter_ignore_asset_query(include_ignored_assets: bool = False) -> str:
     """Create and return the subquery to filter ignored assets. If `include_ignored_assets`
     is true then the filter is returned to include them."""
-    ignored_asset_subquery = "SELECT value FROM multisettings WHERE name='ignored_asset')"
+    subquery = "SELECT 1 FROM multisettings WHERE name = 'ignored_asset' AND value = asset"
```
LefterisJP, line 80: what's this change for?

yabirgb: I measured it while doing the other changes and it was a bit faster but since I measured with other changes too I'll undo this until the follow up PR gets merged

PR #9932, opened 2025-05-12
Store validator indices in extra_data of requests

nicholasyoder:
Sets the affected validator indices in the extra_data of request events in preparation for processing those events to determine validators' historical balances. Also adds new event subtypes in order to more easily query these types of events.


rotkehlchen/accounting/constants.py
```diff
@@ -28,6 +28,10 @@
         HistoryEventSubType.ATTEST: {DEFAULT: EventCategory.ATTEST},
         HistoryEventSubType.MEV_REWARD: {DEFAULT: EventCategory.MEV_REWARD},
         HistoryEventSubType.BLOCK_PRODUCTION: {DEFAULT: EventCategory.CREATE_BLOCK},
+        HistoryEventSubType.CONSOLIDATION_REQUEST: {DEFAULT: EventCategory.INFORMATIONAL},
```
LefterisJP, line 31: Adding subtypes for everything won't work in the long run.

Make one for "CONSOLIDATE".
What is self consolidate? Do you mean turn to accumulating? That would be info + upgrade right?
The info + withdrawal for withdrawal request.

What exactly is the exit request? Is this something happening on the EL?

Then you can use info + subtype + counterparty being CPT_ETH2 to filter

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
```diff
@@ -195,11 +195,15 @@ def _decode_eth2_consolidation_request(self, context: DecoderContext) -> Decodin
         if source_public_key == target_public_key:
             target = self._get_indexes_or_public_keys(public_keys=[target_public_key])[0]
             info_event.notes = f'Request to convert validator {target} into an accumulating validator'  # noqa: E501
+            info_event.event_subtype = HistoryEventSubType.SELF_CONSOLIDATION_REQUEST
+            info_event.extra_data = {'validator': target}
```
LefterisJP, line 199: here in all places you should specify if it's index or if it's public key. And I suppose in all places it should be index?

rotkehlchen/accounting/constants.py
```diff
@@ -28,6 +28,10 @@
...
+        HistoryEventSubType.SELF_CONSOLIDATION_REQUEST: {DEFAULT: EventCategory.INFORMATIONAL},
```
LefterisJP, line 32: Also you can change those and add new `EventCategory` with specific icons to add a bit more detail into the UI.

rotkehlchen/accounting/constants.py
nicholasyoder: > What is self consolidate? Do you mean turn to accumulating? That would be info + upgrade right?

yes. Using `HistoryEventSubType.UPDATE` for this instead now.

> The info + withdrawal for withdrawal request.

there is no withdrawal subtype - I guess you mean REMOVE_ASSET?

> What exactly is the exit request? Is this something happening on the EL?

Its the same as a withdrawal request except is requesting to exit the validator completely rather than just a partial withdraw. But actually it doesn't really need a different subtype, since its event amount will be zero - so will just use `HistoryEventSubType.REMOVE_ASSET` for both the exit and withdrawal request.

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
nicholasyoder: mmm... yeah it needs to be an index here. Actually as it was here it could have been a public key if it was unable to retrieve the index for some reason. Changed it so it will log an error fail to decode if its unable to find the index. Also updated the extra_data keys to specify that its an index.

rotkehlchen/accounting/constants.py
nicholasyoder: Renamed to `HistoryEventSubType.CONSOLIDATE` and for the category used the more generic word `COMBINE` so it can be used for other stuff as well.

rotkehlchen/accounting/constants.py
```diff
@@ -311,6 +312,9 @@
     )}, EventCategory.BURN_NFT: {DEFAULT: EventCategoryDetails(
         label='burn nft',
         icon='lu-flame-kindling',
+    )}, EventCategory.COMBINE: {DEFAULT: EventCategoryDetails(
+        label='Combine',
+        icon='lu-combine',
```
LefterisJP, line 317: Make sure the icon is in the icons list for the frontend. They have a file which defines all icons we use.

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
```diff
@@ -79,16 +79,23 @@ def _query_validator_indexes(self, public_keys: list[Eth2PubKey]) -> list[int]:
 
         return [found_indexes.get(key, UNKNOWN_VALIDATOR_INDEX) for key in public_keys]
 
-    def _get_indexes_or_public_keys(self, public_keys: list[Eth2PubKey]) -> list[int | Eth2PubKey]:
```
LefterisJP, line 82: why remove this? This was used in a specific case. When the validator is depositing there is no index for the first 16-24 hours until it is confirmed by the CL. Only then do we know the index.

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
```diff
@@ -193,13 +200,26 @@ def _decode_eth2_consolidation_request(self, context: DecoderContext) -> Decodin
...
-            info_event.notes = f'Request to convert validator {target} into an accumulating validator'  # noqa: E501
+            if (indexes := self._get_known_validator_indexes(
+                context=context,
+                public_keys=[target_public_key],
+            )) is None:
+                return DEFAULT_DECODING_OUTPUT
+
+            info_event.extra_data = {'validator_index': (validator_index := indexes[0])}
+            info_event.notes = f'Request to convert validator {validator_index} into an accumulating validator'  # noqa: E501
+            info_event.event_subtype = HistoryEventSubType.UPDATE
```
LefterisJP, line 211: So this will get the "Update project" label? It may be a bit misleading. Wonder how it will look like in the UI now.

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
```diff
@@ -218,12 +238,20 @@ def _decode_eth2_withdrawal_request(self, context: DecoderContext) -> DecodingOu
...
         public_key = Eth2PubKey(bytes_to_hexstr(context.tx_log.data[20:68]))
-        validator = self._get_indexes_or_public_keys(public_keys=[public_key])[0]
+        if (indexes := self._get_known_validator_indexes(
+            context=context,
+            public_keys=[public_key],
+        )) is None:
+            return DEFAULT_DECODING_OUTPUT
+
+        info_event.extra_data = {'validator_index': (validator_index := indexes[0])}
+        info_event.event_subtype = HistoryEventSubType.REMOVE_ASSET
```
LefterisJP, line 248: For the INFO/Remove asset where else is it used? The event category is just informational. Perhaps we can get a better icon/label?

rotkehlchen/accounting/constants.py
nicholasyoder: yep its already present: https://github.com/rotki/rotki/blob/develop/frontend/app/src/plugins/rui/index.ts#L324

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
nicholasyoder: mmm... if the first deposit is still being processed there will be no withdrawal or consolidation requests will there? We need the index here to put in the extra_data. If it can't find the index then it can't really decode these request events correctly. `_get_indexes_or_public_keys` was just added for decoding these requests, so its not hurting anything elsewhere to replace it with `_get_known_validator_indexes`.

rotkehlchen/chain/ethereum/modules/eth2/decoder.py
nicholasyoder: Looks like the only other place its used in for queuing Eigenlayer withdrawals: https://github.com/rotki/rotki/blob/develop/rotkehlchen/tests/unit/decoders/test_eigenlayer.py#L420

Changed it to have the `WITHDRAW` category.

